PR #32, opened 2024-11-04
Import

rbuergi:
This is a major refactoring of all the streams, how they interact, how they validate, etc. Any form of initialization was removed and replaced by reactive principles.


rbuergi:
Major: complete reworking of internals

PR #64, opened 2026-01-08
Add Architecture Overview documentation for MeshWeaver

joslat:
Trying out first contribution with a simple draft for an architecture overview.

Also I am curious on why the documentation is inside /samples/documentation/ usually docs are published under /docs - they could also be linked to the meshweaver module I believe.


rbuergi:
Hi @joslat . Thank you so much for the contribution. I am happy to accept the PR. You came in between releases. The samples directory are the samples we created for MW 2. MW 3 is not yet ready. We are migrating to a 100% data based approach, i.e. when Migration to MeshWeaver 3 will be finished, we will delete the Samples folder and we will host an instance where all the documentation can be viewed. The documentation will probably then be hosted there and we won't necessarily even check it in to git. Let's discuss more when we see each other.

PR #84, opened 2026-04-14
organization-layout-areas

sierragolflima:
### Changes to Layout Areas 

**Why**

Todo board views (e.g. /ACME/CustomerOnboarding/TodosByCategory) broke in the portal because the security layer couldn't tell the difference between a user requesting data and a hub requesting data from another hub during initialization. It treated the hub like an unknown user and denied access, causing every child Todo card to fail to load.

 **How**

- Mark hub-to-hub communication as trusted: Added an IsHub flag to AccessContext so the security pipeline can recognize internal hub requests and let them through without applying user-level permission checks. This flag is only set by internal infrastructure code, never by user-facing paths.
- Aligned test configuration with the portal: The existing test was missing several services the portal uses (row-level security, AI, threads, activity tracking, API tokens). Updated TodoBoardThumbnailTest to mirror the portal's full configuration so these kinds of failures get caught in CI.

### DIFF
--- a/memex/Memex.Portal.Shared/OrganizationLayoutAreas.cs
+++ b/memex/Memex.Portal.Shared/OrganizationLayoutAreas.cs
@@ -21,18 +21,18 @@ public static class OrganizationLayoutAreas
     {
         var hubPath = host.Hub.Address.ToString();
 
-        var orgStream = host.Workspace.GetStream<Organization>()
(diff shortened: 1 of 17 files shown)


sierragolflima:
@copilot resolve the merge conflicts in this pull request

sierragolflima:
Superseded by #94 (merged). Root cause was wrong NodeType on sample orgs — simple one-line fix per file.

PR #85, opened 2026-04-14
deployment-test

sierragolflima:
### Enabling meshweaver test deployment via aspire deploy

**Changes** 

Deployment Modes:
- Simplified deployment modes from 6 to 4 by removing local-test and local-prod — hybrid modes that mixed local and cloud infrastructure. The remaining modes (local, test, prod, monolith) have a consistent infrastructure profile.

Project Templates:
- Added a project template (dotnet new meshweaver-memex) that scaffolds a complete portal with pre-configured users, access roles, and sample data — so new portals start working out of the box.

Deployed MeshWeaver/Memex-Test:
- Login not working: Azure Container Apps uses an internal address for the app that differs from the public URL. Microsoft login tried to redirect back to that internal address, which didn't match. Fixed by configuring the app to use the public URL from Azure Container Apps.
- AI chat working: After fresh deployments, users didn't yet have permission to their own chat area in the database. The AI couldn't read or write there. Fixed by granting users access to their own space automatically.
- Azure Foundry models appear (and in the right order) in the AI chat dropdown list.


rosalie1984 (commented):
Deployment nodes: 
- Discussed the implemented changes, reduction of deployment modes from 6 to 4 in line with RB.

Project Templates_
- deployment for specific instance simplified - needed for future client work in proxy environment.

Deployment MeshWeaver/Memex-Test
- Foundry models need to be there
- AI chat is essential - for each user specification
- Login needs to be unified without hickups

Tests run through without failure

Warnings are in connection with "var results = await", this should be checked and commented.

apart from that this is approved.

PR #90, opened 2026-04-15
Acme

sierragolflima:
**Suppress CS4014 warnings on fire-and-forget RegisterCallback calls:** Add _ = discard to unawaited RegisterCallback calls in ThreadExecution and ThreadChatView, matching the established pattern across 13+ existing call sites. The hub's disposal mechanism already handles cleanup of pending callbacks via linked CancellationTokenSource.


rosalie1984 (commented):
As discussed, ACME is expected to improve, and this is achieved with the changes.
The warnings will still be addressed, and with the clean-up, this is approved.

PR #91, opened 2026-04-16
new settings.json file, changes to gitignore and CLAUDE files

sierragolflima:
**Restructured Claude Code configuration for better security and team collaboration.** Moved team-shared rules from .claude/settings.local.json (incorrectly checked into git) to .claude/settings.json (committed), while personal/machine-specific rules remain in the local file (now gitignored).

**Tightened security by narrowing overly broad permissions.** Destructive Azure commands were replaced with read-only patterns, and rm:*, git commit:*, and git push:* now require confirmation. Added a deny list to hard-block irreversible operations like force pushes, hard resets, sudo, and resource group deletions.

**Moved the "never commit/push automatically" rule from CLAUDE.md to the permissions system.** Runtime-enforced rules are deterministic, and this saves tokens on every session. Also cleaned up accidental one-off entries that had accumulated in the allow list.


sierragolflima:
@copilot resolve the merge conflicts in this pull request

rosalie1984 (approved):
discussed with Samuel and it makes sense to not have everything in CLAUDE.md and instead explicitly state hard permission and limits in the setting file - the 3 remaining conflicts can be resolved with git copilot - there are no test fails

PR #96, opened 2026-04-25
Minor fixes to UI: focus steal & overlapping elements

sierragolflima:
### Two fixes for the MeshWeaver UI:

**Main:**
Pressing / or ? anywhere on the page used to jump the cursor into the search bar when you were typing in the AI chat. Now the app checks whether you're already typing somewhere first. If you are, the shortcut is ignored and your cursor stays put. (The problem with the ? seems to concern only US keyboards where / and ? share the same physical key — ? is just Shift+/)

**Other:**
The search dropdown was hidden behind the AI chat's title bar. Raised the header's priority so the dropdown always appears on top, and lowered the chat title bar's priority.


rosalie1984 (commented):
These are simple changes nothing that goes into the deep code base and only fixes the appearance of the of the UI mechanics.

Once check go successfully through this is approved.

PR #107, opened 2026-06-29
ci(base-image-acr): null-safe, assertive multi-arch verify step

rbuergi:
Follow-up to #106. The `base-image-acr.yml` build+push succeeded (the multi-arch base is in ACR — `latest` → OCI image index with linux/amd64 + linux/arm64 children), but the **verify** step turned the job red: `--query "[?contains(tags, 'latest')]"` throws `In function contains(), invalid type for value: None` because the buildx provenance/SBOM attestation child manifests have `tags: null`.

Replace it with a real assertion: read the `:latest` image index's child platform architectures via `az acr manifest show` + `jq`, and fail loudly unless **both** `amd64` and `arm64` are present. Now a successful multi-arch push reports green, and a future single-arch flatten fails the job.

### DIFF
--- a/.github/workflows/base-image-acr.yml
+++ b/.github/workflows/base-image-acr.yml
@@ -73,8 +73,15 @@ jobs:
             --push deploy/base-images/portal-ai
 
       - name: Verify multi-arch manifest list in ACR
+        # Assert :latest is an image INDEX whose children include BOTH linux/amd64 AND linux/arm64,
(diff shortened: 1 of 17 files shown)


rbuergi:
Closing: branch conflicts with main and is contaminated with unrelated MAUI/access/migration changes. The small base-acr verify-query fix will be redone cleanly.

PR #124, opened 2026-06-30
ci+fix: CodeEditRecompile flake — 2-vCPU repro harness, watcher refinement, stuck-state diagnostic

rbuergi:
Driving the `CodeEditRecompileTest.PressingCompileButton_SetsRequestedReleaseAt_AndProducesNewRelease` CI flake (the `WaitForLatestRelease` 50s timeout). Three deliverables — plus an honest scoping of the residual.

## 1. Reusable 2-vCPU repro harness — `.github/workflows/flake-repro.yml`
`workflow_dispatch` job that loops one suspect test on the real `ubuntu-latest` 2-vCPU runner until it fails (mirrors `dotnet-test.yml`'s build), uploads the failing trx + `MonolithMeshTestBase` traces + blame hang-dump. **Reliably reproduces the flake** (caught at iter 4 / 19 / 80 across runs) — the first time this CI-only flake is reproducible on demand. Inputs target any project/test.

🚨 **Heisenbug warning baked into the file:** cranking the `CompileWatcher` log categories to Debug *masks* the flake (runs went green at 40/80 iters). Keep production timing; rely on the timing-neutral stuck-state diagnostic below.

## 2. Stuck-state diagnostic — `CodeEditRecompileTest.WaitForLatestRelease`
On the 50s timeout, dumps the node's `Status` / `LatestReleasePath` / `RequestedReleaseAt` / `LastReleaseRequestHandledAt`. Timing-neutral (fires only on timeout). Pinned the stuck state: **`Status=Ok, RequestedReleaseAt set, LastReleaseRequestHandledAt=null, Latest==knownRelease`** — read via the test's cross-hub mirror.

## 3. Watcher correctness refinement — `InstallReleaseRequestWatcher`
Advance `dispatchHighWater` (+ un-park + log) **only on the Update commit path** (when actually flipping `Pending` + stamping `Handled`), never eagerly before the Update. The eager advance was a real lost-trigger bug: if status flapped to `Pending`/`Compiling` between the `Where` and the lambda, the lambda bailed *without* handling, yet high-water was already advanced past the unhandled trigger → the settled re-emission failed `req > dispatchHighWater` → trigger lost. **Correct + low-risk**, but with one post-fix repro data point it is **not proven to materially move the flake** — landing it as a correctness improvement, not the root fix.

## Residual (honest scope)
The dominant root is a **known cross-hub-mirror DELIVERY / missed-emission race** (the test's mirror freezes at the post-click state and never receives the owner's subsequent emissions) — documented in depth in our prior RCA (`compile-flake-syncstream-staleframe-rca`): a same-node content-clock guard is *neutral*, and it has resisted ~5 prior deep attempts (~2.5M tokens) as a core sync-stream issue. Per that RCA, this is **not release-pressure grindable** and needs a deliberate core-sync delivery-handshake fix — not a guessed change. This PR makes it **reproducible** and adds the diagnostic so that work has a foothold.

### DIFF
--- a/.github/workflows/flake-repro.yml
+++ b/.github/workflows/flake-repro.yml
@@ -99,28 +99,12 @@ jobs:
             -c Release --no-build --no-restore -o dist/packages --nologo
           dotnet pack src/MeshWeaver.BusinessRules.Generator/MeshWeaver.BusinessRules.Generator.csproj \
             -c Release --no-build --no-restore -o dist/packages --nologo
-      # 🚨 Crank the compile/release-watcher log categories in the RUNTIME bin appsettings
-      # ONLY (never the committed src appsettings — that's production cost contract; AGENTS.md
-      # sanctions editing the bin/ copy for a debug run). The committed test appsettings pins
-      # these to Warning, which suppressed the "[ReleaseRequestWatcher] handling…" Information
-      # line that tells us whether the watcher even fired. Debug here surfaces the full
-      # compile→release trace into the failing test's trx Output + the test-logs file.
-      - name: Crank watcher log levels (bin runtime config only)
-        run: |
-          set -euo pipefail
-          app="$REPRO_PROJECT/bin/Release/net10.0/appsettings.json"
-          if [ -f "$app" ]; then
-            tmp="$(mktemp)"
-            jq '.Logging.LogLevel.Default = "Information"
-                | .Logging.LogLevel["MeshWeaver.Graph.CompileWatcher"] = "Debug"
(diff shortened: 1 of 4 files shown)


rbuergi:
Closing: the flake this chased was root-caused and eliminated by #173 (single compile driver + complete terminal write). The watcher refinement here predates that rewrite and is unproven against it. Residuals (eager high-water advance, the standalone flake-repro harness, the timeout diagnostic) are tracked in #185 — the branch stays for cherry-picking.

PR #127, opened 2026-06-30
feat(search): explore the content vector index — rows, highlighted excerpts, block reader, original-doc viewer

rbuergi:
## What & why

The mesh vector-indexes content (uploaded PDFs/DOCX → chunks → embeddings → `Document` nodes), but it was barely explorable: global search showed a card grid of nodes with no descriptions, no content excerpts, and no highlight, and chunk search lived only inside the Space → Content Indexing settings tab. This makes the indexed content **first-class in global search and explorable end to end**.

**Flow:** search a phrase → one row per result (title + description) with an *"In documents"* section of content matches (excerpt + highlighted passage) → click a content hit → the **block reader** (one content-index block at a time, prev/next, highlighted) → **Open original** → the source **PDF/DOCX rendered inline with the passage highlighted**.

## Changes

- **One result per row** — new `MeshSearchRenderMode.List`; `MeshSearchView` renders rows (icon · title · description) reusing the existing click / keyboard / delete plumbing. Empty descriptions show *"Ask the agent to create a description."* The `/search` page defaults to List.
- **Content matches merged into search** — `ContentSearchResultsView` runs the same `ContentChunkSearch` engine the agents' `search_chunks` tool uses and renders an *"In documents"* section: file · block, the excerpt with the **matched passage highlighted**, deep-linking into the block reader.
- **`Document` node views** — `Blocks` (`/{doc}/Blocks?index=N&q=…`): one block at a time, highlighted, with Prev/Next in place + "Open original". `Source` (`/{doc}/Source?…`): the original file rendered inline with the passage highlighted.
- **Reusable controls** — `HighlightControl ↔ HighlightView` (text + match terms → `<mark>` via a `RenderFragment`, **never an HTML string**) and `DocumentSourceControl ↔ DocumentSourceView` (PDF.js / mammoth viewer via the sanctioned `OnAfterRenderAsync` interop, graceful download+passage fallback).
- **Vendored** `pdfjs-dist` + `mammoth` under `wwwroot/lib` (offline-safe, not CDN; `libman.json`), marked `linguist-generated`/`-diff` in `.gitattributes` so they collapse in review.

## Key decisions

- **Render-time highlighting, no reindex.** The chunk text is verbatim from the extracted document text, so the query terms are located at render time — in the excerpt, the block, and the rendered original. No stored offsets, no schema change.
- **Content search is per-partition.** One schema per partition means cross-partition content search isn't a single query, so the *"In documents"* section scopes to a namespace (the `?ns=` param, a `namespace:` token, or the current namespace). This is deliberate, not a limitation of this PR.
- **No hand-rolled HTML for structured UI** — new surfaces use framework controls + the typed `HighlightControl`.

## Testing

- `HighlightControlTest` (term extraction + segment marking, 8 cases) and `ContentSearchHitContractTest` (the content-row data contract: `DocumentPath` deep-link + highlightable snippet, 2 cases) — green.
- Full solution build clean for all touched projects.
- ⚠️ **Needs a manual runtime check I couldn't do headless:** the PDF.js/DOCX viewer happy path (render + in-document highlight) against a real upload. The C#, interop pattern, and graceful fallback are verified; the libraries are vendored.

## Notes for reviewers

- Vendored libs are committed for offline/prod builds. If you'd rather not commit them, swap to `libman restore` in CI/image build and gitignore `wwwroot/lib` — `libman.json` is already the source of truth.
- The existing "Explore index" settings tab still uses hand-rolled `Controls.Html` rows (pre-existing); migrating it onto `HighlightControl` + the Blocks link is a clean follow-up, intentionally out of scope here.


rbuergi:
Thanks @Copilot — addressed all four findings in 0c7514676:

1. **List mode hid the view-options bar** → `MeshSearchRenderMode.List` is now selector-eligible; the global search page keeps its group-by / display bar and the "None" choice preserves the row layout.
2. **Case-sensitive `namespace:`** → the token is now matched with `RegexOptions.IgnoreCase`.
3. **Raw `ex.Message` shown to users** → the block-load exception is logged; a generic "Could not load this block." is rendered.
4. **Unencoded `/static` URL** → each path segment is URL-encoded (slashes preserved) so spaces / # / ? / unicode in a file name are safe.

rbuergi:
### CI status — green except rotating known flakes

The only real test failure was **mine** and is fixed: `MauiControlCoverageTest` needed the two new controls classified (227f43b69) → shard 0 now passes.

The remaining failures are **pre-existing CI flakes in subsystems this PR does not touch**, and they rotate across shards between runs:

| Run | Failing shard | Test | Status |
|-----|--------------|------|--------|
| 1 | shard 3 | `SearchQueryTests.TextSearch_CaseInsensitive_MatchesRegardlessOfCase` (MeshWeaver.Query) | passes locally on this branch; did **not** recur in run 2 |
| 2 | shard 1 | `NodeTypeReleaseTest.CompilationPending_CreatesReleaseMeshNode_WithNotes` (Monolith) | a **known** compile-race flake (named in CLAUDE.md's flake list) |

Build ✅, shards 0/2/3 ✅, Copilot review ✅ (all 4 addressed). I've re-run the failed jobs only. If the NodeTypeRelease flake recurs, this PR is still ready to merge on a green run — none of these failures originate here.

PR #134, opened 2026-06-30
fix(sync): stop heart-beating a dead owner forever — cures the _Activity/import-* NotFound storm

rbuergi:
## The bug you keep hitting: `{Partition}/_Activity/import-{fingerprint}` heart-beaten forever

A sync subscription whose owner **dies after a successful subscribe** kept heart-beating the gone owner **every interval forever**. `JsonSynchronizationStream` posted `HeartBeatEvent` **fire-and-forget**, with the change feed as the **sole** recycled-grain detector — which misses the case where the owner dies **without** a node Created/Deleted change-feed event.

That is exactly a one-shot import-activity lock:
1. The importer writes `{Partition}/_Activity/import-{fingerprint}` through a cached `GetMeshNodeStream(activityPath).Update` handle → opens a heart-beated sync stream.
2. The import finishes; the **dedicated off-router import hub is disposed**.
3. The node persists as history, so there is **no change-feed pulse** to drive resubscribe.
4. The cached handle's heartbeat keeps posting to the gone address → **`[ROUTE] NotFound` every interval, per partition, for the life of the silo** — the storm that pegs CPU and trips `/healthz` (local e2e portal → pod killed).

## Fix (one place: the heartbeat)

The heartbeat now **observes its delivery** instead of fire-and-forget:
- A **healthy** owner has no ack → the observe simply times out → ignored (the normal case).
- A **terminal NotFound** means the owner address is **gone** — a recycled/deactivated grain *reactivates* on the heartbeat post and acks, so **only a permanently-gone owner NotFounds** → tear the keep-alive down.

Keep-alive is preserved (the heartbeat is still posted + processed). We only **add** teardown on a terminal NotFound. Recycled-grain recovery is unchanged — it's change-feed-driven, and a reactivatable owner never NotFounds.

## Test — the dedicated repro at this exact shape

`HeartbeatStopsWhenOwnerDiesAfterSubscribeTest`: the owner serves a real type so the **subscribe succeeds and the heartbeat runs**, then a flag flips it **dead** so every further heartbeat NotFounds **with no change-feed event** (the `_Activity/import-*` shape). After death the heartbeat count must **freeze**.
- **Before the fix:** RED — `Expected 10 to be ≤ 1` (climbs ~one-per-interval).
- **After:** green.

Complements `NonExistentOwnerNoHeartbeatStormTest` (owner *never* existed) — this is the owner-**died-after-subscribe** half. Regression-checked green: `ChangeFeedResubscribeCoalesceTest` and `ResubscribeOnOwnerDisposeTest` (recycled-grain recovery).

## Scope

This is the **storm cure** (the CPU peg / liveness-trip). The related cleanups discussed — bundling the AI static-repo sources (so `Skill` can't be silently un-imported) and the partition→`_Access`(public-read)→write-under-System sequencing — are a separate follow-up PR.

### DIFF
--- a/src/MeshWeaver.Data/Serialization/JsonSynchronizationStream.cs
+++ b/src/MeshWeaver.Data/Serialization/JsonSynchronizationStream.cs
@@ -459,14 +459,20 @@ public static class JsonSynchronizationStream
                     if (hbDelivery is not null)
                         h.Observe(hbDelivery)
                             .Take(1)
-                            .Timeout(heartbeatInterval)
+                            // A healthy owner sends no heartbeat ack, so the observe never emits — switch
+                            // to Empty (COMPLETES) instead of the throwing Timeout overload: the normal
+                            // case must not route through the error path every interval (exceptions-as-
+                            // control-flow is a real CPU/alloc cost given how many heartbeats run). A
+                            // terminal NotFound still arrives FAST as an OnError before this fires.
+                            .Timeout(heartbeatInterval, Observable.Empty<IMessageDelivery>())
                             .Subscribe(
-                                _ => { },   // a healthy owner has no heartbeat ack to send
+                                _ => { },   // no ack to consume
                                 ex =>
                                 {
+                                    // Owner address gone (a recycled grain reactivates on the post and
+                                    // acks — only a permanently-gone owner NotFounds): STOP heart-beating.
                                     if (ex is DeliveryFailureException { Failure.ErrorType: ErrorType.NotFound })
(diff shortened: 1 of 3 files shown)


rbuergi:
Good catch @Copilot — switched to `Timeout(heartbeatInterval, Observable.Empty<IMessageDelivery>())` so the healthy no-ack case completes instead of throwing a TimeoutException every interval. The terminal NotFound still arrives fast as an OnError before the timeout, so teardown is preserved. (3b… pushed.)

PR #146, opened 2026-07-01
feat(speech): centralized Swiss-German Whisper container + MeshWeaver.Speech

rbuergi:
## What

Inverts the **on-device** Whisper Swiss-German model (`memex/Memex.Client`, downloaded + run per client) into **one centralized container** the portal configures and every client calls. This is the foundational layer; the mic UI + client wiring follow.

Why centralize: a browser can't run on-device Whisper at all, replicating a 547 MB model per client is wasteful, and public cloud STT (Azure/OpenAI) doesn't do real Swiss-German *dialect* — our fine-tune (trained partly on Bernese) does, so we must host **our** model.

```
 Blazor chat / RN / MAUI ── record WAV ──▶ (portal endpoint, next) ──▶ ISpeechTranscriber (HTTP IIoPool)
                                                          │ POST /inference
                              configure ────────────────▶ whisper.cpp CONTAINER + ggml-swiss-german-turbo-q5_0
```

## In this PR

- **`deploy/whisper/`** — Dockerfile (builds whisper.cpp's `whisper-server`) + compose + README. Serves the same model the MAUI client ran on-device, behind `POST /inference`. Model is **mounted (swappable), not baked**.
- **`src/MeshWeaver.Speech`** — the centralized mesh-side client:
  - `SpeechConfiguration` (Endpoint, Language, Enabled), read **live** via `IOptionsMonitor` so the portal can repoint the container at runtime.
  - `ISpeechTranscriber` / `WhisperContainerTranscriber` — multipart POST to `/inference`, JSON transcript parse. The HTTP call runs on the **HTTP `IIoPool`** (never a hub/circuit thread); the surface is a **cold `IObservable`** — no `async`/`await` leak.
- **4 unit tests** drive the transcriber against a real in-process server mimicking whisper.cpp `/inference`: transcribe + trim, language forward + per-call override, unconfigured → clear error, server 500 → observable error. ✅
- **Docs** — `Architecture/CentralizedSpeech` (design + diagram + status), cross-linked to the existing `OnDeviceVoice`.

## Not in this PR (next)

The portal `POST /api/speech/transcribe` endpoint, the `ThreadChatView` mic button (browser `MediaRecorder`), and the RN/MAUI record paths — then an end-to-end run against a live container.

## Verification

`MeshWeaver.Speech` + tests build under Release `-warnaserror`; 4/4 tests pass. The **container isn't runtime-verified** here (no Docker in the build sandbox) — the client that speaks its contract is.


rbuergi:
Holding this PR: (1) main's RN app shell evolved substantially (#145/#149 — navigation, theme provider, client screens), so the speech UI (ChatComposer + push-to-talk) needs INTEGRATING into the new App.tsx shell, not a mechanical merge; (2) the centralized path still needs the portal `POST /api/speech/transcribe` endpoint (currently only the container contract exists). Both are follow-ups before this is mergeable. The client-side thread ops it introduced already landed via #149.

PR #156, opened 2026-07-01
fix(react): pin the live layout-area protocol (verified against Memex.LocalMesh)

rbuergi:
## What

`GrpcAreaSource` (the renderer's live data plane) had `WIRE:`-guessed subscribe/change handling. I **verified the real protocol against a running mesh** — the new `Memex.LocalMesh` (monolith runtime + SQLite + gRPC, PR #154) — by subscribing to a Doc area over gRPC-web and capturing exactly what the mesh sends:

- **Reference is polymorphic** — it must carry its `$type`. A `$type`-less reference elicited **no subscription**; `{ $type: "LayoutAreaReference", area }` streams the area back. (Field casing is server-insensitive.)
- **`DataChangedEvent.Change` wraps the EntityStore** as `{ Content: "<store json>" }`; the EntityStore *is* `{ areas, data }` — the exact tree the renderer consumes. The old code treated the wrapper as the tree, so nothing rendered.
- **`ChangeType` `0` = Full** (replace); anything else = RFC 7396 merge-patch.

Captured live (Doc/Architecture, area "Overview"): first a `"Building layout..."` progress store, then a Full store whose `Overview` area is a `StackControl` of `NamedAreaControl` children — the real doc control tree the renderer already knows how to render.

## Verification

react suite green (11 tests), typecheck clean. This turns the client's `WIRE:` guesses into a protocol **verified against a live mesh**, and unblocks the RN/web client rendering real layout areas.

## Depends on / relates to
- **#154** — `Memex.LocalMesh` (the local mesh this was verified against).

### DIFF
--- a/clients/react/src/live/grpcSource.ts
+++ b/clients/react/src/live/grpcSource.ts
@@ -58,7 +58,11 @@ export class GrpcAreaSource implements AreaSource {
   /** Open the area subscription and fold every change into the tree until the stream ends. */
   async start(): Promise<void> {
     for await (const delivery of this.connection.watch(this.address, this.streamId, "SubscribeRequest", {
-      reference: this.reference,
+      // The sync-stream Reference is a polymorphic WorkspaceReference — it MUST carry its $type
+      // discriminator or the mesh can't deserialize it. Verified against Memex.LocalMesh: a $type-less
+      // reference elicits no subscription; with "LayoutAreaReference" the area streams back. Field casing
+      // is server-insensitive, so the lowercase { area, id, layout } passes through.
+      reference: { $type: "LayoutAreaReference", ...this.reference },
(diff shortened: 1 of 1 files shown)


rbuergi:
Superseded by main's grpcSource.ts (aefce93a2, authored after this PR): it pins the same protocol against the server source with strictly broader handling — RFC 6902 patch ops incl. move/copy with wire-encoded pointer decoding, InstanceCollection key normalization, NoUpdate skip, and an RFC 7396 fallback. The {Content: ...} unwrap here matched a transport path that predated the server-side RawJson inlining (RawJsonConverter.WriteRawValue); merging would regress the newer implementation. The commit remains part of #154/#157's history — their merges with main should take main's version of this file.

PR #165, opened 2026-07-01
fix(build): un-red main — remove unused field newly compiled via slnx addition

rbuergi:
Main went red when #149 put the E2E test project into the solution and -warnaserror caught a latent CS0414. CD is skipping (self-update stalled) until this lands.


rbuergi:
Redundant since #163 merged — it carries the same fix (8fbda0751 drops the unused field) plus the serialization-test alignment. Main's run for the #163 merge is in progress.

PR #169, opened 2026-07-01
fix(hosting): idle-release for read streams — stop the sync-stream heartbeat leak

rbuergi:
Root fix for the read-stream accumulation (task: every path ever read heartbeats forever; ~1,650 live streams measured). Refcounted idle release with detach-then-mark; eviction-only sweep (never resubscribes); live subscribers pin their entry; transparent re-open after release. 4 new deterministic tests; Persistence 261/261, Query 342/342, concurrency suites green. Note: DataTest.Delete flaked once in Data.Test (272/273) — evidence says pre-existing ack-vs-commit race (diff unreachable in that process; pristine baseline 273/273); flagged for separate triage rather than re-run-to-green.


rbuergi:
✅ DevAuth regression root-caused + fixed: the storm-breaker re-probe was disposing the upstream every 100ms poll on a freshly-provisioned node → racing UnsubscribeRequest wedged the owner → subscribe/unsubscribe storm → 25s hang. Now re-probes once per window and disposes only the HydrationSub (idle sweep still reaps live upstreams — leak fix intact). DevAuth 109/109, idle-release 8/8, Persistence 276/276, Release -warnaserror clean.

PR #172, opened 2026-07-02
fix(routing): legacy /User/{id} home resolves to the user root partition (no more 'No renderer for area Roland')

rbuergi:
Root-causes the user-home 'Area not found' + a contributor to the multi-round composer-vanish: post-v10 the user node is a root partition at {id}, but /User/{id} matched only the legacy 'User' catalog node → hub=User/area={id} → no renderer. Rewrite to the canonical {id} root at the single resolution chokepoint (nav + message routing), guarded + loop-safe. 5 new resolution/render tests; 101 regression tests green; solution Release -warnaserror clean.

### DIFF
--- a/src/MeshWeaver.Blazor/Pages/AreaPage.razor.cs
+++ b/src/MeshWeaver.Blazor/Pages/AreaPage.razor.cs
@@ -46,8 +46,10 @@ public partial class AreaPage : ComponentBase
     protected override void OnParametersSet()
     {
         // Reactive — Subscribe, never await on PathResolver chain (deadlock surface;
-        // see Doc/Architecture/AsynchronousCalls.md).
-        PathResolver.ResolvePath(Path ?? "").Subscribe(resolution =>
+        // see Doc/Architecture/AsynchronousCalls.md). This is GUI URL→area navigation, so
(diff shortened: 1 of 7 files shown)


rbuergi:
⚠️ Real regression found in triage (not a flake): `UserOnboardingServiceTests.CreateUser_WritesPartitionRootOnly_NoUserMirror` passes on main (745ms) but fails on this branch (10s timeout). Cause: the legacy-`User/` rewrite lives in `PathResolutionService.ResolveOnce`, which is shared by NavigationService AND RoutingServiceBase — so a genuine node read/route of `User/{id}` now rewrites to the user root instead of returning null, breaking the no-mirror invariant. Fix: scope the rewrite to the NAVIGATION path only (GUI URL→area), leaving message routing / node reads untouched. Holding merge until scoped.

rbuergi:
✅ Regression fixed: rewrite scoped to navigation-only (ResolveNavigationPath); ResolvePath (routing/reads) stays literal. CreateUser_WritesPartitionRootOnly_NoUserMirror passes again + new guard test pins the split. 6 home + 41 nav + 24 UserActivity tests green; Release -warnaserror clean.

PR #184, opened 2026-07-02
feat(spaces): delete space drops whole partition; sync direction + multiple sync sources; Partitions admin overview

### LINKED ISSUE #183: Spaces: delete menu must remove the entire partition; sync direction + multiple sync sources; Partitions admin overview
Tracking ticket for the space-administration feature set:

1. **Delete a space = remove the whole partition.** The Delete menu on a Space must not just delete mesh nodes — the partition's backing store (Postgres schema incl. all satellite tables), the `Admin/Partition/{id}` definition, and the provisioning caches must go with it, so the same id can be re-created cleanly.
2. **Partitions overview page.** A global-admin page listing ALL partitions with the main node's properties (namespace empty, id = partition), sync status, and sync sources.
3. **Sync settings manageable from the overview** — edit each source's settings in place, add/remove sync sources.
4. **Sync direction** — each sync source configurable as unidirectional (export-only mesh → repo, or import-only repo → mesh) or bidirectional.

rbuergi:
Closes #183.

### 1. Deleting a Space removes the ENTIRE partition
- `IPartitionStorageProvider.DeletePartition(namespace)` — the inverse of `EnsurePartitionProvisioned`. Postgres: `DROP SCHEMA … CASCADE` through the per-adapter io pool (DDL-race retry), evicting `_schemasInitialized` / `_provisioned` / `_registeredPartitions` so a same-id re-create provisions from scratch. Other providers keep the no-op default (single path-keyed stores — the recursive node delete already removed their rows).
- New `INodePostDeletionHandler` seam (mirror of `INodePostCreationHandler`), invoked in `HandleDeleteNodeRequest` for the ROOT of the operation after the subtree commit; failures surface as Warnings on the deletion activity (the nodes are already gone — the response stays Ok).
- `PartitionDropPostDeletionHandler` (registered with the Space type): drops the backing store on every provider (sequential, like provisioning), then deletes the `Admin/Partition/{id}` definition under system impersonation (store-drop first — a failed drop keeps the definition visible for retry instead of leaving an invisible orphan schema).

### 2. Sync direction — unidirectional or bidirectional
- `SyncDirection` (`Bidirectional` default / `ExportOnly` / `ImportOnly`) on `GitHubSyncConfig`, enforced in `GitHubSyncService.SyncToGitHub` (rejects import-only) and `ReimportAtCommit` (rejects export-only; `UpdateToLatest` inherits).
- The settings tab renders direction-aware sync buttons (an import-only source offers no commit, an export-only source no checkout).

### 3. Multiple sync sources per Space
- Primary stays at `{space}/_GitSync`; additional sources are child config nodes `{space}/_GitSync/{sourceId}`, each with its own repo/branch/subdirectory/direction.
- `AddSyncSource` / `WatchConfigNodes` / `RemoveSyncSource`; optional `sourceId` threaded through the service, the activity extensions and the settings tab ("Additional sync sources" section: per-source data-bound editor + direction-aware actions + remove + add-by-name).

### 4. Partitions admin overview (Global Settings → Administration → Partitions)
- Rewrote `PartitionSyncAdminLayoutArea`: every partition root (top-level `Space` query + `IStaticRepoSource` registrations) in a grid with main-node properties, sync status and a sync-sources summary; per-partition detail panel with the Synced/Not-synced toggle, sync-source editors bound directly to each config node stream, add/remove source, Open and Delete navigation.
- New `IPartitionSyncSourceProvider` DI seam so GitSync contributes its sources to the Graph-hosted page without a circular project reference; the section degrades gracefully when GitSync isn't wired.

## Tests
- `SpaceDeletionPartitionDropTests` (PG, e2e): create Space → schema exists → delete → schema + `Admin/Partition/{id}` gone → re-create same id → fresh schema serves writes.
- `PartitionLifecycleTests.DeletePartition_DropsSchema_AndReprovisionRecreatesIt` (+ idempotent double-drop; promise-cache eviction).
- `SyncDirectionAndSourcesTest`: import-only rejects export, export-only rejects import (after a successful export), add/sync/remove of an additional source with per-source last-sync state.
- `GlobalAdminSpaceCrudTest.DeletingSpace_RemovesPartitionDefinition` (monolith).
- Updated `MeshNodeContentEditorTest` (direction enum field) and `GitHubSyncSettingsTabTest` (reactive sync-button row).

Full solution builds clean with CI flags (`-c Release -p:CIRun=true -warnaserror`). Docs updated: `Spaces.md` ("Deleting a Space"), `GitHubSync.md` (direction + multiple sources); `DocumentationLinkIntegrityTest` green.

## Notes
- The overview page itself has no dedicated GUI render test (its predecessor had none either); the seams beneath it are covered.

### DIFF
--- a/src/MeshWeaver.Documentation/Data/Architecture/GitHubSync.md
+++ b/src/MeshWeaver.Documentation/Data/Architecture/GitHubSync.md
@@ -68,11 +68,38 @@ Under **Repository**:
 |---|---|
 | **Repository URL** | `https://github.com/owner/repo`. |
 | **Branch** | The branch to commit to (default `main`). |
+| **Sync direction** | `Bidirectional` (default), `ExportOnly`, or `ImportOnly` — see below. |
 | **Create the branch if it doesn't exist** | When on, a missing branch is created as a fresh snapshot commit. |
 | **Create the repository (private) if it doesn't exist** | When on, a missing repo is created **private** under the owner/org. |
 | **Subdirectory** (optional) | Mirror the Space into this folder of the repo. Files outside it are left untouched. Empty = repository root. |
 
 Click **Save repository settings**.
+
+### Sync direction — unidirectional or bidirectional
+
+Each source syncs in a configurable direction, enforced by `GitHubSyncService` on every
+operation (the GUI additionally hides what would be rejected):
+
+| Direction | Commit ("Sync now") | Checkout / re-import |
+|---|---|---|
+| **Bidirectional** (default) | ✓ | ✓ |
+| **ExportOnly** — mesh → repo | ✓ | ✗ rejected — the repo can never overwrite the mesh |
+| **ImportOnly** — repo → mesh | ✗ rejected — the mesh can never overwrite the repo | ✓ |
+
+Use `ImportOnly` for a Space that mirrors an upstream repository you don't own, and
+`ExportOnly` for a backup/publishing target that must never feed edits back.
+
+### Multiple sync sources
+
+A Space can sync with **more than one repository**. The Repository section above edits
+the **primary** source (`{space}/_GitSync`); every additional source is its own config
+node at `{space}/_GitSync/{sourceId}` with its own repository, branch, subdirectory and
+direction. Manage them in the **Additional sync sources** section of the same settings
+tab (or, platform admins, on **Global Settings → Administration → Partitions**): add a
+source by name, edit its settings through the same data-bound editor, sync it with its
+own direction-aware buttons, and remove it when no longer needed. Programmatically:
+`GitHubSyncService.AddSyncSource / WatchConfigNodes / RemoveSyncSource`, and every sync
+operation takes an optional `sourceId` (null = the primary).
 
 ---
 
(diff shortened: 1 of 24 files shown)


rbuergi:
CI triage of the two shard-1 failures on this PR:

- Run 1 (baeac519): `CodeEditRecompileTest.NodeType_RequestedReleasePath_PinsToHistoricalRelease` — 50 s timeout. Compile-driver family (the #185 residual population); this PR does not touch that subsystem.
- Run 2 (a4453251): `Acme.TodoDataChangeWorkflowTest.QueryWithStateActive_ExcludesDeletedItems` — 1 min timeout. Exercises `DataChangeRequest` workspace workflows — not the `DeleteNodeRequest` pipeline this PR modifies. Passes locally on this branch in 2 s.
- `main` itself was red this morning (run 28575894420, commit 0cac4e76) on a third test of the same shape (`CompileActivityNoPhantomPathTest`, 40 s timeout).

Different test each run, different suites, all load-timing timeouts, none on this diff's code paths → pre-existing CI timing flakiness, not a regression from this change. Re-running the failed jobs once with this diagnosis on record; if a third distinct failure appears, that's a systematic signal and gets investigated before any merge.

PR #240, opened 2026-07-04
fix(content): SaveFileAsync read-your-writes — kill the ContentCollections 30s-timeout flake

rbuergi:
## The recurring CI flake, root-caused and killed deterministically

`NodeHubContentCollectionTest.CollectionNamedArea_RendersBrowserForFolder_AndContentForFile` fails intermittently on shard 2 with `ObservableAssertionException : Expected the observable to emit a value matching the predicate within 30s, but it did not` — it's been rerun-once past the merge gate repeatedly (including on #224 and #239 in this session).

### Root cause (an "event miss")
A content collection is **cached** and its markdown files scanned **once** at init. `ContentCollection.SaveFileAsync` wrote the file to disk but **did not update the collection's markdown stream** — it relied entirely on the **`FileSystemWatcher`** to notice the new file and merge it (`UpdateArticle`). The file-content render (`GetMarkdown` → reduce that stream) therefore depends on a watcher event that is **asynchronous and lossy**: `FileSystemWatcher` coalesces events and overflows its OS buffer under load. When the event is dropped or delayed, the just-saved file never enters the stream → the render emits null → the 30s assertion times out.

This also bit **production**: an upload immediately followed by a view could show "not found" until (or unless) a watcher event arrived.

### Fix
`SaveFileAsync` now merges the just-written markdown file into the stream **itself** (read-your-writes), through the same path the watcher uses. The watcher remains the mechanism for **out-of-band** (external) changes; a save the process performed is reflected **deterministically**.

### Proven deterministically — both directions
The new `SaveFileReadYourWritesTest` runs the same save→render with the watcher turned **off** (via a new, genuinely-useful `DisableFileWatcher` config — the watcher is unreliable on network mounts / under load), so the file can reach the stream **only** through `SaveFileAsync`'s merge:
- **Without the fix** → times out at **30s** (the exact CI failure reproduced locally).
- **With the fix** → passes in **~370ms**.

All 16 tests in the project pass; Release `-warnaserror -p:CIRun=true` clean.

### DIFF
--- a/src/MeshWeaver.ContentCollections/ContentCollection.cs
+++ b/src/MeshWeaver.ContentCollections/ContentCollection.cs
@@ -141,20 +141,17 @@ public class ContentCollection : IDisposable
     {
         await provider.SaveFileAsync(path, fileName, openReadStream);
 
-        // 🚨 Read-after-write for the collection's OWN writes must NOT depend on the file-system
-        // watcher. The watcher (AttachMonitor) is the only thing that feeds a post-init write into
-        // markdownStream — but on Linux inotify DROPS the event for a file written into a
-        // just-created subdirectory (the recursive watch on the new dir isn't registered before the
-        // write), so the article never lands and a content render stays "not found" until the
-        // collection re-initializes. That is the CI-only CollectionNamedArea flake AND the prod
-        // "upload a file then open it → shows nothing" bug this test class was written for. macOS
-        // FSEvents watches the whole tree so it never misses — which is why it only ever flaked on
-        // the Linux runner. Ingest our own write directly; the watcher remains for EXTERNAL changes.
-        if (fileName.EndsWith(".md", StringComparison.OrdinalIgnoreCase))
-        {
-            var folder = path.Trim('/');
-            UpdateArticle(folder.Length == 0 ? fileName : $"{folder}/{fileName}");
-        }
+        // 🚨 Read-your-writes: merge a just-written markdown file into the stream RIGHT NOW instead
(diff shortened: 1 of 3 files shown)


rbuergi:
Superseded by #238 (merged, main green — "ingest a collection's own write immediately"), which lands the same read-your-writes fix (SaveFileAsync → proactive UpdateArticle) plus a watcher Created handler and path normalization, and its own deterministic regression test (ContentCollectionWriteIngestTest). This PR's core fix is now redundant and would conflict on merge.

Independent same-root-cause confirmation: I reproduced the failure deterministically by disabling the FileSystemWatcher (Linux inotify drops the create event for a file written into a just-created subdir; only the FILE-content render — which reduces the watcher-populated markdown stream — flaked; the folder listing reads the provider live, so it was never affected; CI-only because macOS FSEvents doesn't miss).

The only unique bit here is the optional `ContentCollectionConfig.DisableFileWatcher` knob (useful where the watcher is unreliable — network mounts / high load; the read-your-writes fix is what makes it viable). Happy to re-open a minimal PR with just that config if it's wanted; closing this to avoid redundant churn.

rbuergi:
Superseded by #238, which merged first and is live on main (green): `SaveFileAsync` proactive `UpdateArticle` ingest so read-after-write no longer depends on the `FileSystemWatcher`, plus a `Created` handler for external files and a deterministic `NoMonitorProvider` regression test (`ContentCollectionWriteIngestTest`) that fails without the fix and passes with it. Same root cause and same core fix — closing as duplicate. (If the `DisableFileWatcher` config knob from #240 is wanted, it can be lifted onto main as a small follow-up.)

PR #241, opened 2026-07-04
fix(content-collections): publish written .md articles directly — don't rely on the watcher for our own writes

rbuergi:
Root cause of the twice-observed CI cold-start hang (NodeHubContentCollectionTest, 30s timeout, Linux-only): the file render read from a stream populated only by the FileSystemWatcher, and `SaveFileAsync` relied on the watcher to notice a file it wrote itself. On inotify, a write into a just-created subdirectory races the (subdir-scoped) watch registration and is silently dropped under cold/2-core load — the article never publishes, the view never emits.

Fix: `SaveFileAsync` publishes the written article directly (watcher stays authoritative for external changes only) + the missing `.md` Created handler. No timeout widening, no retry.

Deterministic repro (monitor-less provider): 10s hang with the exact CI signature without the fix, 258ms with it. 16/16 + cold-first @2-core 8/8 + 20/20 under load; Release `-warnaserror` 0/0.

### DIFF
--- a/src/MeshWeaver.ContentCollections/ContentCollection.cs
+++ b/src/MeshWeaver.ContentCollections/ContentCollection.cs
@@ -137,8 +137,36 @@ public class ContentCollection : IDisposable
     /// <param name="path">The destination folder path within the collection.</param>
     /// <param name="fileName">The file name to write.</param>
     /// <param name="openReadStream">The content to persist.</param>
-    public Task SaveFileAsync(string path, string fileName, Stream openReadStream)
-        => provider.SaveFileAsync(path, fileName, openReadStream);
+    public async Task SaveFileAsync(string path, string fileName, Stream openReadStream)
+    {
+        await provider.SaveFileAsync(path, fileName, openReadStream);
+
+        // Proactively publish the markdown we just wrote into the live stream. The application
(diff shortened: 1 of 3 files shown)


rbuergi:
Superseded by #238 (merged, commit 18c1eb805 'ingest a collection's own write immediately — kill the CI-only file-render flake') — an independent, functionally-identical fix landed on main while this was in review: same seam (SaveFileAsync publishes its own write directly), same root cause (Linux inotify drops the change event for a file written into a brand-new subdirectory under IncludeSubdirectories, before the subdir watch is armed), same effect (the reactive file view never emits → the ContentCollections cold-start hang). No point merging a duplicate. The stuck 'pull_request' trigger here was just this PR going DIRTY once #238 touched the same file.

rbuergi:
Superseded by #238, which merged first and is live on main (green): `SaveFileAsync` proactive `UpdateArticle` ingest so read-after-write no longer depends on the `FileSystemWatcher`, plus a `Created` handler for external files and a deterministic `NoMonitorProvider` regression test (`ContentCollectionWriteIngestTest`) that fails without the fix and passes with it. Same root cause and same core fix — closing as duplicate. (If the `DisableFileWatcher` config knob from #240 is wanted, it can be lifted onto main as a small follow-up.)

PR #245, opened 2026-07-04
fix(chat): full-screen home composer; CLI-harness command routing; honest effort

rbuergi:
Follow-up to #215 (composer hit-target/focus). Addresses several chat-composer reports on memex.meshweaver.cloud.

## Changes
1. **Home composer → full-screen** ("should open full screen chat, not node / not side pane"). `UserActivityLayoutAreas.ComposerAreaView` builds the `ThreadChatControl` with `WithHideEmptyState(true)` — the compact/dashboard composer, whose submit navigates the new thread full-screen in the main pane (`ThreadChatView` reads `HideEmptyState` as `isCompact` → `NavigateTo("/{path}")`) instead of opening it in the side panel.

2. **CLI-harness command routing** ("login in claude code should route through to the harness"; "the only command we want to inject is /harness and the custom skills"; "omit the model command"). Under a CLI harness (Claude Code / Copilot) MeshWeaver injects **only** `/harness` (runtime switch) + custom MeshWeaver skills (`nodeType:Skill`). `/login`, `/logout` (harness authenticates itself), `/model` and `/agent` (harness owns model/agent selection) are **forwarded 1:1** to the harness. The inline Connect/login widget is now unreachable — its ~300 lines (fields + `StartHarnessConnect`/`SubmitConnect*`/`CancelConnect`/`DisconnectHarness`/`ResolveHarnessConfigDir`/`TryHandleHarnessCommand` + the razor widget + `@if(_connectProvider is null)` gates) are dead and slated for a focused deletion follow-up (kept out of this diff to avoid churn on a hot file).

3. **Honest effort level** ("effort shows medium which isn't real — must ask the harness"). `ClaudeCodeRuntimeProbe.ReadEffortLevel` no longer falls back to a hardcoded `"medium"`; it returns the value the harness actually has set, or `null` (status bar shows `…`). The probe is now called with the user's **own** per-user config dir (`{ClaudeCode:ConfigDirRoot}/{userId}/.claude` — the exact dir the harness runs the CLI under) instead of `null` (the server's `~/.claude`).

## Not in this PR (root cause pinned, deferred)
- **Model `glm-5.2 unavailable → z-ai/glm-5.2`**: `SelectionId.IdOf` over-strips a registered model id that contains a slash (`z-ai/glm-5.2` → `glm-5.2`) → false stale-model fallback in `AgentChatClient`. Core execution path; needs verification against the live catalog.
- **Collapse side-pane when the SAME thread opens in the main pane**: must key on real navigation vs. background nav-context emissions or it reintroduces the "chat vanishes during execution" bug.
- **Playwright E2E for Claude Code / Copilot.**

## Verification
- `dotnet build -c Release -warnaserror -p:CIRun=true` clean for `MeshWeaver.Blazor.Portal` (pulls Graph) and `MeshWeaver.AI.ClaudeCode`.


rbuergi:
Thanks for the review. Addressed in 549327723:
- **#2** (dispatch comment) — corrected to accurately describe that only harness-owned + node-pick commands are forwarded; everything else is skill-resolved.
- **#3** (effort-probe log) — the resolved dir is now logged instead of the raw (possibly null) parameter.
- **#1** (path separators) — intentionally left as-is: it mirrors the existing `ResolveHarnessConfigDir` and the portal runs in a Linux container (the CLI config dir is POSIX-pathed), so the `/` separator is correct for the runtime.

PR #246, opened 2026-07-04
feat(chat): /navigate skill — pane-aware, resilient navigation + honest NavigateTo

rbuergi:
## What & why

"Take me there" now works. When the assistant points you at a place, you can navigate there — and the resolution is **honest** (never claims to open a node that doesn't exist — the exact `@AiSettings` false-success bug from the reported thread) and **resilient** (finds the best match even when the path is imperfect).

### The `/navigate` skill (pane-aware)
A new `Navigate` skill-action kind + built-in **`/navigate`** skill. Pane-aware exactly like the context chip (`OnContextChipClicked`):
- thread in the **main pane** → target opens in the **side panel**;
- thread in the **side panel** → **changes the URL and navigates the main pane**.

So the conversation and where it sent you sit side by side.

### Resilient resolution (shared by the skill and the agent tool)
- **`NavigationTargetResolver`** — a pure, fully unit-tested algorithm:
  - `Classify`: a single path-shaped argument → **direct path first**; free text → **make sense of the context**.
  - `NormalizePath`: URL correction — strips a leading `@`, quotes, a pasted `https://host/…` prefix, a stray `/node/` segment, doubled slashes, and percent-encoding; preserves query strings.
  - `Score`/`PickBest` (search-fallback ranking) and `ScoreSkill`/`PickBestSkill` (intent→skill).
- **`NavigationResolver`** — reactive orchestration: authoritative `GetMeshNodeStream` existence check (no index lag) → search fallback when a path is off; free text → an **intent→skill** match first (so "change my model" runs **`/model`** — a skill *does* more than a route change), else the best node. Never dead-ends.
- **`MeshPlugin.NavigateTo`** (the agent's server-side tool) now routes through the resolver and **never reports success for a place that doesn't exist** — it returns the closest matches instead.

## Tests
- `NavigationTargetResolverTest` — 55 pure-algorithm cases (classify / normalize / route-vs-node / ranking / skill-intent).
- `MeshPluginTest` — resilient + honest resolution against the live mesh, incl. a non-existent-path honesty test.
- `SkillNodeTypeTest` — catalog + the `/navigate` Navigate action.
- Docs: `ChatCommands.md` documents `/navigate`; `DocumentationLinkIntegrityTest` green.

Built clean under `-warnaserror` (Release) for `MeshWeaver.AI` and `MeshWeaver.Blazor.Portal`, verified in a worktree merged with current `main`.

## Scoped follow-up (not in this PR)
Make the **agent's** tool drive a true opposite-pane open via a server→client navigation directive. Today the agent path resolves honestly and renders the resolved target inline (the existing live-only `DisplayLayoutArea` channel); full agent-initiated pane-aware navigation needs new plumbing through the chat streaming/rendering path and is intentionally deferred to keep this PR's CI green and low-risk. The `/navigate` skill already delivers pane-aware navigation when invoked.

### DIFF
--- a/src/MeshWeaver.AI/Data/Skill/navigate.md
+++ b/src/MeshWeaver.AI/Data/Skill/navigate.md
@@ -1,0 +1,30 @@
+---
+nodeType: Skill
+name: /navigate
+description: Take me there — open a node, doc, or page in the UI. Pane-aware (opens in the pane opposite the thread) and resilient (finds the best match even if the path is off).
+icon: Location
+category: Skills
+order: 3
+action:
+  kind: Navigate
+---
+
+Navigate the user to a place in the UI. Resolution is **resilient** and application is **pane-aware**.
+
+## Resolution — direct path first, else make sense of the row
+
+- **One argument that looks like a path** (`/navigate Doc/AI/ModelProviderSettings`, `/navigate @/rbuergi`, a pasted URL): resolve it as a **direct path** first. The URL is corrected automatically — a leading `@`, a stray `/node/` segment, a pasted `https://…host/…` prefix, doubled slashes, and percent-encoding are all cleaned up. If the exact node isn't there, fall back to the **best search match** rather than dead-ending.
+- **Free text** (`/navigate model settings`, `/navigate my notifications`): make sense of the context on the row — match it to a **skill** where one fits (so the user can *do* something, e.g. `/model` to change the model), otherwise **search** the mesh and open the best-matching node.
+
+Never report success for a place that doesn't exist. If nothing resolves, say so and offer the closest matches.
+
+## Application — always the opposite pane
+
+- **Thread is in the MAIN pane** → open the target in the **side panel** (so the conversation and its subject sit side by side).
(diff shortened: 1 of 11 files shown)


rbuergi:
Thanks for the review, Copilot — both points addressed:

1. **`contextPath` unused** — fixed. `ResolveDirect` now takes the context: a **bare single-segment** token (e.g. `@MyChild`) that isn't found at the root is retried under the thread's context path (`{context}/{token}`) before the search fallback. Multi-segment paths and explicit `@/absolute` inputs stay absolute, so the common case is unchanged.

2. **`PageRoutes` omits `_settings`** — the `PageRoutes` set was replaced by an `IsPageRoute` pattern switch (to satisfy `NoStaticCollectionsTest`). I deliberately did not add `_settings` because I couldn't verify it as a real top-level route in this portal, and mis-adding a route is worse than the fallback: a `/_settings/…` target that isn't a known page simply resolves as a node (exact read → search), which degrades gracefully. Happy to add it if it's confirmed.

PR #291, opened 2026-07-06
feat(events): EventSubscription continuation-handler seam (stage 5 foundation)

rbuergi:
## What

Adds **`IEventSubscriptionContinuationHandler`** — a DI-resolved seam so an `EventSubscription` continuation whose effect lives **above** `MeshWeaver.Graph` runs through a registered handler. The `EventSubscriptionRunner` dispatches any non-native `ContinuationType` (everything except the native `GrantSpaceAccess`) to the handler whose `Handles` matches, wrapped in `AsSystem` (system identity).

## Why

This is the **foundation for the durable delegation backstop** (stage 5 of the event-subscription unification). The delegation continuation (`PostThreadMessage` — post a sub-thread's result back into the parent thread) needs `MeshWeaver.AI`'s thread-submission surface, which `Graph` can't reference. The seam lets AI register that handler; the runner (in Graph) invokes it when a `NodeStatus`-resting subscription for a delegated sub-thread fires.

## Tests

`EventSubscriptionRunnerTest` (5/5): the existing NodeChange/Timer/NodeStatus/legacy coverage **plus** a new test proving a fired `PostThreadMessage` subscription dispatches to a DI-registered handler. Release `-warnaserror` clean.

## Next (the wedge-prone part — deliberately a separate, AI-test-validated pass)
The AI-side `PostThreadMessage` handler + wiring `ExecuteDelegationAsync` to create the durable backstop at delegation dispatch — **keeping the in-memory TCS fast path 100% intact** (the backstop only fires on reboot, guarded against double-delivery via the parent's unfinished-delegation check). That touches the delegation core with a subtle double-delivery/round-resume interaction, so it lands as a focused pass validated by the AI.Test shard.

### DIFF
--- a/src/MeshWeaver.Graph/EventSubscriptionRunner.cs
+++ b/src/MeshWeaver.Graph/EventSubscriptionRunner.cs
@@ -4,6 +4,7 @@ using MeshWeaver.Graph.Configuration;
 using MeshWeaver.Mesh;
 using MeshWeaver.Mesh.Services;
 using MeshWeaver.Messaging;
+using Microsoft.Extensions.DependencyInjection;
 using Microsoft.Extensions.Hosting;
 using Microsoft.Extensions.Logging;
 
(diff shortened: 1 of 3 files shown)


rbuergi:
Closing: the durable delegation backstop this PR's seam was built for rests on a false premise.

Investigation (see `InitializeThreadLifecycle` / `ResumeInterruptedRound` in ThreadExecution.cs) shows delegation is **already reboot-safe**: an interrupted `Executing` parent is re-run into its existing response cell, and its doc states *each child thread's own hub runs the same recovery recursively* — parent and sub-thread both self-heal. There is no stuck state to rescue.

Compounding this, the backstop's natural guard field `MeshThread.StreamingToolCalls` is `[JsonIgnore]` (transient) — empty post-reboot — so a backstop reading it can never match after a restart, and in-process it only duplicates the in-memory TCS fast path.

The generic `IEventSubscriptionContinuationHandler` seam here is clean but has no valid consumer, so we're not merging it as speculative infra (YAGNI). The genuine gap (re-run re-does the parent LLM turn and discards the sub-thread Summary) is a deeper resume-state-machine change with marginal payoff — deferred. Rationale recorded for future reference.

PR #310, opened 2026-07-06
fix(chat): Claude Code login — drop headless browser scrape; anchor credential to owner

rbuergi:
## Problem

On memex-cloud, logging Claude Code in via the **account/browser** method hangs when you paste the code. Cloud logs confirm:

```
CompleteConnect failed for rbuergi/ClaudeCode
System.TimeoutException: Timed out waiting for the Claude token after the code was submitted.
```

`claude setup-token` **masks** the token in stdout (and is TTY-gated), so the server-side scrape can never capture it → guaranteed `TokenTimeout`. The account/browser method cannot work headlessly, by design.

Separately, the apiKey/token **paste** ("Save fails") stored the credential node via `StoreHarnessCredential` with `MainNode` defaulting to **itself** under `_Memex` — diverging from the proven `ModelProviderService.CreateProvider` (which sets `MainNode = ownerPath`). A self-anchored `_Memex` satellite is accepted by the monolith's in-memory store but can be rejected by Postgres partition routing/RLS.

## Fix

- **Drop the "account (browser)" method** from the login dialog. It offered a headless-impossible CLI stdout-scrape. Users now paste a subscription token / API key directly (run `claude setup-token`, paste the `sk-ant-oat…`).
- **`StoreHarnessCredential` sets `MainNode = owner`** so the credential satellite anchors to the owner partition root, matching `CreateProvider`.

## Tests

`ConnectStrategyTest.ApiKeyPaste_StoreHarnessCredential_{WithoutPartitionRoot, WithPartitionRoot_Resolves, AsNonAdminOwner}` — 3/3 green. The store path is correct in-process (admin + non-admin, seeded/unseeded), asserting `MainNode == owner`.

## Notes

The connect stack was audited and is already fully pooled + reactive (`IIoPool` at every leaf) — no async deadlock. Verified live on memex.localhost (real Postgres); rolled + healthy there.


rbuergi:
Addressed both review points in e0398dd5a: (1) account/browser login is now suppressed only for ClaudeCode — Copilot's device-flow account login stays reachable; (2) SubmitLoginCredential classifies the pasted credential (sk-ant-api… → apiKey/ANTHROPIC_API_KEY, otherwise oauth/CLAUDE_CODE_OAUTH_TOKEN) instead of hardcoding apiKey.

PR #317, opened 2026-07-06
fix(create): remove CreateTransient — node Create menu broke on Postgres

rbuergi:
## The bug

The node **Create** menu showed **"Area not found — No renderer is registered for area \`{id}\`"** on every Postgres-backed portal (reported on an Agent at \`/TestSpace/Tom/Create\`, but it hit **every** node type).

## Root cause

The old create flow wrote a **\`Transient\`** placeholder node — \`meshService.CreateTransient\`, which also **bypassed the \`CreateNodeRequest\` access-control pipeline** (it wrote straight to \`IStorageAdapter\`) — then navigated to \`/{path}/Create\`. \`PostgreSqlSqlGenerator\` hardcodes \`n.state = 2\` (Active only), so path-resolution could not see the transient node → it treated the node id as an *area name* on the parent hub. InMemory tests have no state filter, so CI never caught it.

The standard areas were **never** missing — every node hub already inherits them via \`ConfigureDefaultNodeHub → AddDefaultLayoutAreas\`.

## The fix (remove \`CreateTransient\` entirely)

- **\`CreateLayoutArea\`** builds ONE \`Active\` node and persists it with a single **\`CreateNode\`** (through access control), then navigates to \`/{path}/Edit\`.
- **Editors materialise default content** when a fresh node has none — generic \`EditNode\` (via \`MeshDataSource.CreateContentInstance\`), \`AgentView\` (Details + Edit), \`ModelProvider\` — so a new Agent/Model shows an editable form instead of "Loading…".
- **Comment / reply / group** create flows carry the draft in-memory and \`CreateNode\` on submit (also fixes a latent bug where comment-create wrote to the *primary* node's path).
- **\`CreateTransient\` removed** from \`IMeshService\`, \`MeshService\`, \`HubNodePersistence\`, \`MeshServiceExtensions\`. \`MeshNodeState.Transient\` enum value kept (persisted contract). Docs updated.

## Verification

- Full-solution Release \`-warnaserror\` build (merged with main): green.
- Tests: \`CreatableTypesIntegrationTest\`, \`NodeCreationAccessTest\`, \`TodoCreateFlowTest\`, \`CreateLayoutAreaIntegrationTest\`, comment/reply — all pass.
- **Playwright E2E** (new \`CreateMenuE2ETest\`) against a real PG-backed DevLogin portal: Agent **and** Markdown create → land on \`/Edit\`, no "Area not found". 2/2 pass.

### DIFF
--- a/src/MeshWeaver.AI/AgentView.cs
+++ b/src/MeshWeaver.AI/AgentView.cs
@@ -118,9 +118,13 @@ public static class AgentView
                 (h, c) => host.Workspace.GetMeshNodeStream()
                     .Select(node =>
                     {
-                        var agent = AsAgentConfiguration(node, host.Hub.JsonSerializerOptions);
-                        if (node == null || agent == null)
+                        if (node == null)
                             return RenderLoading("Loading agent...");
+                        // A freshly-created Agent has no AgentConfiguration content yet — the create
+                        // flow persists an Active node with null Content and edits it afterward. Show a
+                        // default (empty) config instead of an endless "Loading…".
+                        var agent = AsAgentConfiguration(node, host.Hub.JsonSerializerOptions)
+                                    ?? new AgentConfiguration { Id = node.Id };
                         return BuildDetailsLayout(host, node, agent);
                     }),
                 "Content"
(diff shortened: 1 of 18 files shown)


rbuergi:
Thanks @copilot — addressed in 970cd507a:

- **Comment / reply / group create-failure handlers** now log via `ILogger<LayoutAreaHost>` instead of swallowing, and the comment & reply **draft forms stay OPEN on failure** (no longer clear the state) so the user's text isn't silently lost.
- **`BuildEditNodeContent`** no longer seeds the `/data` projection with the default content instance on every render — the form binds to the node stream, so the seed was redundant and could reset in-progress edits before the first write lands.

Re: the inline `UpdateData(newCommentPathStateId, "")` (CommentsExtensions.cs:288) — that lives in the view-**factory** lambda (`(h, _) => { … }`), which runs once when the area subscribes, not per stream emission, so it doesn't discard the draft on comment-list updates. Left as-is.

PR #333, opened 2026-07-06
fix(messaging): snapshot rule chain — concurrent Remove NREs the dispatch walk (sync-hub flake)

rbuergi:
## The flake

Sync-hub layout tests (\`EditorTest.TestEditorWithDelayed\`, \`EditPersistenceTest.EditState_ShouldSurviveDataUpdates\`, …) intermittently time out on CI — **a different one each run**. It turned main red once post-merge. Reproduced deterministically by running \`MeshWeaver.Layout.Test\` in bulk (whole project, one process): ~1 in 5 runs a sync-hub test fails.

## Root cause

\`MessageHub\` dispatches a delivery by folding over its rule chain (\`rules\`, a \`ThreadSafeLinkedList<AsyncDelivery>\`). \`HandleMessageAsync\` walked the chain via the **raw \`LinkedListNode.Next\`**, *outside* the list's lock:

\`\`\`csharp
current = current.Next;   // MessageHub.cs — outside ThreadSafeLinkedList's lock
\`\`\`

\`ThreadSafeLinkedList\` locks \`Add\`/\`Remove\`/\`First\`, but not this walk. When a handler-disposable's \`rules.Remove(node)\` fires concurrently (rapid sync-hub create/teardown churn under bulk load), \`LinkedList.Remove → Invalidate\` nulls the removed node's owning-list reference **before** its \`next\` pointer, so a racing \`get_Next()\` dereferences \`list.head\` → **\`NullReferenceException\`**. That fails the delivery → \`DeliveryFailure\` → the synchronization stream reports \`[SYNC_STREAM] OnError\` and propagates a \`StreamErrorEvent\` → the client subscriber's \`.Within(Ns)\` never matches → timeout. Which sync hub is being torn down at that instant decides which test flakes.

## The fix

Iterate a **snapshot taken under the read lock** instead of walking live nodes:
- \`ThreadSafeLinkedList.Snapshot()\` — \`CopyTo\` under the read lock.
- \`MessageHub.HandleMessageAsync\` folds over the snapshot array; a concurrent \`Remove\` can no longer invalidate the iteration.

Snapshotting is also semantically correct: a delivery is handled by the rules present when dispatch began.

## Verification

- Bulk repro (\`dotnet test test/MeshWeaver.Layout.Test\` looped): **~1-in-5 failed with the NRE before; 9+ consecutive clean after**.
- Full-solution Release \`-warnaserror\` build (merged with main): green.

## Note — a separate, rarer flake remains

\`TestEditorWithDelayed\`'s final \`.Within(30s)\` wait can still time out under **thread-pool starvation**: its projection does \`Thread.Sleep(100)\` (blocks a pool thread) inside \`ThrottleImmediate(20ms).Select(result)\`. Seen once under a heavily-loaded local box (whole run ballooned 13s→1m4s). This PR fixes the NRE (the likely CI red, since that test uses a sync hub); the thread-pool one is a separate follow-up if it recurs on CI.

### DIFF
--- a/src/MeshWeaver.Messaging.Hub/MessageHub.cs
+++ b/src/MeshWeaver.Messaging.Hub/MessageHub.cs
@@ -675,7 +675,7 @@ public sealed class MessageHub : IMessageHub
 
     private IObservable<IMessageDelivery> HandleMessageAsync(
         IMessageDelivery delivery,
-        LinkedListNode<AsyncDelivery> node,
+        AsyncDelivery[] ruleChain,
         CancellationToken cancellationToken
     )
     {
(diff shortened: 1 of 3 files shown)


rbuergi:
Thanks @copilot — added `test/MeshWeaver.Messaging.Hub.Test/RuleChainConcurrentDisposeTest.cs` in 5a4b6d78a: it hammers dispatch (300 round-trips over a 100-rule chain) while a background task disposes+re-registers rules, i.e. the exact remove-during-dispatch shape. **Proven both ways**: it fails on the pre-fix raw-`LinkedListNode.Next` walk (the delivery NREs → the response never lands → 10s timeout) and passes deterministically (3/3, ~240ms) with the snapshot fix.

(FYI the shard-3 red on the first run was a foreign flake — `MeshHostBuilderTeardownOrderingTest`, unrelated to this dispatch change; passes locally with the fix. The new push re-runs CI clean.)

PR #348, opened 2026-07-06
feat(social): LinkedIn company-Page sync via Community Management API

rbuergi:
## What

Adds the org-scope connect + REST pull engine for LinkedIn **company Pages** the signed-in user administers (e.g. Systemorph, MeshWeaver). Three new endpoints in `LinkedInPageSyncEndpoints.cs`, registered after `UseAuthentication`:

- `GET /connect/linkedin/org?page={path}` — org-scope OAuth start (`r_organization_social` + `rw_organization_admin`)
- `GET /connect/linkedin/org/callback` — stores the org credential + auto-discovers the org URN via `organizationAcls`
- `GET /connect/linkedin/sync?page={path}` — pulls the Page's posts (`/rest/posts`) + `organizationalEntityShareStatistics` → writes `Systemorph/Post` nodes (which flow into the content calendar)

## Why

Personal posts stay export-only — LinkedIn closed `r_member_social` to apps. But an **organization admin** can read a Page's posts and per-post engagement over REST via the Community Management API. This is the only API path to real LinkedIn post data + stats.

## Design

- Mirrors the existing `LinkedInConnectEndpoints` patterns (CSRF state cookie, token exchange, reactive mesh writes via `IMeshService`).
- Uses the versioned `/rest/` surface (`LinkedIn-Version` header).
- **Fails loud**: every upstream non-success surfaces its status + body rather than producing silent zeros.

## ⚠️ Untestable until product approval

The `/rest/` Community Management endpoints return `403` until the LinkedIn app (`780dsuvyxglmc4`) is approved for the **Community Management API** product. Therefore:

- Exact response field names and `LinkedIn-Version` (`202406`) are best-effort from the docs — **verify against the live API at approval-time** before relying on the numbers (flagged in code + XML doc).
- The new endpoints are **additive and inert** until a Page org credential exists (auth-gated, no-op otherwise) — they don't touch existing flows.

Follow-up at approval: add the `/connect/linkedin/org/callback` redirect URI to the app, run the sync end-to-end, adjust field names if LinkedIn's live shapes differ.

### DIFF
--- a/memex/Memex.Portal.Shared/MemexConfiguration.cs
+++ b/memex/Memex.Portal.Shared/MemexConfiguration.cs
@@ -981,6 +981,10 @@ public static class MemexConfiguration
         // UseAuthentication so HttpContext.User is populated.
         app.MapLinkedInConnect();
 
+        // LinkedIn company-Page sync (Community Management API) — org-scope OAuth +
+        // posts/statistics pull. Same ordering requirement (needs HttpContext.User).
+        app.MapLinkedInPageSync();
+
         // GitHub Sync — OAuth authorization-code connect endpoints (same ordering
         // requirement: needs HttpContext.User). Stores the per-user token at
         // {userId}/_Provider/GitHub. See Doc/Architecture/GitHubSync.
--- a/memex/Memex.Portal.Shared/Social/LinkedInPageSyncEndpoints.cs
+++ b/memex/Memex.Portal.Shared/Social/LinkedInPageSyncEndpoints.cs
@@ -1,0 +1,489 @@
+using System;
+using System.Collections.Generic;
+using System.Net.Http;
+using System.Net.Http.Headers;
+using System.Reactive.Linq;
+using System.Reactive.Threading.Tasks;
+using System.Security.Cryptography;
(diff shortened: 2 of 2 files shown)


rbuergi:
Thanks — addressed in df69056:
- **Open-redirect (both endpoints + callback):** added `IsSafePagePath` — rejects leading-slash, protocol-relative `//`, scheme, and `..`; the callback re-validates the cookie-carried pagePath before it hits a Location header.
- **`elements` array guard:** now checks `ValueKind == Array` before `GetArrayLength()` in the posts pull (stats/discovery paths already guarded), so a non-array 403 shape fails loud instead of throwing.
- **platformUrn / dedup:** re-sync is already idempotent by construction — each post is written to a stable path `{page}/posts/{sanitized-urn}` via `CreateOrUpdateNode`, so a re-run updates the same node rather than duplicating. `SocialMediaPost` has no `platformUrn` member (an unmapped key would be dropped), and the telemetry importer that used to join on it was removed with the broken Code pieces. If a durable per-post join key is needed later, adding `PlatformUrn` to the `SocialMediaPost` type is the right follow-up.

PR #394, opened 2026-07-09
Chat composer: slash-popup theme/width fix + price cloud [1m] model variants

rbuergi:
## What
Two small, self-contained chat-composer fixes (no `plugins/` dependency).

### 1. Slash-command / @-reference popup (Monaco suggest widget)
- **Follows the APP theme now**, not the OS. The old `@media (prefers-color-scheme: dark)` block forced a **dark popup in a light app** whenever macOS was in dark mode (reported: dark box with underlined items over a white side panel). Dropped it; the `html[data-theme=…]` rules already handle the app's real theme.
- **Stops spilling out of a narrow side panel.** The hard-coded `width: 550px` overflowed the ~350px side-panel composer into the page. Width is now responsive, plus a per-editor JS clamp caps the widget to the editor's own bounds. The clamp is `try/catch`-guarded and hooks the editor's own change events (no feedback loop) — worst case it's a no-op, it can never break Monaco.

### 2. Token-cost chip showed $0 for cloud models
`ModelPricing.Default` only stripped a `/` prefix, so `claude-opus-4-8[1m]` (the 1M-context variant) missed the `claude-opus-4-8` row and priced at **$0**. It now also strips a trailing `[…]`/`:tag` variant → base rate. Local/self-hosted models (e.g. `mythalion:latest`) correctly stay $0.

## Testing
- `MeshWeaver.AI` builds clean under `-c Release -warnaserror` (ModelPricing).
- Monaco `.razor.js` passes `node --check`.
- The popup's visual behaviour is best-effort (guarded) and will confirm on the running portal.

### DIFF
--- a/src/MeshWeaver.AI/ModelPricing.cs
+++ b/src/MeshWeaver.AI/ModelPricing.cs
@@ -107,17 +107,41 @@ public static class ModelPricing
 
     /// <summary>
     /// The built-in default rate for a model id, or null when the model isn't in
-    /// the table. Case-insensitive; tolerates a leading provider/path prefix by
-    /// also trying the last path segment.
+    /// the table. Case-insensitive; tolerates a leading provider/path prefix (by
+    /// also trying the last path segment) AND a trailing context/variant marker
+    /// (by also stripping a <c>[…]</c> / <c>:…</c> suffix off that segment). So the
+    /// bare-id rows still price the deployed variants — e.g. <c>claude-opus-4-8[1m]</c>
+    /// (the 1M context window) and <c>llama3:latest</c> both resolve to their base
+    /// row instead of falling through to <c>$0</c>.
     /// </summary>
     public static ModelPriceRate? Default(string? modelId)
     {
         if (string.IsNullOrWhiteSpace(modelId))
             return null;
-        if (Defaults.TryGetValue(modelId, out var rate))
-            return rate;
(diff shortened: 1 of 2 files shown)


rbuergi:
Addressed all four review points — thanks, two were real regressions from a stale-branch file copy:

1. **Monaco focus callbacks (`onDidFocus/BlurEditorText` → `HandleFocusChanged`)** — restored. My first push had wholesale-copied this file from a 233-commit-behind branch, which reverted main's anti-wipe guard. Rebuilt the branch as **main's files + only the intended deltas**; the focus wiring (and `editorHasFocus` reconcile suppression) is preserved.
2. **`ModelPriceRate` positional break** — reverted. Kept `CacheRead/WritePerMillion` as **init-only properties** exactly as main has them; the record's positional/`Deconstruct` shape is unchanged. The diff now only touches `Default()` + two private helpers.
3. **Clamp could expand the widget past 550px** — fixed. It now caps at `Math.max(220, Math.min(550, editorWidth))` — shrink-only, never grows past the CSS design width.
4. **`StripVariant` truncating a provider prefix with `:`** — fixed. It's now applied only to the **last path segment** (post-`/`), so a stray `:` earlier in the id is never mistaken for a variant marker.

Rebuilt clean under `-c Release -warnaserror` (0 warnings).

PR #406, opened 2026-07-10
Fix portal wedge: thread-cell write storm + DeleteNodeResponse AccessContext drop

rbuergi:
Two root causes of the *"a successful operation wedges the portal"* failure, found while debugging a live wedge, plus the assistant `create`-node guidance fix. Done against latest `main` in a clean worktree.

## 1. `DeleteNodeResponse` dropped → 60s delete wedge (AccessContext lost)
`HandleDeleteNodeRequest` posted its **success** response with a hand-rolled `WithTarget(sender)+WithProperty(RequestId)` instead of `ResponseFor(request)`. That post runs deep in the fan-out `.Do`/`.SelectMany` continuation, on the workspace emission scheduler where the ambient AsyncLocal `AccessContext` is **wiped** — so the response carried no context and the fail-closed PostPipeline **dropped** it. The caller never got a reply and its `DeleteNodeRequest` timed out at 60s **even though the delete succeeded**:
```
DeleteNodeResponse posted with no AccessContext → [STALE-CALLBACK] DeleteNodeRequest > 30000ms
→ TimeoutException: No response received … within 00:01:00
```
`PostFailed` already used `ResponseFor` (which auto-propagates `request.AccessContext`), so only the **success** path wedged. Fix: the Ok post uses `ResponseFor(request)` too.

**Pinned** by `ResponseFromContinuationAccessContextTest`: a response posted via `ResponseFor` from a wiped-ambient continuation carries the **caller's** identity; the hand-rolled post **loses** it (→ in prod, the fail-closed drop).

## 2. Thread-cell write **storm** → portal wedge (no graceful degradation)
When a response cell's per-node hub never delivers initial state, `cache.Update` throws *"no initial state arrived"* instantly on every push; the streaming loop pushes once per token, so the round **stormed** failed writes (`seq 744,745,746…` + `[STALE-CALLBACK]`), flooding logs and backing up the mesh/cache hub until the portal wedged. Added a round-scoped **circuit breaker** in `ThreadExecution`: latch on the first permanent failure, short-circuit every later push to `Observable.Empty` (no `cache.Update`, no storm), **log once at Info** (a dead cell is graceful degradation, not a per-token warning flood), and end the round normally.

## 3. Assistant `create` = mesh node, not a `.txt` file
`ToolsReference.md` now tells the assistants that **"create" means a mesh NODE** (default `nodeType: Markdown` when unsure, never a `.txt`/content-collection file via `UploadContent`), and to load the matching `/create-space` · `/agent` · `/code` · `/model` · `/layout-area` · `/slide` skill for specialized types.

## Testing
- `MeshWeaver.AI` builds clean under `-c Release -warnaserror` (covers ThreadExecution + Mesh.Contract).
- New `ResponseFromContinuationAccessContextTest` (2 tests) passes.
- The circuit breaker has no unit test — a deterministic dead-cell repro is impractical (needs a per-node hub that activates but never loads); the logic is small and guarded.

### DIFF
--- a/src/MeshWeaver.AI/Data/Agent/ToolsReference.md
+++ b/src/MeshWeaver.AI/Data/Agent/ToolsReference.md
@@ -297,6 +297,17 @@ Action: Call `NavigateTo('@Doc/Architecture')`, respond: "Here's the architectur
 
 Creates a new node in the mesh. The node is validated before being persisted.
 
+### 🚨 "Create" ALWAYS means a mesh NODE — never a `.txt` file
+
+When the user says **"create"** (a page, note, doc, list, plan, space, world, character, …) they mean a **mesh node** made with `Create`. They do **NOT** mean a file. Do **NOT** answer a "create" request by writing a `.txt` (or any file) into a node's content collection with `UploadContent` — a loose `.txt` file is never the right output for "create X". `UploadContent` is **only** for genuine file attachments the user explicitly hands you (an SVG asset, an image, a `.json`/`.csv` data file), never as the way to author content.
+
+- **Default to `nodeType: "Markdown"`.** If you're unsure which node type fits, create a **Markdown** node — it's the general-purpose document/content node and is almost always the right answer for prose, notes, plans, and pages.
+- **For specialized nodes, load the matching skill first**, then follow it — do not guess the shape:
+  - a **Space** / company / team / project / topic workspace → load **`/create-space`**
+  - an **Agent** → **`/agent`** · a **Code** node → **`/code`** · a **model / data type** → **`/model`** · a **layout area** → **`/layout-area`** · a **slide deck** → **`/slide`**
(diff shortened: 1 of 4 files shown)


rbuergi:
Thanks — re the `updateObs` double-subscribe at ThreadExecution:1086: that's **pre-existing** on `main` (the unconditional internal fire-and-forget subscribe + `return updateObs` so terminal-status callers can await the write) and is **not introduced or worsened by this PR** — my change only enriched the existing error handler and added the dead-cell early-return (which, when latched, returns `Observable.Empty` *before* any subscribe, so it can't double-write). A safe fix has to preserve the terminal-await contract (a naive `Publish().RefCount()` would starve terminal callers of the node emission; it needs `Replay(1)`/`AsyncSubject` semantics), so it's a separate, carefully-tested change rather than a rider on this wedge fix. Filing it as a follow-up; keeping this PR scoped to the wedge + AccessContext drop.

PR #407, opened 2026-07-10
AI models show their provider's brand logo in the model picker

rbuergi:
## Problem
Every AI model rendered an identical sparkle in the picker: `BuiltInLanguageModelProvider` stamped `Icon = "Sparkle"` on every `LanguageModel` node and `Icon = "Key"` on every `ModelProvider`. Providers were indistinguishable at a glance.

The picker (`MeshNodePickerControl`) already renders each node's `Icon` (URL / data-URI / inline-SVG / emoji), so the fix is purely to give the nodes real icons — no picker changes.

## What this does
- **13 brand icons** in `src/MeshWeaver.Graph/Icons/` (embedded → served at `/static/NodeTypeIcons/`), built to match the existing icon design exactly: a brand-coloured `rx=4` tile with a white glyph, so they're theme-independent by construction. Marks are from simple-icons (MIT), used nominatively.
- **`ModelProviderIcons`** resolver — maps a model to its maker by substring, **model id first, then provider name**. That ordering means a `claude-*` served through AzureFoundry, or `anthropic/claude-*` via OpenRouter, still resolves to Anthropic. Unknown → `null` → falls back to the neutral sparkle. Lives in `MeshWeaver.AI` (not `Graph`) to keep the generic layer free of provider knowledge.
- **Wired into** `BuiltInLanguageModelProvider` (the two icon assignments).
- Brands: Anthropic, OpenAI, Google (Gemini/Gemma), Mistral, Meta (Llama), DeepSeek, xAI (Grok), Qwen, Perplexity, GitHub Copilot, OpenRouter, Ollama, Azure.

## Tests
`ModelProviderIconsTest` — 23 cases: per-brand model-id mapping, the gateway case (AzureFoundry → Claude → Anthropic), OpenRouter-prefixed ids, provider-name fallback, and unknown → null. Green under `-c Release -warnaserror`.

## Scope note
Covers the **platform catalog** (what fills the picker for everyone). BYO models a user creates via the generic `create` path aren't auto-stamped yet — the picker reads `node.Icon`, so wiring the same resolver into the "add your own provider" flow is a clean follow-up.


rbuergi:
Thanks @copilot. Addressed the first point in 7044244d8 — the fallbacks now use the self-contained `/static/NodeTypeIcons/{key,sparkle}.svg` URLs instead of the legacy Fluent names.

On the second point (retrofit): correct — because `Provider/{name}` nodes are seeded `ExcludeThisAndChildren`, the new logos apply to fresh installs / partitions / local+test, but existing deployed catalogs keep their current icons until a one-time repair patches `Icon` on the existing Provider + child model nodes. Tracking that as a follow-up rather than bundling a migration into this PR.

PR #420, opened 2026-07-11
feat(feedback): /feedback skill + Feedback type; rename create-markdown → markdown

rbuergi:
## What & why

Adds a **`/feedback`** capability and renames a skill, per request.

### 1. Rename `create-markdown` → `markdown`
`content/ai/Skill/create-markdown.md` → `markdown.md` (`name: /markdown`), both skill-catalog test assertions re-sorted.

### 2. Assistant mandates the create-skills
`Assistant.md` now instructs the main agent to **load the matching skill first** before creating content (`/markdown`, `/create-space`, `/create-group`, `/feedback`) — never improvise the node shape.

### 3. `Feedback` type + `/feedback` skill
- **`Feedback` NodeType** — content record (`Feedback` + `FeedbackStatus`), registered in `AddGraph()` and the central type registry.
- **`/feedback` skill** — captures the user's **location** (the app-context node path the agent already receives) + **name** (from the `# Current User` prompt block) and files **one `Feedback` node per submission** into the dedicated **`Feedback`** space.

### 4. Dedicated `Feedback` space — auto-provisioned on every instance
- **`FeedbackStaticRepoSource`** seeds the top-level `Feedback` space (space root + a `Public → Contributor` grant), registered in `StaticRepoSyncExtensions`. **`SyncMode = Additive`** so user-filed feedback is never pruned on re-import; the importer provisions the writable PG partition.
- Result: the whole feature is **zero-touch after deploy** — no first-run bootstrap.

## 🔐 Access model (please eyeball)
- New built-in **`Contributor`** role = `Read | Create | Comment | Thread | Api` — **not** Update/Delete/Compile. It's a strict subset of `Editor` and grants nothing until an `AccessAssignment` references it.
- The `Feedback` space grants **`Public → Contributor`**, which folds into every authenticated user via `PermissionEvaluator` (same mechanism `Commenter` uses for public commenting). So **all users — including all platform admins** (who aren't data superusers and would otherwise be locked out) — can read the shared board and file feedback, but cannot edit/delete others' entries.
- **Note:** this is the shared-inbox model (everyone can read the board). If admins-**only** review is wanted, that's a follow-up (submit-only public role + admin-scoped read grant, or homing the inbox under the `Admin` partition).

## Tests
- Skill-catalog lists (rename + new `feedback`).
- `FeedbackNodeTypeTest` — typed round-trip proving registration.
[...]

PR #431, opened 2026-07-12
feat(feedback): private per-user feedback + admin System-scoped review tab

rbuergi:
## What & why

Follow-up to #420. Reworks the feedback **read model** to what was chosen: **each user sees only their own feedback; platform admins see everything.** (The shipped model was a shared space where every logged-in user could read all feedback.)

### Storage — private per user
`/feedback` now files each entry under the **submitter's own partition**: `{userId}/Feedback/{id}`. The **self-scope-owner rule** (`PermissionEvaluator`) grants a user Admin at the scope equal to their own id, so the entry is readable/writable by that user **only** — no other regular user has any access, and **no grant is needed**. Privacy is structural, not a policy bolt-on.

### Removed the old shared-space machinery
- `FeedbackStaticRepoSource` + its `StaticRepoSyncExtensions` registration (no shared space anymore).
- The now-unused built-in **`Contributor`** role (`Role` / `PermissionEvaluator` / `RoleNodeType`) — it existed only for the public-contribute grant.
- The `Public → Contributor` + `BreaksInheritance` seed.

### Admin review — `FeedbackReviewTab`
A platform-admin **Settings tab** (group "Administration"), gated on `hub.IsGlobalAdmin` at **both** the menu (tab visibility) **and** the content (before the query runs — defense-in-depth). It lists **all** feedback across user partitions via a single `nodeType:Feedback` query run **as System** — the explicit RLS bypass on `MeshQueryRequest.UserId` (`PostgreSqlMeshQuery` returns `""` = no filter for `UserId == WellKnownUsers.System`), not an `ImpersonateAsSystem`/AsyncLocal hop. Read-only `Controls.DataGrid` (submitter, summary, location, status, date).

## Tests
- `FeedbackPrivacyTest` — the owner has `Read | Create` on `{user}/Feedback/{id}` with no grant; another user gets `Permission.None`.
- `FeedbackNodeTypeTest` — type round-trip, updated to the `{user}/Feedback` location.

Full solution `dotnet build -c Release -p:CIRun=true -warnaserror` → **0 warnings, 0 errors**; both suites green locally.

## ⚠️ Note for reviewers
The **admin cross-partition read** is verified by construction (the `UserId == System` bypass is the documented mechanism) and by the gating tests, but the end-to-end "admin sees every user's feedback in the grid" is best confirmed on a **live PG portal** (the InMemory monolith doesn't exercise cross-schema RLS). Happy to validate via the `/playwright` real-portal harness before/after merge if you want.


rbuergi:
Addressed in 7476457ac: gave `ThreadAccessTest` its own test-local `CreateNoUpdate` role (Read+Create+Comment+Thread+Api, no Update) so it genuinely exercises "has Create, lacks Update → thread create rejected" instead of relying on a now-removed built-in. Security.Test 10/10 green.

PR #462, opened 2026-07-13
GitSync: sync git-committed content-collection binaries into the mesh (stop wiping course videos)

rbuergi:
## Problem

Git-committed content-collection assets under `{Space}/content/**` — the real blobs in the repo (course videos, posters, fonts, images) — were **never synced into the mesh content collection** on a GitSync import. They only ever existed as manual uploads and disappeared on every sync (this has repeatedly nuked all course videos on memex).

### Root cause (three-fold)

1. **`RepoFile.Content` was a `string`** and `OctokitGitHubRepoClient.DecodeBlob` UTF-8-decoded every base64 blob — **corrupting any non-text blob** (a `.mp4`/`.png` byte stream is not valid UTF-8; round-tripping it through a UTF-8 string mangles it).
2. **`GitHubSyncService.ParseSnapshot`/`ParseFile`** treated only `*.json`/`*.md` as nodes and dropped everything else — the `content/**` binaries were discarded before they reached the importer.
3. **`InMemoryStaticRepoSource`** (the source every GitHub sync uses) never captured the raw content blobs nor overrode the content-import hook, so nothing synced them, and the import's node-prune left the content unrepresented.

## Fix (end to end)

- **Binary-safe fetch.** `RepoFile` now carries raw bytes for non-UTF-8 blobs (`Binary`, plus `Bytes`/`IsBinary`). The Octokit fetch classifies each base64 blob **strictly** — valid UTF-8 → text, otherwise raw bytes — so a committed video survives byte-for-byte.
- **Owning-node resolution.** `ContentAssetMapper` classifies `{node}/content/**` paths: `content/videos/x.mp4` → owner = Space root; `TDD/content/x.png` → owner = `{Space}/TDD` (nearest node whose folder holds a `content/` dir; nested owners supported). The per-Space `content` collection is **one physical store** (mounted on the Space root, children inherit it), so it builds **one whole-collection `StaticContentSync`** — a single authoritative mirror, not overlapping per-owner mirrors that would prune nested content.
- **Sync + mirror.** New `SyncContentFilesRequest` (+ handler + fluent `hub.SyncContentFiles`) writes inline bytes into a node's content collection and **mirrors** the folder (add/update + prune stale) — binary-safe, idempotent. `IStaticRepoSource.EnumerateInlineContentSyncs` + `StaticRepoImporter.SyncInlineContent` drive it after the node upsert, under System.
- **Result:** after a GitSync import the content collection contains **exactly** the repo's `content/**` (added/updated/deleted); re-importing the same commit is a no-op; content the repo still carries is never pruned.

Reactive throughout (I/O only inside the collection's own pool leaves), immutable, warnings-as-errors, net10.

## Tests (regression pin)

- **`ContentAssetMapperTest`** — path classification, owning-node resolution, single whole-collection grouping, and `RepoFile` text/binary byte fidelity.
- **`GitSyncContentCollectionTest`** — the full `ImportFromGitHub` path over the in-memory fake repo: a committed `{Space}/content/videos/x.bin` lands byte-for-byte, a re-import of the same commit keeps it, and a binary removed from the repo is pruned while one still present survives (incl. nested-module content).

Verified locally: full solution builds in Release with `-warnaserror` (0 warnings); GitSync suite (121), Graph static-repo import (13), and Content import (18) tests all green.

## Deploy note

This is a core change — a **portal image roll is needed** to make it live. After the roll, a single `git_hub_sync {course} update` will restore + keep the videos from git (no more manual `upload-videos.py`).

### DIFF
--- a/src/MeshWeaver.ContentCollections/ContentImportExtensions.cs
+++ b/src/MeshWeaver.ContentCollections/ContentImportExtensions.cs
@@ -26,12 +26,119 @@ public static class ContentImportExtensions
     public static ContentImportBuilder ImportContent(this IMessageHub hub, string nodePath)
         => new(hub, nodePath);
 
+    /// <summary>Begin a fluent inline content sync targeting <paramref name="nodePath"/>'s hub.</summary>
+    public static SyncContentFilesBuilder SyncContentFiles(this IMessageHub hub, string nodePath)
+        => new(hub, nodePath);
+
     /// <summary>
-    /// Registers the <see cref="ImportContentRequest"/> handler. Wired into
-    /// <c>AddContentCollectionsInfrastructure</c> so every content-enabled node hub can receive an import.
+    /// Registers the <see cref="ImportContentRequest"/> + <see cref="SyncContentFilesRequest"/> handlers.
+    /// Wired into <c>AddContentCollectionsInfrastructure</c> so every content-enabled node hub can
+    /// receive a collection→collection import AND an inline (byte-carrying) content mirror.
     /// </summary>
     internal static MessageHubConfiguration AddContentImportHandler(this MessageHubConfiguration config)
-        => config.WithHandler<ImportContentRequest>(HandleImportContent);
+        => config
+            .WithHandler<ImportContentRequest>(HandleImportContent)
+            .WithHandler<SyncContentFilesRequest>(HandleSyncContentFiles);
+
+    private static IMessageDelivery HandleSyncContentFiles(
+        IMessageHub hub, IMessageDelivery<SyncContentFilesRequest> delivery)
+    {
+        var request = delivery.Message;
+        var contentService = hub.ServiceProvider.GetService<IContentService>();
+        if (contentService is null)
+        {
+            hub.Post(ImportContentResponse.Fail("Content collections not configured on this node"),
+                o => o.ResponseFor(delivery));
+            return delivery.Processed();
+        }
+
+        // The hub action block only subscribes + returns; every I/O leaf runs on the
+        // collection's own pool — this layer is pure reactive composition.
+        SyncFiles(contentService, request)
+            .Subscribe(
+                count => hub.Post(ImportContentResponse.Ok(count), o => o.ResponseFor(delivery)),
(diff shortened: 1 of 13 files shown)


rbuergi:
Scope note: this fixes the **GitSync import** path (a Space's own git repo, e.g. course videos). The plugin-registry re-serve path (`NodeRepoPackageSource`/`GitHubPackageSource` → `PackageFile`) is still text-only and would drop a binary a *plugin* ships under `content/**` — a pre-existing limitation, unchanged in practice (plugins ship `Source/*.cs` text, not content binaries). The `RepoFile` change makes that theoretical case fail *empty* rather than *corrupt*; node files (.json/.md, always valid UTF-8) are unaffected and still round-trip as text. Making `PackageFile`/`PackageInstaller` binary-aware is a clean follow-up if plugins ever need to ship content assets.

rbuergi:
Addressed all three Copilot review points in 5db87fd:
- **Path traversal** — the `SyncContentFilesRequest` handler now rejects unsafe paths up front (`IsSafeCollectionPath`: empty/rooted or any `.`/`..` segment in `TargetPath` or a file path), so a caller can never escape the collection root. New test `SyncContentFiles_RejectsPathTraversal`.
- **Permission** — new `SyncContentFilesPermissionAttribute`: always requires `Create`, and additionally `Delete` when `Mirror=true` (a pruning mirror deletes). A Create-only caller can't trigger deletes.
- **Perf** — `ContentAssetMapper` splits path-only classification (`IsContentPath`) from byte materialization; `TryClassify` takes a `Func<byte[]>` realized only for an actual content asset, and `ParseSnapshot` prechecks the path before touching `f.Bytes` — so a large repo pays no per-file allocation for its node files.

rbuergi:
CI note: shard 1 red was a **known, unrelated flake** — `MeshWeaver.Hosting.Monolith.Test` host **segfaulted (exit=139)**, not a test assertion failure (the shard's TRX gate reported "No failed tests in trx"). This PR touches no Hosting.Monolith files (only GitSync / ContentCollections / Mesh.Contract). Re-ran the failed shard; build + all other shards were green.

PR #467, opened 2026-07-14
Fix concurrent hub-construction SIGSEGV: evict Autofac reflection cache on collectible ALC unload

rbuergi:
## Root cause

The intermittent native-host **SIGSEGV (exit 139)** in concurrent per-node hub construction — the crash the #447 exit-marker gate unmasked in `FutuRe.Test` / `Hosting.Monolith.Test` — was **not** a lock/serialization problem. (The earlier global `SetupModules` lock was the wrong fix and is not used.)

`NodeAssemblyLoadContext` compiles nodes into **collectible** load contexts (`isCollectible: true`) and unloads them on recompile / release / teardown. Its `Dispose()` called `Unload()` but **never evicted Autofac's process-static shared reflection cache** (`ReflectionCacheSet.Shared`). Autofac keys that cache on `ConstructorInfo`, so after unload a retained key:

1. **roots the collectible context** → it can never be collected — a memory leak on every recompile; and
2. lets a later, unrelated **concurrent** `GetOrAdd` during framework hub construction (`AddMeshDataSource`) whose key hashes into a bucket holding the stale key compare against **freed metadata** → `AccessViolationException` in `ConcurrentDictionary.TryGetValueInternal` (the observed crash frame).

Autofac performs exactly this eviction automatically for scopes created via `BeginLoadContextLifetimeScope`; MeshWeaver manages its collectible node contexts by hand, so it must run the eviction itself.

## Fix

- New `MeshWeaver.ServiceProvider/ReflectionCacheEviction.EvictFor(AssemblyLoadContext)` — clears every shared-cache entry referencing the context's assemblies.
- `NodeAssemblyLoadContext` subscribes `Unloading += ReflectionCacheEviction.EvictFor` — runs **before** unload, while the metadata the predicate walks is still valid; static handler ⇒ no self-reference that would defeat collection. No lock, no serialization, no convoy. (Graph reaches ServiceProvider transitively; no new project reference.)

## Test (deterministic, not the flaky AVE)

`ReflectionCacheEvictionTest` is a root-cause repro: a collectible `NodeAssemblyLoadContext` activated through Autofac must be **GC-collectable after unload**. It pins `ReflectionCacheSet.Shared` alive (the cache is `WeakReference`-backed — otherwise a plain GC collects the whole cache and false-passes). **RED** pre-fix (context rooted), **GREEN** post-fix — a permanent regression guard, unlike the ~50% qemu-amd64 crash.

## Validation

- Deterministic test: RED → GREEN across the fix.
- **Graph.Test regression: 532/532 passed.**
- **CI-parity `Release -warnaserror`**: Graph.Test and Hosting.Monolith.Test (the crash site) both build with 0 warnings / 0 errors.

### DIFF
--- a/.github/workflows/alc-unload-probe.yml
+++ b/.github/workflows/alc-unload-probe.yml
@@ -1,0 +1,152 @@
+# Manually-triggered ROOT-CAUSE probe for the teardown SIGSEGV (exit=139) in the
+# ALC-heavy native test hosts (Hosting.Monolith.Test / FutuRe.Test). The crash is a
+# native use-after-unload: a process-wide reflection/serialization cache retains an
+# accessor/handle into a collectible node AssemblyLoadContext; when a node-hub disposes
+# it unloads that ALC (MeshDataSource.RegisterForDisposal → UnloadNodeContexts), and a
+# LATER background GC dereferences the freed metadata → SIGSEGV. The post-crash mini-dump
+# is a delayed-detection snapshot (the ALC is already gone → nothing to gcroot), so it
+# can't name the offending cache.
+#
+# This workflow flips MESHWEAVER_ALC_UNLOAD_GC_PROBE=1, which makes NodeAssemblyLoadContext
+# drive a SYNCHRONOUS full GC immediately after each Unload. A dangling native pointer then
+# faults AT that unload — on the Dispose thread, right after a
+# "ALC_UNLOAD_PROBE forcing GC after unloading <node>" log line — so the culprit node (and a
+# corruption-time dump with a managed Dispose→GC.Collect stack) is captured instead of the
+# opaque delayed crash. GCStress would be the classic tool but needs a checked runtime the
+# hosted runner doesn't have; forcing the collection at the unload point is the retail-runtime
+# equivalent for this specific use-after-unload.
+#
+# Run from the Actions tab (workflow_dispatch). Re-dispatch with project=MeshWeaver.FutuRe.Test
+# to probe the other crasher.
+name: ALC unload probe (manual)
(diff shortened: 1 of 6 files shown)


rbuergi:
**Status from a verification pass (driving this via /pullrequest): leak is fixed, but the SIGSEGV persists — not merging.**

- ✅ **Leak half verified.** The new `ReflectionCacheEvictionTest` passes deterministically locally — the unloaded `NodeAssemblyLoadContext` is now GC-collectable, i.e. the stale `ConstructorInfo` key is evicted from `ReflectionCacheSet.Shared`. Build is clean under CI flags.
- ❌ **Crash not closed.** CI run [29307749907](https://github.com/Systemorph/MeshWeaver/actions/runs/29307749907) shard 1 still failed with `[CI] MeshWeaver.Hosting.Monolith.Test exit=139` (the same concurrent-construction SIGSEGV). So evicting on `NodeAssemblyLoadContext.Unloading` is **necessary but not sufficient**.

Reading: consistent with the investigation's note that beyond the ALC keys a *genuine Autofac concurrency bug / Reflection.Emit factory* stayed un-isolated — so there's a second contributor, a residual eviction window, or (given the ~50% qemu intermittency) the fix reduces-but-doesn't-eliminate. One red run can't distinguish those, but it's definitive the crash can still fire with this fix.

Not merged (green-main gate). Keeping this branch — the leak fix is worth landing once the residual crash cause is also closed. Confirmed independently: crash point is disconnected from any active construction (dump: all threads idle at crash), so it's use-after-free detonating on a later read — dump can't isolate the corrupting op further.

PR #475, opened 2026-07-14
Mesh hub: never persist its transient node; never NACK a message addressed to it

rbuergi:
The root mesh hub's address (`Address.Type == \"mesh\"`) is **transient per-process infrastructure** — a fresh guid every process (`CreateMeshAddress`), not an addressable node. It leaked two ways, each a known catastrophe class. This PR fixes both; a third (routing registration) is deferred because the naive version deadlocks — see below.

## B — never persist the transient node (`StoragePostCommitFlush.Flush`)
When the flushing hub is the mesh hub, log an error and skip the write (the ack still resolves `true`, so read-after-write is intact). Persisting `mesh/{id}` orphaned a DB row on every restart — the stray `mesh/{id}` rows found in the database. Same behaviour as portals: routable, never persisted.

## Receive-guard — never NACK back (`MessageHub.FinishDelivery`)
A message addressed to the mesh hub that reaches no handler is **logged as an error** (to expose the abusing sender) and returns **`delivery.Processed()`** — never a `DeliveryFailure` back to the sender. Echoing an error back wedges the sender (it treats it as a fault → retries → storm → the 60s-timeout mesh-wide-outage class).

**Non-obvious trap the test caught:** `delivery.Ignored()` is *not* enough — `MessageService` turns an on-target `Ignored` into a `DeliveryFailure{Ignored}` sent back to the sender (see `UnhandledMessageReportsFailureTest`). Only `Processed()` fully suppresses the echo; a requesting sender simply times out.

## Test
`MeshHubNotAnEndpointTest` — an unhandled `IRequest<T>` addressed to the mesh hub must yield a **`TimeoutException`** (Ignored / no-NACK), never a `DeliveryFailureException`. The exact inverse of `UnhandledMessageReportsFailureTest`. Red → green across the fix.

## Validation
- Dedicated test red→green; **full `Messaging.Hub.Test` 89/89**; a `MonolithMeshTestBase` test green (mesh starts fine).
- `-warnaserror` clean: `Messaging.Hub`, `Hosting`.

## Deferred (separate): register the mesh hub as a routing participant
So routing *resolves* `mesh/{id}` instead of `CreateHub`-ing it. The naive `WithInitialization → RegisterStream(hub)` **deadlocks the mesh at startup** (deterministic SIGABRT at the 60s timeout): `IRoutingService` is mesh-scoped, so resolving it *during the mesh hub's own construction* is circular. The correct fix is a post-construction hook or routing-side self-recognition of the mesh address — a separate, carefully-tested change, not rushed in here.


rbuergi:
Closing: all three approaches in this PR were defective and reverted — (A) registering the mesh hub in its own WithInitialization deadlocks (IRoutingService is mesh-scoped → circular); (B) the never-persist guard checked the injected hub, which is the SINGLETON mesh hub for every node's flush, so it skipped ALL persistence (16 data-change/comment/version/activity tests red); (C) the receive-guard returned Processed() for unhandled mesh-hub messages, but legitimate flows POST-and-await the mesh hub, so they hung → timeout. The stray mesh/{id} DB rows need the actual write path confirmed first, then a fix keyed on the NODE identity, verified locally against the persistence tests. Documented for a proper redo.

Issue #321, opened 2026-07-06
AI agent runs ahead without honoring Stop, and shows no feedback between tool calls


sierragolflima:
## Problem

A high-volume AI agent (any agent that issues many tool calls per round) exhibits three related failures during a multi-step run:

1. **Runs ahead without pausing** — dozens of tool calls in a single round, driven only by the LLM's own judgement, with no framework-enforced cap.
2. **Does not react to Stop immediately** — pressing Stop eventually cancels the round, but the agent keeps executing the current tool call (up to its ~30s timeout) before cancellation takes effect, with no acknowledgement that Stop was received.
3. **No feedback between tool calls** — while the LLM reasons between tool invocations (not streaming text, not inside a tool call), the UI shows only the last `Thread.ExecutionStatus` from the previous tool call, so "thinking" is indistinguishable from "stuck".

### 1. No per-agent tool-call cap
`AgentConfiguration` (`src/MeshWeaver.AI/AgentConfiguration.cs`) has no `MaxToolCallsPerRound` field. `FunctionInvokingChatClient.MaximumIterationsPerRequest` is read/logged at `src/MeshWeaver.AI/ChatClientAgentFactory.cs:~225` but never set per-agent — it falls back to the `Microsoft.Extensions.AI` default of 128. An agent that reads a dozen nodes per item across hundreds of items can issue 400+ tool calls in a single round before hitting that floor.

### 2. Cancellation does not acknowledge immediately
Stop path: user clicks Stop → `RequestedStatus = Cancelled` written to the thread node → `executionCts.Cancel()` fires in `InstallCancellationWatcher` (`src/MeshWeaver.AI/ThreadExecution.cs:~1538`) → the streaming loop receives `OperationCanceledException` on the next iteration, after the current tool call's timeout elapses or it completes. During that window the UI still shows the previous tool-call status with no indication the cancellation was received.

### 3. No feedback during LLM reasoning
Progress updates are emitted via `PushToResponseMessage` sampled at 100ms (`StreamingSampleInterval`, `ThreadExecution.cs:~1636`), covering streaming text and active tool calls. Between tool calls the status is not updated, producing stretches of silence indistinguishable from a stuck agent. The 90s stuck-round watchdog (`StuckGracePeriod`, `ThreadExecution.cs:534`) only fires when the thread emits nothing for 90s; an actively tool-calling agent emits continuously, so it never triggers even if the agent loops far beyond intent.

### 4. No "check in with user" enforcement
[...]

PR #483, opened 2026-07-15
Fix teardown SIGSEGV: MeshQuery 100% reactive + IIoPool cancel+join drain

rbuergi:
## Root cause (finally named)

The intermittent native **`exit=139` (SIGSEGV)** in ALC-heavy test hosts (`Hosting.Monolith.Test`, `FutuRe.Test`) — the flake that has turned `main` red on and off — is a **use-after-unload during teardown**, not the Autofac reflection cache (#467 was orthogonal, as the dump analysis showed).

Named via the `alc-unload-probe.yml` run (`teardown-stragglers.log`):
- `MeshQuery` ran its provider merge on `SubscribeOn(TaskPoolScheduler.Default)` → **unowned thread-pool work that outlives mesh disposal** (58× `MeshQuery.<>c__DisplayClass19_*` still running during teardown).
- Meanwhile `ServiceProvider.Dispose()` → `CompilationCacheService.Dispose()` **unloads the collectible node ALCs**; the stragglers dereference the freed ALC metadata → native crash.
- The teardown *tried* to drain first, but `IoPoolRegistry.WhenDrained().ToTask()` only **waits** — it never **cancels**. A live change-feed leaf never completes on its own, so the wait timed out and the scope disposed under it.
- Bonus: the probe never named the culprit before because its own `LogWarning` resolves a logger from the already-disposed container and throws — blind exactly where it fires.

## Fix

Per the architectural directive (*MeshQuery fully reactive to the IO boundary; `IIoPool` only for external IO; `DisposeAsync ⇒ Dispose`*):

- **`MeshQuery` is now 100% reactive** — deleted all three `SubscribeOn(TaskPoolScheduler.Default)`. Safe because every provider already defers external IO to its own `IIoPool` (`StorageAdapterMeshQueryProvider`) or is pure in-memory `Observable.Return` (`StaticNodeQueryProvider`), so Subscribe never blocks the hub. The owned pool is the only async boundary — no unowned straggler.
- **`IoPool` drains synchronously.** A pool-wide token is linked into every leaf; new `Drain()` **cancels** all in-flight leaves and **joins** by re-acquiring all gate permits (no poll/sleep/signal — it reuses the semaphore the pool already owns). `Dispose()` drains before releasing, so when it returns no pool thread is running and the caller may safely unload node ALCs.
- **Teardown cancels instead of just waiting.** `IoPoolRegistry.DrainAll()` drains every pool; `MeshTeardownExtensions` + `MonolithMeshTestBase` call it synchronously in place of `WhenDrained().ToTask()`.

## Verification

- New deterministic test `IoPoolTest.Drain_cancels_in_flight_leaves_and_joins_synchronously`: a never-completing in-flight leaf is cancelled and joined — `CurrentInFlight == 0` the instant `Drain()` returns. **11/11 IoPool tests pass.**
- All changed projects build green under Release `-warnaserror`.
- The real integration proof is this PR's CI (the full `Monolith.Test` shard that was crashing); I'll also re-run the ALC-unload probe against this branch.

No user-facing change → no What's New entry (pure internal stability fix).

### DIFF
--- a/src/MeshWeaver.Graph/Configuration/CompilationCacheService.cs
+++ b/src/MeshWeaver.Graph/Configuration/CompilationCacheService.cs
@@ -427,7 +427,11 @@ internal sealed class NodeAssemblyLoadContext : AssemblyLoadContext, IDisposable
         // an accessor into a collectible node assembly (the exit=139 teardown crash).
         if (UnloadGcProbe)
         {
-            _logger?.LogWarning("ALC_UNLOAD_PROBE forcing GC after unloading {ContextName}", Name);
+            // Write to Console DIRECTLY, not via _logger: this runs INSIDE ServiceProvider.Dispose()
+            // where the ILogger (XUnitFileLogger.GetMinLogLevel) resolves from the already-disposed
+            // container and throws ObjectDisposedException BEFORE writing — so the probe never named
+            // the culprit. Console is the only sink alive during teardown.
+            Console.Error.WriteLine($"ALC_UNLOAD_PROBE forcing GC after unloading {Name}");
             GC.Collect();
             GC.WaitForPendingFinalizers();
             GC.Collect();
--- a/src/MeshWeaver.Hosting.Monolith.TestBase/MonolithMeshTestBase.cs
+++ b/src/MeshWeaver.Hosting.Monolith.TestBase/MonolithMeshTestBase.cs
@@ -1164,15 +1164,18 @@ public abstract class MonolithMeshTestBase : Fixture.TestBase
             await WaitWithProgressAsync(testName, sw, cts.Token);
             // DisposalCompleted only drains the action blocks + message round-trips.
             // Offloaded I/O (IIoPool) runs on the ThreadPool independently and is NOT
-            // covered — wait for it here, BEFORE base.DisposeAsync tears down the
-            // service scope, so no continuation resolves a disposed Autofac scope
-            // (the unobserved "catastrophic ObjectDisposedException" class).
+            // covered — CANCEL + JOIN it here, BEFORE base.DisposeAsync tears down the
+            // service scope, so no continuation resolves a disposed Autofac scope AND no
+            // ThreadPool leaf dereferences a collectible node ALC's freed metadata after
+            // unload (the teardown use-after-unload SIGSEGV). A live change-feed leaf never
+            // completes on its own, so a WAIT-only drain would time out and let the scope
+            // dispose under it; DrainAll() cancels the leaf so it stops, then joins.
             if (ioPools is not null)
             {
-                await ioPools.WhenDrained(DisposeTimeout).FirstAsync().ToTask();
+                ioPools.DrainAll();
                 if (ioPools.TotalInFlight > 0)
                     FileOutput.WriteLine(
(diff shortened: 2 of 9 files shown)


rbuergi:
**Independent investigation (parallel session)** — while chasing the main-red `Monolith.Test exit=139` I confirmed this branch fixes that SIGSEGV (shard 1 / `Monolith.Test` passes 👍). But the CI run surfaced **two distinct regressions**, both green on `main` and red only here (commit `57435f320`). Sharing the diagnosis so you don't have to re-derive it.

---

### 1. `CancellationAckTest.Stop_ImmediatelyStampsCancellationAck_BeforeRoundTearsDown` — **flaky**, timing race

- `main`: **20/20 pass** locally · `#483`: **19/20 pass** locally (+ the CI failure). Not a hard break; a race this branch **widened**.
- The ack (`ExecutionStatus = CancellationRequestedStatus`) is a *transient* stamped by the async `InstallCancellationWatcher` (`ThreadExecution.cs:2842`), **guarded to `IsExecuting()`**. The test catches that transient before the terminal `Cancelled` write clears it.
- Emission timelines (instrumented both branches):

```
main (PASS):  +21ms cancel visible, Status=Executing
              +29ms Exec="Cancellation requested…"  ← ACK stamped, still Executing → observed ✅

#483 (FAIL):  +167ms cancel visible, Status=Executing
              +188ms Status=Cancelled (TERMINAL)     ← Exec=<null> throughout, ack NEVER emitted ❌
```

- **Cause:** the synchronous cancel+join tears the *hanging* round down fast enough that the ack-watcher loses the race to the terminal write — the transient is never emitted (guard sees `!IsExecuting()`, or it coalesces with the terminal write). Worse under CI load.
- **Fix direction (your call):** with instant teardown, the transient ack is arguably obsolete — its purpose (#321) was to cover the ~30s freeze while a hung tool call drained, which this PR eliminates. Terminal `Cancelled` now *is* the immediate acknowledgment, so the test could accept **either** the transient ack **or** a prompt terminal `Cancelled`. Alternatively, stamp the ack synchronously in the same turn the cancel is observed, before initiating teardown.

---

### 2. `OrleansMarkdownExportTest.ExportPdf_RoundTrips` + `ExportDocx_RoundTrips` — **looks deterministic**, lost AccessContext

Both fail with:

> `Failed to create ActivityLog node: AccessContext must never be null for an application post … hub=Templates/Export/Pdf, message=CreateNodeRequest … was posted with no AccessContext. The post lost the user identity.`

- Green on `main`, red here → this branch drops the user identity on the export's `CreateNodeRequest`.
- **Likely cause:** `AccessService.Context` is an `AsyncLocal` that flows on the captured `ExecutionContext`. Deleting the `SubscribeOn(TaskPoolScheduler.Default)` hops in `MeshQuery` changes which thread/continuation the export's node-create runs under, so the context isn't present at the post (fail-closed then throws, per `AccessContextPropagation.md`). This is a correctness concern beyond the test — worth confirming the export path carries `AccessContext` explicitly (e.g. `AccessContextScope.FromNode` / `ImpersonateAsHub`) rather than relying on the removed scheduler hop.

---

Both are consequences of the (correct!) reactive/teardown change interacting with code that leaned on the old scheduler offloads. Happy to dig further on either if useful — I've left your branch untouched.

PR #488, opened 2026-07-15
Fix teardown SIGSEGV: await in-flight hub construction before disposing the collection

rbuergi:
## Root cause (proven, not guessed)
The endemic `Hosting.Monolith.Test exit=139` (native SIGSEGV, reds ~half of all PRs) is a **teardown race**, pinned from the `alc-unload-probe` teardown-stragglers.log:

A long-lived `MeshQuery` subscription runs on the raw threadpool (`TaskPoolScheduler.Default` — untracked by `MeshTeardownExtensions`' drain). During teardown it routes → `CreateHub` → `serviceProvider.CreateMessageHub` → Autofac `BeginLifetimeScope`, which recursively builds a `SynchronizationStream` + sub-hubs (seconds). The host disposes the Autofac scope after the drain (which awaits `DisposalCompleted`) **while that construction is still running** → the activator derefs freed metadata → native SIGSEGV. `Dispose()` already `CloseCreation()`s new constructions at dispose-start, so it was ~50% (the check-then-act residue).

## Fix
`HostedHubsCollection.DisposeHubsReactive` now **awaits in-flight hub construction** (the bounded `creations` Lazy-dict — only drains once `CloseCreation` ran) **before** snapshotting + disposing children. So `DisposalCompleted` — and the whole teardown drain — cannot complete while a `BeginLifetimeScope` is in flight, and the scope is never torn down under a construction. Snapshotting after the drain also includes any hub a just-finished construction produced (previously leaked past the snapshot).

## Verification
The `alc-unload-probe` on this branch survived **2 iterations that pre-fix crashed on iteration 1** — the primary query-straggler race is closed. (Iteration 3 crashed under the probe, but the probe *forces a `GC.Collect()` inside every ALC unload*, over-stressing teardown far beyond normal CI and amplifying a separate ALC-unload-vs-container-dispose race — that's a diagnostic artifact, not a normal-CI path.) **This PR's normal full-suite CI (no probe env) is the fair test.**

Contained to disposal; does not touch MeshQuery's hot path. `#467` (reflection-cache eviction) was a real leak fix but orthogonal.

### DIFF
--- a/src/MeshWeaver.Messaging.Hub/HostedHubsCollection.cs
+++ b/src/MeshWeaver.Messaging.Hub/HostedHubsCollection.cs
@@ -205,41 +205,57 @@ public class HostedHubsCollection(IServiceProvider serviceProvider, Address addr
     private void DisposeHubsReactive()
     {
         var totalStopwatch = Stopwatch.StartNew();
-        var hubs = messageHubs.Values.ToArray();
-        logger.LogDebug("Starting disposal of {count} hosted hubs: [{hubAddresses}]",
-            hubs.Length, string.Join(", ", hubs.Select(h => h.Address.ToString())));
 
-        var childCompletions = hubs.Select(h =>
-        {
-            var address = h.Address;
-            try
+        // 🚨 FIRST wait for any IN-FLIGHT hub CONSTRUCTION to finish, THEN snapshot + dispose.
+        // A construction runs serviceProvider.CreateMessageHub → SetupModules → Autofac
+        // BeginLifetimeScope, which recursively builds a SynchronizationStream + sub-hubs and can
+        // take seconds. If the owning Autofac scope is disposed (by the host, after TeardownAsync's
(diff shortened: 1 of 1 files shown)


rbuergi:
Closing — falsified under normal CI. Diagnosis: this awaits in-flight hub `creations` at the mesh's DisposeHostedHubs phase, but the Autofac scope is disposed LATER (by the host, after TeardownAsync). Query stragglers keep arriving from the UNTRACKED TaskPoolScheduler.Default in that gap → still crash `Monolith.Test exit=139` (residual stragglers show the same MeshQuery → CreateHub → GetRequiredService → LifetimeScope.ThrowDisposedException path). Wrong layer. The real fix must stop/drain the MeshQuery threadpool stragglers before scope disposal — i.e. move MeshQuery off TaskPoolScheduler.Default onto a drainable pool (IIoPool) AND cancel+join it at teardown (#483's IoPoolRegistry.DrainAll direction). Superseded by that approach.

PR #531, opened 2026-07-18
Markdown nodes: clean @@ embeds (no header/comments) + collapsible sub-node side menu

rbuergi:
Two improvements to the Markdown node's Overview rendering.

## 1. `@@` embeds hide the header + comments by default
An inline `@@` embed rendered the target node's **full page** — its header (title/icon/menu) *and* its comments section. On a page that consolidates several sub-nodes via `@@`, that duplicated the markdown's own heading and dropped a comments box under every section.

- `LayoutAreaMarkdownRenderer` now appends `?hideHeader=true` to the raw path of **inline** (`@@`) references.
- `MarkdownOverviewLayoutArea` reads it (`Reference.HasParameter`) and skips the header + comments.
- Top-level page renders carry no such parameter → unchanged. Authors can force it back on with `@@node?hideHeader=false`. Non-markdown areas ignore the unknown parameter.

## 2. Collapsible side menu of sub-nodes
A markdown node with children now renders a **collapsible left-hand `NavMenu`** of those sub-nodes beside its content — so a top node that splits a document into sub-pages is navigable. Built from a live `ObserveChildren("is:main")` query; internal satellite children (`_Access`, `_Thread`, …) are excluded; suppressed for `@@` embeds and when there are no children.

## Testing
- Built `MeshWeaver.Graph` and `MeshWeaver.Markdown` locally — **0 errors**.
- Not run in a portal yet — behavioural verification (embeds render content-only; the raw-path resolver threads `?hideHeader`; the side menu collapses) needs a portal build/deploy via CI.

Motivating page: `PartnerRe/EslProposal` (a proposal split into six `@@`-embedded sections).

### DIFF
--- a/src/MeshWeaver.Blazor/Components/MarkdownHtmlRenderer.cs
+++ b/src/MeshWeaver.Blazor/Components/MarkdownHtmlRenderer.cs
@@ -133,7 +133,12 @@ public class MarkdownHtmlRenderer
                     var rawPath = node.GetAttributeValue($"data-{LayoutAreaMarkdownRenderer.RawPath}", "");
                     if (!string.IsNullOrEmpty(rawPath))
                     {
-                        RenderLayoutAreaFromPath(builder, rawPath);
+                        // @@ embeds hide the node header by default; the renderer emits
+                        // data-show-header='false' unless the author opted in with ?showHeader=true.
+                        var showHeader = !string.Equals(
+                            node.GetAttributeValue($"data-{LayoutAreaMarkdownRenderer.ShowHeader}", "true"),
+                            "false", System.StringComparison.OrdinalIgnoreCase);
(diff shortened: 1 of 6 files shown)


rbuergi:
Left unmerged while sweeping the open PRs — the three failing shards are real and in this PR's own feature area (`DocumentationEmbedIntegrityTest.PythonCodeNodes_LiveEmbed…`, `RelativeUcrReference_ResolvesToExpectedNode`, `FutuReAnalysisTest.AnnualReport_Overview_ShouldContainLayoutAreaReferences`, `MauiMarkdownAtAtOperatorTest.AtAt_Path_EmitsLayoutAreaElement_NotLiteralText`) — the embed-markup change breaks the existing @@ contract these tests pin. Needs the tests reconciled with the new embed shape (or the shape fixed) before merge.

Issue #471, opened 2026-07-14
User enumeration via search still possible on memex.meshweaver.cloud — User nodes expose email addresses (PII)


rosalie1984:
## Environment
- **Portal:** https://memex.meshweaver.cloud (memexcloud instance)
- **Observed:** 2026-07-14
- **Related:** #385 (Problems 3 and 4 — search and access menu expose users). This is a follow-up: the search-side user enumeration reported there is still reproducible, and additionally the User node itself discloses PII when opened.

## Problem 1 — Search still enumerates all users
A plain search filtered to user nodes (`nodeType:User`) returns **every provisioned user** of the instance (16 User nodes at time of writing), including users the searcher shares no space or content with. This reproduces both through the portal search UI and through the MCP `search` tool.

## Problem 2 — Opening a User node exposes the email address
Clicking any of those search results opens the User node, whose content includes the user's **email address**, along with full name and role. Example shape (own node, PII redacted):

```json
{
  "nodeType": "User",
  "content": {
    "$type": "User",
    "fullName": "…",
    "email": "…@….com",
    "role": "…"
  }
}
```

So any authenticated user can enumerate the full user base **and** harvest every user's email address in two clicks.

## Expected behavior
- Search (portal + MCP) must not return User nodes of users the caller is not entitled to see — same row-level filtering as direct reads (see #385 expected behavior).
- Even where a User node is legitimately visible (e.g. a collaborator on a shared space), the **email address should not be part of the generally readable node content** — it is PII and should be restricted (e.g. visible only to the user themself and platform admins), or moved out of the public User node content entirely.

## Notes
- #385 Problem 3/4 covered the enumeration surfaces; no fix appears to have landed yet (issue still open, no merged PR found). This issue adds the PII-disclosure angle: the payoff of the enumeration is currently every user's email address.
- GDPR-relevant: email + full name + role of all users disclosed to any authenticated account.

Reported by @rsalzmann.

rbuergi:
## RCA (verified against source)

Four independent roots; the email leak is the urgent one.

[...]

PR #535, opened 2026-07-18
Polished owner-editable user profile with curated Showcase

rbuergi:
## What this renders

Enhances the existing visitor-profile surface (`UserActivityLayoutAreas.BuildVisitorProfile`) into a **polished public profile** — it is not a rewrite or a duplicate; the old visitor path now flows through the new `BuildProfile`. The profile has these sections:

- **Header** — avatar + display name (reuses the existing `UserProfileControl`). Email is shown **only** to the owner or a global admin.
- **Bio** — an opt-in short bio, rendered read-only as markdown.
- **Links** — an opt-in markdown links block (one `[label](url)` per line), rendered via `Controls.Markdown` so the framework renders the anchors.
- **Showcase** — the owner's **curated pins** rendered as cards, **with a recent-public-content fallback** so the band is never empty.
- **Recent activity + items** — visibility-filtered so a visitor only ever sees public nodes.

Two new areas are registered on the User hub: `Profile` (`/{user}/Profile`, the read-only showcase + owner preview) and `EditProfile` (`/{user}/EditProfile`, the owner-only editor). Node-menu items ("View public profile", "Edit profile") are added for the owner, and an inline "Edit profile" button appears on the owner's profile.

## Curated pins + fallback model

The Showcase source is the User node's existing `PinnedPaths` — the same list the framework's **Pin/Unpin** surface (`PinLayoutArea`) already writes, so curation is the existing, proven, node-bound mechanism (nothing new to hand-roll). Rendering:

- **Pins present** → `Controls.MeshSearch` scoped `path:(a OR b …)`. In the owner's editor each card carries the inline **unpin** overlay (`PinnedThumbnailArea`) so pins are curated in place.
- **No pins** → fallback to the owner's recent public content (`namespace:{owner} …`), visibility-filtered — the section is never bare.

Design fork (noted per instructions): I **reused `PinnedPaths`** for the Showcase rather than adding a parallel `ShowcasePaths` field — it is already "a list of node paths on the User node's content," already owner-editable via the framework, and already visibility-filtered on render, so a visitor never sees a pinned node they cannot read. **Links** is the one new content field (`User.Links`, a markdown string).

## Getting-started card

On an **empty** profile (no bio, links, or pins) the profile renders a friendly getting-started card instead of empty sections. It explains how to add a bio, links, and pin content; for the owner it links straight to the editor and previews their recent work; for a visitor it is a gentle "nothing here yet" plus the owner's recent public work. This is **seeded for all users behaviourally** — there is nothing to store; the card is inherent in the render path and shows automatically, always, until the profile is filled in (a persistent self-documenting starter, not a one-time step).

## Owner editing (data-bound, no replica)

`EditProfile` binds **directly to the User node**: bio and links are node-bound `MarkdownEditorControl`s (`Value = JsonPointerReference("bio"/"links")`, node-bound `DataContext`), each edit a per-field read-modify-write straight to the node via `GetMeshNodeStream` — one source of truth, no `/data/{id}` replica, no save subscription. Gated on `Permission.Update` (visitors get access-denied).

## Compliance

- **Framework controls only** — composed from `Controls.Stack` / `Controls.Markdown` / `Controls.MeshSearch` / `Controls.Button` / `Controls.H3`/`Label` + the node-bound editor controls. **Zero** `Controls.Html` / string-HTML / `StringBuilder` in the new code (a test asserts no `HtmlControl` in the tree).
- **Data-bound, not hand-rolled** — reads/writes go through the node stream; no `/data` replica; no `.Take(1)` on the live binding.
- **PII (#471)** — email hidden for non-owner/non-admin viewers; fail-closed (`StartWith(false)` on the admin check).
- **No async** — everything is `IObservable<T>` + `.Subscribe`; immutable collections; no static state.

## Tests

- `UserProfileShowcaseTest` (Graph.Test, pure unit, 16 tests): owner-editable bio/links/showcase; visitor read-only **without email** (+ owner/admin sees it); empty profile → getting-started card; pins render in Showcase vs recent-public fallback; and the control tree is layout-area controls with **no** `HtmlControl`.
- `UserActivityAreaTest` (Monolith.Test, integration): the new `Profile` and `EditProfile` areas resolve end-to-end.

Build: `dotnet build -c Release -warnaserror -p:CIRun=true` clean (0 warnings) for Graph, Graph.Test, Monolith.Test. All tests green (16 + 24 unit, 9 integration).

### DIFF
--- a/src/MeshWeaver.Graph/UserActivityLayoutAreas.cs
+++ b/src/MeshWeaver.Graph/UserActivityLayoutAreas.cs
@@ -56,6 +56,18 @@ public static class UserActivityLayoutAreas
     /// <summary>Area that clears the owner's <see cref="User.Body"/> override so the default welcome home returns.</summary>
     public const string ResetHomeArea = "ResetHome";
 
+    /// <summary>
+    /// Area name for the public profile page (<c>/{user}/Profile</c>) — the polished, read-only
+    /// showcase every visitor sees, and the owner's preview + entry point to the editor.
+    /// </summary>
+    public const string ProfileArea = "Profile";
+
+    /// <summary>
+    /// Area name for the owner-only profile editor (bio, links, showcase) — node-bound editors that
+    /// auto-persist to the User node. Access-gated on <see cref="Permission.Update"/> (self-edit only).
+    /// </summary>
+    public const string EditProfileArea = "EditProfile";
+
     /// <summary>Link to the doc page that explains the configurable Body-page + <c>@@</c>-region model.</summary>
     internal const string ConfigGuideLink = "/Doc/GUI/ConfigurablePages";
 
@@ -98,7 +110,10 @@ public static class UserActivityLayoutAreas
             // Space Body editor. See EditHome / BuildHomeBodyEditor.
             .WithView(MeshNodeLayoutAreas.EditArea, EditHome)
             // Clears User.Body → the welcome template returns; reached from the Reset menu item.
-            .WithView(ResetHomeArea, ResetHome))
+            .WithView(ResetHomeArea, ResetHome)
+            // The polished public profile (read-only showcase) and its owner-only, node-bound editor.
+            .WithView(ProfileArea, ProfileAreaView)
+            .WithView(EditProfileArea, EditProfile))
             // Re-enable Edit on the user home (the default node menu HIDES generic Edit on a
             // protected partition root) and add a Reset-to-default item once the owner has
             // authored a Body override.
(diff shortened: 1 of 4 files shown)


rbuergi:
Synced with `main` (advanced ~30 commits). `main` independently added visitor-email redaction to the Activity area (also #471). Reconciled by folding main's superior `canSeeEmail` observable (handles anonymous/virtual/owner) into a shared `CanSeeEmailStream` helper used by both the Activity and Profile areas, and dropped my equivalent `ViewerIsAdmin`. main's `BuildVisitorProfile(…, showEmail)` is superseded by `BuildProfile`. Rebuilt Release/-warnaserror clean; 40 unit + 9 integration tests green on the merged tree.

PR #537, opened 2026-07-18
fix(gitsync): webhook config lookup runs as System

rbuergi:
Live-found rolling out #527 on memex-cloud: the push webhook delivered 200 but triggered no update.

**Root cause:** the webhook HTTP request is anonymous (its authorization is the verified HMAC signature), and `_GitSync` config satellites are not anonymous-readable — the processor's ambient-identity config query matched **nothing** on an access-gated portal. The DevLogin persistent circuit fallback masked this in tests (the same masking the activity-owner-credential tests pinned).

**Fix:** the shared config-node query (push-target matching AND issue-space matching) impersonates System — the identity model the webhook write path already uses. The push E2E test now clears every ambient identity before invoking the processor; revert-proven (fails on the unfixed query, passes fixed). GitSync suite 131/131; Release `-warnaserror` clean.

### DIFF
--- a/src/MeshWeaver.GitSync/GitHubWebhookProcessor.cs
+++ b/src/MeshWeaver.GitSync/GitHubWebhookProcessor.cs
@@ -234,9 +234,7 @@ public sealed class GitHubWebhookProcessor
     private IObservable<IReadOnlyList<PushTarget>> MatchingSyncTargets(string repoUrl, PushEvent push)
     {
         var target = ParseSafe(repoUrl);
-        return meshService
-            .Query<MeshNode>(MeshQueryRequest.FromQuery($"nodeType:{GitHubSyncService.ConfigNodeType}"))
-            .Take(1)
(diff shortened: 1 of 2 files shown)


rbuergi:
CI triage for the red run (29645134874) — both failures are foreign to this diff (GitSync-only change):

- **Shard 1**: `NodeTypeAssemblyLeakTest.NodeTypeAssemblyContext_IsCollected_AfterMeshDisposal` hit the 60s hard watchdog inside `CompileNodeTypeAndWeakRefContexts` (Roslyn compile starved under shard load; three concurrent createdumps ran on the runner during the window). Local run on this branch: **passes in 3s**. No overlap with the GitSync diff.
- **Shard 5**: harness-level only — `Failed to CreateArtifact: ETIMEDOUT` (artifact-upload network timeout); its Test Results job is green.

Re-running the failed jobs.

PR #559, opened 2026-07-19
Serialize user_effective_permissions rebuild with an advisory lock (40P01)

rbuergi:
## The defect (confirmed live on memex-cloud 2026-07-19, twice)

Concurrent mesh-node ACCESS writes from **different partitions** deadlock in Postgres with `40P01: deadlock detected`. Two plugin hubs warming concurrently each ran their access-seeding pass; each pass writes `_Access` grants into its own partition schema's `access` table, and each write's `access_changed` trigger runs the FULL `rebuild_user_effective_permissions()` **inside the originating write's transaction**. Each rebuild:

- takes ACCESS EXCLUSIVE on its **own schema's** `user_effective_permissions` (the atomic-swap `ALTER TABLE … RENAME` trio), and
- writes the **shared** `public.partition_access` rows (upsert + revoke-delete sync).

Two such transactions touch (a) different schemas' tables and (b) the same shared rows in different lock orders → lock cycle → PG kills one → the whole seeding pass of that hub aborts:

```
MeshNode Unknown at 'AgenticEngineering/Start/_Access/Public_Access': PostgresException: 40P01: deadlock detected
```

## The fix

The function's **first body statement** is now

```sql
PERFORM pg_advisory_xact_lock(hashtext('meshweaver_uep_rebuild'));
```

one **global** transaction-scoped advisory lock shared by every schema's rebuild. Concurrent rebuilds queue instead of interleaving into a cycle; the lock releases automatically at commit/rollback (an erroring rebuild can never leak it). No retry loops, no timeouts, no app-side locking.

While landing this, the rebuild-function DDL — which had drifted into two near-identical copies (`GetMeshSchemaScript` and `GetVersionedPartitionDdl`) — is single-sourced in the new `PostgreSqlSchemaInitializer.GetUepRebuildFunctionScript(schemaRef)`. Both schema scripts and (through `GetVersionedPartitionDdl`) the `public.ensure_partition_schema` proc embed it, so every installer ships one identical body.

## How EXISTING databases get the updated function

The schema script only `CREATE OR REPLACE`s the function for the **boot schema** (`InitializeAsync` → `GetSchemaScript`) and for partitions that go through `public.ensure_partition_schema` again (new provisioning). Already-provisioned partition schemas keep their deployed body — so the mechanism for live DBs is the new **`V47_SerializeUepRebuildAdvisoryLock` migration**: it probes `pg_proc` for every schema that has `rebuild_user_effective_permissions` and re-`CREATE OR REPLACE`s it there (same OID — the existing `access_changed` trigger picks the locked body up immediately). The memex-family deployments (memex / memex-cloud / atioz) run `Memex.Database.Migration` on every deploy, so V47 lands on the next roll-out. Fresh DBs skip repairs (`MigrationRunner` fast-forwards) and get the locked body straight from the initializer templates. Snowflake is untouched — it has no SQL rebuild function (C#-side projection).

## Tests

- `test/MeshWeaver.Hosting.PostgreSql.Test` (Testcontainers, full project): **553 passed, 0 failed, 1 skipped** (pre-existing skip).
- New `ConcurrentUepRebuildTests.ConcurrentRebuilds_AcrossTwoPartitions_SerializeInsteadOf40P01`: provisions two prod-shape partitions via `ensure_partition_schema`, seeds a Space + grant in each through the real trigger path, **pins that both provisioned function bodies carry the advisory lock** (deterministic-red if it is ever removed), then fires 8 interleaved full rebuilds across the two schemas and asserts all succeed and `public.partition_access` converges. With the lock this is deterministic-green; without it, the interleave is exactly the prod 40P01 cycle.

`dotnet build -c Release -warnaserror` is clean for `MeshWeaver.Hosting.PostgreSql`, `Memex.Database.Migration`, `Memex.Portal.Distributed`, and the PG/Orleans/indexing test projects. (Note: `tools/MeshWeaver.Tools.PostgreSqlImport` does not compile, but that is pre-existing — it references `StorageImporter`/`ImportHelper` which no longer exist anywhere, and it is in neither the solution nor CI.)

### DIFF
--- a/memex/aspire/Memex.Database.Migration/MigrationRegistry.cs
+++ b/memex/aspire/Memex.Database.Migration/MigrationRegistry.cs
@@ -55,6 +55,7 @@ public static class MigrationRegistry
         new V44_FixMeshNodeHistoryTriggerPerSchema(),
         new V45_AddNodeAuthorshipColumns(),
         new V46_AddExcludeFromContextColumn(),
+        new V47_SerializeUepRebuildAdvisoryLock(),
     ];
 
     /// <summary>
--- a/memex/aspire/Memex.Database.Migration/Migrations/V47_SerializeUepRebuildAdvisoryLock.cs
+++ b/memex/aspire/Memex.Database.Migration/Migrations/V47_SerializeUepRebuildAdvisoryLock.cs
@@ -1,0 +1,87 @@
+using MeshWeaver.Hosting.PostgreSql;
+using Microsoft.Extensions.Logging;
+
+namespace Memex.Database.Migration.Migrations;
+
+/// <summary>
+/// Re-applies <c>rebuild_user_effective_permissions()</c> on every schema that has it so the
+/// function takes the global transaction-scoped advisory lock
+/// (<c>pg_advisory_xact_lock(hashtext('meshweaver_uep_rebuild'))</c>) as its FIRST statement.
+///
+/// <para><b>The bug (memex-cloud 2026-07-19, twice).</b> Concurrent <c>_Access</c> writes from
+/// DIFFERENT partitions deadlocked with <c>40P01 deadlock detected</c>: two plugin hubs warming
+/// concurrently each ran their access-seeding pass; each grant write fired that schema's
+/// <c>access_changed</c> trigger, which runs the FULL <c>rebuild_user_effective_permissions()</c>
+/// inside the originating write's transaction. Each rebuild takes ACCESS EXCLUSIVE on its OWN
+/// schema's <c>user_effective_permissions</c> (the atomic-swap <c>ALTER TABLE … RENAME</c>) AND
+/// writes the SHARED <c>public.partition_access</c> rows — two transactions touching different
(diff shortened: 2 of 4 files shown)


rbuergi:
CI triage: shard 5 failed on MeshWeaver.Content.Test.SourceDocumentDataLoadingTest.NodeType_LoadsDataFromSourceDocuments ("Microsoft 2026 should load PropertyRisks from JSON") — the cold-start PingRequest at SourceDocumentDataLoadingTest.cs:113 missed the default 10 s Emit budget under shard load. MeshWeaver.Content.Test has no reference to MeshWeaver.Hosting.PostgreSql (this PR touches only the PG schema DDL, a memex migration, and a PG Testcontainers test), so there is no dependency edge from this diff to that test. Same cold-start timing class the test's own remarks document (method timeout was previously raised to 30 s for exactly this). Re-running the failed shard; the underlying cold-start activation budget in Content.Test deserves its own root-cause pass, tracked separately from this PR.

PR #589, opened 2026-07-21
deploy(aks): fix portal probe timeout (1s→15s) + raise memory limit (16→28Gi)

rbuergi:
Two memex-cloud production-stability fixes, root-caused live on 2026-07-21 (repeated crash-loops — "TDD page won't load, then crashes").

**1. Probe timeout — the crash-loop cause.** `deploy/aks/manifests/portal-ha-patch.yaml` re-declared readiness/liveness probes on `/` but **omitted `timeoutSeconds`**, so k8s applied the default **1s**. The portal's `/` renders the full Blazor home (~1s normally, more under GC/partition load) → a normal-but-slow response failed the probe → 6 failures killed the pod (exit 139) → crash-loop. It also silently overrode the chart's sensible `/health`+`/alive` probes. Now `timeoutSeconds: 15`, liveness `failureThreshold: 10` (≈300s before a kill).

**2. Memory 16Gi → 28Gi.** .NET server-GC targets ~75% of the container limit → 16Gi put the GC ceiling at ~12GB, exactly the live-set of the three big courses → GC thrash + Orleans health stalls. 32 GiB nodes fit 28Gi + system pods. Updated in both the HA patch and `values.aks.yaml resources.portal` (kept in sync).

Both **already applied live** via `kubectl` and verified (fresh pod stable; `/`=0.4s, TDD=1.1s — the 1.1s that used to trip the 1s probe). This PR lands them durably so the next deploy doesn't revert them.

### DIFF
--- a/deploy/aks/manifests/portal-ha-patch.yaml
+++ b/deploy/aks/manifests/portal-ha-patch.yaml
@@ -51,20 +51,28 @@ spec:
               app.kubernetes.io/component: memex-portal
       containers:
         - name: memex-portal
+          # timeoutSeconds is CRITICAL: the portal's `/` renders the full Blazor home and
+          # legitimately takes ~1s (more under GC pressure or partition load). Omitting it lets
+          # k8s apply the DEFAULT timeoutSeconds: 1 — so a normal-but-slow response failed the
+          # probe, 6 failures killed the pod (exit 139), and it crash-looped (memex-cloud,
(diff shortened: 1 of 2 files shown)


rbuergi:
Superseded by #601 (chart-level probe fix + crash dumps + affinity) — this legacy portal-ha-patch.yaml change would re-apply the heavy GET / probes that caused the unready→HPA-blind death spiral. Closing to prevent an accidental regression.

PR #592, opened 2026-07-21
Portal probes: /alive+/health with 5s timeouts, memex-cloud to 2 replicas

rbuergi:
## The outage (2026-07-21, memex.meshweaver.cloud)

Repeated ~2-minute full outages during a live demo. Root cause was a **probe death spiral**, not a crash:

1. Both k8s probes hit `/` — the full Blazor route served **through the message-hub pool** — with the default **1s timeout**.
2. Any CPU-saturation episode (hub wedge, compile storm, hot render) made `/` miss the 1s window; after 6×20s k8s **killed an alive-but-busy pod**.
3. The restart cold-started caches and live recompiles → more CPU → next kill.
4. memex-cloud's `portal-patch.json` pinned **`replicas: 1`** (overriding the chart's HA default) and the KEDA scaler was paused at 1 — so every kill was a full outage with nothing to fail over to. (An unready sole pod also blinds the HPA's metrics, so autoscaling can't rescue it either.)

## The fix

- **`deploy/aks/envs/memex-cloud/portal-patch.json`**: replicas 1 → 2; hostname `topologySpreadConstraint` (replicas on different silos nodes); readiness → `/health`, liveness → `/alive` (matching the Helm chart's `deployment.yaml`, which was already correct — only the AKS patches had diverged); `timeoutSeconds: 5`.
- **`deploy/aks/manifests/portal-ha-patch.yaml`**: same probe endpoints + timeouts, rationale inline.

`/alive` is the constant hub-pool-independent health check (`ServiceDefaults.cs`) — a wedged hub pool can no longer false-kill the pod; `/health` readiness pulls a struggling pod out of rotation while its peer serves.

## Already live

Applied to memex-cloud via kubectl: 2 pods across 2 nodes, KEDA unpaused min=2/max=4, silos nodepool min=2/max=4. This PR makes the next deploy stop reverting it.

https://claude.ai/code/session_01NGgWKSgL8pgrbg86xqfa6o

### DIFF
--- a/deploy/aks/envs/memex-cloud/portal-patch.json
+++ b/deploy/aks/envs/memex-cloud/portal-patch.json
@@ -1,6 +1,9 @@
 [
-  { "op": "replace", "path": "/spec/replicas", "value": 1 },
+  { "op": "replace", "path": "/spec/replicas", "value": 2 },
   { "op": "add", "path": "/spec/template/spec/nodeSelector", "value": { "workload": "silos" } },
+  { "op": "add", "path": "/spec/template/spec/topologySpreadConstraints",
+    "value": [ { "maxSkew": 1, "topologyKey": "kubernetes.io/hostname", "whenUnsatisfiable": "DoNotSchedule",
+                 "labelSelector": { "matchLabels": { "app.kubernetes.io/component": "memex-portal" } } } ] },
 
   { "op": "replace", "path": "/spec/template/spec/volumes/0",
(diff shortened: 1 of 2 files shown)


rbuergi:
Superseded by #601, which makes the /alive readiness + probe timeouts permanent in the helm chart (the live cluster already runs these via kubectl patch). Closing in favor of #601.

PR #643, opened 2026-07-25
fix(mcp): replica-safe stateful MCP sessions across silos (ISessionMigrationHandler)

rbuergi:
## Problem
Stateful MCP Streamable-HTTP sessions are pinned to the silo that served `initialize`. With >1 silo, a follow-up request for an established `Mcp-Session-Id` can land on a silo that never saw the session → the SDK rejects it **404 "Session not found"** (the "works on one silo, breaks on two" bug). The MCP client carries **no affinity cookie** (unlike the Blazor browser), so ACA/nginx sticky sessions can't pin it. Same class of multi-replica bug `OAuthCodeStore` fixed for the OAuth code exchange (2026-07-23 outage).

## Fix — mirrors `OAuthCodeStore` exactly
- **`McpSessionStore`** persists each session's `InitializeRequestParams` as an `Admin/McpSession/{hashPrefix}` mesh node (System identity, PG-persisted, shared by every replica), read back with the **same resilient `GetMeshNodeStream` poll** `OAuthCodeStore` uses at `/token` to ride out the create→routable lag across silos. Raw session id never stored (SHA-256, same scheme as ApiToken/OAuthCode).
- **`McpSessionMigrationHandler : ISessionMigrationHandler`** re-hydrates the session on whichever replica receives the request — **owner-checked** (only the original authenticated caller may re-bind it).
- **Injection:** `AddMeshMcp` calls `WithHttpTransport((Action<HttpServerTransportOptions>)null)` (decompiled to confirm), so the SDK resolves `ISessionMigrationHandler` from DI. Registering it in `ConfigureMemexServices` next to `OAuthCodeStore` is the entire wiring. Ships with the platform (`memex-portal-ai`) image — not a plugin, no package bump; Memex (deployments) just runs the new image.

## Verified — multiple replicas/silos
`test/Memex.Portal.Shared.Test/McpSessionStoreTest.cs` (7/7 green):
- `TwoStoreInstances_StoreOnOne_ReadOnOther` — initialize on replica A, follow-up on replica B, sharing one mesh (the exact multi-replica shape, mirroring `OAuthCodeStoreTest.TwoStoreInstances_GenerateOnOne_ExchangeOnOther`).
- `Handler_StoreOnOne_MigrateOnOther_RoundTrips` — end-to-end through the SDK handler interface; `Handler_RejectsMigration_ForDifferentOwner`; unknown/expired/round-trip.

(Also validated end-to-end earlier against the real `ModelContextProtocol` 1.2.0 SDK with 3 in-process instances: a session created on one served by another, the no-handler control 404'd.)

### DIFF
--- a/memex/Memex.Portal.Shared/Authentication/McpSessionMigrationHandler.cs
+++ b/memex/Memex.Portal.Shared/Authentication/McpSessionMigrationHandler.cs
@@ -1,0 +1,104 @@
+using System.Reactive.Threading.Tasks;
+using System.Security.Claims;
+using System.Text.Json;
+using Microsoft.AspNetCore.Http;
+using Microsoft.Extensions.Logging;
+using ModelContextProtocol;
+using ModelContextProtocol.AspNetCore;
+using ModelContextProtocol.Protocol;
+
+namespace Memex.Portal.Shared.Authentication;
+
+/// <summary>
+/// Makes stateful MCP sessions pod-independent across horizontally-scaled portal replicas, the
+/// same way <see cref="OAuthCodeStore"/> made the OAuth code exchange replica-safe.
+///
+/// <para>
+/// The MCP Streamable-HTTP transport keeps per-session state in the process that served the
+/// <c>initialize</c> request. With more than one silo the MCP client — which carries no affinity
+/// cookie (unlike the Blazor browser) — can send a follow-up request for an established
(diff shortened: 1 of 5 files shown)


rbuergi:
Closing — not the approach we want. No changes merged; MeshWeaver main untouched.

PR #664, opened 2026-07-27
MeshDataSource: share ONE own-node stream per hub, not one per call

rbuergi:
Chasing the `InboxToolIntegrationTest` flake (2 failures / 8 runs on clean `main`) found a **product defect in core data plumbing** — not a test defect.

## The defect

For an **own-hub** node, `GetMeshNodeStream()` minted a **brand-new reduced stream on every call**: `MeshNodeReference` reduced `primary → InstanceCollection → MeshNode` per call, uncached — unlike the **cross-hub branch of the same reducer**, which shares one stream per path via `Workspace._remoteStreamCache`.

Every `SynchronizationStream` pushes values through its **own hosted `sync/<id>` hub** (`OnNext → Hub.Post(SetCurrentRequest)`). So two readers of the same hub's own node received the **same commit on two independent action blocks, in arbitrary relative order**. Measured: **25–37% inversions** over 40 commits.

The code already *claimed* this property — the reducer's own comment says reducing from the primary "keeps both the watcher and any other own-MeshNodeReference subscriber pinned to the same stream". Reducing from the primary is necessary but not sufficient: the `Reduce` call itself mints the new stream.

## Why it matters beyond the test

It broke every **"react on the own node, then act"** component. In the AI thread, the submission watcher's `ThreadInboxChannel.OfferFromNode` could still be un-run while another own-node reader had already seen a follow-up land — so `check_inbox` answered `"(no new messages)"` for a message demonstrably committed on the node. That is why the test fails in ~400 ms on an assertion mismatch rather than a timeout.

In production terms: **a user's mid-round steering message is provably on the thread node yet invisible to the agent for an unbounded, scheduler-dependent interval.** Not lost — the round-end fold re-dispatches unconsumed pending as a fresh round (verified) — but degraded from "folded into the current answer" to "arrives as a separate round".

Each own-node read also **leaked** a reduced chain plus two hosted `sync/` hubs that nothing ever disposed. There are **157** such call sites in `src/`.

## The fix

Build the chain once per `(workspace, path)` and serve it from a per-hub cache on the **existing hub-scoped `OwnNodeCache` DI singleton** — so its lifetime is the hub's, not the process's, and not a closure over a possibly-shared hub *configuration* (which would retain workspaces). A caller passing an explicit `StreamConfiguration` still gets a bespoke stream and doesn't poison the cache; nothing in the tree does that today.

Fan-out then becomes a single synchronous `ReplaySubject` dispatch **in subscription order** — the same coherence the cross-hub branch already got for free.

## Deterministic proof

`test/MeshWeaver.Hosting.Monolith.Test/OwnMeshNodeStreamCoherenceTest.cs` (new):

| test | before | after |
|---|---|---|
| `OwnNodeStream_IsSharedAcrossCallers` (`ReferenceEquals`) | deterministically **False** | **True** |
| `TwoOwnStreamReaders_ObserveEveryCommitInSubscriptionOrder` (40 commits, 2 readers, counts inversions; waits on a condition, no sleeps) | 15/40, 10/40, 15/40 inversions — **3/3 fail** | **0 inversions, 3/3 pass** |

`InboxToolIntegrationTest.CheckInbox_TwoFollowUps` had a poll-and-accumulate loop whose own comment documented this race ("can lag the test's own-stream read by a beat"). With the root cause fixed that's a band-aid, so it's now a single `check_inbox` call — a stronger assertion and a second guard on the fix.

| | before | after |
|---|---|---|
| `InboxToolIntegrationTest` class | 1 failure / 5 runs | **22/22 clean** (286 test executions) |

Full suites on `0fcc366f2`: **AI 889/0, Monolith 400/0, Data 287/0, Graph 627/0, Layout 337/0**; full solution `-c Release -p:CIRun=true -warnaserror` clean.

## Notes

- **A measurement contaminant worth recording**: enabling `MeshWeaver.AI: Debug` in the test's `bin/appsettings.json` *masks* the flake — it slows the test's own chain so the watcher wins the race. The before/after batches here were re-run with the committed config.
- **Not fixed, follow-up**: `SynchronizationStream.OnNext` posting through a per-stream hosted hub means **any** two distinct reduced streams over the same source are mutually unordered. This fixes the own-`MeshNode` case; other reducers that mint per-call streams have the same latent property.
- **Deliberately not done**: no "read the node as a fallback" path in `check_inbox`. That would be the band-aid; the ordering guarantee is now real.

No What's New entry — the user-visible symptom (a steering message arriving as a separate round) is a latent race, not a documented behaviour change.


rbuergi:
## Do not merge as-is — this change causes the `MeshHubDisposalLeakTest` failure, confirmed on Linux

I initially refuted the leak-test correlation with 22 local executions showing zero failures on both this branch and pristine `main`. **That evidence was worthless, and the reason is its own bug**: `MeshHubDisposalLeakTest` **cannot fail on macOS**. ClrMD's `DataTarget.CreateSnapshotAndAttach` throws `PlatformNotSupportedException`, the analyzer swallows it, `staticRooted` stays `false`, and `staticRooted.Should().BeFalse()` passes unconditionally — while an instrumented probe reports `meshHubAlive=True` on both trees. Filed as a separate test-infrastructure defect (see #673's description).

Re-run on Linux under Docker, full `Hosting.Monolith.Test` assembly in one process (CI's shape):

| build | leak-test failures |
|---|---|
| pristine `main` `0fcc366f2` | **0 / 3** |
| this branch `f1454708d` | **3 / 3** |

### Mechanism

```
ROOT[StrongHandle] → TimerQueue → TimerQueueTimer
  → System.Reactive…ConcurrencyAbstractionLayerImpl+PeriodicTimer
  → System.Reactive.Linq.ObservableImpl.Sample<MeshWeaver.Mesh.MeshNode>+_
  → AnonymousSafeObserver<MeshNode> → … → MessageHub
```

**Sharing one own-node stream also shares its subscribers' lifetime.** The persistence sampler subscribes `GetMeshNodeStream().Sample(…)`, and Rx's `Sample` is driven by a `PeriodicTimer` rooted in the process-wide `TimerQueue`. Before this change that subscription sat on a stream minted for it alone, which the teardown cascade completed — ending the sampler and its timer. Once the stream is shared, nothing completes it at the sampler's teardown point, so the periodic timer outlives disposal and pins the whole mesh graph.

Note the sequel: **disposing the cached streams does not fix it either** — still 3/3. The retention lives in the *subscription*, not the dictionary, so this is not a lifetime tweak on the cache.

### Where that leaves the change

The coherence problem it fixes is real and worth fixing — own-node readers observing the same commit on independent `sync/` hubs in arbitrary order, 25–37% inversions measured, which is what makes `check_inbox` answer "(no new messages)" for a message demonstrably on the node. But as written it trades a 25% test flake for a genuine GC retention of a disposed mesh.

It needs a different shape — e.g. keep per-subscriber streams and achieve watcher ordering another way, or have the shared stream complete its subscribers deterministically at hub teardown. Converting to draft rather than closing, since the analysis and the deterministic coherence tests are worth keeping.

rbuergi:
Closing rather than parking: this change is a proven regression and should not sit in the queue where it might get merged.

On Linux (where `MeshHubDisposalLeakTest` actually works — it cannot fail on macOS, see #674): pristine `main` **0/3** leak-test failures, this branch **3/3**. Sharing one own-node stream also shares its subscribers' lifetime — the persistence sampler's Rx `Sample` runs on a `PeriodicTimer` rooted in the process-wide `TimerQueue`, and nothing completes the shared stream at that subscriber's teardown, so the timer outlives disposal and pins the whole mesh graph. Disposing the cached streams does not fix it either.

The coherence bug it targets is real and still open: own-node readers observe the same commit on independent `sync/` hubs in arbitrary order (25–37% inversions measured), which is what makes `check_inbox` answer "(no new messages)" for a message demonstrably on the node. It needs a different shape — keep per-subscriber streams and get ordering elsewhere, or have the shared stream complete its subscribers deterministically at teardown.

The analysis, the mechanism, the deterministic coherence tests (`ReferenceEquals` on two `GetStream` calls; inversion counting over 40 commits) and the lifetime work all remain readable on this PR and its commits for whoever picks it up.

PR #682, opened 2026-07-27
Compile: stop source discovery idling 90s on a missed Initial, and pin recursive compilation

rbuergi:
## The profile

Straight from the `Store/Plugin` compile activity node that preceded today's outage:

```
20:46:02.486  Invoking compiler…
20:47:32.673  Source query: … (all nine, one timestamp)
20:47:35.288  Compiled assembly written
```

| phase | time |
|---|---|
| source discovery | **90.19 s** |
| Roslyn compiling 30 files | **2.6 s** |
| write + release | 0.2 s |

**97% of the compile was waiting, not compiling.** And 90.19 s is `2 × SyncStreamOptions.HeartbeatInterval` (45 s).

## Root cause — a lost wakeup, not a lock

When the cached synced source query's subscription misses its `Initial`, **nothing re-requests it**. The read makes no progress until the sync stream's periodic heartbeat re-delivers content. Miss it twice and a 3-second compile takes a minute and a half — which is exactly how `Store/Plugin` blew the 60 s settle window and left every plugin root serving the *"did not settle"* fallback.

This failure mode has bitten before: one screen above the fix, an existing comment documents the identity-loss variant — *"the source reads run unauthenticated and stall until the next 45s sync-stream heartbeat — the ~42s freeze."* Same signature, different trigger.

## The fix

Race the cached query against a **direct, uncached** mesh read of the *same* expanded queries:

- the probe is **subscription-delayed (3 s)**, so a healthy compile — cache hit or one cold storage read — never issues it and pays nothing;
- a stalled compile answers in about a read instead of 45 or 90 seconds;
- `Amb` takes whichever speaks first and drops the other;
- when the probe wins it logs at **Warning** — "compiles are slow" was invisible until someone read microsecond timestamps out of an activity node.

The probe **accumulates chunks** rather than taking the first emission: a query's `Initial` arrives chunked, and a partial source set compiles **wrong** — the missing file surfaces as a phantom Roslyn error pointing at code that is perfectly fine.

**Fully reactive**: `Defer` / `CombineLatest` / `Scan` / `Throttle` / `Take` / `Select` / `Catch`. No `await`, no `ToTask`, no blocking bridge.

## Recursive compilation tests

The incident ran through the **cross-type** path: `Store/Plugin` declares `shared=@Store/Coupon/Source`, `shared=@Store/Order/Source`, `shared=@Store/BillingProfile/Source` — which is what turns discovery into nine live queries. `RecursiveCompilationTest` pins both halves:

- **Expansion** — a shared reference yields both the folder node *and* its subtree, for every referenced type; deterministic; no self-recursion; a non-shared type does not leak in; matching agrees with expansion (the compiler and the Sources menu use the same resolver).
- **Chunk accumulation** — chunked `Initial` accumulates instead of truncating, re-delivery is idempotent, removal drops, empty/null changes are inert.

Worth stating plainly: **these pass on `main`.** The recursion is correctly *specified* — what failed is the *execution* of those queries stalling. The tests are there so a regression in either half is caught, and so the next person profiling a slow compile starts from a pinned contract.

## Testing

**644/644** Graph tests pass.

## Follow-ups not in this PR

- Why the `Initial` is missed in the first place (cold subscription racing the source-update burst) — this makes it survivable, it does not remove the race.
- Per-stage compile timing in the activity log, so this is measurable without microsecond arithmetic.
- A readiness gate / deploy-time precompile — worth revisiting *after* this, on real numbers: a 3 s compile changes that calculus completely.

### DIFF
--- a/src/MeshWeaver.Graph/Configuration/MeshNodeCompilationService.cs
+++ b/src/MeshWeaver.Graph/Configuration/MeshNodeCompilationService.cs
@@ -562,14 +562,20 @@ internal class MeshNodeCompilationService(
                         .Take(1)
                         .Select(map => (IReadOnlyCollection<MeshNode>)map.Values.ToList())
                         .Catch((Exception _) => Observable.Return<IReadOnlyCollection<MeshNode>>([]))))
-                .Select(results =>
+                .Select(results => results
+                    .SelectMany(r => r ?? [])
+                    .Where(n => !string.IsNullOrEmpty(n.Path))
+                    .GroupBy(n => n.Path)
+                    .Select(g => g.First())
+                    .ToList())
+                // 🚨 NEVER win the race with an EMPTY set. Every per-query leg degrades to empty on
+                // error, so a probe that hit a cold partition or a permission wall would otherwise
+                // emit "no sources", beat the cached query, and compile the type against NOTHING —
+                // strictly worse than the stall it exists to dodge. No sources found ⇒ stay silent
+                // and let the cached query (and the outer Timeout) decide.
+                .Where(merged => merged.Count > 0)
+                .Select(merged =>
                 {
-                    var merged = results
-                        .SelectMany(r => r ?? [])
-                        .Where(n => !string.IsNullOrEmpty(n.Path))
-                        .GroupBy(n => n.Path)
-                        .Select(g => g.First())
-                        .ToList();
                     logger.LogWarning(
                         "Source discovery for '{SelfPath}' fell back to the UNCACHED probe after {Delay}s — "
                         + "the cached synced query stalled (a missed Initial idles until the {Heartbeat}s "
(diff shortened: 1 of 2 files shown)


rbuergi:
## Reproduced on a warm, healthy mesh — one heartbeat, exactly

Triggered a compile of `Store/Coupon` on memex just now (site healthy, nothing wedged, cache warm). Its activity log:

```
21:35:47.712  Compile started for Store/Coupon
21:35:47.712  Invoking compiler…
21:36:32.909  Source query: … (all nine, one timestamp)
21:36:35.736  Compiled assembly written
21:36:35.972  end
```

| phase | time |
|---|---|
| source discovery | **45.20 s** |
| Roslyn compiling 23 files | **2.83 s** |

**45.20 s is one `HeartbeatInterval` (45 s), to within 200 ms.** The outage compile was 90.19 s — two.

This settles two things:

1. **Not a cold-start artifact.** A warm, healthy mesh pays a full heartbeat on *every* compile. The outage wasn't a freak event; it was this constant plus a 60 s settle window, which any type with enough sources will eventually cross.
2. **It's the cross-type path.** `Store/Coupon` matched **23 Code nodes across four types** (`Store/Coupon`, `Store/Plugin`, `Store/Order`, `Store/BillingProfile`) via `shared=` — nine expanded queries, which is why `RecursiveCompilationTest` in this PR pins that expansion.

Expected after this change: the probe answers at ~3 s (its subscription delay) instead of 45 s, so the same compile lands in about 6 s wall-clock — and the Warning line says the probe won, making the underlying lost-wakeup visible instead of silently costing 45 s a time.

I'll post the same measurement taken after deploy.

PR #684, opened 2026-07-27
Sync: start the heartbeat early — the 45s stall behind the 2026-07-27 outage

rbuergi:
## This is the actual root cause

My two earlier fixes ([#681](https://github.com/Systemorph/MeshWeaver/pull/681), [#682](https://github.com/Systemorph/MeshWeaver/pull/682)) treated it one level too high, and **neither worked in production**: after deploying `ci.1459` the uncached probe never fired (zero log lines) and the self-heal never healed — four plugin roots went to the fallback overlay again and only a scale-to-zero restored the site. Both sit on top of the layer that is actually broken.

## The gap

Every recovery path in `CreateExternalClient` is **event-driven**: the change feed announces an owner restart, the latch re-subscribes. **Nothing** recovers a stream whose owner ACKED the `SubscribeRequest` and then never delivered the first `DataChangedEvent`. No announce ever comes, so the latch never trips and the read just waits — in practice until the 45 s heartbeat shakes something loose.

The existing comment says so outright:

> *"This is not a retry/watchdog: nothing re-posts on a timer — recovery stays event-driven."*

## The measurement

From the compile activity log's microsecond timestamps on memex:

| | source discovery | Roslyn |
|---|---|---|
| healthy mesh | **45.20 s** | 2.83 s |
| during the outage | **90.19 s** | 2.6 s |

**45.20 s is one `HeartbeatInterval` to within 200 ms.** 97% of every compile was a lost `Initial` waiting for a heartbeat. Types then crossed the 60 s settle window and every plugin root served the *"did not settle"* fallback.

The same silence keeps an instance's NodeType stream from ever seeing a fresh build — which is precisely why the overlay self-heal had nothing to fire on, and why only a process restart cleared it. **One bug, three symptoms.**

## The fix

If nothing has arrived shortly after subscribing, re-send the `SubscribeRequest` through the **same proven `Resubscribe` path**. Bounded: 3 attempts, 4 s apart. A missed-wakeup nudge, not a retry loop —

- an owner that genuinely cannot answer already faults the stream through the subscribe's own error path;
- one that is merely slow satisfies the probe with its first delivery;
- `Take()` bounds the timer, so it can never become the process-global TimerQueue root that the heartbeat comment in this same file documents (the `MeshHub_IsCollected` leak).

## The load-bearing test is the negative one

`HealthyStream_ReceivesItsData_AndNeverResubscribes` — a stream that gets its snapshot must **never** re-subscribe. A watchdog that fires anyway re-creates a `sync/{ClientId}` hub **per stream** on the owner's single-threaded action block: the exact storm `ChangeFeedResubscribeCoalesceTest` exists to prevent, and considerably worse than the bug being fixed. Plus a budget test pinning that the probe sits far below the heartbeat and the retry count stays bounded.

**289/289** Data tests pass.

## Honest scope

This removes the 45 s stall. It does **not** remove the need for deploy-time precompile: a new image invalidates every assembly, so types still recompile on first activation — just in ~3 s instead of ~48 s, which is the difference between invisible and an outage.


rbuergi:
## Draft — stopping here deliberately, with the diagnosis intact

Fourth attempt at this tonight. The site is healthy (`ci.1459`, 16/16 roots verified) and **nothing from this branch is deployed**. A fifth speculative change to code every sync stream in the mesh depends on, at 01:00, after three wrong fixes, is worse than a documented open bug. Handing over instead.

### What is PROVEN (measured, not inferred)

From the compile activity node's microsecond timestamps on memex:

| | source discovery | Roslyn |
|---|---|---|
| healthy mesh | **45.20 s** | 2.83 s |
| during the outage | **90.19 s** | 2.6 s |

`45.20 s` = one `SyncStreamOptions.HeartbeatInterval` to within 200 ms; `90.19 s` = two. **97% of every compile is a lost `Initial` waiting for a heartbeat.** Types then cross the 60 s settle window → every plugin root serves the "did not settle" fallback. The same silence stops an instance's NodeType stream from seeing a fresh build, which is why the overlay self-heal has nothing to fire on and only a process restart clears it. One bug, three symptoms.

Also proven: `RecursiveCompilationTest` (merged in #682) passes on `main` — the `shared=@Other/Source` expansion is **correct**. The recursion is not mis-specified; its *execution* stalls.

### What this branch does, and what it got wrong

Re-sends the `SubscribeRequest` through the existing `Resubscribe` path when nothing arrives after a probe. Two mistakes found by CI, both fixed here, both worth recording:

1. **Probe at 4 s fired on healthy streams** on loaded runners — a genuine cold read outran it. Moved to 15 s. Wrong direction is dangerous: under mass cold start (every deploy) an eager probe re-subscribes healthy streams and each re-subscribe creates a `sync/{ClientId}` hub on the owner's single action block — the storm `ChangeFeedResubscribeCoalesceTest` exists to prevent.
2. **It added its own `reduced.Subscribe`** to notice the first delivery. An extra early subscriber changes *when* a ref-counted stream starts, so unrelated tests broke across shards 0 and 4 (`ExportImportAccessControlTest`, `TodoGraphIntegrationTest` — both green on the same code an hour earlier). The flag now piggybacks the existing passive tracker. Those two shards are green again.

### The remaining failure, and why it may be a design problem

`MissingInitialResubscribeTest.HealthyStream_ReceivesItsData_AndNeverResubscribes` still fails on CI while passing locally (Data 289/289, plus the two formerly-broken classes 5/5 and 7/7).

The uncomfortable reading: if a *healthy* CI stream still hasn't delivered after 15 s, then **elapsed time is a poor discriminator between "lost wakeup" and "slow owner"** — and any purely time-based watchdog will either fire on healthy streams or arrive too late to matter. If so, the right fix is at the protocol level (an explicit ack/empty-snapshot response, so silence is unambiguous) rather than a timer.

That is the question the next person should answer before tuning this number again.

### Left open

- **`main` is red independently of this**: shard-3/shard-1 host crashes (`a test host exited non-zero`), which is why CD skipped and `ci.1451` sat on the cluster for hours. That blocks all delivery and needs its own fix.
- Deploy-time precompile is still the structural answer (a new image invalidates every assembly); it's cheap only once this stall is gone.

rbuergi:
## Rewritten: one line, and it is the actual fix

The re-subscribe watchdog is gone. Net effect on the file is **-34 lines**, and the functional change is one line:

```diff
- Observable.Interval(heartbeatInterval)
+ Observable.Timer(FirstHeartbeat, heartbeatInterval)
```

### Why this, and not the probe

The tell was always **where** the data arrives: exactly on a heartbeat. 45.20 s of source discovery on a healthy mesh, 90.19 s during the outage, against 2.83 s of Roslyn — 45.20 s is one `HeartbeatInterval` to within 200 ms. The owner isn't slow; it holds the delivery until something pokes it, and every other recovery path here is event-driven, so a stream whose owner acked and then went quiet produces **no event at all** and nothing runs until the first heartbeat 45 s later.

So poke sooner. First tick at 5 s, cadence unchanged. Same fire-and-forget `[SystemMessage]` the stream already sends forever — one extra cheap message per stream, no new subscription, no new protocol, no state machine.

**The probe was the wrong shape and CI proved it twice:** at 4 s it fired on healthy streams under load, and its extra `reduced.Subscribe` changed *when* ref-counted streams start, breaking `ExportImportAccessControlTest` and `TodoGraphIntegrationTest` in unrelated shards. Each `SubscribeRequest` creates a `sync/{ClientId}` hub on the owner's single-threaded action block, so a probe becomes a storm under mass cold start — i.e. every deploy — which is exactly what `ChangeFeedResubscribeCoalesceTest` guards. A heartbeat creates nothing.

### The test is a deterministic red, not a timing race

With the production 45 s cadence configured, a heartbeat inside the test window can **only** come from the early first tick. Verified by reverting the one line — `FirstHeartbeat_ArrivesLongBeforeAFullInterval` **fails** — then restoring it.

| | |
|---|---|
| `Data.Test` | **290/290** |
| `ExportImportAccessControlTest` | **5/5** (broken by the previous attempt) |
| `TodoGraphIntegrationTest` | **7/7** (broken by the previous attempt) |

Plus a storm guard (subscribe count never grows) and a bounds check on the constant.

rbuergi:
## ⚠️ Post-deploy measurement: this did NOT fix the stall

Deployed as `3.0.0-ci.1467` (commit `aeb2bf0`) and measured the same compile, same method as the baseline in this PR.

| | total compile |
|---|---|
| pre-fix baseline (warm system, 21:35Z) | **48 s** |
| post-fix (settled system, 02:50Z) | **57 s** |

**No improvement.** A compile attempted a few minutes after the rollout also failed outright with the original symptom — `No response received in hub cache/… within 00:01:00 for request SubscribeRequest → target Store/Coupon`.

So the hypothesis this PR implements — *"the owner withholds delivery until a heartbeat pokes it, so poke sooner"* — is **not supported by production**. Starting the first heartbeat at 5 s instead of 45 s changed nothing measurable. Either the delivery is not actually triggered by the heartbeat post, or something else dominates the 45 s.

### What IS still solid

The measurement that started this remains reproducible and unexplained: **45.20 s of source discovery on a healthy mesh, 90.19 s during the outage, against 2.6–2.83 s of Roslyn**, with 45.20 s matching `HeartbeatInterval` to within 200 ms. That correlation is real; my causal story about it was wrong.

### What this PR still legitimately contributes

- `SyncStreamOptions.FirstHeartbeat` as a documented, configurable knob (default 5 s, taken as the sooner of it and the cadence).
- `EarlyFirstHeartbeatTest`, including `EarlyTick_NeverDelaysAShorterConfiguredInterval` — which caught a real regression where an unconditional early tick *delayed* short cadences and broke `HeartbeatFireAndForgetTest`.
- No regression: 15/15 CI green, and the ci.1467 rollout did **not** need the manual scale-to-zero that the previous deploy required (16/16 roots verified rendering).

### Recommendation for whoever picks this up

Do not treat the stall as fixed. The next step is to instrument where those 45 s actually go — per-phase timing inside `SnapshotSources`/`GetSources`, and a log line when a delivery is satisfied by a heartbeat versus an Initial — rather than another causal guess. I made three of those tonight and each was wrong; the measurement is cheap and would have settled it hours ago.

Reverting is also reasonable if the knob isn't wanted: the behaviour change is one line.

Issue #686, opened 2026-07-28
Compile discovery costs ~45s on memex but 0ms on a fresh mesh — it is cold hub activation at scale, not the compile path


rbuergi:
## Controlled experiment settles it

Ran the **same instrumented image** (`3.0.0-ci.1469`, core #685) against a **fresh disposable mesh**, with a NodeType whose sources have the same cross-type `shared=` shape as `Store/Coupon`:

```
Probe/Main   sources: ["namespace:Source scope:subtree", "shared=@Probe/Lib/Source"]
```

Result:

```
⏱ NodeType definition resolved in 0ms (ok); source-freshness scan 0ms.
⏱ Source discovery took 0ms (synced query) — everything below this line is Roslyn.
Source query: namespace:Probe/Main/Source scope:subtree nodeType:Code
Source query: path:Probe/Lib/Source nodeType:Code
Source query: namespace:Probe/Lib/Source scope:subtree nodeType:Code
Source query: namespace:Probe/Main/Test scope:subtree nodeType:Code
Source discovery matched 2 Code node(s)
TOTAL: 109ms
```

**109 ms total, 0 ms discovery** — versus **45.20 s of discovery** for `Store/Coupon` on memex (and 90.19 s during the 2026-07-27 outage), against ~2.8 s of Roslyn in both cases.

## What this rules out

- ❌ The compile path itself. Identical code, identical `shared=` cross-type shape → 0 ms.
- ❌ `CodeQueryResolver` expansion (pinned by `RecursiveCompilationTest`, and visible above: all four queries expand correctly).
- ❌ The heartbeat theory (#684) — already disproved in production: 48 s before, 57 s after.
- ❌ `15s GetMeshNode + 30s SnapshotSources = 45s` — Loki shows those 15 s stalls only on `LocalSyncDemo/_Sync/party-1`, never on the compile path.

## What it points at

The difference between the two environments is **scale and partition topology**, not code. On memex those four-to-nine discovery queries are **cross-partition** (`path:Store/Plugin/Source`, `namespace:Store/Order/Source scope:subtree`, …) and must be answered by **per-node/partition hubs that may be cold**. A cold hub activation on memex is exactly the multi-second-to-minute event seen all night in `MessageHubGrain` timeouts and `SubscribeRequest` 60 s failures.

[...]

PR #688, opened 2026-07-28
Flakes: give hub tests a budget the CI fixture can actually meet

rbuergi:
## The catalogue first

Pulled every red `main` run of the last two days from check annotations. **Four named flakes, and not one is a bad assertion** — all four are timeouts:

| test | red main runs | failure |
|---|---|---|
| `MeshNodeVersionSyncTest.MeshNode_InitialVersion_IsZero` | 1 | `Test execution timed out after 10000 milliseconds` |
| `TwoSiloRecycleConvergenceTest.PostRecycleUpdate_NonOwnerSiloMirror_Converges` | 2 | `System.TimeoutException` |
| `TwoWaySyncTest.TwoWay_Update_KeepsServerAddition_UnlessForced` | 1 | `System.TimeoutException` |
| `ThreadAgentIntegrationTest.FullFlow_CreateThread_SendMessage_NonStreaming` | 1 | `GetMeshNode(…) timed out after 60.0s` |

## What this PR fixes: budgets a correct test could never meet

The same CI log for `MeshNodeVersionSyncTest` shows:

```
19:59:22.595  === TEST CLASS INIT ===
19:59:30.720  [DISPOSE] Mesh.Dispose() invoking
```

**8.1 s of mesh fixture inside a 10 s budget** — 1.9 s left for the body. On a quiet runner it fits; under an 8-shard load it cannot. The test was measuring the runner.

`[HubFact]` carried the same trap globally: **5 s in Release** (i.e. CI), which the fixture alone can exceed. It bit a heartbeat test written the same night — green locally, where no budget is enforced, and impossible on CI.

Both move to **30 s**, matching the runner-wide `methodTimeout`.

**Nothing is masked.** A genuine wedge surfaces its own `GetMeshNode … timed out after 60.0s` first, so a hang still fails the test — just not before the fixture has finished starting.

## Not fixed here — and NOT budget noise

The other three are recorded in the commit so nobody dismisses them as timing:

`ThreadAgentIntegrationTest` fails with

```
GetMeshNode('ACME/ProductLaunch') timed out after 60.0s — the owning per-node hub never
answered the GetDataRequest. NO LOCAL HUB … it never activated here.
```

That is the **same cold-hub-activation defect measured in production this week** ([#686](https://github.com/Systemorph/MeshWeaver/issues/686): 0 ms source discovery on a fresh mesh, 45.20 s on memex). These flakes are that bug reproducing under CI load — real signal, and they should stay red until it is fixed rather than be timed away.

## Testing

409/409 Persistence, 290/290 Data.

### DIFF
--- a/src/MeshWeaver.Fixture/HubFactAttribute.cs
+++ b/src/MeshWeaver.Fixture/HubFactAttribute.cs
@@ -10,17 +10,30 @@ namespace MeshWeaver.Fixture;
 public class HubFactAttribute : FactAttribute;
 #else
 /// <summary>
-/// Marks a hub test method. In non-DEBUG builds a 5 second timeout is applied
-/// so that a wedged hub fails the test fast instead of hanging the run.
+/// Marks a hub test method. In non-DEBUG builds a timeout is applied so that a wedged hub fails
+/// the test instead of hanging the run.
+///
+/// <para>🚨 The budget must exceed what the MESH FIXTURE costs on a loaded CI runner, or the test
+/// measures the runner instead of the code. It was 5s, and 5s is not achievable there: a CI log for
+/// <c>MeshNodeVersionSyncTest</c> (2026-07-27) shows class init at 19:59:22.595 and dispose at
+/// 19:59:30.720 — <b>8.1s of fixture</b> before the body runs at all. That produced a recurring
+/// main-red flake with no defect behind it, and the same trap bit a heartbeat test written the same
+/// night ("Test execution timed out after 5000 milliseconds", passing locally where no budget is
+/// enforced).</para>
+///
+/// <para>30s matches the runner-wide <c>methodTimeout</c> and still fails a genuine wedge well
+/// before the run hangs — a wedged read surfaces its own <c>GetMeshNode … timed out after 60.0s</c>
+/// first, so nothing is masked by the larger budget.</para>
 /// </summary>
 public sealed class HubFactAttribute : FactAttribute
 {
     /// <summary>
-    /// Initializes a new instance and sets the test timeout to 5000 milliseconds.
+    /// Initializes a new instance and sets the test timeout to 30000 milliseconds — above the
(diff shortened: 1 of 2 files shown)


rbuergi:
## Correction to my own framing: raising budgets is a treadmill, not a cure

This PR's CI failed on a test I did not touch and had not catalogued — `FutuReAnalysisTest.EuropeRe_AnnualProfitabilityWaterfall_ShouldRender` — and its failure is the important one:

```
Test execution timed out after 30000 milliseconds
[05:18:35.376] === TEST CLASS INIT: FutuReAnalysisTest ===
```

It **already carried `[Fact(Timeout = 30000)]`**, and it still consumed the entire 30 s in fixture startup. So the progression is 10 s → 30 s → still red. Widening the budget moves the line; it does not cross it.

That sharpens the conclusion in the commit message, and I want it recorded plainly rather than buried:

- **What this PR legitimately fixes** stands: `[HubFact]`'s 5 s Release budget and `MeshNodeVersionSyncTest`'s 10 s are *unmeetable* — an 8.1 s mesh fixture cannot fit in either, so those tests were measuring the runner. Raising them to the runner-wide 30 s removes two false failures.
- **What it does not fix, and cannot**: the tail. `FutuRe` at 30 s, `ThreadAgentIntegrationTest` with `GetMeshNode … NO LOCAL HUB … it never activated here`, `TwoSiloRecycleConvergence`, `TwoWaySync` — these are the **cold-hub-activation defect** ([#686](https://github.com/Systemorph/MeshWeaver/issues/686): identical NodeType shape compiles with 0 ms discovery on a fresh mesh, 45.20 s on memex) reproducing under CI load.

**I am deliberately not raising `FutuRe`'s budget to 60 s.** That would make main greener while deleting the clearest reproduction of the production bug we have. The honest state is: two test-side flakes fixed, and the rest correctly identified as one product defect that CI is reproducing for free.

Re-ran the failed shard to confirm this PR's change is not implicated.

PR #698, opened 2026-07-28
Bake NodeTypes against the image before the pod serves

rbuergi:
## The problem

Every image roll changes Graph's MVID. `FileSystemAssemblyStore` bakes `FrameworkVersion[..8]` into every filename **and** the lookup glob, so a new image misses the whole assembly cache **by design** (ABI safety — it's what prevents the `BadImageFormatException` wedge). The shared `/data/assembly-cache` PVC buys cross-pod and cross-restart reuse *on the same image*; it does nothing for a deploy.

Today that guaranteed miss is paid **lazily, unordered, on user requests, after the pod is already taking traffic**. That's the compile churn behind the "did not settle" overlays.

## What this does

**1. Decide what to build by probing the store, not the record.**

`HasUsableBuild` is deliberately a pure record check — its own doc says *"no store probe, no `File.Exists`"*. Right on the hot path; wrong as a source of truth. Clear or remount `/data/assembly-cache` and every NodeType still claims `Ok` with its bytes gone. Nothing re-drives a compile; the miss surfaces later, one instance at a time, at activation (with a sticky default-config fallback if it exhausts retries).

`NodeTypeBakeStatus` asks the store instead. Level-triggered on reality rather than on history, which buys three things at once: a wiped share re-bakes itself, an interrupted bake **resumes**, and a second replica inherits the first one's work. No checkpoint file, no marker that can lie.

**2. The sweep consumes the report.**

Types already on the share are `AlreadyBaked` and never activated — that's what turns a re-run from hours of Roslyn into a directory listing.

A type whose **bytes** are missing cannot be warmed by activation alone: every existing kickoff is keyed on the record (first-build needs null fields, recovery needs `Compiling`, framework-stale needs a mismatch), and in this state the record is pristine. `WarmOne` would have reported `Compiled` instantly **over an empty share** — the one outcome a bake gate must never produce. `RebuildMissingBytes` drives it properly: flip to `Pending`, then wait for a compile demonstrably newer than the one observed before the flip. Matching on status alone would accept the replayed pre-existing `Ok`.

**3. An opt-in readiness gate — "fail before prod, not in prod".**

The chart already keeps the old pod serving until the new one passes its `startupProbe` on `/health` (`maxSurge: 1`, `maxUnavailable: 0`). Gating `/health` on the bake turns a NodeType that no longer compiles from a production incident into a **stalled rollout**: the new pod never takes traffic, the previous image keeps serving.

Two safety rules, both tested:
- **Fails closed on a regression, open on "not running".** A type already broken before this image cannot gate — otherwise one abandoned NodeType freezes every future deploy, discovered when you urgently need to ship. A gate registered while the sweep is disabled reports healthy rather than black-holing the pod on a config mistake.
- **Never tagged `live`.** A long bake is not a wedged process; failing liveness would have Kubernetes restart the pod mid-bake.

## Deployment

Chart plumbing only — `PreWarm__DynamicTypes` and `PreWarm__GateReadiness` both default `false`, so **no behaviour change until someone flips them**. `probes.startup` is now configurable because the gate is unusable without it: a cold compile is ~90 s per NodeType and the sweep is sequential, so 60-100 types needs 2-3 **hours** of startup budget. The current default (5 × 60 = 5 min) would kill the pod mid-bake on every attempt. Suggested paired setting is in the values comment.

I'd enable this on one env with someone watching rather than land it on by default.

## Tests

- `NodeTypeBakeStatusTest` — 24 cases. Classification is pure, so every state is testable without a fixture, including bytes disappearing under a healthy record and a store that throws (fails safe → rebuild).
- `NodeTypeBakeGateStateTest` — 8 cases pinning the regression-vs-pre-existing rule and "completion is not absolution".
- `DynamicTypePreWarmerTest` — existing 3 still green; two assertions relaxed from `== Compiled` to `ReachedUsableBuild`, since a type found already on the share is equally fine and asserting the status verbatim made the test a race against the probe.

## Not in this PR

The bake-ahead pod (warm the new tag's cache while the old image still serves, then gate the rollout on the result) — safe to build separately because the FrameworkTag is in the filename, so old- and new-image assemblies coexist in the share.

### DIFF
--- a/.github/workflows/main-cd.yml
+++ b/.github/workflows/main-cd.yml
@@ -201,6 +201,20 @@ jobs:
           -p:ContainerRegistry=${{ env.ACR }}
           -p:ContainerRepository=memex-migration
           -p:ContainerImageTags='"${{ steps.vars.outputs.version }};${{ steps.vars.outputs.sha }};main"'
+      # nodetype-bake — compiles every dynamic NodeType against THIS build into the shared assembly
+      # cache before the portal rolls onto it, and fails the release on a regression. It must be
+      # built from the SAME commit as the portal: the framework identity it bakes against is the
+      # Graph MVID of these binaries, so an image built from a different commit would warm a cache
+      # the portal cannot use. Same multi-arch publish; <RuntimeIdentifiers> is project-local.
+      - name: Build + push nodetype-bake
+        run: >
+          dotnet publish memex/aspire/Memex.NodeType.Bake/Memex.NodeType.Bake.csproj
+          -c Release --no-self-contained -t:PublishContainer -p:PublishProfile=
+          -p:ContainerRuntimeIdentifiers='"linux-x64;linux-arm64"'
+          -p:CIRun=true
+          -p:ContainerRegistry=${{ env.ACR }}
+          -p:ContainerRepository=memex-bake
+          -p:ContainerImageTags='"${{ steps.vars.outputs.version }};${{ steps.vars.outputs.sha }};main"'
 
   # The MeshWeaver.Plugins PR gate image: `mw-plugin-test` (tools/MeshWeaver.PluginTester)
   # imports a node-repo checkout into a fresh in-process mesh, asserts every NodeType compiles,
--- a/MeshWeaver.slnx
+++ b/MeshWeaver.slnx
@@ -32,6 +32,7 @@
     <Project Path="memex/aspire/Memex.AppHost/Memex.AppHost.csproj" />
     <Project Path="memex/aspire/Memex.Aspire.Hosting/Memex.Aspire.Hosting.csproj" />
     <Project Path="memex/aspire/Memex.Database.Migration/Memex.Database.Migration.csproj" />
+    <Project Path="memex/aspire/Memex.NodeType.Bake/Memex.NodeType.Bake.csproj" />
     <Project Path="memex/aspire/Memex.Portal.Distributed/Memex.Portal.Distributed.csproj" />
     <Project Path="memex/aspire/Memex.Portal.ServiceDefaults/Memex.Portal.ServiceDefaults.csproj" />
   </Folder>
--- a/deploy/helm/templates/memex-bake/job.yaml
+++ b/deploy/helm/templates/memex-bake/job.yaml
@@ -1,0 +1,88 @@
+{{- if .Values.bake.enabled }}
+---
+# Run-once NODETYPE BAKE job — compile every dynamic NodeType against the image being deployed,
+# into the shared assembly cache, BEFORE the portal rolls onto it. Same shape as memex-migration:
+# a Job (not a Deployment) so it runs to completion and STOPS.
+#
+# 🚨 WHY THIS EXISTS. Every image roll changes the framework identity (Graph's MVID), and that
+# identity is part of the assembly store's filename AND its lookup glob — so a new image misses the
+# whole compile cache BY DESIGN (ABI safety). Left to the portal, that guaranteed miss is paid
+# lazily, unordered, on user requests, after the pod is already serving: ~90 s of cold Roslyn per
+# NodeType on somebody's page load. Paid here it is paid once, on a container nobody is waiting for.
+#
+# 🚦 AND IT IS THE GATE. The job exits non-zero when a NodeType that USED TO compile no longer does,
+# so a broken type is discovered before any traffic moves. A type that was already broken beforehand
+# does NOT fail the bake — it is broken in production right now, and blocking every future rollout
+# on it would freeze deploys entirely.
+#
+# Safe to run while the OLD image is still serving: the framework tag is in the store filename, so
+# old- and new-image assemblies coexist and this writes files the running pods never read.
+#
+# The name embeds .Release.Revision so a fresh bake runs on every `helm upgrade`, and
+# ttlSecondsAfterFinished auto-cleans the completed Job — same reasoning as memex-migration.
+apiVersion: "batch/v1"
+kind: "Job"
+metadata:
+  name: "memex-bake-{{ .Release.Revision }}"
+  labels:
+    app.kubernetes.io/name: "{{ .Chart.Name }}"
+    app.kubernetes.io/component: "memex-bake"
+    app.kubernetes.io/instance: "{{ .Release.Name }}"
+spec:
+  # 🚨 backoffLimit 0 — a bake failure is a VERDICT, not a flake. Retrying a NodeType that does not
+  # compile just burns another full sweep to reach the same answer, and a retried-then-passed bake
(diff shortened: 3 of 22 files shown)


rbuergi:
## Added: one pod bakes, the rest follow

Roland flagged the hole — with `maxSurge`, or any `replicas > 1`, **every** pod on the new image independently finds the same framework-stale cache and starts the same sweep into the same volume. Not just duplicated work: concurrent cold Roslyn compiles of the *same* NodeType against one network volume — precisely the storm the sequential ordered sweep exists to prevent. The readiness gate in the first commit would have made it the common case rather than an edge case.

`NodeTypeBakeLease` — a heartbeat file lease on the shared assembly-cache directory. One pod wins an atomic `CreateNew` and bakes; the others follow and compile nothing.

**The follower is not passive.** Each poll it re-probes the share *and* re-attempts the lease, so it finishes the moment the baker is done, and if the baker **dies** its lease goes stale and the next poll takes the bake over. Same level-triggered rule as the rest of this design — no state needs a human to notice it.

**Keyed per framework version**, so a bake-ahead pod on a new image and the live pods on the old one don't block each other (different files); only same-image replicas contend.

**Fails open throughout** — any I/O problem resolves to "you may bake". A coordination mechanism that can *deny* work turns a transient volume blip into a fleet that never compiles anything, which is strictly worse than the duplicate work it prevents.

Held by `Observable.Using` for the lifetime of the sweep, so it's released on completion, error or unsubscribe rather than outliving its bake.

8 tests: mutual exclusion, immediate handover on release, steal-on-stale (the fleet-wedge guard), no-steal-while-fresh (or the storm returns), per-framework independence, fail-open on an unusable directory, and a guard that the heartbeat stays well inside the stale window so one missed beat under load can't hand the bake to a second pod mid-compile.

All green: 20 in `Hosting.Monolith.Test`, 24 in `Graph.Test`.

rbuergi:
## Added: the bake is its own service, like db-migration

Per Roland — `memex/aspire/Memex.NodeType.Bake`, a run-once Job on the image being deployed, same shape as `memex-migration`. It compiles every dynamic NodeType into the shared assembly cache **before** the portal rolls onto it, and exits non-zero when a type that used to compile no longer does.

**Reference fidelity is why it references `Memex.Portal.Shared`** rather than a hand-picked library list. A NodeType compiles against the assemblies loaded in the process doing the compiling, so a divergent reference set would let a type bake green here and still fail in the portal — and a bake whose result doesn't predict the portal's is *worse* than no bake, because the gate certifies it. Monolith mesh is the one deliberate difference: this process must not join the Orleans cluster, take grain ownership from pods currently serving, or vanish mid-activation when it exits. It uses the **portal's** ConfigMap for the same reason.

**It refuses to run without `Deployment__DataRoot`** (exit 2). Baking into a container-local directory would warm nothing and exit 0 — a green gate over an empty cache is the one wrong answer this must never give.

**`backoffLimit: 0`** — a bake failure is a verdict, not a flake. Retrying burns another full sweep to reach the same answer, and a retried-then-passed bake would be indistinguishable from a clean one.

### What unblocked it

I'd flagged this as blocked earlier, and the blocker was real: the compile write-back stamps `CompiledFrameworkVersion` onto the **shared** NodeType record, so pods still serving the old image would see a mismatch, fire the framework-stale kickoff, recompile back, and storm production.

Fixed at the source — the kickoff now probes the store before rebuilding. `CompiledFrameworkVersion` says which framework the last *write-back* came from; it does not say whether bytes for *our* framework exist, and the store answers the question that matters because its key carries the live tag. The version mismatch stays the cheap per-emission pre-filter; the probe runs at most once per hub lifetime, after the existing `Take(1)`, so the hot path the original comment protected is untouched.

That same rule now holds in three places — the probe, the sweep, and the kickoff: **bytes on the share outrank the record.**

### Deployment

`bake.enabled` defaults false — it only makes sense where `/data` is a shared persistent volume; with an `emptyDir` it warms a directory that dies with the Job. CD builds `memex-bake` from the same commit as the portal, since the framework identity it bakes against is these binaries' MVID.

Chart verified with `helm template` (absent when disabled, rendered when enabled) and `helm lint`. The service builds; it has **not** been run against a real cluster.

rbuergi:
## Ran it on a real cluster — it found two false-greens

Deployed to a k3s cluster (colima, own `bake-test` namespace, images built from this branch, portal + bake sharing one `/data` PVC). Testing earned its keep immediately: **the bake certified an image it had never checked.**

### 1. It raced the migration → silent pass

Bake and migration are separate Jobs with no ordering, both gating only on Postgres TCP. The bake won, came up against an unmigrated database, enumerated **zero** NodeTypes and exited **0 in 0.05s** — while the portal went on to compile four types lazily, the exact thing the bake exists to prevent.

Now it waits for the migration to record a `db_version` (the same condition the portal's `DbVersionGate` enforces, so the two can't disagree about what "migrated" means) before the mesh is started at all.

### 2. Finding nothing counted as passing

An empty outcome set is indistinguishable from the ways enumeration *fails*: unmigrated or wrong database, a query that timed out (`WarmDynamicTypes` deliberately completes empty on enumeration errors — in the portal a failed warm must not be fatal), a mesh still seeding. A gate that certifies "I verified nothing" is worse than no gate. Zero types now exits 3 unless `Bake__AllowEmpty` says the mesh is genuinely empty.

### 3. The whole mechanism was invisible

`appsettings.json` pins `"MeshWeaver"` to `Warning` with an allow-list for low-volume lifecycle markers, and the bake wasn't on it — so a pod held out of rotation by the readiness gate explained itself to **nobody**, and diagnosing run #1 needed the source rather than the logs. Added, with the volume rationale.

## What the cluster confirmed after the fixes

- **Bake**: waits for `db_version=49`, enumerates 4 types, builds them in dependency order, exits 0. The probe classified the wiped cache exactly right — `bytesmissing=3 previouslybroken=1`.
- **Lease** (the single-baker property): a second portal pod logged *"another pod holds the bake lease — following the shared assembly store instead of compiling (4 type(s) outstanding)"*. Verified cross-process, not just unit-tested.
- **Readiness gate**: the pod sat `0/1` for the whole 66s sweep and flipped to Ready on completion.
- **Resume**: a settled boot reports `0 of 4 need building — 4 already on the share. Building: (nothing)` — zero recompiles.

One honest caveat on that last point: the first two boots did **not** hit the cache, because node versions were still churning while the mesh settled after import (four build generations accumulated for one type: v14 from the bake, v18, v25, v29). The store is keyed on the NodeType's MeshNode version, so a build is superseded by any write that bumps the node. Once settled it hits cleanly. A production deploy — existing mesh, stable node versions, only the framework tag changing — is the settled case, but I have not exercised a framework roll end-to-end on the cluster, only a fresh install.

The `bake-test` namespace is still up if you want to poke at it.

PR #706, opened 2026-07-28
Pin what a partition-owning create actually registers (and what onboarding never reaches)

rbuergi:
**Draft on purpose. The fix is reasoned from evidence but NOT proven — the tests in it pass without it.**

## The evidence

On the education e2e mesh: schema `"e2e-admin"` existed, `"e2e-admin".mesh_nodes` held `id=e2e-admin ns='' User`, `auth.mesh_nodes` mirrored it — and `Admin/Partition` held **eleven entries, every one a package or space, not one user.**

Routing learns partitions from those nodes (`PostgreSqlExtensions`: *"Start the `Admin/Partition/*` subscription so writes can route"*). So every address in the partition answered `No node found at 'e2e-admin'`, the hub faulted on activation **thirty times in one test** — because the lookup can never succeed — and the install page hung 3.5 minutes with nothing naming a cause.

## The asymmetry

`OwnsPartitionProvisioningValidator` is documented as *the ONE place a partition's backing store is created*, and it creates the schema for `User` and `Space` alike — but writes **no** `Admin/Partition` definition. `PackageInstaller` writes one alongside a package install; `IncludedPartitionStaticProvider` emits one for configured inclusions.

**Package partitions routed; self-provisioned ones didn't.** That's why it survived — the paths people exercise most were already fine.

## Why this is a draft

`MonolithMeshTestBase` uses **in-memory persistence**, where routing doesn't depend on `Admin/Partition` the way the PostgreSQL path does. The three tests pass with the fix **and with it disabled** — I checked by removing the one `SelectMany` and re-running: still 3/3 green. They're kept as a shape check and labelled as such, not as a regression test.

So: do not merge this on a green run. It needs verification on a **PostgreSQL-backed mesh** — create a top-level `User`/`Space`, assert `Admin/Partition/{id}` exists and an address inside the partition resolves rather than timing out. The repro without the fix is `npx playwright test 10-install.spec.ts` (fails at 3.5m; needs the *dotted* trace key `Logging__LogLevel__MeshWeaver.Messaging=Trace`).

## If it holds

This is likely the production bug behind *"entitled but no home partition = never installed"* — 2 of 19 PartnerRe learners. Same shape: a user whose space exists but is unroutable.

## Dead ends, so nobody repeats them

- **Resource starvation** — reproduces at `/health` 6ms.
- **Bulk-only cross-test interaction** — reproduces with a single test.
- **`NamespacePicker.QueryOnce`** (Plugins #222) — fix served and verified; failure unchanged.


rbuergi:
**Correction — the fix in this PR was wrong and has been removed.** What remains is the test that disproved it.

I tested it against Postgres, which the original Monolith tests could not see, and the theory did not survive:

| | registered at `Admin/Partition/{name}`? |
|---|---|
| `Space` create | **yes — already, with or without the change** |
| `User` create | *n/a — the create is refused outright* |

So the registration I added was redundant on the one path it could reach. And a top-level `User` cannot go through the create pipeline at all — it returns `Access denied: Create permission required`. Users are onboarded by writing the root **straight through the storage adapter** (`FirstUserOnboardingTests` does exactly that), so the create-validation chain — and therefore `OwnsPartitionProvisioningValidator`, where partition provisioning lives — **never runs for a user**.

That is the real explanation for the observation that started this: eleven packages and spaces in `Admin/Partition`, not one user. The user's *schema* is created along the adapter path; nothing on that path registers the partition. Any fix belongs in onboarding, not in a validator a user create never reaches.

There is also a control test that reads an uncreated path and asserts null — because the first version of these tests passed with the fix disabled, and I would rather the next reader see the guard than trust the greens.

**Still open:** whether the missing registration is what hung the install. Two things in the code argue it is not on the read path — `PostgreSqlPartitionSubscriptionHostedService` says *"the router maps a path's first segment to a schema synchronously"*, and `SpaceOnboardingIntegrationTests` says *"No Admin/Partition pre-write needed"*. Recorded rather than resolved.

Ready for review as a test-only change.

PR #707, opened 2026-07-28
Issue every API surface on a portal hub, never the root mesh hub

rbuergi:
## Why

The root `mesh/{id}` hub is transient routing infrastructure, not a call target — request-shaped work issued on it never answers. The caller's queue stays empty at `RunLevel=Started` while the pending callback runs out its timeout, which reads like *"the target never replied"* and sends you hunting a race that doesn't exist. That misreading is exactly what cost a session on the `ThreadAgentIntegrationTest` hang (fixed separately in #705).

**Why grep can't find this class of bug:** `MeshHostApplicationBuilder` installs `MessageHubServiceProviderFactory(BuildHub)`, so the mesh hub's own provider **is** the ASP.NET root provider. Every constructor-injected or `RequestServices`-resolved `IMessageHub` is therefore the root mesh hub. Resolving it is fine — using it as a **call origin** is the defect. `hub.ServiceProvider.GetService<T>()`, `hub.JsonSerializerOptions`, and `hub.GetHostedHub(...)` as a hosting *parent* are all left alone.

## Server surfaces

| Surface | Defect | Fix |
|---|---|---|
| MCP **and** REST | `SessionHubResolver` returned `rootHub` outright when no session id resolved → every anonymous/claimless call ran on the root mesh hub | always materialise a portal hub; those callers share one process-stable `portal/{prefix}-anon-{instance}` (shared, so an anonymous call can't leak a hub + routing registration per request) |
| MCP | `rootHub.CommitToGitHub` / `UpdateToLatestFromGitHub` / `CheckBranchStateOnGitHub` — these call `hub.RunActivity(...)` | issued on `sessionHub`, matching `GitHubActionArea`, which already calls them on its portal host hub |
| gRPC | `ValidateTokenRequest` issued on the root hub — and the surrounding `.Catch` degrades a faulted validation to **Anonymous**, so a valid API token silently lost its identity instead of failing loudly | lazily-materialised `portal/grpc-{instance}` transport hub |

REST was audited endpoint-by-endpoint: all mesh-operation endpoints route through `ResolveSession`, so the resolver fix covers them. `SeoEndpoints` and `PluginRegistryEndpoints` use the hub only as a service-provider/serializer handle. The CLI is clean — it's pure HTTP over `/api/mesh/*` and its session id comes from claims.

## Clients — cache the portal id

The address a client connects with **is** its portal-hub address on the server, so a fresh id per connect mints a fresh server-side hub on every connect/reconnect and abandons the previous one's subscriptions and sync sub-hubs. The Python and TypeScript SDKs minted `py/<uuid>` / `node/<uuid>` per `connect()`; they now reuse one cached `portal/<id>` per process (`MESHWEAVER_CLIENT_ID` pins it across invocations).

The cache is **per-process, not a machine-wide file** — the server lets the *latest* claimant of an address win, so a shared id would make two concurrent client processes steal each other's pushes. Same reason portal-next's `stableClientId()` is per-tab rather than a cookie. Workers are unaffected: they pass explicit stable addresses (`py/python-kernel`, `node/node-kernel`).

`portal` is already a default stream-routed address type (`MeshConfiguration.DefaultStreamRoutedAddressTypes`), so these route on Orleans without any new `AddStreamRoutedAddressType`.

## Verification

- `MeshWeaver.Hosting.Grpc.Test` — 8/8 green
- `DocumentationLinkIntegrityTest` — green
- `Hosting.Grpc`, `Mcp`, `Memex.Portal.Shared`, `Documentation` all build clean under `-c Release -warnaserror`

Includes a What's New entry (user-facing: API/MCP/gRPC reliability + SDK reconnect behaviour).

### DIFF
--- a/clients/python/meshweaver/connection.py
+++ b/clients/python/meshweaver/connection.py
@@ -181,7 +181,15 @@ class MeshConnection:
 #: an address win, so two concurrent client processes sharing one machine-wide id would silently
 #: steal each other's pushes. Set ``MESHWEAVER_CLIENT_ID`` to pin a stable id across process
 #: invocations (a long-lived agent or CLI session reconnecting to its own hub).
-_CLIENT_ID = os.environ.get("MESHWEAVER_CLIENT_ID") or uuid.uuid4().hex
+#:
+#: The value is sanitised to ``[A-Za-z0-9_-]`` (mirroring ``SessionHubResolver.Sanitize`` on the
+#: server): an operator-supplied id containing ``/`` would otherwise claim a multi-segment address
+#: like ``portal/a/b`` — a different address shape than intended, which would not round-trip
+#: through the server's ``Address`` parsing.
+_CLIENT_ID = "".join(
+    c if (c.isascii() and c.isalnum()) or c in "-_" else "-"
+    for c in (os.environ.get("MESHWEAVER_CLIENT_ID") or uuid.uuid4().hex)
+)
 
 
 def client_address() -> str:
--- a/clients/python/meshweaver/connection.py
+++ b/clients/python/meshweaver/connection.py
@@ -9,6 +9,7 @@ from __future__ import annotations
 
 import asyncio
 import json
+import os
 import uuid
 from typing import Any, AsyncIterator, Awaitable, Callable, Optional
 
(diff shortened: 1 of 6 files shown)


rbuergi:
Thanks — the first finding was a real defect in this PR, now fixed in `3c72699`.

**`TransportHub` missing `RegisterStream` — fixed.** Correct and important. `portal` is a stream-routed address type, so on Orleans the RoutingGrain dispatches to that address over the cluster-wide memory stream; without the registration the `ValidateTokenResponse` had nowhere to land cross-silo and the request would time out — reintroducing exactly the failure this hub exists to avoid. Worth noting the monolith masks it (routing short-circuits on `GetHostedHub`), so `Hosting.Grpc.Test` passes either way and only a distributed deployment would have surfaced it. Now wired the same way `SessionHubResolver` already wires its session hubs.

**`MESHWEAVER_CLIENT_ID` sanitisation (Python + TypeScript) — fixed.** Both SDKs now sanitise to `[A-Za-z0-9_-]`, mirroring `SessionHubResolver.Sanitize` on the server, so an id containing `/` can't claim a multi-segment address like `portal/a/b`. Verified both implementations map the same input to the same output.

**"an Mcp-Session-Id header" → "a" — not taken.** "MCP" is initialised ("em-cee-pee"), a vowel sound, so "an" is correct here.

PR #715, opened 2026-07-29
Finish in-flight hub constructions at disposal instead of racing them

rbuergi:
Closes the **finish** half of the teardown contract — *properly finish the requests already started, transparently refuse new ones* — for hosted-hub construction. The refusal half already exists (`CloseCreation` cascade, #635); this PR closes the "check-then-act residue" #488 named and #613's dump analysis pinned.

## The race

A routed emission's hub creation passes the `IsDisposing` check moments before disposal begins. `DisposeHubsReactive` then snapshots `messageHubs` — but the hub under construction isn't there yet (it lands only after `Build`), so the collection signalled `DisposalCompleted` and the owner tore down the container **while `Build` was still resolving services**:

```
MeshQuery emission (still live at teardown)
 → MonolithRoutingService.CreateHub → HostedHubsCollection.CreateHub
   → MessageHubConfiguration.Build → SubscribeToOwnDeletion → InstallReleaseRequestWatcher
     → GetService → disposed Autofac LifetimeScope
```

Locally that's the `ObjectDisposedException` teardown-straggler class; on CI the same construction walking the TypeRegistry over an unloading collectible ALC is the **#613 SIGSEGV**.

Why such emissions exist at all: `hub.Post` is fire-and-forget by design, so a test's (or teardown's own) last write's route resolution can always straddle `DisposeAsync`. The traffic can't be eliminated — only drained.

## The fix

Fully reactive, no async/await, no blocking wait: `GetHub` counts callers inside a creation `Lazy` and pings a `Subject` on completion; `DisposeHubsReactive`'s completion join gains an **inflight leg** that waits for the count to drain and then disposes whatever the late construction produced — inside the join, before `DisposalCompleted` — so the container outlives every in-flight `Build` and the late hub is never a zombie outside the disposal snapshot. Merge order (ping subscribed before the immediate probe) closes the probe/ping race; the existing join `Timeout` still caps a wedged construction.

## Proven, not asserted

- **Deterministic repro** (`InflightHubCreationDrainTest`): a creation parked inside `WithInitialization` — the same hook the straggler stacks die in — while the parent disposes. **Without this change the test FAILS** (disposal completed in ~20 ms with the construction still in flight); with it, disposal waits, then the late hub's own `DisposalCompleted` terminates. A second test pins the refusal half (creation after dispose-start returns null).
- **End-to-end measurement** — `MeshWeaver.AI.Test` via the native xUnit host, the way CI runs it: the hub-construction straggler chain goes **14 → 0** (total 19 → 12; 919/919 green before and after). The remaining 12 are a *different* class — in-flight `ThreadExecution` round work (`AgentChatClient.GetStreamingResponseAsync`, `MeshQuery` reads) touching disposed state — consumer-side, deliberately untouched here.
- `Messaging.Hub.Test` 109/109; both projects build clean under `-c Release -warnaserror`.

No What's New entry: hub-lifecycle/teardown internals, no user-visible change.

Related: #613 (this removes its trigger), #488 (same diagnosis, closed — this is that idea rebased onto the CloseCreation architecture), #708 (independent; unbreaks main).

### DIFF
--- a/src/MeshWeaver.Messaging.Hub/HostedHubsCollection.cs
+++ b/src/MeshWeaver.Messaging.Hub/HostedHubsCollection.cs
@@ -130,8 +130,14 @@ public class HostedHubsCollection(IServiceProvider serviceProvider, Address addr
     // In-flight construction tracking (see GetHub): count of callers currently inside a
     // creation Lazy, and a ping per completion so DisposeHubsReactive can wait reactively
     // for the drain — no async/await, no blocking wait on the disposal path.
+    //
+    // 🚨 Synchronized: concurrent creations of DIFFERENT addresses complete on different
+    // threads, and a bare Subject's OnNext is not safe under concurrent callers — a torn
+    // notification could drop the very ping that reports the count reaching zero, stalling
+    // the drain leg until the join's Timeout. Subject.Synchronize serialises the
+    // notifications (same pattern as MeshNodeStreamCache).
     private int inflightCreations;
-    private readonly Subject<Unit> inflightChanged = new();
+    private readonly ISubject<Unit> inflightChanged = Subject.Synchronize(new Subject<Unit>());
 
     /// <summary>
     /// Per-address construction single-flight (see <see cref="GetHub"/>). Entries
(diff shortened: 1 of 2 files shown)


rbuergi:
Copilot's finding was correct and is fixed in `01d95d0d4`: concurrent creations of different addresses complete on different threads, and a bare `Subject.OnNext` is not safe under concurrent callers — a torn notification could drop the very ping that reports the in-flight count reaching zero, stalling the drain leg until the join's Timeout (a 5s degraded stall, not a hang, thanks to the existing cap — but a real defect). Now `Subject.Synchronize`, the same pattern `MeshNodeStreamCache` uses. 109/109 locally; the try/catch stays as belt-and-braces for process-teardown races, no longer masking concurrency faults.

PR #716, opened 2026-07-29
Hand the test logger a resolved config, not the container it outlives

rbuergi:
Salvages the third commit of #708 (`99577caba`), which landed on that branch ~2 minutes after the PR was merged and therefore never reached `main`.

## The defect

Teardown logs AFTER the DI container is disposed. `XUnitFileLogger` held the `IServiceProvider` and resolved `IConfiguration` on first log per category, so anything logging from a `Dispose` path threw:

```
MonolithMeshTestBase.DisposeAsync → AsyncDisposeQueue.DrainAsync
  → CompilationCacheService.Dispose        (logs during teardown)
    → XUnitFileLogger.GetMinLogLevel → GetService<IConfiguration>
      → disposed Autofac LifetimeScope → ObjectDisposedException
```

Measured on `Acme.Test` via the native xUnit host (the way CI runs it): **23 first-chance `ObjectDisposedException`s per run**, every one this chain. They were caught (`GetMinLogLevel` swallows and defaults to `Information`), so they never failed a test — they burned 23 throws + 23 first-chance stack captures on the teardown path and buried `teardown-stragglers.log` in fault-shaped noise. That log is the working diagnostic for the #613 teardown crashes (it is how the #715 hub-construction chain was found), so keeping it signal-only has direct value.

## The fix

`XUnitFileLoggerProvider` resolves the configuration **object** once, in its constructor, while the container is guaranteed alive, and hands that object to every logger. `IConfiguration` is a plain object once resolved — the reference stays valid after the scope that produced it is gone — so this removes the post-disposal container access rather than guarding it.

Verified: `Acme.Test` 76/76 exit 0 via the native host, stragglers **23 → 0**; `MeshWeaver.Fixture` builds clean under `-c Release -warnaserror` on current main.

Explicitly NOT claimed: that this fixes the transient CI `exit=2` (those exceptions were caught; attempt 2 of the same commit ran clean with no code change).

No What's New entry: test-infrastructure only.

### DIFF
--- a/src/MeshWeaver.Fixture/XUnitFileOutputHelper.cs
+++ b/src/MeshWeaver.Fixture/XUnitFileOutputHelper.cs
@@ -325,9 +325,12 @@ public class XUnitFileLoggerProvider : ILoggerProvider
         {
             _configuration = serviceProvider?.GetService<IConfiguration>();
         }
-        catch
+        catch (ObjectDisposedException)
         {
             // A provider built from an already-disposed scope: no configuration, default levels.
+            // ONLY the disposed case is tolerated — any other resolution failure (misconfigured
+            // DI, a faulting IConfiguration factory) is a real defect and must surface loudly,
+            // not degrade the log level silently.
             _configuration = null;
         }
     }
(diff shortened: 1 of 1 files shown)


rbuergi:
Copilot's finding taken in `a0263fbb1`: the catch-all could mask a real DI failure as a silent log-level downgrade, against this repo's own no-swallow rule. Now catches `ObjectDisposedException` only — the one case the constructor is meant to tolerate (a provider built from an already-disposed scope); anything else surfaces loudly. Acme.Test 76/76, Fixture clean under `-warnaserror`.

PR #747, opened 2026-08-02
fix(auth): make Sign in with Apple actually work (form_post + generated client secret)

rbuergi:
## What

`AddAppleAuthentication` could never complete a sign-in:

1. **Missing `response_mode=form_post`.** Apple requires it whenever scopes are requested (the code requested `name`+`email`), so the authorization request was rejected outright — and the callback would have arrived as a cross-site POST the generic `OAuthHandler` cannot parse anyway.
2. **Static client secret.** Apple issues none: the secret is an ES256-signed JWT minted from the Sign in with Apple `.p8` key, valid ≤6 months.

## How

- Swap the hand-rolled `AddOAuth("Apple")` body for **`AspNet.Security.OAuth.Apple` 10.0.0**, which owns the protocol (form_post challenge + POST callback parsing, cached client-secret generation, id-token validation against Apple's published keys). The extension now only wires configuration.
- Config surface: `Authentication:Apple:{ClientId,TeamId,KeyId,PrivateKey}` — `ClientId` is the Services ID; `PrivateKey` is the `.p8` content, accepted as raw PEM, PEM with literal `\n`, base64 of the PEM file, or bare PKCS#8 base64 (`NormalizePrivateKey`). Static `ClientSecret` remains as fallback for an externally minted JWT. Still a no-op without `ClientId`.
- Correlation cookie forced `Secure` (form_post = cross-site POST; same rationale as the Microsoft handler).
- Helm chart: `Authentication__Apple__{ClientId,TeamId,KeyId}` in the portal ConfigMap, `Authentication__Apple__PrivateKey` conditional in the Secret.
- What's New entry included.

## Tests

`AppleAuthenticationTest` (Memex.Portal.Shared.Test, 11 cases): scheme registers with the Apple handler, generated-vs-static secret selection, no-op without ClientId, and every key-normalization shape proven importable via `ECDsa.ImportFromPem` with a real ES256 key. `dotnet build -c Release -warnaserror` clean; `helm lint` clean.

### DIFF
--- a/Directory.Packages.props
+++ b/Directory.Packages.props
@@ -7,6 +7,7 @@
     <PackageVersion Include="AngleSharp" Version="1.5.2" />
     <PackageVersion Include="Appium.WebDriver" Version="8.3.0" />
     <PackageVersion Include="AspectCore.Extensions.Reflection" Version="2.4.0" />
+    <PackageVersion Include="AspNet.Security.OAuth.Apple" Version="10.0.0" />
     <!-- Open-source (MIT) MAUI UI libraries for the native view pack's rich controls (Phase 4+):
          charts, full data grid, and popups/expanders. All MIT-licensed — no Syncfusion/DevExpress/
          Telerik (commercial). Referenced by MeshWeaver.Maui / Memex.Client as the controls land. -->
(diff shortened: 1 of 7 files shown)


rbuergi:
Both review points addressed in 735b49e69: CallbackPath pinned to /signin-apple (the NonfileRouteConstraint exclusion makes it a routing contract), and PrivateKey-without-TeamId/KeyId now throws at registration time naming the missing config key (covered by two new test cases, 13/13 green).

Issue #748, opened 2026-08-02
Move NodeType compile state off NodeTypeDefinition into the compile activity / a _Compile satellite


rbuergi:
## Decision (maintainer, 2026-08-02)

Compile bookkeeping must not live on the NodeType node's content. Per-run results belong to the compile **Activity** (largely already true — `{type}/_Activity/compile-…` carries the log); the steady state (status, trigger, assembly pointers, source-version maps) belongs in a **fixed-id satellite** (e.g. `{type}/_Compile`), which the `_`-prefix rule already excludes from GitSync export (`GitHubSyncService.Filter`, line ~195) and prune (`StaticRepoImporter.IsGovernance`).

**Why:** the mesh writes these fields into the same node the repo ships. Every compile makes the node "server-modified", exports carried a stale compile verdict into git, and imports stamped it back (stale-green → type parks on a cold cache). The transitional fix (authored-token change detection + operational strip/preserve, PR forthcoming) keeps the seams honest; this issue is the structural fix that makes the seams unnecessary.

## The affected fields (all on `NodeTypeDefinition`, `MeshWeaver.Graph/Configuration/NodeTypeDefinition.cs`)

`CompilationStatus` (:241), `CompilationError` (:249), `CompilationDiagnostics` (:263), `LastCompileStartedAt` (:269), `LastCompileSucceededAt` (:276), `LastCompiledVersion` (:286), `LastCompilationActivityPath` (:296), `LatestReleasePath` (:309), `RequestedReleasePath` (:325), `RequestedReleaseAt` (:340), `RequestedReleaseForce` (:348), `RequestedReleaseBy` (:361), `LastReleaseRequestHandledAt` (:370), `LatestAssemblyCollection` (:387), `LatestAssemblyPath` (:395), `ReleaseNotes` (:404), `CompiledSources` (:431), `CurrentSourceVersions` (:448), `CompiledFrameworkVersion` (:518). `IsDirty` (:473, `[JsonIgnore]`) derives from the two dictionaries and needs re-sourcing. `[JsonExtensionData] UnknownMembers` (:536) makes removal deserialization-safe for stored nodes.

## Writers to migrate (~15 sites)

- `NodeTypeCompilationHelpers.cs` — kickoff/park/recovery/source-watcher/release-watcher/dispatch/terminal writes (:339, :229, :390, :482, :650, :684, :890, :1094, :1264, :1447, :1498)
[...]

PR #749, opened 2026-08-02
NodeType sync owns authored content; compile bookkeeping stays on the mesh

rbuergi:
## The defect (memex, 2026-08-02)

A NodeType node's content mixes the repo-authored definition (configuration, sources, description) with mesh-owned compile bookkeeping (`compilationStatus`, `latestAssemblyPath`, `compiledSources`, …). The sync treated both alike:

- **Export** committed a compile verdict into git — stale the moment the next compile ran.
- **Import** stamped that stale copy back over the live node — a **stale green** claiming a weeks-old assembly is current (the class that parks a type on a cold cache). Operationally this made every plugin-source deploy a **force-sync + manual recompile of every touched type** (today's Store/Chess deploys, live evidence: `Store/Plugin` regressed to a 2026-07-19 v202 verdict by a force import).
- **Change detection** hashed the bookkeeping, so bookkeeping-only diffs read as authored changes.

## The fix — inside the existing import logic, no parallel comparison mechanism

`NodeTypeOperationalContent` (Mesh.Contract) is the ownership split; the three seams consume it:

| Seam | Rule |
|---|---|
| `GitHubSyncService.SerializeOne` (export) | **Strip** operational members — repo files carry the authored definition only. Export-only seam: the file parsers stay shared with filesystem persistence, which must round-trip full nodes. |
| `StaticRepoImporter` (import, incl. force) | **Preserve** the live node's operational members — authored changes land, live compile state stays exactly as the mesh last wrote it, stale embedded copies never land. Authored-identical imports return the same instance → structural no-op upserts. |
| `PartitionSourceFingerprint` (per-node manifest token + partition fingerprint) | **Ignore** operational members — authored content only, and every `Code` node still hashes its full text ("compare all code"). Existing NodeType manifest entries re-hash once; the preserve rule makes that pass write-free. |

Member matching is case-insensitive (stored content is camelCased; writers fall back to PascalCase on hubs without a naming policy — see `StampReleaseRequest`); the member list is pinned against the `NodeTypeDefinition` record by test.

Transitional by design: #748 tracks moving the compile state off the NodeTypeDefinition entirely (compile activity + `_Compile` satellite); this PR keeps legacy repo files and stored nodes honest meanwhile and remains correct after the move.

## Verification

- 11 unit cases (`NodeTypeOperationalContentTest`): strip/preserve/token semantics, typed + degraded-JSON shapes, both naming policies, record pinning both directions.
- Full-loop `NodeTypeOwnershipSyncTest` (fake GitHub): export strips; a legacy stale file imports with the authored change landing and the live compile state kept.
- `MeshWeaver.GitSync.Test` 143/143 · `MeshWeaver.Graph.Test` 773/773 · full solution `-c Release -warnaserror` clean.

What's New entry included (deploys of source-only changes work with a plain Update — no force, no recompile ritual).

### DIFF
--- a/src/MeshWeaver.Documentation/Data/WhatsNew/2026-08-02-nodetype-sync-owns-authored-content.md
+++ b/src/MeshWeaver.Documentation/Data/WhatsNew/2026-08-02-nodetype-sync-owns-authored-content.md
@@ -1,0 +1,20 @@
+---
+Name: GitHub sync now compares what you authored, not compile bookkeeping
+Category: What's New
+Description: Type definitions sync on their authored content and code; compile state stays live on the mesh and out of your repo.
+Icon: Sparkle
+---
+
+# GitHub sync now compares what you authored, not compile bookkeeping
+
+A type definition carries two kinds of data: what you authored (its configuration, sources and
+code) and what the mesh recorded about compiling it (status, timestamps, assembly pointers). The
+sync used to treat both the same, with two bad effects: exported repo files carried a compile
+verdict that was stale the moment the next compile ran, and importing such a file stamped that
+stale verdict back over the live type — which could park the type after a restart and made
+plugin-source deploys need a force-sync followed by manual recompiles.
+
+Now the sync knows the difference. Exports commit only the authored definition, imports always
+keep the live compile state (even from older repos whose files still embed it), and change
+detection compares the authored content — including every code file — so a plain "Update to
+latest" deploys a source-only change correctly, with no force and no recompile ritual.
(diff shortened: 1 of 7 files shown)


rbuergi:
Re the review note: `MeshWeaver.Mesh.Services` is required — the `CompilationStatus` enum lives there (`src/MeshWeaver.Mesh.Contract/Services/CompilationStatus.cs`) and is used by `Strip_HandlesTypedContent_WhateverTheNamingPolicy`; the file failed CS0103 without it, and the full solution builds clean under `-warnaserror` with it.

PR #752, opened 2026-08-02
Compile state gains its satellite: {type}/_Activity/compile-state (#748 phase 1)

rbuergi:
Phase 1 of #748 ("move compile state to activity" — maintainer directive, 2026-08-02).

## What lands

- **`NodeTypeCompileState`** — the operational compile members as a value of their own, projected from `NodeTypeDefinition`. The member set is **pinned 1:1 against `NodeTypeOperationalContent.MemberNames` by test (both directions)**, so the satellite and the #749 sync seams cannot drift apart.
- **`NodeTypeCompileStateMirror`** — dual-writes the state onto the fixed-id satellite **`{type}/_Activity/compile-state`**: an Activity node whose `ReturnValue` carries the state (the import-manifest pattern — same table mapping, GitSync exclusion and prune protection for free). Installed on every per-node hub beside the compile watchers; one subscription, `DistinctUntilChanged` on the serialized state, System-identity writes, concatenated so the final satellite equals the final observed state; a failed write logs and retries on the next change.
- **Coverage argument:** every compile-state *change* happens on an active type hub (all writers run through the hub or its own stream), so the mirror cannot miss one. A dormant pre-existing type gets its satellite on next activation — one PreWarm sweep covers the fleet.
- The node stays the writers' target and single source of truth. Phase 2 flips readers to the satellite; phase 3 stops the node writes and retires the mirror (plan in #748).

## Root-caused along the way

`ContentAs<T>` — the canonical shape-tolerant accessor — handled typed, `JsonElement`, and same-short-named foreign content, but returned **silent null for `JsonNode`/`JsonObject`**: exactly the shape node builders assign and a freshly created node's own stream carries before re-typing. The mirror projected nothing for every just-created NodeType until the DOM case was added. This hardens every `ContentAs` call site, not just the mirror.

## Verification

- Unit: member-set pin (both directions), projection fidelity, `IsEmpty`, satellite round-trip through `StateNode`/`Parse`, `ContentAs` on JsonObject content.
- Integration (full monolith mesh): create a NodeType → satellite appears on hub activation with the seeded state; change the node's compile members → the satellite follows.
- `MeshWeaver.Graph.Test` 779/779 · `MeshWeaver.GitSync.Test` 143/143 · full solution `-c Release -warnaserror` clean.

No What's New entry — internal infrastructure, no user-visible behavior change yet.

### DIFF
--- a/src/MeshWeaver.Graph/Configuration/NodeTypeCompileState.cs
+++ b/src/MeshWeaver.Graph/Configuration/NodeTypeCompileState.cs
@@ -1,0 +1,262 @@
+using System.Collections.Immutable;
+using System.Reactive.Disposables;
+using System.Reactive.Linq;
+using System.Text.Json;
+using MeshWeaver.Data;
+using MeshWeaver.Mesh;
+using MeshWeaver.Mesh.Services;
+using MeshWeaver.Mesh.Services.LanguageServer;
+using MeshWeaver.Messaging;
+using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.Logging;
+
+namespace MeshWeaver.Graph.Configuration;
+
+/// <summary>
+/// The MESH-OWNED compile state of a NodeType, as a value of its own — the exact operational
+/// members the framework persists onto <see cref="NodeTypeDefinition"/> today (status,
+/// timestamps, assembly pointers, source-version maps, release-request bookkeeping), projected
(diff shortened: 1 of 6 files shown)


rbuergi:
Both review points addressed in 3c82a5f2a: the catch-alls are simply removed — `GetMeshNode` already maps absence to null, so `Read` and the test's `WaitForState` now let timeouts/denials propagate instead of collapsing them into "no state".

PR #754, opened 2026-08-02
Coupons administration tab: manage Store coupon codes from Settings

rbuergi:
Adds **Settings ▸ Administration ▸ Coupons** (global admins only, gated and grouped exactly like the Plugin Catalog tab):

- **Live coupon list** over `Admin/Coupons/{CODE}` — code, what it unlocks (an empty allow-list reads "any plugin"), price ("free" for null/0), validity window, redemption tally (`used / cap` when capped), notes. The chunked query is accumulated (creates/edits/deletes all land) and the grid re-renders per settled snapshot.
- **Open** buttons per coupon → the node page, where the `Store/Coupon` type's own Edit / Delete actions live.
- **New coupon** → the standard Create flow (`/Store/Coupon/Create?type=Store%2FCoupon`; the type's `defaultNamespace` places it under `Admin/Coupons`).

The coupon type is **dynamic** (Store plugin), so the tab reads content strictly by shape — typed, `JsonElement`, or `JsonNode` — and never references the plugin's compiled types (`ToRow` + `Accumulate` are pure and unit-tested across all three shapes, plus priced/windowed/capped variants).

Registered in the memex portal config beside `AddPluginCatalogSettingsTab`. What's New entry included.

Verification: 5 new unit tests green, full solution `-c Release -warnaserror` clean.

### DIFF
--- a/memex/Memex.Portal.Shared/MemexConfiguration.cs
+++ b/memex/Memex.Portal.Shared/MemexConfiguration.cs
@@ -873,6 +873,9 @@ public static class MemexConfiguration
                         // Plugin Catalog tab (platform admins only) — browse the registry's modules
                         // and Install / Update them onto this instance. Replaces the old Plugins Space.
                         .AddPluginCatalogSettingsTab()
+                        // Coupons tab (platform admins only) — the Store's typed coupon codes at
+                        // Admin/Coupons: live list, redemption tallies, create/open.
+                        .AddCouponAdminSettingsTab()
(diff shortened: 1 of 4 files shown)


rbuergi:
Both findings addressed in ae26f971c: the grid binds its rows inline via `Controls.DataGrid(rows)` (no per-render data ids, no orphaned entries), and the explanatory text uses `Controls.Markdown`.

PR #757, opened 2026-08-02
Roslyn autocomplete in Code-node editors — overlay completions + the kernel's script environment

rbuergi:
Code-node Monaco editors had **diagnostics but no completions**. Two things were missing: the completion plumbing was shaped for `@`-reference *trigger tokens* (it cannot express "complete C# at line/column"), and the language service could only complete against **saved** text — useless while typing.

## What's in here

**1. Overlay completions.** `IMeshLanguageService` gains a `GetCompletions` overload taking the editor's in-flight buffer: the NodeType's solution is forked with that one document substituted (the `CheckSpeculative` shape), so suggestions reflect what is on screen. Forked solutions are immutable, so concurrent suggest requests never contend and nothing is applied back.

**2. A script environment for cells that aren't part of a NodeType compile** — a course lesson's `{lesson}/Source/{cell}`, whose owner is a Markdown page. Those execute on the kernel, so they now *complete* on the kernel's terms: script-kind parsing, its imports and reference set, and the script globals (`Mesh`, `Log`, `Ct`, `Inputs`) in scope. `MeshScriptEnvironment` is that one description and `KernelExecutor` now builds its `ScriptOptions` from the same values — the editor cannot drift from what ▶ Run accepts.

  The dispatch keys on whether the owner **is** a NodeType. "Did compilation inputs resolve" does *not* distinguish them — a plain Code node resolves inputs too — which silently parsed every lesson cell as regular C# and reported a spurious *"top-level statements must be in an executable"* on all of them. `CheckSpeculative` takes the same fork, so lesson cells get honest live squiggles too.

**3. A position-based Monaco provider** (`registerCodeCompletionProvider` + `GetCodeCompletions`), independent of the trigger-token one, with the LSP→Monaco kind mapping at the boundary and a bounded, never-throwing call so a stalled service cannot hang the suggest widget.

## The bug the tests caught

Filtering now happens **before** truncation. Roslyn returns every symbol in scope, alphabetically — a bare `Take(maxResults)` kept the A's and dropped everything after, so typing `Mes` could never reach `Mesh` no matter how the client filtered afterwards. The prefix comes from the completion list's own span; pinned by a test using a small cap and a late-alphabet match.

## Scope change

With the script environment in place, the old opt-in (C# only under a NodeType's `Source/` subtree) has no reason to exclude standalone cells: **every C# Code node now gets language services**, each answered by the environment it actually compiles in.

## Tests

`MeshNodeLanguageServiceTest` +5: overlay completes against proposed text; script completions resolve both imported types (`Controls.Stack`) and the globals (`Mesh`) as bare identifiers; script diagnostics surface real errors and stay quiet on a clean cell (trailing expression, `#r nuget` stripped with line numbers preserved); prefix filtering survives a small cap. 13/13 green, full solution builds.

### DIFF
--- a/src/MeshWeaver.Blazor/Components/Monaco/CodeCompletionRequest.cs
+++ b/src/MeshWeaver.Blazor/Components/Monaco/CodeCompletionRequest.cs
@@ -1,0 +1,10 @@
+namespace MeshWeaver.Blazor.Components.Monaco;
+
+/// <summary>
+/// One code-completion request from the Monaco editor: the FULL in-flight buffer text and the
+/// caret position, already converted to the LSP convention (0-based line and character — Monaco
+/// is 1-based; <c>MonacoEditorView.GetCodeCompletions</c> converts at the boundary). Distinct
+/// from the trigger-token completion path (<c>CompletionCallback</c>), which matches a typed
+/// <c>@</c>-reference query and knows nothing about positions.
+/// </summary>
+public readonly record struct CodeCompletionRequest(string Text, int Line, int Character);
--- a/src/MeshWeaver.Blazor/Components/Monaco/CodeEditorView.razor
+++ b/src/MeshWeaver.Blazor/Components/Monaco/CodeEditorView.razor
@@ -23,7 +23,8 @@
                   CodeEditMode="true"
                   ValueChanged="@OnValueChanged"
                   DiagnosticsCallback="@diagnosticsCallback"
-                  DiagnosticSourcePath="@diagnosticSourcePath" />
+                  DiagnosticSourcePath="@diagnosticSourcePath"
+                  CodeCompletionCallback="@codeCompletionCallback" />
 
 @code {
     private MonacoEditorView? editorRef;
@@ -41,6 +42,7 @@
     // different Code node in the same component lifetime) picks up the new paths.
     private Func<string, IObservable<IReadOnlyList<DiagnosticInfo>>>? diagnosticsCallback;
     private string? diagnosticSourcePath;
+    private Func<CodeCompletionRequest, IObservable<IReadOnlyList<CompletionEntry>>>? codeCompletionCallback;
 
     // Store the data pointer for two-way binding
     private JsonPointerReference? ValuePointer;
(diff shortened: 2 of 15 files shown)


rbuergi:
Added likely-usage ranking (89e3431fd) — alphabetical was the wrong default: pressing `.` led with `Controls.Badge` while every real cell wants `Stack`/`Markdown`/`DataGrid`.

Mirrors how editors actually do it, in two regimes:

- **Typed something → match quality decides**, via Roslyn's own `CompletionService.FilterItems` (exact > prefix > CamelCase hump > substring, honouring target-typed `MatchPriority`). This replaces the hand-rolled StartsWith/Contains sort. Popularity deliberately cannot reorder across match quality, or typing `Bad` would surface the popular `Stack` above `Badge`.
- **Typed nothing → likely usage decides**: identifier frequency across the Code nodes this mesh already runs, plus a 5× locality bonus for identifiers already in the cell being edited. That is VS Code's locality bonus plus the signal IntelliCode models over a public corpus — except our corpus is the MeshWeaver idiom itself, a much better predictor here.

`CompletionUsageIndex` builds that prior once, lazily, in the background, from a bounded query behind a timeout (400 nodes / 4000 identifiers, 30-min TTL). It is never awaited on the request path and every failure resolves to "no prior" — a global `nodeType:` query is the shape that has wedged this mesh before, so a cold or stuck index costs ranking quality and nothing else.

Verified non-vacuous: with the prior disabled, the ranking test fails with *found "Badge"*.

rbuergi:
Added per-user **suggest memory** (2610e7dd4) — VS Code's `editor.suggestSelection` semantics reproduced rather than approximated:

- **recentlyUsedByPrefix** — keyed on the word you had typed when you accepted. VS Code stores prefixes in a trie and looks up the *longest* stored prefix that still prefixes the current word, so accepting `Stack` after typing `St` keeps selecting it at `Sta`; ties break by most-recent.
- **recentlyUsed** — the fallback when no prefix entry applies: among the items actually offered, the one accepted most recently.
- Identity is **kind+label** (the property `Stack` ≠ the class `Stack`), a remembered item not on offer is never selected, and the memory is bounded at 300 acceptances (VS Code's own limit), dropping the least recent.

It **preselects and never reorders** — the matcher's ranking stays exactly as computed and only the highlighted row moves. That is what lets the memory be aggressive without ever hiding a better match. `CompletionEntry` gains `Preselect` (LSP's own field) and Monaco honours it.

Storage follows the established per-user settings-node shape: `{userId}/_Settings/Completions`, a satellite node beside the notification preferences. Writes are **coalesced** behind a 10s per-user throttle — acceptances arrive a keystroke apart, and a write per acceptance is precisely the storm shape this repo has been burned by; a burst of twenty costs one write. Loads are one-shot, empty-on-absent, behind a timeout — never a point probe of a maybe-absent path.

21/21 tests green (5 new pin the memory semantics), full solution builds.

PR #769, opened 2026-08-03
Syncing a space no longer makes you its admin

rbuergi:
Running `git_hub_sync update` on a space **silently granted the caller Admin on it**. Nobody asked for that access and nobody should have it: a GitSynced space is owned by `system-security` by definition, and AGENTS.md says so — *"only system-security writes a GitSynced space; if a human or agent account can still write one, that is a misconfiguration to remove"*.

## The chain

1. The import passes the invoking `userId` down as the writer, so every imported node lands with `CreatedBy = that person`.
2. The partition bootstrap reads `CreatedBy` as **the creator** and, when no grant exists yet, mints them an Admin `AccessAssignment` at `{space}/_Access/{user}_Access` — `MeshExtensions.CreateCreatorGrant`.
3. It already skips exactly this when the creator **is** System (`isRealCreator` is false for System).

So the whole escalation was one identity threaded through a deploy path. Anyone who could trigger a sync became the space's administrator.

## The fix

The import writes as **System**. The **activity** still carries the real `userId`, so who triggered the deploy is recorded exactly as before — only the node writes change hands.

Observed live: syncing `Skill` created `Skill/_Access/rbuergi_Access` (`"rbuergi Access"`, role Admin) at the same second as the call. That grant has been deleted; `Skill/_Access` is back to `Public → Viewer` + `system-security`.

Pairs with MeshWeaver.Plugins#310, which makes `Skill` a proper `preInstalled` public package rather than the bare root that let this surface in the first place.

### DIFF
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -157,24 +157,20 @@
   <ItemGroup>
     <None Include="README.md" Pack="true" PackagePath="\" />
   </ItemGroup>
-  <!-- Override vulnerable transitive dependencies. Forced as direct PackageReference so
-       the version pin in Directory.Packages.props actually applies (transitive deps
-       otherwise resolve to whatever a higher-up package requested).
-       - System.Security.Cryptography.Xml: CVE in 10.0.0 chain
-       - OpenTelemetry.Api: GHSA-g94r-2vxg-569j (1.13.1 vulnerable)
-       - Microsoft.AspNetCore.DataProtection: GHSA-9mv3-2cwr-p262 (10.0.0 vulnerable)
-       - AngleSharp: GHSA-pgww-w46g-26qg (1.4.0 vulnerable — transitive via bUnit); pinned 1.5.2.
-         The Directory.Packages.props pin was inert without a direct ref (no transitive pinning),
-         so AngleSharp resolved to the vulnerable 1.4.0 and tripped NU1902 as an error under
-         -warnaserror (WarningsNotAsErrors does not exempt NuGet audit advisories from the gate).
-       Conditioned on net10.0 because netstandard2.0 projects (e.g. BusinessRules.Generator)
-       opt out of CPM and don't need these packages. -->
-  <ItemGroup Condition="'$(TargetFramework)' == 'net10.0'">
-    <PackageReference Include="System.Security.Cryptography.Xml" />
-    <PackageReference Include="OpenTelemetry.Api" />
-    <PackageReference Include="Microsoft.AspNetCore.DataProtection" />
-    <PackageReference Include="AngleSharp" />
-  </ItemGroup>
+  <!-- No transitive-dependency overrides are needed any more, and that is deliberate.
+       This block used to force patched versions of System.Security.Cryptography.Xml,
+       Microsoft.AspNetCore.DataProtection, OpenTelemetry.Api and AngleSharp. Every one of those
+       advisories came from something we could fix at SOURCE instead:
+         - Microsoft.Identity.Web 4.9.0 dragged in System.Security.Cryptography.Xml 10.0.7;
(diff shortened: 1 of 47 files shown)


rbuergi:
## Do not merge as-is — the tests found a better fix than mine

Two failures, and the first one changes the diagnosis:

**`PartitionRootBootstrapTest.RepairGrantMissing_CreatingChild_RecreatesCreatorGrant`** — the grant-on-child-write is a **deliberate self-heal**: if a partition's owner grant goes missing, writing a child recreates it. My `!rootExists` condition removes that repair wholesale, so this test is right to fail.

The real defect is narrower than this PR frames it. The repair uses **`creator` = whoever is writing**, not *the partition's owner*. On someone's own home those are the same person and the heal is correct. On `Skill` they are not — a deploy writing one activity node was treated as the owner and handed Admin, reproducibly, on every sync.

**So the condition should be owner-vs-writer, not root-absent:** repair only when the writer actually owns the partition — its root's `CreatedBy`, or the partition being their own home. That keeps the self-heal for the case it was built for and closes the escalation.

**`GitHubWebhookProcessorTest.Process_PushEvent_TriggersUpdateToLatestOnMatchingSpaces`** — from the import-as-System half; it asserts the `userId` threaded into `UpdateToLatest`. Either the test's expectation moves to System, or the import keeps the caller id and the whole fix rests on the ownership condition above.

Marked draft rather than merged. Earlier in this session I pushed a second change here (creating the activity node as System) that broke `ActivityOwnerCredentialTest` because it collapsed *attribution* into *authorization*; it is reverted. I'd rather leave this red with an accurate diagnosis than push a third attempt I cannot run the tests against.

**Reproduction, for whoever picks it up:** `git_hub_sync update` on `Skill` (a partition that has existed for weeks) → `Skill/_Access/{caller}_Access` appears with the Admin role, named `"{caller} Access"`, timestamped to the second of the call. Delete it, sync again, it returns.

rbuergi:
## Third attempt reverted — the data to fix this cleanly does not exist yet

I tried the owner-versus-writer condition I proposed above. It fails, and the failure is informative:

```
RepairGrantMissing_CreatingChild_RecreatesCreatorGrant [FAIL]
  Expected value not to be <null> because the bootstrap re-creates a missing creator grant
```

**The root carries no usable owner.** The test creates its partition through `CreateChild`, so the root is written by `ProvisionAndCreateRoot` — under the System identity, not the creator's. So `root.CreatedBy` is not the owner, and "is this writer the owner?" cannot be derived from what is stored today. The self-heal has no owner record to key on; the *grant itself* was the only ownership record, and it is precisely the thing that has gone missing when the heal runs.

So this needs a design decision, not a condition tweak. Two candidates:

1. **An explicit owner on the partition root** (stamped at provision time), which the heal reads. Correct, but a schema/backfill change.
2. **Key off the partition being system-owned** — a GitSynced space has `{partition}/_GitSync`, and AGENTS.md already states only `system-security` writes those. That covers `Skill` and every plugin/course space exactly, and leaves user partitions untouched. Cheaper, and it needs one more authoritative read in a method that already does two.

Option 2 is what I would build next, but I have run out of room to build and verify it, and I have already broken two things today by pushing changes I could not run the tests against.

**State of this branch now:** the bootstrap change is reverted; `PartitionRootBootstrapTest` is 8/8 green again. What remains is the original import-as-System commit. Note that alone does **not** close the escalation — the activity node create is what triggers the grant — so this PR is not worth merging as it stands. It is worth keeping for the diagnosis.

**Reproduction, unchanged:** `git_hub_sync update` on `Skill` → `Skill/_Access/{caller}_Access` with role Admin, timestamped to the second of the call. Delete it, sync again, it returns.

PR #775, opened 2026-08-04
Price models from the catalog node; label the model chip with it

rbuergi:
## The bug

The cost surfaces only ever called `ModelPricing.Default` — the **compiled-in** table. `ModelPricing.Resolve(id, node)`, the method that honours a price authored on a `LanguageModel` node, had **zero callers**. So every model outside that table (any OpenRouter / OpenAI-compatible / BYO-key model) billed at **$0** in the thread token chip and the Token Usage tab, no matter what an admin stamped on its node.

Live on memex today: `z-ai/glm-5.2` and `moonshotai/kimi-k3` both showed $0.

Two smaller things came out of the same look:

- `TokenUsageSettingsTab` used the non-cache-aware `Cost(in, out)` while the chip used the cache-aware overload — the two surfaces disagreed for cache-heavy threads.
- The composer's selected-model chip rendered `LastSegment(path)` ("kimi-k3"), not the node's label that the `/model` picker lists.

## The change

- **`ModelPriceCatalog`** (new, `MeshWeaver.AI`): `LanguageModel` snapshot → `id`/`path` → authored rate, plus `RateFor(model, catalog)` = **node price → built-in table → null**. Null, never a fabricated $0 — callers show tokens without a cost. Content is read through `ContentAs<ModelDefinition>` so typed, `JsonElement` and `JsonObject` all price (an `is JsonElement` test would silently read nothing for two of the three). A node carrying only one of the two prices contributes nothing rather than shadowing the table with a partial rate.
- **`ThreadTokenChip`** and **`TokenUsageSettingsTab`** resolve through it, off the same live model query the picker uses. The catalog is an **enrichment, not a gate**: seeded empty and `Catch`-ed, so a slow or unreadable catalog degrades pricing to the built-ins instead of blanking the usage display.
- **`TokenUsageSettingsTab`** now bills cache reads/writes at their own rates, matching the chip.
- **`ModelInfo.Label`** carries the model node's `Name`, so the composer chip shows the picker's own text (`Kimi K3 · $3.00/$15.00`) and falls back to the last path segment when there's no node/label.
- OpenRouter's two deployed ids added to `ModelPricing.Defaults` as a backstop (org/model slugs are matched whole, before the last-segment fallback).

## Tests

17 new units in `ModelPriceCatalogTest` — node-price-wins, both id shapes (bare wire id, `org/model` slug, node path), all three content shapes, the null-not-$0 rule, deterministic duplicate-id resolution, cache-aware cost, and the `ToModelInfo` label fallbacks.

**Verified non-vacuous**: neutering the catalog lookup in `RateFor` fails `AuthoredNodePrice_BeatsTheBuiltInTable` and `NodePath_ResolvesToTheSameRateAsTheBareId`; the two shape tests were then re-pinned to off-table prices so they can't pass via the built-ins either.

Local runs: `MeshWeaver.AI.Test` 938 passed / 0 failed · `Memex.Portal.Shared.Test` 222 passed · Monolith `AgentPickerProjection`/`LanguageModel` filters 10 passed. Clean builds (warnings-as-errors) of `MeshWeaver.AI`, `MeshWeaver.Blazor.Portal`, `Memex.Portal.Shared`.


rbuergi:
Triage of the two Copilot findings before merge: both flag the `RateFor(...) ?? 0m` collapse of unknown-model pricing to $0.00. That pattern is pre-existing on main (`ModelPricing.Default(usage.Model)?.Cost(...) ?? 0m` in both ThreadTokenChip.razor.cs and TokenUsageSettingsTab.cs) — this PR does not introduce it, it strictly narrows the set of models that hit it (live catalog + compiled defaults instead of defaults only). Rendering unknown cost as blank rather than $0 is a fair follow-up enhancement, not a blocker for this strict improvement. Merging on green CI.

PR #784, opened 2026-08-04
CI: cut the run from ~24 min to ~14 min

rbuergi:
Stacked on #782 (same file) — review/merge that first.

Measured on run [30903428924](https://github.com/Systemorph/MeshWeaver/actions/runs/30903428924) (green: 23.7 min wall, 72 min runner time):

```
build 9.3 min ──▶ slowest shard 13.8 min ──▶ collector 0.3 min
```

| Shard | free-disk | **download** | extract | tests | total |
|---|---:|---:|---:|---:|---:|
| 2 | 61 | **35** | 25 | 355 | 503 |
| 3 | 64 | **162** | 25 | 240 | 522 |
| 4 | 79 | **143** | 50 | 247 | 552 |
| 0 | 66 | **177** | 26 | 286 | 589 |
| 1 | 62 | **147** | 48 | 445 | 743 |
| 5 | 78 | **318** | 25 | 370 | **827** |

## 1. The shard imbalance was never the tests — it was the artifact

One 3.3 GB artifact, downloaded by all six shards = **20 GB of transfer per run**, and the artifact service's throughput varies ~10× run to run: **35 s … 318 s for the same bytes**. The long-pole shard spent 318 s downloading and a perfectly ordinary 370 s testing; the fastest shard fetched the identical bytes in 35 s.

The build now packs **one tarball per shard** and each shard downloads only its own (~1/6).

The assignment moves to `.github/scripts/shard-assign.sh` because **both** jobs need it now — build to decide what goes in each artifact, test to decide what to run. Two inline copies could silently diverge, and a shard would then try to run a project whose bin it never downloaded. Portable on purpose (no `declare -A`), so it verifies locally:

```bash
SHARD_INDEX=0 SHARD_TOTAL=6 .github/scripts/shard-assign.sh
```

## 2. The weight table had drifted

Measured 2026-07-13; ~100 s of long pole had rotted into it — Monolith 200→**345**, Data 25→**73**, Persistence 165→**97**, Security 90→**34**, PluginTester 60→**8**.

Re-measured as the **max over two green runs** (30903428924, 30903158378) so a shard is sized for the bad case, not the average. LPT now balances to **345 / 340 / 340 / 339 / 339 / 339** (measured spread was 240–445). 345 is `Hosting.Monolith.Test` alone — the floor for any sharding.

## 3. `Free disk space` ran 78 s × 7 jobs

Dropped the two sub-steps that don't pay for themselves:
- **docker-images** prunes the runner's pre-baked images (node, buildpack-deps), none of which is the pgvector image Testcontainers then pulls anyway;
- **swap-storage** deletes the swapfile on jobs running memory-heavy hosts (`Persistence.Test` carries a memory watchdog) and reclaimed **0 B** on the measured run.

Kept android (11 GiB) + haskell (3.6 GiB). A `df -h` probe after extract now reports headroom, so we can decide from data whether the rest can go too.

## Safety

The shard **fails loudly** if `shard-assign` yields no projects. `set +e` is in force there, so a broken script would otherwise leave the loop with nothing to run and the shard would report **green having run zero tests** — exactly the masked-failure class the exit-marker gate exists to catch.

## Verified

- `actionlint` clean, `shellcheck` clean.
- Assignment verified locally: **all 62 projects covered exactly once, zero duplicates**, loads 345/340/340/339/339/339.
- The macOS-awk portability trap is fixed (a multi-line `awk -v` value works on the runner's gawk but errors on macOS awk — the `-v` form would have "worked in CI" while being unverifiable locally).

## Not done

Caching the NuGet restore (39 s). Restoring a multi-GB packages folder plausibly costs as much as it saves; I'd rather not put an unmeasured change on the critical path. Worth a measurement of its own.

Splitting the three heavyweights (Monolith 345 + AI 263 + Orleans 239 = 43% of all test time) would lower the 345 s floor, but it needs class-level splitting across processes and `xunit.runner.json` pins `maxParallelThreads: 1` — semantics-changing, so not bundled here.


rbuergi:
Triage (session working the open-PR queue): merged current main into the branch (8bd2cd3f) after the earlier author-side merge (299d4720) — the branch is not stale against the #796 shard split.

Two consecutive CI runs then failed on ONE test each — a different test on a different shard each time:

- run 30998033944, shard 4: `VersionHistoryTest.RollbackNode_RestoresHistoricalState` — "Test execution timed out after 20000 milliseconds".
- run 31000712491, shard 3: `ExportImportAccessControlTest.Export_EditorFullAccess_ExportsWholeSubtree` — "Expected the observable to emit a value within 60s" (+ soft watchdog at 60s).

The branch's src diff is pid-stamping in trace lines only (MonolithMeshTestBase.TestTrace / MessageHub.DisposeTrace) and cannot plausibly cause either timeout. Both look like timing races surfacing under the new per-shard artifact layout (different download/extract timing → different load profile per shard). Per AGENTS.md these flakes are real races to pin, not to retry away — so not merging on a red suite, and not re-running to fish for green. Leaving OPEN: the CI work itself (per-shard tarballs, ~1/6 artifact download per shard) remains valuable and is not superseded.

PR #794, opened 2026-08-04
Node CRUD off the router: handling ops for your own node ≠ being a work queue

rbuergi:
## The defect

**The mesh hub is the mesh's ROUTER, and every node CRUD in the process was targeted at it.** `MeshService` posted to `hub.GetMeshHub()` because that is where `WithNodeOperationHandlers` lives — so creates, deletes and moves executed on the router's own action block, competing with the routing it exists to do.

A burst starves `SubscribeRequest` traffic and every node op then times out. That is the atioz 2026-06-11 portal-wide wedge, whose signature is recorded in `StaticRepoImporter`:

> "11× CreateOrUpdateNodeRequest + 3× `CreateNodeRequest@mesh/<self>` stale >60s while real user SubscribeRequests starved"

**And the existing mitigation was already defeated.** `StaticRepoImporter` builds a dedicated `import/{id}` hub precisely to keep bulk creates off the router — but anything writing through `IMeshService` walked straight back up to the mesh root, so the isolation bought nothing.

## What changed since the first attempt — and why it was red

The first version of this PR resolved the target as *"the nearest ancestor that registered `WithNodeOperationHandlers`, other than the router."* That is wrong, and CI proved it: `EducationStartExerciseTest` failed with

```
Access denied: user 'learner-user' lacks Create permission on 'TestCourse/Module1/Ex1'
```

Two facts collide:

1. **Every per-node hub registers the handlers** — `MeshDataSource` calls `WithNodeOperationHandlers()` — because a node hub must be able to *receive* ops addressed to its own node (the delete fan-out posts `DeleteNodeRequest` to `new Address(path)`; the create handler forwards its `Argument` the same way). So "nearest handling ancestor" resolved to **the calling hub itself** for the single biggest class of caller.
2. **Per-node hubs are exactly the hubs carrying `AddAccessControlPipeline`**, and `CreateNodePermissionAttribute` anchors its check at the **receiving hub's** path. Retargeting a create onto the caller's own node hub therefore evaluated `Create` against whatever node the caller happened to be *rendering* — a learner copying a course module into their own partition was denied on the read-only source.

That is silent privilege reduction, and it is the exact hazard `EducationLayoutAreas.EnsurePersonalCopy` documents in its xmldoc. It also meant the change did not move that traffic off the router at all — it moved it onto the node's own hub.

**Fix: two markers, because they are two capabilities.** Handling node ops *for your own node* says nothing about being a good place to *execute someone else's write*. `WithNodeOperationExecution()` is the new, explicit opt-in that `NodeOperationTarget` looks for; `WithNodeOperationHandlers()` keeps its original meaning and stays on every node hub. Anything that did not opt in keeps falling back to the router exactly as before — which is still the common case (a per-node hub rendering a layout area), so that traffic is untouched.

## The change

- **`MeshExtensions.WithNodeOperationExecution`** — declares a hub a node-CRUD execution target (implies the handlers; idempotent).
- **`MeshExtensions.NodeOperationTarget`** — walks to the nearest ancestor carrying that marker and not the router; falls back to the mesh root otherwise.
- **`MeshService` / `NodeCopyHelper` / `VUserHelper`** resolve through it instead of `GetMeshHub()`.
- **Opted in**: the dedicated `import/{id}` hub (the wedge source — this is what finally makes its isolation real), the MCP session hub, the Blazor portal hub. None of the three carries a per-node access pipeline.
- **`StaticRepoImport.md`** updated: its claim that the handlers alone kept bulk upserts off the router was not true for anything routed through `IMeshService`.

The self-reference guard from the first version is kept — `GetMeshHub()` has always carried `!ReferenceEquals(parent, current)`, and without it the walk spins forever on the CRUD path (an infinite loop that presents as a hang, not an error).

## Tests

- `NodeOperationTargetTest` — an opted-in hub targets itself and never a `mesh` address; a hub with no opt-in falls back to the mesh root; a child hub walks up to its nearest declaring ancestor; **and a hub with the handlers but no execution opt-in still falls back to the router** — the regression above, pinned.
- `EducationStartExerciseTest` — reproduced red on the first version, green now (31/31 in `MeshWeaver.AI.Test`'s education class).

The detector half of the original PR shipped separately as #802 (`RouterTrafficRule`), so remaining router-targeted paths — `WebhookInboxEndpoints` writes through an injected root hub, and is not addressed here — stay visible as ERRORs rather than silently regressing.

## Release note

Skipped deliberately: this is internal message routing with no user-visible surface.


rbuergi:
**Held as draft — do not merge as-is.** The diagnosis stands, the mechanism is tested, but retargeting CRUD relocates two invariants the mesh hub was silently providing:

1. **The action-block hop.** `mesh/` was always a different action block from the caller; the nearest handling hub often is the caller's own, so the handler runs on its turn. (`ActivityLiveProgressTest` pins this — under Orleans that turn is the grain activation.)
2. **The identity the write runs under.** This is the blocking one: `EducationStartExerciseTest` fails with `Access denied: user 'learner-user' lacks Create permission on 'TestCourse/Module1/Ex1'`. A write that previously ran elevated now runs as the caller — **silent privilege reduction**, breaking learner course installs.

The detector half is split out to **#802**, which is safe to ship on its own and will tell us from production which paths actually hit the router. The migration then follows the data instead of one CI symptom.

Note for whoever picks this up: the earlier shard-5 `ActivityLiveProgressTest` failure on this branch was **not** caused by it — #793 ("Pin the off-turn hop at the seam, not at a thread id — main is red") landed mid-flight; after merging main it passes 3/3 locally.

rbuergi:
Triage (session working the open-PR queue):

**The merge conflict was mechanical and is resolved** — pushed as d49f5a124. Both sides carried a version of the ReportRouterTraffic detector; main's refined one (RouterTrafficRule extraction, landed separately) supersedes the branch's inline version, so MessageHub.cs resolves to main's side byte-for-byte while the branch's actual retargeting work (NodeOperationTarget, MeshService, SessionHubResolver, PortalApplication + test) merged untouched. Local Release/-warnaserror builds of all touched projects are clean; NodeOperationTargetTest 3/3 and RouterTrafficRuleTest 8/8 pass locally.

**But CI on the merged tree fails 3 tests, and they look SEMANTIC, not flaky** (run 31001137070):

1. `EducationStartExerciseTest.StartExercise_Learner_CopiesModuleToOwnNamespace_AndRedirectsToLearn` — the StartExercise area renders an error instead of the redirect: `Access denied: user 'learner-user' lacks Create permission on 'TestCourse/Module1/Ex1'`. Retargeting node CRUD off the router changes WHICH hub executes the learner's module-copy create, and with it the access-context under which the permission check runs.
2. `InboxToolIntegrationTest.CheckInbox_TwoCallsBackToBack_SecondReturnsEmpty` — first call already returns `(no new messages)` where it should find one; notification write/visibility ordering appears changed by the retarget.
3. `NodeHubContentCollectionTest.CollectionNamedArea_UrlDecodesEncodedPathSegments` (shard 1).

These are routing-behavior decisions (where a create executes and whose identity it carries), so they belong to this PR's author rather than mechanical triage. Leaving the PR open with the merge pushed so the failures reproduce on a current tree.

rbuergi:
Triage from a re-run (2026-08-05): this one is **not** a flake — unlike #784, which went green on re-run and has been merged.

**`MeshWeaver.AI.Test.EducationStartExerciseTest.StartExercise` fails on this branch and passes on current `main`** (verified locally against `main`: 31 passed / 0 failed, 9s). Shard 0 red at 16m; `Consolidate test results` red downstream of it.

That points at this PR's change rather than CI infrastructure — plausibly the routing change altering which hub handles the exercise-start CRUD, which is exactly the surface this PR moves.

Leaving the PR open and unmerged. Worth keeping: the `CreateNodeRequest@mesh/<self>` leaked-callback shape this PR addresses is the same one behind the disposal wedge seen earlier today, so the change is still wanted — it just needs that test resolved first.

rbuergi:
Picked this up and it is now green-able rather than closeable — the diagnosis held, the predicate did not.

**Reproduced the blocker locally first** (merged onto current `main`): `EducationStartExerciseTest` fails with `Access denied: user 'learner-user' lacks Create permission on 'TestCourse/Module1/Ex1'`. Confirmed true positive, and the mechanism is structural rather than test-specific:

1. `MeshDataSource` calls `WithNodeOperationHandlers()` on **every per-node hub** — it has to, because a node hub must receive ops addressed to its own node (`FanOutDeleteSubtree` posts `DeleteNodeRequest` to `new Address(path)`; `HandleCreateNodeRequest` forwards its `Argument` the same way). So "nearest ancestor with the handlers" resolved to **the calling hub itself** for the dominant caller class — meaning the change did not move that traffic off the router at all, it moved it onto the node's own hub.
2. Per-node hubs are exactly the ones carrying `AddAccessControlPipeline` (`SecurityServiceExtensions.cs:68`), and `CreateNodePermissionAttribute.GetPermissionChecks` yields `(hubPath, permission)` — anchored at the **receiving hub**. So a learner copying a course module into their own partition had `Create` evaluated on the read-only source they were merely rendering. Your note #2 called this exactly right; this is where it comes from. It is also the hazard `EducationLayoutAreas.EnsurePersonalCopy`'s xmldoc already warns about.

**Fix: two markers, because they are two capabilities.** Handling ops for *your* node says nothing about being a good place to execute *someone else's* write. `WithNodeOperationExecution()` is the explicit opt-in `NodeOperationTarget` now looks for; `WithNodeOperationHandlers()` keeps its meaning and stays on every node hub. Anything that did not opt in falls back to the router exactly as before — so the per-node-hub traffic that broke is untouched, and your concern #1 (the action-block hop) no longer applies to it either.

Opted in: the `import/{id}` hub, the MCP session hub, the Blazor portal hub — none carries a per-node access pipeline. **The import hub is the real prize**: it exists precisely to keep bulk creates off the router, but `IMeshService` posted to `GetMeshHub()` unconditionally, so its isolation bought nothing. `StaticRepoImport.md` asserted otherwise and is corrected. `NodeCopyHelper` and `VUserHelper` hard-coded `GetMeshHub()` for the same reason and now resolve through the same contract.

**Verified locally on the merged tree:**
- `EducationStartExerciseTest` — red before, **31/31 green** after
- `MeshWeaver.Hosting.Monolith.Test` — 471 passed / 0 failed / 2 skipped
- `MeshWeaver.Graph.Test` — 802 / 0
- `NodeCopyHelperTest` + `NodeOperationTargetTest` — 13 / 0 (re-run after the helper edits)
- `DocumentationLinkIntegrityTest` — 1 / 0
- `dotnet build MeshWeaver.slnx -c Release -p:CIRun=true -warnaserror` — **Build succeeded**, zero warnings

`NodeOperationTargetTest` gains the case that pins the regression: a hub with the handlers but no execution opt-in still falls back to the router.

`WebhookInboxEndpoints` still writes through an injected root hub — unchanged here, and #802's `RouterTrafficRule` keeps naming it.

PR #814, opened 2026-08-05
A projected query must answer: never hand a dictionary to Query<MeshNode>

rbuergi:
## The bug

**Every mesh query carrying a `select:` clause hung.** Measured on memex 2026-08-05:

| Query | Result |
|---|---|
| `nodeType:AccessAssignment scope:subtree limit:20` | instant, 20 rows |
| `nodeType:AccessAssignment scope:subtree select:path limit:50` | hang > 120 s |
| `nodeType:NodeType select:path limit:5` (certainly matches) | hang > 120 s |

**The database was never involved.** Measured in-cluster: the 136-schema fan-out is **78 ms**, a 136-branch `UNION ALL` with `ORDER BY … LIMIT 50` is **57 ms**, and every one of the 136 schemas has the `node_type` index.

`select:` makes the query layer project each row to a `Dictionary<string,object>`. That **is** the contract for the untyped surface — `QuerySyntaxTests.Select_SingleProperty` pins it — but a dictionary is not a `MeshNode`, and both typed paths break on it:

- **`MeshQuery`** (the aggregator, the path `mesh.Query<MeshNode>` takes) does `(T)(object)ProjectToSelect(...)` → `InvalidCastException` for **every** item, thrown *inside the merge* where the fault reaches no subscriber. The provider then emits neither an `Initial` nor an error, the all-providers Initial gate starves, and the caller **hangs in total silence** — no error, no empty result, no log line.
- **`PostgreSqlMeshQuery.CollectQueryResultsAsync`** does `if (item is T typed)` → every projected row **silently dropped**: a query that matched returned empty.

So the MCP `search` tool — which is `Query<MeshNode>` — wedged on any projected query. And the documented best practice was a deadlock: `Doc/Architecture/CqrsAndContentAccess` tells callers to add a `select:` clause to avoid whole-node loads.

## The fix

Skip the **object-level** projection when the caller is typed on `MeshNode`, at both boundaries. Nothing is lost — `select:` still narrows the SELECT in SQL, so a projected read costs only what it asked for; the row simply stays a node for the caller that asked for nodes. Untyped callers still get their dictionary.

## Tests

`ProjectedSelectQueryTests` (new): `select:path`, `select:path,version` and `select:path,name` all **answer** — each threw `InvalidCastException` or returned empty before this change; an unprojected query is unchanged; and the untyped surface still projects to a dictionary, so the fix cannot have been achieved by abolishing projections.

- **PostgreSql suite: 611 passed / 0 failed** (1 skipped).

## Note on scope

The deeper issue Roland flagged — `async`/`IAsyncEnumerable` in the query pipeline, where a fault goes silent instead of erroring — is **not** addressed here; that's a dedicated refactor (measured surface: 148 `IAsyncEnumerable` occurrences across 50 files). This change makes projected queries work; it does not make the pipeline reactive. The `IIoPool.Invoke` boundary these paths use is the one sanctioned by `.claude/skills/async`.

### DIFF
--- a/src/MeshWeaver.Hosting.PostgreSql/PostgreSqlMeshQuery.cs
+++ b/src/MeshWeaver.Hosting.PostgreSql/PostgreSqlMeshQuery.cs
@@ -214,10 +214,19 @@ public class PostgreSqlMeshQuery : IMeshQueryProvider, IVectorSearchProvider
     /// <see cref="CollectQueryResultsAsync{T}"/>), never handed to a caller
     /// whose <c>await foreach</c> would pump it on a hub/grain scheduler.
     /// </summary>
+    /// <param name="projectSelect">
+    /// Whether a <c>select:</c> clause also projects each row to a
+    /// <c>Dictionary&lt;string,object&gt;</c>. TRUE for the untyped surface, whose callers expect
+    /// exactly that (pinned by <c>QuerySyntaxTests.Select_SingleProperty</c>). FALSE for a typed
+    /// <c>Query&lt;MeshNode&gt;</c>: a dictionary is not a MeshNode, so projecting there produced
+    /// rows the typed layer could not accept — and it dropped or mis-cast every one of them.
+    /// The SELECT is narrowed in SQL either way, so nothing extra is fetched.
+    /// </param>
     private async IAsyncEnumerable<object> QueryAsync(
         MeshQueryRequest request,
         JsonSerializerOptions options,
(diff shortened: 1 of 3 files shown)


rbuergi:
Triage before merge: CI suite is green (MeshWeaver Build and Test: SUCCESS). The one Copilot finding is a test-coverage suggestion (also exercise the MeshQuery aggregator path where the original hang occurred), not a defect in the change — leaving the thread open as a follow-up signal for the author rather than resolving it. Merging on green.

PR #846, opened 2026-08-06
Agent working files and skills land in the mesh, not on a disk that disappears

rbuergi:
## What

Microsoft Agent Framework 1.0 turned every stateful part of an agent loop into a **pluggable store**. We already build on its `ChatClientAgent`, and we were hand-rolling the pieces it now abstracts. This implements two of those abstractions over **mesh nodes**.

- **`MeshNodeAgentFileStore : AgentFileStore`** — an agent's working files become mesh nodes under a per-thread root (`Markdown` for files, `Group` for directories). MAF's own stores are a directory on disk or a dictionary; both are invisible to the platform and gone with the process. Here they are ordinary content: versioned, permissioned, visible in the portal.
- **`MeshAgentSkillsSource : AgentSkillsSource`** — our existing `nodeType:Skill` nodes served through MAF's interface. Authoring and storage are unchanged.
- **`MeshStoreAccess`** — composed by both. It owns the three things that must be true of every store call and that fail **silently** when they are not: pooled work, the caller's identity, and the canonical reactive mesh surfaces.

## Constraints honoured

| Constraint | How |
|---|---|
| API 100% reactive | Everything we own returns `IObservable<T>`. `Task` appears only in the `*Async` members MAF declares abstract, each a one-line adapter. No `async`/`await`. |
| All I/O in the pool | New `IoPoolNames.AgentStore`, **not** `Ai` — a store call runs inside a tool call inside a round already holding an `Ai` slot; re-entering the same bounded pool from within a slot it holds is a nested-gate deadlock. `Once` (one-shot) vs `Stream` (`SubscribeThroughPool`) for live feeds, which would otherwise leak a slot per call. |
| `stream.Update` for writes | Overwrite uses `GetMeshNodeStream(path).Update(...)`. Creation cannot — see below. |
| `GetQuery` / `GetMeshNodeStream` | Listings and searches are the live synced `GetQuery` (whole nodes incl. content, so search needs **no** per-node reads). Live single-node reads are `GetMeshNodeStream`. |
| CQRS | `SearchAsync` deliberately unimplemented. |
| Tool-call architecture unchanged | `AgentFilesPlugin` is an ordinary `IAgentPlugin` in the existing `ResolvePluginTools` switch. No harness loop, no MAF tool injection. |

## Three decisions worth review

1. **Writes split by lifecycle.** `stream.Update` is the mutation API and overwrite uses it. Creation *cannot*: for a path that does not exist the per-node hub never activates and the write returns `DeliveryFailure: No node found` (the tests proved this). Creation therefore uses `CreateNode`.
2. **`SearchAsync` throws `NotSupportedException`.** Content search is a query, and a query here is a live change feed. MAF's signature can only carry the one-shot snapshot that is stale the moment it returns. `Search` returns `IObservable<IReadOnlyCollection<FileSearchResult>>` and keeps re-emitting. Nothing in MeshWeaver calls the override; it exists only because the base declares it abstract.
3. **`ProbeNode` uses `GetMeshNode`, not `GetMeshNodeStream(...).Take(1)`.** A missing node's stream does not go quiet, it **faults** with a NotFound `DeliveryFailure` — so a timeout never fires. `GetMeshNode` already encodes the distinction: absent → `null`, denied → error, too slow → `TimeoutException`.

## Skill discovery

Keeps the platform's layering — **1.** the user's own, **2.** the context node's partition, **3.** the node type's partition, then platform defaults — via a new `BuildSkillSubtreeQueries` at the canonical `AgentPickerProjection` site, so there is no parallel reconstruction. Layers 2 and 3 now resolve over the **whole partition subtree**, so a space or plugin can author a skill next to the content it belongs to instead of flattening everything into one namespace.

`BuildSkillQuery` is untouched — **the pickers behave exactly as before.**

## Verification

- `dotnet build MeshWeaver.slnx -c Release -p:CIRun=true -warnaserror` → clean (also re-run after merging `origin/main`)
- Full `MeshWeaver.AI.Test`: **1007 passed, 0 failed**, 5 skipped
- 28 new tests pass; `DocumentationLinkIntegrityTest` passes

## Also

Bumps `Microsoft.Agents.AI` 1.16.0 → 1.17.0 (the three packages ship in lockstep), where these abstractions reached their current shape. Adds an architecture doc page and a What's New entry.

https://claude.ai/code/session_01Um7g7UaNMvB1yqdZNM3K4U


rbuergi:
## Update — search implemented, and a real GUI/agent inconsistency fixed

**`SearchAsync` is now implemented**, in the same shape as every other MAF override: `Search(...).Take(1).ToTask(ct)`, one expression, no `async` state machine. `Search` itself stays live (`IObservable<IReadOnlyCollection<FileSearchResult>>`); the `.Take(1)` sits at the MAF boundary where every other single-value override puts it. The earlier `NotSupportedException` is gone — uniformity beat the special case.

**Tooling is covered at both levels.** Unit tests now hit every override and every reactive method, including three **liveness** tests that subscribe *before* the change (a one-shot implementation would hang them), recursive vs non-recursive search, and `CreateDirectoryAsync` idempotence. `AgentFilesToolingTest` (Playwright, host Ollama) walks the whole surface — write → list → search → read → delete — in one conversation, asserting the rendered tool result each turn **and** cross-checking the mesh: the node must exist after the write and be gone after the delete. A second test pins containment: a new conversation cannot see the previous one's files, and the original survives.

**The substantive fix: skill discovery had two definitions.** The GUI resolved `AiSettingsNodeType.DefaultSkillQueryTemplates`; the new agent-framework source resolved `AgentPickerProjection.BuildSkillQueries`. A user would have seen one set of skills in the slash menu while an agent round resolved another — silently. Both now express the same layering, and `MeshAgentSkillsSource` resolves through `AiSettingsNodeType.ObserveSkillQueries` — the exact call the slash autocomplete makes — so user-configured sources and installed skill packages reach the agent framework too. A parameterised test pins the two equal across seven context shapes.

### ⚠️ One finding worth a second opinion

Moving the platform query row from **first** to **last** broke slash autocomplete outright — `SkillAutocompleteTest` timed out with no `/agent`, `/model`, `/harness`. I isolated that single variable to confirm it's the ordering, not the subtree widening.

The platform row appears to be the only one guaranteed to resolve; every other targets a partition that may not exist. I restored it to first — the pre-existing working contract — and pinned it with `PlatformDefaults_AreAlwaysFirst`, with a comment stating *why* and that row order is **not** the precedence signal (precedence is still resolved from each result's own partition, so a user's skill still overrides a platform skill of the same name).

**I deliberately did not chase the underlying behaviour.** If a union query whose first row targets a non-existent partition genuinely fails to converge, that would silently empty *any* picker built on a multi-row union — a framework-level issue well outside this PR.

### Verification

- `dotnet build MeshWeaver.slnx -c Release -p:CIRun=true -warnaserror` → clean, after merging current `main`
- `MeshWeaver.AI.Test`: **1022 passed, 0 failed**, 5 skipped
- `DocumentationLinkIntegrityTest` passes

Four existing tests encoded the old flat skill contract and are updated — a deliberate change to skill discovery, not incidental.

rbuergi:
## Review addressed — three real defects fixed, one deliberately not

Thanks — the regex one in particular would have shipped.

**1. ReDoS (fixed).** `SearchCore` compiled a `Regex` straight from the **model's** pattern with no bound. Catastrophic backtracking would have burned a bounded `AgentStore` pool slot and a core for as long as it ran. Every match is now bounded (2s), and a pattern that trips it surfaces as an error the tool reports rather than a silent empty result. Pinned by a test using the classic `(a+)+$` over a long non-matching run — it genuinely throws `RegexMatchTimeoutException`.

**2. `Delete` removed directories (fixed).** It deleted whatever node sat at the path while `Exists()` carefully distinguishes file from directory. A "delete this file" that silently removed a folder — orphaning everything filed under it — isn't a defensible reading. It now reports `false` for a directory, exactly as for an absent file.

**3. `CreateDirectory` over a file reported success (fixed).** It returned the existing node whatever its type, so `mkdir("x")` "succeeded" with `x` still a file, and the next write beneath it would land somewhere the caller never asked for. Now a conflict.

**4. Stale pool comment (fixed).** The `AgentStore` cap was justified by "a `SearchAsync` over a large store reads many nodes" — untrue once search became one live query carrying content, and doubly so now that `SearchAsync` is implemented. Re-stated around what actually holds slots.

### Not changed: the `MeshStoreAccess` fallbacks

`?? IoPool.Unbounded` is the established convention at every pool call site in the codebase (`WebSearchPlugin`, the Cosmos and Blob storage adapters, `RegistryPackageSource`, `InstanceOAuthService`) and exists for construction outside DI. Diverging here would be the inconsistency.

On the principal: a null `AccessContext` does **not** fail open. `PostPipeline` fails closed when no context is set — the write is denied, not escalated. The "silently stamp hub-self as principal" fallback that *would* have failed open was deleted in May. So the fallback is a no-identity path that denies, not a bypass.

PR #860, opened 2026-08-06
A data change reports through an observable, not a hub-per-write Activity

rbuergi:
## The defect

**Every `DataChangeRequest` constructed an `Activity` — and an `Activity` is not a log, it is a HUB.** Its constructor calls `GetHostedHub` (`Activity.cs:38`), and `StartSubActivity` hosts one more per data source (`Activity.cs:148`), wired together with a C# `event` plus an `autoClose` heuristic that fires when "all sub-activities are no longer Running".

That whole apparatus answered one question — *tell me when the writes committed, and hand me the accumulated messages* — which is exactly what `IObservable<ActivityLog>` expresses natively. It predates the reactive API.

**It was already vestigial.** Of ~20 `RequestChange` call sites, all but three passed `activity: null` — including the mainline cross-hub node write (`ApplyMeshNodePatchInTurn`, `DataExtensions.cs:1259`). The three holdouts were `HandleDataChangeRequest` and the unified-reference update/delete. Meanwhile the validation-failure path already built its `ActivityLog` **by hand, with no Activity at all** (`DataExtensions.cs:1636`) — the proof that the log works standalone — and the sub-activity tree it produced was read by exactly one thing: a `LogDebug` line.

Per Roland's decision 2026-08-04 (*"I'm ok taking the activity wrap away — it is overkill"*): an activity is the unit of **intent** (import, GitSync, compile — a persisted node with live progress), never a per-write record.

## The change

- **`IWorkspace.RequestChange` / `Update` / `Delete` drop the `Activity?` parameter and return `IObservable<ActivityLog>`.** The write is still issued **EAGERLY on call**, so a caller that ignores the result still writes — no call site can silently stop writing, which is why this is not the usual cold `RequireSubscribeObservable` write shape. The observable *reports*: one finished `ActivityLog` after every affected stream applied its part, replayed to late subscribers.
- **Per-stream completion is an `AsyncSubject` scheduled on the stream hub's NEXT turn.** The transform's result is applied by the same turn that runs it, so signalling inline would report "committed" — and post the `DataChangeResponse` — while the store still held the pre-change state. That is ack-on-accept, the shape `ApplyMeshNodePatchInTurn` already documents as having raced read-after-write.
- **`Activity` is untouched** and still used for intent-level work. `ImportManager` folds the save's log into the import activity instead of passing the activity down.

### Three behaviours that were wrong and are now right

1. **Validation failures were dropped for every `activity: null` caller** — they went only to the Activity. They now always land on the returned log, with one consistent `"<members> invalid: <error>"` shape (the three branches previously each formatted differently, one of them as `"Validation error in activityId {message}"`).
2. **A failed stream write reported Succeeded.** The old exception path completed the sub-activity with no error message, so the activity rolled up to Succeeded → `DataChangeStatus.Committed`. It now reports Failed. A **disposed** stream stays a Warning — the benign teardown marker the stream classifies as Debug-only — and Warning still commits.
3. **The activity hub got a synthetic "Succeeded" response** (`activity is null` → post Succeeded immediately), because an Activity created on an activity hub would have recursed. With no Activity involved, every hub reports its real log.

`DataChangeResponse` is unchanged, so consumers reading `.Log` / `.Status` see the same shape — now with the real status behind it.

## Tests

- **`DataChangeLogTest`** (new) pins the contract: reports exactly once then completes; a validation failure is reported *and not applied*; and a `DataChangeRequest` leaves the host hub's activity-typed hosted-hub count unchanged — the direct pin on "no hub per write".
- Green locally, Release + `-warnaserror`: solution build, **Monolith 471** (2 pre-existing skips), **Graph 812**, **Data 304**, **GitSync 153** (the activity/`ActivityLiveProgress` suites), **Layout 343**, **Import 21**, doc link integrity.

## Notes for review

- `MeshWeaver.Reinsurance`'s **`legacy/`** tree has 5 call sites passing an Activity to `RequestChange`. That tree has no solution file and its CI only compile-checks source nodes, so nothing builds it today — but it would need the one-line update if revived.
- Docs updated: `AsynchronousCalls.md` (the read-modify-write example) and `DataMesh/CRUD.md` signatures.

## Release note

Skipped deliberately: internal plumbing with no user-visible surface (same call as #794).

https://claude.ai/code/session_01S3GuX6pWYWHMCP7NbWbcf9

### DIFF
--- a/src/MeshWeaver.Data/DataExtensions.cs
+++ b/src/MeshWeaver.Data/DataExtensions.cs
@@ -1256,8 +1256,7 @@ public static class DataExtensions
                     // Route via the hub's DataChangeRequest pipeline — the workspace
                     // writes through the data-source stream (which owns the typed
                     // InstanceCollection + persistence + reduction fan-out).
-                    hub.GetWorkspace().RequestChange(
-                        DataChangeRequest.Update([merged]), null);
+                    hub.GetWorkspace().RequestChange(DataChangeRequest.Update([merged]));
                 }
                 catch (Exception ex)
                 {
@@ -1654,7 +1653,7 @@ public static class DataExtensions
                 // Every hub now reports its REAL log, including the activity hub, which used to be
                 // handed a synthetic "Succeeded" because an Activity there would have recursed.
                 var changeSubscription = hub.GetWorkspace()
-                    .RequestChange(changeRequest with { ChangedBy = changeRequest.ChangedBy }, request)
+                    .RequestChange(changeRequest with { ChangedBy = changeRequest.ChangedBy })
                     .Select(log => isSatellite || string.IsNullOrEmpty(hubPath)
                         ? log
                         : log with { AffectedPaths = log.AffectedPaths.Add(hubPath) })
@@ -2393,7 +2392,7 @@ public static class DataExtensions
             ChangedBy = changedBy
         };
 
-        return workspace.RequestChange(changeRequest, null)
+        return workspace.RequestChange(changeRequest)
             .Select(log =>
             {
                 hub.ServiceProvider.GetService<ActivityLogBundler>()
@@ -2547,7 +2546,7 @@ public static class DataExtensions
                     ChangedBy = changedBy
                 };
 
-                return workspace.RequestChange(changeRequest, null)
+                return workspace.RequestChange(changeRequest)
                     .Select(log =>
                     {
                         hub.ServiceProvider.GetService<ActivityLogBundler>()
--- a/src/MeshWeaver.Data/DataExtensions.cs
+++ b/src/MeshWeaver.Data/DataExtensions.cs
@@ -1257,7 +1257,7 @@ public static class DataExtensions
                     // writes through the data-source stream (which owns the typed
                     // InstanceCollection + persistence + reduction fan-out).
                     hub.GetWorkspace().RequestChange(
-                        DataChangeRequest.Update([merged]), null, null);
+                        DataChangeRequest.Update([merged]), null);
                 }
                 catch (Exception ex)
                 {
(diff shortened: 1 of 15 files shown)


rbuergi:
Thanks — three of the five were correct and are fixed in `ac61db07c`; two I verified rather than changed.

**Fixed**

- **`UpdateFailed`'s throw removed.** Correct, and stronger than stated: *every* invocation of a stream `exceptionCallback` is wrapped by `SynchronizationStream` in a log-only try/catch (`SignalDisposedToProducer` and the `UpdateStreamRequest` handler), so that `DataException` could reach no caller at all — it only produced a secondary "exceptionCallback threw" ERROR. The failure now travels in the returned log, which is what callers read. `UpdateFailed` is deleted, and with it the `IMessageDelivery? request` parameter it was the only consumer of (it ignored the delivery anyway).
- **`ImportManager` subscriptions** both take an `onError` that logs into the import activity.
- **`_ =` on the `SynchronizationStream` call site** — applied, matching the test call sites.

**Verified, not changed**

- **`.Synchronize()` before `Aggregate`:** not needed — Rx's `Merge` already serializes downstream notifications per the observer grammar. Rather than argue it, I ran a throwaway probe: 200 rounds × 32 `AsyncSubject`s completed from parallel `Task.Run`s through the exact `Merge().Aggregate(...)` shape; every round aggregated all 32 items. Probe deleted, not committed.
- **The synchronous throw for a missing / uninitialised partition stream stays.** It fires at *call* time, and that is precisely what `ApplyMeshNodePatchInTurn`'s `try/catch` around `RequestChange` depends on to ack the patch `false`. Converting it to a log message on the returned observable would leave that patch unacked — a hang rather than an error. It is also unchanged pre-existing behaviour.

**Beyond the review** — the first version reported completion via `stream.Hub.InvokeAsync`, i.e. one extra hub message per data change on the framework's hottest path. `UpdateStreamRequest` now carries an optional `Applied` callback that the handler invokes in the *same turn* right after `SetCurrent`: after the apply (so still no ack-on-accept), and free. That is the completion seam the old `Activity.autoClose` was faking.

PR #861, opened 2026-08-06
Kernel: a control result is rendered, not also dumped as text

sierragolflima:
## The defect

A cell whose last expression is a **control** printed the control's record `ToString()` into its own output pane, beside the thing it describes:

```
StackControl { Id = , Style = , Readonly = , Skins = System.Collections.Immutable.ImmutableList`1[MeshWeaver.Layout.Skin], Class = , IsClickable = False, DataContext = , … }
✓ Done
INFO
StackControl { Id = , Style = , … }
```

Reported on <https://memex.meshweaver.cloud/RiskTransfer/01-GrossToNet>, under the rendered "The 2025 Loss Book" grid.

## The cause

`KernelExecutor` handled a non-null result twice:

```csharp
UpdateView(viewId, returnValue);                                       // renders it — correct
scriptOutputLogger.LogInformation("{Value}", returnValue.ToString());  // …and dumps it as text
```

## The fix

Guard the log on the result not being a control. The log line stays for everything else — for `1 + 1` it is the only output a learner sees — so this checks the type rather than removing the line. `IUiControl` is the interface `UiControl` implements, so it covers every control.

## The test asserts both directions

Silencing the logger outright would suppress the dump and quietly break every scalar cell. A one-sided test would pass on that wrong fix, so `KernelControlResultNotLoggedTest` pins the distinction, measured on one run:

| | Control cell log | Scalar cell log |
|---|---|---|
| **Without the fix** | `StackControl { Id = , Style = , … }` | `scalar-result-42-marker` |
| **With the fix** | *(empty)* | `scalar-result-42-marker` |

### One thing the test encodes deliberately

Each cell gets its **own activity node**, and the assertion waits for a **terminal status** before reading. The activity log is **replaced per submission, not accumulated** — two cells on one activity leaves only the second cell's log. An earlier draft of this test passed twice against a log that never contained the value under test, so that reasoning is written into the test summary rather than left to be rediscovered.

## Verification

- `dotnet build src/MeshWeaver.Kernel.Hub -c Release -warnaserror` — clean, before and after merging current `main`
- The test fails against unfixed code and passes with the fix, re-confirmed on the merged tree
- Branch merged with current `main` (12 commits) before opening

## Not addressed here

Whether these cells should be executable on a read-only course page at all is a separate design question and is deliberately left open.


sierragolflima:
Closing without merging. The approach here is wrong, and the reason is worth recording.

This suppressed the result log when the value implements `IUiControl`. That fixed the notebook cell and **broke script execution**: `ScriptExecutionInUserHomeTest` failed twice — once on the return value no longer landing as the terminal activity-log message, once on the message count dropping from 5 to 4.

That behaviour is a documented contract, not an incidental assertion. `ScriptExecution.md` states it:

> **Log the result, not just "done"** — include counts, paths, durations. The activity log is the primary post-mortem surface.

For a script there is no rendered surface, so the log line is the only output the run produces. The decision therefore cannot be made from the value's **type** — both callers legitimately return controls, and only one of them has somewhere to draw.

Narrowing it by **caller** is the shape that matches reality, but there is no existing discriminator: both the script path (`CodeNodeType.cs:223`) and the cell path (`ExecutionManager.cs:29`) set `Id`. That makes it an API decision rather than a bug fix, which does not belong inside a red pull request.

Continued in #876, with reproduction steps and four candidate approaches.

The test added here, `KernelControlResultNotLoggedTest`, is sound and worth keeping — it pins both directions on one run. The branch `fix/kernel-uicontrol-tostring-log` is left in place as a starting point.

PR #871, opened 2026-08-06
CI cannot report green on zero evidence, and a starved job must not block delivery

rbuergi:
## What today's "flakes" actually were

I went looking for flaky tests across the last 14 failed CI runs. Across all of them there were **two** test-level failures, one occurrence each. Everything else was **GitHub Actions runner starvation**: jobs sat **15 minutes with zero steps executed** and were then cancelled — main's `Check for an already-green tree` (twice) and `Build solution (once)` on #868.

The pipeline turned that infrastructure event into both **false reds** and, worse, a **false green**.

## 1. The skip was fail-CLOSED into silence

A `needs` job that ends `cancelled` skips every dependent **regardless of the dependent's `if`**, unless the condition contains `always()` / `!cancelled()`. So a starved probe silently skipped `build` and all six shards.

`build` is now `!cancelled() && needs.precheck.outputs.skip != 'true'`: an unanswerable probe means *"we don't know"* ⇒ **test it**. Only a genuine workflow cancellation (where `!cancelled()` is false) still stops the run.

## 2. The required check passed with no tests run 🚨

`Consolidate test results` is the single required check (ruleset `main pr protection`). Its trx gate exits 0 when it finds no failures — and **"no trx at all" is not a failure**, so a run in which *nothing executed* published a **green required check**. Verified on main run `31120801452`: probe cancelled, build skipped, all shards skipped, `Consolidate test results: success`.

That is how an untested tree can satisfy the merge gate and ship. It is strictly worse than a flaky check, because it fails silently in the safe-looking direction. The gate now requires **positive evidence**: at least one `.trx`, or the sanctioned green-tree reuse path.

## 3. Delivery keyed on a conclusion any cancelled job poisons

`main-cd.yml` gated on `workflow_run.conclusion == 'success'`. One cancelled job makes the run `failure` even when everything that ran was green — so **main built no image from ~14:50 onward** and the pull-based self-update sat on its last image (CD shows `skipped` on `e365679a7`, `6343b8218`, …).

CD now runs a `gate` job that reads the **required check for the head SHA** and only then builds images. After #2 that check cannot be green without evidence, so this is strictly stronger than the umbrella conclusion — and immune to a starved job.

## VersionHistoryTest — bounds that contradicted each other

`RollbackNode_RestoresHistoricalState` timed out on main shard 4 with `Test execution timed out after 20000 milliseconds`, naming nothing. Cause: its awaits ran on `Emit()`'s 10 s default and the rollback poll claimed `ReadNodeTimeout` — **60 s** — inside a `[Fact(Timeout = 20000)]`. A bound three times the method timeout can never fire, so the method timeout always won and the failure was undiagnosable by construction.

Every wait in the class now uses one declared **4 s step budget** (~10× the observed per-operation cost — 246 tests in 96 s) and names what it was waiting for. This is a *narrowed* bound, not a widened one: a stuck step now fails fast and says which step.

## Not in this PR

`PaywallRealGateShapeTests` — the other test-level failure — is already root-caused and fixed in **#868** (`MeshNodeListPathEquality` compared assignment snapshots by path set while the fold reads content, so `DistinctUntilChanged` permanently discarded content-only changes; in prod a revoked permission kept working until reload). Its CI was one of the runs killed by the runner starvation above; I re-triggered it.

## Tests

- `MeshWeaver.Content.Test`: **245 passed**, 1 pre-existing skip.
- Both workflow files validated as YAML; job graph unchanged apart from the new `gate`.
- `alert-on-failure` still behaves: image jobs *skip* when the gate is not green, and `failure()` does not fire on skipped.

## Release note

Skipped deliberately: CI/delivery plumbing, no user-visible surface.

https://claude.ai/code/session_01S3GuX6pWYWHMCP7NbWbcf9

### DIFF
--- a/.github/workflows/dotnet-test.yml
+++ b/.github/workflows/dotnet-test.yml
@@ -190,10 +190,18 @@ jobs:
   build:
     name: "Build solution (once)"
     needs: precheck
-    # Skipped when this exact tree is already green (see the file header). `test`
+    # Skipped ONLY when this exact tree is already green (see the file header). `test`
     # needs `build`, so it skips with it; `collect-results` is if:always() and
     # reports the reused result.
-    if: needs.precheck.outputs.skip != 'true'
+    #
+    # 🚨 `!cancelled()` is what makes the skip FAIL-OPEN. A `needs` job that ends
+    # cancelled — the probe waiting 15 min for a runner and being killed with zero
+    # steps executed, main 2026-08-06 — skips every dependent REGARDLESS of this `if`,
+    # unless the condition contains always()/!cancelled(). That silently skipped the
+    # build and all six shards, and the run still reported a green required check.
+    # With !cancelled(), an unanswerable probe means "we don't know" ⇒ TEST IT; only a
+    # genuine workflow cancellation (where !cancelled() is false) still stops the run.
+    if: ${{ !cancelled() && needs.precheck.outputs.skip != 'true' }}
     timeout-minutes: 25
     runs-on: ubuntu-latest
     steps:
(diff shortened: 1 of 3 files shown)


rbuergi:
**Auto-merge deliberately DISARMED on this PR** (was armed at open).

While this PR sat waiting for a runner, #868 auto-merged **through the exact hole this PR closes**:

```
Build solution (once):      cancelled  steps=0   18:54:01 → 19:09:02   (never got a runner)
Run tests (shard …):        skipped
Consolidate test results:   success    25s                             ← required check, ZERO tests run
→ auto-merged 19:09:37Z
```

That is defect #2 in the description, live: the one required check reporting green on no evidence, and a PR merging untested. (#868's content was verified locally 28× by the session that wrote it, so the outcome there is fine — the mechanism is not.)

A PR whose entire point is "never trust an evidence-free green" must not merge on one. This waits for a run that actually executes the suite. GitHub Actions is in a declared major outage right now (githubstatus.com, 18:46 UTC: *"Workflow runs are still failing, and jobs may remain queued for an extended period before starting or may time out"*), and repo-wide no `MeshWeaver Build and Test` run has started since 17:45 — so re-triggering now only burns another cycle.

Re-arm with `gh pr merge 871 --auto --merge` once Actions is healthy, or merge normally on a green run.

PR #907, opened 2026-08-07
Gate /static behind the same access check as the canonical read path (#587)

### LINKED ISSUE #587: Hardening: /static/ endpoint serves content collections without authentication or access control
## Summary

## Current behavior

## Why this matters

## Proposed hardening

## Affected code

rbuergi:
## Summary

Fixes #587 — `/static/**` served content collections with **no authentication and no access control**.

## Three distinct holes (all in `BlazorHostingExtensions`)

1. **No authorization anywhere.** `ResolveStatic → ServeFromCollection → ServeFile` resolved a collection and streamed the file. The `/static/storage/content/{node}/{file}` form reads the mesh-level store directly — every partition's uploads flat and world-readable at a **fully predictable** URL (`MeshNodeImageHelper`, `MarkdownFileParser` and `ContentAssetMapper` all build that scheme).
2. **`IsStatic` was declared but read nowhere** — set at 2 sites, consulted at 0 — so every collection registered on the mesh hub was servable by URL.
3. **A cache short-circuit bypass.** The only brush with access control in the address pattern was the collection-config lookup (`GetDataRequest` carries `[RequiresPermission(Read)]`), and `collectionCache.TryGetValue` skipped it on a hit. **Once any caller warmed the cache, every later caller — anonymous included — bypassed the check entirely.**

Found while building the gate: `DecodeContentPath` un-escapes `%2E%2E` **after** the server's URL normalization, and `FileSystemStreamProvider` does a bare `Path.Combine` — so `content/PublicSpace/%2E%2E/PrivateSpace/secret.pdf` would borrow PublicSpace's anonymous grant to read another partition. Guarded.

## The gate

Two explicit declarations, **both defaulting closed**:

- **`IsStatic`** — is the collection mounted on `/static` at all? Not declared ⇒ 404, identically for every caller (a hosting decision must not vary with identity).
- **`IsPublic`** (new) — is that mount world-readable? Not declared ⇒ gated.

**It matches the canonical read path rather than inventing rules.** The ordinary content read posts `GetDataRequest`, whose `[RequiresPermission(Read)]` pipeline checks the owning hub's own path; the gate calls `hub.GetEffectivePermissions(ownerPath, userId)` through the same `PermissionEvaluator`. Partition/RLS scoping, group expansion, `PublicRead` policies and the paywall's per-subject deny/allow therefore apply verbatim. Owner attribution goes through `IPathResolver` (longest node prefix), so a file under `{Course}/{Lesson}` attributes to the lesson and the per-child deny bites, while the cover under `{Course}` stays anonymous; Read folds hierarchically, so deepest-node attribution is never weaker than the root. Fail-open only where `AccessControlPipeline` itself does (no `EffectivePermissionsDelegate` registered) — absence of policy, not bypass of one.

Also: gated content is now `private, max-age=300, must-revalidate` (was `public, immutable` — a CDN or proxy would have kept replaying it), and the large-file streaming path, which previously carried **no** `Cache-Control` at all, is classified too. 401 anonymous / 403 authenticated, mirroring the sibling `CourseAssetEndpoints`.

Deliberately public: `NodeTypeIcons` (compiled into MeshWeaver.Graph, needed before any identity exists — login page, nav, every anonymous card) and `DocContent` (shipped in the nuget package; the Doc partition already carries Anonymous/Public Read grants).

## Verification

`StaticContentAccessTest` drives **real HTTP over TestServer against a real mesh** seeded with the production paywall shape — 11 tests: anonymous denied, signed-in-without-grant forbidden, public cover still anonymous, gated lesson denied to a visitor and served to the entitled buyer, explicitly-public collection served, gated content never shared-cacheable, unattributable file denied, unmounted collection not served, plus both address-pattern cases including *"warming the collection cache does not open the gate"*. `StaticContentGateRulesTest` adds 17 pure-rule tests (attribution, traversal/quote guard, cache classification).

**Fail-without:** disabling the gate flipped 5 of 11 to failures — *"Expected Unauthorized because an anonymous caller holds no grant on PrivateSpace — the URL is not the access control, but found OK"* — and disabling the second pattern failed both `AddressPattern_*` with 0 passed.

**Pass-with:** 28/28 here, ContentCollections.Test 28/28, Content.Test 245 passed / 1 pre-existing skip, Documentation.Test 19/19. Independently re-verified **28/28 after merging current main**. `-c Release -warnaserror` clean on every touched project.

## 🚨 Deployment risks — please review before merge

1. **Login-page logos.** `CorporateIdentity.LogoPath` / `BrandingResolver` build `/static/storage/content/{node}/…`, now gated on Read of the owning node. Correct — but if a portal's logo lives in a space without an Anonymous grant, **the login page shows a broken image**. The fix is a grant or a public collection, not code. Worth a prod smoke-check right after deploy.
2. **`Memex.LocalMesh` now serves public mounts only** — that sidecar resolves no identity at all, so anything else would be ungated. If it was relied on for non-public assets, those 404.
3. **Per-file gating below the owning node is not expressible** where content does not mirror the node tree (e.g. lesson media flat under `content/{Space}/videos/`); those attribute to the Space. `_Entitlements`-gated course media is served by the separate `/assets` endpoint and is unaffected.
4. **5 s `GrantWait` on denial** (matching `CourseAssetEndpoints`) — a denied request holds a request slot that long. The positive-shaped wait is required because the permission fold's first emission can be a premature empty seed.
5. No docs written — `AccessControl.md` has no section on the `/static` contract; worth adding as a follow-up.

No What's New entry: security hardening; no intended user-visible change beyond correctly denying what was never meant to be public.

### DIFF
--- a/memex/Memex.LocalMesh/Program.cs
+++ b/memex/Memex.LocalMesh/Program.cs
@@ -85,6 +85,14 @@ app.MapGet("/static/{**path}", async (HttpContext ctx, string path) =>
     var filePath = string.Join('/', path[(slash + 1)..].Split('/').Select(Uri.UnescapeDataString));
     var content = app.Services.GetRequiredService<IMessageHub>().ServiceProvider.GetService<IContentService>();
     if (content is null) return Results.NotFound();
+    // 🚨 PUBLIC MOUNTS ONLY (issue #587). This sidecar resolves no identity at all — there is no
+    // UserContextMiddleware and no caller to attribute a read to — so it may serve exactly the
+    // collections that are declared both mounted on /static AND world-readable, i.e. the shipped
+    // icon/doc assets this endpoint exists for. Anything access-controlled (user content,
+    // attachments, per-Space collections) is not servable here and answers 404; it is reachable
+    // through the portal's gated /static endpoint instead.
+    var config = content.GetCollectionConfig(collection);
+    if (config is not { IsStatic: true, IsPublic: true }) return Results.NotFound();
     // HTTP-boundary await: the read leaf runs on the collection's IIoPool; the endpoint awaits
     // the replayed single emission only.
     var stream = await content.GetContent(collection, filePath).Take(1);
--- a/memex/Memex.Portal.Shared/MemexConfiguration.cs
+++ b/memex/Memex.Portal.Shared/MemexConfiguration.cs
@@ -832,6 +832,16 @@ public static class MemexConfiguration
                             Name = "content",
                             IsEditable = true,
                             ExposeInChildren = true,
+                            // Mounted on /static: images, thumbnails and PDFs in this Space are
+                            // fetched as /static/{Space}/content/{file} (MeshNodeLayoutAreas'
+                            // image renderer, ContentLayoutArea's PDF viewer) and as
+                            // /static/storage/content/{Space}/{file} through the mesh-level store.
+                            // The mount does NOT make them public — every request is attributed to
+                            // the owning node and gated on Read (issue #587).
+                            IsStatic = true,
+                            // Never inherited from the Storage config section: user content is
+                            // access-controlled, full stop.
+                            IsPublic = false,
                             BasePath = basePath,
                             Settings = contentStorageConfig.Settings is { } src
                                 ? new Dictionary<string, string>(src) { ["BasePath"] = basePath }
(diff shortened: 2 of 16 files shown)


rbuergi:
🚧 **Do not merge — being reworked.** Maintainer direction: we do NOT want any protected content accessible under `/static` at all. Gating it (checking permission per request) is the wrong shape; `/static` must serve ONLY public content and must not carry a permission fold.

Reworking to: a collection is either mounted on `/static` — in which case it is public, cacheable, and never permission-checked — or it is not mounted there at all (404 for everyone), and protected content is served exclusively by the authenticated content endpoint. The `IsPublic`-gating half comes out; the `IsStatic` default-closed half and the `%2E%2E` traversal guard stay.

rbuergi:
Superseded by the unmount-not-gate rework — see the new PR. Its gate design was rejected: protected content must be unreachable under /static, not reachable-then-denied.

PR #910, opened 2026-08-07
Change feed dispatches off the publisher's thread — kills the delete deadlock (#899)

### LINKED ISSUE #899: Per-leaf delete of an _Access grant can park forever in the permission check — recursive space delete wedges (post-#894, 1-in-N)
Observed on PR #896 run 31161307727 shard 5 (`SpaceMenuAndAccessTest.Space_ChildrenAreQueryable`), commit includes #894 via main. Intermittent — main's own post-#894 run and #894's PR run were green.

## Evidence (teardown snapshot from the failing run)

```
Hub TestSpace_b29902812e/_Access/Roland_Access RunLevel=Started Disposal=Pending
  Queue(buffer=1, deferred=0, exec=0, deliveryActionCompleted=False)
  Executing(DeleteNodeRequest, 12897ms)
```

Two per-leaf `DeleteNodeRequest`s (`{space}/Overview`, `{space}/_Access/Roland_Access`) pending 10.4 s at the mesh hub; the access-grant leaf hub has been EXECUTING its delete for 12.9 s with the next delivery queued behind the single-threaded action block. The recursive space delete therefore never completes; the test's query wait starves and the class fails with leaked callbacks.

## Why this path is new

Before #894, recursive delete never visited `_Access` satellites at all (the #839 leak) — this per-leaf shape (delete an AccessAssignment on its own per-node hub, mid-space-teardown) first became reachable when the authoritative plan started enumerating satellites. The last-admin validator refusal was fixed with `CascadeRootPath` (exempt, fast); the remaining park is downstream of validation.

## Prime suspect

`CheckDeletePermissionForNode` → `hub.GetEffectivePermissions(pathToCheck, userId).Take(1)` where `pathToCheck = node.MainNode ?? node.Path` — for an access satellite, MainNode is the SPACE ROOT, i.e. the permission fold is evaluated against the very subtree being torn down, mid-delete, on the leaf's own hub. If any source stream of that fold (the `$security-access:{scope}` cached queries / scope walk) fails to produce its Initial in that state, `Take(1)` parks forever — the #539/#667 'mid-teardown owner never answers' family, now on the delete path itself.

Per the wedges-to-zero rule the fix is the non-emitting source (or bounding the permission check with an explicit, loud timeout that fails the delete rather than parking the hub) — not a widened budget.

rbuergi:
## Summary

Fixes #899 — a recursive delete could park forever, posting **no response at all** (neither success nor failure), wedging both per-node hubs. It had become a general merge blocker: it redded shard 5 on unrelated PRs (#905, #907) and reproduced locally 2-3 times in 12 full-project runs.

## Root cause — a two-thread Rx lock-order inversion, NOT a non-completing read

The earlier narrowing on this issue (a non-completing `Read` inside `PersistenceService.Delete`'s containment probe) is **disproven**: `storage.Delete` does emit. The call that never returns is `changeFeed.Publish(...)` in `DeleteAndPublish` (`StorageAdapterChangeFeedExtensions.cs:126`) → `InProcessMeshChangeFeed.Publish` → `Subject<T>.OnNext`, i.e. **synchronous fan-out on the publisher's thread**. Probe on a failing run: `ADAPTER-DELETE-SUBSCRIBE` → `ADAPTER-DELETE-EMIT` → nothing.

`dotnet-stack` on the wedged process showed exactly two blocked threads, both under `InProcessMeshChangeFeed.Publish`, at `Monitor.Enter` on *different* Rx gates (a shared synced-query `Merge` gate, and a `CombineLatestObserver`). A stack probe supplied the missing half: below `HandleDeleteNodeRequest` sit `CombineLatest…Run` → `ReplaySubject.Subscribe` → `CombineLatestObserver.OnNext` — that is `PermissionEvaluator.GetEffectivePermissions`'s `enriched` fold emitting **synchronously during `Subscribe`** (its three sources are cached `ReplaySubject`s).

Because `CheckDeletePermissionForNode` is `GetEffectivePermissions().Take(1).SelectMany(<entire delete pipeline>)`, the **whole delete** — storage delete, cache invalidation, change-feed publish — executes inside that `CombineLatest` gate. With two per-node hubs deleting concurrently:

- **T1** holds its own fold gate `G1`, publishes → takes the **shared** merge gate `M` → walks downstream into T2's fold → blocks on `G2`.
- **T2** holds `G2`, publishes → blocks on `M`.

Cycle. Both hub action blocks park permanently; the operation's 30 s `.Timeout` later posts a failure from a timer thread, but the threads stay blocked forever.

## Fix

`InProcessMeshChangeFeed` now fans out on **one dedicated serial dispatch loop**, never the publisher's thread: publishers push into a `Subject.Synchronize`d inbox; an `ObserveOn(EventLoopScheduler)` pump (background thread `mesh-change-feed`) drives the subscriber-facing subject. `Publish` only enqueues, so **a publisher can never acquire a foreign Rx gate while holding its own — the inversion is impossible by construction, regardless of what subscribers do.** This mirrors what `OrleansMeshChangeFeed` already does for the cross-silo half. Ordering preserved (single FIFO loop); a throwing subscriber is logged at Error instead of tearing down the bus.

No timeout, no retry, no watchdog was added.

## Verification

**Deterministic test** — `MeshChangeFeedDispatchTests` (new, 3 tests) encodes the exact cycle (two publishers each holding their own gate at a `Barrier`; the subscriber walks a shared gate into the other publisher's gate), plus the off-thread invariant and the FIFO contract.

**Fail-without / pass-with** (same binaries, only `MeshChangeFeed.cs` swapped): pre-fix **2 failed / 1 passed** (deadlock test times out at 10 s); post-fix **112/112** for the whole project.

**Soak** of the wedging project (`MeshWeaver.NodeOperations.Test`, full project, Release):

| build | result |
|---|---|
| baseline (probes only) | **2 FAIL / 12** |
| baseline under extra load | **3 FAIL / 5** |
| with fix | **0 FAIL / 16** |
| with fix, probes removed | **0 FAIL / 16** |

Independently re-verified after merging current main: dispatch tests **3/3**, and **6 × 85/85** clean full-project runs.

Also: full solution `dotnet build MeshWeaver.slnx -c Release -p:CIRun=true -warnaserror` → **0 warnings, 0 errors**; Query 361/361, Graph 814/814, Data 303/303, PathResolution 123/123, Hosting 112/112, NodeOperations 85/85, PG `SpaceDeletionPartitionDropTests` 1/1.

## Notes for review

1. **In-process change-feed delivery is now asynchronous** (ordered, microseconds later) instead of inline. This matches the distributed deployment — the Orleans cross-silo relay was always async — and the delete path already invalidates `IMeshNodeStreamCache` and disposes the per-node hub explicitly. Five tests that assumed same-thread delivery were adapted to wait on the actual condition (no sleeps); **two of those were genuine regressions introduced by the fix** and were verified passing on the pre-fix baseline before adapting.
2. **The other half of the cycle is deliberately untouched**: `GetEffectivePermissions(...).Take(1).SelectMany(<whole handler>)` still runs the entire delete inside the permission fold's gate. That is no longer sufficient for a deadlock via the change feed, but the same shape exists on `IStorageAdapter.Changes`. No wedge observed there in 32 post-fix runs; noted as the nearest remaining inversion candidate.
3. One extra background thread per mesh (`mesh-change-feed`, `IsBackground`), disposed with the feed; Monolith teardown still reports "teardown clean".
4. **Pre-existing red on main, unrelated:** `StaleActivationSeedRollbackTest.PostRecycleWrite_LandsAboveTheDurableVersion_AndDoesNotRollTheStoreBack` fails identically with `MeshChangeFeed.cs` reverted to `HEAD` (`Expected 2 to be greater than 7000`) — a reactivating hub seeding a stale snapshot. Filed separately.

Closes #899


rbuergi:
🚧 **Not merging — the async-delivery blast radius is wider than the five tests adapted.** Run 31172266455: shards **0, 1, 2 and 4** red. Sampled failures:

- `Hosting.Orleans.Test.GrainActivationFailureRegistryTest.ChangeFeedBroadcast_ClearsStoredActivationError_ForExactlyThatPath`
- `Hosting.Orleans.Test.OrleansAutoExecuteTest.AutoExecute_UpdateThreadMessageContent_RoutesToResponseGrain`
- `Hosting.Monolith.Test.SearchResultStormTest.CatalogSubscription_ContentUpdateStorm_AddsNoExtraRerenders`

This is residual risk #1 in the PR body materialising at scale: moving in-process fan-out off the publisher's thread makes delivery ordered-but-asynchronous, and more callers than expected assumed **same-thread, already-delivered-on-return** semantics. The local sweep (Hosting, NodeOperations, Graph, Data, Query, PathResolution, PG) did not include the **Orleans** suites, which is where two of the three samples sit.

The fix itself is not in question — the deadlock is proven by `dotnet-stack` and the soak went 2/12 → 0/16 twice. What has to be settled before this can merge is each failing call site: does it have a genuine ordering requirement (⇒ it must wait on the condition, like the five already adapted), or is it asserting an implementation detail of synchronous fan-out (⇒ the assertion is wrong)?

Next step: enumerate every failure across shards 0/1/2/4 and classify them one by one. Do NOT blanket-adapt the tests to make CI green — two of the five adapted so far were *genuine regressions this fix introduced*, and the same could be true here; each one needs the pre-fix-baseline check that distinguishes the two cases.

rbuergi:
Superseded by #918 (merged): the generator-route fix. The async-delivery approach here was disproven by CI (three synchronous cache-invalidation contracts).

PR #911, opened 2026-08-07
/static serves build assets only; all mesh content moves to /api/content (#587)

### LINKED ISSUE #587: Hardening: /static/ endpoint serves content collections without authentication or access control

rbuergi:
## Summary

Fixes #587. **Supersedes the gate in #907, which was rejected**: protected content must be *unreachable* under `/static`, not reachable-then-denied. `/static` now has **no access control at all** — because nothing that needs protecting is served there.

Two surfaces with opposite contracts:

- **`/static/{mount}/{file}` — build assets only.** Backed by a new `StaticAssetMount(Segment, Assembly, ResourcePrefix)` read straight from the assembly manifest. No `IContentService`, no `IPathResolver`, no hub post, no identity resolution, **no permission check anywhere**. `public, max-age=30d, immutable`. `UserContextMiddleware` re-excludes `/static/` — it must not *have* a caller to check.
- **`/api/content/{node}[/{collection}]/{file}` — all mesh content.** Resolves the owning node with `IPathResolver`, then posts `GetDataRequest` (which carries `[RequiresPermission(Read)]`) to **that node's hub** with the caller's `AccessContext`. `private, no-store`.

The `collectionCache` short-circuit is **deleted** — it was a standing bypass: the config lookup was the only permission-bearing hop, and a cache hit skipped it entirely.

## What stays on `/static`, and why that is safe by construction

| Mount | Justification |
|---|---|
| `NodeTypeIcons` | 42 SVGs `EmbeddedResource`-compiled into `MeshWeaver.Graph`. No user data, identical in every deployment, needed before any identity exists (login page, nav, anonymous cards). ~457 existing references + DB rows (migration V41) keep working unchanged. |
| `DocContent` | Images embedded in `MeshWeaver.Documentation`, shipped on nuget.org. Same class; ~50 references unchanged. |

Both are **build output, not mesh content**. Serving them from the assembly manifest rather than a collection is what lets `/static` be access-control-free *by construction* instead of by policy. Both remain registered as in-mesh collections for listings/reads, but neither declares `IsStatic`, so neither is fetchable on the content route either.

## Unmounted

- **Mesh-level `storage`** — `IsStatic` removed in `Memex.Portal.Monolith` and `aspire/Memex.Portal.Distributed`. This is the collection holding **every partition's content** at `content/{nodePath}/…` — the #587 headline hole. Now unreachable over HTTP.
- **`Memex.LocalMesh`'s own `/static`** — was a reduced copy that resolved *any* collection with **no identity resolution at all**. Build-asset mounts only now.
- `IsStatic` (default-closed) is now actually **read**, as the content route's mount check.

## URL generators redirected

`MeshNodeImageHelper` · `MeshNodeThumbnailControl` · `MeshNodeLayoutAreas.RenderImageAsync` · `MarkdownFileParser` (resolve + serialize strip) · `ContentLayoutArea.RenderPdf` · `DocumentLayoutAreas` · `FileBrowser` (both hand-duplicated download links) · `ContentCollectionsExtensions.AdaptResourceUrl` · **`MarkdownExtensions.ToStaticHref` → `ToContentHref`** — the biggest miss, and **#907 never touched it**: it rewrites *every image in every markdown document* · `BrandingResolver`/`ExportTemplateResolver` (readers, now via `ContentPathReference`, still accepting the legacy prefix persisted on nodes) · 5 sample layout areas · `WithThumbnailsPath` in Northwind/Todo · 200 authored literals across 190 sample/doc/test-data files · 3 marketing `imageLink`s · 6 stale doc pages · `content/ai/Skill/space.md` + `Agent/ToolsReference.md` (agent guidance that told models to write `/static` logos) · plugin-catalog fixture · deploy runbook.

Post-sweep, `/static` survives only as: route/prefix declarations, `NodeTypeIcons`/`DocContent`, prose describing the contract, and two deliberate legacy-read constants.

## Verification

- **`StaticContentUnmountedTest`** (15, real HTTP over TestServer against a real mesh): the backing store is **404 for anonymous AND for a fully-entitled admin** — unreachable, not denied; icons serve identically to both with `public`/`immutable`; the same file **is** served to an authorized caller via `/api/content`; an unpublished collection 404s even for an admin; a gated response is never `public`.
- **`ContentUrlGenerationTest`** (25) pins every generator + the traversal rule.
- **`ContentReferenceIntegrityTest.MeshContentUrls_AreNeverAddressedThroughTheStaticRoute`** — the durable repo-scan guard behind "static should not be referenced by anyone". It **found 9 real offenders on first run** (3 marketing links + 6 stale doc pages), all fixed.

**Fail-without:** with three probes applied (re-mount collections on `/static`, drop the `IsStatic` check, drop the traversal guard) → **5 of 15 fail**, including `BackingStoreUrl_IsNotReachableUnderStatic_ForAnonymous` → *"Expected NotFound … but found OK"* — i.e. the vulnerability reproduced. Probes reverted, **243/243** on the identical tree.

Suites: Hosting.Blazor **243/0** · Content 246 (1 pre-existing skip) · ContentCollections 28 · Markdown 137 · Markdown.Export 19 · Documentation 19 · Autocomplete 150 · Memex.Portal.Shared 238 · ContentCollections.Indexing.Graph 31. Independently re-verified **40/40** across the three new classes after merging current main. `-c Release -warnaserror` clean on all 16 touched projects.

## 🚨 Review before merge — broad blast radius (243 files)

1. **Every content URL in the product changed.** Server-side coverage is green but **no browser E2E was run** — worth a `/playwright` pass over images, PDFs, the file browser and doc diagrams.
2. **Persisted legacy values in live DBs.** Node `Icon` / `CorporateIdentity.LogoPath` rows holding `/static/storage/content/…` still resolve through the readers, but a stored value rendered directly into an `` will 404 until re-saved. **No migration written** — flagged rather than guessed.
3. **Externally-authored plugin repos** carrying `/static/...` URLs (e.g. an AgenticEngineering `poster`) now 404; those need a content-side edit.
4. `/api/content/{node}/{file}`: if the file path's first segment matches a collection name on that node, the explicit-collection branch wins. Documented; no collision in the current tree.
5. `AdaptResourceUrl` with a null address now returns the relative URL unchanged — zero callers, contract-only.

Closes #587


rbuergi:
CI red on a **single** test, and it is not this change: `SpaceMenuAndAccessTest.Space_ChildrenAreQueryable` on shard 5 — the recursive-delete deadlock tracked in #899 and fixed by PR #910. Every other shard is green, and the three new classes here (`StaticContentUnmountedTest`, `ContentUrlGenerationTest`, `ContentReferenceIntegrityTest`) pass 40/40 locally against current main.

#899 now blocks essentially every PR in the repo — it has redded #905, #907, #910's own run and this one. **This PR should merge once #910 lands**, not before; re-running shard 5 here would only re-sample the same deadlock.

Issue #912, opened 2026-08-07
CSharpFileParser serializes only the code — IsExecutable / ActivityParentPath are silently lost on every FS, blob and GitSync round-trip


rbuergi:
Found while fixing #841 (ExecuteScript silent dispatch death), and deliberately **not** fixed there because the blast radius crosses into users' git repositories.

## The defect

`CSharpFileParser.SerializeCodeConfiguration` (`src/MeshWeaver.Hosting/Persistence/Parsers/CSharpFileParser.cs:100`) writes **only `config.Code`** — it emits no `// <meshweaver>` header. The read side then rebuilds a `CodeConfiguration { Code, Language }`, so on any file-system or blob-storage round-trip these are **silently lost**:

- `IsExecutable`
- `ActivityParentPath`
- every `Last*` stamp

## Consequence

On an FS/blob-backed mesh, **every executable Code node comes back non-executable**, and `ExecuteScript` then refuses it. Pre-#841 that refusal was reported to nobody at all, which is why this stayed invisible.

Proven by a controlled A/B: the identical node executes on an in-memory mesh; the same node persisted to the file system reads back `IsExecutable=false`, and the on-disk `Scripts/exec-….cs` contains just the code with no header.

## Why it wasn't fixed in #841

The same serializer is what **GitSync writes into users' git repositories** (`GitHubSyncService.cs:263`, `NodeFileMapper.cs:38`) and what `AzureBlobStorageAdapter.cs:151` persists. Emitting a header is therefore a cross-repo **file-format change**, and it brushes the `.cs`/`.json` authoring contract in AGENTS.md — which already documents the read-side half of this trap:

> the `.cs` header carries ONLY `Id`/`DisplayName`/`NodeType`, so a cell with `isExecutable`/`activityParentPath` (or node-level `description`/`order`) authored as `.cs` silently loses its Run button on import (e.g. `Data/Architecture/PythonCodeNodes/SampleStatistics.json` stays JSON deliberately)

So the documented workaround is "author those as `.json`". This issue is the *write* side of the same gap: a node that was correctly created in the mesh loses those fields when serialized out, with no authoring mistake involved.

## Options (needs a decision)

1. **Extend the `.cs` header** to carry `IsExecutable` / `ActivityParentPath` (and decide about the `Last*` stamps, which are arguably runtime state that should not be in a source file at all). Changes the file format in every synced git repo.
[...]

PR #913, opened 2026-08-07
ExecuteScript dispatch answers the caller — nine swallow points closed (#841)

### LINKED ISSUE #841: ExecuteScript dispatch dies silently — no Activity node is ever created, and no failure mode in the chain is observable by the caller
### Summary

The MCP `ExecuteScript` tool returns `status: Dispatched`, but the documented first step of every script run — "Creates a fresh `Activity` MeshNode for this run" (`ScriptExecution.md`) — never happens: the Activity node at the returned `activityPath` is never created, no `ActivityLog` appears anywhere, and the pod emits nothing at Warning or above. Reproduced 100% on a distributed local deployment, across a pod restart, with a minimal freshly created Code node.

### How to reproduce (local dev portal, two-deployment topology, 2026-08-05/06)

1. Create a minimal Code node: `content: { language: "csharp", isExecutable: true, code: "Log.LogInformation(\"ping\");" }`.
2. Call MCP `ExecuteScript` on it → response `{ status: "Dispatched", activityPath: "{user}/_Activity/{id}", … }`.
3. Poll `get @{activityPath}` → `Not found`, indefinitely. No new `nodeType:Activity` node appears anywhere in the mesh.
4. Restart the serving pod, repeat → identical.

Ruled out: a logged crash (pod stdout at Warning+ clean for the submission window), a transient kernel wedge (fresh pod), a malformed script (minimal ping; a realistic script fails identically).

### The silent-failure chain (code map)

- `MeshOperations.ExecuteScript` (`src/MeshWeaver.AI/MeshOperations.cs:3740-3742`) posts `ExecuteScriptRequest` **fire-and-forget** to `Address(codeNodePath)`. `Dispatched` means only "`hub.Post` did not throw". The ack (`ExecuteScriptResponse`) is deliberately not awaited — the code comment (`:3718-3723`) states routing it back to a hosted MCP session hub is fragile.
- The handler lives on the Code node's **per-node hub** (`CodeNodeType.cs:64`, `HandleExecuteScript`). If that hub never activates on the serving silo, the post is dropped with no diagnostic.
- If it does activate, `HandleExecuteScript` reads its own node via `GetDataRequest` with a strict timeout; the timeout path posts a failure `ExecuteScriptResponse` — which nobody observes (see above).
- All Information-level breadcrumbs in this chain (e.g. `ExecuteScript called with path=` at `MeshOperations.cs:3703`) are suppressed by the deployment's logging defaults (`Memex.Portal.Distributed/appsettings.json`: `Default`/`MeshWeaver` = `Warning`), so none of these cases leaves any trace in stdout.

### What this issue asks for

1. The root cause of the drop on this topology (next diagnostic step per `DebuggingMessageFlow.md`: enable the messaging trace categories in the affected deployment and locate where the `ExecuteScriptRequest` delivery terminates).
2. Making dispatch failure observable: a dispatch whose Activity node does not materialize should surface as an error at Warning+ somewhere — currently the `ScriptExecution.md` step-1 contract can fail with zero trace, and the caller cannot distinguish "pending" from "dropped".

Not asked for: raising default log levels as the fix — the defect is the silent drop, not the logging configuration.

rbuergi:
## Summary

Fixes #841 — the dispatch could die with **no Activity node created and no failure mode observable by anyone**.

## Nine swallow points

| # | Location | Mechanism |
|---|---|---|
| 1 | `MeshOperations.cs:3740` | `hub.Post` **fire-and-forget** — "Dispatched" meant only *"Post did not throw"*. Every `Success=false` response and every routing `DeliveryFailure` reached a caller with no callback and was dropped. |
| 2 | `MeshOperations.cs:3736` | The caller **reconstructed** `activityPath` as `{partition}/_Activity/{id}`, while the owning hub resolves the parent via `ActivityParentPath` / `DefaultActivityParentPath` / `{viewer}`. With any of those, a **fully successful** run left the caller polling a path that can never exist — and it raced `CreateNode`, so even the default case 404s if polled immediately. |
| 3 | `CodeNodeType.cs:172` | `.Take(1)` on a live pathless query with **no timeout, no empty-guard** — if it never converges the chain stops dead: no node, no response, no log. The wedges-to-zero violation. |
| 4-8 | `CodeNodeType.cs` create / self-read / not-executable / stamp / `FailActivity` | Response-only, **no logging**, faults flattened to `.Message`, one bare `catch { }`. |
| 9 | `McpMeshPlugin.cs:781` | Tool description promised *"blocks until the kernel signals completion"*; `timeoutSeconds` was **entirely unused**. |

## Fix

`ExecuteScript` now uses `hub.Observe<ExecuteScriptResponse>` inside `Observable.Defer` (cold), bounded by the caller's `timeoutSeconds`, and returns the hub's **authoritative** activity path. `Success=false` → `Error` + reason; `OnError` → `errorType` + `message` + `detail` **and** `LogError` with the exception object, so type and stack reach the logger — the diagnosability standard set by #892. **"Dispatched" now means an Activity node exists.**

`CodeNodeType` gets two sinks — `RefuseDispatch` (Warning) and `FaultDispatch` (LogError **with the exception**) — both of which always answer the caller *and* log; three distinct refusal verdicts replace the conflated one; the bare `catch` logs. New extracted `ResolveActivityParent` applies `.Take(1).DefaultIfEmpty().Timeout(15s)` naming the partition, so exactly one of OnNext/OnError always reaches the subscriber.

## Verification

**Fail-without** (src reverted, tests unchanged): **6 failed / 0 passed** — the viewer-routed case reported the guessed path; the happy path errored *"No node found at rbuergi/_Activity/…"* because the guess raced creation; all three refusal cases reported `Dispatched` instead of `Error`. Reverting **only** `ResolveActivityParent`'s body: the stall test **timed out after 30 s** (the silent stop reproduced) and the empty-completion test threw *"Sequence contains no elements"*.

**Pass-with:** AI.Test 77/77 · `ResolveActivityParent` 5/5 · Documentation 1/1 · Markdown.Export 19/19. Independently re-verified **10/10** on the new classes after merging current main. `-c Release -warnaserror` clean on all touched projects.

## Notes

- `ExecuteScript` now blocks for the dispatch ack (previously returned instantly), bounded by `timeoutSeconds`; the ack is posted right after `CreateNode` — ~200 ms in-process.
- The 15 s parent-lookup timeout is a **diagnosability boundary, not a retry**: a genuinely stalled `Admin/Partition` query now fails loudly instead of hanging.
- Orleans/gRPC suites use `Mesh.Observe<ExecuteScriptResponse>` directly — contract unchanged, but they need real infra, so CI is their first run.
- **Separate defect exposed, filed as #912 and deliberately not fixed here:** `CSharpFileParser.SerializeCodeConfiguration` writes only the code, so `IsExecutable`/`ActivityParentPath` are silently lost on every FS/blob/GitSync round-trip — a cross-repo file-format decision. A labelled characterization test pins the truthful current behaviour.

Closes #841

### DIFF
--- a/src/MeshWeaver.AI/MeshOperations.cs
+++ b/src/MeshWeaver.AI/MeshOperations.cs
@@ -3692,12 +3692,31 @@ public class MeshOperations
     }
 
     /// <summary>
-    /// Runs an executable Code node's C# through the kernel (Microsoft.DotNet.Interactive)
-    /// and returns status JSON. The target node must have
-    /// <c>CodeConfiguration.IsExecutable == true</c>. Emits once when the kernel signals
-    /// completion (the kernel hub posts a response to <see cref="SubmitCodeRequest"/>
-    /// after the code finishes) or on timeout.
+    /// Dispatches an executable Code node's script to its per-node hub and emits dispatch
+    /// status JSON. The target node must have <c>CodeConfiguration.IsExecutable == true</c>.
+    ///
+    /// <para>The emission is the <b>dispatch acknowledgement</b>, not run completion: on
+    /// success it carries the <c>activityPath</c> the owning hub ACTUALLY created, which the
+    /// caller then polls / subscribes for live messages and the terminal status.</para>
+    ///
+    /// <para>🚨 #841 — every failure mode in the chain surfaces here as
+    /// <c>status: "Error"</c>. This method used to <c>hub.Post</c> the request
+    /// fire-and-forget and return a <c>Dispatched</c> envelope carrying a
+    /// <i>locally guessed</i> activity path, so three whole classes of failure were
+    /// invisible to the caller: (a) the request never routed (the Code node's per-node hub
+    /// did not activate, the path does not exist) — a <c>DeliveryFailure</c> nobody observed;
+    /// (b) the owning hub answered <c>Success = false</c> (not executable, unreadable node,
+    /// activity creation refused) — a response nobody observed; (c) the guessed path was
+    /// simply wrong, because the owning hub resolves the activity parent through
+    /// <c>CodeConfiguration.ActivityParentPath</c> / the partition default / the
+    /// <c>{viewer}</c> sentinel, none of which the caller can see. In case (c) a fully
+    /// SUCCESSFUL run still left the caller polling a path that would never exist.</para>
     /// </summary>
+    /// <param name="path">Path of the executable Code node (a leading <c>@</c> is stripped).</param>
+    /// <param name="timeoutSeconds">
+    /// Wall-clock budget for the dispatch acknowledgement (not for the script run). Elapsing
+    /// is reported as an error, never as a silent success.
+    /// </param>
     public IObservable<string> ExecuteScript(string path, int timeoutSeconds = 120)
     {
         logger.LogInformation("ExecuteScript called with path={Path}", path);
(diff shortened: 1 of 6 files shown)


rbuergi:
CI red on `MeshWeaver.Layout.Test.DataChangeStreamUpdateTest.DataChangeRequest_ShouldUpdateLayoutAreaViews` (shard 2) — **not** obviously this change: #841 touches `MeshOperations`, `CodeNodeType` and `McpMeshPlugin`, none of which are in the layout-render path.

That test is known-sensitive: `SynchronizationStream.OwnerVersion`'s doc comment cites it by name (*"the DataChangeStreamUpdateTest count-view non-emission"*) as the failure that mixing `Host.Version` into the init frame produced. So it is version/ordering-sensitive on the layout stream.

Before re-running: check whether it is failing on `main` independently — several PRs are in flight that touch stream versioning and change-feed delivery (#916 layout-area host dedup, #917/#910 change-feed dispatch). If it is red on main, it belongs to that cluster, not here.

Issue #674, opened 2026-07-27
MeshHubDisposalLeakTest passes unconditionally on macOS (ClrMD unavailable → analysis never runs)


rbuergi:
## Summary

`MeshWeaver.Hosting.Monolith.Test.MeshHubDisposalLeakTest.MeshHub_IsCollected_AfterMeshAndServiceProviderDisposal` **cannot fail on macOS**. It passes unconditionally, so every developer on a Mac gets a green light from a guard that physically cannot fire.

## Mechanism

The test detects retention by attaching ClrMD to a snapshot of its own process. On macOS:

1. `DataTarget.CreateSnapshotAndAttach` throws `PlatformNotSupportedException: … not supported on macOS`.
2. The analyzer's `catch` swallows it.
3. `staticRooted` therefore stays `false`.
4. `staticRooted.Should().BeFalse()` passes — **because the analysis never ran**, not because nothing was rooted.

An instrumented probe run alongside it reports `meshHubAlive=True` on macOS while the assertion still passes. The failure signal and the "everything is fine" signal are indistinguishable.

## Why this matters — it actively misled an investigation

This is not theoretical. During the 2026-07-26/27 flake work:

- A change (#664) was suspected of causing this test to fail on CI.
- 22 executions were run locally on macOS across both the branch and pristine `main` — **zero failures on either side** — and the correlation was refuted on that basis.
- Re-run on Linux under Docker, where ClrMD works: pristine `main` **0/3**, the branch **3/3**. The change did cause it, via an Rx `Sample` `PeriodicTimer` rooted in the process-wide `TimerQueue`.

It also hid a **pre-existing** retention on `main` (fixed in #673): with that class run alone on Linux, `main` fails **7 of 8** runs. No Mac developer could ever have seen it.

So the test currently provides false assurance in exactly the situation it exists for.

## Suggested fix

Make the unavailability explicit rather than silent. Any of:

- **Skip loudly** when ClrMD cannot attach — a skipped test with a clear reason is honest; a passing one is not.
- **Fail** when the analysis cannot run, on the grounds that a leak guard that cannot inspect anything has not done its job (harsher, but no one can misread it).
- At minimum, **assert that the analysis actually ran** before asserting its result — e.g. a tri-state (`Rooted` / `NotRooted` / `Unavailable`) instead of a bare bool, with `Unavailable` never satisfying the assertion.

Whichever is chosen, the same trap applies to any other test using the ClrMD analyzer — worth a sweep.

[...]

PR #916, opened 2026-08-07
Stop stacking a LayoutAreaHost per write — the render leak that killed pods (#606)

### LINKED ISSUE #606: Untyped live-compiled content (PluginContent) render hot-loop leaks memory until the pod is killed
## Production impact

memex-cloud portal pods balloon ~1GB/min while serving and die (liveness 137 / previously Orleans membership suicide) whenever the untyped-content loop is hot. Log signature, repeating:

```
MeshNodeStreamCache.GetStream: Content for ClaimsDeepfield stayed an untyped JsonElement after deserialization (TypeRegistry lacks the $type discriminator) — downstream 'Content is X'/'as X' consumers will fail (renders empty, reactive waits time out). Raw: {"$type":"PluginContent", …}
```

plus `PolymorphicTypeInfoResolver` / `ObjectPolymorphicConverter` unregistered-type warnings and Orleans Watchdog starvation.

## Mechanism (full trace, file:line on main @ 646b86f50)

A recompile of a live NodeType (Store/Plugin) mints a NEW `PluginContent` CLR type in a new collectible assembly. The persisted node (`ClaimsDeepfield` etc.) still carries `$type:"PluginContent"`. That type registers ONLY in the freshly built per-node hub's TypeRegistry (compiled `HubConfiguration` → `WithContentType<PluginContent>()`, applied in `MeshNodeHubFactory.cs:21-41`). There is **no fan-out** into the mesh hub, the deliberately domain-agnostic cache hub (`MeshNodeStreamCache.cs:450-458`), or circuit hubs — and the collectible-assembly guard (`PolymorphicTypeInfoResolver.cs:95-120`) intentionally refuses lazy adoption of dynamic types, so those hubs can never resolve the discriminator.

When such a hub deserializes the content (`MessageHubGrain.cs:244` uses the mesh hub's options; portal renders use circuit options), `ConvertContentJsonElementToTyped` (`MeshNodeStreamCache.cs:1626-1653`) degrades to `JsonElement`, logs, and **returns it as a successful emission**. Because it is not an error:

- the storm-breaker/negative cache (`GetStreamRaw`, 1170-1214) never engages;
- `AreaStreamRetry.RetryAreaWithBackoff`'s bounded retry/give-up never engages (error-only);
- `NamedAreaView`'s give-up handler never engages (untyped content arrives on `OnNext`, renders empty);
- typed reactive waits (`.Where(n => n.Content is T).Take(1).Timeout(…)` — the framework's own shape, e.g. `NodeTypeCompilationHelpers.cs:173/299/371/…`) never see a typed value → **time out** → the outer render re-opens the wait → each cycle allocates a fresh un-memoized `GetStream(path, options).Select` subscription (`MeshNodeStreamCache.cs:1559-1560` — unlike `GetQuery`, which memoizes via `_optionsWrappedQueries`, 1723-1731) plus fresh `LayoutAreaHost`/enrichment state registered for disposal against a host that never reaches a terminal render;
- the per-node grain amplifies with `TryDeactivateOnIdle()` churn (`MessageHubGrain.cs:301/312`) — activate→read→log→deactivate→reactivate per request → Watchdog starvation.

Nothing heals the mismatch; subscriptions + allocations accumulate until the pod dies.

## Fix seams (in rising order of invasiveness)

1. **(b) Stop the silent success** — in `ConvertContentJsonElementToTyped`/`DeserializeContent`, when a `$type` was present but stayed `JsonElement`, record the path in the existing `_negative`/backoff cache and surface `OnError` so the already-built bounded retry + give-up UI engage. Caps the leak; area shows unavailable instead of looping. ⚠️ Regression risk: consumers that today TOLERATE the untyped emission (`ContentAs` recovery is "the default that cures the untyped-JsonElement read") would start seeing errors — needs a consumer audit before shipping.
2. **(a) Read-side resolution fallback** — on discriminator miss, resolve the type from the live compilation/assembly store (`IAssemblyStore`, `NodeTypeDefinition.LatestAssemblyPath`) as a read-side alias (mirror `aliasByName`, `TypeRegistry.cs:62/86-91`), respecting the collectible-assembly pinning concern. Restores typed rendering everywhere; the loop exits because the wait completes.
3. **(c) Fan recompiled type registrations into live hubs** — broadcast on the compile-completion write (`NodeTypeCompilationHelpers.cs:1327-1370`) over the existing change-feed (`OnMeshChange`, `MeshNodeStreamCache.cs:487-518`). Hardest to get right (new CLR identity per recompile × collectible unload).

Recommended order: (b) with a consumer-tolerance audit, then (a) as the real fix.

## Context

Boot-storm half of the crash loop already fixed in #602 (lazy pre-warm/seed, read-first import skip, Orleans membership tolerance); this leak is the remaining crash driver on memex-cloud (2026-07-22, image ci.1301, pod killed 137 after ~26min at ~21GB).

rbuergi:
## Summary

Fixes #606 — the portal leaked until the pod was killed (memex-cloud ballooned ~1 GB/min while serving).

## Root cause: every write stacked another live render pipeline on the owner

**`Workspace.SubscribeToClient` (Workspace.cs:661-665) had no registry and no dedup by `StreamId`.** Every arriving `SubscribeRequest` unconditionally ran `ReduceManager.ReduceStream<TReduced>(...)`, which for a `LayoutAreaReference` reaches `LayoutExtensions.cs:114` → **`new LayoutAreaHost(...)`**. The previously-created stream for that same `StreamId` was never disposed — only an `UnsubscribeRequest` disposes a server-side stream, and a resubscribe never sends one.

**Retaining reference:** each leaked `LayoutAreaHost` is held by its own `SynchronizationStream<EntityStore>` on the owner's `sync/{ClientId}` hub, and itself holds `disposablesByArea` (LayoutAreaHost.cs:784) — node-stream + permission subscriptions that register as `SharedView` subscribers on `MeshNodeStreamCache` entries, so the idle sweep can **never** release those upstream sync streams either. It also keeps rendering and keeps pushing frames.

**The trigger — why it fires on every write:** `JsonSynchronizationStream.cs:615` resolves the mirror's staleness `Version` **by name** (`GetProperty("Version", typeof(long))`). `EntityStore` has no such property, so `receivedVersion` stays `0` forever while `announcedVersion` is fabricated as `receivedVersion + 1` for every version-less event. The gate `receivedVersion < announcedVersion` is therefore **permanently open** for every layout-area stream, so every change-feed event on the owner path calls `Resubscribe`, re-posting `SubscribeRequest` with the **same StreamId** and no unsubscribe.

Net: **one write to any node somebody is viewing = one more permanent `LayoutAreaHost` + full render graph.** Untyped `PluginContent` did not create this; it made the pulses continuous (renders never reach a terminal state), which is why it presented as an untyped-content problem.

**Controlled A/B:** disarming only the change-feed resubscribe took per-write frames from `1,4,5,6,7,8,9,10` to a flat `1 Full + 4 menu patches` on all 8 writes. Instrumentation reverted; no A/B code remains.

## Fix

1. **`Workspace.cs`** — instance `ConcurrentDictionary<(Subscriber, StreamId), ClientSubscription>` with register/lookup. **Instance, not static**, so lifetime is the owner hub (an owner recycle correctly starts empty; a genuinely orphaned mirror still gets a fresh stream). `ClientSubscription` is a **class, not a record**: `SynchronizationStream` is a record whose generated `GetHashCode`/`Equals` recurse into `StreamConfiguration` and stack-overflow the process — the same trap already documented on `_evictedRemoteStreams`. Removal is pair-exact by reference identity.
2. **`JsonSynchronizationStream.CreateSynchronizationStream`** — a repeat `SubscribeRequest` for a `(subscriber, StreamId)` already served under the same reference now **re-asserts the existing stream's state as a `Full`** (delivered by the already-registered outbound forwarder, and the client's `ExpectResubscribeFull` latch is armed for it) and returns that stream instead of building a second host. It deliberately does **not** dispose the old stream: `SynchronizationStream.Dispose` tears down the shared `sync/{ClientId}` hub and `HostedHubsCollection.GetHub` has no disposal check, so dispose-then-recreate races a disposing hub.

**Three secondary accumulators of the same class** — subscriptions created inside a per-emission render body, appended under a synthetic key no area teardown reaps (`DisposeExistingAreas`/`DisposeChildAreas` only remove keys that `StartsWith(context.Area)`) — switched to the framework's existing `LayoutAreaHost.ReplaceDisposable`, documented for exactly this:

- `NodeMenuItemsExtensions.cs:101` — the worst: a **global** predicate renderer (`WithRenderer(_ => true, RenderMenus)`) appending under the constant key `"$Menu"` for every context. Now keyed `menu-subscription:{area}` (deliberately not the bare area name, since `LayoutAreaHost.RenderArea:557` registers rendered controls under the area key and replacing that bucket would dispose the live menu control).
- `OverviewLayoutArea.cs:52` and `MeshNodeLayoutAreas.cs:1781` — content projections; the comment at `MeshNodeLayoutAreas.cs:1749` already notes the body "re-runs on every render".
- `GitHubSyncSettingsTab.cs:190` — same shape.

**Honest attribution:** the host-stacking fix is **measured**; the four `ReplaceDisposable` changes are **code-reasoned and suite-verified**, not independently measured (their accumulated subscriptions are silenced by their own `DistinctUntilChanged`, so they leak memory without producing countable wire traffic).

## Verification

`LayoutAreaResubscribeHostAccumulationTest.RepeatedWrites_DoNotStackLayoutAreaHosts_OnTheOwner` — managed measurement only, no ClrMD, so it works on macOS where those guards now skip (#674). Every live server-side `LayoutAreaHost` emits its own `Full` per node change, so **Full frames per write == live render pipelines**.

- **Fails pre-fix:** `Full frames per write: [1, 4, 5, 6, 7, 8, 9, 10, 11, 12]` → *"Expected 12 to be less than or equal to 6"*.
- **Passes post-fix:** `[1, 2, 2, 2, 2, 2, 2, 2, 2, 2]` (the 2 = 1 render Full + 1 resubscribe re-assert Full).

Suites on the final tree: **Hosting.Monolith 478 passed / 0 failed** (4 skipped) · **Graph 820/820** · **Layout 343/343** · **Data 303/303**. `-c Release -warnaserror` clean on Data, Graph, GitSync and the test project.

## Residual risks

1. **Rebuild-as-recovery is gone.** A resubscribe no longer rebuilds the server-side render; if a host is genuinely stuck with `Current == null`, the guard returns it without asserting anything. The dedicated recovery path (`ArmOverlaySelfHeal` → `DisposeRequest` → recycle) still covers this, but it is a real behavioural change worth a reviewer's eye.
2. **The inert staleness gate is untouched** (`JsonSynchronizationStream.cs:615`). Layout-area streams still resubscribe on every owner-path change — now cheap (a Full re-assert) instead of leaking, but still one extra round-trip per write per viewer. Making the gate evidence-based for version-less reductions is a worthwhile follow-up.
3. **Same defect class not swept exhaustively.** Four further `RegisterForDisposal`-inside-a-render sites were found and deliberately **not** touched: `EditLayoutArea.cs:193`, `DomainCatalogLayoutArea.cs:50` (also mints a fresh `Guid` data id per render), `ThreadLayoutAreas.cs:221` and `:285`, `PartitionSyncAdminLayoutArea.cs:286`. Widening into the chat path without measurement was more blast radius than could be verified here.
4. **Untyped-content half not fixed** (separate finding): `MeshNodeStreamCache.ConvertContentJsonElementToTyped` and `MeshNodeStreamExtensions.EnsureTypedContent` return `node with { Content = deserialized }` over a freshly cloned `JsonElement`. `JsonElement` has no `Equals` override, so `MeshNode` record equality is **always false** for untyped content and every `DistinctUntilChanged` degrades to a no-op — including the persistence sampler's. The sibling `GetQuery` seam already does the right thing. Symmetry looks high-value; it needs a NodeType whose content type is registered only on its per-node hub to repro faithfully.

Closes #606

### DIFF
--- a/src/MeshWeaver.Data/Serialization/JsonSynchronizationStream.cs
+++ b/src/MeshWeaver.Data/Serialization/JsonSynchronizationStream.cs
@@ -671,6 +671,40 @@ public static class JsonSynchronizationStream
         var logger = GetLogger(hub.ServiceProvider);
         var request = delivery.Message with { Subscriber = delivery.Sender };
 
+        // 🚨 RE-SUBSCRIBE IS A REFRESH, NOT A SECOND SUBSCRIPTION (issue #606).
+        //
+        // Resubscribe() re-posts a SubscribeRequest carrying the SAME StreamId — it means
+        // "re-assert my stream", not "open another one". Building a second reduced stream here
+        // left the FIRST one live and rendering forever (nothing but an UnsubscribeRequest ever
+        // disposes a server-side stream, and a resubscribe never sends one). For a
+        // LayoutAreaReference each of those is a whole LayoutAreaHost + render graph, so every
+        // change-feed pulse on the owner path permanently added one more render pipeline —
+        // the unbounded growth that killed the pod.
+        //
+        // Serve the refresh off the stream we ALREADY have: re-assert its current state as a
+        // Full, which the outbound forwarding subscription below (registered when that stream
+        // was created) delivers to the subscriber — exactly the "fresh snapshot" the caller
+        // asked for, and exactly what its ExpectResubscribeFull latch is armed to accept.
+        // A genuinely orphaned mirror (owner recycled ⇒ new workspace ⇒ empty registry) still
+        // takes the full create path below.
+        if (workspace is Workspace registry
+            && registry.GetClientSubscription(request.Subscriber, request.StreamId, request.Reference)
+                is ISynchronizationStream<TReduced> alreadyServing)
+        {
+            logger.LogDebug(
+                "Owner {Owner} already serves stream {StreamId} for {Subscriber} — re-asserting the current snapshot instead of creating a second stream",
+                hub.Address, request.StreamId, request.Subscriber);
+            if (alreadyServing.Current is { Value: not null } snapshot)
+                alreadyServing.Update(
+                    _ => new ChangeItem<TReduced>(snapshot.Value, alreadyServing.StreamId, alreadyServing.Hub.Version),
+                    ex => logger.LogWarning(ex,
+                        "Stream {StreamId}: could not re-assert snapshot for resubscribing subscriber {Subscriber}",
+                        alreadyServing.StreamId, request.Subscriber));
+            // Nothing yet to re-assert (the initial subscribe is still hydrating) — its own
+            // first Full is already on its way to this subscriber; do NOT build a second stream.
+            return alreadyServing;
+        }
+
         var fromWorkspace = workspace
             .ReduceManager
             .ReduceStream<TReduced>(
@@ -756,6 +790,13 @@ public static class JsonSynchronizationStream
         //             reduced.Host.GetWorkspace().RequestChange(e);
         //         })
         // );
+
+        // Record it as THE stream for this (subscriber, StreamId) so a later resubscribe
+        // refreshes it instead of stacking another one (see the guard at the top). Registered
+        // AFTER the outbound forwarding subscription exists, so a re-assert can never find a
+        // stream that has no way to reach the subscriber. Unregisters itself on disposal.
+        (workspace as Workspace)?.RegisterClientSubscription(
+            request.Subscriber, request.StreamId, request.Reference, reduced);
 
         return reduced;
     }
--- a/src/MeshWeaver.Data/Workspace.cs
+++ b/src/MeshWeaver.Data/Workspace.cs
@@ -664,5 +664,88 @@ public class Workspace : IWorkspace
         this.CreateSynchronizationStream<TReduced, TReference>(delivery);
     }
 
+    // 🚨 ONE server-side stream per (subscriber, StreamId) — issue #606.
+    //
+    // A SubscribeRequest carries the SUBSCRIBER's StreamId, and a resubscribe deliberately
+    // REUSES it ("refresh MY stream" — JsonSynchronizationStream.Resubscribe). Before this
+    // registry the owner treated every arrival as a brand-new subscription and built another
+    // reduced stream — for a LayoutAreaReference that is a whole new LayoutAreaHost + render
+    // graph (LayoutExtensions' AddWorkspaceReferenceStream factory) — while the previous one
+    // kept running, kept rendering and kept pushing DataChangedEvents to the same subscriber.
+    // Nothing ever released it: only an UnsubscribeRequest disposes a server-side stream, and
+    // a resubscribe never sends one.
+    //
+    // That made the accumulation unbounded on the ordinary serving path: the layout-area
+    // stream's staleness gate is INERT for EntityStore reductions (EntityStore has no
+    // `long Version`, so `receivedVersion` never advances while `announcedVersion` is
+    // fabricated as received+1 — see the version gate in JsonSynchronizationStream), so
+    // EVERY change-feed event on the owner path resubscribes. Measured: one write to a node
+    // added one more live render pipeline, so a single subsequent write produced 1 → 10 Full
+    // frames after 8 writes, each leaked host holding its EntityStore, its menu/node-stream
+    // subscriptions (which pin MeshNodeStreamCache entries so their upstream sync streams can
+    // never be idle-released) and its whole control tree.
+    //
+    // The registry is an INSTANCE field on the workspace, so its lifetime IS the owner hub's
+    // (no static state, and an owner recycle correctly starts from empty → a genuinely
+    // orphaned mirror still gets a fresh stream).
+    private readonly ConcurrentDictionary<(string Subscriber, string StreamId), ClientSubscription>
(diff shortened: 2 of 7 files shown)


rbuergi:
CI red on a **single** test that this change does not touch: `SpaceMenuAndAccessTest.Space_ChildrenAreQueryable` on shard 5 — the #899 delete wedge, which has now redded #905, #907, #910, #911 and this PR. Every other shard green; the regression test here (`LayoutAreaResubscribeHostAccumulationTest`) passes locally against current main.

**Merge once the #899 fix lands.** Re-running shard 5 would only re-sample the same deadlock. The #899 work is now taking the generator route (releasing the permission-fold gate) rather than the async-fan-out route, which was rejected for breaking three documented synchronous cache-invalidation contracts — see #917's latest comment.

Issue #889, opened 2026-08-07
PaywallRealGateShapeTests: buyer's AccessAssignment update never reaches the cached $security-access stream — 3 PRs redded on shard 5 in one morning (post-#868)


rbuergi:
Tracking the REMAINING paywall gate-fold race — the memory of record says post-#868 shard-5 failures are real signal, and 2026-08-07 delivered three on causally-unrelated PRs:

| PR | run | failing test |
|---|---|---|
| #883 | (Wave-0 rerun) | shard 5, PaywallRealGateShapeTests |
| #884 | 31153582625 | `Cover_VisitorAllowed` |
| #887 | 31154251112 | `AuthenticatedNotEntitled_IsDenied` |

## What the #881 instrumentation shows (new signal)

```
Expected the observable to emit a value matching the predicate within 45s because the buyer's
root grant must survive the child deny on GateShape4fad606c/PaidLesson, but it did not.
Last of 1 emission(s) was: None.
```

**Exactly one emission (`None`), then 45 s of silence.** So the `enriched` CombineLatest fold DID complete once off the initial reads (all sources emitted their Initial), and the buyer's `AccessAssignment` — the last of five seed writes — never produced an UPDATE emission on the cached `$security-access:{scope}` stream. This is the missed-change-feed-update shape, not fold-equality (#868 fixed that) and not the disposed-subscriber starvation (#853).

## Ruled out / already shipped in this family
- #849 (vacuous deny asserts), #853 (OnNext-abort starvation), #868 (path-set equality swallowing content-only changes — CI-verified green on dispatch run 31128621589), #881 (diagnosis: barrier reports last emission).
- `PostgreSqlMeshQuery.ObserveQuery` itself is race-hardened (early-backlog before initial query; live pipeline before backlog swap; per-change re-query serialized via Concat) — the drop is upstream or in the cache layer.

## Prime suspects
1. **Per-schema adapter feed merge race** (partitioned routing): the merged `_changeFeed` includes each per-schema adapter's `Changes` Subject — if the schema's adapter is created lazily by the very write whose notification matters, the publish can precede the merge subscription → notification lost with no reconcile (pg_notify LISTEN fallback is disabled for partitioned PG).
2. **Cached-stream layer** (`$security-access:{scope}`, Replay(1) process cache): a subscriber-lifecycle path that drops one update without killing the stream.

[...]

PR #917, opened 2026-08-07
Eliminate the Rx lock-order-inversion class, not just the one instance (#899)

### LINKED ISSUE #899: Per-leaf delete of an _Access grant can park forever in the permission check — recursive space delete wedges (post-#894, 1-in-N)

rbuergi:
## Summary

**Supersedes #910** (cherry-picked here) and closes #899 **at the class level**.

## An empirical correction to the model

I measured which Rx operators actually hold a gate while delivering `OnNext` (Rx 6.1.0):

| construct | holds a gate during downstream delivery? |
|---|---|
| `Merge` | **yes** |
| `CombineLatest` | **yes** |
| `Concat` / `Replay(1)` | yes |
| bare `Subject<T>` | **no** — lock-free snapshot fan-out |
| `IsolatedChangeFeed` | no |

So a `Subject` **is not a gate** — it is the *conduit* that lets a gate-holding publisher walk into someone else's gate. Every verdict below therefore turns on **which thread publishes**, not on the subject type. That reframing is what made the inventory decidable rather than guesswork.

It also explains why **PG never reproduced the wedge while in-memory did**: any step that hops through `IIoPool` (`SubscribeOn(TaskPoolScheduler)` on every path, including `IoPool.Unbounded`) has already left the fold gate. The in-memory path is the only fully synchronous one in the repo.

## Inventory + fixes

Exactly **one** true inversion source beyond the one #910 fixed: `InMemoryStorageAdapter.Write`/`Delete`/`DeleteIfExists`, publishing on the caller's hub thread through a fully synchronous `Observable.Defer` pipeline. **Fixed.** Sqlite, Snowflake, the Snowflake router and Cosmos publish off-thread via `IIoPool` — safe for inversion, but they carry the #889-shaped isolation gap (plain `Subject` + `catch {}`); left deliberately, because their suites cannot run here and an untested delivery-semantics change to a storage backend is worse than a reported one.

New shared primitive **`DispatchedChangeFeed`** generalises #910's one-off loop and PG's `IsolatedChangeFeed` into one type (enqueue-and-dispatch + per-subscriber isolation). The dispatch shape is deliberately **not** `ObserveOn`: measured, it prefers `ISchedulerLongRunning` and parks **one dedicated thread per live subscription** (64 feeds → +65 threads), and PG creates one adapter per partition schema — a thread-per-schema leak. `Select(v => Observable.Return(v, scheduler)).Concat()` is FIFO, off-thread, **+2 threads for 64 feeds**, and is AGENTS.md's own sanctioned serialisation idiom. `InProcessMeshChangeFeed` is rebuilt on it and **gains per-subscriber isolation it did not have** — #910 caught around the whole fan-out, so the first throwing subscriber still aborted every later one.

## Ingredient 2 — the generator, not just the conduit

New `HubPermissionExtensions.TakeDecisionOutsideGate()`: the permission decision is still **taken** inside the fold, but re-emitted via `Observable.Return(value, Scheduler.Default)` so only the continuation leaves the gate. Applied to `CheckDeletePermissionForNode`.

All **47** `GetEffectivePermissions` call sites classified by a structural rule and verified individually: layout areas / menu providers projecting into a `UiControl` are **safe** (they finish inside the gate and touch nothing shared); message-pipeline, validator and node-operation sites are **dangerous**. Named dangerous and deliberately deferred: **`AccessControlPipeline.cs:123`** — every `[RequiresPermission]` handler still runs its body inside the fold's gate, the broadest exposure in the repo, needing its own full-suite cycle; `RlsNodeValidator` is flagged as the cheapest high-leverage follow-up (one call covers every validated write).

## Tests

`RxFanOutInversionHarness` — the reusable two-publisher / `Barrier` / shared-gate probe, parameterised by `subscribe` + `publish` so it points at any fan-out point; #910's inline version folds into it. **12 tests, 9 fail pre-fix** (three deadlock at 10 s). `Take1_alone_runs_the_continuation_INSIDE_the_folds_gate` is a deliberate **precondition** test asserting the pre-fix shape really does hand the handler body the lock, so the others cannot pass vacuously.

## Verification

Whole solution  → **0 warnings, 0 errors**. Hosting **121/121** · NodeOperations **85/85** · Query **361/361** · Monolith **477 passed / 4 skipped** · Data **303/303**. Independently re-verified **12/12** after merging current main.

Unlike #910, **no existing test needed adapting** — which matters, because in-memory is the default monolith storage and making its feed asynchronous was the highest-blast-radius part of this change.

## Found, not fixed (separate class — worth their own issues)

1. `CachingStorageAdapter`, `PathRemappingStorageAdapter` and `RoutingProxyAdapter` **do not forward `inner.Changes`** — they fall through to the `Observable.Empty` default, so writes behind them are invisible to every synced query.
2. `CosmosStorageAdapter._changes` has **no `OnNext` anywhere** — Cosmos writes emit no change notification at all.

Closes #899

### DIFF
--- a/src/MeshWeaver.Hosting/MeshChangeFeed.cs
+++ b/src/MeshWeaver.Hosting/MeshChangeFeed.cs
@@ -1,6 +1,5 @@
+using System.Reactive;
 using System.Reactive.Concurrency;
-using System.Reactive.Linq;
-using System.Reactive.Subjects;
 using MeshWeaver.Mesh.Services;
 using Microsoft.Extensions.Logging;
 
@@ -50,28 +49,22 @@ namespace MeshWeaver.Hosting;
 public class InProcessMeshChangeFeed : IMeshChangeFeed, IDisposable
 {
     /// <summary>
-    /// Publisher-facing inbox. <see cref="Subject.Synchronize{T}(ISubject{T})"/> because ANY
-    /// hub action block may publish concurrently (two per-node hubs deleting at the same
-    /// time is the normal case in a recursive delete) and a bare <see cref="Subject{T}"/>
-    /// is not safe under concurrent <c>OnNext</c>.
+    /// The bus itself: enqueue-and-dispatch fan-out with per-subscriber isolation. Safe under
+    /// concurrent publish from ANY hub action block (two per-node hubs deleting at the same
+    /// time is the normal case in a recursive delete).
     /// </summary>
-    private readonly Subject<MeshChangeEvent> _inboxCore = new();
-
-    private readonly ISubject<MeshChangeEvent> _inbox;
-
-    /// <summary>Subscriber-facing fan-out, pumped exclusively by <see cref="_dispatcher"/>.</summary>
-    private readonly Subject<MeshChangeEvent> _outbox = new();
+    private readonly DispatchedChangeFeed<MeshChangeEvent> _feed;
 
     /// <summary>
     /// The single dispatch thread. A dedicated <see cref="EventLoopScheduler"/> rather than
     /// the thread pool: change-feed delivery drives cache invalidation and synced-query
-    /// folds, and must not queue behind (or be starved by) unrelated pool work.
+    /// folds, and must not queue behind (or be starved by) unrelated pool work. There is
+    /// exactly ONE of these per mesh, so the dedicated thread is affordable here — storage
+    /// adapters (one feed per partition schema) use the pool-backed default instead.
     /// </summary>
     private readonly EventLoopScheduler _dispatcher =
         new(start => new Thread(start) { IsBackground = true, Name = "mesh-change-feed" });
 
-    private readonly IDisposable _pump;
-    private readonly ILogger<InProcessMeshChangeFeed>? _logger;
     private bool _disposed;
 
     /// <summary>
@@ -80,29 +73,10 @@ public class InProcessMeshChangeFeed : IMeshChangeFeed, IDisposable
     /// <param name="logger">Optional logger; a subscriber that throws is reported here.</param>
     public InProcessMeshChangeFeed(ILogger<InProcessMeshChangeFeed>? logger = null)
     {
-        _logger = logger;
-        _inbox = Subject.Synchronize(_inboxCore);
-        _pump = _inboxCore
-            .ObserveOn(_dispatcher)
-            .Subscribe(
-                change =>
-                {
-                    try
-                    {
-                        _outbox.OnNext(change);
-                    }
-                    catch (Exception ex)
-                    {
-                        // A faulting subscriber must not tear the bus down for everyone
-                        // else — but it is never swallowed silently: it is reported at
-                        // Error so the defect is visible in logs and CI.
-                        _logger?.LogError(ex,
-                            "Mesh change-feed subscriber threw for {Path} {Kind}",
-                            change.Path, change.Kind);
-                    }
-                },
-                ex => _logger?.LogError(ex,
-                    "Mesh change-feed dispatch loop faulted — change events stopped"));
+        // 🚨 The dispatch + isolation semantics live in DispatchedChangeFeed — the ONE
+        // fan-out primitive shared with every storage adapter, so "publishers only enqueue"
+        // is a property of the mesh, not a property of this one class.
+        _feed = new DispatchedChangeFeed<MeshChangeEvent>(logger, "mesh-change-feed", _dispatcher);
     }
 
     /// <summary>
@@ -114,7 +88,7 @@ public class InProcessMeshChangeFeed : IMeshChangeFeed, IDisposable
     public void Publish(MeshChangeEvent change)
     {
         if (!_disposed)
-            _inbox.OnNext(change);
+            _feed.OnNext(change);
     }
 
     /// <summary>
@@ -125,7 +99,7 @@ public class InProcessMeshChangeFeed : IMeshChangeFeed, IDisposable
     public void PublishLocal(MeshChangeEvent change)
     {
         if (!_disposed)
-            _inbox.OnNext(change);
+            _feed.OnNext(change);
     }
 
     /// <summary>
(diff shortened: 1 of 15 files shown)


rbuergi:
🚨 **Do not merge yet — async fan-out breaks three production cache-invalidation contracts.** This came out of the parallel #910 investigation and applies here too, since this PR makes the *same* async conversion and extends it to in-memory storage.

**Three framework caches are invalidated ONLY by the change feed, and each states the synchronous dependency in its own doc comment:**
- `PathResolutionService._resolutionCache` — *"Runs synchronously on the publisher's thread"*; positive-only, and *"Created(P) can deepen the resolution of P and of every descendant"*
- `MeshNodeStreamCache`'s failure-state reset — *"the VERY NEXT read heals"*
- `Workspace._remoteStreamCache` eviction

**Concretely:** creating `TestData/storm` first *resolves* that path to route the create, caching the **shallower ancestor** address. Pre-fix, the create's publish evicted that entry before `CreateNode(...)` returned — read-your-own-writes. Post-fix the eviction lands an unbounded moment later, so an immediately-following query routes against the stale shallower address and starves.

**Evidence** (from the #910 branch, same conversion): `SearchResultStormTest` passes **2/2** full-project runs pre-fix and fails **2 of 4** post-fix locally, **2 of 2** on CI. The failing set *shifts every run* — `SearchResultStormTest`, `PandasExplorerLayoutAreaTest` (leaked `CreateNodeRequest` pending 2.47 s at the root mesh hub), `HarnessInstallGateTest` — i.e. broad timing/staleness, not a few bad assertions. This PR's green Monolith run is a sample, not a refutation: the failure is intermittent at ~50%.

**Recommended fix** (needs a maintainer call): give `IMeshChangeFeed` an explicit **inline invalidation channel** for the pure, non-blocking dictionary-eviction subscribers, keeping the observable fan-out on the dispatch loop. Those handlers take no Rx gate and do no I/O, so they were never part of the #899 cycle — which is specifically {permission-fold gate, synced-query `Merge` gate}, both taken by *subscriber chains*.

**The alternative removes the fallout entirely** and is the half this PR already names as deferred: stop `HandleDeleteNodeRequest` running the whole delete inside `GetEffectivePermissions`' `CombineLatest` gate — i.e. finish ingredient 2 (`AccessControlPipeline.cs:123` and the `RlsNodeValidator` caller) rather than making delivery async at all.

Also worth folding in from #910's branch — a **genuine production regression** that async delivery exposed and that is fixed there: `ThreadExecution.PushToResponseMessage` guards `Status`, `ToolCalls` and `UpdatedNodes` against a late push regressing them, but **`Text` was left open**, so a delayed placeholder could overwrite a completed answer with `"Generating response..."`. That guard belongs in whichever PR lands.

rbuergi:
**CI confirms the predicted fallout — drafting this in favour of the generator route.** The failures are exactly the async-delivery staleness class named in my previous comment:

- `GrainActivationFailureRegistryTest.ChangeFeedBroadcast_ClearsStoredActivationError_ForExactlyThatPath` (shard 0)
- `SearchResultStormTest.CatalogSubscription_ContentUpdateStorm_AddsNoExtraRerenders` (shard 1)
- `ChangeFeedVersionGateTest.ReceivedOwnerWrites_DoNotResubscribe_ButStaleAnnouncementDoes` (shard 1)
- `ObservableQueryIntegrationTests.ScopeDescendants_NotifiesOnAllDescendantChanges` (shard 1)

Every one of those depends on the change feed having delivered before the next read — the contract those three caches document explicitly. So this is not adaptable test fallout; it is the approach.

**Superseded by the generator route**, now in progress: release the permission-fold gate (`CheckDeletePermissionForNode`, `AccessControlPipeline.cs:123`, the `RlsNodeValidator` caller) so ingredient 2 of the cycle cannot form, leaving synchronous delivery intact. That already has **15/15 consecutive clean NodeOperations soak runs**, with repeated full-Monolith runs in progress.

**Worth keeping from this PR, and being carried over:**
- the measured operator-gate table — `Merge` and `CombineLatest` hold a gate during delivery, a bare `Subject` does not; the subject is the *conduit*, not the lock
- the complete fan-out inventory with per-site verdicts
- `RxFanOutInversionHarness` plus the deliberate precondition test
- the `TakeDecisionOutsideGate()` idiom and the 47-site classification

**Still worth their own issues regardless of route** (found here, not fixed): `CachingStorageAdapter`, `PathRemappingStorageAdapter` and `RoutingProxyAdapter` do not forward `inner.Changes` and silently fall through to the `Observable.Empty` default; and `CosmosStorageAdapter._changes` has no `OnNext` anywhere, so Cosmos writes emit no change notification at all.

rbuergi:
Superseded by #918 (merged). The reusable pieces (operator-gate measurements, fan-out inventory, inversion harness, TakeDecisionOutsideGate idiom) were carried into it.

PR #924, opened 2026-08-08
Executable Code nodes survive the FS/blob/GitSync round-trip — .cs claims pure source only

### LINKED ISSUE #912: CSharpFileParser serializes only the code — IsExecutable / ActivityParentPath are silently lost on every FS, blob and GitSync round-trip

rbuergi:
Fixes #912.

## The defect

`CSharpFileParser.CanSerialize` claimed **every** `CodeConfiguration`, and `SerializeCodeConfiguration` writes only the code — no `<meshweaver>` header. On any FS/blob/GitSync round-trip, `IsExecutable`, `ActivityParentPath`, the `Last*` execution stamps, a non-C# `Language`, and `UnknownMembers` were silently stripped. Every executable Code node on a file-backed mesh came back non-executable, and `ExecuteScript` (truthfully, since #841) refused it.

## The fix — option 2 of the issue, at the single selection seam

`CanSerialize` now declines any config carrying state a bare `.cs` file cannot represent; the node falls through to the universal whole-node JSON serializer. This matches the documented authoring contract (AGENTS.md: executable cells are `.json`; pure source stays `.cs`). `Serialize` throws on a non-representable config so future registry-bypassing callers fail loud instead of dropping state.

Everything downstream already handles the extension flip:
- **FS + blob adapters** pick the serializer via `GetSerializerFor` and clean up the other-extension file — a node persisted as `.cs` by pre-fix builds migrates to `.json` on its next write. Pinned by `MarkingAPersistedSourceNodeExecutable_MigratesCsToJson` (the read side prefers `.cs`, so a stale `.cs` would shadow the new `.json` forever).
- **GitSync** `Push` mirrors the export set and prunes absent files — the old `.cs` disappears in the same commit that adds the `.json`. `TypedNodeExportRobustnessTest.CodeSourceUnderSourceSegment_ExportsAsCsAndRoundTrips` (pure source stays `.cs`) still green.

## Tests

- `CodeNodeRoundTripTests` (new, 11 cases): serializer-selection gate per field, lossless FS round-trip as `.json`, pure source keeps `.cs`, and the `.cs`→`.json` migration with stale-file cleanup.
- The #841 characterization test `ExecuteScript_ForIsExecutableCodeNode_SurfacesTheFileSystemRoundTripLoss` flipped to its documented post-fix form (`…_DispatchesAfterFileSystemRoundTrip`): dispatch **accepted** after an FS round-trip, real activity path.
- Full `McpReadYourWritesTest` (12) + `TypedNodeExportRobustnessTest` (3) green locally in Release.

## Out of scope (unchanged)

The partition-objects surface (`SavePartitionObjects`/`GetPartitionObjects` for `Source`/`Test` code partitions) still round-trips bare code via the static `SerializeCodeConfiguration`; compile sources carry no executable metadata there.

### DIFF
--- a/src/MeshWeaver.Documentation/Data/WhatsNew/2026-08-08-executable-code-cells-survive-storage.md
+++ b/src/MeshWeaver.Documentation/Data/WhatsNew/2026-08-08-executable-code-cells-survive-storage.md
@@ -1,0 +1,17 @@
+---
+Name: Executable code cells keep their Run button on file-backed meshes
+Category: What's New
+Description: Executable Code nodes no longer lose their executable flag and execution history when persisted to file, blob or git storage.
+Icon: Sparkle
+---
+
+# Executable code cells keep their Run button on file-backed meshes
+
+Code cells marked executable used to silently lose that setting — together with their
+execution history and activity links — whenever the mesh stored them on a file system,
+in blob storage, or synced them to a git repository. After a restart or re-import the
+Run button was gone and running the script was refused.
+
+Such cells are now stored in a lossless format, so the Run button, the last-run
+information and the activity trail all survive persistence and sync. Plain C# source
+files are unaffected and keep their readable `.cs` form in synced repositories.
--- a/src/MeshWeaver.Hosting/Persistence/Parsers/CSharpFileParser.cs
+++ b/src/MeshWeaver.Hosting/Persistence/Parsers/CSharpFileParser.cs
@@ -90,6 +90,13 @@ public partial class CSharpFileParser : IFileFormatParser
     {
         if (node.Content is not CodeConfiguration codeConfig)
             throw new InvalidOperationException("Cannot serialize node without CodeConfiguration content");
+        if (!IsPureCSharpSource(codeConfig))
+            throw new InvalidOperationException(
+                $"CodeConfiguration for '{node.Path}' carries state a .cs file cannot represent " +
+                "(IsExecutable / ActivityParentPath / Last* stamps / non-C# language / unknown members); " +
+                "serializing it here would silently drop that state on the next read (#912). " +
+                "Select the serializer via FileFormatParserRegistry.GetSerializerFor so the node " +
+                "falls through to the lossless JSON form.");
 
         return SerializeCodeConfiguration(codeConfig);
     }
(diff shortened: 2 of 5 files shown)


rbuergi:
Also carries a test-only de-flake in the same project: `PermissionFoldGateTests.TakeDecisionOutsideGate_runs_the_continuation_OUTSIDE_the_folds_gate` redded this PR's run 31239308980 on its thread-id assertion — an over-strict proxy (the pool may legally reuse the Subscribe thread for the hopped continuation once the test yields). "Outside the gate" is exactly the `Monitor.IsEntered == false` assertion the test already makes; the thread-id check is removed with a comment explaining why. Unrelated to the serializer change, disclosed here for review clarity.

PR #936, opened 2026-08-08
What's New: admins can enroll a user in one step

### LINKED ISSUE #642: Store: expose PluginGate.Enroll as an admin-invokable action (purchase = one primitive)
A purchase is two writes (_Entitlements record + _Access Viewer grant); creating only the record yields 'purchased but no access'. Admins/automation need the single Enroll primitive reachable via API/tool.

From the 2026-07-24 plugin-install retrospective.

rbuergi:
Fixes #642.

The implementation lands in **[MeshWeaver.Plugins#370](https://github.com/Systemorph/MeshWeaver.Plugins/pull/370)** — the enroll primitive it exposes (`PluginGate.Enroll`, wrapped by `Entitlements.GrantEntitlement`) lives in the Store plugin's Roslyn-compiled node source, not in this repo, so this PR carries only the user-facing release note.

## What the issue was

A purchase is two writes — the eternal entitlement record at `{plugin}/_Entitlements/{viewer}` **and** the `_Access` Viewer grant at `{plugin}/_Access/{viewer}_Access`. Coupon redemption and paid orders run both. Admins and automation had no such path, so enrolling by hand meant creating the record and stopping: the store counts that viewer as an owner while the permission fold still denies every gated page.

## What shipped there

`AdminEnrollment.Run(hub, pluginPath, viewer, via)` — validate → gate on `hub.IsGlobalAdmin(userId)` → call the existing `Entitlements.GrantEntitlement`. It re-implements nothing, so a third write added to a purchase tomorrow lands there for free. Two surfaces sit on it: the `/{plugin}/Enroll` admin panel and a `Store/Enrollment` control-plane node an agent creates over MCP (the same shape as `Store/Provision`). 21 in-node tests, green; the plugin compile gate is clean.

## Why the note lives here

`Store` is a Platform package installed on every mesh, so this is a portal-visible change and What's New is where portal users read about it. Strings on the new panel are localized (en + de) in the plugin's own `StoreTexts` catalog, which is that repo's translation mechanism.

Release build with CI's flags: `dotnet build src/MeshWeaver.Documentation -c Release -warnaserror` → 0 warnings, 0 errors.

### DIFF
--- a/src/MeshWeaver.Documentation/Data/WhatsNew/2026-08-08-admin-enroll-action.md
+++ b/src/MeshWeaver.Documentation/Data/WhatsNew/2026-08-08-admin-enroll-action.md
@@ -1,0 +1,22 @@
+---
+Name: Admins can enroll a user in one step
+Category: What's New
+Description: Giving somebody access to a package is now a single action — it writes the entitlement record and the access grant together, so nobody ends up "purchased but locked out".
+Icon: PersonAdd
+---
+
+# Admins can enroll a user in one step
+
+Granting somebody access to a Store package used to take two writes that had to agree: the
+entitlement record that marks them an owner, and the access grant that actually opens the gated
+pages. Buying a package did both. Doing it by hand did not — so an admin who added the record saw
+the package listed as owned while every lesson inside stayed locked, with nothing on screen
+explaining the contradiction.
+
+Enrollment is now one action. A platform admin opens **Enroll** on the package, names the person,
+and clicks once; both records are written together, exactly as a purchase writes them. Automation
+gets the same door — a scripted enrollment runs the same code path, so an onboarding script and a
+checkout can no longer drift apart.
+
+Only platform admins can enroll. Re-enrolling somebody who already has access is safe: it refreshes
+their grant in place rather than duplicating anything, and entitlements never expire.


rbuergi:
🔗 **Do not merge this ahead of its implementation.** This PR is the What's New entry only — `PluginGate.Enroll` / `Entitlements.GrantEntitlement` live in **Systemorph/MeshWeaver.Plugins** (`Store/Plugin/Source/`) as Roslyn-compiled mesh nodes, not framework assemblies, so #642's actual admin surface ships in **MeshWeaver.Plugins#370**. Merging this alone would publish a release note for a feature that is not on the instance.

Merge order: Plugins#370 first (once its CI is resolved), then this.

**Plugins#370 is currently RED** on `Store/Catalog` — `No renderer is registered for area 'Tests' on hub 'Store'`. The failure is attributed to the pinned `mw-plugin-test` image not folding in the freshly released assembly (that repo's AGENTS.md documents the symptom), not to the new code — but the diff does perturb that compile's timing, so 'untouched' could not be proven. It was deliberately NOT allowlisted and NOT re-run. That needs a maintainer call.

Also worth noting for #642's acceptance: the 'enrolled user can then read the gated content' assertion cannot be written in either repo (MeshWeaver's harness cannot call a plugin primitive; the plugin's in-node tests are pure). What is pinned instead is that both written paths derive from `Entitlements.EntitlementPath` / `PluginGate.AccessNodePath`, so they cannot drift from the primitive; the 'root grant unlocks gated children' half is already covered by `PaywallRealGateShapeTests.Buyer_CanReadGatedContent`.

PR #937, opened 2026-08-08
Catalog governance: commercial packages need a Global Admin, orphaned install records can be removed

### LINKED ISSUE #830: Commercial packages must require Global Admin to access/sync; free packages sync without special permission
## Requirement (maintainer, 2026-08-06)

> The other plugins need special permission of Global Admin to be able to access **commercial** packages ⇒ only **free** ones can be synced without permission.

## Current state

- `PluginCatalogSettingsTab` gates the whole catalog **surface** on `hub.IsGlobalAdmin(viewerId)` (`PluginCatalogSettingsTab.cs:65`) — all-or-nothing, and about *visibility*, not about the sync/install action.
- `Package` carries a `Price` (`Package.cs`), and `NodeRepoPackageSource` reads it, but **no code path distinguishes free from commercial when installing or auto-updating**.
- `PluginUpdateWatcher`'s auto-update applies a delta install with no price/permission check at all — a commercial package that is already an install record keeps updating itself regardless of who (if anyone) is entitled.

So today the boundary is drawn in the wrong place: the UI is admin-only, while the machine path is ungated.

## Proposed rule

| Package | Access / sync |
|---|---|
| **Free** (`Price` null or 0) | syncable and auto-updatable with no special permission |
| **Commercial** (`Price` > 0) | requires **Global Admin** on the installing instance to access, install, or auto-update |

Enforce it at the **action**, not just the surface:
1. `PackageInstaller` (and whatever the Store's provision funnel calls) rejects install of a priced package unless the acting principal is a global admin — under the post-#820 "the click authorizes, the SYSTEM executes" pattern, the *authorizing* principal is the one that must be admin.
2. `PluginUpdateWatcher` skips (and surfaces) priced packages whose install record was not admin-authorized, rather than silently updating them.
3. The registry endpoint (`/api/plugins`) should not enumerate commercial package payloads to unauthenticated callers.

## Tests

Production-shape (post-#804/#805/#825, as in `PluginsPartitionReadableTest`): a signed-in non-admin installs a free package → succeeds; the same principal installs a priced package → denied; a global admin installs the priced package → succeeds; auto-update applies to the free package and skips the priced one for a non-admin-authorized record.

## Context

Found while wiring `MeshWeaver.Reinsurance` in as a catalog source on memex-cloud (its modules are unpriced; the education courses are priced at CHF 900, so both classes now sit in the same catalog on one instance).

### LINKED ISSUE #840: An orphaned install record cannot be removed by anyone — no cleanup route once a package leaves the registry
### Summary

When a package disappears from the plugin registry (real case: a course folder renamed `KmuBasics` → `AgenticOffice`, making it a new product), its install record `Plugins/{oldId}` becomes permanently unremovable — by design of each individual mechanism, with no sanctioned route left.

### The lock, step by step (all working as documented)

1. `Plugins/_Policy` is a `PartitionAccessPolicy` with `create/update/delete: false`, `publicRead: true`. Per `AccessControl.md` ("the partition policy capping permissions for everyone at this scope and below"), this blocks Delete for **every** caller — verified: a caller holding an Admin `AccessAssignment` on the `Plugins` partition is denied. This is correct: install records are written only by `PackageInstaller` under scoped system impersonation.
2. The only system-identity removal surface is the catalog tab's Uninstall action — but catalog cards are driven by the registry package list (`CatalogLayoutAreas.RenderFromSource`). A package that left the registry has no card, hence no Uninstall.
3. Result: `Plugins/{oldId}` persists forever — `publicRead` keeps `/Plugins/{oldId}` rendering a phantom "installed" record whose target partition may no longer exist, and whose `id` will never return.

### How to reproduce

1. On a portal with a configured plugin registry, install any package via Settings ▸ PluginCatalog. This writes the install record `Plugins/{id}` (nodeType `Package`) under system impersonation.
2. Remove the package from the registry source — e.g. rename its folder in the backing repo, which makes it a new product id — and refresh the catalog. The catalog no longer lists `{id}`.
3. Delete the `{id}` content partition.
4. Attempt to remove `Plugins/{id}` by every available route:
   - Delete via API/MCP as a regular user → `Delete permission denied for 'Plugins/{id}'`.
   - Grant the caller an Admin `AccessAssignment` on the `Plugins` partition (`Plugins/_Access/{user}_Access`, role `Admin`) and retry → still denied; the `_Policy` caps all roles.
   - Settings ▸ PluginCatalog → no card for `{id}` exists (cards are driven by the registry package list), so no Uninstall action is reachable.
5. Observed: `Plugins/{id}` persists and `/Plugins/{id}` renders publicly (`publicRead: true`) as a phantom install record for a product that no longer exists.

Observed instance: the `KmuBasics` → `AgenticOffice` rename. On the local dev portal (2026-08-05/06), `Plugins/KmuBasics` (v16, `installedNodeCount: 86`) was undeletable by a caller holding a verified Admin assignment on the `Plugins` partition dated 2026-07-15.

### What this issue asks for

A sanctioned, system-identity cleanup route for install records whose package is absent from the registry — e.g. the catalog tab also listing orphaned install records (install record without matching registry package) with an Uninstall/Remove action, following the same scoped-impersonation pattern `PackageInstaller` already uses.

### What this issue does NOT ask for

Weakening `Plugins/_Policy` or granting user identities write access to the install registry. The policy is correct; the gap is the missing surface.

rbuergi:
Fixes #840. Fixes #830.

Two plugin-catalog boundaries drawn in the wrong place. They live in the same files, so they land in one PR.

## #830 — free syncs freely, commercial needs a Global Admin

`Package` carried a `Price` that **no install path read**. The catalog *surface* was admin-only while the machine paths were ungated: the boot-time default install and `PluginUpdateWatcher`'s auto-update applied priced packages with no permission check at all — a priced module that was once an install record kept updating itself regardless of who, if anyone, was entitled.

| Package | Access / sync |
|---|---|
| **Free** (`Price` null or 0) | installs + auto-updates, **no special permission** |
| **Commercial** (non-zero `Price`) | requires **Global Admin** on the installing instance |

The rule now sits on the **action**, not the screen:

- `PackageEntitlement.Authorize` is the single predicate (`hub.IsGlobalAdmin`, waiting for the *positive* confirmation rather than the evaluator's seeded `false`, and leaving the permission fold via `TakeDecisionOutsideGate` before any write).
- It runs inside `PackageInstaller.Install` / `InstallNodeRepoDelta` — which **every** path funnels through, so no caller can bypass it — plus once at the top of `InstallOrUpdate` so a refusal costs no fetch.
- An install runs as System for its whole lifetime (it is provisioning), so the **authorizing** principal is captured *before* the impersonation and threaded explicitly: the clicking user for a catalog install, the install record's new `AuthorizedBy` for an unattended update. **Absence fails closed** — an unattended install authorizes nothing, which is exactly "only free ones can be synced without permission".
- A refusal is **never silent**: it faults with a `PackageAuthorizationException` naming the package, its price and the principal that was asked; the watcher turns that into a notification on the install record and leaves the manual Update button. A non-admin viewer sees "Requires Global Admin" on the card instead of a button that would refuse.

**Point 3 of the issue (`/api/plugins` must not enumerate commercial payloads anonymously) needed no code**: the endpoint already fails closed on a registered instance key and scopes every listing and file fetch to that instance's `PluginGrant`. Verified in `PluginRegistryEndpoints`; noted in the doc rather than duplicated in machinery.

## #840 — a way out for an orphaned install record

A package that leaves the registry (the real case: a course folder renamed `KmuBasics` → `AgenticOffice`, making it a new product id) left `Plugins/{oldId}` stranded. Each mechanism was right on its own: `Plugins/_Policy` caps `create/update/delete` at `false` for **every** caller — the test verifies a platform admin *holding an Admin assignment on `Plugins`* still gets no Delete — and the only system-identity removal lived on a catalog card, which a departed package by definition no longer has. The record then rendered publicly, forever, as an "installed" product that does not exist.

The gap was the missing **surface**, so the policy is untouched and no user-identity delete is introduced:

- `PackageInstaller.RemoveInstalledRecord` — `ImpersonateAsSystem` scoped around the one delete, the same shape the installer's prune already uses.
- The catalog lists **Orphaned install records** (installed, not offered by the source) with a removal action gated on `hub.IsGlobalAdmin()`.
- The list is computed **only against a non-empty available list**: an empty one means "registry offers nothing" and "listing failed" indistinguishably, and an unreachable registry must never offer to remove every record.
- Removing a record does **not** delete the content it installed — separate partition lifecycle.

## Tests

`test/MeshWeaver.PluginCatalog.Test` — `MonolithMeshTestBase`, production admin shape (`ConfigureMeshBase` + an `Admin/_Access` assignment), no mocks:

- **CommercialPackageAuthorizationTest** — free installs for a non-admin *and* unattended; commercial refused for both, with no record, no content and **not one file fetched**; commercial installs for a global admin and stamps `AuthorizedBy`; the auto-update call shape (`InstallOrUpdate(..., record.AuthorizedBy)`) refuses a record with no admin authorizer, applies with one, and re-verifies rather than trusting the record.
- **OrphanedInstallRecordTest** — the lock verified rather than assumed, the orphan selection (and the empty-list guard), and that the removal deletes the record and nothing else.
- **DeclaredAccessInstallTest** now authorizes its priced install with a platform admin; the access shape it pins is unchanged.

Full project: **168/168 green** (Release). `LocalizationTest` 33/33, `DocumentationLinkIntegrityTest` green. Both new user-visible strings are in `strings.en.json` *and* `strings.de.json`.

Docs: What's New entry + a "Free syncs freely, commercial needs a Global Admin" / "Removing an orphaned install record" section in `Doc/Architecture/PluginRegistry`.

## One judgement call to flag

The issue writes commercial as `Price > 0`; this implements **non-zero** price, matching the installer's existing "priced" test (`EnsureDeclaredAccess`), where a negative price is the Store's coupon-only shape. A coupon-only package is not free either, so it needs the same admin. If you want strictly `> 0`, it is a one-line change in `PackageEntitlement.IsCommercial`.


rbuergi:
⚠️ **Cross-repo risk to check before this reaches a portal that sells content.** The #830 gate refuses a commercial install whose authorizing principal is null (fail-closed, by design). The Store plugin's purchase-provisioning path lives in the external `MeshWeaver.Plugins` repo and could not be verified from here: if it provisions priced course content through `PackageInstaller.Install` / `InstallOrUpdate` **without** threading an authorizing principal, those provisioning runs will now be refused rather than silently proceeding.

That is the intended direction (an unattended path must not install commercial content), but it is a behaviour change with a live blast radius, so it wants a look at the Plugins side — either thread the purchaser/admin as the authorizing principal, or route purchase-provisioning through the entitlement primitive (cf. #642) rather than a bare install.

Also flagging the deliberate deviation from the issue text: the gate tests `Price != 0` rather than `Price > 0`, so a coupon-only package (the Store's negative-price shape) counts as commercial and needs the same admin. That matches the installer's existing 'priced' test in `EnsureDeclaredAccess`; say the word if you want strict `> 0` instead — it is a one-line change.

PR #939, opened 2026-08-08
The course index showed one branch, never the course

### LINKED ISSUE #770: Left index shows one branch — it should show the whole course and mark the position
The collapsible left-hand index on a course page lists **only the current node's children**, so a reader standing in lesson 2 sees lesson 2's sub-nodes and nothing else. They cannot tell how long the course is, what is coming, or that they are nearly done — they navigate by the Next button and by guesswork.

**Where it is built:** `MarkdownOverviewLayoutArea` — the page is wrapped in a `Controls.NavMenu` (`WithWidth(240).WithCollapsible(true)`) whose group is fed by `host.ObserveChildren("is:main")`. Children of the current node, by construction. That is the "one branch".

**What it should be:** the whole course, with the current page **marked** rather than the list pruned to it.

- resolve the course ROOT (walk up to the partition / `Store/Plugin` root) instead of using the current node;
- query its **subtree**, not its children, grouped by module, ordered by `Order`;
- render the current path as a **marked, non-link** entry — a link to the page you are on is a dead control, and the marker should be a glyph as well as a colour so position survives for a reader who cannot distinguish the accent;
- keep the existing exclusions (`@@` embeds, `_`-satellites).

Worth pairing with the sibling issue on storing where a learner left off: once the whole index is visible, "resume" has somewhere obvious to point.

rbuergi:
Fixes #770.

The collapsible index on the left of a course page listed the current node's children, so a reader standing in lesson 2 saw lesson 2's sub-nodes and nothing else. They could not tell how long the course was, what was coming, or that they were nearly done, and they navigated by the Next button and by guesswork.

## The seam could not do its job

The core half of this landed in `3eb28c3c5` as `INodeNavigationProvider` — a seam that lets a module supply the menu for pages it OWNS, leaving docs and spaces on core's default. But it was declared **synchronous**, handing the provider only the child list core had already fetched. A whole-course index is a QUERY, and nothing on a hub may block on one (AsynchronousCalls.md), so as written the seam could only re-order or re-label the very children that ARE the bug. It had zero implementations.

So it becomes reactive:

- `GetNavigation(LayoutAreaHost)` returns `IObservable<NodeNavigation?>` — `null` declines synchronously (cheap, opens no query), and a provider that can only decide by querying emits `null` for pages it does not claim. Core `CombineLatest`s it with the node and permission streams, so the index follows adds, renames and re-orders live.
- Entries nest (`NodeNavigationEntry.Children`), so a long course groups its lessons by module instead of listing two hundred flat.
- `@@` embeds never get a side menu, so their providers are not asked at all — no query is opened for a page that cannot show the result.

## Courses supplies it

`CourseNavigationProvider` owns the semantics core must not learn: a course root is a `Course` node, its direct `Module` children are the chapters, and lessons live in the modules' `Theory` / `Example` / `Exercise` sub-namespaces.

**Two shared queries, never a probe per page.** ONE query per PARTITION (`course-roots:{partition}`) resolves which course, if any, a page sits under; ONE `scope:descendants` query per COURSE (`course-index:{course}`) IS the index. Both are keyed so every page of the partition and every page of the course subscribe to the same synced collection — a course with 200 lessons costs two live queries, not two hundred. Satellites (`_`-prefixed segments) and an exercise's own `Source`/`Test`/`Solution` code nodes are filtered the way core filters them.

**Position is stamped last**, onto exactly one entry, so the index is provably the same wherever the reader stands — navigating moves the marker, it never re-scopes or shrinks the list. A page below the index (an exercise's starter code) marks the deepest entry that contains it, so a reader always has a position. The current entry renders as marked TEXT rather than a link — a link to the page you are on is a dead control — with a `▸` glyph **as well as** an accent colour, so the marker survives for a reader who cannot distinguish the colour.

## Two fixes that fall out of making this render

- The module page embedded its theory and example children WITHOUT the `showHeader=false` the `@@` embed carries, so each block brought a full node header (Move/Copy/Delete) and its own inline comments into a page that already framed it — and would now have brought a second course index too. They are embeds; they say so.
- `LabelControl` grew a `Tooltip`, rendered as the element's `title`, so a glyph can say in words what it says visually.

## Internationalization

The glyph is language-neutral. Its `nav.youAreHere` tooltip and the `nav.contents` heading (which replaces a hard-coded English `"Contents"` that was there before) are in **both** `strings.en.json` and `strings.de.json`; `LocalizationTest` passes.

## Tests

`test/MeshWeaver.Courses.Test/CourseIndexNavigationTest.cs` — pinned twice, on the pure builder and end-to-end through the real Overview area as the browser receives it:

- from a DEEP lesson the index lists the WHOLE course, grouped by module, ordered by `Order` (modules seeded out of order to prove it) — **including the other module's lesson**, which children-of-the-current-node can never surface;
- exactly one entry is marked, and a page below the index marks the deepest entry containing it;
- `_`-satellites and exercise internals are excluded;
- the index is byte-identical at every depth — only the marker moves;
- a page outside any course gets no supplied index, so core's default stands;
- end-to-end: the nav renders as a `NavMenu` → course group → one group per module, the current lesson a `LabelControl` starting with `▸` (not a link), its sibling and the other module's lesson still `NavLinkControl`s.

**Local results** (Release, `-warnaserror`):

| Suite | Result |
|---|---|
| `MeshWeaver.Courses.Test` (incl. 7 new) | 24/24 pass |
| `MeshWeaver.Graph.Test` | 824/824 pass |
| `MeshWeaver.Layout.Test` | 343/343 pass |
| `LocalizationTest` | 33/33 pass |
| `DocumentationLinkIntegrityTest`, `MarkdownEmbedAnnotationsTest` | pass |

### DIFF
--- a/src/MeshWeaver.Blazor/Components/Label.razor
+++ b/src/MeshWeaver.Blazor/Components/Label.razor
@@ -11,6 +11,7 @@
         Typo="@Typo"
         Weight="@Weight"
         Style="@Style"
+        title="@InternalTooltip"
         @onclick="OnClick">
         @InternalData
     </FluentLabel>
@@ -22,7 +23,8 @@ else
         Disabled="@Disabled"
         Typo="@Typo"
         Weight="@Weight"
-        Style="@Style">
+        Style="@Style"
+        title="@InternalTooltip">
         @InternalData
     </FluentLabel>
 }
@@ -30,6 +32,7 @@ else
 @code
 {
     private string? InternalData { get; set; }
+    private string? InternalTooltip { get; set; }
     private HorizontalAlignment? Alignment { get; set; }
     private bool Disabled { get; set; }
     private Typography Typo { get; set; } = Typography.Body;
@@ -39,6 +42,7 @@ else
     {
         base.BindData();
         DataBind(ViewModel.Data, x => x.InternalData );
+        DataBind(ViewModel.Tooltip, x => x.InternalTooltip);
         DataBind(ViewModel.Disabled, x => x.Disabled);
         DataBind(ViewModel.Alignment, x => x.Alignment);
         DataBind(ViewModel.Typo, x => x.Typo);
--- a/src/MeshWeaver.Courses/CourseNavigationProvider.cs
+++ b/src/MeshWeaver.Courses/CourseNavigationProvider.cs
@@ -1,0 +1,230 @@
+using System.Reactive.Linq;
+using MeshWeaver.Courses.Configuration;
+using MeshWeaver.Graph;
+using MeshWeaver.Layout.Composition;
+using MeshWeaver.Mesh;
+using MeshWeaver.Messaging;
+
+namespace MeshWeaver.Courses;
+
+/// <summary>
+/// Supplies the left-hand index for pages that sit inside a course: the WHOLE course, grouped by
+/// module, with the reader's position marked — not the branch they happen to be standing in.
+///
+/// <para>Core's default lists the current node's children, so a reader in lesson 2 sees lesson 2's
+/// sub-nodes and nothing else: they cannot tell how long the course is, what is coming, or that
+/// they are nearly done. Courses owns the semantics core must not learn — that a course root is a
+/// <c>Course</c> node, that its direct <c>Module</c> children are the chapters, and that lessons
+/// live in the modules' <c>Theory</c> / <c>Example</c> / <c>Exercise</c> sub-namespaces — so it
+/// hands core the finished index through <see cref="INodeNavigationProvider"/>.</para>
+///
+/// <para><b>Two shared queries, never a per-page probe.</b> Which course (if any) a page belongs to
+/// comes from ONE query per PARTITION, keyed <c>course-roots:{partition}</c> — every page in the
+/// partition subscribes to the same synced collection. The index itself is ONE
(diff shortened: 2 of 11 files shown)


rbuergi:
Note on the local test table: `MeshWeaver.Layout.Test` first came back 342/1 — `EditorTest.TestEditorWithDelayed` blew its 30 s budget while two other `dotnet test` processes and a Release build were competing for the same cores. That test drives five sequential `Thread.Sleep(100)`-blocked remote round-trips and its own comment records a prior de-flake for CI load. The diff cannot reach it (its path is `EditorControl`/`StackControl`/`Controls.Markdown` — no `LabelControl`, and `Tooltip` is never set), and the suite is 343/343 in 31 s when run alone. Recording it rather than quietly re-running it green.

rbuergi:
Checked the `using MeshWeaver.Graph.Configuration;` finding and it is **incorrect** — the directive is load-bearing. Removing it and rebuilding `MeshWeaver.Courses.Test` with `-c Release -warnaserror` produces **4 errors**; restoring it builds clean. Types from that namespace are referenced unqualified in the file, which is why a grep for `Configuration.` finds nothing. Leaving it in place; not resolving this thread as 'fixed' because there was nothing to fix.

Issue #891, opened 2026-08-07
NodeType watchers die silently on own-stream OnError — log-and-die handlers violate wedges-to-zero


rbuergi:
Latent hazard found during the #612 investigation (provably NOT #612's mechanism, so not fixed there).

`InstallCompileWatcher`, `InstallReleaseRequestWatcher`, and `InstallSourcesWatcher` (src/MeshWeaver.Graph/Configuration/NodeTypeCompilationHelpers.cs ~274/~695/~908) all subscribe the per-NodeType own stream with `Subscribe(onNext, ex => log)`. If the own stream ever faults (e.g. a `TypedContentObserver` deserialization fault propagating OnError), the watcher subscription terminates permanently and silently: compiles stop being kicked off, release requests stop being honored, and the only trace is a single warning line. That is a wedges-to-zero violation — the hub keeps running with its control-plane watchers dead.

Per the no-band-aids rule the fix is NOT a resubscribe watchdog. Candidates, in preference order:
1. Prove the own stream cannot OnError (make faults flow as values / route to the hub's failure sink) — then the handlers' death path is unreachable and can assert loudly instead.
2. If the stream can legitimately fault, a watcher death must take its hub down loudly (fail the hub / surface a health signal), never continue half-alive.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #940, opened 2026-08-08
A compile problem now says whether it is your code or a timeout — and links the log

### LINKED ISSUE #641: Compile-failure overlay: show the compiler log; do not render lookup timeouts as compile failures
Users saw 'compilation failed' everywhere with no error text; half were real (CS0246), half were Store/Plugin type-lookup TimeoutExceptions rendered identically. Overlay should show the actual error + link the compile activity, and distinguish timeout from failure.

From the 2026-07-24 plugin-install retrospective.

rbuergi:
Fixes #641.

## Already on main (not redone here)

The audit on the issue established that the compiler output itself is already
surfaced: `BuildCompilationErrorMarkdownText`'s diagnostics fence plus the
per-source Monaco with Roslyn markers (`NodeTypeLayoutAreas`), the
`CompileProgressIndicator`, the `{TypeName}: {Message}` treatment for a
non-Roslyn abort in `RunCompile`, and ONE timeout path that already reads
correctly — the enrichment settle timeout ("has not finished compilation yet…
no code change is needed"). None of that is touched.

### 1. The broken INSTANCE page never linked the compile activity

`WithCompilationErrorOverlay` never received `LastCompilationActivityPath`, so
the "View full compile log →" anchor existed only on the NodeType's own
Overview — the *instance* page, the one a user actually lands on, had no route
to the diagnostics at all. The path is now threaded through to the markdown
builder and rendered as a markdown link (never an embedded `LayoutAreaControl`:
the activity is history and may be unaddressable — subscribing to an inexistent
address is the resubscribe storm). It rides along on every overlay that has a
type node in hand, and it matters most on a timeout, where the log is the only
evidence of how far the build got.

### 2. Timeouts and lookup misses no longer render as compile failures

Every overlay call site that passed no `guidance` silently inherited *"There was
a **compilation error** in this item's code… **Please correct the code.**"*
Four causes reached it that way, none of them the author's code:

- the **type-registration probe** (~3 s) folded a FAULT into "type is not
  registered", so an indeterminate lookup and a genuinely missing type produced
  identical copy. Split into `Registered` / `Missing` / `Indeterminate`;
- **pinned-release / assembly-store misses** — an assembly this node cannot
  fetch is a lookup miss, not broken source;
- **framework-stale** — whose guidance already said "no source changes are
  needed" *under* a lead-in demanding a code fix;
- an **enrichment chain that faulted** outright, before any build was consulted
  (it now also names the exception TYPE, like `RunCompile` does).

The wording is no longer implicit: `NodeTypeEnrichmentHelpers.OverlayCause` +
`OverlayCopy` are the single place that decides it, and `ImpliesCodeFix` pins
which causes may ask for an edit. Every availability cause leads with **"No code
change is needed for this."**

The contract handler's 60 s settle timeout fell straight to the Subscribe error
sink as `Fail("The operation has timed out.")` — naming neither what timed out
nor that it *was* a timeout — and once that reached the write-back it persisted
`CompilationStatus = Error`. It now folds into the same write-back with a
message that names the type and the budget.

### 3. Structural: `CompilationStatus.Unavailable`

The enum had no way to say "could not determine", so a timeout could only be
persisted as `Error`. `Unavailable` is appended **last** (every existing
persisted ordinal unchanged) and is a MARKER, never an answer: the write-back
refuses to let it overwrite a state another driver owns (`Pending`/`Compiling` —
that compile writes its own terminal state; clobbering it would strand the
dispatch), a strictly better one (`Ok`), or a never-compiled `null` the
first-build kickoff gates on. `EnsureCompileDispatched` treats it as "never
determined", so the next **request** — never a timer — dispatches a fresh
compile rather than leaving the type parked.

Every reader was checked (`grep CompilationStatus`) so the new member cannot be
mistaken for a failure or hang a waiter:

| Reader | Handling |
|---|---|
| enrichment `settled` predicate, `IsRecompileSettled` | terminal — waiting would burn the whole no-progress budget for a write that is not coming |
| `ApplyStreamResult` | dedicated branch: retry copy + compile-log link |
| MCP `GetDiagnostics` / `compile` (`MeshOperations`) | settled, and reported as "NOT a compile failure" instead of hanging or blaming the source |
| NodeType progress page / status badges / compile panel (`NodeTypeLayoutAreas`) | own undetermined state (warning tint, "Retry compile"), never the error tint |
| `DynamicTypePreWarmer` | terminal; reported `TimedOut`, never `CompileError` |
| `NodeTypeBakeStatus` | deliberately unchanged — `Unavailable` is not `PreviouslyBroken` and stays probeable |
| `AwaitCompilationSettled` (`MeshDataSource`) | already correct (only Pending/Compiling are in-flight) |

## Tests

- `CompileErrorPageTest` — the copy invariant is **enumerated over
  `OverlayCause`**, so a cause added later cannot quietly inherit the code-fix
  wording; plus the settle-timeout page, the genuine-failure page, and the
  compile-log link (present / absent).
- `RecompileSettlePredicateTest` — `Unavailable` is terminal, but never
  satisfies `requireUsableBuild`.
- `CompileErrorOverviewTest` — end to end: an `Unavailable` type's instance
  Overview reads as retry and links the log and never says "correct the code";
  a genuine Roslyn failure still shows its diagnostics, still asks for the fix,
  and now links its log too.

Local runs (Release): Graph.Test 831 passed · Content.Test/CompileErrorOverview
4 passed · Messaging.Hub.Test Localization 33 passed ·
Documentation link integrity passed · Monolith CompileSingleDriverConsistency +
MeshNodeCompilationIntegration passed. All touched projects build clean with
`-c Release -warnaserror`.

## Known gap

The overlay's prose (headline, guidance paragraphs) remains English-only, as it
was before this PR — localizing it means deferring the whole copy to render time
and migrating the pre-existing constants, which is a separate change. The two
NEW discrete strings this PR adds (`ui.viewCompileLog`,
`ui.compileStateUnknown`) are in both `strings.en.json` and `strings.de.json`,
and the previously hard-coded "View full compile log →" on the NodeType page now
reads from the catalog. `CompileProgressIndicator` (the global banner) still
surfaces only in-flight compiles and hard failures — an `Unavailable` type shows
nothing there rather than an error-styled banner, which the page overlay covers.

### DIFF
--- a/src/MeshWeaver.AI/MeshOperations.cs
+++ b/src/MeshWeaver.AI/MeshOperations.cs
@@ -3261,8 +3261,9 @@ public class MeshOperations
 
     /// <summary>
     /// True when the NodeType's <see cref="NodeTypeDefinition.CompilationStatus"/>
-    /// has reached a terminal state (<see cref="CompilationStatus.Ok"/> or
-    /// <see cref="CompilationStatus.Error"/>). Pending and Compiling are
+    /// has reached a terminal state (<see cref="CompilationStatus.Ok"/>,
+    /// <see cref="CompilationStatus.Error"/> or
+    /// <see cref="CompilationStatus.Unavailable"/>). Pending and Compiling are
     /// transient — readers should keep waiting for the watcher's settle
     /// write rather than report a half-baked state.
     /// </summary>
@@ -3277,7 +3278,11 @@ public class MeshOperations
     /// the compile-settle wait instead of hanging until the 5s timeout.
     /// </summary>
     private static bool IsSettled(CompilationStatus? status)
-        => status == CompilationStatus.Ok || status == CompilationStatus.Error;
+        => status is CompilationStatus.Ok
+                  or CompilationStatus.Error
+                  // Terminal too: a driver already gave up determining the state, so no
+                  // further write is coming and a waiter would hang out its whole budget.
+                  or CompilationStatus.Unavailable;
 
     private string FormatDiagnosticsFromDef(
         Graph.Configuration.NodeTypeDefinition def, string nodeTypePath)
@@ -3286,7 +3291,9 @@ public class MeshOperations
         return FormatDiagnostics(
             status,
             nodeTypePath,
-            error: status == CompilationStatus.Error ? def.CompilationError : null,
+            error: status is CompilationStatus.Error or CompilationStatus.Unavailable
+                ? def.CompilationError
+                : null,
             startedAt: status == CompilationStatus.Compiling ? def.LastCompileStartedAt : null,
             lastCompiledAt: status == CompilationStatus.Ok ? def.LastCompileSucceededAt : null,
             hub.JsonSerializerOptions);
@@ -3417,8 +3424,7 @@ public class MeshOperations
                         .Where(n =>
                         {
                             var status = ReadCompilationStatusFromNode(n);
-                            return (status == CompilationStatus.Ok || status == CompilationStatus.Error)
-                                && IsNewCompileRun(n, before);
+                            return IsSettled(status) && IsNewCompileRun(n, before);
                         })
                         .Take(1)
                         .Timeout(TimeSpan.FromSeconds(60))
@@ -3434,11 +3440,18 @@ public class MeshOperations
                                     path = resolvedPath,
                                     error,
                                     activityPath,
-                                    message = status == CompilationStatus.Ok
-                                        ? "Compile SUCCEEDED."
-                                        : "Compile FAILED — see `error` for Roslyn diagnostics. "
-                                          + "Full source-discovery + matched-Code-paths trace lives at "
-                                          + (activityPath ?? "(no activity log written)") + "."
+                                    message = status switch
+                                    {
+                                        CompilationStatus.Ok => "Compile SUCCEEDED.",
+                                        // NOT a failure — the run never reported an answer.
+                                        CompilationStatus.Unavailable =>
+                                            "Compile state could NOT BE DETERMINED (see `error`) — this is a "
+                                            + "timeout, NOT a compile failure. Nothing is known to be wrong "
+                                            + "with the source; trigger the compile again.",
+                                        _ => "Compile FAILED — see `error` for Roslyn diagnostics. "
+                                            + "Full source-discovery + matched-Code-paths trace lives at "
+                                            + (activityPath ?? "(no activity log written)") + "."
+                                    }
                                 },
                                 hub.JsonSerializerOptions);
                         })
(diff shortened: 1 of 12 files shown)


rbuergi:
On the rolling-deploy exposure you flagged (an old pod deserializing `"Unavailable"` fails the `NodeTypeDefinition` parse): that one is worth closing rather than accepting, because the blast radius is not limited to the new field — `System.Text.Json`'s enum converter throws on an unknown *value*, so the whole node fails to materialize, and a NodeTypeDefinition that will not deserialize is exactly the poisoned-content shape #891 just had to defend the compile watchers against.

Two options, either fine:
1. Make `CompilationStatus` deserialization tolerant — an unrecognized name maps to `Unknown` (which every reader already treats as 'never determined'), so an old pod degrades to 'don't know' instead of failing the parse. This is the schema-evolution twin of `[JsonExtensionData] UnknownMembers`, which already protects unknown *members* on that same type.
2. Or ship the reader tolerance one release BEFORE anything writes the new value, and note the ordering in the PR.

Option 1 is self-contained and removes the window entirely; worth doing here unless it turns out the converter is shared in a way that makes it risky. Not blocking my read of the rest — the enum-as-marker design (never overwrites Pending/Compiling/Ok/null, request-driven redispatch, no timer) is the right shape.

rbuergi:
Worked all three review findings; verdicts differ per finding.

**1. Hand-built HTML anchor — VALID, fixed (6116854a5).** You were right: this PR added an `<a>` string inside an existing `Controls.Html` call, which AGENTS.md forbids absolutely, and it forced manual `HtmlEncode` on both href and label — the escaping drift the rule exists to prevent. Now `Controls.Markdown` with the styling on the control; 19/19 green. Noting for the record that the two SIBLING `Controls.Html` anchors in the same method (release-published link, assembly-store line) are pre-existing and deliberately left alone — same violation, but they want their own sweep rather than widening this PR.

**2. Hard-coded English in the Unavailable status copy — VALID but pre-existing style, deliberately not fixed here.** The whole `switch` it joins is English constants (Compiling, Pending, Error, Waiting, No-compile-required all hard-code their label and prose). Localizing only the new arm would leave the block half-translated and inconsistent; localizing the block means deferring every constant to render time, which is a separate change the PR body already discloses. The genuinely NEW discrete strings (`ui.viewCompileLog`, `ui.compileStateUnknown`) ARE in both catalogs.

**3. The write-back guard — accurate observation, deliberate design.** You are right that in the most likely settle-timeout case the status is `Pending`/`Compiling` and nothing is persisted. That is the intent: a compile IS in flight and owns the terminal write, so stamping `Unavailable` over it would strand the dispatch. `Unavailable` earns its keep on the paths where no compile is running (the type-existence probe, pinned-release misses). Overwriting a stale `Error` is likewise intended — the latest attempt could not determine the state, so 'unknown' is truer than a stale 'failed'.

Issue #476, opened 2026-07-14
Agent rounds have no model with quota; the stale-model fallback is silent for non-interactive rounds and never health-checks its target


rosalie1984:
## Summary

Two related problems, one operational and one in the code:

1. **No language model on `memex.systemorph.com` has usable quota.** Every agent round on the portal currently dies. It does not have to be `glm-5.2` — we just need **at least one deployed model with real quota** that agent rounds can rely on.
2. **The stale-model fallback is effectively silent for non-interactive rounds, and it falls back onto a model it never checks the health of.** A round records the model the caller asked for while a *different* model actually answers — and that different model is the one that is out of quota.

## 1 · Model quota (ops)

Agent rounds on the portal fail with:

```
Error: Your requests to DeepSeek-V4-Flash for DeepSeek-V4-Flash in swedencentral
have exceeded rate limit. Status: 429 (Too Many Requests) ErrorCode: RateLimitReached
```

and, when a non-default deployment is selected, with a 500 from the managed endpoint:

```
Status: 500 (Internal Server Error)
{"object":"error","message":"Internal server error: 'NoneType' object has no attribute 'items'", ...}
```

There are also **zero `LanguageModel` nodes in the mesh** (`namespace:Provider scope:descendants nodeType:LanguageModel` → 0, same for the user namespace), so the catalog is entirely deployment configuration — nothing can be fixed from the mesh side.

**Ask:** one (any) model deployment with quota sized for background agent rounds, not just interactive chat.

## 2 · The silent fallback (code)

`AgentChatClient.ApplyStaleModelFallback()` (src/MeshWeaver.AI/AgentChatClient.cs:1865) does this:

- if the selected model does not resolve (provider/key missing), swap `currentModelName` to `resolver.ResolveDefaultModelId()`;
- set `staleModelNotice`, which is later yielded **as an assistant chat message** ("model X unavailable — using default Y", AgentChatClient.cs:721).

Three gaps:

**(a) The notice only exists in the chat stream.** For a round started programmatically (`hub.StartThread(...)` from code — background jobs, watchers, scheduled gathers) nobody is reading the chat transcript. The swap is, in practice, silent.

**(b) The persisted round metadata still names the requested model.** `ThreadComposer.ModelName` and `ThreadMessage.ModelName` keep the *asked-for* id, so the record claims a model that never ran:

```jsonc
"composer": { "agentName": "Assistant", "modelName": "glm-5.2" },   // what we asked for
[...]

PR #944, opened 2026-08-08
Localize the no-usable-model round failure (follow-up to #938)

rbuergi:
Addresses #476 (code half), follow-up to #938 — which was merged before this fix could land on it. The ops half of the issue (a deployment with real model quota) still stays with the maintainer.

## What #938 got wrong

The Copilot review on #938 was right: `noUsableModelError` was composed as a **hard-coded English sentence** inside `AgentChatClient.CreateAgentsSync` and written verbatim to the response cell's `Text` + `Summary`. That is a user-visible string, so it renders English for a German viewer — the i18n rule in AGENTS.md. It also carried the raw factory error (`"ApiKey is missing for model 'X'"`) into the thread, which is exactly the raw provider dump #476 complains about.

## The fix — split the CONDITION from its PRESENTATION

- `AgentChatClient` now latches only the condition and its inputs: `HasNoUsableModel`, `RequestedModelName`, and `NoUsableModelDetail` (the English factory error, **log-only**).
- `ThreadExecution` builds the prose at write time, where the round's own `AccessContext` supplies the viewer's language:
  `LocalizationCatalog.Get("chat.noUsableModel" | "chat.noUsableModelNoSelection", userAccessContext?.Locale, requested)` — an **explicit** locale, never ambient `CultureInfo` (a round hops schedulers, so an AsyncLocal culture would render one user's failure in another user's language).
- The raw factory detail moves to the `NO_USABLE_MODEL` log line only — it never reaches the thread.
- Keys added to **both** `strings.en.json` and `strings.de.json` (`LocalizationTest` gates the pair).

`ModelSubstitutionTest.NoUsableModelAnywhere_FailsRoundWithSpeakingError` now asserts the cell's `Summary` **is** the catalog message for the requested model, and that no `ApiKey` text reaches the thread — so the localization wiring itself is pinned, not just the English words.

Local (Release, `-warnaserror`): `MeshWeaver.AI` + `MeshWeaver.AI.Test` build clean; ModelSubstitution / ThreadTokenUsage / RoundFaultTerminalState 9/9; `MeshWeaver.Messaging.Hub.Test` localization 33/33. On the pre-cherry-pick branch the same change ran the full `MeshWeaver.AI.Test` (1053 passed), `MeshWeaver.Threading.Test` (128), `MeshWeaver.Hosting.Orleans.Test` (136).

### DIFF
--- a/src/MeshWeaver.AI/AgentChatClient.cs
+++ b/src/MeshWeaver.AI/AgentChatClient.cs
@@ -66,7 +66,7 @@ public class AgentChatClient : IAgentChat
     // keys in factory config, invisible to the credential resolver) — it becomes fatal only
     // when agent creation ALSO produced nothing, which is what latches noUsableModelError.
     private bool modelFallbackExhausted;
-    private string? noUsableModelError;
+    private bool noUsableModel;
     private string? persistentThreadId;
     private IReadOnlyList<string>? currentAttachments;
     private bool isPersistentFactory;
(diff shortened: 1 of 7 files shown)


rbuergi:
The review finding is **correct and load-bearing** — I verified it by making the change it asks for, and the test then FAILED.

Setting a German submitter locale and asserting the German catalog text produces: expected the German message, got the English one. So the `locale: null` comparison was hiding a real defect, exactly as flagged.

**Root cause is upstream of this PR** (filed as **#948**): `ThreadSubmission` reconstructs the round's identity from the submitter rider when the AsyncLocal context is wiped on the Throttle/Subscribe continuation —

```csharp
userCtx = new AccessContext { ObjectId = sid, Name = submitter.SubmitterName ?? sid };
```

— and `Locale` is not carried. Per that code's own comment this is the USUAL path, so every user-facing string a round resolves from `userAccessContext?.Locale` renders English for every viewer. That is broader than #476: it applies to any round-emitted localized text.

I reverted my test change so it does not block this PR, which is otherwise correct — it does route the message through the catalog, which is the necessary half. The fix (a `SubmitterLocale` rider stamped where the submitter is captured) plus the un-vacuumed test belong in #948.

Resolving this thread as 'valid, tracked elsewhere' rather than 'fixed here'.

Issue #701, opened 2026-07-28
Bind plugin access policy to the NodeType instead of materialising _Policy + per-child denies


rbuergi:
## The model we want

For a `Store/Plugin` node, access should be decided **by the type**, not by rows written under each plugin:

1. **Any request to a plugin page by someone without an entitlement → redirect to the plugin's Overview / marketing page**, and from there to checkout. (This is what `marketingPath` + `redirectOnDenied → {plugin}/Subscribe` already express.)
2. **Overview and checkout are anonymous** — no sign-in to see the cover or start a purchase.
3. **No further restrictor is needed.** No per-child `AccessAssignment` denies, no `_Policy` node.
4. **Purchase (or coupon) grants access** — the entitlement record is the single switch that opens the whole subtree.

Note this **inverts** today's logic. Currently the root carries `Public` + `Anonymous` **Viewer** and every non-public child carries an explicit **deny** — allow-then-deny, O(children) rows, re-derived on every sync. The model above is deny-by-default with exactly two declared anonymous exceptions (marketing page, subscribe surface) plus `publicSegments`.

## Why the current materialised approach has to go — measured on memex, 2026-07-28

- **Churn.** `_Policy` version numbers: `AgenticEngineering` **254,760**, `Underwriting` **217,403**, `Claims` **202,448** — all `lastModifiedBy: system-security`. These nodes are rewritten continuously as pure bookkeeping.
- **Drift, right now.** Every one of those `_Policy` nodes contains only `redirectOnDenied` and **no `publicRead: false`** — while `PluginGate.IsGatedPolicy` requires *both* (`publicReadOff && paywall`). Written state and read state disagree in production.
- **Silent non-application.** `Underwriting` and `Claims` were **completely ungated** — every page anonymously readable — because gating keys off `price`, and both shipped with `price: null` and no `contactEmail`. A type-declared rule cannot be "not run".
- **New children are a race.** "A newly synced child is born gated" is a hope that depends on a gating pass, not an invariant.

## The hook already exists

Type-bound access is already first-class in **both** evaluators. The Postgres RLS predicate consults it directly:

```sql
EXISTS (SELECT 1 FROM admin.node_type_permissions ntp
        WHERE ntp.node_type = n.node_type AND ntp.public_read = true)
```

[...]

Issue #946, opened 2026-08-08
CI: ObservableQueryIntegrationTests.ScopeDescendants_NotifiesOnAllDescendantChanges fails on loaded shards, always shard 1


rbuergi:
Split out of #826 (which tracks the different "8 distinct tests, once each" shape). This is a **named test failing repeatedly on unrelated PRs**, always the same shard.

## Evidence (2026-08-08)

`MeshWeaver.Query.Test.ObservableQueryIntegrationTests.ScopeDescendants_NotifiesOnAllDescendantChanges`, **shard 1** both times:

| PR | Run | What the diff touched |
|---|---|---|
| #927 | 31238512895 | NodeType watcher subscriptions |
| #940 | 31245604519 | compile-overlay copy + one `Controls.Html` → `Controls.Markdown` swap |

Neither diff touches query subscription, the change feed, or subtree scoping. Both pass locally on the same head.

## Why it matters beyond CI noise

The test pins that a `scope:descendants` query **notifies on every descendant change**. That is the same liveness contract behind the live-revocation family (#697/#868/#889): a subtree query that misses a change is a security-relevant staleness bug, not a cosmetic one. If the notification is genuinely racy under load, the product property is racy — the test is the messenger.

## Suggested starting points

- Whether the assertion depends on receiving a *specific number* of change events, or on a filtered shape. AGENTS.md is explicit that counting emissions on a pg_notify-backed / change-feed stream races the initial-snapshot path — if this test counts, the fix may be the assertion; if it filters correctly, the race is in the feed.
- Whether the subtree notification path has a window between the initial snapshot and the live subscription where a change can land unobserved (the pre-Initial buffering that `FanOutQuery` and its two siblings each had to fix — see PR #905's write-up for that pattern).
- Same shard every time suggests a specific co-tenant in shard 1 is the load trigger; the shard composition is worth a look.

## Repro status — honest

Not reproduced locally (18-core machine, passes comfortably). It needs a saturated runner. Instrumenting the notification path is likely to be more productive than local re-runs.

## Cost

Blocks a green PR and burns a CI cycle on every hit; it struck two unrelated PRs in one morning.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #945, opened 2026-08-08
CI: CrossHubPatchAtomicityTest drops concurrent cross-hub adds on loaded shards — the ack path is bounded, the merge is not the suspect


rbuergi:
Split out of #826, which tracks a different shape ("8 distinct tests each failed exactly once"). This one is a **named test failing repeatedly on unrelated PRs**, always on the same shard.

## Evidence (2026-08-08)

`MeshWeaver.AI.Test.CrossHubPatchAtomicityTest.ConcurrentCrossHubPatches_DoNotDropAQueuedMessage`, **shard 4** both times:

| PR | Run | What the diff touched |
|---|---|---|
| #927 | 31238512895 | NodeType watcher subscriptions |
| #930 | 31243241054 | agent-picker query resolution |

Neither diff goes near cross-hub patching. Both pass locally on the same head, first attempt.

## Why this is not "just a flaky test"

The test **asserts a product invariant**: 6 mirrors × 48 concurrent cross-hub `stream.Update` adds against one node, four rounds, and every key must survive. A failure means the owner genuinely dropped acked writes — the exact loss class #590/#648/#667 were about. The test doing its job under load is more likely than the test being wrong.

## What I ruled OUT

The **merge is atomic**. `DataExtensions.ApplyMeshNodePatchInTurn` performs the read-merge-write inside `primary.Update(...)` on the owner's action-block turn — a pure, bounded transform with no IO and no nested subscribe, explicitly written that way so back-to-back patches each merge onto the freshest node. Sibling keys cannot be lost there.

## The remaining suspect: the bounded ACK path

In the same function, success is acked OFF-turn by `postSub`:

- it waits for a post-commit emission containing the stamped write, `.Timeout(20s)`;
- then `IPostCommitFlush.Flush(...)`, `.Timeout(10s)`.

On a saturated shard those bounds are reachable. Two things to check, in order:

1. **Does an expired ack turn into a lost write, or only a lost acknowledgement?** If the merge committed and only the ack timed out, the key should still be present and the test would pass — so a MISSING key means either the merge never ran for that patch, or the client never sent it. Instrument which.
2. **Does the client retry on a NACK/timeout, and if so against what base?** A retry that re-diffs against a stale mirror base could emit a patch that replaces `PendingUserMessages` wholesale instead of adding one key — that WOULD drop siblings. Worth confirming the diff is always a minimal per-key delta under contention.

## Repro status — honest

[...]

PR #951, opened 2026-08-08
Bind the plugin paywall to its NodeType instead of materialising _Policy + per-child denies

rbuergi:
Addresses #701.

Binds a gated plugin's access shape to its **NodeType** instead of materialising a `_Policy` node
plus a deny row per child. This is the MeshWeaver-side framework half; the Store plugin in
`Systemorph/MeshWeaver.Plugins` has to adopt it (see **What the Plugins repo must do** below), so
this does not close the issue on its own.

## The model implemented

```csharp
builder.ConfigureNodeTypeAccess(access => access.WithGate(new NodeTypeGate("Store/Plugin")
{
    PublicSurfaces = [NodeTypeGate.Self, "Overview", "Subscribe"],
    RedirectOnDenied = "Subscribe",
}));
```

Every node of the declared type keeps its cover (`Self`), its marketing page and its checkout
surface readable by **everyone including anonymous visitors**; a reader denied anywhere beneath it
is sent to `{plugin}/Subscribe`. Nothing is written per plugin — no `_Policy`, no root grant, no
per-child deny.

The other three requirements fall out of machinery that already exists:

| Requirement | Mechanism |
|---|---|
| Everything except the declared surfaces is closed | The framework's deny-by-default — no grant, no Read |
| Purchase / coupon opens the whole subtree | ONE `Viewer` `AccessAssignment` at the plugin root; grants inherit downward |
| Gated even with `price: null` | The gate is on the TYPE — there is no price in the rule at all |

### Two properties this leans on

**A gate only ever GRANTS.** It never denies, never caps, never removes a permission a role or an
entitlement confers. A declaration that can only widen a short, explicitly listed set of paths
cannot lock anyone out and cannot regress an existing deployment — which is why the older
allow-then-deny gate keeps working unchanged beside it, and why an actual `_Policy` still WINS over
the type-declared redirect (a deployment mid-migration keeps every hand-tuned policy it carries).

**`Self` opens the node and nothing beneath it.** That asymmetry is *why* this has to be on the
type: an `AccessAssignment` at the plugin root inherits strictly downward, so opening the cover
that way opens the whole subtree — precisely why the materialised shape needed a deny on every
non-public child to claw it back.

### Cost

`PermissionEvaluator` resolves a path's nearest gated ancestor-or-self from one process-wide cached
query **per gated node type** (`$security-gated:{type}`) — bounded by the number of gated *nodes*,
not their children — seeded from the static providers so a statically declared plugin resolves on
the first emission. **A mesh that declares no gate subscribes nothing and runs the exact fold it
ran before**; every call site short-circuits on `gates.Count == 0`. That is also what makes this
safe to merge ahead of the Plugins-repo half: with no declaration anywhere, behaviour is unchanged.

The per-type query is `StartWith`-seeded because it feeds the `CombineLatest` gating every
permission check — an unseeded source that never matched (no such node yet) would stall the fold
and hang every read. Seeding also starts *stricter* (no gated nodes known ⇒ no public surface ⇒
deny), so the pre-load window can only be more restrictive, never a bypass.

## Two landmines found while mapping this — reported, deliberately not tripped

**1. `NodeTypePermission` never reaches the database.** The issue cites
`node_type_permissions` / `ConfigureNodeTypeAccess(a => a.WithPublicRead(...))` as a hook that is
"already first-class in both evaluators". It is not. The only DI→DB bridge,
`PostgreSqlExtensions.InitializePostgreSqlSchemaAsync`, has **zero callers** across `src/`,
`memex/` and `tools/` — every `SyncNodeTypePermissionsAsync` call in the repo is from a test. So the
`public_read` clause in all three RLS predicates evaluates against an empty table in production.
The in-memory `PermissionEvaluator` doesn't consult `NodeTypePermission` either (0 references).
The hook is inert on **both** sides.

**2. …and wiring it up would open paid content.** ~25 node types declare `WithPublicRead`,
including `Course`, `Module`, `Exercise`, `Thread`, `Skill` and `Agent`. Populating the table today
would grant public read on all of them at once to every authenticated user with partition access —
i.e. it would unlock exactly the paid course content this issue is trying to protect. That is why
this PR builds a **separate, purely additive** declaration rather than extending `PublicRead`, and
why the dead bridge is left dead pending a decision on each of those 25 types.

## Deliberate gap: the SQL fold does not know about gates

Postgres RLS decides *query listing*; `PermissionEvaluator` decides *exact reads and every UI gate*
(`AnonymousGate`, `RlsNodeValidator`, `MeshOperations.Get`, `NamedAreaView`, layout areas). Only the
latter is implemented here.

Consequence: a declared public surface is readable by path but will not appear in an anonymous
`search`. **The asymmetry is strictly in the safe direction — SQL is stricter, never looser — so it
cannot become a paywall bypass**, but it is a real gap and it is the one acceptance criterion of
#701 not met. Closing it needs a new per-schema table + migration + the clause in three SQL sites
(`PostgreSqlSqlGenerator.GenerateAccessControlClause`, `BuildPerSchemaAccessClause`, and the
`search_across_schemas` stored proc); it cannot reuse `node_type_permissions` for landmine 2 above.
Documented inline in `AccessControl.md` rather than left to be discovered.

## What the Plugins repo must do (`Systemorph/MeshWeaver.Plugins`)

`PluginGate` lives at `Store/Plugin/Source/PluginGate` and is not visible from this repo. To take
this up:

1. **Declare the gate** on the `Store/Plugin` NodeType configuration — `PublicSurfaces` from
   `MarketingPath` + the subscribe surface + `publicSegments`, `RedirectOnDenied = "Subscribe"`.
2. **Delete the materialisation**: stop writing `{plugin}/_Policy` and stop writing the
   Public/Anonymous Viewer `denied: true` rows per child. This is what takes the measured churn
   (`AgenticEngineering` `_Policy` at version 254,760, every write by `system-security`) to zero.
3. **Stop writing the root Public/Anonymous Viewer GRANT.** It is what makes the per-child denies
   necessary; without it the subtree is closed by default and the cover is opened by the type.
4. **Delete `IsGatedPolicy`.** Its two-condition read (`publicReadOff && paywall`) against policies
   that only ever carried `redirectOnDenied` is the writer/reader disagreement measured in
   production — there is no policy to interrogate any more, so the predicate has nothing to drift
   from. This is the "fix the read/write agreement explicitly" item: it is fixed by removing the
   second condition's subject entirely, not by relaxing the check.
5. **Keep `Enroll` exactly as it is** — one root `Viewer` grant is still the entitlement, and it
   still opens the whole subtree.

Until step 1 lands, this PR changes nothing at runtime (no gate declared anywhere).

## Tests

New:
- `test/MeshWeaver.Security.Test/NodeTypeGateEvaluatorTests.cs` — 16 pure path-arithmetic tests:
  surface matching, `Self` opening the node but never its subtree, prefix collisions
  (`Overviewer` is not `Overview`), `..` traversal refusal, nearest-gated-ancestor-wins for nested
  plugins, and the no-gate / no-instances short-circuits.
- `test/MeshWeaver.Security.Test/NodeTypeGateTests.cs` — 12 integration tests against the REAL
  `PermissionEvaluator`, on a mesh whose entire gate is **one plugin node plus one entitlement
  grant**: cover + marketing + checkout anonymously readable; gated child closed to anonymous *and*
  to a signed-in non-buyer with no deny row anywhere; the single root grant opening `PaidLesson`
  and `Module/Deep/Lesson`; the redirect resolving from the type with no `_Policy` node; nothing
  outside the gated subtree affected; and the signed-out `AnonymousGate` decision.

Every deny-shaped assertion is preceded by a **positive** wait on the same mesh. A `!HasFlag(Read)`
assertion is satisfied by the first emission of a cold fold (`Permission.None` before anything has
loaded), so on its own it passes vacuously — establishing that the cover IS readable is what makes
the denial evidence rather than a race.

Results (local, Release):
- `MeshWeaver.Security.Test` — **373/373 passed** (the full suite, incl. `AnonymousGateTests`,
  `RedirectOnDeniedTests`, `UserPublicReadTest`, `AdminPartitionAdminTests`, `SecurityServiceTests`)
- `MeshWeaver.slnx` — `-c Release -warnaserror`: **0 warnings, 0 errors**
- `MeshWeaver.Hosting.PostgreSql.Test` (Docker-backed) — **18/18 passed**:
  `PaywallRealGateShapeTests`, `PaywallIdentityMatrixTests`, `GatedNodeExactPathReadTests`,
  `PerSubjectAccessFoldTests`. These exercise `PermissionEvaluator` against the REAL materialised
  gate shape and are the surface most at risk from the changed fold, since this PR adds a fourth
  source to the `CombineLatest` in the hot path.
- `MeshWeaver.Graph.Test` — **48/48 passed** (`AccessAssignmentGuardTest`,
  `UserNodeTypePermissionTest`)
- `MeshWeaver.Documentation.Test` link integrity — **passed** (`AccessControl.md` + the new
  What's New page)

### DIFF
--- a/src/MeshWeaver.Documentation/Data/Architecture/AccessControl.md
+++ b/src/MeshWeaver.Documentation/Data/Architecture/AccessControl.md
@@ -653,6 +653,76 @@ securityService.HasPermission("Anonymous", Permission.Read)
 
 ---
 
+# Type-declared subtree gates (`NodeTypeGate`)
+
+A node type that owns an entitlement-gated subtree — a storefront plugin, a paid course —
+declares its access shape **once, on the type**, instead of materialising it per instance:
+
+```csharp
+builder.ConfigureNodeTypeAccess(access => access.WithGate(new NodeTypeGate("Store/Plugin")
+{
+    PublicSurfaces = [NodeTypeGate.Self, "Overview", "Subscribe"],
+    RedirectOnDenied = "Subscribe",
+}));
+```
+
+Read as: *every* node of type `Store/Plugin` keeps its cover (`Self`), its marketing page and its
+checkout surface readable by everyone — anonymous visitors included — and a reader denied anywhere
+beneath it is sent to `{plugin}/Subscribe`. Nothing else is written. No `_Policy` node, no
+per-child deny, no root grant.
+
+**The rest of the model falls out of what the framework already does:**
+
+| Requirement | Mechanism |
+|---|---|
+| Everything except the declared surfaces is closed | The framework's deny-by-default — no grant, no Read |
+| Purchase / coupon opens the whole subtree | ONE `Viewer` `AccessAssignment` at the plugin root; grants inherit downward |
+| Denied reader is redirected | `NodeTypeGate.RedirectOnDenied`, resolved relative to the gated node |
+
+## Two properties worth relying on
+
+**A gate only ever GRANTS.** It never denies, never caps, and never removes a permission a role or
+an entitlement confers. A declaration that can only widen a short, explicitly listed set of paths
+cannot lock anyone out and cannot regress an existing deployment — which is why an actual `_Policy`
+node still wins over the type-declared redirect, and why the older allow-then-deny gate keeps
+working unchanged next to it.
+
+**`Self` opens the node and nothing beneath it.** That asymmetry is the reason the gate must live
+on the type at all: an `AccessAssignment` at the plugin root inherits strictly downward, so opening
+the cover that way opens the whole subtree — which is exactly why the materialised shape had to
+write a deny for every non-public child to claw it back.
+
+## Why not materialise it per instance
+
+Measured on memex, 2026-07-28 (issue #701), the per-instance shape failed three separate ways:
+
+- **Churn.** The reconcile pass rewrote `_Policy` until its version counter reached six figures
+  (`AgenticEngineering` 254,760), every write by `system-security`, as pure bookkeeping.
+- **Writer/reader drift.** The written policies carried only `redirectOnDenied`, while the reader
+  additionally required `publicRead: false` — the gate read as *not gated* in production.
+- **Silent non-application.** Gating keyed off a `price` field, so two plugins that shipped
+  `price: null` were completely ungated: every page anonymously readable.
+
+A declaration on the type has no version counter to churn, no second condition to drift from, and
+cannot be "not run" for an instance. A plugin that declares no price is gated for the same reason
+every other one is — its type.
+
+## Cost, and the evaluators
+
+`PermissionEvaluator` resolves a target path's **nearest gated ancestor-or-self** from one
+process-wide cached query per gated node type (`$security-gated:{type}`) — bounded by the number of
+gated *nodes*, not their children, and seeded with the static providers so a statically declared
+plugin resolves on the first emission. A mesh that declares no gate subscribes nothing and runs the
+exact fold it ran before.
+
+> ⚠️ **The SQL fold does not yet know about gates.** Postgres RLS decides *query listing*; the
+> evaluator above decides *exact reads and every UI gate*. Until the gate lands in the SQL
+> predicate, a declared public surface is readable by path but will not appear in an anonymous
+> `search`. The asymmetry is strictly in the safe direction — SQL is stricter, never looser, so it
+> cannot become a bypass — but it is a real gap, not a design choice.
+
+---
+
 # Hierarchical access pattern
 
 ```mermaid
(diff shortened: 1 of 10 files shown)


rbuergi:
CI red on `DelegationCompletionTest.SubmitMessage_ResponseCellGetsTerminalText` (Orleans, shard 0) — a completion Notification that never arrived within 20 s.

I checked whether this PR could be the cause, since it adds a fourth `CombineLatest` source to the permission hot path and a delayed fold would delay the notification write. **It cannot be**, and the reason is in the code: `PermissionEvaluator` short-circuits before any of the new machinery when no gate is declared —

```csharp
var gates = CollectGates(hub);
if (gates.Count == 0)
    return fromPolicy.DistinctUntilChanged();
```

The Orleans delegation harness declares no gate, so it takes the byte-identical pre-PR path: no extra query, no extra source, no added latency. Re-running the failed job on that basis rather than merging past it or assuming a flake without checking.

If it reds again on the same test I will treat it as a real signal and investigate the delegation notification path itself — noting that today already produced two other reproducible cross-PR flakes (#945, #946), so a third would fit the pattern rather than contradict it.

Issue #902, opened 2026-08-07
memex: platform Agents catalog is GONE — 'Agent' is a ghost partition root (exists to create, 404s to read, no version history); preInstalled is dead code


rbuergi:
**User-visible outage** (reported by sglauser): no platform agents are available from the portal.

## State on memex (verified 2026-08-07)

| Probe | Result |
|---|---|
| `get @Agent` | **Not found: Agent** |
| `get @Agent/*` | `[]` |
| `search namespace:Agent scope:descendants` | 0 |
| `create` node id=Agent | **"Node already exists: Agent"** |
| `get_versions Agent` | **"No version history found"** |
| after `recycle @Agent` | still Not found |

So the root **exists to the uniqueness check, is unreadable, and has no history** — a content-less ghost partition root, i.e. the shape of #638 (*top-level Space create is not atomic — churn leaves content-less, grant-less ghost roots*). It is NOT ordinary index lag (a direct `get` 404s and survives a hub recycle) and NOT a soft-delete with recoverable history (there are no versions to restore).

Consequence: the platform agent catalog cannot be read — and it cannot be re-created either, because `create` is blocked by the ghost.

## Why nothing restored it automatically — `preInstalled` is dead metadata

`Systemorph/MeshWeaver.Plugins` already ships both catalogs as store packages: `Agent/index.json` declares `nodeType: Store/Plugin` named **"Agents"**, `Skill/index.json` declares **"Skills"**, and *both* set `"preInstalled": true`.

But a grep for `preInstalled` / `PreInstalled` across `src/` (all `*.cs`) returns **no matches**. Nothing in the platform reads that flag, so neither package is ever installed by default.

Confirming the consequence: memex's store lists **38** `Store/Plugin` nodes (including `Essentials`) — **neither Agents nor Skills is registered**.

## And the Skill catalog only works by improvisation

`Skill` exists on memex solely because someone hand-created `Skill/_GitSync` (a `GitHubSyncConfig` → MeshWeaver.Plugins, branch `main`, subdirectory `Skill`) plus `Skill/_Policy` = `PartitionAccessPolicy { publicRead: true }`. `Agent` never got that hand-treatment, so when its root was lost there was no mechanism to bring it back.

Two sibling catalogs, two different provisioning stories — one improvised, one non-existent.

## Required fix (maintainer: "see it installs properly. don't hack")

[...]

PR #954, opened 2026-08-08
The Store self-heal could not write its own progress into the mesh it was healing

### LINKED ISSUE #919: Store self-heal cannot run on the mesh it needs to heal — the repair's activity write is RLS-blocked by the missing grants it exists to create
Found while restoring the Store on memex.systemorph.com (2026-08-07). The mesh **could not self-heal, by construction** — and no upstream merge could have fixed it, because the fix cannot execute on an affected mesh.

## The deadlock

The Store self-heal (GitSync `update` → recompile `Store/Catalog` → `CatalogGate` seeds the public grants) records its progress in an **activity node inside the partition it is repairing**: `Store/_Activity/import-{sourceId}`. When the partition has **no grants**, RLS rejects that create, so the sync fails before it can import anything:

```
Update aborted: no initial state arrived for 'Store/_Activity/import-378cec0cd08c9c76' within 30s.
Likely causes — (1) RLS silently rejected the prior CreateNode …
```

The repair requires the grants it exists to write. MeshWeaver.Plugins #200/#204 and PR #212 fixed the *seeding* for a fresh Store; none of them can reach a Store already locked out.

## Two aggravating factors observed on the same incident

1. **The import activity id is deterministic per source** (`import-{sourceId}`), so once one attempt leaves a poisoned row behind (here: v166, 16 KB of content, unloadable by its own hub — the #590/#902 activation-family shape, sighted live in prod), **every subsequent attempt re-targets the same broken node** and fails identically. There is no way to get a clean retry without deleting the row out-of-band.
2. Recovery required the full out-of-band sequence: hand-write `{partition}/_Policy` (breaks the RLS circle) → delete the poisoned activity row via SQL + portal restart (the AGENTS.md-sanctioned pairing) → `force` update (two-way preservation otherwise keeps every server-modified node and imports 0). Three separate manual interventions for what should be one self-heal.

## Suggested fixes (independent, any subset helps)

1. **The sync's own bookkeeping must not be subject to the partition's access state it is repairing** — write the import activity under the System scope that already executes the sync (it runs as `system-security`), or exempt `_Activity` satellite creation by the sync identity from RLS the same way `_Access` writes by the gate are possible.
2. **Never reuse a deterministic activity id across attempts once the prior node is in a faulted/unloadable state** — mint a fresh id (the ordinary activity pattern) or repair-in-place on upsert (the #909 ghost-repair now does this for partition roots; extend to satellites).
3. A partition-level self-heal entry point that does not depend on the partition being readable — e.g. the preInstalled reconciler pattern from #909, which runs from outside the partition and repaired the Agent ghost on both portals today.

## Evidence

Portal logs (memex ns, 2026-08-07 16:20–21:15 UTC), activity nodes `Store/_Activity/{088c4342,b123af6b,f6685947,e666aa36}`, and the row inspection `store.activities` path `Store/_Activity/import-378cec0cd08c9c76` (v166, contentlen 16819) prior to deletion.

rbuergi:
Fixes #919.

memex's Store lost its grants, and the sync sent to restore them died before it imported anything:

```
Update aborted: no initial state arrived for
'Store/_Activity/import-378cec0cd08c9c76' within 30s
```

The repair needed the grants it existed to write. Getting the Store back took three manual interventions — a hand-written `_Policy` to break the RLS circle, a SQL row delete plus a portal restart, and a `force` update. Two independent defects turned a bad minute into a permanent deadlock. Both are fixed here, and both have a repro that **fails on the parent commit** (verified by reverting `src/` and re-running: see *Evidence* below).

## 1. The bookkeeping escaped System where no call-site review could find it

Every import-activity write already runs inside `using (accessService.ImpersonateAsSystem())`. That looked airtight — and wasn't, because **a cross-hub MeshNode write is not performed on the enqueuing thread**.

`IMeshNodeStreamCache` serialises writes per path through a `Concat`. Only the **first** request is subscribed synchronously inside `Subject.OnNext` (still inside the caller's `using`). Every later one is subscribed when its **predecessor settles** — on the owner's response thread, where that scope is long disposed. Both halves of the write then read whatever ambient identity happens to be on that thread: `UpdateRemote` captures it for the outbound `PatchDataRequest` **and** opens the initial-state `SubscribeRequest` under it.

On a grant-less partition the **read** direction is what kills you: the initial state never arrives and the write dies with the 30 s abort above. The repro pins the exact shape — the import reports `"Imported"` while its activity is stuck at `Running` forever, because the terminal write was refused and swallowed.

**Fix:** the identity rides the request. `UpdateRequest` carries the enqueuer's `AccessContext`, captured synchronously at enqueue where the scope is still live, and `BuildUpdateQueueObservable` re-establishes it around the dispatch. A caller that genuinely had no identity is unchanged — a null capture is **not** re-stamped, so context-less writes keep failing closed as they must.

## 2. A deterministic id was doing two jobs, and only one of them wanted a stable id

`import-{fingerprint}` was both the lock and the progress log. So one poisoned row at that id (v166, 16 KB, unloadable by its own hub) made **every** retry re-target the same broken node — each progress line burning the full 30 s abort — with no clean retry short of deleting the row by hand in SQL.

The roles are now split:

- **The lock** keeps `import-{fingerprint}`. That id **is** the create-wins serialization between replicas and the durable "already imported" marker the boot short-circuit reads, so it cannot be minted fresh — two replicas holding different lock ids both import. It is now written **only** through the idempotent upsert, which floors the version on the durable row it just read (#902/#909), so a forked/ghost lock row heals in place instead of being silently discarded.
- **The attempt** is minted fresh per run — `import-{fingerprint}-{timestamp}-{rand}` — and takes every progress line, per-file `⚠`/`🗑` diagnostic and terminal summary. Nothing ever writes to a prior attempt's node again, so a faulted attempt cannot poison the next one. It also gives operators the import history they actually want: one node per run rather than one node overwritten in place.

## Evidence

Two repros in `StaticRepoImporterTests` (Postgres, real RLS):

- `Reimport_AsPlainUserOnGrantlessPartition_WritesItsOwnBookkeeping` — a GitSync-shaped `ImportSource` into a grant-less partition with an ordinary user as the ambient identity. Asserts, straight from storage (RLS-free, so the caller's own grants can't fool it), that the lock **and** the attempt reached terminal `Succeeded`, and the content healed.
- `Reimport_MintsAFreshAttemptId_AndHealsAPoisonedDeterministicLock` — plants a content-less row at v166 straight into storage at the deterministic lock id (same shape as `GhostPartitionRootRecoveryTest`), then imports. Asserts the lock heals in place with its version moving forward, that the run logged to its own node, and that a second run mints a different attempt id.

Control run with `src/` reverted to the parent commit, tests unchanged:

```
Reimport_MintsAFreshAttemptId_AndHealsAPoisonedDeterministicLock [FAIL]
  Expected collection not to be empty because every run opens its own attempt node.
Reimport_AsPlainUserOnGrantlessPartition_WritesItsOwnBookkeeping [FAIL]
  Expected value to be Succeeded, but found Running.
Total: 2, Failed: 2
```

With the fix: `StaticRepoImporterTests` 11/11 green. Regression net around the touched seams, all green: `MeshWeaver.Data.Test` 303/303 · `MeshWeaver.Hosting.Monolith.Test` cache + circuit-context classes 39/39 (1 pre-existing skip) · `MeshWeaver.Graph.Test` importer classes 13/13 · `MeshWeaver.Documentation.Test` 19/19.

## Notes

- No What's New entry — this is ops-internal: it changes how a sync records its own progress, with no user-visible surface.
- `Doc/Architecture/StaticRepoImport.md` is updated for the lock/attempt split and the queued-write identity carry.
- Trade-off worth naming: an import that actually runs now writes **two** bookkeeping nodes instead of one (lock + attempt). Boot re-imports at an unchanged fingerprint still short-circuit and write nothing, so the growth is per-run-that-imports, not per-boot.


rbuergi:
🚫 **Holding this one — the review finding is correct, and it removes the mutual exclusion the lock exists to provide.** Not merging until it is resolved.

The class doc still states the invariant (`StaticRepoImporter.cs:94-95`):

> lock — `IMeshService.CreateNode` makes the first caller win and concurrent replicas get "already exists"

but the new comment at `:551-554` says the lock node is now written **only** through `Upsert` (`CreateOrUpdateNodeRequest`), chosen so a forked/ghost lock row heals in place. `Upsert` treats "already exists" as success and continues. So two replicas starting close together both miss the Succeeded short-circuit, both stamp `Running`, and **both import** — the create-wins serialization is gone, replaced by a doc comment asserting it still holds.

Both properties are wanted and they conflict in the obvious implementation:
- **create-wins** needs a failing create to be the mutex;
- **ghost-heal** needs an upsert that can rewrite a poisoned row.

A shape that keeps both: attempt `CreateNode` for the ACQUISITION (a failure means someone else owns the run → defer, which is today's behaviour), and reach for `Upsert` only on the repair path — i.e. when the create failed AND the existing row is provably unusable — plus for the terminal stamp, where the owner is already established. That keeps the mutex on the create and confines the upsert to the case it was added for.

I am deliberately not implementing that here: it is a concurrency invariant on the boot-import path, the ghost-detection half is the piece the report already flags as untested ("I could not construct a deterministic repro for that exact activation failure"), and guessing at it late is how the original wedge got written. Worth a focused pass.

Everything else in this PR is strong — the root cause is excellent (the `ImpersonateAsSystem` scope lost in `MeshNodeStreamCache`'s serial write queue, where only the first request subscribes synchronously and later ones subscribe on the owner's response thread after the caller's `using` disposed), and it is backed by a real control run. The attempt/lock split is the right idea; only the lock's acquisition primitive needs to go back to create-wins.

rbuergi:
✅ **Unblocking — resolved in `08a53c1`.**

Both review findings are correct, and they are the same defect I held this on: the deterministic node was documented as a create-wins lock while `Upsert` treats "already exists" as success.

I took the **documentation** side of the reviewer's "or" rather than reintroducing a real lease, because obeying a `Running` marker is exactly the behaviour that was deliberately removed — it wedged atioz into `AlreadyRunning, 0 nodes` forever, and the guard now re-imports on any status that is not `Succeeded` precisely so a crashed or raced marker gets reclaimed. Adding a lease back would reopen that wedge.

So the node is now described for what it is: a **content-addressed short-circuit marker**, not a lock. Two replicas on one fingerprint can import concurrently, and that is safe only because every write on the path is an idempotent upsert of identical content.

Corrected: the class doc, the `THE LOCK` block, the `Reimport` step comment, the sentinel-miss comment, and three places in `StaticRepoImport.md`.

One of them mattered beyond wording — a comment was using the imagined lock to justify accepting an eventually-consistent read (*"no concurrent import races this read"*), which is the sort of borrowed guarantee a later change quietly leans on. It now states the real reason (idempotency), and the doc carries an explicit "do not add a non-idempotent step here, and do not treat 'the lock protects me' as an available argument" note.

PR #983, opened 2026-08-09
A client, a customer and the cluster's real names were in the public repo

rbuergi:
The rule is **generic in MeshWeaver, specific in Systemorph/Memex**. The public repo was carrying a lot of the specific.

| What | Where | Now |
|---|---|---|
| **Customer name** `atioz` | **173 tracked files** | SDK examples → the public `memex.meshweaver.cloud`; the seeded "Client" entry in the React Native instance list **deleted**; ~150 provenance comments say `prod` |
| **Client name** `PartnerRe` | 24 files, as the sample node path `PartnerRe/AIConsulting/FinalReport` | `Acme`, matching the neutral sample orgs already in the repo |
| **Instance inventory** | `Instances.md` — every portal with owner, purpose, database, who may log in, and a client engagement | Explains what an instance *is*; points at the private repo for which ones exist |
| **Infrastructure** | cluster, resource group, Postgres server, VPN gateway, portal identity, database names, a private VNet address (~60 files) | Placeholders |

## Two deliberate decisions

**`memex.meshweaver.cloud` stays.** It is the public instance the README already advertises as "see it live", so it is the honest example URL for client SDKs and docs. Removing it would cost the demo and protect nothing.

**The mechanism half of `Instances.md` stays.** Reading a running version, how self-update works, instance lifecycle — that is generic and applies to any installation. Only the inventory left.

## Not touched, and why

`rbuergi@systemorph.com` appears in **39 files** — but it is the documented DevLogin test identity (AGENTS.md, `MonolithMeshTestBase`, `TestUsers`), not a leak. `rsalzmann` (14 files) and `Markus` (3) are colleagues' names in test fixtures. Changing the test identity is a decision about the test contract, so it is raised rather than done.

## Verification

- Full solution builds clean with CI's flags — **0 errors, 0 warnings**
- `MeshWeaver.AI.Test`: **1090 passed** (5 files used the client name as a fixture path)
- `DocumentationLinkIntegrityTest` passes
- Zero occurrences remain of any swept identifier

One trap worth knowing: angle-bracket placeholders (`<pg-server>`) are XML tags inside `///` doc comments and broke the build in three places. Code now uses prose placeholders; only markdown and shell keep the bracket form.

https://claude.ai/code/session_01AEZMJc1GJRjnRsxnvCWRmK


rbuergi:
Review pass — all six threads addressed in 036c5101b. Three of these were real breakage, not style:

**`memex-aca.bicep` / `mainTemplate.json` (`<pg-server>` in resource names)** — agreed, and worse than cosmetic: `<` and `>` are invalid in an Azure Storage share name, so the deployment would have failed at validation. Restored the original `memexpostgres` rather than substituting a new neutral token: that string is not the Postgres Flexible Server hostname this PR set out to scrub — it is the Aspire container resource name (`memex-postgres`) flattened into a generated Azure resource name — and keeping it preserves the **existing** share, where a rename would have silently orphaned the Postgres data volume behind a new one. Fixed in both files (all 5 ARM occurrences + both bicep ones).

**`InstancesOptions` defaults (both threads)** — agreed. These are substituted verbatim into copy/pasteable `az` commands, so `-g the AKS resource group` splits into three arguments and `az` fails pointing at the wrong flag. Now `<aks-cluster>` / `<aks-resource-group>` / `<pg-server>`, matching the placeholder convention used elsewhere in this PR, with a comment recording *why* prose cannot be the default here.

**`MeshQueryMergeContractTest` / `PartitionSegmentValidationTest`** — agreed, both are rename artifacts. `(prod 2026-07-03)` was `(atioz 2026-07-03)`, now just `(2026-07-03)`; `the memex and the cloud database databases` was `the memex and memexcloud databases`, now `the memex and memex-cloud databases`.

**`UserDashboardThreadQueryTests` display name** — correct, and it was the tip of a real bug rather than a consistency nit. `PartnerRe` had been renamed to `Acme` in a test that already seeded a separate `ACME` root; a top-level segment **is** a partition and its schema is `segment.ToLowerInvariant()`, so the two were the same node and the second seed threw `Node already exists: ACME` — this was the shard-1 CI failure. Renamed to `Contoso`, which also clears the `Name` and thread-title strings you flagged.

Verified: `Memex.Portal.Shared.Test` 250/250 green, Release `-warnaserror` clean.

Issue #956, opened 2026-08-08
Feature request: a versioning indicator so users can verify a deployment is up to date


sierragolflima:
## Caveat on this report

This was investigated externally, without direct access to the AKS cluster or the
Azure subscription it runs on — via the GitHub API, the public HTTP surface of
`memex.meshweaver.cloud`, and browser checks in Settings. Everything below is what
was observable from outside; there may be a straightforward explanation we simply
couldn't see from this vantage point (e.g. a config difference on this specific
deployment, a rollout in progress, or something particular to how this environment is
set up). Filing this as a feature request with the observations attached, rather than
as a confirmed bug report, for that reason.

## The need

It would help both users and operators to have a reliable way to tell **how current a
live deployment is** — whether a recently-merged fix on `main` is expected to be live
yet, or still in flight. As far as we could tell, there is currently no way to answer
"is this deployment up to date?" without someone with direct cluster access checking
the running pod's image tag by hand on the production subscription on Azure.

## What may already exist (this could just be a visibility gap on our end)

An "About" settings tab (global, ungated) appears to already exist in the codebase,
added in `1bd78e931` (2026-07-08, *"About page, What's New feed, top-bar theme toggle
& build-commit stamping"*): it's described as showing the platform version and the
git commit the build came from, linking to it on GitHub. A "What's New" feed appears
to have landed in the same commit. Pull request #897 (2026-08-07) appears to extend
the About tab with an installed-plugin inventory.

**If this is working as intended, this report is really about reachability/reporting
accuracy, not about needing new functionality.** We're genuinely unsure which of the
possibilities below applies — flagging both rather than assuming.

## What we observed, for context

On `memex.meshweaver.cloud`, we could not find an About tab or a What's New tab in
global Settings, after a hard refresh and again with a completely fresh
session/cookie (so far as we could tell, this ruled out simple client-side caching
and being pinned to one specific backend pod via session affinity — though we can't
rule out possibilities we didn't think to test).

This seemed worth mentioning alongside a possibly-related data point: `Systemorph/
Memex`'s GitHub Deployments record (which that repo documents as the registry of
record for this deployment) shows one status entry:

> `memex-cloud` — success — "Running 3.0.0-ci.1842 (rolled by in-pod self-updater)" —
> posted 2026-08-06T06:40:38Z

`ci.1842` appears to correspond to `MeshWeaver`'s CD workflow run #1842, created
[...]

Issue #981, opened 2026-08-09
Callbacks still pending past the teardown quiescing budget (FutuRe EuropeRe, Cession, PandasExplorer) — a CreateNodeRequest the mesh hub never answers


rbuergi:
## Summary

Split out of **#826**. Two tests fail intermittently because a callback is still pending when teardown's quiescing budget expires — work that should have been joined, wasn't.

| Test | Detail |
|---|---|
| `FutuReAnalysisTest.EuropeRe` | leaked `CreateNodeRequest` callback pending **2,738 ms past the 2 s Quiescing budget**; latest run `31244442523`, 2026-08-08 |
| `CessionLayoutAreaTest` | currently dormant (clean across 385 attempts) but its signature is **identical** to EuropeRe's, so it is not an independent data point — expect it to move with the same fix |

Also seen in the same family and worth fixing together or immediately after:

- **`MeshHubDisposalLeakTest.MeshHub_IsCollected_AfterMeshAndServiceProviderDisposal`** — **5 occurrences on 2026-08-08 alone, including on `main`**. It was *measured* to be unaffected by the #613/#967 teardown-ordering fix (identical before and after, 12 forced GCs either way), so it needs its own diagnosis. A concrete lead was recorded during that work: **the test hand-rolls its teardown** — `hub.Dispose()` → `await DisposalCompleted` → `sp.Dispose()`, setting `_selfDisposed = true` — so `base.DisposeAsync()`, which is what runs `ioPools.DrainAll()` and fires the terminal teardown signal, **never executes**. Pooled leaves can therefore still be live when it takes its GC snapshot.
- `SourceDocumentDataLoadingTest` was seen failing with the same "left Observe subscriptions pending past the Quiescing budget" message on shard 4.

Evidence base: a sweep of **600 runs / 640 run-attempts** (2026-08-04 → 2026-08-09) reading every attempt, with all 247 failing shard logs verified non-empty.

## Why this is a real defect, not test noise

#826's premise — that these were environment wobble — was **disproven**: concurrent-run density around failing attempts is identical to the whole population, and every member of that cluster which ever recurred turned out to be a genuine defect (#834, #968, #975, #955, #932, #977, #618).

[...]

PR #985, opened 2026-08-09
Tell every user whether their build is current, not just admins

rbuergi:
`Refs #956` — **do not close that issue on merge.**

#956 names two items. This PR addresses the first only:

- ✅ **Running-vs-latest is admin-gated** — surfaced on the ungated About tab, below.
- ⬜ **No unauthenticated version surface** — `/health` and `/alive` carry no version and there is no `/api/version`, so nothing is verifiable without signing in. **Still open, deliberately.** It is an information-disclosure decision for the maintainer, not a presentation change, and it is left untouched here rather than folded in silently. See "Not in scope" at the bottom.

Most of what the issue asks for was already shipped. The About tab is ungated and already names the exact build — version, git commit (linked to the commit on GitHub), repository, runtime, plugin inventory. What was missing is the half that makes the version *mean* something: whether it is the newest one. That comparison existed, was already being made by the self-update poller a few times a day, and was visible only on the platform-admin-gated Updates tab.

This adds the verdict to the ungated About page. No new machinery, no new poller, no new state.

## The access question — verified, not assumed

The issue flagged one open question: *can a non-admin actually read `Admin/UpdatePolicy`?* **No.** Pinned against a real mesh in `NonAdminUpdateStatusTest`:

- an identity with no grants resolves to `Permission.None` on `Admin/UpdatePolicy`, and the per-user read gate in `MeshNodeStreamCache` refuses the read with `UnauthorizedAccessException` — it does not hang, which is what lets the About page degrade instead of stalling;
- a platform admin reads the same node fine (positive control, so the denial is a denial and not an empty mesh).

So the grant was **not** widened. Instead `PlatformUpdateStatus.Observe` reads the node as System and lets exactly two things out — a verdict, and the newer version only when one exists. This is the same shape the anonymous `/privacy` page already uses to read `Admin/Privacy`.

Deliberately **not** projected: the update strategy, the CI-verified-builds flag, and the poller's `CheckedAt` bookkeeping. Those are decisions about the deployment rather than facts about it, and they stay on the admin tab.

## What a user sees

| State | Line |
|---|---|
| Nothing newer detected | ✅ **Update status:** Up to date — no newer build has been detected |
| Newer build recorded | ⬆️ **Update status:** Update available — `3.0.0-ci.2400` |
| Unknown | *(no line at all)* |

Language-neutral glyph plus one translated phrase. Three new keys in **both** `strings.en.json` and `strings.de.json`, read via `host.Localize` — `LocalizationTest` green.

**Unknown renders nothing, on purpose.** An install that never wired self-update, or one with automatic checks switched off, has no evidence either way; a reassuring "up to date" on the strength of a check that never ran would be worse than silence. Absence is detected with the storm-safe existence query, never a point-read of a missing path.

## Two things worth a reviewer's eye

**The impersonation scope is `Observable.Create` + a synchronous `using`, not `Observable.Using`.** Impersonation is an `AsyncLocal`, and `Using` disposes its resource on whichever thread the sequence *terminates* on — so the restore never reaches the thread that opened it, leaving the caller (here, a user-facing layout render) running as System afterwards. This is not theoretical: it is exactly what made the first draft of the access test read as `system-security` and assert nothing. Both mesh tests now carry a non-vacuity guard on the identity actually in force.

**`PrivacyStatementNode.GetStatement` uses the `Observable.Using` shape** this PR avoids. I have not verified whether it actually leaks at its Blazor call site (different threading), so I have deliberately left it alone rather than change a shared helper on a hunch — flagging it as worth a look.

## Not in scope

The issue's second item — nothing is reachable outside an authenticated session (`/health` and `/alive` carry no version; there is no `/api/version`) — is untouched, and **#956 should stay open for it after this merges**. An unauthenticated version endpoint is an information-disclosure decision (who may learn which build a deployment runs, and what that tells an attacker about unpatched CVEs), not a presentation one. That is a maintainer call, and quietly bundling it into a UI change would be the wrong way to make it. Happy to take it as its own issue if wanted.

## Verification

- `dotnet build -c Release -warnaserror` on `Memex.Portal.Shared` + its test project: **0 warnings, 0 errors**
- `test/Memex.Portal.Shared.Test`: **250/250 green** (14 of them new/touched)
- `LocalizationTest`: **33/33 green**
- What's New entry included (user-facing change)


rbuergi:
## CI is red on an unrelated pre-existing crash — not this diff

`MeshWeaver Build and Test` run [31299547995](https://github.com/Systemorph/MeshWeaver/actions/runs/31299547995), shard 2:

```
[createdump] Crashing thread 0bfc signal 11 (000b)
Segmentation fault  ( cd "$dir" && timeout … dotnet MeshWeaver.FutuRe.Test.dll … )
[CI] MeshWeaver.FutuRe.Test exit=139
```

**No test failed.** The trx gate found nothing; the shard failed on the exit-marker gate alone — the host segfaulted mid-run, which is the #613 signature exactly.

### Why this is not the change in this PR

- **No reachability.** `MeshWeaver.FutuRe.Test` does not reference `Memex.Portal.Shared`, the only project whose C# this PR changes. The one shared project it does reference is `MeshWeaver.Messaging.Hub`, where the diff is three added keys and one reworded string in the embedded localization JSON — that cannot produce a native SIGSEGV during teardown.
- **The branch is not stale.** It contains all of `origin/main` (`906fc8827`); `git log HEAD..origin/main` is empty.
- **The same tree passed on main 45 minutes earlier.** Main run [31298376804](https://github.com/Systemorph/MeshWeaver/actions/runs/31298376804) (06:12Z) ran `Run tests (shard 2)` for real — not a reused green tree — and it succeeded. This run crashed at 06:58Z.

### The part worth someone's attention: #613 has recurred

#613 was closed **2026-08-09 05:48Z**, ~70 minutes before this run. Both of its fixes are ancestors of this branch:

| Commit | Date | |
|---|---|---|
| `33c44e8e7` | 2026-08-08 20:55Z | Unload node ALCs at the END of teardown, not before the I/O drain (PR #967) |
| `74bc3a07c` | 2026-08-06 09:24Z | The FutuRe exit=139 is a background-GC fault, not an ALC use-after-unload |

So the crash survived both. #981 tracks the *adjacent* teardown family (callbacks pending past the quiescing budget) but that is a test failure, not a signal death — it does not cover this. **I would suggest reopening #613**; I have left the same evidence there rather than reopening someone else's issue unilaterally.

I have deliberately **not** re-run CI. Re-running to see whether it goes green is what hides an intermittent crash, and this one is now known to survive its own fix.

### What was verified locally

- `dotnet build -c Release -warnaserror` on `Memex.Portal.Shared` + its test project — 0 warnings, 0 errors
- `test/Memex.Portal.Shared.Test` — 250/250 green
- `LocalizationTest` — 33/33 green

PR #988, opened 2026-08-09
A fresh replica's first write to a path was guarded by nothing — the invariant moves into the store

### LINKED ISSUE #971: Monotonic write guard is per-process, and the PG upsert has no version condition — a fresh replica's first write can roll a node back
## Summary

`MonotonicWriteGuardStorageAdapter` enforces "the store never moves backward" using a **per-process** in-memory high-water mark. PR #968 fixes a real window in it (the mark was claimed *after* the row was mutated, so a writer on the `IStorageAdapter.Changes` feed could overwrite a newer row with a stale snapshot — acked writes dying silently).

That fix is correct and worth landing on its own, but it leaves a **structural gap that #968 deliberately does not close**, and the gap matters in production rather than in tests.

## The gap

Two parts, and they compound:

1. **The mark is per-process.** A second replica starts with an empty mark table, so its first write for a given path has nothing to compare against.
2. **`PostgreSqlStorageAdapter`'s upsert carries no version condition.** Nothing at the store level refuses a write whose version is below the durable row's.

So on a fresh replica's **first write to a path**, monotonicity is enforced by nothing at all — not by the in-process guard (no mark yet) and not by the database (no condition on the UPDATE). Any stale snapshot that replica happens to hold will be written straight over newer durable state.

This is invisible in single-process tests, which is why it has not shown up: the guard looks sound because in one process the mark is always populated by the time it matters.

## Why it deserves a real fix rather than a wider in-memory guard

The invariant is a property of **the row**, not of a process's memory. Any in-process structure re-derives it and will be wrong again the next time the topology changes (a rollout briefly running two pods, a KEDA scale-out, a restarted replica). The durable fix is a **store-level compare-and-set**: make the write conditional on the version it believes it is replacing, and refuse (or re-read and retry the merge) when the condition fails.

Note the same reasoning applies to `SnowflakePartitionStorageProvider`, which has not been checked.

## Impact

Silent data loss, and the shape is nasty: the stale write is **acked**, so nothing surfaces. The next write is minted one above the rolled-back row and is accepted for a perfectly good reason, so the lineage continues from the wrong place. What an operator sees is a node that quietly reverted — the `Version=12 → Version=2` shape the guard exists to prevent — with no error anywhere.

## Provenance

Split out of #826 / PR #968, which root-caused and fixed the in-process window. Recorded there as "residual, not fixed — needs a store-level compare-and-set; deserves its own issue."

rbuergi:
Fixes #971.

## Verdict up front

**On a fresh replica's first write to a path, "the store never moves backward" was enforced by nothing at all.** Not by the in-process guard — its high-water table is empty on a new process, so the write took the unverified fast path. Not by the database — the Postgres upsert carried no version condition, and the in-memory store was a plain `_nodes[path] = node`. #968 closed the *in-process* window; this closes the *cross-process* one it explicitly recorded as residual.

The failure is silent by construction: the stale write is **acked**, and the next write is minted one above the rolled-back row and accepted for a perfectly good reason — so the wrong lineage persists with no error anywhere.

### 1. The invariant moves into the store

`MeshNode.Version` is the node's forward-only revision counter, so a correctly-produced write is always at or above what is stored. Every backend that can express that now refuses a strictly-regressing write **at the row** and reports it by emitting the STORED node instead of the one it was handed:

| backend | condition |
|---|---|
| Postgres | `INSERT … AS target … ON CONFLICT (namespace, id) DO UPDATE SET … WHERE target.version <= EXCLUDED.version` — row count 0 ⇒ refused |
| in-memory | one `AddOrUpdate` keeping the higher `Version` — atomic against every concurrent writer, no read-then-write window |
| Snowflake | `WHEN MATCHED AND t."version" <= s."version" THEN UPDATE SET …`; the no-MERGE emulator fallback gets a conditional `DELETE … AND "version" <= :version` plus a `WHERE NOT EXISTS` guard on the `INSERT` |

Equal versions still apply — re-persisting an unchanged node is a legitimate, common shape. A refused write publishes **no** change notification (nothing changed, and a notification carrying the loser would hand every subscriber the state the store just rejected), and on Snowflake none of the trigger-replacement steps run either, since they all derive from a row that was not written. `WriteMany` reads per-command `NpgsqlBatchCommand.Rows`, so a refused node in a batch is neither announced nor reported as committed, and the caller gets durable truth back in its original order.

The contract is written into `IStorageAdapter.Write`'s doc, including the explicit licence for backends that cannot express it to behave as before.

### 2. The conflict resolves by MERGING, never by refusing

Per the maintainer's policy on the issue. Callers of `stream.Update` do not write conflict-retry loops and must not have to, so `MonotonicWriteGuardStorageAdapter` now funnels **both** conflict detectors — its own verification read, and a store that refused the write — into one resolution: re-read, merge the loser into durable truth, write once. Deliberately **one attempt, never a retry loop**: if the row moved again during the merge, the second refusal is logged and the newer row is kept.

The merged node carries the **latest** row's version, specifically so re-persisting it cannot itself read as a rollback. That also preserves the `saved.Version > written.Version` signal `RefusedWriteOwnerRebaseTest` pins, so the owner-rebase path is unchanged.

## How each policy rule maps onto the shared merge

You asked me not to write a second merge. I did not — but one rule genuinely could not be expressed through the existing entry point, and I want to be plain about which.

**`MeshNodePatchMerge.Apply` (three-way) cannot serve this at all.** Its whole classification turns on `live == base`, and handed `baseValues: null` it takes the "no conflict signal" branch and applies the patch **wholesale** — i.e. the stale snapshot overwrites the newer row, which *is* the defect. There is no argument shape that makes it behave two-way. So the addition is a second **entry point on the same class** (`ApplyTwoWay`), sharing the recursion, `StringDelta`, and `MergeArray`, rather than a second merge implementation that could drift. `MeshNodeConflictMerge` on top of it owns only the MeshNode member mapping (which members are user data, which are the store's own bookkeeping, and how `Content` is projected to JSON and materialised back into its CLR type — the node as a whole is never serialized, because `HubConfiguration`/`GlobalServiceConfigurations` are delegates).

| rule | how it lands | shared with the three-way merge? |
|---|---|---|
| **non-strings → latest wins** | every non-string, non-object, non-array leaf keeps `live` and reports the member | **identical** — `Apply` already refuses conflicting scalars and keeps live |
| **strings → try to merge** | `StringDelta.Compute(latest, stale)` reduces the pair to one splice; a pure **insertion** means stale ⊃ latest (take the superset), a pure **removal** means latest ⊃ stale (keep latest); a splice that both removes and inserts is a genuine divergence ⇒ not auto-resolvable | **new rule, shared primitive.** `Apply`'s string merge rebases the writer's splice over the intervening edit — that needs a base. With none, only the superset case is decidable. |
| **arrays → element identity, not RFC 7396 replace** | calls the **same `MergeArray`** with an empty base, which is exactly "neither side can be shown to have removed anything, so keep every element" — union by whole-element value equality | **same code path** |
| **unresolvable → latest wins + a warning in the activity log** | `onOverwritten` → `OverwrittenMembers` → a `LogWarning` **and** a durable `ActivityLog` MeshNode | **new** — `Apply`'s `onRefuse` has no durable sink today |

**Which MeshNode members take the string merge:** only `Name` and `Description` — the prose. `NodeType`, `Category`, `Icon` and `DesiredId` are wire identifiers, not text; the superset rule would happily splice `"Code"` and `"CodeCell"` into one of them and silently **retype** the node, which drives routing and per-node hub activation. `PreRenderedHtml` is derived from `Content`, so it follows the merged content rather than being spliced into broken markup. All five take latest-wins-**and-report**, so the policy's fallback still covers them and nothing is dropped silently. The user data that actually benefits from merging lives in `Content`, which is merged leaf by leaf.

Two ambiguities the base-less merge cannot resolve, decided deliberately and documented at the call site rather than hidden:

- **A superset string, and a union array, can resurrect content the newer writer deleted.** Without a common ancestor, "the stale writer inserted this" and "the newer writer deleted it" are literally the same observation. The alternative silently destroys content an acked write added, which is the failure mode of record — so the policy resolves toward keeping content. Resurrection is visible in the content and reported; destruction is not.
- **A key absent from the latest object is never adopted from the stale one.** The serializer suppresses defaults, so absence is ambiguous between "removed" and "is the default" — and under both readings the newer state is the one to keep. Adopting there would let an older snapshot re-assert a field the newer write had just defaulted.

## The warning is the load-bearing half

Latest-wins is acceptable; latest-wins **invisibly** is the #968 defect — an acked write that rolled a row back with no error anywhere. So every member the merge could not resolve produces both:

- a structured `LogWarning` naming the path, both versions, and the merged / dropped member lists; and
- a durable `ActivityLog` MeshNode at `{path}/_Activity/write-conflict-{latestVersion}`, category `WriteConflict`, `ActivityStatus.Warning`, with one `LogLevel.Warning` message **per dropped member, naming it**.

The id is keyed on the **durable** version alone, never the loser's. That is the right granularity ("this revision had a conflict") and it is what bounds the litter: a wedged owner mints an *increasing* version on every retry — the #725/#872 fork shape, 834, 835, 836 … against a stuck 2423 — so keying on the losing version would accumulate one satellite per attempt, forever. Re-recording overwrites the same node at `Version = 1`, which the store's equal-version rule accepts, so the record can never conflict with itself. It is best-effort by contract (a failed trace must not fail the resolution it describes) and skipped for paths already inside an `_Activity` namespace, so a conflict on a trace can never write a trace about a trace.

## Snowflake

**Fixed, not deferred — but unverified against a live endpoint.** `SnowflakePartitionStorageProvider` has **no write path of its own** (grep for `INSERT`/`MERGE`/`UPDATE`/`Write(` in it returns nothing): it exposes `SnowflakePathRoutingAdapter`, whose `Write` delegates straight to the per-schema `SnowflakeStorageAdapter.Write`. That is the path now made version-conditional, so the provider's gap closes with it.

Its *other* per-process dictionaries (`_registeredPartitions`, `_schemasInitialized`, `_provisioned`) are **not** the same defect and are deliberately untouched: they are idempotent provisioning promise-caches, not an authority on durable state. A second replica starting with an empty `_provisioned` just re-runs an idempotent `CREATE SCHEMA IF NOT EXISTS`; nothing is lost.

I could not execute this against Snowflake or its emulator — that backend has no CI coverage here. The change is compile-clean and shape-identical to the Postgres one, and that is the whole of my evidence for it.

## Not covered, and why

FileSystem, Caching, Sqlite, Cosmos and AzureBlob adapters keep their previous behaviour, which the interface doc now explicitly sanctions. FileSystem/Caching/Sqlite are single-process dev and test hosts — there is no second replica, and a file-level read-then-write "CAS" would be racy anyway. Cosmos has ETag optimistic concurrency but no `version <=` predicate: it would need a read plus `IfMatch` on every write, i.e. an extra round-trip on the hot path, against a backend that is emulator-unverified with a known contract gap. In every one of these the in-process guard remains the protection it always was, which is sound precisely because they are not multi-replica.

## Verification

**Fail-without, run against current `main`** (the two behaviour-bearing files reverted to `origin/main`, rebuilt, re-run) — both new tests fail, and the output names the defect exactly:

- `AWriterWithNoHighWaterMark_CannotRollTheDurableRowBack` → `[replica write] emitted=v1/stale-snapshot`, `Expected 1 to be greater than or equal to 7000`. On main the fresh replica's write **landed**: the v1 snapshot went straight over the v7000 row. That is the issue's silent rollback, reproduced.
- `AConflictIsMerged_NotRefused_AndTheDroppedValueReachesTheActivityLog` → `[durable] v9000 name=durable-advance description=alpha beta`, `Expected "alpha beta gamma" … but found "alpha beta"`. On main the conflict was **refused**, so the losing write's text was destroyed rather than merged, and no activity-log record existed.

**Pass-with:** 2/2. The property is **constructed, not raced** — "a second writer with no prior mark" is a fresh `MonotonicWriteGuardStorageAdapter` instance over the same store, which is exactly a newly started replica's state. No load, no sleeps, no timing budget.

**Regression**, biased toward everything that writes an explicit `MeshNode.Version` (found by grepping the test tree for `Version = ` alongside a `.Write(`):

| scope | result |
|---|---|
| `MeshWeaver.Hosting.Test` (whole project) | **129 / 129** |
| write-integrity pins: `MonotonicWriteGuardWindowTest`, `RefusedWriteOwnerRebaseTest`, `StaleActivationSeedRollbackTest`, `StaleActivationDurableFirstSeedTest` + the 2 new | **8 / 8** |
| Monolith `Version|Storage|Persist|Recycle|NodeOps` subset | **33 / 33** |
| `PartitionRootBootstrapTest`, `OwnerlessPartitionRepairTest` | **12 / 12** |
| `GhostPartitionRootRecoveryTest`, `BulkSaveInstallTest` | **6 / 6** |

One existing test needed a **setup** change, and it is worth reading rather than skimming. `MonotonicWriteGuardTests.StaleHighWaterMark_DoesNotRefuse_WhenTheStoreMovedOutOfBand` simulated "another replica deleted + recreated the node" by writing v1 straight over v50 on the leaf, with no delete. The leaf is now version-conditional, so the **store** refuses that in-place rewind — which is precisely the behaviour this PR exists to add, so the setup, not the guard, was what broke. It now does what its own comment says: `Delete` then recreate at v1, which is how a legitimate rewind has always been expressed (the `RecreateAfterDelete*` cases beside it, and the guard's "Legitimate rewinds" note — the row is gone, so the recreate faces nothing to regress against). The property it pins is untouched and still fails if the guard ever trusts its mark over a verification read. **An in-place rewind through `IStorageAdapter` is no longer possible on a conditional backend** — that is a deliberate, load-bearing consequence, not a side effect: an out-of-band restore is done with database tooling, not through this interface.

**The Postgres SQL was executed against real Postgres**, not merely compiled — the aliased-target + conditional-`ON CONFLICT` shape on a scratch temp table: create `INSERT 0 1`, advance to 7000 `INSERT 0 1`, regressing write at v1 **`INSERT 0 0` with the row untouched**, equal-version rewrite `INSERT 0 1`. That zero row count is exactly what `WriteAsyncCore` now reads to report the refusal.

`dotnet build -c Release -warnaserror` clean on `MeshWeaver.Hosting`, `MeshWeaver.Hosting.PostgreSql`, `MeshWeaver.Hosting.Snowflake` and `MeshWeaver.Hosting.Monolith.Test`, after merging current main.

**What I could not verify locally:** the Postgres and Snowflake test suites. The PG suite needs its container fixture and this machine was already carrying heavy contention; the Snowflake one has no CI coverage at all. The SQL probe above is the strongest evidence I could get for the PG statement short of that suite — CI is the check for the rest, and the PG change touches every write, so it is the thing to watch on this PR.

### DIFF
--- a/src/MeshWeaver.Data.Contract/ActivityCategory.cs
+++ b/src/MeshWeaver.Data.Contract/ActivityCategory.cs
@@ -19,6 +19,15 @@ public static class ActivityCategory
     /// </summary>
     public const string Compilation = nameof(Compilation);
     /// <summary>
+    /// Activity recording that a write lost a version race against the durable row and was merged
+    /// into it rather than refused. Raised by the write-integrity chain
+    /// (<c>MonotonicWriteGuardStorageAdapter</c>) so a latest-wins resolution that DROPPED a value is
+    /// visible instead of silent — the failure mode of record was an acked write that rolled a row
+    /// back with no error anywhere.
+    /// </summary>
+    public const string WriteConflict = nameof(WriteConflict);
+
+    /// <summary>
     /// Activity whose category is unknown or unclassified.
     /// </summary>
     public const string Unknown = nameof(Unknown);
--- a/src/MeshWeaver.Data/Serialization/MeshNodePatchMerge.cs
+++ b/src/MeshWeaver.Data/Serialization/MeshNodePatchMerge.cs
@@ -110,6 +110,149 @@ public static class MeshNodePatchMerge
     }
 
     /// <summary>
+    /// TWO-WAY sibling of <see cref="Apply"/>, for the one caller that has no base to rebase against: a
+    /// write that lost a version race AT THE STORE. There the write-integrity chain holds only the two
+    /// full objects — the durable row and the refused write — and no common ancestor, because the losing
+    /// writer never shipped one (issue #971).
+    ///
+    /// <para><b>Why the three-way <see cref="Apply"/> cannot serve this.</b> Its whole classification
+    /// turns on <c>live == base</c>. Handed <c>baseValues: null</c> it takes the "no conflict signal"
+    /// branch and applies the patch WHOLESALE — i.e. the stale snapshot would overwrite the newer row,
+    /// which is exactly the defect. There is no argument shape that makes <see cref="Apply"/> behave
+    /// two-way, so the entry point is separate; the SEMANTICS are shared (same recursion, same
+    /// <see cref="StringDelta"/>, same array-identity merge) so the two can never drift on the parts that
+    /// are genuinely common.</para>
+    ///
+    /// <para><b>Resolution order</b> (the project's stated conflict policy): equal → nothing; nested
+    /// object → recurse; array → UNION by element identity (<see cref="MergeArray"/> with an empty base,
+    /// so no element is ever dropped — RFC&#160;7396's replace-the-array is what silently loses them);
+    /// string where one side is a single-splice SUPERSET of the other → keep the superset, both writers'
+    /// text survives; everything else (numbers, bools, dates, shape changes, and strings that diverged on
+    /// both sides) → LATEST wins and <paramref name="onOverwritten"/> is called. That callback is the
+    /// load-bearing half: latest-wins is acceptable, latest-wins invisibly is the bug.</para>
+    ///
+    /// <para>🚨 A key ABSENT from <paramref name="latest"/> is never adopted from <paramref name="stale"/>.
+    /// The serializer suppresses defaults, so absence is ambiguous between "the newer writer removed it"
+    /// and "its value is the default" — and under both readings the newer state is the one to keep.</para>
+    /// </summary>
+    /// <param name="latest">The durable object. Mutated IN PLACE into the merged result.</param>
+    /// <param name="stale">The losing write's object.</param>
+    /// <param name="prefix">Dotted member prefix used when reporting (e.g. <c>Content</c>).</param>
+    /// <param name="onMerged">Called with the member path for each leaf where BOTH writers' content survived.</param>
+    /// <param name="onOverwritten">Called with the member path for each leaf that was NOT auto-resolvable.</param>
+    public static void ApplyTwoWay(
+        JsonObject latest,
+        JsonObject stale,
+        string prefix,
+        Action<string>? onMerged = null,
+        Action<string>? onOverwritten = null)
+    {
+        foreach (var (key, staleVal) in stale.ToArray())
+        {
+            // Absent on the newer side → the newer state stands (see the default-suppression note).
+            if (!latest.TryGetPropertyValue(key, out var liveVal))
+                continue;
+            if (JsonNode.DeepEquals(liveVal, staleVal))
+                continue;
+
+            var member = string.IsNullOrEmpty(prefix) ? key : $"{prefix}.{key}";
+
+            if (liveVal is JsonObject liveObj && staleVal is JsonObject staleObj)
+            {
+                ApplyTwoWay(liveObj, staleObj, member, onMerged, onOverwritten);
+                continue;
+            }
+
+            if (liveVal is JsonArray liveArr && staleVal is JsonArray staleArr)
+            {
+                // Union by element identity — MergeArray with an EMPTY base, which is precisely
+                // "neither side can be shown to have removed anything, so keep every element".
+                var union = MergeArray(liveArr, staleArr, []);
+                if (!JsonNode.DeepEquals(union, liveVal))
+                {
+                    latest[key] = union;
+                    onMerged?.Invoke(member);
+                }
+                continue;
+            }
+
+            if (TryGetString(liveVal, out var liveStr) && TryGetString(staleVal, out var staleStr))
+            {
+                if (TryMergeTwoWay(liveStr, staleStr, out var resolved))
+                {
+                    if (!string.Equals(resolved, liveStr, StringComparison.Ordinal))
+                    {
+                        latest[key] = JsonValue.Create(resolved);
+                        onMerged?.Invoke(member);
+                    }
+                }
+                else
+                {
+                    onOverwritten?.Invoke(member);
+                }
+                continue;
+            }
+
+            if (liveVal is null)
+            {
+                // An explicit JSON null on the newer side carries no content the stale writer could
+                // have destroyed, so adopting its value keeps strictly more data.
+                latest[key] = staleVal?.DeepClone();
+                onMerged?.Invoke(member);
+                continue;
+            }
+
+            onOverwritten?.Invoke(member);
+        }
+    }
+
+    /// <summary>
+    /// The base-less string rule, shared by <see cref="ApplyTwoWay"/> and the MeshNode-level members it
+    /// cannot see (Name, Description, …). Returns <c>true</c> when the two strings auto-resolved, with
+    /// <paramref name="resolved"/> set to the value to keep; <c>false</c> when they diverged on BOTH
+    /// sides and the caller must apply latest-wins plus a warning.
+    ///
+    /// <para><see cref="StringDelta.Compute"/> reduces the pair to ONE splice (common prefix + common
+    /// suffix stripped). A splice that only INSERTS means <paramref name="stale"/> contains
+    /// <paramref name="latest"/> in full — take the superset and nothing is lost. A splice that only
+    /// REMOVES means the reverse — keep latest, again nothing is lost. A splice that does both is a
+    /// genuine divergence: with no common ancestor there is no way to tell an insertion by one writer
+    /// from a deletion by the other, so it is not auto-resolvable.</para>
+    ///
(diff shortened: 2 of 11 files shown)


rbuergi:
Copilot's review caught a real one, now fixed in 7750ac8c5.

`JsonSerializer.SerializeToNode` / `Deserialize` throw **`NotSupportedException`** when no compatible converter exists for the type or one of its members — not only `JsonException`. The merge caught only the latter, so content this serializer cannot project would have escaped conflict resolution and **faulted the whole write chain** — precisely what the guard promises never to do, and worse than the refusal it replaced.

Both documented serialization failures now report `Content` as an overwritten member (so it lands in the activity-log entry and the warning) and keep the newer row.

Deliberately **not** widened to a blanket catch: an arbitrary exception from a custom converter is a genuine fault, not an unmergeable input, and it must still propagate. Swallowing it would hide a defect behind a plausible-looking latest-wins — the exact shape this change exists to eliminate. A loud failure is acceptable; a silent wrong one is not.

Write-integrity pins re-run after the change: **8/8**.

PR #1000, opened 2026-08-09
Cold means cold: the credential snapshot stops warming itself on read

rbuergi:
## The failure

`MeshWeaver.AI.Test.AutoRouterDispatchTest.RouterIsRecognisedAgainstAColdCatalog` landed in #992 and
reddened the next unrelated tree that actually ran it — PR #989, run
[31305125041](https://github.com/Systemorph/MeshWeaver/actions/runs/31305125041), shard 3:

```
AssertionException : Expected value to be false because a cold snapshot cannot tell 'gone' from
'not loaded yet' — …(ValidMasterModel), but found True.
   at AutoRouterDispatchTest.RouterIsRecognisedAgainstAColdCatalog()
```

It failed its own **precondition**. `ReadTierCandidates().Should().BeEmpty()` passed and
`HasReadableCatalog.Should().BeFalse()` on the very next line failed — the whole test lived 67 ms
(class init `09:20:00.134` → dispose `09:20:00.205`). So the catalog went from unreadable to readable
between two adjacent statements.

## Root cause — the API, not the test

`ChatClientCredentialResolver.ReadSnapshot()` opened the live catalog subscription **as a side effect
of being read**:

```csharp
private IReadOnlyList<MeshNode> ReadSnapshot()
{
    if (snapshotSubscription is null) RebuildSubscription();   // ← a read that warms
    return cachedSnapshot;
}
```

The test's own precondition read therefore *started* the warm-up it was asserting had not happened.
Whether the first emission landed before or after the next line was a coin flip; under CI load it
landed in between. (The test's comment — "deliberately NO `EnsureSubscription()` — nothing has warmed
the snapshot" — was simply not true of the code.)

That is a defect in the API, not just in the test: `HasReadableCatalog` exists **precisely** to report
the pre-warm window, and asking it ended that window. A query that mutates what it reports cannot be
observed by anyone — the test just happened to be the first caller to try.

## The fix — reads are pure, warming is an owner's act

- `ReadSnapshot()` no longer subscribes. Watching the catalog starts only when an owner says so:
  `EnsureSubscription()` / `WatchPartition()` / `WatchSharedProvider()`.
- The DI registration in `AddLanguageModelType` now performs that act when it builds the mesh's shared
  resolver. Every consumer resolves it from DI (verified: no code anywhere constructs one), so every
  consumer is warm exactly as before — marginally earlier, at service resolution rather than at first
  lookup. **No behavioural change for a running deployment.**
- The test builds its **own** resolver. Nothing else holds a reference to it and a read cannot warm it,
  so the snapshot is empty at the first assertion, at the last, and at every one in between. A closing
  re-check of `HasReadableCatalog` **brackets** the precondition — proof the router assertions were
  answered by the static rule and not by a catalog that warmed midway.

The `HasReadableCatalog.Should().BeFalse()` precondition is **kept, unweakened, and now asserted
twice**. Nothing was skipped, retried, slept on, or widened.

## Coldness is deterministic, not "more likely"

A new sibling test, `ReadingNeverWarms_OnlyTheOwnerDoes`, pins it on a **positive** signal rather than
on the passage of time (no sleep, no window): warm the mesh's *shared* resolver and wait until the
catalog is demonstrably readable through it — at that instant a private resolver that has been read
repeatedly and never asked to subscribe must **still** read empty. Then one explicit
`EnsureSubscription()` makes the same instance readable.

Verified by counter-proof: with `ReadSnapshot`'s warm-on-read temporarily restored, that test fails
**deterministically** —

```
Expected collection to be empty because the catalog is readable — so an unasked resolver reading empty
can only mean reads do not subscribe …, but found 1 item(s).
Failed! - Failed: 1, Passed: 4
```

— while `RouterIsRecognisedAgainstAColdCatalog` still *passed* on that same old-behaviour run. That is
the point: on this machine the old race falls the lucky way, which is exactly why it only ever showed
up on CI. The new test converts a probabilistic failure into a deterministic one.

## Siblings checked

- `ModelTierCatalogTest`, `BuiltInLanguageModelProviderTest` — pure POCO suites over an explicitly
  constructed catalog, no mesh, no snapshot. No ambient dependency.
- The other three `AutoRouterDispatchTest` tests go through `SeedCatalogAndWarm()`, which warms
  explicitly and then waits on the *actual* condition (seeded models + router visible). Sound.

## Verification

- `dotnet build src/MeshWeaver.AI -c Release -warnaserror` — clean (0 warnings)
- `dotnet build test/MeshWeaver.AI.Test -c Release -warnaserror` — clean
- `dotnet test test/MeshWeaver.AI.Test -c Release` — **1120 passed**, 0 failed, 5 skipped
- `dotnet test test/MeshWeaver.Threading.Test -c Release` — **128 passed**, 0 failed (the other suite
  that drives the resolver)

## What's New

Skipped deliberately — internal test-determinism fix plus an internal warm-up seam with no
user-visible effect.


rbuergi:
Correct, and worth being exact about — fixed in 1740076.

`EnsureSubscription` governs the long-lived **watch**, not whether the snapshot can ever be
populated: `Resolve` (hence `HasUsableCredential`) is a read-**through**, and a miss fires the
single-flight `TriggerAuthoritativeRefresh`, which merges into `cachedSnapshot` with no subscription
in play. So the rule is: a resolver stays cold as long as its owner neither warms it **nor asks it
for a credential**; the snapshot projections (`ReadTierCandidates`, `ReadTiers`,
`HasReadableCatalog`, `IsRouterSelection`) never end the window.

The cold tests already sit on the right side of that line — they call only projections — and that is
now stated at both call sites so a later edit cannot re-open the window by accident. Worth noting for
the one non-obvious case: `ResolveDefaultModelId()` *does* pass `HasUsableCredential` down, but as a
predicate over the candidate list — `ModelTierCatalog.Eligible` filters an empty list, so the
predicate is never invoked and no read-through happens.

Issue #999, opened 2026-08-09
Orleans tests cannot parallelize: SharedOrleansFixture statics + ResetSharedState() Clear() at every cluster init


rbuergi:
Blocks intra-project parallelism on `MeshWeaver.Hosting.Orleans.Test` (weighted 239s in `shard-assign.sh`). Measured, with a precise mechanism.

## Measurement

Release build, native xUnit v3 host, `parallelizeTestCollections: true` + `maxParallelThreads: 4`:

```
MeshWeaver.Hosting.Orleans.Test   Total: 136, Failed: 8, Time: 81.0s
```

All 8 failures share one shape:

```
No node found at 'TestUser/_Thread/parent-for-delegation-test-eeb1/e3c1d11b'
No node found at 'TestUser/_Thread/portal-flow-orleans-test-bc7a/e3433644'
No node found at 'TestUser/_Thread/rapid-submits-test-459e/89eb49eb'
```

Affected: `Delegation_CreateCellsThenThread_ExecutionStartsAndCompletes`, `PortalFlow_CreateThread_CreateCells_Submit_ExecutionCompletes`, `RapidSubmits_PileUpAndAllIngest`, `AutoExecute_UpdateThreadMessageContent_RoutesToResponseGrain`, `Delegation_NodeChanges_PropagateFromSubThread`, `ThreadHub_LocalWorkspaceWrite_VisibleViaGetDataRequest`, `SubmitChat_FromSidePanel_CellsAppearAndStream`, `DelegationSubThread_RoutesAfterCreate`.

## Mechanism — not ports, not grain directory

Paths are GUID-suffixed per test, so this is **not** a name collision. It is process-wide static state being wiped mid-test.

`SharedOrleansFixture` holds:

```csharp
internal static readonly ConcurrentDictionary<string, MeshNode> SharedNodes
internal static readonly ConcurrentDictionary<string, List<object>> SharedPartitionObjects
internal static SwappableChatClientFactory SwappableFactory { get; } = new()
```

and wipes them at the **start of every cluster init** — called from both `SharedOrleansFixture.InitializeAsync` and `OrleansTestBase`:

```csharp
internal static void ResetSharedState()
{
    SharedNodes.Clear();
    SharedPartitionObjects.Clear();
    SwappableFactory.Reset();
}
```

Under parallelism, class B's init calls `SharedNodes.Clear()` while class A is mid-test — deleting A's nodes. Hence "No node found".

The fixture documents its own precondition:

[...]

Issue #980, opened 2026-08-09
StaleActivationSeedRollbackTest: out-of-band IStorageAdapter write not visible to immediate durable read under concurrent load


rbuergi:
Surfaced while measuring intra-project test parallelism (#979). **The race exists on `main` today** — single-threaded execution just never schedules the two operations closely enough to expose it.

## Symptom

`MeshWeaver.Hosting.Monolith.Test.StaleActivationSeedRollbackTest`, both tests:

```
MeshWeaver.Reactive.Assertions.AssertionException :
Expected value to be 5000 because the out-of-band advance must be durable
before we reactivate, but found 1.
   at StaleActivationSeedRollbackTest.cs:100
```

- `ReactivatedOwner_SeedsFromDurableStorage_NotFromTheRoutingCache`
- `PostRecycleWrite_LandsAboveTheDurableVersion_AndDoesNotRollTheStoreBack`

## Reproduction

Release build, native xUnit v3 host, `test/MeshWeaver.Hosting.Monolith.Test`:

| Config | Result |
|---|---|
| full suite, `maxParallelThreads: 1` | 499 tests, **0 failed**, 458.3s |
| full suite, `maxParallelThreads: 4` + `parallelizeTestCollections: true` | 499 tests, **2 failed**, 146.0s |
| that class alone, single-threaded | 2 tests, **0 failed**, 3.0s |

To reproduce, drop this at `test/MeshWeaver.Hosting.Monolith.Test/xunit.runner.json` (the per-project opt-in added in #979) and run the project's native host:

```json
{
  "parallelizeAssembly": false,
  "parallelizeTestCollections": true,
  "maxParallelThreads": 4,
  "methodTimeout": 30000
}
```

## What it is not

**Not a path collision.** The node path is GUID-suffixed per test (`TestData/stale-seed-<guid>`), so concurrent classes are not sharing a node.

## The actual question

The test deliberately writes **straight through `IStorageAdapter`**, bypassing `IMeshChangeFeed`, so the path-resolution cache keeps serving the pre-recycle snapshot:

```csharp
const long durableVersion = 5000L;
await Storage.Write(created with { Name = "durable-advance", Version = durableVersion }, JsonOptions)
    .Should().Within(30.Seconds()).Emit();

(await ReadDurable(path).Should().Within(30.Seconds()).Emit())!
    .Version.Should().Be(durableVersion, "the out-of-band advance must be durable before we reactivate");
```

[...]

PR #1008, opened 2026-08-09
perf(ci): re-measure shard weights — worst shard 7.4 → 5.6 min

rbuergi:
The shard weights in `shard-assign.sh` had drifted again. The file warns *"Re-measure when a project's runtime changes materially"* — this is that re-measure.

## The drift

Measured from run `31311680838` (2026-08-09) using the per-project `[CI] <name> exit=` markers:

| Project | committed | measured | |
|---|---|---|---|
| `PluginCatalog.Test` | 30 | **87** | **+190%** |
| `Hosting.PostgreSql.Test` | 130 | 161 | +24% |
| `GitSync.Test` | 51 | 62 | |
| `Query.Test` | 49 | 57 | |
| `AI.Test` | 263 | 177 | −33% |
| `Hosting.Monolith.Test` | 345 | 270 | −22% |
| `Persistence.Test` | 97 | 73 | −25% |
| `Content.Test` | 91 | 74 | its new intra-project parallelism |

The **under**-estimates are what cost time. Across four green main runs, shard 1 was the pole in every one (8.0–8.3 min) and shard 0 the idler in every one (5.4–5.7 min).

## Effect

Re-running the LPT against the refreshed table:

```
before:  max 7.40 min   min 5.20 min   spread 2.20
after:   max 5.57 min   min 5.42 min   spread 0.15
```

**~1.8 min off the critical path for a table edit.** Verified the script still emits 66 disjoint entries with no duplicates across the six shards.

## Two stale header claims corrected

- The sharding floor is no longer "345 s" — it is 270 s, and Monolith is split 2, so the real floor is ~135 s.
- Recorded that **cutting shard count makes the critical path worse** while tests are single-threaded: total measured work is 32.5 min, so ideal per-shard is 5.4 min at 6 shards, 8.1 at 4, 10.8 at 3. Fewer shards only becomes attractive once intra-project parallelism shrinks the 32.5 min — which is the point of the opt-in added in #979.

## ⚠️ One-run measurement, deliberately

The previous table used the max over two green runs. This one is a single run: the second green run's logs no longer carried the `[CI]` markers by the time I measured. A fresh single measurement still beats a stale two-run max — LPT absorbs modest weight error; what it cannot absorb is a 190% under-estimate like PluginCatalog's. The header now says to take the max over two at the next re-measure.

## No What's New entry

Pure CI infrastructure, no user-visible effect — the `pullrequest` skill's step 0.5 says to skip and say so.

### DIFF
--- a/.github/scripts/shard-assign.sh
+++ b/.github/scripts/shard-assign.sh
@@ -24,16 +24,37 @@
 # ALC-heaviest projects — onto shard 0 and made it the ~19-minute long pole of
 # every run.
 #
-# Weights are wall-clock SECONDS, the MAX over two green runs (30903428924 and
-# 30903158378, both 2026-08-04) so a shard is sized for the bad case, not the
-# average. The per-project `[CI] <name> exit=` markers in the test job's log ARE
-# the measurement: take the delta between each `::group::<name>` and its marker.
-# Re-measure when a project's runtime changes materially — the previous table
-# (2026-07-13) had drifted far enough to cost ~100 s of long pole (Monolith
-# 200→345, Data 25→73, PluginTester 60→8, Persistence 165→97).
+# Weights are wall-clock SECONDS. The per-project `[CI] <name> exit=` markers in
+# the test job's log ARE the measurement: take the delta between successive
+# markers (the first project's start is the `Run Tests` step's own start_at).
 #
-# The floor for ANY sharding is the heaviest single project: today
-# Hosting.Monolith.Test at 345 s. Adding shards below that buys nothing.
+# 🚨 Re-measure when a project's runtime changes materially. This table has now
+# drifted twice, and BOTH times the cost was paid on the long pole:
+#   * 2026-07-13 → 2026-08-04: Monolith 200→345, Data 25→73, Persistence 165→97.
+#   * 2026-08-04 → 2026-08-09: PluginCatalog 30→87 (+190%), PostgreSql 130→161,
+#     AI 263→177, Monolith 345→270, Content 91→74. The under-weighted entries are
+#     what hurt: shard 1 ran 7.4 min against a 5.4 min ideal, i.e. 2.0 min of pure
+#     imbalance on EVERY run, while shard 3 idled at 5.2.
+#
+# Current table: measured from run 31311680838 (2026-08-09). Simulating the LPT
+# below against those weights gives max 5.57 / min 5.42 min — a 0.15 min spread.
+#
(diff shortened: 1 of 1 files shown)


rbuergi:
Closing — this **failed its own success criterion**, and it introduced a regression. Recording the evidence because the underlying measurement is still useful.

## It did not deliver

| | predicted | actual (run 31313671216) |
|---|---|---|
| worst shard | 5.57 min | **8.5 min** |
| spread | 0.15 min | **2.6 min** (was 2.2 before) |
| total | ~15.4 min | 16.8 min vs 17.2 baseline — inside noise |

Shard times: 6.7 / 6.0 / **8.5** / 7.2 / 5.9 / 8.0.

## And it regressed the AI split

`MeshWeaver.AI.Test` is declared `2`-way split so its halves land on different shards. With the committed weights they do (shards 3 and 4). With mine, **both land on shard 5**, undoing the split and putting all ~177s on one shard:

```
shard 3: old=1 new=0
shard 4: old=1 new=0
shard 5: old=0 new=2
```

## The real lesson: weight-balance ≠ time-balance

The script reported near-perfect weighted loads — `326 325 325 325 325 325` — and wall time still spread 2.6 min. So the failure is not "the numbers were stale", it is that **per-project weight is a weak predictor of shard wall time**. Tuning the table harder will not fix that. Contributing factors, unseparated:

- per-project host startup and Testcontainers pulls that the weights do not model
- run-to-run variance on shared runners (the reason the header asks for a max over two runs, which I could not honour — the second run's `[CI]` markers were gone)
- LPT packing being sensitive to small weight changes, as the AI split shows

## What IS worth keeping from this

The drift is real and measured (run 31311680838), even if re-weighting alone does not pay:

```
PluginCatalog.Test    30 -> 87   (+190%)
Hosting.PostgreSql   130 -> 161
AI.Test              263 -> 177
Hosting.Monolith.Test 345 -> 270
Persistence.Test      97 -> 73
Content.Test          91 -> 74   (its new intra-project parallelism)
```

Total measured test work: **32.5 min across 63 projects**, so the ideal per-shard load is 5.4 min at 6 shards — and observed shards run 5.9-8.5. The gap between 5.4 ideal and ~7 typical is overhead and variance the weights do not capture.

## Recommendation

Do not re-attempt this as a weights-only change. The leverage is in **shrinking the 32.5 min**, not redistributing it: #999 (Orleans statics, ~239s) and #980 (Monolith, 458->148s). Once intra-project parallelism lands, both the shard count and the weights should be re-derived together against the new reality.

PR #1010, opened 2026-08-09
Delete the Orleans fixture statics that made the assembly un-parallelisable

### LINKED ISSUE #999: Orleans tests cannot parallelize: SharedOrleansFixture statics + ResetSharedState() Clear() at every cluster init

rbuergi:
Fixes #999

## The defect

`SharedOrleansFixture` kept the test cluster's mesh-node store and its chat-client factory in process-wide statics, and wiped them with `ResetSharedState()` at the **start of every cluster init**:

```csharp
internal static readonly ConcurrentDictionary<string, MeshNode> SharedNodes
internal static readonly ConcurrentDictionary<string, List<object>> SharedPartitionObjects
internal static SwappableChatClientFactory SwappableFactory { get; } = new()

internal static void ResetSharedState()
{
    SharedNodes.Clear();
    SharedPartitionObjects.Clear();
    SwappableFactory.Reset();
}
```

That wipe **was** the per-class isolation, and it is exactly why the assembly could not be parallelised: with two classes in flight, B's init calls `SharedNodes.Clear()` while A is mid-test, deleting A's nodes — the eight `No node found at 'TestUser/_Thread/…'` failures in the issue. The fixture documented its own precondition: *"Safe because classes run strictly sequentially."*

AGENTS.md names this shape precisely: *"the moment you add a `Clear()` method 'for test isolation', that method IS the proof of the bug."* So the fix is to delete the statics, not to lock them — a lock would re-serialise the very thing we want parallel.

## Why they existed, and what replaced them

Orleans instantiates `ISiloConfigurator` / `IHostConfigurator` types via `new()`, so a configurator can never be handed an instance. The issue called that constraint "the real problem to solve".

Most of it dissolves once you notice **a silo's DI container is already per-cluster**:

- The silo's `InMemoryStorageAdapter` is a singleton of that silo's own container. Per-cluster by construction, disposed with the cluster. Nothing to inject.
- Same for the chat factory, once `SharedSiloConfigurator` registers `SwappableChatClientFactory` as a **type** instead of an instance. The silo's container builds it; `SharedOrleansFixture.ChatFactory` reads it back out.

What was genuinely missing was a way to point the **Orleans client** at what the silo already owns. `TestCluster.InitializeClientAsync` offers no hook, but `TestClusterHostFactory.CreateClusterClient` takes a post-configure `Action<IHostBuilder>` — a closure. So `OrleansTestCluster.DeployAsync` sets `InitializeClientOnDeploy = false` and builds the client host itself, and `ShareSiloNodeStore` hands it the silo's adapter.

`ResetSharedState()` and `SwappableChatClientFactory.Reset()` are **deleted**, along with the eight `try/finally` blocks that existed only to put the chat factory back.

### Two consequences, both deliberate

**1. `Cluster.Client` is null on these clusters.** The client host belongs to `OrleansTestClusterHost`; callers use `ClientServices` / `ClusterClient`. This also closes a teardown gap in passing: `TestCluster.DisposeAsync` disposes the client through `IHost.DisposeAsync`, which never runs `StopAsync` — the new drain always stops the client host first, before the silos, before anything is disposed.

**2. The cluster declares `ConnectionTransportType.TcpSocket`.** `TestCluster`'s default transport is `InMemory`, wired through an `InMemoryTransportConnectionHub` that is a **private field of `TestCluster`**. A client we construct ourselves can never join that hub — it would dial the silos' TCP endpoints and get `ConnectionRefused`. Real loopback sockets are what `TestCluster` used before the in-memory transport existed. Measured cost: **131.1s → 139.8s** serial for the whole assembly, locally.

`TwoSiloCacheUpdateFixture` is the one place a *silo* genuinely needs an injected instance (two silos, one store, no client to route through). It takes `TestCluster`'s only silo-host hook, `CreateSiloAsync`, with the transport consequence spelled out at the call site — that setter flips `ConnectionTransport` to `TcpSocket` because replacing `DefaultCreateSiloAsync` also drops the in-memory transport wiring.

## Parallelism

#979 left this project out, noting it "boots a real cluster per test class, where port collisions are a genuine risk". Port collisions were never the mechanism: the paths in all eight failures are GUID-suffixed, so nothing could collide by name, and Orleans' port allocator already handles concurrent clusters. The statics were.

18-core Mac, Release, native xUnit v3 host, whole assembly:

| run | tests | failed | time |
|---|---|---|---|
| serial, before this branch | 136 | 0 | 131.1s |
| serial, after | 136 | 0 | 139.8s |
| parallel ×4 | 136 | 0 | 91.9s |
| parallel ×4 | 136 | 0 | 91.6s |
| parallel ×4 | 136 | 0 | 77.9s |

All eight tests named in #999 pass in all three parallel runs.

**CI then answered the question the local box could not — with more variance.** `MeshWeaver.Hosting.Orleans.Test` on the 4-vCPU `ubuntu-latest` runner:

| | tests | failed | time |
|---|---|---|---|
| main, serial ([31313253809](https://github.com/Systemorph/MeshWeaver/actions/runs/31313253809)) | 136 | 0 | 221.7s |
| this PR, parallel ×4 ([31314161999](https://github.com/Systemorph/MeshWeaver/actions/runs/31314161999)) | 136 | 0 | 155.4s |
| this PR, parallel ×4 ([31315000994](https://github.com/Systemorph/MeshWeaver/actions/runs/31315000994)) | 136 | 0 | 197.6s |

**136/136 green on CI every time — that is the claim #999 actually rests on.** The speed-up is not the 1.5× the 18-core box suggested: two parallel samples span 155.4–197.6s against one 221.7s serial sample, i.e. 1.43× on the low sample and 1.12× on the high one. A shared runner has enough variance that no single number here is worth quoting, so read the CI gain as real but modest and under 1.5×.

The `shard-assign.sh` weight is left at 239. Re-weighting wants more than three samples, and a stale-high weight only under-fills that shard — it never mis-assigns one.

## Out of scope

`OrleansClusterDisposal.Pending` is still a `static readonly ConcurrentBag`. It is an append-only completion sink drained once at assembly teardown — never read as state, never cleared, and parallel-safe — but it is a static collection. Making it an instance means threading the assembly fixture through ~50 test class constructors, which does not belong in this change.

## CI red is not this branch's

Run [31314161999](https://github.com/Systemorph/MeshWeaver/actions/runs/31314161999) (`7a4909ab`) was green on all six shards. The later run [31315000994](https://github.com/Systemorph/MeshWeaver/actions/runs/31315000994) (`9e8fc86c`, which differs only by a comment edit and a restored `ILogger` argument) is red on shards 2 and 4:

- **`WhatsNewEntryIntegrityTest`** fails on `2026-08-09-check-a-deployment-is-current.md`, an entry that came from main. This branch adds no What's New entry, and **main is red on the identical assertion** ([31314807609](https://github.com/Systemorph/MeshWeaver/actions/runs/31314807609)).
- **`FutuReAnalysisTest.EuropeRe_LineOfBusiness_Search_ShouldReturn8LoBs`** misses a `MonolithMeshTestBase` teardown quiescing budget by 289 ms (2289 ms against 2.00 s), in a different assembly and shard. This branch touches only `test/MeshWeaver.Hosting.Orleans.Test/`.

The Orleans shard is green in both runs.

## What's New

Skipped — test infrastructure only, no user-visible effect.


rbuergi:
## CI status — do not read the red as this branch's

Two runs on this branch:

| run | head | result |
|---|---|---|
| [31314161999](https://github.com/Systemorph/MeshWeaver/actions/runs/31314161999) | `7a4909ab` | **green, all 6 shards** |
| [31315000994](https://github.com/Systemorph/MeshWeaver/actions/runs/31315000994) | `9e8fc86c` | red on shards 2 and 4 |

The only difference between those two commits is `aab1f094d`, which edits one comment and restores an `ILogger` argument. The Orleans shard (0) is green in **both** — that is where this PR's entire diff lives.

**Shard 4 — `WhatsNewEntryIntegrityTest.EveryEntry_CarriesTheFrontmatterTheFeedRendersFrom`**

```
2026-08-09-check-a-deployment-is-current.md: Category is 'What's New', must be one of Feature | Fix
2026-08-09-check-a-deployment-is-current.md: Order is '(none)', must be '-20260809'
```

That entry is main's, not this PR's — this branch adds no What's New entry at all. **main is red on it right now**: [run 31314807609](https://github.com/Systemorph/MeshWeaver/actions/runs/31314807609) (`fd277b36`) fails the identical assertion. Merging main again cannot help; the entry has to be fixed on main.

**Shard 2 — `FutuRe.Test.FutuReAnalysisTest.EuropeRe_LineOfBusiness_Search_ShouldReturn8LoBs`**

```
FutuReAnalysisTest left Observe subscriptions pending past the Quiescing budget
  1 pending callback(s) after 2.00s: …CreateNodeRequest…(2289ms)
```

A `MonolithMeshTestBase` teardown budget missed by 289 ms, in a different assembly and a different shard. This branch touches only `test/MeshWeaver.Hosting.Orleans.Test/`, so there is no path from the diff to it.

Not merging. Per AGENTS.md the gate is absolute, and neither red is mine to make green.

### The number the PR was opened to get, reported with its variance

`MeshWeaver.Hosting.Orleans.Test`, CI, 4-vCPU `ubuntu-latest`:

| | tests | failed | time |
|---|---|---|---|
| main, serial ([31313253809](https://github.com/Systemorph/MeshWeaver/actions/runs/31313253809)) | 136 | 0 | 221.7s |
| this branch, parallel ×4 ([31314161999](https://github.com/Systemorph/MeshWeaver/actions/runs/31314161999)) | 136 | 0 | 155.4s |
| this branch, parallel ×4 ([31315000994](https://github.com/Systemorph/MeshWeaver/actions/runs/31315000994)) | 136 | 0 | 197.6s |

**136/136 green every time — that is the part that is not in doubt, and it is what #999 was about.**

The speed-up is not the 1.5× the 18-core measurement suggested. Two parallel samples span
155.4s–197.6s against a single 221.7s serial sample; on the low sample it is 1.43×, on the high one
1.12×. A shared 4-vCPU runner has enough variance that neither number should be quoted alone, and
three samples cannot separate the runner from the change. Treat the CI gain as "real but modest,
somewhere under 1.5×", and re-measure over more runs before re-weighting
`shard-assign.sh` (still 239).

PR #1014, opened 2026-08-09
Three idioms, nine sites, rediscovered by chance — the convention was never written down

### LINKED ISSUE #995: Seven sites, one defect: Rx timers whose subscription reaches no owner's disposal chain
## Summary

Seven sites, one defect: **an Rx timer whose subscription is not owned by anything that dies with the thing its closure captured.** An `Observable.Timer` / `Observable.Interval` entry lives on the process-wide `TimerQueue`, which is a strong GC root — so while it is pending it holds its tick closure and everything that closure transitively captured. When that tail is a `MessageHub`, the hub cannot be collected.

Each occurrence has been found by `MeshHubDisposalLeakTest` firing at random on an unrelated PR, months apart, and fixed in isolation. This issue is about the pattern, not another instance.

## The ledger

Two sub-shapes.

**(1) The subscription is discarded outright** — nothing can cancel the pending timer, so teardown does not release it:

| Site | Fixed in |
|---|---|
| `ActivityControlPlaneExtensions.SubscribeWithReEstablish` | #991 |
| `ThreadSubmission.InstallSubmissionWatcher` | #991 |
| `ThreadExecution.InstallExecRoundWatcher` | #991 |
| `ThreadExecution.InitializeThreadLifecycle` | #991 |

**(2) The subscription is held, but strongly pins an owner that may never be disposed** — fine for an owner that always tears down, fatal for an abandoned one (a hub created at `RunLevel=1` that never reaches teardown):

| Site | Shape | Fix at the time |
|---|---|---|
| `MessageHub` disposal watchdog | uncancelled `Task.Delay(25s)` rooted the whole hub graph after **every** dispose | `Observable.Timer` + `TakeUntil(disposalCompleted)` |
| `MessageHub.InstallStaleCallbackScanner` | `Observable.Interval` closure captured `this` | `WeakReference<MessageHub>` + self-dispose on dead target |
| `JsonSynchronizationStream` heartbeat | `Observable.Interval` closure captured `hub` | `WeakReference<IMessageHub>` + self-dispose on dead target |

Each of those three carries a comment naming the leak it caused, which is good — but a comment on the fixed site cannot prevent the next one.

## Why the current detection is not enough

`MeshHubDisposalLeakTest` is what caught all of these, and it is a **discovery** probe, not a regression test (now documented on the test itself, in #991):

- **It samples.** A root live only for a bounded window — #991's was ~1 s — is caught only if the probe's forced GC lands inside that window. Otherwise: green, defect fully present.
- **It cannot attribute.** It reports the first `MessageHub` reachable from any non-stack root, so it names whatever chain it walked, not the one you are chasing.
- **It is inconclusive off Linux.** ClrMD snapshot-attach throws on macOS, so a survivor SKIPs (#674) — a developer cannot reproduce locally on a Mac.

Net effect: the defect is found late, at random, on someone else's PR. #991 red-flagged PR #984, whose diff was five files with no C# in it at all.

## What to decide

Audit every `Observable.Timer` / `Observable.Interval` in `src/` and classify each: does its subscription reach an owner's disposal chain, and does its closure strongly capture something that might outlive that owner? (`grep -rn "Observable\.Timer\|Observable\.Interval" --include="*.cs" src/` is ~35 hits, most of which are fine — composed inside a stream, or already assigned to a `SerialDisposable` / field the owner disposes.)

Then decide whether this is worth enforcing rather than remembering. Options, roughly in cost order:

1. **Nothing** — accept that the probe finds one every few months, now that its limits are documented.
2. **A convention plus a review habit**: a timer subscription is either assigned into an owner-disposed disposable, or composed into a stream that is; a periodic timer that outlives its subject holds it weakly. Write it down next to the existing guidance in `AsynchronousCalls.md` / `ControlledIoPooling.md`.
3. **A test that enumerates the sites** — cheap and deterministic, but only as good as its list.
4. **An analyzer** (Roslyn) for "`Observable.Timer(...)`/`Interval(...)` whose `.Subscribe(...)` result is discarded". Catches sub-shape (1) mechanically. Sub-shape (2) is a capture-lifetime question and is probably not statically decidable here.

My read: (4) covers exactly the shape that caused #991 and is mechanical, while (2) is what actually covers both. Worth a maintainer's call rather than an agent's — the cost is small either way, but so is the incident rate, and the audit in step one should probably come first to size it.

## Provenance

Split out of #991 at review, where fixing the fourth instance made the pattern visible.

rbuergi:
Implements the audit's recommendation on #995: **option (2), write the convention down.** No analyzer, no enumerating test, no behaviour change — one new architecture page and four cross-links.

## Why not the analyzer

The audit produced the number that settles it. A sub-shape-1 Roslyn rule — "`Observable.Timer`/`Interval` whose `.Subscribe(...)` result is discarded" — would have flagged **1 of the 46** timer/gate sites in `src/` and **1 of the 6** findings. The second live defect (`MeshNodePickerView`) was held in a field and reads as correct to any static rule, as do all four ambiguous sites, as does the majority of the prior art. Widening it is worse: an unrestricted discarded-`Subscribe` rule fires on **388 of the 697** `Subscribe`/`Connect` invocations in `src/`, ~386 of which are the cold-observable `stream.Update(...).Subscribe(...)` pattern AGENTS.md **mandates**.

The page says all of that, with the numbers, because someone will propose the analyzer again.

## New page, not a section — and why

`Doc/Architecture/SubscriptionOwnership`, a **sibling** of `AsynchronousCalls` rather than a section inside it. Two reasons: `AsynchronousCalls.md` is already 1456 lines and its scope is *composing async work without `await`*, whereas this is a **lifetime/disposal** rule; and it needs to be linkable from `HubDisposalModel` and `DebuggingDisposalAndLeaks`, which a subsection of a 1456-line page serves badly. `AsynchronousCalls` gets a short new subsection under "Subscribe is mandatory" — the rule that makes this class easy to get wrong — that states the pending-vs-completed distinction and hands off.

## What it covers

- **The rule**, applicable without the backstory: a timer/interval/delayed subscription must reach its owner's disposal chain, on every path. **"Held in a field" is not ownership.**
- **Both sub-shapes with real before/after**, drawn from the actual fixes — `ActivityLogLogger`'s discarded `Subscribe` → `SerialDisposable` + `hub.RegisterForDisposal` (#1007), and `MeshNodePickerView`'s field-that-only-the-next-keystroke-disposed → `Disposables.Add(...)` registered once in `OnInitialized` (#1007), the sibling of `MauiViewPack`'s `Disposables.Add(_searchSub)`. Plus the two idioms for owners whose disposal is *not* guaranteed: `WeakReference` + self-dispose (`InstallStaleCallbackScanner`, the sync-stream heartbeat, `KernelContainer`) and the severable cell (`MeshNodeTypeSource`, with its measured 9-pinned-hubs-vs-0).
- **The measured truth table**, verbatim — the most valuable part, because it stops people reasoning from memory. The three counter-intuitive rows are called out: `Task.WhenAny` **removes its continuation from the losing task** so it does not root (this cleared `DataContext.OpenInitializationGate`, which by inspection reads as a 120 s hub-wide leak and is not one); a **fired** `Observable.Timer(...).Subscribe(...)` handle, still held, does **not** retain its callback closure; a bare `Task.Delay(...).ContinueWith(...)` **does** root. That last fact is also *why* discarding `stream.Update`'s handle is safe — which is the same fact that makes the analyzer's false-positive rate what it is, so the page ties the two together rather than leaving them as unrelated trivia.
- **Why a green `MeshHubDisposalLeakTest` is not evidence** — it samples, it cannot attribute, it SKIPs on macOS (#674) — and what the right pin is instead: a targeted, timing-free ownership test plus a **negative control**, with #1007's two negative-control results as the worked example. Also: assert *both* halves (teardown disposes the handle **and** the working path arms into it), or the registration can guard a slot nothing writes to and the test still passes.
- **How to re-audit**: three passes, only the first mechanical. The ~130-line Roslyn scanner stays a **one-off script re-run at audit time**, out of the tree — the reasoning (allowlist rot, coupling a unit test to repo layout) is recorded so the next person does not re-litigate it.

Written as running prose with real code, not a bullet list of platitudes — a developer who has never seen #995 should be able to apply it, and the four-clause summary at the end is what fits in a review comment.

## Cross-links

Added from the four places a reader would actually start: `AsynchronousCalls` ("Subscribe is mandatory"), `DebuggingDisposalAndLeaks` (the ClrMD probe section — which now also carries the "a pass here pins nothing" warning inline, matching the remarks already on the test), `HubDisposalModel` ("adding disposal work"), and the `Architecture` topic map under **Reactive core**.

## Verification

- `dotnet build src/MeshWeaver.Documentation -c Release -warnaserror` — **0 warnings, 0 errors**.
- `DocumentationLinkIntegrityTest` — **passed** (every link is absolute `/Doc/…`, per `AuthoringDocumentation`).
- No H1 in the new page: the frontmatter `Name` is the page title, so the body opens on the lead paragraph.

## Notes

- **No What's New entry.** The feed is for changes a portal user notices; this is a contributor-facing engineering convention with no behaviour change. Same call `ReleaseToProductionPipeline.md` made, and the same reason #1007 skipped one for the fixes themselves.
- **`Fixes #995`** — #995 asked for a decision between four options and the audit recommended (2). The two live defects it turned up were closed by #1007; writing the convention closes the class item, which is what the issue is about.
- One thing worth flagging separately rather than fixing here: the audit's **four ambiguous sites** (`MeshNodeStreamCache` idle sweep, `EventSubscriptionRunner`'s unbounded `FireAt − now` delay, `SnowflakeChangeFeedPoller`, `MessageService`'s 30 s deferred-delivery delay) remain as the audit left them — safe *if* the owner is disposed, which is the assumption that failed in all three ledger cases. The page now states the rule they'd be measured against; whether to harden them is a separate call.

Fixes #995

### DIFF
--- a/src/MeshWeaver.Documentation/Data/Architecture.md
+++ b/src/MeshWeaver.Documentation/Data/Architecture.md
@@ -67,7 +67,7 @@ The catalog below lists every page; these are the load-bearing reads per theme.
 
 | Theme | Start here | Go deeper |
 |---|---|---|
-| **Reactive core** | [Asynchronous Calls](AsynchronousCalls) — the no-`await` rulebook | [Actor Model](ActorModel) · [Controlled I/O Pooling](ControlledIoPooling) · [Initialization Gates](InitializationGates) · [Aggregating Providers](AggregatingProviders) · [Hub Disposal Model](HubDisposalModel) |
+| **Reactive core** | [Asynchronous Calls](AsynchronousCalls) — the no-`await` rulebook | [Actor Model](ActorModel) · [Controlled I/O Pooling](ControlledIoPooling) · [Subscription Ownership](SubscriptionOwnership) — a pending timer is a GC root · [Initialization Gates](InitializationGates) · [Aggregating Providers](AggregatingProviders) · [Hub Disposal Model](HubDisposalModel) |
 | **Reading & writing nodes** | [CQRS — Queries vs. Content Access](CqrsAndContentAccess) | [MeshNode Stream Cache](MeshNodeStreamCache) · [Request via Stream Update](RequestViaStreamUpdate) · [Data Access Patterns](DataAccessPatterns) · [Workspace References](WorkspaceReferences) · [Content Chunk Navigation](ContentChunkNavigation) |
 | **Storage & partitions** | [Postgres Schema Architecture](PostgresSchemaArchitecture) | [Partition Storage Routing](PartitionStorageRouting) · [Partition Storage Hubs](PartitionStorageHubs) · [Partitioned Persistence](PartitionedPersistence) · [Static Repo Import](StaticRepoImport) · [Vector Search](VectorSearch) |
 | **Security** | [Access Control](AccessControl) | [Granting Access](GrantingAccess) · [AccessContext Propagation](AccessContextPropagation) · [Owner Injection](OwnerInjection) · [Permission API](PermissionApi) · [Invitation-Only Onboarding](InvitationOnlyOnboarding) |
--- a/src/MeshWeaver.Documentation/Data/Architecture/AsynchronousCalls.md
+++ b/src/MeshWeaver.Documentation/Data/Architecture/AsynchronousCalls.md
@@ -260,6 +260,23 @@ The legacy `workspace.UpdateMeshNode(update)` extension is `[Obsolete]` and poin
 - `meshService.MoveNode(...)` / `meshService.CopyNode(...)` — cold; subscribe.
 - `remoteStream.Update(current => updated, ex => …)` — the `ex` callback fires on the stream's hub; the returned `void` IS the subscription.
 
+### …but a subscription that stays PENDING needs an owner
+
+Discarding the `IDisposable` from `stream.Update(...).Subscribe(...)` is fine — those observables complete promptly, and the handle of a *completed* sequence roots nothing (Rx detaches its observer on completion). A **timer** subscription is the opposite case: `Observable.Timer` / `Observable.Interval` / `Task.Delay` park an entry on the process-wide `TimerQueue`, which is a **strong GC root**, so while the timer is pending it holds the tick closure and everything that closure captured — past its owner's disposal.
+
+```csharp
+// ❌ The timer's IDisposable goes nowhere — nothing can cancel a flush still pending
+//    when the hub tears down, so the closure roots the hub for the whole delay.
+Observable.Timer(TimeSpan.FromMilliseconds(100)).Subscribe(_ => Publish());
+
+// ✅ The pending timer is the HUB's: teardown disposes the composite, cancelling it.
+pendingFlush.Disposable = Observable.Timer(TimeSpan.FromMilliseconds(100))
+    .Subscribe(_ => Publish());          // pendingFlush: a SerialDisposable passed to
+                                         // hub.RegisterForDisposal(...) at construction
+```
+
+Holding it in a field is **not** the same as owning it — the field has to be disposed by the owner's teardown, on every path. Full treatment, with both leak shapes, the measured truth table of which primitives actually root, and why there is deliberately no analyzer: [Subscription Ownership](/Doc/Architecture/SubscriptionOwnership).
+
 ---
 
 ## 🚨 `MeshNode.Content` is always typed at the `GetMeshNodeStream` boundary
--- a/src/MeshWeaver.Documentation/Data/Architecture/DebuggingDisposalAndLeaks.md
+++ b/src/MeshWeaver.Documentation/Data/Architecture/DebuggingDisposalAndLeaks.md
@@ -129,6 +129,17 @@ GC-handle** root (a true leak that accumulates) from a **stack** root — a disp
 continuation the snapshot froze mid-flight, which clears on resume. Assert on the
 former; tolerate the latter.
 
+🚨 **A PASS HERE PINS NOTHING.** This probe *samples* (a root live only for a bounded
+window — 1 s, 100 ms — is caught only if the forced GC lands inside it), it *cannot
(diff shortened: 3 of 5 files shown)


rbuergi:
## CI attribution — both shard-4 failures are inherited from main, neither is in this diff

This branch is docs-only (four markdown files) and CI is red. Checked before assuming, per AGENTS.md → "if CI fails on something your diff does not touch, merge main before you investigate":

**1. `WhatsNewEntryIntegrityTest.EveryEntry_CarriesTheFrontmatterTheFeedRendersFrom`** — fails on `2026-08-09-check-a-deployment-is-current.md`, which carries `Category: What's New` (must be `Feature|Fix`) and no `Order`. That file is on `origin/main`, added by #1006; the test that rejects it landed on main just before it, in #1003. Two PRs raced, so **main itself is red**: run `31314807609` on `fd277b36f` — the exact commit this branch merged — fails on the same single test. `#1017` and `#1018` are already open to repair it, so this one is not mine to fix and is deliberately left alone.

**2. `MeshWeaver.AI.Test.MeshNodeAgentFileStoreTest.Search_StaysLive_AndReEmitsWhenMatchingContentAppears`** — `The operation has timed out.` Unrelated to a markdown diff; not re-run, per the no-retry rule.

Everything this PR touches is green locally:

- `dotnet build src/MeshWeaver.Documentation -c Release -warnaserror` → **0 warnings, 0 errors**
- `DocumentationLinkIntegrityTest` → **passed** (the test that guards every link on the new page)
- `Build solution (once)` and shards 0, 1, 2, 3, 5 → **pass** on this branch

This needs a re-merge of main plus a fresh run once the What's New frontmatter fix is in. Not merging until the suite is green on its own merits.

PR #1017, opened 2026-08-09
fix(docs): repair the What's New frontmatter that turned main red

rbuergi:
🔴 **main is currently red** on `fd277b36f` (run 31314807609), and every open PR inherits the failure.

```
WhatsNewEntryIntegrityTest.EveryEntry_CarriesTheFrontmatterTheFeedRendersFrom

2026-08-09-check-a-deployment-is-current.md: Category is 'What's New',
  must be one of Feature | Fix — the feed renders Features in full and
  bundles Fixes into the day's summary
2026-08-09-check-a-deployment-is-current.md: Order is '(none)',
  must be '-20260809' (negative ship date) — otherwise the doc tree sorts
  it alphabetically by title
```

One entry, shipped with the older frontmatter shape.

```diff
  Name: Checking that a deployment is up to date, from inside or outside
- Category: What's New
+ Category: Feature
  Description: …
  Icon: ArrowSync
+ Order: -20260809
```

Its sibling from the same day, `2026-08-09-export-works-without-being-an-admin.md`, carries both fields — so this one slipped through rather than the rule being new.

**`Feature`, not `Fix`**: the entry describes a new anonymous `/api/version` endpoint plus an About-page capability, not the repair of a defect.

## Verified

`WhatsNewEntryIntegrityTest` passes 1/1 against the corrected tree (Release, native xUnit host).

## Why this is worth its own PR

A red main blocks the pull-based self-update the deployments roll forward on (AGENTS.md → the merge gate), and it is reddening unrelated work: **#1013 is a NuGet-cache change that cannot possibly affect frontmatter**, and it failed on exactly this and nothing else. Anyone diagnosing their own PR right now is chasing someone else's breakage.

## No What's New entry for this PR

It repairs a What's New entry's metadata; it is not itself a user-visible change.


rbuergi:
Flagging a collision: **#1018 was opened 90 seconds after this and fixes the same file** — two sessions reached the same red main independently. Both are correct on the entry itself (`Category: Feature`, `Order: -20260809`); whichever merges second will conflict on it.

**Suggest landing #1018 instead**, because it also repairs what produced the bad entry. The `pullrequest` skill contradicted itself: its procedure block (step 0.5) had been updated to `Category: <Feature|Fix>` with `Order: -${DATE//-/}`, but the prose section further down still read:

> **Front-matter**: `Name`, `Category: What's New`, `Description`, `Icon`

— no `Order` at all, and a `Category` the test rejects. That is the half a reader quotes when hand-writing an entry, and it is how this one came to be written. Fixing only the entry leaves the next author with the same instructions.

Worth recording either way, because it makes this failure mode structural rather than careless: **a What's New entry does not meet `WhatsNewEntryIntegrityTest` until it has already merged.** The PR that adds one is green; `main` goes red on the merge commit. So this always lands on main and never on the PR that caused it — #1003 (the test) and #1006 (the entry) simply raced.

No objection to landing this one first if it goes green sooner — in that case I will reduce #1018 to the `SKILL.md` half so the generator still gets fixed. Closing this in favour of #1018 is also fine. Either way, please do not merge both.

rbuergi:
Closing this as superseded — **#1018 merged first** (`e38ad0c6e`) and carries the identical entry fix (`Category: Feature`, `Order: -20260809`) plus the repair to the `pullrequest` skill that generated the bad frontmatter. Your diagnosis here was right and independently reached; the only reason to prefer #1018 was that it also fixed the generator, so the next author is not handed the same wrong instructions.

Nothing is lost — the entry is repaired on main. Thanks for catching it.

PR #1025, opened 2026-08-09
A failed status write must never block the self-update roll-forward

### LINKED ISSUE #1020: atioz self-update stalled ~37h — a rollout restart masks it, so it will drift again
## What was observed

Deployed portal images at 2026-08-09 13:00Z, with the ACR build time of each tag:

| Portal | Was on | Tag built | Staleness |
|---|---|---|---|
| memex | `ci.2371` | 09-Aug 09:11Z | ~4 h — consistent with the 6 h poll |
| memex-cloud | `ci.2354` | 09-Aug 07:42Z | ~5 h — consistent with the 6 h poll |
| **atioz (prod)** | `ci.2151` | **08-Aug 00:31Z** | **~37 h — roughly 6 missed polls** |

memex and memex-cloud sit exactly where a 6-hour `SelfUpdateOptions.PollInterval` should put them.
atioz did not, and it is the production portal.

## Why this is still open even though atioz now looks healthy

A `kubectl rollout restart` was issued during unrelated work. The self-update poll fires on startup
(`StartWith(-1L)`), so atioz jumped straight to `ci.2417` and now reports healthy — one ready pod,
zero restarts, HTTP 200.

**That masked the fault rather than fixing it.** Nothing was diagnosed, so the poller is expected to
stall again and the portal to drift silently. The restart also destroyed the in-pod log evidence
from the stalled period.

## Where to look

- **`[SelfUpdate]` log lines on the atioz pod.** memex is known to log at `Warning`, which hides the
  Information-level `[SelfUpdate]` lines; check whether atioz does the same before concluding from
  their absence. (Do not change log levels in code to investigate — see AGENTS.md.)
- **The federated credential** for subject `system:serviceaccount:atioz:memex-portal-sa` on UAMI
  `memexaks-portal-mi`, and its **AcrPull** grant on the `meshweaver` registry. A silently-expired or
  mismatched-subject credential makes every ACR poll fail without crashing the portal.
- **`VersionSelect.PickTarget`** — it has previously mis-selected tags (per-RID `-linux-x64` /
  `-linux-arm64` suffixes sorting above the manifest list, fixed in #120). Worth confirming it is not
  rejecting or mis-ranking the newer `3.0.0-rc1.ci.N` tags.

## Trap to record

`kubectl rollout restart` on a portal is **not** a neutral "restart the same version" — the startup
poll pulls it forward to the newest image. Capture `[SelfUpdate]` logs **before** restarting a portal
suspected of a stalled poller, or the evidence goes with the pod.

rbuergi:
Fixes #1020

## The poll never stopped — every tick failed on a status write

The restart masked the fault, but Loki kept the stalled pod's logs. `memex-portal-deployment-7d477dd8b6-7xzcq` (the `ci.2151` pod) shows the poller ticking on schedule for its whole 37 h life, and **failing every single time**:

```
t0        [SelfUpdate] starting; version=3.0.0-rc1.ci.2151 … canPatch=True, interval=06:00:00.
t0+60s    [SelfUpdate] policy stream faulted; re-establishing in 06:00:00.
t0+6h00m30s   [SelfUpdate] check failed (policy=Continuous).
t0+12h00m30s  [SelfUpdate] check failed (policy=Continuous).
…every 6 h, six times, until the restart
```

The `30 s` offset from each tick is the tell. Unfiltered, the exception is:

```
warn: MeshWeaver.Mesh.MeshNodeStreamHandle[0]
      [UpdateRemote] ERROR hub=cache/hp4PbexBOkqjtxhnolk9aQ target=Admin/UpdatePolicy type=TimeoutException
warn: MeshWeaver.Hosting.MeshNodeStreamCache[0]
      [UpdateQueue] FAILED path=Admin/UpdatePolicy seq=5 elapsedMs=30005.2
      System.TimeoutException: Update aborted: no initial state arrived for 'Admin/UpdatePolicy' within 30s.
warn: Memex.Portal.Shared.SelfUpdate.SelfUpdateHostedService[0]
      [SelfUpdate] check failed (policy=Continuous).
```

So everything the issue suspected was **fine**: the federated credential worked, ACR answered, and `VersionSelect.PickTarget` picked the right newer tag every time. The tick got all the way past the `IsNewer` gate and then died writing two cosmetic fields.

`RunOnce` chained the k8s PATCH *after* `RecordAvailable` with `.SelectMany`. `RecordAvailable` writes `LatestAvailableTag` + `CheckedAt` to `Admin/UpdatePolicy` — bookkeeping for the admin tab. **A status field held the deploy hostage.**

That inverts the invariant #611 established for the policy *read*: neither mesh touch may gate the roll-forward, because a fresh image is exactly what recovers a degraded pod. The read was decoupled then; the write was left on the critical path.

## Why only atioz

`FAILED path=Admin/UpdatePolicy` appears in **no other namespace** all week — only atioz, exactly six times, one per stalled tick. Its `Admin/UpdatePolicy` node hub was unreachable because the silo's Orleans routing was wedged (`RoutingGrain`, `[StatelessWorker(1)]`, one `RouteMessage` executing for `06:00:22` with `NonReentrancyQueueSize=541`). **That wedge is a separate, deeper defect and is not fixed here** — see the note below.

## The fix

`RecordAvailable`'s failure is surfaced by name (tag + node) and no longer aborts the tick; `Apply` runs regardless, over `Concat` so a successful record still lands first. Nothing is swallowed — the warning carries the exception.

`Apply` also announces the tag **before** the attempt. Previously the only trace a stalled install left was `[SelfUpdate] check failed`, which reads like a registry problem while the registry was fine — that is why this took 37 h and a manual version comparison to spot.

No watchdog, no retry, no shortened interval, no restart job. The poller's cadence is untouched.

## Proof

`Failed_availability_write_never_blocks_the_roll_forward` injects the prod fault at `RecordAvailable` (now a fault-injection seam alongside `ReadPolicyStream`) with the exact prod exception text, against a real monolith mesh.

Reverting only the coupling — `.Catch(…).Concat(Apply(…))` back to `.SelectMany(_ => Apply(…))` — reproduces atioz byte-for-byte and the test fails:

```
[Warning] [SelfUpdate] check failed (policy=Continuous).
Exception: System.TimeoutException: Update aborted: no initial state arrived for 'Admin/UpdatePolicy' within 30s.
…× every tick, no patch ever applied…
=== TEST FAILED: The operation has timed out.
Failed! - Failed: 1, Passed: 0
```

With the fix, all 3 tests in the class pass in 4 s. `Memex.Portal.Shared` and `MeshWeaver.Hosting.Monolith.Test` build clean at `-c Release -warnaserror`.

## What's New

`Category: Fix` — an operator can notice this: a portal that silently stops taking updates.

## What actually ships this

Merging this PR is not what deploys it. `main-cd.yml` gates every image job on
`workflow_run.event == 'push'`, so the image that carries this fix is published by the **real push to
main that this merge creates** — and only then. A `workflow_dispatch` re-run of Build and Test would
turn main green while publishing nothing.

So after merge, verify the IMAGE, never the green tick:

```bash
az acr repository show-tags -n meshweaver --repository memex-portal-ai --orderby time_desc --top 5 -o tsv
```

The three portals then pick it up on their own 6 h poll — which, with this change, no longer depends
on `Admin/UpdatePolicy` being writable.

## ⚠️ Still open — the underlying wedge

atioz's routing wedge is untouched by this PR. It was confined to that one pod: lines containing
`NonReentrancyQueueSize` came **only** from `memex-portal-deployment-7d477dd8b6-7xzcq` (~8.4 M on
08-Aug, ~6.0 M on 09-Aug up to the restart) and the current pod has emitted **zero** in the five
hours since. So the restart did clear it and nothing is bleeding right now — but nothing explains
what started it either, so it can recur.

`RoutingGrain` is `[StatelessWorker(1)]` and non-reentrant, so a single `RouteMessage` whose
memory-stream `OnNextAsync` never completes wedges the whole silo's routing, with no timeout on that
path. Worth its own issue.

One upside of this PR: when it does recur, the install now rolls forward on the next 6 h tick, and
rolling forward restarts the pod — so the wedge self-clears instead of needing a human to notice.

### DIFF
--- a/memex/Memex.Portal.Shared/SelfUpdate/SelfUpdateHostedService.cs
+++ b/memex/Memex.Portal.Shared/SelfUpdate/SelfUpdateHostedService.cs
@@ -19,7 +19,14 @@ namespace Memex.Portal.Shared.SelfUpdate;
 /// its own portal + migration Deployments to the new version so k8s rolls them. Outside Kubernetes it
 /// records the available version for detect-and-notify. Mirrors <c>ShippedReleaseSeedHostedService</c>
 /// (raw <see cref="IHostedService"/>, <c>SubscribeOn(TaskPoolScheduler.Default)</c>, one subscription).
-/// Not sealed: <see cref="ReadPolicyStream"/> is the fault-injection seam for the resilience test.
+/// Not sealed: <see cref="ReadPolicyStream"/> and <see cref="RecordAvailable"/> are the
+/// fault-injection seams for the resilience test — they are the only two mesh touches this poller
+/// makes, and therefore the only two ways a degraded mesh could ever stop an install rolling forward.
+///
+/// <para>🔁 wedges-to-zero, the standing invariant here: NEITHER mesh touch may gate the roll-forward.
+/// The registry poll and the k8s PATCH speak only to ACR and the API server, so they keep working
+/// while the mesh is degraded — and a fresh image is precisely what recovers a degraded pod. The
+/// policy READ was decoupled in #611; the availability WRITE in #1020.</para>
 /// </summary>
 public class SelfUpdateHostedService : IHostedService
 {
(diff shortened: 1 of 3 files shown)


rbuergi:
## CI: two runs, two different unrelated flakes — not this diff

**Run 1** (`31322505388`, before syncing main) — shard 2:
`MeshWeaver.FutuRe.Test.FutuReAnalysisTest.EuropeRe_LineOfBusiness_Overview_ShouldRender`, a leaked
`CreateNodeRequest` callback (`REGISTERED_AFTER_POST`).

Merged `origin/main` (#1024 had landed) per AGENTS.md — *"if CI fails on something your diff does not
touch, merge main before you investigate anything."*

**Run 2** (`31323317051`, after the merge) — shard 2 now **passes**; shard 0 fails instead:
`MeshWeaver.Hosting.Orleans.Test.OrleansStaticRepoImportTest.Reimport_ChangedContent_RepairsContentNullPage`,
`TaskCanceledException` (test-level timeout).

Different test, different project, different shard each run — the ambient flake cluster, and `main`
is showing it independently: **3 of its last 12 `Build and Test` runs are red** (`0758b90c`,
`043c0d33`, `3345b501` today).

Why neither can be this diff: the branch touches exactly three files —
`Memex.Portal.Shared/SelfUpdate/SelfUpdateHostedService.cs`,
`test/MeshWeaver.Hosting.Monolith.Test/SelfUpdatePollerResilienceTest.cs`, and one What's New
markdown. Neither `MeshWeaver.FutuRe.Test` nor `MeshWeaver.Hosting.Orleans.Test` references
`Memex.Portal.Shared` or `MeshWeaver.Hosting.Monolith.Test`.

`Build solution (once)` is green on both runs, and `MeshWeaver.Hosting.Monolith.Test` — the project
carrying the new test — passed in both.

**Not merging.** The gate is green CI; these need to be either waited out or fixed at their own root
cause, which is a separate race from #1020.

Issue #990, opened 2026-08-09
Deck PDF export is content-faithful, not pixel-faithful — gradient/HTML-authored slides need headless Chromium


rbuergi:
## Summary

Split out of **#423**, whose access defect is fixed and merged (#986). Deck → PDF export works and is content-faithful; what it is not is **pixel-faithful**.

The QuestPDF renderer (`Pdf/PdfDocumentRenderer.cs`) reconstructs slide content into a PDF document model. That is deliberate and it is why the feature works offline, quickly, and without a browser in the image. But it means anything whose fidelity depends on **browser rendering** does not survive:

- gradient and image backgrounds
- slides authored as raw HTML
- CSS-driven layout, custom fonts, transforms

## Why this was left out of #423

Closing the gap needs **headless Chromium in the portal image** — a materially different change: image size, startup cost, a sandbox/security surface, and a new failure mode when the browser is missing or crashes. That is a product decision about what the export is *for*, not a bug fix, so folding it into an access-control fix would have been wrong.

Recorded here rather than left implicit in a closed issue.

## What to decide before building anything

1. **Is content-faithful actually insufficient?** For most decks the QuestPDF output is the better artifact — smaller, faster, text-selectable, no browser. Pixel fidelity matters for a specific class of deck (design-led, HTML-authored). Worth confirming that class is real and worth the cost before adding Chromium.
2. **If yes: replace or augment?** A second renderer alongside the existing one (chosen per deck, or per export) keeps the fast path intact and confines the browser dependency to the decks that need it. Replacing outright would make every export pay for Chromium.
3. **Where does it run?** In the portal image, or a separate rendering service. The portal is the shared `memex` deployment — adding a browser there affects every tenant.

## Note on scope

The existing export paths (`DeckExportMenuProvider`, `ExportPdf.csx` emitting one page per slide, `SendDocumentDispatch` for send-to-contacts) all work and are covered by e2e tests. This is purely about rendering fidelity — nothing about the plumbing needs revisiting.

PR #1027, opened 2026-08-09
Every stored image in a pixel-faithful deck export was printing blank

rbuergi:
Refs #990

## The gap PR #1009 flagged, and what was hiding in it

#1009 shipped pixel-faithful deck export and closed with one open item, in its own words:

> **Asset inlining has no end-to-end test against a real content collection.** … "a slide with a stored image exports with that image embedded" is unverified. This matters more under the new CSP: a reference that fails to resolve is now a **missing image** rather than a silently-remote one.

I wrote that test. It failed. **Every stored slide image printed blank**, and had done since the feature merged.

## The defect

The markdown pipeline rewrites a slide's `logo` onto `api/content/{nodePath}/{file}` — the same href the live view renders (`MarkdownView` passes `Stream.Owner`, the composer passes `slide.Path`; they agree by construction). The **first segment of that href is the node's partition**.

`ParseAssetReference` read the first segment as the **collection name**. So for a slide at `Space123/pitch/intro` the export asked `IContentService.GetContent("Space123", "pitch/intro/logo.png")`. There is no collection named `Space123`; `GetContent` throws for an unknown collection; the export template caught it, logged a warning and moved on.

Before #1009 that produced a slow-but-correct PDF, because the un-inlined `api/content/…` reference was still fetchable. Under the new CSP (`default-src 'none'; img-src data:`) plus the no-DNS process flags, it cannot be fetched. So the visible result changed from "slow" to **blank**, which is exactly the risk #1009 named.

The convention itself is not wrong, it is *the wrong reader for this shape*: first-segment-names-the-collection is right for an authored `content:{collection}/{path}` reference (what `BrandingResolver` reads) and wrong for every URL the product renders. The two readings agree on the file path and disagree on the collection — which is why it fails as "collection not found" rather than as anything that looks like a bug.

## The fix: one server-side reading of a content reference, not two

`ContentFileResolver` (new, in `MeshWeaver.ContentCollections`) holds the algorithm the HTTP content route already implemented privately: resolve the owning node with `IPathResolver` (longest real-node prefix), ask **that node's hub** for its collections via the `[RequiresPermission(Read)]`-carrying `GetDataRequest`, then take the named collection when the first remaining segment is one and the node's default `content` collection otherwise, putting the node's own path back in front of the file when the collection is inherited (`CombineOwnerRelative`).

- **`BlazorHostingExtensions.ResolveContentFile` now delegates to it.** Same permission hop, same `IsStatic` mount check, same 404 texts — the logic moved, it did not change.
- **The export uses it too**, through a new `SlideAssetInliner`, which also lifts the inlining loop out of `ExportPdf.csx` — where it was untestable, which is *why* the gap existed.
- `ParseAssetReference` is replaced by **`ToContentReference`**: strips the route prefix and percent-decodes per segment, and deliberately does **not** split off a collection. Which segments name the node cannot be read off the string, and pretending otherwise is the whole bug. (It also no longer folds the `/content/` segment out — that fold is correct for an authored `content:` reference and would, on a route URL, let a folder named like a collection win over the collection.)

## Unresolved is now a result, not a log line

`SlideAssetInlining` returns `Inlined` and `Unresolved` (each with a reason), and the template logs one warning per unresolved reference **saying it will print as a blank image**. The old code had a bare `continue` for an unparseable reference — no log at all — and a `LogWarning` for the rest, with the export still reporting success. A user got a PDF with a hole in it and, at best, a line buried in the activity log.

## Tests — and proof they bite

`DeckAssetInliningTest` (`MeshWeaver.Hosting.Monolith.Test`, `MonolithMeshTestBase`, no mocks) runs a real mesh with real FileSystem content collections and real stored PNGs:

1. **Happy path** — a stored image is embedded: the document carries `data:image/png;base64,{its own bytes}` and no `api/content/` reference survives.
2. **Unresolved is reported** — a reference to a file that isn't there comes back in `Unresolved` with a reason, and is left in the document rather than dropped.
3. **Every reference shape**, each against a real collection: node-relative `{node}/{file}`, explicit default `{node}/content/{file}`, a named collection `{node}/attachments/{file}`, a `~`-encoded qualified name `{node}/Media~2026/{file}` → collection `Media/2026`, and percent-escaped segments `{node}/a%20folder/logo%20two.png`.
4. **The whole path through the mesh** — a real `ExportDocumentRequest` with `Fidelity = Pixel`. Never skipped: it asks the renderer's own probe and asserts whichever contract the machine supports — with a browser, a PDF whose page carries an **embedded image** (PdfPig `GetImages()`) and an activity log saying `Inlined 1/1`; without one, the actionable refusal. Verified locally against Chrome 151: `pages=1 images=1`.

**The suite was checked against the defect.** Reinstating #1009's first-segment split fails 3 of the 4, and the export test's failure carries the cause verbatim:

```
[Warning] Slide asset api/content/Space…/pitch/intro/logo.png could not be inlined
          (Collection 'Space…' not found); it will print as a blank image
```

The fourth (unresolved-is-reported) passes either way, correctly: an unresolvable reference is unresolvable under both readings.

Test writing also had to establish where a node's file actually lives, so the fixture stores every asset **through the product's own resolution** rather than by guessing a disk layout — a test that agreed with itself about the layout would have proved nothing.

## Not test-only, so there is a What's New entry

The task was scoped as verification, and I would have said "test-only, no What's New". It isn't: this repairs a user-visible defect in a feature that shipped today, so `2026-08-09-pictures-in-a-printed-deck-are-actually-there.md` (`Category: Fix`) is included.

## 🔒 A hole this PR opened, found in review and closed

Copilot caught it and it was real. The content route validates the **decoded** path with `StaticAssetMount.IsSafeRelativePath` before resolving anything; my extraction left that check in the route, so the route stayed safe — but the **export** path had none, and its references come straight out of user-authored slide markup with raw-HTML passthrough. `FileSystemStreamProvider` resolves a collection-relative path with a bare `Path.Combine`, so `%2E%2E` — which survives URL normalisation and only becomes `..` once decoded per segment — walked out of the collection's BasePath into another node's files, under the portal's own filesystem access. Exactly the server-side-fetch hazard #1009 spent a review round closing, re-opened through a different door.

The guard now lives in `ContentFileResolver.Resolve`, so every caller inherits it rather than each remembering it — which is the entire point of there being one resolver. The route keeps its own check in front of the call as well.

**Proven, not assumed.** `A_traversal_reference_is_refused_rather_than_read` stores a PNG in a *different node's* collection and has a slide reach two levels up for it. With the guard disabled the traversal succeeds and the file is inlined (`Inlined` = 1); with it, the reference is refused with `Invalid content path` and no byte reaches the document.

Copilot's other comment was also correct: the doc claim that per-segment decoding "cannot introduce or remove a `/`" is false — `Uri.UnescapeDataString("%2F")` is `/`. The comment (inherited wording) now says so and points at the guard that makes it safe, and `SlidePrintComposerTests` pins that behaviour instead of the false claim.

## One observation I did not act on

Node hubs are hosted flat under the mesh hub, so `ExposeInChildren` means *hub* children, not *path* children. `MemexConfiguration` mounts the `content` collection only on partition roots (`!nodePath.Contains('/')`), so a **child** node's hub reports no `content` collection at all and `/api/content/{Space}/{child}/{file}` finds nothing to serve — independently of anything in this PR, and equally in the live view. `CombineOwnerRelative` exists for exactly the inherited case that this mount shape never produces. The tests here mount `content` per node, which is the shape the route demonstrably serves. Worth a look, but it is a separate question from this fix and I did not want to change a content-mount policy inside a bug fix.

## Verification

Release `-warnaserror` builds clean for `MeshWeaver.ContentCollections`, `MeshWeaver.Markdown.Export` and `MeshWeaver.Hosting.Blazor`. Locally green:

| Suite | Result |
|---|---|
| `DeckAssetInliningTest` | 5/5 (real Chrome: `pages=1 images=1`) |
| `MeshWeaver.Markdown.Export.Test` | 52/52 |
| `MeshWeaver.Hosting.Blazor.Test` | 263/263 |
| `MeshWeaver.Persistence.Test` | 503/503 |
| `MeshWeaver.Content.Test` | 257/258 (1 pre-existing skip) |
| `MeshWeaver.ContentCollections.Test` | 29/29 |
| `DeckExportScriptRelayTest` + `DeckPixelExportScriptRelayTest` | 6/6 |

Plus the two deliberate-breakage checks: 3/4 red against the reinstated collection split, and the traversal test red with the guard disabled.

**Not merged** — reporting back for your call.

### DIFF
--- a/src/MeshWeaver.ContentCollections/ContentFileResolver.cs
+++ b/src/MeshWeaver.ContentCollections/ContentFileResolver.cs
@@ -95,6 +95,19 @@ public static class ContentFileResolver
         string reference,
         AccessContext? caller = null)
     {
+        // 🚨 TRAVERSAL GUARD, on the DECODED reference, before anything else. `%2E%2E` and `%2F`
+        // survive URL normalisation and only become `..` and `/` once a caller has decoded per
+        // segment — and FileSystemStreamProvider resolves a collection-relative path with a bare
+        // Path.Combine, so an un-guarded `..` reads outside the collection's BasePath, i.e. another
+        // partition's files. It lives HERE rather than in each caller because the export's
+        // references come out of user-authored slide markup with raw-HTML passthrough: a slide could
+        // otherwise point the server at anything the portal can read. (The content route also checks
+        // this before calling us; belt and braces on a path where the cost of being wrong is a file
+        // disclosure.)
+        if (!StaticAssetMount.IsSafeRelativePath(reference))
+            return Observable.Return(
+                ContentFileResolutionResult.NotFound("Invalid content path"));
+
         // Cold: nothing is resolved and nothing is posted until somebody subscribes.
         return Observable.Defer(() => hub.ServiceProvider
             .GetRequiredService<IPathResolver>()
--- a/src/MeshWeaver.ContentCollections/ContentFileResolver.cs
+++ b/src/MeshWeaver.ContentCollections/ContentFileResolver.cs
@@ -1,0 +1,236 @@
+using System.Reactive.Linq;
+using System.Text.Json;
+using MeshWeaver.Data;
+using MeshWeaver.Mesh.Services;
+using MeshWeaver.Messaging;
+using Microsoft.Extensions.DependencyInjection;
+
+namespace MeshWeaver.ContentCollections;
+
+/// <summary>
+/// One <c>/api/content/…</c> reference, resolved to the collection and the collection-relative file
+/// path a server-side read needs.
+/// </summary>
+/// <param name="Collection">
+/// The owning node's collection config, exactly as that node reported it — <b>unrenamed</b>, so a
+/// caller can still read <see cref="ContentCollectionConfig.IsStatic"/> and report the collection's
+/// real name.
+/// </param>
+/// <param name="QualifiedName">
+/// <c>{nodePath}/{collectionName}</c> — the name the config must be registered under before it is
+/// read, so two nodes inheriting the same ancestor collection cannot collide in one content
+/// service's cache.
+/// </param>
+/// <param name="FilePath">The path of the file <b>within that collection</b>.</param>
+/// <param name="Owner">The node the reference resolved to.</param>
+public sealed record ContentFileResolution(
+    ContentCollectionConfig Collection,
+    string QualifiedName,
+    string FilePath,
+    Address Owner)
+{
+    /// <summary>The config renamed to <see cref="QualifiedName"/> and attributed to <see cref="Owner"/>.</summary>
+    public ContentCollectionConfig QualifiedConfig =>
+        Collection with { Name = QualifiedName, Address = Owner };
+}
+
+/// <summary>
+/// Outcome of <see cref="ContentFileResolver.Resolve"/>: either a resolution or the reason there
+/// isn't one. The reason is a plain sentence a caller can put in a 404 or a log line.
+/// </summary>
+/// <param name="Resolution">The resolution, or <c>null</c>.</param>
+/// <param name="Reason">Why there is none, or <c>null</c> when there is one.</param>
+public sealed record ContentFileResolutionResult(ContentFileResolution? Resolution, string? Reason)
+{
+    /// <summary>An unresolvable reference, with the reason.</summary>
+    public static ContentFileResolutionResult NotFound(string reason) => new(null, reason);
+
+    /// <summary>A resolved reference.</summary>
+    public static ContentFileResolutionResult Found(ContentFileResolution resolution) =>
+        new(resolution, null);
+}
+
+/// <summary>
+/// 🚨 THE ONE server-side reading of a content reference: <c>{node}/{collection}/{file}</c> or
+/// <c>{node}/{file…}</c> → the collection to open and the path inside it.
+///
+/// <para><b>Two shapes, one rule.</b> The owning node is resolved with
+/// <see cref="IPathResolver"/> (longest real-node prefix). When the first remaining segment names a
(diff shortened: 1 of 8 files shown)


rbuergi:
Both review comments were right, and the second one was a real hole I opened — thank you.

**Traversal (the important one).** The content route validates the *decoded* path with `StaticAssetMount.IsSafeRelativePath` before it resolves anything; my extraction left that check in the route, so the route stayed safe — but the **export** path had none, and its references come straight out of user-authored slide markup with raw-HTML passthrough. `FileSystemStreamProvider` resolves a collection-relative path with a bare `Path.Combine`, so `%2E%2E` (which survives URL normalisation and only becomes `..` once decoded per segment) walked out of the collection's BasePath, under the portal's own filesystem access.

The guard now lives in `ContentFileResolver.Resolve`, so every caller inherits it rather than each remembering it — which is the whole point of there being one resolver. The route keeps its own check in front of the call as well.

Not taken on faith: `A_traversal_reference_is_refused_rather_than_read` stores a PNG in a **different node's** collection and has a slide reach two levels up for it. With the guard disabled the traversal succeeds and the file is inlined (`Inlined` = 1); with it, the reference is refused with `Invalid content path` and no byte reaches the document.

**The doc claim.** Correct — `Uri.UnescapeDataString("%2F")` is `/`, so per-segment decoding *can* split a segment. The comment asserted the opposite (inherited wording). It now says decoding can introduce a separator and points at the guard that makes it safe, and `SlidePrintComposerTests` pins that behaviour instead of the false claim.

Fixed in 0b8fbeecd.

PR #1033, opened 2026-08-09
perf(test): parallelise Hosting.Monolith.Test — 314.7s → 122.5s

rbuergi:
The heaviest project in the suite, and the one that sets the shard floor — `shard-assign.sh`: *"The floor for ANY sharding is the heaviest single project."*

```
threads=1   512 tests, 0 failed, 314.7s
threads=4   512 tests, 0 failed, 122.5s     2.57x
```

Held back in #979 because `threads=4` red two tests. Both are now resolved, neither by suppression.

## 1. The durable-seed failure is gone

`StaleActivationDurableFirstSeedTest` was the blocker in #979. #980 closed it as **already fixed** by the store-level write CAS (#971 / PR #988): the read-back of a value *below* the one written was never a visibility lag, it was a concurrent **rollback**, which the CAS now refuses at the store. Verified — it does not reproduce at `threads=4` on current main.

## 2. The GC-reachability family is scoped — and this is the part worth reading

Four classes assert **process-wide** reachability ("after disposing X, nothing still roots it") using `WeakReference` + a forced full `GC.Collect`:

- `MeshHubDisposalLeakTest`
- `NodeTypeAssemblyLeakTest`
- `KernelScriptMemoryLeakTest`
- `ConsoleCaptureExecutionContextLeakTest`

That assertion is only meaningful in a **quiet process**. With neighbours concurrently allocating and holding live object graphs, a forced collect cannot be relied on to reclaim the target — so the test reports a "leak" that is really a busy heap.

**Under parallelism such a test is not flaky, it is wrong**: its precondition no longer holds. Declaring that precondition with `[CollectionDefinition(DisableParallelization = true)]` *is* the fix. A retry, a widened timeout, or a `[Skip]` would have been the band-aid.

All four are tagged, not just the one that failed locally — `NodeTypeAssemblyLeakTest` red here, while `MeshHubDisposalLeakTest` red identically on CI run 31300348689 this morning and passed this run purely on timing. Same defect class; scoping one and waiting for the others to bite would be worse than useless.

🚨 The collection is documented as being for **reachability assertions only**. A test that is merely slow, or that races on shared *state*, must not be added — that is a real defect, and serialising it would hide it.

Everything else in the assembly still runs at `maxParallelThreads: 4`.

## Why this matters beyond one project

Total measured test work is 36.7 min across 8,817 tests. Monolith is the single largest block and the floor under any sharding change — cutting it 2.57x is what makes reducing shard count viable later (ideal per-shard is 5.4 min at 6 shards today, 8.1 at 4; that arithmetic only improves once the 36.7 min shrinks).

## No What's New entry

### DIFF
--- a/test/MeshWeaver.Hosting.Monolith.Test/ConsoleCaptureExecutionContextLeakTest.cs
+++ b/test/MeshWeaver.Hosting.Monolith.Test/ConsoleCaptureExecutionContextLeakTest.cs
@@ -30,6 +30,7 @@ namespace MeshWeaver.Hosting.Monolith.Test;
 /// scope, keeps the timer alive, and asserts the capture target is collectible
 /// after the scope is disposed.</para>
 /// </summary>
+[Collection(GcReachabilityCollection.Name)]
 public class ConsoleCaptureExecutionContextLeakTest
 {
     [MethodImpl(MethodImplOptions.NoInlining)]
--- a/test/MeshWeaver.Hosting.Monolith.Test/GcReachabilityCollection.cs
+++ b/test/MeshWeaver.Hosting.Monolith.Test/GcReachabilityCollection.cs
@@ -1,0 +1,36 @@
+using Xunit;
+
+namespace MeshWeaver.Hosting.Monolith.Test;
+
+/// <summary>
+/// The collection for tests that assert PROCESS-WIDE GC reachability — "after disposing X,
+/// nothing still roots it" — via <c>WeakReference</c> + a forced full <c>GC.Collect</c>.
+///
+/// <para>🚨 <b>This is scoping, not a band-aid.</b> Those assertions are statements about the
+/// whole process, so they are only meaningful when the rest of the process is quiet. With other
+/// classes concurrently allocating and holding live object graphs, a forced collect cannot be
+/// relied on to reclaim the target, and the test reports a "leak" that is really just a busy heap.
+/// Under parallelism such a test is not flaky — it is WRONG, because its precondition no longer
+/// holds. Declaring that precondition is the fix; suppressing, retrying or widening a timeout
+/// would be the band-aid.</para>
+///
+/// <para><c>DisableParallelization</c> makes xUnit run this collection on its own rather than
+/// alongside the parallel ones, which restores the quiet process these tests require. Everything
+/// else in the assembly keeps running at <c>maxParallelThreads</c> (see
+/// <c>test/MeshWeaver.Hosting.Monolith.Test/xunit.runner.json</c>).</para>
+///
+/// <para>Measured: enabling parallelism on this assembly without this collection reds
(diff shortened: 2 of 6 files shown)


rbuergi:
Closing — **not ready.** My GC scoping was necessary but not sufficient, and CI found two more parallelism-caused failures that my local run did not.

## What CI red on (run 31327487555)

```
PixelFidelityOfferedOnlyWhenAvailableTest.DeckExportDialog_OffersPixelFidelity_ExactlyWhenABrowserResolves
StoreLevelWriteConflictTest.AConflictIsMerged_NotRefused_AndTheDroppedValueReachesTheActivityLog
```

Neither is a GC-reachability test, so neither is covered by the `GcReachabilityCollection` this PR adds. Both are **caused by this change**: checked the last four red runs on `main` and neither test appears — main's failures are all Orleans plus one `AI.Test`.

## Why my local run missed them

Local: 512 tests, 0 failed, 122.5s at `threads=4` on an **18-core** machine. CI runners have **4 vCPU**, so at the same `maxParallelThreads: 4` the contention is completely different — four threads genuinely competing for four cores rather than spread across eighteen. Any race whose window is proportional to CPU pressure shows up there and not here.

That is a reusable lesson for this repo: **a local parallel green on a many-core dev machine is not evidence for a 4-vCPU runner.** It is the same mistake shape as trusting a macOS green for `MeshHubDisposalLeakTest`, whose ClrMD analysis cannot even run on macOS.

## The two failures are worth someone's attention, separately

- `StoreLevelWriteConflictTest.AConflictIsMerged_NotRefused_AndTheDroppedValueReachesTheActivityLog` — asserts conflict-merge semantics. Parallelism creates genuinely concurrent writers, which is exactly what that code path is for, so this may be a **real defect** the serial suite never exercised rather than a test artefact. It sits on the same store-level write path as the CAS from #971 / #988.
- `PixelFidelityOfferedOnlyWhenAvailableTest…ExactlyWhenABrowserResolves` — turns on whether a browser resolves; under CPU pressure that resolution likely times out. More plausibly a test-side assumption, but it needs looking at rather than assuming.

## What survives

The measurement stands and is worth keeping: **314.7s → 122.5s (2.57x)** on the heaviest project in the suite, and the two blockers from #979 are now one — the durable-seed failure is genuinely gone, fixed by the write CAS.

`GcReachabilityCollection` is also correct as written and will be needed by whoever finishes this: four classes assert process-wide reachability with `WeakReference` + forced `GC.Collect`, and under parallelism such a test is not flaky but **wrong**, because its precondition (a quiet heap) no longer holds.

The branch `perf/share-mesh-cluster` is left in place. Finishing this means resolving the two above — on CI, not locally.

## Meanwhile

#1035 parallelises `AI.Test` (198.2s → 93.7s, 2.11x) and is independent of all of this — no GC assertions, no shared Testcontainers, clean at 4 threads first try.

PR #1035, opened 2026-08-09
perf(test): parallelise AI.Test — 198.2s → 93.7s

rbuergi:
Second-heaviest project after Monolith — `shard-assign.sh` weight 177s, 1,138 tests, 39 classes carrying a `ShareMeshAcrossTests` opt-in.

```
threads=1   1138 tests, 0 failed, 198.2s
threads=4   1138 tests, 0 failed,  93.7s     2.11x
```

**One config file, nothing else.** Unlike `Hosting.Monolith.Test` (#1033) this needed no scoping:

- no GC-reachability assertions — no `WeakReference` + forced `GC.Collect`, so nothing here has a precondition that a busy heap invalidates
- no shared Testcontainers instance of the kind that keeps `Hosting.PostgreSql.Test` serial (25 classes on one `[Collection("PostgreSql")]`)

Ran clean at 4 threads first try.

## Correcting an earlier reading

An AI.Test run this morning showed 1 failure at `threads=1`, which I reported as "no clean local baseline, needs CI to judge". That was the nuget.org metadata flake this machine hit repeatedly today — not a defect. The baseline above is green at `threads=1` on the same machine, so the comparison is sound.

## Why this is safe

The isolation already exists: `MonolithMeshTestBase` keys its `IServiceProvider` cache by test-class `Type`, and xUnit's parallelism unit is the collection — one per class by default. Concurrent classes therefore already get independent meshes. Opt-in stays **per project**, because parallel safety is a property of the tests, not of the runner.

## Together with #1033

| project | 1 thread | 4 threads |
|---|---|---|
| `Monolith.Test` (#1033) | 314.7s | 122.5s |
| `AI.Test` (this) | 198.2s | 93.7s |
| **total** | **513s** | **216s** |

~5 minutes off the suite's 36.7 min of test work, from the two heaviest projects. That also moves the shard-count question: the ideal per-shard load only makes fewer/denser shards worthwhile once the total shrinks.

## No What's New entry

### DIFF
--- a/test/MeshWeaver.AI.Test/xunit.runner.json
+++ b/test/MeshWeaver.AI.Test/xunit.runner.json
@@ -1,0 +1,6 @@
+{
+  "parallelizeAssembly": false,
+  "parallelizeTestCollections": true,
+  "maxParallelThreads": 4,
+  "methodTimeout": 30000
+}


rbuergi:
Closing — same outcome as #1033, and the same cause. Not shipping red.

## CI red on

```
AI.Test.HarnessInstallGateTest.RequiresInstall_WithInstalledNode_ResolvesHarness
AI.Test.ThreadTokenUsageTest.CompletedRound_WithPromptCache_RecordsCacheTokens_OnCellAndSatellite
Orleans.Test.OrleansContentImportSyncTest.Import_SyncsContentFile_IntoNodeContentCollection_OnGrain
```

The Orleans one is **not** from this change — the diff is a single JSON file inside `test/MeshWeaver.AI.Test/`, and Orleans runs as its own process on another shard. It is the Orleans flake family that has been failing on `main` independently all day (`OrleansStaticRepoImportTest`, `OrleansMeshTests.HubWorksAfterDisposal`, `DelegationCompletionTest`, …), which #1010 addresses.

The two `AI.Test` failures are mine.

## The finding that matters

Local: **1138 tests, 0 failed, 93.7s** at `maxParallelThreads: 4`. CI: two failures at the same setting.

That is now **twice in one session** — #1033 the same way. The cause is the hardware:

| | cores | 4 threads means |
|---|---|---|
| my dev machine | 18 | four threads spread across eighteen cores |
| CI runner | 4 | four threads genuinely contending for four |

Any race whose window scales with CPU pressure reproduces on CI and not locally. **A local parallel green on a many-core machine is not evidence for a 4-vCPU runner** — the same shape of error as trusting a macOS green for `MeshHubDisposalLeakTest`, whose ClrMD analysis cannot run on macOS at all.

## What this means for the parallelism programme

Intra-project parallelism is still the largest available win — measured 2.57x on Monolith and 2.11x here, against 36.7 min of total test work. But the rollout cannot be validated locally. Each project needs:

1. a **CI** run at `threads=4`, not a local one, and
2. the failures it surfaces triaged individually — some are test assumptions (`GcReachabilityCollection` in #1033 is the correct shape for one whole class of them), and at least one looks like a genuine defect (`StoreLevelWriteConflictTest`, on the same store path as the #971/#988 CAS, which a serial suite would never exercise).

The mechanism from #979 (per-project `xunit.runner.json` + the build-time guard) is already merged and correct — enabling a project is one file. What is missing is the per-project CI validation loop, not the plumbing.

`Content.Test` remains enabled and green in production, so the mechanism itself is proven.

PR #1047, opened 2026-08-09
perf(test): per-class test partition — fixes the parallelism blocker, re-lands AI.Test at 2.11x

rbuergi:
Fixes the root cause behind #1040 and re-lands #1035.

## The defect

`MonolithMeshTestBase.TestPartition` was `public const string = "TestData"`, so **every test class in every project wrote into one partition** — and a partition has a single owning hub that serialises every write routed through it. Harmless while classes run one at a time; a contention point the moment they do not.

That is why #1035 red on CI: waits that never emitted, in a class that is **green in isolation under identical CPU pressure** (`ThreadTokenUsageTest`: 8/8 in 5.6s alone; fails alongside 1,130 neighbours).

🚨 Not a timeout budget — the failing shape is "observable emitted nothing at all", i.e. state changed underneath. Widening those waits would have hidden it.

## The fix — one line, opt-in per project

`TestPartition` becomes `public virtual` with the **default unchanged** (`"TestData"`), so no other project moves. Isolation is opted into by a project's own base class:

```csharp
// AITestBase — all 39 AI.Test classes get their own partition, from one line
public override string TestPartition => $"TestData{GetType().Name}";
```

## The cascade the compiler forced

Every one is a hard error, never a silent change:

| | |
|---|---|
| 42 static-qualified refs → instance access | 15 files |
| 3 hardcoded node paths → `$"{TestPartition}/…"` | `ThreadComposerFlowTest` |
| 2 × `CS1736` default-parameter values | → null-default + coalesce |
| 2 × `CS0236` static field initialisers | → instance properties |

The other 20 `"TestData"` strings in AI.Test are `Path.Combine(AppContext.BaseDirectory, "TestData")` — **on-disk directories**, unrelated to the partition — and are deliberately untouched.

🚨 The `CS1736` sites needed the `?? TestPartition` coalesce. `string? ns = null` alone would have passed `null` where the partition used to be — a silent behaviour change **the compiler could not have caught**.

## Verified on a representative machine

A dev Mac has 18 cores; the runner has 4 — so `maxParallelThreads: 4` means something completely different on each. That is precisely why #1033 and #1035 passed locally and red on CI. `DOTNET_PROCESSOR_COUNT=4` makes .NET size its thread pool and GC as though it had 4 cores:

```
before fix, DOTNET_PROCESSOR_COUNT=4 : 1138 tests, 2 FAILED (ThreadTokenUsageTest)
after  fix, DOTNET_PROCESSOR_COUNT=4 : 1138 tests, 0 failed, 117s
```

Serial baseline 198.2s → **2.11x**.

## Scope

`Hosting.Monolith.Test` (#1033, measured **2.57x**) is deliberately **not** re-enabled here — it additionally needs `GcReachabilityCollection` from branch `perf/share-mesh-cluster`. One project at a time, each independently revertible.

## No What's New entry

### DIFF
--- a/src/MeshWeaver.Hosting.Monolith.TestBase/MonolithMeshTestBase.cs
+++ b/src/MeshWeaver.Hosting.Monolith.TestBase/MonolithMeshTestBase.cs
@@ -57,7 +57,10 @@ public abstract class MonolithMeshTestBase : Fixture.TestBase
     /// (e.g., "TestData/mynode") and they'll have proper mesh node hubs.
     /// Registered as a Markdown node so the hub gets AddMeshDataSource + WithNodeOperationHandlers.
     /// </summary>
-    public const string TestPartition = "TestData";
+    /// <summary>Per-class overridable so concurrently-running classes do not contend on one
+    /// partition hub. Default is unchanged ("TestData"): only a project that opts in — by
+    /// overriding in its own base — gets isolation, so no existing test moves. See #1040.</summary>
+    public virtual string TestPartition => "TestData";
 
     /// <summary>
     /// Quiescing budget for test-mesh hubs.
--- a/test/MeshWeaver.AI.Test/AITestBase.cs
+++ b/test/MeshWeaver.AI.Test/AITestBase.cs
@@ -10,6 +10,12 @@ namespace MeshWeaver.AI.Test;
 /// </summary>
 public abstract class AITestBase(ITestOutputHelper output) : MonolithMeshTestBase(output)
 {
+
+    /// <summary>Per-class partition, so concurrently-running AI.Test classes do not contend on one
+    /// partition hub — the root cause behind the parallel failures in #1040 (a partition has ONE
+    /// owning hub that serialises every write routed through it). Overriding here covers all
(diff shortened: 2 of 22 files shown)


rbuergi:
Closing — the fix is proven but this form of it does not compile solution-wide. Full detail on #1040.

**Proven:** per-class partitions fix the contention. `DOTNET_PROCESSOR_COUNT=4` (which reproduces the 4-vCPU runner): **1138 tests, 2 failed → 0 failed, 117s**. Root cause confirmed and remedied — no timeout widened, no test skipped.

**Blocked:** virtualising the shared `MonolithMeshTestBase.TestPartition` breaks the wider solution. Full CI-parity build gives 10 errors, and while `CS0236` (static initialiser → instance property) is mechanical, **`CS0120` is not** — static *methods* consume these values, and in xUnit some of those must stay static (`[MemberData]` sources and similar). Those need per-site design decisions.

**The right scope** is on #1040: leave the const alone, add `protected virtual string ClassPartition => TestPartition`, override once in `AITestBase`, and switch only AI.Test's references. One assembly instead of the solution.

Branch `perf/per-class-partition-ai` is preserved — it already contains the solved `CS1736` and `CS0236` cases inside AI.Test, which the scoped version still needs.

PR #1050, opened 2026-08-09
A declined write no longer hangs the create — both storage paths answer alike

### LINKED ISSUE #981: Callbacks still pending past the teardown quiescing budget (FutuRe EuropeRe, Cession, PandasExplorer) — a CreateNodeRequest the mesh hub never answers

rbuergi:
## The defect

`IStorageAdapter.Write` emits `null` as a **documented try-then-claim sentinel** — *"this adapter does not own this path"*, **not** *"the write succeeded"*. `HandleCreateNodeRequest` filtered that null away with `.Where(n => n is not null)`, so the detached chain that owes the reply **completed empty, posted nothing, and left the requester's `hub.Observe` callback pending forever**.

**The asymmetry is the defect, independent of any capture.** The composite `PersistenceService.Write` turns the *same* condition into a **throw** — *"Could not save '{path}': no writable storage provider accepted the node."* — which the handler's `onError` arm answers correctly. So one condition either **failed cleanly or hung forever**, decided purely by which adapter the hub resolved. The non-composite wirings resolve a single decorated adapter, and several decline by contract:

| Adapter | Declines when |
|---|---|
| `PathFilteringStorageAdapter` | the path doesn't match its predicate |
| `PostgreSqlPathRoutingAdapter` / `SnowflakePathRoutingAdapter` | the partition isn't routable |
| `RoutingProxyAdapter` | no partition hub claims the path |
| `StaticNodeStorageAdapter` | always (read-only) |

Whether a caller gets an answer must not depend on the storage wiring underneath it.

## The fix — two changes, in order of specificity

**1. `RequireClaimedWrite` replaces both save paths' `.Where(n => n is not null)`.** The sentinel now **faults** with an `InvalidOperationException` in the same message family `PersistenceService` already throws, so both paths produce an identically shaped `CreateNodeResponse.Fail(…, ValidationFailed)` from the one `onError` arm. It also records `CREATE_SAVE_DECLINED adapter=<type> path=<p>` on #1011's `RequestFateLedger` — the piece every #981 capture lacked: **which** adapter declined.

**2. The terminal `onCompleted` arm answers instead of only recording.** The save is not the only empty-capable upstream: the existence `Read`, `EnsurePartitionBootstrap`, the validators and the NodeType `Exists` probe are all adapter/permission-supplied observables free to complete empty, and a chain that terminates unanswered is a hang however it got there.

It is gated on **both** flags, and each guard is load-bearing:
- `emitted` — the chain produced a node, so the reply is owed by the post-creation `Subscribe`, whose own arms both answer. Answering here would race a success into a failure.
- `responded` — a branch already posted its own, more specific rejection (already-exists, validation, unknown NodeType). Answering again would post a second response for one correlation.

Only the genuinely unanswerable case is left, and the only correct answer for it is a failure: nothing emitted ⇒ no node was created ⇒ no `Ok` can ever be right.

Every terminal post now goes through one local `Respond(…)` that records the answer. That is what makes (2) **exact** rather than approximate — and it makes a future branch that posts-and-returns-`Empty` suppress the backstop automatically, without anyone having to remember which branches can precede an empty completion.

## Verification — negative control run explicitly

`CreateNodeAlwaysAnswersTest` (4 tests, `HubTestBase` + a scripted storage adapter). With the fix: **4/4 pass in 411 ms**. With **only** `MeshExtensions.cs` reverted:

```
Create_WhenAdapterDeclinesTheWrite_AnswersWithFailure [FAIL] [15 s]
  Expected the observable to emit a value within 15s … but it did not.
  The observable emitted nothing at all.
Create_WhenSaveCompletesWithoutEmitting_AnswersWithFailure [FAIL] [15 s]
  … The observable emitted nothing at all.
Failed! - Failed: 2, Passed: 2
```

A hang is the current behaviour, so a timeout is what the assertion reports — that is the pin.

The other two tests **pass in both states**, which is the point: they are guards, not the pin.
- `…KeepsItsOwnRejectionReason` — already-exists posts a `Fail` then returns `Observable.Empty`; the specific `NodeAlreadyExists` reason must survive rather than being flattened into the backstop's `Unknown`.
- `…AnswersOkEvenThoughTheOkIsPostedAfterCompletion` — a `LatePostCreationHandler` completes on `TaskPoolScheduler`, so the `Ok` is posted **strictly after** the chain's `onCompleted` has run. A backstop that only checked "was a response posted?" (and not "did the chain emit?") would answer `Fail` first and win that race. This is a scheduler hop, not a sleep.

Also green: `MeshWeaver.NodeOperations.Test` **89/89**, `WhatsNewEntryIntegrityTest` + `DocumentationLinkIntegrityTest` 2/2, `-c Release -warnaserror` clean.

## Blast radius — enumerated

Callers that previously **hung** now receive a failure. Every one of them already has a live branch for exactly this response shape, because the composite path produces it today:

| Caller | Behaviour on the new failure | Retry / cascade? |
|---|---|---|
| `MeshService.CreateNode` | maps `ValidationFailed` → `UnauthorizedAccessException`, else `InvalidOperationException` — **the identical mapping the composite throw already gets** | no |
| `HubNodePersistence.CreateNode` | same mapping | no |
| `NodeCopyHelper` | `NodeAlreadyExists` → skip; otherwise throws with the message | no |
| `VUserHelper` | logs a warning | no |
| `ThreadSubmission` (response cell) | "already exists" → commit; otherwise `onFailure` | no |
| `HubThreadExtensions.StartThread` | **has a one-shot fallback** — but `TryBuildUserPartitionFallback` only fires when the error contains `"Access denied"`, which neither new message does, and it is already `canFallBack: false` on the second attempt | no |
| `HandleCreateOrUpdateNodeRequest.DispatchInnerCreate` | `NodeAlreadyExists` → update path; otherwise `PostFail` | no |
| `hub.Post(new CreateNodeRequest(…))` fire-and-forget (`ThreadMessageLayoutAreas`, `ThreadChatView`) | nothing awaited | n/a |

Net: **a wedge is replaced by an already-exercised error path.** No new branch is introduced at any call site.

Two test adapters decline writes (`FreshMeshPartitionProvisioningTest`, `StorageAdapterWriteManyTests`) — both are registered as `IPartitionStorageProvider`s *inside* `PersistenceService`, so their nulls were already folded by the composite and are untouched.

## `Refs #981`, deliberately not `Fixes` — and why

Three previous investigations declined to close #981 on inference; this does not close it either. The honest position:

**The decline fix provably does NOT explain #981's captures.** All five reported members derive from `MonolithMeshTestBase` → `AddInMemoryPersistence()`, whose resolved adapter is `SubtreeDeletionGuard(MonotonicWriteGuard(VersionWriting(InMemoryStorageAdapter)))`. `InMemoryStorageAdapter.Write` **never emits null** — it returns the node or the version winner — and the three decorators only forward. So the sentinel branch is unreachable in the exact wiring the captures came from. This is a real adjacent defect, not the captured one.

**The backstop *might* explain them, and that is exactly what the next capture will say.** If the create chain is what completes empty, the backstop now answers, the pending callback resolves, and the `[QUIESCE-TIMEOUT]` symptom disappears — with an `ERROR` log naming the path and a `CREATE_CHAIN_COMPLETED_EMPTY` ledger stage. If the chain is instead still *running* (a leaf that never completes), the backstop never fires and #981 is untouched. Nothing in the captures distinguishes those yet.

**One lead worth recording for whoever picks #981 up.** In the Monolith wiring the strongest remaining empty-completion candidate inside the create chain is `EnsurePartitionBootstrap` → `hub.CheckPermission(...).TakeDecisionOutsideGate()`. That is `decision.Take(1).SelectMany(HopOffGate)`: if the permission fold **completes without emitting**, `Take(1)` yields nothing, the `SelectMany` chain goes empty, and the whole create completes empty. Note the `.Timeout(15s)` above it does **not** catch this — `Timeout` faults on a stall, not on an empty completion — which fits the captures' 0.7–3.2 s ages being well under 15 s. I have not proven it; it is a code reading, and it is now instrumented rather than silent.

Fixes nothing else; supersedes nothing.

Refs #981


rbuergi:
## CI red on shard 0 — it is the known Orleans teardown-starvation flake, not this diff

Run [31335940378](https://github.com/Systemorph/MeshWeaver/actions/runs/31335940378): build green, shards 1-5 green, **shard 0 red on exactly one test**:

```
MeshWeaver.Hosting.Orleans.Test.OrleansBrokenNodeTypeAccessTest
  .Subscribe_ToLayoutAreaOf_NonCompilingNodeTypeInstance_SurfacesError_NotWedge [FAIL]
  at OrleansBrokenNodeTypeAccessTest.cs(198,0)
MeshWeaver.Hosting.Orleans.Test  Total: 141, Failed: 1
```

**That is the cross-test interaction #1051 root-caused**, and it names this very class as one of its six aliases — most recently on #1045, "whose entire diff adds `Class=\"@Class\"` to four buttons in a `.razor`". `OrleansClusterDisposal` hands each `TestCluster` teardown to `IoPool.Unbounded`, so every cluster the assembly built can be shutting down underneath the class currently running and starve its render on a 4-vCPU runner. #1051's `IoPool(2)` bound is the fix and is still open.

**This diff cannot reach that failure.** Line 198 is the layout-area subscription's `.Timeout(45.Seconds())` — reached only *after* both `meshService.CreateNode(...)` calls returned successfully, which the test's own output confirms:

```
NodeType compile settled at Error.
Instance created.
```

Both creates were **answered**, so neither `RequireClaimedWrite` nor the terminal backstop fired: the changed code paths are the ones that produce a *failure* where there was previously *no response at all*. What timed out is the Overview render, which this PR does not touch.

Per AGENTS.md I have not re-run it — a re-run to "see if it flakes" hides races, and this one is already reproduced deterministically (`DOTNET_PROCESSOR_COUNT=4`) and root-caused elsewhere. **Once #1051 merges, merge main into this branch and re-check** rather than retrying this run.

rbuergi:
## CI red again on shard 0 — same Orleans family, and **#1051 did not fully eliminate it**

Run [31358460359](https://github.com/Systemorph/MeshWeaver/actions/runs/31358460359) (branch re-synced with main, so **#1051's `IoPool(2)` disposal bound IS in this tree**): build green, shards 1-5 green, shard 0 red on one test — a *different* member of the same family:

```
MeshWeaver.Hosting.Orleans.Test.OrleansMarkdownExportTest
  .SubHub_WithoutExplicitRegistration_ResolvesExportDocumentControlViaRegistryChain [FAIL]
  at OrleansMarkdownExportTest.cs(282,0)
MeshWeaver.Hosting.Orleans.Test  Total: 141, Failed: 1
```

`OrleansMarkdownExportTest` is one of the six aliases #1051 itself enumerates ("*…`OrleansMarkdownExportTest` (twice, two different methods)…*"), and the shape is identical to the previous run on this PR: 141 tests, exactly one failure, a `TimeoutException` on a render/resolve path, a different name each time.

**The observation worth recording: bounding cluster disposal at `IoPool(2)` reduced but did not remove it.** That is new information about #1051, not about this PR — flagging it rather than silently re-running.

**This diff cannot reach it.** The test's own output shows the create completed and was answered:

```
Created Markdown node: TestUser/subhub-chain
```

The timeout is on the *subsequent* export/registry-chain resolution at line 282. Both changed paths in this PR (`RequireClaimedWrite`, the terminal backstop, and the new `BOOTSTRAP_PERM_*` stages) only alter what happens when a create is **unanswered** or a write is **declined**; here neither applies. The new stages are `NoteRequestStage` calls, which short-circuit on an empty ledger and allocate nothing when no one is awaiting the id.

Per AGENTS.md I have **not** re-run it — a re-run to "see if it flakes" hides the race, and this one is already reproduced deterministically (`DOTNET_PROCESSOR_COUNT=4`) and root-caused in #1051. The right next step is on #1051's mechanism, not here.

PR #1051, opened 2026-08-09
Bound Orleans cluster teardown so it stops starving the running test

rbuergi:
The shard-0 Orleans flake is **one cross-test interaction wearing six names**. Across four runs it has failed as `OrleansStaticRepoImportTest`, `OrleansMeshTests.HubWorksAfterDisposal`, `DelegationCompletionTest`, `OrleansMarkdownExportTest` (twice, two different methods) and `OrleansBrokenNodeTypeAccessTest` — the last of those on #1045, whose entire diff adds `Class="@Class"` to four buttons in a `.razor`. None of them were anyone's diff.

## Root cause

`OrleansClusterDisposal` hands each `TestCluster`'s stop→dispose to `IoPool.**Unbounded**` and makes it hot immediately. Both of those are right on their own: teardown must not block the xUnit thread (awaiting it inline was the original deadlock this file was written to fix), and the next class should start at once.

With no bound, though, **every cluster the assembly has built can be shutting down at once underneath the class currently running.** Silo shutdown is not cheap. On a 4-vCPU runner those overlapping teardowns starve the live test's render, which then dies on its own 30s `Timeout`. The victim is whoever happens to be running when enough teardowns pile up — hence a different name every time, and hence it never correlated with any change.

## Reproduced deterministically

Not a stress harness — just the runner's shape, via `DOTNET_PROCESSOR_COUNT`:

| build | processors | result |
|---|---|---|
| unbounded | 18 (native) | 141/141 pass |
| unbounded | **4** | **1 FAILED** — 30s timeout |
| unbounded, victim class **alone** | 4 | 5/5 pass in 5s |
| bounded(2) | **4** | 141/141 pass |
| bounded(2) | **2** | 141/141 pass |

The third row is the one that settles it: the class that fails in the suite passes by itself under the same constraint, so it is the overlap and not the test.

## The fix

`IoPool(2)` instead of `IoPool.Unbounded` for the disposal legs.

This is the framework's own prescription rather than a tuning knob — AGENTS.md: *"concurrency bounding … channels through `IIoPool`"* — and `IoPool(int maxConcurrency)` already exists. Teardown stays **off** the teardown thread, so the deadlock the background disposal exists to avoid stays avoided; it simply stops competing with the suite for the runner. Legs of different clusters never wait on one another, so the bound cannot deadlock the final drain.

No What's New entry: test infrastructure, no user-visible change.

### DIFF
--- a/test/MeshWeaver.Hosting.Orleans.Test/OrleansClusterDisposal.cs
+++ b/test/MeshWeaver.Hosting.Orleans.Test/OrleansClusterDisposal.cs
@@ -113,12 +113,34 @@ internal static class OrleansClusterDisposal
     }
 
     /// <summary>
-    /// Runs one <see cref="Task"/>-returning leaf on <see cref="IoPool.Unbounded"/> and projects it to
+    /// The gate every cluster teardown passes through — BOUNDED, not <see cref="IoPool.Unbounded"/>.
+    ///
+    /// <para>🚨 Unbounded was the cross-test starvation. Each disposal is made hot immediately so the
+    /// next class's cluster can start, which is correct — but with no bound, every cluster this
+    /// assembly ever built could be shutting down AT ONCE underneath the class currently running.
+    /// Silo shutdown is not cheap, and on a 4-vCPU runner those overlapping teardowns starve the
+    /// live test's render: it waits on a stream that never gets scheduled and dies on its own
+    /// 30s Timeout. That is the shard-0 Orleans flake — a DIFFERENT test each run, because the
+    /// victim is simply whoever is running when enough teardowns pile up.</para>
+    ///
+    /// <para>Reproduced with <c>DOTNET_PROCESSOR_COUNT=4</c> (the runner's shape): the whole project
+    /// fails there and passes unconstrained, and the victim class passes ALONE under the same
+    /// constraint — so it is the overlap, not the test.</para>
+    ///
+    /// <para>Bounding is the framework's own prescription for this ("concurrency bounding channels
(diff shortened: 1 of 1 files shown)


rbuergi:
Reporting a residual, from a branch that has your `IoPool(2)` bound merged in.

**#1051 reduced this but did not eliminate it.** On a re-synced branch (PR #1050), build green and shards 1-5 green, **shard 0 red on `OrleansMarkdownExportTest.SubHub_…ViaRegistryChain`** — another of the six aliases this issue itself enumerates, and the same shape every time: **141 tests, exactly 1 failure, `TimeoutException`, a different test name on each run.**

That "different name each run, always exactly one" signature is the thing worth acting on — it says the failure is a property of the *assembly's* teardown/startup contention rather than of any test, so chasing whichever name surfaced is chasing an alias.

Attribution for that particular occurrence, so it is not mistaken for the PR under test: the failing test's own output shows `Created Markdown node: TestUser/subhub-chain`, i.e. the create was answered; the timeout is the later resolution at line 282. The PR's diff cannot reach it.

Not re-run to fish for green, per the standing rule.

Two other data points from tonight that may help size it:
- PR #1043 hit the same shard-0 wall on `OrleansBrokenNodeTypeAccessTest` ×2 and `OrleansMeshTests.HubWorksAfterDisposal` before your fix landed.
- PR #1050 hit `OrleansBrokenNodeTypeAccessTest` earlier. Three different alias sets, one shape.

Worth reopening or filing a follow-up for the residual — the bound helped, but the assembly still loses exactly one test per run.

PR #1052, opened 2026-08-09
perf(test): parallelise Threading.Test — 65.5s → 35.3s (1.86x)

rbuergi:
One config file. No base-class change, no scoping, no cascade.

```
threads=1   128 tests, 0 failed, 65.5s
threads=4   128 tests, 0 failed, 35.3s     1.86x
```

## Both numbers measured at CI's core count, not this machine's

Taken with `DOTNET_PROCESSOR_COUNT=4`, which makes .NET size its thread pool and GC as though it had 4 cores — i.e. the runner, not an 18-core dev box.

That distinction is why #1033 and #1035 failed: both were green locally at `maxParallelThreads: 4` and red on CI at the *same setting*, because "4 threads" means four competing for four cores there and four spread across eighteen here. A like-for-like 4-CPU pair is the only local measurement worth quoting.

## It answers an open question from #1040

`Threading.Test` writes into the **same shared `"TestData"` partition** as everything else and parallelises cleanly. So the shared partition is **not a universal blocker** — it only bites where a project has concentrated cross-class state:

| project | why it fails |
|---|---|
| `AI.Test` | 39 `ShareMeshAcrossTests` opt-ins + a shared test user |
| `Hosting.Monolith.Test` | process-wide GC-reachability assertions |
| `Hosting.PostgreSql.Test` | 25 classes on one Testcontainers instance |
| **`Threading.Test`** | **nothing — 1 opt-in, no GC assertions, no shared container** |

So the route forward is **per-project enablement verified with the 100s loop**, not the base-class refactor I attempted and closed in #1047. That is a materially cheaper path.

## Verification

- Green at 4 simulated CPUs, with a like-for-like serial baseline under identical conditions
- **Full solution build** clean under CI's flags — building only the touched project is exactly what let `CS0236` reach CI on an earlier attempt

## No What's New entry


rbuergi:
CI red, but **not from this change** — inherited from main.

```
Hosting.Orleans.Test.OrleansBrokenNodeTypeAccessTest.Instance_OfNonCompilingNodeType_AnswersTerminalError_NotSilence
```

This PR's entire diff is one file: `test/MeshWeaver.Threading.Test/xunit.runner.json`. Orleans.Test runs as its own process on a different shard, so there is no channel by which it could be affected.

`main` is currently unstable on Orleans — the same test fails there twice in the last four red runs, alongside a cluster of others:

```
2x OrleansStaticRepoImportTest.Reimport_ChangedContent_RepairsContentNullPage
2x OrleansMarkdownExportTest.ExportPdf_RoundTrips_AsTypedResponse
2x OrleansBrokenNodeTypeAccessTest.Instance_OfNonCompilingNodeType_AnswersTerminalError_NotSilence
1x OrleansResubmitDeadlockTest.Resubmit_AfterExecution_DoesNotDeadlock
```

Timeline suggests why: Orleans parallelism has landed on main (`test/MeshWeaver.Hosting.Orleans.Test/xunit.runner.json` is present), and #1051 — *"Bound Orleans cluster teardown so it stops starving the running test"* — is open for the fallout.

**Holding this PR until #1051 lands** rather than re-running, which would only re-roll someone else's instability. The change itself stands on its own evidence:

- `threads=1` → 128 tests, 0 failed, 65.5 s
- `threads=4` → 128 tests, 0 failed, 35.3 s (**1.86x**)
- both at `DOTNET_PROCESSOR_COUNT=4`, i.e. the runner's core count rather than an 18-core dev box
- full solution build clean under CI's flags

rbuergi:
Second run on a **fully synced** branch (was 16 commits behind; now current with main) — red again on Orleans, a *different* test:

```
run 1: OrleansBrokenNodeTypeAccessTest.Instance_OfNonCompilingNodeType_AnswersTerminalError_NotSilence
run 2: OrleansMarkdownExportTest.ExportPdfArea_RendersExportDocumentControl_ClientDeserializes
```

Both are `Hosting.Orleans.Test`; this PR's entire diff is one file inside `test/MeshWeaver.Threading.Test/`, which runs as a separate process on a different shard. There is no channel between them.

Both also fail on `main` independently — `OrleansBrokenNodeTypeAccessTest` ×4 and `OrleansMarkdownExportTest` ×2 across its recent red runs, with `main` itself alternating failure/failure/success/success.

**Two different Orleans tests across two synced runs is not a flake this PR can wait out — it is a repository-wide instability.** Blocking on #1051 ("Bound Orleans cluster teardown so it stops starving the running test"), which is the open fix for it.

Not re-running again: the sync was the one legitimate reason to re-trigger CI, and it has been done. Further re-runs would just be sampling someone else's intermittency.

The change itself needs nothing:

- `threads=1` → 128 tests, 0 failed, 65.5s
- `threads=4` → 128 tests, 0 failed, 35.3s (**1.86x**)
- both at `DOTNET_PROCESSOR_COUNT=4` (the runner's core count, not an 18-core dev box)
- full solution build clean under CI's flags
- one file, no base-class change, no cascade

Ready to merge the moment main is green.

PR #1058, opened 2026-08-09
perf(test): parallelise Hosting.Monolith.Test — 317.6s → 110.2s (2.88x)

### LINKED ISSUE #980: StaleActivationSeedRollbackTest: out-of-band IStorageAdapter write not visible to immediate durable read under concurrent load

rbuergi:
The heaviest project in the suite, and the one that sets the shard floor — `shard-assign.sh`: *"The floor for ANY sharding is the heaviest single project."*

```
threads=1   513 tests, 0 failed, 317.6s
threads=4   513 tests, 0 failed, 110.2s     2.88x
```

Both measured with `DOTNET_PROCESSOR_COUNT=4` — the runner's core count, not this 18-core dev box. That distinction is exactly why the earlier attempt (#1033) passed locally and red on CI: "4 threads" means four competing for four cores there, four spread across eighteen here.

## `GcReachabilityCollection` — scoping, not a band-aid

Four classes assert **process-wide** reachability ("after disposing X, nothing still roots it") using `WeakReference` + a forced full `GC.Collect`:

`MeshHubDisposalLeakTest` · `NodeTypeAssemblyLeakTest` · `KernelScriptMemoryLeakTest` · `ConsoleCaptureExecutionContextLeakTest`

Test classes do **not** share a mesh — `ShareMeshClusterEnabled` is `false`, so every `[Fact]` builds its own — but they **do share the heap**. With neighbours allocating, a forced collect cannot be relied on to reclaim the target, so the test reports a "leak" that is really a busy heap.

Under parallelism such a test is not flaky, it is **wrong**: its precondition no longer holds. `[CollectionDefinition(DisableParallelization = true)]` declares that precondition. A retry, a widened timeout or a `[Skip]` would have been the band-aid.

🚨 Documented on the type: the collection is for **reachability assertions only**. A test that is merely slow, or races on shared *state*, must not be added — that is a real defect and serialising it would hide it.

## The other two #1033 failures are gone

`PixelFidelityOfferedOnlyWhenAvailableTest` and `StoreLevelWriteConflictTest` do not reproduce on current main at `threads=4`. Consistent with the store-level write CAS (#971 / #988) having hardened the path the latter exercises — the same fix that closed #980.

## Verification

- Like-for-like pair at 4 simulated CPUs, both green
- **Full solution build** clean under CI's flags — building only the touched project is what let `CS0236` reach CI on an earlier attempt

## Diff

Six files: the collection, four one-line `[Collection]` attributes, and one config. No base-class change, no cascade.

## No What's New entry

### DIFF
--- a/test/MeshWeaver.Hosting.Monolith.Test/ConsoleCaptureExecutionContextLeakTest.cs
+++ b/test/MeshWeaver.Hosting.Monolith.Test/ConsoleCaptureExecutionContextLeakTest.cs
@@ -30,6 +30,7 @@ namespace MeshWeaver.Hosting.Monolith.Test;
 /// scope, keeps the timer alive, and asserts the capture target is collectible
 /// after the scope is disposed.</para>
 /// </summary>
+[Collection(GcReachabilityCollection.Name)]
 public class ConsoleCaptureExecutionContextLeakTest
 {
     [MethodImpl(MethodImplOptions.NoInlining)]
--- a/test/MeshWeaver.Hosting.Monolith.Test/GcReachabilityCollection.cs
+++ b/test/MeshWeaver.Hosting.Monolith.Test/GcReachabilityCollection.cs
@@ -1,0 +1,36 @@
+using Xunit;
+
+namespace MeshWeaver.Hosting.Monolith.Test;
+
+/// <summary>
+/// The collection for tests that assert PROCESS-WIDE GC reachability — "after disposing X,
+/// nothing still roots it" — via <c>WeakReference</c> + a forced full <c>GC.Collect</c>.
+///
+/// <para>🚨 <b>This is scoping, not a band-aid.</b> Those assertions are statements about the
+/// whole process, so they are only meaningful when the rest of the process is quiet. With other
+/// classes concurrently allocating and holding live object graphs, a forced collect cannot be
+/// relied on to reclaim the target, and the test reports a "leak" that is really just a busy heap.
+/// Under parallelism such a test is not flaky — it is WRONG, because its precondition no longer
+/// holds. Declaring that precondition is the fix; suppressing, retrying or widening a timeout
+/// would be the band-aid.</para>
+///
+/// <para><c>DisableParallelization</c> makes xUnit run this collection on its own rather than
+/// alongside the parallel ones, which restores the quiet process these tests require. Everything
+/// else in the assembly keeps running at <c>maxParallelThreads</c> (see
+/// <c>test/MeshWeaver.Hosting.Monolith.Test/xunit.runner.json</c>).</para>
+///
+/// <para>Measured: enabling parallelism on this assembly without this collection reds
+/// <c>NodeTypeAssemblyLeakTest.NodeTypeAssemblyContext_IsCollected_AfterMeshDisposal</c> and
+/// <c>MeshHubDisposalLeakTest.MeshHub_IsCollected_AfterMeshAndServiceProviderDisposal</c> —
+/// both purely because a neighbour was allocating.</para>
+///
+/// <para>Add a class here ONLY if it asserts reachability/collection of an object after disposal.
+/// A test that is merely slow, or that races on shared STATE, does not belong — that is a real
+/// defect and serialising it would hide it.</para>
+/// </summary>
+[CollectionDefinition(Name, DisableParallelization = true)]
+public sealed class GcReachabilityCollection
+{
+    /// <summary>The collection name; referenced by <c>[Collection(GcReachabilityCollection.Name)]</c>.</summary>
+    public const string Name = "GC reachability (must run without parallel neighbours)";
+}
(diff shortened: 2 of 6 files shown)


rbuergi:
Closing. The measurement is excellent and the GC scoping is right, but Monolith has a **broad** parallelism sensitivity rather than a fixable set of defects.

## Measurement (stands)

```
threads=1   513 tests, 0 failed, 317.6s
threads=4   513 tests, 0 failed, 110.2s     2.88x
```
Both at `DOTNET_PROCESSOR_COUNT=4`, plus a clean full-solution build.

## Why it is still not shippable

Each CI attempt surfaces a **different** test:

| attempt | new failure |
|---|---|
| #1033 | `PixelFidelityOfferedOnlyWhenAvailableTest`, `StoreLevelWriteConflictTest` |
| #1058 | `CessionLayoutAreaTest.MotorXL_LayoutArea_ReturnsContent` |

None recur; none appear on `main` (whose current reds are Orleans — `OrleansBrokenNodeTypeAccessTest` ×4 etc., pending #1051). Three distinct tests across two runs is a population of tests that quietly assume exclusivity, not one bug.

## Why the local 4-CPU loop did not catch it — worth knowing

`DOTNET_PROCESSOR_COUNT=4` fixes the *core count*, but not the *neighbour set*. `shard-assign.sh` splits this project across two shards (`…csproj#1/2`, `#2/2`), so on CI each process runs a **different subset** of the 513 tests, with different classes running concurrently. Locally all 513 run in one process. Same core count, different concurrency composition — so different races surface.

To reproduce CI faithfully for a split project you must also run the split:

```bash
SHARD_INDEX=n SHARD_TOTAL=6 .github/scripts/shard-assign.sh   # gives the `#1/2` class filter
DOTNET_PROCESSOR_COUNT=4 ./MeshWeaver.Hosting.Monolith.Test -class <that subset>
```

## What is preserved

`GcReachabilityCollection` on branch `perf/parallelise-monolith` — still correct and still required. Test classes do not share a mesh (`ShareMeshClusterEnabled` is false, every `[Fact]` builds its own) but they **do share the heap**, so a `WeakReference` + forced `GC.Collect` assertion is a process-wide claim that a busy heap falsifies. Those four classes need `DisableParallelization` regardless of what else is fixed.

## Recommendation

Monolith is the biggest prize (**2.88x** on the project that sets the shard floor) and the hardest. It needs per-test triage using the shard-aware repro above, not another enable-and-see attempt. `Threading.Test` (#1052, 1.86x) is the cheap one and is blocked only on main's Orleans instability.

PR #1060, opened 2026-08-10
A dead sibling subscriber silenced the change feed, so live views stopped updating

### LINKED ISSUE #1053: A live children query stops re-emitting after a create completes — missed Added on the change-feed → synced-query path
Characterised while proving what #1040 was **not**. This is a distinct defect with no issue of its own until now.

## The family

`MeshNodeAgentFileStoreTest.*_StaysLive_AndReEmitsWhen*` — at least:
- `ListChildren_StaysLive_AndReEmitsWhenAFileAppears`
- `Search_StaysLive_AndReEmitsWhenMatchingContentAppears`

Both have failed on `main` today.

## The shape

A **30 s timeout where a live children query never re-emitted after a create completed.** The create succeeds; the subscriber that should observe the new child never sees it. That points at a **missed `Added` on the change-feed → synced-query path**, not at the query itself.

## What it is NOT — each ruled out with evidence, so nobody re-treads it

- **Not cross-class shared state (#1040).** It failed on `main` under **serial** execution, where no two classes run concurrently. `AI.Test` also turns out to have no cross-class shared state at all: `ShareMeshClusterEnabled` is `false`, so every `[Fact]` builds its own mesh.
- **Not the deleted-address hang (#1029 / PR #1038).** Those three tests contain **no deletes at all**. The #1038 mechanism requires a delete tombstone.
- **Not the token-usage satellite ordering defect** fixed in PR #1043 — different tests, different mechanism (that one was a point-read of a satellite that may not exist yet).
- **Not observed in the 14 runs** of the #1043 investigation, so it is intermittent and not provoked by parallelism.

## Why it is worth its own issue

It is a **liveness** failure in the data-binding path, which is the mechanism every layout area and every Blazor view depends on. AGENTS.md is explicit that `.Take(1)` on a stream feeding a live view freezes the binding — this is the same class of symptom arriving by a different route, and if a change-feed `Added` can be missed, a user-facing view can silently stop updating.

## Where to look

- The change-feed → synced-query hand-off: does an `Added` published between the initial snapshot and the live subscription have a drop window? A prior investigation established that `StorageAdapterMeshQueryProvider` subscribes `persistence.Changes` into an early backlog **before** the initial query and registers the live `changeBuffer` **before** `initialDone = true`, replaying through the same `Concat` — so that specific window is closed and the gap is elsewhere.
- Whether the create's write publishes `IMeshChangeFeed.Updated` on the path the children query is watching (an emit-on-start that does not publish is a documented failure mode here).
- `PathMatcher.ShouldNotify` / namespace-scoped change relevance for a **children** query specifically.

🚨 Do **not** fix by widening the 30 s timeout, adding a retry, or polling. The create completed; an event was owed and not delivered.

rbuergi:
## The defect

`InMemoryStorageAdapter` published `Write`/`Delete` through a plain `Subject<T>` wrapped in
`catch { /* never throw */ }`.

`Subject<T>.OnNext` delivers **synchronously, in subscription order**, on the caller's thread — so
the first observer that throws **aborts delivery to every observer after it**. The `catch` then
turned that into silence: no error, no log line, and the write itself reported success.

**The thrower is not hypothetical.** Every live synced query owns a
`persistence.Changes → changeBuffer` pipeline (`StorageAdapterMeshQueryProvider`), and a pipeline
caught in its teardown window has a **disposed `changeBuffer`** while the feed subscription is still
delivering into it → `ObjectDisposedException`. One-shot queries — `IMeshService.QueryAsync`,
autocomplete, path resolution — open and tear one of those down constantly, so on any busy mesh the
window is hit whenever a write lands during a teardown. That is exactly why #1053 is intermittent
and why it does not need parallelism to appear.

This is the **same defect the Postgres feeds killed in #889** (`IsolatedChangeFeed`,
`MergedFeedFanoutIsolationTests`). It was never fixed on the in-memory adapter — the feed behind
every monolith test mesh and every in-memory partition in a real portal.

## Evidence, not inference

`SyncedQueryChangeFeedStarvationTest` stages the teardown window deterministically and runs a plain
`hub.GetQuery` children listing over a real monolith mesh. **Negative control** (revert only the
adapter's feed, nothing else):

```
[FAIL] ALiveChildrenQuery_ReEmits_EvenWhenATornDownPipelineSitsOnTheChangeFeed
Failed: 1, Duration: 30 s
```

A 30-second timeout on a live children query that never re-emitted after a create that completed —
the reported signature verbatim. Restore the feed and it passes in 0.5 s. All four new tests were
confirmed red before the fix and green after.

That control also **rules out the two other live leads on the issue**, because it holds them fixed:

- *"Does the create's write publish on the watched path?"* — it does. Same publish, same path; only
  the fan-out changed between red and green.
- *"Does `PathMatcher.ShouldNotify` disagree for a children-scoped watcher?"* — it does not. Same
  matcher, same `scope:children` query; the notification is dropped **after** relevance matching, in
  delivery.

So the gap is neither publication nor relevance nor the synced-query fold: it is the fan-out.

## The fix

`IsolatedChangeFeed` moves from `MeshWeaver.Hosting.PostgreSql` to `MeshWeaver.Mesh.Contract`, next
to the `IStorageAdapter.Changes` contract it enforces, and every adapter now publishes through it —
in-memory, Sqlite, Cosmos and both Snowflake adapters carried the identical plain-Subject feed. The
guarantee belongs to the interface, not to the two backends that happened to get debugged.

Publish sites drop their `catch { }` per the Postgres precedent: the feed isolates each observer and
**logs** it, so a throw escaping there would be a bug in the feed rather than a subscriber's fault to
swallow. A disposed observer is dropped (provably dead); any other throw is isolated but the
subscriber is **kept**, since permanently disabling a live subscriber over a transient fault would
just relocate this very bug.

Snowflake's partition provider gains the same internal `Logger` accessor the Postgres one has, so
the router's feed reports an isolated fault instead of constructing with a null logger.

## Tests

| | |
|---|---|
| `InMemoryChangeFeedFanoutIsolationTest` (3, new) | the fan-out contract at the adapter |
| `SyncedQueryChangeFeedStarvationTest` (1, new) | the #1053 symptom end to end |
| `MeshNodeAgentFileStoreTest` | 24/24 — the family that reported it |
| `MeshWeaver.Hosting.Test` | 134/134 |

No timeout was widened, no retry or poll added, no assertion weakened.

## What's New

Shipped as `Category: Fix` — a view that silently stops updating is user-visible, and per the
skill's note that fixes are the entries that go missing, "would a user notice?" is the bar.

Fixes #1053

### DIFF
--- a/src/MeshWeaver.Documentation/Data/WhatsNew/2026-08-09-live-views-stop-going-quiet.md
+++ b/src/MeshWeaver.Documentation/Data/WhatsNew/2026-08-09-live-views-stop-going-quiet.md
@@ -1,0 +1,22 @@
+---
+Name: Live views no longer go quiet after a save
+Category: Fix
+Description: A list, table or search result could silently stop updating after new content appeared; it now always catches the change.
+Icon: Sparkle
+Order: -20260809
+---
+
+# Live views no longer go quiet after a save
+
+Everything you see in the portal is live: create a page and the folder listing next to it grows, add
+content and an open search result picks it up. Occasionally that stopped happening — the item was
+saved correctly, but a view already on screen kept showing the world as it was a moment earlier, and
+only a reload brought it back in line.
+
+The cause was one notification reaching some watchers and not others. Whenever anything was saved,
+every open view was told about it in turn, and a single watcher that was in the middle of closing
+could cut the announcement short — so every view further down the line was never told at all, with
+nothing reported anywhere.
+
+Each view is now told independently, so no view can be silenced by another, and a watcher that does
+run into trouble is recorded instead of disappearing.
--- a/src/MeshWeaver.Documentation/Data/WhatsNew/2026-08-10-live-views-stop-going-quiet.md
+++ b/src/MeshWeaver.Documentation/Data/WhatsNew/2026-08-10-live-views-stop-going-quiet.md
@@ -3,7 +3,7 @@ Name: Live views no longer go quiet after a save
 Category: Fix
 Description: A list, table or search result could silently stop updating after new content appeared; it now always catches the change.
 Icon: Sparkle
-Order: -20260809
+Order: -20260810
 ---
 
 # Live views no longer go quiet after a save
--- a/src/MeshWeaver.Hosting.Cosmos/CosmosStorageAdapter.cs
+++ b/src/MeshWeaver.Hosting.Cosmos/CosmosStorageAdapter.cs
@@ -1,6 +1,5 @@
 using System.Reactive;
 using System.Reactive.Linq;
-using System.Reactive.Subjects;
 using System.Text.Json;
 using Microsoft.Azure.Cosmos;
 using MeshWeaver.Mesh;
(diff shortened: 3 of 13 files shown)


rbuergi:
## Instrument output

The test log sink is confirmed alive (the `=== TEST START: … ===` marker is present, per #997/#1005),
and the fixed feed now emits the line that was missing for this entire bug class:

```
=== TEST START: SyncedQueryChangeFeedStarvationTest.ALiveChildrenQuery_ReEmits_EvenWhenATornDownPipelineSitsOnTheChangeFeed ===

warn: Change-feed observer Subject`1 was disposed while still subscribed (in-memory);
      dropped from the feed at TestData/live-872ac2314cfc4d7cbda79c6df3a16993/two.
      Delivery to the other 13 observer(s) was NOT affected.
```

Two things worth reading off it:

1. **The path is the child that was created** — `…/two`. The notification was published, and published
   on the path the `scope:children` watcher matches. Publication and `PathMatcher.ShouldNotify` are
   both doing their job; the loss was purely in delivery.
2. **13 other observers.** That is a single test on a fresh monolith mesh. On the old plain `Subject`
   the abort would have starved up to thirteen live subscribers of that one notification — the
   children query under test among them — and the `catch { }` meant not one character of the above
   was ever printed. That density is why this reproduces intermittently under a real suite and why it
   took a staged teardown window to make it deterministic.

Issue #683, opened 2026-07-27
AiSettingsNodeType.AddSkillSource is fire-and-forget — a fast uninstall can be undone by a late install write


rbuergi:
## The race

`AiSettingsNodeType.AddSkillSource(hub, services, user, sourceNamespace)` returns `void` and subscribes to its own write internally:

```csharp
hub.GetWorkspace()
   .GetQuery(...)
   .Take(1)
   .SelectMany(nodes => meshService.CreateOrUpdateNode(target with { Content = merged }))
   .Subscribe(_ => { }, ex => logger.LogWarning(...));
```

The caller gets no completion signal, so the Store's install plan can write its manifest — and report the install finished — **before** the settings update has landed.

Sequence that loses the fix:

1. install → `AddSkillSource` subscribes, write in flight
2. install plan writes the manifest, reports success
3. viewer uninstalls promptly → deregistration finds **no row** to remove, completes cleanly, clears the record
4. the step-1 write lands → the viewer now has a skill source for a package whose nodes are gone, and no record that says they ever installed it

The result is exactly the stale composer source that MeshWeaver.Plugins#199 set out to eliminate, reachable whenever install and uninstall happen close together.

## Why it isn't fixed in #199

That PR fixes the *uninstall* side (withdrawal now propagates failure instead of clearing the record on a silent miss). This one is the *install* side, and fixing it properly means giving `AddSkillSource` a completion signal — an `IObservable<Unit>` overload the install plan can sequence on, so registration is a phase like every other rather than a side effect racing the manifest.

That is a core API change with its own semantics and tests; smuggling it into an uninstall PR would have been the wrong shape.

## Suggested fix

Add an observable-returning overload (keeping the `void` one for existing fire-and-forget callers) and have `Localizer`'s `Register skills` phase sequence on it, so the manifest — which is deliberately the LAST phase — cannot be written before the skill source is actually registered.

Found by the Copilot review on MeshWeaver.Plugins#199.

PR #1062, opened 2026-08-10
Grain activation no longer times out waiting on itself

rbuergi:
## The defect

On **memex.systemorph.com** (AKS ns `memex`) Orleans grain activations faulted and retried forever:

```
fail: MeshWeaver.Hosting.Orleans.MessageHubGrain[0]
      [ACTIVATE] Grain Edu: activation faulted for Edu
System.TimeoutException: No response received in hub cache/tYwK_RNWZUSCQGvEPNvvzg within 00:01:00
for request SubscribeRequest (id=...) → target Edu.
```

The error names the activating grain as an unreachable target — because the request was the grain's own.

## Root cause

`MessageHubGrain.OnActivateAsync` merged `streamCache.GetStream(addressPath)` — the grain's **OWN**
address — into the activation source. The cache opens that upstream with
`GetMeshNodeStreamBypassCache(path)`, i.e. a `SubscribeRequest` addressed at `path`, which the
RoutingGrain delivers straight back to **this** grain, where `DeliverMessage` parks it on
`_hubReadyRaw` until `CompleteActivation` runs.

So on a cold entry the branch is a **strict self-loop**: it cannot answer until the very activation
it feeds has finished. Its only possible terminal is "the grain did not answer itself" — no
information about the node. `Observable.Merge` forwarded that terminal anyway, so the cache hub's
60 s `RequestTimeout` became a **hard ceiling on TOTAL activation**.

That ceiling pre-empted both deliberate slow-path designs downstream, 100% of the time:

- `NodeTypeEnrichmentHelpers.WaitForCompileSettled` **disarms** its wall clock while a compile is in
  flight — a cold pod compiling many dynamic types is meant to run past the budget.
- `BuildEnrichmentChain` catches a stuck type into a visible compilation-error **overlay** so the hub
  activates anyway.

Neither could be reached. Orleans retried; the cache's transient-streak breaker then replayed the
cached `TimeoutException` into each fresh activation, so the retries faulted in milliseconds
(measured burst at 03:57Z: +0 s/+1 s/+2 s/+3 s). The `Edu` partition was unreadable to users and to
MCP with the identical exception.

## Why only this portal (same image on all three)

Measured over 30 h of Loki (ns `memex`): ~36 activation faults an hour, and **every faulting grain
was a `nodeType:Store/Plugin` node** — Edu (231), Feedback, Skill, Essentials, Publish,
Collaboration, DataModelling, Agent, Instances. Exactly the `nodeType:Store/Plugin` set.

- **memex.systemorph.com**: `Store/Plugin` is installed as a dynamic in-mesh NodeType sitting at
  `CompilationStatus = Error` — `CS0117: 'AiSettingsNodeType' does not contain a definition for
  'AddSkillSource'` (that method was deleted in #683; the in-mesh plugin source still calls it, and
  no build or test can see it). That is precisely the case where enrichment legitimately runs long.
- **memex.meshweaver.cloud** (control): `Store/Plugin` does not exist at all, so enrichment's 3 s
  existence probe answers immediately and `Take(1)` disposes the self-loop long before 60 s.

**Data-dependent, not version-dependent** — consistent with all three portals running
`3.0.0-rc1.ci.2509`. The Monolith routing path has no equivalent self-loop, which is why no existing
test caught it.

## The fix

The cache branch may still supply a **VALUE** (a warm entry lets a reactivation skip the storage
read) but never a **TERMINAL**. `ComposeActivationSource` collapses a faulted self-referential branch
to `Empty` and logs the exception in full — it is not swallowed.

Activation keeps a complete set of terminal outcomes without it: the path resolver's node drives
enrichment; nothing within `FirstNodeResolutionTimeout` throws the precise "no MeshNode emitted / no
query provider claims its partition" error; both branches finishing with no node completes the
source (`Empty`, not `Never`, is what keeps the missing-node answer **prompt**); enrichment faults
surface as the overlay.

`ComposeActivationSource` / `BuildActivationChain` are extracted as `internal static` so the whole
chain runs in virtual time — same seam as `WaitForCompileSettled`, same precedent as
`LongRunningOperationCapExceeded`.

## Verification

`MessageHubGrainActivationSourceTest` — 7 tests, virtual time, 40 ms. Reverting only the composition
to plain `Observable.Merge` fails 3 of them with the prod signature verbatim:

```
Activation faulted, but no branch of it had anything to fault ON.
TimeoutException: No response received in hub cache/tYwK_RNWZUSCQGvEPNvvzg within 00:01:00
for request SubscribeRequest (id=...) → target Edu.
```

- `dotnet build -c Release -warnaserror` — `MeshWeaver.Hosting.Orleans` + its test project: 0 warnings, 0 errors.
- 20 grain tests green under Release/`-warnaserror`, including the 9 Orleans **cluster** tests that
  guard the semantics touched here — `GrainActivationCompletesFastTest.NonExistentPath_ActivationFailsFast_NotThirtySecondTimeout`,
  `OrleansBrokenNodeTypeAccessTest` (×2), `OrleansIdleReactivationNoWedgeTest` (×2),
  `OrleansUnresolvableNodeNoWedgeTest` (×3), `PartitionRootActivationTest`.
- `WhatsNewEntryIntegrityTest` + `DocumentationLinkIntegrityTest` green.

## Follow-up (separate, not in this PR)

The aggravator is a data defect in the **plugins** repo: `Store/Plugin`'s in-mesh source calls
`AiSettingsNodeType.AddSkillSource`, deleted from the framework in #683. It needs porting to
`AiSourcesInstallHook`. This PR makes the platform survive it (the type shows its compilation-error
overlay instead of wedging every plugin partition), but the type still needs to compile.

### DIFF
--- a/src/MeshWeaver.Documentation/Data/WhatsNew/2026-08-10-plugin-pages-stop-timing-out.md
+++ b/src/MeshWeaver.Documentation/Data/WhatsNew/2026-08-10-plugin-pages-stop-timing-out.md
@@ -1,0 +1,26 @@
+---
+Name: Plugin pages stop timing out on themselves
+Category: Fix
+Description: Installed plugins such as courses could become unreachable, retry forever, and never recover on their own.
+Icon: Sparkle
+Order: -20260810
+---
+
+# Plugin pages stop timing out on themselves
+
+An installed plugin — a course area, a skills or agents section — could stop opening altogether.
+The page never rendered, tools reading the same content answered "unavailable", and the portal kept
+retrying without ever getting further. Nothing was actually broken: the content was there, and the
+part of the portal that starts a section up was working on it.
+
+The cause was a section asking the portal for its own page while it was still starting up. That
+question could only be answered once start-up had finished, so it always ran out of time — and the
+portal treated running out of time as proof that the section had failed. Start-up was abandoned at
+exactly the moment it was about to succeed, and each retry inherited the same stale answer, so the
+section could never get going again. It hit hardest where a section legitimately takes a little
+longer to prepare, which is precisely where the extra patience was designed to go.
+
+Now a section no longer treats a question it can only answer itself as a verdict on whether it
+started. Sections that need longer to prepare are given that time, a genuinely broken one shows the
+explanatory page it was always meant to show, and a section that is missing is reported straight
+away instead of after a wait. Recovery no longer needs a restart.
--- a/src/MeshWeaver.Hosting.Orleans/MessageHubGrain.cs
+++ b/src/MeshWeaver.Hosting.Orleans/MessageHubGrain.cs
@@ -1,5 +1,6 @@
 ﻿using System.Collections.Immutable;
 using System.Reactive;
+using System.Reactive.Concurrency;
 using System.Reactive.Linq;
 using System.Reactive.Subjects;
 using System.Reactive.Threading.Tasks;
@@ -221,9 +222,12 @@ public class MessageHubGrain(ILogger<MessageHubGrain> logger, IMessageHub meshHu
         else
         {
             logger.LogDebug("[ACTIVATE] Grain {StreamId}: no static node with HubConfig, merging path resolver + mesh-node cache", streamId);
-            // Path resolver gives a fast in-process answer (no SubscribeRequest round-trip)
-            // for routable paths; the mesh-node cache backs it up for dynamic / freshly-
-            // created nodes that the path resolver hasn't indexed yet.
+            // Path resolver gives the AUTHORITATIVE in-process answer (no SubscribeRequest
+            // round-trip) for routable paths; the mesh-node cache is an ACCELERATOR that
+            // replays an already-hydrated entry (a reader kept this path warm across an idle
+            // collection) so a reactivation can skip the storage read. It can only ever
+            // contribute a VALUE — see ComposeActivationSource for why its terminal is
+            // meaningless by construction and must not fault this activation.
             var pathResolver = meshHub.ServiceProvider.GetRequiredService<IPathResolver>();
             var accessService = meshHub.ServiceProvider.GetService<AccessService>();
             // 🚨 Grain activation is INFRASTRUCTURE — reading the node to learn its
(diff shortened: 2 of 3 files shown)


rbuergi:
## The shard-0 red is pre-existing on main — not this PR

`OrleansBrokenNodeTypeAccessTest.Subscribe_ToLayoutAreaOf_NonCompilingNodeTypeInstance_SurfacesError_NotWedge` failed on run 31358456796. It sits in the code this PR touches, so I did not wave it off — I checked main directly:

| Run | Branch | When (UTC) | Result |
|---|---|---|---|
| 31336834864 | **main** | 08-09 21:26 | ❌ `Instance_OfNonCompilingNodeType_AnswersTerminalError_NotSilence` |
| 31356206042 | **main** | 08-10 04:40 | ❌ `Subscribe_ToLayoutAreaOf_…` |
| 31357286162 | **main** | 08-10 05:01 | ❌ `Instance_OfNonCompilingNodeType_…` |
| 31358230786 | **main** | 08-10 05:19 | ❌ `Subscribe_ToLayoutAreaOf_…` |
| 31359416118 | **main** | 08-10 05:41 | ✅ |
| 31358456796 | this PR | 08-10 05:23 | ❌ `Subscribe_ToLayoutAreaOf_…` |

Both tests in that class have been flapping on **main** since at least **2026-08-09 21:26Z** — roughly eight hours before this branch existed. It is a live main-red, not a regression here.

### Why this change cannot be the cause

The test accepts *either* an `OnError` carrying a `CompilationFailed` NACK *or* an `OnNext` that renders the compilation-error overlay, with a **45 s** budget. The self-referential cache subscription this PR neutralises fires at the cache hub's **60 s** `RequestTimeout` — past the budget — so the `OnError` branch was never the path that made this test green. Before and after, the passing path is the overlay, which this PR does not touch. The failure mode is a wedge: nothing surfaced at all within 45 s.

### What I did

Merged current main (this branch was behind, which `strict_required_status_checks_policy` requires anyway), reconciled with a concurrent main-merge someone else had pushed to this branch at 05:23Z (merged, not force-pushed), and re-verified locally:

- `dotnet build -c Release -warnaserror` — 0 warnings, 0 errors
- 20/20 grain tests green, **including both `OrleansBrokenNodeTypeAccessTest` tests** and the 7 new virtual-time ones

Fresh CI is running on `aed123c55`. Not merging — sequencing is yours.

PR #1063, opened 2026-08-10
The bake gate logged a stall it could not enforce

rbuergi:
## The defect

`NodeTypeBakeGateState` is registered **unconditionally** (diagnostics should always be collected). The health check that **enforces** it registers only when `PreWarm:GateReadiness` is true. Those two were independent — so a recorded regression logged:

```
REFUSING READINESS — N NodeType(s) regressed on this image.
The rollout will stall with the previous image still serving.
```

…on a pod with nothing gating it, which then went Ready and took traffic.

That flag is **false on all three production portals** right now. So the line has been claiming a protection that is not wired up — and it was read as evidence the portal was protected while a production outage was diagnosed against it for hours. A message that overstates enforcement is worse than no message.

## The fix

The host **declares** whether it gates. One parse of the config key drives both the state registration and the health check:

```csharp
var gateBake = bool.TryParse(config[EnabledConfigKey], out var p) && p;
builder.Services.AddNodeTypeBakeGate(gateBake);
if (gateBake) builder.Services.AddHealthChecks().AddCheck<NodeTypeBakeHealthCheck>("nodetype_bake");
```

Deliberately **not** re-parsing the key inside the framework: that would be a second source of truth able to disagree with the wiring it claims to describe.

`GatesReadiness` **defaults to false**, so the honest branch is the fallback — a host must opt *in* to claiming enforcement. Unarmed, the message says plainly that nothing consumes the state, startup continues, and instances of those types will fail.

Worded for a batch host as well: `Memex.NodeType.Bake` also registers the gate and never takes traffic, so "this pod will take traffic" would have introduced a fresh inaccuracy into the fix for an inaccuracy.

## Scope

This makes the gate *honest*. It does **not** arm it — that needs `maxUnavailable: 0` and a `progressDeadlineSeconds` ≥ the cold-bake budget first, and on `memex` the strategy is currently `maxUnavailable: 1` at `replicas: 1`, where readiness refusal protects nothing. Arming before fixing that would create false confidence. Those are deployment changes and are deliberately not in this PR.

## Verification

- `NodeTypeBakeGateStateTest` — 16/16 pass, including one that pins **arming changes only what may be claimed, never which outcomes gate**
- `MeshWeaver.Hosting` and `Memex.Portal.Distributed` both build clean under `-c Release -p:CIRun=true -warnaserror`


rbuergi:
Attribution result for the two `Plugins compile against this PR` failures (`Store/Catalog tests=FAILED`):

**This PR's diff is exonerated, with evidence rather than a re-roll.** Ran `mw-plugin-test` locally — this branch's core, Store staged from plugins `origin/main` (verified unchanged since the failing run; only a comment references the deleted `AddSkillSource`): **ALL GREEN, exit 0**. All four files in this diff are also unreachable from the tester's mesh (no `AddNodeTypeBakeGate`/pre-warmer registration there).

So the failure is a **CI-conditions flake in `Store/Catalog`'s Tests area** — most plausibly one of the nine tests added with MeshWeaver.Plugins #382 (cold/warm progress + grace-timer logic, i.e. timing-sensitive by construction) racing on a loaded runner. It will red unrelated core PRs at some rate until pinned. Tracked for follow-up in the plugins repo; the right repro is the Tests area under `DOTNET_PROCESSOR_COUNT`-constrained CPU, not a retry.

(First local repro attempt was invalid — staged from the plugins **primary checkout**, which is parked off-main and still carried the pre-`b10c981` Store. Restaged from `origin/main` via `git archive`.)

PR #1065, opened 2026-08-10
A hub that answered the sender then declared it hadn't, so the sender got two answers

rbuergi:
## The failure

CI shard 0 has been flapping on `MeshWeaver.Hosting.Orleans.Test` (3 of main's last 8 runs). The brief for this work said the cause was teardown starvation. **The evidence refutes that** — the three red runs have three *different* signatures, and two of them die at the FRONT of the assembly, where no teardown backlog exists:

| Run | Test | Signature |
|---|---|---|
| 31336834864 | `Instance_OfNonCompilingNodeType_AnswersTerminalError_NotSilence` | **assertion** `Expected CompilationFailed, but found Unknown` — 28 s in |
| 31336448666 | `OrleansMarkdownExportTest.ExportPdf_RoundTrips` | 60 s timeout, **first class in the assembly** |
| 31334386600 | `OrleansResubmitDeadlockTest.Resubmit_AfterExecution` | stuck at `Status = Executing` after 45 s |

This PR fixes the first — an assertion, not a timeout, and the only signature to appear twice.

## The defect

`MessageService.DeserializeDelivery`'s `UnhandledMessageNack` path — the fallback hub standing in for a node whose NodeType did not compile — **posts** its own typed `DeliveryFailure` (`CompilationFailed` + `NodeTypePath`), then returned `delivery.Failed(reason)`. That overload's documented meaning is the opposite of what just happened: *"the failing site did NOT answer the sender; whoever finishes this delivery owes it a `DeliveryFailure`."*

So the on-target reporter posted a **second** `DeliveryFailure` for the same request — same message text, but the generic `ErrorType.Unknown` and no `NodeTypePath`. Both are `ResponseFor` the one request, so the caller's `Observe(...).FirstAsync()` resolved on whichever arrived first. The error a user saw for a broken page was decided by scheduling.

That explains every observation: an assertion rather than a timeout, byte-identical message text in passing and failing runs, and a hit 28 s in where nothing could have starved it.

`IMessageDelivery` already states the contract: *"A site that posts its own NACK must use `FailedAndNacked` instead, or the sender gets two."* `ReportRoutingFailure` honours both markers a failing site records; the on-target branch never did — it called `ReportFailure` raw, discarding `SenderWasNacked` and any recorded classification.

## The change

- Fallback-NACK path returns `FailedAndNacked` — it *did* answer.
- `ReportOnTargetFailure` mirrors `ReportRoutingFailure`: skip when already answered, keep the verdict the site reached. `Unknown` stays the fallback, preserving unclassified sites' behaviour.
- `ReportFailure` records the marker on the delivery it answered, so the generic registry-hint path stops double-posting too.
- The `DeliveryFailure` ping-pong guard said *"swallow without responding"* but returned plain `Failed`, leaving the reporter owing an answer — and since that delivery's `Message` is `RawJson`, `ReportFailure`'s type guard could not see what it was. It logged its own suppression while a `DeliveryFailure` went out anyway, feeding the ping-pong it names. Now `FailedAndNacked`.

## Evidence

`FallbackNackDuplicateTest` pins the **cause** (two answers) rather than the **symptom** (which won), so it is deterministic — no Orleans, no load, no CPU constraint:

- **Before:** FAILS in 247 ms — `but it emitted DeliveryFailure { … ErrorType = Unknown, NodeTypePath = }`, same request Id and same message text as the good one.
- **After:** PASSES.

Running the Orleans victim **alone passed 10/10 before the fix** (`DOTNET_PROCESSOR_COUNT=4`) — the flip needs the loaded, multi-class assembly, which is why chasing the test name never reproduced it. That is why the proof is the deterministic repro rather than a local flake rate.

- `MeshWeaver.Messaging.Hub.Test`: **162/162** pass (run twice — after the main fix, and again after the ping-pong fix).
- `dotnet build -c Release -warnaserror` on `MeshWeaver.Messaging.Hub`: **0 warnings, 0 errors**.

## Scope — this does NOT make shard 0 green on its own

At least three distinct defects are reddening shard 0. This is the most frequent one. `ExportPdf_RoundTrips_AsTypedResponse` (60 s timeout as the first class, having printed only "Created Markdown node") and `Resubmit_AfterExecution_DoesNotDeadlock` (stuck at `Status=Executing`) are **separate and still open**. Merging this reduces the flap rate; it does not eliminate it.

### DIFF
--- a/src/MeshWeaver.Documentation/Data/WhatsNew/2026-08-10-broken-nodetype-error-names-the-cause.md
+++ b/src/MeshWeaver.Documentation/Data/WhatsNew/2026-08-10-broken-nodetype-error-names-the-cause.md
@@ -1,0 +1,25 @@
+---
+Name: Opening a broken page now tells you what is broken
+Category: Fix
+Description: A page whose code does not compile now reliably reports the compilation error and names the type at fault, instead of sometimes showing a generic failure.
+Icon: Sparkle
+Order: -20260810
+---
+
+# Opening a broken page now tells you what is broken
+
+Opening a page whose underlying code does not compile is supposed to tell you exactly
+that: a compilation error, naming the type at fault, with a route to the compile log.
+Most of the time it did. Sometimes, for the very same broken page, you got an
+unhelpful generic failure instead — no error kind, no type name, nothing to act on.
+Reloading could give you the good message, or the bad one again.
+
+The page was being answered **twice**. The part of the system that stands in for a
+broken type sent a proper diagnosis — "this NodeType did not compile, here it is" —
+and then a second, generic "delivery failed" answer was sent for the same request.
+Whichever arrived first is the one you saw, so the quality of the error depended on
+timing rather than on what was wrong.
+
+A request now gets exactly one answer, and it is the one from the place that actually
+knows what went wrong. The compilation error, the name of the broken type, and the
+link to the compile log are what you get, every time.
--- a/src/MeshWeaver.Messaging.Hub/MessageService.cs
+++ b/src/MeshWeaver.Messaging.Hub/MessageService.cs
@@ -480,6 +480,12 @@ public class MessageService : IMessageService
                 // every other caller's behaviour. See /async + the no-handler path in MessageHub.FinishDelivery.
                 Post(new DeliveryFailure(delivery, message) { ErrorType = errorType },
                     new PostOptions(Address).ResponseFor(delivery));
+                // The sender has now been answered. Record that ON the delivery — the same marker
+                // IMessageDelivery.FailedAndNacked sets — so no later stage answers the same request a
+                // second time. Only on the path that actually posted: the shutdown short-circuit and the
+                // suppressed-control-traffic branch below answer nobody, and a Post that threw did not
+                // either, so none of them may claim the sender was nacked.
+                delivery = delivery.WithProperty(IMessageDelivery.SenderNackedProperty, true);
             }
             catch (Exception ex)
             {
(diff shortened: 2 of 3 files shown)


rbuergi:
Superseded by #1061, which shipped this exact root cause (two `DeliveryFailure`s for one request id) while this was in flight. Closing rather than merging — one mechanism, not two.

Three duplicate-answer sites are **not** covered by #1061's fallback-hub fix and remain open on main. Preserved on `fix/orleans-shard0-errortype` (commit 0d518a39f), unverified by a failing-first test, for someone to pick up deliberately:

1. **Generic registry-hint path** — `UnpackIfNecessary` answers via `ReportFailure(delivery.Failed(failureMessage))`, returns a still-Failed, unmarked delivery, and the on-target caller reports it again. Both carry `Unknown`, so the duplicate is invisible to any assertion about the winner — it shows up only as a count.
2. **DeliveryFailure ping-pong guard** — logs "Suppressing DeliveryFailure-on-DeliveryFailure ping-pong" and then returns plain `Failed`, so a `DeliveryFailure` goes out anyway. `ReportFailure`'s own type guard cannot catch it because that delivery's `Message` is `RawJson`, which is precisely why the discriminator check exists. The guard has never actually suppressed anything.
3. **On-target classification loss** — the on-target branch calls `ReportFailure` raw, discarding the `ErrorType` the failing site recorded via `Failed(message, errorType)`. `ReportRoutingFailure` honours it; the on-target branch does not.

Also worth a decision: `FailureAlreadyReported` (#1061) and the pre-existing `FailedAndNacked`/`SenderWasNacked` now express the same "already answered" concept through two independent markers.

PR #1068, opened 2026-08-10
The plugin gate stopped covering every commercial package, silently

rbuergi:
## The hole

`#830` put the free/commercial boundary **on the action**: `PackageInstaller.Install` refuses a
priced package unless an explicit `authorizingUserId` is a global admin. That is right, and it
deliberately closed the machine paths — the unattended default install and `PluginUpdateWatcher`'s
auto-update.

**The plugin GATE is a machine path too, and it was not moved with it.** `PluginGateRunner` calls

```csharp
PackageInstaller.Install(harness.Mesh, package, files, snapshot.CommitSha)
```

— `authorizingUserId` defaults to `null`, i.e. *"nobody authorized this"*. So every priced package
is refused before a single node is written, and the gate reports it as failed on install having
compiled, rendered and tested **none** of it. A gate that cannot test a whole class of package is
not a strict gate; it is a hole that happens to be loud.

## It is live, not hypothetical

`MeshWeaver.Plugins` [PR #383](https://github.com/Systemorph/MeshWeaver.Plugins/pull/383) bumped its
image pin onto `eb330e1a2`, and `Manufacturing` (`price: -1 CHF`, coupon-only) went red with
`PackageAuthorizationException` — nothing wrong with the package, no principal to authorize it.
Every other package in that repo passed.

Measured on that repo's `Store` + `Manufacturing` subset, one variable at a time:

| tester / image | `Manufacturing` | exit |
|---|---|---|
| old pin `631b3eed` (pre-#830) | `[PASS]` | 0 |
| new pin `e772e609`, unpatched | `[FAIL] install: PackageAuthorizationException` | 1 |
| new pin `e772e609`, `price` forced to `0` | `[PASS]` | 0 |
| **this fix**, locally | installs 86 nodes, **9/9 NodeTypes compile + render** | — |

Row 3 is the control: with the price the only thing changed, the same image installs the same
package happily. The refusal is about authorization, nothing else.

## The fix — two halves, because the harness was missing both

1. **`GateMesh` seeded its Admin grant at ROOT `_Access` only.** `IsGlobalAdmin` reads
   `Permission.All` at scope `Admin` — an `AccessAssignment` in `Admin/_Access` — and a root grant
   is deliberately *not* that shape (AccessControl.md → "The Admin partition"). So even a principal
   passed in would not have been an admin. Now seeded at both scopes, mirroring
   `TestUsers.PublicAdminAccess`.
2. **Both `Install` call sites pass that principal explicitly** (the install and the idempotence
   re-install). A CI gate is an attended, operator-run check; it presents an admin rather than
   installing as nobody.

## Pinned by a test

`PluginGateRunnerTest.CommercialPackage_InstallsAndGatesGreen` — a fixture package carrying
`price: -1` must install and compile green through the gate, and the run must not contain
`PackageAuthorizationException`. It fails without **either** half of this change.

```
Discovered 1 package(s), install order: Priced
── Priced: installed (3 written, 0 unchanged); checking 1 NodeType(s)…
   ok  Priced/Thing [compile:Passed render:Passed tests:Skipped]
[PASS] Priced (3 node(s), 1 type(s))
ALL GREEN.
```

`dotnet build … -c Release -warnaserror` clean for both touched projects.

## One thing this change UNCOVERED but does not cause

With the package finally installing, the gate reaches the idempotence pin — and locally, on this
branch, `Manufacturing` re-install writes 4 nodes instead of 0:

```
idempotence: re-install of the unchanged snapshot wrote 4 node(s) (expected 0):
  Manufacturing/EnergyPrice/DE20260423, .../DE20260424, .../DE20260425Forecast,
  Manufacturing/Operator/MarcoRossi
```

**Not caused by this change** — it reproduces identically with the price forced to `0` (no
authorization involved at all), and it does **not** reproduce in the pinned image
(`e772e609` / core `8e7d862`, same package, same free variant → `[PASS]`). `src/MeshWeaver.PluginCatalog/`
has no commits between `8e7d862` and `main`, so the remaining candidates are a serialization/typing
change elsewhere in that window (e.g. `32883fc70`) or a host difference (macOS vs the linux
container). Filed here as an observation with its evidence rather than chased — it deserves its own
attribution, and it is exactly the kind of finding that stayed invisible while the gate could not
install the package at all.

### DIFF
--- a/test/MeshWeaver.PluginTester.Test/PluginGateRunnerTest.cs
+++ b/test/MeshWeaver.PluginTester.Test/PluginGateRunnerTest.cs
@@ -97,6 +97,16 @@ public class PluginGateRunnerTest(ITestOutputHelper output)
         }
         """;
 
+    // ── a COMMERCIAL package: identical to the good one except it carries a price. `price: -1`
+    //    is the coupon-only shape the plugins repo's `Manufacturing` ships; any non-zero price
+    //    makes PackageEntitlement.IsCommercial true, which is the whole point of the fixture. ──
+
+    private const string PricedIndexJson =
+        """{"$type":"MeshNode","id":"Priced","namespace":"","path":"Priced","mainNode":"Priced","name":"Priced Plugin","nodeType":"Space","state":"Active","content":{"$type":"PluginManifest","description":"A commercial plugin.","price":-1,"currency":"CHF"}}""";
+
+    private const string PricedThingNodeTypeJson =
+        """{"$type":"MeshNode","id":"Thing","namespace":"Priced","path":"Priced/Thing","mainNode":"Priced/Thing","name":"Thing","nodeType":"NodeType","state":"Active","content":{"$type":"NodeTypeDefinition","description":"A thing.","configuration":"config => config.WithContentType<Thing>().AddDefaultLayoutAreas()","includeGlobalTypes":true}}""";
+
     [Fact(Timeout = 300_000)]
     public async Task GoodPackage_CompilesRendersAndExecutesTestsGreen_ExitsZero()
     {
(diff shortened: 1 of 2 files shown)


rbuergi:
**Blast radius is wider than the one package that surfaced it.**

Scanned the node repos for priced packages (`content.price != 0` on the package root):

| repo | priced packages | its pin today |
|---|---|---|
| MeshWeaver.Plugins | `Manufacturing` (-1) | bumped past #830 → **red now** |
| **MeshWeaver.Reinsurance** | `Reinsurance` (900), `LossModelling` (900), `Planning` (900), `ClaimsDeepfield` (-1), `ReinsuranceDemo` (-1) | still `sha256:10462f9a`, **pre-#830** |
| MeshWeaver.SocialMedia | none | — |

Reinsurance is green only because its pin predates the entitlement gate. Bumping it — which is exactly what a stale pin is supposed to be — turns **five** of its packages red at once, none of them broken, all of them refused before a node is written. So the practical effect of the hole is that the gate quietly stops covering the most valuable packages in the catalog, and the repo most affected is the one that has not yet noticed.

That is also why the fix belongs here rather than in any node repo: no amount of allowlisting in a plugins repo restores coverage — it only records that coverage is gone.

rbuergi:
**Tightening the control on the idempotence observation** — my first pass did not actually isolate it. Both local runs I compared had this fix applied (the tester passes the principal regardless of price), so "price forced to 0" varied the price, not the change.

Redone properly: same host, same core `main`, same free variant, only `authorizingUserId` varied — suppressed back to `null` at both call sites (the pre-fix behaviour), rebuilt, re-run:

| local run | authorizingUserId | idempotence |
|---|---|---|
| this fix | `GateMesh.AuthorizingUserId` | `wrote 4 node(s)` — EnergyPrice/DE20260423, /DE20260424, /DE20260425Forecast, Operator/MarcoRossi |
| principal suppressed | `null` | **`wrote 4 node(s)` — the same four** |

Identical failure with the change backed out, so it is not caused by threading the principal (and the paths are content nodes, not the install record that `AuthorizedBy` is stamped on). It still does not reproduce in the pinned image (`e772e609` / core `8e7d862`), so the remaining candidates are unchanged: something in core between `8e7d862` and `main` outside `MeshWeaver.PluginCatalog/`, or a macOS-vs-linux host difference. Working tree restored; nothing from this experiment is in the diff.

PR #1071, opened 2026-08-10
Remove the skip-trapdoors: a gate with a missing input must go RED, never grey

rbuergi:
## The defect class

A gate that **skips** when one of its inputs is missing. GitHub renders a skipped job with the same grey/green tick as a passed one, so *"the gate never ran"* and *"the gate passed"* are indistinguishable on the PR. A required check that passes on no evidence is worse than a flaky one.

This is not theoretical. The `Plugins compile against this PR` gate has **never run** — here it is on main today (run 31358162940, `run_conclusion=success`):

```
JOB  Plugins compile against this PR: success          ← green
  - Check out MeshWeaver.Plugins:        success       ← its OUTCOME was failure;
                                                          continue-on-error rewrote
                                                          the conclusion to success
  - Compile + run the VITAL plugins ...: skipped       ← the gate itself
  - Warn — gate not yet armed:           success
```

That is how #683 deleted `AiSettingsNodeType.AddSkillSource` with a live caller in `Store/Plugin/Source/Localizer.cs` and left **nine** plugin partitions on the compilation-error overlay in production, and how a separate `AddTracking` deletion broke `SocialMedia/Post`. `Store` is in `VITAL_MODULES` precisely so an armed gate would have caught it.

## What the defensive conditions actually bought

Nothing that mattered. The job **already** excludes forks at the job level (`github.event.pull_request.head.repo.fork != true`), so the one legitimate skip was handled. The `continue-on-error` + `steps.token.outputs.present` dance only covered the case where the credential is missing on a run that could and should have used it — i.e. exactly the case that must be loud.

## The concept change

One **`preflight`** job asserts every CI input this workflow needs from outside the tree, fails RED naming exactly what to provision, and the gates then depend on it and run **unconditionally**. The fork exemption is expressed **once**, as a fork check — never as a "is the secret set?" check, because at the job level those two are indistinguishable and only one of them is safe. Adding an input is one line in `missing`.

`collect-results` gains `preflight` as a `needs` plus an explicit fail step. It runs with `always()` and is the repo's **only** required status check, so without that a preflight failure would skip `plugin-gate` and still report a green, merge-ready PR — the same trapdoor one level up.

## The credential

`PLUGINS_REPO_TOKEN` **never existed** in this repo's secrets — that is why the gate was inert. The credential that exists is `PLUGINS_REPO_SSH_KEY` (read-only deploy key `159776507` on the plugins repo), so the checkout moves to `ssh-key:`.

## Evidence

Ran the BEFORE and AFTER shapes side by side on real Actions, against inputs that genuinely do not exist (throwaway branch, since deleted — run [31360767677](https://github.com/Systemorph/MeshWeaver/actions/runs/31360767677)):

| Job | Conclusion |
|---|---|
| **BEFORE** — `continue-on-error` + `if: steps.plugins.outcome == 'success'` | ✅ **success**, gate step `skipped` — the trapdoor |
| **BEFORE** — job `if` on an unset variable | ⬜ **skipped** — grey tick |
| **AFTER** — preflight, input **absent** | ❌ **failure** |
| **AFTER** — preflight, input **present** | ✅ success — fail-closed does not red every PR |
| **AFTER** — checkout with the deploy key, **no** `continue-on-error` | ✅ success (`Store: 97 files`, `Essentials: 36 files`) |

The failing preflight renders as an actionable annotation:

> **Missing required CI input:** `PLUGINS_REPO_SSH_KEY` (secret) — private half of the read-only deploy key on Systemorph/MeshWeaver.Plugins. Used by the 'Plugins compile against this PR' gate to check that repo out.

YAML parses, job list gained only `preflight` and dropped none, every `needs`/`if` resolves.

## Overlap

- **#1067 is open and edits this exact block** — it arms the same `PLUGINS_REPO_SSH_KEY` and keeps a "credential absent → warn" branch. This PR removes that branch (the preflight replaces it). **Either order works, but they will conflict**; whichever lands second needs the other's block reconciled. Suggested: merge **#1067 first** (smaller, already reviewed), then this rebased on top — or close #1067 as subsumed, since this PR carries its ssh-key switch.
- Companion PR in the plugins repo removes the two `if: ${{ vars.MW_TEST_IMAGE != '' }}` trapdoors with the same preflight concept.

## Deliberately left alone

The four `continue-on-error`s that are legitimate and documented: the `Publish Test Results` reporter (×2 — the TRX summarize step is the real gate, a GitHub-API 429 must not fail the run), and the green-marker push and prune (losing a marker costs a redundant run, never correctness).

### DIFF
--- a/.github/workflows/dotnet-test.yml
+++ b/.github/workflows/dotnet-test.yml
@@ -186,6 +186,69 @@ jobs:
           echo "- TTL: \`${CI_GREEN_TTL_SECONDS}s\`"
         } >> "$GITHUB_STEP_SUMMARY"
 
+  # ─────────────────────────── PREFLIGHT ───────────────────────────
+  # ONE place that asserts every CI INPUT this workflow needs from outside the tree
+  # (repo secrets and variables). It exists so that no gate downstream ever has to
+  # defend itself against a missing input.
+  #
+  # 🚨 THE BUG CLASS IT REMOVES — "skip-trapdoors". The shape it replaces was a
+  # per-gate `continue-on-error:` plus an "is the secret set?" branch, which turns an
+  # unprovisioned credential into a SKIPPED job — and GitHub renders a skipped job
+  # with the same grey/green tick as a passed one. "The gate never ran" and "the gate
+  # passed" become indistinguishable, and a required check that passes on no evidence
+  # is worse than a flaky one. That is not theoretical: the cross-repo plugin gate
+  # below sat inert-and-green from the day it was added (PLUGINS_REPO_TOKEN was never
+  # provisioned) — long enough for #683 to delete `AiSettingsNodeType.AddSkillSource`
+  # with a live caller in the plugins repo and take out nine plugin partitions in
+  # production, and for a separate `AddTracking` deletion to break SocialMedia/Post
+  # the same way.
+  #
+  # The contract, in three lines:
+  #   * a missing input is a RED job naming exactly what to provision — never a skip;
+  #   * gates then run UNCONDITIONALLY. No gate may test its own inputs.
+  #   * the ONE legitimate exemption — a fork PR cannot read org secrets — is expressed
+  #     HERE, once, as a FORK check. Never as a "the secret is empty" check: at the job
+  #     level those two are indistinguishable, and only one of them is safe.
+  #
+  # Adding an input? Add a line to `missing` below and nothing else.
+  preflight:
+    name: "Required CI inputs"
+    timeout-minutes: 5
+    runs-on: ubuntu-latest
+    steps:
+    - name: "Assert every required secret / variable is present"
+      env:
+        PLUGINS_REPO_SSH_KEY: ${{ secrets.PLUGINS_REPO_SSH_KEY }}
+        IS_FORK: ${{ github.event.pull_request.head.repo.fork }}
+      run: |
+        set -euo pipefail
+        # The one sanctioned exemption. GitHub withholds org secrets from fork runs BY
+        # DESIGN, and every gate that needs one carries the same fork check at its job
+        # level, so nothing downstream runs half-armed.
+        if [ "${IS_FORK:-}" = "true" ]; then
+          echo "Fork PR — org secrets are withheld by design; the gates that need them do not run."
+          echo "Fork PR: secret-dependent gates are not applicable." >> "$GITHUB_STEP_SUMMARY"
+          exit 0
+        fi
+        missing=()
+        [ -n "${PLUGINS_REPO_SSH_KEY:-}" ] || missing+=("PLUGINS_REPO_SSH_KEY (secret) — private half of the read-only deploy key on Systemorph/MeshWeaver.Plugins. Used by the 'Plugins compile against this PR' gate to check that repo out.")
+        if [ "${#missing[@]}" -ne 0 ]; then
+          for m in "${missing[@]}"; do
+            echo "::error title=Missing required CI input::$m"
+          done
+          {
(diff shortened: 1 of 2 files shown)


rbuergi:
## Status after merging current main (incl. #1067)

Merged `origin/main` at `b2d69b5d6` and resolved the conflict in the `plugin-gate` block **in #1067's favour** on everything load-bearing:

- **kept** `ssh-key: ${{ secrets.PLUGINS_REPO_SSH_KEY }}` and its full deploy-key-vs-PAT-vs-App rationale, including how to rotate;
- **kept** the `tools/MeshWeaver.PluginTester` artifact packing — arming the gate revealed it could never have run, since the artifact packs TEST project bins and the tester is a *tool*;
- **kept** #1067's fuller account of the #683 outage, moved into the `preflight` comment where the rule is now stated;
- **dropped** the `continue-on-error`, the "is the credential configured?" probe, and the three outcome-gated branches — those are the trapdoor this PR exists to remove. The job already excludes forks at job level, so they only ever covered the case that must be loud.

### The gate now genuinely runs — and passes

```
=== mw-plugin-test summary ===
[PASS] Store (96 node(s), 9 type(s))
    ok  Store/Catalog:    compile=Ok render=ok tests=ok    44/44 passed
    ok  Store/Plugin:     compile=Ok render=ok tests=ok  290/290 passed
    ok  Store/Enrollment: compile=Ok render=ok tests=ok    13/13 passed
    ok  Store/Provision:  compile=Ok render=ok tests=ok      9/9 passed
[PASS] Essentials (34 node(s), 0 type(s))
ALL GREEN.                                            real 0m37.228s
```

130 nodes, 9 NodeTypes, 356 tests executed — with **no** `continue-on-error` and **no** step-level `if`. `Required CI inputs` also green. This is the first time this gate has tested anything.

### CI is red on a pre-existing main flake, not on this change

This branch's diff is `.github/workflows/dotnet-test.yml` + `AGENTS.md` — **zero source files** — so it cannot cause a test failure. The collector's own step results confirm the red is the test gates, not this PR's gates:

```
Summarize test failures:                 failure   <- pre-existing test flake
Fail on non-zero project exit:           failure   <- same
Fail if the plugin gate failed:          skipped   <- gate PASSED
Fail if a required CI input is missing:  skipped   <- preflight PASSED
```

Failing tests are the documented `MeshWeaver.Hosting.Orleans.Test` cross-test race, plus one `ContentCollectionReferenceTest`. **main is red on the same cluster right now**, with a *different* Orleans test name each run — the signature of one shared race, not a per-PR defect:

| run | branch | failing test |
|---|---|---|
| 31361913384 | this PR | `OrleansMarkdownExportTest.SubHub_WithExportTypesRegistered_…`, `OrleansStaticRepoImportTest.Import_CreatesSpaceRoot_AndChildContent`, `ContentCollectionReferenceTest.GetDataRequest_WithEmptyContentCollectionReference_…` |
| 31361446933 | **main** | `OrleansSubmitFromIdleTest.SubmitFromIdle_ThreeSequential_EachReachesIdle` |
| 31358230786 | **main** | `OrleansBrokenNodeTypeAccessTest.Subscribe_ToLayoutAreaOf_NonCompilingNodeTypeInstance_SurfacesError_NotWedge` |

Not re-running to "see if it was a flake" — per AGENTS.md that hides the race. Flagging it rather than papering over it; this PR should not merge until that cluster is addressed or the merge is taken deliberately.

### Companion

Systemorph/MeshWeaver.Plugins#384 removes the two `if: ${{ vars.MW_TEST_IMAGE != '' }}` trapdoors with the same preflight concept — **green**.

PR #1075, opened 2026-08-10
Isolate the five machine-hungry AI.Test classes, then let the rest run in parallel

### LINKED ISSUE #1040: AI.Test is not parallel-safe: cross-class shared state (repro in 100s via DOTNET_PROCESSOR_COUNT=4)

rbuergi:
Fixes #1040.

`AI.Test` runs single-threaded today. This lands the opt-in that #1043 measured and backed out —
but only for the classes that can take it, with the five that cannot isolated into their own serial
collection.

## The premise of #1040 stays disproven; the blocker is contention, not shared state

`MonolithMeshTestBase.ShareMeshClusterEnabled` is `false`, so every `ShareMeshAcrossTests` override
in the assembly is ignored and **every `[Fact]` builds its own mesh**. Two classes never share a
workspace, hub, `IStorageAdapter` or `IMeshNodeStreamCache`. The 502-call-site `TestPartition`
refactor proposed in the issue thread would have made per-class what is already per-*test*, and is
not done here.

What actually failed when #1043 opted in were three concurrency-**stress** tests on shard 4.
`CrossHubPatchAtomicityTest`'s own diagnostic settled what kind of failure that was:

```
WRITE ERRORS (0): []
OWNER-TRUTH probe: version=289 present=288 missing=[]
MIRROR SEQUENCE (192 emissions), REGRESSIONS (0): []; converged=NEVER
```

No write failed, the owner held all 288, the mirror advanced monotonically with zero regressions and
was still climbing when the bound expired. **Starved, not wrong.**

## Membership rule — structural, not "whatever failed last time"

Choosing the set off one CI run would be guessing. A class joins `ConcurrencyStressCollection` only
when **both** hold:

1. it **creates concurrency of its own** — N operations deliberately in flight at once, or a
   dedicated pump thread — rather than doing one thing at a time; **and**
2. its verdict is a **wall-clock bound on that burst**: the assertion that fails is a deadlock /
   lost-write detector, not a functional comparison.

(1) alone is just a slow test, which is exactly what parallelism is for. (2) alone is a generous
budget on sequential work, which survives sharing a box. Together they describe a test that measures
the machine.

| class | what it does | the bound that decides it |
|---|---|---|
| `CrossHubPatchAtomicityTest` | 6 mirrors × 48 writes × 4 rounds, `Observable.Merge`d into one burst so they hit the owner together | 30 s settle on "no entry lost" |
| `PatchWorkspaceAckTest` | 10 patches fired in parallel (`Patch_ConcurrentUpdates_NoDeadlock`) | `< 45 s`, set deliberately so lock contention short of a full deadlock still fails |
| `AgentChatClientDeadlockTest` | 8 concurrent `GetOrderedAgentsAsync` chains | `WaitAsync(15 s)` on all eight |
| `InboxToolIntegrationTest` | 12 submits fired back-to-back at a thread already mid-round (`Hammer_ConcurrentSubmits_AllIngested_NoneLost`) | 150 s settle on "none lost" |
| `DelegationDeadlockTest` | runs the tool on its own single-threaded pump thread | `WaitAsync(10 s)` / `WaitAsync(5 s)` |

Deliberately **not** included, and why the rule matters:

- `ActivityTerminalSettleTest.HammeredAppends_…` — "hammer" in the name, but the appends are
  sequential on one thread and the bound is 30 s over ~3 s of work. Fails (1).
- `ThreadSubmissionIntegrationTest.Submit_ThreeRapidSubmissions_…` — a burst of three against 45 s
  and 30 s bounds. ~15× headroom; fails (2).
- `McpReturnTimingTest` — 10 s per-call hang detectors, but one call at a time. Fails (1).
- `ConnectStrategyTest` — spawns the fake-CLI subprocess, but one at a time, on 8–20 s bounds.
  Fails (1).

## Why the budgets are not the thing to widen

Every bound listed above is a deadlock or loss detector. Padding one so a starved run passes leaves
a detector that detects nothing — the failure mode it exists for would sail straight through.
Scheduling is the correct lever precisely because the tests are right and the contention is the
artefact.

`DisableParallelization = true` rather than merely sharing a collection: a shared collection alone
would stop two stress classes overlapping **each other** while still letting three ordinary classes
run alongside one of them. These bounds are sized against an idle box, so "alone" is the property
they need. Same idiom `Content.Test` already runs in production alongside its own parallel opt-in.

## The opt-in mechanism

`test/MeshWeaver.AI.Test/xunit.runner.json` — the per-project override from #979, identical to
`Content.Test`'s and `Hosting.Orleans.Test`'s. The two mutually-exclusive `Content` includes in
`test/Directory.Build.props` switch to the project-local file on `Exists()`, and
`VerifyXunitRunnerConfigCopied` fails the build if neither branch lands a config in `$(TargetDir)`
— so this cannot regress into xUnit's own defaults (unbounded parallelism) silently.

The CI shard split is untouched: `shard-assign.sh` still splits `AI.Test 2`, and the split enumerates
classes and takes every Nth, which is orthogonal to collection membership. Weight left at 263.

## Sequencing — this branch contains #1043

#1043's own conclusion was "land the ordering defects, **then** opt in". Both of its fixes are
prerequisites for this change and are not merged yet, so this branch is stacked on top of it:

- the token-usage satellite point-read (`ThreadTokenUsageTest.WaitForUsage`), which errors rather
  than waits when the satellite has not landed — a race parallelism raises the rate of;
- the second door into #1029 in `MessageService` (an *accepted* delivery throwing
  `HubDisposingException` mid-handler hitting a `Catch` that hardcoded `ShuttingDown`).

If #1043 merges first this diff shrinks to just the collection split on its own.

`ToolTimeoutAttributeTest` is deliberately left in the parallel bulk. #1043 identified a genuine
one-line race there (`Task.WaitAsync(timeout, ct)` can report the tool budget for a *cancelled*
call) and chose not to ship it because nothing can pin it deterministically without injecting the
timeout decision. Moving the test into the serial collection would hide that product defect behind
test scheduling, which is the band-aid, not the fix.

## Local measurement is not offered as evidence

`DOTNET_PROCESSOR_COUNT=4` sizes the thread pool and GC as if there were four cores but **does not
take the other fourteen away**, so a test that spawns its own concurrency still gets real
parallelism on an 18-core box. That is why #1043's five consecutive green local runs did not predict
CI. The numbers reported on this PR are CI numbers.

## No What's New entry

Pure-internal: test scheduling and CI wall time only, no user-visible behaviour change.


rbuergi:
## Measured on CI — 1.15×, with both AI.Test halves green

Run [31362879486](https://github.com/Systemorph/MeshWeaver/actions/runs/31362879486) (head `d3b2440d`).
Both AI.Test shard jobs **executed** (`Run tests (shard 3)` and `(shard 4)` = `completed/success`) — not a
reused-green-tree skip — so these numbers describe tests that actually ran.

The figure is the assembly's own `Time:` line, which is directly comparable across runs regardless of
runner/artifact overhead. Serial baselines are two same-day green **main** runs whose shards also
executed.

| AI.Test half | main `cf38cf6a`<br>(31361180771) | main `c88a019a`<br>(31359626059) | **this PR** |
|---|---|---|---|
| part 1/2 → shard 3 (540 tests) | 169.4 s | 149.7 s | **125.1 s** |
| part 2/2 → shard 4 (599 tests) | 118.8 s | 110.9 s | **113.8 s** |
| **combined** | **288.2 s** | **260.7 s** | **238.9 s** |

- vs the two-sample mean serial baseline (274.5 s): **1.15×**
- vs the closest-in-time baseline (288.2 s): 1.21×; vs the faster one (260.7 s): 1.09×
- for reference, #1043's full (and red) opt-in measured 275.4 s → 212.0 s = 1.30×

**Where the gain is, and where it isn't.** Part 1 moved 169.4 → 125.1 s; part 2 barely moved
(118.8 → 113.8 s). That is not the stress collection's doing — #1043's *full* opt-in measured the
same lopsidedness on the same half (105.2 → 99.0 s, 1.06×). The 599-test half is serialised by its
own long tests, so there is little for a fourth thread to do there.

`shard-assign.sh` weight left at 263 — one sample either side is not grounds to re-weight.

## The serial bucket costs almost nothing

Per-class wall span, measured from the trx of a full local parallel run (1139 tests, 0 failed):

```
  38.0s   13 tests  InboxToolIntegrationTest
  25.0s    1 test   CrossHubPatchAtomicityTest
   1.2s    7 tests  PatchWorkspaceAckTest
   0.5s    3 tests  AgentChatClientDeadlockTest
   0.0s    2 tests  DelegationDeadlockTest
  64.7s            TOTAL
```

Three of the five cost under 2 s combined — isolating them is free. The CI split puts
`InboxToolIntegrationTest` in part 1 (0-based sorted index 54, even) and the other four in part 2
(indices 3, 39, 41, 97), so each half carries ~27–38 s of serial bucket.

## The isolation is verified, not assumed

From the same trx (`startTime`/`endTime` per result):

```
tests=1139  wall=201.7s  sum-of-test-time=581.8s
peak concurrency = 4                      ← the opt-in is live
stress/non-stress overlapping pairs = 0   ← DisableParallelization holds
```

Peak concurrency 4 proves the project-local `xunit.runner.json` took effect; zero overlapping pairs
proves no stress test ever ran alongside another class. (The 201.7 s local wall is **not** a timing
result — a Release build of the solution graph was running concurrently. Local timing is not
evidence here anyway, for the `DOTNET_PROCESSOR_COUNT` reason in the PR body.)

## The one red shard is main's, not this branch's

`Run tests (shard 0)` failed on `Hosting.Orleans.Test.OrleansMarkdownExportTest.ExportDocx_RoundTrips_AsTypedResponse`
(60 s method timeout). Shard 0 runs no AI.Test, and this branch changes nothing outside AI.Test —
`MessageService.cs` is #1043's, whose own run ([31361525723](https://github.com/Systemorph/MeshWeaver/actions/runs/31361525723))
had **shard 0 green**. Main run [31361446933](https://github.com/Systemorph/MeshWeaver/actions/runs/31361446933)
(`bd250ef0`) failed shard 0 in the same project on a *different* test
(`OrleansSubmitFromIdleTest.SubmitFromIdle_ThreeSequential_EachReachesIdle`) — a different test each
run in one project is the live-race signature, and it is on main.

Merged current main (`df1ed76e1`) and re-pushed rather than re-running the same input.

rbuergi:
## Second CI sample — 1.15×, and it lands on the same number

Run [31364357007](https://github.com/Systemorph/MeshWeaver/actions/runs/31364357007) (head `0330e6c1`,
current main merged in). Both AI.Test shards **executed** and both are `success`.

| AI.Test half | serial: main `cf38cf6a` | serial: main `c88a019a` | PR run 1 `d3b2440d` | PR run 2 `0330e6c1` |
|---|---|---|---|---|
| part 1/2 → shard 3 | 169.4 s | 149.7 s | 125.1 s | **110.9 s** |
| part 2/2 → shard 4 | 118.8 s | 110.9 s | 113.8 s | **127.4 s** |
| **combined** | **288.2 s** | **260.7 s** | **238.9 s** | **238.3 s** |

The halves swing several seconds run to run (runner noise), but the **combined figure is stable to
0.6 s across two runs**: 238.9 s and 238.3 s, mean **238.6 s** against a two-sample serial mean of
274.5 s.

**1.15×.** Lower than the 1.30× #1043 measured for the full opt-in, which is the expected price of
giving five classes the box — and the full opt-in was red. Across the two runs that is **2,278
AI.Test executions under parallelism with zero failures**.

`shard-assign.sh` weight left at 263.

## CI is red on shard 0, and it is main's

`Run tests (shard 0)` fails on `MeshWeaver.Hosting.Orleans.Test`. Shard 0 runs no AI.Test, and this
branch changes nothing outside AI.Test. **A different test fails every run, including on main:**

| run | branch | failing test |
|---|---|---|
| [31361446933](https://github.com/Systemorph/MeshWeaver/actions/runs/31361446933) | **main** `bd250ef0` | `OrleansSubmitFromIdleTest.SubmitFromIdle_ThreeSequential_EachReachesIdle` |
| [31363873251](https://github.com/Systemorph/MeshWeaver/actions/runs/31363873251) | **main** `df1ed76e1` | `OrleansBrokenNodeTypeAccessTest.Subscribe_ToLayoutAreaOf_NonCompilingNodeTypeInstance_SurfacesError_NotWedge` |
| 31362879486 | this PR `d3b2440d` | `OrleansMarkdownExportTest.ExportDocx_RoundTrips_AsTypedResponse` (60 s method timeout) |
| 31364357007 | this PR `0330e6c1` | `OrleansDelegationTest.Resubmit_AfterDelegation_DoesNotDeadlock` |

Four runs, four different tests, one project — and main's own tip run is red the same way (it also
reds shard 2 on `Hosting.Monolith.Test.PandasExplorerLayoutAreaTest`). #1043's run, carrying the same
`MessageService.cs` change this branch does, had **shard 0 green**, so that change is not implicated
either.

Worth noting for whoever picks it up: `Hosting.Orleans.Test` is one of only two projects that
**already** ships the parallel opt-in (#979) — `parallelizeTestCollections: true`,
`maxParallelThreads: 4` — and it has no stress collection. A deadlock test, an export round-trip on a
60 s bound and a "reaches idle" sequence failing one-per-run is the same shape this PR isolates for
AI.Test. Filed separately rather than widened here.

`Plugins compile against this PR` also fails (`Store/Catalog` tests RED). It is not a required check,
it fails on #1043 too, and it is about plugin content, not this diff.

**Not merging.** The gate is a green CI and it is main that is red.

Issue #1081, opened 2026-08-10
Hosting.Orleans.Test reds shard 0 with a different test every run — on main too


rbuergi:
`Hosting.Orleans.Test` reds shard 0 on a **different test every run**, on main and on PRs alike. Four consecutive runs, four different tests, one project:

| run | branch | failing test |
|---|---|---|
| [31361446933](https://github.com/Systemorph/MeshWeaver/actions/runs/31361446933) | main `bd250ef0` | `OrleansSubmitFromIdleTest.SubmitFromIdle_ThreeSequential_EachReachesIdle` |
| [31363873251](https://github.com/Systemorph/MeshWeaver/actions/runs/31363873251) | main `df1ed76e1` | `OrleansBrokenNodeTypeAccessTest.Subscribe_ToLayoutAreaOf_NonCompilingNodeTypeInstance_SurfacesError_NotWedge` |
| [31362879486](https://github.com/Systemorph/MeshWeaver/actions/runs/31362879486) | PR #1075 | `OrleansMarkdownExportTest.ExportDocx_RoundTrips_AsTypedResponse` — `Test execution timed out after 60000 milliseconds` |
| [31364357007](https://github.com/Systemorph/MeshWeaver/actions/runs/31364357007) | PR #1075 | `OrleansDelegationTest.Resubmit_AfterDelegation_DoesNotDeadlock` |

**Main is currently red on this**, so every open PR inherits it and every PR author spends a cycle deciding whether the failure is theirs. That misattribution cost is the reason to file rather than shrug: PR #1075 changes nothing outside `MeshWeaver.AI.Test`, and shard 0 runs no AI.Test at all.

## The lead worth checking first

`MeshWeaver.Hosting.Orleans.Test` is one of only **two** projects that already ship the per-project parallel opt-in from #979 — `test/MeshWeaver.Hosting.Orleans.Test/xunit.runner.json` sets `parallelizeTestCollections: true`, `maxParallelThreads: 4` — and it has **no serial collection for its machine-hungry classes**. The only collection it defines is `OrleansClusterCollection` (the shared cluster fixture).

The failure shapes fit contention rather than a functional defect:

- `Resubmit_AfterDelegation_DoesNotDeadlock` — a deadlock detector,
- `ExportDocx_RoundTrips_AsTypedResponse` — dies on a 60 s **method timeout**, not an assertion,
[...]

PR #1086, opened 2026-08-10
The settle gate read the record one way, the code acting on it another

rbuergi:
Closes nothing yet. This is the #1081 investigation. It **refutes the premise the next session was
told to build on**, ships a diagnostic so the next occurrence is attributable, and fixes one real
defect on the failing edge that the traces show is **not** what failed. Every claim below is backed
by a captured trace; I have been deliberate about which parts are proven and which are not.

## 1. The premise is wrong: the request DOES reach the grain

PR #1073 concluded, from a silo log reading `Node created → CreateNodeResponse → 45 s of complete
silence → TimeoutException`, that "the request never reaches the grain". With
`MESHWEAVER_MSG_TRACE=1` (the env-gated tracer already in `RoutingGrain`) the failing run reads:

```
07:36:10.214  OrleansRoutingService.Deliver DISPATCH_TO_GRAIN  addr=type/…/broken-sub-instance
07:36:10.217  RoutingGrain.RouteMessage RESOLVE_EMIT  remainder=(null)  →  GRAIN_CALL
07:36:10.258  hub=type/…/broken-sub-instance  InitializeHubRequest   ← the grain ACTIVATED
07:36:10.415  hub=type/…/broken-sub-instance  SubscribeAck
07:36:10.444  hub=type/…/broken-sub-instance  DataChangedEvent → client/broken-sub-…
   … 160 DataChangedEvents reach the client inside the first second …
07:36:55.6    === TEST FAILED: The operation has timed out.
```

Routing, activation, subscribe, ack and the render round-trip complete in **~260 ms**. The failing
edge is the CONTENT of the render, not its delivery.

### 🚨 The general rule this exposed: silo-side `ILogger` output is unreliable in Orleans tests

Two independent reasons, and together they mean "silence" carries no information here:

1. `DynamicCompilationSiloConfigurator` never wired `AddXUnitLogger` (#1073's finding; that
   one-liner is included here).
2. **Even with it wired, silo-side records are still dropped.** `AddXUnitLogger()` registers a
   *fresh* `TestOutputHelperAccessor` in the silo's own container and nothing ever populates it —
   `ServiceSetup.SetOutputHelper` only touches the TEST's provider. So every silo record falls
   through to `XUnitFileOutputRegistry.GetAnyActiveOutputHelper()`, an **`AsyncLocal`**, empty on
   any thread whose ExecutionContext does not descend from the test class's constructor: the
   Orleans grain scheduler, timers, and pool threads.

Proof: with the logger wired, a file trace records the enrichment running at `07:56:56.876` —
squarely inside the "45 s of silence" — while the silo log for that window contains two lines,
neither of them the `LogInformation` calls that ran.

**So: diagnose this path with a file-based trace, never `LogInformation`.** This is the third
instance of one defect class (an `AsyncLocal` empty on a different scheduler): #1005 (`TestBase`
registering from an `async InitializeAsync`, dropping every `ILogger` record of every
`MonolithMeshTestBase` test), #997 (duplicate appsettings copy), and now this one.

## 2. It reproduces locally — a correction to the brief

Whole project, one process, Release, on a dev Mac — **5 red of 9 runs**, ~2 min each, victims
rotating exactly as on CI:

```bash
dotnet build test/MeshWeaver.Hosting.Orleans.Test -c Release -warnaserror
MESHWEAVER_MSG_TRACE=1 DOTNET_PROCESSOR_COUNT=4 \
  dotnet test test/MeshWeaver.Hosting.Orleans.Test -c Release --no-build --no-restore
```

Traces land in **`$TMPDIR`, not `/tmp`**, on macOS. Victims seen locally:
`OrleansBrokenNodeTypeAccessTest.Subscribe_ToLayoutAreaOf_…` ×2, `OrleansMarkdownExportTest.ExportPdf`
×2, both `SubHub_…` methods, `OrleansDelegationStartTest.Delegation_CreateCellsThenThread…`,
`OrleansSubThreadAutoResumeTest.SubThreadDelegation_LiveProgress…`, `OrleansMeshTests.HubWorksAfterDisposal`.

## 3. The witness: what the failing page actually contained

The 45 s bound is **unchanged** — it is the detector — but it now leaves a witness. A reproduction
with it armed prints:

```
WEDGE WITNESS: 6 store emission(s) arrived and none rendered the compilation-error overlay.
Last store: data.progress = "Rendering Overview... awaiting first data"
            areas         = { "$Menu", "$Menu:Node", "$Menu:Mesh" }   ← no "Overview" key at all
```

The hub activated on the **correct** overlay configuration (enrichment trace: `EXIT
COMPILE_ERROR_OVERLAY`, `status=Error`), the `Overview` area is registered and seeded its
placeholder, the client is subscribed and receiving frames — and the `Overview` area never appears
in any of them. (My first reading of this was "the generator never delivered"; §6 shows that is
wrong — the generator delivered and the SERVER pushed. The witness is the client half of the
evidence; §6 adds the server half.)

## 4. What is in this PR

**(a) A real defect, fixed — explicitly NOT #1081's cause.** The compile-settle predicate in
`BuildSlowPath` and `IsRecompileSettled` decided on `Content is NodeTypeDefinition` (a CLR type
test) while `ApplyStreamResult`, which acts on whatever they admit, reads the same node with
`ContentAs<>`. When the mirror holds Content un-materialized — the normal shape off a sync stream
or right after creation — every guard was skipped and the trailing escape admitted the emission as
settled; `ApplyStreamResult` then deserialized the same node and acted on a state the gate exists
to reject (`CompilationStatus` null/Unknown → `ApplyDefaultConfig`, binding the instance hub to the
mesh DEFAULT config for the grain's whole lifetime, since enrichment binds once and never
re-observes). `CompileSettlePredicateTest` pins it: **5 cases red against the previous predicate,
green after**, verified by reverting only the fix, re-running, and restoring. The failing run took
the *correct* branch, so this is not #1081.

**(b) A diagnostic for the never-delivering area — and it is a diagnostic, not a bound.** An area
whose generator neither emits nor faults keeps its "awaiting first data" placeholder forever
(nothing here is bounded, deliberately — a legitimately slow area must never be cut short) and the
server logs *nothing*. That is a wedges-to-zero violation in its own right, and it is why four
sessions could not see this. `LayoutAreaHost` now counts the render results an area actually
delivered and reports zero at the two moments knowable **without a clock**: a terminal notification
having pushed nothing (Error — provably stuck), and disposal having pushed nothing (Warning — a
subscriber navigating away early is legitimate). Nothing is bounded, cancelled, retried, or turned
into an error that wasn't one; a timeout here would swap one silence for another and leave the
generator unfixed.

> Trap for whoever builds on this: the render observable completing *after* it has emitted is the
> normal shape (`OnCompleted` follows `OnNext`). The signal is the **count of pushed results**,
> never the completion itself — a diagnostic keyed on "completed" reports every healthy area.

**(c) The silo logger** (`DynamicCompilationSiloConfigurator`, per #1073) and **the witness-on-timeout**.

## 5. Ruled out, with the evidence

- **Routing / activation** — trace in §1; ~260 ms end to end.
- **Cold Roslyn compile** — the victim reaches its final wait ~5 s in, having already observed
  `CompilationStatus.Error`.
- **The enrichment settle gate (this PR's own fix)** — the failing run took `EXIT
  COMPILE_ERROR_OVERLAY`. Fixed anyway, on its own merits, with its own pin.
- **`host.Localize` / the AsyncLocal `AccessContext.Locale` hop** — traced the overlay's view
  factory: `OVERLAY-FACTORY-ENTER → OVERLAY-FACTORY-LOCALIZED (2 ms) → PUSH area=Overview`. It
  neither throws nor stalls, the chain is null-safe end to end (`GetService`, `AccessService?`,
  `Locales.Resolve(null)` → English), and `LayoutAreaHost` **captures the subscriber's
  `AccessContext` and restores it on the render scope** — the mechanism AGENTS.md's rule
  prescribes, working as designed.
- **`LayoutDefinition.RenderingGate`** — its "Still-Running" branch renders `Observable.Empty` and
  would produce exactly this witness, but nothing in `src/` installs one (`gate=none` in every
  trace line).
- **`Layout` `IIoPool` exhaustion** — cap 256, permit released in a `finally`.
- **`maxParallelThreads: 4` / #1010** — not re-tested; #1073's controlled before/after already
  refuted it and nothing here contradicts that. Not reverted.

## 6. Where the defect actually is — measured on both sides of the hop

Caught with the render path probed (5th local reproduction). The server side is **completely
healthy** and identical to a passing run:

```
08:29:48.713  INIT area=Overview  gate=none
08:29:48.714  DEFER-SUBSCRIBED → SAFERENDER-ENTER
08:29:48.715  SAFERENDER-EXIT null=False
08:29:48.716  OVERLAY-FACTORY-ENTER
08:29:48.718  OVERLAY-FACTORY-LOCALIZED  label=View full compile log →
08:29:48.771  PUSH area=Overview     ← the overlay IS merged into the stream, 58 ms end to end
```

And the client, over the following 45 s:

```
WEDGE WITNESS: 7 store emission(s) arrived and none rendered the compilation-error overlay.
Last store: data.progress = "Rendering Overview... awaiting first data"
            areas         = { "$Menu", "$Menu:Node", "$Menu:Mesh" }   ← no "Overview" key
```

Seven frames land, carrying the `$Menu*` areas written to the **same stream**, through the **same
serialized queue**, in the **same 60 ms window**. The frame carrying `Overview` (and the
progress-clear `PushRenderResult` performs) never does.

| hop | evidence | verdict |
|---|---|---|
| client → RoutingGrain → grain activation | `DISPATCH_TO_GRAIN → GRAIN_CALL → SubscribeAck`, ~260 ms | healthy |
| NodeType enrichment → hub configuration | `EXIT COMPILE_ERROR_OVERLAY`, `status=Error` | healthy |
| layout render of `Overview` | `SAFERENDER → OVERLAY-FACTORY-LOCALIZED → PUSH`, 58 ms | healthy |
| server stream → client mirror | 7 frames land; the Overview frame never does | ❌ **the wedge** |

This is a **mirror frame-loss** — a known defect class here ("the owner never loses a write, the
READER does", #945: a monotonicity/version guard on the mirror dropping a frame the owner
committed). I am **not** claiming #945 specifically, and not naming the guard: I have not traced
the frame's version stamps, and inventing the mechanism is exactly the trap this issue has sprung
five times. What is claimed is the boundary, and it is measured on both sides.

**Next step, concretely:** instrument the frame across that hop — the `ChangeItem`'s
`Version`/`ChangeType` at `PushRenderResult` on the owner, and what the client's
`JsonSynchronizationStream` does with the corresponding `DataChangedEvent` (applied, or dropped by
a version gate). One trace line each side closes it.

Note the new zero-render diagnostic would **not** have fired here, correctly: the server did render
and did push. Its silence is itself informative — it proves the server side completed its job. It
closes the sibling hole (a generator that delivers nothing), which is a real wedges-to-zero
violation on its own terms.


rbuergi:
## The witness fired — and it names the gap exactly

Reproduced again locally with the diagnostic from this branch in place (4th run of the batch). The
failure is no longer "The operation has timed out." It is:

```
WEDGE WITNESS: 6 store emission(s) arrived and none rendered the compilation-error overlay.
Last store (3739 chars):
{"$type":"MeshWeaver.Data.EntityStore",
 "data":  {"progress": {"message":"Rendering Overview... awaiting first data"}},
 "areas": {"$Menu": …, "$Menu:Node": …, "$Menu:Mesh": …}}
```

Read the `areas` map: `$Menu`, `$Menu:Node`, `$Menu:Mesh` — and **no `Overview` key at all**. The
area is registered, `LayoutAreaHost.WriteFrameworkMilestones` has written its
`"Rendering Overview... awaiting first data"` placeholder, and then **`UpdateArea` never fires for
it** — for the full 45 s, and (nothing bounds it) presumably forever. `FailRendering` did not fire
either: a faulted render would have put an error control in `areas`, and there is none.

So the wedge is now stated as narrowly as the evidence allows, and it is NOT any of the five
mechanisms proposed so far:

> The per-instance hub activates on the compile-error overlay configuration (trace:
> `EXIT COMPILE_ERROR_OVERLAY`), the `Overview` area is registered and seeds its progress
> placeholder, the client is subscribed and receiving store frames — and the area's view generator
> never delivers a first emission, nor an error. The area sits on "awaiting first data" with
> nothing to time it out.

That is also exactly the shape of the sibling victims — `SubHub_…ResolvesExportDocumentControlViaRegistryChain`
waits for a control at an area pointer that never renders; the export round-trips wait for a
response produced downstream of a render. "Different test every run" is which area happened to be
the one that never delivered.

### Where to look next (not yet established — do not treat as diagnosis)

The overlay's view is `Observable.Return(BuildCompilationErrorMarkdown(…, host.Localize("ui.viewCompileLog")))`
built inside the `(host, ctx) =>` factory in `NodeTypeEnrichmentHelpers.CreateCompilationErrorConfiguration`.
An `Observable.Return` cannot fail to emit, so the candidates are (a) the factory itself never runs,
or (b) it throws while being constructed — `host.Localize` is evaluated eagerly there and resolves
off `AccessContext.Locale`, which is AsyncLocal and the render subscribe deliberately hops threads
(`LayoutAreaHost.ScheduleRenderSubscribe` → the `Layout` `IIoPool`). Either way the framework's
answer is the same and is the actual defect to fix: **an area whose generator neither emits nor
faults leaves the page on "awaiting first data" with no bound and no diagnostic.** I did not
establish which of (a)/(b) it is, and I am not going to guess it — the witness now lands that
evidence on the next occurrence.

The `Layout` pool is not the constraint, for the record: cap 256, permit released in a `finally`
(`IoPool.SubscribeThroughPool`).

rbuergi:
## `host.Localize` is NOT the cause — refuted with a render trace

I named it and was asked to chase it. Chased; it is wrong, and I would rather say so than
let it become the sixth mechanism. Both sub-cases were asked for, and both are answered.

**It does not throw.** I traced the overlay's view factory itself. In a healthy render of the
very same compile-error overlay, on the very same hopped thread:

```
08:22:11.069  SAFERENDER-ENTER          hub=type/OrleansBrokenSub…/broken-sub-instance area=Overview
08:22:11.070  OVERLAY-FACTORY-ENTER     hub=…/broken-sub-instance area=Overview
08:22:11.072  OVERLAY-FACTORY-LOCALIZED label=View full compile log →      ← 2 ms, no throw
08:22:11.107  PUSH                      area=Overview                      ← the overlay lands
```

**It cannot return something that stalls the generator either.** The generator is
`Observable.Return(...)` — it cannot fail to emit once constructed — and the whole resolution
chain is null-safe by construction, so there is no null-deref to find:

- `LayoutAreaLocalizationExtensions.Localize` resolves `AccessService` with **`GetService`**
  (null-tolerant), and `LocalizeExtensions.Localize` is declared on `AccessService?`.
- `ResolveViewerLocale` reads `accessService?.Context?.Locale`, falls back to
  `accessService?.CircuitContext?.Locale`.
- `Locales.Resolve(null)` falls back to English; the catalog returns the key when a string is
  missing. Worst case is English text, never an exception.

**And the AGENTS.md rule I invoked is already satisfied structurally, not by luck.** The rule
("a layout-area render hops the hub scheduler and an ambient AsyncLocal does not survive it —
resolve explicitly off `AccessContext.Locale`") is *why this works*:
`LayoutAreaHost.BuildInitialization` **captures the subscriber's `AccessContext` at construction
and restores it on the render scope** (`accessService?.SetContext(capturedAccessContext)`)
before the render subscribe, clearing it on teardown. That capture-and-restore is exactly the
mechanism the rule prescribes, and the pool hop the rule warns about is covered by it. My
suspicion was that the hop defeated the read; the trace shows the design holds.

I also ruled out the other structural candidate on the same path while I was there:
`LayoutDefinition.RenderingGate` — whose "Still-Running" branch renders
`Observable.Empty` and would produce precisely this witness — is **not installed by anything**
in `src/` (`gate=none` in every trace line). Not it either.

### What that leaves, stated exactly

The healthy sequence is `SAFERENDER-ENTER → OVERLAY-FACTORY-ENTER → OVERLAY-FACTORY-LOCALIZED
→ PUSH area=Overview`, all inside ~40 ms. In the failing case one of those steps is missing,
and which one is now a single line of trace away. The probe is written and the repro loop is
hunting for it; I have not yet caught the canonical victim with it armed, and I am not going to
name the step before I see it.

### One trap worth recording for whoever gets there

`RENDER-COMPLETED-NO-EMISSION` fires on **healthy** renders too — `OnCompleted` follows
`OnNext` on this pipeline. So the defect signal is the **count of pushed results**, never the
completion itself; a diagnostic keyed on "completed" alone reports every area in the mesh. The
committed diagnostic counts (see 870171fcf).

rbuergi:
## Caught it: the server RENDERS the overlay and PUSHES it — the client's mirror never gets that frame

Fifth local reproduction, this time with the render path probed. The server side is **completely
healthy**, and identical to a passing run:

```
08:29:48.713  INIT                       hub=type/OrleansBrokenSub…/broken-sub-instance area=Overview gate=none
08:29:48.714  DEFER-SUBSCRIBED
08:29:48.714  SAFERENDER-ENTER
08:29:48.715  SAFERENDER-EXIT            null=False
08:29:48.716  OVERLAY-FACTORY-ENTER
08:29:48.718  OVERLAY-FACTORY-LOCALIZED  label=View full compile log →
08:29:48.771  PUSH area=Overview         ← the compile-error overlay IS merged into the stream
```

58 ms, start to finish. `PushRenderResult` ran: the overlay control is in the server's
`EntityStore`, and that same call clears the progress marker.

Now the client, over the following 45 s:

```
WEDGE WITNESS: 7 store emission(s) arrived and none rendered the compilation-error overlay.
Last store: data.progress = "Rendering Overview... awaiting first data"
            areas         = { "$Menu", "$Menu:Node", "$Menu:Mesh" }   ← no "Overview" key
```

Seven frames arrive. They carry the `$Menu*` areas — which were written to the **same stream**,
through the **same serialized update queue**, in the **same 60 ms window**. The frame carrying the
`Overview` area and the progress-clear does not, ever.

### So the defect is in the sync path, not the render and not the routing

Restating the whole chain with what is now measured at each hop:

| hop | evidence | verdict |
|---|---|---|
| client → RoutingGrain → grain activation | `DISPATCH_TO_GRAIN → GRAIN_CALL → InitializeHubRequest → SubscribeAck`, ~260 ms | healthy |
| NodeType enrichment → hub configuration | `EXIT COMPILE_ERROR_OVERLAY`, `status=Error` | healthy — correct config |
| layout render of `Overview` | `SAFERENDER → OVERLAY-FACTORY-LOCALIZED → PUSH area=Overview`, 58 ms | **healthy — content produced** |
| server stream → client mirror | 7 frames land; the Overview frame never does | ❌ **the wedge is here** |

That is a *mirror frame-loss*, and it is a known defect class in this repo rather than a new one:
"the owner never loses a write, the READER does" (#945 — a monotonicity/version guard on the mirror
dropping a frame the owner had committed). I am **not** claiming #945 specifically or naming the
guard: I have not traced the frame's version stamps, and inventing the mechanism here is exactly
the trap this issue has sprung five times. What I am claiming is the boundary, and it is measured
on both sides.

### Why every previous theory pointed elsewhere

Because both ends *look* fine and every one of them was, in fact, fine. The grain activates, the
config is right, the area renders, data flows continuously — the client even keeps receiving
frames throughout the 45 s. Only one specific frame is missing, and the only artifact that could
ever have shown it is a side-by-side of the server's push and the client's store, which nothing
produced until now.

### Next step, concretely

Instrument the frame itself across the boundary — the `ChangeItem`'s `Version`/`ChangeType` at
`PushRenderResult` on the owner, and what the client's `JsonSynchronizationStream` does with the
corresponding `DataChangedEvent` (applied, or dropped by a version gate). One trace line on each
side of that hop closes it. Both bounds involved are ≥ 60 s while the tests' detectors are 30–45 s,
which is why this always surfaces as the test's own timeout and never as a framework diagnostic.

### What is already committed here

- The **witness** on the 45 s bound (bound unchanged — it is the detector) that produced the client
  half of the evidence above.
- The **zero-render diagnostic** in `LayoutAreaHost` — note it would *not* have fired for this
  case, and correctly so: the server DID render and DID push. That diagnostic closes the sibling
  hole (a generator that delivers nothing), which is a real wedges-to-zero violation on its own,
  and its absence here is itself informative: it proves the server side completed its job.
- The settle-gate fix, still explicitly not this bug.

rbuergi:
Checked this rather than acting on it, and the `using` is required — leaving it in place.

`OrleansDynamicCompilationTest.cs:397` calls `.ConfigureLogging(logging => logging.AddXUnitLogger())`. `AddXUnitLogger` extends `ILoggingBuilder`, which lives in `Microsoft.Extensions.Logging` — so the directive at line 17 is what makes that line resolve. Removing it would break the build rather than avoid a break.

The empirical check agrees: this PR's CI is `COMPLETED/SUCCESS` with `Build solution (once)` green under `-c Release -warnaserror`, which is exactly the configuration CS8019-as-error would fire in. An unused directive could not have compiled there.

Worth noting the wider context, since it is what the directive is for: adding `AddXUnitLogger` to `DynamicCompilationSiloConfigurator` is the substance of the fix. Without it the silo had no logger wired, so **every CI log of this failure class contained two `WriteLine`s and a timeout** — which is how four separate investigations theorised about a "silent silo" without ever seeing a silo trace.

Resolving as verified-not-a-defect.

Issue #369, opened 2026-07-08
Delegation sub-thread token usage is recorded under model (unknown) — sub-thread cost is unpriced ($0)


sierragolflima:
## Summary

When an agent delegates work to a worker (a **sub-thread**), the sub-thread's `TokenUsage` satellite records `model: (unknown)` instead of the model that actually ran. Because cost is derived from `ModelPricing` keyed by model, `(unknown)` matches no price → the sub-thread's tokens are reported at **$0** and are invisible in per-model roll-ups. The parent thread records its model correctly; only delegated work is affected.

## Observed (two independent runs, 5 sub-threads — 5/5 affected)

**Run 1** — orchestrator on a premium model, two delegations:

| Thread | `_Usage` key (model) | Priced? |
|---|---|:--:|
| Parent (orchestrator) | `Provider/…/<real-model-A>` | ✅ |
| Worker sub-thread (write) | **`(unknown)`** | ❌ $0 |
| Worker sub-thread (verify) | **`(unknown)`** | ❌ $0 |

**Run 2** — orchestrator on a mid-tier model, three delegations:

| Thread | `_Usage` key (model) | Priced? |
|---|---|:--:|
| Parent (orchestrator) | `Provider/…/<real-model-B>` | ✅ |
| Worker sub-thread (notify) | **`(unknown)`** | ❌ $0 |
| Worker sub-thread (write) | **`(unknown)`** | ❌ $0 |
| Worker sub-thread (correction) | **`(unknown)`** | ❌ $0 |

Every sub-thread carries **real, non-zero token counts** but model `(unknown)` — across different orchestrator models and different tasks.

## Impact

- Sub-thread token **cost is reported as $0** (no `ModelPricing` entry matches `(unknown)`).
- Per-model usage **roll-up is broken** for all delegated work.
- The headline cost/usage figure **undercounts** any delegation-heavy task twice over: sub-thread usage lives on a separate `_Usage` node *and* is unpriced. In Run 2 above, roughly 13% of the run's total tokens were priced at $0.

## Root cause (hypothesis)

[...]

Issue #364, opened 2026-07-08
Content collection route does not URL-decode path segments — files/folders with spaces (or other encoded characters) return "not found"


sierragolflima:
## Summary

When browsing a content collection in the portal, clicking a file or folder whose name contains a space (or any other character the browser percent-encodes) produces a "not found" error. The collection browser lists such items correctly, but opening them fails.

## How to reproduce

1. On any node that has a content collection (for example the default `content` collection), add a file whose name contains a space, such as `my report.md`. A folder with a space in its name, such as `Data Extraction`, reproduces the same way.
2. Open the collection browser in the portal at `/{node}/content`. The item `my report.md` (or the `Data Extraction` folder) is listed correctly, with its space intact.
3. Click the item. The browser navigates to `/{node}/content/my%20report.md` (the space percent-encoded as `%20`).
4. Observe the error instead of the file content or folder listing.

## Actual behavior

The page renders:

```
'my%20report.md' not found in collection 'content'.
```

Note the literal `%20` in the message — the encoded form is compared directly against the collection's real item names.

## Expected behavior

The file renders (or the folder opens) as it does for items whose names contain no encoded characters. Percent-encoded URL segments are decoded before they are matched against the collection's stored item names.

## Root cause

The content-collection layout areas take the path from the raw URL id (`host.Reference.Id`) and use it as a collection path **without URL-decoding it**. The stored item name contains a real space (`my report.md`), while the path from the URL contains `my%20report.md`, so the case-insensitive equality match never succeeds and the fallback "not found" branch is returned.

The only decodes in the resolution chain are `WorkspaceReference.Decode` (which reverses the custom `%9Y`→`.` escaping, not percent-encoding) and query-string parsing — neither touches `%20` in the path. The shared file renderer `ContentLayoutArea.RenderFile` re-encodes the path with `Uri.EscapeDataString` when it builds the static URL, which confirms the internal contract expects an already-decoded path; the callers violate that contract.

This is not related to nesting depth. It occurs whenever any path segment contains a character the browser percent-encodes; names without such characters resolve correctly, which is why most navigation works.

## Affected code

- `src/MeshWeaver.ContentCollections/CollectionNamedLayoutArea.cs` — `RenderCore`, where `path` is derived from `host.Reference.Id` and matched against `item.Path`.
[...]

PR #1098, opened 2026-08-10
Docs accuracy (2): deleted services, a menu that never existed, and twelve stale FromAsync comments

rbuergi:
Second pass of the non-Architecture docs accuracy audit (the first landed as #1090). Every claim below was checked against the code it describes; a page that turned out accurate was left alone and is listed at the bottom.

## Claims that would cause a defect if believed

- **`MessageHubConfiguration.WithInitialization` and `MeshHubBuilder.WithInitialization`** told callers to "wrap a Task-returning method via `Observable.FromAsync(() => method())`". That observable is subscribed on the hub's **init turn**, so following the advice runs the method's synchronous prologue on the hub scheduler, unbounded — the exact deadlock the rule exists to prevent. `IEmailSender`, `IStorageAdapter` and `StorageAdapterChangeFeedExtensions` made the same recommendation to implementers; all four shipped implementations use a bounded `IIoPool`.
- **`MeshPlugin.md`** did not say `Delete` is recursive. It is — deleting a parent removes every descendant. This repo has already had a node-menu delete wipe a partition root.
- **`Observables.md`** taught an `async` view delegate as the way to load data, contradicting its own later section banning `await` on the render path.
- **`CRUD.md` / `CreatingNodeTypes.md`** documented a `Task`-returning `WithInitialData` overload. No such overload exists on any builder; `CreatingNodeTypes.md` contradicted itself forty lines later ("never `async`/`await`/`Task<T>`").

## Dead references

- **`IMeshExportService` / `IMeshImportService`** (NodeOperations.md) — deleted in the persistence cull 2026-05-12, documented with `GetRequiredService<T>()` samples that throw. Replaced with `MeshOperations.Export`/`.Import`, and the archive shape corrected (`manifest.json` + `files/`, not a `.md`/`.cs`/`.json` tree).
- **`DefaultMenuProvider`** and its **"Actions" sub-menu** (NodeMenu.md) — neither exists. `DefaultNodeMenuProvider` emits a flat, `_separator`-grouped list; every row of the reference table was wrong on order, permission or menu.
- **`IChatClientCatalog`** (ModelProviderSettings.md) — exists nowhere; the real call is `ProviderModelLister.ListModels`.
- **`NodeTypeWithNuGetCompilationTest.cs`** (NodeTypes/Testing.md) — no such file; the MathNet coverage is `NuGetAssemblyResolverTest`.
- **GitHub issues #369 and #364** (AgentDesign.md) — both closed as COMPLETED. #364's caveat still told readers to avoid non-ASCII upload filenames; that shipped 2026-07-08.
- **Twelve `Observable.FromAsync` doc comments** across `src/` describing code that uses `FromViewBuilder`, `IIoPool.Run/Invoke`, `BoundLeg(OnThreadPool(...))`, or no Task bridge at all. `LayoutAreaHost` alone had five, one calling `FromAsync` "the only sanctioned Task→observable seam".

## Behaviour described wrongly

- **`source:activity` vs `source:accessed`** — ToolsReference.md described `source:activity` as "ordered by user's last access time". That is `source:accessed` (the caller's visit log); `source:activity` is the change feed. `QuerySource.Activity`'s own doc comment also claimed it "implies nodeType:Activity filter", which no backend does.
- **`ModelProviderSettings.md`** was a design spec whose "Current state" said the CLI connect flow "is not yet implemented — no session manager, no login-status check, no UI". All of it shipped.
- **`ToolsReference.md`** satellite table named `comments`, `approvals` and `tracking` tables. None exist — all three share `annotations`.
- **`MeshPlugin.md`** listed `RenderArea` (an `McpMeshPlugin` tool) in the in-portal agent's tool set, and omitted `Patch`/`EditContent`, which the plugin's own tool descriptions say to prefer over `Update`.
- **`ChatCommands.md`** declared `SkillActionKind Kind` as `required`; it defaults to `Pick`, so an `action:` block without a `kind:` silently becomes a Pick.

## Flagged, not fixed (code defects, not doc defects)

- **`SynchronizationStream.Initialize` calls `Observable.FromAsync`** (`SynchronizationStream.cs:1187`) — a real occurrence outside `IoPool`, on the stream-init path. The doc comment describes it accurately, so the comment is not the bug.
- **`ImportNodesRequest` has no handler.** The types are declared and type-registered, but nothing handles the request, so `hub.Observe<ImportNodesResponse>(...)` never answers. Documented as such rather than presented as the API.
- **The in-portal Export download is still stubbed** (`NodeExportView`) — the menu action opens the area, the download reports the persistence-cull migration.
- **`AgentPluginReference.Methods` is a mutable `List<string>`**, against the immutable-collections policy.
- `AGENTS.md` names a `LayoutAreaPlugin` that does not exist in `src/`.

## Verified accurate — left alone

`ModelTiers.md` (tier ids, ranks, aliases, Auto's `order:-10`/`isRouter` dispatch and the whole `ModelTierSource` chain match `ModelTierDefaults`/`ModelTierCatalog`/`BuiltInLanguageModelProvider`), `Skill/thread.md`, `DataBindingMaui.md`, `QuerySyntax.md`'s `source:` section, the `AzureAIS__*` latent-bug notes in `ProviderConfiguration.md`/`ModelProviderSetup.md` (the AppHost really does emit keys nothing binds), every `*Control` type named across the GUI docs, and every other test file named in these docs.

## Verification

- `dotnet build -c Release -warnaserror` clean on `MeshWeaver.Graph`, `MeshWeaver.AI`, `MeshWeaver.Mesh.Contract`, `MeshWeaver.Hosting.PostgreSql`, `MeshWeaver.Documentation`.
- `DocumentationLinkIntegrityTest` — passed (1/1).
- `WhatsNewEntryIntegrityTest` — passed (1/1).
- `DocExecutableBlocksTest` — passed (54/54). This matters: the NodeMenu.md live cell was rewritten from a hand-built markdown table to a `DataGridControl`, and that block is executed through a real kernel by this test.

The 27-page duplicate-H1 pattern was observed and deliberately left for the restructure pass.


rbuergi:
Good catch — fixed in `060f18698`.

`MeshQuery`'s class doc did still say *"source:activity implies nodeType:Activity filter"* after `ParsedQuery` was corrected, which is exactly the failure mode this pass exists to remove: fixing one site and leaving its twin to keep teaching the falsehood.

Both are now stated as what they are — **JOIN-backed ORDERING sources, neither implying a nodeType**: `source:activity` orders by activity recency, `source:accessed` JOINs `UserActivity` to order by last-access time.

I also swept `src/` for the phrasing to check there was no third site. These two were the only ones, so the claim is gone rather than relocated. `MeshWeaver.Hosting` builds clean under `-c Release -warnaserror`.

PR #1102, opened 2026-08-10
"Area not found" right after a compile is a transient frame, not a verdict

rbuergi:
## The flake that redded three PRs in one day

The plugin gate intermittently failed with:

```
RED Store/Catalog: compile=Ok render=ok tests=FAILED
    No renderer is registered for area `Tests` on hub `Store`.
```

`AreaProbe` latched the **first** classifiable frame of the Tests-area sync stream (`Take(1)`). Right after a (re)compile the instance hub re-registers its layout, so the stream legitimately serves a frame in which the type's custom areas don't exist *yet* — only the defaults. Latching that frame produced a verdict claiming the tests failed **when they never ran**.

The window only widens under CI runner load: **16 straight local runs** — macOS and Linux, CPU-constrained, the exact vital-set input — could not reproduce it. It was misattributed twice before the CI log's own detail settled it: first to a core diff whose files the tester never loads, then to "timing-sensitive tests" that were in fact never executed.

## The fix — the platform's own rule

*Wait on the actual condition via the stream.* A not-found frame records transient detail and the pipeline keeps waiting for a frame that carries the area. The existing `Timeout` stays the backstop, upgraded to **name the last transient state** — so a genuinely absent Tests area still fails with the precise reason rather than a generic "no verdict". Same treatment for the default-area probe (same latch). Real render errors and ❌ rows stay immediately terminal — transience applies to not-found only.

## Tests

The pipeline is now a public seam over `IObservable<JsonElement>` (`ClassifyTestsFrames`), driven with synthetic frames:

- not-found → green table ⇒ **Passed** (the regression pin)
- only not-found ⇒ timeout naming `Area not found`
- ❌ row ⇒ fails immediately
- no frames ⇒ generic no-verdict

4 new; 22/22 in the project; Release `-warnaserror` clean.

No What's New entry — CI tooling only, no user-visible portal change.

### DIFF
--- a/test/MeshWeaver.PluginTester.Test/AreaProbeTest.cs
+++ b/test/MeshWeaver.PluginTester.Test/AreaProbeTest.cs
@@ -1,0 +1,92 @@
+using System.Reactive.Linq;
+using System.Reactive.Threading.Tasks;
+using System.Text.Json;
+using MeshWeaver.PluginTester;
+using Xunit;
+
+namespace MeshWeaver.PluginTester.Test;
+
+/// <summary>
+/// The Tests-area verdict pipeline, driven with synthetic frames through the
+/// <see cref="AreaProbe.ClassifyTestsFrames"/> seam.
+///
+/// <para>The one behaviour these tests exist to pin: <b>an "Area not found" frame is transient,
+/// never terminal</b>. Right after a (re)compile the instance hub re-registers its layout, so the
+/// sync stream legitimately serves a frame in which the type's custom areas do not exist yet.
+/// Latching that frame as a verdict turned the re-registration window into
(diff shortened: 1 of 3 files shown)


rbuergi:
Second commit: the honest timeout from the first commit exposed the real defect — the shipped Tests host activates mid-import, before the type's release exists, and never rebinds (`Tests area never became available within 120s`; `render=ok` was masked by the framework's own `Catalog` area). Recycle-then-probe was tried and reverted: it races subscribe against teardown on the raw stream, which bypasses the cache that absorbs exactly that race. Landed design: the probe always runs on a fresh `GateProbe` instance — first access is the probe's own subscription, post-compile, so it binds the verified release by construction. Local: 22/22 + vital-set e2e ALL GREEN.

PR #1103, opened 2026-08-10
The plugins gate blamed a compile break for a failure that compiled fine

### LINKED ISSUE #1077: The plugins gate reports 'does not compile — a public API changed' when nothing failed to compile
Two separate problems, one of which sends investigators to the wrong place.

## 1. The failure message misattributes its own cause

The gate fails with:

> `MeshWeaver.Plugins does not compile against this PR — see the 'Plugins compile against this PR' job. A public API changed…`

while the job log says the opposite:

```
RED Store/Catalog: compile=Ok render=ok tests=FAILED
    **Area not found**
    No renderer is registered for area `Tests` on hub `Store`.
GATE FAILED — 1 new failure(s), 0 stale allow entr(ies).
```

**`compile=Ok`. Nothing failed to compile.** The failure is a `Tests`-area render.

This is not a cosmetic complaint. A public-API break in this repo genuinely does break the Plugins repo — plugin node source is Roslyn-compiled from mesh content, so a repo-local `rg` can never prove a deletion safe (that is how `AddSkillSource` broke 8 NodeTypes). So the message names the **most plausible and most expensive** wrong cause, and an investigator will go diff public signatures that are fine. The message should reflect which phase actually failed: compile / render / tests.

## 2. The underlying failure looks like the instance-vs-type-hub quirk, not plugin rot

**A NodeType's `Tests` layout area is served by INSTANCE hubs, not the type node** — the type node's hub is the NodeType editor (`AddNodeTypeView`). So probing a *type* path for `area/Tests` always answers "Area not found"; that is by design. The gate is supposed to find a shipped instance (`nodeType:{typePath}`) or create a throwaway probe instance under the type path.

Here the probe landed on hub `Store` — a type path — and got the expected "Area not found". Note the available-areas list in the log contains `Catalog`, `Chat`, `Settings`, … but no `Tests`, which is exactly what a type-node hub looks like.

**It is intermittent, which points at the probe path rather than at the plugin:** on `main`, run `31361180771` = **success**, `31361446933` = **failure**, `31361055089` = **success**. Same plugins checkout, same node, different outcome. If `Store/Catalog` simply had no shipped instance, it would fail every time.

Worth checking whether the throwaway-probe creation can fail or race and then fall back to probing the type hub, which would produce exactly this signature.

## Why it matters now

It currently reds PRs whose diffs cannot reach it — observed on #1043 (three files: `MessageService.cs` and two test files) with the identical `Store/Catalog` signature, at the same time main was red on it. Every such occurrence costs an investigation that starts from a false lead supplied by the error message itself.

## Suggested fix order

1. Make the failure message name the failing **phase** (and the node), not a guessed cause. Cheap, and stops the wrong-lead problem immediately.
2. Then diagnose the intermittent probe-instance path.

rbuergi:
Fixes #1077

## 1. The message named a cause the report contradicted

The gate judges three checks per NodeType — **compile**, **render**, **tests** — and its own log says which failed:

```
RED Store/Catalog: compile=Ok render=ok tests=FAILED
    **Area not found**
    No renderer is registered for area `Tests` on hub `Store`.
```

CI annotated that as *"MeshWeaver.Plugins does not compile against this PR … A public API change here breaks plugin node source that is Roslyn-compiled at runtime."* One message for all three checks, asserting a cause the report contradicts — and the **most expensive wrong lead available**, because it is plausible: a public-API break here genuinely does break that repo (plugin node source is Roslyn-compiled from mesh content, so a repo-local `rg` can never prove a deletion safe — that is how `AddSkillSource` broke 8 NodeTypes). So the claim reads as authoritative and the investigator goes to diff public signatures. It did exactly that on #1043, whose whole diff was `MessageService.cs` and two test files.

Now the tester emits one verdict line naming the failing check and the node(s):

```
GATE FAILED — tests: Store/Catalog — 1 new failure(s), 0 stale allow entr(ies).
```

CI lifts that line **verbatim** — it does not re-derive or re-word a cause — and the public-API sentence is kept, **conditional on the phase actually being `compile`**. Two failures previously reported as "0 new failure(s)" now name themselves: a fatal error says `fatal: …`, a stale allow entry says so. A run that fails with nothing recorded says `no failing check was recorded` rather than picking a phase.

`GateSummaryTest` pins each of these **two-sided**: the line must name the check that failed *and* must not name one that did not — a message that can drift back is the thing being fixed. CI's grep is anchored on `^GATE FAILED` (the `GateReport.FailedPrefix` constant, ASCII), so re-punctuating the line cannot silently unhook the annotation.

## 2. The `Store/Catalog` failure is a framework defect, not plugin rot — and not the probe path

The hypothesis in the issue (the probe landed on a *type* path, where "Area not found" is by design) does not hold. `Store` is the Store package's **root**, retyped to `Store/Catalog` by the installer — a correctly-typed instance. The evidence is the available-areas list the failure printed: item for item it is `ConfigureDefaultNodeHub`'s set — `AddDefaultLayoutAreas` (incl. `AddEducationLayoutAreas`' CourseNav/Learn/StartExercise/GoToMyCopy and `AddDomainLayoutAreas`' Catalog/Details/DataModel) plus Invite. **Not one area from `Store/Catalog`'s compiled configuration, and not the Space placeholder's either.** Hub `Store` was serving the mesh **default** configuration: the known partition-root defect (#1055, and 2026-07-29 before it).

It recurred on a commit that already carried #1055's fix (run `31361446933` at 06:18Z; #1055 merged 04:39Z). **NOT cached is not NOT pinned**: a hub is pinned by address in `Mesh.GetHostedHub` and never re-reads its node's NodeType, so fixing the *resolution* does not un-pin the hub a bad resolution already activated. The only thing that does is `PackageInstaller`'s `DisposeRequest` — fire-and-forget, and conditional on the placeholder dance having run. That residual is **not fixed here** (it needs a repro and a design decision, not a guess); it is written into both comment blocks that carry this ledger, which now ask three questions instead of two.

The gate now prints **which** node hosted the Tests run and how it was picked (an instance already in the mesh vs. a throwaway probe it created) — the two are not equivalent, and that difference is what separates the same commit's green and red runs. That line is what makes the third question answerable from CI alone. The host resolution is also **bounded** now: it queries the index and may create a node, neither with a budget of its own, so a wedge there spent the whole job timeout and reported nothing.

## Not done

No `continue-on-error`, no retry around the probe, no widened timeout, no allow-list entry for `Store/Catalog`. The gate still fails exactly when it failed before — it just says what failed.

## Verification

- `MeshWeaver.PluginTester.Test`: 28/28 pass (12 new).
- `mw-plugin-test` run end-to-end against an empty root: emits `GATE FAILED — fatal: …`, exit 1 — the line CI greps, produced by the real binary.
- `-c Release -warnaserror` clean: `MeshWeaver.PluginTester`, `MeshWeaver.PluginTester.Test`, `MeshWeaver.PluginCatalog`, `MeshWeaver.Hosting`.

### DIFF
--- a/.github/workflows/dotnet-test.yml
+++ b/.github/workflows/dotnet-test.yml
@@ -784,6 +784,12 @@ jobs:
     # bar. The timeout is deliberately well below the shard job's 20 min — if the gate ever becomes
     # the long pole it should fail and be re-scoped, not silently stretch every PR.
     timeout-minutes: 12
+    # The failing CHECK and the node(s) it failed on, lifted verbatim from the tester's verdict
+    # line, so the required check downstream can name the real cause instead of guessing one.
+    # Empty whenever the gate died before producing a verdict (checkout fault, missing tester) —
+    # consumers must degrade to "see the job", never to an invented cause.
+    outputs:
+      failure: ${{ steps.gate.outputs.failure }}
     steps:
     - uses: actions/checkout@v6
       with: { fetch-depth: 1 }
@@ -829,6 +835,7 @@ jobs:
         ssh-key: ${{ secrets.PLUGINS_REPO_SSH_KEY }}
         fetch-depth: 1
     - name: "Compile + run the VITAL plugins against this PR"
+      id: gate
       env:
         # 🚨 The vital organs only — NOT the whole repo. This gate sits in the PR critical path,
         # so it buys the most breakage-detection per minute rather than maximum coverage:
@@ -865,7 +872,40 @@ jobs:
         # plugin-gate.allow is the plugins repo's known-debt ratchet: listed failures are
         # tolerated, NEW failures fail this PR. Same one-way contract the plugins CI uses, so the
         # two gates cannot disagree about what is acceptable.
-        time dotnet "$TESTER" "$STAGE" --allow "$STAGE/plugin-gate.allow"
+        #
+        # 🚨 The run is TEED, not swallowed: the log still streams to the job, and the copy is read
+        # afterwards for the tester's verdict line. `if !` (not `set +e`) so `set -e` stays armed
+        # for everything else, and PIPESTATUS[0] is the TESTER's code — tee always exits 0.
+        log="${RUNNER_TEMP:-/tmp}/plugin-gate.log"
+        status=0
+        if ! time dotnet "$TESTER" "$STAGE" --allow "$STAGE/plugin-gate.allow" 2>&1 | tee "$log"; then
+          status="${PIPESTATUS[0]}"
+          # tee always exits 0, so a non-zero pipeline with a zero head means tee itself failed.
+          if [ "$status" = "0" ]; then status=1; fi
+        fi
+        # The tester's verdict line IS the message — `GATE FAILED — tests: Store/Catalog — 1 new
+        # failure(s), …` (GateReport.FailedPrefix). Lifted verbatim: nothing here re-derives or
+        # re-words a cause, which is exactly how this job came to annotate a `tests` failure as
+        # "does not compile — a public API changed" and send investigators to diff public
+        # signatures that were fine (#1077).
+        verdict="$(grep -m1 -- '^GATE FAILED —' "$log" || true)"
+        # Multi-line-safe even though the verdict is one line: a bare KEY=VALUE with an unexpected
+        # newline would corrupt every later output in the file.
+        {
+          echo "failure<<GATE_EOF"
+          echo "$verdict"
+          echo "GATE_EOF"
+        } >> "$GITHUB_OUTPUT"
+        if [ "$status" != "0" ]; then
+          if [ -n "$verdict" ]; then
+            echo "::error::$verdict"
+          else
+            # No verdict = the gate died before judging anything (missing tester, install fault,
+            # crash). Say THAT — an unexplained failure is cheaper than a wrong explanation.
+            echo "::error::The plugins gate failed before it produced a verdict — no check was judged. See this job's log."
+          fi
+          exit "$status"
+        fi
 
   collect-results:
     name: "Consolidate test results"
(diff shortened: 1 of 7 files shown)


rbuergi:
## This PR's own CI red is the proof

The `Plugins compile against this PR` job failed on this run — on the pre-existing `Store/Catalog` defect, which this diff cannot reach (all 6 test shards pass; the diff is the tester's reporting, the workflow's annotation, and three comment blocks). That is exactly the phenomenon #1077 describes: *"it currently reds PRs whose diffs cannot reach it."*

**Before**, that would have been annotated as:

> MeshWeaver.Plugins does not compile against this PR … A public API change here breaks plugin node source that is Roslyn-compiled at runtime.

**What this run actually printed** ([job](https://github.com/Systemorph/MeshWeaver/actions/runs/31383870908/job/93441658490)):

```
##[error]GATE FAILED — tests: Store/Catalog — 1 new failure(s), 0 stale allow entr(ies).
```

And the new host line settles the second half of the issue outright:

```
RED Store/Catalog: compile=Ok render=ok tests=FAILED
    Tests host: Store — an instance of Store/Catalog already in the mesh (1 candidate(s)); its hub may predate this install
    **Area not found**
    No renderer is registered for area `Tests` on hub `Store`.
```

Every other type in the same run reports its host too, and the pattern is unambiguous:

| Type | Host | Result |
|---|---|---|
| `Store/Catalog` | `Store` — instance already in the mesh | **FAILED** |
| `Store/Enrollment` | `Store/Enrollment/GateProbe` — probe the gate created | 13/13 passed |
| `Store/Provision` | `Store/Provision/GateProbe` — probe the gate created | 9/9 passed |
| `Store/Plugin` | `GateEnroll120df…` — instance already in the mesh | 290/290 passed |

So it is **not** "the probe landed on a type path" (the hypothesis in the issue), and **not** "a pre-existing instance is bad" either — `Store/Plugin` used one and passed. It is specifically the **partition root** `Store`, whose hub was activated during the install/provisioning window and never re-binds, serving `ConfigureDefaultNodeHub`'s area set instead of `Store/Catalog`'s. Same defect as #1055, still live after it, and the residual is written up in `PathResolutionService` and `PackageInstaller`.

Not fixed here, deliberately — un-pinning an already-activated hub needs a repro and a design decision, not a guess. This PR makes the failure legible enough that the next occurrence starts from the right place.

rbuergi:
## Rebased on main — the gate that blocked this PR is now green

`Plugins compile against this PR: **pass** (59s)`. #1143 fixed #1104 (a per-node hub binds `HubConfiguration` from its NodeType once at activation and is then pinned by address, since `RoutingServiceBase.RouteInMesh` short-circuits on `Mesh.GetHostedHub`; a rebind watcher off `IMeshChangeFeed` now un-pins it). That is the residual this PR documented as open — the two comment ledgers now point at the resolution instead of posing the question.

### Merge resolution (3 conflicts)

**`PathResolutionService` + `PackageInstaller`** — the comment ledgers, merged rather than picked. Main's account is the correct spine; what this branch contributed and main did not is the **fingerprint**: the available-areas list being exactly `ConfigureDefaultNodeHub`'s set (`AddDefaultLayoutAreas` + Invite) and not one area of the node's real type. That is what lets someone *recognise* the defect rather than re-derive it, so it is folded in. The "check THREE things … is the live hub the one activated inside the provisioning window?" framing is deleted — #1143 answered it.

**`PluginGateRunner`** — main's design wins outright. #1102 deleted the shipped-instance preference: the probe host is now ALWAYS a freshly created `GateProbe`, which binds the just-verified release by construction. `ResolveTestsHost` and its two-branch "shipped vs probe" description are obsolete and gone with it.

Kept on top, because neither is about branch selection:
- the host still goes into the report — with one branch it no longer discriminates a cause, but `No renderer … on hub X` is still unreadable without knowing which node X was, and that ambiguity is precisely what made the original RED get read as "the probe landed on a type path";
- probe creation is still **bounded** — `CreateNode` is a round-trip with no budget of its own, and unbounded it spends the whole 12-min job timeout reporting nothing.

Also drops a stale duplicate `<summary>` on `CreateTestsProbe` left by #1102 (the old "Prefers an instance the package ships" text survived above the new one, contradicting it), and corrects this branch's own test fixture, which still fed the summary a shipped-instance description the runner can no longer produce.

### Verification

33/33 tester tests. Both two-sided cases — `TestsFailure_NamesTests_AndNeverCompile` and `TestsFailure_WithAllowlist_NamesTests_AndNeverCompile` — confirmed `outcome="Passed"` **by name** in the trx, with both assertion directions present in source: the line must name the check that failed AND must not name one that didn't. `-c Release -warnaserror` clean.

### Remaining red is main's, not this branch's

`Run tests (shard 4)` fails on `StaleStampRootBindingTest`, added by #1146. It fails on main's own runs at `09cd35739` and `269dcbc76`, and reproduces locally (`Failed: 1, Passed: 2`). This diff cannot reach it — its only `PackageInstaller.cs` change is a comment block. Filed as #1151 with the fault site (`Workspace.SubscribeToClient` throws inside the `SubscribeRequest` handler; the handler's error arm posts nothing, so the requester burns its full 60s budget).

**Not merging** — the gate is absolute and `Consolidate test results` is red on main's regression.

rbuergi:
⚠️ **The previous conflict resolution silently reverted #1153 — I've restored it in `5b0686a09`.**

The commit *"Merge origin/main — keep both ledger accounts, adopt main's probe design"* got `PluginGateRunner.cs` right (main's `CreateTestsProbe` is the correct side — a shipped instance's hub activates mid-import, before the type's compile produced its release, and never rebinds). But it resolved `PackageInstaller.cs` in the **branch's** favour, and this branch predates #1153:

- `SettleRetypedRoot` — #1153's ordered recycle-and-wait — was **absent from the merged result entirely**.
- The write-time fire-and-forget `hub.Post(new DisposeRequest(), …)` that #1153 **removed** was back.

#1153 is what greened main earlier today. Merging that resolution would have re-red main on **both** halves of `StaleStampRootBindingTest` — the exact test main had just been rescued from.

**The tell, for next time:** `git diff origin/main HEAD --stat` read `PackageInstaller.cs | 158 +++------` on a PR whose only intent in that file was to enrich a comment. A negative delta that large in a file you meant to *add* to is the signature of a resolution that swallowed someone else's fix. `git merge-base --is-ancestor origin/main HEAD` also returned false after a commit titled "Merge origin/main", meaning an older main was merged.

**Nothing of this branch's own work was dropped.** The gate ratchet (`GateAllowlist`/`GateVerdict`), `Headline`, the `GateSummaryTest` suite, the workflow wiring, and the gate-run `31361446933` fingerprint comments in **both** `PackageInstaller.cs` and `PathResolutionService.cs` are all preserved — layered on top of main's mechanism rather than instead of it.

`MeshWeaver.PluginCatalog` and `MeshWeaver.PluginTester` both build clean under `-c Release -warnaserror`.

Issue #1144, opened 2026-08-10
LogIncident filer re-files on every recurrence — the incident never records its issue


rbuergi:
## What is failing

The first live run of the log watcher (2026-08-10 12:03–12:17Z) filed **37 issues for ~12 distinct defects**. Two separate mechanisms:

1. **Re-filing on recurrence.** The same incident filed repeatedly, minutes apart — `ROUTER_TRAFFIC` filed **8 times** (#1113, then #1117, #1118, #1121, #1136, #1138, #1139, #1140 at 12:06:59 → 12:13:28). The chain: a recurrence bumps the incident → re-triage → the agent drafts → `RequestedStatus = File` → the filer opens a **new** issue. Nothing on the incident records the already-opened issue, and nothing in the File path checks for one.
2. **Same-second doubles.** #1124/#1125 and #1111/#1112 were each created within one second — either two fingerprints for one burst or a race in the filer allowing two concurrent File requests for one incident.

## Probable cause

`LogIncident` has no `issueUrl`/`issueNumber` written back after filing, so the File handler is not idempotent. For (2), the File handler likely doesn't claim-before-filing (cf. the claim-the-guard-BEFORE-the-mutation pattern).

## Impact

The alert channel is GitHub notifications; every duplicate is a false notification. First sweep: ~25 junk issues (closed by hand). The watcher is scaled back to **0 replicas** until this is fixed — the alerting rollout is blocked on it.

## Where to look

- The filer in `MeshWeaver.Observability` — where `RequestedStatus = File` is consumed and the issue opened: write `issueNumber` back onto the incident in the same update, and refuse to file when it is already set (a recurrence should **comment or bump** the existing issue, or stay silent — design choice).
- The re-triage path from #1094's reconcile: a `Failed` incident that recurs re-enters triage; after this fix it must land on "already filed" rather than a fresh draft.
- The same-second doubles: make the File consumption claim the incident before opening the issue.

## What went RIGHT (do not regress it)

Filing stopped by itself when the catch-up drained (bounded, no infinite storm); every issue was App-attributed with the evidence footer; the drafts are real defect descriptions, several immediately actionable. The identity dedup at the *incident* level held — 602 occurrences of ROUTER_TRAFFIC remained ONE incident; the duplication is in the *filing*, not the fingerprint.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1149, opened 2026-08-10
A claim that never reached GitHub parked the incident in Filing forever

rbuergi:
Follow-up to #1144 / #1147. Not a duplicate of the idempotency fix — it closes the dead end that fix introduced.

## What is failing

#1147 made `Filing` a **written** status: the claim stamps it before the GitHub round-trip so a second work item can tell a mid-file incident from a mid-triage one. That closed the duplicate correctly. It also created a state nothing can leave.

When the claim is taken and nothing comes back — the portal dies in that window, or the write-back itself fails — the incident is left at:

```
Status = Filing,  RequestedStatus = None,  IssueNumber = null
```

and every exit is closed:

| Candidate | Why it does not fire |
|---|---|
| the claim | keys on `RequestedStatus`, which the claim itself already cleared to `None` |
| `NeedsTriageReconcile` | requires `Status == Triaging`; `Filing` is not covered |
| `NextRequest` → `{ IssueNumber: not null }` | there is no link |
| `NextRequest` → `New or Failed` | the status is `Filing` |
| `NextRequest` → `_` | returns `RequestedStatus`, i.e. `None` |

So the incident counts occurrences forever and is never ticketed — invisible and un-ticketed, which is exactly what `Triaging` used to do before #1094 (nineteen incidents accumulated that way on memex.systemorph.com), now in a status the repair does not cover. `LogIncidentControlPlane.InFlightStatus`'s doc comment already claims this case is recoverable "via the stranded-triage reconcile" — for `File` it is not, because that reconcile only reads `Triaging`.

## The fix

A recurrence re-asks for `File` when the incident is `Filing` **with no issue link**.

```csharp
// Filing with no link ⇒ the claim was taken and nothing came back. Re-ask.
{ Status: LogIncidentStatus.Filing } => LogIncidentRequest.File,
```

It sits *below* `{ IssueNumber: not null }`, so it only ever fires for an incident that has no ticket. And re-asking is safe **because of** #1147's claim: a `File` request arriving on an incident that has since been ticketed is granted as a rate-limited `Comment`, never as a second issue. `Filing` **with** a link is not in-flight at all — the write-back lands the link before the status settles — so the issue-link rule keeps it quiet.

No timer, no watchdog, no retry loop: the trigger is the same recurrence report that already drives every other transition.

## Pinned

`LogIncidentMergeTest`:

- `Merge_ReAsksForAFilingIncidentThatNeverGotAnIssue` — the strand is re-asked.
- `Merge_DoesNotReAskOnceTheFilingProducedItsIssue` — the same status **with** a link stays quiet.
- `Merge_LeavesAnInFlightTriageAlone` — replaces the `[Theory]` half that asserted `Filing` must be left alone (that assertion is what described the dead end). `Triaging` is unchanged: only the thread can tell "still running" from "died", so nothing may re-request there.

Negative control: removing only the new arm turns `Merge_ReAsksForAFilingIncidentThatNeverGotAnIssue` red — `Expected value to be File …, but found None`. Restored, all 106 tests in `MeshWeaver.Observability.Test` pass; `MeshWeaver.Observability` and the test project build clean at `-c Release -warnaserror`.

### DIFF
--- a/src/MeshWeaver.Documentation/Data/Architecture/LogWatchTriage.md
+++ b/src/MeshWeaver.Documentation/Data/Architecture/LogWatchTriage.md
@@ -131,8 +131,17 @@ A recurrence is therefore always folded onto the ticket that exists:
 
 The claim keys on `RequestedStatus`, never on the status, so the in-flight statuses (`Triaging`,
 `Filing`) are not dead ends: a crash between the claim and the write-back parks the incident there,
-and asking again — an admin in the incident view, or the stranded-triage reconcile below — is
-honoured.
+and asking again is honoured. **Something has to do the asking**, though, and that differs by status:
+
+- `Triaging` — the stranded-triage reconcile below, which asks the thread whether the round is over.
+  Only the thread can tell "still running" from "died", so nothing may re-request on a timer.
+- `Filing` **with no `IssueNumber`** — the next recurrence re-requests `File` (`NextRequest`). The
+  claim was taken and nothing came back, and unlike `Triaging` there is no in-mesh authority to ask:
+  the only record of whether the issue was opened is GitHub. Re-asking is safe because the claim
+  itself is the guard — if the incident has since been ticketed, the request is granted as a
+  `Comment`, never as a second issue.
+- `Filing` **with** an `IssueNumber` is not in-flight at all: the write-back lands the link, so that
+  state is a completed file whose status simply settled, and the issue-link rule keeps it quiet.
 
 ### The one state nothing requests: `Triaging`
 
--- a/src/MeshWeaver.Observability/LogIncidentIngestService.cs
+++ b/src/MeshWeaver.Observability/LogIncidentIngestService.cs
@@ -181,6 +181,17 @@ public sealed class LogIncidentIngestService(
     /// and the agent's fresh draft then asked to <c>File</c> again. That chain is how
     /// <c>ROUTER_TRAFFIC</c> was filed EIGHT times in seven minutes on the watcher's first live run.
     /// Once an issue exists, a recurrence belongs on it, never in a new one.</para>
+    ///
+    /// <para>🚨 <b><c>Filing</c> without an issue is the one in-flight status that is re-asked.</b>
+    /// The claim writes <c>Filing</c> BEFORE the GitHub round-trip, so a portal that dies in that
+    /// window — or a write-back that fails — leaves the incident at <c>Filing</c> with no link.
+    /// Nothing else can move it: the claim keys on <c>RequestedStatus</c>, which is already
+    /// <c>None</c>; the stranded-triage reconcile only ever touches <c>Triaging</c>; and every rule
+    /// above needs either an issue or a different status. That is precisely the dead end
+    /// <c>Triaging</c> used to be — nineteen incidents accumulated there, invisible and un-ticketed —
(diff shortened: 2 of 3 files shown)


rbuergi:
CI is red, and neither failure is reachable from this diff. Recording the analysis rather than re-running.

**Shard job conclusions** (run [31392662916](https://github.com/Systemorph/MeshWeaver/actions/runs/31392662916)): shards 0, 1, 3, 5 green; shards 2 and 4 red.

| Shard | Failing test | Verdict |
|---|---|---|
| 4 | `MeshWeaver.PluginCatalog.Test.StaleStampRootBindingTest.StaleStampSelfTypedRootShippingContent_RootServesItsTypesArea` (+ `StaleStampSelfTypedRoot_…`) | **Pre-existing red on main.** The same test fails on main's own tip run [31390882509](https://github.com/Systemorph/MeshWeaver/actions/runs/31390882509) at `269dcbc7` — the commit this branch is based on. |
| 2 | `MeshWeaver.Hosting.Monolith.Test.SearchResultStormTest.CatalogSubscription_ContentUpdateStorm_AddsNoExtraRerenders` | Flake — `Expected value to be 1 …, but found 3`. Did not fail on main's tip run; the preceding main runs at `09cd3573` and `23859fe2` were also red on shard 4 with a *different* test each time (`CrossHubPatchAtomicityTest`), which is the shard flake cluster, not this branch. |

**Why neither can be this change.** The diff touches exactly one code path: `LogIncidentIngestService.NextRequest`, a pure function reached only when a red-log report is folded into an incident. `MeshWeaver.PluginCatalog.Test` and `MeshWeaver.Hosting.Monolith.Test` have **no reference to `MeshWeaver.Observability`** at all (`find . -name '*.csproj' -exec grep -l MeshWeaver.Observability {} +` → only `LogWatcher`, `Memex.Portal.Shared`, `MeshWeaver.Observability.Test`), so there is no path from either assembly to the changed code.

The branch is not stale: it is based on `269dcbc76` (main's tip) and `git merge origin/main` reports up to date, so this is not the re-sampled-flake-on-a-stale-branch case.

Not merging — the green-merge gate stands. This needs main's `StaleStampRootBindingTest` red fixed first.

rbuergi:
Addressed in `97352dfcf` — the reviewer is right, and the fix is exactly the one suggested.

The `Filing => File` arm matched on `Status` alone, so it overwrote whatever sat in `RequestedStatus`. An admin's `Suppress` — or any future control-plane verb — issued while the incident was at `Filing` would be stomped back to `File` on the next occurrence report. And because reports *repeat*, the instruction could never survive long enough to be acted on: the incident would file itself regardless of what anyone asked for.

Scoped to `RequestedStatus: LogIncidentRequest.None`. This costs the stranded-Filing repair nothing, and the PR's own doc comment already contains the argument for why:

> Nothing else can move it: the claim keys on `RequestedStatus`, **which is already `None`**

The stranded case is None by construction — the claim consumes the request before it writes `Filing`, which is precisely why the incident gets stuck and why this arm was added. So an incident at `Filing` *with* a request pending is not stranded at all; somebody is steering it. The trailing `_ => incident.RequestedStatus` now preserves that.

**Negative control**: with the arm left unscoped, `Merge_LeavesAnExplicitRequestAloneWhileFiling` fails and nothing else does (1 of 16). Restored, 16/16 pass — so the new test pins this specific behaviour rather than passing incidentally.

PR #1153, opened 2026-08-10
The installer started a teardown and handed back without waiting for it

rbuergi:
**This greens main.** `StaleStampRootBindingTest` has been red on main in *both* its cases since it
merged (#1146, 12:54 today) — the PR itself was green, on a merge-ref whose tree main had already
moved past. One defect, two victims, so this is one fix rather than two.

## The defect

`UpsertNodes` recycled the retyped partition root with a fire-and-forget `DisposeRequest` at write
time. That is wrong twice over, and the two compound.

**Mistimed.** At that point the in-package NodeType still advertises the compile stamp the node repo
*committed*, so the re-activation the recycle invites binds the **fallback** configuration for the
new hub's lifetime — precisely what #1101 stopped the warm from doing, arriving by the other door.

**Outlives the install.** `Install` returned while that teardown was still running, so the next touch
of the root landed mid-teardown and got `HubDisposingException`. Both red tests are exactly that, and
the log names them:

- `cannot create '/content/'` — the package that ships assets
- `cannot create 'LayoutAreaReference { Area = Tests }'` — the one that does not

In both, the handler's reply died with the hub, so the caller had nothing to wait for and burned its
full 60 s request timeout on work that never happened. Whether a package's files arrived came down to
which side of that dead heat the machine fell on — which is why the same package installed cleanly on
one portal and half-empty on another.

## The fix

The recycle moves to where it can do its job. `SettleRetypedRoot` waits for the in-package type's
rebuild (`MayPublishIntoRoot` is that wait, and it answers promptly when the rebuild produced nothing
loadable — nothing worth rebinding to, so the recycle is skipped and the rebind watcher takes it once
the type compiles), **then** posts the `DisposeRequest`, **then** waits for the root to answer.

A `PingRequest` is the whole primitive and **ordering** is why it works: it is enqueued on the root
*behind* the `DisposeRequest`, so it cannot be answered by the hub that is going away — only by the one
that comes after, and answering it *is* that re-activation. A rejected or undelivered ping means the
teardown is still running, so it retries; the budget is bounded and its `Timeout` is a graceful sink
that lets the install finish rather than wedging. `SyncPackageContent` keeps the same wait for the call
sites that do not run `SettleRetypedRoot` — one ping when the root is already up.

## Verification

- `MeshWeaver.PluginCatalog.Test` **236/236**, from 235/1.
- The suite drops from **3 m 17 s to 1 m 19 s** — the stalls it was absorbing were these 60 s timeouts.
- `dotnet build -c Release -p:CIRun=true -warnaserror` clean.

Ruled out along the way, so it does not get re-guessed: the no-content case is **not** the NodeType-ALC
eviction of #1148 — it fails identically with that fix applied.

### DIFF
--- a/src/MeshWeaver.Documentation/Data/WhatsNew/2026-08-10-a-plugins-files-arrive-even-when-its-root-restarts.md
+++ b/src/MeshWeaver.Documentation/Data/WhatsNew/2026-08-10-a-plugins-files-arrive-even-when-its-root-restarts.md
@@ -1,0 +1,42 @@
+---
+Name: A freshly installed plugin is ready the moment the install finishes
+Category: Fix
+Description: An install could hand back a package whose root was still restarting, so the first thing to open it got nothing — a page that never rendered, or files that were never written.
+Icon: CloudArrowUp
+Order: -20260810
+---
+
+# A freshly installed plugin is ready the moment the install finishes
+
+Installing a package could leave it in a state where the very next thing to touch
+it came away empty-handed. Two ways you might have seen it: a package that ships
+files — course videos, posters, images — installed without them, after a pause of
+about a minute and a note that its binaries were not being served. Or a freshly
+installed plugin's own page came up blank, showing none of the views the package
+brought with it.
+
+## What was happening
+
+Installing a package changes what kind of thing its root is, and the portal
+restarts that root so it picks up its new behaviour. The install did not wait for
+that restart. It started it, then handed back.
+
+So everything that came next — publishing the package's files, the first person to
+open one of its pages — could arrive while the root was still on its way down.
+Nothing was written, nothing rendered, and because the reply died with the
+restarting root, whatever was waiting waited out a full minute before giving up.
+Which side of that dead heat a machine landed on is why the same package could
+install perfectly on one portal and arrive half-empty on another.
+
+The restart was also happening too early to be useful. It ran before the package's
+own code had finished building, so the root that came back could not yet see what
+the package had brought — and, having come back once, it kept that empty view.
+
+## What changed
+
+The restart now happens once the package's code is ready, and the install waits for
+the root to come back before it finishes. It asks the root a question that can only
+be answered *after* the restart, so there is no window left to race.
+
+An install therefore hands back a package that is actually ready: its files are
+published, its pages render its own views, and the minute-long pause is gone.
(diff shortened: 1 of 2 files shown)


rbuergi:
Independent verification (coordinator session driving the merge train) — ran this PR's branch merged on current main, `DOTNET_PROCESSOR_COUNT=4`, class runs of `StaleStampRootBindingTest`:

**Root cause confirmed, fix necessary — but PARTIAL: 3/5 green (2 failures, both `StaleStampSelfTypedRoot_RootServesItsTypesArea` at exactly 2m6s).** The write-time fire-and-forget `DisposeRequest` this PR removes is real and was tearing ShopA/ShopB down ~100 ms after activation in every decoded failure trace; with it gone, green runs drop from 1–2.5 min to 6–8 s. The residual mode: after install returns, the rebind recycle on the freshly landed build tears the root down under the client's `SubscribeRequest`; the transient outcome reaches the client, and then nothing ever re-probes — the change-feed latch only fires on a post-recycle node *write* (there is none) and the heartbeat is fire-and-forget, so the client parks the full 120 s frame wait.

**The missing piece is client-side** (`JsonSynchronizationStream.CreateExternalClient`): a bounded, NACK-triggered re-probe (6 attempts, 500 ms→8 s backoff, cancelled by keep-alive disposal; the re-sent `SubscribeRequest` is itself what reactivates the recycled owner — same sanctioned shape as `OrleansRoutingService`'s transient retry, not a timer watchdog). **This PR + that re-probe = 10/10 consecutive green class runs.** The 94-line patch (one file) is ready — happy to hand it over or open it as a stacked follow-up the moment this merges.

**Shard-3 AI.Test failures on this PR's run are FLAKE, not regression:** `grep -rln "PackageInstaller|AddPluginCatalog|InstallPackage" test/MeshWeaver.AI.Test/` is empty (no path from this diff to those tests); both failing classes are OUTSIDE #1075's isolation set and today (11:56Z) is their first day running parallel on a 4-vCPU runner; both classes ran 4/4 green on this branch locally. `FullCrudWorkflow`'s mode is the post-delete read-verdict race (#1029 tombstone family) — a separate pre-existing defect. Rerunning the failed jobs now.

Note for sequencing: this PR semantically conflicts with #1154 in `PackageInstaller.cs` (recycle placement). Recommendation: **this PR merges first** (its post-rebuild recycle + message-based probe is cross-silo-correct); #1154 then rebases keeping its `MessageService` seams + `DisposalRaceNackTest` and reconciling its installer half.

PR #1154, opened 2026-08-10
A hub that was going down answered nobody, so every caller waited out its full minute

rbuergi:
## Main is red

`StaleStampRootBindingTest.StaleStampSelfTypedRootShippingContent_RootServesItsTypesArea`, run 31390882509 (the #1147 merge):

```
AssertionException : Expected value to be true because the committed asset must still be
published into the root's content collection (…/ShopB/images/cover.png), but found False.
[PackageInstaller] publishing 1 content asset(s) to ShopB failed — the package's nodes are
  installed but its binaries are not being served
  System.TimeoutException: No response received in hub mesh/… within 00:01:00 for request
  SyncContentFilesRequest → target ShopB.
install returned at +60701ms
```

**Not a #1147 interaction** — `git log 09cd35739..269dcbc76` is #1147 plus one commit, touching only `MeshWeaver.Observability` and docs. The test is genuinely racy: on `origin/main`, under `DOTNET_PROCESSOR_COUNT=4` in bulk, **the class failed in 4/4 runs**. #1146's single green was luck.

## Root cause, in two halves

The fate ledger names the first: `ROUTED onTarget=True state=Submitted@ShopB(+87ms) → NOT_PROCESSED_DISPOSING runLevel=Dead@ShopB(+94ms)`. A root hub is recycled during install, and **a hub that is tearing down answers nobody.** Three seams in `MessageService`, all silent:

1. **A delivery the hub ACCEPTED whose turn comes after `ShutDown`** was returned unchanged. Neither existing NACK covers it — the intake gate never saw it, and it is in the main queue, not the deferred one the #672 drain answers for.
2. **The `HubDisposingException` branch** posted its correctly-classified transient through `ReportFailure`, which **declines to post once `RunLevel >= DisposeHostedHubs`** — precisely the state that branch exists for. Computed, logged, dropped.
3. **An outbound REPLY** was refused by `PostImpl`'s teardown guard, so a handler that ran and posted its answer still stranded the caller — *"a reply WAS posted for this correlation and the callback is STILL pending."*

All three now go through the **live parent**. That killed the 60 s park — and immediately exposed the second half: the honest *"I am recycling, ask again"* verdict was being delivered to consumers that all read it as **permanent failure**, so the fast correct answer lost the same work the silence did, just sooner. (Both variants failed on the first CI run of this branch — that is this bug, not a regression.)

4. **`SynchronizationStream`** kept a rejected subscribe alive and waited for the change-feed latch — but **a recycle is not a write**, so no mesh change event is ever published and the latch's only trigger never fires. The subscriber is orphaned until something unrelated writes the node. The rejection itself is now the event: it re-arms through the **existing** `Resubscribe` (in-flight guard, `ExpectResubscribeFull`, System impersonation), bounded so a recycle loop degrades to today's orphaned stream rather than a storm. The re-ask fetches the **current** snapshot — it never re-asserts a cached frame (the #945 shape).
5. **`ContentService`** collapsed a `HubDisposingException` into a null collection, so callers got the terminal *"Target content collection 'content' not found"* for a hub that was merely restarting. It rethrows the transient; the cache entry is evicted either way.
6. **The content-sync handler** flattened the same fault into an application-level `ImportContentResponse.Fail`, indistinguishable from a real error. It answers with the typed `ErrorType.ShuttingDown`.
7. **`PackageInstaller`**: the root recycle was fire-and-forget, so the install's own warm and content publish raced the teardown they had just asked for — now sequenced on that instance's `DisposalCompleted`. And the publish re-asks on the typed transient, gated on the rejecting instance's `DisposalCompleted`. **Only** the typed transient — an application failure still fails on the first answer.

## Evidence

| | before | after |
|---|---|---|
| PluginCatalog.Test, bulk, `DOTNET_PROCESSOR_COUNT=4` | **4/4 and 5/5 runs red** | **green** |
| `StaleStampRootBindingTest` | a 60 s park + a 120 s one | **3/3 in 6 s** |
| installer content publish | `install returned at +60701ms` | `+289ms` |
| `DisposalRaceNackTest` (new) | both seams fail with the exact CI signature | 2/2 pass, 2 min → 16 s |
| `MeshWeaver.Messaging.Hub.Test` | — | **175/175** |
| `MeshWeaver.Data.Test` | — | **305/305** |

`dotnet build -c Release -warnaserror` clean on every touched project.

Same family as today's memex-cloud production wedge (a `DeliverMessage` executing 37+ min with `NumRunning=147` behind it): a recycle mid-request orphaning in-flight deliveries.

## Note

Main is red, so **this PR's own shard result is the proof** — see the `MeshWeaver.PluginCatalog.Test` shard.

### DIFF
--- a/src/MeshWeaver.ContentCollections/ContentImportExtensions.cs
+++ b/src/MeshWeaver.ContentCollections/ContentImportExtensions.cs
@@ -57,10 +57,31 @@ public static class ContentImportExtensions
         SyncFiles(contentService, request)
             .Subscribe(
                 count => hub.Post(ImportContentResponse.Ok(count), o => o.ResponseFor(delivery)),
-                ex => hub.Post(ImportContentResponse.Fail(ex.Message), o => o.ResponseFor(delivery)));
+                ex => hub.Post(FailureFor(delivery, ex), o => o.ResponseFor(delivery)));
 
         return delivery.Processed();
     }
+
+    /// <summary>
+    /// The answer to send when a sync failed.
+    ///
+    /// <para>🚨 A hub-disposal fault is TRANSIENT and must NOT be flattened into
+    /// <see cref="ImportContentResponse.Fail"/>. This hub is being recycled — the collection could
+    /// not even be created (<c>ContentCollection.CreateStream</c> cannot host its
+    /// <c>SynchronizationStream</c> once creation is frozen) — and the node is coming straight
+    /// back. Reported as an application failure, the caller can no longer tell "your request is
+    /// malformed" from "ask me again in a moment", so it gives up on work that would have
+    /// succeeded: the plugin installer declared a package's committed binaries lost and the assets
+    /// were never served (StaleStampRootBindingTest, the test that turned main red). Answering with
+    /// the typed <see cref="ErrorType.ShuttingDown"/> hands the caller the framework's own verdict
+    /// — "the address may reactivate; retry to get the authoritative answer" — which it can act on.
+    /// The reply still reaches the sender because <c>MessageService</c> forwards a correlated reply
+    /// through the live parent when this hub can no longer post it itself.</para>
+    /// </summary>
+    private static object FailureFor(IMessageDelivery delivery, Exception exception)
+        => HubDisposingException.IsHubDisposal(exception)
+            ? new DeliveryFailure(delivery, exception.Message) { ErrorType = ErrorType.ShuttingDown }
+            : ImportContentResponse.Fail(exception.Message);
 
     /// <summary>
     /// Writes each inline file under <c>TargetPath</c> (binary-safe — the bytes are streamed straight
--- a/src/MeshWeaver.ContentCollections/ContentService.cs
+++ b/src/MeshWeaver.ContentCollections/ContentService.cs
@@ -230,8 +230,30 @@ public class ContentService : IContentService
             CreateCollection(config)
                 .Catch((Exception ex) =>
                 {
-                    logger.LogWarning(ex, "Creating content collection '{Collection}' failed", name);
                     collections.TryRemove(name, out _);
+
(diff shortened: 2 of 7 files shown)


rbuergi:
Independent verification (coordinator session driving the merge train):

**The three `MessageService` seams are genuine product defects and the fixes are sound** — independently found and pinned two of them before reading this PR (deterministic repro: stall the victim's action block → watchdog advances RunLevel out of band; red-before 2/2 with each sender burning its full 60 s on the exact `NOT_PROCESSED_DISPOSING runLevel=Dead` / `HANDLER_FAULT HubDisposingException` signatures; green-after 2/2 in 16 s). The third seam (correlated replies forwarded through the parent) checks out: `PostImpl`'s teardown guard provably kills any self-post from `DisposeHostedHubs` on.

**But the PR is insufficient for the StaleStamp red, and its own CI shows it** (run 31403272620, shard 4: BOTH variants failed — the no-content variant in the 120 s client ride-out mode, the content variant still dropping `cover.png`). The installer half here sequences the recycle on a local `GetHostedHub(...)` instance (null cross-silo — Monolith-only) and neither half closes the client-side gap: after a rebind recycle with no post-recycle node write, nothing ever re-probes, so the client parks the full 120 s (measured on three builds). The missing piece is a bounded NACK-triggered re-probe in `JsonSynchronizationStream.CreateExternalClient` — patch ready, see the verification comment on #1153 (that combination is 10/10 green).

**Semantic conflict with #1153** in `PackageInstaller.cs` (recycle placement — `git merge-tree` confirms). Recommendation: let #1153 merge first (its post-rebuild recycle + message-based probe is cross-silo-correct), then rebase this PR **keeping the `MessageService` seams + `DisposalRaceNackTest`** and reconciling/dropping the installer half.

Issue #1104, opened 2026-08-10
A hub activated on the wrong NodeType stays wrong — fixing resolution does not un-pin it


rbuergi:
Found while fixing #1077. **This is the defect that intermittently reds PRs whose diffs cannot reach it**, and it currently blocks PR #1103.

## What happens

The plugins gate fails on `Store/Catalog` with `compile=Ok render=ok tests=FAILED`, *"No renderer is registered for area `Tests` on hub `Store`"*.

The decisive evidence is the available-areas list in the failure: it is, **item for item, `ConfigureDefaultNodeHub`'s set** (`AddDefaultLayoutAreas` + Education + Domain areas + Invite) — and **not one area from `Store/Catalog`'s compiled configuration**. Hub `Store` was serving the mesh **default** configuration rather than its own.

Two hypotheses are ruled out by the same run, which printed which node hosted each Tests run:

| Type | Host | Result |
|---|---|---|
| `Store/Catalog` | `Store` — instance already in the mesh | **FAILED** |
| `Store/Enrollment`, `Store/Provision` | `…/GateProbe` — probe the gate created | passed |
| `Store/Plugin` | `GateEnroll120df…` — instance already in the mesh | passed |

- **Not "the probe landed on a type path."** `Store` is the Store package's **root**, retyped to `Store/Catalog` by the installer — a correctly-typed instance.
- **Not "pre-existing instance vs fresh probe."** `Store/Plugin` used a pre-existing instance and passed.

## Why #1055 did not close it

It recurred on a commit that **already carried #1055's fix** (run `31361446933` at 06:18Z; #1055 merged 04:39Z).

**Not cached is not not-pinned.** A hub is pinned by address in `Mesh.GetHostedHub` and **never re-reads its node's NodeType**. So repairing the *resolution* path does nothing for a hub that a bad resolution has already activated — the wrong configuration is latched for the hub's lifetime. The only existing thing that un-pins it is `PackageInstaller`'s `DisposeRequest`, which is **fire-and-forget** and conditional on the placeholder dance having run.

## Why it deserves its own issue

- **It is intermittent and cross-cutting**: main's gate was green at 10:53Z and the same plugins checkout failed at 11:39Z. It reds PRs by lottery, and the failing PR's diff is irrelevant — one victim's diff was three files (`MessageService.cs` plus two tests).
- **It is a framework defect, not a plugins defect.** A hub serving the default configuration for a correctly-typed node is wrong regardless of who installed the node.
[...]

PR #1159, opened 2026-08-10
A stray recycle killed the hub the install was still using

### LINKED ISSUE #1151: StaleStampRootBindingTest reds main: SubscribeRequest handler faults and never answers, requester stalls 60s
`StaleStampRootBindingTest` (added by #1146) fails on main and reds every PR since.

## Evidence it is main, not any branch

| commit | run | shard 4 |
|---|---|---|
| `09cd35739` (#1146's own merge) | [31390266052](https://github.com/Systemorph/MeshWeaver/actions/runs/31390266052) | **both** tests FAIL |
| `269dcbc76` (#1147, main tip) | [31390882509](https://github.com/Systemorph/MeshWeaver/actions/runs/31390882509) | `…ShippingContent_RootServesItsTypesArea` FAILS |
| PR #1103 merged with `269dcbc76` | [31395351596](https://github.com/Systemorph/MeshWeaver/actions/runs/31395351596) | **both** tests FAIL |

`MeshWeaver.PluginCatalog.Test`:
- `StaleStampRootBindingTest.StaleStampSelfTypedRoot_RootServesItsTypesArea`
- `StaleStampRootBindingTest.StaleStampSelfTypedRootShippingContent_RootServesItsTypesArea`

Which of the two fails varies; at least one fails every run. **Reproduces locally** on `269dcbc76`: `Failed: 1, Passed: 2, Total: 3` in 2m 6s (`dotnet test test/MeshWeaver.PluginCatalog.Test --filter "FullyQualifiedName~StaleStampRootBindingTest"`).

## The failure

The visible symptom is a 60s request timeout, but that is the **downstream** effect — the handler faults and never answers:

```
System.TimeoutException : No response received in hub client/e841599f2320 within 00:01:00
for request SubscribeRequest (id=Frf4NBvPNkiXt0pU74K38w) → target ShopA.
```

The message-flow trace names the real event:

```
SubscribeRequest: … → ROUTED onTarget=True state=Submitted@ShopA(+96ms)
                    → HANDLER_ENTER@ShopA(+99ms)
                    → HANDLER_FAULT TargetInvocationException@ShopA(+100ms)
  ⇒ the chain FAULTED and no reply was posted — the fault is the cause;
    find why its error arm does not answer the requester.
```

Fault site, from the local run:

```
at InvokeStub_Workspace.SubscribeToClient(Object, Span`1)
--- End of inner exception stack trace ---
at MeshWeaver.Data.Workspace.MeshWeaver.Data.IWorkspace.SubscribeToClient(IMessageDelivery`1 delivery)
     src/MeshWeaver.Data/Workspace.cs:650
at MeshWeaver.Data.DataExtensions.<HandleSubscribeRequest>b__0(DataValidationResult validationResult)
     src/MeshWeaver.Data/DataExtensions.cs:1660
at MeshWeaver.Data.DataExtensions.HandleSubscribeRequest(IMessageHub hub, IMessageDelivery`1 request)
     src/MeshWeaver.Data/DataExtensions.cs:1644
```

So: `SubscribeToClient` throws inside the `SubscribeRequest` handler on the `ShopA` hub, the reflective invoke wraps it as `TargetInvocationException`, and the handler's error arm posts nothing — the requester then burns its full 60s budget. Two things worth separating: **why `SubscribeToClient` throws for this node**, and (independently, per the trace's own hint) **why a faulted handler does not NACK the requester** — the second turns any such fault into a 60s stall instead of an immediate, legible error.

## Possibly related, from the same run's output

```
[Warning] [ObjectPolymorphicConverter] Received '$type':'FrontContent' which is NOT registered
in this (receiving) hub — it can only be read as an untyped JsonElement (the value renders
empty / reactive waits time out).
```

Per AGENTS.md → "Type-registry mismatch", the fix shape for that warning is
`WithType(typeof(FrontContent), nameof(FrontContent))` on the receiving hub. Unverified as the
cause here — flagging it because "reactive waits time out" is exactly the observed shape.

## Why it matters

Main is red, so it reds every PR built on it, and each occurrence costs an investigation that
starts on the wrong branch. It also blocks the merge gate (`Consolidate test results` is the
required check) and therefore the pull-based self-update.

Found while verifying #1103 (the gate-message fix for #1077), whose own diff cannot reach this —
its only change to `PackageInstaller.cs` is a comment block.

rbuergi:
Fixes #1151. **main is red on this** — `StaleStampRootBindingTest` fails on main's own runs at `09cd35739` and `269dcbc76`, and reproduces locally.

## What was wrong

`PackageInstaller` posted a fire-and-forget `DisposeRequest` to un-pin the retyped root's hub. That hub genuinely does need un-pinning — it was activated as the Space placeholder, so the live instance still carried the placeholder's configuration (default areas, none of the package type's).

But the installer's recycle was **addressed, not instance-scoped**. Routing resolves the address at *delivery* time, so it disposes whichever instance happens to be live when it lands — and it lands **late**, queued behind the recycle the framework watcher already started. The install then keeps touching that same address (`WarmInstalledRoots`, and `SyncPackageContent`'s `SyncContentFilesRequest`, both of which ACTIVATE the root), so the stray `Dispose` tore down the **fresh, correctly-bound** instance in the middle of the handler serving the install's own content publish.

Measured on `StaleStampRootBindingTest`'s ShopB fixture: **two ShopB disposals 426 ms apart** — the watcher's at +2 ms after the retype, the stray one at +428 ms, landing on the instance the content sync had activated 105 ms earlier. The handler faulted with `HubDisposingException` ("cannot create '/content/'"), answered `ImportContentResponse.Fail` — and **that reply never left the disposing hub**, so the installer burned its whole 60 s `RequestTimeout` and published 0/1 assets.

## Why deleting it is safe now

Since **#1104** the framework owns this guarantee, and owns it better. `NodeTypeRebindWatcher` is armed on every activation and recycles the hub the first time the mesh change feed reports a NodeType different from the one the instance bound. Two properties matter:

- it fires on the **post-commit** feed event, so it cannot race the debounced persist;
- it posts to its **own instance** (`instanceHub.Post(…, WithTarget(instanceHub.Address))`), so the signal can only ever kill the mis-bound instance that armed it.

The installer's copy had **neither** property. That is what made it a defect rather than a redundant fast path.

## Also: the diagnostic that made this cost a whole investigation

`postFate` keys on the **response** delivery's own id, which nobody tracks — so a reply refused by the responder's teardown guard was recorded nowhere, and the request's trail ended at `RESPONSE_POSTED`, reading *"the reply was lost between responder and requester"* with no hint that the responder's own shutdown ate it.

The refusal is now recorded on the **request's** trail (`RESPONSE_REFUSED_SHUTTING_DOWN`), and `RequestFateLedger` reports that verdict ahead of `RESPONSE_POSTED` — a refused reply always carries both stages, and "the responder's own shutdown ate the reply" is strictly sharper than "lost in transit". Diagnostics only; the refusal behaviour is unchanged.

**Left open deliberately, and named in the code:** actually delivering (or NACKing) a reply a disposing hub cannot post. That has to go out through the live **parent** — our own `Post` re-enters the very gate that refused it — which is the `NackThroughParent` contract applied to a third path. Own change, own blast-radius review.

## Verification

| | result |
|---|---|
| Fix applied | **3 passed, 0 failed — 9 s** |
| **Negative control** (revert only `PackageInstaller.cs`) | **1 failed, 2 passed — 2 m 7 s** |
| Restored | 3 passed, 9 s |

The duration is itself the signature: the failing run burns the full 60 s `RequestTimeout`. `MeshWeaver.PluginCatalog` builds clean under `-c Release -warnaserror`.

### DIFF
--- a/src/MeshWeaver.Messaging.Hub/MessageService.cs
+++ b/src/MeshWeaver.Messaging.Hub/MessageService.cs
@@ -1540,6 +1540,24 @@ public class MessageService : IMessageService
             && message is not ShutdownRequest and not DisposeRequest)
         {
             postFate?.Add($"POST_REFUSED_SHUTTING_DOWN runLevel={hub.RunLevel}", Address);
+            // 🚨 …and record it on the REQUEST's trail when the message being refused IS the answer
+            // to an awaited request. `postFate` keys on the RESPONSE delivery's own id, which nobody
+            // tracks, so without this the refusal is recorded nowhere and the request's trail ends
+            // at RESPONSE_POSTED — reading "the reply was lost between the responder and the
+            // requester", with no hint that the responder's own teardown is what ate it. That
+            // ambiguity cost a full investigation on #1151, where an install's SyncContentFilesRequest
+            // burned its whole 60 s RequestTimeout because the answering hub crossed
+            // DisposeHostedHubs between HANDLER_ENTER and the reply.
+            //
+            // Diagnostics only — the refusal itself is unchanged. Actually DELIVERING (or NACKing)
+            // a reply a disposing hub cannot post is the open defect, tracked separately: it has to
+            // go out through the live PARENT (our own Post re-enters this very gate), which is the
+            // NackThroughParent contract applied to a third path — one this guard, not the intake
+            // gate or the disposal drain, is responsible for.
+            if (correlatedRequestId is not null)
+                requestFates?.Find(correlatedRequestId.ToString())?.Add(
+                    $"RESPONSE_REFUSED_SHUTTING_DOWN type={message.GetType().Name} runLevel={hub.RunLevel}",
+                    Address);
             return ((IMessageDelivery)delivery).Failed("Hub is shutting down");
         }
 
--- a/src/MeshWeaver.Messaging.Hub/RequestFateLedger.cs
+++ b/src/MeshWeaver.Messaging.Hub/RequestFateLedger.cs
@@ -233,10 +233,9 @@ internal sealed class RequestFateLedger
             // shutdown ate the reply" is a strictly sharper answer than "it was lost in transit".
             if (Has("RESPONSE_REFUSED_SHUTTING_DOWN"))
                 return "the handler REPLIED but the responding hub was already past "
-                     + "DisposeHostedHubs, so its own teardown guard refused the post — nothing "
-                     + "reached the requester and nothing NACKed it, so the caller burns its whole "
-                     + "RequestTimeout. The responder is the one shutting down; look at what "
-                     + "recycled/deleted that address mid-handler, not at the requester.";
+                     + "DisposeHostedHubs, so its own teardown guard refused the post and the "
+                     + "caller burns its whole RequestTimeout un-NACKed — chase what "
+                     + "recycled/deleted the RESPONDER's address mid-handler, not the requester";
             if (Has("RESPONSE_POSTED"))
                 return "a reply WAS posted for this correlation and the callback is STILL pending — "
                      + "the reply was lost between the responder and the requester, so chase the "
--- a/src/MeshWeaver.Messaging.Hub/RequestFateLedger.cs
+++ b/src/MeshWeaver.Messaging.Hub/RequestFateLedger.cs
@@ -228,6 +228,15 @@ internal sealed class RequestFateLedger
         {
             bool Has(string token) => snapshot.Any(s => s.StartsWith(token, StringComparison.Ordinal));
 
+            // Ahead of RESPONSE_POSTED on purpose: a refused reply always carries BOTH stages (the
+            // post is recorded, then the teardown guard refuses it), and "the responder's own
+            // shutdown ate the reply" is a strictly sharper answer than "it was lost in transit".
+            if (Has("RESPONSE_REFUSED_SHUTTING_DOWN"))
+                return "the handler REPLIED but the responding hub was already past "
+                     + "DisposeHostedHubs, so its own teardown guard refused the post — nothing "
+                     + "reached the requester and nothing NACKed it, so the caller burns its whole "
+                     + "RequestTimeout. The responder is the one shutting down; look at what "
+                     + "recycled/deleted that address mid-handler, not at the requester.";
             if (Has("RESPONSE_POSTED"))
                 return "a reply WAS posted for this correlation and the callback is STILL pending — "
                      + "the reply was lost between the responder and the requester, so chase the "
--- a/src/MeshWeaver.PluginCatalog/PackageInstaller.cs
+++ b/src/MeshWeaver.PluginCatalog/PackageInstaller.cs
@@ -1668,46 +1668,47 @@ public static class PackageInstaller
                     .SelectMany(rest => RootRetypeReconciled()
                         .SelectMany(_ => RootRetypePersisted())
                         .Select(_ => rest))
-                    // 🚨 RECYCLE the retyped root's hub. It was ACTIVATED as the Space placeholder
-                    // (RootRetypeReconciled reads the stream, and readers race the install anyway),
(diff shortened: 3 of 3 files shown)


rbuergi:
## ⚠️ This does NOT green main — do not merge expecting it to

Correcting my own PR body. CI run on `169672e11` (my exact commit, verified by SHA):

- **shard 4 — `StaleStampRootBindingTest.StaleStampSelfTypedRoot_RootServes` still FAILS**, which is the very test this PR set out to fix.
- shard 2 — `MeshWeaver.FutuRe.Test exit=139`, the CoreCLR `background_sweep` fault closed as external in #613; not this diff.

Locally the whole class is **3 passed in 9 s**, and the negative control is real — reverting only `PackageInstaller.cs` gives **1 failed in 2 m 07 s**, the failure burning the full 60 s `RequestTimeout`. So the stray addressed `Dispose` **is** a genuine defect and removing it **is** a genuine fix.

It is just **not sufficient**. Local-passes/CI-fails is the documented signature of this class: whatever CI adds beyond processor count — slower cores, disk, the other projects sharing the shard — is part of the repro and is not reproduced here. My earlier claim that this fixes main was wrong, and I would rather say so than have someone merge it on that basis.

**What is still true and worth keeping:**
- The installer's recycle was **addressed, not instance-scoped** — routing resolves at delivery time, so it disposed whichever instance was live when it landed, which (queued behind the framework watcher's recycle) was the fresh, correctly-bound one. Measured: two disposals 426 ms apart, the second landing on an instance activated 105 ms earlier.
- Since #1104 the framework owns that guarantee properly (`NodeTypeRebindWatcher` fires post-commit and posts to its **own** instance), so the installer's copy is redundant *and* harmful.
- The `RESPONSE_REFUSED_SHUTTING_DOWN` trail closes a real diagnostic gap: `postFate` keys on the response delivery's own id, which nobody tracks, so a reply eaten by the responder's teardown was recorded nowhere and the trail read "lost between responder and requester" — pointing at the wrong end.

**Next step for whoever picks this up:** the remaining failure needs a CI-side capture, not another local run. The `RESPONSE_REFUSED_SHUTTING_DOWN` stage added here should make the next CI red say on sight whether the reply was refused by a disposing responder or genuinely lost — which is the discriminator the original investigation lacked.

Also addressed in `dd1501767`: the review finding that the new verdict returned two sentences while `Verdict()` documents "ONE sentence a reader needs". Reworded to one, keeping both facts (the caller burns its timeout un-NACKed; the address to chase is the responder's).

rbuergi:
Superseded on both halves — closing rather than leaving a third overlapping fix for #1151 on the board.

**PackageInstaller half → #1153 (merged, greens main).** This PR removed the fire-and-forget `DisposeRequest` and relied on `NodeTypeRebindWatcher` to un-pin the root. #1153 replaced the same post with an *ordered* recycle-and-wait, which is the better fix: it keeps the un-pinning the installer genuinely needs while guaranteeing the teardown finishes before `Install` returns, so the next touch of the root cannot land mid-teardown.

**MessageService half → #1154.** This PR only added a `RESPONSE_REFUSED_SHUTTING_DOWN` breadcrumb and said explicitly in its own comment that *"actually DELIVERING (or NACKing) a reply a disposing hub cannot post is the open defect, tracked separately"*. #1154 is that fix — it routes the `HandleMessage` disposal-race NACK through the live parent, because `ReportFailure` posts through the hub's own `Post` and declines once `RunLevel >= DisposeHostedHubs`, which is precisely the state the NACK is computed in. Diagnosing the hole is worth less than closing it, and #1154 closes it.

For the record, my earlier claim on this PR that it greens main was wrong and I corrected it here at the time: `StaleStampSelfTypedRoot_RootServes` still failed on CI on the SHA-verified commit. #1153 is what actually greened main (run 31401910253, all six shards).

PR #1164, opened 2026-08-10
fix(delete): recursive delete denies atomically up front — no more half-deleted subtrees on permission loss

### LINKED ISSUE #1128: MeshNode handler fails to recognize Instances paths for DeleteNode, logs DeliveryFailure on permission denial
**What is failing**

The `MeshNode` message handler received a `DeleteNodeRequest` targeting `Instances/Deployment` and logged a `DeliveryFailureException` with the message `[DeleteNode] unexpected path=Instances partial-deleted=31`. The underlying failure is a permission check: user `rbuergi` lacks `Delete` permission on `Instances/Deployment`.

**Probable cause — confident on the permission part, the "unexpected path" part needs investigation**

The permission denial is **working as designed**: `rbuergi` is a platform admin (global admin scoped to the `Admin` partition via `Auth:GlobalAdmins` config), which grants `Permission.All` only on the `Admin` partition — not on `Instances/Deployment`. Deleting a deployment under `Instances` requires an explicit `Delete` grant at the `Instances` scope.

However, the `[DeleteNode] unexpected path=Instances` diagnostic in the log is suspicious: it suggests the `MeshNode` message handler treats `Instances` as an unexpected or unrecognized path for `DeleteNode`. Combined with `partial-deleted=31`, this hints that a bulk delete partially succeeded (31 nodes deleted) before the permission check caught and stopped further processing — which means the permission check runs **after** some mutations have already been applied. The "unexpected path" log itself may be a code issue (the handler not recognizing the `Instances` partition as a valid delete target) that should be investigated separately.

**Impact**

Single occurrence, single pod (`memex-portal-deployment-575dc48bfb-wq7s8`), ~4.5 hours ago. Low severity. The user was denied an unauthorized operation, which is correct behavior. The partial-delete count (31) and the "unexpected path" diagnostic are the actionable concerns.

**Where to look**

- `MeshNode` delete handler in `MeshWeaver.Mesh` — specifically the `[DeleteNode]` logging to understand why `Instances` is considered "unexpected"
- The delete pipeline to understand whether permission checks run before or after mutation begins (partial-delete=31 suggests they may be out of order)
- The `Instances/Deployment` node type currently has a compilation error (source files missing) — unrelated but worth noting

---

### Evidence

| | |
|---|---|
| Fingerprint | `173a7da836430e52` |
| Category | `MeshWeaver.Mesh.MeshNode` |
| Severity | Error |
| Exception | `MeshWeaver.Messaging.DeliveryFailureException` |
| Namespace | `memex` |
| Pods | `memex-portal-deployment-575dc48bfb-wq7s8` |
| Occurrences | 1 |
| First seen | 2026-08-10 07:34:22Z |
| Last seen | 2026-08-10 07:34:22Z |

```
2026-08-10 07:34:22Z memex-portal-deployment-575dc48bfb-wq7s8 fail: MeshWeaver.Mesh.MeshNode[0]
      [DeleteNode] unexpected path=Instances partial-deleted=31
      MeshWeaver.Messaging.DeliveryFailureException: Access denied: user 'rbuergi' lacks Delete permission on 'Instances/Deployment'
```

</details>

<sub>Opened automatically from `Admin/_LogIncident/173a7da836430e52`. Recurrences are folded into this issue rather than opening new ones.</sub>

rbuergi:
Fixes #1128

## What was happening

A recursive `DeleteNodeRequest` on `Instances` aborted with `DeliveryFailureException: Access denied: user 'rbuergi' lacks Delete permission on 'Instances/Deployment'` **after 31 descendants were already deleted**, and the handler logged the denial as `[DeleteNode] unexpected path=Instances partial-deleted=31`.

## Root cause — measured order

The pipeline already decides a lot before mutating (root permission check → root validators → plan → per-descendant `ValidateDeleteRequest` pre-flight, which carries the same `[RequiresPermission(Delete)]` gate). The defect is that the **commit re-takes the authorization decision per-leaf, mid-mutation, against state the cascade itself is deleting**:

1. The deletion plan (authoritative storage enumeration, incl. satellite tables) contains the subtree's own `_Access` grant nodes — the very rows that authorize the caller.
2. The bottom-up fan-out (`Observable.Merge` over sibling branches) deletes those grants early.
3. The caller's effective Delete permission evaporates mid-commit; the next per-leaf `DeleteNodeRequest` is refused — by the delivery gate and/or by `RlsNodeValidator`, which derives its principal from the `DeletedBy` audit field.
4. The cascade aborts half-done (no rollback — actor model), and the error branch labels an expected permission denial "unexpected".

The 31 deleted nodes were physically removed from storage; there is no recycle path in this pipeline (recovery only via version history / point-in-time restore where PG backs the partition).

## Fix — decide once up front (caller), execute the decided cascade (system)

- **Atomic up-front denial, legible**: an Unauthorized refusal in the per-descendant pre-flight now aborts the whole operation **before any mutation** with `NodeDeletionRejectionReason.Unauthorized` and the gate's message naming the denied path (previously mislabelled `ValidationFailed`).
- **Commit immune to self-revocation**: `DeleteSubtreeUntilDrained` stamps the SYSTEM `AccessContext` on the per-leaf deletes. `DeletedBy` still carries the caller for audit; per-leaf validators still run.
- **`RlsNodeValidator`**: a cascade leg (`CascadeRootPath` set — only the fan-out stamps it) executed under the **delivery-stamped** system context acts as system instead of re-deciding authorization from `DeletedBy` mid-commit. Both signals are required, so the leaked ambient `system-security` AsyncLocal residue never matches — `DeletedBy` stays authoritative for every direct request (RlsIntegrationTests' shape unchanged, all 23 green).
- **Diagnostics name the real condition**: permission denial → Warning `[DeleteNode] permission-denied path=…`; a denial arriving with partial deletions stays a LOUD error canary (`permission-denied MID-COMMIT … N node(s) already deleted`); the response passes the legible "Access denied …" message through without the "Unexpected error:" prefix.

## Test evidence

New `RecursiveDeletePermissionTest` (MonolithMeshTestBase + per-user `SetCircuitContext`):

- `RecursiveDelete_DeniedOnDescendant_RefusedAtomically_NothingDeleted` — caller holds Delete on the root but is denied on a descendant scope → structured Unauthorized response naming the denied path; storage-authoritative assert that **nothing** was deleted.
- `RecursiveDelete_PlanContainsCallersOwnAccessGrant_CompletesFully` — the incident mechanism (caller's only grant inside the subtree). **Red before the fix** with the exact incident signature (`[DeleteNode] unexpected path=… partial-deleted=2`, mid-commit denial), green after.

Regression: `DeleteNodeBehaviorTest` + `RecursiveDeleteAuthoritativeTest` (13/13), `RlsIntegrationTests` (23/23), `PartitionRootDeletionGuardTest` (13/13), `MeshNodeAuditingTest` (3/3). Touched projects build clean with `-c Release -warnaserror`.

https://claude.ai/code/session_01UYLgoPK1qLtyxZ2ixdFtj5

### DIFF
--- a/src/MeshWeaver.Documentation/Data/WhatsNew/2026-08-10-recursive-delete-atomic-permission-denial.md
+++ b/src/MeshWeaver.Documentation/Data/WhatsNew/2026-08-10-recursive-delete-atomic-permission-denial.md
@@ -1,0 +1,20 @@
+---
+Name: Recursive delete now refuses permission denials up front — never half-deletes a subtree
+Category: Fix
+Description: A recursive delete that hits a missing Delete permission is now refused atomically before anything is removed, with a clear "permission denied" answer instead of an unexpected error after a partial deletion.
+Icon: Sparkle
+Order: -20260810
+---
+
+# Recursive delete now refuses permission denials up front — never half-deletes a subtree
+
+Deleting a folder tree you were not fully allowed to delete could previously remove part of
+the tree before stopping: the permission check for each node ran while the deletion was
+already underway, so dozens of nodes could be gone by the time the denial hit — and the error
+was reported as an unexpected failure rather than a permission problem.
+
+Recursive deletes now decide permissions for the whole subtree before touching anything. If
+you lack Delete on any node in the tree, nothing is deleted and you get a clear message naming
+the node you are not allowed to delete. A delete that was fully authorized when it started can
+also no longer trip over its own access records mid-way — it either refuses up front or
+completes fully.
--- a/src/MeshWeaver.Graph/Security/RlsNodeValidator.cs
+++ b/src/MeshWeaver.Graph/Security/RlsNodeValidator.cs
@@ -243,11 +243,33 @@ public class RlsNodeValidator : INodeValidator, IOwnerEnforcedNodeValidator
 
     /// <summary>
     /// Extracts the user ID from the validation context.
-    /// First checks explicit request identity (CreatedBy/UpdatedBy/DeletedBy),
-    /// then falls back to AccessContext (the logged-in user).
+    /// A recursive-delete cascade leg executed under the delivery-stamped SYSTEM
+    /// context acts as system; otherwise the explicit request identity
+    /// (CreatedBy/UpdatedBy/DeletedBy) is checked first, then the AccessContext
+    /// (the logged-in user).
     /// </summary>
     private static string? GetUserId(NodeValidationContext context)
     {
+        // 🚨 A CASCADE-LEG delete executed under the SYSTEM context acts as system, even
+        // though DeletedBy carries the original caller for the audit trail. The recursive
+        // delete authorizes the WHOLE subtree up front under the caller (root permission
+        // check + the [RequiresPermission(Delete)] gate on every descendant's pre-flight
+        // ValidateDeleteRequest) and then commits under the system identity — because the
+        // plan contains the subtree's own _Access grant satellites, and re-deriving the
+        // principal from DeletedBy here re-decided that authorization MID-COMMIT against
+        // state the cascade itself was deleting, aborting the subtree half-deleted
+        // (issue #1128, "partial-deleted=31"). BOTH signals are required:
+        //  • CascadeRootPath — only the subtree fan-out stamps it; and
+        //  • the SYSTEM AccessContext — carried on the DELIVERY by the fan-out's explicit
+        //    WithAccessContext stamp and threaded into this context by the delete handler.
+        // A leaked ambient "system-security" (the bootstrap-ImpersonateAsSystem AsyncLocal
+        // residue AccessControlPipeline.ResolveIdentity documents) has CascadeRootPath null
+        // and therefore never matches — DeletedBy/CreatedBy stay authoritative for every
+        // direct request, exactly as before (RlsIntegrationTests' DeletedBy shape).
+        if (context.Request is DeleteNodeRequest { CascadeRootPath: not null }
+            && context.AccessContext?.ObjectId == WellKnownUsers.System)
+            return WellKnownUsers.System;
+
         // Check explicit request identity first
         var requestUserId = context.Request switch
         {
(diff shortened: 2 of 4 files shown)


rbuergi:
Independent verification of this diff (separate session, no shared state) — the diagnosis holds. Two confirmations and one residual.

**Confirmed by elimination, not just by the log text.** `PreValidateDescendantsObs` already aborts *before* any storage side effect, and its `.Catch` converts a delivery-gate `Unauthorized` into a pre-flight failure. So a *pre-existing* denial on `Instances/Deployment` could never have produced `partial-deleted=31` — it would have posted `pre-validation failed` with nothing deleted. The permission must therefore have evaporated **between** pre-flight and commit, which only the cascade deleting the subtree's own `_Access` rows explains. That independently corroborates the self-revocation mechanism.

**The system-stamp guard is not reachable from a direct request.** `CascadeRootPath` is pre-existing and written in exactly one place — the fan-out at `MeshExtensions.cs:2704` (`baseRequest with { … CascadeRootPath = rootPath }`). Combined with the delivery-stamped `AccessContext`, a remote caller cannot forge the bypass.

**Also worth noting for the record:** the issue title ("fails to recognize Instances paths") is a misreading of the log format — `unexpected` is the *error-kind* classifier in `"[DeleteNode] {Kind} path={Path} partial-deleted={Partial}"`, not a statement about the path. The diff's relabeling fixes exactly that.

---

**Residual (narrow, not a blocker): drain-pass survivors are deleted as SYSTEM without ever passing pre-flight.**

`DeleteSubtreeUntilDrained` re-enumerates after each pass and deletes anything that "appeared mid-flight" (`MeshExtensions.cs:2579-2587`). Those paths are by definition *not* in `collected.ToDelete`, so they never went through `PreValidateDescendantsObs` — yet after this change they commit under the system identity.

Before this PR they were permission-checked per-leaf against the caller; now they are not. The blast radius is bounded (they must be descendants of a root the caller was authorized to delete recursively), and leaving orphans behind would be worse — so this is probably the right trade. But it is a genuine widening rather than a no-op, and the "authorization decided once up front" invariant in the comments doesn't literally hold for that set. Might be worth either saying so explicitly in the `DeleteSubtreeUntilDrained` comment, or pre-validating survivors on drain passes.

PR #1165, opened 2026-08-10
A scan that lost its load context bound the hub to nothing

rbuergi:
**Not a fix for the main red — that is gone (#1153).** This is a *third*, independent
failure mode on the same `StaleStampRootBindingTest` cluster, captured on CI and fixed at
its root. It is orthogonal to both #1153 (installer teardown outliving the install) and
#1163 (a raced subscribe parking on an un-re-probed `ShuttingDown` NACK).

## The evidence

CI run [31404522444](https://github.com/Systemorph/MeshWeaver/actions/runs/31404522444),
shard 4, commit `169672e1`. The test does **not** time out — it completes in **831 ms** and
fails the *content* assertion:

```
[type +609ms] v6  Status=Ok       asm='ShopA_Front/v3-93d7ee2d-694c899111a5.dll'
rebuild settled at +609ms
[type +642ms] v7  Status=Pending
[type +643ms] v9  Status=Compiling
[Warning] Failed to extract NodeTypeConfigurations from …/v3-93d7ee2d-694c899111a5.dll:
    Cannot scan an assembly whose context is unloading
    Object name: 'DynamicNode_ShopA_Front'
    at NodeAssemblyLoadContext.Pin()  … at MeshNodeCompilationService.CompileResultFromAssembly
[type +697ms] v10 Status=Ok       asm='ShopA_Front/v8-93d7ee2d-694c899111a5.dll'
Tests frame at +831ms:  … Area not found
    at StaleStampRootBindingTest.cs(319)  — NotContain("Area not found")
```

Note the two assembly names: `v3-93d7ee2d-694c899111a5` and `v8-93d7ee2d-694c899111a5` —
**same framework hash, same source hash**. A byte-identical recompile is what evicted the
load context out from under a concurrent reader.

## The defect

A per-instance hub activation hydrates its NodeType's `HubConfiguration` by reflecting over
the compiled assembly (`GetConfigurationsFromExistingAssembly` →
`CompileResultFromAssembly`). Both scan sites resolved the load context and pinned it as
**two statements**:

```csharp
var context = cacheService.GetOrCreateLoadContextForPath(nodeName, assemblyLocation);
using var scanPin = context.Pin();
```

A recompile landing in between registers its own path-keyed context, and
`EvictSupersededContexts` → `UnloadContext` disposes every superseded one — including the
reference the reader had *just* resolved. The pin then throws on a context that was live one
instruction earlier.

`Pin()` documents exactly what to do about that, in its own XML doc:

> Throws `ObjectDisposedException` if the context is already unloading — **the caller must
> then re-resolve against the current context** rather than scan a doomed assembly.

Neither caller implemented it. The `catch` below recorded an **empty configuration list**
instead, and the activation path reads that as authoritative: no matching configuration →
`ApplyEntry(…, hubConfig: null)` → the hub binds the mesh defaults. A hub resolves its
configuration exactly once, at activation, so the freshly installed root then serves
"Area not found" **for its whole lifetime** — from a rebuild it merely ran next to. The
neighbouring branch in `NodeTypeEnrichmentHelpers` already calls this outcome out by name as
"the silent default-config fallback … cached for the grain's lifetime".

## The fix

Resolve and pin become **one** operation, `ICompilationCacheService.PinForScan`. The cache
service owns it because it owns `_loadContexts` — it is the only component that *can*
re-resolve. Both scan sites go through it, so two readers racing each other's evictions
both converge instead of both failing.

It terminates rather than spins: the evictor removes the key **before** it disposes, so the
next resolve constructs a fresh context; the defensive `TryRemove` covers the remove-less
ordering so the loop can never be handed the same doomed context twice. Exhaustion rethrows,
so a genuinely dying cache still surfaces.

## Verification

- `ScanPinSupersessionTest` (new, deterministic, dependency-free) pins three facts: a
  recompile really does doom a concurrent reader's already-resolved context (the
  precondition, asserted so the fix is pinned against a race that happens); `PinForScan`
  survives it; and a disposed context left in the cache is replaced rather than returned.
  **3/3 in 346 ms.**
- `dotnet build -c Release -warnaserror` clean for `MeshWeaver.Graph` + `MeshWeaver.Graph.Test`.
- `MeshWeaver.PluginCatalog.Test`, `DOTNET_PROCESSOR_COUNT=4`, Release — reported in full because
  one run was red and the number matters:

  | tree | result |
  |---|---|
  | #1159's tree + this fix | 236/236, 1 m 21 s |
  | `origin/main` (0e4cc931) + this fix | **235/1**, 2 m 1 s |
  | `origin/main` (0e4cc931) alone | 236/236, 1 m 22 s |

  The red is **`SelfTypedRootWhoseTypeNeverCompiles_SkipsThePublishPromptly`** — a different test
  from the one this PR is about — and it is the pre-existing installer flake, not this change:

  - **This change is not on that test's code path.** ShopC's type never compiles, so
    `assemblyLocation` is empty and `CompileAndGetConfigurations` takes the
    `BuildFailureDiagnostics` branch, never reaching `CompileResultFromAssembly`. Its
    `HasUsableBuild` is false (the committed stamp carries a stale framework hash), so the
    hydration path never calls `GetConfigurationsFromExistingAssembly` either, and nothing emits,
    so `LoadAndExtractConfigurationsFromReleaseAsync` never runs. **Neither of the two call sites
    this PR touches executes for that test.**
  - #1163 independently measured this class at **2 of 5 class runs red** on `origin/main + #1153`.
    A single green baseline run is exactly what a 3-in-5 green rate produces, so it exonerates
    nothing by itself — the mechanism argument above is the load-bearing one.
  - The failure's own shape puts it in the installer: the log goes silent from **+2.3 s** (the
    expected `CS0103` compile failure for `ShopC/Front`) to **+30.2 s**, then an Rx
    `TimeoutException: The operation has timed out.` surfaces at `Install`. 30 s is not the test's
    own bound (`StepTimeout` is 120 s) — `PackageInstaller` has several uncaught
    `.Timeout(TimeSpan.FromSeconds(30))` (lines 465, 561, 1544, 1680, 1702, 1965), one of which
    fired. Handed to the installer cluster rather than guessed at here.

## Triage discriminator for the next red of this test

Three modes, told apart on sight:

| Fails at | Signature | Owner |
|---|---|---|
| ~60 s, on the **install** | reply died with a hub torn down mid-handler | #1153 (merged) |
| ~120 s, on the **frame wait** | `TimeoutException`, subscribe parked on an un-re-probed `ShuttingDown` NACK | #1163 |
| **< 1 s**, at `StaleStampRootBindingTest.cs:319` | `NotContain("Area not found")` + log carries `Cannot scan an assembly whose context is unloading` | **this PR** |

#1163's re-probe does not cover this one: a re-probe reactivates the owner, but if that
reactivation's configuration lookup hits an unloading context again it binds the defaults
again and answers *promptly* with the wrong content. If anything the re-probe schedules the
reactivation closer to the rebuild, so the two changes are complementary.

## Named, deliberately deferred

- **`EvictSupersededContexts` ping-pong.** Two readers on different assembly paths of the
  same NodeType each evict the other on registration. Bounded and convergent with this fix,
  but it is avoidable churn (an ALC re-created and re-loaded per hydration of a superseded
  path). Not touched here — narrowing it would only make the race rarer, which is not a fix.
- **The redundant recompile itself.** The second compile had an identical framework hash
  *and* source hash; `CompiledSources` keys on `LastModifiedTicks`, not content. Skipping it
  would shrink this window, but again only rarer, not gone — a real source edit produces the
  same eviction.
- **The silent default-binding still has no self-heal.** With this fix the scan no longer
  fails for supersession, but a genuine reflection failure (missing dependency) still lands
  on `ApplyEntry(hubConfig: null)` with no overlay and no `WithOverlaySelfHeal`, unlike its
  sibling byte-resolution branch — which the sibling's own comment says is mandatory-gated.
  Getting that version gate subtly wrong hot-loops a recycle, so it is not being done
  alongside a fix that already stands on its own.

### DIFF
--- a/src/MeshWeaver.Documentation/Data/WhatsNew/2026-08-10-a-page-no-longer-loses-its-views-to-a-rebuild-next-door.md
+++ b/src/MeshWeaver.Documentation/Data/WhatsNew/2026-08-10-a-page-no-longer-loses-its-views-to-a-rebuild-next-door.md
@@ -1,0 +1,37 @@
+---
+Name: A page no longer loses its own views to a rebuild happening next door
+Category: Fix
+Description: A page could come up with only the generic menu — none of its own views — and stay that way, because it opened at the instant its code was being rebuilt.
+Icon: ArrowSync
+Order: -20260810
+---
+
+# A page no longer loses its own views to a rebuild happening next door
+
+A page whose behaviour comes from code you wrote could open showing only the
+generic node menu — none of the views that code defines — and reporting that the
+view you asked for could not be found. Reloading did not help. The page stayed
+that way until somebody restarted it by hand, even though the code it was missing
+had built perfectly well seconds earlier.
+
+## What was happening
+
+A page works out which code serves it exactly once, when it first opens. That
+lookup reads the compiled result of your code.
+
+Rebuilds swap that compiled result out. If a page happened to open in the instant a
+rebuild was swapping — and a rebuild can be triggered by something entirely
+unrelated to that page — the lookup found itself reading a result that was being
+retired underneath it. Instead of looking again at the current one, it gave up and
+reported that your code defines no views at all.
+
+That answer was then treated as the truth. The page bound itself to the generic
+defaults, and because the lookup only ever happens once, it kept them for as long
+as it stayed open.
+
+## What changed
+
+A lookup that finds itself reading a result being retired now simply looks again at
+the current one, which is what it should have done all along. The rebuild next door
+no longer costs a page its own views, and a page that opens during one comes up
+complete.
--- a/src/MeshWeaver.Graph/Configuration/CompilationCacheService.cs
+++ b/src/MeshWeaver.Graph/Configuration/CompilationCacheService.cs
@@ -195,11 +195,51 @@ internal interface ICompilationCacheService
     NodeAssemblyLoadContext GetOrCreateLoadContextForPath(string nodeName, string dllPath);
 
     /// <summary>
+    /// Resolves a LIVE context for a scan and returns it already pinned — the atomic form of
+    /// "GetOrCreateLoadContext… then <see cref="NodeAssemblyLoadContext.Pin"/>".
+    /// <para>🚨 Resolve-then-Pin as two steps is a RACE, and losing it silently mis-binds a hub
+    /// for its whole lifetime. A concurrent recompile evicts the superseded context
+    /// (<c>EvictSupersededContexts</c> → <c>UnloadContext</c>) between the caller's resolve and
+    /// its pin, so the pin throws <see cref="ObjectDisposedException"/> on a reference that was
+    /// live one instruction earlier. <see cref="NodeAssemblyLoadContext.Pin"/> documents the
+    /// answer — "the caller must then re-resolve against the current context rather than scan a
+    /// doomed assembly" — but no caller implemented it, and the swallow downstream turned a
+    /// transient supersession into an authoritative EMPTY configuration list (issue #1151:
+    /// the freshly installed root hub bound the mesh defaults and served "Area not found" for
+    /// good, because a hub resolves its configuration exactly once, at activation).</para>
+    /// <para>Only this service can close that window: it owns the dictionary, so it is the only
+    /// component that can re-resolve. The loop terminates because the evictor removes the key
+    /// BEFORE disposing, so the next resolve constructs a fresh context.</para>
+    /// </summary>
+    PinnedScanContext PinForScan(string nodeName, string? dllPath);
+
+    /// <summary>
+    /// <see cref="PinForScan(string, string?)"/> for a release-keyed context.
+    /// </summary>
+    PinnedScanContext PinForScanOfRelease(NodeTypeRelease release, string releaseFolder);
+
+    /// <summary>
     /// Registers NuGet probing directories for a node. The node's AssemblyLoadContext
     /// consults these directories during Resolving events so transitive dependencies
     /// of resolved NuGet packages can be loaded at runtime.
     /// </summary>
     void RegisterProbingDirectories(string nodeName, System.Collections.Generic.IReadOnlyList<string> directories);
+}
+
+/// <summary>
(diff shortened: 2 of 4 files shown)


rbuergi:
## CI: [run 31408256551](https://github.com/Systemorph/MeshWeaver/actions/runs/31408256551)

**Shard 4 — the shard that carries `MeshWeaver.PluginCatalog.Test` / `StaleStampRootBindingTest` — passed** (6 m 36 s), as did shards 0, 1, 3, 5 and `Plugins compile against this PR`.

| job | result |
|---|---|
| Build solution (once) | pass 9 m 40 s |
| Run tests (shard 0 / 1 / **4** / 5) | pass 6 m 51 s / 7 m 5 s / **6 m 36 s** / 9 m 10 s |
| Run tests (shard 3) | pass 8 m 40 s |
| Run tests (shard 2) | **fail** 9 m 0 s |
| Test Results (shard 0–5) | pass (all six) |
| Consolidate test results | fail — downstream of shard 2's job |
| Plugins compile against this PR | pass |

**Shard 2's failure is not this diff, and not a test failure at all.** `Test Results (shard 2)` is
**pass** — zero failed tests. The job failed on a process crash:

```
line 27: 3166 Segmentation fault  ( cd "$dir" && timeout … 8m dotnet MeshWeaver.FutuRe.Test.dll … )
[CI] MeshWeaver.FutuRe.Test exit=139
```

That is the CoreCLR `background_sweep` fault closed as external in #613. Two things rule this diff
out directly:

- **It reproduces without this change.** #1159's own CI on `169672e1` — a tree that does not contain
  this fix — failed shard 2 identically, and that PR recorded it as "the CoreCLR `background_sweep`
  fault closed as external in #613; not this diff".
- **This diff's own project passed in that very shard.** `MeshWeaver.Graph.Test  Total: 921,
  Failed: 0` ran on shard 2 alongside the crash, matching the local run (921/921, 29 s).

Not merging — CI is not fully green, and the remaining red is the #613 external fault rather than
anything this branch can fix.

PR #1177, opened 2026-08-10
Composer load no longer surfaces a transient owner miss as a raw error

rbuergi:
## Symptom

Navigating to the home page on memex.meshweaver.cloud popped a raw framework error in the portal modal:

> Loading the composer: No response received in hub cache/gmIIqhF220CKKxqxgY_J7Q within 00:01:00 for request SubscribeRequest (id=…) → target rbuergi/_Thread/ThreadComposer. The request may have been undeliverable or the target hub was not found.

Nothing was actually broken — dismissing it and carrying on worked.

## Root cause

The per-user composer grain had idle-collected, and its reactivation outran the 60 s request budget. That is the exact failure class `MeshNodeStreamCache.IsTransientOwnerFailure` documents: the error is forwarded to the current subscriber **for the caller to retry**, and the cache stays clean so a resubscribe re-probes. Every layout area implements that contract via `RetryAreaWithBackoff(AreaErrorClassifier.ShouldRetryArea)` — but the two composer-node readers never did:

- `ThreadChatView.OpenComposerProjection` surfaced the very first failure verbatim (`SurfaceError` → portal error modal).
- `ThreadSidePanelContent.OpenComposerObserver` died silently on the first miss (Debug log), leaving Send-click navigation inert for the circuit's whole lifetime.

## Fix

Both readers now use the same bounded, backoff-throttled retry as every layout area. A transient failure that survives the whole retry budget is logged at Warning and stays silent (the raw banner is hostile and unactionable; the next rebind resubscribes naturally). Non-transient failures still surface. The side panel's give-up log is promoted Debug → Warning permanently — a dead navigation observer is a real defect worth seeing, and it only fires after the retry budget is exhausted.

No new classification or retry logic: the transient predicate is already pinned by `AreaErrorClassifierTest` (including the "No response received in hub … SubscribeRequest" banner) and the backoff by `AreaStreamRetryTest`. This PR is the missing wiring around those tested helpers, so there is no unbounded-resubscribe storm risk (the 2026-06-14 wedge class).

## Verification

- `dotnet build src/MeshWeaver.Blazor.Portal -c Release -warnaserror` — clean, 0 warnings.
- `WhatsNewEntryIntegrityTest` — passed (entry `2026-08-10-the-chat-input-no-longer-greets-you-with-a-technical-error.md`, Category: Fix).

### DIFF
--- a/src/MeshWeaver.Blazor.Portal/Chat/ThreadChatView.razor.cs
+++ b/src/MeshWeaver.Blazor.Portal/Chat/ThreadChatView.razor.cs
@@ -823,6 +823,14 @@ public partial class ThreadChatView : BlazorView<ThreadChatControl, ThreadChatVi
     /// so the read never hits a missing node. <see cref="MeshWeaver.AI.ThreadComposerNodeType.ComposerOf"/>
     /// discriminates by NodeType: a ThreadComposer node's own content, or the thread's embedded
     /// <c>Thread.Composer</c>.
+    /// <para>An idle-collected composer grain whose reactivation outruns the request budget answers
+    /// the subscribe with a TRANSIENT owner failure ("No response received in hub … target hub was
+    /// not found"). The stream-cache contract forwards that to the subscriber for the CALLER to
+    /// retry (<c>MeshNodeStreamCache.IsTransientOwnerFailure</c>) — the bounded backoff here is that
+    /// retry, the same <see cref="AreaStreamRetry"/> every layout area uses. Only a NON-transient
+    /// failure is surfaced; a transient one that survives the whole budget is logged and stays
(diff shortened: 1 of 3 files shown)


rbuergi:
This PR resolves the actionable defect in issue #1175 (auto-filed tonight from the same banner on `rsalzmann/_Thread/ThreadComposer`, pod `…6bd9c46f8b-z9t6r`, 14:13–14:20Z). Evidence-based triage on the issue: https://github.com/Systemorph/MeshWeaver/issues/1175#issuecomment-5245863789 — consider adding `Fixes #1175` to the body so the merge closes it (the server-side starvation that made the owner probe slow in that window is tracked separately in #1172).

PR #1193, opened 2026-08-10
A live search feed admitted rows its own snapshot excludes (and the storm test had no quiescence signal)

rbuergi:
`SearchResultStormTest.CatalogSubscription_ContentUpdateStorm_AddsNoExtraRerenders` fails
intermittently on CI and takes the whole shard with it — it red **#1176** (a diff confined to
`MeshWeaver.Observability`), **#1238**, and `main` at `9a05081`.

It turned out to be **two** defects with the same symptom. The first two commits fixed the test;
the third fixes the product, and that is the one that was actually producing "found 4".

---

## 1. The test had no quiescence signal (commits 1–2, test-only)

The re-render baseline was latched on a count threshold:

```csharp
await WaitUntil(() => accumulator.Nodes.Count >= 3, ReadNodeTimeout);
var rerendersAfterLoad = Volatile.Read(ref rerenders);
```

**A count threshold is not a signal that the load has finished.** The test's own comment said so —
*"Initial, plus any trailing `Added` if the index lagged a create"* — but then treated reaching 3 as
the end of it. Each trailing `Added` is a structural render, so the baseline was whatever the counter
held at an arbitrary point *inside* the churn, and the closing `Be(rerendersAfterLoad)` blamed the
storm for renders that happened before it. Reproduced locally as `Expected 1, but found 3`.

The fix is ordering, not time: a marker node created **after** the three real leaves, with the
baseline latched inside the subscription when that marker is applied. The feed is ordered, so by then
every structural frame for `a`/`b`/`c` is applied *and* counted. The latch sits after the increment,
which also closes a non-atomic window where a reader could see `Nodes`/`rerenders` half-updated.

The review finding on the first commit was correct and is fixed in `e3781b6`: `IMeshService.Query<T>`
establishes its subscription + initial snapshot through the IO pool, so creating the marker straight
after `.Subscribe(...)` raced it — the marker could land *inside* the Initial payload, and the latch
would fire on frame 1, baselining before the very frames it exists to wait for. It is now created
only after the first structural render, which makes it demonstrably post-Initial.

**Negative control** (with `SearchResultAccumulator.Apply` returning `true` for `Updated` — the storm
regression this test exists to catch, reintroduced): `Expected value to be 1 … but found 28`. So the
fix costs the test none of its power.

## 2. A live search feed admitted rows its own snapshot excludes (commit 3, PRODUCT)

Necessary but not sufficient: with both test fixes in, 5 local runs still failed 2. Instrumenting the
subscription to print every frame showed the two extra structural renders are one node being **added
and then removed**:

```
[FRAME] Initial structural=True  TestData/storm/b@v1, TestData/storm/a@v1, TestData/storm/c@v1
[FRAME] Added   structural=True  TestData/storm/zz-quiesce@v1     <- baseline latches here (2)
[FRAME] Updated structural=False TestData/storm/a@v2 … @v12
[MonotonicWriteGuard] CONFLICT on TestData/storm/a: a write at Version=11 lost the race …
[FRAME] Added   structural=True  TestData/storm/a/_Activity/write-conflict-12@v1   <- +1
[FRAME] Updated structural=False TestData/storm/a@v13
[FRAME] Removed structural=True  TestData/storm/a/_Activity/write-conflict-12@v1   <- +1  → found 4
[FRAME] Updated structural=False TestData/storm/a@v14 … @v26
```

Nothing about that node changed between the two frames. Only which code path judged it did.

**Mechanism.** `StorageAdapterMeshQueryProvider` emits from two paths that must agree on what a
result is:

- **Snapshot** — `RunQueryNodes` excludes satellite-path nodes (they live in their own storage
  tables, so a content query never sees them on PG) and partition roots for a bare descendants
  listing.
- **Live** — `ProcessBatch`'s write-lag supplement admitted any notified entity that passed
  `QueryEvaluator.Matches`.

`QueryEvaluator.Matches` consults only the query's **filter** and **free-text term**. A catalog
listing — `path:X scope:descendants` — has neither, so it returns `true` for *everything* notified
under `X`. The satellite went in as a structural `Added`; the next change re-queried, correctly
excluded it again, and emitted `Removed`.

This is not test-only. Any live search over an area where a comment, access grant, activity or
notification is written flashes a phantom row and re-renders the whole grid twice. It was already
visible in the tree: `ObservableQueryIntegrationTests` carries comments recording that a framework
`{p}/_Access/…` satellite create was landing on a `scope:subtree` live feed and silently satisfying a
count gate — worked around at the test level, never fixed at the source.

**The fix.** Both exclusions now live in one predicate, `IsExcludedFromResults`, applied by the
snapshot **and** by the supplement — the shared predicate rather than a second copy, since
duplicating them is exactly how they drifted apart. `Deleted` still removes unconditionally: dropping
a path that is gone can never be wrong, and an excluded path was never in the set.

`CatalogSubscription_SatelliteWriteUnderTheRoot_NeverEntersTheResultSet` pins it **deterministically**
— it writes the satellite straight through storage (the exact notification the live pipeline
consumes) and uses an ordinary node created afterwards as the ordered proof the feed processed it. No
conflict has to be provoked.

## Verification

| | |
|---|---|
| `SearchResultStormTest`, 12 consecutive runs (both tests) | **12 pass, 0 fail** — before, on this same tree: 5 runs → 2 failed |
| Fail-without for the new test | `found 2` (the Added/Removed pair above) |
| `test/MeshWeaver.Hosting.Monolith.Test` | 577 passed, 0 failed, 3 skipped |
| `test/MeshWeaver.Graph.Test` | 978 passed, 0 failed |
| `test/MeshWeaver.Query.Test` | 368 passed, 0 failed |
| `test/MeshWeaver.Persistence.Test` | 603 passed, 0 failed |
| `test/MeshWeaver.Search.Test` | 5 passed, 0 failed |
| `-c Release -warnaserror` | clean for `MeshWeaver.Hosting` and `MeshWeaver.Hosting.Monolith.Test` |

## Filed, not fixed here

- **#1249** — why a *strictly sequential* writer provokes that conflict at all: every patch-driven
  write is persisted twice (the post-commit flush the ack chains off, and the per-node persistence
  sampler whose 200 ms-sampled value goes stale in the hub inbox), by two routes that are not ordered
  against each other. No data is lost, but it fires the guard's *"find the writer that adopted a
  stale own-node snapshot"* alarm on a workload with no fork — which is what makes a genuine rollback
  hard to spot. Deliberately out of this PR: that is the persistence path behind #826/#971/#980.
- **#1250** — the same supplement also bypasses RLS, so a live *secure* query can transiently emit a
  node the subscriber cannot read. Fixing it needs an async permission check inside a currently
  synchronous batch step.


rbuergi:
The review finding was correct and is fixed in e3781b6 — but it is **not the whole story**, and this PR should not merge as-is.

## The review finding (fixed)

`IMeshService.Query<T>` establishes its subscription + initial snapshot through the IO pool, so creating the marker straight after `.Subscribe(...)` races it. If the create wins, the marker arrives INSIDE the Initial payload rather than as an ordered post-snapshot change; the latch then fires on frame 1 and the trailing `Added` frames it exists to wait for get counted against the storm — the very failure this test exists to catch. Now the marker is created only after the first structural render, which makes it demonstrably post-Initial.

## Why that is necessary but not sufficient — 5 local runs, 2 failed

```
run 1: FAILED   run 2: passed   run 3: passed   run 4: passed   run 5: FAILED
Expected value to be 2 … but found 4.
```

The residual has a **different mechanism**, visible in the failing runs' log:

```
[MonotonicWriteGuard] CONFLICT on TestData/storm/a: a write at Version=11 lost the race against
the durable Version=12 … Resolved by merging into the durable row: merged=(none);
latest-wins (stale values DROPPED)=Name
[SaveMeshNode] write at Version=11 for TestData/storm/a was refused — the durable row is at
Version=12 (forked lineage: stale activation seed or a second writer). Rebasing this owner onto
the durable truth so subsequent writes land.
```

So the 25-update storm **forks the write lineage** on `storm/a`, and the rebase that resolves it produces additional STRUCTURAL renders (2 extra, hence `found 4`). That is not a baseline-quiescence problem at all, and no amount of baseline work will fix it.

Note what the guard itself says: *"Find the writer that adopted a stale own-node snapshot; do not relax this guard."* Either the test drives updates in a way that legitimately forks the lineage (test defect — it awaits each `Emit`, so worth understanding why that is not enough), or something adopts a stale own-node snapshot under rapid sequential updates (**product defect**, and the more interesting possibility). Until that is answered, weakening or re-baselining the assertion would be papering over it.

Related: this test also just red `main` (`9a05081`) and #1238. It is distinct from #1242's host-death class — that one kills a random test per run with `a test host exited non-zero`; this one fails the same test with a real assertion.

rbuergi:
## The residual is a PRODUCT defect, and here is the frame that settles it

I instrumented the subscription to print every change frame and ran the test until it failed. The
two extra structural renders are not the storm and not the baseline — they are one node being
**added and then removed** from the result set:

```
[FRAME] Initial structural=True  TestData/storm/b@v1, TestData/storm/a@v1, TestData/storm/c@v1
[FRAME] Added   structural=True  TestData/storm/zz-quiesce@v1     <- baseline latches here (2)
[FRAME] Updated structural=False TestData/storm/a@v2 … @v12
[MonotonicWriteGuard] CONFLICT on TestData/storm/a: a write at Version=11 lost the race …
[FRAME] Added   structural=True  TestData/storm/a/_Activity/write-conflict-12@v1   <- +1
[FRAME] Updated structural=False TestData/storm/a@v13
[FRAME] Removed structural=True  TestData/storm/a/_Activity/write-conflict-12@v1   <- +1  → found 4
[FRAME] Updated structural=False TestData/storm/a@v14 … @v26
```

Nothing about that node changed between the two frames. Only which code path judged it did — which
is what makes it a defect rather than a race to be tolerated.

### Mechanism (file:line)

`StorageAdapterMeshQueryProvider` emits from two paths that must agree on what a result is:

- **Snapshot** — `RunQueryNodes` excludes satellite-path nodes (they live in their own storage
  tables, so a content query never sees them on PG) and partition roots for a bare descendants
  listing. `src/MeshWeaver.Hosting/Persistence/Query/StorageAdapterMeshQueryProvider.cs:210-251` (pre-fix).
- **Live** — `ProcessBatch`'s write-lag supplement admitted any notified entity that passed
  `QueryEvaluator.Matches`. `…/StorageAdapterMeshQueryProvider.cs:1355-1366` (pre-fix).

`QueryEvaluator.Matches` consults only the query's **filter** and **free-text term**
(`QueryEvaluator.cs:32-47`). A catalog listing — `path:X scope:descendants` — has neither, so
`Matches` returns `true` for *everything* notified under `X`. The satellite went in as a structural
`Added`; the next change re-queried, correctly excluded it again, and emitted `Removed`.

The `{path}/_Activity/write-conflict-{n}` satellite is written by `MonotonicWriteGuardStorageAdapter`'s
conflict resolution, so it lands right inside a `path:{root} scope:descendants` subscription.

This is not test-only: any live search over an area where a comment, access grant, activity or
notification is written flashes a phantom row and re-renders the whole grid twice.

### Fix

Both exclusions now live in one predicate, `IsExcludedFromResults`, applied by the snapshot **and**
by the supplement — the shared predicate rather than a second copy, since duplicating them is how
they drifted apart. `Deleted` still removes unconditionally.

`CatalogSubscription_SatelliteWriteUnderTheRoot_NeverEntersTheResultSet` pins it **deterministically**
— it writes the satellite straight through storage (the exact notification the live pipeline
consumes) and uses an ordinary node created afterwards as the ordered proof the feed processed it.
No conflict needs to be provoked. Fail-without on this tree reports `found 2` — the Added/Removed
pair above.

### Verification

- 12 consecutive runs of `SearchResultStormTest` (both tests): **12 pass, 0 fail**.
  Before, on this same tree: 5 runs → 2 failed.
- `test/MeshWeaver.Hosting.Monolith.Test` full suite green.
- `dotnet build -c Release -warnaserror` clean for `MeshWeaver.Hosting` and the test project.

### What this does NOT fix — filed as a separate issue

Why a **strictly sequential** writer provokes that conflict in the first place. Every patch-driven
write is persisted twice, by two routes that are not ordered against each other:

1. the post-commit flush the `PatchDataResponse` ack chains off
   (`DataExtensions.cs:1405-1413` → `StoragePostCommitFlush`), and
2. the per-node persistence sampler (`MeshDataSource.cs:1101-1166` → `HandleSaveMeshNode`), whose
   `Sample(200 ms)` value is current when sampled and goes **stale in the hub inbox** while the row
   advances.

The log names the losing writer itself: `[SaveMeshNode] … was refused` is only reachable from
`HandleSaveMeshNode`, and the sampler is the only poster of `SaveMeshNodeRequest` for a per-node hub
— whose writes are ascending by construction, so it cannot have produced the durable v12 ahead of
its own v11. That leaves the flush as the winner.

No data is lost (the merge keeps the newer row), but it fires the guard's *"find the writer that
adopted a stale own-node snapshot"* alarm on a workload with no fork — which is precisely what makes
a genuine rollback hard to spot. Not something to fix by relaxing the guard or widening a window;
one of the two routes has to become authoritative. Kept out of this PR deliberately: that is the
persistence path behind #826/#971/#980, and it does not belong in a test-stability change.

### Adjacent gap noticed, not addressed

The same supplement also bypasses the **RLS** filtering that `RunQueryNodes(useSecurityFilter: true)`
applies, so a secure live query can momentarily emit a node the subscriber cannot read (removed
again on the next re-query — the same add-then-remove shape). Filed separately; fixing it needs an
async permission check inside a currently-synchronous batch step.

PR #1206, opened 2026-08-11
Initial bake sweeps at full speed — only a serving pod's warm-up trickles

rbuergi:
## The measurement (memex-cloud, 2026-08-11, healthy single silo)

162 dynamic NodeTypes compiled in a 7.0-minute sweep — of which the compiles themselves summed to **51 seconds** (p50 0.2s, p90 0.5s, max 3.4s `Store/Plugin`). The fixed `BetweenTypes = 2s` pacing accounted for ~5.4 of the 7 minutes. Roslyn is not the bake's bottleneck; the pacing is.

## The distinction

The trickle exists to keep a **serving** pod's background warm-up from looking like load. An **initial bake** on a readiness-gated pod (the rollout path armed by `PreWarm:GateReadiness`) serves nobody while it bakes — pacing there is pure added rollout time, during which the fleet self-update sits still.

- Gate arms readiness → sweep runs with `TimeSpan.Zero` pacing → whole fleet ready in ~2 minutes.
- Ungated background warm → keeps the 2s trickle, unchanged.
- `PreWarm:BetweenTypes` (TimeSpan) overrides either explicitly; invalid/negative values are logged and ignored.
- Per-type lazy compiles (the "usual compile" path) untouched.

Mechanics: `WarmDynamicTypes` gains an optional `betweenTypes` threaded through `BakeOrFollow` → `WarmPending` to the one `DelaySubscription` site; the hosted service resolves gated-vs-serving and the config override. Pure plumbing — no behavioral change for any existing caller that passes nothing.

What's New entry included (`Category: Feature`).

### DIFF
--- a/src/MeshWeaver.Documentation/Data/WhatsNew/2026-08-11-initial-bake-full-speed.md
+++ b/src/MeshWeaver.Documentation/Data/WhatsNew/2026-08-11-initial-bake-full-speed.md
@@ -1,0 +1,19 @@
+---
+Name: Platform updates prepare in about two minutes instead of twenty
+Category: Feature
+Description: The preparation step of a platform update now runs at full speed, since nobody is being served from the updating side while it runs.
+Icon: Sparkle
+Order: -20260811
+---
+
+# Platform updates prepare in about two minutes instead of twenty
+
+When the platform updates itself, it first rebuilds every dynamic content type against the new
+version. That preparation deliberately paced itself so a busy portal would not slow down — but
+during an update the preparing side isn't serving anyone yet, so the pacing only made updates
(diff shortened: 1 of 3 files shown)


rbuergi:
Post-merge review-thread disposition: (1) the follower re-entry call site was the CS7036 CI round 1 caught — fixed in e6051be02 (pacing now threads through all three BakeOrFollow call sites). (2) A negative betweenTypes cannot throw: Rx.NET normalizes negative relative due-times to zero (Scheduler.Normalize) on the DelaySubscription path, and DynamicTypePreWarmerHostedService already rejects negative PreWarm:BetweenTypes values with a logged warning — the 'never throws' contract holds without a defensive clamp.

Issue #1216, opened 2026-08-11
Batch bake's source discovery is truncated by the default query limit (50) — 169/237 types got zero sources in production


rbuergi:
## Evidence (memex-cloud, 2026-08-11, image ci.2947 — first production run of #1207/#1211)

From Loki, the sweep's own diagnostic line:

```
BatchBake: source discovery resolved 50 Code node(s) into source sets for 237 pending type(s) in one pass
BatchBake: AgenticOffice/Buchungsjournal → NoSources
BatchBake: AgenticPrimer/DiceGame       → NoSources
BatchBake: BusinessRules/Scope          → CompileError
...
DynamicTypePreWarmer: warm-up complete in 00:00:19.36 — compiled=4 compileErrors=59 timedOut=0 skipped=5 contentBroken=169
```

**50** is the default query limit. `NodeTypeBatchBake.ResolveSources` issues ONE global fetch:

```csharp
RunQuery(meshService, $"nodeType:{CodeNodeType.NodeType}")   // → MeshQueryRequest.FromQuery(query)
```

`FromQuery` leaves `Limit` null, so `MeshQuery` falls back to the parsed/default limit and returns the first 50 Code nodes of a mesh that has thousands. Every type whose sources fell outside those 50 resolved to an EMPTY source set (`NoSources`, 169 of 237) or a PARTIAL one (`CompileError`, 59) — types that compile fine on the activation path.

The unit/integration tests pass because a test mesh has fewer than 50 Code nodes, so the cap never binds. This is a pure works-in-test / fails-at-scale defect.

## Why it was contained, and why it still matters

The bake gate refused readiness and the previous image kept serving — no outage. But:

- `PreWarm:BatchBake` defaulted **ON for any readiness-gated pod**, so every gated portal that self-updated would stall its own rollout the same way (live `PreWarm__BatchBake=false` patches applied to memex-cloud / memex / atioz as the immediate stopgap; a helm upgrade would revert them).
- Worse on an UNGATED pod: a wrongly-empty source set classifies as `PreWarmStatus.NoSources`, which by design does **not** gate — so the same bug would quietly bake a fleet of empty assemblies with nothing refusing readiness.

## Fix

[...]

Issue #1207, opened 2026-08-11
Bake as one linked build: batch-drive all NodeType compiles into many DLLs without per-type hub activations


rbuergi:
## Motivation (measured, memex-cloud 2026-08-11)

A healthy sweep compiled 162 dynamic NodeTypes with the Roslyn work summing to **51s** (p50 0.2s, p90 0.5s, max 3.4s) inside a 7-minute wall clock: ~5.4 min was the fixed 2s pacing (removed for gated bakes in #1206), the rest per-type overhead. The base `MetadataReference` set is already process-wide; references are not the cost. What remains structurally is that the bake drives each compile **through the type's own hub activation** (`WarmOne` → `GetMeshNodeStream` → grain activation → compile-watcher settle):

- 237 grain activations + stream settles that exist only to trigger compiles;
- cross-silo sensitivity: on 2026-08-10/11 every one of these round-trips could land on a wedged peer silo and eat a 5-minute timeout — the 20-min/5-h bakes were exactly this, not compilation;
- per-type activity/release bookkeeping serialized behind the activations.

## Proposal — "link it all up in one workspace, build as one big thing into many DLLs"

Initial-bake mode (`initial bake` ≠ `usual compile` — the distinction now wired into the pre-warmer):

1. **One enumeration, one discovery pass**: batch-resolve every type's source queries in one sweep (the enumeration query already returns all definitions; add a single batched Code-node fetch instead of 2 queries × N types).
2. **One dependency graph** (exists: `NodeTypeDependencyGraph`), topological order (exists).
3. **Drive `MeshNodeCompilationService` directly** with pre-resolved inputs — no hub activation, no stream settle, no cross-silo hop. Emit **per-type DLLs exactly as today** (the usual-compile path, ALC unload granularity, and release model stay untouched — "many dlls"). Sibling references resolve from the just-emitted images (in-memory `MetadataReference.CreateFromImage`) instead of waiting for the share.
4. **Bulk state writes**: compile stamps/releases written per type as today but without the watcher round-trip; the gate consumes the same `PreWarmOutcome`s.
5. Optional: parallelize independent graph branches (Roslyn is thread-safe; the 2026-07-28 storm was concurrent *activations*, which this path no longer performs).

🚨 Constraint: this must NOT fork the compiler — one `CompileAsyncCore`, fed batched inputs; the lazy per-type path (`usual compile`) keeps identical behavior and artifacts.

[...]

PR #1217, opened 2026-08-11
Batch bake is opt-in: its source discovery read 50 of thousands of Code nodes in production

rbuergi:
## What happened

First production run of #1207/#1211 (memex-cloud, image `ci.2947`, gated pod):

```
DynamicTypePreWarmer: warm-up complete in 00:00:19.36 —
  compiled=4 compileErrors=59 timedOut=0 skipped=5 contentBroken=169
REFUSING READINESS — 64 NodeType(s) regressed on this image
```

Loki carried the cause in the sweep's own diagnostic:

```
BatchBake: source discovery resolved 50 Code node(s) into source sets for 237 pending type(s) in one pass
```

**50 is the default query limit.** `ResolveSources` issues one global `nodeType:Code` fetch through `MeshQueryRequest.FromQuery`, which leaves `Limit` null, so on a mesh with thousands of Code nodes it returns the first 50. 169 types resolved an **empty** source set (`NoSources`), 59 a **partial** one (`CompileError`) — all of them types that compile fine on the activation path. The tests pass because a test mesh has fewer than 50 Code nodes, so the cap never binds.

## Why this PR flips the default rather than only fixing the limit

- The mode defaulted **ON for any readiness-gated pod**, so every gated portal that self-updates stalls its own rollout the same way. The live `PreWarm__BatchBake=false` patches on memex-cloud / memex / atioz are a stopgap that a helm upgrade would revert.
- On an **ungated** pod it is worse than a stall: a wrongly-empty source set classifies as `PreWarmStatus.NoSources`, a content verdict that deliberately does **not** gate — so the same bug would bake a fleet of empty assemblies with nothing refusing readiness.

Correctness before speed: batch bake now has to be asked for explicitly, and the code says why. The measured win that *did* hold — #1206's zero pacing on gated bakes — is untouched.

## Follow-up (#1216)

Unbounded/paginated discovery, verified through the Postgres provider rather than assumed; a scale test with more Code nodes than the default limit; and — the deeper fix — discovery asserting its own invariant, so a type with a non-empty `Sources` list that resolves to nothing fails the BATCH (falling back to the activation path) instead of being reported as content-broken, which currently makes a discovery bug indistinguishable from genuinely deleted content.

### DIFF
--- a/src/MeshWeaver.Documentation/Data/WhatsNew/2026-08-11-batch-bake-opt-in.md
+++ b/src/MeshWeaver.Documentation/Data/WhatsNew/2026-08-11-batch-bake-opt-in.md
@@ -1,0 +1,19 @@
+---
+Name: Platform updates keep using the proven preparation path
+Category: Fix
+Description: The new batched preparation is now opt-in after it mis-read content at production scale; updates use the proven path by default.
+Icon: Sparkle
+Order: -20260811
+---
+
+# Platform updates keep using the proven preparation path
+
+A new, much faster way of preparing content types during a platform update was introduced earlier
+today. Its first run against a large production mesh read only part of the available source
+content, and reported many healthy types as broken. Nothing was lost and no page changed — the
+update safety-check spotted it and kept the previous version serving — but the fast path is not
+trustworthy yet.
+
+Updates now use the proven preparation path by default, and the faster one has to be switched on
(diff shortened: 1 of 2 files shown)


rbuergi:
Review finding arrived after the merge; fixed forward in #1219.

PR #1219, opened 2026-08-11
"misread" is the past tense; the hyphen was wrong

rbuergi:
Addresses the Copilot review finding on #1217, which landed after the merge: the What's New description used "mis-read" where "misread" is the standard form.

Documentation copy only — no code, no behavior change. No separate What's New entry (this edits one).


rbuergi:
CI red on `MeshWeaver.AI.Test.MeshPluginTest.FullCrudWorkflow_CreateGetUpdateDelete` (shard 3). This PR changes ONE word in a markdown file — it cannot affect that test. main is green at 8c4d21bc1 (the commit this branch is cut from) and this branch is not behind, so it is a flake in that test, i.e. a real race worth pinning rather than re-running away. Not re-running per the repo's no-rerun-to-green rule; leaving this cosmetic PR open until the flake is understood or someone re-runs it deliberately.

rbuergi:
Second CI failure on this PR, and it is **not** this diff — which changes one word in one markdown file.

| run | failing tests | shard |
|---|---|---|
| before main sync | `MeshWeaver.AI.Test.MeshPluginTest.FullCrudWorkflow_CreateGetUpdateDelete` | 3 |
| after main sync | `MeshWeaver.AI.Test.ScriptExecutionInUserHomeTest.Run_StreamsProgressTimely_NotBuffered` · `MeshWeaver.PathResolution.Test.HierarchicalBrowsingTests.Query_WithSkipAndLimit_ReturnsPaginatedResults` | 3, 4 |

**Different tests each time**, and `main` is green on the last five runs (`9398713`, `dad8615`, `eb283c4`, `56421a5`, `9be51fc`) with the same code. The shard messages are `a test host exited non-zero (failure, crash, or timeout kill)` — the signature of a killed/starved test host rather than a specific assertion, so the tests named are whichever ones happened to be running when the host died.

I checked the one that looked suspicious: `Query_WithSkipAndLimit_ReturnsPaginatedResults` ran right after #1232 changed `QueryParser`, but main carries #1232 and is green, so it is not a regression from it.

Not re-running (the repo's no-rerun-to-green rule: a flake is a real race, and re-running hides it). Leaving this cosmetic PR open — it is worth nothing next to the signal it is carrying, and whoever picks up the shard-host instability can use the table above.

Issue #1218, opened 2026-08-11
Starved cross-silo source discovery produces FALSE CompileErrors that the bake gate reads as image regressions


rbuergi:
## Evidence (memex-cloud, 2026-08-11, ci.2947, activation-path bake during a rollout)

Sampled 56 outcomes of an in-flight gated bake:

```
outcome mix: TimedOut=16  CompileError=14  Compiled=26
gap s: p50=0.25  p90=60.00  max=60.2   total=908s
gaps >60s: 15, costing 900s   (i.e. 99% of the elapsed bake is 60s timeouts)
```

Two distinct defects show up here.

### 1. The 60-second timeouts dominate the bake, and they are a ROLLOUT artifact

Every slow gap is exactly ~60.0 s — the `SubscribeRequest` timeout, not the 5-minute per-type budget. During a rollout the OLD pod still holds the existing activations for these types, so the new pod's per-type activation round-trips to a busy peer and times out. `[PreferLocalPlacement]` (#1204) only biases where a NOT-yet-activated grain is placed; it cannot move an activation that already exists on the other silo.

At 16 timeouts per 56 types, a full 237-type sweep costs ~68 × 60 s ≈ **68 minutes of pure timeout**. This is the same cost centre as the 2026-08-10/11 incident, and it is inherent to driving compiles through hub activations while another silo is serving.

### 2. Starved source discovery is reported as a COMPILE ERROR, and the gate treats it as a regression

```
DynamicTypePreWarmer: BusinessRules/Scope → CompileError
  CS0246: The type or namespace name 'ScopeLibrary' could not be found
  CS0103: The name 'ScopeTestsArea' does not exist in the current context
```

`BusinessRules/Scope` **compiled successfully in the same day's baseline run** (18 m 55 s sweep, `compiled=228`). Nothing about the type changed. Its source queries returned empty/partial under the same cross-silo starvation as #1 (core #694), so Roslyn compiled it against nothing and produced a genuine-looking `CS0246`.

That is indistinguishable, at the gate, from a real image regression — so `BakePhase.Regressed` sticks, readiness is refused, and the rollout stalls on types that are perfectly healthy. 14 of 56 sampled types were in this state.

## Why this matters beyond one rollout

[...]

PR #1220, opened 2026-08-11
Batch bake's global source fetch could not see a single Code node Postgres stores (#1216)

rbuergi:
Fixes the batch bake's source discovery so it reads the whole mesh — and, more importantly, makes it
**refuse to answer** when it cannot. Closes the defect in #1216.

## Production evidence (memex-cloud, 2026-08-11, image ci.2947)

```
BatchBake: source discovery resolved 50 Code node(s) into source sets for 237 pending type(s) in one pass
BatchBake: AgenticOffice/Buchungsjournal → NoSources
BatchBake: AgenticPrimer/DiceGame       → NoSources
BatchBake: BusinessRules/Scope          → CompileError
...
DynamicTypePreWarmer: warm-up complete in 00:00:19.36 — compiled=4 compileErrors=59 timedOut=0 skipped=5 contentBroken=169
```

## What was actually wrong — TWO mechanisms, not one

The issue identified the row cap. Building the scale test found a second, larger one: **a cap can only
ever produce a PARTIAL source set, never an empty one**, so it does not explain 169 empties.

**1. The global fetch was capped at 50.** `ResolveSources` issued
`MeshQueryRequest.FromQuery("nodeType:Code")`, leaving both `MeshQueryRequest.Limit` and
`ParsedQuery.Limit` null. "No limit" is not read the same way by every backend — see the table below.

**2. The global fetch could not see Postgres Code nodes AT ALL.** On the partitioned backend a node's
table is chosen by **path segment**: a `Source/` or `Test/` segment routes to the per-schema `code`
satellite table. Resolving a table from a `nodeType:` filter uses a *different* map, and `Code` is
deliberately absent from it — `SatelliteTableMapping.Defaults` says so outright: *"`Source`/`Test` are
primary code content sharing the `code` table (no leading underscore — matched as a path segment, and
**not nodeType-resolvable**)"*. So an unpinned `nodeType:Code` query fans out over every schema's
`mesh_nodes` and returns none of the Code nodes that live where Code nodes actually live.

Measured on a two-partition PG mesh holding 60 Code nodes (all in `{schema}.code`, `mesh_nodes` empty):

| query | rows |
|---|---|
| `nodeType:Code` | **2** (both from the static catalog; zero from Postgres) |
| `namespace:*/Source scope:subtree nodeType:Code` | **60** |
| `namespace:*/Test scope:subtree nodeType:Code` | 0 (nothing seeded under `Test/`) |

That is the shape of the production numbers: the ~59 `CompileError` types were the static/doc ones
that the capped `nodeType:Code` query partially covered, and the 169 `NoSources` were the PG-stored
ones it could not see at all.

## Limit semantics, verified per backend (not assumed)

| provider | absent limit | `request.Limit` honoured? |
|---|---|---|
| `StorageAdapterMeshQueryProvider.LoadCap` (`:381`) | **unbounded** | yes (`request.Limit ?? parsed.Limit`) |
| `MeshQuery.ClipMergedInitial` (`:817`) | **unbounded** | yes |
| `PostgreSqlMeshQuery` (per-schema, pinned) (`:276`, `:455`) | **unbounded** | yes — copies it into `parsedQuery` |
| `CosmosMeshQuery` (`:121`) | **unbounded** | yes |
| `PostgreSqlCrossSchemaQueryProvider.QueryAcrossSchemasAsync` (`:364`) | **50** | **no — dropped** |
| `SnowflakeCrossSchemaQueryProvider` (`:408`) | **50** | **no — dropped** |

Paging is not available as an alternative: `public.search_across_schemas` takes a LIMIT but **no
OFFSET**, and its `ORDER BY n.last_modified DESC` is not a total order, so a Skip/Limit loop would
silently drop and duplicate rows across pages — worse than one big page.

## The fix

1. **Discovery states an explicit ceiling and then ASSERTS it was not reached.** Every discovery query
   carries `limit:25000` in the query **text** (the cross-schema fan-out reads only `ParsedQuery.Limit`)
   *and* on the request. A result that reaches the ceiling is a discovery **failure**, not a result.
2. **The global fetch is the union of the three shapes that actually cover both tables** —
   `nodeType:Code` (mesh_nodes + the static catalog) plus the two wildcard-namespace shapes that route
   to `code`. Still O(1) in the number of pending types: three queries whether the mesh has 3 types or 300.
3. **`request.Limit` is no longer dropped by the cross-schema fan-out** (`PostgreSqlPartitionedMeshQuery`,
   and the identical twin in `SnowflakePartitionedMeshQuery`) — the same propagation the per-schema
   delegate already does. The `?? 50` default for a request that states *no* limit is untouched; an
   unanchored UNION over every partition schema needs some bound, and changing that default is a
   separate decision.
4. **The invariant — the half that stops the CLASS, not the instance.** If a type with a NON-EMPTY
   declared `Sources` list resolves to an EMPTY source set, and the type's own persisted
   `CurrentSourceVersions` snapshot does not corroborate it as explicitly empty, that is a **discovery
   failure**: the whole batch is abandoned, loudly, naming every offending type, and the run falls back
   to the activation-driven sweep. It never emits a per-type verdict.

   This matters because `NoSources` is deliberately **non-gating** (it means "the content was deleted",
   #1204) — so a broken discovery pass did not look broken, it looked like content breakage. On an
   **ungated** pod the same pass would have baked a fleet of empty assemblies and stamped every one of
   them good, with nothing refusing readiness. The genuinely-deleted case (#1204) still works: it has
   the corroborating empty snapshot.

## What is now guaranteed, and what is not

**Guaranteed (proven by test):**
- Discovery resolves every type's full source set on a real partitioned Postgres mesh holding more Code
  nodes than the default cap, across two partitions, and resolves **exactly** what the activation path
  resolves (compared node-for-node).
- A truncated or unestablished source set can no longer become a per-type verdict — it fails the batch.
- The genuinely-deleted-content case still reports the non-gating `NoSources`.

**Not verified:**
- **Nothing has run at production scale.** The scale test uses 60 Code nodes across 2 partitions; prod
  has thousands across ~136 schemas. The 25 000 ceiling is a judgement, not a measurement, and the
  memory cost of holding every Code node's content in one map is unchanged from the original design.
- **Snowflake is compile-verified only** — the fan-out fix there is a line-for-line twin of the
  Postgres one, but no Snowflake instance ran it.
- The `nodeType:Code`-is-not-nodeType-resolvable asymmetry is left as-is: it is documented, deliberate,
  and any query outside the batch bake that filters `nodeType:Code` unpinned still cannot see the `code`
  table. Widening it is a routing change with a much larger blast radius and belongs in its own PR.

## `PreWarm:BatchBake` stays OFF by default

`DynamicTypePreWarmerHostedService.KickWarmup` still defaults `batchBake = false` (PR #1217).
**Turning it back on is a separate decision for @rbuergi after this is verified at scale in prod** —
this PR makes the path correct and self-refusing, it does not re-enable it.

## Tests

- `BatchBakeSourceDiscoveryScaleTests` (new, Postgres/Testcontainers) — 60 Code nodes across two
  partitions. **Verified RED before the fix** (`batch discovery: TypeA=0, TypeB=0`, "Expected collection
  to have 30 item(s) … but found 0") and green after. Also asserts, directly against the database, that
  Code nodes live in `{schema}.code` and not `mesh_nodes` — so a future satellite-routing change that
  re-breaks the batch fails *here*, naming the mechanism.
- `NodeTypeBatchBakeDiscoveryInvariantTest` (new, Monolith) — the invariant fails the batch with no
  corroboration, accepts an empty set when the snapshot corroborates it, and does not over-fire on a
  healthy type.

Local: `MeshWeaver.Hosting.Monolith.Test`, `MeshWeaver.Graph.Test` (956/956),
`MeshWeaver.Hosting.PostgreSql.Test`, `MeshWeaver.Documentation.Test` (20/20) green; every touched
project builds clean under `-c Release -warnaserror`.

Related: #1218 (the activation path's twin of "a starved source set becomes a CompileError"), #1207
(design), #1211 (implementation), #1204 (content-vs-image verdicts), #1217 (opt-in default).

### DIFF
--- a/src/MeshWeaver.Documentation/Data/WhatsNew/2026-08-11-update-preparation-reads-all-of-your-content.md
+++ b/src/MeshWeaver.Documentation/Data/WhatsNew/2026-08-11-update-preparation-reads-all-of-your-content.md
@@ -1,0 +1,25 @@
+---
+Name: Update preparation now reads all of your content
+Category: Fix
+Description: The faster update-preparation path used to read only the first page of source content on a large mesh; it now reads all of it, and refuses to guess when it cannot.
+Icon: Sparkle
+Order: -20260811
+---
+
+# Update preparation now reads all of your content
+
+When the platform updates, it prepares every content type on your mesh so the first person to open
+a page does not have to wait. The faster preparation path introduced earlier today asked the mesh
+for its source content in a single request — and on a large mesh that request came back with only
+the first page of results. Types whose content sat beyond that page were prepared against nothing,
+and then reported as broken even though nothing was wrong with them.
+
+The request now asks for everything, and — this is the part that matters — it **checks** that it
+got everything. If the answer could be incomplete, or if a type that says it has source content
+comes back with none and nothing else on the mesh agrees, preparation stops and hands the whole
+update back to the slower, proven path. It no longer reports a type as broken on evidence it never
+actually gathered.
+
+You would have seen this as an update that took much longer than usual, or as content types
+briefly showing a preparation error after a platform update. Nothing was lost either way: the
+update safety-check kept the previous version serving.
--- a/src/MeshWeaver.Hosting.PostgreSql/PostgreSqlPartitionedMeshQuery.cs
+++ b/src/MeshWeaver.Hosting.PostgreSql/PostgreSqlPartitionedMeshQuery.cs
@@ -883,7 +883,13 @@ public sealed class PostgreSqlPartitionedMeshQuery : IMeshQueryProvider
         // this shape and resolved 50 Code nodes out of thousands). A request that states no limit at
         // all still gets the fan-out's default: an unanchored UNION over every partition schema
         // needs SOME bound, and changing that default is a separate decision.
-        if (request.Limit.HasValue)
+        // 🚨 POSITIVE limits only. `Limit <= 0` means "do not clip" upstream —
+        // MeshQuery.ClipMergedInitial applies a limit only `if (effectiveLimit is int limit &&
+        // limit > 0)` — but propagated into SQL a zero becomes a literal `LIMIT 0`, i.e. ZERO ROWS
+        // returned for a caller that asked for everything. That is the same silent-empty-result
+        // failure this very PR exists to fix (#1216: discovery that cannot see rows reports the
+        // content as missing), so it must not be re-introduced one layer down.
+        if (request.Limit is > 0)
             queryForSql = queryForSql with { Limit = request.Limit };
 
         var userId = GetEffectiveUserId(request);
--- a/src/MeshWeaver.Hosting.PostgreSql/PostgreSqlPartitionedMeshQuery.cs
+++ b/src/MeshWeaver.Hosting.PostgreSql/PostgreSqlPartitionedMeshQuery.cs
@@ -874,6 +874,18 @@ public sealed class PostgreSqlPartitionedMeshQuery : IMeshQueryProvider
         if (!string.IsNullOrEmpty(queryForSql.Path) && queryForSql.Path.Contains('*'))
             queryForSql = queryForSql with { Path = null, Scope = QueryScope.Exact };
 
+        // 🚨 Carry the REQUEST-level limit into the parsed query — the same propagation the
+        // per-schema delegate does (PostgreSqlMeshQuery: `if (request.Limit.HasValue) parsedQuery =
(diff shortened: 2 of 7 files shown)


rbuergi:
Both findings addressed in 696a3276a: the fan-outs now guard with `request.Limit is > 0`, so a non-positive limit keeps its upstream "do not clip" meaning instead of becoming a literal SQL `LIMIT 0`. Good catch — that is the same silent-empty-result class this PR fixes, one layer down.

Issue #974, opened 2026-08-08
Three more sites report an availability failure as a definitive negative (permission-fold fault → "Access denied", read timeout → "Not found", cold cache → "user unknown")


rbuergi:
## Summary

Split out of **#637**. PR #947 fixed the API-token leg and PR #970 fixes the identity/authorization leg, but **three more sites collapse an availability failure into a definitive negative answer** — the same defect class, each needing its own change.

An availability failure answered as a definitive negative is worse than an error, because it is *actionable-looking*: the user is told something specific and false, and acts on it. #637's headline case was a storage stall telling a signed-in user they were not onboarded and redirecting them to the sign-up form.

## The remaining three

| Site | Reports a … | As |
|---|---|---|
| `AccessControlPipeline` | permission-fold **fault** | "Access denied" |
| `MeshOperations.FetchNode` | read **timeout** | "Not found" |
| `UserIdentityCache` | **cold cache** | "user unknown" |

Each is a lie with a different cost. "Access denied" sends a correctly-entitled user to ask for permissions they already have. "Not found" invites deleting or recreating a node that exists. "User unknown" is the same trap #637 opened — it can drive onboarding or provisioning logic for a user who is already onboarded.

## Why these are not a quick follow-up to #970

Unlike the identity reads, these cannot be fixed by classifying at a single bounded seam. Each needs either a **tri-state carried through the permission fold** (so "we could not determine" survives the fold instead of collapsing into "no") or **a new `ErrorType` on the bus** so the condition can cross a hub boundary without being flattened. That is why #970 deliberately stops short of them rather than reaching for a partial fix.

## The rule these should follow

Decide the distinction **where the condition is known** — inside the one place a budget or a fold is applied — and never reconstruct it upstream by sniffing a message string. That is the shape #970 uses (`IdentityReadOutcome<T>` decided inside `IdentityRead.Bounded`), and it is what makes the classification survive rewording.

Also worth applying #970's lesson about its own fix: after removing a `catch`, sweep for edges that can now throw *past* the classifier. That sweep found a third instance in #970 itself — `LoadDbRolesAsync` resolving the hub and workspace outside the classified chain, where a hub mid-disposal would have escaped as a 500.

## Not band-aids

Widening a timeout, retrying, or catching-and-defaulting are all excluded: the goal is to name the condition correctly, not to make it rarer.

PR #1223, opened 2026-08-11
fix(content): a denied file answers like a missing one, not a 500 that names the reason

### LINKED ISSUE #587: Hardening: /static/ endpoint serves content collections without authentication or access control

rbuergi:
## What

A permission denial on `/api/content` was reported as an unexpected error instead of as "not accessible". The read check itself is fine — an anonymous caller **is** denied Read on the owning node. What leaked was how that denial came back.

`/api/content` is the route content was **moved to in order to close #587**, where `/static/storage/content/{node}/{file}` "exposed every partition's uploads at a fully predictable URL to anyone who asked". The reporting bug re-opened the *enumeration* half of that hole.

## Why it happened

A denial reaches the endpoint only as a `DeliveryFailureException` — the owning hub refuses the `GetDataRequest` by posting a `DeliveryFailure`, which `hub.Observe` surfaces through `onError`. So the documented *"config reads back null ⇒ 404"* path never ran for a denial: the stream faults instead of delivering null, and the generic `.Catch` mapped it to `Results.Problem`.

That gave any unauthenticated caller:

- **a 500-vs-404 oracle** — `500` = the node exists and you are denied, `404` = no such file. One request per guess maps every private partition by URL alone, no credential needed.
- **the denial text verbatim**, live on prod today:
  ```json
  {"status":500,"detail":"Error retrieving content: Access denied: user 'Anonymous' lacks Read permission on 'Doc'"}
  ```
  — the permission model, the principal, and confirmation the node exists.

## The fix

A refused read now answers exactly like a missing one: `404`, no detail. `AccessControlPipeline` already logs the real reason server-side. Genuine faults keep `Problem` but no longer echo `ex.Message`.

## Tests

Two new, and **neither existed before** — every other `/api/content` assertion in `StaticContentUnmountedTest` passes `asAdmin: true`, so the suite would have stayed green if the route had stopped denying anonymous callers entirely.

- `ContentRoute_RefusesAnAnonymousCaller` — 404, and the body leaks neither the secret nor the denial reason.
- `ContentRoute_DeniedAndMissing_AreIndistinguishable` — the oracle itself.

`StaticContentUnmountedTest` 17/17. `MeshWeaver.Hosting.Blazor` clean under `-c Release -p:CIRun=true -warnaserror`.

## ⚠️ Separate issue this does NOT fix

While testing I uploaded a file to my own **private** partition and it serves **200 image/png to an anonymous caller**, stable across repeated requests:

```
curl https://memex.meshweaver.cloud/api/content/rbuergi/content/_icontest.png  →  200 image/png
```

`rbuergi` has three named grants and no anonymous grant. Since `500` turns out to be the denial signature, a `200` means the check **passed** for Anonymous there — unlike `Doc`, which correctly denies. That points at effective-permission data granting Anonymous Read on a user partition root, and it needs a query against prod's permission tables. **This PR does not address it.**

### DIFF
--- a/src/MeshWeaver.Documentation/Data/WhatsNew/2026-08-11-content-denial-is-not-an-error.md
+++ b/src/MeshWeaver.Documentation/Data/WhatsNew/2026-08-11-content-denial-is-not-an-error.md
@@ -1,0 +1,15 @@
+---
+Name: Files you may not read answer like files that are not there
+Category: Fix
+Description: A denied content file now returns the same 404 as a missing one, instead of a 500 that named the reason.
+Icon: Sparkle
+Order: -20260811
+---
+
+# Files you may not read answer like files that are not there
+
+Requesting a file from a space you do not have access to used to fail differently from requesting a file that simply does not exist: the first returned a server error, the second a plain "not found". Anyone could tell the two apart without signing in, which made it possible to work out which spaces exist and are private just by trying addresses.
+
+The server error also quoted the reason back to the caller, naming the permission that was missing and the space it belonged to.
+
+Both are fixed. A file you are not allowed to read now answers exactly like a file that is not there — a plain "not found", with no explanation attached. The real reason is still recorded in the server log, where administrators can see it. Nothing changes for files you are allowed to read.
--- a/src/MeshWeaver.Hosting.Blazor/BlazorHostingExtensions.cs
+++ b/src/MeshWeaver.Hosting.Blazor/BlazorHostingExtensions.cs
@@ -388,8 +388,30 @@ public static class BlazorHostingExtensions
             // No await on hub round-trips — chain via SelectMany; deadlock surface
             // (await pathResolver.ResolvePath / hub.Observe) eliminated.
             return ResolveContentFile(context, mainHub, path, caller)
-                .Catch<IResult, Exception>(ex =>
-                    Observable.Return(Results.Problem($"Error retrieving content: {ex.Message}")))
+                // 🚨 A DENIAL IS NOT AN ERROR — it must be indistinguishable from absence.
+                // The owning hub refuses a GetDataRequest the caller lacks Read on by posting a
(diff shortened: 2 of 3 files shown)


rbuergi:
Superseded by #1229, which carries forward everything here that still applies after #1224 (`4874ed33e`) landed.

**Kept** — both tests, adapted: `ContentRoute_RefusesAnAnonymousCaller` and `ContentRoute_DeniedAndMissing_AreIndistinguishable`. They pin properties nothing else in the suite covers (every other `/api/content` test is `asAdmin: true`), and I verified they pass unmodified against #1224's implementation — so there was no hidden defect to report.

**Kept** — the no-`ex.Message` hardening. #1224 did **not** carry this part: its fallback arm still returned `Results.Problem($"Error retrieving content: {ex.Message}")`. That message is hub-authored diagnostics (`"Access denied: user 'Anonymous' lacks Read permission on 'Doc'"` — the permission model, the principal, and proof the node exists) going to an anonymous caller over a public route. The detail now goes to the server log instead. Worth noting the two typed arms would not have covered it anyway: they are property patterns over `DeliveryFailure.Failure`, which is `null` for the string-only constructors and `Unknown`/`NotFound`/`Exception` for anything a hub reports unclassified.

**Dropped** — the blanket `ex is DeliveryFailureException ? Results.NotFound()`. #1224 deliberately preserves the issue-#974 tri-state (`Unauthorized` → 404, `Unavailable` → 503), and collapsing everything to 404 would turn "the permission probe is degraded" into a confident, cacheable "this does not exist" — a correctness regression dressed as a security fix, and one that discards a distinction the access pipeline goes to real trouble to carry. #1229 adds `ContentFailureMappingTest`, which pins the tri-state arm by arm so this cannot be reapplied silently.

Superseded rather than rebased because this PR's title and description are about the change being dropped. Closing — no work is lost.

PR #1224, opened 2026-08-11
fix: gate content-collection files on the owning node's real permissions

### LINKED ISSUE #587: Hardening: /static/ endpoint serves content collections without authentication or access control

rbuergi:
## The disclosure

`/api/content/{node}/{file}` served a **private user partition's uploaded files to callers who held nothing on that partition — anonymous HTTP callers included**. Reproduced against production before the fix:

```
$ curl -s -o /dev/null -w '%{http_code} %{content_type} %{size_download}' \
    https://memex.meshweaver.cloud/api/content/rbuergi/content/_icontest.png
200 image/png 1389
```

No cookie, no token. The URL scheme is fully predictable, and this is the route `UserContextMiddleware` calls "the access-controlled route" — the one built to close #587, where `/static/storage/content/{node}/{file}` exposed every partition's uploads.

## Root cause

`UserNodeType.WithUserNodePublicRead` (`src/MeshWeaver.Graph/Configuration/UserNodeType.cs:224-226`, applied at :117) grants "every signed-in user" Read on a user-partition hub, spelled:

```csharp
.AddHubPermissionRule(Permission.Read, (_, userId) => !string.IsNullOrEmpty(userId));
```

Two compounding defects:

1. **An unauthenticated caller is not nameless.** `UserContextMiddleware` resolves it to `WellKnownUsers.Anonymous` — the string `"Anonymous"` — which is perfectly non-empty, so it satisfied the rule. `AccessControlPipeline` then *drops* the pending check (`:133-138`) and returns `next.Invoke(...)` (`:163`) — the real `PermissionEvaluator` is **never subscribed**. The bitter irony: a caller with a *null* context would have been denied; naming the anonymous caller correctly is what opened the door. Because `userId` is non-empty, the pipeline's "ANONYMOUS delivery" warning never fired either — no log signal.
2. **The rule is path-blind.** Its predicate takes only `(delivery, userId)`, so it satisfied **every** `[RequiresPermission(Read)]` message on that hub — including the collection-config `GetDataRequest` the content route relies on as its gate. That is why an **authenticated non-member** was served the bytes too.

The route's own contract ("the file is gated by exactly the decision an ordinary node read makes") was therefore true only while no hub rule short-circuited the check. One did.

## Blast radius

`UserNodeType` is the **only** live carrier — `NodeAccessExtensions.WithPublicRead` has zero call sites. So the exposure is precisely **user-partition roots**: `/api/content/{user}/content/…` and `/api/content/{user}/attachments/…`, both registered `IsStatic = true` (`MemexConfiguration.cs:858,871`). Space partitions and child nodes under a user partition were correctly gated throughout.

## The fix

* **`WellKnownUsers.IsAuthenticated`** — one shared definition of "signed in" that excludes the anonymous and public pseudo-subjects, used by both rule sites so neither can drift back to the emptiness spelling.
* **The content route asks the evaluator directly**, about the **resolved owning node** (`resolution.Owner`, not the first path segment), instead of inheriting the answer from a config read a hub-wide rule can satisfy on its own. Anonymous callers go through `AnonymousGate` — the same fail-closed predicate the crawler-facing `/api/og` route uses, so the two surfaces now answer the same question the same way.
* **Fail closed**: a refused read answers exactly as a missing one (no existence oracle over a predictable URL scheme); an *undetermined* verdict refuses with 503 rather than serving, keeping the tri-state #974 established.

Public pages are unaffected, **verified rather than assumed**: `/api/og/BusinessRules.png` returns 200 in production, which means `AnonymousGate` already allows that partition — and that is now the same predicate the content route uses.

## Tests

`ContentFileAnonymousAccessTest` drives the **real** route over HTTP against a real monolith mesh and the real `PermissionEvaluator` (no mocks), on a **real `User` node** — the shape a Markdown fixture cannot reproduce, and the reason this was never caught: the existing suite exercised `/api/content` only as an admin.

Against the unfixed tree 3 of 5 fail, including the bytes being served to anonymous *and* to an authenticated stranger; all 5 pass after.

| Case | Before | After |
|---|---|---|
| anonymous, partition without an anonymous grant | **served** | 404 |
| anonymous, via explicit collection segment | **served** | 404 |
| authenticated non-member | **served** | 404 |
| anonymous, partition **with** an anonymous grant | 200 | 200 |
| the grant holder | 200 | 200 |

Local suites green: Hosting.Blazor 306, Security 388, Graph 975, Memex.Portal.Shared 291, Documentation 20. Release `-warnaserror` clean.

### DIFF
--- a/src/MeshWeaver.Documentation/Data/WhatsNew/2026-08-11-private-partition-files-are-no-longer-served-to-strangers.md
+++ b/src/MeshWeaver.Documentation/Data/WhatsNew/2026-08-11-private-partition-files-are-no-longer-served-to-strangers.md
@@ -1,0 +1,36 @@
+---
+Name: Private partition files are no longer served to strangers
+Category: Fix
+Description: A file uploaded to a private partition could be fetched by anyone with the URL, without signing in. Content files now require exactly the same permission as the node that owns them.
+Icon: Sparkle
+Order: -20260811
+---
+
+# Private partition files are no longer served to strangers
+
+A file stored in a partition's content collection — an upload, an attachment, an image — could be
+fetched over `/api/content/{partition}/content/{file}` by a caller who had never signed in, and by
+a signed-in caller who held no permission on that partition. Only the URL was needed, and the URL
+scheme is entirely predictable.
+
+The route was always meant to be the access-controlled one. It asked the owning node's hub for the
+collection's configuration, a request that carries a read-permission requirement, and treated that
+as the gate. The gate was real, but one rule quietly satisfied it: user partitions grant read
+access to "every signed-in user", and the rule that decided who counted as signed in tested only
+that the caller had a name. An unauthenticated caller does have one — it is called `Anonymous` —
+so it passed, and with it the whole hub's read check, the collection's configuration, and the file
+itself. A caller with no identity at all would have been refused; naming the anonymous caller
+correctly is what opened the door.
+
+Two things changed. "Signed in" is now a single shared definition that excludes the anonymous and
+public placeholder identities, so no rule can drift into admitting them again. And the content
+route now asks the permission evaluator directly, about the node that owns the file, instead of
+inheriting the answer from a configuration read that a hub-wide rule can satisfy on its own. For a
+logged-out visitor that is the same check the public share-card and crawler pages already use: a
+file is served anonymously only where its partition carries an explicit grant for anonymous
+readers.
+
+Public pages are unaffected — plugin covers, course landing pages and share cards carry that
+explicit grant and keep serving to logged-out visitors. What changes is everything else: a refused
+file now answers exactly as a missing one, so the response cannot be used to discover which files
+exist, and a permission check that cannot reach a verdict refuses the read rather than allowing it.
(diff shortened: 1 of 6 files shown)


rbuergi:
### Coordination note — overlap with the sibling error-handling fix

This PR **overlaps** a fix another agent developed in parallel: mapping a denied content read to a bare `404` instead of a `500` that echoes `ex.Message` (an enumeration oracle — `500` = exists-but-denied vs `404` = missing — plus disclosure of the permission model to anonymous callers).

That work is real and correct, and I do not want to duplicate or race it. The overlap here is `ContentFailure(...)` in `BlazorHostingExtensions.cs`, which maps `DeliveryFailureException`:

- `ErrorType.Unauthorized` → `404` with a fixed, neutral body (no `ex.Message`)
- `ErrorType.Unavailable` → `503` (fail closed, but preserves the #974 tri-state: "could not check" is not "you may not")

**Proposed resolution**, whichever the coordinator prefers:

1. **That PR ships first** → I rebase this branch onto it and drop my `ContentFailure` hunk, keeping only the authorization change; or
2. **This PR ships first** → that branch rebases and keeps its two tests (`ContentRoute_RefusesAnAnonymousCaller`, `ContentRoute_DeniedAndMissing_AreIndistinguishable`), which are worth having regardless — they pin the indistinguishability property from a different angle than the suite here.

Either way the two fixes are complementary, not competing: theirs stops the **response** from disclosing, this one stops the **bytes** from being served. Their own assessment is correct that the error-handling fix does not close the `200`-on-a-user-partition case — that is what the authorization change here does.

PR #1234, opened 2026-08-11
fix(export): PDF and DOCX render embedded layout areas instead of losing them

rbuergi:
## The defect

A document can embed a live view with `@@("…/area/OgCard/…")`. **Exported to PDF or DOCX, it did not survive** — and both failure modes were silent.

**Content-fidelity path — printed the embed's SOURCE TEXT.** `Ast/DocumentBuilder` built its own Markdig pipeline:

```csharp
_pipeline = new MarkdownPipelineBuilder()
    .UseAdvancedExtensions()
    .UsePageBreaks()
    .Build();          // ← no LayoutAreaMarkdownExtension
```

The framework registers `LayoutAreaMarkdownExtension` at `MarkdownExtensions.cs:68`; this pipeline omitted it, so `@@(…)` was never recognised as a block. It fell through to a `ParagraphBlock` and the renderer faithfully printed the author's markdown source into the PDF.

**Pixel path — printed a BLANK.** It parses with the framework pipeline (`SlidePrintComposer.cs:77` → `MarkdownViewLogic.Render`), so it correctly emitted the empty `<div class='layout-area'>` anchor a live client fills. But the print document is loaded from `file://` under a CSP of `default-src 'none'`, with `--host-resolver-rules=MAP * ~NOTFOUND` and a dead proxy (`HeadlessChromiumPdfRenderer.cs:230,236`). The browser could never fill it.

Neither logged anything, and **no test anywhere covered `@@(…)` in either export path** — which is why it rotted.

## The fix — remove the drift, not just the symptom

Two pipelines and two control walks were the root cause, so there is now one of each:

- **`Ast/ExportMarkdownPipeline`** — one pipeline used both to *find* embeds and to *render* the document. Whatever the resolver can find, the builder can render, by construction.
- **`Html/AreaMarkupRenderer`** — one walk of the live control tree, producing a `MarkupNode` tree that is serialized **either** to HTML (email / browser print) **or**, via the new **`Html/MarkupToDocument`**, to the `DocumentElement` model. A card grid becomes a real `TableElement`: `w:tbl` in Word, a real table in QuestPDF — selectable and editable, not text pretending to be a table.
- **`Html/DocumentAreaResolution`** — reading an area is reactive and cross-hub; the builder walk is synchronous. So areas resolve in a **prior pass** and the walk looks them up by key — the same split the export already uses for client-captured Mermaid/Math SVGs. Nothing blocks on a stream mid-walk.
- **Pixel path** resolves anchors *before* printing and *before* asset inlining, so content the area itself references gets inlined too. The DOM is round-tripped **only when an anchor is present**, so a deck with no embeds prints byte-identically to today.

### Reuse, not duplication — the shared lift

The area resolver, control renderer, table sizer, markup tree and sanitizer added by #1222 for email are **not email-specific**. They move `Email/` → `Html/` with format-neutral names (`LayoutAreaResolver`, `AreaMarkupRenderer`, `DocumentHtmlComposer`, `MarkupNode`, `HtmlDowngrade`, `TableSizer`, `MarkupStyles`, `DocumentHtmlOptions`), via `git mv` so history follows. Only `EmailImageInliner` (cid: parts) stays email-only. `MarkupNode` rather than `HtmlNode` deliberately — `HtmlNode` would collide with HtmlAgilityPack.

### Unresolvable areas

Previously the email path **removed** the anchor — a silent gap. An area that cannot be resolved (permission denied, unknown area, fetch failure) now leaves a short **localized notice naming it** (`export.areaUnavailable`, en + de, plus the React vendored catalogs). Never a blank, never an error dump, and one bad embed never fails the export.

## Tests

The defect existed because nothing asserted it, so this is the substance of the PR.

- **`DocumentExportLayoutAreaTest`** (real mesh, `MonolithMeshTestBase`, no mocking) — drives the real `ExportDocumentRequest` → `.csx` template → `ActivityLog` pipeline (the templates compile at **runtime**, so `dotnet build` cannot check them; only a test that executes them can). Asserts the **PDF bytes** read back with PdfPig and the **DOCX `word/document.xml`** both contain the area's content, contain no `@@(`, and that the DOCX carries a native `<w:tbl>`. Plus the unresolvable-area contract and the pixel print HTML (asserted **without** a browser, since the browser is deliberately not in the image).
- **`ExportMarkdownPipelineTests`** — pins that the export pipeline parses every embed form **including the reported `area:OgCard?urls=…`**, that an embed never renders as its own source, and that `\newpage` still works after the pipeline change.
- **`MarkupToDocumentTests`** — pins the card grid → 2-column table mapping, per-cell line structure, and clickable links.

All green; full solution builds clean with CI's flags (`-c Release -p:CIRun=true -warnaserror`).

## Scope note: why PDF is not routed through the browser

The alternative considered was routing document PDF through the existing headless-Chromium renderer. **Not viable today, on evidence:**

1. **Chromium is not in the deployed image.** `deploy/base-images/portal-ai/Dockerfile:29-38` is the sole `apt-get` in the image ancestry (`curl ca-certificates gnupg git nodejs` + two npm CLIs); `main-cd.yml:526` proves it is the base of `memex-portal-ai`. `Probe()` returns `null` in prod, so pixel is never even offered. This is stated design, not oversight (`PixelRenderingOptions.cs:8-14`).
2. **There is no HTML→DOCX converter** anywhere in the repo or its packages — DOCX comes only from the `Document` model via OpenXml, and Chromium prints only PDF.
3. A general document composer would need new CSS with real pagination (`SlidePrint.css` is a fixed 16:9 box with `overflow: hidden` — it *clips*) plus `@page` margin boxes for running headers, since the Chromium CLI has no header/footer template.

So the fix lands on **both** paths from one shared resolver, which is what the reported bug actually requires. Making the browser route primary needs a base-image decision first; recorded in the README and `PixelFaithfulExport.md`.

## Known limitation (documented, pre-existing)

The content-fidelity renderers draw **no images at all** — `PdfDocumentRenderer.cs:349` and `DocxDocumentRenderer.cs:441` both emit bracketed alt text and never fetch bytes. So an area's *pictures* still do not appear in PDF/DOCX; its text, structure and links do, absolutised against the portal base URL so they work after the file is mailed on. Not introduced here; now written down.

## 🚨 Dependency on #1222

This branch is **based on #1222** (per instruction, to avoid forking the area-resolution logic), so the diff currently includes that PR's commits. **#1222 should merge first.** I will not merge this until #1222 lands or its author agrees — flagging rather than landing someone else's feature via my PR.

---

## 🔐 Security: an information-disclosure bug fixed in review

Resolving embedded areas server-side means the **export now reads content the reader never explicitly asked for**, which makes the identity it reads under a security boundary rather than an implementation detail. Review found that boundary broken:

`LayoutAreaResolver` captured the caller's `AccessContext` and applied it with `Observable.Defer` + `using var`. The scope therefore closed **when the factory returned** — so the area stream was *created* under the caller's identity but *subscribed and rendered* without it, and the owner's read gate evaluates on the **subscription**. This is the same family as the `/api/content` hole fixed today in #1224: **a permission check that appears to run but is not the one that decides.**

Fixed with `Observable.Using`, which binds the scope to the subscription's lifetime.

**Regression test** (`DocumentExportAreaAccessTest`, real row-level-security mesh, no mocking): on one document embedding a restricted node, `carol` (granted) gets the card in her PDF and `bob` (no grant) does not, while the rest of the document he may read still renders. A/B on the same document, because "the secret is absent" passes trivially if area resolution is broken outright.

**What that test does not prove — stated rather than implied.** It pins the user-visible outcome, and that outcome holds because each export runs in its own script hub where the ambient identity *is* the requester. It does **not** discriminate `Defer` from `Using`: with caller and ambient identical, both shapes pass. Attempting to discriminate them — resolving with an explicit caller differing from the ambient identity — produced **order-dependent** results on a shared workspace: the same `(ambient, caller)` pair returned the full card or nothing depending only on which identity read that area **first**. That points at the layout-area result being cached per `(address, reference)` **without identity**. That is a real question and a plausible disclosure vector in any path where two users' reads share a workspace, but it is not this PR's, and it deserves its own investigation rather than an assertion bolted on here. **The scope change is therefore carried as defence-in-depth — it can only ever hold the caller's identity longer than before — not as a mechanism this PR proves operative.** Flagging it explicitly so it is not mistaken for a closed question.


rbuergi:
Rebased on main — **#1222 and #1225 are now merged**, so the dependency note in the description is resolved and this PR's diff is now exactly its own work (28 files). Conflict resolution took main's deletion of `BuiltInSkillCatalog.cs` and its file-derived skill expectations from #1225; nothing resurrected.

**Visual verification added.** I generated real PDFs and looked at them rather than trusting the assertions, and it caught a defect the tests could not see: the resolved card grid was structurally perfect and **visually flat** — title, description and link all rendered as identical body text, so a card did not read as a card.

Cause: `AreaMarkupRenderer` carries emphasis as **inline CSS** (`MarkupStyles.CardTitle` = `font-weight:700`), never as `<strong>`, because inline CSS is the only styling a mail client obeys. `MarkupToDocument` read emphasis from tags only — and computed it *after* the `<a>` branch had already returned, so a styled link (which is exactly what a card title is) could never keep it. Fixed in 6fb9e4a, pinned by two new tests in `MarkupToDocumentTests`.

Note the `@meshweaver/react` check is red repo-wide right now (`@fluentui/react-icons` resolution), unrelated to this PR and non-required.

rbuergi:
Addressed all five Copilot findings — four were real, and two of them were genuine bugs in my own new code. Thanks, these were good catches.

1. **`LayoutAreaResolver` — caller AccessContext disposed before subscription.** Correct and the most serious one. `Observable.Defer` + `using var` closed the scope when the *factory* returned, so the area stream was **created** under the caller's identity but **subscribed and rendered** without it — and the owner's read gate is evaluated on the subscription. Exactly the shape `AccessContextPropagation.md` warns about; it also meant the unavailable-area notice could resolve in the wrong locale. Switched to `Observable.Using`, which ties the scope to the subscription's lifetime, with stream disposal still on completion.

2. **`TableSizer` — `role="presentation"` on content tables.** Correct. Those are the author's own data tables; marking them presentational strips their semantics from a screen reader and buys nothing, since the sizing is what Word needs and Word does not consult `role`. Removed, with a comment explaining that only `AreaMarkupRenderer`'s *layout* tables (the card-grid scaffolding) legitimately carry it.

3. **`MarkupToDocumentTests` — test contradicted its own name.** Correct, and it was worse than cosmetic: my helper didn't reproduce the renderer's padding at all, so it asserted `HaveCount(1)` against an input the renderer never produces. The helper now pads like `CardTable` does, and the test asserts every row keeps the same column count — which is what actually matters, since QuestPDF and Word derive the column layout from the row's cell count and a short final row makes the last card stretch full width.

4. **`DocumentAreaResolution` — one pipeline for all chapters.** Correct on both counts. With `IncludeChildren`, descendant chapters are different nodes, so a relative `@@("area:Foo")` was resolved against the *root* address, and two descendants writing the same relative embed collided on the raw-path key and would have rendered each other's content. Introduced `ExportChapter(Title, Markdown, NodePath)`; resolution and `DocumentBuilder` now use a per-path pipeline (cached, instance-local) and the node path is part of the cache key. Templates pass each node's own `Path`.

5. **`SendDocumentLayoutArea` — hard-coded English subject.** Correct, and `ui.sendDocument.defaultSubject` already existed in both catalogs. Now localized.

Verified: `-c Release -warnaserror` clean, 71 unit + 8 export integration + 3 relay/send tests green. The integration tests matter here because the `.csx` templates compile at **runtime** — `dotnet build` cannot type-check the `ExportChapter` change, only a test that executes them can.

rbuergi:
Synced main (now includes #1243, #1239, #1233, #1236) and re-triaged both red tests on head `7cbe0ff`.

**`CodeCellOutputCaptureTest.Progress_Of_Succeeded_Activity_Renders_Output_For_PublicRead_Viewer` — stale branch, not this diff.** All 5 tests in that class pass on the synced tree. The branch predated #1243, so it was running that test without the companion setup changes #1243 made when it deleted `AccessService.persistentCircuitContext`. Merging main is the entire fix — no test setup of mine needed changing, and no permissive behaviour was restored.

Verified the reverse direction too, since `DocumentExportAreaAccessTest` uses `SetCircuitContext` and was squarely in #1243's blast radius: **all 9 export integration tests still pass** on the synced tree.

**`SearchResultStormTest.CatalogSubscription_ContentUpdateStorm_AddsNoExtraRerenders`** — known main flake, #1193 targets it. Not chased.

**The open question I flagged in the security section above is now closed upstream — by #1239, not by me.** I reported that resolving an area with an explicit caller differing from the ambient identity gave *order-dependent* results on a shared workspace, and that this pointed at the layout-area result being cached per `(address, reference)` without identity. #1239 `f86034d` adds identity to the `Workspace._remoteStreamCache` key, which is exactly that. So the caching vector is fixed; the `Observable.Using` change in this PR remains correct and is still carried as defence-in-depth rather than as the operative gate, which is how the description states it.

PR #1238, opened 2026-08-11
Batch bake is the default again for a gated bake — verified on all three portals

rbuergi:
Restores the default that #1217 walked back, now that the defects behind that walk-back are fixed and the path is proven in production.

## Why it was turned off, and what changed

Its first production run (memex-cloud, `ci.2947`) finished in 19s with `compiled=4 compileErrors=59 contentBroken=169` of 237 — batched discovery was resolving almost nothing. Two real defects, both now fixed and pinned:

- **#1216 / #1220** — the global fetch was truncated by a default query limit, *and* `nodeType:Code` could not see Postgres-stored Code nodes at all (Code lives in the per-schema `code` satellite table, routed by path segment, and is deliberately not nodeType-resolvable).
- **#1232** — `scope:subtree` was silently **dropped** for a WILDCARD namespace, so `namespace:*/Source scope:subtree` matched only namespaces *ending* in `/Source` and missed nested sources such as `Claims/SampleData/Source/Fixtures/…`.

Discovery also no longer guesses: a type that declares sources but resolves none fails the **whole batch** to the activation path instead of emitting the non-gating `NoSources` verdict — so a discovery bug can no longer disguise itself as content breakage.

## Production evidence (2026-08-11)

| portal | image | sweep |
|---|---|---|
| memex-cloud | `ci.2982` | **237/237 in 62s** (same fleet: 18m55s on the activation path) |
| memex | `ci.2990` | 83 already baked + 1 pending in **6.9s** |
| atioz | `ci.2990` | **39/39 in 38s**; follower pod **1.1s** off the shared store |

All three with `compileErrors=0 timedOut=0 contentBroken=0`, gate armed throughout.

## Why the default and not just the env var

The default is what every deployment gets. Carrying batch bake as three live `kubectl set env` patches meant a helm upgrade, a new namespace, or a fresh deployment silently fell back to the 18-minute sweep — the failure being invisible is exactly what makes it worth fixing at the default. `PreWarm:BatchBake` goes back to being the **escape hatch** (`false` forces the activation sweep on a gated pod), not the enabler.

No change is needed in the tracked helm values: they already carry `PreWarm__GateReadiness`/`PreWarm__DynamicTypes`, and batch bake follows the gate. Adding an explicit key there would be redundant and would pin the mode against any future default change.

Tests: 35/35 across the pre-warmer, bake-gate and batch-bake suites; Release `-warnaserror` clean. What's New entry included (`Category: Feature`).


rbuergi:
CI round 2 is red on `NodeTypeModelProbeTest.ProbeInstanceModel_SnapshotsContentTypeAndSchema` (shard 1) and `SearchResultStormTest.CatalogSubscription_ContentUpdateStorm_AddsNoExtraRerenders` (shard 2) — neither of which this PR can affect.

Round 1 of this same branch, carrying the entire code change, was **GREEN**. The only delta since is one word in the What's New entry ("nearly nineteen" → "nearly nineteen minutes", the Copilot finding above). A doc-only edit turning a green run red on two unrelated tests in two different shards is not this PR.

The shard message is `a test host exited non-zero (failure, crash, or timeout kill)` — a killed host, not an assertion, so the tests named are whichever were executing when it died. Filed with the full four-run evidence table as #1242.

Not re-running (the repo's no-rerun-to-green rule). The substantive verification for this PR stands: round 1 green, plus 35/35 locally across the pre-warmer / bake-gate / batch-bake suites and a Release `-warnaserror` clean build.

PR #1241, opened 2026-08-11
fix(#1214): retract a bake regression when the type rebuilds on the same image

### LINKED ISSUE #1214: The NodeType bake can compile a half-applied plugin update and report it as an image regression
## What happened (memex-cloud, 2026-08-11 ~11:04–11:12 UTC)

A gated pod's bake reported:

```
REFUSING READINESS — 4 NodeType(s) regressed on this image: Store/Catalog, Store/Coupon, Store/Enrollment, Store/Order
CS0117: 'Localizer' does not contain a definition for 'InstalledLocationOf'
CS0117: 'Localizer' does not contain a definition for 'ReadInstallRecord'
```

Nothing was wrong with the image. The plugins repo had **reverted** `fix/install-record-read` (which had added those members), and the portal's plugin auto-update was applying that revert **while the bake was compiling**:

| Node | last_modified |
|---|---|
| `Store/Plugin/Source/Localizer` (v22, post-revert — members gone) | 11:04:59 |
| `Store/Plugin/Source/PluginLayoutAreas`, `InstallLayoutAreas` | 11:06:57 |

The bake ran 10:47→11:06 and compiled `Store/*` against a snapshot where `Localizer` had already been reverted but its callers had not. Verified after the fact: **no** Store source references those members any more (0 rows), i.e. the content converged at 11:06:57 and is self-consistent — the compile just sampled the middle of the update.

## Why this matters

The bake gate exists to stop a bad IMAGE from rolling. Here it stalled the rollout on a transient CONTENT state, and the failure is terminal in two ways: the types are PARKED after a compile failure, and `BakePhase.Regressed` is sticky (completion is not absolution — by design). So the pod never becomes Ready and the rollout hangs until a human intervenes, even though the content is fine seconds later.

This is the same *class* as #1209 and the KmuBasics content-vs-image confusion (#1204): a content-state problem wearing an image-regression costume — but with a new trigger, an in-flight package install, which will recur on any portal whose `PluginCatalog:AutoUpdateByDefault` is on (the default).

## Proposals

1. **Sequence the bake behind the install** — the `PreWarmCompletion` barrier (#1114) already exists to sequence boot flows against the bake; the inverse ordering (bake waits for an in-flight default-install/auto-update to settle) is the natural cure and removes the race rather than detecting it.
2. **Re-verify before gating**: before a `CompileError` on a previously-healthy type is promoted to a regression, re-read the type's source versions; if `CurrentSourceVersions` changed since the compile started, the snapshot was torn — recompile once instead of stalling the rollout.
3. At minimum, name the suspicion in the log when a compile fails while a package install is in flight, so an operator is not told "this image is broken" about content that repaired itself.

Recovering an already-stalled pod is just a restart once the content has converged — but that requires someone to notice.

rbuergi:
Fixes #1214.

## The incident (memex-cloud, 2026-08-11 ~11:04–11:12 UTC)

A gated pod refused readiness with:

```
REFUSING READINESS — 4 NodeType(s) regressed on this image: Store/Catalog, Store/Coupon, Store/Enrollment, Store/Order
CS0117: 'Localizer' does not contain a definition for 'InstalledLocationOf'
CS0117: 'Localizer' does not contain a definition for 'ReadInstallRecord'
```

Nothing was wrong with the image. The plugins repo had reverted `fix/install-record-read`, and the portal's plugin auto-update was applying that revert **while the bake compiled**:

| node | last_modified |
|---|---|
| `Store/Plugin/Source/Localizer` (post-revert, members gone) | 11:04:59 |
| `Store/Plugin/Source/PluginLayoutAreas`, `InstallLayoutAreas` | 11:06:57 |

The bake ran 10:47→11:06 and compiled inside that two-minute gap, so Roslyn's verdict described a source set that existed briefly and never again — verified afterwards, no Store source references those members.

## Root cause

**The platform already treats such a failure as provisional.** `NodeTypeCompileParkRegistry.ShouldRetryForSourceChange` un-parks and recompiles a failed type the moment its source snapshot changes, so those four types went green seconds after the content converged at 11:06:57.

The only thing that stayed broken was the **gate's record** — latched, sticky, and never re-read. So the pod never became Ready and the rollout hung until a human noticed. This recurs on any portal with `PluginCatalog:AutoUpdateByDefault` on (the default), and on any concurrent content write — a git sync or a bulk edit, not just plugin installs.

## The fix — the latch becomes level-triggered on the truth it claims to represent

- **`NodeTypeBakeGateState.RetractRegression(path, reason)`** — a regression is retracted when the type it names is afterwards observed reaching a usable build **on this image**. This is *not* "completion is absolution" (`MarkComplete` still refuses that); it is a **later, better measurement of the same type on the same image superseding an earlier one**. `MarkComplete` now remembers its message so a retraction arriving *after* the sweep lands on the Complete verdict instead of leaving the pod stuck mid-sweep. Retractions stay visible in the health payload in **both** the red and the green wording.
- **`DynamicTypePreWarmer.WatchForRecovery`** — `MarkOutcome` now reports whether it recorded a regression, and the pre-warmer starts one watch per condemned type. It **observes only**: nothing is triggered, retried, polled or timed out. Holding the subscription also keeps the condemned type's hub activated, which is what keeps the sources watcher that notices the repair alive.
- **The recovery must be a compile that is demonstrably FRESH** — settled `Ok` with `LastCompileSucceededAt` strictly newer than the baseline the watch observed. `HasUsableBuild` alone is **not** sufficient and matching on it silently disables the whole gate: `ApplyCompileFailure` keeps the previous build's assembly coordinates and framework stamp, so any type that ever compiled on this image would satisfy it the instant its regression was recorded. The first draft did exactly that, the integration test caught it, and the test now pins it.
- **Proposal 3** — `PreWarmOutcome.SourcesMovedDuringCompile` carries the torn-snapshot evidence (a source's `LastModified` at or after `LastCompileStartedAt`) and the log names the suspicion instead of telling an operator the image is broken. Diagnostic only; it never downgrades a verdict.

### Altitude: why not proposal 1, and why the evidence alone is not enough

**Sequencing the bake behind the install would deadlock.** `InstanceAutoRegistrationService.cs:216` already waits on `PreWarmCompletion.Completed` before installing (that is what #1114 added), so the inverse ordering makes each wait on the other.

**The timestamp evidence cannot carry the fix by itself.** It is one-sided: in this incident the callers landed at 11:06:57, *after* the failures, so at the moment each verdict was formed no source had yet moved past that type's compile-start stamp. A tear is often only demonstrable in hindsight — which is exactly why the cure is the retraction (level-triggered, no deadline) and the evidence is used for the log line.

## Verification

**The torn-snapshot test fails against the current behaviour.** With `RetractRegression` neutered to today's "a regression is immortal" semantics:

```
BEFORE: === TEST FAILED: The operation has timed out.
        Failed: 1, Passed: 0 — Duration 2 m 32 s
AFTER:  Passed! - Failed: 0, Passed: 35, Skipped: 0, Total: 35 — Duration 3 s
```

The retraction lands ~200 ms after the content converges. The test's log shows the mechanism end to end:

```
[22:44:51] RETRACTING the regression recorded for PreWarmTestTorn/TornType — it has since
reached a usable build on THIS image … Gate now: 1 NodeType(s) regressed on this image:
PreWarmTestTorn/StaysBroken (1 further regression(s) retracted after the type rebuilt on
this image — PreWarmTestTorn/TornType)
```

`RegressedType_ThatRebuildsAfterTheContentConverges_IsRetracted_ButARealBreakStillGates` reproduces the incident's shape in one mesh — a helper losing a member its caller still calls, then the caller converging — and pins **both** halves: the torn type is retracted, and a genuinely broken type on a source set nobody touched **still gates**. The negative half has a real synchronisation point (the positive retraction) rather than a sleep.

Full suites:

```
test/MeshWeaver.Hosting.Monolith.Test   Passed: 563, Failed: 0, Skipped: 4  (5 m 53 s)
test/MeshWeaver.Graph.Test              Passed: 978, Failed: 0, Skipped: 0  (31 s)
```

Release `-warnaserror` clean, one project per invocation: `MeshWeaver.Hosting`, `MeshWeaver.Hosting.Monolith.Test`, `MeshWeaver.Graph.Test`, `Memex.Portal.Distributed` (the gate's consumer — `MarkOutcome`'s `void`→`bool` is source-compatible).

#1204's `NoSources` and #1226's unestablished-source-set buckets are untouched — `SourceSnapshotEstablishmentTest.TheThreeCases_GateDifferently` still passes unchanged.

### DIFF
--- a/src/MeshWeaver.Documentation/Data/WhatsNew/2026-08-11-updates-no-longer-stall-on-a-type-that-already-repaired-itself.md
+++ b/src/MeshWeaver.Documentation/Data/WhatsNew/2026-08-11-updates-no-longer-stall-on-a-type-that-already-repaired-itself.md
@@ -1,0 +1,35 @@
+---
+Name: Platform updates no longer stall on a type that has already repaired itself
+Category: Fix
+Description: If a custom type fails to build because a package update was landing at that exact moment, the update now finishes as soon as the type builds again — instead of stopping until someone noticed.
+Icon: Sparkle
+Order: -20260811
+---
+
+# Platform updates no longer stall on a type that has already repaired itself
+
+While a platform update is being prepared, every custom type is rebuilt, and a type that used to
+work but no longer does will hold the update back. That safety-check is deliberate: it is what
+keeps a bad version from reaching you, and it has caught real breakage.
+
+But a type can also fail to build for a reason that has nothing to do with the new version. A
+plugin update rewrites several files, and it does not rewrite them all in the same instant. If the
+rebuild happens to run between two of those writes, it sees a half-changed set of files — one file
+already updated, the files that use it not yet — and reports errors about code that, moments later,
+no longer exists anywhere. It reads exactly like real damage: an error message, a line number, a
+type that stopped working.
+
+Until now that verdict was final. Seconds later the remaining files landed, the platform rebuilt
+the affected types by itself, and everything was healthy again — but the update stayed stopped,
+because nothing ever went back to look. It took someone noticing and restarting to get moving
+again.
+
+Now the platform keeps watching every type it held the update for. The moment such a type builds
+successfully on the new version, its objection is withdrawn and the update carries on by itself.
+Nothing is retried or timed out to make this happen — the type building is simply better evidence
+than the earlier failure, and it wins.
+
+A genuine problem is unaffected. A type that is really broken never builds, so nothing withdraws
+its objection and the update stays held exactly as before. And when this does happen, the platform
+now says so in the log while it waits: if a type's source files changed while it was being built,
+the log names that as the likely cause instead of blaming the new version.
--- a/src/MeshWeaver.Hosting/DynamicTypePreWarmer.cs
+++ b/src/MeshWeaver.Hosting/DynamicTypePreWarmer.cs
@@ -112,6 +112,21 @@ public record PreWarmOutcome(string TypePath, PreWarmStatus Status, string? Deta
     /// would let one abandoned NodeType block every future deploy.</para>
     /// </summary>
     public bool WasHealthyBeforeBake { get; init; } = true;
+
+    /// <summary>
+    /// Evidence that this compile sampled a TORN source snapshot: at least one of the type's
+    /// sources was last modified AT OR AFTER the compile started
+    /// (<see cref="DynamicTypePreWarmer.SourcesMovedDuringCompile(NodeTypeDefinition)"/>), so the
+    /// diagnostics describe a source set the mesh was in the middle of replacing.
+    ///
+    /// <para>Diagnostic only — it deliberately does NOT reclassify the outcome. A source moving
+    /// mid-compile makes a failure SUSPECT, not innocent, and a torn compile of genuinely broken
+    /// code must still gate. What clears the gate is the type actually rebuilding on this image
+    /// (<c>NodeTypeBakeGateState.RetractRegression</c>); this flag is what lets the log say WHY the
+    /// pod is refusing readiness on content that may repair itself, instead of telling an operator
+    /// at 3am that the image is broken (issue #1214, proposal 3).</para>
+    /// </summary>
+    public bool SourcesMovedDuringCompile { get; init; }
 }
 
 /// <summary>
@@ -678,8 +693,7 @@ public static class DynamicTypePreWarmer
                                     var d = (NodeTypeDefinition)n!.Content!;
                                     return d.CompilationStatus switch
                                     {
-                                        CompilationStatus.Error => new PreWarmOutcome(
-                                            typePath, ClassifyCompileFailure(d), d.CompilationError),
+                                        CompilationStatus.Error => FromFailedCompile(typePath, d),
                                         // The rebuild never reported an answer — not a
                                         // compile failure, so never labelled one.
                                         CompilationStatus.Unavailable => new PreWarmOutcome(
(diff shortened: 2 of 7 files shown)


rbuergi:
Addressed in `c59fb898` — thank you, this was a real defect.

You were right on both counts, and the second one changed the design. Watching an `UpstreamFailed` dependent doesn't just cost activations — it is the **wrong mechanism**. A derived regression's entire evidence is *"my upstream failed"*; the sweep never compiled that type, deliberately. So the thing to do when the blocker's verdict is withdrawn is to withdraw the derived one **with** it, not to go and re-measure a type we chose not to measure. That is the same principle the `Upstream*` statuses already encode: a derived verdict is only ever as strong as the one it derives from.

So, your option (b):

- `PreWarmOutcome.BlockedBy` carries the blocker as **data** (the sweep already names it in the detail text; parsing it back out would break on the next reword).
- `MarkOutcome` now returns `true` only for a **directly-measured** regression, so only those start a watch. A derived one is still recorded and still gates — it just isn't watched.
- `RetractRegression` cascades **transitively** to every regression blocked by the type being retracted, keyed on the actual blocker so unrelated regressions keep standing.

Net effect: one broken upstream now activates **zero** hubs, and its whole fan-out clears the instant the upstream is retracted. Pinned by `RetractingABlocker_AlsoRetractsTheRegressionsDerivedFromIt` (including the transitive hop) and `RetractingABlocker_LeavesUnrelatedRegressionsStanding`.

PR #1257, opened 2026-08-12
Drop the dead JsonPath.Net reference; correct the JsonPatch.Net rationale

### LINKED ISSUE #1231: Decision needed: json-everything (JsonPatch.Net/JsonPath.Net) carries an Open Source Maintenance Fee
Surfaced by the dependency licence audit (2026-08-11). **This one needs a product decision, not a mechanical removal.**

Four packages in the shipping graph carry `OSMFEULA.txt` — the **Open Source Maintenance Fee** EULA from the json-everything project:

| Package | Version | Scope |
|---|---|---|
| `JsonPatch.Net` | 5.0.3 | direct |
| `JsonPath.Net` | 3.0.2 | direct |
| `Json.More.Net` | 3.0.1 | transitive |
| `JsonPointer.Net` | 7.0.1 | transitive |

## What the EULA actually says
It is **not** the QuestPDF shape, and the difference matters:
- The **source is MIT** (an OSI-approved licence) and remains freely distributable — §3.
- The fee attaches to the **Binary Release** (the published NuGet), for users who use it in revenue-generating activities **with annual gross revenue ≥ US$10,000** — §1.
- §3 states explicitly: *"The Fee is not a license fee."*
- §4: *"To the extent any term of this Agreement conflicts with User's rights under the OSI License, the OSI License shall govern"*, and a user may self-compile binaries from source without the Agreement.

So we are **not** in breach of a licence by consuming it, and the code is MIT. But the published binary we restore does come with a maintenance-fee expectation for commercial users — the same *economic* shape the user objected to, via a different legal mechanism.

## Why it is not a quick swap
`JsonPatch.Net` is referenced by `MeshWeaver.Data`, `MeshWeaver.Data.Contract` and `MeshWeaver.Messaging.Hub` — it implements the **RFC 7396 JSON-merge-patch** path that every cross-hub `GetMeshNodeStream(path).Update(...)` write goes through. Replacing it touches the core mutation API, so it must be a deliberate, separately-tested change.

## Options
1. **Pay the maintenance fee** — cheapest in engineering time; contradicts the "no commercial dependencies" rule.
2. **Replace** with a permissively-licensed JSON Patch/Pointer implementation (or implement RFC 7396 merge-patch in-house — it is a small, well-specified algorithm, and merge-patch is the only part the hub actually needs).
3. **Accept and document** — rely on §3/§4 (MIT source governs) and record the position.

Until this is decided the four packages sit in `EXCEPTIONS` in `.github/scripts/check-licenses.py` with this issue as the reason, so the licence gate stays green for everything else while this stays visible.

rbuergi:
Closes #1231.

## Decision recorded: accept and document

`JsonPatch.Net` (+ transitive `Json.More.Net`, `JsonPointer.Net`) carries an Open Source Maintenance Fee. The source is MIT, and the OSMF EULA itself states the fee is **not** a licence fee and that the OSI licence governs on conflict — so this is a fee *expectation on the published binary*, not a licence breach. We accept it and keep it documented as dated debt in the licence gate.

## Two concrete changes

**1. `JsonPath.Net` was dead weight — removed.** Referenced by `MeshWeaver.Data` but never used: no `using Json.Path`, no symbol from the package anywhere in the tree. (The only `JsonPath` grep hits are an unrelated `existingJsonPath` local in `FileSystemStorageAdapter`.) That is one OSMF package out of the shipping set for free.

**2. The gate's rationale for `JsonPatch.Net` was factually wrong — corrected.** It read:

> Core dependency: the RFC 7396 merge-patch path behind every cross-hub stream.Update.

It is not on that path. **RFC 7396 is hand-written in-tree** on `System.Text.Json.Nodes` — `DataExtensions.MergePatchRecursive` and `Serialization/MeshNodePatchMerge.cs` — and neither imports anything from the json-everything family. What the package actually provides is **RFC 6902** (patch) + **RFC 6901** (pointer), concentrated in `Serialization/JsonSynchronizationStream.cs` and `StandardReducers.cs`.

This mattered beyond tidiness: the misstatement made the dependency look load-bearing for the core mutation API and priced any replacement as core-risk. It is not — should we ever revisit, the remaining work is a pointer type plus `CreatePatch` (diff, 3 call sites), because the RFC 6902 **apply** is *already* hand-rolled in-tree (the library's escaping is wrong — see the comment at `JsonSynchronizationStream.cs:1147-1149`), and only `move`/`copy`/`test` still call the library, which MeshWeaver's own diff never emits.

## Verification

```
$ python3 .github/scripts/check-licenses.py
Licence gate passed: 391 shipping packages checked (40 declared-unused ignored).
9 documented exception(s) still carried   # was 10 — jsonpath.net is gone
```
`dotnet build src/MeshWeaver.Data -c Release -warnaserror --no-incremental` → `0 Warning(s), 0 Error(s)`, DLL emitted (checked non-incrementally so this is a real compile, not an up-to-date no-op).

No What's New entry: nothing here changes observable behaviour.


rbuergi:
Superseded by PR #1292, which replaced the whole json-everything family with our own implementation.

Everything this PR did is already on main, verified rather than assumed:

```
$ git grep -c 'JsonPath.Net'  origin/main -- Directory.Packages.props   -> 0
$ git grep -c 'JsonPatch.Net' origin/main -- Directory.Packages.props   -> 0
$ git show origin/main:.github/scripts/check-licenses.py | grep -cE '"json(path|patch|pointer)|json.more'  -> 0
```

So the dead `JsonPath.Net` reference is gone, and so are all four licence-gate exceptions — #1292 went further and removed `JsonPatch.Net`, `Json.More.Net` and `JsonPointer.Net` too.

The one part of this PR that was **not** purely deletions — correcting the licence gate's false claim that `JsonPatch.Net` sits on "the RFC 7396 merge-patch path behind every cross-hub `stream.Update`" — is moot now that the exception text it lived in no longer exists. For the record, since that sentence caused a peer session to size the replacement as core-risk: RFC 7396 was always hand-written in-tree on `System.Text.Json.Nodes` (`DataExtensions.MergePatchRecursive`, `Serialization/MeshNodePatchMerge.cs`), and what the package actually supplied was RFC 6902 + RFC 6901.

Closing unmerged. Its CI red was inherited (`ExportTemplateCompilationTest`, unreachable from a package-reference deletion), so nothing here needs chasing.

PR #1168, opened 2026-08-10
A parked NodeType no longer makes every instance of it un-writable

rbuergi:
### These four issues are THREE defects, not one. The evidence

I went into this expecting the four "startup readiness timeout" issues to share a cause. They do not, and the timestamps say so plainly. **All four are startup-adjacent but only two are the same defect** — and the sibling shutdown lead (#1107) does not apply to any of them:

| Issue | Pod lifecycle at the moment of failure | Defect |
|---|---|---|
| #1114 | boot (memex, ci.2604) | **A** — parked NodeType blocks every instance write |
| #1133 + #1137 | boot (memex + memex-cloud, three pods) | **B** — boot-time HTTP to a peer portal that is rolling in the same second |
| #1122 | neither — a transient probe hub's own create/dispose, mid-pod-life | **C** — post-mortem watchdog on a disposed hub (PR #1160, verified) |

### Defect A — what #1114 actually is (this PR)

Measured on `memex-portal-deployment-575dc48bfb-wq7s8` (Loki, namespace `memex`):

```
07:09:08  CS0117: 'AiSettingsNodeType' does not contain a definition for 'AddSkillSource'
07:09:09  NodeType 'Store/Plugin' PARKED after compile failure
07:10:17  EnrichWithNodeType slow path for 'Store/Plugin' faulted — overlay for 'Agent'
          TimeoutException: NodeType 'Store/Plugin' recompile did not settle and no
                            compile is in flight (within 60s)
07:10:45  [DefaultInstall] installing package Agent failed
          … Update aborted: no initial state arrived for 'Agent' within 30s
07:12:06  DynamicTypePreWarmer: warm-up complete (compiled=74, compileErrors=2)
07:13:04  … overlay for 'DataModelling'   ← AFTER the bake
07:14:14  … overlay for 'Edu'             ← AFTER the bake
07:15:24  … overlay for 'Essentials'      ← AFTER the bake
07:16:35  … overlay for 'Publish'         ← AFTER the bake
```

The chain:

1. A node-repo `.cs` inside the `Store/Plugin` NodeType calls `AiSettingsNodeType.AddSkillSource`, a framework method that had been deleted. CI cannot see in-mesh source, so it detonated on deploy — the exact class AGENTS.md warns about (#683).
2. The type parks. Parking is correct and deliberate: it stops Roslyn being re-dispatched for a type that would fail identically.
[...]

PR #1266, opened 2026-08-12
fix: storm breaker key carries payload identity, so a wide fan-out is not dropped as a loop

### LINKED ISSUE #1200: Import activity triggers unbounded message storm to cache hub via RawJson/DataChangedEvent loop
## What is failing

Import activities (e.g. `UWDeepfield/_Activity/import-*`) flood the cache hub with >2000 `RawJson` messages in under 1 second, triggering the message-storm guard which then drops the key at ingestion for 2s. The same burst also sends `DataChangedEvent` messages. The storm guard message explicitly warns of an "UNBOUNDED retry/resubscribe/repost loop."

## Probable cause

**Confidence: high** — the evidence ties directly to a known defect.

The sender is `UWDeepfield/_Activity/import-*`, and the target is `cache/gmIIqhF220CKKxqxgY_J7Q`. The same hub ID (`gmIIqhF220CKKxqxgY_J7Q`) appears in the concurrently-firing ROUTER_TRAFFIC incident (#1140) where the mesh router hub is being used as the sender address by portal hubs instead of their own hub identity. The import activity's hub is likely misaddressing messages with the router hub's identity; the router then re-processes or re-broadcasts them, creating a positive feedback loop that generates the storm. This is the exact pattern described in the log: "UNBOUNDED retry/resubscribe/repost loop."

## Impact

- **Low immediate operational impact**: 3 occurrences across ~31 seconds on a single pod (memex-portal-deployment-59479dc7d9-kmwvr). The storm guard mitigated it by dropping the offending key at ingestion for 2s. No crash, no cascade.
- **Medium risk going forward**: This will recur each time the same import activity runs, because the underlying hub-identity miswiring (issue #1140) is still present. If the import processes more data or multiple imports run concurrently, the burst could intensify.
- **This is a downstream symptom of #1140**, not an independent defect.

## Where to look

- `MessageService` in `MeshWeaver.Messaging` — the storm guard threshold (2000 messages / 1s) and key-drop logic are here; confirm the guard fired correctly (it did).
- Trace back to the import hub at `UWDeepfield/_Activity/import-*` — what message handler is producing the `RawJson` bursts? Likely an event handler that receives a `DataChangedEvent` and replies with `RawJson`, creating a ping-pong cycle.
- The hub identity miswiring is tracked in **issue #1140** (`ROUTER_TRAFFIC`). The storm will stop reproducing once the hub identity is fixed.
- Import hub lifecycle: check whether the import activity's hub correctly resolves its own address and whether the cache hub subscription setup could cause re-subscription loops.

---

### Evidence

| | |
|---|---|
| Fingerprint | `7a92a4f2b427a9ef` |
| Category | `MeshWeaver.Messaging.MessageService` |
| Severity | Error |
| Namespace | `memex-cloud` |
| Pods | `memex-portal-deployment-59479dc7d9-kmwvr` |
| Occurrences | 3 |
| First seen | 2026-08-11 07:28:14Z |
| Last seen | 2026-08-11 07:28:45Z |

```
2026-08-11 07:28:14Z memex-portal-deployment-59479dc7d9-kmwvr fail: MeshWeaver.Messaging.MessageService[0]
      MESSAGE STORM detected in hub cache/gmIIqhF220CKKxqxgY_J7Q: the tuple (sender=UWDeepfield/_Activity/import-f2bb979af363573d-20260811072327418-f4dc71b1, target=cache/gmIIqhF220CKKxqxgY_J7Q, type=RawJson) exceeded 2000 messages within 1s (observed 2001 in-window). Dropping this key at ingestion for 2s to keep the hub responsive — this indicates an UNBOUNDED retry/resubscribe/repost loop somewhere; find and fix it.
2026-08-11 07:28:45Z memex-portal-deployment-59479dc7d9-kmwvr fail: MeshWeaver.Messaging.MessageService[0]
      MESSAGE STORM detected in hub mesh/gmIIqhF220CKKxqxgY_J7Q: the tuple (sender=UWDeepfield/_Activity/import-f2bb979af363573d-20260811072327418-f4dc71b1, target=cache/gmIIqhF220CKKxqxgY_J7Q, type=DataChangedEvent) exceeded 2000 messages within 1s (observed 2001 in-window). Dropping this key at ingestion for 2s to keep the hub responsive — this indicates an UNBOUNDED retry/resubscribe/repost loop somewhere; find and fix it.
2026-08-11 07:28:45Z memex-portal-deployment-59479dc7d9-kmwvr fail: MeshWeaver.Messaging.MessageService[0]
      MESSAGE STORM detected in hub cache/gmIIqhF220CKKxqxgY_J7Q: the tuple (sender=UWDeepfield/_Activity/import-f2bb979af363573d-20260811072327418-f4dc71b1, target=cache/gmIIqhF220CKKxqxgY_J7Q, type=RawJson) exceeded 2000 messages within 1s (observed 2001 in-window). Dropping this key at ingestion for 2s to keep the hub responsive — this indicates an UNBOUNDED retry/resubscribe/repost loop somewhere; find and fix it.
```

</details>

<sub>Opened automatically from `Admin/_LogIncident/7a92a4f2b427a9ef`. Recurrences are folded into this issue rather than opening new ones.</sub>

rbuergi:
Closes #1200

## The defect

`MessageStormBreaker` keys its per-second rate counters on `(Sender, Target, TypeName)` — **no
component identifying what the message is about**. Its own justifying comment claims real traffic
is "DIVERSE tuples". That is false wherever the mesh funnels traffic through a dispatcher: **one
sender talks to one target with one message type about MANY DIFFERENT THINGS.**

- Every sync stream an owner hub holds open to the shared mesh-node cache posts
  `DataChangedEvent` (→ `RawJson` once the router packages it) over that single tuple.
- A bulk importer sends one `CreateOrUpdateNodeRequest` per source node from one import hub to the
  one mesh hub.

Keyed on the routing tuple alone, a wide **legitimate** fan-out is arithmetically identical to one
thing looping. The guard therefore crosses its bar on ordinary bulk work and `ShouldDrop`
**discards real writes at ingestion** — imported content and progress-log lines silently missing.

**That is data loss, not mitigation.** The issue text credits the guard with "mitigating" the
storm and rates the impact "low"; it is backwards. The trip is the guard eating legitimate traffic,
and the reason nobody could tell is the same missing component: with no payload identity in the
key or the log line, "N things fanning out" and "one thing looping" produce an identical message.

**The issue's stated cause is refuted.** It blames router-identity miswiring (#1140). Its own log
disproves that: `sender=UWDeepfield/_Activity/import-…-f4dc71b1` is the import **attempt node's own
per-node hub** (the fresh-per-run node `StaticRepoImporter` logs progress to), not the router. The
router appears only as the hub that *packages and forwards* the frames — which is exactly why the
cache hub sees them as `type=RawJson`.

## The fix

The key gains a fourth component: the payload's own identity, resolved **without ever parsing a
payload on the ingestion path** (the breaker sees every inbound delivery, before the turn loop — a
per-message parse would be its own performance defect).

| Where the message is | Where the identity comes from | Cost |
|---|---|---|
| Typed, in-process | `IDiagnosticKeyed.DiagnosticKey` — an already-computed property | one interface check |
| `RawJson`, after a hub hop erased the type | envelope property stamped by `MessageDelivery.Package` | one property lookup, only for `RawJson` |
| Exposes no identity | `null` → key degrades to **exactly the old tuple** | — |

`IDiagnosticKeyed` already existed for the pending-callback diagnostic, for precisely this
"N distinct things vs one thing repeated" ambiguity — this makes it load-bearing. It is now
implemented by `StreamMessage` (`StreamId`, covering `DataChangedEvent`, `PatchDataChangeRequest`,
`UnsubscribeRequest`, `StreamErrorEvent`) and by the node-lifecycle verbs
`Create`/`CreateOrUpdate`/`Delete`/`Copy`/`Move` (the node path).

`Package()` is the one seam where the type is erased for a cross-hub hop, and it is already doing a
full `JsonSerializer.Serialize` — stamping the key there is free by comparison, and delivery
properties already survive both the in-process router hop and the gRPC/SignalR wire (the same
mechanism that carries `PostOptions.RequestId`).

**No band-aid**: threshold (2000), window (1 s) and cooldown (2 s) are untouched, no exemption
list, no widened bound. A genuine single-thing loop still trips exactly as before — and the `Error`
now names *which* stream/path was looping instead of only which two hubs were talking.

### Bounded key space

The payload component makes the key space unbounded in principle: the existing eviction only fires
when a key is *touched again* after an idle window, so a key seen once would live forever. Crossing
a soft cap of 10 000 live counters arms **one inline sweep per window** that drops counters idle
beyond window+cooldown. The live set then tracks *recent* traffic instead of everything the hub
ever saw — no timer, no background thread, no static state, and no effect on detection (a key idle
for seconds is by definition not storming).

## Tests

Deterministic, injected logical clock — no sleeps, no wall-clock races
(`MessageStormBreakerTest`):

1. `WideFanOut_OneTuple_ManyDistinctPayloadKeys_NeverTrips` — **the bug.** 40× the threshold through
   ONE tuple, each message about a different path → must not trip.
2. `RepeatedSamePayloadKey_StillTrips_AndNamesTheThing` — the guard is intact; the trip names the
   thing; and a fan-out running *alongside* the loop keeps flowing (the drop is scoped to the one
   storming key, not to everything sharing its sender/target/type).
3. `PackagedDelivery_KeepsPayloadIdentity_AcrossTheTypeErasure` — the same fan-out survives
   `Package()`; a packaged loop still trips, reported as `type=RawJson` with the payload named.
4. `NoPayloadIdentity_FallsBackToTheRoutingTuple_AndStillTrips` — the fallback is **fail-closed**.
5. `TrackedKeys_AreBounded_ByRecentTrafficNotAllTimeTraffic` — 500 one-shot keys are swept.

**Red-before proof** (payload component neutered, everything else identical):

```
Failed WideFanOut_OneTuple_ManyDistinctPayloadKeys_NeverTrips
  Expected value to be false because a write to a DISTINCT path is not a loop,
  however many of them one importer issues, but found True.
Failed RepeatedSamePayloadKey_StillTrips_AndNamesTheThing
  Expected "UWDeepfield/_Activity/import-f2bb979af363573d" …, but found <null>.
Failed PackagedDelivery_KeepsPayloadIdentity_AcrossTheTypeErasure
Failed TrackedKeys_AreBounded_ByRecentTrafficNotAllTimeTraffic
Failed: 4, Passed: 6
```

## Local verification (CI parity)

- `dotnet build MeshWeaver.slnx -c Release -p:CIRun=true -warnaserror` → **0 Errors, 0 Warnings**
  (re-run after merging current `main`).
- `MeshWeaver.Messaging.Hub.Test` → 191/191
- `MeshWeaver.Data.Test` → 311/311
- `MeshWeaver.Hosting.Monolith.Test` → 582 passed / 3 pre-existing skips
- `MeshWeaver.Documentation.Test` → 20/20

What's New entry included (`Category: Fix` — a user notices imported content going missing).

### DIFF
--- a/src/MeshWeaver.Data.Contract/Messages.cs
+++ b/src/MeshWeaver.Data.Contract/Messages.cs
@@ -124,7 +124,20 @@ public enum ChangeType
 /// Base type for messages carried over a synchronization stream.
 /// </summary>
 /// <param name="StreamId">The identifier of the stream the message belongs to.</param>
-public abstract record StreamMessage(string StreamId);
+public abstract record StreamMessage(string StreamId) : IDiagnosticKeyed
+{
+    /// <summary>
+    /// The stream id — the thing this message is ABOUT.
+    ///
+    /// <para>🚨 Load-bearing, not cosmetic. One owner hub holds a SEPARATE sync stream per
+    /// subscriber/reference, and every one of them posts to the SAME subscriber address with the
+    /// SAME message type. Without this component the hub-ingestion <c>MessageStormBreaker</c>
+    /// folds all of them into one rate bucket, so a wide, legitimate change fan-out (a bulk import
+    /// driving many streams) is indistinguishable from ONE stream in a resubscribe/repost loop —
+    /// and the breaker DROPS the fan-out's frames at ingestion. See <see cref="IDiagnosticKeyed"/>.</para>
+    /// </summary>
+    string IDiagnosticKeyed.DiagnosticKey => StreamId;
+}
 /// <summary>
 /// Base type for stream messages that carry a versioned JSON change.
 /// </summary>
--- a/src/MeshWeaver.Documentation/Data/Architecture/ActionBlockWedgePrevention.md
+++ b/src/MeshWeaver.Documentation/Data/Architecture/ActionBlockWedgePrevention.md
@@ -48,6 +48,18 @@ for any *future* defect: a per-hub **aggregate** backpressure breaker.
   saw was **many distinct keys** — each phantom path, each failed area is a different key — so
   no single key crossed the threshold while the *aggregate* saturated the thread. **Per-key is
   the gap.**
+
+  🚨 "Per-key" means `(sender, target, message-type, **payload identity**)`. The fourth component
+  is not optional decoration — without it the claim above is false, because the mesh funnels
+  traffic through dispatchers where one sender talks to one target with one message type about
+  **many different things** (every sync stream an owner holds to the shared node cache; every
+  `CreateOrUpdateNodeRequest` a bulk importer sends to the mesh hub). Keyed on the routing tuple
+  alone, a wide *legitimate* fan-out is arithmetically identical to one thing looping, and the
+  breaker drops it — real writes discarded at ingestion (#1200). The identity comes from
+  `IDiagnosticKeyed.DiagnosticKey` on the message (a stream id, a node path); once a hub hop has
+  erased the payload type to `RawJson`, it comes from the envelope property `MessageDelivery.Package`
+  stamped there, so the breaker never parses a payload on the ingestion path. A message exposing no
+  identity keys on the bare tuple — the fallback is the old, stricter behaviour, never "allow".
 - The fix: a per-action-block watermark on inbound depth/rate. When one block's queue exceeds
   the watermark, **shed `[CanBeIgnored]`/failure-class messages** (never user-facing or
   lifecycle messages) to keep it draining. The breaker is keyed on the **hub**, aggregated
--- a/src/MeshWeaver.Documentation/Data/WhatsNew/2026-08-12-bulk-imports-no-longer-lose-content-to-the-storm-guard.md
+++ b/src/MeshWeaver.Documentation/Data/WhatsNew/2026-08-12-bulk-imports-no-longer-lose-content-to-the-storm-guard.md
@@ -1,0 +1,25 @@
+---
+Name: Bulk imports no longer lose content to the storm guard
+Category: Fix
+Description: A large import or sync could have some of its writes silently discarded by the runaway-message guard; the guard now tells a wide fan-out apart from a genuine loop.
+Icon: Sparkle
+Order: -20260812
+---
+
+# Bulk imports no longer lose content to the storm guard
+
+The platform carries a safety guard that watches for runaway message loops and cuts them off
+before they can freeze a workspace. The guard judged traffic only by who sent it, who it was for,
+and what kind of message it was — it could not see *what each message was about*. A large import or
+sync sends thousands of writes that share all three of those, one per page or record, so the guard
+mistook the whole batch for a single message repeating itself and discarded part of it.
+
+The result was content that simply never appeared: an import would report itself finished while
+some of its pages, or some of its progress log, were missing, with nothing in the interface to say
+anything had been dropped.
+
+The guard now also considers which page, node or stream each message concerns. Thousands of writes
+to thousands of different places are recognised as normal work and pass through untouched, while a
+genuine loop — the same one thing repeating thousands of times a second — is still stopped just as
+before. As a bonus, when the guard does fire it now names the exact item that was looping, so the
+underlying cause can be found instead of guessed at.
--- a/src/MeshWeaver.Mesh.Contract/CreateNodeRequest.cs
+++ b/src/MeshWeaver.Mesh.Contract/CreateNodeRequest.cs
@@ -14,8 +14,16 @@ namespace MeshWeaver.Mesh;
 /// </summary>
 /// <param name="Node">The MeshNode to create</param>
 [CreateNodePermission]
-public record CreateNodeRequest(MeshNode Node) : IRequest<CreateNodeResponse>
+public record CreateNodeRequest(MeshNode Node) : IRequest<CreateNodeResponse>, IDiagnosticKeyed
 {
+    /// <summary>
+    /// The path of the node being created — so a bulk fan-out (N creates for N DISTINCT paths from
+    /// one importer to the mesh hub) is told apart from ONE create retried in a loop. Both the
+    /// pending-callback diagnostic and the ingestion storm breaker key on it; without it the whole
+    /// fan-out is one bucket and the breaker drops it. See <see cref="IDiagnosticKeyed"/>.
+    /// </summary>
+    string IDiagnosticKeyed.DiagnosticKey => Node is { } node ? node.Path : string.Empty;
+
     /// <summary>
     /// The user or system requesting the creation.
     /// </summary>
@@ -216,8 +224,14 @@ public class CreateNodesPermissionAttribute() : RequiresPermissionAttribute(Perm
 /// </summary>
 /// <param name="Path">The path of the node to delete</param>
 [RequiresPermission(Permission.Delete)]
-public record DeleteNodeRequest(string Path) : IRequest<DeleteNodeResponse>
+public record DeleteNodeRequest(string Path) : IRequest<DeleteNodeResponse>, IDiagnosticKeyed
 {
+    /// <summary>
+    /// The path being deleted — tells a wide prune (N distinct paths) apart from one delete
+    /// retried in a loop. See <see cref="IDiagnosticKeyed"/>.
+    /// </summary>
+    string IDiagnosticKeyed.DiagnosticKey => Path ?? string.Empty;
+
     /// <summary>
     /// If true, also delete all descendant nodes.
     /// </summary>
@@ -361,8 +375,15 @@ public enum NodeDeletionRejectionReason
 /// Both checks run through the standard permission pipeline.</para>
 /// </summary>
 [CreateOrUpdateNodePermission]
-public record CreateOrUpdateNodeRequest(MeshNode Node) : IRequest<CreateOrUpdateNodeResponse>
+public record CreateOrUpdateNodeRequest(MeshNode Node) : IRequest<CreateOrUpdateNodeResponse>, IDiagnosticKeyed
 {
+    /// <summary>
+    /// The path being upserted. This is THE bulk verb — <c>StaticRepoImporter</c> issues one per
+    /// source node from a single import hub to a single mesh hub, so without this component the
+    /// entire import is one storm-breaker bucket. See <see cref="IDiagnosticKeyed"/>.
+    /// </summary>
+    string IDiagnosticKeyed.DiagnosticKey => Node is { } node ? node.Path : string.Empty;
+
     /// <summary>Optional JSON Patch payload (Json.Patch.JsonPatch) to apply to
     /// the existing node. When null, <see cref="Node"/> is the full instance
     /// to upsert. Typed as <c>object?</c> so the patch type is owned by the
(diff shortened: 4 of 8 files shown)


rbuergi:
Both review findings were real defects in this change and are fixed in 4097afd5a:

1. **`GetOrAddCounter` could return an unpublished counter.** On a lost `TryAdd`, if the winner's entry was evicted (window roll / sweep) before the re-read, the fallback handed back the locally-built instance — that delivery would be counted on an orphan nobody else can see, silently resetting a storming key's count and cooldown. Replaced with a retry loop that only ever returns a counter it has observed in, or successfully published to, the dictionary.

2. **`SweepStaleKeys` exempted `Tripped` counters**, so a key that stormed once and was never seen again lived for the hub's lifetime — defeating the very bound the sweep documents. Staleness is now the only criterion, and it is the correct one: *stale* means no message for a full window+cooldown, which an actively storming key can never be (every message restamps its window) and which is already past the cooldown — so releasing the counter is behaviourally identical to the self-heal its next message would perform. New test `TrackedKeys_Sweep_AlsoReleases_TrippedThenIdleKeys` pins it, including that the formerly-storming key flows again afterwards.

The opportunistic in-`ShouldDrop` eviction deliberately keeps its `!Tripped` guard — that path fires on a zero-count window roll for a key still being touched, which may legitimately be mid-cooldown.

`MeshWeaver.Messaging.Hub.Test` 192/192; full solution `-c Release -p:CIRun=true -warnaserror` 0 errors / 0 warnings on the tree merged with current main.

rbuergi:
## CI is RED on shard 2 — analysis: not caused by this diff

**What actually failed.** `MeshWeaver.Hosting.Monolith.Test (part 2/2) exit=124 TIMEOUT (6m wall-clock cap)`. Every other project in shard 2 is green (Graph 1025, Layout 402, Auth 143, FutuRe 59, Hosting 145, …). The cap was hit because *every* dynamic NodeType compile in that runner failed and each compile test then waited out its 40–60 s budget:

```
System.NullReferenceException
  at Microsoft.CodeAnalysis.CSharp.Symbols.NamedTypeSymbol.Microsoft.Cci.ITypeDefinitionMember.get_ContainingTypeDefinition()
  at Microsoft.Cci.MetadataWriter.<GetConsolidatedTypeParameters>g__getConsolidatedTypeParameters|104_0(...)
  at Microsoft.Cci.FullMetadataWriter.CreateIndicesForNonTypeMembers(...)
  ...
  at Microsoft.CodeAnalysis.Compilation.Emit(...)
```

That is a crash **inside Roslyn's metadata writer** while emitting the NodeType assembly — StaleBarrierType, CodeEditType, EmptyRaceType, ButtonType, DirtyType and BrokenType all die identically, i.e. a process-wide condition, not per-type source.

**Why this diff cannot be the cause — provable, not asserted.** The job log contains **zero** `MESSAGE STORM` lines. In `ShouldDrop`, a drop is only reachable once `Counter.Tripped` is set, and the only place that sets it is the trip transition, which *always* logs that Error. Zero trip logs ⇒ `Tripped` was never set ⇒ **the breaker dropped nothing in this job, before or after the change**. The key's composition is therefore behaviourally inert here; a finer key can only ever lower a per-key count, never raise one.

The only other behavioural delta in the diff is `Package()` stamping one extra string onto the delivery envelope. That cannot reach Roslyn's emitter.

**Local evidence on this exact commit (4097afd5a):**
- `CodeEditRecompileTest` — 11/11 pass
- `MeshWeaver.Hosting.Monolith.Test` (full) — 582 passed / 3 pre-existing skips
- full solution `-c Release -p:CIRun=true -warnaserror` — 0 errors / 0 warnings

**Controls checked** (same signature searched, not found): main runs 31580689228 and 31577940567 (shard 2 red for unrelated reasons — "Area not found", observable timeouts), `fix/1250-live-supplement-rls-bypass` (FutuRe exit=1), `fix/nodetype-recompile-alc-leak` shards 0 and 3. Branch is 0 commits behind main, so the stale-branch path is excluded too.

⚠️ **The Roslyn emit NRE looks like a genuine defect worth its own issue** — when it happens, *every* NodeType in the process fails to compile, which in production is the condition that refuses portal readiness. `CodeEditRecompile` is already a named flake family in AGENTS.md. Not re-running to see whether it reproduces (project rule: a re-run to test flakiness hides the bug).

Not merging — a coordinator runs the merge train, and the gate is green CI.

rbuergi:
## Shard 2 attribution — checked reachability both ways, then merged main

**What failed:** `MeshWeaver.Hosting.Monolith.Test (part 2/2) exit=124 TIMEOUT (6m wall-clock cap)`. Every other project in the shard was green (Graph 1025, Layout 402, Auth 143, FutuRe 59, Hosting 145, ContentCollections 29, …). The cap was hit because *every* dynamic NodeType compile in that runner failed and each compile test then sat out its 40–60 s budget — StaleBarrierType, CodeEditType, EmptyRaceType, ButtonType, DirtyType, BrokenType, all identically:

```
System.NullReferenceException
  at NamedTypeSymbol.Microsoft.Cci.ITypeDefinitionMember.get_ContainingTypeDefinition()
  at Microsoft.Cci.MetadataWriter.<GetConsolidatedTypeParameters>g__getConsolidatedTypeParameters|104_0
  at Microsoft.Cci.FullMetadataWriter.CreateIndicesForNonTypeMembers
  ... at Microsoft.CodeAnalysis.Compilation.Emit(...)
```

i.e. a crash **inside Roslyn's metadata writer**, process-wide rather than per-type.

**I did not assume "known flake".** This diff touches `MessageStormBreaker`, `MessageDelivery` and five node-lifecycle request types, all extremely widely reachable, so I traced both behavioural deltas to their consumers:

1. **Five request types now implement `IDiagnosticKeyed`.** The only consumer is `MessageHub.Observe` → `GetOrAddResponseSubject(messageId, requestType, target, DiagnosticKey)`. `responseSubjects` is keyed on `messageId` (a fresh `Guid`); the key is stored as metadata on the `PendingCallback` record and read only by the stale-callback report. **Request/response correlation is untouched.**
2. **`Package()` stamps one extra envelope property.** Consumers of `Properties` wholesale are `PostOptions.ResponseFor`/`WithProperties` (copy-through) and `HierarchicalRouting` (forward). Nothing keys on, counts, or validates properties. **Inert metadata.**
3. **The breaker key itself is strictly finer**, so it can only ever *lower* a per-key count — it cannot cause a drop the old key would not also have caused. And that job logged **zero** `MESSAGE STORM` lines; since `Tripped` is only ever set on the trip transition, which always logs, the breaker dropped nothing there under either key.

None of those paths reaches `Compilation.Emit`.

**Controls** (same Roslyn signature searched, absent in all): main runs 31580689228 and 31577940567 — shard 2 red for *different* reasons — plus `fix/1250-live-supplement-rls-bypass` and `fix/nodetype-recompile-alc-leak` (shards 0/3).

**Action taken:** merged current `origin/main` (9 commits — #1168 startup-readiness, #1264 code-cell result, #1269 AiSettings seed, the parked-NodeType fix, plus new `CompileWaitDoesNotTimeoutTest` / `NodeTypeParkedException` / `DynamicTypePreWarmerTest`, all squarely in the failing area) and re-ran — a real change, not a flake retry.

**Local on the merged tree:** `MeshWeaver.Hosting.Monolith.Test` 582 passed / 3 pre-existing skips / **0 failed**; `CodeEditRecompileTest` 11/11. Touched projects each built individually `-c Release -warnaserror` → 0 errors, 0 warnings.

PR #1267, opened 2026-08-12
fix(query): the live change feed no longer bypasses row-level security (#1250)

### LINKED ISSUE #1250: Live secure query: the change-feed supplement bypasses RLS, so a subscriber can transiently receive a node it cannot read
Noticed while fixing the satellite half of this in #1193 — same code path, same shape, different
filter.

## The gap

`StorageAdapterMeshQueryProvider.ObserveQuerySecure` runs its SNAPSHOT through
`RunQueryNodes(request, options, useSecurityFilter: true)`, so the initial payload is RLS-filtered.
The live path's write-lag supplement in `ProcessBatch` is not: it admits `change.Entity` straight
from the process-wide change feed on `_evaluator.Matches` alone.

`src/MeshWeaver.Hosting/Persistence/Query/StorageAdapterMeshQueryProvider.cs` — `ProcessBatch`,
the `changesByPath` loop.

`QueryEvaluator.Matches` consults only the query's filter and free-text term
(`QueryEvaluator.cs:32-47`) — it performs no permission check of any kind. So for a live secure
subscription, a write to a node **under the query's base path that the subscriber cannot read** is
emitted to that subscriber as an `Added`, carrying the node (and its `Content`). The next change
re-queries with the security filter applied, does not find it, and emits `Removed` — the same
add-then-remove flicker #1193 fixes for satellites, except here the transiently-visible row is one
the caller was never entitled to see.

Scope-limited but real: `PathMatcher.ShouldNotify` bounds it to the query's base path and scope, so
it needs a subscriber with read on some ancestor but not on the specific descendant that changed —
which is exactly what a per-node deny / narrower grant under a readable space produces.

## Why it was not fixed in #1193

That PR made the supplement apply the snapshot's *result-set* exclusions (satellite paths,
partition roots) — pure, synchronous predicates over the node. The RLS filter is neither: it is an
observable permission evaluation, and `ProcessBatch` is a synchronous step inside a
`Concat`-serialised batch pipeline. Doing it properly means either lifting the supplement into the
reactive chain or dropping the supplement for the secure variant and letting the re-query be the
only source (the supplement exists for backends whose notification can outrun row visibility —
PG's LISTEN/NOTIFY — so dropping it is not free).

## Suggested check when fixing

A test in the shape of `ObserveQueryFreshnessTest` / the access-control suites: subscriber with read
on a space but denied on one child, live secure subscription over the space, write the denied child,
assert the subscription never emits it. It should fail on current `main`.

rbuergi:
Closes #1250.

## The bypass

`StorageAdapterMeshQueryProvider`'s live pipeline re-queries storage on every change
notification, then **supplemented** that re-query with the notification's raw `Entity`,
admitting it on `QueryEvaluator.Matches` alone. `Matches` consults only the query's filter
and free-text term (`QueryEvaluator.cs:32-47`) — **no permission check of any kind**.

The snapshot leg is RLS-filtered (`CollectMatched` → `ValidateRead`). The live leg was not.
So on the **secure** surface — `IMeshQueryProvider.Query`, which `MeshQuery.Query<T>` and
therefore `IMeshService.Query<T>` / `workspace.GetQuery` route to — a write under a live
subscription's base path to a node the subscriber cannot read was emitted to them as an
`Added` **carrying the node and its `Content`**.

Nothing downstream re-filtered it: `MeshQuery.TryFilterDuplicateLiveChange` forwards duplicate
`Added`s by design (#889).

**The subscriber picks the base path freely.** The query need not be over anything they can
read — an RLS-filtered Initial over an unreadable path simply comes back empty. The exposure
needed nothing but a live subscription plus a same-process write under it.

**It was not necessarily transient.** The injected node is written into `currentItems`, so it
stays in the caller's result set — and on screen — until the next in-scope change emits a
`Removed`. On a quiet subtree that is indefinite.

## Who was exposed

`AddSingleton<IMeshQueryProvider>` (not `TryAdd`) — this provider is always in the merge,
alongside any native one. Exposure is the intersection of "the adapter populates `Entity`"
and "this provider serves the query shape":

| Backend | `Entity` on the feed | Defer | Exposure |
|---|---|---|---|
| **In-memory** | yes (`InMemoryStorageAdapter:131`) | none | **every live secure query** |
| **SQLite** | yes (`SqliteStorageAdapter:161`) | none | **every live secure query** |
| **File system** | yes (`FileSystemChangeWatcher:201-202`) | none | **every live secure query** |
| **PostgreSQL** | yes, in-process (`:463`, `:579`) | `DeferToNativeProvider = true` | **scoped SATELLITE reads only** — `_Access`, `_Thread`, `_Activity`, `_Comment`, satellite `nodeType:` filters, `source:activity|accessed` (`DefersToNativeProvider` returns false for exactly those) |
| **Snowflake** | yes (`:485`) | `= true` | same satellite residue as PG |
| **Cosmos** | never (`Updated(path, null)`) | n/a | not reachable via this path |

So hosted PostgreSQL portals were **not** clear of this: the shapes the pedestrian still
serves there are the access grants, threads, activities and comments — a live
`path:{space}/_Thread scope:children` or a thread-list subscription would flash a satellite
the viewer has no grant on. Unscoped and scoped-primary content queries on PG go to
`PostgreSqlPartitionedMeshQuery`, which has no supplement.

## Which fix, and why

**Branch taken: delete the supplement.** The issue left open whether to lift `ProcessBatch`
into the reactive chain so `ValidateRead` applies. That is unnecessary — the supplement is
dead code for its stated purpose. Verified across every adapter and change source:

- **Every source that populates `Entity` commits to the store BEFORE publishing.**
  `InMemoryStorageAdapter.Write` (`AddOrUpdate` → `OnNext`), `SqliteStorageAdapter.Write`
  (`ExecuteNonQuery` → `OnNext`), `PostgreSqlStorageAdapter.Write` (`await WriteAsyncCore`
  → `OnNext`) and `PublishChanges` (post-commit), `SnowflakeStorageAdapter.Write` (same
  shape). `FileSystemChangeWatcher` is the strongest case: it notifies with the node it just
  **read back out of the adapter**. In every one of them the re-query that the notification
  triggers is guaranteed to see the row.
- **Every source whose notification can outrun row visibility passes `Entity = null`** —
  `PostgreSqlChangeListener:130` (the LISTEN/NOTIFY path the supplement's own comment cited as
  its justification), `CosmosChangeFeedProcessor:149`, `SnowflakeChangeFeedPoller:177-179`, and
  the `Deleted` legs of the SQL adapters. The supplement could not fire for them in principle.
- **The `Deleted` branch was already non-load-bearing.** Only the in-memory adapter carries a
  tombstone entity; PG/SQLite/Snowflake deletes pass `null`, so delete propagation has always
  come from the re-query alone on every real backend.
- **Whenever the supplement DID change the outcome, the re-query was the one that was right.**
  The row it re-added had been excluded by RLS, by satellite-table routing (#1193), by a path
  filter, or clipped past the load cap. It could also install an *older* notification entity
  over a newer re-queried row.

The notification is now strictly a **trigger**: `RunQuery` is the only source of rows, so
every frame — Initial and every live delta — carries exactly what the (RLS-filtered, on the
secured surface) read returned. The `Concat` serialisation and ordering are untouched; only
the tuple carrying the notification alongside the snapshot is gone.

This also **retires the #889 provider race** documented at
`MeshQuery.TryFilterDuplicateLiveChange`: the pedestrian no longer emits a raw-entity `Added`
ahead of the per-schema PostgreSQL delegate's authoritative row. The forward-duplicates
contract stays (independent providers can still race), with the doc updated to say so.

## Test

`test/MeshWeaver.Security.Test/LiveQueryRowLevelSecurityTest.cs` — `MonolithMeshTestBase`,
real access control, no mocking. A subscriber holds an Editor grant on **one child only**;
they open a live **secure** subscription over the parent namespace; a denied sibling is then
written, followed by a touch of the readable node.

- **Red on `main`:** `Expected collection to be empty … leaked as: Added, Removed, but found
  2 item(s)` — exactly the add-then-remove flicker the issue predicts.
- **Green with the fix:** no frame of any change type carries the denied node, and its content
  string never appears in any emission.
- The positive half is asserted too: the live `Updated` for the readable node still arrives.
- No sleep. The denied write is ordered before the readable one and the pipeline is
  `Concat`-serialised, so observing the readable node's `v2` frame is a deterministic barrier
  proving the denied node's frame has already been fully processed. No "exactly N events"
  assertion — the check filters on emission shape.

## Verification

Release + `-warnaserror`, per project: `MeshWeaver.Hosting`, `MeshWeaver.Documentation`,
`MeshWeaver.Security.Test`, `MeshWeaver.Hosting.Monolith.Test` — all `0 Warning(s) 0 Error(s)`.

| Suite | Result |
|---|---|
| `MeshWeaver.Security.Test` (full) | 399 / 399 |
| `MeshWeaver.Query.Test` | 388 / 388 |
| `MeshWeaver.Persistence.Test` | 617 / 617 |
| `MeshWeaver.Hosting.Test` | 145 / 145 |
| `MeshWeaver.Documentation.Test` (WhatsNew + link integrity) | 2 / 2 |
| Monolith `SearchResultStormTest` + `ObserveQueryFreshnessTest` | 11 / 11 |

`SearchResultStormTest.CatalogSubscription_SatelliteWriteUnderTheRoot_NeverEntersTheResultSet`
(#1193's regression pin) still passes — the invariant it guards is now held by construction.

What's New: `Category: Fix` —
`2026-08-12-a-live-list-can-no-longer-flash-a-node-you-cannot-open.md`.

### DIFF
--- a/src/MeshWeaver.Documentation/Data/WhatsNew/2026-08-12-a-live-list-can-no-longer-flash-a-node-you-cannot-open.md
+++ b/src/MeshWeaver.Documentation/Data/WhatsNew/2026-08-12-a-live-list-can-no-longer-flash-a-node-you-cannot-open.md
@@ -1,0 +1,28 @@
+---
+Name: A live list can no longer flash a node you cannot open
+Category: Fix
+Description: A live search or catalog list could briefly show a node the viewer has no permission to read — including its content — when someone else edited it, before dropping it again on the next update.
+Icon: Sparkle
+Order: -20260812
+---
+
+# A live list can no longer flash a node you cannot open
+
+A list that stays live — a search page, a catalog, a folder view — is built twice over: once when
+the page loads, from the full answer to your query, and then continuously, as changes come through.
+The load side has always applied your permissions, so a node you may not read simply is not in it.
+
+The live side did not. It admitted whatever had just been written under the area you were watching,
+checking only whether it matched your search words and filters — never whether you were allowed to
+see it. So when someone edited a node you have no access to, and that node sat under an area your
+list covered, it appeared in your list as a new row, carrying its name and its content. The next
+update took it back out again, because that frame was rebuilt from the permission-checked answer.
+A row you were never entitled to, visible for as long as nothing else changed.
+
+The live side now shows only what the permission-checked read returns, for every update, not just
+the first one. Nothing else about live lists changes: an edit to something you *can* see still
+arrives immediately.
+
+This affected portals backed by the in-memory, SQLite, or file-system stores. On PostgreSQL — which
+every hosted portal uses — the main content queries are served by a different, unaffected component,
+so the exposure there was limited to scoped queries over attached records.
--- a/src/MeshWeaver.Hosting/Persistence/Query/MeshQuery.cs
+++ b/src/MeshWeaver.Hosting/Persistence/Query/MeshQuery.cs
@@ -858,10 +858,14 @@ public class MeshQuery : IMeshQueryCore
     ///
     /// <para>🚨 An <c>Added</c> for a path that is ALREADY live is NOT dropped — it flows through
     /// exactly like an <c>Updated</c>. Dropping it was the terminal dropped-update of issue #889:
-    /// two providers legitimately race the same write (the pedestrian
-    /// <c>StorageAdapterMeshQueryProvider</c> supplements its re-query with the change
-    /// notification's raw entity and usually emits FIRST; the per-schema PostgreSQL delegate
-    /// re-queries storage and emits the authoritative row a beat LATER — both as <c>Added</c>).
+    /// two providers legitimately race the same write and both announce it as <c>Added</c>, at
+    /// different times and with different content. (The #889 instance was the pedestrian
+    /// <c>StorageAdapterMeshQueryProvider</c> supplementing its re-query with the change
+    /// notification's raw entity and emitting FIRST, while the per-schema PostgreSQL delegate
+    /// re-queried storage and emitted the authoritative row a beat LATER. #1250 deleted that
+    /// supplement — the pedestrian now only ever emits what its own read returned — so that
+    /// particular race is retired, but the merge still fans in independent providers and the
+    /// contract below is what keeps their races safe.)
     /// The old dedup forwarded whichever arrived first and DISCARDED the second — discarding the
     /// authoritative content correction. When the raced write was the LAST write touching the
     /// query (PaywallRealGateShapeTests' buyer grant), no later change ever healed the snapshot:
--- a/src/MeshWeaver.Hosting/Persistence/Query/StorageAdapterMeshQueryProvider.cs
+++ b/src/MeshWeaver.Hosting/Persistence/Query/StorageAdapterMeshQueryProvider.cs
@@ -208,9 +208,10 @@ internal class StorageAdapterMeshQueryProvider : IMeshQueryProvider, IMeshQueryC
                 var (matched, parsedQuery, _) = collected;
 
                 // Satellite paths and partition roots are not content-query results.
-                // 🚨 The rules live in IsExcludedFromResults — one predicate, documented there —
-                // precisely so the LIVE path (ProcessBatch's notification supplement) applies the
-                // same set. Duplicating them here is how the two drifted apart (#1193).
+                // 🚨 The rules live in IsExcludedFromResults — one predicate, documented there.
+                // This read is now the SOLE source of rows for both the Initial snapshot and every
+                // live delta (#1250 removed the live path's parallel admission of raw notification
+                // entities), so applying them here applies them everywhere.
                 IEnumerable<object> matchedNodes =
                     matched.Where(n => !IsExcludedFromResults(n, parsedQuery));
 
@@ -328,12 +329,12 @@ internal class StorageAdapterMeshQueryProvider : IMeshQueryProvider, IMeshQueryC
     }
 
     /// <summary>
-    /// The result-set exclusions this provider applies to EVERY emission — the snapshot
-    /// (<see cref="RunQueryNodes"/>) and the live change feed
-    /// (<see cref="ProcessBatch{T}"/>'s notification supplement) alike.
+    /// The result-set exclusions this provider applies to EVERY emission. They live in
+    /// <see cref="RunQueryNodes"/> — the single read behind the Initial snapshot AND every live
+    /// delta (<see cref="ProcessBatch{T}"/> only diffs that read's output).
     ///
-    /// <para>🚨 <b>One predicate, both paths (#1193).</b> The snapshot used to own these rules
-    /// inline while the live path's write-lag supplement admitted any notified entity on
+    /// <para>🚨 <b>One predicate, one path (#1193, #1250).</b> The snapshot used to own these rules
+    /// inline while the live path ran a parallel admission of raw change-notification entities on
     /// <c>_evaluator.Matches</c> alone — and <c>Matches</c> is <c>true</c> for EVERY node when the
     /// query carries no filter and no free-text term (a bare <c>path:X scope:descendants</c>
     /// catalog listing is exactly that shape). So a write to a node the snapshot deliberately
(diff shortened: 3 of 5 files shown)


rbuergi:
## CI triage — the red on runs 1 and 2 is three pre-existing `main` flakes, none in this diff's blast radius

Recording the evidence so nobody re-triages it. This diff changes exactly one thing: which rows the
pedestrian query provider's **live delta** emits. It touches no write path, no node lifecycle, no
layout-area render, and no activity state machine.

**Run 1** (`4953958e`) — shards 2 + 3:

| Failure | Also on `main` |
|---|---|
| `FutuRe.Test.FutuReAnalysisTest.EuropeRe_AnnualReport_EmbeddedCharts_ShouldRenderViaPathResolution` | run [31580689228](https://github.com/Systemorph/MeshWeaver/actions/runs/31580689228) (`5dc532a1`, same wall-clock window) and [31573558342](https://github.com/Systemorph/MeshWeaver/actions/runs/31573558342) |
| `AI.Test.SkillAutocompleteTest.Autocomplete_ListsBuiltInSkills_FromCatalog` | run [31574636901](https://github.com/Systemorph/MeshWeaver/actions/runs/31574636901) |

**Run 2** (`3422dc74`, after merging main) — shards 3 + 4:

| Failure | Verdict |
|---|---|
| `AI.Test.SkillAutocompleteTest.Autocomplete_ListsBuiltInSkills_FromCatalog` | same flake again |
| `AI.Test.CodeCellOutputCaptureTest.Progress_Of_Succeeded_Activity_Renders_Output_For_PublicRead_Viewer` | **documented** CI-only flake — the test's own source names three prior sightings (runs `31401910253` / `31407254207-2` / `31414605989`) and carries a diagnostic block written for exactly this |

Across `main`'s last 25 Build-and-Test runs the recurring failures are `EuropeRe_AnnualReport…` (×3)
and `Autocomplete_ListsBuiltInSkills_FromCatalog` (×3).

### Why `CodeCellOutputCaptureTest` is not this change

Its own failure diagnostic answers the question it was written to answer:

> `FAILURE-TIME NODE STATE: Status=Running, Messages=1, End=.` Node is NOT Succeeded although this
> test already observed Succeeded before subscribing ⇒ **a later write regressed the node
> (write-order bug)**.

The **activity node itself** rolled back from `Succeeded` to `Running`. That is a write-ordering
defect on the node, upstream of any render. This PR deletes a read-side row admission; it cannot
move a node's persisted `Status` backwards. Note also that the captured output line *did* render for
the PublicRead viewer (`Message = out-line-public`) — the access-shaped half of that test passed.

### Local runs on this branch (all green)

`FutuReAnalysisTest.EuropeRe_AnnualReport_EmbeddedCharts_ShouldRenderViaPathResolution` 1/1 ·
`SkillAutocompleteTest` 1/1 · `CodeCellOutputCaptureTest` 5/5 — plus the full suites in the PR body
(`Security.Test` 399/399, `Query.Test` 388/388, `Persistence.Test` 617/617, `Hosting.Test` 145/145).

Now re-run on a fresh merge of `main` (which had moved 14 commits ahead).

rbuergi:
## Run 3 (`bb138f9d`, merge of main @ `e4de9129`) — red on a break that is in `main`, proven by bisect

Shard 4: `PluginCatalog.Test.StaleStampRootBindingTest.SelfTypedRootWhoseTypeNeverCompiles_SkipsThePublishPromptly`

> Expected **92898** to be less than 30000 … the install must not post into a root it cannot
> activate and then wait out the hub's whole request timeout

This one is **not** a load flake — it reproduces deterministically, and it is **not** this PR.
Bisected locally in this worktree:

| Tree | `install of the broken package returned at` | Result |
|---|---|---|
| This branch (`bb138f9d` = main @ `e4de9129` + the fix) | **93 189 ms** | fail |
| Same tree with `StorageAdapterMeshQueryProvider.cs` + `MeshQuery.cs` checked out from `origin/main` — i.e. **pure main** | **93 048 ms** | fail |

A 141 ms delta on a 93 s measurement: the break is entirely in `main`, and this PR's diff (a
read-side row admission, ~50 lines, no write path) accounts for none of it. CI's 92 898 ms matches
the local numbers to within 0.3 %, so it is a real bounded wait, not runner contention.

`StaleStampRootBindingTest` arrived on main in `75c68f618` and reached this branch only via the
merge; the install now spends ~93 s where the test's own doc records 393 ms. That looks like a
regression in the NodeType compile-wait path that landed alongside it — worth its own issue, but
it belongs to whoever owns that change, not here.

### Running total across four CI runs on this branch

Every failure so far is reproducible on `main` or is a documented pre-existing flake; none is in
this diff's blast radius, and all of them pass locally on this branch except the one above, which
fails identically without the diff:

| Failure | Verdict |
|---|---|
| `FutuReAnalysisTest.EuropeRe_AnnualReport_EmbeddedCharts_ShouldRenderViaPathResolution` | main flake, 3× in main's last 25 runs |
| `SkillAutocompleteTest.Autocomplete_ListsBuiltInSkills_FromCatalog` | main flake, 3× in main's last 25 runs |
| `CodeCellOutputCaptureTest.Progress_…_PublicRead_Viewer` | documented CI-only flake; its own diagnostic reports the **activity node regressed `Succeeded`→`Running`** — a write-order bug, and this PR writes nothing |
| `StaleStampRootBindingTest.SelfTypedRootWhoseTypeNeverCompiles_SkipsThePublishPromptly` | **reproduces on pure main** (bisect above) |

PR #1281, opened 2026-08-12
The cell's output pane leads with the RESULT; the status is a footer

rbuergi:
## The report

> "still shows Done => you can show this status at bottom. but output must show control and not Done!!"

#1264 made the returned control render. It rendered it **under** an `H4` "✓ Done", so the pane still *read* as a status report with the answer somewhere below — which is the same complaint one layer up.

## The change

`Progress` (the code cell's output pane) now renders **[Result, Log, Status]**:

- the **result leads** — it is what the reader pressed Run for;
- then whatever the script printed;
- then **one quiet status line at the foot**: the terminal badge drops from `H4` to 0.8rem small print when it sits under a result.

The **running** bar is deliberately untouched — while a run is in flight it is the only content there is, so it stays as-is. `Overview` follows the same order; its header already carries a status badge, so nothing is lost there.

## Named areas, and why

The three slots are now named areas — `Result`, `Log`, `Status` — and the tests address them **by name**. This order is a product decision that has now moved once; index-based assertions (`Areas[0]` = indicator, `Areas[1]` = log) turn the next move into unrelated red, which is exactly what this PR had to repair in `CodeCellOutputCaptureTest`.

`Progress_Renders_The_Control_A_Script_Returned` now pins the ORDER itself — result at index 0, status last — plus the footer's small-print styling, so a regression fails on the thing that was actually reported rather than on an incidental index.

## Verified live

The previous fix is confirmed running on `memex.meshweaver.cloud` (reflection on the loaded assembly: `ResultArea = Result`), and a fresh run of `ThinkInStreams/01-TheCombinatorialTrap/Source/CaseCount` renders `Progress/Result` with its H4, six-row DataGrid and markdown. This PR is the ordering on top of that.

⚠️ Unrelated, found while verifying and worth its own issue: the image tagged `3.0.0-rc1.ci.3103` (`f748d16`) ships `/app` binaries built from `b99c744` — 38 commits earlier. The portal self-updated onto it and ran pre-fix code while every tag and tick said otherwise.

Tests: `MeshWeaver.AI.Test` CodeCellOutputCaptureTest 6/6, `MeshWeaver.Graph.Test` Activity* 71/71, full `dotnet build -c Release -warnaserror` clean.

### DIFF
--- a/src/MeshWeaver.Graph/ActivityLayoutAreas.cs
+++ b/src/MeshWeaver.Graph/ActivityLayoutAreas.cs
@@ -52,10 +52,20 @@ public static class ActivityLayoutAreas
 
     /// <summary>
     /// Area id of the script RESULT inside <see cref="Progress"/> / <see cref="Overview"/> —
-    /// the control a script returned, rendered live. Named (not auto-numbered) so the
-    /// indicator and log keep their positions and so tests address it by name.
+    /// the control a script returned, rendered live. It comes FIRST: it is what the reader
+    /// pressed Run for.
     /// </summary>
     public const string ResultArea = "Result";
+
+    /// <summary>Area id of the message log inside <see cref="Progress"/> / <see cref="Overview"/>.</summary>
+    public const string LogArea = "Log";
+
+    /// <summary>
+    /// Area id of the run's status inside <see cref="Progress"/> / <see cref="Overview"/> — the
+    /// live bar while running, the terminal "✓ Done" line afterwards. LAST, and quiet: see
+    /// <see cref="Progress"/>.
+    /// </summary>
+    public const string StatusArea = "Status";
 
     /// <summary>
     /// The control a script RETURNED, as a live stream — the missing half of #915.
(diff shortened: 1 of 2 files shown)


rbuergi:
Superseded by 330b1f545 ("The cell's execution state lives in its toolbar, not above the result"), which landed on main from a parallel session while this was in CI — same report, better answer.

Where they differ: this PR demoted "✓ Done" to small print at the FOOT of the pane; main removes it entirely for the case that prompted the complaint (a success whose script returned a control shows the result alone, `ActivityLayoutAreas.ShowsStatusLine`) and puts the idle/done state in the cell TOOLBAR as a localized execution chip. A status line that carries no information is better absent than quiet, and en/de beats the hardcoded English this PR kept.

Main's version is covered — `ShowsStatusLine` is pinned in `ActivityProgressViewTest` and asserted end-to-end in `CodeCellOutputCaptureTest` — so nothing from here needs porting. The rendering fix underneath both (#1264) is merged and verified live.

Closing rather than resolving the conflict: merging this would re-add chrome main just removed.

PR #1287, opened 2026-08-12
The coupon list says what a coupon really unlocks

rbuergi:
The label half of **Systemorph/MeshWeaver.Plugins#321**. The code it describes lives in the Store
package (**Systemorph/MeshWeaver.Plugins#422**); the *label* is here, so it ships separately.

## What was wrong

`CouponAdminSettingsTab.cs:127` rendered an empty package list as **"any plugin"**:

```csharp
Unlocks = plugins.Length == 0 ? "any plugin" : string.Join(", ", plugins),
```

That reads as *"this code opens the whole store"*, and the Store has never behaved that way. An
empty list means the coupon may be **entered** on any package — but redeeming it grants only the
package it was used on (`CouponRedemption.GrantTargets`). The label promised the exact opposite of
what the code does, and an administrator scanning the grid had no way to tell a genuinely-generous
code from a one-package one: they printed the same words.

## What it says now

| coupon | before | after |
|---|---|---|
| no list | `any plugin` | `the package it is used on` |
| `grantsAll: true` | `any plugin` (or the list) | `everything` |
| `grantsAll: true` + a list | the list | `everything` |
| a list | the list | the list — unchanged |

`grantsAll` wins the cell deliberately: it is the one thing that unlocks more than the list names,
so showing the list there would understate the coupon.

## Localization

Both labels resolve through the viewer's catalog — `ui.couponUnlocksEverything` and
`ui.couponUnlocksRedeemedOn`, added to `strings.en.json` **and** `strings.de.json` ("alles" / "das
Paket, auf dem er eingelöst wird").

`ToRow` stays a **pure** function: the localizer arrives as an optional `Func<string,string>`,
because the projection has no `LayoutAreaHost` and its unit tests call it directly. The one caller
that *has* a host passes `host.Localize`, so a German admin reads German. Never
`CultureInfo.CurrentUICulture` — a layout-area render hops the hub scheduler and an ambient culture
does not survive it.

The other five cells in this grid (`free`, `always`, `until …`, the currency default) are
pre-existing English literals and are left alone — widening this into a full localization pass of
the tab would bury the one-line semantic fix that the issue is about.

## Reading `grantsAll`

By SHAPE, like every other field here — `Store/Coupon` is a dynamically compiled node type, so this
assembly never sees it as a class. Absent property ⇒ `false`, which is what every coupon written
before the flag existed always meant.

## Verified

* `dotnet test test/MeshWeaver.PluginCatalog.Test --filter CouponAdminSettingsTabTest` — **7/7**
  (the empty-list case renamed to say what it now asserts, plus two new cases: `grantsAll` wins the
  cell even beside a list, and the labels come from the localizer when one is supplied)
* `dotnet test test/MeshWeaver.Messaging.Hub.Test --filter LocalizationTest` — **33/33** (no
  language is missing either new key)
* `dotnet test test/MeshWeaver.Documentation.Test --filter WhatsNewEntryIntegrity` — **1/1**
* `dotnet build -c Release -warnaserror` on `MeshWeaver.PluginCatalog`, `MeshWeaver.Messaging.Hub`
  and `MeshWeaver.Documentation` — **0 warnings, 0 errors**

### DIFF
--- a/clients/react/src/i18n/strings.de.json
+++ b/clients/react/src/i18n/strings.de.json
@@ -425,6 +425,8 @@
   "ui.iconName": "Symbolname:",
   "ui.language": "Sprache:",
   "ui.openCoupon": "Gutschein öffnen:",
+  "ui.couponUnlocksEverything": "alles",
+  "ui.couponUnlocksRedeemedOn": "das Paket, auf dem er eingelöst wird",
   "ui.typeDeleteToConfirm": "Zum Bestätigen DELETE eingeben:",
   "ui.createRelease": "Release erstellen",
   "ui.runTests": "Tests ausführen",
--- a/clients/react/src/i18n/strings.en.json
+++ b/clients/react/src/i18n/strings.en.json
@@ -425,6 +425,8 @@
   "ui.iconName": "Icon Name:",
   "ui.language": "Language:",
   "ui.openCoupon": "Open a coupon:",
+  "ui.couponUnlocksEverything": "everything",
+  "ui.couponUnlocksRedeemedOn": "the package it is used on",
   "ui.typeDeleteToConfirm": "Type DELETE to confirm:",
   "ui.createRelease": "Create Release",
   "ui.runTests": "Run Tests",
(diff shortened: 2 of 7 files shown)


rbuergi:
**The red `@meshweaver/react` check was mine, and is fixed** (`dbf247ff6`).

`clients/react/src/i18n/localize.test.ts` is a catalog **drift guard**: it asserts the React client's `strings.{en,de}.json` have keys *identical* to the server's and compares every value. Adding the two new keys to the server catalog alone therefore split the two apart — and a JS client would have rendered the raw key instead of the label.

Both keys are now in the client catalogs with the same values.

Verified locally: `npx vitest run src/i18n/localize.test.ts` — **15/15 passed**; 902 keys on each side, en and de, zero value drift, no language missing a key.

Not a pre-existing main failure: the guard is exactly reachable from this diff, which is what makes it the right gate.

Issue #1230, opened 2026-08-11
Remove QuestPDF — its Community licence is pay-to-use above a revenue threshold


rbuergi:
MeshWeaver ships dual-licensed Apache-2.0 / MIT. **QuestPDF** (`Directory.Packages.props`, 2026.7.2) is dual-licensed: the Community tier is free only below a revenue threshold; above it a paid commercial licence is required. `MarkdownExportExtensions.cs:60` accepts that Community licence programmatically at mesh-builder time, and `MeshWeaver.Blazor` pulls it transitively, so **QuestPDF ships in every deployed portal image**.

The user's decision (2026-08-11) is to dereference it — not to buy a licence and not to swap in another restrictively-licensed package. PDF rendering moves to the browser via the existing `Pixel/HeadlessChromiumPdfRenderer.cs` (Playwright, MIT).

## Footprint (verified on main)
- `src/MeshWeaver.Markdown.Export/Pdf/PdfDocumentRenderer.cs` — 368 lines, 100% of the QuestPDF rendering, the only file with `using QuestPDF`.
- `src/MeshWeaver.Markdown.Export/Configuration/MarkdownExportExtensions.cs:60` — `Settings.License = LicenseType.Community`.
- `src/MeshWeaver.Markdown.Export/MeshWeaver.Markdown.Export.csproj:8` — the only `PackageReference`.
- Tests using it: `PdfHeaderLogoTests.cs`, `RendererOutputTests.cs`.
- `ExportFidelity.cs`, `ExportFormat.cs`, `Branding/ExportTemplateResolver.cs` mention it in **comments only** — no compile dependency.

## Why it is not a one-line deletion
Fork point `Templates/ExportPdf.csx:126`: Pixel fidelity goes to Chromium (**Deck only**); `ExportPdf.csx:193-196` warns and **falls back to QuestPDF for every non-Deck node**. Per `PixelFaithfulExport.md:70` the Chromium path today has **no cover page, no TOC, no running header/footer and no page-break rules** — those exist only in `PdfDocumentRenderer.cs`. `README.md:23` also states headless Chromium is deliberately not shipped in the portal image.

So the deletion must land **with** the document-export migration onto Chromium, otherwise it is a functional regression rather than a cleanup.

## Done when
- [ ] Markdown→PDF, cover page, TOC and running header/footer render through the browser path
- [ ] `PdfDocumentRenderer.cs`, the licence line and the `PackageReference` are deleted
[...]

PR #1290, opened 2026-08-12
feat(image): ship a headless Chromium in the portal-ai base image

rbuergi:
Step 1 of 2 for #1230. **Base image only** — no renderer changes, no package removal. Those follow in a second PR so this one can be judged on its own cost.

## Why the browser has to be in the image

MeshWeaver's PDF export prints with a headless browser. Today that serves only the **opt-in pixel-faithful deck export**, which resolves an *already-installed* browser — so on the deployed image the feature is simply never available. #1230 makes the browser the renderer for **every** PDF, replacing QuestPDF, whose Community tier is free only below a revenue threshold. Either way, a portal whose PDF export depends on an operator having installed Chrome by hand has no PDF export.

## Measured size delta

Both figures on `linux/arm64`, same method, both images present:

| | before | after | delta |
|---|---|---|---|
| On disk (`docker images`) | 1.64 GB | 2.57 GB | **+0.93 GB** |
| Manifest — what a node pulls | 440 MB | 682 MB | **+242 MB** |

The added layer is **689 MB uncompressed** (`docker history`). Inside it: the browser 344 MB, the font set 91 MB, the shared libraries Chromium links (mesa, X, GTK stack) ~206 MB, the rest apt metadata. The `fonts-local.conf` layer is 16 kB.

## Multi-arch evidence

Both architectures were built and the resulting browser executed:

| | binary | `--version` |
|---|---|---|
| `linux/arm64` (native) | `chromium_headless_shell-1234/chrome-linux/headless_shell` | `Chromium 151.0.7922.34` |
| `linux/amd64` (QEMU) | `chromium_headless_shell-1234/chrome-headless-shell-linux64/chrome-headless-shell` | `Google Chrome for Testing 151.0.7922.34` |

Same Chromium version on both. The Dockerfile ends the install step with `/usr/bin/chromium --version`, so an architecture that resolves no runnable browser **fails the build** instead of shipping a base image whose PDF export is broken.

## Three findings that shaped this, each measured rather than assumed

**1. `apt-get install chromium` is a trap on this base.** `mcr.microsoft.com/dotnet/aspnet:10.0` is Ubuntu 24.04, where `chromium-browser` is version `2:1snap1-0ubuntu2` — a snap shim that `PreDepends: snapd` and installs no browser at all in a container. Noble has no plain `chromium` deb. Debian has one, but .NET 10 GA publishes no Debian tag (`10.0-trixie-slim` is preview-only), so changing the base is not on the table. Google Chrome and Microsoft Edge have **no Linux arm64 build**, so either would have produced a green amd64 leg and an arm64 leg with no browser. Playwright's Chromium is the one build published for both.

**2. The full desktop browser never finishes a print in a container** — so `--only-shell` is not a size trim, it is the thing that works. Measured in this exact image: the full browser blocks in its desktop start-up path (dbus, UPower, GSettings schemas, the component updater, none of which exist there) — **150 s in, no PDF, process still alive**. Through the renderer that is a hung export failing on its two-minute timeout. The headless shell has none of that machinery, printed the same document in about two seconds, and is 340 MB instead of 641 MB. Same Blink engine, same version.

**3. The binary is named differently per architecture** — `headless_shell` in Playwright's arm64 Chromium build, `chrome-headless-shell` in the Chrome-for-Testing build it uses on amd64. Matching only the first name built a perfectly green arm64 leg and an amd64 leg with no browser. The `--version` guard caught exactly this during development; the build now matches both names and prints the tree before failing if neither is there.

## `fonts-local.conf` — without it the image renders nothing in bold

Playwright's dependency list installs a broad CJK/Thai/Cyrillic font set so any content prints. One of them, WenQuanYi Zen Hei, claims the generic `sans-serif` and `system-ui` families in its own fontconfig snippet **and ships a Regular face only**. Before this file:

```
$ fc-match "sans-serif:bold"
wqy-zenhei.ttc: "WenQuanYi Zen Hei" "Regular"
```

so `font-weight: 700` matched a family with no bold, Chromium did not synthesise one, and every heading, table header and bold run printed at regular weight — silently. `prefer` puts Liberation Sans at the front of the match list without removing anything, so per-glyph fallback still reaches the CJK fonts. It is its own layer, deliberately after the browser install, so a font tweak never invalidates a 20-minute multi-arch browser layer.

## 🚨 Merging this does NOT put Chromium in production

`main-cd.yml` does not rebuild the base image — it layers the app image on `memex-portal-ai-base:latest` **already in ACR**. So after this merges, someone must run **`base-image-acr.yml`** (workflow_dispatch) to build and push the new multi-arch base. Until that runs, the deployed base has no browser.

**That ordering is load-bearing for PR 2**: PR 2 makes every PDF export depend on the browser. If PR 2's image rolls out on a base without Chromium, PDF export fails in production. Sequence: merge this → run `base-image-acr.yml` → verify → merge PR 2.

One thing to watch on that first rebuild: `base-image-acr.yml` has `timeout-minutes: 60` and the job now additionally installs ~60 OS packages and downloads a 111 MB browser **per architecture**, with the arm64 leg under QEMU. I have not raised the bound because I have no measurement of it on a GitHub runner and would rather not widen a limit speculatively — but if that first run times out, that is why.

Refs #1230


rbuergi:
Both review points addressed.

**`find … | head -n 1`** → now `-executable … -print -quit`, so it stops at the first hit instead of walking the whole ~350 MB browser tree (which is emulated on one of the two legs), and requires the match to actually be a program. Verified against the real tree in the built image — resolves `/opt/ms-playwright/chromium_headless_shell-1234/chrome-linux/headless_shell` — and a full arm64 rebuild is running to confirm the layer end to end.

**Docs still saying the browser is not shipped** — correct catch, and it reached further than the two files quoted: the XML doc on `PixelRenderingOptions` said it too. All three are fixed in #1291 (`PixelRenderingOptions`, `MeshWeaver.Markdown.Export/README.md`, `Doc/Architecture/PixelFaithfulExport`), which is where the statement actually becomes false — this PR only *adds* the browser, while #1291 is what makes it the renderer for every PDF. They are merged back to back in that order, so the window where the docs lag is the gap between the two merges. The updated text distinguishes the flavours explicitly: `portal-ai` ships the browser, the lean `portal` image does not.

PR #1291, opened 2026-08-12
feat(export): print PDFs with the browser and delete QuestPDF

### LINKED ISSUE #1230: Remove QuestPDF — its Community licence is pay-to-use above a revenue threshold

rbuergi:
Step 2 of 2 for #1230. **Depends on #1290** (the base image that ships the browser).

QuestPDF's Community tier is free only below a revenue threshold, above which a paid commercial licence is required. It was the *only* PDF renderer, which is why removal needed a replacement first. With the browser now in the image, the PDF is printed from HTML + CSS Paged Media by the engine the portal already runs for pixel-faithful deck export — two PDF back ends become one.

The `Document` model is untouched and still feeds DOCX. Only the last step changes:

```
Markdig AST ─▶ Document model ─▶ DocumentPrintComposer ─▶ IPixelPdfRenderer ─▶ PDF
               (unchanged,        (new: model → print      (existing browser leaf,
                shared with DOCX)  HTML + CSS Paged Media)   IIoPool Process pool)
```

## Parity, verified by rendering PDFs and reading the pages back

Not by asserting on markup. Each row below was checked on a real printed PDF — text extracted per page, and pages rasterised and inspected visually.

| Feature | Status | Evidence |
|---|---|---|
| **Cover page** | ✅ full | Page 1 carries logo, title in brand primary, tagline, rule, name, website, date — and no header, no footer, no page number |
| **Table of contents** | ⚠️ **partial** | Entries, indentation by level and working internal links all present. **Page numbers are gone** — see below |
| **Running header** | ✅ full | Brand text left, logo right, continuous rule across the content width, on every page but the cover |
| **Running footer** | ✅ full | Brand text left, `N / M` right, on every page but the cover; present even when the brand supplies no footer text, as before |
| **Page-break rules** | ✅ full | `PageBreakBeforeH1` / `H2` / `PageBreakBetweenChildren` / explicit `\newpage` all start a new page |
| **A4 / landscape** | ✅ full | Page box 595×842 pt portrait, wider than tall in landscape |
| **Header suppression** | ✅ full | No header text and no logo ⇒ no band and no rule at all, matching the old gate |
| Table header repeat | 🎁 better | A table spanning pages now repeats its header row; the document model could not |
| Heading orphans | 🎁 better | `break-after: avoid` stops a heading stranded at the foot of a page |

Rendered locally against the image from #1290 and inspected page by page (cover; contents with header/footer/`2 / 5`; body pages; plus a landscape / no-cover / no-TOC / no-header variant to prove the named-page split and the header-suppression gate are independent). The same assertions are pinned in `RendererOutputTests` and `PdfHeaderLogoTests`, which read the printed pages back with PdfPig.

### ⚠️ The one regression: no page numbers in the contents list

Entries are real PDF links — clicking one jumps to the section — but the `..... 12` column is gone.

Reading a target's page from CSS requires `target-counter()` from CSS Generated Content for Paged Media, and **Chromium implements no part of it**. I verified this directly against the browser this image ships, alongside the features that *do* work and that the rest of the furniture is built on: `@page` margin boxes, named pages, `counter(page)` / `counter(pages)`.

The alternative is a two-pass print — print once, read back from the PDF's link annotations which page each anchor landed on, inject the numbers, print again. Rejected deliberately: it doubles the cost of every export with a contents list, and its failure mode is *silently wrong page numbers* when the inserted digits reflow the list onto another page. An honestly absent column beats a confidently wrong one. Called out in the What's New entry so users are not left guessing.

## Two defects found by looking at the output, not at the assertions

**1. Every export would have died in one second in production.** Chromium refuses to start as root without `--no-sandbox`, and the portal container runs as uid 0:

```
ERROR:zygote_host_impl_linux.cc:101] Running as root without --no-sandbox is not supported.
```

`NoSandbox` defaults to `false`, so nothing would have printed. The renderer now passes the flag when the sandbox is unavailable *by construction* — `geteuid() == 0`, read from the OS rather than inferred from a user name so a container with no passwd entry answers correctly. This does not override an operator decision: where a decision exists (a non-root deployment, where the namespace sandbox works) `NoSandbox` is still theirs. Where the browser cannot use its sandbox at all, the flag is the only way to run, and the isolation that holds is the layer above — the print document's `default-src 'none'` CSP plus the process-level network denial, both already tested with the other neutralised.

**2. Chained `string.Replace` re-scans substituted text.** A document titled with a placeholder token had the next value spliced into its `<title>`; a brand header containing one could pull the document body into the stylesheet. The composer now fills each template in **one pass over the template**, so substituted content is inert by construction. Pinned by a test.

## Reuse, not re-derivation

Markup is built with `Html/MarkupNode` from #1234 — "the ONE place that turns a tree into a string of HTML", escaping text and attribute values on the single path out. There is no HTML-string building anywhere in the path, and the page furniture lives in real `DocumentPrint.{html,css}` template files, the same convention `SlidePrint` already uses.

I checked whether an existing translation already covered this and it does not: `Html/` goes control→markup (`AreaMarkupRenderer`), markup→model (`MarkupToDocument`) and markdown→email-HTML (`DocumentHtmlComposer`). **Model→markup did not exist.** Routing PDF through `DocumentHtmlComposer` instead would have meant re-implementing the page-break rules, TOC collection and Mermaid/Math SVG substitution that currently live in `DocumentBuilder` and are shared with DOCX — more duplication, not less. Embedded layout areas need no new code at all: #1234 already lands them in the `Document` model via `MarkupToDocument`, so they arrive as ordinary elements.

## Threading

`PdfDocumentRenderer.Render` returns a cold `IObservable<byte[]>` — no `Task` in the signature, no `Observable.FromAsync`, no `Task.Run`. The browser is a `Process` leaf reached through `IIoPool.InvokeBlocking` on the `Process` pool, unchanged from the pixel path.

## CI

The PDF leg now needs a browser, so CI provisions **the exact build the image ships** (same Playwright version, same `--only-shell` Chromium) as an explicit infrastructure step beside the Testcontainers pre-pull. No `continue-on-error` and no input-shaped `if:` — a failed install reddens the shard rather than leaving the tests quietly asserting the "no browser here" branch and calling it green.

## Licence gate

The `questpdf` entry is deleted from `EXCEPTIONS` with the package, exactly as that entry instructed. Verified locally after a solution restore:

```
431 packages (391 shipping, 40 declared-unused), 0 violations
Licence gate passed: 391 shipping packages checked
```

## Verified locally

`dotnet build -c Release -warnaserror` clean (`0 Error(s)`, `0 Warning(s)`) per touched project individually: Markdown.Export, Markdown.Export.Test, MeshWeaver.Blazor, Memex.Portal.Shared, Hosting.Monolith.Test, Hosting.Orleans.Test, Security.Test, Documentation.Test.

Tests, all green: Markdown.Export.Test **99**; the seven export-script integration tests across Monolith (`ExportDocumentScriptRelay`, `DocumentExportLayoutArea`, `DocumentExportAreaAccess`, `DeckExportScriptRelay`, `DeckPixelExportScriptRelay`, `SendDocumentToContacts`, `DeckAssetInlining`) **15**; Orleans `OrleansMarkdownExportTest` **5**; Security `ScriptTemplateExecuteAccessTest` **3**; `DocumentationLinkIntegrityTest` **1**.

## 🚨 Merge ordering

`main-cd.yml` layers the app image on `memex-portal-ai-base:latest` **already in ACR** — it does not rebuild the base. So after #1290 merges, **`base-image-acr.yml` must be run** to publish the new base. If this PR's image rolls out on a base without Chromium, PDF export fails in production.

Sequence: merge #1290 → run `base-image-acr.yml` → verify the base carries `/usr/bin/chromium` → merge this.

## What's New

`Category: Fix` — the output changes visibly (contents page numbers gone, repeated table headers), so it is an entry a user would otherwise find missing.

Closes #1230


rbuergi:
#1290 is merged, so the Dockerfile now carries the browser — but **the base image in ACR does not yet**. `main-cd.yml` layers the app image on `memex-portal-ai-base:latest` already in ACR and never rebuilds it; the rebuild is `base-image-acr.yml` (workflow_dispatch).

So before this merges into a deployed image: run **`gh workflow run base-image-acr.yml --ref main`**, then confirm the new base actually carries the browser, e.g.

```
docker run --rm meshweaver.azurecr.io/memex-portal-ai-base:latest /usr/bin/chromium --version
# expect: Chromium 151.0.7922.34  (arm64) / Google Chrome for Testing 151.0.7922.34 (amd64)
```

If this PR's image rolls out on a base without Chromium, every PDF export fails — the renderer refuses loudly rather than returning a malformed file, so it would be visible immediately, but it would still be an outage of the feature.

Watch that first base rebuild's duration: the job installs ~60 OS packages and a 111 MB browser **per architecture**, the arm64 leg under QEMU, against a `timeout-minutes: 60` that was sized for the old (node-only) content. I deliberately did not raise it without a measurement.

PR #1294, opened 2026-08-12
fix(delete): name the STAGE a delete timeout died in, and which descendant went silent

rbuergi:
Refs #1198. **Diagnostics only — this does NOT close the issue**, and it corrects the issue's
central piece of evidence.

## `partial-deleted=0` was never evidence of anything

The issue reads `[DeleteNode] timeout path=KmuBasics/Buchungsjournal partial-deleted=0` as "the
operation made zero progress", and the triage used that to narrow the failure to the pre-commit
stages. That reading does not hold. The commit is wrapped in `.Timeout(opts.Timeout)`, a timeout
**unsubscribes**, and the deleted-path list is attached only to the exception
`HierarchicalPathDeletion` raises *itself*. The freshly-constructed `TimeoutException` carries no
`Data`, so `partial-deleted` printed `0` for **every** timeout regardless of how much of the
subtree was already gone. The commit stage was never excluded.

## What the log could not say

`HandleDeleteNodeRequest` bounds **six** independent stages with the same
`MeshOperationOptions.Timeout` — and three of them had no bound at all, so a wedge there produced
no `[DeleteNode]` line whatsoever (only the caller's request timed out). Every bounded one threw
`new TimeoutException()` → `"The operation has timed out."` → one indistinguishable log line.

## What changed

- **Every stage is bounded and named**: `read-root`, `check-permission`, `validate-root`,
  `collect-paths`, `pre-validate-descendants`, `commit`. The three previously-unbounded stages
  (root read, permission fold, root validators) now fail loudly *inside the budget the operation
  already declares* instead of hanging silently. **No budget was raised.**
- **The fan-out names who went silent.** `PreValidateDescendantsObs` posts `ValidateDeleteRequest`
  at *every* descendant and waits for *all* of them under **one** timeout — so a single
  unresponsive per-node hub refuses the whole delete. It now tracks which paths have answered and,
  on timeout, reports the ones that did not (first 10 in the message, first 20 in the log line).
- **`partial-deleted` is truthful.** A sink records each path as its delete commits, so the commit
  timeout carries the real set. The response's `AffectedPaths` is fixed by the same change:
  "nothing happened, safe to retry" and "the subtree is half gone" are finally distinguishable.
- **The caller sees it too** — the response `Error` carries the stage and the detail, not just the
  log.

Real output from the two new tests:

```
Delete of 'TestData/prevalidate' exceeded 15s timeout in stage 'pre-validate-descendants':
  [DeleteNode:pre-validate-descendants] 1 of 1 descendant(s) of 'TestData/prevalidate' did not
  answer ValidateDeleteRequest within 15s, so the delete was refused with the subtree untouched:
  TestData/prevalidate/child-silent

Delete of 'TestData/commitstall' exceeded 15s timeout in stage 'commit':
  [DeleteNode:commit] the bottom-up delete of 'TestData/commitstall' did not drain within 15s
  — 1 of 3 planned path(s) were already removed from storage
affected: TestData/commitstall/healthy          ← was empty, and partial-deleted=0
```

## Why the root cause is not fixed here

Two stages fit the single line we have, and with `partial-deleted` proven uninformative the commit
stage is a third. Choosing between them from one occurrence would be a guess, and the plausible
fixes differ in kind (a per-leaf report for an unresponsive descendant vs. a stalled storage
enumeration vs. a leaf hub wedged mid-commit). The next occurrence now names the stage and the
silent hub — exactly the evidence that decision needs. **#1198 stays open** with that note.

Deliberately NOT done: no timeout raised, no validation made optional, no retry, no watchdog.

## Test

`test/MeshWeaver.Hosting.Monolith.Test/DeleteTimeoutStageDiagnosticsTest.cs` — the stall is a real
registered `INodeValidator` that never emits for one node, which is the framework-legal way to make
a single per-node hub go silent (validators run through `Concat`, so that hub posts no
`ValidateDeleteResponse` at all).

1. **Pre-flight fan-out** → the response names `pre-validate-descendants` **and** the silent
   descendant's full path.
2. **Commit** → a healthy sibling is genuinely deleted before the stalled leaf holds up the
   cascade, and the failure reports that sibling in `AffectedPaths` instead of the old zero.

Both bounded by the mesh operation timeout; no sleeps, no polling. Green locally
(`-c Release -warnaserror`, 2/2 passed, 30 s).

### DIFF
--- a/src/MeshWeaver.Documentation/Data/WhatsNew/2026-08-13-a-delete-that-times-out-now-says-what-it-was-waiting-for.md
+++ b/src/MeshWeaver.Documentation/Data/WhatsNew/2026-08-13-a-delete-that-times-out-now-says-what-it-was-waiting-for.md
@@ -1,0 +1,26 @@
+---
+Name: A delete that times out now says what it was waiting for
+Category: Fix
+Description: When deleting a node runs out of time, the error now names the step it was stuck on, the child that never answered, and everything it had already removed.
+Icon: Sparkle
+Order: -20260813
+---
+
+# A delete that times out now says what it was waiting for
+
+Deleting a node with a large subtree is a multi-step operation: the node is read, your permission
+is checked, its validators run, the subtree is listed, every child is asked whether it may be
+deleted, and only then is anything removed. Each of those steps has the same time budget — so when
+a delete ran out of time, all six looked identical from the outside. The message said only that the
+delete had timed out, which left no way to tell a slow database from a single child that had stopped
+responding.
+
+It now names the step. If the delete was waiting on the children, it also names the ones that never
+answered — that step asks every child at once and waits for all of them, so one unresponsive child
+stops the whole delete, and knowing which one is the entire diagnosis.
+
+The count of what had already been deleted was worse than unhelpful: it always read zero after a
+timeout, whatever had actually been removed, because the parts already deleted were discarded along
+with the timed-out step. A delete that gets partway through now reports exactly which nodes it
+removed, so "nothing happened, safe to retry" and "the subtree is half gone" are finally
+distinguishable.
--- a/src/MeshWeaver.Mesh.Contract/MeshExtensions.cs
+++ b/src/MeshWeaver.Mesh.Contract/MeshExtensions.cs
@@ -2045,6 +2045,56 @@ public static class MeshExtensions
     }
 
     /// <summary>
+    /// <see cref="Exception.Data"/> key naming the delete STAGE a <see cref="TimeoutException"/>
+    /// came out of. Issue #1198: the delete pipeline bounds SIX independent stages with the same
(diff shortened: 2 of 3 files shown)


rbuergi:
Copilot review addressed:

- **"FIVE independent stages"** — real catch, fixed in 76dc9f7. The list (and the production code) has six: `read-root`, `check-permission`, `validate-root`, `collect-paths`, `pre-validate-descendants`, `commit`.
- **`ITestOutputHelper` needs `Xunit.Abstractions`** — not applicable on xUnit **v3**, where the type moved into the `Xunit` namespace (which the file already imports). `dotnet build -c Release -warnaserror` on this project is clean (`0 Error(s)`), and the two tests run green, so the compile concern does not hold.

PR #1302, opened 2026-08-12
Node-native course pages get the whole course as their side menu + progress in the learner's home

rbuergi:
## What

Three pieces, one experience:

1. **`EducationNavigationProvider`** (MeshWeaver.Graph) — an `INodeNavigationProvider` that claims pages of node-native (plugin) courses by SHAPE: any subtree holding `…/Lesson`, `…/Module` or exercise-typed nodes. It supplies the WHOLE course as the left menu: every lesson as a group ordered by `Order`, reading pages first / exercises last, `Source`–`Test` internals excluded, exactly one entry marked current at any depth. A page inside the learner's own copy (`{viewer}/{course}/…`) indexes **the copy** — the menu navigates the material that is theirs to edit and run. Registered in `MemexConfiguration` next to `AddCourses()`.

2. **`CourseProgress`** — progress captured in the learner's home as they read: one record per course at `{viewer}/_Progress/{course}` (`lastPath`, `lastVisitedAt`, first-visit stamp per lesson; CENTRAL paths so copy + template share one record). Written fire-and-forget off the render path under the System identity (deferred-write ambient identity is a race), idempotent by construction — `MergeVisit` returns null when the record already says everything, so render-driven capture cannot storm, and the ✓-decoration re-render it triggers converges. The menu marks visited lessons with a leading `✓`.

3. **`NodeNavigation` standalone area** — the supplied menu registered on every node, so layouts that are not the markdown Overview (the Edu plugin's own lesson/module areas — follow-up PR in MeshWeaver.Plugins) can embed the same course index by name. Null when no provider claims the page, never an error box.

## Why

"I asked already many times to have ALL the lessons included in the side menu, not just the one" — the 2026-08-08 fix (8d54b26ed, "the course index showed one branch, never the course") repaired this for core-Courses types only; the provider seam was never implemented for the Edu-plugin course shape that every live course actually uses. Plus: "highlight where we are, and capture progress in user home."

## Deliberately NOT claimed

Core's own `Course`/`Module` node types (no module prefix) — those belong to `CourseNavigationProvider`; claiming both would race two menus for one page (pinned in tests).

## Tests

`EducationNavigationProviderTest` — 12 pure tests: whole-course structure + ordering, Source/Exercise-folder handling, index invariance under navigation (only the marker moves), deepest-ancestor marking, learner-copy indexing, docs-space + core-Courses decline, `MergeVisit` idempotence + shape-tolerant reads (typed/JsonElement/JsonObject), ✓ decoration incl. copy→central mapping. Full Graph suite green (1044).

### DIFF
--- a/memex/Memex.Portal.Shared/MemexConfiguration.cs
+++ b/memex/Memex.Portal.Shared/MemexConfiguration.cs
@@ -743,6 +743,9 @@ public static class MemexConfiguration
                 // Interactive courses: Course/Module/Exercise/ExerciseAttempt
                 // node types + the stream-update validation control plane.
                 .AddCourses()
+                // Whole-course left-hand index for NODE-NATIVE courses (Edu/Lesson-shaped
+                // plugin partitions) — the same menu AddCourses gives its own course type.
+                .AddEducationNavigation()
                 .AddPortalType()
                 .AddAI(serveFromPartition);
 
--- a/src/MeshWeaver.Graph/Education/CourseProgress.cs
+++ b/src/MeshWeaver.Graph/Education/CourseProgress.cs
@@ -1,197 +1,117 @@
 using System.Collections.Generic;
 using System.Linq;
-using System.Reactive.Disposables;
-using System.Reactive.Linq;
-using System.Text.Json;
 using System.Text.Json.Nodes;
 using MeshWeaver.Mesh;
-using MeshWeaver.Mesh.Services;
-using MeshWeaver.Messaging;
-using Microsoft.Extensions.DependencyInjection;
-using Microsoft.Extensions.Logging;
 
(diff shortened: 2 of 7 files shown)


rbuergi:
Reworked after discovering MeshWeaver.Plugins#428 (Resume beside Start) in flight: its resume record is a blind later-visit-wins upsert at `{viewer}/_Progress/{course}`, and this PR's first commit put a merge-written record at the same path — two writers clobbering each other. Visited history now accumulates as CHILD markers at `{viewer}/_Progress/{course}/Visited/{lesson}` (idempotent no-read upserts, same render-turn-identity + 3s-dwell discipline as #428's writer), so the two features compose without ever touching the same node.

PR #1307, opened 2026-08-12
Make the content-import identity declarable — an ambient scope cannot reach every pipeline stage

rbuergi:
Follow-up to #1263, which was merged, deployed (ci.3152 / de54cd0) and did **not** fix
Systemorph/MeshWeaver.Reinsurance#46 — the import re-ran and produced the byte-identical failure:

```
412 writes applied.
409 write(s) failed — first: files → ReinsuranceDemo/Cedents/Adriatica:
AccessContext must never be null for an application post — no identity, no delivery.
hub=ReinsuranceDemo/Setup, message=SyncContentFilesRequest
```

## What #1263 got right, and where it aimed one level too shallow

#1263 correctly stopped the builders reading the ambient at *Subscribe*. But eager capture means
"at `Post()`" — and the importer was calling `Post()` from inside the **second stage** of

```csharp
Observable.Using(() => access.ImpersonateAsSystem(),
                 _ => nodeWrites.Concat(fileWrites))
```

`Observable.Using` opens the AsyncLocal on the **subscribing thread**, which covers only what runs
synchronously there — the first stage. A `Concat`'s second stage is subscribed from the first's
*completion*, on a pool thread the scope never reached, so there is nothing for an eager capture to
capture.

Instrumenting that exact pipeline (probe now folded into the test as an assertion):

| stage | built | thread | ambient | stamped |
|---|---|---|---|---|
| `nodeWrites` (`CreateOrUpdateNode`) | synchronously at the outer Subscribe, **in** scope | 13 | `content-importer` | `content-importer` ✅ |
| `fileWrites` (`SyncContentFiles`) | from stage one's completion, **outside** it | 19 | `(null)` | unstamped → **refused** ❌ |

So the 412/409 split was never node-vs-file: it was **which stage built the post**, which is why it
was total on both sides rather than intermittent. One correction to the going hypothesis: it is
**not** the `Observable.Defer` either — the probe shows the enclosing LINQ projection already ran at
`ambient=(null)` on thread 19, so hoisting `Post()` out of the `Defer` would not have helped.

## The change

No capture strategy can fix this, because the identity genuinely is not on that thread. Make the
correct thing **expressible** instead: both content-import builders accept the identity as a
**value**.

```csharp
hub.SyncContentFiles(path).To("content").ImpersonateAsSystem().Post();   // platform infrastructure
hub.SyncContentFiles(path).To("content").WithAccessContext(user).Post(); // a specific caller
```

A declared identity rides the delivery, so no thread hop — and no later re-ordering of a Subscribe —
can lose it.

**Nothing is defaulted.** With no identity declared and no ambient one, the post stays unstamped and
still **fails closed**. Re-adding a fallback is exactly what the 2026-05-21 hub-self deletion
removed, and it masked a real prod bug.

Supporting surface:

- `WellKnownUsers.SystemContext` — the one `AccessContext` value for "this is platform
  infrastructure", for callers that must **stamp** rather than **scope**.
- `CarryAccessContext(source, services, captured)` — restores an **explicit** identity around each
  emission. Deliberately kept separate from the ambient overload, which re-reads `Context` **only**:
  feeding it the post's `Context ?? CircuitContext` capture would synthesise a Blazor circuit
  identity into background-Subscribe callbacks (the 757d2a296 anti-pattern that helper's own xmldoc
  warns about). `CarryPostIdentity` picks the route; existing ambient behaviour is byte-identical.

## The test

`ContentImportAccessContextTest.DeclaredIdentity_SurvivesAPipelineStageTheAmbientScopeNeverReaches`
reproduces the #46 pipeline and asserts the stamped `ObjectId` of **both** stages
(`nodeStamp=content-importer`, `fileStamp=pack-attacher`, using two distinct identities so a match
cannot be coincidence).

Two things make it real rather than decorative:

- **It asserts on identity, never on "it landed".** The xUnit host sets a standing `hostIdentity`, so
  `CircuitContext` resolves to the DevLogin admin on every thread — unfixed, this pipeline still
  wrote all six files, as `Roland`, and every landed/`Success` assertion passed. In production
  `CircuitContext` is null on a hub thread, which is why prod refuses and a naive test would not.
  Only the stamped ObjectId separates the two. (Measured pre-fix on this very test:
  `ok=6/6 nodeStamp=content-importer fileStamp=Roland`.)
- **It guards its own premise.** `secondStageAmbient.Should().NotBe(Importer.ObjectId)` fails loudly
  if the second stage ever *can* see the enclosing scope, so the test can never quietly decay into a
  single-threaded tautology that pins nothing.

## Verification

- `dotnet build test/MeshWeaver.Content.Test -c Release -warnaserror` — clean, on current main
- All 3 tests in `ContentImportAccessContextTest` pass (the two #1263 pins plus the new one)
- Branch merged with `origin/main` (743995dcb) before pushing

## Companion

Systemorph/MeshWeaver.Reinsurance#55 fixes the live import and closes
Systemorph/MeshWeaver.Reinsurance#46. It does **not** depend on this PR: the demo's CI pins the
platform image by digest (`MW_IMAGE_DIGEST`), so `compile-check.py` cannot see this API yet. It
fixes the import by building every write inside one real `using` on one thread, and names
`.ImpersonateAsSystem()` as the follow-up once the pin advances.

No What's New entry: this PR adds framework surface and a regression test with no behaviour change
to any existing call path — the user-visible fix ships in the Reinsurance repo.


rbuergi:
### The shard-2 red on `5ecbc20` was not this PR

`MeshWeaver.FutuRe.Test.FutuReAnalysisTest.EuropeRe_AnnualReport_EmbeddedCharts_ShouldRenderViaPathResolution` —
the hub never activated:

```
Failed to create hosted hub for address FutuRe/EuropeRe/Analysis
Expected the observable to emit a value matching the predicate within 50s … Last of 2 emission(s) was: <null>
```

That is the known FutuRe/EuropeRe activation-and-teardown cluster (#981, #826), and nothing in this
diff can reach it. The one way an additive change like this could have perturbed an unrelated test
is by shifting **overload resolution** on the new
`CarryAccessContext(source, services, AccessContext?)`, so I checked every call site rather than
assuming: all are either 1-arg (`serviceProvider` / `accessService`) or pass `restoreNullCapture:`
**by name**. None can bind to the new overload. The other two additions are a new static property
and two new builder methods, none of which existed to be called.

Per AGENTS.md ("if CI fails on something your diff does not touch, merge main before you investigate
anything") the branch is now merged with `main` (`8b16e25`, it was 6 behind) and re-running — a new
input, not a bare retry.

Re-verified locally on the merged tree:

- `dotnet build test/MeshWeaver.Content.Test -c Release -warnaserror` — `0 Error(s)`
- all 3 `ContentImportAccessContextTest` tests pass:
  `nodeStamp=content-importer  fileStamp=pack-attacher  stage2Ambients=[(null) ×4, pack-attacher ×2]`

That last line is the whole finding in one row: stage two's ambient is whatever the previous
delivery happened to leave behind — null before any post has landed — and never once the enclosing
scope's `content-importer`.

Issue #1297, opened 2026-08-12
25 of 72 control views silently drop Style/Class — the API lets you set what the view ignores


rbuergi:
`Style` and `Class` are public API on `UiControl` (`src/MeshWeaver.Layout/UiControl.cs:16,52`), and `WithStyle(...)` is used widely — but **25 of 72 control views never apply them**. An author sets a style, the view drops it, and nothing reports anything.

## Why this is filed rather than fixed

PR #1288 fixed `HtmlView`, the worst instance: `Controls.Html(x).WithStyle("width:100%")` was silently a no-op, which is why three Reinsurance charts rendered at ~6% of column width. The audit below came out of that work and is deliberately **not** included in it — a 25-view sweep needs its own review.

## The audit (72 control views under `src/MeshWeaver.Blazor/`)

| Bucket | Count |
|---|---|
| 🔴 **Silently drops** — 17 drop both, 8 drop `Class` only | **25** |
| 🟡 Unclear (`NamedAreaView`) | 1 |
| 🟢 Applies both | 39 |
| ⚪ Deliberately N/A (no meaningful box) | 7 |

Plus three outside the buckets: **`SearchBoxView`, `MeshSearchView`, `Monaco/CodeEditorView`** are registered as control views in `BlazorViewRegistry` but derive from plain `ComponentBase`, so `Style`/`Class` are never *bound* at all.

Standouts:

- **The 8 `Class`-only drops** look like plain omissions rather than design — they bind `Style` and forget `Class`.
- **`MeshNodePickerView`** is a `FormComponentBase` and drops `Width`/`Height` too.

## What "fixing" should and should not mean

🚨 **Not a blanket wrapper.** #1288 established why, with a case that would have shipped undetected: `samples/Graph/Data/ACME/Project/Todo/Source/TodoLayoutAreas.cs:117` sets `flex: 1` on its fragment's own root **because that root is the direct flex child**. A wrapper makes it inert — and that file is **in-mesh source, so no build and no test would catch it**. Each of the 25 needs its own judgement about where the style belongs.

Also from #1288, worth reusing: gate on the **declaration** (`ViewModel.Style`), not the resolved value. Either can be a `JsonPointerReference` that lands a frame later, so gating on the value renders bare and then restructures the DOM mid-flight.

## The general defect, which outlives any individual view

[...]

PR #1323, opened 2026-08-12
fix: the eight views that bound Style and dropped Class (#1297, part 1)

rbuergi:
Part 1 of the #1297 sweep — the **8 `Class`-only drops**, landed on their own because they are the
subset with no design question in them.

## Why these eight are the cheap subset

The 25 views in #1297's audit split cleanly. Seventeen drop **both** `Style` and `Class`, and each of
those needs a judgement about *where* the box goes — the case #1288 caught (`TodoLayoutAreas.cs:117`
puts `flex: 1` on the fragment's own root **because that root is the direct flex child**) proves a
blanket wrapper is wrong, and that file is in-mesh source no build would catch.

These eight are not that. Every one of them already writes `Style="@Style"` onto **a real root element
of its own** and simply never wrote the class beside it. So there is no "where should it go" question
and no wrapper to introduce: **the element that already carries the author's Style is by definition the
element that must carry the author's Class.** No new element is created anywhere in this PR.

## Per view — what its root already had, and what it now carries

| View | Root before | Fix |
|---|---|---|
| `EditorView` | `<div style="@Style">` — no class attribute at all | `class="@Class"` on that div |
| `NumberFieldView` | wrapper `<div style="…@ComputedStyle">` — no class attribute; the inner input's `class` is a local appearance token | `class="@Class"` on the wrapper, i.e. the same element as `ComputedStyle` |
| `Label` | `FluentLabel Style="@Style"` ×2 branches; `Class` is a parameter it accepts | `Class="@Class"` on both branches |
| `TabsView` | `FluentTabs … Style="@Style"`; `Class` accepted, not passed | `Class="@Class"` |
| `MarkdownView` | `class="markdown-body"` fixed | `class="markdown-body@(ClassSuffix)"` |
| `MeshNodeCollectionView` | `class="mesh-node-collection"` fixed | `class="mesh-node-collection@(ClassSuffix)"` |
| `LayoutAreaView` | `class="layout-area-container @(loaded…) @(fill…)"` | composed in `@code` as `ContainerClass` |
| `ThreadMessageBubbleView` | `class="thread-msg-bubble @(IsUser…)"` | composed in the code-behind as `BubbleClass` |

Two of them compose in code rather than markup for a mechanical reason worth stating so the next
person doesn't "simplify" it back: **Razor cannot parse two adjacent explicit expressions in one
attribute value** (`@(a)@(b)` → `CS1073: Unexpected token`), and both of those roots end in a
conditional. The literal parts and their spacing are reproduced exactly.

## `ClassSuffix`, and why not just append `" @Class"`

`BlazorView.ClassSuffix` is `" " + Class`, or the **empty string** when the author declared nothing.
Writing `class="markdown-body @Class"` instead would leave a **trailing space in that attribute on
every page in the portal** for the overwhelming majority of controls that set no class. Hence the
suffix, and hence the backwards-compatibility tests assert the **exact attribute**
(`class="markdown-body"`) rather than `Contain`.

⚠️ A trap found on the way, in case it recurs: the natural `class="markdown-body@ClassSuffix"` is
**silently wrong** — Razor's email-address heuristic treats `word@word` as literal text, so it would
have rendered the string `markdown-body@ClassSuffix` into the DOM with no error. The explicit
`@(ClassSuffix)` form is required.

## Gating: deliberately NOT on the declaration here

#1288 gates `HtmlView`'s wrapper on `ViewModel.Style` (the declaration) rather than the resolved value,
because either can be a `JsonPointerReference` landing a frame later and gating on the value would
render bare and then **restructure the DOM** mid-flight.

That reasoning does not transfer to this PR and applying it anyway would be cargo-culting: no element
is created or removed here. A pointer that lands late simply updates an attribute value — exactly what
`Style` already does on these same elements today.

## Tests — verified by falsification

Nine tests added to `ControlStyleRenderingTest`, which renders through the **real** `HtmlRenderer`
against a real monolith mesh (asserting `control.Class` proves nothing — the value was always set
correctly; the markup dropped it).

Reverting the five rendered views and re-running:

```
Failed  MarkdownControl_Class_JoinsTheFixedClassList
Failed  MeshNodeCollectionControl_Class_JoinsTheFixedClassList
Failed  LabelControl_Class_ReachesTheDom
Failed  EditorControl_Class_ReachesTheDom
Failed  TabsControl_Class_ReachesTheDom
Failed! - Failed: 5, Passed: 17, Total: 22
```

- The **five positive** tests fail without the fix → they genuinely pin the bug, one per view.
- The **four backwards-compatibility** guards pass **with and without** it → real guards, not
  tautologies. They assert the unstyled attribute exactly (`class="markdown-body"`,
  `class="mesh-node-collection"`, and `NotContain("class=\"\"")` for the two roots that had no class
  attribute at all), so a trailing space or an empty `class=""` fails the build. `.markdown-body` is
  the hook for the entire markdown stylesheet — a mangled attribute there restyles every rendered
  document in the portal.

One harness addition: `IErrorBoundaryLogger` is now registered as a no-op alongside the existing
`IJSRuntime`/`NavigationManager` stubs, because `EditorView` and `TabsView` re-dispatch through
`DispatchView`, which mounts its per-control `ErrorBoundary`.

**Three of the eight are not rendered**, and the test class doc says so rather than leaving it
implicit: `NumberFieldView` is a form input needing a live pointer stream (the same reason the existing
doc already gives for Checkbox/Switch/Combobox/…), `LayoutAreaView` needs a real area stream, and
`ThreadMessageBubbleView` a live thread node. All three take the identical one-line change to the root
element they already style.

## Verification

| Check | Result |
|---|---|
| `dotnet build src/MeshWeaver.Blazor -c Release -warnaserror` | 0 Warning(s), 0 Error(s) |
| `dotnet build test/MeshWeaver.Hosting.Blazor.Test -c Release -warnaserror` | 0 Warning(s), 0 Error(s) |
| `ControlStyleRenderingTest` (22) | Passed |
| `WhatsNewEntryIntegrityTest` + `DocumentationLinkIntegrityTest` | Passed |

What's New entry added (`Category: Fix`).

## Still open on #1297 after this

- **17 views drop both** — the ones needing the per-view "where does the box go" judgement. Next PR.
- **`MeshNodePickerView`** also drops `Width`/`Height`; it is a `FormComponentBase` whose root ignores
  `ComputedStyle`, so that is the *same* one-line fix, not a separate one — it lands with the 17.
- **`SearchBoxView` / `MeshSearchView` / `Monaco/CodeEditorView`** are a different defect (plain
  `ComponentBase`, so `Style`/`Class` are never *bound*) and want their own issue.

Closes nothing on its own — #1297 stays open until the remaining views land.


rbuergi:
Copilot review — two of three taken, one answered.

**Taken (158d1104c):**

1. **`LayoutAreaView` empty class token.** Correct, and I had already accepted the same argument for `CollaborativeMarkdownView` in part 2 without carrying it back here. The interpolated literal already ended in a space before the empty `fill-area` conditional, so `ClassSuffix`'s own leading space made it a double space plus an empty token. Now joined as a token list, which also drops the pre-existing trailing space in the not-`Top`/not-`Fill` case. A class attribute is a whitespace-separated token list, so neither the old space nor its removal can change a selector match.
2. **The doc example was the dangerous form.** `class="markdown-body@ClassSuffix"` is exactly what Razor's email-address heuristic emits as **literal text with no error** — I hit it while writing this PR. Copying the doc would have reintroduced the bug. It now shows `@(ClassSuffix)` and says why in one line.

**Answered, not changed:**

3. The claim that `Contain("class=\"markdown-body\"")` would still pass on a regression to `class="markdown-body "` does not hold — **the needle includes the closing quote**. `class="markdown-body "` and `class="markdown-body something"` both fail to contain `class="markdown-body"`, because the character after `markdown-body` is a space, not `"`. That closing quote is doing exactly the anchoring the comment asks for; making it a regex would add machinery without adding coverage. (It is also load-bearing in the other direction: the *positive* test uses `class="markdown-body probe-class"`, so a dropped or reordered token fails there.)

22/22 green locally after the change.

PR #1325, opened 2026-08-12
fix: key the synced-query cache on the query set, not the id alone (#1311)

### LINKED ISSUE #1311: Live NodeType compile takes a type-subtree-scoped source snapshot — shared=@Other/Type/Source resolves to zero nodes on a running mesh
## What happened

MeshWeaver.Plugins#432 added `"sources": ["namespace:Source scope:subtree", "shared=@Edu/Lesson/Source"]` to `Edu/Module`. Every gate passed — `compile-check.py` (resolves shared sources from the repo) and the ACR mesh gate (fresh mesh import; the type reached Ok). The Edu sync on memex at 2026-08-12 13:43 then **parked `Edu/Module`** with `CS0103 — LessonLayoutAreas does not exist`.

## The trace

The compile activity (`Edu/Module/_Activity/compile-20260812141153…`) shows why:

```
⏱ NodeType definition resolved in 0ms (ok); source snapshot 0ms (9 node(s), taken ONCE and reused by every stage).
Executed source queries (4):
  - namespace:Edu/Module/Source scope:subtree nodeType:Code
  - path:Edu/Lesson/Source nodeType:Code
  - namespace:Edu/Lesson/Source scope:subtree nodeType:Code
  - namespace:Edu/Module/Test scope:subtree nodeType:Code
Matched Code nodes (9): [all Edu/Module/* — zero from Edu/Lesson]
```

The snapshot the queries execute against contains only the type's OWN subtree, so any `shared=@{OtherType}/Source` entry matches nothing. `search namespace:Edu/Lesson/Source scope:subtree` over MCP returns the two Code nodes fine — they exist and are typed Code; the compile just can't see them. Same family as MeshWeaver.Plugins#430 ("hub-scoped synced queries are partition-local").

## Why this is nasty

- It contradicts both gates: repo-side compile-check AND the fresh-mesh ACR gate resolve shared sources, so a cross-type source ships green and parks the type only on a LIVE mesh at the next sync/restart — the UWDeepfield failure shape, this time with all gates green.
- `Chess/GambitHunt` (`shared=@Chess/Game/Source`) reports Ok (lastCompiledAt 2026-08-12 12:58) on the same mesh — either its compile ran through a differently-scoped path, or that green is from a snapshot that did resolve; either way behaviour is inconsistent across triggers.

## Mitigation shipped

MeshWeaver.Plugins#433 makes `Edu/Module` self-contained again (embeds the `NodeNavigation` area by name instead of compiling in Lesson's code) — unparks the type. But the platform question stands: **either the live compile's source snapshot must span every declared source root (incl. `shared=@` targets), or `shared=@` must be rejected at authoring time** — a directive that silently resolves to zero sources on exactly one class of mesh is the worst of both.

rbuergi:
Fixes #1311.

## The diagnosis in the issue was right about the symptom, wrong about the cause

The issue reads the trace as "the live compile takes a **type-subtree-scoped** source snapshot", i.e. the snapshot is scoped narrowly by construction. It is not. Query expansion is correct at both legs — the activity log's own "Executed source queries (4)" lists the two `Edu/Lesson` queries, and `CodeQueryResolver.Expand` produces them from `shared=@Edu/Lesson/Source` exactly as intended.

What is wrong is that the compile's **cached** leg never executed them.

`IMeshNodeStreamCache.GetQuery(id, queries)` is a get-or-create keyed on **`id` alone**: on a hit it returned the existing entry and **discarded the `queries` argument**. That is only sound if every caller's id fully determines its query set — and the compiler's does not. `NodeSources.CacheId` is `nodetype-sources:{nodeTypePath}`, while the queries are expanded from the NodeType's author-editable `Sources`/`Tests`.

So on a running portal:

1. `Edu/Module`'s entry materialised with the pre-#432 query set.
2. The Edu sync added `shared=@Edu/Lesson/Source`.
3. `InstallSourcesWatcher`'s `DistinctUntilChanged` on `(Sources, Tests)` fired — its entire purpose is to re-resolve on exactly this change — called `NodeSources.GetSources` with the new queries, and got the **old entry** back. A no-op.
4. The compile's cached leg answered from that frozen entry: **established, authorised, and too narrow**. Non-empty, so it won `RaceSourceSnapshot` instantly — the reported `source snapshot 0ms`. `DirectSourceProbe` recomputes its queries every call and *would* have found the shared files, but it cannot answer before its ~1 s quiet window, so its correct wider answer was discarded.
5. Roslyn, shown a short set, emitted a completely genuine-looking `CS0103 — LessonLayoutAreas does not exist` about code sitting right there in the mesh.

This also explains both inconsistencies the issue flags. **Every gate was green** because the failure needs a cache entry that *predates* the edit: a fresh mesh (the ACR gate), the repo-side `compile-check.py`, and any newly started process all materialise the entry from the new declaration. And `Chess/GambitHunt` read `Ok` on the same mesh for the same reason — its declaration was already in place when its entry was first built.

It is the third member of the #1226 / #1279 / #1218 family. Not unestablished (#1226), not unauthorised (#1218's `Anonymous` probe) — **established, authorised and short**. Same lethal shape: a partial source set that reads as a verdict on the author's code.

## The fix

Index each id's entries **by their query set**. Same id + same set = the same shared entry as before; same id + a different set = a stream that answers the set you asked for.

Indexing rather than **replacing** on mismatch is deliberate: two callers holding different declarations would thrash under a replace, re-opening the upstream `SyncedQueryMeshNodes` on every alternating call — the subscription-storm shape this cache exists to prevent. The superseded entry stays connected exactly as it always has (`AutoConnect(1)` never disconnects), so a sources edit costs one extra resident synced query for the life of the process, bounded by declarations authored rather than by traffic. The options-wrapper memo is keyed on the signature too, or the fix would be undone one layer up.

I checked every other `GetQuery` call site: the rest encode their varying inputs in the id (`nav-below:{root}:{includeDocuments}`, `instsync-cfgs:{spacePath}`, …), so none of them can accumulate entries. `NodeSources` was the only violator.

## Tests — fail before, pass after

Both were run against the unfixed tree first:

- `GetQuery_AskedForAWiderQuerySet_DoesNotServeTheNarrowerCachedAnswer` — the mechanism in isolation; deterministic, no compile, no race. Before: the widened set never emits the library node (30 s timeout, last emission holds only the own-subtree node).
- `AddingASharedSourceToAWarmNodeType_IsVisibleToTheNextCompile` — the reported symptom end to end. Before: `source snapshot 0ms (1 node(s))` and `CompiledSources` = `{type/SharedConsumer/Source/model}`, the shared file absent.

`SyncedQueryProjectionContractTest.SharedId_FirstRegistrationWins_EvenWhenTheSecondAsksForContent` pinned the **old footgun as contract** — its `because` reads *"the synced-query cache is keyed by id alone and ignores the queries on a hit … sharing an id across differently-projected reads is a silent-null defect"*. It was the only failure in the whole `MeshWeaver.Query.Test` suite after the change. Rewritten to pin the fixed contract, plus the new claim that widening one caller's set does not evict another's.

The two `MeshWeaver.Courses` comments and the `QuerySyntax.md` callout that documented the id-only key are updated: a `select:` drift between two call sites sharing an id is no longer a silent null, only waste, and the guidance now says so.

## Verification

- `dotnet build -c Release -warnaserror` clean: `MeshWeaver.Hosting` (covers `Graph` + `Mesh.Contract`), `MeshWeaver.Courses`, `MeshWeaver.Documentation`.
- `MeshWeaver.Query.Test` — 388/388 (was 387/388 with the contract test pinning the old behaviour).
- `MeshWeaver.Hosting.Monolith.Test`, targeted at everything that leans on this cache — `CompileSourceSnapshot{Reuse,Wedge}`, `SourceSnapshotEstablishment`, `SourceDiscoveryUnavailable`, `CodeEditRecompile`, `MeshNodeStreamCacheConcurrency`, `LanguageModelSyncedQuery`, `SyncedQueryChangeFeedStarvation`, `AgentSyncedQueryFromHostedHub`, `CompileLegBoundWedge`, and the new class — 31/31.
- `WhatsNewEntryIntegrityTest` + `DocumentationLinkIntegrityTest` — 2/2.


rbuergi:
## CI status — the one red job is a pre-existing shard-2 wedge, not this diff

Merged current main (`408ed73fd`) and re-ran. Everything is green except **shard 2**, which fails on:

```
[CI] MeshWeaver.Hosting.Monolith.Test (part 2/2) exit=124 TIMEOUT (8m wall-clock cap hit — likely fixture/init hang)
```

`Test Results (shard 2)` **passes** — the trx records no failing test. The shard fails purely on the wall-clock cap, i.e. the host did not finish, not that anything asserted false.

**Why this is not attributable to this change**, in order of weight:

1. **PR #1331 hit the identical marker** (`Monolith.Test (part 2/2) exit=124 TIMEOUT`, run 31615938816). Its entire diff is one comment in `LogIncidentIngestService.cs` plus a new test file in `MeshWeaver.Observability.Test` — it adds nothing at all to `MeshWeaver.Hosting.Monolith.Test` and cannot lengthen or wedge it. Two unrelated PRs, same wedge.
2. **This PR's first run passed shard 2** (9m29s, run 31615459925) with these same new tests already present. What changed between the two runs is the base, not the diff.
3. **The two tests this PR adds to that assembly cost ~1 s** (`Passed! … Duration: 1 s` locally for the pair). They cannot be what pushes an 8-minute budget over.
4. Main's own run at `408ed73fd` passed shard 2 — but in **10m04s** for the shard, so the assembly is sitting close to the cap and tips over under load. That is consistent with a load-dependent wedge rather than a deterministic regression on either side.

Per AGENTS.md I am not raising the cap or re-running for luck: a `timeout 8m` trip means something is not completing, and the fix is to find what. That is its own investigation, outside this change set — flagging it here rather than papering over it.

Locally, on the merged tree: `MeshWeaver.Query.Test` 388/388; the 31-test targeted set across `CompileSourceSnapshot{Reuse,Wedge}`, `SourceSnapshotEstablishment`, `SourceDiscoveryUnavailable`, `CodeEditRecompile`, `MeshNodeStreamCacheConcurrency`, `LanguageModelSyncedQuery`, `SyncedQueryChangeFeedStarvation`, `AgentSyncedQueryFromHostedHub`, `CompileLegBoundWedge` and the new class 31/31; `-c Release -warnaserror` clean on every touched project.

The earlier shard-0 failure on the pre-merge run (`OrleansChatTest.CreateThread_AndSubmitMessage_ProducesThreadMessages`, a 50 s timeout in the streaming-response poll) did **not** recur after the merge — shard 0 now passes in 6m28s. I had also ruled that one out by inspection: every `GetQuery` id in the AI/thread path already fully encodes its query set (`AgentPickerProjection` even SHA-fingerprints the queries into the id, and `ChatClientCredentialResolver` derives both id and queries from the same partition set), so keying the cache on the query set is a no-op there and cannot add a subscription to an agent round.

rbuergi:
**Update — sharpening the previous comment.** PR #1331 has since gone **fully green** after it merged the same main (all six shards, including shard 2 at 9m43s). So the `Monolith.Test (part 2/2) exit=124 TIMEOUT` is **intermittent**, not a deterministic consequence of current main. My earlier note leaned on "#1331 hit it too" and should be read with that correction.

The full picture across the four runs is what actually rules this diff out — each PR has one pass and one fail on shard 2, in opposite runs:

| run | base | Monolith.Test additions | shard 2 |
|---|---|---|---|
| #1325 run 1 | `ad177d1c2` | this PR's 2 tests | **pass** (9m29s) |
| #1325 run 2 | `408ed73fd` | this PR's 2 tests | **fail** — part 2/2 `exit=124` |
| #1331 run 1 | pre-merge | none | **fail** — part 2/2 `exit=124` |
| #1331 run 2 | `408ed73fd` | none | **pass** (9m43s) |

A diff-attributable regression cannot produce that pattern: the wedge appears once with the added tests and once without them, and disappears once with them and once without them. It is a load-dependent host hang in `MeshWeaver.Hosting.Monolith.Test` part 2/2 sitting at the 8-minute cap — `Test Results (shard 2)` passes in the failing run, so no test asserts false; the host simply does not finish. Main's own run at `408ed73fd` passed the shard, but took 10m04s, which is the same story from the other side.

The two tests this PR adds to that assembly measure ~1 s together, so they are not the marginal second that tips it.

I am deliberately **not** re-running this PR to get a green tick, and **not** raising the cap: per AGENTS.md a `timeout` trip means something is not completing, and "make the bound bigger" is the band-aid the repo forbids. The part-2/2 hang looks like a genuine open item worth its own investigation — flagging it rather than absorbing it into this change set.

rbuergi:
### Correcting my own record: the canonicalisation is reverted

I took the Copilot review suggestion above (build `SyncedQueryMeshNodes` from the canonical/sorted query set so the stream matches the set-based signature), pushed it, and **it broke two tests**. Reverted in e1f56af4e. Resolving that thread was premature — reopening the observation here.

**The suggestion's premise is right, its remedy is not: query ORDER is load-bearing for multi-query consumers.**

A multi-query synced collection does not deliver its union atomically — each query opens an independent upstream subscription and results land as they resolve. Consumers wait on the *first* snapshot matching a predicate, so the order the queries are issued in decides what is already present when that predicate first trips.

`AgentPickerProjection.ObserveAgents` issues `[baseUnion, ...packageQueries]`, and `AgentPickerProjectionTest.ObserveAgents_PackageSourceInSettings_SurfacesPackageAgent` waits for the package agent, then asserts the base union is in the same snapshot — its `because` says it outright: *"the canonical base union stays first — package rows extend it, never replace it."* Sorting ordinal puts `namespace:PickerPkg/Agent…` ahead of `namespace:rbuergi/Agent|ACME/Agent|Agent…`, so the predicate matched a snapshot holding only the package row.

Measured, both directions, locally:

| | AgentPickerProjectionTest | SkillAutocompleteTest |
|---|---|---|
| with canonicalisation | **5/6** (`Expected {"Package Agent"} to contain "Assistant"`) | **FAIL** on CI (`Last of 1 emission(s) was: []`) |
| reverted | **6/6** | **1/1** |

**What remains true from the review.** A caller that repeats a query still opens a redundant upstream (`SyncedQueryMeshNodes` de-duplicates resulting *paths*, not queries), and which redundancy the mesh carries depends on who materialised the shared entry first. That is worth fixing — but with an **order-preserving** distinct, not a sort, and separately from this PR, since "the arrival order of a multi-query union is part of the contract" deserves to be pinned by a test of its own rather than discovered by one.

Unchanged and still verified here: the signature stays canonical (order- and duplicate-insensitive) as the cache **key** — that is #1311's actual fix, and it never touched the order the stream is built in.

PR #1327, opened 2026-08-12
fix: the ten views whose literal root dropped Style and Class (#1297, part 2)

rbuergi:
Part 2 of the #1297 sweep — **10 of the 17 views that drop both `Style` and `Class`**: the ones whose
box is a single literal root element.

> 🔗 **Stacked on #1323** (base is `fix/1297-class-only-drops`, not `main`) — it needs the `ClassSuffix`
> helper that PR lands. GitHub will re-target this to `main` automatically once #1323 merges.
> **Merge order: #1323 → this → part 3.** The diff shown here is this PR's own 10 views.

## The judgement, made per view rather than mechanically

#1288 established that a blanket wrapper is wrong, with a case that would have shipped undetected
(`TodoLayoutAreas.cs:117` puts `flex: 1` on its fragment's own root **because that root is the direct
flex child** — in-mesh source no build type-checks). So the question for each of the 17 is *where does
the box go*, not *what transformation do I apply*.

For these ten the answer is the same and it is not a wrapper: each renders a panel / page / editor
with **one root element of its own**, carrying only hard-coded declarations and no author input at all.
**That root IS the control's box.** Nothing is created, nothing is nested, nothing moves.

| View | Root before | Where Style / Class now go |
|---|---|---|
| `NodeExportView` | `class="node-export-view" style="max-width: 800px;"` | that root |
| `NodeImportView` | `class="node-import-view" style="max-width: 800px;"` | that root |
| `ExportDocumentView` | `class="export-document-view" style="max-width: 720px;"` | that root |
| `AppearanceView` | `style="max-width: 500px;"`, no class attribute | that root (`class="@Class"`) |
| `CatalogView` | `class="catalog-container" style="@ContainerStyle"` — a code-behind literal | `ContainerStyle` composition + class list |
| `CommentableView` | `class="commentable-wrapper"`, no style attribute | the wrapper — it renders child areas and has no other element, so this is the only place its own Style can live |
| `MarkdownEditorView` | `class="collaborative-editor-container"`, no style | that root |
| `CollaborativeMarkdownView` | `class="collab-md-container @ViewModeClass"` | that root, joined as a token list |
| `MeshNodeContentEditorView` | `class="mesh-node-content-editor" style="display:flex;…"` | that root |
| `Monaco/DiffEditorView` | `style="height: @(ViewModel.Height ?? "500px"); width:100%"` | that host box |

## Two placement decisions, both load-bearing

**The author's Style goes LAST.** `BlazorView.StyleSuffix` appends after the view's own declarations, so
a declared `max-width: 400px` **overrides** the built-in `max-width: 800px`. Dropping the declaration was
the bug; putting it where the default wins would be the same bug with extra steps. The tests assert the
order (`"max-width: 800px; max-width: 400px"`), not just presence.

`DiffEditorView` is the sharpest case: it read `ViewModel.Height` straight into its host box and ignored
`Style` entirely, so `WithStyle("height: 200px")` was a no-op. The control's own `Height` stays the
default and the declared style now follows it.

**The author's Class joins the fixed list, never replaces it.** `.node-export-view`, `.catalog-container`
etc. are the hooks their own stylesheets match on.

## Backwards compatibility

Both suffixes are the **empty string** when nothing was declared, so every root is byte-identical. The
guards assert the exact attribute — `class="node-export-view" style="max-width: 800px;"` — not
`Contain`, so a trailing space or an empty `class=""` fails.

⚠️ **One deliberate exception, stated because it is a real (if tiny) difference:**
`CollaborativeMarkdownView`'s container previously emitted `class="collab-md-container "` — a trailing
space — whenever its view-mode class was empty, which is the default mode. Preserving that byte-for-byte
would have produced `"collab-md-container  probe-class"` (double space) once a class was declared, so
the tokens are now joined and the stray space is gone. A class attribute is a whitespace-separated token
list; that space never participated in any selector match.

Three views compose their attribute in code rather than markup, for a mechanical reason worth recording
so nobody "simplifies" it back: **Razor cannot parse two adjacent explicit expressions in one attribute
value** (`@(a)@(b)` → `CS1073: Unexpected token`).

## Tests — verified by falsification

Thirteen tests added to `ControlStyleRenderingTest`, rendering through the **real** `HtmlRenderer`
against a real monolith mesh. Reverting all ten views:

```
Failed  NodeExportControl_StyleAndClass_ReachTheLiteralRoot
Failed  NodeImportControl_StyleAndClass_ReachTheLiteralRoot
Failed  ExportDocumentControl_StyleAndClass_ReachTheLiteralRoot
Failed  AppearanceControl_StyleAndClass_ReachTheLiteralRoot
Failed  CatalogControl_StyleAndClass_ReachTheContainer
Failed  CommentableControl_StyleAndClass_ReachTheWrapper
Failed  MarkdownEditorControl_StyleAndClass_ReachTheContainer
Failed  CollaborativeMarkdownControl_StyleAndClass_ReachTheContainer
Failed  DiffEditorControl_StyleAndClass_ReachTheHostBox
Failed! - Failed: 9, Passed: 26, Total: 35
```

- The **nine positive** tests fail without the fix — one per rendered view.
- The **four backwards-compatibility** guards pass with and without it.

**`MeshNodeContentEditorView` is the one not rendered**, and the test class doc says why rather than
leaving it implicit: its first render is a progress ring, and the styled root appears only once the
node's editable fields have loaded from the mesh. Its change is the same one-line edit to a root it
already styles.

## Verification

| Check | Result |
|---|---|
| `dotnet build src/MeshWeaver.Blazor -c Release -warnaserror` | 0 Warning(s), 0 Error(s) |
| `dotnet build test/MeshWeaver.Hosting.Blazor.Test -c Release -warnaserror` | 0 Warning(s), 0 Error(s) |
| `ControlStyleRenderingTest` (35) | Passed |
| `WhatsNewEntryIntegrityTest` + `DocumentationLinkIntegrityTest` | Passed |

What's New entry added (`Category: Fix`).

## Left for part 3

The remaining **7 of the 17**, which are the ones with a genuine "which element" question rather than a
single obvious root: `MeshNodeCardView` (three branches, one of which delegates to another control),
`MeshNodeThumbnailView` and `LayoutAreaDefinitionView` (hard-coded `Class`/`Style` parameters shadowing
the bound values), `NavLink` (its `Class` is overwritten by the active-state token), `DialogView` (its
style slot composes CSS variables from `Size`), `MeshNodePickerView` (a `FormComponentBase` that ignores
`ComputedStyle`, hence also dropping `Width`/`Height`), and `MeshNodeRoleEditorView` (four branch roots).


rbuergi:
Re-targeted to `main`: CI (`dotnet-test.yml`) only triggers on `pull_request: branches: [main]`, so with the stacked base this PR would have carried **no checks at all** — and a PR with no checks looks the same as one whose checks passed. The branch still contains #1323's commit (it needs the `ClassSuffix` helper), so the diff below shows both until #1323 merges, at which point it collapses to this PR's own 10 views. **Merge order is still #1323 → this → #1332.**

Issue #1314, opened 2026-08-12
LogIncidentIngestService wraps a TWO-stage chain in one Observable.Using — the second stage's eager capture runs off-scope


rbuergi:
Found while fixing the demo-import AccessContext defect (MeshWeaver.Reinsurance#46 / PR #1307). **Not fixed — it needs its own repro, and guessing is what made #46 take two attempts.**

`src/MeshWeaver.Observability/LogIncidentIngestService.cs:63`:

```csharp
Observable.Using(ImpersonateAsSystem, _ =>
    ReadExisting(path).SelectMany(existing => existing is null ? Create(...) : Fold(...)))
```

`ReadExisting` is a `GetMeshNodeStream` read, so the `SelectMany` **selector** — which is where `Create`/`Fold` take their eager `CaptureContext()` — runs on the *emission* thread, outside the `Using` scope.

**Why the precedent it cites does not cover it.** The comment there points at `PackageInstaller` / `CatalogLayoutAreas` as prior art. Those wrap a **single post** per `Using`, which is correct: the post is constructed synchronously on the subscribing thread, inside the scope. This wraps a **two-stage chain**, and that is the difference that breaks it.

**The general rule, established by measurement in #46 rather than by reasoning:** `Observable.Using` opens its AsyncLocal on the *subscribing* thread and therefore covers only what runs **synchronously** there. Everything after the first stage of a `Concat`/`SelectMany` is off-scope — including the enclosing LINQ projection, not merely a `Defer` body. In #46 that cost the **first** group too, not just the later ones, which is why "hoist the call out of the `Defer`" was the wrong fix.

🚨 **The reason this is worth a ticket rather than a glance: the failure is invisible to an outcome assertion.** In #46, the unfixed run reported `ok=6/6` — every write landed — but stamped with the wrong identity (`fileStamp=Roland` instead of the impersonated one), because `AccessService.CircuitContext` falls back to `hostIdentity`, a plain field written by test hosts. So a test that checks "did the write succeed" passes either way. Any repro here must assert **the identity on the write**, not its success.

**If it reproduces**, the fix shape is available now: PR #1307 made explicit stamping expressible — `WithAccessContext(ctx)` / `ImpersonateAsSystem()` on the builders, plus `WellKnownUsers.SystemContext` — so the post carries the identity rather than depending on which thread subscribed it. Nothing defaults: no identity and no ambient still fails closed.

rbuergi:
[...]

Issue #1242, opened 2026-08-11
CI shard hosts are dying and taking random tests down with them — 4 different tests failed across 4 runs of inert diffs while main stayed green


rbuergi:
## Evidence (2026-08-11, all on `main`-synced branches, all green on `main` itself)

Four CI runs, on diffs that cannot affect any of the tests named, failing **different tests every time**:

| PR | diff | run | failing tests | shards |
|---|---|---|---|---|
| #1219 | one word in one `.md` | 1 | `MeshWeaver.AI.Test.MeshPluginTest.FullCrudWorkflow_CreateGetUpdateDelete` | 3 |
| #1219 | same, after main sync | 2 | `MeshWeaver.AI.Test.ScriptExecutionInUserHomeTest.Run_StreamsProgressTimely_NotBuffered` · `MeshWeaver.PathResolution.Test.HierarchicalBrowsingTests.Query_WithSkipAndLimit_ReturnsPaginatedResults` | 3, 4 |
| #1238 | one-line code default + `.md` | 1 | — **GREEN** | — |
| #1238 | previous + **one word** in the `.md` | 2 | `MeshWeaver.Graph.Test.NodeTypeModelProbeTest.ProbeInstanceModel_SnapshotsContentTypeAndSchema` · `MeshWeaver.Hosting.Monolith.Test.SearchResultStormTest.CatalogSubscription_ContentUpdateStorm_AddsNoExtraRerenders` | 1, 2 |

The #1238 pair is the cleanest control available: **the same branch went green, then red after a one-word documentation edit**, on different tests in different shards.

`main` was green throughout — last five runs at the time of writing (`9398713`, `dad8615`, `eb283c4`, `56421a5`, `9be51fc`).

## The signature

The shard failures are reported as:

```
##[error]Shard N: a test host exited non-zero (failure, crash, or timeout kill)
```

That is a **killed or crashed test host**, not an assertion. The tests named are whichever happened to be executing when the host died — which is why they differ every run and why they span unrelated assemblies (`AI.Test`, `PathResolution.Test`, `Graph.Test`, `Hosting.Monolith.Test`).

## Why it matters

- It reds PRs that cannot have caused it, so every author has to first prove innocence. Two separate PRs today spent a full CI cycle each on this.
[...]

PR #1330, opened 2026-08-12
fix: the hosted-hub disposal join must not out-run the answers it joins (#1317)

### LINKED ISSUE #1317: Hosted hubs disposal times out at shutdown, risking leaked hub state
## What is failing

`HostedHubsCollection.DisposeAsync()` is not completing within the 10-second timeout. At least one hosted hub (a message-bus gossip hub) fails to dispose in time, leaving its connections or subscriptions unclosed.

## Probable cause

The timeout (`5029ms` elapsed, against a 10-second limit) suggests a hub is stuck awaiting a network operation or an in-flight message handler during shutdown. This is a **race at teardown** — the host is signalling disposal but the hub has not drained its work queue. Confidence: **moderate** — the evidence is a single warning with no exception trace, so the hub's exact hang point is unknown without instrumentation.

## Impact

**Low.** 1 occurrence on 1 pod with no repeat since `2026-08-12T14:39Z`. A single hub disposal that times out does not block the pod from shutting down (the disposal already awaits the timeout aggressively), so the impact is limited to: (a) a 5-second delay in shutdown, and (b) the risk that the hub's resources (TCP connections, message subscriptions) leak. No customer-facing effect reported.

## Where to look

- Class: `MeshWeaver.Messaging.HostedHubsCollection` — the `DisposeAsync` (or `StopAsync`) method that awaits all hosted hubs with a `CancellationToken` timed at 10 seconds.
- Each hub's `DisposeAsync` implementation — look for any operation that does not honour the passed cancellation token (e.g. a `Task.Delay`, a blocking network call, or a semaphore wait that outlives disposal).
- A good first step would be adding structured logging inside the hub's dispose path to identify which specific hub type is slow.

---

### Evidence

| | |
|---|---|
| Fingerprint | `be4ec56e4c2c4870` |
| Category | `MeshWeaver.Messaging.HostedHubsCollection` |
| Severity | Error |
| Namespace | `memex-cloud` |
| Pods | `memex-portal-deployment-7d4b5684bb-9lrxl` |
| Occurrences | 1 |
| First seen | 2026-08-12 14:39:02Z |
| Last seen | 2026-08-12 14:39:02Z |

```
2026-08-12 14:39:02Z memex-portal-deployment-7d4b5684bb-9lrxl fail: MeshWeaver.Messaging.HostedHubsCollection[0]
      Hosted hubs disposal timed out after 10 seconds (5029ms). Some hubs may not have disposed properly.
```

</details>

<sub>Opened automatically from `Admin/_LogIncident/be4ec56e4c2c4870`. Recurrences are folded into this issue rather than opening new ones.</sub>

rbuergi:
Fixes #1317.

## Root: **ordering** — a join that pre-empts the terminal answers it is joining

The reported line is self-refuting: `Hosted hubs disposal timed out after 10 seconds (5029ms)`. The cap is `Timeout(TimeSpan.FromSeconds(5))`, and the "10 seconds" string has never matched the code since it was written (`a6a9e4cdf`). That is the small half.

The real defect: **the cap is a second budget over work whose bound is owned one level down and is LONGER.**

| Bound | Owner | Value |
|---|---|---|
| `HostedHubsCollection` join cap | the *waiting* parent | **5 s** |
| `DisposalWatchdogTimeout` → force-teardown → `DisposalCompleted` | each *child* | **8 s** |

`MessageHub` is the only `IMessageHub` implementation (verified: `grep -rn ": IMessageHub" src/ test/`), and its `Dispose` arms that watchdog. So a child's answer **is guaranteed terminal** — just not inside 5 s. The join therefore gave up **3 s before** the mechanism that produces a clean answer, in precisely the wedged case it was written for.

Nesting makes it fire with nothing wedged at all: a child's own disposal is its quiesce budget (2 s default) + its own hosted-subtree join, so a busy two-level tree exceeds a flat 5 s while every individual step stays inside its own budget.

## Is state actually leaked, or does the timeout just report badly? — **both, and the leak is real but bounded**

On expiry the collection calls `SignalDone()` anyway → the owner advances to ShutDown → `DisposeImpl()` + `messageService.Dispose()` tear the DI container down — **while children are still mid-disposal, resolving services from it.** That is the post-teardown straggler class (#613) this file's own comments exist to prevent: the cap was *manufacturing* the leak the rest of the file works to avoid. It also silently voided the in-flight-construction contract `InflightHubCreationDrainTest` asserts.

Bounded, not permanent: the children's own watchdogs still collect them ~3 s later — but against a container that is gone. So: **not urgent, and not merely cosmetic.**

## Change

1. **The join carries no deadline of its own.** This is *not* moving the bound from 5 s to 8 s — it removes a duplicate, wrongly-ordered one. The owning hub keeps its **own** `DisposalWatchdogTimeout` over this whole phase, which is where a "the shutdown path wedged" deadline belongs, and which `ForceTeardownAfterWatchdog`s rather than abandoning children.
2. **A child whose `Dispose()` threw is answered immediately from that known-terminal state.** `MessageHub.Dispose` flips `disposalStarted` *before* arming the watchdog, so a throw in between (e.g. an `AggregateException` out of `CancelExecution`'s cancellation callbacks) leaves a hub that can never signal *and* refuses to retry — subscribing to its `DisposalCompleted` would park the join forever. The removed cap was the only thing hiding that. This is #1298's rule: **a gate that can never open answers now; it does not sit on a timeout.** The pre-existing `catch` around `h.Dispose()` already conceded this branch exists — it just then waited on a signal the faulted call may never send.
3. The lying message goes with the timeout arm; genuine faults still log.

## Verification — fail-before reproduces the incident's own log line

```
[Error] [MeshWeaver.Messaging.HostedHubsCollection] Hosted hubs disposal timed out
        after 10 seconds (5003ms). Some hubs may not have disposed properly.
=== TEST FAILED: the owner declared its hosted hubs disposed while a construction
    it started was still running
```
(prod: `5029ms`)

`HostedHubDisposalJoinTest` extends the contract `InflightHubCreationDrainTest` already pins — "disposal FINISHES the requests it started", currently proven for 1 s only — to past the cap. It holds the window open with a **gate**, not a timer, so nothing is sampled; the one wall-clock number (6.5 s) sits in the band between the removed 5 s cap and the 8 s watchdog, which *is* the defect's footprint. A second test keeps an ordinary teardown prompt, so removing a cap cannot quietly become "wait forever".

- Before: **FAIL**. After: **PASS** (with `InflightHubCreationDrainTest` still green).
- `MeshWeaver.Messaging.Hub.Test` full suite: **212/212**.
- `MeshWeaver.Data.Test` (teardown-heavy regression for a shared disposal-path change): **325/325**.
- `-c Release -warnaserror --no-incremental`: 0 warnings, 0 errors.

## Named, not fixed: the watchdog ladder

`DisposalWatchdogTimeout` is a flat 8 s at **every** depth, and a child's `Dispose()` is called *after* its parent's (the parent quiesces first), so children's deadlines land *later* than their parents'. In a deep enough tree a parent's watchdog can still fire while its subtree is legitimately draining. That is a separate structural defect from the one here — fixing it means making the deadline inherited down the cascade rather than restarted per hub — and I have not observed it in production, so I am reporting it rather than half-fixing it.

## Relationship to the sibling issues

Investigated with #1300 and #1317 as a suspected shared root. **They do not share one.** All three are teardown-adjacent and that is where the resemblance ends:

| Issue | Root | Fix shape |
|---|---|---|
| #1300 | **classification** — a legitimate terminal shutdown reported as an error, because the predicate tested the outermost exception type and the real chain is wrapped | recognise the state |
| #1308 | **ordering** — a stream callback writing an accumulator the renderer is draining | make the accumulator own it |
| #1317 | **ordering** — a join pre-empting terminal answers it is waiting for | remove the duplicate bound |

They need opposite fixes and cannot be merged into one change. Three PRs.


rbuergi:
## Re: the `exit=139` SIGSEGV on shard 0 — reachability established by measurement, not attribution

The first CI run was red with `Run tests (shard 0)` **`exit=139 SIGSEGV`**. Because this PR *is* the disposal-ordering change, "known flake" is the wrong reflex — a disposal fix that introduces a segfault must not merge. So I tried to falsify the opposite.

### Two corrections to the initial reading

1. **The crashing project is `MeshWeaver.Hosting.Orleans.Test`, not `MeshWeaver.FutuRe.Test`.** Shard 0 does not contain `FutuRe.Test` at all — it is on shard 2 (`SHARD_INDEX=0 SHARD_TOTAL=6 .github/scripts/shard-assign.sh`). So this is not #613's project.
2. **`Test Results (shard 0)` passed** — the trx recorded zero failures. The host died at `Starting:` + 9.7 s; the job failed via the exit-marker gate, not a test.

Two arguments I tried and **discarded as wrong**:
- *"The crash predates any disposal."* False. The trx shows tests completing from +2.75 s, with eight done by +9.59 s — fixtures are being disposed inside the crash window.
- *"The framework never logs the cap firing in CI."* Unreliable instrument: `Orleans.Test` surfaces no framework log lines at all (its whole local stdout is 7 lines), so absence proves nothing.

### The measurement

I instrumented **main's unmodified `HostedHubsCollection`** (locally, uncommitted, in a throwaway worktree; reverted after) and ran the exact assembly that crashed. This measures whether my change would have had *anything to act on*:

| | |
|---|---|
| Disposal joins executed | **18,430** |
| Joins that reached the `Timeout(5s)` this PR removes | **0** |
| `Dispose()` throws (the other branch this PR changes) | **0** |
| Join duration | p50 **0 ms** · p95 **0 ms** · p99 **30 ms** · max **2,557 ms** |
| Joins ≥ 1 s | 20 of 18,430 (0.1%) |

This PR is behaviourally observable in exactly two situations: a join exceeding 5 s, or a `Dispose()` that throws. Both have a **measured occurrence count of zero across 18,430 opportunities** in this assembly. The diff is inert here, so it cannot be the cause.

Corroborating: `MeshWeaver.Messaging.Hub.Test` (which contains the only test that deliberately constructs the >5 s condition) **is not on shard 0**. And `Orleans.Test` passes locally on this branch — 156/156, exit 0 — as it does on `origin/main`.

### Direction of the change

The removed cap could only ever cause the owner to stop waiting *sooner*. Removing it makes the owner wait for its children **longer** before `DisposeImpl` unloads ALCs — strictly *narrowing* the window in which pooled work touches a freed `LoaderAllocator`, which is the #613-sibling mechanism. The one branch that settles *earlier* (a child whose `Dispose()` threw) requires a throw that was measured zero times.

### Incidental finding, worth its own issue

`max = 2,557 ms` against a 5,000 ms cap. On an unloaded Mac the join already reaches **51% of the budget** — so on a contended CI runner the cap was closer to firing than anyone assumed. That is independent support for removing it, and it is why the reported incident happened at all.

### Two things for whoever is on #613

1. **The core dump is not recoverable.** `DOTNET_DbgMiniDumpName` writes to `/tmp/coredumps/`, but the artifact step only copies `*.dll`/`*.pdb` into `collected-logs/symbols-<proj>/`. The `.dmp` dies with the runner — so the artifact ships symbols with nothing to symbolicate against. That is the same wall the 2026-08-04 investigation hit; adding symbols did not fix it, because the dump is the missing half. **Uploading the dump (or running `dotnet-dump analyze` in-job) is the change that would make these falsifiable.**
2. **`exit=139` is not exclusive to `FutuRe.Test`.** This one is `Hosting.Orleans.Test` — the other ALC-heavy assembly. If #613 is being tracked as a FutuRe-specific crash, that scope is too narrow.

rbuergi:
## Correction to the comment above — I could NOT establish that this crash is pre-existing

A census of CI history came back and it cuts **against** the reading above, so I am correcting it rather than leaving the more comfortable version standing.

**Window 2026-08-06 → 2026-08-12, 1,367 workflow runs.** Every shard-0 log that could possibly carry a non-zero marker was fetched and grepped — 81 of them, a census rather than a sample (18 on `main`, 63 across 35 PR branches; the other 277 failed runs had a green shard 0, which the gate step proves cannot carry one).

`MeshWeaver.Hosting.Orleans.Test exit=139` occurs **exactly once in that window, and it is this branch.** Base rate elsewhere: **0/80.** On `main`: **0/18** inspected failures plus ~764 green runs.

So "known pre-existing flake" is **not supported**. The recurring `exit=139` class in this repo is attached to `MeshWeaver.FutuRe.Test` on shard 2 (issue #613, a CoreCLR background-GC fault) — a different project. What Orleans.Test does carry chronically is `exit=1 TESTFAIL`: **78 of those same 81 logs**, a rotating single-test flake, plus #1242 ("CI shard hosts are dying and taking random tests down with them", closed 2026-08-11).

### The hole in my measurement

The 18,430-join figure was measured on **macOS/arm64, unloaded**. CI is **Linux/x64, contended**. The max join I saw was **2,557 ms against a 5,000 ms cap** — within 2× — so I cannot claim from that data that the cap never fires on a CI runner. I said the 2,557 ms was "closer than anyone assumed"; that observation cuts against my own conclusion and I should have said so there.

### What I still think, and why — stated as a direction, not a clearance

Even granting that the cap *does* fire on CI, the direction still argues against this PR:

- **Before:** cap fires at 5 s → `SignalDone()` → the owner advances to ShutDown → `DisposeImpl` unloads node ALCs **while children are still disposing and resolving services**. That use-after-free window exists on `main`.
- **After:** no cap → the owner waits → ALCs unload only once children have finished. The window is **closed**.

In the only scenario where this diff is active at all, it *removes* the unload-during-live-work window rather than creating one. The single path by which it could plausibly make a segfault more likely is indirect — holding hubs alive longer raises heap pressure, and #613's mechanism is a background-GC fault — but the crash landed 9.7 s in with roughly eight tests completed, which is very little accumulated state for that story.

### Position

**Not cleared, and I am not asking for this to merge on the argument above.** n=1 is n=1. The re-pushed head is running now; that gives a second sample. My recommendation:

- **Green on the re-run** → flake, merge normally.
- **`exit=139` on shard 0 again** → treat it as this PR's, do not merge, and I will chase the heap-pressure path (the honest next step there is pushing the join instrumentation to a throwaway branch so the cap-firing question is answered on a *Linux CI runner* instead of my laptop).

The two notes for #613 in the previous comment stand, and the second one gains weight: **`exit=139` is not FutuRe-exclusive**, and **the core dump is never uploaded** (`DOTNET_DbgMiniDumpName` writes to `/tmp/coredumps/`, but the artifact step copies only `*.dll`/`*.pdb`), so no one can symbolicate any of these. That missing dump is why this had to be argued from reachability at all.

rbuergi:
## CI attribution: the failing test is not this PR, and the assertion was wrong

`LogIncidentIngestAccessContextTest.FirstSighting_IsAttributedToSystem_NotTheEmissionThreadsAmbient` (shard 4) is a **pre-existing flake on `main`** — recorded on `c39288499` (merge of #1332, run 31634660324), before this PR's CI run. The test file was added by `cb40296af` (#1314) and is untouched by this diff.

It is also not merely a flake to wait out. The failing line was:

```
Did not expect value to be 4 because the absent-node read completes off the subscribing thread …
```

— `hop.SelectorThread.Should().NotBe(hop.SubscribeThread)`, on a run whose own probe output shows the pipeline behaved correctly: `subscribeThread=4@0.2ms selectorThread=4@6.2ms`. The stage deferred by 6 ms and the pool handed the continuation back to the subscribing thread.

Two things were wrong, so this is not a green-washing edit:

1. **Wrong property.** Thread identity cannot separate "ran inline" from "ran later". This file's own negative control already records that an `AsyncLocal` reaches a pool hop scheduled from inside the scope — so a thread change never testified about the identity either.
2. **Wrong probe.** The guard sat on `hop`, which makes no identity assertion. The identity assertion is on `hopNoFlow`, which had **no** non-vacuity guard at all: a stage running inside the still-open scope would have seen `system-security` for free.

Replaced with `InlineDetector`, which measures the property the probes actually need — *not run synchronously nested inside the Subscribe call* — exactly rather than by proxy, and cannot flake (same thread ⇒ program order; other thread ⇒ not nested by construction). The guard now sits on `hopNoFlow`, the fold path gains the guard it never had, and `ProbeInlineDetectorSeesInline` is a **positive control** so a "did not happen" assertion cannot be vacuously green.

Mutation-tested: making the read inline turns both guards red with their intended messages. Reverted, green again. `MeshWeaver.Observability.Test` 124/124, `MeshWeaver.Messaging.Hub.Test` 214/214. Current `main` merged. Not merging — serial merge train.

Issue #435, opened 2026-07-12
Silent loss of uploaded collection files after hub teardown; mesh search returns zero results on deployed portal


rosalie1984:
# Silent loss of uploaded collection files + mesh search returning zero results

**Environment:** deployed portal `memex.systemorph.com`, platform `3.0.0-ci.866+build.639194691516252957.b367183f6923bdbbba994ede177e21aca7519ca1`. Platform redeploy (migration stamp) at **2026-07-12 16:12 UTC**. All observations below from the same day, via the MCP tool surface (`get` / `search`). Space/node names anonymized; exact paths available on request.

## Symptom 1 — uploaded collection files vanish silently

Timeline (2026-07-12, UTC):

| Time | Observation |
|---|---|
| 18:25:18–19 | 7 files (docx/pdf) bulk-uploaded into a node's `content` collection via the portal Files view — all `lastModified` stamps within ~1 s |
| ~18:30–18:45 | `get @{node}/content/` lists all 7 files; fetching an individual docx works (content conversion served correctly) |
| ~2 h later | Same listing returns `[]`; direct fetch returns `File 'X' not found in collection 'content'` |

Additional facts:

- **No delete was issued.** The node's children are unchanged, no new/moved nodes in the space, and the parent node document itself was not modified (no version bump attributable to the loss).
- All 7 files disappeared **as a unit**, not individually.
- Node-document writes on the same partition the same day (creates + many `edit_content` rounds on sibling nodes) persisted fine — the loss is specific to the **collection-file path**.

### Hypothesis

The upload wrote only into the owning hub's **in-memory state** (which served the successful reads for ~20+ min) and was **never durably persisted**. The next grain teardown — idle release, or process churn from an unrelated deployment/recompile in the same silo — erased it. Reads succeeding right after upload means users get **no signal** that their files are on borrowed time. This looks like the same silent-drop family as the 2026-07-02 virtual-data-source stream outage (RouteStreamMessage silent-drop / stream-cache idle-release suspects).

### Suggested repro / verification

1. Upload files into a node's `content` collection; confirm the listing shows them.
2. Recycle the node's hub (`DisposeRequest`) — or wait for idle release.
3. List again. If the files are gone, that's a deterministic repro pinning the persistence defect.
4. In parallel: check the partition schema in Postgres (or the blob store, wherever collection files are supposed to land) **immediately after step 1** — if no row/blob exists at that point, the write-side persistence never happened and the teardown is only the messenger.

[...]

PR #1337, opened 2026-08-12
test: pin GetDataStream semantics for a never-set id

rbuergi:
## The gap

No test anywhere in `test/` asserted what `GetDataStream<T>(id)` does for an id that was **never written**. The semantics lived only in a code comment in `EditorExtensions.cs` — and that absence is why three copies of the same defect shipped in MeshWeaver.Plugins (#435, fixed in PR #441), two of them the doors a standard pack reaches a user through, both of which could stay shut forever.

## The behaviour (now asserted, not described)

A **written** id emits immediately. A **never-set** id emits **nothing at all** — not `null`, not a default — and never completes.

`EntityStore.ReduceImpl(EntityReference)` returns `null` for an absent id → `WorkspaceStreams.CreateReducedStream` filters it (`Where(x => x is { Value: not null })`) → the reduced stream's replay buffer never receives an `OnNext`, so there is nothing to replay → `LayoutAreaHost.GetStream` / `LayoutExtensions.GetDataStream` filter nulls a second time.

**The trap is the failure DIRECTION.** `GetDataStream(id).Take(1)` reads like a once-guard. On a never-set id it never fires, so it does not prevent a *duplicate* run — it prevents the **first** one, silently, with no exception to grep.

## What is pinned — `GetDataStreamUnsetIdTest` (test/MeshWeaver.Layout.Test)

| Test | Pins |
|---|---|
| `SetId_EmitsItsValue` | baseline — a written id emits |
| `NeverSetId_EmitsNothing_WhileASecondWriteLandsOnTheSameHost` | the core semantic |
| `TakeOneOnNeverSetId_NeverFiresAndNeverCompletes_SoAOnceGuardBlocksTheFIRSTRun` | the trap: the guarded action never runs, and the stream never completes |
| `ValueSetAfterSubscription_Arrives_SoTheStreamIsEmptyNotDead` | the stream is empty, not dead |
| `StreamOverload_NeverSetBoolId_EmitsNothing_NotFalse` | same on the `host.Stream` overload `EditorExtensions` uses, where a never-set `bool` is not `false` |

## Asserting absence without a vacuous pass

Every negative is measured against a **positive control on the same host**. No assertion ends on a bare clock: each observation window is closed by an **observed** write landing on a *different* id of the same `LayoutAreaHost` (`.TakeUntil(controlLanded)`, with `Replay(1).Connect()` so the subscription is live *before* the control write). When the window closes, the data plane has demonstrably delivered while the never-set subscription was open — so the silence is a measurement, not a gap.

Verified non-vacuous by **mutation**: seeding the never-set id at render turns all three negative tests red; reverting it turns them green again.

## Docs now point at the assertions

- XML `<remarks>` on both `GetDataStream` overloads (`LayoutAreaHost`, `LayoutExtensions`) — the place IntelliSense actually shows a caller.
- The `EditorExtensions.MapToToggleableControl` comment now names the test instead of restating the rule.
- `GUI/Observables.md` had *endorsed* the `.Take(1)` form ("a one-shot FORM READ … fine") with no caveat; it now says why that case is safe (the form id is definitely written) and what happens when it isn't.

No behaviour change — test + documentation only.

## Verification

- `dotnet build -c Release -warnaserror` clean: `MeshWeaver.Layout`, `MeshWeaver.Layout.Test`, `MeshWeaver.Documentation.Test` (0 warnings, 0 errors).
- `MeshWeaver.Layout.Test`: **407/407 pass** (includes the 5 new tests).
- `DocumentationLinkIntegrityTest`: pass.

## What's New

Skipped deliberately — test + doc-comment change, no user-visible effect.

### DIFF
--- a/src/MeshWeaver.Documentation/Data/GUI/Observables.md
+++ b/src/MeshWeaver.Documentation/Data/GUI/Observables.md
@@ -265,7 +265,14 @@ The most important rule: keep dynamic regions small. A large re-render is always
     return Task.CompletedTask;  // the handler itself is synchronous
 })
 // Note: the `.Take(1)` above is a one-shot FORM READ inside a click action —
-// fine. NEVER `.Take(1)` a stream that feeds a live-bound view: the binding
+// fine, BECAUSE the form id has definitely been written by the time the user
+// can click it. 🚨 On an id that was NEVER written, `GetDataStream(id)` emits
+// NOTHING (not null, not a default) and never completes, so `.Take(1)` never
+// fires: as a "run this once" guard it does not block a duplicate run, it
+// blocks the FIRST one — silently, with no exception to grep. Need a value for
+// an id that may be unset? `.StartWith(default)` or seed it with
+// `host.UpdateData(id, …)` first. Pinned by `GetDataStreamUnsetIdTest`.
+// And NEVER `.Take(1)` a stream that feeds a live-bound view: the binding
 // freezes on the first emission (see Data Binding).
 
 // ❌ Wrong — async handler deadlocks the pump under load
--- a/src/MeshWeaver.Layout/Composition/LayoutAreaHost.cs
+++ b/src/MeshWeaver.Layout/Composition/LayoutAreaHost.cs
@@ -987,6 +987,14 @@ public record LayoutAreaHost : IDisposable
     /// <paramref name="id"/>, deserialized as <typeparamref name="T"/>. Skips nulls; applies
     /// DistinctUntilChanged.
     /// </summary>
+    /// <remarks>
+    /// 🚨 An id that was <b>never written</b> emits <b>nothing at all</b> — not <c>null</c>, not a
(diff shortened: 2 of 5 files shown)


rbuergi:
Added a sixth test in `ed2a262`: **`UpdateDataWithNull_WritesNothing_LeavingTheIdNeverSet`**.

`LayoutAreaHost.UpdateData` is a **silent no-op for null**, so `host.UpdateData(id, node.Description)` on a null Description writes nothing and every downstream `.Take(1)` guard on that id is dead — while the calling code reads as though the id were seeded. That is the bridge between "I set it" and "it was never set", i.e. the most likely way the guard bug actually gets authored.

Found while sweeping this repo for live instances of the trap: `AgentView` `CombineLatest`s nine `.Take(1)` form streams (a single unset arm would kill the whole Save), and it is correct **only** because all nine ids are seeded `node.X ?? ""`. The hazard was known and defended against, but never pinned. No live instance of the bug found in this repo.

Verified by mutation like the others (write a non-null value → red). Local: `MeshWeaver.Layout.Test` **408/408**.

PR #1338, opened 2026-08-12
fix(iopool): the blocking join needs its OWN counter, and Dispose never drained

rbuergi:
Follow-up to #1334, **which I merged before reading its review**. Two of the three findings were real defects — one mine, one pre-existing — so this fixes forward rather than leaving them.

## 1. Mine: the idle signal was driven by a shared counter

`_blockingIdle` was set/reset on 0↔1 transitions of `_inFlight`, which is **shared** with `Invoke`, `InvokeStream` and `SubscribeThroughPool`. Those transitions therefore do not correspond to *blocking* work starting or stopping, and it broke in both directions:

- a blocking leaf starting **while an async leaf ran** incremented to 2, so the `Reset` never fired and `Drain` saw "idle" — **re-introducing the exact bug #1334 existed to fix**;
- a blocking leaf finishing while an async leaf ran decremented to 1, so the `Set` never fired and `Drain` waited out its full 30 s budget before reporting a survivor that had already unwound.

A dedicated `_blockingInFlight` now drives the signal, and the residual is read from it — `_inFlight` would double-count an async survivor already reported via the gate residual.

## 2. Pre-existing, and worse: `Dispose()` never drained

```csharp
if (_disposed) return;
_disposed = true;
Drain();          // ← Drain starts with `if (_disposed) return 0;`
```

So `Dispose()` joined **nothing**, while its own summary promises:

> when it returns, no pool thread is running, so the caller may safely unload the node ALCs whose types that work referenced

The disposal path *is* the path that unloads every ALC — the one place that promise had to hold is the one place it did not. `Drain()` now runs before the flag flips.

## 3. The repro was slow

It let `Drain` sit out the leaf's 10 s hold to observe the outcome. It now runs `Drain` on another thread and asserts it does **not** return while the leaf executes — faster, and a stronger statement: it pins the **join**, not just the reporting.

**The class is 15 tests in 1 s**, up from 13 in ~10 s.

## Tests

- `Drain_joinsABlockingLeaf_evenWhileAnAsyncLeafIsRunning` — pins defect 1: a concurrent async leaf must not make the blocking join believe the pool is idle.
- `Dispose_actuallyJoins_beforeItReturns` — pins defect 2. The leaf respects its token and then delays deliberately *after* cancellation, to construct a definite ordering rather than to wait for propagation: a `Dispose` that does not join returns in microseconds and would otherwise pass by luck.

```
dotnet build src/MeshWeaver.Mesh.Contract -c Release -warnaserror   # clean
dotnet test test/MeshWeaver.Hosting.Test --filter IoPoolTest        # 15/15, 1s
```

## Process note

#1334 was green with all shards passing and mergeable, and I merged it without reading the review that had already landed on it. The gate I skipped is "address findings before merge", not the CI gate. Recording it here because the same haste is what produced defect 1.

### DIFF
--- a/src/MeshWeaver.Mesh.Contract/Threading/IoPool.cs
+++ b/src/MeshWeaver.Mesh.Contract/Threading/IoPool.cs
@@ -26,6 +26,15 @@ public sealed class IoPool : IIoPool, IDisposable
     // A ManualResetEventSlim here is consistent with this file being the ONE sanctioned home for
     // such a primitive: IoPool IS the boundary between the turn-based schedulers and blocking I/O.
     private readonly ManualResetEventSlim _blockingIdle = new(initialState: true);
+
+    // 🚨 A DEDICATED counter, not _inFlight. _inFlight is shared with Invoke / InvokeStream /
+    // SubscribeThroughPool, so 0↔1 transitions on it do NOT correspond to blocking work starting or
+    // stopping — and the signal was driven off exactly those transitions. Both directions broke: a
+    // blocking leaf starting while an async leaf ran incremented to 2, so the Reset never fired and
+    // Drain saw "idle" (the original bug, unfixed), and a blocking leaf finishing while an async leaf
+    // ran decremented to 1, so the Set never fired and Drain waited out its whole budget then
+    // reported a survivor that had already unwound. (Copilot review, #1334.)
+    private int _blockingInFlight;
     private volatile bool _disposed;
 
     // Teardown safety net for the drain join: cancellation makes in-flight leaves release in ms,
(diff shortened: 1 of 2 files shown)


rbuergi:
## Split: the `Dispose`-actually-drains half is out of this PR

The first CI run carrying it saw `OrderedRouteDispatcherTest.SameDestination_SubscribesLegsInOrder_OneAtATime` **hang for its full 30 s budget**. That test passes locally **4/4 in 1 s** on this branch at `DOTNET_PROCESSOR_COUNT=4`, and Orleans tests have timed out spuriously in CI twice today — so the attribution is genuinely unproven in both directions.

Which is precisely why it should not ship next to anything else. Swapping those two lines changes **teardown semantics repo-wide**; if it does cause a hang, that has to be attributable to one commit and nothing else. It gets its own PR.

**What remains here is the unambiguous half:**
- the dedicated `_blockingInFlight` counter — the defect I introduced in #1334, where the signal was driven by 0↔1 transitions of a counter shared with the async paths;
- the faster, stronger repro (asserts the **join**, not just the reporting).

14/14 in 861 ms, built with CI's exact flags this time (`-c Release -p:CIRun=true -warnaserror` — omitting `CIRun` is what let the xUnit1031 errors through earlier).

**The `Dispose` defect is not dropped** — it is now documented at the site, in code, saying plainly that `Dispose()` joins nothing today, why (its own idempotence guard on `_disposed`), and that the fix is deferred rather than forgotten. So it cannot be lost if this thread is.

Issue #604, opened 2026-07-22
GitHub sync: git→mesh import silently FullReplace-prunes mesh-only nodes (no paths logged, no opt-out for bidirectional Spaces)


rosalie1984:
## Summary

A `git → mesh` import of a GitHub-synced Space runs in **`FullReplace` (mirror) mode** and, in doing so, **silently deletes any mesh node that is absent from the branch tree** — logging only an aggregate *count*, never the paths. For a Space that is edited both on the mesh AND via GitHub sync (the default `SyncDirection.Bidirectional`), a node created on the mesh but not yet committed to the branch is destroyed by the next unrelated "Update to latest" / catch-up import, with no record of what was removed.

Two independent gaps combine here: a **visibility** gap (the prune names nothing) and a **default-policy** gap (GitSync inherits `FullReplace` where the AI catalogs deliberately opt out of it). Either alone would have made the incident far less damaging.

## 1 · Prune is silent — it logs a count, not the paths (visibility)

The importer computes prunable nodes and deletes each one, but the only per-path logging is on the *failure* branch. A **successful** prune produces just an aggregate count in the activity summary:

```csharp
// StaticRepoImporter.cs — the prune loop
toPrune.Select(t => AsSystem(hub, () => meshService.DeleteNode(t.Path))
    .Select(_ => 1)
    .Catch<int, Exception>(ex => {
        logger?.LogWarning(ex,
            "[StaticRepoImport] {Partition}: prune of {Path} failed (continuing).",
            source.Partition, t.Path);            // ← path logged ONLY on failure
        return Observable.Return(0);
    }))
...
var summary = $"Imported {count.Imported} node(s), pruned {prunedCount}, synced {contentCount} content file(s).";
//                                              ^^^^^^^^^^^ count only — which nodes? unknown
```

So the activity log for a destructive import reads `… pruned 7 …` with no way to learn *which* seven nodes were deleted. When a page/feature "disappears," there is no breadcrumb tying its removal to the import.

**Ask:** emit the pruned paths into the activity log (an `Information`/`Warning` line per pruned path, or a single line listing them). A prune is a destructive act on user data — it should be as auditable as the `⚠ Failed to import …` lines already are.

Ref: `src/MeshWeaver.Graph/StaticRepoImporter.cs` — prune loop (~line 639) and the summary message (~line 674).

## 2 · GitHub sync imports default to `FullReplace` (policy)

[...]

PR #1341, opened 2026-08-12
fix: batch the import activity log per phase — the O(n²) CPU burner throttling memex-cloud

rbuergi:
## The live production CPU burner

`memex-cloud` sits at **51% of CFS periods throttled**. The dominant term is the static-repo
importer writing its own progress log **once per item**.

`NodeTypeCompilationActivity.AppendLog` does **one `stream.Update` per message**, and
`StaticRepoImporter` called it per pruned / kept / failed node at four sites. Each `stream.Update`
runs `CreateSingleObjectPatch`, which **`SerializeToNode`s the entire node content (O(n)) and diffs
all n elements (O(n))** — on *every* write. Over n appends that is **O(n²)** CPU and allocation.

Live attribution, by node path:

| path | writes | content | est. serialised |
|---|---:|---:|---:|
| `AgenticEngineering/_Activity/import-69f6248b…` | **5,239** | 141 kB | **719 MB** |
| `UWDeepfield/_Activity/import-af545232…` | 2,581 | 114 kB | 288 MB |
| `MeshWeaver/_Activity/import-1d1a65…` | 637 | 204 kB | 127 MB |

230,674 writes / ~7.1 GB serialised across 25,579 activity nodes; **no non-import node in the top
20**. Downstream: ~7.8 cores demanded against a 6-core quota → CFS throttling → turns, timers and
`Dequeue` all slip together (thread-pool delay #1284, routing pile-up #1172). No deadlock.

**The patch *output* was never the problem.** `PatchExtensions.DiffArray` is element-wise, so an
append walks the equal prefix and emits a single O(1) `add` op. The cost is on the **input** side of
the patcher — which is exactly why replacing the patch library in #1292 (1.2× faster diff, 17–19%
fewer allocations) did not move the needle: the dominant cost is re-serialising a growing document.

## The fix — batch per phase

- **New `NodeTypeCompilationActivity.AppendLogs(...)`** writes N messages in **one** `Update`
  (`Messages.AddRange`). `AppendLog` becomes the one-message call into it, and its doc comment now
  states the cost model so the per-item shape does not come back.
- **`StaticRepoImporter.Run`** carries each per-file line along with its per-file result tuple —
  the same trick `WrittenPaths` already used — and appends the whole **upsert phase** in one write.
  The **prune phase** likewise appends its kept-from-prune + pruned lines in one write once the
  deletes have settled.
- **`Complete` was not the right vehicle on its own.** It writes the *terminal* status, so folding
  the phase lines into it would defer every line past the prune, content-sync and manifest-write
  phases — no live progress at all. The phase-level append sits alongside it.

Nothing leaves the audit trail: every kept and every pruned path is still **named** on the activity
log (issue #604's requirement), and still lands before the terminal status — the per-path write
queue is serial, so a reader that sees a non-Running status has already seen every line.

**Accepted cost (owner's call):** per-item live progress during a long import is replaced by
per-phase progress. The satellite-node design that would keep per-item progress alters the
`ActivityLog` contract, its readers and the GUI, and was deliberately deferred as a separate change.

No band-aids: no quota change, no scheduler hop, no recursion-depth bound, no throttle/debounce on
the writes. The work itself is gone, not moved.

## Verification — before / after write counts

New test `StaticRepoImportActivityWriteCountTest` runs a real import into a fresh partition, then
re-imports the same paths under a two-way policy so every node takes the "kept local change" branch
— a phase that writes **no content at all**, so the attempt-activity node's revisions are exactly
the import's own logging. It asserts on **`MeshNode.Version`** (the node's own "+1 per real
modification" counter), never on log text, so a reworded message cannot satisfy it.

| items in the phase | activity writes **before** | **after** |
|---:|---:|---:|
| 4 | 8 | **5** |
| 16 | 20 | **5** |

Writes grew 1:1 with the item count; they are now a constant (opening line, root line, one phase
line, terminal summary). Fail-before/pass-after was verified by reverting the `src/` diff, rebuilding
and re-running: RED with `4 -> 8; 16 -> 20`, GREEN with `4 -> 5; 16 -> 5`.

Extrapolating the measured production shape: the 5,239-write / 719 MB activity becomes a handful of
writes, and the ~7.1 GB of serialisation across the 25,579 activity nodes collapses by the same
order.

## Test runs (local, this branch)

- `MeshWeaver.Graph.Test` — **1060 passed, 0 failed** (includes the new test)
- `MeshWeaver.GitSync.Test` — **160 passed, 0 failed** (covers the prune/preserve paths and the
  "activity log NAMES the pruned node" assertion of #604)
- `MeshWeaver.Documentation.Test` — `WhatsNewEntryIntegrityTest` + `DocumentationLinkIntegrityTest` green
- `dotnet build -c Release -warnaserror` clean for `MeshWeaver.Graph`, `MeshWeaver.Graph.Test`,
  `MeshWeaver.Documentation`

What's New entry: `Category: Fix` — users notice imports being far cheaper.

### DIFF
--- a/src/MeshWeaver.Documentation/Data/WhatsNew/2026-08-12-imports-no-longer-burn-cpu-on-their-own-progress-log.md
+++ b/src/MeshWeaver.Documentation/Data/WhatsNew/2026-08-12-imports-no-longer-burn-cpu-on-their-own-progress-log.md
@@ -1,0 +1,23 @@
+---
+Name: Imports no longer burn CPU on their own progress log
+Category: Fix
+Description: Large repository imports are dramatically cheaper, so the portal stays responsive while one is running.
+Icon: Sparkle
+Order: -20260812
+---
+
+# Imports no longer burn CPU on their own progress log
+
+While a large import was running, the portal could slow to a crawl — pages that took seconds to
+appear, chats that stalled mid-answer — and it got worse the bigger the import was.
+
+The cause was the import's own progress log. Every file it kept, pruned or failed to write was
+recorded as a separate update to the import's activity record, and each of those updates re-read and
+re-compared the entire record. So the hundredth line cost a hundred times what the first one did, and
+a single import on one portal ended up rewriting hundreds of megabytes just to describe what it had
+done. That work was competing with everything else the portal had to do.
+
+The import now records each stage once, with all of that stage's lines together. The activity log
+still names every file it kept and every node it pruned — nothing is lost from the audit trail — but
+the cost no longer grows with the size of the import. The one visible difference is that progress now
+appears stage by stage rather than file by file.
--- a/src/MeshWeaver.Graph/Configuration/NodeTypeCompilationActivity.cs
+++ b/src/MeshWeaver.Graph/Configuration/NodeTypeCompilationActivity.cs
@@ -98,15 +98,38 @@ internal static class NodeTypeCompilationActivity
     }
 
     /// <summary>
-    /// Append an Information-level <see cref="LogMessage"/> to the activity
+    /// Append a single <see cref="LogMessage"/> (Information-level by default) to the activity
     /// log so callers can see progress in real time. No-op when
     /// <paramref name="activityPath"/> is null. Best-effort: failures log and
     /// swallow — observability must never break a compile.
+    ///
+    /// <para>🚨 ONE write per call, and a write is <b>O(size of the whole activity node)</b>: the
+    /// patcher <c>SerializeToNode</c>s the entire content and diffs every element on EVERY update.
+    /// So N appends to one activity cost <b>O(N²)</b> CPU + allocation. Never call this in a loop —
+    /// collect the messages and use <see cref="AppendLogs"/> (per phase) or <see cref="Complete"/>
+    /// (terminal). The per-item shape burned 719 MB of serialisation on a single memex-cloud import
+    /// activity (5,239 writes over a 141 kB node) and was half of the CFS throttling on that pod.</para>
     /// </summary>
     public static void AppendLog(IMessageHub hub, string? activityPath, string message, ILogger logger,
-        Microsoft.Extensions.Logging.LogLevel level = Microsoft.Extensions.Logging.LogLevel.Information)
+        Microsoft.Extensions.Logging.LogLevel level = Microsoft.Extensions.Logging.LogLevel.Information) =>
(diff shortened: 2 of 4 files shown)


rbuergi:
## Correction: "the patch output was never the problem" holds for the own-stream path only

Recording this here because it is the sentence a future reader will rely on, and it is true for
exactly half the write paths — the half that is **not** the common case for these writes.

Surfaced while root-causing #1351 (`OutOfMemoryException` serializing a `MessageDelivery<RawJson>`
onto an Orleans memory stream).

### The two patch paths differ

**Own-stream — RFC 6902 JSON Patch.** As this PR says: `PatchExtensions.DiffArray` is element-wise,
so an append walks the equal prefix and emits a single O(1) `add` op. The cost is entirely on the
*input* side (re-serialising a growing document), which is what batching fixed. ✅

**Cross-hub — RFC 7396 JSON *merge* patch.** `MeshNodeStreamExtensions.ComputeMergePatchDiff`
recurses into JSON **objects only**; any changed **array** is emitted whole
(`patch[key] = updatedValue?.DeepClone()`). `MeshNodePatchMerge.ExtractBaseValues:303` then mirrors
that shape and clones the **previous** array whole into `BaseValues`.

So a single append to an N-element collection in `MeshNode.Content` produces a `PatchDataRequest`
carrying **~2N elements**. Here the patch *output* very much is the problem: **message size is
O(collection size)**, and it is unbounded because `ActivityLog.Messages` has no cap — unlike every
other log-shaped collection in node content (`LogIncident.Samples` → `MaxSamples`,
`CompletionMemory.Entries` → `MaxEntries = 300`).

This matters because cross-hub is the *common* case for activity writes: the writer is rarely the
activity node's own hub.

### A closed door, so nobody re-opens it

**`BaseValues` cannot simply be dropped for arrays.** It is genuinely consumed —
`MeshNodePatchMerge.MergeArray(liveArr, patchArr, baseArr)` performs a real three-way array merge by
element identity, and the base arm is what distinguishes "the writer added this" from "the writer
never had it". Dropping it would silently regress cross-hub array merges into last-write-wins. The
obvious halving is not available.

### Where this goes

Not a regression in this PR — batching was the right fix for the CPU term and it holds. The message-size
term is a separate axis that batching cannot reach (fewer writes, but each still O(N)), and it is
what actually OOMs a serializer. The cure is the satellite migration for growing in-content
collections, which is now scoped and in progress for **activities** first. Cross-referencing #1359,
#1351 and #1200.

PR #1343, opened 2026-08-12
A successful recompile now rebinds the instances — the healthy binding was the one nobody watched

rbuergi:
## The incident

memex, 2026-08-12: a GitSync update recompiled three NodeTypes **successfully** and every instance carried on executing the **previous** assembly. Recycling six hubs by hand fixed it. Nothing was broken, and nothing said so.

## Why it was invisible

Every signal reported that the deploy had landed:

| | |
|---|---|
| `compilationStatus` | `Ok` |
| `lastCompileSucceededAt` | fresh, after the import |
| `latestAssemblyPath` | the **new** dll |
| `compiledSources` | `== currentSourceVersions` |
| the source nodes | the new code |
| the type's `Tests` area | rendered green |

…while the running code was the old assembly. The only way to see it from outside was to look for a string the deploy had changed — a renamed test label. **A green status and a green test table are both produced by whatever assembly is loaded**, so neither can testify about itself.

## The gap

Every `WithOverlaySelfHeal` call site is a **degraded** branch: no registration, slow-path timeout, unresolved pin, missing bytes, ABI-stale, `Unavailable`, compile `Error`, compiling-in-progress. An instance that **succeeded** in binding a good assembly got no watcher at all — and `EnrichWithNodeType`'s re-enrichment short-circuit then caches that binding for the grain's lifetime.

So the case nobody watched was the case that worked, and the only exits were idle-deactivation or a manual recycle.

`NodeTypeRecompileExtensions` closes the half where the recompile is never **triggered** — *"the nodes read current while the mesh keeps executing the previous assembly"*, in its own words. This closes the half where it succeeds and nobody rebinds.

## The fix

`WithStaleAssemblySelfHeal` / `ArmStaleAssemblySelfHeal`, armed on the three healthy returns. It posts one self-`DisposeRequest` — the `RecycleLayoutArea` idiom the overlay watcher already uses — when the type publishes a usable build whose `LatestAssemblyPath` differs from the one this instance bound.

**The gate is the assembly identity, deliberately not the node `Version`.** A version gate is what failed here before: memex 2026-07-27, where a recompile of an already-`Ok` type need not advance the version, the watcher waited for a bump that never came, and `ArmOverlaySelfHeal` had to grow a grace timer to escape it. Comparing the published assembly path against the bound one is exact in both directions — it changes **iff** a different build was produced, so it cannot miss a deploy and cannot fire on an unrelated node write (a `RequestedReleaseAt` stamp, release notes, a pin edit). It cannot hot-loop either: after the recycle the instance rebinds the new path, which becomes the next watcher's baseline, so **each published build costs each instance at most one self-recycle**.

Not armed when the type publishes no assembly path, or when no hub configuration would be composed — the same null-config carve-out `WithOverlaySelfHeal` already documents, so the fail-fast NACK-fallback hub is never swapped for a bare one.

## How it was tested

- **`StaleAssemblySelfHealWatcherTest` — 5 cases green, and proven NON-VACUOUS**: neutering the predicate fails the two healing assertions (the other three keep passing, which is correct — a dead watcher never fires). Pinned: a new assembly heals exactly once; republishing the **same** assembly never fires however far the version advances; a new assembly at an **unchanged** version still heals (the 2026-07-27 regression); unsettled and foreign content ignored; disposal stops it.
- `MeshWeaver.Graph.Test` **1064/1064** green
- Enrichment/compile integration tests (Monolith — `SelfHeal|Recompile|NodeTypeCompile`) **20/20** green
- Graph, Hosting and Graph.Test build clean under `-c Release -p:CIRun=true -warnaserror`


rbuergi:
## CI attribution: the failing test is not this PR

`ThreadAgentIntegrationTest.FullFlow_CreateThread_SendMessage_StreamResponse_SaveReply` (shard 4) fails with an **identical signature on `main`**, on a commit that predates this branch's merge base:

| | this PR (run 31638881449) | main `dabfecba9` (run 31611972654) |
|---|---|---|
| shard | 4 | 4 |
| site | `ThreadAgentIntegrationTest.cs:190` | `ThreadAgentIntegrationTest.cs:190` |
| error | `GetMeshNode('ACME/ProductLaunch') timed out after 60.0s` | `GetMeshNode('ACME/ProductLaunch') timed out after 60.0s` |

`dabfecba9` is an ancestor of this branch's merge base `2fea35426` (`git merge-base --is-ancestor` ⇒ 0), so this PR's code is definitively not in it.

The check was worth making rather than assuming: `ACME/ProductLaunch` **is** `nodeType: ACME/Project`, a dynamic NodeType, so it does go through the enrichment path this PR touches. The evidence rules that out anyway — the same failure happens without the change.

Root cause is filed separately as #1362. In short, the trace refutes the "NO LOCAL HUB" wording: the hub activated at +6.7 s, `HANDLER_ENTER`/`HANDLER_EXIT state=Processed` at +7.27 s, then no reply for the remaining 53 s — `HandleGetDataRequest` has no terminal answer for a workspace stream that never emits, and the hub had come up with `$type: Project` unregistered.

Pushed here: **current `main` merged** (the branch was ~57 files behind), no code change. `MeshWeaver.Graph` clean under `-c Release -warnaserror`. Not merging — serial merge train.

rbuergi:
Good catch — fixed. `AssertDisposedExactlyOnce` proved only that *a* `DisposeRequest` was posted, so the test would have passed with the post carrying no options at all, or naming the wrong hub — the worse of the two, since it would bounce someone else while leaving this instance stale.

Now pinned the way `OverlaySelfHealWatcherTest` and `NodeTypeRebindWatcherTest` do it: capture the `Func<PostOptions, PostOptions>` and compare against `baseOptions.WithTarget(new Address(InstancePath))`, derived from the same base so the auto-generated `MessageId` doesn't perturb record equality.

**Proven non-vacuous**: retargeting the post to a foreign address fails both firing cases with *"the self-heal recycle must target the instance hub's own address"*.

11/11 self-heal tests green.

Issue #1324, opened 2026-08-12
Is the per-compile _Activity hub retention BOUNDED? DisposeOnTimeout(15 min) should now reach it — measure past the horizon


rbuergi:
## The measurement

A **15-second local repro** — `NodeTypeRecompileAlcLeakTest` in `test/MeshWeaver.Hosting.Monolith.Test`, extended in PR #1321 to measure bytes rather than only load contexts. A **trivial** NodeType (`config => config`, a few KB of IL) recompiled three times, `GC.GetTotalMemory(forceFullCollection: true)` after three aggressive blocking collections at each step:

```
managed baseline = 114 MB      (after the first compile settles)
after recompile 1: 132 MB      live contexts = 1
after recompile 2: 153 MB      live contexts = 1
after recompile 3: 174 MB      live contexts = 1
FINAL: 62 MB over 3 recompiles = ~20 MB PER RECOMPILE, linear (+18, +21, +21)
```

Reproduce:

```bash
dotnet test test/MeshWeaver.Hosting.Monolith.Test -c Release \
  --filter NodeTypeRecompileAlcLeakTest      # ~15s, prints the numbers
```

## What this is NOT

- **Not the assembly load contexts.** They unload correctly — 1 live throughout with no instance hub, 2 with a live one, stable across N recompiles (#1272). The ALC bookkeeping, the pin drain and the lease deferral are all right.
- **Not transient allocation.** Measured after three aggressive blocking full GCs; this is what SURVIVES collection.
- **Not the batch bake in aggregate.** One linked build of 279 types left the pod at 2.5 GB total.
- **Not page cache** (the production symptom shows RSS 23.6 GB).
- **Not the LSP workspace cache** (keyed per NodeType, disposed on version change).

## Why it matters

This is very likely the memex-cloud growth: a pod went 2.5 GB → 24.5 GB at **130–340 MB/min**, which is 7–17 recompiles a minute — exactly what a morning of merges produces, since each merge fires a GitSync re-import that recompiles the touched types. The portal reaches its 28 GiB limit in a couple of hours and OOMs.

## Where to look

Untested hypotheses, in the order I would try them:

[...]

Issue #539, opened 2026-07-18
Thread wedges at StartingExecution forever when the _Exec commit is lost (e.g. NodeType release recycles the hub mid-claim) — no recovery on activation


rosalie1984:
## Symptom

A thread round submitted via `hub.StartThread(...)` can wedge at `Status = StartingExecution` **forever**: the user message stays in `pendingUserMessages`, no cell is allocated, no agent round runs, and nothing ever recovers it. The thread only executes when its hub is explicitly re-activated (e.g. a `DisposeRequest` recycle of the main node followed by an access).

## Production evidence (memex portal, UW Deepfield Industry News auto-update)

Two gather rounds (agent `Assistant`, model `DeepSeek-V4-Flash`, submitted through the canonical `hub.StartThread` surface from a layout-area click):

| Thread (under `UWDeepfield/IndustryNews/_Thread/`) | Submitted | Wedged state | Recovered |
|---|---|---|---|
| `gather-…-2958` | 2026-07-16 05:27:46 | `StartingExecution`, message in `pendingUserMessages`, thread version 1 — for **2.5 days** | 2026-07-18 ~21:50, within seconds of an explicit recycle of `UWDeepfield/IndustryNews` + next access; thread ran to completion (v1 → v90), items landed correctly |
| `gather-…-b0f2` | 2026-07-18 21:41:11 | same | same moment, same way (v1 → v90) |

Trigger correlation — both wedged rounds were submitted **~60–70 s after a NodeType compile + release recycled the main-node hub**:

- Release `20260716052635` at 05:26:35 → thread submitted 05:27:46 → wedged.
- Release `20260718214041` at 21:40:41 → thread submitted 21:41:11 → wedged.

(Opening the feed triggers the stale-assembly recompile/release, which recycles the per-node hub; the user's toggle click then lands right on that boundary.)

The model was exonerated: both rounds executed quickly and correctly on `DeepSeek-V4-Flash` once the hub was re-activated.

## Root cause

The round dispatch is a two-edge handshake:

1. **Claim** — `InstallSubmissionWatcher` (on the thread hub's action block) writes `Idle → StartingExecution`.
2. **Commit** — the `_Exec` round watcher's first emission commits `StartingExecution → Executing` and launches the round.

[...]

PR #1345, opened 2026-08-12
fix: share a workspace's local reduced streams instead of minting a hub per call (#1324)

rbuergi:
Part of #1324 — the dominant retainer, not all of it (see *What remains* below), so this deliberately does **not** auto-close the issue.

## The measurement, and what it ruled out

Reproduced the issue's own repro (`NodeTypeRecompileAlcLeakTest`) on this machine — 110 MB → 171 MB
over three recompiles, ~20 MB each, load contexts steady at 1 — then took `dotnet-dump` heap dumps
around the loop and diffed `dumpheap -stat`.

**It is not Roslyn.** Not one `CSharpCompilation`, `MetadataReader` or symbol-graph type appears in
the top of the delta. The retained objects are:

| type | before → after (3 recompiles) | per recompile |
|---|---|---|
| `MeshWeaver.Messaging.MessageHub` | 217 → 406 | **+63** |
| `Autofac.Core.Lifetime.LifetimeScope` | 218 → 407 | +63 |
| `MeshWeaver.Messaging.Serialization.TypeRegistry` | 218 → 407 | +63 |
| `System.Text.Json.JsonSerializerOptions` | 435 → 813 | +126 |
| `…Serialization.TypeDefinition` | 12 367 → 22 228 | +3 287 |

**63 message hubs per recompile, ~140 KB apiece.** Dumping the hosted-hub tree by address shows what
they are: `sync/{id}` sub-hubs, one per `SynchronizationStream`, hosted on the NodeType hub, on
`_Activity/compile-state`, on the fresh `_Activity/compile-<ts>`, and on the cache hub.

## Root cause

`Workspace.GetStream(reference)` was **not memoized**. Every call runs
`ReduceManager.ReduceStream` → `WorkspaceStreams.CreateReducedStream`, which constructs a
`SynchronizationStream` — and a `SynchronizationStream` constructs a hosted `sync/{id}` hub with its
own Autofac scope, `TypeRegistry` and `JsonSerializerOptions` — then registers itself for disposal
**on its PARENT** (the data source's primary stream), which lives as long as the hub. So an uncached
`GetStream` leaks a hub per *call*, reclaimed only when the whole hub dies.

Stack-sampling the streams created during one recompile names the callers:

```
x12  Workspace.GetStream <- DataExtensions.HandlePatchDataRequest      ← every cross-hub write
x24  Workspace.GetStream <- MeshNodeStreamHandle.GetStream <- Subscribe ← every own-node read
x6   Workspace.GetStream <- MeshNodeStreamHandle.UpdateOwn             ← every own-node write
```

Nothing about a plain reduce is caller-specific, and **remote** reduced streams were already cached
per `(owner, reference, identity)` — the local twin of that cache simply did not exist.

## The fix

Cache local reduced streams on the workspace, keyed on the reference, with the same liveness
re-check `GetExternalClientSynchronizationStream` uses and the same split: a caller-supplied
`StreamConfiguration` (client id, subscriber, property bag) makes the stream caller-specific, so
those stay uncached. A faulted `Lazy` is evicted rather than poisoning the reference for the
workspace's life.

## Measured after

| | before | after |
|---|---|---|
| retained per recompile | 20 MB | **9.3 MB** |
| hubs per recompile | 63 | **22** |
| mesh steady-state hubs | 187 | **48** |

The steady-state drop is the pleasant surprise: the same waste was being paid on ordinary reads and
writes, not only on compiles.

## What remains (honestly)

The remaining ~9 MB is a **second retainer with the same shape**, and it is the one the issue's own
third comment named: `Workspace.EvictForPath` parks the evicted remote stream in
`_evictedRemoteStreams` **without disposing it**, so every change event on a subscribed path mints a
fresh client/owner `sync/` pair (6 on the cache hub + 8 on `compile-state` per compile here) that is
reaped only by the 10-minute idle sweep with zero subscribers, or by `Workspace.Dispose`. Fixing it
needs a decision about *whose* lifetime a parked stream belongs to, so it is deliberately not in
this change. The per-compile `_Activity/compile-<ts>` node hub is a third, smaller one (~1 hub +
6 streams per compile).

The guard therefore moves from **32 MB → 16 MB**, not to single digits: a bound nothing can
currently pass is a red test, not a guard. The test comment records the exact remaining retainers
and says to tighten it again with them.

## Verification

- `MeshWeaver.Data.Test` — 329/329
- `MeshWeaver.Hosting.Monolith.Test` — 629/629 (includes `NodeTypeRecompileAlcLeakTest` at the new 16 MB bound)
- `dotnet build -c Release -warnaserror` clean for both touched projects, after merging current main

### DIFF
--- a/src/MeshWeaver.Data/Workspace.cs
+++ b/src/MeshWeaver.Data/Workspace.cs
@@ -483,18 +483,79 @@ public class Workspace : IWorkspace
             new DataChangeRequest { Deletions = instances.ToImmutableList(), ChangedBy = null }
         );
 
+    // 🚨 LOCAL reduced streams are cached exactly like the REMOTE ones above — because a reduce
+    // is NOT free and NOT garbage-collectable.
+    //
+    // Every ReduceManager.ReduceStream constructs a SynchronizationStream, and a
+    // SynchronizationStream constructs a hosted `sync/{id}` sub-hub (its own Autofac lifetime
+    // scope, TypeRegistry and JsonSerializerOptions — ~140 KB apiece), then registers itself for
+    // disposal ON ITS PARENT (WorkspaceStreams.CreateReducedStream). The parent is the data
+    // source's primary stream, which lives as long as the hub. So an UNCACHED GetStream leaks a
+    // hub per CALL, released only when the whole hub dies.
+    //
+    // Nothing about a plain reduce is caller-specific, and the callers are hot: the
+    // PatchDataRequest handler resolves the target stream this way on EVERY cross-hub write, and
+    // MeshNodeStreamHandle does it on every own-node read and every own-node Update. One NodeType
+    // recompile made ~60 of these; a portal morning of GitSync re-imports made enough to drive
+    // memex-cloud from 2.5 GB to 24.5 GB at 130–340 MB/min (Systemorph/MeshWeaver#1324).
+    //
+    // A caller-supplied `configuration` DOES make the stream caller-specific (client id,
+    // subscriber, initialization callback, property bag), so those stay uncached — the same
+    // split the remote cache draws by keying on the subscribing identity.
+    private readonly ConcurrentDictionary<WorkspaceReference, Lazy<ISynchronizationStream>> _localStreamCache = new();
+
     /// <inheritdoc />
     public ISynchronizationStream<TReduced> GetStream<TReduced>(
         WorkspaceReference<TReduced> reference,
         Func<StreamConfiguration<TReduced>, StreamConfiguration<TReduced>>? configuration
         )
     {
-        return (ISynchronizationStream<TReduced>?)ReduceManager.ReduceStream(
-            this,
-            reference,
-            configuration
-        ) ?? throw new InvalidOperationException("Failed to create stream");
+        if (configuration is not null)
+            return ReduceLocalStream(reference, configuration);
+
+        while (true)
+        {
+            var lazy = _localStreamCache.GetOrAdd(reference,
+                _ => new Lazy<ISynchronizationStream>(
+                    () => ReduceLocalStream(reference, null),
+                    LazyThreadSafetyMode.ExecutionAndPublication));
+
+            ISynchronizationStream stream;
+            try
+            {
+                stream = lazy.Value;
+            }
+            catch
+            {
+                // A Lazy in ExecutionAndPublication mode CACHES the exception, so a transient
+                // failure (e.g. HubDisposingException from a hub that is winding down) would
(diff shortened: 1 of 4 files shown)


rbuergi:
## CI red on the first run — diagnosed, not retried

`OrphanedStartingExecutionRecoveryTest.OrphanedStartingExecutionClaim_RecoversOnActivation_AndRoundRuns`
failed on shard 5:

```
Expected dictionary to contain key "68da7c42" because the queued message must still be present on the orphaned claim.
```

**It is a race the test was always exposed to, and this change legitimately shifts it.** That test seeds
an orphaned `StartingExecution` claim straight to storage, then *subscribes* to activate the thread hub
cold. Subscribing is what wakes the hub — so the recovery under test (roll the claim back to `Idle` →
the submission watcher drains `PendingUserMessages` into a round) runs on the same hub that must also
serve the `SubscribeRequest` snapshot. Nothing in the product orders those two. The assertion required
the snapshot to win.

This PR removes ~40 stream+hub constructions from a cold activation, so init now finishes sooner and
the drain wins more often. The state being asserted is precisely the one the fix under test exists to
consume — an intermediate, not a specification.

**Checked before touching it**, rather than assuming a flake:
- Passes locally on this branch, including under `DOTNET_PROCESSOR_COUNT=2` (`status=StartingExecution`
  — the snapshot wins here).
- The snapshot path is unaffected by the caching change: `HandleSubscribeRequest` →
  `SubscribeToClient` → `CreateSynchronizationStream` passes a `StreamConfiguration` (client id +
  subscriber), and configured streams stay **uncached** by design in this PR. So what moved is
  activation *cost*, not snapshot *content*.
- Every assertion that pins issue #539 is downstream of the failure and never ran on the red build:
  drained `PendingUserMessages`, `Status` back to `Idle`, `ExecutionStartedAt` cleared, the Echo reply
  in the response cell, and the revert-proven log pin on the recovery branch itself.

**Fix:** the checkpoint now asserts `UserMessageIds.Contain(userMsgId)` — durable by construction, since
the drain moves entries into `IngestedMessageIds` while `CommitRoundAndExecute` / `FoldConsumedInbox`
explicitly restore `UserMessageIds ⊇ IngestedMessageIds`. It proves what the checkpoint is *for* (the
cold activation loaded the SEEDED node) and is strictly stronger against the failure mode that matters —
content materialising as a defaulted record — which the old assertion could not tell apart from a won
race. Read through `ContentAs<T>` for the same reason.

`MeshWeaver.Threading.Test` 128/128 locally; `MeshWeaver.Data` + the test project build clean under
`-c Release -warnaserror` after merging current main.

Issue #1326, opened 2026-08-12
GitSync reports "Skipped (0 nodes)" while a space is behind (_Activity/import-{hash} fingerprint), and force-prune eats Release records (AddIfUnderPrefix case-sensitivity)


rbuergi:
Found by a fleet-wide staleness sweep on 2026-08-12 (compare each `{Space}/_GitSync.lastSyncCommitSha` against `git log <sha>..origin/main -- <subdirectory>`):

**On memex-cloud, with WORKING push webhooks on both repos** (MeshWeaver.Plugins and education deliver `push` there, and 34 of 43 spaces were current), these spaces sat stale:

- `DoublePendulum`, `FractalStars`, `ThreeBody` — missed the same-day Plugins merge `c3519b2` (2 commits touching their folders).
- `AgenticPrimer` (6 behind, since 08-07), `AgenticPrimerDe` (6), `AgenticOffice` (5), `AdvancedBusinessRules` (3), `SwissSolvencyTest` (2), `ReinsurancePractice` (1) — education merges including learner-facing fixes (the `/static`→`/api/content` media fix live-broken since 08-09, the exercises-grid mount fix).

So the headless push-update matched SOME spaces of a repo and not others, silently — the classic shape: the delivery fans out per sync source and something (per-space matching, an error swallowed per space, subdirectory casing, a bounded batch?) drops part of the set with no log anyone sees. Second, known-adjacent defect confirmed again: a manual `update` on stale `ThreeBody` reported **"Imported Skipped (0 node(s))"** while behind — only `force: true` converged it (and the force PRUNED 47 mesh-minted `Release` bookkeeping nodes as 'absent from the repo', which sync ownership #749/#752 arguably should protect).

Asks:
1. Per-space outcome logging on the push fan-out — a push that updates 3 of 9 matching spaces must say which 6 it skipped and why.
2. Root-cause the non-matching spaces (all six education courses + the three Plugins showcases).
3. The stale-skip ("Skipped (0 nodes)" while behind) is the same lie in the manual path — one fix likely covers both.
4. Force-prune should not delete operational satellites (`Release` records) the repo never carries.

All ten spaces were converged by hand today (updates + force where skipped). Sweep method + matrix context: Systemorph/Memex `docs/deployments.md` ("How MODULE updates reach instances").

🤖 Generated with [Claude Code](https://claude.com/claude-code)

rbuergi:
[...]

PR #1346, opened 2026-08-12
fix: a repository's green build now updates EVERY space that follows it (#1326)

rbuergi:
Part of #1326 — asks #1 (per-space outcome logging) and #2 (root-cause the non-matching spaces). Asks #3 and #4 are a different mechanism and are left open, so this deliberately does **not** auto-close the issue.

## First, a correction to the framing

**A `push` event imports nothing, by design.** `GitHubWebhookProcessor.ProcessPush` logs and returns
0 — the sync is CI-gated, and the trigger is a green `workflow_run`. So "push webhooks are live" is
not the premise that matters; the 34 spaces that *did* update were updated by the build-completion
path, and that path works. The question is only why the same 9 never were.

## Root cause

`QueryConfigNodesAsSystem` asks "which spaces follow this repository?" with an **unpinned**
`nodeType:GitHubSyncConfig` query and **no stated limit**. On Postgres an unpinned query is served
by `PostgreSqlCrossSchemaQueryProvider`, which answers a limitless request with a **page**:

```csharp
var orderBy = query.OrderBy != null ? … : "n.last_modified DESC";
var limit = query.Limit ?? 50;      // ← one global LIMIT over the UNION of every partition schema
```

The repo filter runs in C# *after* that clip, so configs that fell outside the 50 most recently
modified rows were never even considered — and the result is indistinguishable from the complete
set.

**Why it is always the same 9, and why they never recover.** The order is `last_modified DESC`, and
a successful sync *rewrites the config node* (`RecordSeenCommit` / `RecordLastSync`). So every space
that updated moves to the front of the window and every space that missed out sinks further out of
it. The omission is self-reinforcing and completely silent — which matches the reported pattern
exactly: a fixed set of stragglers, each a different number of commits behind depending on when it
fell off.

This is the same line of code as **#1216** (the batch bake resolving 50 `Code` nodes for 237 pending
types), seen from a second caller. That is what makes it a framework fix and not a one-line patch at
the call site.

## The fix

**1. An explicit way to say "this is an enumeration, not a page."**
`MeshQueryRequest.Complete()` / `MeshQueryRequest.NoLimit`. The merge layer already treated a
non-positive limit as "do not clip" (`ClipMergedInitial` clips only when `limit > 0`); the storage
providers now *translate* it instead of mistranslating it:

| provider | before | after |
|---|---|---|
| `PostgreSqlSqlGenerator` | could emit `LIMIT -1` (syntax error) | no `LIMIT` clause |
| `PostgreSqlCrossSchemaQueryProvider` | `query.Limit ?? 50` folded it to 50 | `int.MaxValue` |
| `PostgreSqlMeshQuery` | `count >= -1` returned one row | folded away before the guard |
| `StorageAdapterMeshQueryProvider` | load cap of `skip - 1` | no cap |
| `CosmosMeshQuery` | `TOP -1` | no `TOP` |

`PostgreSqlPartitionedMeshQuery` now propagates the request limit **whatever its sign**. Refusing to
propagate `<= 0` (added by #1216) avoided a literal `LIMIT 0`; substituting the fan-out's default of
50 instead is the bug in the other direction, and the one that shipped.

**2. A default that clipped is no longer silent.** The cross-schema provider warns — naming the
query — whenever it hands back exactly the default page to a caller that stated no limit. This is
the generalisation of the issue's ask #1: the *next* caller to make this mistake is findable instead
of invisible.

**3. The fan-out itself.** It says `Complete()`, and it waits for the **merged Initial** rather than
whatever emission arrives first (`MeshQuery` emits exactly one merged Initial once every provider
has reported; a bare `Take(1)` can capture a pre-Initial empty snapshot — the hazard
`GitHubSyncService` already guards against three files over).

**4. Per-space outcome logging** (issue ask #1, literally). The fan-out now reports how many configs
exist in the mesh, how many it selected, and for every one it passed over, its path and the exact
predicate that dropped it — `branch 'main' != built branch 'release'`, `already at this commit`,
`direction is ExportOnly`. `ConfigMatchesBuild` is re-expressed as "there is no reason to skip it"
so the predicate and the log line cannot drift apart.

## Verification

`CompleteEnumerationFanOutTests` — 66 probe rows across two partition schemas, against a real
Postgres:

```
default page: 50 of 66     ← no limit stated: a page, as documented
complete:     66 of 66     ← Complete(): every match
limit 10:     10           ← a stated positive limit still clips
```

**Fail-before verified** by reverting only the provider translation (`query.Limit ?? DefaultFanOutLimit`):
the `Complete()` leg fails with `2201W: LIMIT must not be negative` — i.e. before this change there
was no way for a caller to ask for the whole set at all.

- `MeshWeaver.Hosting.PostgreSql.Test` — 697/698 (1 pre-existing skip)
- `MeshWeaver.GitSync.Test` — 160/160
- `dotnet build -c Release -warnaserror` clean for every touched project

## Not fixed here

The issue's asks #3 (`"Imported Skipped (0 node(s))"` while genuinely behind) and #4 (force-prune
deleting mesh-minted `Release` records) are a different mechanism — the content-fingerprint marker
at `{space}/_Activity/import-{hash}` and `ComputePrunableNodes` — and are left for a separate
change. Notably `Release/` is spared today only by `SyncIgnore.Default`, not by the prune's
governance rule (`_`-prefixed segments), so under `FullReplace` a force *is* free to remove those
nodes; that is the thread to pull for #4.

### DIFF
--- a/src/MeshWeaver.Documentation/Data/WhatsNew/2026-08-13-a-repository-update-now-reaches-every-space-not-most-of-them.md
+++ b/src/MeshWeaver.Documentation/Data/WhatsNew/2026-08-13-a-repository-update-now-reaches-every-space-not-most-of-them.md
@@ -1,0 +1,34 @@
+---
+Name: A repository update now reaches every space, not most of them
+Category: Fix
+Description: When a repository built successfully, some spaces were quietly left on old content — always the same ones, and the more they fell behind the less likely they were to catch up. Every matching space is now updated, and any that is skipped says so by name.
+Icon: Sparkle
+Order: -20260813
+---
+
+# A repository update now reaches every space, not most of them
+
+Spaces that follow a repository refresh themselves whenever that repository builds successfully.
+Most of them did. Some never did — and the ones that never did were always the same ones.
+
+The refresh starts by asking a simple question: which spaces follow this repository? That question
+was being answered the way a *search box* answers — with a first page of results, the most recently
+changed ones, capped at fifty. For a search that is exactly right. For "go through all of them" it
+is quietly wrong, and nothing in the answer says it was only a page.
+
+What made it stick was the ordering. Refreshing a space marks it as recently changed, so every
+space that *did* refresh moved to the front of the queue, and the ones that missed out drifted
+further towards the back. A space that fell off the end could never get back on: it simply stopped
+following the repository, while every build reported success. On one installation nine spaces out
+of forty-three had been sitting on stale content for up to a week that way — including course
+material with fixes their readers never saw.
+
+A read that needs *everything* can now say so, and this one does. The same mistake had already cost
+us once elsewhere — a build pass that saw fifty source files out of thousands — so the platform now
+also warns whenever it hands back a capped page to a caller that never asked for one, naming the
+query, rather than letting it pass for the whole set.
+
+The refresh has also learnt to account for itself. It now reports how many spaces follow the
+repository, how many it updated, and for each one it passed over, which one and why — "on a
+different branch", "already on this commit". A space that is being skipped is now something you can
+read in the log instead of something you have to notice.
--- a/src/MeshWeaver.GitSync/GitHubWebhookProcessor.cs
+++ b/src/MeshWeaver.GitSync/GitHubWebhookProcessor.cs
@@ -253,15 +253,10 @@ public sealed class GitHubWebhookProcessor
     /// <paramref name="headSha"/>: the branch must match, the source must be allowed to import, and
     /// the source must not already sit on that commit.
     /// </summary>
+    /// <remarks>Expressed as "there is no reason to skip it" so the predicate and the log line
+    /// (<see cref="SkipReason"/>) can never disagree about why a Space was left behind.</remarks>
     internal static bool ConfigMatchesBuild(GitHubSyncConfig? cfg, string branch, string headSha)
-    {
-        if (cfg is null || cfg.Direction == SyncDirection.ExportOnly)
-            return false;
-        if (!string.Equals(cfg.Branch, branch, StringComparison.OrdinalIgnoreCase))
-            return false;
-        // Already imported this exact commit — a re-run of the same green build must be a no-op.
-        return !string.Equals(cfg.LastSyncCommitSha, headSha, StringComparison.OrdinalIgnoreCase);
-    }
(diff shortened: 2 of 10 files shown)


rbuergi:
## Red on the rerun is #613, not this change — failed jobs re-run

`MeshWeaver.FutuRe.Test exit=139 SIGNAL SIGSEGV (host died on a signal, not an assertion)` on shard 2.
**Zero failing tests in the trx** — the host segfaulted; `createdump` wrote a 1.4 GB core.

That is [#613](https://github.com/Systemorph/MeshWeaver/issues/613), still OPEN: *"MeshWeaver.FutuRe.Test
exit=139 is a CoreCLR background-GC fault (null MethodTable in background_sweep) — not teardown, not
ALC unload"*. Our own source calls it out by name — `NodeAssemblyLoadContext.UnloadGcProbe`: *"Every
FutuRe exit=139 dump so far faults on the CONCURRENT collector (bgc_thread_function → gc1 →
background_sweep)"*.

**Causation is excluded by construction, not by assumption:**

- The immediately preceding run on this branch — the *same* tree minus the last commit — was
  **green**, and that green run already contained every change reachable from `MeshWeaver.FutuRe.Test`
  (`MeshWeaver.Mesh.Contract` + `MeshWeaver.Hosting`). The only delta since is
  `PostgreSqlCrossSchemaQueryProvider.cs`.
- `test/MeshWeaver.FutuRe.Test/MeshWeaver.FutuRe.Test.csproj` references Fixture, Hosting,
  Hosting.Monolith(+TestBase), Layout, Graph, Mesh.Contract, Messaging.Hub — **no
  `MeshWeaver.Hosting.PostgreSql`, no Cosmos, no GitSync**. It cannot load the changed code.
- The one file in the diff that FutuRe *can* reach is `StorageAdapterMeshQueryProvider.LoadCap`, and
  it changed behaviour only for a NON-POSITIVE limit. Nothing outside the webhook fan-out sets one,
  so for every existing caller the computation is unchanged.

A host that dies on a signal with no recorded test failure is the harness-crash case, so the failed
jobs are re-run rather than the change being retried blind. Locally on this branch:
`MeshWeaver.Hosting.PostgreSql.Test` 697/698 (1 pre-existing skip), `MeshWeaver.GitSync.Test` 160/160,
`MeshWeaver.Query.Test` 388/388, all after the truncation-probe commit.

PR #1359, opened 2026-08-12
fix: batch the indexing and import progress logs — the O(n²) shape #1341 left behind

rbuergi:
Follow-on to #1341, same defect class, two more sites. Root-cause context for #1172 / #1284 is in
#1358 and in https://github.com/Systemorph/MeshWeaver/issues/1172#issuecomment-5273208458.

## The shape

Every `stream.Update` on a node re-serialises that node's **whole content** to compute its patch. So
a writer that appends one line per item to a growing collection serialises 1 + 2 + … + n copies of a
growing document — **O(n²)** CPU and allocation. That class is what put memex-cloud at 51% of CFS
periods throttled, and the process-wide freezes it caused are what both #1284's thread-pool delay and
#1172's routing back-pressure report were measuring.

## 1. `ContentIndexingActivity` — one write per collection, not per file

`ContentIndexingObserver.ReindexCollection` called `ctx.Log` per indexed file **and** per failed file.
The lines are now carried out of the walk alongside each file's failure count (`FileOutcome`) and
written **once** for the whole collection with the closing summary — the same trick
`StaticRepoImporter.WrittenPaths` used in #1341.

`ContentIndexingActivityContext` gains `LogRange(...)`; `Log(...)` remains the single-line call and now
documents that it is one `stream.Update` and must never be used per item.

**Liveness is kept on purpose.** One short heartbeat every 200 files is the only write made *during*
the walk, and it lands while the document is still small (opening line + a few heartbeats), so it
cannot reintroduce the quadratic cost. Accepted trade — the same one taken in #1341: per-collection
detail instead of per-file.

## 2. `GitHubSyncService` — parse failures reported together

`ParseFile` invoked `progress` per unparseable file, and those parses run under `.Merge(8)` — so the
writes were concurrent on one activity node as well as quadratic. `ParseFile` now **returns** the
problem and `ParseSnapshot` reports them in a single entry that still names every skipped file and why.

## Verification

`ReindexActivityWrites_DoNotGrowWithTheNumberOfFiles` asserts the activity node's own revision counter
(`MeshNode.Version`) — never log text, so rewording a message cannot satisfy it:

| | writes |
|---|---|
| **RED before** | 3 files → **7**, 12 files → **16** (1:1 with file count) |
| **GREEN after** | identical for both sizes, and below the file count |

Suites: indexing **32/32**, GitSync **160/160**, Documentation What's New + link integrity **2/2**.
All touched projects clean under `-c Release -warnaserror`.

---

# Design for routing — NOT implemented here

## 🚨 The finding that changes the plan: a delta field alone does not work

I said earlier that `ThreadExecution.PushToResponseMessage` "needs a delta field rather than batching".
Working the design through, **that is not sufficient, and it is worth stating plainly because it is the
intuitive fix**:

`stream.Update` serialises the **entire `MeshNode.Content`** to compute its patch — *regardless of which
field changed*. So adding an append-only `Segments` list, or a small `PendingDelta` field, still
re-serialises the full accumulated `Text` on every write. Per-write cost stays **O(total transcript)**.

This is the same lesson #1341 recorded from the other side: *"the patch OUTPUT was never the problem"*
— `DiffArray` already emits a single O(1) `add` op for an append. The cost is on the **input** side.

Therefore only two levers actually exist:

| Lever | Effect | Cost |
|---|---|---|
| **Write less often** (coarser `Sample` window, consolidate on terminal) | reduces the CONSTANT only — still O(n²) | visibly choppier streaming; a pure bound change, which is the band-aid shape |
| **Move the growing text OFF the node** (satellite) | genuinely **O(n)** | contract + reader change |

So for chat streaming, the satellite shape is not the "nicer" option — it is the *only* one that changes
the asymptotics. I did not implement either; picking is the scope call.

## What the satellite shape would take

Applies identically to `ActivityLog.Messages` and `ThreadMessage.Text`, which is why it retires the whole
class rather than patching five sites:

1. **Segment nodes.** Appends write a NEW child node per batch (`{path}/_Activity/{id}/{seq}` or a
   dedicated `_Log` segment) holding only that batch. Each write is O(batch), never O(total).
   Batching still applies on top — one node per batch, not per line — so node count stays bounded.
2. **The owning node keeps a bounded head** (status, counts, first/last lines) so the common
   "did it succeed?" read stays a single node read.
3. **Readers must compose** — head + ordered segments. This is the real cost and the real risk:
   `ActivityLog.Messages` is read by the GUI activity view, `CompileProgressIndicator`, MCP `get`,
   the compile/import diagnostics paths, and a number of tests. Chat adds the streaming view and
   the transcript-to-model path.
4. **Storage routing already exists** for `_Activity` → the `activities` table; a new segment prefix
   needs one routing entry (see the satellite table map in AGENTS.md).
5. **Retention becomes natural** — old segments can be pruned without rewriting the head, which the
   current unbounded `ImmutableList` cannot do at all.
6. **Resubmit / branch / delete-from-message** stay on the head node, but must cascade-delete or
   re-anchor the segments; that is the piece to design carefully for chat.

Honest assessment: (1)–(2) are small, (3) is the bulk of the work, and chat's (6) is where the
regression risk lives. Doing it for **activities only** first is tractable and low-risk (fewer readers,
no live-typing UX), and it would validate the shape before chat inherits it.

## Deliberately not taken, with the reason

- **`InstanceSyncService.AppendPending`** — one write per changed node into a growing `PendingChanges`.
  Batching it means buffering the change feed, which trades **durability**, not progress granularity: a
  pending entry not yet flushed is lost on a crash, and the push direction has no reconciling sweep to
  recover it (the pull sweep only covers the other direction), so the peer silently drifts until that
  node changes again. Different trade from #1341's — flagging rather than taking it. Mitigating factor:
  the list is drained continuously, so growth is bounded by burst size, not by total node count.
- **Thread heartbeats and `ActivityLogLogger`** — already rate-capped (`Sample`/100 ms throttle), so
  there is nothing to batch. Their residual cost is per-write re-serialisation of an unbounded list,
  i.e. exactly the satellite problem above.


rbuergi:
## CI attribution: the batching is exonerated by the failure's own output

The prior was that batching changed the `MeshNode.Version` sequence and broke a version-arithmetic assertion. The failure output falsifies that directly:

```
After Create: versions = [1]
After all updates: versions = [3, 2, 1]
ObservableAssertionException : … it errored: The input does not contain any JSON tokens.
  Expected the input to start with a valid JSON token, when isFinalBlock is true. Path: $
  at VersionHistoryTest.cs(165)
```

The version list is exactly `[3, 2, 1]` — the count the test expects. Line 165 is the diagnostic loop that **reads back** each enumerated version. Nothing about write counts.

Nor could this PR reach it: `MeshWeaver.ContentCollections.Indexing.Graph` is not in `MeshWeaver.Content.Test`'s dependency closure (absent from its `bin/`), and the test runs no indexing and no GitSync.

## Real defect found and fixed

`FileSystemVersionStore.GetVersions` globs `{id}_*.json`, so **the file name is the publication**. `WriteVersion` used `File.WriteAllTextAsync(filePath, …)` — `FileMode.Create` creates the target first and streams the bytes after, so a version is enumerable while it is still 0 bytes. List-then-read inside that window gets `""`, which is precisely the parse error above.

The sibling writer in the same folder already writes temp-then-rename and its comment names this exact hazard; the version store hand-rolled a second, unprotected implementation. Both now go through one `AtomicFileWrite` helper — the duplication is what let the defect exist in one while the other documented it.

**RED before, GREEN after, proven non-vacuous:** the new `FileSystemVersionStoreAtomicityTest` fails against the unfixed store naming ~20 distinct versions enumerated at 0 bytes (`node_12.json` … `node_38.json`), and passes with the fix. It asserts it made at least one observation, and a deterministic full-history round-trip runs regardless, so a timing probe that samples nothing still cannot pass it vacuously.

`MeshWeaver.Hosting.Test` 149/149, `MeshWeaver.Content.Test` 261/261 (including the previously failing test), Documentation What's New + link integrity 2/2. Current `main` merged. Not merging — serial merge train.

rbuergi:
## The original blocker is fixed — shard 4 is green. Shard 2 is a different, known defect (#890)

Run [31649727371](https://github.com/Systemorph/MeshWeaver/actions/runs/31649727371):

| shard | before | now |
|---|---|---|
| **4** (`VersionHistoryTest`) | **fail** | **pass** |
| 0, 1, 3, 5 | pass | pass |
| 2 | pass | **fail** |

So the atomic-publication fix does what it claims. Shard 2's failure is unrelated and I am **not** re-running it, because doing so would bury a genuine recurrence.

**It is [#890](https://github.com/Systemorph/MeshWeaver/issues/890)** — "process-wide compile failures with bare NRE message — every compile in the shard (incl. BrokenType) fails 'Object reference not set…'" — closed on 2026-08-09 after a sweep found no recurrence in a window that ended 2026-08-08. Reopened with the new evidence.

The recurrence is textbook: all eight compiles in `MeshWeaver.Hosting.Monolith.Test (part 2/2)` fail with a bare NRE, `BrokenType` included, then the project hits `exit=124 TIMEOUT`. And this time the diagnostics carried a stack, which the issue had explicitly given up on getting:

```
at Microsoft.CodeAnalysis.CSharp.Symbols.NamedTypeSymbol.Microsoft.Cci.ITypeDefinitionMember.get_ContainingTypeDefinition()
at Microsoft.Cci.MetadataWriter.GetConsolidatedTypeParameters(ITypeDefinition typeDef)
…
at Microsoft.CodeAnalysis.Compilation.SerializeToPeStream(…)
```

Entirely inside Roslyn's PE metadata writer during `Emit`, with no MeshWeaver frame on the stack — which also falsifies #890's premise that the NRE strikes *pre*-Roslyn.

Why it is not this PR: the diff is progress-log batching plus a change to how a text file is published to disk (temp + rename rather than writing the final name directly). Neither can reach Roslyn's emitter. The same eight tests pass locally on this exact tree in 18 s, every other shard is green, and PRs #1330 and #1343 — same `main` base, same runner pool, without this diff — passed shard 2 in the same window.

Leaving this for a human call rather than pressing re-run.

Issue #1348, opened 2026-08-12
Zero-disruption rolls: bake-gated silo generations, activity-pinned grains


rbuergi:
Roland's policy for platform rolls — the aim is to **always serve fast and never disturb the user**:

1. **As long as there is activity on a grain, we don't tear it down.** Deactivation is for idle grains only; a roll must not evict pages people are looking at.
2. **We bake, and until baking is finished, every grain allocation goes to an OLD silo.** The new generation earns traffic by having every assembly warm; the old generation is warm by definition.
3. **If the bake fails, we never switch away from the old silos.** A broken new generation is a non-event for users, not an outage — fail closed toward OLD, never toward cold/broken NEW.
4. **Only a successful and finished bake opens the new generation** — from then on, new grains are allocated on the new silos. Old silos drain naturally (rule 1) and retire.
5. **A recycled grain re-hashes** onto the new generation — the explicit early-move. Everything else transitions smoothly: activity-pinned until idle, re-placed on next activation.

## Gap analysis (memex, 2026-08-12)

- The roll is a single-Deployment **rolling update**: the old pod dies grains-and-all — rule 1 has no mechanism at all (preStop/grace only softens the kill).
- The **image bake contributes zero** (`alreadyBaked=0`, #1347): the new pod boots cold and the 76 s prewarm sweep races the first visitors — rules 2–4 exist only as the readiness gate, which is currently **off**, and even on it fails toward "no pod ready" rather than "old generation keeps serving".
- The sweep warms the most user-visible types (the `Store/Coupon → Store/Order → Store/Plugin` cycle trio) **last** (#1347).
- Observed same day: /SST/Subscribe "took forever with many redirects" — cold hub activations through the paywall redirect chain after a roll + syncs.

## Shape of the implementation

Two-generation topology during a roll (old ReplicaSet kept serving, not scaled to zero), placement director that routes NEW activations by bake state (old until bake-success, new after), activity-based deactivation as the only drain, and `recycle` as the explicit re-hash. The readiness gate (DynamicTypePreWarmer) becomes the bake-success signal that OPENS the new generation rather than a boolean that blocks the only pod.

Policy is documented in `Doc/Architecture/ReleaseToProductionPipeline` → "The roll policy — serve fast, never disturb" (PR to follow).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

rbuergi:
## Verdict: SUPERSEDED — closing

[...]

PR #1366, opened 2026-08-13
Bake takeover: cluster membership, not a 10-minute clock (#1355)

### LINKED ISSUE #1355: NodeType bake single-silo guard: the lease works, but failover is a 10-minute timer where Orleans membership already knows
## The guard EXISTS — and it engaged in production today

`NodeTypeBakeLease` + `BakeCoordination` (`src/MeshWeaver.Hosting/NodeTypeBakeLease.cs`) already
implement "ONE POD BAKES; the rest follow the share". It is invoked from
`DynamicTypePreWarmer.BakeOrFollow`, not from `NodeTypeBatchBake.cs` or
`DynamicTypePreWarmerHostedService.cs` — which is why grepping those two for
leader/primary/elected finds nothing.

It is **live in every deployed configuration**, not just on paper. `BakeCoordination` is registered
whenever `Deployment:Backend=Filesystem`, and that is what AKS
(`deploy/aks/values.aks.yaml`), ACA (`deploy/aca/memex-portal/memex-portal.bicep`), compose and
compose-ha all set. The `Azure` branch of `Memex.Portal.Distributed/Program.cs` — the only path that
does NOT register it — is selected by no deployment.

Proof from the current memex pod (2026-08-12):

```
Bake lease ACQUIRED for framework 22825f59 by memex-portal-deployment-77c58c4469-g2ngb
    — this pod bakes; others follow
```

Acquisition itself is sound: an atomic `FileMode.CreateNew` on the shared `/data/assembly-cache`
volume, keyed per framework version, held for the bake's lifetime by `Observable.Using` and released
on completion/error/unsubscribe. Exactly one process wins.

## What is actually weak

Three residual weaknesses, in descending order of how much they matter. None has been observed to
bite, and all three portals currently run **one** replica, so this is not urgent — but "single silo"
is currently *best-effort*, not a guarantee, and it is worth being explicit about which.

**1. Failover is a fixed 10-minute timer, where Orleans already knows the answer.**
`StaleAfter = 10 min` + `FollowPollInterval = 60 s`: if the baking pod dies, the fleet waits out a
budget before anyone steals the lease. That budget is a stand-in for "is the holder still alive?",
and Orleans cluster membership answers that authoritatively and immediately. The lease already
stamps its holder as `Environment.MachineName` — the pod name — so the holder IS resolvable against
membership. Replacing the staleness clock with a membership check (or moving the election onto a
fixed-key grain, whose single activation *is* an election and whose placement survives a silo
death) makes takeover level-triggered on the truth instead of on a clock.

**2. The liveness signal is an SMB file timestamp.** `File.SetLastWriteTimeUtc` /
`File.GetLastWriteTimeUtc` on Azure Files. Metadata caching can make a heartbeat read stale when it
is not, or fresh when the holder is gone. This is the weakest link in the mechanism and it is
invisible when it misfires — the symptom would be two pods baking, which today only shows up as
duplicated work in the logs.

**3. Every I/O error fails OPEN.** The `catch` in `TryAcquire` returns a lease, i.e. "you may bake".
That is a deliberate and defensible trade ("duplicate work is a cost, a fleet that never compiles is
an outage") — but it is exactly where the single-silo property stops being a guarantee. If it must
become one, that branch is the place, and the replacement has to be something that cannot fail
ambiguously.

## Why concurrent bakes are harmful here (beyond wasted CPU)

- Concurrent cold compiles of the SAME type are the storm the sequential ordered sweep exists to
  prevent — four of them on memex (2026-07-28 04:05) dropped six plugin roots to the "did not
  settle" overlay and needed a scale-to-zero.
- Each compile loads a collectible ALC, and `MessageHubGrain.OnDeactivateAsync` carries a documented
  use-after-unload gap for a grain deactivating on a LIVE silo. More concurrent ALC churn is more
  exposure to that class — a **mitigation**, emphatically not a fix; the fix is the per-ALC
  in-flight tracking named in that comment (#613).
- There is also a real correctness hazard, not just waste: two bakers both run
  `NodeTypeBatchBake.WriteStamp`, which reads the freshest node, applies the compile stamp and
  writes at `NextVersion`. The monotonic write guard makes the loser's stamp bounce (it already logs
  `compile-state stamp … was REFUSED`), so the bytes land on the share while the record does not
  name them — the type stays pending and is re-baked next boot. Self-healing, but it is precisely
  the "the bake contributed nothing" shape from #1347.

## Suggested scope

Keep the file lease as the fallback for hosts with no cluster (monolith, tests, dev) and put the
election on Orleans where a cluster exists. Note the overlap with #1348's "bake-gated silo
generations" — an election grain should stay self-contained so the two compose rather than collide.

rbuergi:
Closes #1355.

## The election is kept — only the three named weaknesses change

`NodeTypeBakeLease` + `BakeCoordination` are production-proven (memex: `Bake lease ACQUIRED for framework 22825f59 by memex-portal-deployment-77c58c4469-g2ngb — this pod bakes; others follow`). The atomic `CreateNew`, the per-framework key, and the `Observable.Using` lifetime are untouched.

## 1. Failover was a timer for something Orleans already knows

`StaleAfter = 10 min` was a stand-in for *"is the holder alive?"* — and cluster membership answers that authoritatively and immediately, with probes, indirect probes and a membership table already running for exactly this question. The holder now stamps its `IClusterMembership.LocalIdentity` into the lease, so the answer is **looked up, not guessed at**:

| Membership says | Result |
|---|---|
| **Gone** | take over **immediately** — no staleness budget to wait out |
| **Alive** | **never** take over, however old the heartbeat looks |
| **Unknown** | fall back to the `StaleAfter` clock |

`Unknown` is now the clock's **only** remaining job: no cluster at all (monolith, tests, dev), a pre-membership lease with no identity stamped, or an identity membership cannot resolve.

> **Absence from the snapshot is `Unknown`, never `Gone`.** Orleans keeps departed silos as `Dead` until the defunct-cleanup window elapses, so a silo that really died IS in the snapshot and resolves to Gone. One missing entirely usually means our snapshot is not hydrated — and reading that as "gone" on a freshly-started silo would evict a perfectly live baker.

**Identity, not name.** `SiloOptions.SiloName` is set in none of this repo's deployments, so a pod-name → member lookup would be guesswork. `LocalIdentity` is opaque by contract: the implementation stamps it and the same implementation resolves it (`SiloAddress.ToParsableString()` ↔ `FromParsableString`), and a string it cannot resolve is `Unknown`, never `Gone`.

## 2. Liveness was an SMB file timestamp

The heartbeat is now a **write, not a touch** — the instant lives in the lease file's **content**. Azure Files can serve cached metadata, and a falsely-stale `GetLastWriteTimeUtc` is exactly the misreading that puts two pods on one compile. Membership already makes the timestamp non-authoritative; moving the instant into the bytes closes it on the no-cluster path too.

A partial read is possible in principle (the beat truncates and rewrites in place) and degrades safely: an unparseable instant falls back to the file time, an unparseable identity falls back to the clock — both strictly more conservative than the truth.

## 3. Every I/O error failed open

That blanket rule is precisely where "single silo" stopped being a guarantee. It is now split by **what the failure actually tells you**:

- **No coordination substrate** (the shared directory cannot be created; no usable lease path) → **bake**, unchanged. There is no fleet to coordinate with, and denying work here would turn a volume blip into a fleet that never compiles.
- **Substrate works, holder indeterminate** (a takeover we decided on whose write failed) → **follow**. Following is not "never compile": the follower re-probes and re-attempts the lease every `FollowPollInterval`, so being wrong costs one poll — while baking costs the concurrent-compile storm the class exists to prevent.

A lease that **vanishes** between the create attempt and the read is now handled as the observed state transition it is — one re-attempt of the atomic create, then follow — instead of falling through to "bake without coordination". An `IOException` from `CreateNew` is also no longer treated as unambiguous: it covers both "already exists" and a real I/O failure, and the two are now separated by asking whether the file is actually there.

## The `WriteStamp` race: defended, not closed — and now written down

`NodeTypeBatchBake.WriteStamp` reads the freshest node, stamps, and writes at `NextVersion`; the monotonic guard bounces the loser (`compile-state stamp … was REFUSED`), so the bytes land on the share while the record does not name them.

**The election removes one of the two writers — a second baker — and cannot remove the other.** The other is the type's **own per-node hub**, which stamps compile state from the activation-driven path, the sources watcher and the release watcher, and is legitimately live while a batch bake runs. So this is not an election problem, and a single baker does not make it go away.

It is defended because the loser's outcome is already the correct one: the bytes are durable and content-addressed, the record is left pending, and the level-triggered probe — which asks the **store**, not the record — re-bakes and re-stamps on the next pass. A compare-and-re-apply here would buy one saved recompile at the cost of a second write path into a record this driver does not own, which is the shape that produced racing writers in the first place. The reasoning now lives at the call site and in `NodeTypeCompilation.md` rather than in an issue.

## Scope

`IClusterMembership` lives in `MeshWeaver.Mesh.Contract` because `MeshWeaver.Hosting` has **no Orleans reference** (Orleans is confined to `MeshWeaver.Hosting.Orleans` / `.Connection.Orleans` / the Distributed portal). `OrleansClusterMembership` is registered **silo-side** by `ConfigureMeshWeaverServer` — deliberately *not* by `AddOrleansMeshServices`, which also runs on the Orleans **client** host where `IClusterMembershipService` is not in the container.

The election stays self-contained. **#1348 (bake-gated silo generations) is not touched** — the overlap is that both want to reason about which silos are on which framework; keeping the election a plain "who bakes" decision is what lets the two compose rather than collide.

## Verification

**Fail-before / pass-after.** With membership neutered to `Unknown` and the content-borne instant removed — i.e. the pre-#1355 regime behind the new surface — exactly **four** tests fail and the **12** pre-existing invariants pass:

```
Failed  LiveHolder_IsNeverTakenOver_EvenWhenTheHeartbeatLooksAncient
Failed  APreMembershipLease_StillParses_AndIsGovernedByTheClock
Failed  TheHeartbeatInstantComesFromTheContent_NotTheFileTimestamp
Failed  DeadHolder_IsTakenOverImmediately_WithoutWaitingOutTheClock
Failed! - Failed: 4, Passed: 12, Total: 16
```

Restored: 16/16, and `NodeTypeBakeLeaseTest` + `DynamicTypePreWarmerTest` 26/26.

Release + `-warnaserror` clean: `MeshWeaver.Mesh.Contract`, `MeshWeaver.Hosting`, `MeshWeaver.Hosting.Orleans`, `MeshWeaver.Hosting.Monolith.Test`, `MeshWeaver.Hosting.Orleans.Test`, `MeshWeaver.Documentation.Test`. `WhatsNewEntryIntegrityTest` + `DocumentationLinkIntegrityTest` green.

Both regimes are tested, because the fallback is what every monolith and test host actually runs — including that a **pre-membership two-field lease still parses** and is governed by the clock, so a rollout does not turn every in-flight lease into an unreadable one.


rbuergi:
## Rebased onto the post-lease `main` — the mechanism moved, the point did not

`main` deleted the substrate this PR was written against. d5426f566 (*"the protocol is THE coordination — lease deleted, default on"*) removed `NodeTypeBakeLease`, `BakeCoordination` and the 60 s share-poll follower outright and made the Build-node protocol the default, so both of this branch's core files came back from the merge as **modify/delete** conflicts.

Neither side was taken wholesale. Resolved per weakness from #1355:

| #1355 weakness | On the new substrate | Resolution |
|---|---|---|
| **1. Failover is a 10-minute timer** | **Still live** — `BuildNodeType.ClaimStaleAfter` is the same 10 min stand-in for "is the holder alive?", just on node state instead of a file | **Ported** onto `Arbitrate` |
| **2. Liveness is an SMB file timestamp** | **Moot** — the heartbeat is already a `BuildState` field written via `stream.Update`; the Azure Files caching failure left with the lease | not ported (nothing to move into "content" — there is no file) |
| **3. Every I/O error fails open** | **Moot** — there is no lease path and no substrate that can be unusable, so "no substrate → bake / holder indeterminate → follow" has no referent | not ported (82b754735's review fix resolved a lease-file condition that no longer exists) |

### The rule, now on the claim arbiter

The candidate stamps `IClusterMembership.LocalIdentity` into its `BuildClaimRequest`; the arbiter copies it to `BuildState.ClaimedByIdentity` on grant and asks membership rather than aging a timestamp:

- **Gone** → take over **immediately**, no budget to wait out
- **Alive** → **never** take over, however old the heartbeat looks
- **Unknown** → the `ClaimStaleAfter` clock — now *only* for hosts with no cluster (monolith, test, dev, the Orleans client) and for an identity that cannot be resolved

Absence from the snapshot stays **Unknown, never Gone**.

`IClusterMembership` and `OrleansClusterMembership` are unchanged from this branch and merged without conflict — they now have a consumer. `DynamicTypePreWarmer.BakeOrFollow` and the two file deletions were taken from `main` unchanged.

### The `WriteStamp` race is still DEFENDED, not closed

Unchanged on purpose, and the comment survived the merge intact — only the name of the mechanism that excludes the *second* baker changed (the claim arbiter, not the lease). The other writer is still the type's own per-node hub, legitimately live during a batch bake, so the read-modify-write at `NextVersion` stays racy by construction and the monotonic guard stays what makes it safe.

### Verified, fail-before included

`main`'s four existing pure `Arbitrate` tests pass **verbatim** — they pass no membership, so they pin the fallback. Four new cases pin the rule that supersedes it, each written so the clock cannot satisfy it: the gone holder's heartbeat is **one second** old, the live holder's is **an hour past** `ClaimStaleAfter`. Neutering the membership consultation to a constant `Unknown` fails **exactly those two** and leaves the six clock-governed ones green.

`BuildCoordinationTest` 10/10 · 65/65 over `DynamicTypePreWarmerTest` + `NodeTypeBakeGateStateTest` + `NodeTypeBatchBakeDiscoveryInvariantTest` + `BuildCoordinationTest` (main's own prewarmer/gate suites, run to prove the side taken from main is intact) · `MeshWeaver.Documentation.Test` 20/20 · `-c Release -warnaserror` clean.

`git diff origin/main...HEAD --stat` is **503 insertions / 29 deletions** across 11 files, and every one of those 29 removed lines is a line deliberately rewritten — no incidental revert of anyone's merged work.

Issue #464, opened 2026-07-13
Runtime-compile pipeline: self-update leaves stale NodeType assemblies, CompileWatcher wedges permanently, activation-fault loop surfaces raw Orleans rejection


rosalie1984:
# Runtime-compile pipeline: platform self-update leaves stale NodeType assemblies, CompileWatcher can wedge permanently, and activation-fault loops surface raw Orleans rejections to the GUI

## Incident

memex portal (AKS `memex` namespace), 2026-07-13/14, `UWDeepfield` space (mesh-compiled plugin). After the continuous self-update rolled a main where the ContentCollections surface went reactive (`GetFiles(path)` / `GetFolders(path)` → pooled `IObservable`, one argument; `GetContentAsync`/`DeleteFileAsync`/`DeleteFolderAsync`/`GetCollectionAsync` removed in favour of `GetContent`/`DeleteFile`/`DeleteFolder`/`GetCollection`), every NodeType in the space stopped compiling — their shared compat source (`UWDeepfield/Source/DeepfieldMeshCompat`) still used the old two-argument `await foreach` shape.

User-visible symptom on every typed node (e.g. `UWDeepfield/Attention`, `UWDeepfield/Clients`):

```
Delivery to 'UWDeepfield/Clients' failed: Forwarding failed: tried to forward message
Request [S127.0.0.1:11111:… sys.client/hosted-…]→[… messagehub/UWDeepfield/Clients]
MeshWeaver.Connection.Orleans.IMessageHubGrain.DeliverMessage(…) #… [ForwardCount=2]
for 2 times after "DeactivateOnIdle was called." to invalid activation. Rejecting now.
```

plus an endless "Subscribing…" spinner on the resubscribing layout areas.

Causal chain: broken NodeType compile → instance grain cannot materialize a hub config → activation faults instantly → `TryDeactivateOnIdle` (retry-on-next-access) → every delivery (and every one of `RoutingGrain.DeliverToGrainWithRetry`'s 6 transient retries) lands in a deactivation window → Orleans forwards twice, then rejects → the raw rejection text is NACKed to the GUI.

The plugin-side fix (rebridging the compat shim onto the new reactive surface) is applied on-mesh and verified: `UwAttention`, `UwLossWatch`, `UwDecisionLog`, `TreatySubmission` rebuilt green and `UWDeepfield/Attention` resolves again. The remaining NodeTypes cannot be rebuilt because of the framework defects below; only a pod restart clears them.

[...]

PR #1367, opened 2026-08-13
fix: a faulted synced query is evicted, not replayed forever (#1316)

rbuergi:
Closes #1316.

## What the limit actually was

Not a query that was slow, and not a connection another operation was holding. The
Npgsql failure in the report is at **`PoolingDataSource.OpenNewConnector` → `NpgsqlConnector.ConnectAsync`** — the physical connect
(TCP/TLS/auth) exceeding its 15 s budget, which Npgsql only reaches when the pool is *below*
`MaxPoolSize`. So the pool was not exhausted; one handshake was slow.

The bound that was actually reached is a **cache with no eviction**. `MeshNodeStreamCache._queries`
holds each synced query as `Replay(1).AutoConnect(1)` in a dictionary that never expires an entry:

- `ReplaySubject` (backing `Replay(1)`) **latches** a terminal and replays it to every later subscriber,
- `AutoConnect(1)` connects once and **never reconnects**,
- the dictionary entry lives for the life of the process.

So the transient 15 s connect timeout became a **permanent** per-id outage clearable only by
restarting the pod. That is why the report shows the watcher re-establishing and failing again and
again: every re-establish re-subscribed to an instantly-replayed terminal.

## Why it mattered more than a log line

The poisoned id was `nodetype-sources:Edu/Module` — the NodeType's own source set, read through
`NodeSources.GetSources`. With it dead, `NodeTypeDefinition.CurrentSourceVersions` could never be
written, and a **null** snapshot classifies as a gating `PreWarmStatus.CompileError`
(`DynamicTypePreWarmer.ClassifyCompileFailure`). A few slow seconds against the database therefore
blocked the rollout of a type whose sources had never been read at all.

That null-snapshot classification is deliberately fail-closed ("a real regression cannot hide behind
'not seeded'") and is **left as-is** — this change removes what was making the snapshot null. Both
were the "two distinct defects" 58a1ceb6b named when it closed #1310/#1312/#1313 and left #1316 open.

## The fix

The per-**path** sibling cache in this same file already learned this from this exact error class —
`EvictFaultedEntry`: *"an Npgsql connect failure fell through both predicates and replayed forever
until a manual recycle."* The **query** cache never got the treatment. This is that discipline, one
dictionary over.

- `EvictFaultedQuery` drops the entry **pair-exact** (never a chain another caller already replaced),
  so the next `GetQuery(id, …)` builds a fresh `SyncedQueryMeshNodes` with fresh provider subscriptions.
[...]

Issue #455, opened 2026-07-13
Render timestamps in each viewer's local time zone (store UTC, display per-user)


sierragolflima:
## Type

Modification / feature request (not a defect). Timestamp **storage stays UTC and is
unchanged**; this adds a per-user **display** conversion so each viewer sees timestamps in
their own local time, correctly, wherever they are.

## Motivation

Timestamps in the portal (e.g. atioz.meshweaver.cloud) currently render in UTC — 1–2 hours
behind Swiss wall-clock time (CET/CEST). Today the display code calls `.ToLocalTime()` /
`.LocalDateTime`, but under Blazor Server "local" resolves to the **server process** zone,
and the deployment container runs UTC, so the conversion is a no-op.

The user base is in Switzerland today but may operate from other regions (e.g. the US) in
future. A single fixed display zone would then show Swiss time to a US user. The goal is a
solution that is stable across **all** time zones: correctness follows the viewer, not the
deployment or the server.

## Proposal

Store UTC (unchanged); resolve the display zone **per user**; convert through one central
seam. Named IANA zones only — never fixed offsets — so DST is automatic and per-region
(US and EU DST switch on different dates).

1. **Source of truth — a `TimeZoneId` (IANA) field on the User node** (`UserNodeType.cs`),
   e.g. `America/New_York`, `Europe/Zurich`. Mesh data: auditable, cross-device, editable
   in user settings, deterministically testable, and available anywhere the user identity
   is (including server-side/hub render paths that have no browser, via `AccessContext`).

2. **Default population by browser detection** — when the field is empty (first sign-in),
   detect `Intl.DateTimeFormat().resolvedOptions().timeZone` via JS interop in
   `OnAfterRenderAsync` (the sanctioned DOM-side async carve-out) and write it **once** to
   the profile. Fallback chain: configured `Display:TimeZone` default → UTC. "Write only if
   empty" prevents a travelling user's device from overwriting their chosen home zone; an
   explicit setting always wins.

3. **One central seam** `ToDisplayTime(DateTimeOffset)` (+ a small read-only timestamp
   format helper): resolves the current user's `TimeZoneId` (`AccessContext` → profile,
   cached per circuit in an **instance** scoped service — not static), then
   `TimeZoneInfo.ConvertTime(utc, zone)`. Synchronous; no async on the render path.

4. **Route every render site through the seam** (replaces the current server-local reads):

   | File | Line | Current |
   |---|---|---|
[...]

PR #1370, opened 2026-08-13
Compiled code is filed under the name it is looked up by (#1368)

### LINKED ISSUE #1368: FrameworkStaleProactiveRebuildTest fails on three unrelated PRs while main stays green — a load-widened race blocking the merge train
`MeshWeaver.Hosting.Monolith.Test.FrameworkStaleProactiveRebuildTest.FrameworkStaleNodeType_ProactivelyRebuilds_WithoutInstanceActivation` has failed on **three unrelated PRs** within a few hours, while **main is green** across every merge in the same window.

| PR | subject | shares code with the test? |
|---|---|---|
| #1359 | batch indexing/import progress writes | no |
| #1364 | silent-read `ShuttingDown` NACK | no |
| #1366 | bake lease membership takeover | adjacent (bake), plausible |

Only #1366 is even arguably related. Three independent branches hitting the same assertion is not three coincidences.

Two more single sightings in the same window, listed in case they share a cause rather than as separate reports:
- `MeshWeaver.Graph.Test.LayoutAreaRetrievalTest.LayoutAreasUnifiedReference_MatchesTheTypedRequest` (#1364)
- `MeshWeaver.Hosting.Monolith.Test.StaleActivationDurableFirstSeedTest.ReactivationViaRead_ServesDurableTruth_AndWritesNothingBack` (#1325)

## Neither test is new

`FrameworkStaleProactiveRebuildTest` came in with #464 (`e0e8ab91b`); `StaleActivationDurableFirstSeedTest` with `af7836c44`. So this is a pre-existing test newly destabilised, not a test shipped broken.

## The leading hypothesis, stated so it can be falsified

**CI was running ~7 concurrent workflows during this window** (an unusually deep merge train). A test whose name is *"proactively rebuilds **without** instance activation"* is asserting that something happens on its own within some bound — exactly the shape that fails first when runners are contended. The same load-widened-window pattern showed up twice elsewhere today.

If that is right, the defect is the test's dependence on a bound rather than on a condition, and the cure is to wait on the actual signal (`stream.Where(...).FirstAsync().Timeout(...)`) rather than on elapsed time.

🚨 **What must not happen:** do not raise the bound, do not add a retry, and **do not re-run to see whether it passes**. The repo's standing rule is that flakes are real races; a test that fails under load and passes when idle is reporting something true about the code or about itself. Establish which before changing anything.

## Two candidate roots worth checking first

1. **The test genuinely races** — a bounded wait rather than a condition wait. Cheapest to check, and the most likely given the name.
2. **Today's bake changes moved the ground under it.** #1357 retired the pre-run bake image entirely and #1361 changed the readiness gate to distinguish a faulted enumeration from an empty one. `FrameworkStaleProactiveRebuild` is squarely in that area. Main passing does not clear this — main may simply be winning the race on an idle runner.

`DOTNET_PROCESSOR_COUNT=4` is known to reproduce CI-shaped races on the maintainer's macOS host and is the obvious first experiment.

## Why it matters now

This is currently the **binding constraint on the merge train** — three otherwise-ready PRs are held by it, and every re-queue costs a full CI cycle and re-samples the same race.

rbuergi:
Closes #1368.

## Neither hypothesis in the issue survived

**Not a bounded wait.** The test already waits on a condition, not on elapsed time. The 90 s expiry was a symptom: the node received exactly **one** emission in the whole window — byte-identical on all three PRs (`Last of 1 emission(s)`, `LastCompiledVersion = 4`, `Version = 8`). Nothing was late. The framework-stale kickoff looked at the state, logged

```
Framework-stale kickoff SKIPPED … the assembly store already holds a build for
the live framework — nothing to rebuild
```

and, having consumed its `Take(1)`, never looked again. There is no bound here to widen.

**Not today's bake work.** #1357 / #1361 / #1363 / #1343 are all innocent. The two commits that matter are `5c86076bb` (2026-07-28, the store probe) and #145 (2026-07-02, the defect).

## Two writers, one field, different values

`LastCompiledVersion` is one half of the `IAssemblyStore` key `(nodeTypePath, version)`. Two places write it:

| writer | value | correct? |
|---|---|---|
| `NodeTypeCompilationHelpers.ApplyCompileSuccess` | `result.Version` — the version the upload used | ✅ and its own comment says why it must be: *"A different version here would point activation at a store key that has no bytes"* |
| `NodeTypeContractHandler` (`GetCompilationPathRequest` write-back) | `curr.Version` — the node's version at write-back time | ❌ since #145 |

The node has advanced past the upload by write-back time (every compile-status write bumps it), and on a hydrate that version was never the upload version at all. Whichever landed last won, so whether a compile left a valid record was decided by **scheduling** — which is exactly why contention moved it and a quiet `main` kept producing the other outcome.

## The test was the detector, wired backwards

Since `5c86076bb` the kickoff asks the store before rebuilding, keyed by `LastCompiledVersion`:

- record **correct** → probe hits the real bytes → kickoff rightly declines → **test RED**
- record **corrupt** → probe misses an empty key → kickoff fires → **test GREEN**

It was green only when the record was broken. It is the only test that reads this field as a store key and asserts on the consequence, which is why it was over-represented among the failures.

## What it costs in production

Every reader uses this field as the store key: activation (`MeshDataSource`, `NodeTypeEnrichmentHelpers`) — which on a miss falls back to the default config, no `MeshNodeReference` reducer, so **the instance page renders empty** — the MCP compile tool, and `NodeTypeBakeStatus.HasBytes`, **the pre-update bake gate**, which files a genuinely baked type as having no bytes.

## The fix

Both writers now record the version the bytes were filed under. `BuildResponseFromLocal` echoes the store key it resolved with instead of the live node version, and the write-back takes the version off the response — which the contract already defines as *"echoed so the consumer's cache key matches what was actually compiled"*.

## The test

Its fixture could not express the state it claimed. A real self-update changes the framework tag the store key embeds, so record and store go stale **together**; forging only the record leaves genuine live-framework bytes on the volume, and declining to rebuild is then the *correct* answer. Both directions are now pinned, each arranging the store to match the record it forges:

- record stale + **no** live-framework bytes ⇒ rebuilds (the #464 defect)
- record stale + live-framework bytes ⇒ declines (the recompile-storm guard — previously unpinned, and the behaviour this test had been calling a failure)

Both also assert the record's store key names bytes that exist: the regression pin for the defect above.

## Verification

- **fail-before**, unconstrained, 3 s: `LastCompiledVersion=6 must name assembly-store bytes that exist; the record points at '…/v4-…dll'`
- **pass-after**: green unconstrained *and* under `DOTNET_PROCESSOR_COUNT=4`
- the **old** test under `DOTNET_PROCESSOR_COUNT=4` reproduced CI byte-for-byte before any change
- regression sweep `Compile|NodeType|SelfHeal|PreWarm|Bake|StaleActivation`: **142/142**
- `MeshWeaver.Graph` and `MeshWeaver.Hosting.Monolith.Test` clean under `-c Release -warnaserror`

## The other two sightings do NOT share this root

- **`LayoutAreaRetrievalTest`** on #1364 failed with `the data stream for this reference completed without ever producing a value — its owner is shutting down`. That string is the NACK **#1364 itself introduces**; it is that PR's own diff surfacing on a shutdown race, not shared contention. Passes locally under 4 cores.
- **`StaleActivationDurableFirstSeedTest`** failed on a version count (`expected 2 … found 3` — an extra activation write-back) against a **`Markdown`** node, so this fix cannot touch it. It passed in the sweep; I could not reproduce its failure and make no claim about it.

### DIFF
--- a/src/MeshWeaver.Documentation/Data/WhatsNew/2026-08-13-compiled-code-is-filed-under-the-name-it-is-looked-up-by.md
+++ b/src/MeshWeaver.Documentation/Data/WhatsNew/2026-08-13-compiled-code-is-filed-under-the-name-it-is-looked-up-by.md
@@ -1,0 +1,34 @@
+---
+Name: Compiled code is filed under the name it is looked up by
+Category: Fix
+Description: After compiling code that lives inside the portal, the portal sometimes wrote down the wrong location for the result. Later it looked there, found nothing, and quietly fell back to a blank configuration — so the page rendered empty and the pre-update readiness check called a perfectly good build missing.
+Icon: Sparkle
+Order: -20260813
+---
+
+# Compiled code is filed under the name it is looked up by
+
+Code that lives inside the portal is compiled by the portal itself, and the result is filed away so
+that everything which needs it later — opening a page, activating content, checking before an update
+that everything still builds — can fetch the same result instead of compiling it again.
+
+Filing it away means writing down two things: the compiled code, and a note saying where it was put.
+The note has to match. If it does not, every later lookup goes to an empty shelf.
+
+Two different parts of the portal write that note, and one of them was writing the wrong location. It
+recorded where the content stood *at the moment it finished writing* rather than where the compiled
+result had actually been filed — and those drift apart, because the act of recording the compile
+moves the content on. Which of the two wrote last was decided by nothing more than timing, so the
+same compile could produce a correct note or a broken one on different runs of the same code.
+
+When the broken note won, nothing announced it. The compile had genuinely succeeded, and everything
+you could see said so: the status read fine, the timestamp was fresh, the build was real and sitting
+exactly where it had been put. Only the note pointed elsewhere. What you saw instead was a page for
+that content rendering empty — the portal had looked for the compiled code, not found it, and fallen
+back to a configuration that does nothing, without saying why. The check that runs before a platform
+update was misled the same way: it asked whether each piece of in-portal code had a usable build,
+followed the note, found the empty shelf, and reported a healthy build as missing.
+
+Both writers now record the location the result was actually filed under. A note that cannot be
+trusted to point at the thing it names is worse than no note at all, so this is also now pinned by a
+test that reads the note and insists the shelf it names is not empty.
--- a/src/MeshWeaver.Graph/Configuration/NodeTypeContractHandler.cs
+++ b/src/MeshWeaver.Graph/Configuration/NodeTypeContractHandler.cs
@@ -140,7 +140,7 @@ internal static class NodeTypeContractHandler
                                         hubPath, requestedReleasePath, localPath);
                                     return compilationService.GetConfigurationsFromExistingAssembly(localPath!, hubPath)
                                         .Select(result => new ResolvedResponse(BuildResponseFromLocal(
-                                            hubPath, node, localPath!, release.AssemblyCollection,
+                                            hubPath, releaseVersion, localPath!, release.AssemblyCollection,
                                             release.AssemblyContentPath, result), false));
                                 });
                         });
@@ -176,7 +176,7 @@ internal static class NodeTypeContractHandler
                                 .SelectMany(result => SurfaceActivityLog(
                                     hub, def.LastCompilationActivityPath, result,
                                     res => BuildResponseFromLocal(
-                                        hubPath, node, localPath!, def.LatestAssemblyCollection,
+                                        hubPath, compileVersion, localPath!, def.LatestAssemblyCollection,
                                         def.LatestAssemblyPath, res)))
                                 .Select(response => new ResolvedResponse(response, false));
                         });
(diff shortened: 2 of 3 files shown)


rbuergi:
## The plugin-gate red is a pre-existing `Store/Catalog` flake — measured, not argued

**Answer: option 3.** `Store/Catalog` is flaky and unrelated to this diff. Filed as
Systemorph/MeshWeaver.Plugins#455. Nothing added to `plugin-gate.allow`.

I reproduced the gate locally rather than reasoning from the log — cloned MeshWeaver.Plugins,
staged `Store` + `Essentials` exactly as the workflow does, and ran `mw-plugin-test` in a loop.

### The A/B

Ten consecutive runs each, same idle machine, same staged plugin content:

| condition | failures |
|---|---|
| this PR's diff | **3 / 10** |
| this PR's diff **reverted** (plain `main`) | **3 / 10** |

Identical rate. The failure reproduces on unmodified `main`, so it is not this change.

### It is the same failure CI hit

Local run 4 produced CI's verdict line byte-for-byte:

```
GATE FAILED — tests: Store/Catalog — 1 new failure(s), 0 stale allow entr(ies).
```

and the actual assertion — which the gate log carries and which none of the quoted warnings name:

```
| Paid (live): a card payment INSTALLS, not just entitles | ❌ 🚨 #450: the card payment
  granted the entitlement and copied NOTHING — 'paidbuyer…/PaidPack…/Exercises/Lesson1'
  is not on the mesh |          53/54 passed.
```

Confirmed identical in this PR's CI log (job 94319256886).

### The three warnings are noise — they are in the GREEN runs

This was the hypothesis worth killing first, and it is dead. `#1367`'s **passing** gate run
(job 94316748975) contains:

```
4 '$type':'PurchaseContent'      7 '$type':'StoreContent'      2 '$type':'PluginContent'
2 '$type':'OrderContent'         2 '$type':'InstallManifest'
```

`PurchaseContent` — 4 occurrences in a run that reported `ALL GREEN`, the same count as in my
green local runs. `EnrollmentControlPlane … run failed` and `PackageInstaller … SyncContentFilesRequest`
likewise appear identically in green and red runs (4 occurrences in both). All three are the
loudest thing in the log and none is diagnostic.

So the inversion hypothesis — that this fix makes a type resolve where it previously did not, and
`Store/Catalog` had been passing on the broken behaviour — does not hold: the warning is unchanged
with the fix, without the fix, and in runs that pass.

### Why it flakes

`PaidFulfilmentTests.RequireVisible` polls `MeshQueries.FindNode` — the **query index**, eventually
consistent by design — under a 20 s budget, for a cross-partition copy into a buyer partition that
did not exist a moment earlier. It races the indexer, not the fulfilment. The fix belongs in the
Plugins repo and is to wait on `GetMeshNodeStream(path)` (the authoritative primitive) rather than
to raise the budget; #455 says so explicitly.

### Also in this push

Reviewing my own diff against the inversion hypothesis turned up a real gap in it, now fixed in
`707a62e66`: the first cut took `LastCompiledVersion` from the response whenever the response
carried a parseable version, but `LatestAssemblyPath` immediately above falls back to the
**persisted** value when the producer uploaded nothing (memory:// compile, no configurations, Null
store, unreadable bytes). That paired retained bytes with a key that was never theirs — the same
record-names-an-empty-shelf defect reached the other way round. The path and the version are one
reference and now come from one source.

Re-verified after the change: both `FrameworkStaleProactiveRebuildTest` cases green unconstrained
and under `DOTNET_PROCESSOR_COUNT=4`; `MeshWeaver.Graph` clean under
`-c Release -warnaserror --no-incremental`.

`main` has not moved since this branch was cut (`042ab3f6a`), and the repo's ruleset is
`strict: false`, so there is nothing to rebase onto.

PR #1372, opened 2026-08-13
Restore PDF contents page numbers by measuring the print (#1309)

### LINKED ISSUE #1309: PDF export lost TOC page numbers when QuestPDF was replaced — Chromium implements no part of CSS target-counter
Accepted regression from #1230 / PR #1291, filed so it is tracked rather than quietly forgotten.

## What changed

Replacing QuestPDF with the headless-Chromium renderer reached **full parity on everything except one thing**: the table of contents no longer prints page numbers. Entries, indentation and working links all survive.

Parity verified on rendered PDFs that were rasterised and looked at, not on assertions:

| Feature | Status |
|---|---|
| Cover page (logo, title, tagline, rule, name, website, date) | ✅ full |
| Running header (brand text + logo, continuous rule) | ✅ full |
| Running footer (brand text + `N / M`) | ✅ full |
| Page-break rules (H1/H2/children/explicit `\newpage`) | ✅ full |
| A4 / landscape; header suppressed when blank | ✅ full |
| TOC entries, indentation, working links | ✅ full |
| **TOC page numbers** | ❌ **gone** |
| Table header repeat / heading orphans | 🎁 better than before |

## Why it is not a quick fix

Chromium implements **no part of** CSS `target-counter`, which is the compliant mechanism. There is no flag, no polyfill at print time, and no paged-media fallback that yields a correct number.

A **two-pass render** (measure, then re-render with numbers injected) is the obvious workaround and was **deliberately rejected**: its failure mode is *silently wrong* page numbers when pagination shifts between passes, and a TOC that confidently points at the wrong page is worse than one that omits numbers. If this is revisited, the bar is that it must fail **loudly** — detect a pagination mismatch between passes and refuse rather than print a number it cannot stand behind.

## Where it is already disclosed

The loss is named in the What's New entry (`Category: Fix`), in the export documentation, and in PR #1291's body — so a user meeting it finds an explanation rather than assuming a bug.

## Related, found during the same work and NOT fixed

Two pre-existing defects, both faithfully reproduced rather than silently corrected:

- `DocumentBuilder.ReadInlinesInto` renders `~~strike~~` as **bold + strike** — `DelimiterCount >= 2` also sets `isStrong`. Affects DOCX export too.
- All three export `.csx` templates use `node.Content is MarkdownContent`, which the `ContentAs<T>` rule now bans: a degraded `JsonElement` silently yields an **empty document**.

rbuergi:
Closes #1309.

`#1230` replaced QuestPDF with the headless browser and reached parity on everything but one item: contents entries kept their links and lost the `..... 12` column. Chromium implements no part of CSS `target-counter`, so nothing *in the page* can name the page a heading landed on.

**The print can.** Every contents entry is a link annotation whose `GoTo` destination is the page the browser itself chose — not a model of the pagination but *the* pagination, recorded by the engine that performed it. `TocPageNumbers` reads it back with PdfPig, which is already in this project's graph (via `MeshWeaver.ContentCollections`) and already carried by the licence gate, so **no dependency is added** and `check-licenses.py` is untouched.

## The objection in #1309, answered

#1309 rejected a two-pass because its failure mode is *silently wrong* numbers when the inserted digits reflow the list, and set the bar at failing **loudly**. Both halves:

1. **The digits cannot reflow anything.** The number column is emitted on **both** prints at a **constant** width (`flex: 0 0 40pt` — the same 40pt right-aligned column the document model drew), so the measuring print and the published print are the same layout by construction. Sizing the column to its content (`min-width`, `fit-content`, a width derived from the page count) is precisely how a naive two-pass goes wrong.
2. **That argument is not trusted anyway.** The renderer re-reads the **published** PDF and releases it only if every destination still equals the number printed beside it.

Any refusal — count mismatch, destinations out of document order, a destination not after the contents list, or a verification disagreement — falls back to the numberless list of #1230 and logs the reason with both readings. A contents list that lies about a page is worse than one that is silent about it.

## Why not the other approaches

| Approach | Why not |
|---|---|
| Measure in the page with script (`offsetTop` / content height) | Arithmetic over the *viewport*; the print box is not the viewport. `break-after: avoid`, `orphans`/`widows`, a repeated `<thead>` and unbreakable figures all move a heading to a page no per-page constant predicts — plausible numbers, occasionally off by one, indistinguishable from correct ones by eye. Also needs `script-src` in a document whose CSP is deliberately `default-src 'none'`. |
| Post-process the PDF (stamp numbers into the produced file) | Exact for *measuring*, which is why measuring is done exactly this way. Writing needs a PDF writer that faithfully round-trips an existing document; PdfPig has none, and adding one is a new dependency and a new licence decision. |
| Pure two-pass, no verification | The thing #1309 rejected, and rightly. |

## Cost

One extra browser print, paid **only** by documents that have a contents list (measured: 3.1 s vs 1.4 s for the same document without one). Reading stops at the page that completes the count, so a 50-page export parses two pages and the read-back costs ~1 ms. The read-back is a CPU leaf and goes through `IIoPool`'s `Process` pool — the same pool as the browser — so a burst of exports bounds its parsing as well as its browsers. Cold and reactive throughout: no `Task`, no `Observable.FromAsync`.

## A real bug the refusal check caught

Two headings sharing a title slugified to the **same** anchor, so the second contents entry jumped to the **first** section — in the rendered page and in the PDF's link annotation alike. It predates #1230 and was invisible only because there was no number to expose it; the "destinations must be in document order" check surfaced it as `3, 3, 4, 6, 4, 8, 11`. Fixed at the root: anchors are made unique per document (`overview`, `overview-2`), which is what every markdown renderer does.

## Verification

Verified the way #1230's parity table was — by rendering PDFs and **looking** at them — and then by assertion.

**By eye** (rasterised, real renderer):
- A 10-page report: first entry, a section spanning three pages, a heading landing **mid-page**, and the last entry on the last page. Every printed number matches the section's first page and the running footer.
- A wrapping contents title with duplicate section names: the number sits on the entry's first line; the two `Overview` entries now read 4 and 7.
- A 46-entry contents list spanning **two** pages: numbered correctly across its own page break, body starting at page 4.

**By assertion**: `RendererOutputTests` pins it against an **independent oracle** — the page whose *glyphs* carry the heading — and requires all three to agree: what the list **prints**, where the heading is **drawn**, and where the link **jumps**. An error in the annotation reading therefore cannot also produce the expectation. `DocumentPrintComposerTests` pins the composer contract with no browser installed (column reserved on both prints, constant width, digits the only difference between the two documents, a mismatched list ignored outright).

Local: `MeshWeaver.Markdown.Export.Test` 109/109, `MeshWeaver.Documentation.Test` 20/20, monolith export tests 16/16, `-c Release -warnaserror` clean on the touched projects and their dependents, licence gate green.

## Not included (from #1309's "Related, NOT fixed")

The two pre-existing defects that issue names — `ReadInlinesInto` rendering `~~strike~~` as bold+strike, and the export `.csx` templates using `node.Content is MarkdownContent` — are unrelated to the contents column and are left for their own change so they are not buried in this one.

### DIFF
--- a/src/MeshWeaver.Documentation/Data/Architecture/PixelFaithfulExport.md
+++ b/src/MeshWeaver.Documentation/Data/Architecture/PixelFaithfulExport.md
@@ -163,13 +163,44 @@ Every decision about *what the page looks like* happens in a composer that is pu
 
 The browser tests (`HeadlessChromiumRenderTests`, `RendererOutputTests`, `PdfHeaderLogoTests`, `DeckPixelExportScriptRelayTest`) ask the renderer's own probe and then assert **whichever contract applies to the machine**: with a browser, a real PDF — cover, contents and body on separate pages, the running header on every page but the cover, `N / M` in the footer; without one, a loud, actionable refusal. They are never skipped, and a missing browser can neither hide a regression nor redden a build. CI provisions the same browser build the image ships, as an explicit infrastructure step, so a shard cannot pass by taking the "no browser" branch.
 
-### 🚧 The one thing that did not survive: page numbers in the contents list
+## Page numbers in the contents list: measure the print, then print the measurement
 
-The document model could print `Chapter 3 ......... 12`. **The browser cannot**, and the contents list therefore carries clickable links but no page numbers.
+The contents list prints `Chapter 3 …… 12` again. **CSS still cannot produce that number**, so it is not produced in CSS: it is read out of a printed PDF and put into a second print, which is then read back and checked before it is released. `#1230` shipped without the column and `#1309` restored it this way.
 
-The reason is specific: reading a target's page from CSS needs `target-counter()` from CSS Generated Content for Paged Media, and Chromium implements no part of it — verified directly against the browser this image ships, alongside the features that *do* work (`@page` margin boxes, named pages, `counter(page)` / `counter(pages)`), which is what the rest of the furniture is built on.
+### Why nothing in the page can answer the question
 
-The alternative is a two-pass print: print once, read back from the PDF's link annotations which page each anchor landed on, inject the numbers and print again. It was rejected deliberately — it doubles the cost of every export with a contents list, and its failure mode is *silently wrong page numbers* when the inserted digits reflow the list. An honestly absent column beats a confidently wrong one, and in a PDF the clickable entry is the affordance a reader actually uses.
+Reading a target's page from CSS needs `target-counter()` from CSS Generated Content for Paged Media, and Chromium implements **no part of it** — verified directly against the browser this image ships, alongside the features that *do* work (`@page` margin boxes, named pages, `counter(page)` / `counter(pages)`), which is what the rest of the furniture is built on.
+
+Measuring in the page with script cannot be made right either, and it is worth being precise about why, because it is the approach that looks easiest. A heading's `offsetTop` divided by a content height is arithmetic over the *viewport*, and the print box is not the viewport: `break-after: avoid` moves a heading that would be stranded, `orphans`/`widows` move the line before it, a repeated `<thead>` and an unbreakable figure both consume height that no per-page constant accounts for. The result is a number that is plausible, occasionally off by one, and impossible to distinguish from a correct one by looking at it. (It would also need `script-src` in a print document whose CSP is deliberately `default-src 'none'`.)
+
+### What the PDF already knows
+
+Every contents entry is a **link annotation**, and its `GoTo` destination names the page Chromium itself put that heading on. That is not a model of the pagination — it *is* the pagination, recorded by the engine that performed it. `TocPageNumbers` reads it back with PdfPig, which the platform already carries for content indexing, so no dependency is added and the licence gate is untouched.
+
+Entries are matched to annotations positionally, and each assumption that rests on is checked rather than trusted:
+
+| Assumption | Why it holds | Checked by |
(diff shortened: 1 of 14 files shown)


rbuergi:
Both review findings addressed in 8d18077b2 — and the first one turned out to be a genuine correctness bug, not just a robustness nit.

**`break-inside: avoid` on `.mw-toc-entry`.** Verified rather than assumed: sliding a contents list past a page break in 8pt steps, **10 of 15 layouts printed 15 link annotations for 14 entries** without the rule, and exactly 14 with it. So a fragmented entry really does emit a rect per fragment, `TocPageNumbers` really would see more links than entries, and the list really would lose its numbers to an avoidable refusal.

One detail worth recording, and now in the CSS comment: **reproducing it needs a title of four or more lines.** `orphans` and `widows` both default to 2, so a two- or three-line entry cannot be split at all — every half would breach one of them. My first attempt at a repro used ordinary wrapped (two-line) titles and showed *no* difference across 15 layouts, which would have been a convincing case for calling the rule cosmetic and deleting it. The comment names the condition so the next person does not reach that false conclusion.

**Whitespace.** Right — the assertion I added to the existing PDF test was inconsistent with the new boundary test, which already stripped. Both now go through a shared `Packed()` helper.

Local after the change: `MeshWeaver.Markdown.Export.Test` 110/110, `-c Release -warnaserror` clean.

Issue #1371, opened 2026-08-13
Install writes its completion manifest with zero nodes copied — 'entitled but empty' for AgenticEngineering/Exercises


rbuergi:
Surfaced by education PR #147, and **masked until now** rather than caused by it.

## What happens

`15-install-lands › Install lands — AgenticEngineering › the copies are REAL nodes` fails with:

```
the copy of declared root 'AgenticEngineering/Exercises' must exist
```

The portal log shows the learner receiving **everything except the content**: the entitlement, the `_Access` grant, the ledger row, the home root — and the install **manifest written last**, i.e. the install engine reported completion. Yet there are **no nodes under `…/AgenticEngineering/Exercises`**.

This is the 2026-07-27 "entitled but empty" shape: server-side, in the install engine.

## Why it was invisible until today

Shard 2's `bootstrap` project was failing on an unrelated defect (the harness read the funnel's German labels as English — MeshWeaver#1349, fixed by education#147). **`bootstrap` is a project dependency, so its failure skipped the entire spec suite** — shard 2 had *never once reached* `15-install-lands`.

So a red gate was hiding a second, unrelated red. That is worth noting beyond this bug: a failing dependency silently removed a whole suite from the run, and nothing distinguished "these specs failed" from "these specs never executed". This is the same family as the `tests=skipped`-reads-as-pass problem (MeshWeaver.Plugins#451) and several others found this week.

## Why this is not #1349

Education#147 changes **only** files under `e2e/`. It cannot cause a server-side install to write a manifest without content. #1349's own signature — a paywall that never advances because the harness cannot read German labels — is fixed and verified: shard 3 (`AgenticOffice`, red 3/3 across three prior runs) is now green.

Do not fold this back into #1349.

## Also ruled out: the memory/GC family

The same run shows **zero** `Platform stalled`, **zero** `timed out after`, **zero** `never came live`. The GC-stall family (5 stalls, heap 4702 → 6289 MB, pauses to 12.8 s) that earlier explanations rested on is gone since MeshWeaver#1345, and this failure persists regardless. It is not resource pressure.

## What to establish

1. **Why the install engine writes its completion manifest when no content was copied.** Whatever the copy failure is, reporting success is the more serious half — an install that copies nothing must not be indistinguishable from one that had nothing to copy. (Compare MeshWeaver#1365, which stopped the plugin-gate report fabricating `0 node(s)` counts for the same reason.)
[...]

Issue #1299, opened 2026-08-12
Plugin gate: Reinsurance+SST fail the install unchanged-check on every image, and the idempotence check is nondeterministic on one digest


rbuergi:
Two separate defects in the plugin-gate install check, found while converting hand-built SVG onto framework chart controls (MeshWeaver.Reinsurance#52). Both are **core**, neither belongs to the PR that exposed them.

## 1. A real regression: Reinsurance + SST fail the unchanged-check on every image carrying the controls

`Reinsurance` and `SST` fail the install **unchanged-check** on *every* candidate image that carries the chart controls. The controls themselves (`TowerControl`, `KpiStripControl`, `ComparisonBarsControl`) landed in core on 2026-08-08 for exactly this purpose; the regression window is **08-08 → 08-11**.

This is not the plugin's defect — it reproduces across images independent of the plugin change under test, and it blocks a PR that only replaces SVG strings with the controls core added for that job.

🚨 **Do not resolve this by adding `Reinsurance`/`SST` to `plugin-gate.allow`.** That file is a **shrink-only ratchet**; widening it to unblock a UI change would park a live core defect behind an allow entry and remove the signal that found it. The gate is doing its job — the thing it is reporting is real.

## 2. The `idempotence` check is nondeterministic

Uncovered while trying to attribute defect 1, and worth its own attention because it corrupts every attribution attempt made through this gate:

**Re-running the exact same image digest gives different results** — 2 failures on one run, 3 on the next, with Reinsurance reporting **80 nodes one run and 77 the next**. Same digest, same inputs.

That nondeterminism is what made two earlier attributions wrong (#1205, then #1273 — both blamed, both innocent). Any reasoning of the form "rank images by failure count" is over-fitting a flake. Establishing defect 1 required re-running one digest repeatedly rather than comparing digests to each other.

A gate whose output varies on identical input cannot be used to attribute a failure to a change, which is its entire purpose.

## Suggested order

Fix 2 before 1 if they are independent — while the node count wobbles, any verification of 1 is unfalsifiable. If they share a root (a snapshot read that races, of which this codebase has several this week — see the `QuerySnapshot` quiet-window race fixed in MeshWeaver.Plugins#416), one fix may settle both.

## Provenance

[...]

PR #1380, opened 2026-08-13
fix(#1379): a NodeType definition must not claim the content-type key its instances read

### LINKED ISSUE #1379: A per-node instance serves its NodeType's NodeTypeDefinition as its own Content after a stale-assembly recycle
## Symptom

A freshly created instance of a dynamic NodeType intermittently ends up serving its **NodeType's**
`NodeTypeDefinition` as its own `Content`. Every downstream reader then sees the instance as having
no domain content at all.

Observed on a `Store/Plugin` instance:

```
path  = PaidPackd08a16babe9240f798191a0960ccabc1     ← the instance, correct
type  = Store/Plugin
main  = PaidPackd08a16babe9240f798191a0960ccabc1     ← correct
v1                                                   ← never rewritten
raw   = MeshWeaver.Graph.Configuration.NodeTypeDefinition   ← should be PluginContent
```

- **Both** reads agree: `hub.GetMeshNodeStream(path)` (authoritative, the owning hub) and the synced
  query. So it is not a projection artefact of one transport.
- **Persistent for the whole run** — still wrong 20 s later, and `Version` never advances past the
  create, so nothing rewrote it afterwards.
- `ContentAs<PluginContent>()` therefore returns null (the runtime type's short name differs, so
  `As<T>`'s same-name recovery correctly declines).

## Trigger

It correlates with the instance hub activating while its NodeType is still compiling, followed by
the stale-assembly self-heal recycling it:

```
[COMPILE-TRACE] Slow-path typeStream emission for Store/Plugin (instance=PaidPack…):
    HubConfig=False Status=Compiling Coll=local Path=Store_Plugin/v202-…dll
… Status=Ok  Path=Store_Plugin/v207-…dll
Stale-assembly self-heal: NodeType 'Store/Plugin' published a new build
    ('Store_Plugin/v212-…' supersedes 'Store_Plugin/v207-…') — recycling instance 'PaidPack…'
Stream …: owner PaidPack… rejected our SubscribeRequest while shutting down — resubscribing
```

Note the NodeType publishes **twice within ~50 ms** (v207 then v212), so the instance rebinds
almost immediately after its first bind. Whether that double-publish is itself a defect is worth a
look — removing it would likely close the window.

The enrichment path is the obvious suspect: `NodeTypeEnrichmentHelpers` reads the type node
(`typeStream = meshHub.GetWorkspace().GetMeshNodeStream(nodeType)`) alongside the instance, and the
instance is what comes back carrying the type's definition. Every `typeStream.Update(...)` in that
file targets `nodeType` correctly, so the crossing is more likely on the hydration/read side than on
a write.

## Why it matters beyond a flaky test

`AGENTS.md` notes that a framework-version bump **recompiles every dynamic NodeType**, so this window
opens for all of them at once on a deploy. Anything that reads an instance's typed content to make a
decision in that window gets "the content is absent" — and absence is usually a legal answer, so the
failure is silent. The instance that surfaced this was a paid Store package: the fulfilment read the
declaration, found nothing, and reported "nothing to install", leaving a buyer charged, entitled,
stamped delivered, and holding nothing (Systemorph/MeshWeaver.Plugins#450 / #455).
Systemorph/MeshWeaver.Plugins#456 makes that case fail loudly instead of silently, but it cannot fix
the content itself.

## Reproducer — ~9 s per run, ~5% hit rate

```bash
# 1. build the gate CLI from this repo
dotnet build tools/MeshWeaver.PluginTester/MeshWeaver.PluginTester.csproj -c Release

# 2. stage the vital plugin modules exactly as .github/workflows/dotnet-test.yml does
STAGE=$(mktemp -d)
cp -R <plugins-checkout>/Store <plugins-checkout>/Essentials "$STAGE/"
cp <plugins-checkout>/plugin-gate.allow "$STAGE/"

# 3. loop it — roughly 1 run in 20 fails on the Store/Catalog paid-fulfilment case
dotnet tools/MeshWeaver.PluginTester/bin/Release/net10.0/mw-plugin-test.dll \
    "$STAGE" --allow "$STAGE/plugin-gate.allow"
```

With Systemorph/MeshWeaver.Plugins#456 merged, a hit prints the diagnosis directly:

```
'PaidPack…' carries content that did not read as a package declaration (got 'NodeTypeDefinition')
```

⚠️ `MW_LOG_LEVEL=Information` makes the run 0/25 — the extra logging perturbs the timing enough to
close the window. Reproduce at the default level and add targeted instrumentation instead of raising
the global level.

rbuergi:
Closes #1379.

## What was wrong

A per-node instance intermittently served its NodeType's `NodeTypeDefinition` as its own
`Content` — `Path`, `NodeType` and `MainNode` all correct, `Version` never advancing, persistent
for the rest of the process, and identical on both the authoritative stream and the synced query.

It is a **read-side aliasing, not a write** — which is exactly why the Version never moved, and
why the recycle in the issue's trace is a red herring as a *cause*: it only re-orders two
activations. The stale-assembly self-heal is correct and is untouched here.

`MeshNodeHubFactory` stamps `NodeTypePathHolder` so that the `WithContentType` inside a NodeType's
configuration records its content type under an exact, mesh-unique key (#1299's exact route). For
a NodeType **definition** node it stamped the node's **own path**:

```csharp
var nodeTypePath = string.Equals(enriched.NodeType, MeshNode.NodeTypePath, StringComparison.Ordinal)
    ? enriched.Path        // ← "Store/Plugin"
    : enriched.NodeType;
```

The reasoning in the comment — *"that is the type every instance of it will declare"* — got the
**key** right and the **value** cannot be. A definition node's hub applies `NodeTypeNodeType`'s
configuration, `WithContentType<NodeTypeDefinition>()`, and never loads the compiled assembly of
the type it declares. So the only thing it can publish under `Store/Plugin` is
`NodeTypeDefinition` — while that key answers a different question: every read seam resolves by
`node.NodeType`, so the entry at `Store/Plugin` is what an **instance** of Store/Plugin gets, and
its content is `PluginContent`.

`MeshContentTypeRegistry.Register` is last-writer-wins per key, so a definition hub cold-activating
after an instance overwrote the instance's correct entry. The degrade seams (`MeshNodeStreamCache`,
`MeshNodeTypeSource.ResolveJsonElementContent`, `EnsureTypedContent`) then recovered instance
content through `TryRecoverForNodeType(node.NodeType, …)` and got the definition back — **silently**,
because System.Text.Json ignores members the target does not declare and materialises defaults for
the ones it does, so deserialising into the wrong record *succeeds*.

Downstream, `ContentAs<PluginContent>` correctly declines a differently-named type, the paid
fulfilment read no declaration, and the buyer was charged, entitled, stamped delivered, and holding
nothing.

## The change

- **`MeshNodeHubFactory`** stamps `enriched.NodeType` unconditionally — the type whose
  configuration is actually being applied. Instances and both schema probes already write the
  correct entry under that same key, so the special case only ever published a wrong one. The
  literal `"NodeType"` a definition node now stamps is itself a correct entry: a node whose
  NodeType *is* `"NodeType"` does carry `NodeTypeDefinition` content.
- **`TryRecoverForNodeType`** refuses a type the content's own `$type` contradicts. The bad entry
  is fixed at its writer, but *"the map said so"* must not be enough on its own to reshape content:
  an unresolvable read stays a `JsonElement` that `ContentAs<T>` can still recover and the seams
  already warn loudly about, where a wrongly-typed one is silent and lossy. The comparison is on the
  **short name**, matching `ObjectAsExtensions.As<T>`'s existing rule — so two packages that each
  ship a `Currency` are unaffected (#1299's whole point is preserved), a rebuild into a new
  collectible assembly matches, and absent `$type` contradicts nothing and still resolves.

## Verification

Fail-before / pass-after, and a measured rate rather than a single green run.

**Deterministic pin** — `NodeTypePathRegistrationTest.ActivatingANodeTypeDefinition_DoesNotClaimItsOwnPathForItsInstances`
fixes the activation order (definition second), so it is not a race. Red before, naming the
claimant:

```
Expected value to be false because the definition node's activation must not claim
'TestData/declared-…' … Claimed: MeshWeaver.Graph.Configuration.NodeTypeDefinition, but found True.
```

Green after. It also asserts the *correct* entry (`"NodeType"` → `NodeTypeDefinition`) still gets
written, so a naive "just drop the stamp" fix does not pass.

**Measured rate**, `mw-plugin-test` over Store + Essentials staged exactly as the `plugin-gate` job
stages them (~9 s/run, default log level — `MW_LOG_LEVEL=Information` closes the window):

| | runs | gate failures | of which #1379's signature | of which a distinct 15 s residual |
|---|---|---|---|---|
| before | 60 | 2 (3.3%) | **1** (`got 'NodeTypeDefinition'`) | 1 |
| after | 130 | **0** | **0** | 0 |

Both before-failures are `GATE FAILED — tests: Store/Catalog` paid fulfilment, but they are **not**
the same defect and I am not claiming both:

- run 1 — `'PaidPack…' carries content that did not read as a package declaration (got
  'NodeTypeDefinition')`. This is #1379, and it is what this PR fixes: **1/60 → 0/130**.
- run 37 — `'PaidPack…' did not answer within 15s — the package's own hub is not serving its node`.
  That is Plugins#456's `DeclarationReadBudget` elapsing, a serving/activation delay rather than a
  content-typing fault. **This PR does not target it**, and 0 occurrences in 130 runs is too few to
  say anything about a 1-in-60 event either way.

**Regression net** (all green): `MeshContentTypeRegistryTest` 14/14 including the two new ones,
`TypedContentUpdateTest` 6/6 (notably `UntypedUpdate_WithTheAsOrDefaultIdiom_IsRescuedByTheExactContentTypeRoute`,
the case most exposed to the new guard), `ReimportTypedContentRecoveryTest` 5/5,
`NodeTypePathRegistrationTest` 3/3, `WhatsNewEntryIntegrityTest` + `DocumentationLinkIntegrityTest`.

`dotnet build -c Release -warnaserror` clean for both touched projects (`MeshWeaver.Graph`,
`MeshWeaver.Messaging.Hub`), DLL mtimes confirmed newer than the edits.

## Not in this PR

**#1371 does not share this root**, and the evidence is decisive rather than circumstantial: if
#1379 had degraded the package node's content, `ContentAs<PluginContent>` returns null →
`HasInstallableContent == false` → `NothingToInstall`, which writes **no manifest at all**. #1371's
defining observation is a manifest that *was* written. Same reasoning rules out Plugins#456 as its
fix — #456's failure mode also ends in no manifest. Details in the issue.

### DIFF
--- a/src/MeshWeaver.Documentation/Data/WhatsNew/2026-08-13-a-node-no-longer-shows-you-its-type-instead-of-itself.md
+++ b/src/MeshWeaver.Documentation/Data/WhatsNew/2026-08-13-a-node-no-longer-shows-you-its-type-instead-of-itself.md
@@ -1,0 +1,32 @@
+---
+Name: A node no longer shows you its type instead of itself
+Category: Fix
+Description: Just after a package's type was rebuilt, some of its nodes served the type's own definition in place of their content — so pages read as empty and a paid install reported there was nothing to install.
+Icon: Sparkle
+Order: -20260813
+---
+
+# A node no longer shows you its type instead of itself
+
+Every node says what kind of thing it is, and the platform keeps one list of what each kind's
+content looks like — so that a page rendering a node knows what shape to expect. Entries in that
+list were being written by two different things: by a node of that kind, which knows the right
+answer, and by the kind's own definition, which does not. Whichever wrote last won.
+
+When the definition wrote last, every node of that kind started reading back as **a copy of its own
+type definition** rather than its content. Nothing looked broken from the outside: the node kept its
+name, its address and its version, no error was logged, and the substitution held for as long as the
+process ran. Pages simply showed nothing, and anything that asked a node a question about itself got
+an answer that belonged to something else.
+
+The visible cost was on the Store. A purchase whose package was in that state read its own
+declaration, found a type definition where the list of things to install should have been, and
+concluded there was nothing to install — leaving a buyer charged, entitled, marked delivered, and
+holding an empty space. It happened only in the moments right after a package's type was rebuilt,
+which is why it looked random; a platform update rebuilds every type at once, so the window opened
+for everything simultaneously.
+
+Only the nodes themselves describe their content now, and the definition no longer overwrites what
+they said. Alongside that, content is never reshaped into a type it says it is not: if the two ever
+disagree again, the platform declines to answer and says so, instead of handing back a confident
+wrong one.
--- a/src/MeshWeaver.Graph/Configuration/MeshNodeHubFactory.cs
+++ b/src/MeshWeaver.Graph/Configuration/MeshNodeHubFactory.cs
@@ -30,12 +30,31 @@ internal class MeshNodeHubFactory(
                 // through — MonolithRoutingService and MessageHubGrain — so stamping here covers
                 // both hosts; the transient probe hubs stamp their own.
                 //
-                // For a NodeType DEFINITION node, `NodeType` is the literal "NodeType" and the
-                // semantically-correct value is the node's own Path — that is the type every
-                // instance of it will declare.
-                var nodeTypePath = string.Equals(enriched.NodeType, MeshNode.NodeTypePath, StringComparison.Ordinal)
-                    ? enriched.Path
-                    : enriched.NodeType;
+                // 🚨 The stamp names the type whose CONFIGURATION IS BEING APPLIED — always
+                // `enriched.NodeType`, never the node's own Path. A NodeType DEFINITION node
+                // (`NodeType` == the literal "NodeType") is the case that makes the distinction
+                // load-bearing, and it used to be special-cased to `enriched.Path` on the reasoning
+                // that the node's path "is the type every instance of it will declare". The KEY was
+                // right and the VALUE cannot be: a definition node's hub applies
+                // NodeTypeNodeType's configuration — `WithContentType<NodeTypeDefinition>()` — and
+                // never loads the compiled assembly of the type it declares, so the only thing it
+                // can publish under its own path is NodeTypeDefinition. That is the answer to a
+                // DIFFERENT question than the one the key asks: every read seam resolves by
+                // `node.NodeType`, so the entry at `Store/Plugin` is what an INSTANCE of
+                // Store/Plugin gets, and its content is PluginContent.
+                //
+                // MeshContentTypeRegistry.Register is last-writer-wins per key, so a definition
+                // node cold-activating after an instance overwrote the instance's correct entry,
+                // and the degrade seams then recovered instance content as the type's own
+                // definition — at its unchanged Version, on every read, for the rest of the process
+                // (Systemorph/MeshWeaver#1379: a paid Store package whose fulfilment read the
+                // declaration, got a NodeTypeDefinition, and reported "nothing to install").
+                // Instances and the schema probes already write the correct entry under exactly
+                // this key, so dropping the special case loses nothing; the literal "NodeType" that
+                // a definition node now stamps is itself a correct entry — a node whose NodeType is
+                // "NodeType" does carry NodeTypeDefinition content.
+                // Repro: NodeTypePathRegistrationTest.ActivatingANodeTypeDefinition_DoesNotClaimItsOwnPathForItsInstances.
+                var nodeTypePath = enriched.NodeType;
 
                 var defaultConfig = meshConfiguration.DefaultNodeHubConfiguration;
                 var nodeConfig = enriched.HubConfiguration;
--- a/src/MeshWeaver.Messaging.Hub/Serialization/IMeshContentTypeRegistry.cs
+++ b/src/MeshWeaver.Messaging.Hub/Serialization/IMeshContentTypeRegistry.cs
@@ -269,14 +269,55 @@ public sealed class MeshContentTypeRegistry(ILogger<MeshContentTypeRegistry>? lo
             return null;
         // Exact first: keyed on the node's own NodeType, so a name two packages share is not a
         // problem to be refused — each package's nodes resolve to their own type.
+        //
+        // 🚨 …but only when the content's OWN discriminator does not contradict it. Materialize
+        // deserialises to whatever type it is handed, and System.Text.Json ignores members that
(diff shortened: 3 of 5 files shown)


rbuergi:
Worth naming the kinship, because a third route probably exists.

The property that made this nasty is not the wrong registry entry — it is that **System.Text.Json
deserialising into the wrong record SUCCEEDS**. Unknown members are ignored and the declared ones
get defaults, so `Materialize` returns a plausible, fully-formed object rather than throwing. There
is no exception, no log line, and the result is indistinguishable from a correct read at every
consumer that does not already know what it should have got.

That makes this the same family as #1306 (merged earlier today): the bare-name last-writer-wins
claim in `MeshContentTypeRegistry`, where two packages sharing a `Currency` meant one package's
content deserialised into the other's record. Same silent-degradation mechanism, **two different
routes into it**:

| | route | key | fixed by |
|---|---|---|---|
| #1299 / #1306 | discriminator collision — two declarations claim one bare CLR name | `_byDiscriminator` | refuse a contested name |
| #1379 / #1380 | the hub-factory's path stamp — a definition node claims its instances' key | `_byNodeType` | stamp the config's own NodeType |

#1306's remedy was to make the *name* route refuse rather than guess. This one's is to stop the bad
entry being written at all. Neither addresses the shared enabling condition, which is why #1380 also
adds the `$type`-contradiction guard in `TryRecoverForNodeType`: whatever the next route turns out
to be, the registry can no longer reshape content into a type the content's own discriminator
disagrees with. An unresolvable read stays a `JsonElement` that `ContentAs<T>` can still recover and
the seams already warn loudly about; a wrongly-typed one is silent and lossy.

Both known routes write through the single call site `MeshDataSource.WithContentType`
(`MeshDataSource.cs:2116-2117`), so that is where a third would appear too. There are exactly three
callers supplying the key today — `MeshNodeHubFactory` (real activations, both hosts), the schema
probe at `MeshDataSource.cs:1532`, and `NodeTypeDataModelAreas.ProbeInstanceModel` via
`GetInstanceModelStream` — and after #1380 all three derive it from the NodeType whose configuration
is being applied. A fourth caller that derives the key from anything else (a node path, an address,
a display name) would re-open the same hole.

PR #1382, opened 2026-08-13
~~strike~~ no longer exports as bold + strike (#1373)

### LINKED ISSUE #1373: DocumentBuilder renders ~~strike~~ as bold + strike (affects PDF and DOCX export)
Split out of #1309's "Related, found during the same work and NOT fixed", so it is not lost when that issue closes for the contents-page-number regression.

`DocumentBuilder.ReadInlinesInto` sets `isStrong` from `DelimiterCount >= 2`, which is also true for `~~strike~~` — so strikethrough text is emitted as **bold + strike** rather than strike alone.

The delimiter *character* has to be part of the test: `**`/`__` is strong, `~~` is strikethrough, and only the former should set `isStrong`.

Affects both renderers, since both consume the same document model:
- PDF — `DocumentPrintComposer.RenderInline` wraps the run in `<strong><s>…</s></strong>`
- DOCX — `DocxDocumentRenderer` writes both run properties

Pre-existing; faithfully reproduced by #1230 rather than silently corrected.

**Where:** `src/MeshWeaver.Markdown.Export/Ast/DocumentBuilder.cs`
**Test home:** `test/MeshWeaver.Markdown.Export.Test/DocumentBuilderTests.cs`

rbuergi:
Closes #1373.

## The cause

`DocumentBuilder.ReadInlinesInto` decided a span was strong from `DelimiterCount >= 2`.

Markdig's `EmphasisExtras` — enabled here by `UseAdvancedExtensions` — folds **six** different
constructs into the one `EmphasisInline` node and tells them apart by the delimiter **character**,
never the count. Probed against the export's own pipeline:

| markdown | `DelimiterChar` | `DelimiterCount` |
|---|---|---|
| `**b**` / `__b__` | `*` / `_` | 2 |
| `*i*` / `_i_` | `*` / `_` | 1 |
| `~~s~~` | `~` | **2** |
| `~sub~` | `~` | 1 |
| `^sup^` | `^` | 1 |
| `++ins++` | `+` | 2 |
| `==mark==` | `=` | 2 |

`~~` is doubled, so strikethrough satisfied "strong" as well and the run was emitted **bold and
struck**. The same slip mis-styled four more nobody had noticed:

- `++ins++` → **bold**, `==mark==` → **bold** (doubled delimiter)
- `~sub~` → struck through (shares `~` with strikethrough)
- `^sup^` → *italic* (fell through to "not strong, not strike")

Emphasis is now keyed on `DelimiterChar`. Not post-processed, not special-cased for `~~`: the
character was always the discriminator and the code was reading the wrong field.

## The four with nowhere to go

Subscript, superscript, inserted and marked have no representation in the document model, which
carries bold / italic / strike / code and nothing else. They render as their **plain text** — every
word kept, no emphasis invented. That is a deliberate change from today's behaviour, where they wear
a style at random; giving them styles of their own means extending the model and both renderers,
which is a feature and not this fix.

## Verified on the artefact

The complaint is that a word came out bold *on a page*, so a model assertion cannot close it alone.

**Fail-before, read straight out of the produced files:**

- **PDF** — the struck word's glyphs were drawn with `DAAAAA+Arial-BoldMT`. Chromium embedded the
  *bold face* for it; that is the browser's own answer to "did you print this bold?".
- **DOCX** — the run for the struck word carried `<w:b/>`.

**And by rasterising both prints and looking at them:**

| | `~~Retired~~` | `~sub~` `^sup^` `++ins++` `==mark==` |
|---|---|---|
| before | **bold** + struck | struck / italic / bold / bold |
| after | struck, normal weight | plain text |

Real `**bold**` is still bold, `*italic*` still italic, `***both***` still both, and
`**bold with ~~strike~~ inside**` keeps both on the inner phrase — the fix removes a phantom, not a
style.

**Both artefact tests carry a genuinely bold word in the same document as their control**, so
"not bold" is a finding rather than an absence: if the pipeline had stopped producing bold at all,
the control fails first. The DOCX test reads run properties out and asserts them non-null before
inspecting children — a `?.` chain there would short-circuit on a run without properties and make
the assertion unfailable. (`-warnaserror` caught the null-deref that led me to it.)

## Verification

- `MeshWeaver.Markdown.Export.Test` **113/113** — 8 red before the fix (both artefact tests + 6
  AST cases), 0 after.
- Export e2e suites in `MeshWeaver.Hosting.Monolith.Test` 16/16 · `MeshWeaver.Documentation.Test`
  20/20.
- `-c Release -warnaserror` clean on `MeshWeaver.Markdown.Export` and the test project.

The email/HTML export is unaffected and always was: it renders markdown through Markdig's own HTML
renderer rather than the document model, so `~~` has always produced `<del>` there. That matches the
issue's scope — PDF and DOCX only.

## Overlap with #1372

Both touch `Ast/DocumentBuilder.cs`, but **not the same code**: #1372 changes the `WalkBlocks`
signature and `AnchorFromInlines`; this changes the `EmphasisInline` case in `ReadInlinesInto` and
adds a private helper before `ExtractPlainText`. The nearest hunks are ~11 lines apart in the
original file, so a 3-way merge takes both. Either order works — no sequencing needed.


rbuergi:
### CI red on an unrelated live-model test — recorded so it is not re-derived

Run [31667783996](https://github.com/Systemorph/MeshWeaver/actions/runs/31667783996) (shard 0),
**one** failing test, on the tree that already has `origin/main` @ `44c0a510d` merged in:

```
MeshWeaver.Hosting.Orleans.Test.OrleansSubThreadAutoResumeTest
  .SubThreadDelegation_LiveProgressOnToolCall_TerminalSuccess_AndFcReentry

Expected 2 to be greater than or equal to 3 because sub-thread response cell must emit
≥3 distinct growing Text snapshots during streaming (saw 2).
```

**It cannot be this diff.** The assertion counts how many intermediate streaming writes a **live
model** produced for one sub-thread round — a function of that run's chunking and timing. This PR
changes one private static method (`DocumentBuilder.EmphasisStyle`) reachable only from
`DocumentBuilder.Build(...)`; the test never builds an export document. (`MeshWeaver.Markdown.Export`
is on the Orleans test project's reference list, so the assembly loads — but nothing calls into it
here.)

**It belongs to an existing population, not to this branch.** Scanning the last 10 failed
Build-and-Test runs across all branches: six failed, each on a *different* single integration test,
none related to the PR it ran on —
`SilentReadNackTest`, `FrameworkStaleProactiveRebuildTest` (×2), `CodeCellOutputCaptureTest`,
`ThreadAgentIntegrationTest`, `LayoutAreaRetrievalTest` / `StaleActivationDurableFirstSeedTest` /
`OrleansGrainTeardownStragglerTest`. The sibling PR #1381 hit the same shape on its first run
(`DelegationSubThreadUsageTest`, `TimeoutException` against `claude-haiku-4-5`) and went **fully
green** on the next run after merging main — same tests, same diff.

**Deliberately not re-run.** Re-running a red test to see whether it goes green hides races
(AGENTS.md → "No band-aids"), and `origin/main` has already moved past what this branch merged
(`e23dd02b3`). The strict-required-status-checks policy requires this branch to be up to date at
merge time anyway, so the next `git merge origin/main` is both required and a genuinely changed
tree — that is the legitimate re-run, and it should be taken as part of sequencing the merge train
rather than manufactured now.

Local verification on the merged tree is unchanged: `MeshWeaver.Markdown.Export.Test` **113/113**,
export e2e suites **16/16**, `MeshWeaver.Documentation.Test` **20/20**,
`-c Release -warnaserror` clean.

PR #1385, opened 2026-08-13
Two CI flakes that were neither flaky nor live-model — plus the bucketed sweep (#1384)

rbuergi:
Closes part of #1384.

## The sweep first — the population is bigger than "six of ten", and it is a long tail

Every failed `MeshWeaver Build and Test` run since 2026-08-08 (300 runs), bucketed by the
failing test name taken from each shard's `Test failed` annotations, deduped per run.

| window | failed runs | with ≥1 test failure | **failed on exactly ONE test** |
|---|---|---|---|
| 2026-08-08 19:15 → 08-13 05:05 (4.4 d) | 300 | 242 | **168 (56 %)** |
| 2026-08-12 00:00 → 08-13 05:05 (29 h) | 94 | 72 | **51 (54 %)** |

So the shape the issue describes is real and steady: **just over half of all reds are a single
test.** But those 51 singleton runs spread across **30 distinct tests** — this is not one race
with several faces, it is a long tail. Top of it, over the last 29 h (occurrences | distinct
branches):

| # | branches | test | status |
|---|---|---|---|
| 8 | 3 | `FutuReAnalysisTest.EuropeRe_AnnualReport_EmbeddedCharts` | live — `No renderer is registered for area KeyMetrics`, preceded by `'$type':'AnalysisContent' is NOT registered in this hub`. A dynamic-NodeType registration race, **not** #613 (that is `FutuRe.Test exit=139 SIGSEGV`, a different signature in the same project) |
| 6 | 4 | `SkillAutocompleteTest.Autocomplete_ListsBuiltInSkills_FromCatalog` | **appears fixed** — last seen 08-12T09:36, nothing since |
| 4 | 4 | `FrameworkStaleProactiveRebuildTest.…_ProactivelyRebuilds_…` | live (08-12T21:51 → 08-13T03:09), and on four *unrelated* branches incl. `main` |
| 3 | 3 | `ThreadAgentIntegrationTest.FullFlow_…` | live (→ 08-13T01:24) |
| 2 | 2 each | `OrphanedStartingExecutionRecoveryTest`, `NodeTypeModelProbeTest`, `LogIncidentIngestAccessContextTest`, `CrossHubPatchAtomicityTest` | live |
| 1 | 1 | 22 further tests | tail |

The two biggest buckets over the *full* 4.4 days are already dead and must not be re-chased:
`OrleansBrokenNodeTypeAccessTest` (16, all on 08-09/08-10) and `StaleStampRootBindingTest`
(6, all on 08-10). Excluded as instructed and not double-counted: #890 (bare NRE + `exit=124
TIMEOUT`), #613 (`FutuRe.Test exit=139`), #1379 (plugin gate).

**Answer to "is the Orleans-activation half one race or several": several.** Nothing in the tail
shares a root with anything else in it.

## The two tests this PR fixes — and why the issue's premise about them is wrong

#1384 classes both as live-model tests "whose outcome depends on an external API's latency and
chunking", and concludes "no change in this repo can make those deterministic". **Neither test
calls a model.** Both construct a fake in-process `IChatClient`:

- `DelegationSubThreadUsageTest` — `claude-haiku-4-5` is a *string*, used only to look up a
  `ModelPricing` rate. Its seeded provider endpoint is literally `https://example.invalid/v1`.
- `OrleansSubThreadAutoResumeTest` — `StreamingSubAgentClient` emits a fixed 320-char report in
  5-char chunks 20 ms apart, ≈1.3 s of streaming through `ThreadExecution`'s `Sample(100 ms)`,
  so ≈13 writes. `≥3 snapshots` is an assertion about **this repo's** streaming path, not about
  how a model chose to chunk.

Both are ordinary in-repo races with a deterministic form. No bound raised, no assertion
loosened, nothing moved out of the gating path.

### 1. `DelegationSubThreadUsageTest` — the test's own read broke the write it was waiting for

It polled `GetMeshNodeStream({subThread}/_Usage/{model})` — a **point read of a node that does
not exist yet** — behind a `.Catch`. Two consequences:

- `RecordUsage` is subscribed as an *independent* side effect, deliberately not chained before
  the round's terminal status, so "the sub-thread settled" never implied "the satellite exists".
  A point read of an absent node answers with an authoritative routing `NotFound` and
  **terminates** the stream; it cannot wait for a node to appear.
  `ThreadTokenUsageTest.WaitForUsage` documents precisely this, and #1040 removed the same shape
  from that class. This test reintroduced it.
- Worse: that `NotFound` opens `MeshNodeStreamCache`'s storm-breaker window **on the exact path
  `RecordUsage` is about to accumulate into**, and the breaker fast-fails *writes* as well as
  reads. Runs `31666562025` and `31645997822` log the whole chain in order:

  ```
  RouteMessage: NotFound for RawJson → …/_Usage/claude_haiku_4_5
  [SYNC_STREAM] OnError … Owner=…/_Usage/claude_haiku_4_5
  [TokenUsage] RecordUsage failed for …/_Usage/claude_haiku_4_5
      DeliveryFailureException: No node found at '…/_Usage/claude_haiku_4_5'
  … 45 s later: TimeoutException
  ```

  `RecordUsage`'s Phase 1 `CreateNode` succeeded, so the satellite existed — with **zero
  tokens**, because Phase 2's accumulate was suppressed by the window the test's own poll had
  just opened. Exactly one router `NotFound` in the whole run and no further ones, which is what
  you would expect once the node exists: the poll then found it, forever failing its
  `InputTokens == 100` predicate.

Now reads through the live children query of `{subThread}/_Usage` — empty-on-absent, so it
neither `NotFound`s nor poisons — **opened before the round can finish**, so the losing ordering
is covered rather than sampled. The redundant "wait for the sub-thread to settle" hop is deleted;
it carried the same unsound predicate as #2 below.

### 2. `OrleansSubThreadAutoResumeTest` — the progress window closed before the round started

The window was bounded by `IsExecuting: false && Messages.Count >= 2`. But `IsExecuting` is
`Status is StartingExecution or Executing`, so it is false for the **initial `Idle` state** too —
and the snapshot that satisfied the earlier `Messages.Count >= 2` wait *is* Idle-with-two-cells,
which the shared stream handle's `Replay(1)` hands straight to this subscription. The wait could
therefore return before the sub-agent emitted a token. (A finished round also returns to `Idle`,
so the flag genuinely cannot tell "not started" from "finished".)

Run `31667783996` is exactly that. The two snapshots it saw:

```
[progress 1] 19 chars, last line: "Allocating agent..."
[progress 2] 22 chars, last line: "Generating response..."
```

Framework placeholders — not one character of the sub-agent's report. The test then failed
`2 >= 3` and got attributed to "the model chunked differently".

The window now closes on the sub-thread's **response cell** reaching
`ThreadMessageStatus.Completed`, which cannot be true before the round starts. `>= 3` is
unchanged and now clears with margin — 16 snapshots locally, against 2 in the failing CI run —
and a new assertion requires at least one snapshot to carry the sub-agent's streamed report, so
"placeholders only" can never pass silently again.

Both files also drop `content as T` casts on the paths they touch in favour of
`ContentAs<T>(hub.JsonSerializerOptions)`.

## One framework hazard found, deliberately NOT fixed here

The storm breaker fast-failing writes is *by design* (its own comment: "a missing-node path can
never storm the mesh from either direction… Only a real `CreateNode` brings a non-existent node
into being"), and it relies on the change-feed `ResetFailureState` to heal the path after a
create. In the failing runs the negative window was still open when `RecordUsage`'s post-create
`Update` ran, so `RecordNegative` either latched a probe failure that landed *after* the create's
invalidation, or the invalidation never reached it. I could not discriminate the two from the CI
logs, and guessing at a fix inside `MeshNodeStreamCache` is not something to do on a hunch —
filing it separately with the repro recipe (point-read `{thread}/_Usage/{model}` while absent,
then run a usage-emitting round, then assert the satellite accumulated). Production is not
currently exposed: `ThreadTokenChip` and `ThreadChatView` read `_Usage` only through the children
query. The test was the only point-reader.

## Why this is worth the time

A red run that is not the PR's fault trains everyone to re-run reds, and that habit is what hides
real races. Two genuine defects were caught this week precisely because someone declined to
re-run — a store-key defect that was green only while production was broken, and a NACK that
slandered healthy hubs 18 ms after startup. At 54 % of reds being a single unrelated test, that
discipline stops paying. Both fixes here are the kind that only exist because someone read the
log instead of pressing re-run: one test was suppressing the very write it was waiting for, and
the other was asserting on two placeholder strings.

## Verification

- `dotnet build -c Release -warnaserror` clean (0 warnings, 0 errors) for both test projects.
- `DOTNET_PROCESSOR_COUNT=4` local runs of both tests pass; the Orleans one now records 16
  progress snapshots.


rbuergi:
## A third instance, found by this PR's own CI — and the population biting the PR that fixes it

Two CI runs on this branch went red, neither on the tests it changes. Both are members of the
#1384 population. Rather than re-run (the habit this PR exists to make unnecessary), I read them.

### Run 1 — `Hosting.Monolith.Test (part 1/2) exit=124 TIMEOUT`

No `Test failed` annotation and `Test Results (shard 1)` reported **pass**; only the raw log names
`ReleaseRequestWatcherHighWaterTest`, which burned 50 s on
`Could not load type 'HighWaterLayoutAreas' from assembly 'DynamicNode_TestData_HighWaterType'
because the format is invalid` — a torn read of a just-emitted dynamic assembly. Untouched by this
diff. Branch was also 24 commits behind, so per AGENTS.md I merged main before investigating.

**This also proved my own sweep under-counts:** of the 300 failed runs, 58 had no trx test failure
at all, because a project killed at the 8 min cap loses its in-flight tests from the trx entirely.
The singleton figures in the description are a floor. Written up on the issue.

### Run 2 — `OrleansChatTest` + `PluginDependencyOrderTest`

Both shards carrying this PR's two tests **passed** (0 and 5); the two failures were elsewhere.

`OrleansChatTest.CreateThread_AndSubmitMessage_ProducesThreadMessages` turned out to be a **third
instance of the exact defect this PR is about**, so it is now fixed here. Step 7 verified the
response cell with:

```csharp
for (var i = 0; i < 50; i++) {
    responseContent = await GetHubContentAsync<ThreadMessage>(client, path, ct);
    var len = responseContent?.Text?.Length ?? 0;
    if (len > 0 && len == prevLen && ++stable >= 2) break; else stable = 0;
    prevLen = len;
    await Task.Delay(200, ct);
}
```

- "the text stopped growing for 400 ms" is a **sleep-based guess** at "streaming finished".
  `ThreadExecution` pushes through `Sample(100 ms)`, so a scheduling hiccup yields two equal-length
  reads mid-stream (breaks early) — or never two in a row, and it runs all 50 rounds.
- 50 × (a `GetDataRequest` round-trip + 200 ms) overruns the test's own 50 s token, surfacing as a
  bare `TaskCanceledException` out of the helper instead of a useful assertion. That is exactly how
  run 31672644866 reads.
- Hand-rolled `while + Task.Delay` poll loops are forbidden outright (AGENTS.md → Testing
  Guidelines).

Now waits on the response cell reaching `ThreadMessageStatus.Completed` — created at `Streaming`,
reaches `Completed` exactly when the streaming loop exits. No heuristic, no sleep, cannot pass
early.

**Attribution check, not an assumption.** `OrleansChatTest` lives in the same assembly as the test
this PR changes and xUnit runs them serially, so "unrelated" needed proving rather than asserting:
the sweep records it failing **three times across three unrelated branches (08-10 → 08-12)**, all
predating this branch. Pre-existing.

### Not fixed here

`PluginCatalog.Test.PluginDependencyOrderTest.InstallingTheClosureInOrder_LandsTheDependentTyped`
failed the same run with `HubDisposingException: Hub Dependent/Item is shutting down` during
install — a hub-lifecycle race in the plugin-install path. Different root, unrelated to this diff,
and not one to guess at. Reported on the issue instead.

Local verification of all three: `-c Release -warnaserror` clean, and `DOTNET_PROCESSOR_COUNT=4`
runs pass.

PR #1402, opened 2026-08-13
CI: run the .NET suite on every PR, not only those based on main

### LINKED ISSUE #1401: Stacked PRs never run the .NET suite — an untested PR is indistinguishable from a passing one
`.github/workflows/dotnet-test.yml` triggers on:

```yaml
on:
  pull_request:
    branches: [main]
```

So a PR whose **base is another PR's branch** — a stacked PR — never runs `MeshWeaver Build and Test` at all. It shows the `Clients` suite green and simply *no* .NET suite.

**The failure mode is the dangerous half: absence looks like nothing to wait for.** A stacked PR renders as a short, all-green check list. It is visually indistinguishable from a PR that ran the full suite and passed — there is no "skipped", no pending, no marker of any kind. Someone reviewing the checks sees green and merges.

This is the same family as the ~16 other "success on absent evidence" defects found this week (a payment-path gate reporting `tests=skipped` as a pass; `Test Results (shard 1)` reporting *pass* on a shard killed at its cap; a leak test asserting collection after a disposal production never performs). It is arguably the worst of them, because it applies to **every stacked PR in the repo** rather than one subsystem, and stacking is the recommended way to split a large change into reviewable pieces.

## Why it has not bitten yet

The suite does attach as soon as GitHub **retargets** the PR to `main` — which happens automatically when its parent merges. So a stacked series merged strictly in order gets tested, one PR at a time, just later than it appears. The hazard is merging out of order, or reading a stacked PR's green as evidence before its parent lands.

Found while stacking #1398 → #1399 → #1400 (the activity-log satellite migration); both children were called out explicitly in their PR bodies so nobody merged on the false green.

## What to fix

Make an untested PR **look** untested. Options, roughly in order of preference:

1. **Run the suite on stacked PRs too.** Widening the trigger beyond `[main]` is the direct fix; weigh the CI cost, since a stacked series would then test each layer twice (once stacked, once retargeted).
2. **Emit an explicit blocking status** on any PR whose base is not `main`, saying the .NET suite has not run and why. Cheap, and it converts silence into a visible statement.

🚨 Whatever the shape, it must not be an `if:` that skips a job — GitHub renders a skipped job with the same tick as a passed one, which reproduces the exact defect one level up. The repo's own rule: a gate never tests its own inputs, and a check that cannot fail is worse than no check.

Do **not** solve this by documenting "remember stacked PRs aren't tested". That is the state that produced it.

rbuergi:
Closes #1401.

## The change

One line: `dotnet-test.yml`'s `pull_request` trigger loses its `branches: [ "main" ]` filter, so **every PR runs `MeshWeaver Build and Test` whatever its base**. The rest of the diff is the comment block explaining why, so nobody re-adds the filter.

This is option 1 from the issue (run the suite on stacked PRs), not option 2 (emit a blocking "not tested" status). Option 2 would be a check that, after this change, can never fire — and the repo's own rule is that a check which cannot fail is worse than no check. There is no `if:`-shaped skip anywhere in the diff: a skipped job renders with the same tick as a passed one, which would reproduce the defect one level up.

## Why the pre-fix state read as CLEAN

Four things stacked up, and all four were confirmed on #1398 → #1399 → #1400:

| | |
|---|---|
| **No run at all** | The trigger filtered on base = `main`, so a PR based on another PR's branch never matched. |
| **No retarget** | GitHub retargets a child PR to `main` only when the parent's branch is **deleted** on merge. Branch deletion is off here, so children keep pointing at merged branches indefinitely — "it gets tested later, just out of order" is not real. Left alone a stacked PR is tested at **no** point, including at merge. |
| **No trigger on a hand retarget** | `gh pr edit --base main` fires `pull_request.edited`, which is not in the default type set (`opened`, `synchronize`, `reopened`). The base changed, the checks stayed absent. Only close + reopen ever queued a run. |
| **No required check either** | The `main pr protection` ruleset is scoped to `~DEFAULT_BRANCH` (`conditions.ref_name.include: ["~DEFAULT_BRANCH"]`), so a PR whose base is not `main` sits outside it. `Consolidate test results` is not required there, so its *absence* does not block the merge. |

So there was no red, nothing pending, nothing skipped, and nothing required — an untested PR was byte-for-byte indistinguishable from a passing one. Two of the three PRs in that split reached `MERGEABLE/CLEAN` with **zero** runs in their entire history.

`dotnet-test.yml` was the only workflow with the filter — `clients.yml` already triggers on all bases, which is why the stacked PRs showed a green `Clients` suite and simply *no* .NET suite.

## What a stacked run actually tests

GitHub builds `refs/pull/N/merge` = the branch merged with **its base**. For a stacked PR that base is the parent's branch, so the run tests the layer's own diff on top of the parent — real coverage of the code under review. That is the same "tested against a base that is not main's tip" posture the ruleset already sanctions repo-wide (`strict_required_status_checks_policy: false`).

## CI cost

Bounded, because the green-tree reuse is keyed on the **tree**, not the commit. When the parent merges and main lands the identical tree, `precheck` finds the marker the stacked run already wrote and skips build + test. A stacked series pays one full suite per layer, not two.

`push:` stays `main`-only on purpose — a full suite on every feature-branch push would double every PR's cost for no extra evidence.

## Left deliberately undone (needs a repo-settings decision, not a PR)

The ruleset's `~DEFAULT_BRANCH` scope means a stacked PR still has **no required status check**: the suite now runs and is visible, but a red one does not mechanically block the merge. Widening `conditions.ref_name.include` to `~ALL` would close that, and it is repo configuration rather than anything in this tree — flagging it rather than changing it silently.

## Verification

`.github/workflows/dotnet-test.yml` parses (`yaml.safe_load` → `on: {'push': {'branches': ['main']}, 'pull_request': None, 'workflow_dispatch': None}`, all seven jobs intact), and no job in the file keys off `github.base_ref` / `head_branch` / `refs/heads/main`, so nothing downstream assumed a `main` base.

**This PR is itself the test**: it is based on `main`, so it proves only the unstacked path. The stacked path is proven by the follow-up PR for #1384, which I will stack on this branch — if the suite attaches there, the fix works.

No What's New entry: CI-only, no user-visible effect.

### DIFF
--- a/.github/workflows/dotnet-test.yml
+++ b/.github/workflows/dotnet-test.yml
@@ -65,11 +65,49 @@
 
 name: MeshWeaver Build and Test
 
+# ──────────── EVERY PR, WHATEVER ITS BASE — including a STACKED one ────────────
+# 🚨 The `pull_request` trigger deliberately carries NO `branches:` filter. Do not
+# re-add one. It used to say `branches: [ "main" ]`, and the consequence (#1401) was
+# that a PR based on ANOTHER PR's branch — a stacked PR, which is the recommended way
+# to split a large change into reviewable pieces — never ran this suite at all.
+#
+# The failure mode is the dangerous half: an ABSENT check looks like nothing to wait
+# for. There is no "skipped", no pending, no marker of any kind — a stacked PR renders
+# as a short, all-green check list, visually IDENTICAL to one that ran the full suite
+# and passed. #1398 → #1399 → #1400 were split exactly as intended and two of the three
+# reached `MERGEABLE/CLEAN` with zero runs in their entire history.
+#
+# Three things make that state stickier than it looks, all confirmed on those PRs:
+#   * GitHub retargets a child PR to `main` only when the parent's branch is DELETED
+#     on merge. Branch deletion is off here, so a child keeps pointing at a merged
+#     branch indefinitely — "it gets tested later, just out of order" is not real.
+#     Left alone, a stacked PR was never tested at ANY point, including at merge.
+#   * `gh pr edit --base main` fires `pull_request.edited`, which is NOT in the default
+#     type set (opened/synchronize/reopened). The base changed, the checks stayed
+#     absent, and the PR still read CLEAN. Only close+reopen ever queued a run.
+#   * The `main pr protection` ruleset is scoped to `~DEFAULT_BRANCH`, so a PR whose
+#     base is not `main` is outside it: the `Consolidate test results` required check
+#     is not required there, and its absence therefore does not even block the merge.
+# So the pre-fix state had no signal at all: nothing ran, nothing was pending, nothing
+# was required. Running unconditionally is what turns that silence back into evidence.
+#
+# What a stacked run tests: GitHub builds `refs/pull/N/merge`, i.e. the branch merged
+# with ITS BASE. For a stacked PR that base is the parent's branch, so the run tests
+# the layer's own diff on top of the parent — real coverage of the code being reviewed.
+# That is the same "tested against a base that is not main's tip" posture the ruleset
+# already sanctions repo-wide (`strict_required_status_checks_policy: false`).
+#
+# Cost is small and bounded, because the green-tree reuse below is keyed on the TREE:
+# when the parent merges and main lands the identical tree, `precheck` finds the marker
+# the stacked run wrote and skips build+test. A stacked series pays for one full suite
+# per layer, not two.
+#
+# `push:` stays `main`-only on purpose — a full suite on every feature-branch push
+# would double every PR's cost for no extra evidence (the PR run already covers it).
 on:
   push:
     branches: [ "main" ]
-  pull_request:
-    branches: [ "main" ]
+  pull_request:      # ← NO `branches:` filter. See the block above (#1401) before adding one.
   workflow_dispatch:
 
 # A force-push to a PR branch obsoletes the in-flight run — cancel it instead of


rbuergi:
## Live proof: the stacked path now runs the suite

#1408 is opened with `base = fix/1401-stacked-pr-untested` — a non-`main` base, i.e. exactly the shape that ran **nothing** before this change. With this branch as its base, the run attaches:

```
$ gh run list --branch fix/1384-pinned-scan-parks-nodetype
MeshWeaver Build and Test    pull_request    in_progress
```

Both paths are now covered by evidence rather than by reasoning:

| | base | `MeshWeaver Build and Test` |
|---|---|---|
| this PR (#1402) | `main` | runs — the unstacked path, unchanged |
| #1408 | `fix/1401-stacked-pr-untested` | **runs** — the path that previously produced no check at all |

Worth noting *why* the stacked run picks up the fix at all: for a `pull_request` event GitHub resolves the workflow from the **merge ref** (head merged into base), and the base branch here carries the new trigger. So the fix becomes effective for children as soon as it is on the branch they stack on — and for everything once it is on `main`.

⚠️ #1408 must not be merged before this PR — merging it first would put its commits on `main` via a base that no longer exists.

rbuergi:
**Superseded by #1408** — do not merge this one.

#1408 was retargeted from this branch to `main`, so its branch now carries all three files: this workflow change plus the ALC teardown fix and its test. Merging #1408 lands both.

That retarget also broke a genuine deadlock worth recording, because it is the defect this PR is about biting the fix for itself: this PR was red on `StaleStampRootBindingTest` — the very bug #1408 fixes — while #1408 was stacked on it and therefore could not land first. Neither could go green ahead of the other.

Two details from the recovery that match the issue body exactly, now observed a second time:

- changing the base fires `pull_request.edited`, which is **not** in the default type set, so no run queued;
- **close + reopen** is what queued one.

Which is the whole argument for #1401: the retarget alone looks like it worked.

I will close this once #1408 merges, rather than now — until then it is the fallback if #1408 needs rework.

Issue #1387, opened 2026-08-13
A NodeType assembly can be loaded mid-write — 'format is invalid' parks the type, and the emit path skips AtomicFileWrite


rbuergi:
A NodeType's freshly-emitted assembly can be **read while it is still being written**, producing a load failure that looks like a compile failure and parks the type.

Observed on a CI shard (`ReleaseRequestWatcherHighWaterTest`, ~50 s burned before the shard hit its cap):

```
ReflectionTypeLoadException: Could not load type 'HighWaterLayoutAreas' from assembly
  'DynamicNode_TestData_HighWaterType' because the format is invalid.
NodeType 'TestData/HighWaterType' PARKED after compile failure
```

**"Format is invalid" on an assembly the process itself just emitted is a torn read, not a compile error.** The bytes were incomplete at the moment of load. The downstream consequence is worse than a slow test: the type is **`PARKED` after compile failure**, and a parked NodeType refuses portal readiness and stalls every instance hub for the full activation budget.

## Why this is very likely real and not exotic

The identical hazard was fixed **today** in two other places — `FileSystemStorageAdapter` and `FileSystemVersionStore` now write temp-then-rename through a shared `AtomicFileWrite` helper (MeshWeaver#1359). That fix exists because `File.WriteAllTextAsync` uses `FileMode.Create`, which **creates the target and then streams bytes**, so a concurrent reader can observe a 0-byte or partial file. The version store had hand-rolled an unprotected second copy of the same write; the sibling adapter's comment already named the hazard.

**The assembly write does not appear to go through `AtomicFileWrite`.** That is the lead: same defect class, third site, and this one publishes *executable* bytes where a partial read is guaranteed to fail rather than merely return empty content.

## What to establish

1. Confirm the emit path writes the assembly (and its PDB) non-atomically, and route it through the same temp-then-rename helper. The rename must be atomic on the target filesystem — note the assembly cache lives on **Azure Files (SMB)** on AKS, where rename semantics deserve an explicit check rather than an assumption.
2. Establish whether the reader can also observe a *renamed but not yet flushed* file, and whether the load path has any retry/park behaviour that converts a transient torn read into a **persistent** `PARKED` state. A torn read that parks permanently is far worse than one that fails once — and "parked after compile failure" reads as a source defect to everyone who sees it, which is how this could survive.
3. Check the same for the **share** copy (`/data/assembly-cache`), not just the local one.

## Explicitly NOT the fix

[...]

Issue #1405, opened 2026-08-13
Flake: PluginDependencyOrderTest.InstallingTheClosureInOrder_LandsTheDependentTyped times out when the Dependent/Item hub recycles mid-install


rbuergi:
## Failure

`MeshWeaver.PluginCatalog.Test.PluginDependencyOrderTest.InstallingTheClosureInOrder_LandsTheDependentTyped` failed on CI run [31686390017](https://github.com/Systemorph/MeshWeaver/actions/runs/31686390017) (shard 4, PR #1403 — a portal-only Blazor change that does not touch the plugin catalog, install pipeline, or mesh data plane; the branch was exactly one commit atop current main `86aab1132`).

```
System.TimeoutException : The operation has timed out.
  at PluginDependencyOrderTest.cs(142): GetMeshNodeStream("Dependent/Item").Where(n => n is not null).FirstAsync().Timeout(30s)
```

## Evidence from the shard log (job 94405501866)

The install raced a recycle of the `Dependent/Item` per-node hub:

```
[09:37:51.507] [Warning] [MeshWeaver.Graph.MeshDataSource] Own-node refresh from change notification failed on Dependent/Item
MeshWeaver.Messaging.HubDisposingException: Hub Dependent/Item is shutting down — cannot create '/MeshNode'.
The address may reactivate (recycle / restart); retry to get the authoritative answer.
   at SynchronizationStream`1..ctor (SynchronizationStream.cs:983)
   …
   at MeshNodeStreamHandle.UpdateOwn (MeshNodeStreamExtensions.cs:678)

[09:37:56.379] [Warning] [MeshNodeTypeSource] MeshNodeTypeSource[Dependent/Item]: DROPPED a stale own-node
emission (Version=0) below the version this hub already holds (Version=1). The routing-supplied own-node
stream is cache-backed … and can replay a snapshot from before a recycle.

[09:38:21.371] === TEST FAILED: The operation has timed out.
```

So the test's single `GetMeshNodeStream("Dependent/Item")` subscription was opened against a hub that was mid-shutdown; the exception's own text says "retry to get the authoritative answer", but nothing retried, and the fresh post-recycle emission never reached the already-wedged stream → 30 s timeout. This is the store/plugin **root-churn / recycle race** family, not a slow Roslyn or runner problem.

## Correlation — load-coupled, not diff-coupled

Three PR branches failed in the same CI window on three DIFFERENT tests while main was green on the same tree:

| run | branch | failing test |
[...]

PR #1408, opened 2026-08-13
A teardown race permanently parks a NodeType (#1384)

rbuergi:
Part of #1384 — one root, found by reading a red instead of re-running it.

> 🧪 **This PR is deliberately STACKED on #1402** (`fix/1401-stacked-pr-untested`), because it is the live proof of that fix. Before #1401, a PR with a non-`main` base ran **no** .NET suite at all and still read `CLEAN`. If `MeshWeaver Build and Test` appears on this PR, #1402 works. **Do not merge this before #1402.**

## The bug

A scan that had legitimately **pinned** a NodeType's `AssemblyLoadContext` could be handed an `ObjectDisposedException` by the very teardown that was standing there **waiting for that scan to finish** — and the NodeType was then parked permanently.

```
ObjectDisposedException: Cannot load assembly from disposed context
Object name: 'DynamicNode_ShopA_Front'
   at NodeAssemblyLoadContext.LoadNodeAssembly()          CompilationCacheService.cs:446
   at MeshNodeCompilationService.CompileResultFromAssembly(...)
NodeType 'ShopA/Front' PARKED after compile failure — further activations serve the
cached error without recompiling (failure contained, no retry storm).
```

`Pin()` and `LoadNodeAssembly()` guarded on the **same** `_disposed` flag, and `Dispose()` raised it on **entry**, before draining pins:

1. `Pin()` succeeds — it re-checks the flag after incrementing, so a pin that was *handed out* proves the context was live.
2. `Dispose()` sets `_disposed`, then **blocks in the pin drain, waiting for us**.
3. `LoadNodeAssembly()` reads `_disposed` and throws — against a LoaderAllocator that is provably still alive, because `Unload()` has not been called and *cannot* be until our pin releases.

The pin bought exactly the window it exists to buy, and the load refused to use it. Nothing was wrong with the assembly; teardown had merely been *requested*.

**Why a millisecond race becomes permanent damage.** `CompileResultFromAssembly` catches the throw, records `CompilationStatus.Error`, and the compile watcher parks the type with no retry — by design, to contain real compile failures. So one spurious `ObjectDisposedException` kills the NodeType until someone rebuilds it by hand. Same *"one mis-read is not transient"* shape as #1387, reached from the opposite side: there the bytes were incomplete; **here the bytes were perfect and we refused to read them.**

## The fix

Split the one flag into the two states it was conflating:

| state | meaning | effect |
|---|---|---|
| `_unloading` | `Dispose` has **started** | `Pin()` refuses **new** scans, so the drain can finish. The assembly is still loadable. |
| `_disposed` | the drain is **over**, `Unload()` imminent | loads refused. |

`LoadNodeAssembly` now consults `IsClosedToLoads()`, which deliberately keeps the **old, strict behaviour for an UNPINNED caller** (`_unloading && _pins == 0`): without a pin an `Unload()` really can land between the load and the caller's `GetTypes()`, and that is the "…format is invalid" corruption the pin exists to prevent. Only a **pinned** scan is let through. `Lease()` / `ReleaseLease()` consult both flags, so the deferred-unload path for live instance hubs is unchanged.

No retry, no sleep, no widened bound — the context is either still alive when the scan reads it, or it is not.

## Evidence this is live

`StaleStampRootBindingTest.StaleStampSelfTypedRoot_RootServesItsTypesArea`, which #1384's tail lists as **dead** ("6, all 08-10 — do not re-chase"), failed **twice on 08-13**:

| run | commit / branch | when |
|---|---|---|
| `31683003325` | `891aae06a`, **main** | 08:39Z |
| `31686258015` | #1402 (an unrelated CI-only diff) | 09:22Z |

The second one is what surfaced the stack — it reddened my own workflow-only PR 40 minutes after main. The 08:39 occurrence had a *different* signature (a >20 s hub-teardown wedge, `root ShopA never answered after its recycle`), so that test name currently covers **two** roots; this PR fixes the `ObjectDisposedException` one. Details and the rest of the sweep are in [my comment on #1384](https://github.com/Systemorph/MeshWeaver/issues/1384#issuecomment-5278651814).

## Verification — fail-before / pass-after

`PinnedScan_StillLoadsTheAssembly_WhileDisposeIsDraining` (new, in `AssemblyLoadContextLeakTest`) is **deterministic, not timing-hopeful**: the pin is taken first and released only after the load has been attempted, so the interleaving it asserts is the one that always happens.

```
before  FAIL — System.ObjectDisposedException : Cannot load assembly from disposed context
               (verbatim the CI failure)
after   pass
```

It also asserts the other half — once the drain is over, a later load still throws — so the fix cannot be "just delete the guard".

| | |
|---|---|
| `MeshWeaver.Graph.Test` | **1098/1098** |
| `MeshWeaver.PluginCatalog.Test` | **270/270** (carries the flaky test) |
| Release `-warnaserror` clean | `MeshWeaver.Graph` (`--no-incremental`), `Graph.Test`, `PluginCatalog.Test`, `FutuRe.Test` |

All runs `DOTNET_PROCESSOR_COUNT=4`, Release. `MeshWeaver.Hosting.Monolith.Test` + `FutuRe.Test` regression runs were in flight at push time; CI covers them.

No What's New entry: the user-visible symptom (a NodeType stuck on the compilation-error overlay) is real, but it is reached only through a teardown race that no released build has been observed to hit in production — this is a CI-integrity fix. Say the word and I will add a `Category: Fix` line.

### DIFF
--- a/.github/workflows/dotnet-test.yml
+++ b/.github/workflows/dotnet-test.yml
@@ -65,11 +65,49 @@
 
 name: MeshWeaver Build and Test
 
+# ──────────── EVERY PR, WHATEVER ITS BASE — including a STACKED one ────────────
+# 🚨 The `pull_request` trigger deliberately carries NO `branches:` filter. Do not
+# re-add one. It used to say `branches: [ "main" ]`, and the consequence (#1401) was
+# that a PR based on ANOTHER PR's branch — a stacked PR, which is the recommended way
+# to split a large change into reviewable pieces — never ran this suite at all.
+#
+# The failure mode is the dangerous half: an ABSENT check looks like nothing to wait
+# for. There is no "skipped", no pending, no marker of any kind — a stacked PR renders
+# as a short, all-green check list, visually IDENTICAL to one that ran the full suite
+# and passed. #1398 → #1399 → #1400 were split exactly as intended and two of the three
+# reached `MERGEABLE/CLEAN` with zero runs in their entire history.
+#
+# Three things make that state stickier than it looks, all confirmed on those PRs:
+#   * GitHub retargets a child PR to `main` only when the parent's branch is DELETED
+#     on merge. Branch deletion is off here, so a child keeps pointing at a merged
+#     branch indefinitely — "it gets tested later, just out of order" is not real.
+#     Left alone, a stacked PR was never tested at ANY point, including at merge.
+#   * `gh pr edit --base main` fires `pull_request.edited`, which is NOT in the default
+#     type set (opened/synchronize/reopened). The base changed, the checks stayed
+#     absent, and the PR still read CLEAN. Only close+reopen ever queued a run.
+#   * The `main pr protection` ruleset is scoped to `~DEFAULT_BRANCH`, so a PR whose
+#     base is not `main` is outside it: the `Consolidate test results` required check
+#     is not required there, and its absence therefore does not even block the merge.
+# So the pre-fix state had no signal at all: nothing ran, nothing was pending, nothing
+# was required. Running unconditionally is what turns that silence back into evidence.
+#
+# What a stacked run tests: GitHub builds `refs/pull/N/merge`, i.e. the branch merged
+# with ITS BASE. For a stacked PR that base is the parent's branch, so the run tests
+# the layer's own diff on top of the parent — real coverage of the code being reviewed.
+# That is the same "tested against a base that is not main's tip" posture the ruleset
+# already sanctions repo-wide (`strict_required_status_checks_policy: false`).
+#
+# Cost is small and bounded, because the green-tree reuse below is keyed on the TREE:
+# when the parent merges and main lands the identical tree, `precheck` finds the marker
+# the stacked run wrote and skips build+test. A stacked series pays for one full suite
+# per layer, not two.
+#
+# `push:` stays `main`-only on purpose — a full suite on every feature-branch push
+# would double every PR's cost for no extra evidence (the PR run already covers it).
 on:
   push:
     branches: [ "main" ]
-  pull_request:
-    branches: [ "main" ]
+  pull_request:      # ← NO `branches:` filter. See the block above (#1401) before adding one.
   workflow_dispatch:
 
 # A force-push to a PR branch obsoletes the in-flight run — cancel it instead of
--- a/src/MeshWeaver.Graph/Configuration/CompilationCacheService.cs
+++ b/src/MeshWeaver.Graph/Configuration/CompilationCacheService.cs
@@ -277,6 +277,29 @@ internal sealed class NodeAssemblyLoadContext : AssemblyLoadContext, IDisposable
     // here: a live context roots its own assemblies natively, so the target only dies when the
     // whole context does.
     private WeakReference<Assembly>? _loadedAssembly;
+    // 🚨 TWO states, not one, and the difference is load-bearing.
+    //
+    //   _unloading — Dispose has STARTED. New scans are refused from here (Pin), but the
+    //                LoaderAllocator is still fully alive: Dispose drains the in-flight pins
+    //                BEFORE it calls Unload(), so anything already pinned is still loadable.
+    //   _disposed  — the drain is over and Unload() is imminent/underway. Now a load is
+    //                genuinely unsafe.
+    //
+    // Collapsing the two into `_disposed` is what made a pin worthless at the one moment it
+    // mattered. Pin() re-checks the flag after incrementing, so a pin that was HANDED OUT proves
+    // the context was live — and then Dispose could set the flag one instruction later and
+    // LoadNodeAssembly, which guards on the same flag, threw ObjectDisposedException into a
+    // scan that Dispose was at that very moment WAITING for. Nothing was wrong with the
+    // assembly; the teardown had simply been requested.
+    //
+    // That throw is not survivable downstream. CompileResultFromAssembly catches it, records
+    // CompilationStatus.Error, and the compile watcher PARKS the NodeType — "further activations
+    // serve the cached error without recompiling" — so a millisecond-wide teardown race
+    // permanently kills a type until someone rebuilds it by hand (StaleStampRootBindingTest,
+    // 6× on 08-10 and twice on 08-13 including on main). Same "one mis-read is not transient"
+    // shape as #1387, reached from the opposite side: there the bytes were incomplete, here the
+    // bytes were perfect and we refused to read them.
+    private volatile bool _unloading;
     private volatile bool _disposed;
     // In-flight SCANS of this context's loaded assembly (the GetTypes / MeshNodeProviderAttribute
     // reflection / Activator block in MeshNodeCompilationService). Dispose() drains these before
@@ -329,13 +352,13 @@ internal sealed class NodeAssemblyLoadContext : AssemblyLoadContext, IDisposable
     /// </summary>
     public IDisposable Pin()
     {
-        if (_disposed)
+        if (_unloading || _disposed)
             throw new ObjectDisposedException(Name, "Cannot scan an assembly whose context is unloading");
         Interlocked.Increment(ref _pins);
-        // Re-check after the increment: Dispose sets _disposed then drains _pins, so a pin taken
+        // Re-check after the increment: Dispose sets _unloading then drains _pins, so a pin taken
         // concurrently with Dispose must not slip past the drain. If Dispose already started, back
         // out and throw — Dispose's drain will observe the decrement and proceed.
-        if (_disposed)
+        if (_unloading || _disposed)
         {
             Interlocked.Decrement(ref _pins);
             throw new ObjectDisposedException(Name, "Cannot scan an assembly whose context is unloading");
@@ -387,7 +410,10 @@ internal sealed class NodeAssemblyLoadContext : AssemblyLoadContext, IDisposable
     {
         lock (_loadLock)
         {
-            if (_disposed)
+            // Both flags: once Dispose has passed its lease check it is committed to unloading
+            // (draining scans, then Unload), so a new lease could not keep the context alive.
+            // This is the pre-split behaviour — Dispose used to set _disposed at that same point.
+            if (_disposed || _unloading)
                 return NullLease.Instance;
             _leases++;
             return new LeaseScope(this);
@@ -398,7 +424,7 @@ internal sealed class NodeAssemblyLoadContext : AssemblyLoadContext, IDisposable
     {
         bool unloadNow;
         lock (_loadLock)
-            unloadNow = --_leases == 0 && _unloadRequested && !_disposed;
+            unloadNow = --_leases == 0 && _unloadRequested && !_disposed && !_unloading;
 
         if (unloadNow)
             Dispose();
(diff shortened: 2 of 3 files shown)


rbuergi:
Regression runs that were in flight at push time have landed — all green, `DOTNET_PROCESSOR_COUNT=4`, Release:

| project | result |
|---|---|
| `MeshWeaver.Graph.Test` | 1098 / 1098 |
| `MeshWeaver.PluginCatalog.Test` | 270 / 270 (carries the flaky test) |
| `MeshWeaver.Hosting.Monolith.Test` | 647 passed, 3 skipped, **0 failed** (7 m 4 s) |
| `MeshWeaver.FutuRe.Test` | 59 / 59 |

Release `-warnaserror` clean for `MeshWeaver.Graph` (`--no-incremental`), `Graph.Test`, `PluginCatalog.Test`, `FutuRe.Test`. (`Hosting.Monolith.Test` reports 2 × `NU1903` — the pre-existing SSH.NET advisory, present on `main` and unrelated to this diff.)

One note for whoever runs the merge train: **#1402 is currently red on the very bug this PR fixes** (`StaleStampRootBindingTest`, run `31686258015`) — which is how the stack trace was found. Since this PR is stacked on #1402, the order still has to be #1402 first; #1402 will need a green run of its own, and after this fix lands that flake should stop producing them.

rbuergi:
## CI red — and I am not going to claim it is unrelated

Run `31690007487`, shard 4. **269 of 270** in `MeshWeaver.PluginCatalog.Test`.

**The good half — the fix does what it says.** `StaleStampRootBindingTest.StaleStampSelfTypedRoot_RootServesItsTypesArea`, which failed on `main` (`31683003325`) and on #1402 (`31686258015`), **passed**, and the string `Cannot load assembly from disposed context` does not appear anywhere in the shard log.

**The failure** is a different test with a different signature — `NodeRepoInstanceOrderingTest.TypedInstances_UnderAndBesideTheirInPackageType_InstallOnFreshMesh`:

```
System.TimeoutException  at NodeRepoInstanceOrderingTest.Read(String path)          line 86
[Warning] [MeshDataSource] Own-node refresh from change notification failed on Pack/Widget/Nested
  ---> HubDisposingException: Hub Pack/Widget/Nested is shutting down — cannot create '/MeshNode'.
```

Hub lifecycle, not assembly loading — a different subsystem from this diff, and the same family as the *other* root under `StaleStampRootBindingTest` (`root ShopA never answered after its recycle`, main at 08:39Z). `Read()` is `GetMeshNodeStream(path).Where(n => n is not null).FirstAsync().Timeout(30s)`, which times out if the instance hub recycles underneath it.

**But there is one causal path from this diff that I cannot exclude, so I am naming it rather than burying it.** The fix makes a NodeType that previously PARKED reach `Ok` instead. A type reaching `Ok` is what makes the rebind/overlay watchers **recycle its instances** — so more of the install now proceeds and more recycles now fire than on a run where the type died early. That is the fix working, but it does mean this diff can change *which* recycles happen, and `Read()` races exactly that.

What the evidence supports:

| | result |
|---|---|
| with the fix, locally (`MeshWeaver.PluginCatalog.Test`, ×2) | this test **passed both times** — 270/270 |
| without the fix, locally | 270/270 |
| without the fix, CI (#1402) | passed (that run died on `StaleStampRootBindingTest`) |
| with the fix, CI (this run) | **failed**, once |

So it is not deterministic, and it is not a new or fragile test (added 2026-07-19, long green on `main`). One occurrence of one test is precisely the shape of the population #1384 is about — 30 distinct tests failing once each across the sweep window. I am **not** re-running to find out: per AGENTS.md a re-run to "see if it was a flake" is what hides real races, and a single green would not distinguish the two explanations anyway.

**Not merging.** If a reviewer wants this attributed before it lands, the discriminating experiment is to run `NodeRepoInstanceOrderingTest` against `main` under load — if it fails there too, it is population; if it only ever fails with this diff, the recycle path above is the thing to fix next, and it is a pre-existing unsound wait in `Read()` either way.

rbuergi:
## Running the discriminator, and a confound I had to remove first

Branch synced with `main` (`bad2d2726`) — it was 4 PRs behind, and two of them (#1404/#1406 build-coordination, #1407 circuit-drain) touch `MeshWeaver.Graph` and `MeshWeaver.Hosting`'s pre-warmer. An A/B where the control arm carries those and the treatment arm does not is not an A/B. Diff against `main` is still exactly the same three files. `MeshWeaver.Graph.Test` re-verified on the merged baseline: **1098/1098**, Release `-warnaserror` clean.

Three `flake-repro.yml` arms now looping `NodeRepoInstanceOrderingTest.TypedInstances_UnderAndBesideTheirInPackageType_InstallOnFreshMesh` ×30 on the real 2-vCPU runner:

| run | ref | baseline | arm |
|---|---|---|---|
| [31692623885](https://github.com/Systemorph/MeshWeaver/actions/runs/31692623885) | `main` | `882475426` | **control** — no fix |
| [31693193614](https://github.com/Systemorph/MeshWeaver/actions/runs/31693193614) | this branch | `bad2d2726` | **treatment** — same baseline + the fix |
| [31692635186](https://github.com/Systemorph/MeshWeaver/actions/runs/31692635186) | this branch | `fd3ebc369` | pre-sync, kept as a bonus data point |

## Meanwhile, the documentary answer is already in

**#1405**, filed at 09:47Z — *before this branch existed* (first pushed 09:55Z) — from PR #1403, a **portal-only Blazor change** that touches neither the plugin catalog nor the mesh data plane. It is the same failure, down to the source line:

| | #1405 | this run (31690007487) |
|---|---|---|
| exception | `HubDisposingException: Hub Dependent/Item is shutting down — cannot create '/MeshNode'` | `HubDisposingException: Hub Pack/Widget/Nested is shutting down — cannot create '/MeshNode'` |
| frame | `MeshNodeStreamHandle.UpdateOwn` — `MeshNodeStreamExtensions.cs:678` | `MeshNodeStreamHandle.UpdateOwn` — `MeshNodeStreamExtensions.cs:678` |
| preceded by | `[MeshDataSource] Own-node refresh from change notification failed on Dependent/Item` | same, `Pack/Widget/Nested` |
| test wait | `GetMeshNodeStream(path).Where(n => n is not null).FirstAsync().Timeout(30s)` | identical helper |

#1405's own correlation table already lists **run 31686258015 — #1402's red** — as a member of this family. So the maintainer had independently attributed my other failure here before I looked.

Recorded on #1405 as two further occurrences: this one, and the *second* root under `StaleStampRootBindingTest` (`root ShopA never answered after its recycle`).

## On the causal path I refused to rule out

It is still the right thing to have flagged, and I am not retracting it — but it is now bounded on both sides:

- **The mechanism pre-exists this diff**, demonstrated on a portal-only PR that cannot have reached the ALC code at all.
- **A parked type produces a different failure, not this one.** When a type parks, its instances never land typed — the test fails on the `NodeType.Should().Be("Pack/Widget")` assertion, not on a recycle timeout. This signature requires the type to have reached `Ok`, which is the path `main` takes on every green run. So the fix moves runs from an abnormal park **toward** main's normal path; it does not open a path main does not already take.

The arms will say whether that holds empirically. Result posted here when they land.

rbuergi:
## Discriminator: **it fails on `main`.** Not caused by this diff.

`flake-repro.yml`, `NodeRepoInstanceOrderingTest.TypedInstances_UnderAndBesideTheirInPackageType_InstallOnFreshMesh` ×30 on the real 2-vCPU runner.

| arm | ref / baseline | result |
|---|---|---|
| **control — no fix** | [`main` @ `882475426`](https://github.com/Systemorph/MeshWeaver/actions/runs/31692623885) | 🔴 **FAILED at iteration 24/30** (23 passes, then the flake) |
| treatment — with fix | [branch @ `fd3ebc369`](https://github.com/Systemorph/MeshWeaver/actions/runs/31692635186) | 30/30 passed |
| treatment — with fix, synced baseline | [branch @ `bad2d2726`](https://github.com/Systemorph/MeshWeaver/actions/runs/31693193614) | running |

The control-arm failure is the **identical** signature, on code that does not contain a line of this diff:

```
DeliveryFailureException: Hub Pack/Sibling is shutting down (RunLevel=ShutDown)
   at MeshNodeStreamHandle.UpdateOwn (MeshNodeStreamExtensions.cs:678)
[10:59:36] === TEST FAILED: The operation has timed out.
Failed! - Failed: 1, Passed: 0 - Duration: 31 s
```

**The harness genuinely ran** — this workflow's own header warns that an earlier-step failure leaves the loop `skipped` while the job reports `failure`, which is indistinguishable from a real reproduction. It is not that: `Set up job` → `checkout` → `Free disk space` → `Setup .NET` → `Restore workloads` → `Restore dependencies` → `Build` all **succeeded**, and the step that failed is `Loop the suspect test until it fails`, at a named iteration, with a trx.

### What I am *not* claiming

30/30 on the treatment arms is **not** evidence the fix helps this test. At the control arm's observed rate (1 in 24), a clean run of 30 has roughly a 1-in-4 chance of happening anyway. The claim the experiment supports is the narrow one, and it is the one that was in question: **the failure occurs without this change, so this change did not cause it.**

### The causal path I flagged, resolved

I said I could not exclude that the fix, by letting a type reach `Ok` where it previously parked, starts exercising recycle paths a parked type never reached — and that *"my diff doesn't touch that subsystem" is not a defence when the diff changes which code runs*. That was right to raise, and it is now settled from three directions:

1. **Empirically** — the control arm reproduces it with the fix absent.
2. **Documentarily** — #1405, filed 09:47Z from PR #1403, a portal-only Blazor change, records the same failure down to `MeshNodeStreamExtensions.cs:678`.
3. **Mechanically** — a parked type produces a *different* failure. Its instances never land typed, so the test dies on `NodeType.Should().Be("Pack/Widget")`, not on a recycle timeout. This signature requires the type to have reached `Ok` — which is exactly what `main` does on every green run. The fix moves runs from an abnormal park **toward** main's normal path; it does not open a path main does not already take.

Home for the failure: **#1405** (recycle-during-install race, `MeshNodeStreamHandle` leaves the caller's stream silently empty instead of resubscribing on reactivation). Logged there as a further occurrence.

rbuergi:
Third arm landed — completing the table:

| arm | ref / baseline | result |
|---|---|---|
| **control — no fix** | `main` @ `882475426` | 🔴 **failed at iteration 24/30** |
| treatment — with fix | branch @ `fd3ebc369` (pre-sync) | 30/30 |
| treatment — with fix | branch @ `bad2d2726` (same baseline as control) | **30/30** |

Same runner class, same filter, same iteration count, and the treatment arm now shares the control's exact `main` baseline — so the only variable left between the red arm and the green one is this diff, and the **red** one is the arm **without** it.

Restating the limit so nobody over-reads the table: 60 clean iterations across two treatment arms is not proof the fix *reduces* this flake — at the control's ~1-in-24 rate, 30 clean runs happen by luck about a quarter of the time. The supported claim is only the one that was in dispute: **this diff did not cause it.**

rbuergi:
## Green on the synced commit — and the shards genuinely ran

[Run 31693189761](https://github.com/Systemorph/MeshWeaver/actions/runs/31693189761) on `bad2d2726`: `MeshWeaver Build and Test` = **COMPLETED/SUCCESS**, `mergeStateStatus` = **CLEAN**.

Checked against the green-tree-reuse trap, because this PR **adds a test** and a suite can conclude `SUCCESS` with `Run tests` skipped:

```
Build solution (once): success
Run tests (shard 0..5): success, success, success, success, success, success   ← none skipped
Consolidate test results: success          ← the required check
Plugins compile against this PR: success
```

and the assembly that carries the new test actually executed, on shard 2:

```
MeshWeaver.Graph.Test  Total: 1098, Errors: 0, Failed: 0, Skipped: 0, Not Run: 0, Time: 52.964s
[CI] MeshWeaver.Graph.Test exit=0
```

1098 is the same total as the local runs, i.e. it includes `PinnedScan_StillLoadsTheAssembly_WhileDisposeIsDraining` — so the new regression test ran on CI and passed, rather than riding a reused green tree.

`StaleStampRootBindingTest` — the test that failed on `main` and on #1402 with the `ObjectDisposedException` this PR fixes — is green here, and the string `Cannot load assembly from disposed context` does not appear anywhere in the run.

Not merging — handing back per the merge train.

Issue #648, opened 2026-07-25
Acked-success write loss via stale second per-node activation (MergeGuard refuses keys, flushes stale state, acks Success)


rbuergi:
Third (and last known) mechanism behind the `TwoSiloRecycleConvergenceTest.PostRecycleUpdate_NonOwnerSiloMirror_ConvergesInsteadOfOrphaning` flake, surviving #635 and #647.

**Evidence** (identical stack + stdout in both):
- Pre-fix main: run 30159928718 **attempt 1**, shard 2 (job 89684164075, 2026-07-25 13:47Z).
- Local repro during #647 verification (silo logs, `Secondary_1` lines 209-210).

**Mechanism:** after the test disposes the owner silo's hosted hub, a **stale second per-node activation on the other silo** receives the post-recycle `PatchDataRequest`. The three-way MergeGuard refuses every content key ("refused stale/reordered cross-hub write to 'apiKey' … kept the newer live value") — and then **commits + flushes the stale state over the store and acks Success**. The resilient caller stops retrying: the loss is acked as success, so no NACK-based machinery (#647's disposal-NACK + late-NACK re-enqueue) can catch it.

**Fix direction:** single-writer activation invariant for per-node hubs (#325 split-brain family) — a per-node patch must only ever be applied by the CURRENT owner activation; a superseded activation must refuse-and-redirect (NACK with a retryable code), never merge-and-ack. Also: a MergeGuard that refuses every key of a patch must not ack that patch as Success.

Related: #325, #635, #647.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

rbuergi:
## A reproduced instance of this class, with the store dump — and it is a **rollback**, not just a lost ack

Chasing `TwoSiloRecycleConvergenceTest` as a CI flake led straight back here. It is not a flake: **an owner grain's idle-recycle durably rolls the node back to its creation state.** Captured live at the moment of failure:

```
[pre-recycle] mirror at ApiKey=sk-v6, node.Version=12   ← 6 writes acked; store verified >= v12
[recycle]     disposed 1 owner hosted hub(s) on silo 0
[PROBE] UpdateApiKeyResilient returned after 93 ms      ← the write reports SUCCESS
[PROBE] STORE-DUMP present=True version=2 apiKey=sk-v0  ← store is back at the CREATION state
```

[...]

PR #1414, opened 2026-08-13
perf: cross-hub patch ships a string SPLICE, not the whole value (chat half of #1284/#1351)

rbuergi:
Closes the **chat half** of the quadratic-write class — #1284, #1351. The activities half shipped in #1398/#1399/#1400 (bounded window + segment satellites).

## Chat is a different shape, and the satellite migration does not apply

`Thread.Messages` already holds only ids (~26 B each): a 200-message thread is ~10 kB of head against the activity log's 88.7 kB at 500 — two orders of magnitude from being the problem. Message bodies are already `ThreadMessage` satellites. And a streaming cell **must** show all the text so far, so its value legitimately grows; no satellite can change that.

**The cost is the streaming write, not the content shape.** `ThreadExecution` streams through `Sample(100 ms)` into `PushToResponseMessage`, which rewrites the cell's whole content per tick. Cross-hub that is a double whammy:

- `ComputeMergePatchDiff` emits the **entire new text**, and
- `MeshNodePatchMerge.ExtractBaseValues` emits the **entire previous text** next to it.

So one round costs `O(ticks × final length)`. Measured below: **3.84 MB for a 20 kB answer**, all of it Large-Object-Heap traffic on the writer *and* the owner. Widening `Sample(100 ms)` or dropping ticks would be the band-aid; the patch simply does not need to carry text the owner already has.

## The change

A changed **string** leaf at least `PatchStringSplice.MinSpliceLength` (1 kB) long, whose change is smaller than the value, now travels as a splice:

| where | shape |
|---|---|
| in the patch | `{ "$sd": [start, removedLength, "inserted"] }` |
| in `BaseValues`, same leaf | `{ "$sdb": [baseLength, "fingerprint"] }` |

Both halves are load-bearing — splicing only the patch leaves the base re-shipping the previous value every tick and the round stays quadratic.

Three files carry it: the producer (`ComputeMergePatchDiff`), the base extractor + owner-side resolver (`MeshNodePatchMerge`), and the base-less RFC 7396 fallback (`MergePatchRecursive`, so a marker can never be written verbatim into a string field on *any* apply path). New codec: `PatchStringSplice`, next to the existing `StringDelta` / `StringDeltaPatch` it shares the `$sd` marker with.

Below 1 kB, and for arrays and scalars, nothing changes. Arrays in particular still ship their full base — `MergeArray` genuinely consumes it, so that door stays shut exactly as #1351 established.

## Measured: bytes serialised per streaming round

Real runs of the production cross-hub write path (`GetMeshNodeStream(path).Update` → shared stream cache → `PatchDataRequest` → owner merge), counting the actual `PatchDataRequest` payloads as the owning node hub receives them, at two answer sizes. The "before" column is the same test against `main`'s code, not a model — it **fails** there and passes here.

|                              | 5,000 chars / 50 ticks | 20,000 chars / 200 ticks |
|------------------------------|-----------------------:|-------------------------:|
| **before** — patch bytes     |                131,394 |                2,025,576 |
| **before** — base bytes      |                126,394 |                2,005,576 |
| **before** — total wire      |    **257,788** (0.25 MB) |  **4,031,152** (3.84 MB) |
| **after** — patch bytes      |                 14,072 |                   43,408 |
| **after** — base bytes       |                  9,722 |                   26,473 |
| **after** — total wire       |     **23,794** (0.02 MB) |     **69,881** (0.07 MB) |
| **reduction**                |                 10.8× |                **57.7×** |
| before — bytes per answer char |                  51.6 |    **201.6** (grows 4× with length) |
| after — bytes per answer char |                    4.8 |      **3.5** (falls — amortises) |

The shape assertion is the important one: quadrupling the answer must not quadruple the cost *per character*. Before, it does exactly that (51.6 → 201.6); after, the per-character cost falls as the per-tick constant amortises.

## Correctness — chat is the most user-visible surface in the product

**The owner applies a splice only when it can prove the offsets still address the right characters.** The fingerprint (exact length + truncated SHA-256) is compared against the owner's live text:

- **match** → the live text *is* the text the writer diffed against, so the splice yields byte-for-byte the string a whole-value patch would have written;
- **mismatch** → **refuse**. Keep the newer live value, NACK `Conflict` — which re-runs the writer's update lambda against fresh state and re-diffs, the same self-healing route a refused scalar already takes.

**Concurrent splices therefore cannot corrupt the string.** A splice is never applied at an offset the fingerprint did not vouch for, so two mirrors splicing the same text do not interleave: the first lands whole, the second is refused whole and re-diffs. Stated plainly rather than left implicit — this is *not* "last splice wins on that field"; it is "the loser never lands, and is told to start again from current".

The cost is honest and worth naming: a **disjoint** concurrent edit to a big string, which the full-value three-way merge would have rebased and landed in one hop, now costs a round trip. For streaming cells, markdown bodies and prerendered html that trades a rare extra hop against a constant per-tick, per-viewer megabyte.

Tests:

- `PatchStringSpliceTest` (15) — exactness against the true base; 200 sequential splices reproducing the whole streamed text; fingerprint mismatch refuses rather than splicing blind; two mirrors concurrently; non-string live value; no base signal; a full base carried with a splice still gets the three-way rebase; the strict decoder; and both fallback-path cases.
- `StreamingCellWriteByteCountTest` — the measurement above, plus a byte-identity assertion on the settled cell at 20 kB / 200 splices.
- `StreamedTextIntegrityTest` — a **real streamed round** through `ThreadExecution` with a 6 kB answer over 24 chunks: the finished text is byte-identical to what the model streamed, every observed state is a **prefix** of it and never shrinks, and a **late subscriber** on a fresh client reads the whole text. (The framework's short pre-stream placeholders are exempt by a length bound, never by matching their strings — they are localizable, and a length rule cannot excuse a long corrupted state.)

## Rolling deploys — why the marker rides inside the patch

During a rollout a new writer can reach a per-node hub still on the old image. That owner does not decode `$sd`, merges the marker into a string-typed field, and the merged node then **fails to deserialize** — the write is NACKed and never commits. Loud, non-destructive, and self-healing: the mirror never advanced, so the next splice is computed against the same base and lands the whole missing span once the owner is upgraded.

The obvious alternative — carrying splices in a new sibling property an old owner would simply ignore — degrades instead into an owner that **acks a write it only partly applied**, i.e. the acked-write-loss class of #648. A visibly stalled field beats a silently lost one, so the in-patch marker is the deliberate choice.

## Deliberately out of scope, stated rather than implied

- **Owner → subscriber fan-out.** `JsonSynchronizationStream.CreateSingleObjectPatch` still emits an RFC 6902 `replace` carrying the whole string, per tick, per subscriber. Different wire format, different consumers (including the JS clients) — its own change.
- **#1351's ~2× transport constant.** `MessageDelivery.Package` serialises the payload to a string on the root-mesh route and the Orleans codec transcodes it back to UTF-8 through a doubling `PooledByteBufferWriter`. Real waste, but a transport concern rather than a patch-shape one; folding it in here would have made this measurement ambiguous. With the payload now ~58× smaller it is no longer material for chat, and it should be fixed where it matters — large payloads generally.
- **`Thread.SessionState`** is declared (`src/MeshWeaver.AI/Thread.cs:173`) and **never read or written anywhere** — the only hit in `src`, `test`, `clients`, `samples` and `content` is the declaration itself. Left in place rather than deleted: it is a public property on a content record, and in-mesh C# is invisible to the compiler (AGENTS.md), so removing it belongs in a change that can own that sweep.

## Verification

Local, `-c Release -warnaserror` on every touched project and their test assemblies:

| suite | result |
|---|---|
| `MeshWeaver.Data.Test` | 357 / 357 |
| `MeshWeaver.Graph.Test` | 1097 / 1097 |
| `MeshWeaver.Threading.Test` | 130 / 130 |
| `MeshWeaver.Hosting.Monolith.Test` | 653 passed, 3 skipped, 0 failed |
| `MeshWeaver.Documentation.Test` | 20 / 20 (link integrity + What's New) |


rbuergi:
Both taken — thanks, the first one in particular was a real precision bug in the measurement's own metric.

**UTF-8 byte counts.** `PatchWireRecorder` called its metric "bytes" while summing `string.Length`. For this test's ASCII prose the two coincide, so the conclusion never moved, but the payload travels as UTF-8 and a char count silently understates anything non-ASCII — which real chat is full of. Now `Encoding.UTF8.GetByteCount`, and I re-measured **both** columns under it (the "before" column is the same test against reverted src, which still fails there):

|                              | 5,000 chars / 50 ticks | 20,000 chars / 200 ticks |
|------------------------------|-----------------------:|-------------------------:|
| before — total wire          |    257,788 B (0.25 MB) |  4,031,144 B (3.84 MB) |
| after — total wire           |     23,795 B (0.02 MB) |     69,893 B (0.07 MB) |
| before — bytes per answer char |                  51.6 |   201.6 (grows 4× with length) |
| after — bytes per answer char |                    4.8 |     3.5 (falls) |

**Thread creation target.** Correct that `Mesh.Address` was the wrong choice, though the convention is more mixed than "consistently" suggests — `ChatHistoryTest`, `ThreadResumeTest` and `UpdateEmitsDuringRoundTest` also use `Mesh.Address`. The substantive point stands: `ThreadCreationTest`, the whole Orleans thread suite and `DelegationCompletionTest` target the context namespace, that is what a real client does, and routing node lifecycle through the root mesh hub is discouraged here. Switched to `new Address(ContextPath)`; the test passes unchanged.

Both in `bdc218cb7`, and the branch is now merged up to current `main`. Re-verified on the merged tree: `MeshWeaver.Data.Test` 359/359, `MeshWeaver.Threading.Test` 130/130.

Issue #458, opened 2026-07-13
Side-panel / new-chat composer stuck at ~80 px — #178 auto-grow inline height overrides the .no-messages fill layout


sierragolflima:
## Summary

In the right-hand side-panel AI chat, the message composer is stuck at roughly 80 px
(≈3.5 lines) and does **not** grow — whether you type or paste more than ~3.5 lines — even
though the side panel is much taller and has ample empty space above/below the box. This is
a **new-chat / side-panel-specific** defect, distinct from #178 (which correctly fixed
auto-grow for an ongoing thread with messages).

## Screenshot

<img width="300" height="500" alt="Image" src="https://github.com/user-attachments/assets/32e2fb84-150b-409d-a6f2-80f40d3996e8" />

## Steps to reproduce

1. Open the AI chat in the right-hand side panel (a fresh/new chat).
2. Type or paste content longer than ~3.5 lines.
3. Observe: the editable box stays at ~80 px and scrolls internally; it never expands to
   use the visible empty space.

## Root cause (code)

The side panel always mounts the **new-chat** `ThreadChatView`
(`src/MeshWeaver.Blazor.Portal/Chat/ThreadSidePanelContent.razor:65`), which runs the view
in its **`.no-messages` state**. That state applies a deliberately different composer
layout than a thread-with-messages:

- **`.no-messages`** (side panel / new chat): `.input-container { flex: 1 1 auto }` —
  intended to let the input **fill the tall empty slot**
  (`src/MeshWeaver.Blazor.Portal/Chat/ThreadChatView.razor.css:20-39`, per its own comment:
  "the input field grows to occupy the space rather than sitting as a thin bar").
- **with messages** (main pane): `.input-container { flex: 0 0 auto }` — a natural-height
  bottom bar, which is the case #178's auto-grow was written and tested against.

The #178 auto-grow change pins the editor's outer container to an inline
`height: 80px; max-height: min(400px, calc(100vh - 280px)); min-height: 80px`
(`src/MeshWeaver.Blazor/Components/Monaco/MonacoEditorView.razor:158`, the `AutoGrow`
`ContainerStyle`), applied by
`src/MeshWeaver.Blazor.Portal/Chat/ThreadChatView.razor:947-949`
(`Height="80px" MaxHeight="min(400px, calc(100vh - 280px))" AutoGrow="true"`).

That inline `height: 80px` wins over the `.no-messages` fill intent, so the editable box
[...]

PR #1417, opened 2026-08-13
fix: seed AccessContext.Locale for anonymous sessions from Accept-Language

rbuergi:
## The problem

"The viewer's language wins for chrome" is implemented in the plugins (Store #458, Edu/CourseInvite #463) and resolves off `AccessContext.Locale`. That field is populated from the signed-in user's **profile** — and an anonymous visitor has no profile. So it was `null` for every one of them, and the portal answered in English whatever their browser asked for.

The decision was therefore **inert for exactly the audience it was written for**: a first-time visitor hitting a paywall, an invite or a public course page is anonymous *by definition*.

`Accept-Language` was read nowhere in the framework — no middleware, no `RequestLocalization`, no culture cookie, no `?lang=`.

## The fix

Seed `AccessContext.Locale` from the request's `Accept-Language`, negotiated against `Locales.Supported`, on **both** entry paths.

| Path | Where | Reads the header from |
|---|---|---|
| SSR / HTTP request | `UserContextMiddleware` | `HttpContext.Request.Headers.AcceptLanguage` |
| Blazor circuit | `CircuitAccessHandler` **constructor** | `CircuitRequestLanguage`, published per hub invocation by the global `CircuitRequestLanguageFilter` |

`Locales.Negotiate` parses the full RFC 9110 header (comma list, `q=` weights, descending, ties in header order) and matches each tag through **`Locales.TryMatch`** — so region variants fold onto the **primary subtag** exactly as everywhere else (`en-GB` → `en`, `de-CH` → `de`), and the harness and the product agree by construction rather than coincidence. `q=0` (an explicit refusal) and `*` (an absence of preference) are skipped. Nothing matched returns **`null`, not `"en"`**, so "unsupported" stays distinguishable from "asked for English".

### Why the circuit reads the SignalR connection, not `IHttpContextAccessor`

The first cut used `IHttpContextAccessor`. Extending the test to cover **both** SignalR transports showed that it works over WebSockets — the upgrade request stays in flight for the connection's life — and returns **nothing under long polling**, where every poll is a separate request that ASP.NET disposes, nulling the accessor's holder for every flow that captured it. A browser behind a proxy that blocks WebSockets falls back to long polling, i.e. exactly a corporate network, so shipping that would have meant the fix reaches most anonymous visitors and silently misses the rest.

SignalR keeps an `IHttpContextFeature` on the **connection** and refreshes it per request, so `HubCallerContext.GetHttpContext()` answers for every transport. `CircuitRequestLanguageFilter` reads it in the hub invocation that *creates* the circuit handlers and publishes it on `CircuitRequestLanguage` — an `AsyncLocal` held as an instance field on a mesh-scoped singleton (the same shape `AccessService` uses for per-circuit identity; never `static`, so two circuits opening concurrently cannot see each other's language), memoised per connection so the parse happens once rather than per message. `IHttpContextAccessor` remains only as a fallback for a host that runs the Blazor hub without the filter.

The filter is registered **globally** because Blazor's `ComponentHub` is `internal` and cannot be named in a per-hub registration; `ComponentHub` declares no filters of its own, so the global list applies. On any other hub it is a memoised no-op (the repo has no other hub filters).

### Does the seed land before the first render? **Yes.**

This is the question the change lives or dies on, because `LayoutAreaHost` captures the access context in its **constructor** and `CircuitContext` is a plain non-observable `AsyncLocal` — a locale arriving mid-circuit re-renders nothing.

- The header is read in the handler's **constructor**. Blazor resolves the circuit's `CircuitHandler`s from inside a `ComponentHub` hub method — `StartCircuit` for a classic Blazor Server payload, and the first `UpdateRootComponents` for a **Blazor Web App** (which this portal is), where handler creation is *deliberately* deferred so a handler can see the restored application state. That invocation is exactly what the filter wraps.
- The framework then runs `OnCircuitOpenedAsync` → `OnConnectionUpAsync` → **only then** `AddComponentAsync` (the first render). So the language is on the circuit identity before anything draws.
- The circuit is also the path that matters: the portal's SSR pass renders no page content at all (`Routes.razor` returns early until `BrowserDimensionWatcher` resolves the viewport), so a middleware-only fix would have been inert for the paywall.

### A stored preference still wins

Both paths now share **one** projection, `MeshUserProjection.Apply`: identity from the node, preferences from the profile *when set*, falling back to the seed when not. The two paths used to project differently — the circuit read locale + time zone, the middleware only id + name. That was invisible while nothing seeded a locale, and would have rendered **German SSR chrome for an English-profile user** the moment one existed. `CircuitAccessHandler.ApplyMeshUser` is now a thin alias, so its existing tests still pin the same rule.

`CultureInfo.CurrentUICulture` is **not** consulted anywhere. The header is read once, at identity time, and never becomes a second ambient resolution mechanism.

## Verification

**`AnonymousCircuitLocaleSeedTest`** (`MeshWeaver.Hosting.Blazor.Test`) drives a **real SignalR connection into Blazor's real `ComponentHub`** on a real `WebApplication` over a real monolith mesh — real `CircuitAccessHandler`, real per-circuit `ICircuitContextAccessor`, real filter. A probe `CircuitHandler` ordered after it reads the resolved identity in `OnConnectionUpAsync`, i.e. at the last moment before the first render. No mocks.

Nothing short of a real connection can answer the question this change turns on; a unit test of the negotiation alone would stay green while the browser saw nothing — and it is precisely the real-connection test that caught the long-polling gap.

- `Accept-Language: en-GB` → `Locale == "en"` — **the paywall case: English chrome on a `de-CH` course.** Run over **WebSockets and long polling**.
- `Accept-Language: de-CH,de;q=0.9,en;q=0.8` → `"de"` — the mirror, which is what stops "always English" from passing. Both transports.
- `Accept-Language: fr-FR,fr;q=0.9` → `null`, no throw — unsupported falls back cleanly.
- No header → `null` — unchanged.
- Plus two pure precedence tests: a stored `en` beats a German browser; an unset profile falls back to the browser.

**Fail-before confirmed, twice.** (1) With `Locale = null` restored in the anonymous branch, the en-GB and de-CH cases fail with `Expected value to be "en" … but found <null>`; the two null-expectation cases stay green (handler byte-restored afterwards, verified by `diff`). (2) Against the accessor-only version, the two **long-polling** rows fail with the same `found <null>` — which is what drove the connection-scoped redesign.

`Locales.Negotiate` is additionally pinned in `LocalizationTest` over real header shapes (weights beating textual order, `q=0`, `*`, malformed weights, whitespace/casing, unsupported).

### Runs

- `MeshWeaver.Hosting.Blazor.Test` — **402/402** ✅ (8 in the new class, across both transports)
- `MeshWeaver.Messaging.Hub.Test` — **56/56** ✅
- `MeshWeaver.Hosting.Monolith.Test` `~LocalePreference` — **17/17** ✅
- `MeshWeaver.Documentation.Test` `~DocumentationLinkIntegrity` — **1/1** ✅
- Release + `-warnaserror`: `MeshWeaver.Messaging.Contract`, `MeshWeaver.Blazor`, `MeshWeaver.Hosting.Blazor`, `Memex.Portal.Monolith` (`-p:CIRun=true`), `MeshWeaver.Hosting.Monolith.Test`, both touched test projects — all `0 Error(s)`.

No new localization **keys** were added, so the `clients/react/src/i18n/` mirror is untouched by design.

## Noted, not fixed — for the education repo

The education harness has a latent version of this bug and should be picked up there (**not** touched by this PR):

- `80-primer.spec.ts:32` loops `CONTENT_COURSES` (which **does** include `AgenticPrimerDe`) and asserts the English `STORE.priceLine` / `STORE.enrollRedirect` at lines 126-131, while `StoreTexts.German.PricePaid` is `"**{0}** — voller Zugang zu allen Inhalten."`. Those pass today only because a `/Subscribe/` heading regex carries them — i.e. the assertion that would catch a language regression is not actually asserting language.
- `STORE.enrollRedirect = 'Taking you to enroll'` exists nowhere in the vendored plugin source — a stale literal.

Both become live once anonymous visitors genuinely render in their own language.

### DIFF
--- a/src/MeshWeaver.Blazor/Infrastructure/MeshUserProjection.cs
+++ b/src/MeshWeaver.Blazor/Infrastructure/MeshUserProjection.cs
@@ -1,0 +1,60 @@
+using System.Text.Json;
+using MeshWeaver.Mesh;
+using MeshWeaver.Mesh.Security;
+using MeshWeaver.Messaging;
+
+namespace MeshWeaver.Blazor.Infrastructure;
+
+/// <summary>
+/// The ONE projection of an authoritative mesh <c>User</c> node onto a seeded
+/// <see cref="AccessContext"/> — shared by every entry path that resolves a caller, so the SSR
+/// request path and the Blazor circuit path can never disagree about who a viewer is or what
+/// language and time zone they see the portal in.
+///
+/// <para>🚨 <b>Why shared rather than duplicated.</b> The two paths used to project differently:
+/// <c>CircuitAccessHandler</c> read <c>TimeZoneId</c> and <c>Locale</c> off the profile, while
+/// <c>UserContextMiddleware</c> took only <c>Id</c> and <c>Name</c> and left both null. That was
+/// invisible while nothing seeded a locale — every server-rendered string was English anyway. The
+/// moment a request-derived locale exists (<see cref="Locales.Negotiate"/>), the divergence becomes
+/// user-visible and WRONG: a signed-in user whose profile says English, browsing from a German
+/// browser, would get German chrome on the server-rendered shell and English inside the circuit.
+/// One projection, applied by both, makes "the stored preference wins" true by construction rather
+/// than by two call sites happening to agree.</para>
+///
+/// <para>Pure: a function of (seed, node, options) with no hub, circuit or HTTP state, which is why
+/// it can be pinned directly by tests.</para>
+/// </summary>
+public static class MeshUserProjection
+{
+    /// <summary>
+    /// Projects <paramref name="meshUser"/> onto <paramref name="seed"/>.
+    ///
+    /// <para>Identity (<c>ObjectId</c>, <c>Name</c>) always comes from the node — it is the
+    /// authoritative partition key, and a claim-derived guess must never outrank it. Preferences
+    /// (<c>TimeZoneId</c>, <c>Locale</c>) come from the node's profile WHEN SET, and otherwise leave
+    /// the seed untouched: an unset profile field means "this user never expressed a preference",
+    /// which must fall back to whatever the caller already knew (the request's
+    /// <c>Accept-Language</c>, say) rather than erase it to null.</para>
+    /// </summary>
+    /// <param name="seed">The context built from the authentication claims (and the request).</param>
+    /// <param name="meshUser">The authoritative mesh <c>User</c> node for that identity.</param>
+    /// <param name="options">Serializer options used to read the node's <c>User</c> content.</param>
+    public static AccessContext Apply(
+        AccessContext seed,
+        MeshNode meshUser,
+        JsonSerializerOptions options)
+    {
+        ArgumentNullException.ThrowIfNull(seed);
+        ArgumentNullException.ThrowIfNull(meshUser);
+        var profile = meshUser.ContentAs<User>(options);
+        var timeZoneId = profile?.TimeZoneId;
+        var locale = profile?.Locale;
+        return seed with
+        {
+            ObjectId = meshUser.Id,
+            Name = meshUser.Name ?? meshUser.Id,
+            TimeZoneId = string.IsNullOrWhiteSpace(timeZoneId) ? seed.TimeZoneId : timeZoneId,
+            Locale = string.IsNullOrWhiteSpace(locale) ? seed.Locale : locale
+        };
+    }
+}
--- a/src/MeshWeaver.Blazor/Infrastructure/UserContextMiddleware.cs
+++ b/src/MeshWeaver.Blazor/Infrastructure/UserContextMiddleware.cs
@@ -58,6 +58,15 @@ public class UserContextMiddleware(RequestDelegate next, ILogger<UserContextMidd
         var hub = context.RequestServices.GetRequiredService<PortalApplication>().Hub;
         var userService = hub.ServiceProvider.GetRequiredService<AccessService>();
 
+        // 🚨 THE ONLY THING WE KNOW about an ANONYMOUS visitor's language. Everything downstream
+        // resolves text off AccessContext.Locale (never an ambient culture — a hub render hops the
+        // scheduler and an AsyncLocal culture does not survive it), and an anonymous visitor has no
+        // profile to read it from. Without this seed the field is null for EVERY first-time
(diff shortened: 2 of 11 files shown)


rbuergi:
## Follow-up: the WebSocket-only gap is closed

While verifying, I extended the circuit test to run over **both** SignalR transports — and the long-polling rows **failed**. `IHttpContextAccessor` works over WebSockets (the upgrade request stays in flight for the connection's life) and returns nothing under long polling, where every poll is a separate request that ASP.NET disposes, nulling the accessor's holder for every flow that captured it.

That is not a cosmetic gap: a browser behind a proxy that blocks WebSockets falls back to long polling — exactly a corporate network. Shipping it would have meant the fix reaches most anonymous visitors and silently misses the rest.

**Root cause fixed rather than documented.** The header now comes off the SignalR **connection**, which is transport-independent: SignalR keeps an `IHttpContextFeature` on the connection and refreshes it per request, so `HubCallerContext.GetHttpContext()` answers for every transport. A global `IHubFilter` (`CircuitRequestLanguageFilter`) reads it in the hub invocation that *creates* the circuit handlers and publishes it on `CircuitRequestLanguage` — an `AsyncLocal` held as an instance field on a mesh-scoped singleton, the same shape `AccessService` uses for per-circuit identity (never static, so two circuits opening concurrently cannot see each other's language).

The filter is registered **globally** because Blazor's `ComponentHub` is `internal` and cannot be named in a per-hub registration; `ComponentHub` declares no filters of its own, so the global list applies. On any other hub it is a two-field no-op. `IHttpContextAccessor` remains only as a fallback for a host that runs the Blazor hub without the filter.

`AnonymousCircuitLocaleSeedTest` now runs the English and German cases over **both** transports; the long-polling rows are the regression pin.

### Re-verified
- `MeshWeaver.Hosting.Blazor.Test` — **402/402** ✅ (8 in the new class, both transports)
- `MeshWeaver.Documentation.Test` `~DocumentationLinkIntegrity` — **1/1** ✅
- Release + `-warnaserror`: `MeshWeaver.Hosting.Blazor`, `Memex.Portal.Monolith` (`-p:CIRun=true`), both touched test projects — `0 Error(s)`

PR #1427, opened 2026-08-13
The starvation stand-in was starving the SECURITY reads too, and hanging the create

rbuergi:
Shard 1 of #1418 failed `SourceDiscoveryUnavailableTest` and I re-ran it green, calling it a flake. Following up: **it is not a flake**, and the same root cause was silently costing this test ~26 seconds on every run.

## What actually happened

The failure was at **line 153** — the `CreateNode` of the *healthy* source node — with the starve exception and *"emitted nothing at all"*. The stand-in's own comment claims that cannot happen:

> 🚨 Matched on the `namespace:` FORM, not on the bare path, so only source DISCOVERY starves. Node creation and single-node reads under the same subtree use `path:`-shaped queries and must keep working

That is **false**. The security layer reads its policy and grants with `namespace:`-shaped queries over the *same* subtree:

```
namespace:…/StarvedSourceType/Source id:_Policy       nodeType:PartitionAccessPolicy
namespace:…/StarvedSourceType/Source/_Access          nodeType:AccessAssignment
```

so the marker starved those too. The CI log shows both evicted as faulted synced queries, and the hub snapshot shows the consequence:

```
Hub portal/nodeops-… Executing(CreateNodeRequest, 33014ms)
```

The create sat **33 seconds** waiting on security reads that could never answer, and the 30 s assertion gave up first. It passed whenever those queries were already cached and failed when they were cold — which is exactly why it reproduced on a CI shard and never locally (**10/10 green** on main before this change).

## The fix

`CodeQueryResolver` ANDs `nodeType:Code` into every source query it emits, and the security queries carry their own node types — so that clause separates them exactly. The starve now requires it.

## Proof it does not MASK the thing under test

- **Narrowed**: the test passes and the starved type still reaches `Unavailable`.
- **Starve disabled entirely**: the test **fails** — the type compiles `Ok` — so the source-discovery leg is provably still starved.

It also stops the test paying for a hang it never meant to cause: **30.2 s** (the CI soft-deadline warning, and ~30 s locally) → a stable **4 s** across 5 runs.

## 🚩 Worth a separate look

The product behaviour underneath is that **a starved security-policy read leaves `CreateNodeRequest` executing for 33 s** rather than failing fast. In a real cross-silo starvation — the very scenario this suite exists for (#1218) — node creation into the affected namespace would hang the same way. The faulted-query eviction machinery already recovers the *query*; the in-flight caller still waits. That is a robustness gap of its own and I have not touched it here.

### DIFF
--- a/src/MeshWeaver.Graph/Configuration/CodeQueryResolver.cs
+++ b/src/MeshWeaver.Graph/Configuration/CodeQueryResolver.cs
@@ -25,8 +25,10 @@ namespace MeshWeaver.Graph.Configuration;
 ///   <item>A <c>namespace:X</c> value with no <c>/</c> (e.g. bare <c>Source</c>
 ///     or <c>Test</c>) is rebased onto the NodeType's own path so the defaults
 ///     read as "my own Source / Test folder".</item>
-///   <item>Every emitted query is ANDed with <c>nodeType:Code</c> so non-code
-///     children never leak in.</item>
+///   <item>A query that does not already name a node type is ANDed with
+///     <c>nodeType:Code</c>, so non-code children never leak in. One that names its
+///     own (<c>nodeType:Scope</c>, say) is left as authored — the filter is a
+///     DEFAULT, not an override.</item>
 /// </list>
 /// </summary>
 public static class CodeQueryResolver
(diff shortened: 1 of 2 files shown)


rbuergi:
Right, and thank you — I had copied that claim rather than checked it. `CodeQueryResolver` line 145 appends `nodeType:Code` **only when the query does not already name a node type**, so "ANDs it into every source query it emits" is an overstatement.

The marker is still correct here, and the test *proves* that rather than asserting it: the starved type declares no `Sources`, so its discovery runs the **default** query (`namespace:{path}/Source scope:subtree`) and does carry the clause. The note now says exactly that, and names the case it would **not** cover — a declared source query naming some other node type.

I also corrected `CodeQueryResolver`'s own summary, which is where the overstatement came from and where the next person would copy it from, as I did.

Comments only; `SourceDiscoveryUnavailableTest` still green in 4s.

PR #1430, opened 2026-08-13
fix(#1422): a compiled source no longer prunes its own release history

### LINKED ISSUE #1422: Release-snapshot cleanup fails: MeshNode.Delete cannot route Release paths to a writable storage provider
**What is failing**

The Release-snapshot cleanup pipeline (likely a retention/pruning job) tries to delete old `/Release/...` nodes from DataMesh areas, but `MeshNode.Delete` throws `InvalidOperationException` because no writable storage provider claims those paths. All 4 deletion attempts in the same batch failed at the same moment.

**Probable cause**

The paths follow the pattern `Doc/DataMesh/{Category}/{Type}/Release/{timestamp}-{id}`. The `Release/` segment appears to denote a snapshot or compiled-output node — likely read-only or stored under a different storage adapter than normal editable nodes. The `PersistenceService` iterates writable storage adapters to find one that owns the path, and none match. This could be because:

- The Release storage provider is not registered in the writable-adapter list (it may be a separate read-only or archive provider).
- The Release nodes are stored with a different storage key/hierarchy than the delete operation expects.
- The cleanup job is incorrectly targeting the wrong base path, stripping the `Release/` segment and hitting nonexistent nodes — though the samples show the full `/Release/` path, so this is less likely.

**Confidence: Medium-High.** The error message is explicit ("no writable storage provider has this node") and the path pattern is consistent. The root fix is in the storage provider resolution or the cleanup job's path construction.

**Impact**

Low — 4 failures in a single batch on one pod, no recurrence. Likely a release-cleanup job that ran once and failed to expire old snapshots. No user-facing impact detected, but retention of Release snapshots may not be working as intended.

**Where to look**

- `MeshWeaver.Mesh.MeshNode.[DeleteNode]` — the method that logs the error and throws the exception.
- `PersistenceService.Delete` / the writable storage adapter resolution — where the list of writable providers is iterated and the "no writable storage provider" check occurs.
- The Release-snapshot cleanup job/trigger — something issued these 4 deletes; find what calls `DeleteNode` on Release paths.
- Storage provider registration for the `Release` segment under DataMesh areas.

---

### Evidence

| | |
|---|---|
| Fingerprint | `9515a1b73193ce77` |
| Category | `MeshWeaver.Mesh.MeshNode` |
| Severity | Error |
| Namespace | `memex-cloud` |
| Pods | `memex-bake-1786623419-kwv2c` |
| Occurrences | 4 |
| First seen | 2026-08-13 12:17:13Z |
| Last seen | 2026-08-13 12:17:13Z |

```
2026-08-13 12:17:13Z memex-bake-1786623419-kwv2c fail: MeshWeaver.Mesh.MeshNode[0]
      [DeleteNode] unexpected path=Doc/DataMesh/SocialMedia/Post/Release/20260813115710-4VyYozA9 partial-deleted=0
      System.InvalidOperationException: Cannot delete 'Doc/DataMesh/SocialMedia/Post/Release/20260813115710-4VyYozA9': no writable storage provider has this node.
2026-08-13 12:17:13Z memex-bake-1786623419-kwv2c fail: MeshWeaver.Mesh.MeshNode[0]
      [DeleteNode] unexpected path=Doc/DataMesh/PythonPandasNode/PandasExplorer/Release/20260813115710-F_006AEx partial-deleted=0
      System.InvalidOperationException: Cannot delete 'Doc/DataMesh/PythonPandasNode/PandasExplorer/Release/20260813115710-F_006AEx': no writable storage provider has this node.
2026-08-13 12:17:13Z memex-bake-1786623419-kwv2c fail: MeshWeaver.Mesh.MeshNode[0]
      [DeleteNode] unexpected path=Doc/DataMesh/SocialMedia/Profile/Release/20260813115711-UtXMgXv0 partial-deleted=0
      System.InvalidOperationException: Cannot delete 'Doc/DataMesh/SocialMedia/Profile/Release/20260813115711-UtXMgXv0': no writable storage provider has this node.
2026-08-13 12:17:13Z memex-bake-1786623419-kwv2c fail: MeshWeaver.Mesh.MeshNode[0]
      [DeleteNode] unexpected path=Doc/Architecture/BusinessRules/Cession/Release/20260813115710-c2ebmEFR partial-deleted=0
      System.InvalidOperationException: Cannot delete 'Doc/Architecture/BusinessRules/Cession/Release/20260813115710-c2ebmEFR': no writable storage provider has this node.
```

</details>

<sub>Opened automatically from `Admin/_LogIncident/9515a1b73193ce77`. Recurrences are folded into this issue rather than opening new ones.</sub>

rbuergi:
Fixes #1422.

## The exact failing path

`StaticRepoImporter.ComputePrunableNodes` — the ONE prune decision the static-repo import runs after upserting a source's nodes.

The prune rule is *"absent from the source ⇒ mirror it away"*. A compile-minted release record at `{nodeTypePath}/Release/{version}` is absent from **every** source **by construction**: it does not exist until the mesh compiles the type, so no repo and no embedded assembly can ever carry it. None of the existing guards covered that:

- `IsGovernance` spares `_`-prefixed segments only — `Release` is not one.
- #1326 put the protection behind `IStaticRepoSource.IsExcludedFromMirror`, but that hook **defaults to `false`** and only GitSync's `InMemoryStaticRepoSource` implements it.

So every **compiled** source kept deleting its own release history. `DocumentationStaticRepoSource` overrides neither the hook nor `SyncMode`, so the `Doc` partition ran the default `FullReplace` and pruned its releases on each import — the four `Doc/…/Release/…` deletes in the report were minted 20 minutes earlier by the same run.

**Corroborated against the live mesh**, not inferred: source-owned partitions hold exactly **one** release per NodeType (`Doc/DataMesh/SocialMedia/Post/Release/*` → 1 result), while an ordinary partition accumulates dozens (`AgenticOffice/*/Release/*` → 40+ over three days). Only something that prunes explains that difference, and only source-owned partitions have one.

### About the reported exception

`InvalidOperationException: no writable storage provider has this node` is the *symptom*, not the defect. The prune fires its deletes in parallel (`.Merge(BatchSize)`), so a path can pass `DeleteNode`'s not-found gate (`PersistenceService.Read`, which fans out over **all** providers) and then fail at the commit (`PersistenceService.Delete`, which fans out over **writable** providers only) — hence `partial-deleted=0` on all four at the same second. It stops once the prune no longer targets these nodes.

Two things I deliberately did **not** change, and why:
- **`Release` is not missing from the satellite mapping.** `PartitionDefinition.ResolveTable(path)` (read) and `ResolveTable(path, nodeType)` (write) both resolve a `Release` path to `mesh_nodes` — no `_`-segment match, no `NodeTypeTableMappings` entry, `SegmentForNodeType("Release")` is null. Read and write are symmetric; a release is an ordinary `mesh_nodes` row, correctly.
- **The `PersistenceService` read/delete provider-set asymmetry stands.** Widening `Delete` to fan out over read-only providers would let `EmbeddedResourceStorageAdapter.Delete` tombstone shipped documentation — a cleanup that starts working and takes too much. It is a real diagnostic wart (an unexpected `InvalidOperationException` where a clear refusal belongs) but it is a separate change with its own blast radius.

## What bounds the fix

**It can only ever delete less, never more.** The change adds one conjunct to the prune predicate:

```csharp
&& !IsMeshMintedRelease(t)
```

There is no path by which a delete that did not happen before starts happening. The segment match is anchored after the partition root and case-insensitive (the #1326 `release/` vs `Release/` lesson), so a node merely *named* like a release is untouched — `APartitionOrNodeMerelyNamedLikeARelease_IsStillPruned` pins that `AI/ReleaseNotes` and `AI/Releases/page` are still pruned.

Release records already accumulate unbounded in every non-source-owned partition, so this makes `Doc` match existing platform behaviour rather than introducing new growth. A real retention policy is a separate feature.

## Verification

Fail-before/pass-after, no mocking — `ComputePrunableNodes` is pure and public, so the decision itself is driven directly.

`MeshMintedReleaseRecord_IsNeverPruned_EvenWhenTheSourceHasNoMirrorExclusions` passes `isExcludedFromMirror: null` (the compiled-source case) across all three sync modes. **Without the guard** it fails with `found 2 item(s)` under `FullReplace` — exactly the production deletion — while every other test in the file still passes.

- `MeshWeaver.Graph.Test` — 15/15 `StaticRepoImporter*`
- `MeshWeaver.GitSync.Test` — 22/22 including `ForcePrunePreservesIgnoredNodesTest` (the #1326 regression guard)
- `MeshWeaver.Documentation.Test` — What's New + link integrity green
- `dotnet build -c Release -warnaserror` clean on `MeshWeaver.Graph.Test`, `MeshWeaver.GitSync.Test`, `MeshWeaver.Documentation.Test`

### DIFF
--- a/src/MeshWeaver.Documentation/Data/WhatsNew/2026-08-13-release-history-stops-being-erased-on-every-import.md
+++ b/src/MeshWeaver.Documentation/Data/WhatsNew/2026-08-13-release-history-stops-being-erased-on-every-import.md
@@ -1,0 +1,35 @@
+---
+Name: Release history stops being erased on every import
+Category: Fix
+Description: Built-in areas of the platform kept only their newest release record — every earlier one was deleted the next time the content was refreshed. The whole history is kept now.
+Icon: Sparkle
+Order: -20260813
+---
+
+# Release history stops being erased on every import
+
+Every time a live-compiled area of the platform builds successfully, it writes down a small record
+of that build: when it happened, what went into it, and the notes its author wrote. Those records
+are the release history you see on the area's page.
+
+For the parts of the platform that ship with the product — the documentation, and the built-in
+catalogs — that history was being wiped. Only the newest record ever survived; the moment the
+shipped content was refreshed, everything older vanished. Areas outside those built-in sections kept
+their full history, so the difference was easy to miss unless you happened to look at both.
+
+The cause was a disagreement about what "missing" means. Refreshing shipped content works by
+comparing what is in the product against what is in the platform, and removing anything the product
+no longer carries — that is what keeps a deleted page from lingering forever. A release record can
+never be in the product, though: it does not exist until the platform compiles the area, minutes or
+days after the product was built. So every record looked like something that had been deleted, and
+was duly removed.
+
+A related version of this was fixed earlier for content synchronised from a repository, but the fix
+had to be opted into and the built-in content never did — so it kept deleting its own history, and
+the failed cleanup attempts showed up as errors nobody could place.
+
+The platform now treats a release record as something *it* produced rather than something the
+product supplied, so its absence from the product is no longer read as a deletion. This can only
+result in fewer things being removed, never more: ordinary content that genuinely disappeared from
+the product is still cleaned up exactly as before, and a page whose name merely resembles a release
+is untouched.
--- a/src/MeshWeaver.Graph/Configuration/ReleaseNodeType.cs
+++ b/src/MeshWeaver.Graph/Configuration/ReleaseNodeType.cs
@@ -27,6 +27,17 @@ public static class ReleaseNodeType
     public const string NodeType = "Release";
 
     /// <summary>
+    /// The path SEGMENT release records are minted under — <c>{nodeTypePath}/Release/{version}</c>.
+    /// One declaration shared by the writer (<c>MeshDataSourceExtensions.TryCreateReleaseNode</c>)
+    /// and by every reader that has to recognise a mesh-minted release from its path alone: the
+    /// static-repo import's prune guard (a release is absent from EVERY source by construction, so
+    /// its absence is never evidence of a deletion — issues #1326 / #1422) and the sync-ignore
+    /// default. It happens to equal <see cref="NodeType"/>; it is declared separately because they
+    /// are different facts, and code that means "the path segment" must not read as a type name.
+    /// </summary>
+    public const string ReleaseSegment = "Release";
+
+    /// <summary>
     /// Registers the Release node type on the mesh builder by adding its MeshNode definition.
     /// </summary>
     /// <typeparam name="TBuilder">The mesh builder type.</typeparam>
--- a/src/MeshWeaver.Graph/MeshDataSource.cs
+++ b/src/MeshWeaver.Graph/MeshDataSource.cs
@@ -722,7 +722,7 @@ public static class MeshDataSourceExtensions
                 .Replace('+', '-').Replace('/', '_').TrimEnd('=')[..8];
             var version = $"{DateTime.UtcNow:yyyyMMddHHmmss}-{hash}";
 
-            var releaseNamespace = $"{nodeTypePath}/Release";
+            var releaseNamespace = $"{nodeTypePath}/{ReleaseNodeType.ReleaseSegment}";
             var releasePath = $"{releaseNamespace}/{version}";
 
             // Partition the compiler's combined {path → version} snapshot into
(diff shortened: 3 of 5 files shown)


rbuergi:
Filed the scoped-out finding as #1433 — the `PersistenceService` read/delete provider-set asymmetry that turns a benign concurrent-delete race into an `[DeleteNode] unexpected` ERROR. Deliberately not fixed here: the obvious repair (fan `Delete` out over read-only providers too) would let the embedded-resource adapter tombstone shipped documentation, so it needs its own decision rather than a drive-by.

PR #1436, opened 2026-08-13
Regression from #1343: an install publishes TWICE, and I recycled on both

rbuergi:
**I shipped this.** #1343's watcher recycles an instance whenever its type publishes a new assembly — and an **install publishes more than once**: the type node lands and compiles, then its `Source/` lands and it compiles *again*, seconds apart, while its instances are being activated and **read**. Each publication put a `DisposeRequest` through those hubs mid-read.

## The measurement

Same machine, same commit, one predicate different:

| | result |
|---|---|
| the watcher as merged in #1343 | 2 pass / **6 fail** |
| watcher disabled (control) | **8 pass** / 0 fail |
| **with this fix** | **8 pass** / 0 fail |

The test is `NodeRepoInstanceOrderingTest.TypedInstances_UnderAndBesideTheirInPackageType_InstallOnFreshMesh`, and the CI failure names the mechanism outright:

```
Hub Pack/Widget/Nested is shutting down — cannot create '/MeshNode'.
The address may reactivate (recycle / restart); retry to get the authoritative answer.
```

`main`'s CI passing twice with #1343 in it was luck, not evidence — I had taken it as evidence, which is how this got a second chance to bite.

## The fix

Wait for the type to go **quiet** (`AssemblySettleWindow`, 10 s) before acting, so a publication burst collapses into **one** recycle, after the burst. That is also what the watcher meant all along — *"a new build superseded mine"*, not *"the type is still settling"*. A real deploy publishes once and goes quiet, so the case #1343 exists for still heals, one window later.

Pinned by `APublicationBurst_CostsExactlyOneRecycle_AndOnlyAfterItGoesQuiet`, which walks the install shape (five builds a second apart) and asserts **nothing** is disposed until quiet — it fails without the throttle. The existing five cases keep their contracts, now advanced through the window on a `TestScheduler`.

## How it was tested

- The install test: **8/8** green (from 2/8)
- `MeshWeaver.Graph.Test` **1105/1105** green
- Clean under `-c Release -p:CIRun=true -warnaserror`

### DIFF
--- a/src/MeshWeaver.Graph/Configuration/NodeTypeEnrichmentHelpers.cs
+++ b/src/MeshWeaver.Graph/Configuration/NodeTypeEnrichmentHelpers.cs
@@ -1567,12 +1567,17 @@ internal static class NodeTypeEnrichmentHelpers
         IMessageHub instanceHub,
         string nodeType,
         string boundAssemblyPath,
-        ILogger? logger)
+        ILogger? logger,
+        IScheduler? scheduler = null)
         => typeStream
             .Where(t => t?.Content is NodeTypeDefinition d
                 && NodeTypeCompilationHelpers.HasUsableBuild(t, d)
                 && !string.IsNullOrEmpty(d.LatestAssemblyPath)
                 && !string.Equals(d.LatestAssemblyPath, boundAssemblyPath, StringComparison.Ordinal))
+            // 🚨 Wait for the type to stop publishing — see AssemblySettleWindow. An install
+            // publishes twice (type, then its Source/) while instances are being read, and
+            // recycling on each one disposes hubs mid-read.
+            .Throttle(AssemblySettleWindow, scheduler ?? Scheduler.Default)
             .Take(1)
             .Subscribe(
                 t =>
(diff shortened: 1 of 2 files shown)


rbuergi:
Keeping this one — it does compile.

C# 7.2 introduced **non-trailing named arguments**: a named argument may be followed by positional ones as long as it sits in its own correct position. `logger` is parameter 5 and is passed 5th; `scheduler` then follows positionally as 6th, so the call is legal.

The decisive evidence is that it built and ran: `StaleAssemblySelfHealWatcherTest` **6/6**, `MeshWeaver.Graph.Test` **1105/1105**, and the project is clean under `-c Release -p:CIRun=true -warnaserror`. A call that did not compile could not have produced those numbers.

Not renaming it purely for style right now: this PR fixes a regression that is **currently live in `main`** (every PR branched off it inherits the failure — #1427's shard 4 is failing for exactly this reason), so I would rather land it than spend another CI cycle on a cosmetic change. Happy to follow up with the naming if you would prefer it explicit.

PR #1437, opened 2026-08-13
fix(build): the build claim is exclusive across clusters, not just within one

### LINKED ISSUE #1424: Build claim is exclusive per Orleans cluster, not across clusters — bake Job and serving pod both baked
Found by the bake-mode pilot on memex-cloud (2026-08-13 ~12:17 UTC, image `ci.3446`/`9228a6d`): the ephemeral bake Job (#1416, own `-bake` ServiceId + localhost clustering) and the rolling serving pod EACH claimed `Admin/Build` and EACH ran the full 268-type bake (~2 min each) for fingerprint `8a1d5716…`.

## Root cause

The claim arbiter's exclusivity comes from the owning per-node hub serialising `Update` lambdas — which holds **per cluster**. The Job's cluster and the live cluster each activate their **own** hub for `Admin/Build` against the same durable row; each ran its own arbiter against its own mirror, each granted its own candidate, and the durable row took both write streams (last-write-wins on the shared fields). Registrations (`RequestedClaims`) are merge-safe by design (per-candidate keys); the **grant decision** is what raced.

## Impact

Bounded and non-corrupting, by construction: every downstream write is idempotent/content-addressed (assembly store keyed by content, release versions minted from content hashes so replicas agree, GO keyed by fingerprint — both publishers wrote the same record). The cost is one redundant full bake per rollout — including the ~2 GB bake allocation in the serving pod that the Job exists to spare it.

## Fix direction

The claim needs ONE arbitration point across clusters. Candidates (ordered by preference):
1. A **conditional patch** primitive at the storage layer — apply the claim patch only if `ClaimedBy` is currently null/stale (compare-and-set semantics on the durable row), used ONLY by the claim path. This keeps arbitration in the protocol and makes it cluster-count-independent — the property the parallel-builders phase needs anyway.
2. Alternatively, verify-after-grant against the durable row (re-read authoritative, revoke on mismatch) — narrower but still has a race window without CAS.

The 2026-08-13 handoff predicted exactly this: "the content-addressed store on disk does most of the coordinating and the protocol only has to arbitrate claims" — this issue IS that arbitration, now with a concrete prod repro.

Until fixed: running the bake Job is still net-positive (the GO lands earlier; nothing corrupts) but does not yet spare the serving pod its bake — treat the Job as optional.

rbuergi:
Closes #1424.

## What was wrong

The claim arbiter's exclusivity came from the owning per-node hub serialising its `Update` lambdas — which holds **per cluster**. A second Orleans cluster over the same database activates its *own* hub for `Admin/Build`, arbitrates against its *own* mirror, and grants its own candidate. Three things then made the collision invisible:

- both grants were minted at the same next `MeshNode.Version`, and the store's monotonic condition **applies at equal versions** (re-persisting an unchanged node is a legitimate shape), so both landed last-write-wins;
- neither writer was told it lost — the refusal signal is `saved.Version > written.Version`, and the versions were *equal*;
- nothing propagates a rival's write. Orleans membership and its memory streams are per-cluster by construction, and `PostgreSqlChangeListener` is registered but never started in either partitioned-PG wiring.

On top of that, the grant is observed off the RAM mirror, which commits **~200 ms before** the persistence sampler reaches storage — so adjudicating after the write lands can never stop the second builder starting.

## The fix

**1. `IStorageAdapter.WriteIfVersion` — an atomic compare-and-set.** The write-side twin of `DeleteIfExists`: applies only while the durable row still carries the version the caller read (`0` = "must not exist"), so at most one of N concurrent callers is told `true`. Postgres implements it as one statement whose row count is the verdict (`ON CONFLICT … DO UPDATE … WHERE target.version = @expected`, or `DO NOTHING`); in-memory via `TryAdd`/`TryUpdate`. Every decorator forwards explicitly — the interface default is a non-atomic read-compare-write, correct only for single-writer backends, exactly the contract `DeleteIfExists` already documents.

**2. The claim moved onto a lock node, `Admin/Build/_Claim`.** Making the *grant* exclusive turned out to be necessary and not sufficient, and the test caught it: every cluster mirrors `Admin/Build` and flushes it **wholesale** through the persistence sampler, and `MeshNode.Version` is a per-node counter rather than a cross-cluster logical clock — so a *losing* cluster's flush overwrote the winner's `ClaimedBy` with its own `null`, and the next arbitration pass found a free build. The lock has no hub, no mirror and no sampler; it is written only by the compare-and-set grant, a compare-and-set heartbeat, and a rowcount-gated `DeleteIfExists` on release. `BuildState.ClaimedBy` stays as the observable projection `ObserveBuildClaim` reads — a clobber there now costs a stale view, never a second builder.

A refused claim **writes nothing and retries nothing**: the arbiter is level-triggered and re-decides on the next real change. The lock's own change feed is now one of its triggers, which is what hands a released claim on without waiting for the slow tick.

**Fails OPEN.** With no adapter owning the path (or no persistence at all — monolith, test, dev box) the grant is taken on the mirror exactly as before. Being wrong that way costs one duplicated bake — bounded and non-corrupting, because every downstream write is content-addressed. Being wrong the other way leaves the fingerprint with no GO at all, which holds every silo's readiness probe down and stalls the rollout. Same asymmetry the bake gate applies; the opposite of `AssemblyCacheRetention`, where the wrong answer deletes bytes a running pod needs.

## Also, since we were in this code

`NodeTypeBatchBake.WriteStamp` now uses the same primitive. Its racing writer is the type's **own per-node hub** — not a baker, so the claim does not and cannot exclude it, and the existing comment was right that election cannot close this. But the comment's conclusion ("the guard above stays the thing that makes it safe") was wrong at *equal* versions: both sides read `v` and minted `v+1`, the store applied both, one stamp was silently destroyed, and the "REFUSED" warning never fired because it tests for a strictly greater version. Conditioning on the version we read makes the loser deterministic and loud. Its outcome is unchanged and still correct (bytes durable and content-addressed, record pending, level-triggered probe re-bakes). Deliberately **no** compare-and-re-apply loop — that would be a second write path into a record this driver does not own.

## Verified

- **`CrossClusterBuildClaimTest`** — two independent meshes (own hubs, own mirrors, own adapter instances, so no change feed crosses) over **one shared backing store**, both claiming concurrently. **Fails before with `found 2`; passes after with exactly 1**, and the durable lock names the winner. Run with `DOTNET_PROCESSOR_COUNT=4`.
- `BuildCoordinationTest` (incl. the full protocol-driven bake), prewarm/bake/write-integrity filter: **80/80**.
- `MeshWeaver.Hosting.Test` **163/163**, `MeshWeaver.Graph.Test` **1104/1104**, `MeshWeaver.Documentation.Test` **20/20**.
- `-c Release -warnaserror` clean on every touched project plus `Memex.Portal.Distributed`.

## Follow-up found while here (not fixed)

`BuildCoordination.md` claimed the GO broadcast rides the PG `LISTEN/NOTIFY` feed "cross-process and therefore cross-ServiceId by construction". That listener is **never started** in either partitioned-PG overload, so across clusters nothing propagates the GO, and `BuildProtocolDriver.FollowGo` subscribes with no bound. The doc now says so. This does not affect the fix (the arbiter reads the lock rather than waiting to be told), but it means a *separate-cluster* bake host should still be treated as "publishes the GO earlier", not "spares its peers the bake". Starting the listener, or giving the follower a bounded fall-back to its own sweep, is the outstanding work.

### DIFF
--- a/src/MeshWeaver.Documentation/Data/Architecture/BuildCoordination.md
+++ b/src/MeshWeaver.Documentation/Data/Architecture/BuildCoordination.md
@@ -33,7 +33,9 @@ The protocol below removes all three **by construction** rather than by compensa
 
 ```
 Admin/Build                       ← the build ROOT (durable node, nodeType Build)
+Admin/Build/_Claim                ← the root's claim LOCK (durable; only atomic writes)
 Admin/Build/{chunkName}           ← one node per CHUNK (durable, nodeType Build)
+Admin/Build/{chunkName}/_Claim    ← the chunk's claim LOCK
 Admin/Build/{chunkName}/_Activity ← the chunk's execution activity (standard Activity)
 {nodeTypePath}/Release/{version}  ← unchanged: releases minted per compiled NodeType
 ```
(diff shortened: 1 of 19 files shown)


rbuergi:
Two refinements landed after the opening description, both found by reviewing the primitive's edges:

**`60f18c1` — a compare-and-set expecting a version must not resurrect a deleted row.** Postgres implemented `expectedVersion > 0` as `INSERT … ON CONFLICT DO UPDATE … WHERE target.version = @expected`, which reports **success when the row is absent** — the INSERT simply lands. "The row still carries exactly v" is false when there is no row, and for the build claim the difference is a live defect: a heartbeat racing a release re-creates the lock its holder just dropped and blocks the next candidate for the whole staleness budget. That case is now a plain `UPDATE` with the version predicate, so the row count is the verdict. The in-memory adapter already refused it; the two backends now agree.

`CompareAndSetWriteTests` pins the contract **through the production decorator chain** — 8 concurrent creates yield exactly one winner and seven *told* losers, likewise 8 concurrent successions of a held row, a mismatched version is refused and leaves the row, and a delete is not undone by a later compare-and-set. The chain is the point: the interface default is a non-atomic read-compare-write, so a decorator that forgets to forward silently downgrades every caller to the behaviour the primitive replaces and nothing else would notice.

**`f55dbdd` — skip the lock read when this cluster has nothing to grant.** The pending-candidates check ran *after* the pass had already read the lock, so the 2-minute tick cost a storage round-trip on every quiet Build node for a decision that was always "nothing to do".

Verification on the final tree: full `MeshWeaver.Hosting.Monolith.Test` locally **661 passed / 3 skipped / 0 failed**, `MeshWeaver.Hosting.Test` 167/167, and CI green across all six shards plus the cross-repo plugin gate.

Issue #612, opened 2026-07-23
CI wedge: CodeEditRecompileTest StaleBarrierType first compile ends Error, re-trigger not picked up (sub-case b) — shard timeout exit=124


rbuergi:
Occurred on PR #611 CI run 30004790036 (shard 1), on a branch whose diff is causally excluded from the failure (SelfUpdateHostedService is registered by no test mesh, and the new test never ran in the killed process per the shard phase trace).

Evidence from the shard-1 `_meshweaver-test-trace.log`:
- Monolith.Test started 12:07:34; `CodeEditRecompileTest` began wedging at its second CTOR (12:08:04 → 12:09:19, first 75s watchdog kill); every subsequent class CTOR accounted for until the 12:15:34 wall-clock kill (exit=124, no trx).
- The test's own triage diagnostics: `WaitForLatestRelease stuck for TestData/StaleBarrierType: known=(null), MIRROR: Status=Error, Latest=(null), INDEX Release children: [(none)], RE-TRIGGER: STILL STUCK after re-trigger (sub-case b: persistent clobber / dead subscription)`.
- So the FIRST compile of StaleBarrierType ended `CompilationStatus.Error` under CI load, and the forced re-trigger (`RequestedReleaseAt` + `Force`) produced nothing for 25s — release watcher dead/clobbered. That wedged the whole class (6 tests × 50–75s) plus `ReleaseRequestWatcherHighWaterTest`, blowing the 6-min cap.

This is the compile-wedge family previously addressed in #580/#583/#584 — this occurrence shows a residual order/load-sensitive race (sub-case b) that will intermittently red any branch, including main. Needs its own pinned repro + root-cause fix; starting point: `test/MeshWeaver.Hosting.Monolith.Test/CodeEditRecompileTest.cs` (`WaitForLatestRelease`, ~lines 880–920), first-compile-Error path and the release-watcher subscription lifecycle.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

rbuergi:
Fresh occurrence with a new data point — PR #884 run 31153582625 shard 2 (2026-08-07 06:37-06:40Z), branch causally excluded (diff = ClrMD leak-test files only, and those tests passed on the shard):

```
WaitForLatestRelease stuck for TestData/StaleBarrierType: known=(null)
  MIRROR: Status=Error, Latest=(null), ReqAt=, Handled=, Force=False
[...]

PR #1438, opened 2026-08-13
fix(#890): own the Roslyn reference set per mesh — HELD pending the canary verdict

rbuergi:
# 🚨 HELD — do not merge until the canary reports

**Merge only when #890's emit canary returns `canary=REFERENCES`. If it returns `canary=BELOW-ROSLYN`, close this unmerged** — no reference-set change can fix a CLR heap or JIT fault, and closing saves the memory cost entirely. At ~1.4% per run the verdict is roughly a day away.

Stacked on **#1445**, which carries the two halves of #890 that cost nothing and do not depend on this hypothesis: the **detector** (`NoStaticCollectionsTest` keying on the element type) and the **instrument** (the canary's second leg). This PR is the half that does depend on it.

## The change

`MeshNodeCompilationService._references` was a `private static readonly IReadOnlyList<MetadataReference>`, twice classified in review as a "write-once constant lookup". Only the **list** is write-once: a `PortableExecutableReference` owns lazily memory-mapped, `IDisposable` metadata and Roslyn caches derived assembly/symbol tables against that instance.

It becomes `CompilationReferenceSet` — a mesh-scoped singleton, **lazily** materialized so a mesh that never compiles never builds one. Deliberately **not** disposed at teardown: a compile can still be in flight when a mesh dies, and disposing metadata underneath one is a use-after-free on exactly the object graph this is chasing. Removing the field also empties the `CACHE` debt #1445 introduces.

Two tests pin what the change delivers: two sets share no `MetadataReference` instance, and **neither the set nor its references** survive a full collect. The second arm is the load-bearing one — a process-wide memo keyed by assembly path would leave the wrapper collectible while sharing the references again, so a check on the wrapper alone would pass while the very state #890 is about survived.

## Why it is held, not merged

Roslyn 5.6's own source (`dotnet/roslyn@c0573ed`) shows the NRE's guard admits a type only when `ContainingModule == moduleBeingBuilt.SourceModule` — so the symbol whose `ContainingType` reads null is a **source symbol of the compilation being emitted**, never a PE symbol arriving from a reference. The reference set is a **vector at best, not a demonstrated cause**.

Against that uncertain benefit, the cost is certain and measured — A/B, same base, same machine, same tests passing both sides:

| slice | baseline | this PR | delta |
|---|---|---|---|
| `CodeEditRecompileTest` (11) | 662 MiB | 1012 MiB | +53% peak RSS |
| `~Compile\|~NodeType` (133) | 1071 MiB | 1701 MiB | +59% peak RSS |
| **full project (667)** | **1157 MiB / 460.8 s** | **2292 MiB / 449.8 s** | **+98% peak RSS, wall clock unchanged** |

≈15 MiB and ~73 ms per set; a **high-water mark, not a leak** (repeated rounds plateau: 937 → 1042 → 1058 → 1083 → 1078 MiB). Still under the harness's 4 GiB `MEM_PRESSURE` / 6 GiB `MEM_CRITICAL_FAILFAST`, and CI splits the project in two. **No production cost** — a portal hosts one mesh, so one set per process exactly as today. macOS numbers; on Linux the metadata is kernel-reclaimable `RssFile`, so the CI magnitude is untested.

Paying a doubled memory ceiling *now* to pre-empt an answer that is one observation away is the wrong trade. The canary was built precisely so we would not have to reason for a week — its first leg killed the entire generated-source half of the search space in 95 minutes.

## One test failure still on the ledger

The with-change full run had `SilentReadNackTest.OwnerDisposedWithReadOutstanding_IsNacked_NotLeftHanging` fail (expected `DeliveryFailureException`, got `GetDataResponse`); the baseline run passed it, it passes in isolation, and it passed on Linux CI with the change. Its test was churned on main the same day (`46be4a096`/`82d62519c`/`3e49feb85`, one literally *"a read whose owner is recycled RECOVERS"*), and this diff compiles no NodeType. **But** one run each is not evidence of a pre-existing flake, and +1.1 GiB of peak RSS genuinely perturbs GC timing. Not claiming unrelated — claiming unproven.

## What's New

Skipped — pure-internal (architecture rule compliance). No user-visible effect: a portal hosts one mesh, so this is a no-op in production.

### DIFF
--- a/src/MeshWeaver.Documentation/Data/Architecture/NoStaticState.md
+++ b/src/MeshWeaver.Documentation/Data/Architecture/NoStaticState.md
@@ -138,11 +138,12 @@ The build guard classifies permitted static fields into four buckets. Everything
 > `NoStaticCollectionsTest`, whatever the declared collection type, because the declared-type check
 > could not see an `IReadOnlyList<>`.
 >
-> The field itself is still there, listed as a `CACHE` debt with its issue number rather than
-> silently exempt — removing it gives every mesh its own reference set, which measurably costs ~15 MiB
-> per compiling mesh, and whether that buys anything is decided by #890's emit canary. **Landing the
-> detector does not require paying for the fix**, and an allowlist entry with a reason is what turns
-> "nobody noticed" into "we know, and here is the open question".
+> The field is now the mesh-scoped `CompilationReferenceSet`, so the reference set dies with the mesh
+> that built it. That cost ~15 MiB of resident PE metadata per compiling mesh (+98% peak RSS across
+> the full `Hosting.Monolith.Test` project) — a real price, paid because a process-wide Roslyn cache
+> is the shape this rule exists to forbid. It shipped **after** the detector, not with it: the guard
+> landed first carrying the field as a visible `CACHE` debt, which is the cheap half, and the memory
+> was paid only once the evidence justified it.
 
 ### MEMO — Pure memoization on process-global keys
 
--- a/src/MeshWeaver.Graph/Configuration/CompilationReferenceSet.cs
+++ b/src/MeshWeaver.Graph/Configuration/CompilationReferenceSet.cs
@@ -1,0 +1,118 @@
+using System.Collections.Immutable;
+using Microsoft.CodeAnalysis;
+
+namespace MeshWeaver.Graph.Configuration;
+
+/// <summary>
+/// The default Roslyn <see cref="MetadataReference"/> set every dynamic-NodeType compilation
+/// binds against — the trusted-platform-assembly list plus a few well-known additions — owned by
+/// the <b>mesh</b>, not by the process.
+///
+/// <para>🚨 Why this is a mesh-scoped instance and not a <c>static readonly</c> list. It used to
+/// be <c>MeshNodeCompilationService._references</c>, a <c>private static readonly
+/// IReadOnlyList&lt;MetadataReference&gt;</c>, and was twice written off in review as "a
+/// write-once constant lookup" — the sanctioned exemption in
+/// <see href="Doc/Architecture/NoStaticState.md">NoStaticState.md</see>. That reading is wrong,
+/// and the exemption does not apply: the <i>list</i> is write-once, but its <i>elements</i> are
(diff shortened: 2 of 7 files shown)


rbuergi:
## The canary verdict this PR was held for has arrived — and it is `BELOW-ROSLYN`

New #890 occurrence on run [31810904642](https://github.com/Systemorph/MeshWeaver/actions/runs/31810904642/job/94803361047) (shard 4, `CodePackageInstallTest`), with the #612 diagnostics attached:

```
canary=BELOW-ROSLYN
  shared:   THREW NullReferenceException
  pristine: THREW NullReferenceException
```

A **pristine** compilation — brand-new references, freshly parsed source, nothing shared with the poisoned one — failed identically, in `Microsoft.Cci.MetadataWriter.GetConsolidatedTypeParameters` via `NamedTypeSymbol.ITypeDefinitionMember.get_ContainingTypeDefinition()` returning null.

That is a direct answer to this PR's premise: **owning the reference set per mesh cannot fix a fault that reproduces with references that were never shared.** The broken state is below Roslyn (CLR heap / JIT / GC), and the symptom looks like #613's zeroed-word corruption surfacing as an NRE instead of a SIGSEGV.

Full reasoning and suggested next step (core dump + tiering disabled, not a reference-set change) posted on #890: https://github.com/Systemorph/MeshWeaver/issues/890#issuecomment-5294851497

Flagging rather than closing — that is the author's call, and the change may still be worth having on its own merits. It just should not be merged *as the fix for #890*.

rbuergi:
Closing unmerged per this PR's own gate: #890's emit canary has reported **`canary=BELOW-ROSLYN`** — on a fresh occurrence on `main` itself (run 31841598894, shard 1, merge commit 058595b, 2026-08-15), with the full signature in the [#890 comment](https://github.com/Systemorph/MeshWeaver/issues/890#issuecomment). That is the verdict this PR was held for: the fault sits below Roslyn (the CLR heap/JIT class also tracked as #613), so owning the reference set per mesh cannot fix it, and closing saves the per-mesh memory cost entirely. The zero-cost halves (detector + canary instrument) already shipped via #1445 and stay. #890 continues as a below-Roslyn investigation paired with #613.

PR #1487, opened 2026-08-14
fix(clients): RN search works, RN REST ops carry the token, move/copy send real field names

### LINKED ISSUE #1473: RN search never returns: Mesh.search posts a QueryRequest that has no server handler

`clients/grpc-web/src/mesh.ts:70` posts `QueryRequest` to `mesh/main` — no such record exists anywhere server-side (`git grep "record QueryRequest"` is empty; `MeshQueryRequest` is a service DTO, not a hub message). portal-next explicitly routed around it (`clients/portal-next/src/client/live.ts:391-397` points `ops.search` at REST `query-nodes`), but react-native still calls it (`clients/react-native/src/liveOps.ts:216`), so every RN search, `MeshSearch`/`MeshNodeCollection` control, `NodeExport`, and the ThreadChat agent/model selectors wait out the 30 s observe timeout (`clients/grpc-web/src/connection.ts:169`) and render empty.

Short-term fix: RN uses REST `/api/mesh/query-nodes` like portal-next. Proper fix: a `MeshQueryReference : WorkspaceReference` + producer so queries are live-subscribable over the existing `SubscribeRequest` channel, then delete `Mesh.search`s dead `QueryRequest`.

### LINKED ISSUE #1474: RN REST calls carry no Bearer token; LocalMesh sidecar does not map content/list + upload — FileBrowser broken in both RN modes

`buildMeshOps(connection, baseUrl, partition)` (`clients/react-native/src/liveOps.ts:206`) takes no token and `App.tsx:132` passes only the instance URL, so `renderMarkdown`, `listContent` and `uploadContent` fetch with no `Authorization` header. Against a remote portal, `/api/mesh/*` is Bearer-only (`memex/Memex.Portal.Shared/Api/MeshApiEndpoints.cs:58`) so these 401. Against the sidecar, `memex/Memex.LocalMesh/Program.cs` maps only `render-markdown` (:111), speech (:133) and `/static` (:87) — `content/list` and `upload` fall through `MapFallbackToFile("index.html")` (:205) and return the SPA HTML with a 200, so `JSON.parse` throws.

Net: FileBrowser is broken in both RN modes and markdown rendering silently degrades to the limited client-side parser. Only speech gets the token because it is plumbed separately (`App.tsx:150-155`).

### LINKED ISSUE #1475: TS SDKs send {source,target} for move/copy but the records are SourcePath/TargetPath

`clients/grpc-web/src/mesh.ts:155` posts `MoveNodeRequest { source, target }` and `:161` posts `CopyNodeRequest { source, target }`; the C# records are `MoveNodeRequest(string SourcePath, string TargetPath)` (`src/MeshWeaver.Mesh.Contract/CreateNodeRequest.cs:560`) and `CopyNodeRequest(string SourcePath, string TargetPath)` (`:483`) with no `JsonPropertyName` aliases — case-insensitive matching cannot map `source` to `SourcePath`. Same defect in the Node SDK (`clients/typescript/src/mesh.ts:96, :101`).

Currently latent (no JS shell calls them today) but a live landmine on a published SDK surface. Root fix beyond the rename: a machine-readable $type message catalogue checked in CI so hand-mirrored shapes cannot drift.

rbuergi:
Fixes #1473 (client half), fixes #1474 (client half), fixes #1475 — the JS-client defects from the modularization investigation.

**Search** — `Mesh.search` in both SDKs posted a `QueryRequest` no server handles; RN searches spun 30 s and rendered empty. Now REST `POST /api/mesh/query-nodes` with Bearer auth (portal-next's proven route); `basePath` folds into the mesh query syntax. The fake transport's `QueryRequest` responder is **deleted** — a fake handler for a message the real server lacks is exactly how this shipped. Protocol-native follow-up stays open in #1473 (live-subscribable `MeshQueryReference`).

**RN auth** — `buildMeshOps` threads the instance token into search / render-markdown / content-list / upload (previously all 401'd against real portals; the anonymous sidecar is unchanged — empty token adds no header). The sidecar's missing `content/list`+`upload` mappings remain open under #1474.

**Wire shape** — `move`/`copy` sent `{source,target}` against records declaring `SourcePath`/`TargetPath` (never bound). Fixed in both SDKs; pinned by a new wire-shape test asserting the field names and the absence of the old ones.

**Verification** (Clients CI recipe, locally): grpc-web typecheck + 28/28 vitest; `@meshweaver/client` typecheck; react-native typecheck + 153/153 tests.

What's New entry included (Category: Fix).

### DIFF
--- a/clients/grpc-web/src/mesh.test.ts
+++ b/clients/grpc-web/src/mesh.test.ts
@@ -5,11 +5,31 @@ import { fakeMeshTransport } from "./testTransport";
 // Drives the ergonomic Mesh ops surface against the same in-memory Connect+Deliver fake the connection
 // tests use — proving search/get/watch/patch/create/delete/move/copy compose correctly over the transport.
 describe("Mesh ops (gRPC-web, in-memory)", () => {
-  it("search returns the query results", async () => {
-    const mesh = await Mesh.connect("memory://", { transport: fakeMeshTransport() });
+  it("search posts REST query-nodes with Bearer auth (the gRPC QueryRequest has no server handler — #1473)", async () => {
+    const calls: { url: string; init: RequestInit }[] = [];
+    const fakeFetch = (async (url: unknown, init?: RequestInit) => {
+      calls.push({ url: String(url), init: init ?? {} });
+      return new Response(JSON.stringify({ results: [{ path: "ACME/Stories/1" }] }), { status: 200 });
+    }) as typeof globalThis.fetch;
+    const mesh = await Mesh.connect("https://portal.example", {
+      transport: fakeMeshTransport(),
+      token: "mw_test",
+      fetch: fakeFetch,
+    });
     const results = await mesh.search("nodeType:Story namespace:ACME");
     expect(results).toHaveLength(1);
     expect(results[0].path).toBe("ACME/Stories/1");
+    expect(calls[0].url).toBe("https://portal.example/api/mesh/query-nodes");
+    const headers = calls[0].init.headers as Record<string, string>;
+    expect(headers["authorization"]).toBe("Bearer mw_test");
+    expect(JSON.parse(String(calls[0].init.body))).toEqual({ query: "nodeType:Story namespace:ACME", limit: 50 });
+
+    // basePath folds into the mesh query syntax rather than a body field QueryNodesBody lacks.
+    await mesh.search("laptop", "ACME/Products", 10);
+    expect(JSON.parse(String(calls[1].init.body))).toEqual({
+      query: "laptop path:ACME/Products scope:subtree",
+      limit: 10,
+    });
     mesh.close();
   });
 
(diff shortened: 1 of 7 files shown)


rbuergi:
Heads-up on an overlap: a parallel session opened #1488 for the same cluster (#1473 / #1474 / #1475) about 25 minutes after this PR. They are not redundant, but they touch some of the same lines.

**What #1488 adds that this PR does not**

- The `{source,target}` class is not limited to move/copy. Sweeping the whole SDK surface against the C# records found three more of exactly the same shape, all still live after this PR:
  - the Node SDK's `patch` posts `{ path, change }` for `PatchDataRequest(WorkspaceReference Reference, RawJson Patch)` — **neither** member matches, so nothing is ever identified to patch;
  - the Node SDK's `watch` sends a reference with no `$type`, which the polymorphic converter cannot resolve to a `MeshNodeReference`;
  - `execute()` flips `requestedStatus` at the **node** level in both SDKs, but `MeshNode` has no such member — `RequestedStatus` lives on the `ActivityLog` **content**.
- A drift guard rather than a hand-fix: the field names move into one import-free `wire.ts`, mirrored byte-identically into both SDKs (the discipline `clients/react/src/i18n` uses for the server catalog), and `wireContract.test.ts` parses the C# contract sources and asserts every name is a declared member of the record it mirrors — a merge patch is checked against `MeshNode`, which is what catches the `execute()` shape. `clients.yml` gains those contract sources so a server-side rename reddens the client.

**Overlap to resolve.** Both PRs edit the `move`/`copy` lines in `clients/grpc-web/src/mesh.ts` and `clients/typescript/src/mesh.ts`, and both add cases to `clients/grpc-web/src/mesh.test.ts`. Whichever merges second needs a small manual resolution there. The two also differ on `Mesh.search`: this PR keeps it and gives it a REST reach-back; #1488 leaves it alone. Worth a deliberate call rather than a merge-order accident.

**Note for the sidecar.** Pointing `Mesh.search` at `POST /api/mesh/query-nodes` only works where that route exists. `Memex.LocalMesh` does not map it — an unmapped `/api/mesh/*` there falls through `MapFallbackToFile` and answers `index.html` with a **200**, so `JSON.parse` throws and RN's `.catch(() => [])` swallows it. Against the local sidecar (the RN app's default instance) search therefore still returns nothing after this PR. #1492 maps `query-nodes`, `content/list` and `upload` on the sidecar and adds a parity check over both backends' endpoint maps.

PR #1488, opened 2026-08-14
fix(clients): send the C# records' own field names from both TypeScript SDKs

### LINKED ISSUE #1475: TS SDKs send {source,target} for move/copy but the records are SourcePath/TargetPath

rbuergi:
Closes #1475.

## 🔗 Reads with #1487 and #1494 — do not undo one and keep the others

This PR is the third of three that together fix the JS-client cluster, and they are load-bearing on
each other:

- **#1487** (merged) points `Mesh.search` at `POST /api/mesh/query-nodes`. That is the right call:
  a mesh query is `IMeshQuery.Query`, a **service call with no hub message**, so REST is the surface
  that exists. This PR does not compete on `search` — #1487's reach-back stands untouched.
- **#1494** (merged) is what makes that sound *everywhere*. Before it, `Memex.LocalMesh` did not map
  `query-nodes`, and an unmapped `/api/mesh/*` route there does not 404 — `MapFallbackToFile`
  answers it with `index.html` and a **200**, so the caller fails inside `JSON.parse` and the shell
  swallows it. **Reverting #1494 silently re-breaks #1487** against the RN app's default instance.
- **This PR** fixes the remaining wire-shape defects and adds the guard that stops the whole class
  coming back.

## The defect

Both SDKs posted `MoveNodeRequest { source, target }` / `CopyNodeRequest { source, target }`, but the
records are `(string SourcePath, string TargetPath)` with no `JsonPropertyName` aliases.
Case-insensitive matching cannot map `source` → `SourcePath`, so System.Text.Json **drops** both
members, the record is constructed with empty paths, and the operation completes having done nothing.
No error, anywhere. (#1487 hand-fixed these two; this PR keeps them correct *and* guarded.)

## The sweep

Checking the rest of the SDK surface against the records found three more of the same class, none of
which #1487 covers:

| call | sent | record |
|---|---|---|
| Node SDK `patch` | `{ path, change }` | `PatchDataRequest(WorkspaceReference Reference, RawJson Patch)` — **neither** member matched |
| Node SDK `watch` | `{ reference: { path } }` | the reference is polymorphic; without `$type: "MeshNodeReference"` it does not resolve |
| `execute()` (both SDKs) | `{ requestedStatus }` at the NODE level | `MeshNode` has no `RequestedStatus` — it lives on the `ActivityLog` **content** |

Every other `patch` caller in `clients/` already nests under `content`; `execute()` was the odd one out.

## The fix, and why it cannot drift again

Field names live in ONE import-free module, `wire.ts`, **mirrored byte-identically** into both SDKs
(the discipline `clients/react/src/i18n` uses for the server string catalog). Three guards:

- **`wireContract.test.ts`** parses the C# contract sources and asserts every field name is a declared
  member of the record it mirrors. A merge patch is validated against `MeshNode` — that is what
  catches `execute()`'s shape.
- **Byte-identity** between the two copies, so a fix can never land in one SDK only — the other half
  of how #1475 shipped. It reports the first differing **line** with both sides, not two truncated
  strings.
- **No inline bypass**: every message type `wire.ts` owns must be named only there. The list is
  **derived from `wire.ts`**, so a new builder extends the guard automatically and a type it does not
  own (a REST-backed verb) stays correctly out of scope rather than needing an exception list.

`clients.yml` gains the five contract sources the test reads (guarded by `ciTrigger.test.ts`), so a
server-side rename turns the client red rather than turning a client feature into a no-op.

## Fail-before, all by mutation rather than assertion

| mutation | result |
|---|---|
| `wire.ts` reverted to `{source,target}` + node-level `requestedStatus` | 5 red: `MoveNodeRequest has no member(s) source, target — the hub would drop them SILENTLY` |
| a contract source **moved** | file fails at import — `ENOENT`, "no tests" (cannot pass vacuously) |
| a record **renamed** server-side | `no C# record 'MoveNodeRequest' in …` listing the files searched |
| parser degraded to **zero members** | 9 of 12 red: `parsed no members off record X — the parser broke` |
| the mirrored copy **diverged** by one line | names the file, the line number and both sides |
| a message built **inline** in `mesh.ts` | names the file and the type it bypassed |
| one `clients.yml` path dropped | `ciTrigger.test.ts` red |

## Verification

- `clients/grpc-web`: `tsc --noEmit`, `tsc -p tsconfig.build.json --noEmit`, `vitest run` — 52/52
- `clients/typescript`: `tsc --noEmit` clean
- `clients/react`: `ciTrigger.test.ts` 4/4 · `clients/react-native`: `tsc --noEmit`, 153/153
- `dotnet build test/MeshWeaver.Documentation.Test -c Release` → 0 Errors; `WhatsNewEntryIntegrityTest`
  + `DocumentationLinkIntegrityTest` pass with the new release note embedded


rbuergi:
**Overlaps with #1487**, which a parallel session opened ~25 minutes earlier for the same issue. Both fix the `move`/`copy` field names; they touch the same lines in `clients/grpc-web/src/mesh.ts`, `clients/typescript/src/mesh.ts` and `clients/grpc-web/src/mesh.test.ts`, so whichever merges second needs a small manual resolution. Flagging rather than closing either, because the two are not the same change:

**What this PR has that #1487 does not**

1. **Three more mismatches of the same class**, found by sweeping the whole SDK surface against the C# records rather than fixing the two the issue named — all still live after #1487:
   - the Node SDK's `patch` posts `{ path, change }` for `PatchDataRequest(WorkspaceReference Reference, RawJson Patch)`: **neither** member matches, so nothing is ever identified to patch;
   - the Node SDK's `watch` sends a reference with no `$type`, which the polymorphic converter cannot resolve to a `MeshNodeReference`;
   - `execute()` flips `requestedStatus` at the **node** level in both SDKs, but `MeshNode` has no such member — `RequestedStatus` lives on the `ActivityLog` **content** (every other caller in `clients/` already nests it correctly, so this one was the odd one out).
2. **A guard instead of a hand-fix.** The names move into one import-free `wire.ts`, mirrored byte-identically into both SDKs, and `wireContract.test.ts` parses the contract sources in `src/` and asserts each name is a declared member of the record it mirrors — with a merge patch checked against `MeshNode`, which is what catches the `execute()` shape. `clients.yml` gains those five sources so a server-side rename reddens the client instead of turning a client feature into a no-op. The byte-identity assertion is the other half: #1475 shipped in both SDKs at once because each hand-mirrored the other, and a one-sided fix is now impossible.
3. `mesh.test.ts` asserts the payload `Mesh` actually delivers, not only its target address — the existing test checked *where* move/copy were routed and never looked inside the envelope, which is the exact blind spot.

**One difference to decide deliberately, not by merge order**: #1487 keeps `Mesh.search` and gives it a REST reach-back inside `Mesh`; this PR leaves `search` untouched. Both are reasonable; they should not be resolved by whoever merges last.

Note also that #1487's search fix only works where `POST /api/mesh/query-nodes` exists, and `Memex.LocalMesh` does not map it — #1494 does, plus `content/list` and `upload`, with a parity check over both backends' endpoint maps.

rbuergi:
Rebased onto main now that #1487 and #1494 have merged. Resolution keeps **both** sides' value rather than picking one:

- **`Mesh.search` — #1487's REST reach-back stands, untouched.** A mesh query is `IMeshQuery.Query`, a service call with no hub message, so REST is the surface that exists. This PR does not compete on it.
- **`move`/`copy` resolved to the `wire.ts` builders** rather than #1487's inline field names. Identical bytes on the wire; the difference is that the names are now checked against the C# records and cannot be edited in one SDK only. (The new inline-literal guard fails on the inline form *by design* — see below.)
- **`mesh.test.ts` kept our broader payload test AND adopted #1487's negative assertion** that `source`/`target` are gone. `toMatchObject` alone would pass if the old names rode along beside the right ones — that is the shape a half-applied fix leaves behind, and our test did not cover it.
- **`clients.yml` took both trigger additions** (our five contract sources, #1494's two endpoint maps) — they guard different tests.

**🔗 Dependency worth stating so a later reader does not undo one and keep the other:** #1487's `search` is only correct *because* #1494 maps `query-nodes` on `Memex.LocalMesh`. Before #1494 the sidecar answered that route with `index.html` and a **200**, so search against the RN app's default instance failed inside `JSON.parse` and was swallowed. Reverting #1494 silently re-breaks #1487.

**Two guard holes closed since review**, both verified by mutation rather than asserted:

1. The byte-identity check compared two 77-line strings, so its failure output was two identical-looking truncated prefixes. It now reports the first differing **line** with both sides:
   `../typescript/src/wire.ts diverged from src/wire.ts at line 52:` + both lines.
2. The identity check only guarded what is *inside* `wire.ts` — an edit building a message inline in `mesh.ts` would bypass both guards, which is the exact state #1475 shipped from. Every message type `wire.ts` owns must now be named **only** there, with the list **derived from `wire.ts`** so a new builder extends the guard automatically and a type it does not own stays correctly out of scope. Re-introducing `observe(path, "DeleteNodeRequest", { path })` fails with the file and type named.

Also confirmed by mutation, for the "a guard that passes on no evidence" concern: a **moved** contract source fails the file at import (`ENOENT`, "no tests"); a **renamed record** throws `no C# record 'MoveNodeRequest' in …` listing the files searched; and a parser degraded to **zero members** turns 9 of 12 red with `parsed no members off record X — the parser broke`. A vacuous pass is not reachable.

Issue #1472, opened 2026-08-14
A shard that never ran leaves the required check green — consolidation counts failures, not whether every shard reported


rbuergi:
On 2026-08-14 ~07:00 a GitHub codeload outage made `setup-dotnet` / `download-artifact` tarballs unfetchable. **Shard 0 never ran.**

`Consolidate test results` — the **sole required check** — reported **success** anyway. `mergeStateStatus` was merely `UNSTABLE`, which is not blocking.

So a PR could have merged with one sixth of the suite never executed, and nothing in the required signal said so.

## Why this is the forbidden shape

AGENTS.md: *a gate never tests its own inputs*, and a check that passes on **no evidence** is worse than one that fails. `collect-results` runs with `always()` precisely so a crashed shard still reds the run — but "the shard produced no results because it never started" is not the same input as "the shard produced no failures", and the consolidation currently cannot tell them apart.

This is the same family as the trapdoors already closed this week: `tests=skipped` reported as a pass (Plugins#451), stacked PRs running no .NET suite at all while showing `CLEAN` (#1401), and `Test Results (shard 1)` reporting *pass* for a shard killed at its cap.

## What it needs

The consolidation must assert **that every expected shard reported**, not merely that no reported shard failed. A missing shard is an `unverifiable` outcome and must red the required check.

🚨 Two ways to get this wrong, both of which reproduce the defect one level up:
- an `if:` that skips the assertion when the shard artifact is absent — GitHub renders a skipped job with the same tick as a passed one;
- `continue-on-error` on the step that fetches the shard results.

The expected shard count is known statically (the matrix), so the check can be a plain comparison against it rather than an inference from what happened to arrive.

## Provenance

Caught because the agent re-ran the failed jobs rather than accepting the green — an infrastructure failure with the same input is the sanctioned re-run exception. Shard 0 then passed for real. Had the green been trusted, the PR would have merged on five sixths of the suite.

PR #1490, opened 2026-08-14
fix(test): CodeCellOutputCaptureTest waits for the post-run cell — un-reds main

rbuergi:
**Main is red** since `41d662b23` on `CodeCellOutputCaptureTest.Run_Stamps_LastActivityPath_And_OutputSegment_Embeds_That_Path` (all three latest main runs, shard 4, deterministic local repro in 2 s).

**RCA — the blamed merge is innocent.** #1468 (`f1dba1216`) deliberately made an unrun cell render **no output segment** (absence instead of the italic 'Not yet run.' strip), updated `CodeCellLayoutTest`, and missed this AI.Test twin. #1468's merge run **reused an already-green tree and skipped every shard**, so the failure first executed on #1400's run and wore its blame. Verified both directions locally: the test fails at `f1dba1216` (pre-#1400) and the whole class passes 6/6 at `5582ec33f` with this fix.

**The defect in the test**: it snapshotted the *first* cell StackControl emission (post-#1468, legitimately the pre-run shape without an output child) and probed it with a non-reactive `.First()`. The fix folds the output segment's existence into the reactive `Match` — wait on the actual condition — leaving the embed assertion unchanged.

Test-only change; no What's New entry.

Unblocks: CD image builds, and PRs #1481 / #1483 / #1484 which all failed on this main red.

### DIFF
--- a/test/MeshWeaver.AI.Test/CodeCellOutputCaptureTest.cs
+++ b/test/MeshWeaver.AI.Test/CodeCellOutputCaptureTest.cs
@@ -157,8 +157,20 @@ public class CodeCellOutputCaptureTest(ITestOutputHelper output) : MonolithMeshT
             .First(a => a is not null
                 && (a == CodeLayoutAreas.CellArea
                     || a.EndsWith("/" + CodeLayoutAreas.CellArea, StringComparison.Ordinal)))!;
+        // Since #1468 an unrun cell renders NO output segment at all (the pane appears only once
+        // LastActivityPath is stamped), so the FIRST cell emission after subscribing may still be
+        // the pre-run shape. Wait for the emission that carries the output segment instead of
+        // snapshotting the first StackControl and probing it non-reactively — the old shape of
+        // this assertion turned the deliberate absence into "Sequence contains no matching
+        // element" and reddened main.
         var cell = (StackControl)(await stream.GetControlStream(cellArea)
-            .Should().Within(30.Seconds()).Match(c => c is StackControl))!;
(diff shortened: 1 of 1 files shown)


rbuergi:
Independently reproduced and reached the same RCA from a parallel session — confirming rather than contesting, and standing down on my duplicate branch so this one lands unopposed.

Two extra data points that pin #1400's innocence beyond inference:

- **#1400's branch provably never contained #1468.** `git merge-base --is-ancestor f1dba1216 266541be0` is false, and `266541be0`'s `CodeLayoutAreas.cs` still carries the pre-#1468 *unconditional* output segment. That is why its own CI was green — it tested a tree in which the stale frame still had the area NAME to find.
- **Clean three-point before/after**, each built and run locally on the exact commit:

  | commit | contains | result |
  |---|---|---|
  | `9fece595f` (pre-#1468) | neither | **PASS** |
  | `f1dba1216` (#1468, without #1400) | #1468 only | **FAIL** |
  | `5582ec33f` (main tip) | both | **FAIL** |

Also worth noting the failure took **~840 ms** in CI (07:31:48.909 → 07:31:49.753) — it never spent its 30 s window, which is the tell that a non-reactive `.First()` was probing a snapshot rather than a wait timing out.

The marker-reuse mechanism your body describes is now filed as #1491 — the reuse itself is sound (`96562caf1^{tree}` == `f1dba1216^{tree}`), the gap is that the bisect window has to come from the last run that *executed* the suite, not the last run that *concluded* success. Same family as #1472, different mechanism.

PR #1494, opened 2026-08-14
fix(#1474): the local sidecar serves the mesh REST verbs its own shells call

### LINKED ISSUE #1474: RN REST calls carry no Bearer token; LocalMesh sidecar does not map content/list + upload — FileBrowser broken in both RN modes

rbuergi:
Closes #1474 (the sidecar half; #1487 has the client half and explicitly leaves this open).

## The defect

`Memex.LocalMesh` serves the packaged JS shells out of `wwwroot`, and those shells post four
`/api/mesh/*` verbs the gRPC message bus does not carry — a mesh query is `IMeshQuery.Query`, a
service call with no request type to post; content listing and upload move bytes that are not mesh
nodes; the markdown render is the one Markdig pipeline. The sidecar mapped **one** of them.

The reason nobody noticed is the failure mode. The host ends with `MapFallbackToFile("index.html")`,
so an unmapped `/api/mesh/*` POST does not 404 — it answers the SPA's own HTML with a **200**. The
caller fails inside `JSON.parse`, and the file browser reports a parse error rather than a missing
endpoint. A 404 would at least have been diagnosable.

This also completes #1473 for the sidecar: #1487 points the clients at
`POST /api/mesh/query-nodes`, which the sidecar did not serve — so against the RN app's *default*
instance, search would still have come back empty (200 + HTML → `JSON.parse` throws → the shell's
`.catch(() => [])` swallows it).

## The fix

- **`LocalMeshApiEndpoints.MapLocalMeshApi()`** maps `render-markdown`, `query-nodes`,
  `content/list` and `upload`, anonymously (this host resolves no identity), running
  `MeshOperations` on a session hub — never the root `mesh/{id}` hub, on which request-shaped work
  never answers.
- **`SessionHubFactory`** (`MeshWeaver.Hosting`) is the hub-materialisation half of
  `SessionHubResolver`, moved down below the transports. `SessionHubResolver`'s own doc says both
  transports must share it "so their routing semantics stay identical" — the headless sidecar has no
  `HttpContext` and so could not, and writing a third copy of `AddData` + `WithNodeOperationExecution`
  + `RegisterStream` is exactly what that comment warns against. `SessionHubResolver` keeps the HTTP
  half (deriving a session id from the request) and delegates.

## The guards — one per side

- **`LocalMeshApiRoutesTest`** asserts the sidecar's route table directly. No mesh, no SQLite, no
  gRPC: the defect IS the absent route, so the test builds the endpoint map and reads it.
- **`clients/grpc-web/src/restContract.test.ts`** holds the other end: every verb the JS shells post
  to must appear in **both** backends' endpoint maps (the portal's `MeshApiEndpoints` and the
  sidecar's). `clients.yml` is triggered by both sources, and `ciTrigger.test.ts` now recognises
  `memex/` as a cross-tree input root so that trigger cannot go stale.

## Fail-before

Renaming the three new routes to `/DISABLED-…` and rebuilding:

```
Failed LocalMeshApiRoutesTest.LocalMeshSidecar_MapsTheVerbTheShellsCall(route: "/api/mesh/upload")
  Expected collection {"/api/mesh/render-markdown", "/api/mesh/DISABLED-query-nodes", …}
  to contain "/api/mesh/upload" because the local sidecar serves the same shells the portal does…
Failed! - Failed: 3, Passed: 3
```

and `restContract.test.ts` goes red the same way, naming the file and explaining the 200-with-HTML
symptom. Restored: 6/6 and 10/10 green. Dropping one of the new `clients.yml` paths turns
`ciTrigger.test.ts` red, so the trigger genuinely covers the new cross-tree reads.

## Verification

- `dotnet build memex/Memex.LocalMesh -c Release -warnaserror --no-incremental` → 0 Warnings, 0 Errors
- `dotnet build src/MeshWeaver.Mcp -c Release -warnaserror` → 0 Warnings, 0 Errors
- `dotnet build test/Memex.Portal.Shared.Test -c Release -warnaserror` → 0 Warnings, 0 Errors;
  `LocalMeshApiRoutesTest` 6/6
- `clients/grpc-web`: `tsc --noEmit`, `tsc -p tsconfig.build.json --noEmit`, `vitest run` 37/37
- `clients/react`: `ciTrigger.test.ts` 4/4

### DIFF
--- a/.github/workflows/clients.yml
+++ b/.github/workflows/clients.yml
@@ -33,6 +33,10 @@ name: Clients
 #                                                         (control/skin vocabulary is DERIVED from it)
 #   src/MeshWeaver.Hosting.Grpc/Protos/**               → the canonical mesh.proto the python,
 #                                                         grpc-web and typescript clients generate from
+#   memex/Memex.Portal.Shared/Api/MeshApiEndpoints.cs   → clients/grpc-web/src/restContract.test.ts
+#   memex/Memex.LocalMesh/LocalMeshApiEndpoints.cs        (BOTH backends must serve every /api/mesh
+#                                                         verb the client SDK posts to — an unmapped
+#                                                         one answers the SPA fallback with a 200)
 #
 # Filtering on "clients/**" alone made every one of those guards blind to the side that actually
 # changes: a server-only PR adds a key (or a control, or a proto field), CI stays green because
@@ -56,6 +60,8 @@ on:
       - "src/MeshWeaver.Messaging.Contract/Localization/Locales.cs"
       - "src/MeshWeaver.Blazor/BlazorViewRegistry.cs"
       - "src/MeshWeaver.Hosting.Grpc/Protos/**"
+      - "memex/Memex.Portal.Shared/Api/MeshApiEndpoints.cs"
+      - "memex/Memex.LocalMesh/LocalMeshApiEndpoints.cs"
   push:
     branches: [main]
     paths:
@@ -65,6 +71,8 @@ on:
       - "src/MeshWeaver.Messaging.Contract/Localization/Locales.cs"
       - "src/MeshWeaver.Blazor/BlazorViewRegistry.cs"
       - "src/MeshWeaver.Hosting.Grpc/Protos/**"
+      - "memex/Memex.Portal.Shared/Api/MeshApiEndpoints.cs"
+      - "memex/Memex.LocalMesh/LocalMeshApiEndpoints.cs"
 
 jobs:
   react:
--- a/clients/grpc-web/package-lock.json
+++ b/clients/grpc-web/package-lock.json
@@ -15,6 +15,7 @@
       "devDependencies": {
         "@bufbuild/buf": "^1.47.0",
         "@bufbuild/protoc-gen-es": "^2.2.0",
+        "@types/node": "^22.20.1",
         "typescript": "^5.6.0",
         "vitest": "^4.1.9"
       }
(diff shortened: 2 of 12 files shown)


rbuergi:
Two mechanical overlaps to expect, both from the same session:

- **#1488** also adds `@types/node` to `clients/grpc-web` (its `wireContract.test.ts` reads the C# contract sources the same way `restContract.test.ts` reads the two endpoint maps). Whichever merges second conflicts on `package.json` + `package-lock.json` — take either side, they are the same edit.
- **#1487** fixes the CLIENT half of #1473/#1474; this PR is the server half it explicitly leaves open. They touch disjoint files. Worth landing #1487 with this, not without it: pointing the clients at `POST /api/mesh/query-nodes` only helps where that route exists, and until this merges the local sidecar answers it with `index.html` and a 200.

PR #1495, opened 2026-08-14
fix(#1476): the JS packs honour AutoSaveAddress instead of dropping the edit

### LINKED ISSUE #1476: JS packs ignore AutoSaveAddress — markdown/code edits in auto-save views silently do not persist

Blazor implements `MarkdownEditorControl.AutoSaveAddress` (and the newer `CodeEditorControl.AutoSaveAddress`) via the process-wide `IMeshNodeStreamCache` (`src/MeshWeaver.Blazor/Components/MarkdownEditorView.razor:159-174`). `clients/react/src/controls/monaco.tsx` has no `autoSaveAddress` handling — it emits an area-pointer `update` event which, for an auto-save editor, has no bound pointer. Edits in such views do not persist on the JS shells and nothing fails visibly.

The control-parity ratchets (`clients/react/src/render/parity.test.ts`) cannot catch this class: they assert $type registration, not per-control semantics. Consider a semantics-level parity checklist for controls whose contract includes a write path.

rbuergi:
Closes #1476.

## The defect

`MarkdownEditorControl.AutoSaveAddress` / `CodeEditorControl.AutoSaveAddress` name a mesh node the
editor edits **in place**; Blazor persists the debounced text through the process-wide
`IMeshNodeStreamCache` (`MarkdownEditorView.razor`, `Monaco/CodeEditorView.razor`).

`clients/react/src/controls/monaco.tsx` had no handling for it. It emitted the standard area-pointer
`update` event — and an auto-save control carries a **literal** value and no binding pointer, so
`useBindingPointer` returned undefined and the write was a no-op. Nothing threw, nothing logged; the
text stayed on screen until the view reloaded, at which point the edit was simply gone. The React
Native pack had the same gap, and additionally mapped `MarkdownEditor` and `CodeEditor` to the *same*
component, so it could not have written to the two different content fields even with the plumbing.

## The fix

`useAutoSave` in `clients/react/src/controls/autoSave.ts`, exported from `@meshweaver/react/core` so
both leaf packs share ONE implementation:

| control | patch |
|---|---|
| MarkdownEditor | `{ content: { content, prerenderedHtml: null, codeSubmissions: null } }` |
| CodeEditor | `{ content: { code } }` |

A merge patch touches only that field, so every other content member and the polymorphic `$type`
survive — stricter than the Blazor markdown path, which builds a fresh `MarkdownContent` and drops
the metadata with it. The two derived markdown fields are cleared **deliberately**: they are caches
of the previous text, and a reader prefers `prerenderedHtml` when present (`ContentLayoutArea`), so
leaving them would render the text the user just replaced.

A pending edit is **flushed on unmount** rather than dropped — navigating away mid-debounce is
exactly when "my last sentence did not save" happens, and a merge patch is idempotent.

The RN pack's two editors are now separate components (they differ only in which content field they
write).

## Why a new test rather than the parity ratchet

`clients/react/src/render/parity.test.ts` asserts `$type` registration, not per-control semantics, so
it is blind to this class — as the issue notes. Both packs therefore get a behavioural test that
drives the editor and asserts the write: debounce, coalescing a keystroke burst into one write of the
final text, flush-on-unmount, and no write when the control did not opt in.

## Fail-before

Removing the `autoSave?.(v)` call from `monaco.tsx` → **4 of 8** red in
`clients/react/src/controls/autoSave.test.tsx` (`expected [] to deeply equal [ { …(2) } ]`).
Removing it from `rnPack.tsx` → **2 of 3** red in `clients/react-native/src/autoSave.test.tsx`.
Restored: 8/8 and 3/3.

## Verification

- `clients/react`: `npm run typecheck`, `npm test` (35 files / 261 tests), `npm run build`
- `clients/react-native`: `npx tsc --noEmit`, `npx vitest run` (16 files / 156 tests)

No server-side change, so no localization surface; no new user-visible string (the editors' own
labels are unchanged).

### DIFF
--- a/clients/react-native/src/autoSave.test.tsx
+++ b/clients/react-native/src/autoSave.test.tsx
@@ -1,0 +1,113 @@
+import { afterEach, describe, expect, it, vi } from "vitest";
+import React from "react";
+import TestRenderer, { type ReactTestRendererJSON } from "react-test-renderer";
+import {
+  AUTO_SAVE_DEBOUNCE_MS,
+  MeshOpsProvider,
+  RegistryProvider,
+  RenderArea,
+  ScopeProvider,
+  StaticAreaSource,
+  type AreaTree,
+  type MeshOps,
+  type UiControl,
+} from "@meshweaver/react/core";
+import { rnPack } from "./rnPack";
+
+// Issue #1476 — the native editors must persist an auto-saving control's edits too.
+//
+// `MarkdownEditor` and `CodeEditor` used to be THE SAME component in this pack, bound only to a
+// layout-area pointer. A control carrying `autoSaveAddress` has no pointer (it edits a mesh node in
+// place), so an edit went nowhere and nothing failed visibly. They now differ by which content field
+// they write, through the same `useAutoSave` the web pack uses.
+
+type Json = ReactTestRendererJSON;
+
+const NODE = "rbuergi/Notes/today";
+
+function fakeOps(patchCalls: { path: string; fields: Record<string, unknown> }[]): MeshOps {
+  return {
+    watch: () => ({ async *[Symbol.asyncIterator]() { await new Promise(() => {}); } }),
+    startThread: async () => ({ path: "p/_Thread/t1" }),
+    submitMessage: async () => "m1",
+    patch: (path: string, fields: Record<string, unknown>) => patchCalls.push({ path, fields }),
+  } as MeshOps;
+}
+
+function tree(control: Record<string, unknown>): AreaTree {
+  return { data: {}, areas: { main: control as unknown as UiControl } };
+}
+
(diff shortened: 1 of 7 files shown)


rbuergi:
Adjacent gap found while writing this, filed separately as #1498 rather than folded in: the RN pack's editors resolve `control.data`, but `CodeEditorControl`/`MarkdownEditorControl` carry their text on `Value` (the web pack reads `control.value ?? control.data`). So a server-sent editor renders EMPTY on RN regardless of auto-save. It is a different defect — it affects those editors whether or not they opted in — and `editors.test.tsx` misses it because its hand-built tree uses the `data` shape the pack happens to support rather than the shape the server sends.

PR #1501, opened 2026-08-14
feat(blazor): self-contained Radzen view pack — no shell tags, one entry point

rbuergi:
First slice of the injectable view-pack lane from the modularization program (artifact: WS2, third lane).

**What changes**
- New `_content/MeshWeaver.Blazor/assetLoader.js`: once-per-document memoized loader for pack assets that cannot ride per-component `<HeadContent>` (classic third-party scripts). Failed loads evict so a later render retries.
- `RadzenViewBase` self-loads `material-base.css` + `Radzen.Blazor.js` on first interactive render; both views gate on `AssetsReady` so Radzen JS interop cannot race the script. Pages without charts/pivots no longer fetch Radzen at all.
- `App.razor` loses both Radzen tags — the shell is pack-free, so a pack can be added or dropped without touching it.
- `AddRadzenViews()` becomes the pack's single hub entry; the pack (and GoogleMaps) leave `AddMeshViews()` — that list is routing discovery only, and neither assembly has `@page` components (verified by grep).

**Why**: the seam prototype for injectable packs (next: null-returning `DefaultFormatting` fallback + `InstallAssemblies` wiring, then the Analysis-views pack, settings tabs, and the Edu extraction per the month-scan).

**Verification**: `Memex.Portal.Shared` builds clean with `-c Release -warnaserror`. Behavior note: the first Radzen view on a page renders one beat later (asset load); no other user-visible change — no What's New entry.

### DIFF
--- a/memex/Memex.Portal.Shared/App.razor
+++ b/memex/Memex.Portal.Shared/App.razor
@@ -31,7 +31,9 @@
     <link href="@(Configuration["Styles:StylesheetName"]).styles.css" rel="stylesheet" />
     <link href="_content/MeshWeaver.Blazor/fonts.css" rel="stylesheet" />
     <link href="_content/MeshWeaver.Blazor/standard-page-layout.css" rel="stylesheet" />
-    <link href="_content/Radzen.Blazor/css/material-base.css" rel="stylesheet" />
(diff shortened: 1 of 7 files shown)


rbuergi:
Both review findings addressed in follow-up #1506 (duplicate-tag guard + retry-on-failure).

Issue #1486, opened 2026-08-14
RoutingGrain.PostFailure is the only inbound path needing the Orleans stream subscription, and it is unbounded: a NACK to a subscriber-less stream succeeds silently


rbuergi:
`RoutingGrain.PostFailure` (`src/MeshWeaver.Hosting.Orleans/RoutingGrain.cs:424`) returns the NACK to the sender by publishing to an Orleans memory stream:

```csharp
streamProvider.GetStream<IMessageDelivery>(delivery.Sender.ToString())
    .OnNextAsync(failureDelivery)
    .ContinueWith(t => { if (t.IsFaulted) … });
```

Two properties make this the weakest inbound path to a co-hosted hub, and they compound.

## 1. It is the ONLY inbound path that requires the stream subscription

Every *other* same-process delivery to a co-hosted hub short-circuits on the local route table (`OrleansRoutingService.cs:162-166`) and never touches a stream. The failure leg does not. So a hub whose stream subscription was never attached — or was attached and then lost — is reachable for forward traffic and unreachable for NACKs. `SubscribeWhenStreamingReadyAsync` can give up (`OrleansRoutingService.cs:614-626`) while the local route stays live, which is exactly that state.

A publish to a stream with **no live subscriber succeeds**. There is no error, nothing faults, and the `ContinueWith` never sees `IsFaulted`. The NACK is simply gone.

## 2. It has no timeout at all

The forward leg (`PostToStream`) is bounded — 60 s, then a NACK. This one is unbounded: a never-completing `OnNextAsync` logs nothing anywhere, because the only diagnostic is the faulted-continuation branch.

The combination means a requester can wait forever for an answer the router believes it sent. That is the shape of `[STALE-CALLBACK]` observations, and it is worth noting the *forward* leg was given a bound and a NACK precisely because unbounded was found to be wrong — the failure leg was left behind.

## Why this is not "just add a timeout"

A timeout here would only convert a silent hang into a logged one; it would not deliver the NACK. The real questions, in order:

1. **Why does the failure leg use the stream at all** when the forward leg for the same target short-circuits to the local route table? If the local route is live, the NACK should take it. That would remove the dependency entirely for co-hosted hubs, which is the majority case.
2. If a stream publish genuinely is required for some targets, **a publish with no subscriber must be detectable** — otherwise this path can never distinguish "delivered" from "discarded", and no amount of bounding fixes that.
[...]

Issue #1249, opened 2026-08-12
Every patch-driven write is persisted twice, and the two routes are unordered — the write guard fires a false 'stale snapshot' alarm on a sequential writer


rbuergi:
Found while root-causing #1193 (`SearchResultStormTest` reding unrelated PRs). That PR fixes the
consequence — a satellite node leaking into a live search feed — but the thing that *created* the
satellite is a second, independent defect, and it is the more serious of the two because it
**devalues a data-integrity alarm**.

## What happens

A single-threaded writer doing sequential updates to one node produces `MonotonicWriteGuard`
CONFLICT warnings and durable `write-conflict-{n}` ActivityLog satellites. Reproduced locally by
`SearchResultStormTest.CatalogSubscription_ContentUpdateStorm_AddsNoExtraRerenders`, which awaits
every `IMeshService.UpdateNode(...)` before issuing the next:

```
[MonotonicWriteGuard] CONFLICT on TestData/storm/a: a write at Version=11 lost the race against
the durable Version=12 … Resolved by merging into the durable row: merged=(none);
latest-wins (stale values DROPPED)=Name … Find the writer that adopted a stale own-node snapshot;
do not relax this guard.
[SaveMeshNode] write at Version=11 for TestData/storm/a was refused — the durable row is at
Version=12 (forked lineage: stale activation seed or a second writer). Rebasing this owner onto
the durable truth so subsequent writes land.
```

It fires on roughly 2 runs in 5 of that test. Nothing about the workload is concurrent — which is
exactly why the message is misleading: there is no forked lineage and no stale activation seed.

## Mechanism: every patch-driven write is persisted TWICE, by two unordered paths

An own MeshNode reaches storage by two independent routes, and both fire for the same change:

1. **The post-commit flush**, per patch, `DataExtensions.ApplyMeshNodePatchInTurn` →
   `IPostCommitFlush.Flush(committed.Value)` → `StoragePostCommitFlush.Flush` →
   `storage.WriteAndPublishUpdated(...)`. Runs OFF the action-block turn, on the reduced stream's
   post-commit emission. This is what the `PatchDataResponse` ack chains off, so it must be
   durable before the ack — deliberate, and it means the row is already at version *n* by the
   time `UpdateNode` returns.
   `src/MeshWeaver.Data/DataExtensions.cs:1405-1413`

2. **The per-node persistence sampler**, `MeshDataSource.InstallPersistenceSampler` — the own
[...]

PR #1502, opened 2026-08-14
fix(#1485): carry answerability across the packaging boundary — the routers stop NACKing heartbeats

### LINKED ISSUE #1485: Both Orleans routers' answer-once guards are dead code: Package() erases the payload type before they run, so the router NACKs heartbeats and can NACK a NACK
Both routers guard their NACK path with an answer-once contract that **cannot ever match on the routed path**, because the payload has already been type-erased to `RawJson` before it gets there.

## The mechanism

`MeshBuilder.cs:238` packages every delivery before handing it to the router:

```csharp
return routes.Hub.ServiceProvider.GetRequiredService<IRoutingService>()
    .DeliverMessage(delivery.Package(routes.Hub.JsonSerializerOptions));
```

`MessageDelivery.Package` replaces the payload with `new RawJson(serialized)`. So at both guard sites, `delivery.Message` is **always** `RawJson`:

- `RoutingGrain.cs:415` — `delivery.Message is DeliveryFailure || delivery.Message?.GetType().HasAttribute<CanBeIgnoredAttribute>() == true`
- `OrleansRoutingService.cs:380` — `MayAnswer(...)`, same two clauses

Neither clause can be true. Confirmed empirically: the trace reads `RouteMessage: NotFound for RawJson → …`.

## What the guards were for, and therefore what is happening instead

The comments state the contract precisely, and it is the right contract:

> `[CanBeIgnored]` messages (HeartBeatEvent, Shutdown/Dispose) are fire-and-forget: there is NO awaiting Observe callback to fail […] It would be dropped as unhandled at the sender, or — for a permanently-gone owner that is heart-beaten every interval — **re-posted forever, which IS the NotFound storm**.

> Never answer a `DeliveryFailure`: the answer is itself a `DeliveryFailure`, so **answering one loops**.

Since neither clause fires on the routed path, the router **does** NACK heartbeats to a departed owner, and **can** NACK a NACK. The monolith `RoutingServiceBase` guards (`:161`, `:359`) sit before packaging and do work — so the two routers, which the comment says "both agree", **disagree in production**.

## Severity

Not theoretical: the heartbeat cadence is 45 s and `SyncStreamOptions.FirstHeartbeat` is 5 s, so a permanently-gone owner generates NACK traffic indefinitely, and it does so during teardown — "exactly the worst moment, since the paths that give up are shutdown and dispatch failure, when the volume of control messages is highest and the process has least capacity", per the doc comment on `MayAnswer`. That is a plausible contributor to several open teardown/storm items (#1470's neighbourhood, the doc-recompile NotFound-storm wedge).

## The fix is already precedented in the same method

`Package` **already** solves the identity-across-type-erasure problem once, and documents why:

```csharp
// 🚨 Carry the payload's IDENTITY across the type erasure. From here on the delivery
// is `type=RawJson` and nothing downstream can tell WHAT it is about without parsing
// the payload — which the receiving hub's storm breaker runs far too hot to do […]
return message is IDiagnosticKeyed keyed && keyed.DiagnosticKey is { Length: > 0 } key
    ? packaged.SetProperty(IDiagnosticKeyed.DeliveryProperty, key)
    : packaged;
```

The storm breaker got its identity stamp in #1200. The answer-once guards need the same stamp — the pre-packaging payload type (or a computed `mayAnswer` flag) written as a delivery property, and both guard sites reading the property rather than `delivery.Message`.

Prefer stamping the *fact* over the *type name* if the guards only ever ask this one question: it keeps the property from becoming a general-purpose type oracle that invites parsing decisions downstream.

## Acceptance

- A test that packages a `HeartBeatEvent` and a `DeliveryFailure`, routes each to a NotFound target through **both** the Orleans router and the monolith router, and asserts no NACK is emitted — failing on the pre-fix tree for the Orleans paths and passing for the monolith ones, which is the "both routers agree" claim made executable.
- Sweep for any other guard that inspects `delivery.Message`'s CLR type *after* the packaging boundary; this defect class is not obviously limited to these two sites.

Found while investigating #1469. Verified by reading `MeshBuilder.cs:222-239`, `MessageDelivery.cs:182-208`, `RoutingGrain.cs:405-417`, `OrleansRoutingService.cs:370-382`.

rbuergi:
Closes #1485.

## The defect

`MeshBuilder` is the **only** call site of `IRoutingService.DeliverMessage`, and it passes
`delivery.Package(routes.Hub.JsonSerializerOptions)`. `Package` replaces the payload with
`RawJson`. Every answer-once guard in the routing layer was a CLR-type test on
`delivery.Message` — `is DeliveryFailure` / `HasAttribute<CanBeIgnoredAttribute>()` — so on the
routed path it inspected `RawJson` and **could never match**.

Consequence: the routers DID NACK `[CanBeIgnored]` heartbeats to a departed owner — re-posted
every heartbeat interval, which *is* the NotFound storm the guards' own comments describe — and
COULD answer a `DeliveryFailure` with a `DeliveryFailure`.

## One thing in the issue was wrong

#1485 read the monolith `RoutingServiceBase` guards as sitting **before** packaging and therefore
still working, making the two routers "disagree in production". They do not. Both routers are
behind the same single packaging call, so the comment claiming they "both agree" was true only in
that **both were uniformly dead**. `MonolithRouterAnswerOnceAfterPackagingTest` is that claim made
executable — it fails on the pre-fix tree exactly like the Orleans one.

## The sweep found seven sites, not two

| Site | Note |
|---|---|
| `RoutingGrain.PostFailure` | in the issue |
| `OrleansRoutingService.MayAnswer` (2 call sites) | in the issue |
| `RoutingServiceBase.NackRouteFailure` | issue said this worked — it did not |
| `RoutingServiceBase.PostNotFound` | issue said this worked — it did not |
| `MonolithRoutingService.PostNotFound` | **also missing the `[CanBeIgnored]` clause outright** |
| `MessageService.NackThroughParent` | `[CanBeIgnored]` half dead behind its RawJson clause |
| `MessageService.ReportFailure` | **the routing TAIL** |

That last one matters most: `ReportRoutingFailure` reports whatever the mesh's route handler hands
back, and that handler *is* `DeliverMessage(delivery.Package(...))` — so fixing only the routers
would have moved the identical storm one level up, on the one path where a router returns `Failed`
synchronously (the Orleans shutdown short-circuit, i.e. every pod shutdown).

## The fix

Follows the precedent already inside `Package` for `IDiagnosticKeyed` (#1200): stamp the **fact**
onto the envelope immediately before the type is erased, and have every guard read the envelope via
`delivery.MayAnswer()` (new `AnswerPolicy`, in `MeshWeaver.Messaging.Contract`).

The *fact*, not the type name — the one question the routing layer asks is "may I answer this", and
a general-purpose type oracle on the envelope would invite the payload-parsing decisions this
codebase deliberately avoids. Only the **suppressed** case is stamped, so ordinary traffic pays
nothing and an unstamped delivery (never packaged, or pre-serialised by an external client) degrades
to exactly the old CLR-type behaviour — never to "answer something you must not". The property is
**presence-checked, never value-compared**: a delivery property crosses the gRPC/SignalR wire as
JSON and returns as a `JsonElement` rather than the `bool` that was stamped, which is the trap
`MessageStormBreaker.ResolvePayloadKey` documents for the sibling stamp.

## Fail-before, demonstrated

Each test was run on the pre-fix tree and is RED there — all three deliveries (heartbeat,
DeliveryFailure, ordinary control) were answered where exactly one should be:

- `OrleansRouterAnswerOnceAfterPackagingTest` — real `IHostApplicationLifetime`, real hub at the
  sender's address; `found 3`
- `MonolithRouterAnswerOnceAfterPackagingTest` — real monolith mesh, real `IRoutingService`;
  `found 3`
- `RoutingTailAnswerOnceAfterPackagingTest` — the Orleans shutdown branch's exact return shape
  (packaged, then `Failed`); `found 3` (verified by reverting just that guard and re-running)

Each also carries a positive control dispatched LAST through the same serial path, so "the
suppressed ones were processed" is proven rather than waited for. `Packaging_ErasesTheGuardsInput`
pins the mechanism itself.

## Verification

- `dotnet build -c Release -warnaserror`: `MeshWeaver.Messaging.Hub.Test`,
  `MeshWeaver.Hosting.Orleans.Test`, `MeshWeaver.Hosting.Monolith.Test` — 0 warnings, 0 errors
  (these transitively cover every touched src project and its dependents)
- `MeshWeaver.Messaging.Hub.Test` — 242/242
- `MeshWeaver.Hosting.Orleans.Test` routing suite — 19/19
- `MeshWeaver.Hosting.Monolith.Test` routing/NACK subset — 32/32
- `MeshWeaver.Documentation.Test` — 21/21 (doc edit + What's New node)

## Not in this PR

**#1486 is deliberately left open** — see the analysis posted there. Short version: its natural fix
(route the NACK through the same path forward traffic takes, instead of publishing straight to the
Orleans stream) is only *safe* once this PR has landed, because without the answer-once guard it
would create a real NACK loop. That is very likely why the direct stream publish exists at all.

### DIFF
--- a/src/MeshWeaver.Connection.Orleans/OrleansRoutingService.cs
+++ b/src/MeshWeaver.Connection.Orleans/OrleansRoutingService.cs
@@ -9,7 +9,6 @@ using MeshWeaver.Mesh;
 using MeshWeaver.Mesh.Services;
 using MeshWeaver.Mesh.Threading;
 using MeshWeaver.Messaging;
-using MeshWeaver.Reflection;
 using MeshWeaver.Utils;
 using Microsoft.Extensions.DependencyInjection;
 using Microsoft.Extensions.Hosting;
@@ -191,14 +190,14 @@ public class OrleansRoutingService : IRoutingService, IDisposable
                 logger.LogDebug("Orleans: {MessageType} → {Address} rejected as {ErrorType} — host is shutting down",
                     delivery.Message?.GetType().Name, address, nameof(ErrorType.ShuttingDown));
 
-                // Same NACK-once contract as RoutingServiceBase.PostNotFound (see MayAnswer), but
-                // DO classify the returned delivery either way so whoever finishes it can.
+                // Same NACK-once contract as RoutingServiceBase.PostNotFound — see AnswerPolicy —
+                // but DO classify the returned delivery either way so whoever finishes it can.
                 //
                 // 🚨 senderNacked is the POST's verdict, not the permission to post. FailedAndNacked
                 // means "the sender has been answered" and suppresses downstream reporting, so
                 // claiming it when SendDeliveryFailure could not post (no mesh hub) would leave an
                 // Observe(...) caller waiting out its full budget with the failure recorded nowhere.
-                var senderNacked = MayAnswer(delivery)
+                var senderNacked = delivery.MayAnswer()
                                    && SendDeliveryFailure(delivery, shutdownMessage, ErrorType.ShuttingDown);
 
                 return Observable.Return(senderNacked
@@ -264,7 +263,7 @@ public class OrleansRoutingService : IRoutingService, IDisposable
                         // no one waiting, so the NACK is pure added traffic. Both matter most
                         // precisely here: dispatch fails in bulk while a silo is leaving, which is
                         // when the mesh can least afford an answering storm.
-                        if (MayAnswer(delivery))
+                        if (delivery.MayAnswer())
                             SendDeliveryFailure(delivery, $"Failed to deliver to {address}: {ex.Message}",
                                 shuttingDown ? ErrorType.ShuttingDown : ErrorType.Failed);
                         return Observable.Empty<IMessageDelivery>();
@@ -364,22 +363,6 @@ public class OrleansRoutingService : IRoutingService, IDisposable
                 }
             });
     }
-
-    /// <summary>
-    /// Whether a NACK may be sent for <paramref name="delivery"/> AT ALL — the routing layer's
-    /// answer-once contract, stated in one place so every path that gives up on a delivery
-    /// applies it identically (the same contract as <c>RoutingServiceBase.PostNotFound</c>).
-    ///
-    /// <para>Never answer a <see cref="DeliveryFailure"/>: the answer is itself a
-    /// <see cref="DeliveryFailure"/>, so answering one loops. Never answer a
-    /// <see cref="CanBeIgnoredAttribute"/> message: nobody is awaiting it, so the NACK is pure
-    /// added traffic — and it is added at exactly the worst moment, since the paths that give up
-    /// are shutdown and dispatch failure, when the volume of control messages is highest and the
-    /// process has least capacity. That is how a teardown turns into a failure storm.</para>
-    /// </summary>
-    private static bool MayAnswer(IMessageDelivery delivery) =>
-        delivery.Message is not DeliveryFailure
-        && delivery.Message?.GetType().HasAttribute<CanBeIgnoredAttribute>() != true;
 
     /// <param name="delivery">The delivery that could not be routed.</param>
     /// <param name="message">The failure message returned to the sender.</param>
--- a/src/MeshWeaver.Documentation/Data/Architecture/ActionBlockWedgePrevention.md
+++ b/src/MeshWeaver.Documentation/Data/Architecture/ActionBlockWedgePrevention.md
@@ -28,6 +28,38 @@ triggers a resubscribe/repost that fails again.
   it holds at *all* emit sites, permanently, enforced by a test — not re-checked by hand.)
 - Root-aligned, not a band-aid: it removes the *cause* of amplification.
 
+### 🚨 The exemption must be read off the ENVELOPE, never the payload's CLR type
+
+**Issue #1485 — the exemption above was written seven times as `delivery.Message is DeliveryFailure
+|| delivery.Message.GetType().HasAttribute<CanBeIgnoredAttribute>()`, and on the routed path not one
+of them could ever match.** Every mesh delivery reaches a router as
+`delivery.Package(hub.JsonSerializerOptions)` (`MeshBuilder`) — the single call site of
+`IRoutingService.DeliverMessage` — and `Package` replaces the payload with `RawJson`. So the guards
+were inspecting `RawJson`, which is neither a `DeliveryFailure` nor `[CanBeIgnored]`. The trace tell
+is `RouteMessage: NotFound for RawJson → …`.
+
+The consequence is exactly the amplification this invariant forbids: a departed owner that is
+heart-beaten every `SyncStreamOptions` interval was answered with a `DeliveryFailure` every time, and
+a `DeliveryFailure` could be answered with another one. It affected **both** hosts equally — the
+comments claiming the two routers "both agree" were true only in that both were dead — and it reached
+one level above the routers too, in `MessageService.ReportFailure`, which reports whatever the route
+handler hands back and therefore also sees `RawJson`.
+
+The fix is `AnswerPolicy` (`src/MeshWeaver.Messaging.Contract/AnswerPolicy.cs`):
+`Package` stamps the *fact* "this payload must not be answered" onto the envelope's properties
+immediately before the type is erased (the same mechanism `IDiagnosticKeyed` already uses for the
+storm breaker), and every guard calls `delivery.MayAnswer()`, which reads the stamp and falls back to
+the CLR type for a delivery that never crossed a packaging boundary. Only the *suppressed* case is
+stamped, so ordinary traffic costs nothing and an unstamped delivery degrades to the old behaviour —
+never to "answer something you must not".
+
+**So when adding a new emit site: call `delivery.MayAnswer()`. A hand-written CLR-type check is a
+silent no-op anywhere downstream of a hub hop.** Pinned by
+`OrleansRouterAnswerOnceAfterPackagingTest`, `MonolithRouterAnswerOnceAfterPackagingTest` and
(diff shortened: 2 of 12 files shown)


rbuergi:
Follow-up commit `0a1de82` — tests rewritten to the standing reactive constraint (no polling, no
`FirstAsync`/`Take(1)`, no `TaskCompletionSource`).

The three tests now push every NACK the router posts onto a `ReplaySubject` and assert by
**subscribing** to a running fold of that stream through the repo's idiom
`.Should().Within(...).Match(...)`. `DeliverMessage(...).FirstAsync().ToTask()` became the sanctioned
cold-observable `.Subscribe(...)` — which *is* the dispatch — and the `TaskCompletionSource` +
`WaitAsync` pair is gone.

The assertion stays a **positive signal, not a wait**: the ordinary control message is dispatched
LAST down the same serial path, so the fold at the moment it is answered already contains anything
the two suppressed deliveries produced. `ContainSingle` on that fold is an assertion about a
completed sequence, never "sleep, then look".

Fail-before re-established for the rewritten bodies (the earlier evidence covered the old ones):
disabling only the stamp in `Package` turns **all five** assertions red — Orleans 2/3, monolith 1/1,
routing tail 2/2, each reporting `found 3` where one is expected.

Re-verified locally on the new commit: `MeshWeaver.Messaging.Hub.Test` 243/243,
`MeshWeaver.Hosting.Orleans.Test` **160/160** (the full suite, including the real-cluster routing
tests), `MeshWeaver.Data.Test` 373/373, plus `-c Release -warnaserror` clean on all three test
projects.

Production code is unchanged by this commit and carries no async at all: `AnswerPolicy` is pure
predicates and the guards are synchronous boolean checks.

rbuergi:
## ⛔ Not merging yet — CI went red on an unrelated test and I am not satisfied it is unrelated

`MeshWeaver.Hosting.Monolith.Test.PatchWriteRouteCollapseTest.SequentialWriter_DeletedText_IsNotResurrectedByADuplicateWrite`
failed on shard 1 of commit `0a1de82`. It is #1249's regression guard: a deleted word came back
because one patch got persisted by two routes.

**Why I did not just re-run it.** Re-running would have been green (the next commit's full CI *was*
green), and that would have hidden the question rather than answered it.

### What I established

- The **immediately preceding commit `504a0db` has a byte-identical `src/` tree** and passed, with
  **all six shards executed** (verified job-by-job, not just the suite conclusion). The only delta
  between green and red is three test files of mine, none of which touches the persistence path.
- No code path: every guard I changed is on a **failure/NACK** path, and the one hot-path addition
  (`Package`) adds no property for non-suppressed payloads, so an ordinary `PatchDataRequest`
  delivery is byte-identical to before. The three `[CanBeIgnored]` types in the tree are
  `HeartBeatEvent`, `ShutdownRequest`, `DisposeRequest` — none is on this test's write path.
- Locally **0 failures in 40** runs (20 default, 20 at `DOTNET_PROCESSOR_COUNT=2`).
- The race is real and *timed*, not structural: `PostCommitFlushRegistry`'s own docs say the
  sampler's mark is read at handler time and the conflict "needs the row to advance inside the queue
  window — milliseconds". So the duplicate route always exists; what varies is whether the flush
  raises its high-water mark before the sampler's 200 ms debounce fires. A regression guard over a
  timed window is flaky by construction under load.

### The measurement that stopped me

`flake-repro.yml` in `rate` mode, 60 iterations per arm on the 2-vCPU runner:

| arm | result |
|---|---|
| `main` ([run](https://github.com/Systemorph/MeshWeaver/actions/runs/31791387918)) | **0 / 60** |
| this branch ([run](https://github.com/Systemorph/MeshWeaver/actions/runs/31791458666)) | **3 / 60** (iterations 9, 31, 60) |

That is *suggestive but not conclusive* — under the null hypothesis of an equal rate, all three
failures landing in one arm has probability ≈ 12.5%. It is nowhere near enough to dismiss, and
"deleted text comes back" is data loss, so I am not merging on a green that could be luck.

Re-running both arms at **110 iterations** to separate the hypotheses
([main](https://github.com/Systemorph/MeshWeaver/actions/runs/31793764438),
[branch](https://github.com/Systemorph/MeshWeaver/actions/runs/31793777089)). If the branch rate is
genuinely higher I will find and fix the mechanism before this merges; the leading candidate is the
per-message reflection `AnswerPolicy.IsAnswerSuppressed` adds inside `Package()`, which is the only
thing in this PR that touches a hot path.

rbuergi:
## ✅ Resolved — the branch was exonerated by the larger sample, and merged on a genuine green

Re-ran both arms at 110 iterations. The verdict inverts the first sample:

| ref | 60 iters | 110 iters | pooled |
|---|---|---|---|
| `main` | 0 / 60 | **7 / 110** (15, 17, 26, 53, 61, 89, 109) | **7 / 170 ≈ 4%** |
| this branch | 3 / 60 | **1 / 110** (24) | **4 / 170 ≈ 2%** |

So `main` flakes *more* than the branch, and the initial 0-vs-3 was precisely the small-sample luck
`flake-repro.yml`'s own header warns about — the same trap in the other direction. Fisher exact on
the pooled counts gives p ≈ 0.5: no difference.

`PatchWriteRouteCollapseTest` is therefore a **pre-existing ~4% flake on main**, filed with the rate
data and a root cause as #1557: #1249's route-collapse is a *timed* window (the sampler's 200 ms
debounced write versus the post-commit flush raising its high-water mark), while its guard asserts
the deterministic claim "there is no duplicate write to hold at all". Each occurrence is also a real
resurrection of deleted content, so it is not only a test problem.

Merged on `9d39fed`, whose `MeshWeaver Build and Test` was `COMPLETED/SUCCESS` with **all six shards
executed** (checked job-by-job, not just the suite conclusion — a green suite can coexist with
skipped shards on a reused tree).

PR #1507, opened 2026-08-14
fix(test): wait for the emission that carries the node, not the first Initial

rbuergi:
Refs #1384 — fixes the `Persistence_StoryContentIsPreserved` member of that flake population, plus four more instances of the same shape.

## The defect

```csharp
.Should().Match(c => c.ChangeType == QueryChangeType.Initial)).Items.FirstOrDefault()
```

The catalog hydrates **asynchronously** from the storage adapter, so a query subscribed early gets an
`Initial` that does not yet contain `graph/story1`; the node then arrives as an `Added` delta, and
`FirstOrDefault()` on the empty snapshot asserted on `null`. The sibling in the same file,
`MeshNode_ChildrenAvailable_ViaPersistence`, already guards exactly this with `Items.Any(...)` — this
instance was missed.

**The 10 s budget is not the defect and is unchanged.** It only decided how often the window was wide
enough to lose. No retry, no widened timeout, no loosened assertion.

**The `ChangeType` check is dropped, not combined** with the item check: the node may arrive in a
later `Added` rather than a second `Initial`, so requiring both would wait for an emission that never
comes. That is why the sibling filters on items alone.

Folding the assertion into the predicate does not cost diagnostics — `Match` reports
`Last of N emission(s) was: …` on timeout, so a genuinely-missing node still shows what the source
actually produced.

## Fail-before — forced, not observed

I could not reproduce the CI timing on demand, so I constrained the interleaving instead: a synthetic
source emitting an **empty `Initial`, then the node 50 ms later as `Added`** — exactly the ordering
the race produces at random.

| predicate | result |
|---|---|
| old (`ChangeType == Initial`) | **fails**, with the CI message verbatim: `Expected value not to be <null> because Story node should exist in persistence` |
| new (`Items.Any(n => n.Path == …)`) | passes |

The scratch proof was removed before commit (it would test the assertions library, not the product).

## The sweep

Same shape across the file and its neighbours — **four more**, all waiting on one specific item, all
converted:

- `CommentWithRepliesViewTest` — parent, reply, subtree
- `NewCommentFlowTest` — the least defensible: it has a **write** before it, so genuine CQRS lag on
  top of the hydration window
- `MarkdownNodeIntegrationTest`

Three carried the *same copy-pasted premise* — "Static node read — no write before, catalog read is
correct (no CQRS lag)". That is one propagated mistake, not five independent ones, and the premise is
simply wrong: CQRS lag was never the only way an `Initial` can be incomplete.

## 🚨 One deliberately NOT converted, and why

`CommentNodeLoadingTest` asserts a **count over a namespace**, and `Added` carries only the *changed*
items (`MeshQuery.TryFilterDuplicateLiveChange`; consumers accumulate — `MeshNodeCollectionView` does
`current.Concat(change.Items)`). So "wait for one emission holding ≥ 6" is **unsatisfiable** if the
six trickle in as separate deltas: it would swap a fast, clear failure for a 20 s hang. I converted
it, realised the delta semantics made it unsound, and reverted it — the reasoning is recorded in-line
so the next reader does not repeat the attempt. That sub-shape needs accumulation *across* emissions,
which changes what the test asserts on.

**Not swept here, reported precisely instead:** `MeshWeaver.FutuRe.Test/FutuReAnalysisTest.cs` has
**10** sites of this shape (lines 222, 247, 270, 293, 316, 339, 412, 472, 843, 869, 1166). Most are
count-over-a-set (`results.Count.Should().Be(6)`), i.e. the same unsound-to-convert sub-shape — so
they need the accumulation treatment, not this one. That project also carries a live, *differently
caused* flake in #1384's ledger (`EuropeRe_AnnualReport_EmbeddedCharts`, a renderer-registration
race), so sweeping it here would make any FutuRe red after this PR ambiguous.

## Verification

- `dotnet build test/MeshWeaver.Persistence.Test -c Release -warnaserror` → **0 Warning(s), 0 Error(s)**
- `dotnet build test/MeshWeaver.Content.Test -c Release -warnaserror` → **0 Warning(s), 0 Error(s)**
- `dotnet test … --no-build --logger trx`, fresh `.trx` each:
  - `DataContextIntegrationTest` — **5/5 passed**
  - the four touched Content classes — **60 passed, 1 skipped** (the skip is pre-existing)

No What's New entry: test-only, no user-visible effect.

### DIFF
--- a/test/MeshWeaver.Content.Test/CommentNodeLoadingTest.cs
+++ b/test/MeshWeaver.Content.Test/CommentNodeLoadingTest.cs
@@ -86,6 +86,13 @@ public class CommentNodeLoadingTest(ITestOutputHelper output) : MonolithMeshTest
     [Fact(Timeout = 20000)]
     public async Task CommentNodes_AreDiscoverableByNamespaceQuery()
     {
+        // 🚨 NOT converted to a wait-for-the-items predicate like its neighbours (#1384). This
+        // asserts a COUNT over a whole namespace, and `Added` carries only the CHANGED items
+        // (MeshQuery.TryFilterDuplicateLiveChange; consumers accumulate — MeshNodeCollectionView
+        // does `current.Concat(change.Items)`). So "wait for one emission holding >= 6" is not
+        // satisfiable if the six trickle in as separate deltas — it would swap a fast, clear
+        // failure for a 20 s hang. Fixing this shape needs accumulation ACROSS emissions, which
+        // changes what the test asserts on; left deliberately for that separate change.
         var comments = (await MeshQuery
             .Query<MeshNode>(MeshQueryRequest.FromQuery(
                 $"namespace:{DocPartitionNamespace} nodeType:{CommentNodeType.NodeType}"))
--- a/test/MeshWeaver.Content.Test/CommentWithRepliesViewTest.cs
+++ b/test/MeshWeaver.Content.Test/CommentWithRepliesViewTest.cs
@@ -242,9 +242,12 @@ public class CommentWithRepliesViewTest(ITestOutputHelper output) : MonolithMesh
         // No hub initialization needed â€” RoutingMeshQueryProvider discovers
         // partitions via DiscoverNewProvidersAsync during the query.
 
-        // Static node read â€” no write before, catalog read is correct (no CQRS lag).
+        // Wait for the emission that CONTAINS the node, never the first Initial: the catalog
+        // hydrates asynchronously, so an early Initial can be empty and the node arrive as Added.
+        // ("No write before, so no CQRS lag" was the premise this carried; CQRS lag is not the only
+        // way an Initial is incomplete — see #1384.)
         var parentNode = (await MeshQuery.Query<MeshNode>(MeshQueryRequest.FromQuery($"path:{CommentC1Path}"))
-            .Should().Match(c => c.ChangeType == QueryChangeType.Initial)).Items.FirstOrDefault();
+            .Should().Match(c => c.Items.Any(n => n.Path == CommentC1Path))).Items.First(n => n.Path == CommentC1Path);
         parentNode.Should().NotBeNull("Parent comment c1 should exist");
         parentNode!.Id.Should().Be("c1");
         parentNode.Namespace.Should().Be(DocPath + "/_Comment");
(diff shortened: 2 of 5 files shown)


rbuergi:
Resolving the two mojibake threads without acting on them here, with the reason.

**Neither is in this diff.** `gh pr diff --patch | grep -c '^+.*â€'` returns **0** — these are pre-existing lines in files this PR touches for an unrelated reason (the unfiltered-emission sweep).

**It is also not two lines.** A repo-wide scan finds **131 files** under `src/`, `test/` and `clients/` carrying double-encoded sequences. Fixing the two that happen to sit near this diff would leave 129 and make the next PR in any of those files carry the same review comment — which is how this has already surfaced twice today (also on #1482, likewise pre-existing).

Filed as its own sweep so it can be done once, mechanically, and verified as a whole. One detail that matters and rules out a blind find-and-replace: the corrupted runs are **not all em dashes** — #1482's five included an *en* dash (`"Final Report – AI Readiness Assessment"`), which a blanket replace with `—` would have silently flattened. The repair has to round-trip each run through cp1252 so it resolves to whatever character it originally was.

rbuergi:
Re the two Copilot comments (mojibake em dashes at `CommentWithRepliesViewTest.cs:242` and `NewCommentFlowTest.cs:144`) — correct observation, but **not introduced here and deliberately not fixed here.**

Both flagged lines are **pre-existing** and untouched by this diff; Copilot surfaced them because they sit adjacent to my hunks. My own added lines are clean — 0 mojibake sequences, 7 proper em dashes:

```
git diff origin/main...HEAD | grep "^+" | grep -c "â€"   # 0
git diff origin/main...HEAD | grep "^+" | grep -c "—"    # 7
```

More importantly it is **systematic, not local**: `grep -rl "â€" test/ --include='*.cs'` matches **117 files**, and 13 occurrences in these two files alone (3 + 10). Fixing the 2 the reviewer's eye happened to land on would leave both files — and the tree — inconsistent, and would put a cosmetic encoding change in a PR whose diff is otherwise a behavioural test fix, making a future `git log -S` on either harder to read.

Filed separately for a single mechanical pass (it is a UTF-8-read-as-Latin-1 artifact, so it wants one sweep with a verified encoding, not per-PR patches).

Merging on that basis: all six shards green on `df9b71e766` — this PR's exact head, verified against `headSha` rather than the tick — with shards genuinely executed (5–9 min each, not a reused-tree skip).

rbuergi:
Encoding sweep tracked as #1539.

PR #1570, opened 2026-08-14
feat: AI providers become boot-loaded module packs — zero provider types in the composition root

rbuergi:
The AI-provider extraction (per the modularization sweep — all six registrations proved pure DI):

- **Enabler**: `MeshNodeProviderAttribute` gains `HubConfigurations`/`DefaultNodeHubConfigurations` virtuals folded by `InstallAssemblies` — also unblocks Courses/Observability/GitSync/InstanceSync packs later.
- **Four boot attributes** register exactly what the old builder calls did (OpenAI family = one DLL/one gate; Anthropic+Foundry; ClaudeCode; Copilot incl. its ConnectStrategy — previously a direct portal type ref).
- **Portal purge**: every provider using/type reference and all 10 call sites gone; the dead `ModelsSettingsTab`/`AiSettingsTab` (unregistered since the AI menu) deleted (~930 lines). `Modules:Assemblies` in both portals' appsettings is the roster — drop a provider family by removing a line.
- **Refs stay as ships-the-bits only** (`ReferenceOutputAssembly=false` verified to drop the DLLs from output — rejected; the modules-folder publish step is the documented follow-up for hard decoupling).

**Verification**: new `ProviderPackBootTest` pins the contract; **full `MeshWeaver.AI.Test`: 1181 passed / 0 failed**; Monolith + Portal.Shared `-c Release -warnaserror` clean; all four pack DLLs verified in app output.

No What's New — identical defaults and behavior.


rbuergi:
All five findings addressed in the follow-up PR (ordering fix, shared CliSkillsDirectory, restored configSection params, doc merge, loader test).

Issue #1568, opened 2026-08-14
A red Clients workflow merged and silently stopped ALL image publication for ~1h — it is not a required check


rbuergi:
On 2026-08-14, **PR #1555 merged with the entire `Clients` workflow RED** — all 7 client jobs failing, including `Portal-next (typecheck + test + build)`. Nothing blocked it, because `Consolidate test results` is the only **required** status check and it is green independently of `Clients`.

The consequence was not confined to the clients.

## What it cost

`main-cd.yml` builds four images and its `promote` job applies real tags **only after every leg succeeds** (all-or-nothing, by design — a partial image set is worse than none). The portal-next leg builds by running `npm run build` in the image. #1555 bumped `next` `^15.5.22` → `^16.3.0`, where `next build` defaults to Turbopack and hard-errors on a `webpack` config with no `turbopack` config. That leg has died in ~1.1 s on every CD run since.

So **CD published nothing at all** from 12:05Z onward:

| CD run | commit | promote |
|---|---|---|
| #3681 | `66822b3f` (11:35Z) | ✅ succeeded — last published tag `3.0.0-rc2.ci.3681` |
| — | `fcbfe8b2a` (12:05Z, PR #1555 merged) | — |
| — | `3cf807f3` (12:25Z) | ❌ portal-next leg failed, promote skipped |
| — | `de1d108c` (13:00Z) | ❌ portal-next leg failed, promote skipped |

`.github/scripts/check-image-set.sh de1d108c` → all four images missing.

Every self-updating install stayed on the previous image, and **a production fix was stranded**: `de1d108c6` fixes `42P18` (which wedges a portal's NodeType bake — see the memex-cloud incident the same day), and it could not ship. That namespace ran on a live config override instead.

## Why this is a required-checks problem, not a dependabot problem

The bump itself is fine and probably wanted. The defect is that **a workflow whose failure can stop image publication is not a merge gate**. `Clients` had *exactly the signal needed* — `Portal-next (typecheck + test + build)` was red, naming the build that CD would later run — and it was advisory.

This is the same family as #1472 (a shard that never ran left the required check green) and #1491 (three honest greens, none covering the tree that went red): the required check is not measuring what we actually depend on.

## What to decide

Make `Clients` (or at least its `Portal-next` job) a **required** status check on `main`.

[...]

Issue #1496, opened 2026-08-14
Node SDK Mesh.watch decodes the DataChangedEvent as the node, so get() always returns empty content and submitMessage drops earlier message ids


rbuergi:
Found while sweeping the TS SDK surface against the C# records for #1475 (the field-name defects are fixed in #1488; this one is a different mechanism and is left open deliberately).

`clients/typescript/src/mesh.ts` — `Mesh.watch` yields `meshNodeFromChange(delivery.message)`, i.e. it treats the **DataChangedEvent itself** as the node:

```ts
for await (const delivery of this.conn.watch(path, streamId, sub.type, sub.message)) {
  yield meshNodeFromChange(delivery.message);
}
```

`DataChangedEvent` carries `{ changeType, change, streamId }`, not `{ path, name, nodeType, content }`. `meshNodeFromChange` (`types.ts`) looks up exactly those four keys, so every emission is

```
{ path: undefined, name: undefined, nodeType: undefined, content: {}, raw: <the event> }
```

Consequences, all silent:

- `Mesh.get(path)` — implemented as the first emission of `watch` — never returns the node. It resolves with an object whose `content` is `{}`, so a caller reads "the node exists but is empty" rather than getting an error.
- `Mesh.submitMessage` reads `current.content["userMessageIds"]` off that value, so it always sees `[]` and sends a patch that **replaces** the thread's id list with a single id — dropping every earlier message id under RFC 7396 array semantics.

The browser twin gets this right (`clients/grpc-web/src/mesh.ts`): it reads `change`/`changeType`, JSON-parses a string `change`, and folds `changeType === "Patch"` through `applyJsonPatch` (RFC 6902 at the node root, per `JsonSynchronizationStream.CreateSingleObjectPatch`) before yielding. The Node SDK has no `jsonPatch.ts` at all.

**Fix**: port the grpc-web decode (and `jsonPatch.ts`) into `@meshweaver/client`, or extract the fold into a shared pure module the way `wire.ts` now shares the request shapes.

**Why it survived**: `clients/typescript` has no test for `Mesh`, and its CI job (`clients.yml` → `node-sdk`) runs `npm run typecheck` only — never `npm test`, though the package defines one and `test/worker.test.mjs` exists. Worth fixing alongside: a decode test is only useful if CI runs it.

PR #1583, opened 2026-08-14
fix: migrate four client packages to TypeScript 7 (Clients workflow)

rbuergi:
The npm-major bump in `688ce589b` (PR #1555) moved TypeScript 5.9 → 7 across the client packages and left **six** `Clients` jobs red. #1567 already fixed `Portal-next`. This fixes **four** of the remaining five. The two react-native jobs need a scope call — see the bottom.

Verified locally by running each job's own commands (`npm ci` from the committed lock, then typecheck / test / build).

## The TS 7 change behind three of them

TypeScript 7 no longer auto-includes every package under `node_modules/@types` the way 5.x did. Reproduced in a pristine project — same tsconfig, same `@types/node@26.2.0`, only the compiler version varying:

| typescript | `import { readFileSync } from "node:fs"` + `process.cwd()` |
|---|---|
| 5.9.3 | clean |
| 7.0.2 | `TS2591` on `node:fs`, `node:path`, `process` |

| job | real error | migration |
|---|---|---|
| `@meshweaver/react` | 11 × `TS2591` — the tests that read the server catalogs off disk (`i18n/localize`, `render/parity`, `ciTrigger`) lost `node:fs` / `node:path` / `process` | `"types": ["node"]` |
| `@meshweaver/client-web` | same `TS2591` in the wire/rest contract guards, **plus** `TS7006` implicit-any that follows from an untyped `readFileSync` | `"types": ["node"]` |
| `Portal example` | `TS5102: Option 'baseUrl' has been removed` | dropped `baseUrl`; the `paths` entries are already relative to the tsconfig, which is exactly what `baseUrl "."` resolved them against — the same one-liner portal-next took in #1567 |

`"types": ["node"]` is the fix the compiler's own diagnostic prescribes. **Nothing is silenced** — the files stay fully checked, and naming one package is *stricter* than 5.x's include-everything.

## `@meshweaver/client` — a real API migration

TS 7 **removed `ts.transpileModule`**. The package's `"."` export is now `lib/version.cjs` (a version string); the API moved to `typescript/unstable/*`, which is `Project`/`Program`/`Checker` shaped with **no single-string transform anywhere in it** (`grep -r transpile` over the package: zero hits). So this is not a config problem and there is no in-package replacement.

The node kernel's type-stripper now uses **`amaro`** — the SWC-based stripper Node itself bundles for `--experimental-strip-types`.

`mode: "transform"`, not `"strip"`, is deliberate: strip-only **rejects** the TS constructs that carry runtime semantics, which `transpileModule` used to compile. Node's own `module.stripTypeScriptTypes` is strip-only, which is why it is not enough. Verified through the real `executeCode`:

| snippet | result |
|---|---|
| `const answer: number = 6*7` | 42 |
| `enum Color { Red, Green } Color.Green` | 1 |
| `namespace N { export const v = 42 } N.v` | 42 |
| `class C { constructor(private readonly x: number){} … }` | 42 |

`typescript` drops to a **devDependency** (it is only `tsc` now), so the worker no longer ships the whole compiler. `amaro` declares `engines: node>=22`, hence the `engines` field and that job moving to **node 22** — Node 20 went end-of-life 2026-04-30.

That job also **gains `npm test`**: it previously ran typecheck only, so the js/ts execution core this PR changes had *no CI coverage at all*.

## Nothing was suppressed

No `ignoreBuildErrors`, no new `skipLibCheck`, no `@ts-nocheck` / `@ts-ignore` / `@ts-expect-error`, no `--legacy-peer-deps` / `--force`, no narrowed `include`, no deleted or skipped test. Net change to how much is checked: **more** (one added test step).

## Local verification

| job | result |
|---|---|
| `@meshweaver/react` | typecheck clean · **261 tests** in 35 files · lib build clean |
| `@meshweaver/client-web` | typecheck clean · **52 tests** in 6 files |
| `@meshweaver/client` | clean `npm ci` → typecheck clean · **5/5 tests** |
| `Portal example` | `npm ci` → `tsc --noEmit` clean · vite build clean |

The i18n catalog drift guard (15 tests) and the `ciTrigger` workflow guard (4 tests) both pass — the latter re-run *after* the `clients.yml` edit.

## NOT fixed here — RN connector / RN web need a scope call

Dependabot bumped **expo 52→57 and react 18→19 but pinned `react-native` at 0.76.5**, whose peers are `react: ^18.2.0` / `@types/react: ^18.2.6`. That set is genuinely unresolvable — the `npm ci` `ERESOLVE` is the drift guard working correctly, not a packaging nit.

Three candidate fixes, all rejected on evidence:

- **`--legacy-peer-deps` / an `overrides` entry** — installs a combination upstream does not support (RN 0.76.5 predates React 19), and the RN web job actually *renders* through Playwright. Cannot be proven, so it cannot be forced.
- **Pin `@types/react` back to 18 for RN** — tested: RN typechecks `../react/src`, which resolves React **19** types from `clients/react/node_modules`, so two React type identities collide → `TS2786` at every JSX call site.
- **Revert `clients/react` to React 18 as well** — would re-break `Portal-next`, which is green *on* React 19.

The coherent direction is forward: **`react-native@0.82` peers are exactly `react: ^19.1.1` / `@types/react: ^19.1.1`** — i.e. the Expo SDK 52→57 upgrade, six RN minors, which also has to deal with `expo-av`. The repo already flagged this in `clients/react-native/package.json`: *"the real fix is the Expo SDK upgrade … which is its own migration."* Tracked separately rather than half-done here.

These two jobs do **not** gate image publication: `main-cd.yml` triggers on *MeshWeaver Build and Test*, and its image legs build `grpc-web` + `portal-next` only — never `react-native`. (The #1568 stranding was the portal-next leg, fixed in #1567.)

No What's New entry: CI/build plumbing, and no workflow publishes these npm packages.

### DIFF
--- a/.github/workflows/clients.yml
+++ b/.github/workflows/clients.yml
@@ -100,7 +100,13 @@ jobs:
       - uses: actions/checkout@v7
       - uses: actions/setup-node@v7
         with:
-          node-version: 20
+          # 22, not 20: jsdom 30 declares engines ^22.22.2 || ^24.15.0 || >=26.0.0 (jsdom 29 still
+          # allowed 20). On Node 20 its undici dependency hits
+          # `webidl.util.markAsUncloneable is not a function` — markAsUncloneable landed in Node 22 —
+          # and EVERY jsdom worker fails to start, so vitest collects ZERO tests and exits 1.
+          # "no tests" is not a pass, which is why this must be the node version and not a fallback.
+          # (Node 20 also went end-of-life 2026-04-30; portal-next has been on 22 all along.)
+          node-version: 22
       - run: npm ci --no-audit --no-fund
       - run: npm run typecheck
       - run: npm test # vitest: renderer core, control gallery, ThreadChat (live thread-node stream), Blazor-parity ratchet
--- a/.github/workflows/clients.yml
+++ b/.github/workflows/clients.yml
@@ -131,9 +131,12 @@ jobs:
       - uses: actions/checkout@v7
       - uses: actions/setup-node@v7
         with:
-          node-version: 20
+          # 22, not 20: the worker's runtime type-stripper (amaro — see package.json's _amaro_note)
+          # declares engines node>=22, and Node 20 went end-of-life 2026-04-30.
+          node-version: 22
       - run: npm ci --no-audit --no-fund
       - run: npm run typecheck
+      - run: npm test # node --test: the js/ts Code-node execution core (type stripping → vm → REPL value)
 
   grpc-web:
     name: "@meshweaver/client-web (typecheck + test)"
--- a/clients/grpc-web/tsconfig.json
+++ b/clients/grpc-web/tsconfig.json
@@ -2,6 +2,11 @@
   "compilerOptions": {
     "target": "ES2020",
     "lib": ["ES2020", "DOM", "DOM.Iterable"],
+    // TypeScript 7 no longer auto-includes every package under node_modules/@types the way 5.x did,
+    // so @types/node has to be named explicitly or the contract tests that read the C# sources off
+    // disk (wireContract, restContract) lose `node:fs`, `node:path` and `process` (TS2591) — and
+    // with readFileSync untyped, their callbacks then degrade to implicit `any` (TS7006).
+    "types": ["node"],
     "module": "ESNext",
     "moduleResolution": "Bundler",
     "declaration": true,
--- a/clients/portal/tsconfig.json
+++ b/clients/portal/tsconfig.json
@@ -9,7 +9,8 @@
     "esModuleInterop": true,
     "skipLibCheck": true,
     "noEmit": true,
-    "baseUrl": ".",
+    // No "baseUrl": TypeScript 7 removed it (TS5102). The "paths" entries below are already
+    // relative to this tsconfig, which is exactly what baseUrl "." resolved them against.
     "paths": {
       "@meshweaver/react/core": ["../react/src/core.ts"],
       "@meshweaver/react": ["../react/src/index.tsx"],
--- a/clients/react/tsconfig.json
+++ b/clients/react/tsconfig.json
@@ -3,6 +3,12 @@
     "target": "ES2020",
     "useDefineForClassFields": true,
     "lib": ["ES2020", "DOM", "DOM.Iterable"],
+    // TypeScript 7 no longer auto-includes every package under node_modules/@types the way 5.x did,
+    // so @types/node has to be named explicitly or the test files that read the server catalogs off
+    // disk (i18n/localize, render/parity, ciTrigger) lose `node:fs`, `node:path` and `process`
+    // (TS2591). This is the fix the compiler's own diagnostic prescribes, not a suppression — the
+    // files are still fully checked, and listing it is stricter than 5.x's include-everything.
+    "types": ["node"],
     "module": "ESNext",
     "skipLibCheck": true,
     "moduleResolution": "Bundler",
--- a/clients/typescript/package-lock.json
+++ b/clients/typescript/package-lock.json
@@ -10,13 +10,17 @@
       "dependencies": {
         "@grpc/grpc-js": "^1.12.0",
         "@grpc/proto-loader": "^0.8.1",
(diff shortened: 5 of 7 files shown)


rbuergi:
Arrived at the same four fixes independently in #1585 before spotting this — closing mine in favour of this + #1587. Two findings from that pass that are not in either PR:

## 1. 🚨 The `worker.ts` TS 7 migration will ship UNTESTED

`clients/typescript` depends on `typescript` at **runtime** — the Node kernel worker transpiles user TS snippets — so migrating `worker.ts` off `ts.transpileModule` changes **product behaviour**, not just a typecheck.

And the `node-sdk` job runs `npm run typecheck` **only**. The package defines `test` and ships `test/worker.test.mjs`, whose second case is literally *"typescript node: types are stripped, then executed"* — the one thing that would prove the migrated transpile still works — and CI has never run it. So this PR can go green with TypeScript Code-node execution broken, exactly as #1555 did.

#1588 adds `- run: npm test` to that job (for #1496, which is the same "a test nobody runs is documentation" defect). Worth taking the same line here, or stacking so the two do not both edit that job.

I took the other route in #1585 — pinning `typescript` back to `^5.9.3` for that package only, on the grounds that TS 7's `"."` export is now `./lib/version.cjs` and the compiler moved behind `typescript/unstable/*`, which is explicitly unstable for a runtime dependency. **Your route is better if it holds**, since it keeps the bump — it just needs the test running to know that it does. With `npm test` wired in, `worker.test.mjs` answers it in one run.

## 2. The `@meshweaver/react` job can still fail at TEST time on the runner

Not a typecheck failure — the vitest **forks pool** fails to start its workers:

```
Error: [vitest-pool]: Failed to start forks worker for test files .../threadChat.test.tsx
Caused by: TypeError: webidl.util.markAsUncloneable is not a function
```

35 unhandled errors, job dead in 34 s. That is undici's webidl against the runner's Node 20.

It hit my branch at 15:26:16 while yours passed at 15:25:05 on a near-identical tree, so it is plausibly runner flakiness rather than something in either diff — but it is the kind that reads as "the PR broke the tests". If it recurs, `node-version: 20` → `22` in `clients.yml` is the likely fix, and it is worth knowing about before it is attributed to a diff.

rbuergi:
Both points are correct, and I'm taking one and deliberately not taking the other. Recording why rather than resolving silently.

**The duplicated compile is real.** `"test": "tsc && node --test test/*.test.mjs"` and `"typecheck": "tsc --noEmit"` — so with both steps the package compiles twice. They are not identical (one emits, because the tests run against the emitted output; one doesn't), but `tsc` type-checks either way, so the `typecheck` step adds no coverage. It costs a few seconds and buys a cleaner failure attribution — a red `typecheck` step vs a red `test` step that happens to be a compile error. I'm leaving it, and I'd rather state that as a judgement call than pretend it isn't redundant.

**The stale job name is a genuine defect and it gets fixed** — a job labelled `(typecheck)` that also runs tests misreports what it did, which is the precise failure mode this whole change set exists to remove. I'm not fixing it *here*, because **#1588 already renames it to `(typecheck + test)`** and that PR is open now. Renaming it in this PR would burn a full CI cycle on a label and deepen the `clients.yml` conflict between the two branches instead of resolving it — see the overlap note I left on #1588.

So the sequence is: this merges as-is, #1588 lands the rename it already carries. If #1588 stalls for any reason, the rename comes back here as a one-liner.

Worth noting the two PRs independently added `npm test` to this same job within hours of each other, having each found that the package shipped `test/worker.test.mjs` and never ran it. That is the argument for #1568.

rbuergi:
Runtime proof for the `ts.transpileModule` migration (the concern raised in review — the node-sdk CI job only typechecks): ran `clients/typescript` `npm test` locally on this branch — **5/5 pass**, including "typescript node: types are stripped, then executed", "a throwing snippet is captured as an error", and the JSON round-trip. #1588 adds `npm test` to the CI job so this stays enforced. The two red RN jobs on this branch's Clients run are #1587's domain (mirror image: #1587's red jobs are the TS7 six fixed here); merging both restores the workflow.

PR #1585, opened 2026-08-14
fix(clients): restore the Clients workflow — TypeScript 7 and React 19 broke six of eight jobs

rbuergi:
**`main` is red.** PR #1555 (the Dependabot npm-major group) merged at 12:05Z with the **entire Clients workflow failing**, and it has stayed that way — six of eight jobs. Nothing blocked it because `Clients` is advisory, which is precisely what #1568 is about. #1567 fixed the portal-next half; this fixes the rest.

## Four breakages, one bump

**1. TypeScript 7 stopped auto-including `@types`.** The Node globals the cross-tree guard tests read (`node:fs`, `node:path`, `process` — in `ciTrigger`, `localize`, `parity`, `restContract`, `wireContract`) no longer resolved: `TS2591` across `@meshweaver/react`, `@meshweaver/client-web` and the RN connector. Fixed by naming `"types": ["node"]` in those tsconfigs.

**2. TypeScript 7 removed `baseUrl`** (`TS5102`), failing the portal example and the RN connector before a single file was checked. Both files' `paths` were already written relative to their own tsconfig — which is exactly how TS resolves `paths` without a `baseUrl` — so dropping the option changes nothing about what they point at.

**3. 🚨 TypeScript 7 removed the JS compiler API from the stable entry — and that is a PRODUCT break, not a build one.** The `typescript` package's `"."` export is now `./lib/version.cjs`; the compiler moved behind `typescript/unstable/*`. `clients/typescript` depends on `typescript` **at runtime**: the Node kernel worker transpiles user TS snippets with `ts.transpileModule`. So the bump broke **TypeScript Code-node execution**, silently.

Pinned back to `^5.9.3` there. This is not reverting a major for convenience — the major deleted a public API the product calls, and the replacement is explicitly `unstable`. The package's own test (*"typescript node: types are stripped, then executed"*) proves it works again, and it **would have caught the break** — except the `node-sdk` job runs `typecheck` only. That half is fixed in #1496's PR.

**4. `react-native@0.76.5` declares peer `react ^18.2.0`**, so bumping react to 19.2.8 made the RN package internally inconsistent and `npm ci` died with `ERESOLVE` before anything ran, taking out **both** RN jobs. `react`, `react-dom`, `react-test-renderer`, `@types/react` and `@types/react-test-renderer` go back to the 18.x they were on; both lockfiles regenerated.

## The part that cannot be fixed by aligning versions

portal-next needs **React 19** (Next 16 requires it) and shares the **same** renderer source (`../react/src`) through a tsconfig alias — so `clients/react` must stay on 19 while `react-native` must be on 18. Compiling that shared source inside the RN program then saw two incompatible `@types/react` and rejected every shared component as *"not a valid JSX element type"*.

The RN tsconfig now pins the `react` specifier to its **own** `@types/react`, so the shared source compiles against the React the RN app actually runs on. The reasoning is at the site, not just here.

## Verified locally, per job, exactly as `clients.yml` runs them

| job | result |
|---|---|
| `@meshweaver/react` | typecheck ✅ · 261 tests ✅ · build ✅ |
| `@meshweaver/client-web` | typecheck ✅ · 52 tests ✅ |
| `@meshweaver/client` | typecheck ✅ · 5 tests ✅ (incl. the TS transpile) |
| `Portal example` | `npm ci` ✅ · typecheck ✅ · build ✅ |
| `RN connector` | `npm ci` ✅ · typecheck ✅ · 156 tests ✅ |
| `RN web (Playwright)` | `npm ci` ✅ · `expo export --platform web` ✅ |

(Playwright itself was not driven locally; its `npm ci` step — the one that failed on main — and its export step both pass.)

## Worth a separate decision

`clients/react-native` pins `react-native@0.76.5` while `expo` is on `~57.0.12`. Those are different SDK generations, and it is why React 19 cannot land there. Aligning react-native with the Expo SDK is the real fix for that split — a deliberate migration, not something to smuggle into a red-main restore.

No **What's New** entry: build/CI only. (The Code-node TypeScript execution break in item 3 never shipped in a release image — CD has published nothing since 12:05Z, which is #1568's other consequence.)


rbuergi:
Closing as a duplicate: #1583 covers the same four jobs (TS 7 `types: ["node"]` + `baseUrl` removal) and #1587 covers the same two react-native jobs (react 18 revert + deduping the React type identity in the RN tsconfig). Both were reached independently from the same evidence; three PRs on the same files would only conflict.

The two findings from this pass that are not in either PR are posted there instead:
- #1583 — the `worker.ts` TS 7 migration ships **untested**: `node-sdk` runs `typecheck` only, and `worker.test.mjs`'s *"typescript node: types are stripped, then executed"* is the case that proves the migrated transpile works. #1588 wires `npm test` into that job.
- Both — the `@meshweaver/react` job can die at test time with `webidl.util.markAsUncloneable is not a function` (vitest forks pool vs undici on Node 20), in 34 s, on a tree that is otherwise green.

Local verification from this branch, in case it is useful to either author: every job's own commands pass — react 261/261 + build, client-web 52/52, client 5/5 (incl. the TS transpile), portal ci+typecheck+build, RN ci+typecheck+156/156, and `expo export --platform web`.

Issue #1584, opened 2026-08-14
Finish the Expo SDK migration for clients/react-native: RN 0.76.5 → ~0.82 and expo-av → expo-audio/expo-video (currently held back at SDK 52)


rbuergi:
The `npm-major` bump in `688ce589b` (PR #1555) moved **expo 52 → 57 and react 18 → 19** in `clients/react-native` but left **`react-native` pinned at 0.76.5**. That combination is not resolvable, so both RN jobs in the `Clients` workflow fail at `npm ci` and have been red since 12:05Z on 2026-08-14:

- `RN connector (typecheck + test)`
- `RN web (Playwright)`

```
npm error ERESOLVE could not resolve
npm error Found: @types/react@19.2.18
npm error Could not resolve dependency:
npm error peerOptional @types/react@"^18.2.6" from react-native@0.76.5
```

`react-native@0.76.5` declares `peerDependencies: { react: "^18.2.0", "@types/react": "^18.2.6" }` — the `react` one is a **hard** peer. RN 0.76 predates React 19 support entirely (RN 0.78 was the first React 19 release).

#1583 fixed the other four red client jobs (TypeScript 7 migration) and deliberately left this one alone, because every small fix is either unprovable or breaks something else.

## Why the cheap fixes don't work

| candidate | why not |
|---|---|
| `npm ci --legacy-peer-deps` / an `overrides` entry for `@types/react` | Forces an install of a combination upstream does not support. `RN web` is a **Playwright** job that actually exports and renders the app — "it installed" is not evidence it works, and RN 0.76.5 + React 19 cannot be shown to work because it isn't supported. |
| Pin `@types/react` back to `~18` for the RN package only | **Tested — fails.** RN typechecks the source-aliased renderer (`../react/src`), which resolves its React types from `clients/react/node_modules/@types/react` (v19), while RN's own files use v18. Two React type identities in one program → `TS2786: 'RenderArea' cannot be used as a JSX component … Type 'bigint' is not assignable to type 'ReactNode'` at every JSX call site. |
| Revert `clients/react` to React 18 too | Would re-break `Portal-next`, which is green **on** React 19 after #1567. |

## The coherent direction is forward

`react-native@0.82` declares peers `react: ^19.1.1` / `@types/react: ^19.1.1` — exactly what the tree already has. So the fix is to finish the upgrade dependabot half-started:

[...]

PR #1587, opened 2026-08-14
fix: make both react-native jobs green — dedupe the React type identity, hold React/Expo back

rbuergi:
Takes the last two red `Clients` jobs green, without the Expo SDK jump. Complements #1583 (the other four).

The npm-major sweep (`688ce589b`) bumped **expo 52→57, react 18→19 and typescript 5→7** in `clients/react-native` but **pinned `react-native` at 0.76.5**. Both RN jobs failed at `npm ci` — before running a single check. Behind that were **three** defects, each established on CI or by direct probe rather than inferred.

## 1. Install — a real version incompatibility

`react-native@0.76.5` **hard**-peers `react: ^18.2.0`, and the sweep put `react@19.2.8` next to it.

Probed rather than read off the CI log, because the log surfaces the `peerOptional @types/react` conflict and that is *not* the binding one. Pinning `@types/react` back to 18 changes nothing:

```
Found: react@19.2.8
Could not resolve dependency: peer react@"^18.2.0" from react-native@0.76.5   ← hard peer
```

**Not** taken: `--legacy-peer-deps`, `--force`, an `overrides` entry. Each installs a combination upstream does not support, and `RN web` is a **Playwright** job that actually renders. An install succeeding is not evidence it works.

## 2. Web export — TypeScript 7 breaks `expo export`

**This one CI found, and it corrected me.** An earlier revision of this branch kept the TS 7 bump on the theory that only the React/Expo half was incoherent. `RN web` went red on exactly this:

```
TypeError: Cannot read properties of undefined (reading 'getCurrentDirectory')
  at evaluateTsConfig (@expo/cli/src/utils/tsconfig/evaluateTsConfig.ts:7:33)
```

`@expo/cli` (SDK 52) reads tsconfig through the **classic TypeScript compiler API** — the same API TS 7 deleted along with `transpileModule` (the package's `"."` export is now just `lib/version.cjs`). It is the identical removal that forced `clients/typescript` onto `amaro` in #1583.

So this is a **whole-package holdback** (expo + react + typescript), not a partial one. Splitting it was tried and does not work.

## 3. Typecheck — a resolution defect, not a version one

This was hiding *behind* (1). RN source-aliases the shared renderer (`../react/src`), whose own `clients/react/node_modules` holds a different `@types/react`. Two React type identities in one program:

```
TS2786: 'RenderArea' cannot be used as a JSX component
  ... Type 'bigint' is not assignable to type 'ReactNode'
```

(React 19's `ReactNode` admits `bigint`; React 18's does not) — **28 errors, none a real incompatibility.**

Fixed with `paths` entries pinning `react` / `react/jsx-runtime` / `react-dom` to one copy: the tsc equivalent of the bundler dedupe `portal-next` already does with webpack `resolve.modules`.

**This silences nothing, and it is not masking a version requirement** — the control being that pinning to the *other* copy (`@types/react` **19**) also yields **0 errors**. The shared core compiles cleanly against either major; the collision was about there being *two*, never about which. That control is why this is a fix rather than a workaround, so it is worth stating explicitly.

## Verification

Each job's own commands, from a clean `npm ci` against the committed lock:

| gate | before | after |
|---|---|---|
| `npm ci` | `ERESOLVE` | exit 0 |
| `tsc --noEmit` | 28 × `TS2786` | **0 errors** |
| `vitest run` | never ran | **16 files / 156 tests passed** |
| `expo export` (RN web) | `getCurrentDirectory` crash | **exit 0**, `dist/index.html` produced |

## What this is not

It is not the Expo upgrade, and does not pretend to be. **#1584 stays open**: `react-native@0.82` peers are exactly `react: ^19.1.1`, i.e. the Expo SDK 52→57 migration, which also has to port **`expo-av`** (used by the speech / push-to-talk pipeline these tests cover). That is a deliberate migration, not a dependency sweep — but it is no longer blocking a red workflow, and the `paths` dedupe here is needed regardless of which React version RN lands on.

The reasoning is recorded where the next sweep will hit it: `_npm_major_holdback_note` in `package.json`, and inline in `tsconfig.json`.

No What's New entry: CI/build plumbing, and no workflow publishes these npm packages.


rbuergi:
Independent confirmation: I reached the **same two conclusions** in #1585 before spotting this PR, from the same evidence — closing mine in favour of this + #1583.

- **Install**: `react-native@0.76.5` hard-peers `react ^18.2.0`; react/react-dom/react-test-renderer/@types back to 18.x, lockfile regenerated. I also had to move `@types/react-test-renderer` (`~19.1.0` → `~18.3.0`) — the second ERESOLVE only surfaces *after* the first is fixed (`peer react@"^19.2.8" from react-test-renderer@19.2.8`), so it is easy to stop one step early. Worth checking this PR carries it.
- **Typecheck**: two `@types/react` identities in one program because the RN tsconfig aliases `../react/src`, which resolves `react` from `clients/react/node_modules`. Pinning the `react` specifier to RN's own copy is what I landed on too.

Verified locally after both, exactly as the workflow runs them: `npm ci` ✅, `npx tsc --noEmit` ✅, `npx vitest run` 156/156 ✅, and `expo export --platform web` ✅ (the RN web job's build step — the export bundles 457 modules cleanly on React 18).

One thing to watch that is not a version problem: the `@meshweaver/react` job can fail at TEST time on the runner with `webidl.util.markAsUncloneable is not a function` (vitest forks pool vs undici on Node 20) — details on #1583. It killed that job in 34 s on my branch while an equivalent tree passed a minute earlier, so it can look like a regression when it is not.

Issue #1497, opened 2026-08-14
Three shells hand-roll the same four /api/mesh fetches — the drift generator behind #1474 (token, error semantics and basePath already differ)


rbuergi:
The generalisation behind #1474, filed with the evidence rather than folded into the fix.

Three shells each hand-roll the SAME four `/api/mesh/*` fetches, once per shell:

| verb | portal-next | portal | react-native |
|---|---|---|---|
| `query-nodes` | `src/client/live.ts` `queryNodes` | `src/live.ts` (via `runQuery`) | `src/liveOps.ts` |
| `render-markdown` | `src/client/live.ts` `renderMarkdown` | `src/live.ts` | `src/liveOps.ts` |
| `content/list` | `src/client/live.ts` `listContent` | `src/live.ts` | `src/liveOps.ts` |
| `upload` | `src/client/live.ts` `uploadContent` | `src/live.ts` | `src/liveOps.ts` |

They already differ in ways nobody chose:

- **The token.** RN's three carried no `Authorization` header at all — that IS #1474, and it is exactly the failure a duplicated fetch invites: the two web shells were written with the token, the third copy was not, and a 401 here reads as "the file browser is empty", not as an auth error.
- **Error semantics.** `queryNodes` returns `[]` on failure in portal-next but throws in the RN/`Mesh.search` shape; `listContent` throws in all three. Neither convention is written down; each copy picked one.
- **The `basePath` composition.** portal-next folds it as `namespace:<base>`; #1487's `Mesh.search` folds it as `path:<base> scope:subtree`. These are different queries, so the same call returns different rows depending on which shell you are in.

**Proposal**: one `MeshRest` in `@meshweaver/client-web` — `{ baseUrl, token?, fetch? }` with `queryNodes` / `renderMarkdown` / `listContent` / `uploadContent`, `Authorization: Bearer` applied in ONE place, and the error convention decided once. Each shell keeps only the verbs genuinely local to it (portal-next's `deleteNode`, `transcribe`, `autocomplete`).

A working implementation plus its tests (header on every verb, the anonymous-sidecar case, the multipart no-content-type case, the basePath composition, the `[]`-not-throw rule) was written while fixing #1474 and then dropped to avoid colliding with #1487, which threads the token through the existing per-shell copies instead. That is the right call for the immediate fix and leaves this open: the copies are still three.

[...]

PR #1588, opened 2026-08-14
fix(#1498, #1496): bind `value`-shaped controls, and decode the change event instead of the node

### LINKED ISSUE #1498: RN editors read control.data, but CodeEditorControl/MarkdownEditorControl carry their text on Value — a server-sent editor renders empty
Noticed while fixing #1476 (auto-save) and left out of that PR deliberately — it is a different defect and affects these editors whether or not they auto-save.

`CodeEditorControl` and `MarkdownEditorControl` carry their text on **`Value`** (`public object? Value { get; init; }` on both records; the Blazor views bind `ViewModel.Value`). The web pack reads that, with `data` tolerated for hand-written trees:

```ts
// clients/react/src/controls/monaco.tsx
const bound = control.value ?? control.data;
```

The React Native pack does not. Its editors go through `useField(control)`, which binds

```ts
// clients/react/src/controls/common.ts
const bound = control.data ?? control.isChecked;
```

so a control that arrives from the server with `value` and no `data` renders an **empty** editor on RN. The user sees a blank box where the page's text should be, with nothing to indicate the value was simply not read.

`clients/react-native/src/editors.test.tsx` does not catch it: it builds its tree by hand with `data: ptr("/data/code")`, which is the shape the pack happens to support rather than the shape the server sends.

**Fix**: resolve `control.value ?? control.data` for display in the RN text editors (the same precedence the web pack uses), and drive the test from a `value`-shaped control so the harness stops encoding the bug. The write path is unaffected — a `value`-bound control has no binding pointer either way, which is what `autoSaveAddress` exists for (#1476).

Same question is worth asking of the other RN leaves that go through `useField`: `useField`'s `data ?? isChecked` is right for form fields, but not for the controls whose C# record names the member `Value`.

rbuergi:
> ⚠️ **Stacks behind #1585.** The `Clients` workflow is red on main (the TypeScript 7 / React 19 bump), so this PR's own `Clients` run cannot go green until that restore lands. Merge #1585 first.

## #1498 — `value`-shaped controls rendered an empty editor on RN

`CodeEditorControl`, `MarkdownEditorControl` and `SearchBoxControl` carry their text on **`Value`** (the Blazor views bind `ViewModel.Value`), but `useField` resolved `control.data ?? control.isChecked`. A control arriving from the server with `value` and no `data` therefore resolved to `undefined` and rendered an **empty** editor, with nothing on screen to say the text had simply not been read.

The precedence now lives in the **shared** helper — `value ?? data ?? isChecked` — rather than in the RN leaves. The web pack's monaco leaf already applies exactly that locally, and putting it in one place is what stops the two packs disagreeing again. It is a no-op for every `Data`-shaped control.

`editors.test.tsx` was building its tree with `data: ptr(...)` — the harness encoding the bug. It exercised the shape the pack *happened* to support instead of the shape the server sends, so an empty RN editor passed. The tree is now `value`-shaped for the three controls whose C# record names the member `Value`. **Verified RED on the previous helper** (2 failures), green after.

## #1496 — the Node SDK decoded the EVENT as the NODE

`clients/typescript`'s `Mesh.watch` yielded `meshNodeFromChange(delivery.message)`. `DataChangedEvent` carries `{ changeType, change, streamId }` and **none** of the four keys `meshNodeFromChange` reads, so every emission was `{ path: undefined, name: undefined, nodeType: undefined, content: {} }`.

Both consequences were silent:
- `get()` resolved to *"the node exists but is empty"* rather than erroring;
- `submitMessage` read `content["userMessageIds"]` off that value, saw `[]`, and sent a patch that **replaced** the thread's id list with a single id — dropping every earlier message under RFC 7396 array semantics.

The browser twin had the fold right all along, so the fix is to stop having two implementations. The fold moves into `changeFold.ts`, mirrored **byte-identically** into both SDKs alongside `jsonPatch.ts` — the same discipline `wire.ts` already uses for the request shapes — and `wireContract.test.ts`'s drift guard now covers all three mirrored files.

> The guard earned its keep immediately: it caught an edit I had made to only the grpc-web copy.

**One defect found while extracting it:** `message["change"] ?? message["Change"]` tested for `undefined`, which collapses **ABSENT** with **NULL** — an event carrying an explicit null `Change` read as "no change member" and the whole *event* was returned as the node. A membership test keeps them apart.

**And why it survived:** `clients/typescript` had no test for `Mesh` at all, and its CI job ran `npm run typecheck` only — never `npm test`, though the script and the runner were both already there. Both halves are fixed: a decode test in each SDK, and the `node-sdk` job now runs the tests.

## Not in this PR

**#1497** — three shells hand-rolling the same four `/api/mesh` fetches. It is a cross-shell refactor rather than a live defect, and folding it in would make this diff unreviewable. It follows as its own change.

## Verification

| package | result |
|---|---|
| `@meshweaver/client-web` | 63/63 |
| `@meshweaver/react` | 261/261 |
| `react-native` | 157/157 |
| `@meshweaver/client` | decode test 6/6 |

Fixes #1498, #1496


rbuergi:
Heads-up on an overlap, from the session fixing the red `Clients` workflow (#1583 + #1587) — no action needed from you yet, just so whoever merges second isn't surprised.

**We independently made the same change.** Both this PR and #1583 add `- run: npm test` to the **`@meshweaver/client`** job in `.github/workflows/clients.yml`, for the same reason you state so well — the package defined `test` and shipped `test/worker.test.mjs`, and the job ran `typecheck` only. Two workstreams found that gap within hours of each other, which is probably the strongest argument for #1568 (making `Clients` a required check).

**Expect a textual conflict there**, and note a naive resolution could end up running `npm test` twice in that job. #1583 also bumps that job to `node-version: 22` in the same region.

**Two ordering notes**, because the toolchain moves under you:

- #1583 migrates `clients/typescript` **off `transpileModule` onto `amaro`** (TypeScript 7 deleted that compiler API). You change `clients/typescript/src/{mesh,changeFold,jsonPatch}.ts` — those want verifying against the migrated toolchain, not the current one.
- #1587 holds `clients/react-native` back to a coherent **expo + react + typescript** set and adds a `paths` React-type dedupe. You add `clients/react-native/src/editors.test.tsx` — that test will run under #1587's versions, and it's a package where `npm ci` currently fails outright on `ERESOLVE`.

Cleanest sequence is probably #1583 → #1587 → then merge main into this branch and re-run, so your new tests execute on the fixed toolchain rather than the broken one. Both of ours are locally verified and only waiting on their required check.

Happy to resolve the `clients.yml` conflict from our side instead if you'd rather — say the word.

rbuergi:
CI red, attributed — neither failure is this diff, which touches only `clients/**` TypeScript and one line of `clients.yml`.

**RN connector + RN web (Playwright)** — inherited from main. Those are the two jobs #1587 fixes (react-native 0.76.5 hard-peers react ^18.2.0, so `npm ci` dies with ERESOLVE before anything runs). This PR will go green on them once #1587 lands; I'll merge main in then.

**`Clients workflow verdict`** — same cause, one level up: the gate is correctly reporting the red Clients run for this commit. Working as intended.

**Shard 3 — `MeshWeaver.Persistence.Test.DataContextIntegrationTest.Persistence_StoryContentIsPreserved`** — unreachable from this diff by construction. The change set is `changeFold.ts`, `jsonPatch.ts`, `mesh.ts` (both SDKs), `common.ts`, two test files and a workflow line; there is no C# in it at all, and no path by which a TypeScript decode helper reaches a .NET persistence test. Recent main runs have all six shards green, so this is not a standing main red either — it looks like the flake population #1384 describes. Not re-running it blind: it will re-run with the main merge for #1587, and if it recurs on that clean input it is worth its own issue with two data points rather than one.

Verified locally on the merged tree, against the amaro-based worker from #1583: `clients/typescript` `npm ci` ✅ `typecheck` ✅ `npm test` 11/11 ✅, grpc-web 63/63 ✅, react-native 157/157 ✅.

rbuergi:
CI caught a real defect the local runs could not, and it is worth naming because it is a three-resolver trap.

**`RN web (Playwright)` failed with:**

```
Error: Unable to resolve module ./changeFold.js from clients/grpc-web/src/mesh.ts
```

grpc-web's source is bundled by **Metro** for the React Native web export, and Metro does not remap a `.js` specifier to `.ts` the way TypeScript's own resolution does. So the mirrored module had **no import spelling that works everywhere**:

| resolver | `./jsonPatch` | `./jsonPatch.js` |
|---|---|---|
| Node SDK (NodeNext) | ✗ extension required | ✓ |
| grpc-web (Bundler) | ✓ | ✓ |
| **Metro** (RN web export) | ✓ | **✗** |

Since the file has to be **byte-identical** in both SDKs for the drift guard, an SDK-specific import was not available either.

**`changeFold.ts` is therefore import-free** — it takes the RFC 6902 applier as a parameter and each SDK passes its own `applyJsonPatch`. That is precisely why `wire.ts` is import-free, and it keeps the mirroring (and its guard) intact. The comment at the top of the file now says so, with this failure named, so the next person does not reintroduce the import.

The *importing* files are not mirrored and differ as they always did: grpc-web extensionless, Node SDK `.js`.

**Re-verified locally, per job, including the one that failed:**

| | |
|---|---|
| `@meshweaver/client-web` | typecheck ✅ · 63/63 ✅ |
| `@meshweaver/client` | typecheck ✅ · 11/11 ✅ |
| `RN connector` | 157/157 ✅ |
| `RN web` | `expo export --platform web` ✅ (1.1 MB bundle, 457 modules) |

The `Clients workflow verdict` failure was the same root — it was reporting that red.

rbuergi:
Review comment applied — and it is a real defect in a byte-identical file, not a nitpick.

The mirror note named ONE package's path (`mirrored into clients/typescript/src/changeFold.ts`), which is wrong precisely *in that copy*: there the sentence describes the file you are already reading. Byte-identity is what turns a path-naming comment into a bug — any sentence in this file has to be true on both sides.

Reworded to "the OTHER TypeScript SDK's copy of the same path", with the reason stated inline so it is not re-specialised later.

Verified: grpc-web 63/63, `clients/typescript` 11/11, and the two copies still compare byte-identical — which is the drift guard's own assertion.

<sub>Note for later, out of scope here: `wire.ts` carries the same one-sided phrasing ("mirrored BYTE-IDENTICALLY into clients/typescript/src/wire.ts") and has the same property, so it is worth the same one-line treatment when someone next touches it.</sub>

rbuergi:
Merged `main` in — this is the sanctioned case for doing so, not a catch-up.

The shard-0 red was `InstanceSyncPushTest.A_pending_content_change_pushes_a_DELETE_when_the_local_node_is_being_deleted`, which is **my regression from #1598** (I rewrote `WaitForRemote` to subscribe to a new `FakeRemoteMesh.Changed` source instead of polling, and it reddened main on two consecutive runs). It is reverted in #1605, now merged.

A `pull_request` run tests `refs/pull/N/merge` as computed when the run started, so re-running the failed job alone would re-sample the OLD merge commit — the one that still contains the defect. Merging the fixed main in is what gets a tree that actually has the fix. That is AGENTS.md's second case verbatim: *"CI fails on something your diff does not touch — merge main BEFORE investigating; a stale branch re-samples flakes main has ALREADY fixed."*

Revert-check done: `git diff origin/main...HEAD --stat` shows only my 12 intended files, no negative delta.

Everything else on the previous run was green, including all four client jobs and the `RN web (Playwright)` job that caught the Metro resolution defect.

Issue #1557, opened 2026-08-14
PatchWriteRouteCollapseTest is flaky at ~4% on main: #1249's route-collapse is a timed window, and its guard asserts a deterministic outcome over it


rbuergi:
`MeshWeaver.Hosting.Monolith.Test.PatchWriteRouteCollapseTest.SequentialWriter_DeletedText_IsNotResurrectedByADuplicateWrite`
— #1249's regression guard — fails on **main** at roughly 4%. Since shard 1 runs it once per CI
run, that reddens on the order of 1 in 25 PR builds repo-wide.

## Measured, on the 2-vCPU runner (`flake-repro.yml`, `rate` mode)

| ref | iterations | failures | failing iterations |
|---|---|---|---|
| `main` | 110 | **7** | 15, 17, 26, 53, 61, 89, 109 ([run](https://github.com/Systemorph/MeshWeaver/actions/runs/31793764438)) |
| `main` | 60 | 0 | — ([run](https://github.com/Systemorph/MeshWeaver/actions/runs/31791387918)) |

Pooled: **7 / 170 ≈ 4%**. Note the two `main` samples on their own read as "0%" and "6%" — this is
exactly the small-sample trap `flake-repro.yml`'s own header warns about, and it is why a single
clean loop must not be read as evidence.

It does **not** reproduce locally: 0/40 on an M-series Mac (20 default, 20 at
`DOTNET_PROCESSOR_COUNT=2`). It needs the real 2-vCPU runner, like the rest of that flake class.

## Root cause: the collapse is a race with a window, not a structural guarantee

`PostCommitFlushRegistry`'s own documentation says it:

> The mark is therefore read at HANDLER time, once the flush has settled.
> … Only a write that regresses can produce a conflict, and that needs the row to advance inside the
> queue window — milliseconds, not seconds.

So the second persistence route (the `MeshDataSource` own-stream `Sample(200 ms)` sampler) **always
exists**; #1249 suppresses its write by having the sampler's handler observe a high-water mark that
the post-commit flush raises. Under CI load the flush can fail to raise that mark before the
sampler's debounced `SaveMeshNodeRequest` reaches its handler — the sampler then writes, the row has
meanwhile advanced, `MonotonicWriteGuardStorageAdapter` refuses it as a version regression and
resolves by merging, and the base-less merge keeps the string SUPERSET, re-adding the deleted word.

The failing runs show precisely that ordering, with the held duplicate appearing *before* the first
[...]

PR #1591, opened 2026-08-14
fix(#1557, #1569): order the route-collapse suppression, and give the token sweep a signal to wait on

### LINKED ISSUE #1569: ApiTokenServiceTests.CreateToken_SweepsTheUsersAlreadyExpiredTokens is CI-load flaky
Failed on PR #1565's run (a diff that deletes education layout areas — cannot reach ApiTokenService), shard 2. Evidence it is not the PR: passes at main tip `daaed54fc` locally, and 10/10 further runs under `DOTNET_PROCESSOR_COUNT=4` (the CI-race reproducer) are green — 11/11 local, red only under real CI load.

The test is new today (the #1477 token-sweep fix). Suspected shape: the sweeps deletion is asynchronous relative to the test's assertion — under CI load the assert samples before the sweeps write lands. Needs the wait-on-the-actual-condition pattern (stream/query filtered on the deleted state) rather than asserting immediately after CreateToken, or the sweep completing observably before CreateToken returns.

Filed instead of blind-rerunning per the working agreement; #1565 is being re-run with this issue as the evidence link.

rbuergi:
Two flakes, both of which were **waiting on the wrong thing**. One of them is also a live data-loss bug.

## #1557 — the route collapse is a timed window, and its guard asserts a deterministic outcome over it

#1249 collapsed the two durable-write routes for a patch-driven own-node change by having the post-commit flush record the version it made durable, so the per-node persistence sampler drops the duplicate. But the flush recorded on the write's **emission** — after the storage round-trip. The sampler's queued `SaveMeshNodeRequest` can reach its handler inside that window, read a high-water that has not been raised yet, and write.

By then the row has advanced, so `MonotonicWriteGuard` refuses the write as a version regression and resolves by merging — and a **base-less merge keeps the string SUPERSET**, silently re-adding text the newer write deleted.

> **This is not only a test problem.** ~4% of runs on the 2-vCPU runner, and each occurrence is a real resurrection of deleted content, reachable in production under load.

### The fix: claim before the write, resolve after

The flush now **`Claim`s** the version *before* its durable write and resolves the claim afterwards — `Record` on success, `Release` on a fault, an empty sequence, or the null *"this adapter does not own this path"* sentinel. Because the claim exists before the write starts, no sampled state at or below that version can slip past, whenever the handler runs. Ordered by construction, not by winning a race.

### 🚨 The part that keeps this from becoming a lost-write bug

A claim is **provisional**, so the sampler must *not* simply drop against one — that would lose the write outright the moment a flush failed. So:

- against a **confirmed** mark, the row is already durable and dropping is right;
- against an **unresolved claim**, the handler **defers**: it waits for the claim to resolve and writes only if it reports that nothing was persisted.

`Forget()` answers any outstanding waiter *"not persisted"* rather than abandoning it, and the TTL sweep never prunes an entry someone is waiting on. The handler's write is extracted to `WriteSampledNode` so the deferred path goes through exactly the same code — including the `MonotonicWriteGuard` refusal contract and `AdoptDurableTruth`.

**Nothing about the guard is relaxed.** Its alarm was always correct; its *input* was ours. `GenuineSecondWriter_StillTripsTheGuard_AndTheRowNeverRegresses` still passes — that is the point of that test.

### 🚨 Local green is not the gate here

This never reproduced locally (0/40 on an M-series Mac, both default and `DOTNET_PROCESSOR_COUNT=2`). The issue names its own gate: `flake-repro.yml` in **`rate` mode over ≥100 iterations** on the 2-vCPU runner, because at ~4% a clean short loop happens by luck. Dispatched against this branch — [run 31816216876](https://github.com/Systemorph/MeshWeaver/actions/runs/31816216876), 110 iterations. **The measured before/after is what should decide this PR**, and I will post it here; main's baseline is 7/170 ≈ 4% from the issue.

## #1569 — the sweep test had no signal to wait for

`CreateToken_SweepsTheUsersAlreadyExpiredTokens` sampled the lagged query index right after the mint. The expiry sweep runs **off** the mint's critical path (deliberately — a user with a backlog must not pay for it at sign-in), so `CreateToken`'s own emission says nothing about whether the sweep has finished. It passed 11/11 locally and failed under real CI load: a flake waiting on the wrong thing.

There was no source to subscribe to, so this adds one. `ApiTokenService.ExpirySweepCompleted` fires as each sweep finishes — **on the fault path too**, because a signal that only fires on success turns a failed sweep into an indefinite wait, which is the same defect one level down. The three sweep tests now arm the wait **before** the mint that triggers it, through a `SweepOf` helper that documents why.

## Verification

| | |
|---|---|
| `PatchWriteRouteCollapseTest` | 3/3 (incl. the genuine-second-writer guard) |
| `MeshWeaver.Auth.Test` | 146/146 |
| the three sweep tests, `DOTNET_PROCESSOR_COUNT=4` | 30/30 |
| `-c Release -warnaserror` | `Mesh.Contract`, `Graph`, `Hosting`, `Hosting.Monolith.Test` — 0 errors, 0 warnings |

Fixes #1569 (and #1557 pending the rate measurement)

### DIFF
--- a/memex/Memex.Portal.Shared/Authentication/ApiTokenService.cs
+++ b/memex/Memex.Portal.Shared/Authentication/ApiTokenService.cs
@@ -1,6 +1,7 @@
 using System.Collections.Concurrent;
 using System.Diagnostics;
 using System.Reactive.Linq;
+using System.Reactive.Subjects;
 using System.Runtime.CompilerServices;
 using System.Security.Cryptography;
 using MeshWeaver.Data;
@@ -46,6 +47,25 @@ internal class ApiTokenService(
     private const int TokenByteLength = 32;
     private const string NodeTypeApiToken = "ApiToken";
     private const string ApiTokenNamespace = "ApiToken";
+
+    /// <summary>
+    /// Fires the namespace of each expiry sweep as it FINISHES (see <see cref="CleanupExpired"/>).
+    ///
+    /// <para>The sweep is deliberately off the mint's critical path — a user with a backlog must
+    /// not pay for it at sign-in — which means <c>CreateToken</c>'s own emission says nothing about
+    /// whether the sweep has run. Without this signal there is no source to subscribe to, so an
+    /// observer can only SAMPLE and hope: that is what made
+    /// <c>CreateToken_SweepsTheUsersAlreadyExpiredTokens</c> fail under CI load (#1569) while
+    /// passing 11/11 locally — the assertion read the lagged query index before the sweep's deletes
+    /// had landed, and the test had no way to wait for the actual condition.</para>
+    ///
+    /// <para>Instance field on the mesh-scoped singleton — never static, so it dies with the mesh.
+    /// Synchronized because sweeps for different users complete on different threads.</para>
+    /// </summary>
+    public IObservable<string> ExpirySweepCompleted => expirySweepCompleted.AsObservable();
+
+    private readonly ISubject<string> expirySweepCompleted =
+        Subject.Synchronize(new Subject<string>());
 
     /// <summary>
     /// Single-flight recency memo for the LastUsedAt stamp, keyed by token hash:
@@ -260,8 +280,17 @@ internal class ApiTokenService(
                     }))))
             .Subscribe(
                 _ => { },
-                ex => logger.LogWarning(ex,
-                    "Expired API token cleanup sweep failed for {Namespace}", userTokenNamespace));
+                ex =>
+                {
+                    logger.LogWarning(ex,
+                        "Expired API token cleanup sweep failed for {Namespace}", userTokenNamespace);
+                    // Fire on the FAULT path too. An observer waiting for this sweep must be told it
+                    // is over either way — a signal that only fires on success turns a failed sweep
+                    // into an indefinite wait, which is the same "no source to subscribe to" defect
+                    // one level down.
+                    expirySweepCompleted.OnNext(userTokenNamespace);
+                },
+                () => expirySweepCompleted.OnNext(userTokenNamespace));
     }
 
     /// <summary>
--- a/src/MeshWeaver.Documentation/Data/WhatsNew/2026-08-14-no-resurrected-edits.md
+++ b/src/MeshWeaver.Documentation/Data/WhatsNew/2026-08-14-no-resurrected-edits.md
@@ -1,0 +1,17 @@
+---
+Name: Deleted text stays deleted
+Category: Fix
+Description: Under load, a rapid sequence of edits could re-add text you had just deleted. It can't any more.
+Icon: Sparkle
+Order: -20260814
+---
+
+# Deleted text stays deleted
+
+When edits arrived in quick succession, the portal could save the same change twice by two different
+internal routes. If the second one was slightly behind, the two versions were merged — and a merge
+with no common starting point keeps everything from both sides, so a word or a paragraph you had
+just deleted quietly came back.
+
+The two routes are now properly ordered rather than racing, so the duplicate save never happens and
+there is nothing to merge. Deletions stick, however fast you type.
--- a/src/MeshWeaver.Graph/MeshDataSource.cs
+++ b/src/MeshWeaver.Graph/MeshDataSource.cs
@@ -330,11 +330,57 @@ public static class MeshDataSourceExtensions
         var alreadyFlushed = flushed?.HighWater(node.Path) ?? 0L;
         if (node.Version <= alreadyFlushed)
         {
+            // 🚨 CONFIRMED versus CLAIMED (#1557). A confirmed mark means the row IS durable at or
+            // above this state, so dropping is right. An unresolved CLAIM means the flush has taken
+            // the version but its write is still in flight — whether it lands is not yet known, and
+            // dropping there would lose the write outright if the flush then failed or answered the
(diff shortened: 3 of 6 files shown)


rbuergi:
## The gate measurement: **0 failures in 110 iterations**

`flake-repro.yml`, `rate` mode, on the 2-vCPU runner — the same instrument and the same iteration count the issue used for its baseline. [Run 31816216876](https://github.com/Systemorph/MeshWeaver/actions/runs/31816216876), against `fix/flake-roots @ 84f79fea5`:

```
RESULT: 0 failure(s) in 110 iteration(s) run (requested 110); failing iterations: none
```

| ref | iterations | failures |
|---|---|---|
| `main` (from #1557) | 110 | **7** |
| `main` (from #1557) | 60 | 0 |
| **this branch** | **110** | **0** |

Pooled main baseline is 7/170 ≈ 4%. At that rate, 110 iterations would be expected to produce ~4.4 failures, and observing **zero** has probability ≈ `0.96¹¹⁰` ≈ **1%**. So this is evidence, not a lucky loop — which is exactly the distinction the issue's own header insists on ("a single clean loop must not be read as evidence"; a rate over ≳100 iterations is).

🚨 Stated precisely: the run measured `84f79fea5`. The commit after it (`85f279b8c`) only *strengthens* the same mechanism — it resolves the claim on **unsubscribe** as well as on the terminal notifications, closing the one window where a claim could be stranded (a subscription torn down mid-write, i.e. hub teardown) and a deferred sampler could then wait forever. That is strictly more claim-resolution, never less, so it cannot re-open what the measurement closed. I have not re-measured on top of it; say the word if you want the 110 re-run against the tip before merging.

rbuergi:
## The shard-2 red is not this diff — measured, not assumed

`SilentReadNackTest.OwnerDisposedWithReadOutstanding_IsNacked_NotLeftHanging` fails **21 of 60 times on `main` itself**, unmodified, on the same 2-vCPU runner:

```
flake-repro.yml, rate mode, --ref main  → run 31818285657
RESULT: 21 failure(s) in 60 iteration(s); failing iterations: 3 7 14 15 18 20 22 23 25 29 30 35 36 37 40 41 42 43 53 56 58
```

35% on main. Filed as #1599 with the mechanism (two disposal arms race; the test's own doc says it pins one of them specifically).

Attribution, rather than assumption:
- this diff touches the post-commit flush ordering, the flush registry, the sampler's save handler and the ApiToken sweep signal — none of which `HandleGetDataRequest` or hosted-hub creation goes through;
- the same test passed **15/15** on this branch locally under `DOTNET_PROCESSOR_COUNT=4`;
- and it fails at 35% on a tree that does not contain this change at all.

Re-running the failed shard on that basis.

rbuergi:
All three review comments addressed — two applied, one already in the branch.

**1. The deferred write bypassed the hub's inbox — applied, and this was the important one.**

The claim resolves on whatever thread the flush completes on (a storage/emission thread), and the deferred path called `WriteSampledNode` straight from there. That bypasses the single-threaded action block every other handler is serialised by, and the subscription could outlive teardown — the two properties the actor model exists to guarantee.

It now re-posts `SaveMeshNodeRequest`, so the retry goes back in the queue where it is ordered against every other message and refused outright once the hub is shutting down, and the subscription is registered for disposal so it dies with the hub.

It cannot loop: a resolved claim is **removed** from the registry, so the re-posted request reads a high-water of 0 for that version and writes. If a newer claim exists by then it defers once more against *that* one, which resolves in its turn.

**2. `Finally()` for a claim left unresolved by an unsubscribe — already there.** Landed in `85f279b8c`, before the review ran. Same reasoning, independently: Rx does not deliver `OnError`/`OnCompleted` on disposal, so without it a teardown mid-write strands the claim and a deferred sampler waits forever. Good catch either way — it is the failure mode that worried me most about the deferral.

**3. Two consecutive `<summary>` blocks — applied.** `SweepOf`'s doc had been inserted between `ObserveNode`'s doc comment and `ObserveNode` itself, so both bound to the same member. Moved below `ObserveNode`.

Re-verified: `PatchWriteRouteCollapseTest` 3/3, `MeshWeaver.Auth.Test` 146/146, and `MeshWeaver.Graph` / `MeshWeaver.Auth.Test` / `Hosting.Monolith.Test` build `-c Release -warnaserror` with 0 errors.

🚨 Since the deferral changed shape, the 110-iteration rate measurement above no longer covers the tip. Re-dispatching `flake-repro` in `rate` mode against it — I will not merge this on the older run's evidence.

rbuergi:
## Re-measured on the tip: **0 failures in 110 iterations**

The earlier run measured `84f79fea5`; the review then changed the deferral's shape, so that number no longer covered the code being merged. [Run 31820447173](https://github.com/Systemorph/MeshWeaver/actions/runs/31820447173) against `d020cb5a8` — the tip, with the `Finally()` claim-resolution and the re-enqueue-through-the-hub:

```
RESULT: 0 failure(s) in 110 iteration(s) run (requested 110); failing iterations: none
```

| ref | iterations | failures |
|---|---|---|
| `main` (from #1557) | 110 | **7** |
| branch @ `84f79fea5` | 110 | 0 |
| **branch @ `d020cb5a8` (tip)** | **110** | **0** |

Main's pooled baseline is 7/170 ≈ 4%, so 110 clean iterations has probability ≈ `0.96¹¹⁰` ≈ 1%. Both measurements agree, and the second one covers the code that actually merges.

Merging on that.

Issue #1391, opened 2026-08-13
Edu/Course and BinaryClickerV2/BinaryToggle are at CompileError on the running production image


rbuergi:
Two NodeTypes fail to compile on the image currently running in production (`memex-portal-ai:3.0.0-rc1.ci.3367`, namespace `memex`), on **every** pod boot:

```
fail: MeshWeaver.Graph.Configuration.MeshNodeCompilationService[0]
      Failed to compile assembly for node Edu/Course. Executed source queries (2):
      MeshWeaver.Graph.Configuration.CompilationException: Compilation failed for 'Edu/Course':
      BatchBake: Edu/Course → CompileError

      BatchBake: BinaryClickerV2/BinaryToggle → CompileError
```

## Why this is not cosmetic

A NodeType left at `CompileError` **refuses portal readiness** and parks every instance hub for the full activation budget. That is visible in the same window as repeated readiness failures:

```
fail: Microsoft.Extensions.Diagnostics.HealthChecks.DefaultHealthCheckService[103]
```

firing every ~10 s while the startup probe returned **503**. So these two types are a plausible contributor to the slow/unavailable window users saw during the 2026-08-13 06:14 UTC roll — not merely a log-noise item. (The roll's severed *sessions* were a separate cause, fixed by #1389; this is about how long the new pod took to become ready.)

## Why CI cannot see it

🚨 These are **in-mesh NodeTypes**, compiled at RUNTIME in the portal. `dotnet build` never type-checks them and no test in this repo exercises them, so a clean `-c Release -warnaserror` build proves nothing about either. The only gates that would catch this are the per-NodeType compile sweep and the cross-repo plugin gate.

## What to establish

1. **The actual diagnostic.** The Loki line is truncated before the CS error is reachable — `Executed source queries (2):` is followed by the detail, which is cut. Note #1378 (in flight) fixes a truncation of exactly this shape in the test-harness fault record; check whether the same cap applies here, because right now **the log names the failure but not its reason**, which is the thing that makes this hard to action.
2. **Whether they are one root or two.** `Edu/Course` and `BinaryClickerV2/BinaryToggle` are unrelated content; two independent breakages appearing together is more likely a framework-surface change than two coincidences. Deleting or renaming any public framework symbol is a breaking change to code the compiler cannot see — check recent removals against these two types' sources.
3. **Whether more types are affected** and simply not logged in the sampled window — sweep every NodeType's status rather than trusting these two to be the whole set.

## Not the fix

[...]

PR #1593, opened 2026-08-14
fix(#1462): a stranded `Compiling` claim is re-driven by the sweep that already runs

### LINKED ISSUE #1462: A NodeType row can sit at compilationStatus='Compiling' indefinitely — nothing reconciles a non-terminal compile state
Split out of #1391, where it was found but does not belong.

## The observation

A row in `public.mesh_nodes` has sat at **`compilationStatus = 'Compiling'` for ten weeks**, with **3 `currentSourceVersions` keys and 0 actual Code rows** under its `Source/`.

**In that specific spot it is harmless**: it is an orphan the prewarmer never enumerates, so it is in no bake set and gates nothing. That is exactly why it should not stay filed inside #1391 — nothing there is actionable about *this row*.

## The general fact, which is not harmless

**`Compiling` is a non-terminal state with no reconciler.** Nothing sweeps a row that entered it and never left. This row demonstrates that a compile state can persist indefinitely; it just happens to be an orphan, so nobody notices.

Apply the same mechanism to a row that **is** in the bake set and the consequences are ones this repo has already paid for:

- `HasUsableBuild` and the readiness gate reason about `compilationStatus`. A type wedged at `Compiling` is neither `Ok` (usable) nor `Error` (classifiable by #1396's discriminator, which keys on `CurrentSourceVersions` for a type in a *failed* state).
- A NodeType that never reaches a terminal state parks its instance hubs for the full activation budget — the readiness-refusal shape #1391 opens with.
- The recovery pattern already exists elsewhere and is deliberately NOT applied here: `ActivityControlPlane`'s wake-up recovery flips a `Running` activity to `Failed(interrupted)` on first emission, but only where the owner hub is **distinct** from the executor. `KernelContainer` documents exactly why it must not do that for the kernel's own activity.

## What to establish

1. **How a row enters `Compiling` and never leaves.** Process death mid-compile is the obvious candidate (the flip to `Compiling` is durable, the terminal write never happens) — but confirm it rather than assume; this row's 3 recorded source versions against 0 Code rows suggests its sources went away *during or after* the flip.
2. **Whether any row currently in a bake set is affected**, on any portal — the one found is an orphan, and a sweep is the only way to know it is the only one.
3. **Where reconciliation belongs.** Candidates: the compile watcher's own wake-up path (the owner re-requests from its own state, the pattern `ActivityControlPlane.md` sanctions when owner ≠ executor), or the prewarmer treating a `Compiling` row older than the compile budget as stale and re-queueing it.

## Not the fix

Do not add a timer that polls for stale `Compiling` rows and flips them — that is the watchdog-shaped band-aid AGENTS.md rules out. The question is why the terminal write is missing, and the recovery should be driven by the same event that already re-establishes state on activation.

🤖 Split from #1391 so it has a home; the row that revealed it is inert, the mechanism is not.

rbuergi:
`Compiling` is a **non-terminal state that nothing reconciles.** The flip to it is durable; the terminal write is not guaranteed — a process death mid-compile leaves the row there for good.

`DynamicTypePreWarmer`'s *"don't clobber an in-flight compile someone else already started"* guard then declines to touch it on **every subsequent sweep**:

```csharp
if (def.CompilationStatus is CompilationStatus.Pending or CompilationStatus.Compiling)
    return node;   // ← forever, for a claim whose owner died
```

So the type never reaches a terminal state. It is neither `Ok` (usable) nor `Error` (classifiable by #1396's discriminator), it **holds portal readiness**, and it parks every instance hub for the full activation budget. The row that revealed this sat at `Compiling` for **ten weeks** — harmless only because it was an orphan the prewarmer never enumerates. Apply the same mechanism to a row that *is* in the bake set and it is the readiness-refusal shape #1391 opened with.

## The fix

The guard now defers only to a claim that **can still be in flight**:

| state | verdict | why |
|---|---|---|
| `Pending` | honoured | it is what this prewarmer itself writes to *ask* for a compile — re-driving it would fight our own request |
| `Compiling`, started within the per-type budget | honoured | it can still be in flight; clobbering restarts work that is about to finish |
| `Compiling`, started longer ago than the budget | **stranded** | no driver is still working on it — it either finished (and would have written a terminal state) or it died |
| `Compiling`, **no** start timestamp | **stranded** | a live compile always stamps one, so this is the shape a row left over from an older write carries — and honouring it forever is how this became permanent |

The bound is the sweep's own per-type budget, not a new number.

## 🚨 What this is not

**No timer, no poller, no background sweep for stale rows.** The issue rules those out explicitly (*"Do not add a timer that polls for stale `Compiling` rows and flips them — that is the watchdog-shaped band-aid AGENTS.md rules out"*), and so does AGENTS.md.

The recovery **rides the enumeration that already runs**, and only ever reinterprets a claim that has *provably expired*. `StrandedCompileClaimTest` pins exactly that distinction — including the case that must NOT change (a compile 5 s old is still honoured), because a fix here that clobbered live compiles would be worse than the bug.

## Verification

- `MeshWeaver.Hosting.Test` — **176/176**, 8 of them new.
- `-c Release -warnaserror` — `MeshWeaver.Hosting`, `MeshWeaver.Hosting.Test`: 0 errors, 0 warnings.

No **What's New** entry: the user-visible effect (a portal that will not become ready) is a failure that this prevents rather than a change they can notice.

Fixes #1462

### DIFF
--- a/src/MeshWeaver.Hosting/DynamicTypePreWarmer.cs
+++ b/src/MeshWeaver.Hosting/DynamicTypePreWarmer.cs
@@ -801,9 +801,27 @@ public static class DynamicTypePreWarmer
                                 {
                                     if (node?.Content is not NodeTypeDefinition def)
                                         return node!;
-                                    // Don't clobber an in-flight compile someone else already started.
-                                    if (def.CompilationStatus is CompilationStatus.Pending
-                                                              or CompilationStatus.Compiling)
+                                    // Don't clobber an in-flight compile someone else already started…
+                                    // …but ONLY while it can still be in flight (#1462).
+                                    //
+                                    // 🚨 `Compiling` is a non-terminal state and nothing else reconciles
+                                    // it. The flip to `Compiling` is DURABLE; the terminal write is not
+                                    // guaranteed — a process death mid-compile leaves the row there for
+                                    // good. This guard then declines to touch it on every subsequent
+                                    // sweep, so the type never reaches a terminal state: it is neither
+                                    // `Ok` (usable) nor `Error` (classifiable), it holds portal
+                                    // readiness, and it parks every instance hub for the full activation
+                                    // budget. One row on `public.mesh_nodes` sat at `Compiling` for TEN
+                                    // WEEKS this way; it was harmless only because it was an orphan the
+                                    // prewarmer never enumerates.
+                                    //
+                                    // A claim older than the per-type budget cannot still be in flight —
+                                    // whoever made it would have written a terminal state or been given
+                                    // up on long ago — so it is re-driven rather than deferred to.
+                                    // Note what this is NOT: no timer, no poller, no background sweep for
+                                    // stale rows. The recovery rides the enumeration that already runs,
+                                    // and only ever reinterprets a claim that has provably expired.
+                                    if (IsLiveCompileClaim(def, budget))
                                         return node;
                                     return node with
                                     {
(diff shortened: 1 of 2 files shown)


rbuergi:
Review comment checked and **not applied** — `using MeshWeaver.Mesh.Services;` is required, not unused.

`CompilationStatus` lives in `MeshWeaver.Mesh.Services` (`src/MeshWeaver.Mesh.Contract/Services/CompilationStatus.cs`), not in `MeshWeaver.Graph.Configuration` where `NodeTypeDefinition` lives — the test names both. Removing the line and building with the same flags:

```
StrandedCompileClaimTest.cs(27,43): error CS0246: The type or namespace name 'CompilationStatus' could not be found
StrandedCompileClaimTest.cs(75,55): error CS0246: …
StrandedCompileClaimTest.cs(71,17): error CS0103: The name 'CompilationStatus' does not exist in the current context
StrandedCompileClaimTest.cs(72,17): error CS0103: …
```

Restored, and `dotnet build test/MeshWeaver.Hosting.Test -c Release -warnaserror` is back to **0 Error(s), 0 Warning(s)** — which is also what the green CI on this PR already showed, since CS8019 would have failed the build under `-warnaserror` exactly as the comment describes.

Resolving the thread.

Issue #1477, opened 2026-08-14
Every portal-next SSR page render mints two persistent ApiToken nodes; no expiry sweep exists


rbuergi:
Found during the 2026-08-14 modularization investigation.

`clients/portal-next/src/server/snapshot.ts:71-91` POSTs `/api/tokens` per request (`expiresInDays: 1`); `ApiTokenService.CreateToken` (`memex/Memex.Portal.Shared/Authentication/ApiTokenService.cs:102-160`) writes two nodes per mint: `{userId}/ApiToken/{hash12}` and the global index `ApiToken/{hash12}`. `React.cache` dedups within one render only, and no expiry/cleanup job exists anywhere (grep over the auth folder finds only `OAuthCodeStore`). A user browsing 50 pages leaves 100 permanent token nodes in their partition; a crawler is worse. The browser mints another per tab (`live.ts:111`).

Root fix: accept cookie-or-Bearer on the non-MCP `/api/mesh/*` verbs so SSR never needs to mint, plus signed short-lived URLs for `/api/content`; independently, an expiry sweep for `ApiToken` nodes whose `expiresAt` has passed.

PR #1595, opened 2026-08-14
fix(#1493): a re-authorization replaces the client's credential instead of adding one

### LINKED ISSUE #1493: Two ApiToken mint paths still accumulate: per-page-load browser mint, and a fresh 365-day token per OAuth authorization
Follow-up to #1477, which removed the SSR mint (two permanent nodes per *page render*) and added an expiry sweep. Two mint paths remain; neither is unbounded any more, but both still accumulate.

**Evidence.** `search 'nodeType:ApiToken'` on the live memex portal returns **86** user-scoped rows across two users (the global `ApiToken/` index rows are additional and not visible to a normal user's query). The population is dominated by `API Token: OAuth: <code>` labels — the MCP OAuth `/token` exchange — with the same `client_id` recurring under many different codes. Eight `portal-next-ssr` / `react-portal-next` rows date from a single afternoon of local development against it.

### 1. The browser mints one token per page load

`clients/portal-next/src/client/live.ts:111` (and the Vite SPA's `clients/portal/src/live.ts`) POSTs `/api/tokens` on every full page load, because gRPC-web `Connect` is identified by a Bearer `mw_…` token rather than the session cookie. `acquireLiveMesh` shares one connection per tab, so this is per load, not per component — but a reload is another two nodes.

Now **bounded**: those tokens carry `expiresInDays: 1`, so #1477's sweep removes them on the user's next mint. Eliminating it means letting the gRPC transport authenticate from the same-origin session cookie, which is a materially larger change than #1477's REST-verb policy split and touches both React shells.

### 2. The OAuth exchange mints a fresh 365-day token per authorization

`OAuthConnectController` issues a new token with `TokenLifetime = 365 days` on every successful `/token` exchange (`expiresAt` is set, so the sweep does eventually reach it — but only after a year). There is no refresh-token flow, so a client that re-authorizes — a reinstall, a new device, a revoked-and-reconnected integration — leaves the previous year-long credential live and listed. That is the shape behind most of the 86 rows above.

Worth deciding between: reusing a still-valid token for the same `(userId, client_id)` instead of minting, revoking the previous token for that `client_id` on a fresh authorization, or implementing the refresh flow the comment already anticipates. The middle option is the smallest and also the best security answer — one live credential per client, not N.

Neither is a regression from #1477; both are named here so they are not lost.

rbuergi:
The OAuth `/token` exchange minted a fresh **365-day** token on every authorization and left the previous one **live and listed**. A reinstall, a new device, or a revoked-and-reconnected integration therefore each left another usable key behind — for a year.

That is the shape behind most of the **86 `ApiToken` rows** found on the live portal, and every one of them still opens the door.

## Which of the three options

The issue weighs reusing a still-valid token, revoking the previous one, or implementing the refresh flow, and calls the middle one *"the smallest and also the best security answer — one live credential per client, not N."* That is what this is.

- **Identity is the label**, `OAuth: {client_id}`, and `client_id` is a random 24-byte value issued per client registration — a shared label means the same client.
- **The new token is excluded by PATH**, which also makes the listing's lag harmless: a read that has not caught up simply leaves an older row for the *next* authorization to collect, and can never take the new one.

## 🚨 The ordering is half the fix

**Mint first, supersede after — never the reverse.** Revoking before the mint would strand the client with *no* credential at all if the mint then failed.

The test asserts the **surviving** token is the one from the second exchange, so it pins that order rather than just the count.

## Delete, not just mark revoked

This issue is *both* "one live credential" and the unbounded accumulation behind it, and a revoked row keeps accumulating. Deleting matches what #1477's expiry sweep already does with a dead credential. Self-paced (`Concat`, one at a time, the same shape as that sweep), and it can never fail the exchange — the client has a valid token by that point, so housekeeping that could not complete is a Warning and the next authorization tries again.

## What is deliberately NOT here

The **other** mint path in #1493 — the browser's one-token-per-page-load — is untouched. It is already bounded (`expiresInDays: 1`, swept by #1477), and eliminating it means letting the gRPC transport authenticate from the same-origin session cookie: a materially larger change across both React shells, and not something to fold in behind a security fix.

## Verification

- `MeshWeaver.Auth.Test` — **147/147**, including a new end-to-end case that runs register → authorize → token **twice** for one client.
- **Proven RED without the fix.** The failure output is the bug, stated exactly:

```
Expected the observable to emit a value matching the predicate within 30s … (2 item(s)).
  ApiTokenInfo { NodePath = carol/ApiToken/50429e471485, Label = OAuth: ua07k8QOn…, ExpiresAt = 8/14/2027, IsRevoked = False }
  ApiTokenInfo { NodePath = carol/ApiToken/78aa9aad49fb, Label = OAuth: ua07k8QOn…, ExpiresAt = 8/14/2027, IsRevoked = False }
```

Two live year-long credentials for one `client_id`. Green with the fix.

- `-c Release -warnaserror`: `Memex.Portal.Shared`, `MeshWeaver.Auth.Test` — 0 errors, 0 warnings.

Fixes #1493

### DIFF
--- a/memex/Memex.Portal.Shared/Authentication/OAuthConnectController.cs
+++ b/memex/Memex.Portal.Shared/Authentication/OAuthConnectController.cs
@@ -297,12 +297,26 @@ public class OAuthConnectController(
                 // days surprises users who connect once and come back months later.
                 // Refresh-token flow isn't implemented yet; until it is, default to
                 // 1 year. Bump if needed via TokenLifetime below.
+                var label = $"OAuth: {request.client_id}";
+
                 return TokenService.CreateToken(
                         userId: entry.UserId,
                         userName: entry.UserName,
                         userEmail: entry.UserEmail,
-                        label: $"OAuth: {request.client_id}",
+                        label: label,
                         expiresAt: DateTimeOffset.UtcNow.Add(TokenLifetime))
+                    // 🚨 ONE live credential per client, not N (#1493). Every authorization used to
+                    // mint a FRESH year-long token and leave the previous one live and listed, so a
+                    // reinstall, a new device, or a revoked-and-reconnected integration each left
+                    // another usable credential behind — for a year. That is what dominated the 86
+                    // ApiToken rows found on the live portal, and each of them is a key that still
+                    // opens the door.
+                    //
+                    // Ordered mint-THEN-supersede, never the reverse: if the mint fails the client
+                    // must keep the credential it already has, so revoking first could strand it
+                    // with none at all.
+                    .SelectMany(creation => SupersedePreviousTokens(entry.UserId, label, creation.Node.Path)
+                        .Select(_ => creation))
                     .Select(creation =>
                     {
                         logger.LogInformation("Issued OAuth access token for user {Email}, client {ClientId}", entry.UserEmail, request.client_id);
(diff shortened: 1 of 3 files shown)


rbuergi:
Review comment **applied** — it found a real defect, and a worse one than the issue itself.

A `delete everything with this label except mine` rule is not convergent: two concurrent exchanges for the same `(user, client_id)` each see the other's freshly-minted token in the listing and delete it, so **both clients walk away holding a credential that was removed moments later**. That is strictly worse than the accumulation this PR set out to fix.

Superseding is now ordered by `(CreatedAt, path)`: every participant deletes only strictly **below itself**, so the newest token survives whichever exchange evaluates last, and the outcome still converges on exactly one live credential.

The path tiebreak is load-bearing, not decoration — `CreatedAt` is a UTC timestamp and two exchanges can land on the same tick, where "strictly older by time" alone would let both survive.

Pinned by `Token_TwoAuthorizationsRacing_LeaveExactlyOneWORKINGCredential`, which obtains both authorization codes **before** either is exchanged so the two genuinely overlap. It asserts the property that actually matters: not that one row survives, but that **a surviving row still validates** — mutual deletion satisfies a count assertion on its way to zero.

`MeshWeaver.Auth.Test`: 148/148.

rbuergi:
CI red on shard 4 attributed — **#890**, not this diff.

`MeshWeaver.PluginCatalog.Test.IncrementalUpdateTest` failed on a NodeType compile status (`Expected value to be Ok, but found Error`), and the log says why:

```
[Error] [MeshWeaver.Graph.CompileWatcher] NodeType 'Widget/Thing' PARKED after compile failure
  Error: Failed to load the compiled assembly — Could not load type 'Thing' from assembly
  'DynamicNode_Widget_Thing, …' because THE FORMAT IS INVALID.
```

The compile succeeded and the **emitted assembly would not load** — a corrupt PE image for a trivial dynamic NodeType. That is #890's process-poisoning condition in its second manifestation (the first being the `NullReferenceException` inside `Cci.MetadataWriter`); posted there with the reasoning: https://github.com/Systemorph/MeshWeaver/issues/890#issuecomment-5296172431

Attribution, rather than assumption:
- this diff is `OAuthConnectController.cs`, its test file, and a What's New markdown — there is no path from an OAuth controller to a NodeType compile in `MeshWeaver.PluginCatalog.Test`;
- `MeshWeaver.Auth.Test` (the only suite this change can reach) is **148/148** locally;
- and `flake-repro` in `rate` mode over 40 iterations of that PluginCatalog test **on `main`** shows **0 failures** — i.e. it is not a standing per-test flake either, it is the shard-wide condition #890 describes ("one Roslyn Emit poisons the whole process — every LATER compile"), which by construction does not reproduce in an isolated filtered loop.

Re-running the failed shard on that basis — an environment-level fault with the same input is the sanctioned re-run case, and the reasoning is recorded rather than assumed.

PR #1614, opened 2026-08-14
fix: seal an impersonation scope into the operation it was opened for (#1444)

### LINKED ISSUE #1444: Package install writes as system-security outside any impersonation scope — while _UserActivity in the same flow attributes correctly
Observed while diagnosing #1371, in education CI run 31704948831 (shard 2, `mesh-diagnostics/portal.log`, 14:09:55–57Z).

**Every write the package install makes is logged `by system-security`** — including the synthesized home root and the completion manifest — while `_UserActivity` written moments later in the same flow is correctly attributed to the acting user (`by e2e-get-agenticengineering`).

None of those writes is inside an `ImpersonateAsSystem` scope. And `CopyRoot`'s own comment claims the opposite of what the log shows.

## Why this matters

`AGENTS.md` is explicit: **never write as hub.** Every framework write primitive carries the caller's `AccessContext` across `.Subscribe()` boundaries precisely so an operation runs under the calling user's identity. Writing as `system-security` grants `Permission.All` unconditionally — so an install performed *for* a user is being recorded, and authorised, as the platform itself.

Two distinct consequences, and they want separating:

1. **Attribution.** The audit trail says the platform created these nodes, not the learner who asked for them. `_UserActivity` disagreeing with the node writes in the same flow is the tell.
2. **Authorisation.** If any of these writes would have been *refused* under the acting user's permissions, the system identity silently succeeded where the user should have failed. That is the more serious half and should be established first — it is the difference between a logging defect and a permission bypass.

The 2026-05-21 removal of the "silently stamp hub-self as principal" fallback exists for exactly this class; a path that reaches `system-security` without an explicit impersonation scope suggests something is re-acquiring it.

## What to establish

- **Where the identity is acquired.** Is an `ImpersonateAsSystem` scope open further up (legitimately, for infrastructure) and never closed before the content writes? Or is the context lost across a scheduler hop and re-defaulted?
- **Whether it is load-bearing.** `StandardPacks.InstallIfMissing` deliberately uses a system write for "install into a home that is not the caller's", and Plugins#452 established that impersonating the buyer is wrong there because the webhook holds an id, not an identity. So *some* system writes on this path may be correct by design — the question is whether *these* are, and whether the comment or the code is the thing that is wrong.
- **Whether `_UserActivity`'s correct attribution is the anomaly or the norm** — it is the one write here that behaves as documented, which is itself informative.

## Not the fix

Do not "fix" this by wrapping the writes in an explicit `ImpersonateAsSystem` to make the code match the log. That would settle the contradiction in favour of the behaviour without establishing whether the behaviour is right.

## Provenance

Found while fixing #1371 (Plugins#466 — parent-relative embeds in copied install roots). Unrelated to that fix, unpinned, and deliberately not addressed there.

rbuergi:
Closes #1444.

## What was established first

The issue asks to establish where the identity is acquired, whether it is load-bearing, and whether `_UserActivity`'s correct attribution is the anomaly or the norm — **before** changing anything. All three, in order:

**Where.** Not in an `ImpersonateAsSystem` scope anyone wrote around the writes — that observation in the issue is correct. The idiom `Observable.Using(access.ImpersonateAsSystem, _ => work)` is correct for the work itself and is **not self-contained**: Rx forwards `OnNext` to the subscriber **before** `Using` disposes its resource, so everything the subscriber composes on that emission — the next phase, a write, a permission check — is built while the impersonation is still open.

**Why it does not stop at one hop.** The mesh write primitives eager-capture `AccessService.Context` when they are CALLED and re-stamp it around their own emissions (`CarryAccessContext`). So a leaked System identity is not inherited once — it is **re-acquired at every hop after it**. That is a whole install chain, from one scope.

**`_UserActivity` is the norm, not the anomaly.** It is the one write on that path posted directly from the request thread rather than composed on an emission. It behaved as documented because nothing had leaked into it. Every other write was downstream of a system scope.

The first test in this PR pins the raw Rx ordering all of that rests on, so if a future Rx ever disposes before forwarding, that fact is learned from a red test rather than from a production log.

## Why this is fixed at the root and not in the log

The issue separates attribution from authorization and asks for the second to be established first. It is not hypothetical, and the framework already carries the scar:

> `AccessControlPipeline.HandleGetPermission` — *"That AsyncLocal at handler-entry holds `system-security` from SecurityService's bootstrap-time `ImpersonateAsSystem` scope (it leaks past the using-block because the bootstrap action-block thread captured the context at construction). Trusting it returned `Permission.All` for every caller — including anonymous deliveries."*

That call site defends **itself**, by resolving the identity explicitly instead of reading the ambient. This PR closes the class at the source so the next call site does not have to know Rx's disposal order to be safe.

## The fix

`ImpersonationScopeExtensions` — `RunAsSystem` / `RunAsHub` / `ContainIdentity`. The scope is entered at Subscribe (so the cold work is covered — the `Observable.Using`, never `Defer`+`using`, rule) and every notification reaches the subscriber under the **subscriber's own** identity.

Restore semantics are not re-implemented: it reuses `CarryAccessContext`'s per-callback `RestoringObserver`, so the three ways an identity attaches to a pipeline cannot drift apart.

```csharp
// ❌ the write is CONSTRUCTED while the scope is still open → lands as system-security
Observable.Using(access.ImpersonateAsSystem, _ => ReadGatedSource())
    .SelectMany(rows => meshService.CreateNodes(Plan(rows)));

// ✅ the scope covers the read and stops at its boundary → lands as the CALLER
access.RunAsSystem(ReadGatedSource)
    .SelectMany(rows => meshService.CreateNodes(Plan(rows)));
```

🚨 **It never invents an identity.** What is restored is exactly what was ambient at Subscribe: a real user for a user-driven flow, `system-security` for a genuinely system-initiated one, and **nothing at all** when the subscriber had no identity. That last case is what makes adoption safe at an existing call site — a background worker that never had a user is not fail-closed by it. The only behaviour that changes is the one the issue describes.

## Adopted by the eight copies of the same helper

`AsSystem` had been hand-written **eight times** across `src/` — Graph ×3, Mesh.Contract, GitSync ×2, InstanceSync, PluginCatalog — several carrying a comment saying they mirror one of the others. They now delegate to the one sealed primitive.

## What this PR deliberately does NOT claim

The reported install path lives in the **Plugins repo** and is already fixed there (`Store/Plugin/Source/SystemIdentity.cs`, which cites this issue). Its analysis settles the authorization question for that path: nothing landed anywhere the learner could not have written — every path was under their own partition root, where onboarding had just granted them Admin — so the realised damage is an audit trail saying the platform did what a learner asked for. The one permission check on that path resolved its subject from the same ambient context, and no shipped path handed it a user-chosen destination; that is a fact about Rx's disposal order rather than a defence anyone designed, which is why it is fixed anyway.

That plugin-local helper is what this framework primitive should have been all along. This PR is the framework half.

## Verification

| | |
|---|---|
| `SystemScopeDoesNotEscapeTest` | **8/8** — including `leaked == system-security` vs `sealed == alice` on the same composed write |
| `MeshWeaver.Messaging.Hub.Test` | 251/251 |
| `MeshWeaver.AccessControl.Test` | 29/29 |
| `MeshWeaver.InstanceSync.Test` | 26/26 |
| `MeshWeaver.GitSync.Test` | 163/163 |
| `MeshWeaver.Graph.Test` | 1127/1127 |

`-c Release -warnaserror` clean for all six touched projects.

### DIFF
--- a/src/MeshWeaver.Documentation/Data/Architecture/AccessContextPropagation.md
+++ b/src/MeshWeaver.Documentation/Data/Architecture/AccessContextPropagation.md
@@ -370,6 +370,38 @@ A `SynchronizationStream.OnNext` that unconditionally stamps `ImpersonateAsHub(H
 
 > Every `ImpersonateAsSystem` callsite is a deliberate choice to bypass all access checks. Grep for them; review each. If a narrower identity would suffice, use that instead.
 
+### 🚨 An impersonation scope must not ESCAPE the operation it was opened for
+
+`accessService.RunAsSystem(work)` — **not** a hand-written `Observable.Using(access.ImpersonateAsSystem, _ => work)` — wherever the scoped observable is **returned to a caller**.
+
+The hand-written idiom is correct for the work itself and leaks everywhere after it:
+
+```csharp
+// ❌ the scope escapes: Rx forwards OnNext to the subscriber BEFORE Using disposes its resource,
+//    so the write below is CONSTRUCTED while the impersonation is still open — and the write
+//    primitives eager-capture AccessService.Context when they are CALLED.
+Observable.Using(access.ImpersonateAsSystem, _ => ReadGatedSource())
+    .SelectMany(rows => meshService.CreateNodes(Plan(rows)));   // lands as `system-security`
+
+// ✅ the scope covers the read and stops at its boundary
+access.RunAsSystem(ReadGatedSource)
+    .SelectMany(rows => meshService.CreateNodes(Plan(rows)));   // lands as the CALLER
+```
+
+Because those primitives also **re-stamp** the captured identity around their own emissions (`CarryAccessContext` — "MessageHub sets, framework primitive preserves"), a leaked System identity is not merely inherited one hop — it is **re-acquired at every hop after it**. That is how a package install wrote its home root, five copy batches and its manifest as `system-security` while `_UserActivity` — the one write on that path posted directly from the request thread rather than composed on an emission — attributed to the real user (#1444).
+
+**It is an authorization concern, not only an audit one.** `AccessControlPipeline.HandleGetPermission` carries the scar: `SecurityService`'s bootstrap-time system scope leaked past its using-block onto the action-block thread, and trusting the ambient there returned `Permission.All` for **every** caller, anonymous included. That call site defends itself by resolving the identity explicitly; `RunAsSystem` fixes the class at the source so the next one does not have to know Rx's disposal order.
+
+| API | Use it for |
+|---|---|
+| `access.RunAsSystem(work)` | A system read/write whose result a caller composes on. Enters the scope at Subscribe (so the cold work is covered), delivers every notification under the **subscriber's** identity. |
+| `access.RunAsHub(hub, work)` | The same seal for a hub identity. |
+| `source.ContainIdentity(access)` | Sealing an **already-composed** chain — several system calls where the caller builds on the LAST one's emission. |
+
+🚨 **It never invents an identity.** What is restored is exactly what was ambient at Subscribe: a real user for a user-driven flow, `system-security` for a genuinely system-initiated one, and **nothing at all** when the subscriber had no identity — so a background worker that never had a user is not fail-closed by adopting it. The only behaviour that changes is the defect: a caller who *had* an identity, silently continuing as System.
+
+Pinned by `SystemScopeDoesNotEscapeTest` (test/MeshWeaver.Messaging.Hub.Test), whose first case asserts the raw Rx ordering every other case rests on.
+
 ### Implementation: define + grant + test
 
 For each sanctioned identity, you need all three:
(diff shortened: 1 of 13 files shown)


rbuergi:
CI red on `InstanceSyncPushTest.A_pending_content_change_pushes_a_DELETE_when_the_local_node_is_being_deleted`, which this change does not touch. Investigated before re-running, per AGENTS.md:

- **It fails on `main`.** Sweeping the last 30 failed runs by failing test (posted on #1384): this test failed **4 times today** — runs 31793987399 (**main**), 31823380161 (**main**), 31824306190 (`fix/clients-sdk`), and this one. Two of the four are main and neither branch run touches InstanceSync's path.
- **Filed as #1616** with the RCA direction: the test's own output shows `Dropping DataChangedEvent … the target stream is gone` at **5.5 s**, and the wait then burns the remaining 25 s of its 30 s budget on something already discarded. The bound is not the defect.
- **A/B measured**, `flake-repro.yml` rate mode, 40 iterations each: **0/40 on main, 0/40 on this branch** — an isolated loop does not reproduce it at all, so the loop cannot attribute it either way. The population data is what attributes it, and it points at main.

Re-running the failed shard on that basis — not because "it might be a flake", but because it is a measured one with an issue and an RCA direction of its own.

Issue #1655, opened 2026-08-16
DESIGN: endpoint contributions on the module lane — the sixth MeshNodeProviderAttribute hook


rbuergi:
**The wall this removes** (documented in [UiExtensibility](src/MeshWeaver.Documentation/Data/Architecture/UiExtensibility.md) → "What cannot be extended today"): there is no endpoint-contribution seam, so every HTTP surface belonging to a MODULE'S concern stays compiled in the portal, resolving its module service optionally and answering an actionable 503 when the module is delisted. The live inventory of exactly this shape:

- `/api/speech/transcribe` — compiled in Portal.Shared; resolves `ISpeechTranscriber` optionally (MeshWeaver.Speech module).
- `/api/log-incidents` — compiled; resolves the `ILogIncidentIngest` contract seam optionally (MeshWeaver.Observability module).
- `/api/og/{**path}` — compiled, while the OgCard AREA rides the module (#1643): the module's own HTTP half could not follow it.
- The course-asset endpoints — the Wave-2 falsification that kept education runtime code compiled.
- The `Social/*Endpoints` families — candidates only if their concerns ever modularize.

**Proposal**: a sixth hook on `MeshNodeProviderAttribute`:

```csharp
public virtual IEnumerable<Action<IEndpointRouteBuilder>> EndpointConfigurations => [];
```

The host applies every installed module's endpoint configurations at its endpoint-mapping stage (after auth middleware, before the Blazor catch-all — modules load at boot via `InstallAssemblies`, well before `MapEndpoints`, so ordering is structurally safe). Delisting a module then removes its routes wholesale — a 404 instead of today's compiled 503 — and the compiled optional-service endpoints migrate INTO their modules.

## Design questions for review (no code before sign-off)

1. **Contract width**: raw `IEndpointRouteBuilder` (maximum power, host-framework coupling) vs a narrower record (pattern + verb + handler) the host maps itself. Raw is simpler and modules are TRUSTED CODE (unlike `UiContribution` data — the security model difference is load-bearing and must be documented); the narrow form would ease a future non-ASP.NET host but has no consumer today.
2. **Auth defaults**: contributed endpoints should default to the authenticated policy with explicit opt-out (`AllowAnonymous` stays a deliberate per-route act — the `/api/og` fetcher and webhook inboxes need it), so a module cannot accidentally publish an open route.
[...]

PR #1663, opened 2026-08-16
feat(hosting): the endpoint-contribution hook — modules map HTTP routes (#1655)

rbuergi:
Implements #1655 (maintainer-approved). New `MeshWeaver.Hosting.AspNetCore` with the assembly-level `MeshEndpointProviderAttribute` + `app.MapMeshModuleEndpoints()` wired into `StartMemexApplication` beside the other `Map*` calls.

- **Layering**: separate attribute, not a `MeshNodeProviderAttribute` hook — the mesh contract never references ASP.NET, and endpoints apply at endpoint-mapping time (after auth middleware), not mesh build.
- **Auth by default**: contributions map inside a `RequireAuthorization()` group; a route is anonymous only via explicit per-route `AllowAnonymous()`.
- **Loud collisions**: duplicate (verb, pattern) across the WHOLE endpoint table refuses the app at `ApplicationStarted` naming both endpoints (module-vs-module AND module-vs-platform); same pattern on different verbs stays legitimate. ASP.NET alone only surfaces ambiguity at request time.
- **Security model documented**: a module is trusted code the deployment listed — unlike `UiContribution` mesh data, which still cannot contribute routes.
- Docs: Modules.md + UiExtensibility (the 'HTTP endpoints' wall retired). DocumentationLinkIntegrityTest green.

Verification: Hosting.AspNetCore / Portal.Shared / Portal.Monolith clean `-c Release -warnaserror`; `ModuleEndpointContributionTest` 2/2 (discovery + auth default + anonymous opt-out; collision predicate with both parties named). First consumers land as separate PRs: Social (#59), the python/node gRPC link (#60), then speech/log-incidents/og.

No What's New entry: platform capability, no user-visible change.

### DIFF
--- a/MeshWeaver.slnx
+++ b/MeshWeaver.slnx
@@ -81,6 +81,7 @@
     <Project Path="src/MeshWeaver.Fixture/MeshWeaver.Fixture.csproj" />
     <Project Path="src/MeshWeaver.Reactive.Assertions/MeshWeaver.Reactive.Assertions.csproj" />
     <Project Path="src/MeshWeaver.Maps/MeshWeaver.Maps.csproj" />
+    <Project Path="src/MeshWeaver.Hosting.AspNetCore/MeshWeaver.Hosting.AspNetCore.csproj" />
     <Project Path="src/MeshWeaver.OgCard/MeshWeaver.OgCard.csproj" />
     <Project Path="src/MeshWeaver.Graph/MeshWeaver.Graph.csproj" />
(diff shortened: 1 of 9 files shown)


rbuergi:
Review addressed: the collision-refusal comment now states the actual intent (log + StopApplication + throw — belt and braces). The `using MeshWeaver.Hosting.Monolith.Test;` in the test file is kept deliberately: the assembly-level `[assembly: TestModuleEndpoints]` sits OUTSIDE the namespace declaration, so resolving the attribute type (declared inside the namespace) requires it — removing it breaks the build.

Issue #1660, opened 2026-08-16
NodeType assemblies are built in CI and LOADED at runtime — no runtime Roslyn, no warm-up window (maintainer-directed)


rbuergi:
**Direction from the maintainer (2026-08-16, during the Store outage): "we should not dynamically build at all … should be building on CI and then re-use dll … it should not even be warmer, it should be literally built on CI … and just loading into assembly cache."** This issue captures the incident that proved the point, the root causes, and the workstreams to get there.

## The incident (memex prod, 2026-08-16 morning)
- `/Store` served the compile fallback (*"NodeType 'Store/Catalog' build did not settle within 60s"*) for ~3 hours; `Store/Coupon`, `Store/BillingProfile`, `Store/Order`, `Store/Install`, `Store/Provision`, and `Publish/Slide` hubs answered reads with no verdict; recovery required manual per-type recycle+compile and a pod restart.
- Boot log 05:31Z: `DynamicTypePreWarmer: 85 of 86 dynamic NodeType(s) need building … framework=beb17498 baked=1 pending=85 frameworkstale=83`. The batch bake then **successfully compiled every type, Store family included** (log lines per type, 0.2–3.4s each) — but the SERVED state stayed pre-roll: the query index held `Store/Catalog` v1829 while the node hub served v1828 with Aug-14 timestamps, and several per-type hubs wedged (`SubscribeRequest → no response within 00:01:00; target hub not found`).
- Restart at 08:52Z booted on a DIFFERENT image — CD had re-pinned `ci.3894 → ci.3903` from the merge backlog — and the new boot read `framework=4dfb942a baked=0 pending=86, 0 already on the share`: **every image roll mints a new framework MVID, invalidates all 86 types, empties the share key-space, and forces a full runtime re-bake with a public fallback window.** Three rolls in ~12h kept prod cycling through that window.

## Root causes
1. **Runtime baking is the prod path.** Every deploy recompiles the world in-process; the fallback window is roll-frequency × bake-duration, and users live inside it.
2. **Framework identity churns per CI build.** The per-run version stamp is a compile input, so even near-identical trees produce a new Graph MVID per image — prebuilts/bundles/share entries can never be produced AHEAD of the image that invalidates them.
[...]

PR #1696, opened 2026-08-16
feat(#1660 WS3): API-surface framework identity — rebuild only when we need to

rbuergi:
## #1660 WS3 refinement — rebuild only when we need to: the API-surface identity

Rides on #1687 (commit-deterministic identity). The commit key made the CI bake adoptable but
OVER-invalidated: every platform merge — internal-only included — minted a new identity and
rebaked all content (measured: `pending=82` on the routine ci.3979 roll). Content must rebake only
when the API SURFACE it compiles against changed — a breaking change.

### The identity (one property, new resolution — `FrameworkBuildIdentity`)

1. **API-surface identity `s<hash>`** — hosts that ship a `meshweaver-surface.manifest`
   (portals + the CI bake host, opt-in `<MeshWeaverSurfaceManifest>true</>`): the manifest records,
   per MeshWeaver compile reference, the SHA-256 of its **reference assembly** — the compiler's own
   surface oracle (byte-stable under body-only/private edits; flips on any public/protected — and,
   for IVT'd assemblies, internal — surface change; a hand-rolled API walk would miss the IVT
   subtlety, which is why the ref-asm is the oracle). The identity hashes the sorted
   `(name, surface-id)` pairs of a **canonical content-surface set** — the bake gate's own
   framework closure (36 assemblies; content referencing anything outside it cannot pass the gate),
   which is what makes the hash EQUAL across the bake host and the portals despite their different
   TPAs (the portal carries 26 more MeshWeaver assemblies — Blazor/Orleans/hosting — deliberately
   outside the content-staleness key). 🚨 **Generator exception**: `MeshWeaver.Graph` contributes
   its FULL implementation MVID — its code (DynamicMeshNodeAttributeGenerator, the skeleton/
   aggregation/`@@`-include/NuGet-directive pipeline in MeshNodeCompilationService +
   CodeQueryResolver) shapes the GENERATED INPUT of every NodeType compile, so a body-only change
   there must rebake without any API change. (Swept: the only other generator, SkeletonGenerator,
   lives in MeshWeaver.Plugin.Build — plugin build tooling, not in the portal compile path.)
2. **Commit identity `g<sha>`** — fallback for manifest-less CI processes (test hosts), and
   retained everywhere as DIAGNOSTIC PROVENANCE (logged at warm-up start:
   `framework identity {id}, build {g<sha>}`).
3. **Graph MVID** — local builds, unchanged.

All consumers follow the ONE property exactly as wired in #1687 (store tag, `HasUsableBuild` /
`CompiledFrameworkVersion`, bake report, build-claim fingerprint, seeder gate, artifact key);
`AssemblyCacheGenerations.TagOf` learns the `s`+7-hex tag shape.

### Build plumbing

- `Directory.Build.props`: `_MeshWeaverWriteSurfaceManifest` (built-in `GetFileHash` over
  `@(ReferencePathWithRefAssemblies)` filtered to `MeshWeaver.*`) + publish hookup. 🚨 The
  multi-arch container publish's `ProduceReferenceAssembly=false` optimization is REMOVED — ref
  assemblies are load-bearing now (no ref-asms ⇒ impl hashes ⇒ forked identity ⇒ every bake
  declines); the PR #552 race stays killed by the per-leg bin/obj isolation.
- `mw-plugin-test --print-framework-identity` — one-line diagnostic (identity + provenance); the
  proof scripts drive the REAL resolution path through it.
- Publish skip: `publish-bake-bundles.sh` skips-with-notice when
  `prebuilt-bundles/<identity>/<source>/` already holds files — an internal-only merge is a no-op
  publish ("rebuild only when we need to" applies to the publish too).
- `notify-dependents` payload drops the (now wrong) `identity` field — receivers resolve their
  own.

### Proof (measured through the real binary, `mw-plugin-test --print-framework-identity`)

- Ref-asm oracle: byte-identical under method-body-only edits, changed by a public const add
  (sha256 of `obj/ref/MeshWeaver.Data.dll` across from-clean rebuilds); internal-member adds flip
  it for IVT'd assemblies — correct conservatism, documented.
- Live identity (all via the real binary): baseline `s9f38b6456177317dd2c6389469811f6b`;
  - internal body-only edit in MeshWeaver.Data ⇒ identity UNCHANGED
    (`s5cfe9fff…` == `s5cfe9fff…`, and reproducible across independent runs);
  - public member added to MeshWeaver.Layout ⇒ identity CHANGED (`s6adbb981…`);
  - body-only edit in MeshWeaver.Graph, ref-asm CONFIRMED byte-stable across the edit ⇒ identity
    CHANGED (`se1518a33…` → `sfbfd8b4d…`) — the emitter full-MVID rule, isolated;
  - revert ⇒ baseline `s9f38b645…` restored — **after merging #1691/#1694/#1695 mid-proof**:
    three real merged commits (CD workflow + ContentCollections internals) moved the identity
    NOT AT ALL, live evidence that routine merges now cause zero rebakes.
- Unit pins: surface computation (stability, listed-assembly change, host-extra invisibility,
  emitter rule, absence marker), manifest parsing, canonical-list invariants, a closure pin test
  that recomputes the tester's csproj closure and fails naming the drift, the process resolution
  chain, tag attributability.
- Full solution `-c Release -warnaserror --no-incremental` green; affected test projects green.

### Ops notes

- First roll after this merges bakes once under the new `s<hash>` identity (a one-time full bake —
  the key scheme changed); after that, internal-only merges keep `pending≈0` end to end and skip
  the publish.
- `BAKE_PUBLISH_TARGETS` + storage role are already provisioned (see #1687).

### DIFF
--- a/.github/scripts/publish-bake-bundles.sh
+++ b/.github/scripts/publish-bake-bundles.sh
@@ -52,6 +52,13 @@ if [ "${#BUNDLES[@]}" -eq 0 ]; then
   exit 1
 fi
 
+# The completeness sentinel (must match ShippedPrebuiltBundles.CompletionSentinelFileName): its
+# PRESENCE means "every bundle of this publication landed", because it is uploaded strictly LAST.
+# Its content lists the bundle set, so the reader can detect a listed-but-missing bundle too.
+SENTINEL="_complete"
+SENTINEL_LOCAL=$(mktemp)
+for zip in "${BUNDLES[@]}"; do basename "$zip"; done | sort > "$SENTINEL_LOCAL"
+
 publish_one_target() { # <account> <share> <dest-dir>
   local account="$1" share="$2" dest="$3" path="" part
   # az storage directory create is not recursive and errors on an existing directory on some CLI
@@ -74,6 +81,12 @@ publish_one_target() { # <account> <share> <dest-dir>
       --auth-mode login --backup-intent --only-show-errors > /dev/null
     echo "published: $account/$share/$dest/$(basename "$zip")"
   done
+  # LAST write — the atomic completeness marker. Anything that dies before this line leaves the
+  # directory sentinel-less: unreadable to portals, re-published wholesale by the next run.
+  az storage file upload --account-name "$account" --share-name "$share" \
+    --path "$dest/$SENTINEL" --source "$SENTINEL_LOCAL" \
+    --auth-mode login --backup-intent --only-show-errors > /dev/null
+  echo "sealed: $account/$share/$dest/$SENTINEL (${#BUNDLES[@]} bundle(s))"
 }
 
 for target in $BAKE_PUBLISH_TARGETS; do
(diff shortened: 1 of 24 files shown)


rbuergi:
Both review findings were real and are fixed in 3fd2b656e:

1. **Partial-publish freeze** — the publisher now writes a `_complete` sentinel (listing the bundle set) strictly LAST; the skip keys on the sentinel only, and the read side honours the same contract: `SeedPublishedRoot` seeds only sealed source directories and only the bundles the sentinel lists — unsealed/torn publications are refused loudly (pinned by `PublishedRoot_RefusesUnsealedAndTornPublications`).
2. **Unguarded manifest read** — identity resolution is now the pure `ResolveProcessIdentityWithDiagnostics`: any manifest failure degrades to the stamp/MVID fallback with the warning cached beside the identity and logged by the pre-warmer where it announces the identity (pinned by `ProcessIdentity_DegradesToTheFallback_*`).

PR #1723, opened 2026-08-17
docs: satellite bake-publish credentials are provisioned; dispatch still dormant

rbuergi:
Docs-only update recording verified 2026-08-17 state across the bake/publish lane.

### What is now stated as fact

1. **Satellite bake-publish credentials are PROVISIONED.** Managed identity `github-actions-bake` (RG `memex-aks-rg`) holds *Storage File Data Privileged Contributor* on the portals' storage account, with **8 federated credentials — four satellite repos × two subject formats** — and `AZURE_CLIENT_ID` / `AZURE_TENANT_ID` / `AZURE_SUBSCRIPTION_ID` set on all four repos. **A red publish-bake was designed debt until 2026-08-17 and is a real failure after it.**
2. **Register BOTH OIDC subject formats per repo.** GitHub's immutable subjects embed numeric ids and the format **varies per repo inside the org**, so normalising to one broke a satellite with `AADSTS700213`. Adds the provisioning commands, the list command that shows the expected 8 rows, and the copy-the-presented-subject remedy.
3. **The dispatch cascade is still dormant** (`BAKE_SUBSCRIBER_REPOS` / `DEPENDENT_DISPATCH_TOKEN`) — satellites re-bake on their own main pushes; a release-triggered rebuild can be hand-fired via `repository_dispatch`.
4. **Reusable-workflow adoption (#1721) is per job and in progress** — SocialMedia + Plugins merged and green; Reinsurance and Education in flight; Manufacturing deferred. Also: unpinned images are an explicit `allow-unpinned` opt-in, and adoption renames required contexts to `<caller job> / <name>` (SocialMedia's are recorded as the concrete example).
5. **The adopt path is measured in prod for satellite content** — memex on `3.0.0-rc4.ci.4049` adopted 68 prebuilt assemblies from 31 sealed bundles in 18.9 s and compiled zero healthy types (`compiled=0`, `alreadyBaked=84`), against 80 compiles / 64.8 s before the satellite bakes existed.
6. **#1725 is flagged wherever the docs claim boot adopts the platform's OWN bake** — CI bakes platform content from build output while the pod resolves its identity inside the image, so no pod can adopt it today. Satellites are unaffected (they bake inside the image).

### Files

- `ContinuousDeliveryContract.md` — notify-dependents provisioning state, the OIDC subject-format section, the #1725 caveat, the prod measurement.
- `CiContentBake.md` — node-repo section only (adoption state, context rename, credentials, measurement, #1725). **The delivery-lane section is untouched** — it is owned by a concurrent change.
- `OnboardingNewEnvironment.md` — a new env's Azure Files share must reach `BAKE_PUBLISH_TARGETS`, else its pods compile every shipped NodeType at boot.
- `AGENTS.md` — satellite CI = thin callers of this repo's reusable workflows.

### Verification

`MeshWeaver.Documentation.Test` built `-c Release -warnaserror` (0 errors, 0 warnings) and `DocumentationLinkIntegrityTest` passed after each edit round. Copilot review addressed (bold spans kept on one line) and threads resolved.

No What's New entry — internal docs change, no user-visible product effect.

The first run's shard-4 red is the open **#890** (Roslyn emit NRE, canary `BELOW-ROSLYN`), not this diff — see the attribution comment below.

Companion PR: Systemorph/MeshWeaver.Plugins#504.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -328,6 +328,17 @@ summarize step is the real gate; a GitHub-API 429 must not fail the run) and the
 push/prune (losing a marker costs a redundant run, never correctness). The test is *what does a
 failure here hide?* — nothing, for a reporter; everything, for a gate.
 
+### Satellite CI = thin callers of THIS repo's reusable workflows
+
+**Never hand-roll (or copy-paste) a node repo's CI.** The shared jobs live here as `workflow_call`
+workflows — `.github/workflows/node-repo-{validate,compile-check,gate,tag-modules,publish-bake}.yml`
+— and MeshWeaver.Plugins / .Education / .Reinsurance / .SocialMedia call them, keeping only
+repo-specific policy (digest pin, gating, `repository_dispatch` receiver, their own `scripts/`).
+Adopting one renames that repo's required-status-check contexts to `<caller job> / <name>` — do it
+in the same change. Full contract: [CiContentBake.md](src/MeshWeaver.Documentation/Data/Architecture/CiContentBake.md)
+and [ContinuousDeliveryContract.md](src/MeshWeaver.Documentation/Data/Architecture/ContinuousDeliveryContract.md)
+(which also carries the GitHub OIDC subject-format rule: register BOTH subject formats per repo).
+
 ## 🚨 Postgres: One Schema Per Partition
 
 **`public.mesh_nodes` is empty by design.** Data lives in per-partition schemas (`acme.mesh_nodes`, `rbuergi.mesh_nodes`, etc.).
--- a/src/MeshWeaver.Documentation/Data/Architecture/CiContentBake.md
+++ b/src/MeshWeaver.Documentation/Data/Architecture/CiContentBake.md
@@ -157,14 +157,30 @@ The design rules the extraction preserves:
   the per-repo vendored copies. This repo is public, so private satellites call the workflows
   and read the script with their default token.
 - **Repo-specific policy stays in the caller**: the digest pin (`MW_IMAGE_DIGEST`) and its bump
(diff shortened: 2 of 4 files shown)


rbuergi:
**CI attribution for the first run (3da23fdeb → run 32019897724, shard 4 red).**

Failing test: `MeshWeaver.PluginCatalog.Test.CodePackageInstallTest.CodePackage_InstallsNodeTypeAndSource_CompilesLive`. Not caused by this diff — this PR is markdown only, and the failure is the signature of the open **#890** (`NullReferenceException` in `Microsoft.Cci.MetadataWriter.GetConsolidatedTypeParameters` during Roslyn `Emit`).

The evidence is the product's own canary (`EmitPipeline`, built for #890): it re-emitted a trivial nested-generic compilation against both the shared reference set **and** a pristine one, both threw, and it therefore classified the fault `canary=BELOW-ROSLYN` — *"a trivial compilation with BRAND-NEW references and freshly parsed source cannot emit either ⇒ nothing about the reference set explains this; the broken state is below Roslyn (CLR heap / JIT / GC)"*.

Cross-check: shard 4 executed and passed on the exact tree this branch is based on (PR #1721's run 32016868555, all six shards green); main's own run for e0408a053 reused that green, so the base tree is proven. Two files under `src/MeshWeaver.Documentation/Data/Architecture/` cannot reach a CLR-level fault inside Roslyn's metadata writer.

No blind re-run was issued — the new run is for genuinely new content (the fold-in commit). If the same test fails again on this new tree, that is a reproducible signal to escalate on #890 rather than to retry.

Issue #1616, opened 2026-08-14
InstanceSyncPushTest DELETE-push test times out on main and unrelated branches — the change event is dropped at 5.5s, the wait gives up at 30s


rbuergi:
`MeshWeaver.InstanceSync.Test.InstanceSyncPushTest.A_pending_content_change_pushes_a_DELETE_when_the_local_node_is_being_deleted` fails on **main** and on unrelated branches, so every PR it lands on has to spend a fresh investigation proving it is not theirs. Second-largest bucket in the last-30-failed-runs sweep on #1384.

## Occurrences (2026-08-14 alone)

| run | branch |
|---|---|
| 31793987399 | **main** |
| 31823380161 | **main** |
| 31824306190 | fix/clients-sdk |
| 31834879042 | fix/1444-install-attribution |

Two of the four are main, and neither of the two branch runs touches InstanceSync's path — which is what makes this an attribution tax rather than a signal.

## The failure

```
System.TimeoutException : The operation has timed out.
  at InstanceSyncTestBase.WaitForRemote(Func`2 predicate, Nullable`1 timeout)   InstanceSyncTestBase.cs:98
  at InstanceSyncPushTest.A_pending_content_change_pushes_a_DELETE_when_the_local_node_is_being_deleted()   InstanceSyncPushTest.cs:136
```

The test's own output is the interesting part — the give-up is at 30 s, but the thing it is waiting for was **dropped at 5.5 s**:

```
[19:59:48.387] === TEST START ===
[19:59:48.827] [Warning] InstanceSyncWorker  theta/_Sync/partner: remote unreachable
               (Connection refused (fake remote is unreachable)) — accumulating; next probe in 00:00:00.1
[19:59:48.934] [Warning] InstanceSyncWorker  … next probe in 00:00:00.2
[19:59:53.902] [Warning] MeshWeaver.Data.DataExtensions
               Dropping DataChangedEvent for stream FPJp7O20i0exSl4PyXP-Qw on hub cache/J9hKUio…:
               no synchronization hub found on this hub or any parent — the target stream is gone
               (disposed circuit, released read stream, or never-created sync hub).
               Sender: theta/_…
[20:00:18.995] === TEST FAILED: The operation has timed out.
```

So the push does happen; its change event is delivered to a stream whose sync hub is **gone**, and the remaining 25 seconds are spent waiting for something already discarded. 🚨 A larger timeout buys nothing here and would make the drop invisible — the bound is not the defect.

[...]

PR #1731, opened 2026-08-17
docs: which routes ride a module, and which stay in the host

rbuergi:
A rule the codebase had already decided twice without writing down. Falsification output from Tier 1 item 3 — see below.

## The gap

The endpoint hook (#1655/#1663) made two statements look identical: *"a module's feature has HTTP routes"* and *"those routes belong in the module"*. They aren't the same, and we already have worked examples on both sides:

| Rides the module | Stays in the host |
|---|---|
| LinkedIn connect/publish/page-sync callbacks | `POST /api/log-incidents` (Observability) |
| the `meshweaver.v1.Mesh` gRPC service | `POST /api/speech/transcribe` (Speech) |

The right-hand column is deliberate — `Memex.Portal.Shared.csproj` states it as intent for the Observability seam ("resolves `ILogIncidentIngest` optionally and answers an actionable 503 when the module is not listed") — but it was recorded only as a comment on one project file, so the next person scoping a module extraction reads the hook's docs and concludes the opposite.

## The rule

The dividing question is **whose API is it**: the module's own protocol surface, or the portal's client API that happens to delegate to an optional engine.

Both failure modes of getting it wrong are documented, because both pass CI:

1. **The caller loses the diagnosis** — a 503 naming the missing module becomes an indistinguishable 404.
2. **The route loses the host's authorization policy** — the sharp one. The hook's group applies the **default** policy; a route needing the portal's Bearer-only `McpAuth` (whose challenge forwarding is what answers `401 + WWW-Authenticate` instead of `302` to an HTML login) would have to name a host-registered policy by string across an assembly boundary, throwing at request time in any host that never registered it. Symptom in production: "the mobile app logs me out."

## Provenance

Written while scoping — and then **abandoning** — a move of `SpeechEndpoints` onto the `MeshWeaver.Speech` module. The move looked obviously right from the outside and is wrong; this page is what stops it being re-proposed.

Docs only. `MeshWeaver.Documentation.Test` 21/21 green (link integrity included).


rbuergi:
### The red here is #1616, not this PR — measured, not asserted

CI failed on one test: `MeshWeaver.InstanceSync.Test.InstanceSyncPushTest.A_pending_content_change_pushes_a_DELETE_when_the_local_node_is_being_deleted`.

This PR is **26 added lines in one markdown file**. Rather than assert "unrelated" (and "it's only markdown" is explicitly not a valid check in this repo — embedded content ships as loadable nodes with tests over it), I measured the test on `main`:

```
RESULT: 8 failure(s) in 100 iteration(s) run (requested 100);
failing iterations: 12 33 39 47 50 61 64 96
```

`flake-repro.yml`, `rate` mode, 100 iterations, **run against `main`** — https://github.com/Systemorph/MeshWeaver/actions/runs/32024085067

So the test fails ~8% of the time on `main` independently of this branch. It is already tracked as **#1616** ("the change event is dropped at 5.5s, the wait gives up at 30s"), where I have posted the rate measurement.

Supporting checks, for completeness:

- `test/MeshWeaver.InstanceSync.Test` has no reference to `MeshWeaver.Documentation` and no coupling to the changed file.
- The doc-specific gate on this very run — **"Doc content compiles and runs" — passed**, as did every other shard (1–5) and `Build solution (once)`.

Merging `main` to re-sample on a current base; no rerun of the failing job by itself.

PR #1733, opened 2026-08-17
fix: bake the platform's own content INSIDE the shipped image (#1725)

### LINKED ISSUE #1725: Platform content bake publishes under a DIFFERENT framework identity than the runtime image resolves — no pod can ever adopt it
## Evidence (prod, 2026-08-17)

Portal storage, listed from inside a running memex pod:

```
ga0a0288d5e3fd6ada32890c1585e7e1d44d0a1c8: socialmedia
s377941f549f721e01ac764e0fb8db84a:        plugins reinsurance socialmedia   ← the identity pods resolve
s6f7d50f43040910339fd87f9d88763a1:        education
sd0d0daa9398c4d810e0e730a0495b9db:        meshweaver-content                ← where CD publishes
sff0502a097d98897b489df9891075726:        meshweaver-content
```

Running pods on `3.0.0-rc4.ci.4049` log `framework=s377941f`. CD run 32018411051's publish-bake step published to `prebuilt-bundles/sd0d0daa9398c4d810e0e730a0495b9db/meshweaver-content/` (8 bundles) — taken from the Build-and-Test bake artifact `baked-assemblies-sd0d0daa9398c4d810e0e730a0495b9db`.

Consequence, measured on the first boot after the roll:

```
DynamicTypePreWarmer: warm-up complete in 00:01:04.8 — compiled=80 alreadyBaked=4 compileErrors=1 ...
```

**80 types Roslyn-compiled on every pod start, forever** — the platform's own shipped content can never be adopted, because the directory keyed to the pods' identity does not contain `meshweaver-content`. This is precisely the regression the CI-bake lane exists to prevent (#1347 → #1660), and it defeats "rebuild only when we need to" for the platform's own content.

## Why the two hosts disagree

The satellites do NOT have this problem: they bake **inside the shipped container** (`docker run … mw-plugin-test`), so their bake host is bit-identical to the runtime and resolves the same identity. The platform's own bake runs in the **Build-and-Test job's own build output**, a different compilation of the same sources — so any part of the identity that depends on the build invocation rather than the source diverges.

Since #1712 the surface identity mixes ref-assembly hashes (deterministic) with the **implementation MVIDs of the toolchain closure** (`FrameworkBuildIdentity.FullMvidAssemblies` = MeshWeaver.Compiler + MeshWeaver.NuGet + their transitive MeshWeaver closure). MVIDs are per-compilation; the CI test build (`-c Release -p:CIRun=true`) and the container publish are two compilations, so the MVID half diverges even when the source is identical. The two historical `meshweaver-content` directories show CI's own identity moving between builds as well.

## Fix direction (root cause — no fallbacks, no "publish under both")

Preferred: **bake the platform's own content inside the container image**, exactly as every satellite already does — one producer, one toolchain, one identity, and the producer/consumer question stops existing. `mw-plugin-test` is built in the same CD run and already carries `--bake-output`.

Alternative, if the artifact lane must stay: make the identity independent of build-invocation-specific MVIDs (hash the toolchain's ref-assembly surface + generated-input content rather than impl MVIDs — the #1707 S4 direction), so two compilations of one source agree by construction.

🚨 Do NOT "fix" this by publishing under several identities or by having the pod scan for a nearest match: the identity IS the compatibility proof, and adopting bytes from an identity you did not resolve is exactly what it exists to prevent.

Related: #1707 (compile cleanup), #1660 WS3, #1696 (surface identity), #1705 (manifest in the published container).

rbuergi:
Fixes #1725.

## What I measured (the issue's hypothesis, tested)

For commit `babb3bc` — same sources, same runner path — I compared the **Build-and-Test host** (`build-output-0`'s `mw-plugin-test`) against the **shipped image** (`mw-plugin-test:babb3bc`, both arches), reading the real `meshweaver-surface.manifest` and the real assembly MVIDs out of each:

| half of the identity | CI build host vs image (same commit) |
|---|---|
| implementation MVIDs (`FullMvidAssemblies` = the MeshWeaver.Compiler + MeshWeaver.NuGet closure) | **all differ** — including the controls `MeshWeaver.Data`, `.Layout`, `.AI`, `.Utils` |
| reference-assembly hashes (the other canonical entries) | 33/37 identical, **4 differ**: `MeshWeaver.Graph`, `MeshWeaver.Hosting`, `MeshWeaver.Kernel`, `MeshWeaver.Markdown.Collaboration` |

So the issue's hypothesis is **confirmed but incomplete**: the impl-MVID half guarantees divergence, *and* the surface half diverges independently. That second finding matters — the issue's alternative fix ("hash the ref-assembly surface instead of impl MVIDs") would **not** have worked.

Two more measurements:

* **amd64 vs arm64 of the SAME image**: the same four reference assemblies differ ⇒ a multi-arch image carries **two identities**.
* **CI's identity moves every run**: the last four green main runs produced `sd0d0daa…`, `sff0502a…`, `s576bb8d…`, `sf9e5b20…`, while the container identity `s377941f…` is shared by three satellite repos *and* the running pods — the satellites bake in the image, which is exactly why they were never affected.

Determinism itself is fine (a from-clean `--no-incremental` rebuild at the same path is byte-identical). The identity is a property of the **binaries a host ships**, not of the source they were built from — which is what makes it a real ABI proof. Nothing here is a defect in the identity; the defect was a lane that fed it bytes from another build.

## The fix

`main-cd`'s `publish-bake` now bakes the platform's own content **inside the image this run promoted**, exactly as every satellite repo does via `node-repo-publish-bake.yml`:

* stages the `Doc` tree and the `samples/Graph/Data` trees (the two scripts the PR gate uses), then `docker run --platform linux/amd64 … mw-plugin-test … --bake-output` against `mw-plugin-test:<staging_tag>` — the exact, immutable tag this run's `plugin-test-image` leg pushed, never the moving `latest`/`main`;
* `--platform` is **pinned, not inherited**: architecture is part of the identity and every AKS node is amd64;
* publication is unchanged — the canonical `publish-bake-bundles.sh`, every bundle first, `_complete` sentinel strictly LAST;
* the cross-run artifact hop is gone, and with it `actions: read` and the "no artifact ⇒ warn and skip" branch. The bake happens in the job: it produces bundles or fails red. It is also a **stronger gate** than the artifact hop — it proves the shipped content compiles, renders and passes its `Tests` areas against the binaries that actually ship.

`doc-gate` keeps `--bake-output` as a PR-time **smoke test** of the bake stage (the identity postcondition still fails red on a regression) but no longer uploads `baked-assemblies-*` — an artifact that looks like delivery and delivers nothing is what made this invisible.

🚨 Explicitly **not** done, per the issue: no publishing under several identities, no "nearest identity" scan, no relaxed sentinel/identity check, no fallback that adopts bytes from an identity the pod did not resolve.

## Guard

`PlatformBakeLaneGuard` (test/MeshWeaver.Documentation.Test) pins the invariant: `publish-bake` must run `mw-plugin-test` in the image with an explicit `--platform`, and **no** workflow may publish prebuilt bundles taken from a cross-run artifact. Verified it fails on the exact regression (I reinstated `gh run download` and both assertions went red with the right messages), then restored.

## Verified locally

* Both stages baked green **inside** `mw-plugin-test:babb3bc` — 8 bundles, 28 assemblies, `framework-mvid.txt` written by the non-root container user through the bind mounts.
* `dotnet build test/MeshWeaver.Documentation.Test -c Release -warnaserror` → 0 warnings, 0 errors; **23/23** tests green (incl. `DocumentationLinkIntegrityTest`, `WhatsNewEntryIntegrityTest`).
* Both workflows re-parse; every new shell block passes `bash -n`; `Directory.Build.props` is valid XML.

## Honest scope note (measured, documented, not fixed here)

A bundle is matched to a deployment by **node path**. On `memex` the `Doc/…` types match and will adopt; the sample trees live there under `MeshWeaver/samples/Graph/Data/ACME/…` while the bundles are keyed `ACME/…`, so those seven bundles are published but inert on that portal (they do adopt where the samples sit at their canonical roots). That is a content-layout question, not an identity one, and it is written up under "What this step does not do yet" in `CiContentBake.md` rather than silently bundled into this change.

## Docs

`CiContentBake.md` (new "The identity is a property of the BINARIES" section + the corrected delivery lane), `ContinuousDeliveryContract.md`, and the false determinism claim in `Directory.Build.props` that the old lane was built on.

### DIFF
--- a/.github/scripts/stage-doc-gate.sh
+++ b/.github/scripts/stage-doc-gate.sh
@@ -1,0 +1,28 @@
+#!/bin/bash
+# stage-doc-gate.sh <src-data-dir> <stage-dir>
+#
+# Stages src/MeshWeaver.Documentation/Data as the `Doc/` package root mw-plugin-test imports —
+# the shape BOTH consumers need, which is why it lives here rather than inline in one workflow:
+#
+#   * dotnet-test.yml's `doc-gate` — the PR verdict ("the Doc tree compiles, renders, tests"),
+#   * main-cd.yml's `publish-bake` — the SHIPPED IMAGE re-running the same content to produce the
+#     bundles the pods adopt.
+#
+# Those two must stage byte-identical trees: the bake publishes what the gate proved. A second,
+# drifting copy of this staging is exactly how a bake could publish something the gate never judged.
+#
+# The tree has no index.json of its own (index.md is the partition's root PAGE, and committing a
+# JSON root would ship a duplicate root node in the embedded partition), so the root is synthesized
+# HERE and index.md is dropped from the STAGE only — shipped content is untouched. NodeFileMapper
+# then maps Doc/DataMesh/SocialMedia/Post.json et al. to the exact namespaces the files themselves
+# declare, so the gate compiles the tree under its canonical paths.
+set -euo pipefail
+SRC="${1:?usage: stage-doc-gate.sh <src-data-dir> <stage-dir>}"
+STAGE="${2:?usage: stage-doc-gate.sh <src-data-dir> <stage-dir>}"
+if [ ! -d "$SRC" ]; then echo "::error::doc gate: source tree '$SRC' is missing"; exit 1; fi
+rm -rf "$STAGE"
+mkdir -p "$STAGE/Doc"
+cp -R "$SRC/." "$STAGE/Doc/"
+rm -f "$STAGE/Doc/index.md"
+printf '%s' '{"$type":"MeshNode","id":"Doc","path":"Doc","mainNode":"Doc","name":"MeshWeaver Documentation","nodeType":"Space","state":"Active"}' > "$STAGE/Doc/index.json"
+echo "staged: $(find "$STAGE" -name '*.cs' | wc -l | tr -d ' ') .cs files under $STAGE/Doc"
--- a/.github/workflows/dotnet-test.yml
+++ b/.github/workflows/dotnet-test.yml
@@ -1305,15 +1305,20 @@ jobs:
         set -euo pipefail
         tar -I 'zstd --long=27' -xf build-output-0.tar.zst
         rm -f build-output-0.tar.zst
-    # 🚨 Since #1660 WS1 this job is not verdict-only: the SAME tester run that proves the content
-    # compiles also PERSISTS the compiled assemblies (--bake-output) as identity-keyed prebuilt
-    # bundles, uploaded below as `baked-assemblies-<identity>`. Since #1660 WS3 the framework
-    # identity is COMMIT-DETERMINISTIC (for CI builds: g<sha>, stamped by Directory.Build.props —
-    # no run number or ticks reach any compiled attribute), so consumers are no longer limited to
-    # this run: main-cd's publish-bake job copies these bundles to the portals' shared storage,
-    # where every pod built from the SAME COMMIT seeds them at boot
-    # (ShippedPrebuiltBundles.SeedPublishedRoot) instead of recompiling. The seeder's identity
-    # gate still declines anything from another commit — that is the ABI gate, not a bug.
+    # 🚨 `--bake-output` here is a SMOKE TEST of the bake stage, NOT a delivery lane (#1725).
+    # It costs almost nothing — the tester has already compiled everything, this only persists it —
+    # and it makes a bake-stage regression fail on the PR that causes it rather than in CD.
+    #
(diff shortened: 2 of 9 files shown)


rbuergi:
Good catch — fixed in f57f15d01. The guard now strips comment lines before matching and probes `--entrypoint /app/mw-plugin-test` + `--bake-output /bake`, details that only exist on a real command line. The cross-run-artifact assertion strips comments too, in the other direction (prose explaining why the hop was removed must not read as the hop). Verified both ways: green as written, and red when exactly those executable lines are deleted with every comment left in place.

PR #1734, opened 2026-08-17
feat: course delivery becomes its own module

rbuergi:
Tier 2 of the modular factor-out — the clean one. Independent of #1730/#1731.

## What moves

`GET /assets/{Space}/{path…}` and its two collaborators leave `Memex.Portal.Shared` for a new `MeshWeaver.Courses` module: the route, the pure entitlement gate, and the GitHub-contents resolver. Course delivery is a **product** concern — a deployment that hosts no courses now carries neither the resolver nor the route, and asks for none of the entitlement machinery.

The move needed no rewriting: every dependency was already a framework assembly (GitSync for the Space's config + App installation token, Mesh/Messaging for identity and queries, Blazor's `PortalApplication` for the request's `AccessService`). Nothing in the platform referenced these types.

## Two things that would have broken silently

**1. The route must be explicitly `AllowAnonymous`.** It was *implicitly* anonymous in the host, and `CourseAssetGate` is what distinguishes anonymous-but-allowed (a public course) from `NotAuthenticated` (401) and `Forbidden` (403). Module contributions map inside a group defaulting to `RequireAuthorization()`, which would challenge an anonymous viewer **before the gate ran** — turning every public course's assets into a login redirect, with nothing failing at build time. Same shape as the LinkedIn OAuth callbacks relying on their CSRF cookie instead of the group policy.

**2. `CourseAssetService` stays a singleton.** My first draft registered it via `AddHttpClient<CourseAssetService>()` — which registers the type as **transient**, giving every request its own empty promise cache and silently defeating the per-file round-trip sharing the class is built around. Reverted, with the reason in the code.

## Verification

- `-c Release -p:CIRun=true -warnaserror` clean (`0 Warning(s)`) for the module, `Memex.Portal.Monolith`, and `Memex.Portal.Shared.Test`.
- 32/32 tests green against the moved assembly (fresh `.trx`).
- **Boot smoke test**, which is the only place the auth posture is observable: the Monolith starts with the module listed, and an anonymous `GET /assets/NoSuchSpace/none.png` answers **404 with no redirect** — proving the request reached the gate rather than the auth challenge. A malformed path answers 404 too. No endpoint-collision refusal at startup.

## Follow-on

The module is a natural candidate to relocate to `MeshWeaver.Education` later, the way Social moved to `MeshWeaver.SocialMedia` — double-ship first, flip after. Nothing here presumes that.


rbuergi:
Good catch — fixed in deda939bb, and you were right about the mechanism: `CourseAssetService` takes `HttpClient? httpClient = null` and falls back to `new HttpClient()`, so a singleton pins one `HttpMessageHandler` for the process lifetime (no connection recycling, no DNS refresh).

It now takes a **named** client from `IHttpClientFactory` while staying a singleton, which satisfies both constraints at once: the factory rotates handlers, and the service keeps the per-file promise cache as an instance field (a typed `AddHttpClient<T>` would register it transient and give every request an empty cache — the reason it is not one).

Constructed with named arguments, since the ctor's parameter order is `(IoPoolRegistry, IOptions<GitHubAppOptions>, GitHubAppTokenService?, ILogger?, HttpClient?, TimeSpan?)` and a positional factory would silently rebind if that ever changed.

For the record this was pre-existing — the host registered it identically before the move — but it travels with the code, so it is fixed here rather than relocated as-is. Release/`-warnaserror` clean, 32/32 course tests green.

PR #1736, opened 2026-08-17
fix: /api/content hung ~50% of the time on multi-replica portals (#1729)

### LINKED ISSUE #1729: Content route /api/content intermittently hangs forever (~50% of requests) on existing assets — anonymous reads never answer
## Symptom (prod memex, reproducible right now)

`GET https://memex.meshweaver.cloud/api/content/AgenticEngineering/content/og.png` — an asset that EXISTS — alternates between a fast 200 and a request that **never answers at all**: no status, no headers, zero bytes, until the client gives up.

```
$ for i in $(seq 10); do curl -o /dev/null -w "%{http_code} %{time_total}\n" --max-time 8 <url>; done
200 0.44   200 0.39   200 0.35   000 8.00(timeout)   000 8.00   200 0.38   000 8.00   000 8.00   200 0.41   000 8.00
→ ok=5  hang=5
```

Discriminating facts, all measured in the same minutes:

- **A MISSING file answers correctly and fast**: `…/content/does-not-exist.xyz` → `404` in 0.16 s, every time. So routing, path resolution and the not-found path are healthy.
- **A large existing file can succeed**: a 9.9 MB `.mp4` served 200 in 1.65 s while `og.png` was hanging. Not size- or bandwidth-related.
- **Single replica**: the `memex` namespace runs ONE portal pod, and it both serves and hangs — so this is per-REQUEST, not a wedged replica behind a load balancer.
- **Survives everything**: persists across `3.0.0-rc4.ci.4045` → `ci.4049`, across pod restarts, and across today's roll.
- **Anonymous caller** (plain curl, no auth).

## Impact

- Course/doc pages render with broken images for real users, intermittently.
- It is what keeps `MeshWeaver.Education`'s `Read-only smoke against a live instance` red (`smoke.spec.ts` → `apiRequestContext.get: Timeout 30000ms exceeded`), i.e. a red CI gate on a healthy repo.

## Where to look

`BlazorHostingExtensions.MapContentFiles` composes the whole endpoint as `IObservable<IResult>` and does `.FirstAsync().ToTask(context.RequestAborted)`. A hang means **one of the composed observables never emits and never errors** — the request then waits on `RequestAborted` forever. Candidates, in the order the route runs them:

1. `ResolveContentFile` → `IPathResolver` resolution of the owning node.
2. The owning hub's `GetDataRequest` for the collection (the permission-bearing hop).
3. `AllowContentRead` → for an anonymous caller this is `AnonymousGate.AllowAnonymous(hub, ownerPath)`; for an authenticated one `hub.CheckPermission(...).Take(1)`. Note the anonymous arm is the one exercised here.

A 50/50 split on identical input smells like a race between two paths that both feed one stream (e.g. a cached/replayed permission or collection stream where one producer completes and the other silently doesn't), or a lost hub round-trip whose failure is never surfaced as an OnError. There is no timeout on the composition by design, so a dropped emission becomes an infinite hang rather than an error.

🚨 The fix is the missing/dropped emission — **not** a `.Timeout(...)` on the route. A timeout would convert a silent hang into a silent 500 and lose the defect.

Related: #1701 (hub recycle wedges), #1669/#1672 (rebind blindness — same "never emits" family), the `DeliveryFailureException`/tri-state handling already in `ContentFailure`.

rbuergi:
Fixes #1729.

## What was actually wrong (the issue's premise was falsified)

The issue reported a **per-request** ~50/50 hang on a **single replica**. Both halves are wrong, and finding that out is what located the defect:

* `memex.meshweaver.cloud` is served by namespace **`memex-cloud`**, not `memex` — and its portal Deployment runs **2 replicas**, both in the Service.
* Probing each pod directly (in-cluster, bypassing the ingress) is **fully deterministic**, not 50/50:

  | pod | `AgenticEngineering/content/og.png` | `Chess/content/zz.png` |
  |---|---|---|
  | `10.244.2.139` | **hang** (5/5) | 404 in 2.3 s |
  | `10.244.3.119` | 200 in 18 ms | **hang** |

  Each pod serves `/api/content` **only for the nodes whose per-node hub grain it hosts**, and hangs on every other node. Round-robin across the two replicas is what turns that into "~half of requests to any given asset hang".
* It is **not an infinite hang** — a request with a long client budget returns **HTTP 500 after 60.009 s**. Browsers and the smoke test simply give up first.

## Root cause

`MapContentFiles` resolves `services.GetRequiredService<IMessageHub>()` — which, because `MeshHostApplicationBuilder` makes the mesh hub's provider the ASP.NET root provider, **IS the root `mesh/{id}` hub, the ROUTER** — and hands it to `ContentFileResolver.Resolve`, which posts the owning node's collection-config `GetDataRequest` from it.

That makes the router an END of the delivery in both directions: the request reaches the per-node hub stamped `Sender = mesh/{id}`, and the `GetDataResponse` is addressed straight back at `mesh/{id}`. Same-silo the reply short-circuits on the routing service's local table (healthy); **cross-silo it never arrives**, so the request runs out its 60 s budget.

Both pods' own diagnostics name it precisely. Requesting pod:

```
[STALE-CALLBACK] mesh/IJ1R4… : 1 callback(s) pending > 30000ms: GetDataRequest@AgenticEngineering(55672ms)
  → ROUTED onTarget=False state=Forwarded ⇒ the delivery reached a hub but no handler was ever entered
System.TimeoutException: No response received in hub mesh/IJ1R4… within 00:01:00 for request GetDataRequest → target AgenticEngineering
```

Owning pod, same request — the reply **was** produced and had nowhere to go:

```
AccessControlPipeline: Access denied: user 'Anonymous' lacks Read permission on 'MeshWeaver'
  (triggered by message=GetDataRequest sender=mesh/IJ1R4… hub=MeshWeaver)
ROUTER_TRAFFIC: GetDataResponse has the mesh hub as target (sender: MeshWeaver, target: mesh/IJ1R4…)
```

This is a documented, absolute rule the rest of the codebase already follows — `SessionHubResolver`: *"This helper NEVER returns the root `mesh/{id}` hub … request-shaped work issued on it never answers … Every API surface (MCP, REST, gRPC, CLI) issues on a portal hub."* The content route is the **last unmigrated caller**.

## The fix

One seam, the same one every other caller uses:

```csharp
return hub.NodeOperationIssuingHub().Observe(new GetDataRequest(...), …)
```

`NodeOperationIssuingHub()` hops a root-hub caller onto `portal/nodeops-{meshId}` — routing-registered so replies land cross-silo, sharing the mesh's type registry and permission evaluator — and returns any **non-router** hub unchanged, so in-mesh callers (the deck export's `SlideAssetInliner`, a portal hub, a test client) keep their identity byte-for-byte. `GetMeshNode`/`GetMeshNodeOutcome` and every node-CRUD path already went this way in `2c796d297`.

No timeout, no retry, no watchdog, no `catch` — the dropped emission is fixed at its source.

## How the gap survived

`OrleansCrossSiloReplyTest` was written for exactly this failure (core#694 layer 2) and its docstring still names the static-content endpoint. But `2c796d297` moved `GetMeshNode` onto `NodeOperationIssuingHub`, so since 2026-08-10 that test has exercised the `portal/nodeops-…` reply path and **no longer posts from `mesh/{id}` at all** — it silently stopped guarding the endpoint it names. A docstring now records that, and points at the new test.

## Regression test

`ContentRouteIssuingHubTest` drives the **real** `/api/content` endpoint over HTTP against a real monolith mesh and asserts the **sender the owning node hub sees** is not the router.

* **Fails on the unfixed code** — `Did not expect value to be "mesh"`, and the run reproduces the exact production `ROUTER_TRAFFIC` pair.
* **Passes with the fix.**

It deliberately does **not** live in the two-silo Orleans suite: an in-process `TestCluster` does not reproduce the cross-PROCESS reply loss. I wrote that version first and **it passed with and without the fix** — worse than no test — so it was removed rather than shipped as false assurance.

## Verification

* `dotnet build -c Release -p:CIRun=true -warnaserror` (whole solution): 0 warnings, 0 errors
* `MeshWeaver.Hosting.Blazor.Test`: **403/403** (includes every existing `/api/content` authorization test)
* `DeckAssetInliningTest` (the other `ContentFileResolver.Resolve` consumer): 5/5
* `DocumentationLinkIntegrityTest`: 1/1

Also unblocks `MeshWeaver.Education`'s `Read-only smoke against a live instance`, which is red solely because it is correctly catching this bug.

### DIFF
--- a/src/MeshWeaver.ContentCollections/ContentFileResolver.cs
+++ b/src/MeshWeaver.ContentCollections/ContentFileResolver.cs
@@ -1,6 +1,7 @@
 using System.Reactive.Linq;
 using System.Text.Json;
 using MeshWeaver.Data;
+using MeshWeaver.Mesh;
 using MeshWeaver.Mesh.Services;
 using MeshWeaver.Messaging;
 using Microsoft.Extensions.DependencyInjection;
@@ -133,7 +134,37 @@ public static class ContentFileResolver
                 ? new[] { explicitCollection, defaultCollection }
                 : [defaultCollection];
 
-            return hub.Observe(
+            // 🚨 NEVER issue this read on the ROOT MESH HUB — the router. The /api/content endpoint
+            // holds the DI-injected IMessageHub, which in the mesh's root container IS mesh/{id}
+            // (MeshHostApplicationBuilder makes the mesh hub's provider the ASP.NET root provider),
+            // so this GetDataRequest used to make the router an END of the delivery in BOTH
+            // directions: the request reached the per-node hub stamped Sender = mesh/{id}, and the
+            // GetDataResponse (or, for a denied/missing node, the DeliveryFailure) was addressed
+            // straight back at mesh/{id}. Same-silo that reply short-circuits on the routing
+            // service's local table and everything looks fine; CROSS-silo it has to arrive over the
+            // cluster-wide memory stream, and it does not — so the request never answers, the
+            // caller waits out its full 60 s budget and the route 500s with a TimeoutException.
+            //
+            // Prod signature (memex-cloud, 2 replicas, issue #1729): each pod served /api/content
+            // ONLY for the nodes whose per-node hub grain it happened to host and hung for ~60 s on
+            // every other node, so round-robin across the two replicas made ~half of all requests to
+            // ANY given asset hang — broken images on course/doc pages, and a red live-smoke gate.
+            // The pod's own diagnostics named the caller exactly as documented:
+            //   [STALE-CALLBACK] mesh/IJ1R4… : GetDataRequest@AgenticEngineering(55672ms)
+            //     → ROUTED onTarget=False state=Forwarded ⇒ no handler was ever entered
+            //   System.TimeoutException: No response received in hub mesh/IJ1R4… within 00:01:00
+            // while the OWNING silo logged the matching pair:
+            //   ROUTER_TRAFFIC: GetDataResponse has the mesh hub as target (sender: MeshWeaver,
+            //   target: mesh/IJ1R4…)
+            // i.e. the reply WAS produced and had nowhere to go.
+            //
+            // NodeOperationIssuingHub is the one shared seam for this: it hops a root-hub caller onto
+            // portal/nodeops-{meshId} — routing-registered so responses land on it cross-silo, and
+            // sharing the mesh's type registry and permission evaluator — and returns any NON-router
+            // hub unchanged, so in-mesh callers (the deck export's SlideAssetInliner, a portal hub,
+            // a test client) keep their identity byte-for-byte. GetMeshNode/GetMeshNodeOutcome and
+            // every node-CRUD path already went this way (2c796d297); this read was the straggler.
+            return hub.NodeOperationIssuingHub().Observe(
                     new GetDataRequest(new ContentCollectionReference(candidates)),
                     o =>
                     {
--- a/src/MeshWeaver.Documentation/Data/WhatsNew/2026-08-17-content-route-hang.md
+++ b/src/MeshWeaver.Documentation/Data/WhatsNew/2026-08-17-content-route-hang.md
@@ -1,0 +1,19 @@
+---
+Name: Images and downloads no longer hang
+Category: Fix
+Description: Content files served from /api/content could hang instead of loading on portals running more than one replica.
+Icon: Sparkle
+Order: -20260817
+---
+
+# Images and downloads no longer hang
+
+Pictures, videos and file downloads stored on a node are served through the content route. On a
+portal running more than one replica, roughly half of those requests never answered — the page
+showed a broken image or an empty player, and a download simply stalled. Reloading sometimes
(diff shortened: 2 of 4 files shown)


rbuergi:
Both review points addressed in b913c3d17 — thanks, the second one was a real hole.

- **Disposal**: `ContentRouteIssuingHubTest` now overrides `DisposeAsync` to dispose the `HttpClient` and the TestServer-backed `WebApplication` and to delete its per-test storage fixture, matching `StaticContentUnmountedTest`.
- **False pass**: `collectionConfigSender` is now cleared immediately before the HTTP call. Mesh start-up also reads collection configs, so an earlier observation from a legitimately non-router hub would have satisfied the assertion while the route itself still posted from the router — i.e. the one test that exists to catch this bug could have passed without testing anything.

Re-verified after the change that the test still **discriminates**: it fails on the unfixed code (`Did not expect value to be "mesh"`) and passes with the fix. Full `MeshWeaver.Hosting.Blazor.Test`: 403/403.

PR #1737, opened 2026-08-17
feat: mail over Microsoft Graph becomes its own module

rbuergi:
Tier 2 of the modular factor-out — the biggest dependency win available.

## The prize, measured

`Microsoft.Graph` + Kiota was the heaviest dependency in the portal image: **43 MB across ten assemblies**, carried by *every* deployment for four files of code. It also costs at runtime — `KernelScriptReferences` documents `Microsoft.Graph.dll` as a **41 MiB native metadata block** in every Roslyn script reference set, named there as a direct cause of CI memory-pressure flakes.

After the move, measured on a Release build of the Monolith:

- **Zero** `Microsoft.Graph*` / `Microsoft.Kiota*` assemblies in the app output.
- All ten in `modules/MeshWeaver.Mail.MicrosoftGraph/` (43.6 MB).

## What moved, and what deliberately did not

**Moved:** `GraphMail`, `GraphEmailSender`, `EmailInboundProcessor`, `GraphSubscriptionService`, `ExecutiveAssistantPlugin`, and the change-notification webhook.

**Stayed in the host** — everything mail-shaped that never touched the SDK: `InvitationEmailSender`, `OutboundEmailSender`, `NoOpEmailSender`, and `EaGraphAuth` with its consent controller (that flow is raw OAuth over HTTP). The seam already existed: `IEmailSender`/`EmailOptions` live in the mesh contract and `HubEmailExtensions` resolves the sender optionally.

## Three couplings designed rather than moved

**1. `IEmailSender` resolution is order-independent.** Two host services call `GetRequiredService<IEmailSender>()`, so "module absent + `Email:Enabled=true`" would throw at startup. The host now registers the no-op with `TryAddSingleton` while the module uses plain `AddSingleton` — whichever runs first, the last registration wins for `GetRequiredService` and `TryAdd` declines when one exists.

**2. 🚨 The internal-query back-door, and the one security-relevant change here.** `EmailInboundProcessor` read through `IMeshQueryCore` — an **`internal`** interface documented as *"raw queries without user context … must not be exposed to application code"*. Reaching it from a module means granting `InternalsVisibleTo` on the mesh contract, re-opening precisely the IVT coupling #1686 retired. Both call sites now use the public `IMeshService.Query` with `WellKnownUsers.System` — the same viewer-independent read `CourseAssetEndpoints` already performs, and System holds `Permission.All` so the result set is unchanged.

**This changes HOW an unauthenticated inbound email finds its recipient, not WHICH recipient it finds — please review that specifically.** If you would rather it not travel with a packaging change, the alternative is leaving intake in the host; the SDK still leaves the image, since the sender and EA tools are the other SDK users.

**3. The consent route had exactly one definition, and still does.** The EA's "please connect" link took its path from `EaConsentController` (a host MVC controller). `IEaGraphAuth` now exposes `ConnectPath`, with the host implementation surfacing the controller's constant — so the module references no host type and the route is still defined once. `IEaGraphAuth` itself moved to the mesh contract; it is deliberately SDK-free (strings and bools).

## The webhook

An MVC controller became a minimal-API endpoint, because the module endpoint hook maps route handlers, not controllers (a module's controller would additionally need its assembly registered as an MVC ApplicationPart). Same route, verb, shapes and status codes. `AllowAnonymous` because Graph posts unauthenticated and the shared `clientState` is the guard — and because module routes are otherwise authenticated by default.

## Verification

- `-c Release -p:CIRun=true -warnaserror` clean for the module, `Memex.Portal.Monolith` and `Memex.Portal.Shared.Test`.
- 6/6 email tests green.
- **Boot smoke test**: `POST /api/email?validationToken=mw-probe-123` answers `200 text/plain` with the token echoed and no redirect — the Graph subscription handshake working anonymously through the module contribution. No endpoint-collision refusal at startup.


rbuergi:
Fixed in fc372d33d — and you were right that it needed reconciling rather than picking one.

Both numbers were measured, from different globs. **Nine** matched `Microsoft.Graph*`/`Microsoft.Kiota*`, which is what the app-root `du` counted. The module folder receives **ten**: the tenth is `Microsoft.IO.RecyclableMemoryStream`, a private dep of the Graph client that the glob simply didn't match.

It now says ten everywhere, with the composition spelled out so the two counts can't silently diverge again. A measurement that appears twice with different values stops being evidence, which is the real cost here.

PR #1743, opened 2026-08-17
fix(#1616): a push that lands after a local delete no longer reports drained

### LINKED ISSUE #1616: InstanceSyncPushTest DELETE-push test times out on main and unrelated branches — the change event is dropped at 5.5s, the wait gives up at 30s

rbuergi:
Closes #1616.

## Measurement first

Same instrument, same filter, same iteration count, ~40 minutes apart:

| arm | result |
|---|---|
| `main` | **8 failures in 100** — iterations 12, 33, 39, 47, 50, 61, 64, 96 ([run 32024085067](https://github.com/Systemorph/MeshWeaver/actions/runs/32024085067)) |
| this branch | **0 failures in 100** ([run 32026294545](https://github.com/Systemorph/MeshWeaver/actions/runs/32026294545)) |

🚨 **This contradicts something recorded on #1616 and it matters for anyone else working there.** The issue concludes that a filtered `flake-repro` loop does *not* reproduce and that "the instrument is a full shard" — measured 0/40 on main and 0/40 on a failing branch on 2026-08-14. That is no longer true: the same loop measured 8/100 on main three days later. The rate moved, so the A/B this issue asks for is possible again. Its warning still stands in principle — a green loop is not evidence when the loop cannot go red — which is exactly why the before-arm was measured on `main` first rather than assumed.

## Root cause

`DrainOne` resolves the local node **before** the push; `DrainPending` removes the entry **after** the push reports success. Those two are separated by a full remote round trip:

1. While the remote is unreachable, attempts retry every 100–400 ms, each reading the local node and then failing at the network.
2. An attempt reads the node as `Present` **just before** the delete tombstone lands.
3. `MarkDeleted` is written **synchronously** by the delete handler and emits **no change event** — so nothing re-enters the manifest to catch it.
4. The remote comes back. That in-flight attempt's push now succeeds, pushing content that is already stale, and reports the entry drained.
5. The entry is removed. The remote keeps a node that was deleted locally, and there is nothing left in the manifest to correct it.

That is why the test waits out its full 30 s: the DELETE it waits for is never produced. It also explains the `Dropping DataChangedEvent … the target stream is gone` at ~5.5 s quoted in the issue — the same window seen from the other side.

The bound was never the defect. The issue says so, and the test's own intervals (`DrainDebounce` 50 ms, `PullInterval` 200 ms, `RetryInitial` 100 ms) cover ~150 pull attempts inside the wait.

## The fix

The drain re-resolves the local node after a successful push and converts it into a delete when the node went away while it was in flight.

Deliberately narrow — **only `Absent` and `DeleteInProgress` change the outcome**. A still-present node, an unreadable one, or a status this predates all report the push as drained exactly as before. Never destroy the remote copy on evidence we could not gather: that is the rule `DrainOne`'s first read already states, and an earlier fix in this same file exists to enforce it.

Not fixed by raising a bound, adding a retry, or a `Task.Delay`.

## Blast radius

One added local node read per successfully pushed entry, against a push that is already a remote round trip. No change to the delete fast-path, the pull sweep, or the connectivity/retry classification.

### DIFF
--- a/src/MeshWeaver.InstanceSync/InstanceSyncWorker.cs
+++ b/src/MeshWeaver.InstanceSync/InstanceSyncWorker.cs
@@ -330,10 +330,11 @@ public sealed class InstanceSyncWorker : IDisposable
 
         // Push the node's CURRENT content; a node deleted since the change was observed
         // degrades to a delete.
-        return hub.GetMeshNodeOutcome(change.Path, TimeSpan.FromSeconds(15), ReadTimeoutBehavior.EmitNull)
+        return hub.GetMeshNodeOutcome(change.Path, LocalReadTimeout, ReadTimeoutBehavior.EmitNull)
             .SelectMany(outcome => outcome.Status switch
             {
-                NodeReadStatus.Present => PushNode(remote, cfg, outcome.Node!).Select(_ => true),
+                NodeReadStatus.Present => PushNode(remote, cfg, outcome.Node!)
+                    .SelectMany(_ => ConfirmNotDeletedDuringPush(remote, change.Path, remotePath)),
                 NodeReadStatus.Absent => DeleteRemote(remote, remotePath),
                 // The local delete is in flight; the authoritative next state is "gone", and the
                 // manifest holds a Deleted entry for it too. Pushing the delete now is idempotent.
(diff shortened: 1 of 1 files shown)


rbuergi:
Correct, and this was the most valuable finding on any of today's PRs — the fix contained the bug it was fixing.

The `.Catch` was there so an unreadable *local* node wouldn't fail an entry whose push had already succeeded. But it wrapped the whole chain, so a failing `DeleteRemote` was swallowed and the entry still reported drained — leaving the remote holding a locally-deleted node, which is exactly the defect #1616 describes, reintroduced one layer down through the error path.

Fixed in d5a0784fe: the catch is scoped to the re-read only.

```
.Select(after => after.Status is Absent or DeleteInProgress)
.Catch<bool, Exception>(_ => Observable.Return(false))   // read failed → leave as drained
.SelectMany(mustDelete => mustDelete ? DeleteRemote(...) : Observable.Return(true));
```

A failing corrective delete now propagates into `DrainPending`'s existing classifier: a connectivity fault aborts the pass for the retry probe, anything else surfaces on the config as `LastError`. Either way the entry stays pending and the next pass retries — which is the point of having noticed the delete at all.

Full `MeshWeaver.InstanceSync.Test` suite 26/26 green, Release/`-warnaserror` clean. The A/B rate measurement in the PR body was taken on the previous commit; the change is in the error path only (the success path is byte-identical), but say the word if you'd rather I re-run the 100-iteration arm against this head before merge.

Issue #1613, opened 2026-08-14
RenderArea zero-timeout leaks a pending SubscribeRequest when the target hub is still Starting


rbuergi:
## Symptom

`MeshWeaver.AI.Test.RenderAreaOperationTest.RenderArea_Timeout_FaultsWithTimeoutException` fails in CI with a **teardown** error, not a test-body failure. Seen on run 31832029430 (shard 4), on a branch whose diff is two files in `InstanceKeys.cs` — no path to RenderArea.

The test body passes: `TEST START 19:20:51.403` → `TEST END 19:20:51.632`, and `Mesh.Dispose` reports *"teardown clean — all pooled I/O joined, async dispose queue drained"*. The throw comes from `MonolithMeshTestBase.DisposeAsync()` line 1469 (`throw disposeException`).

## Root cause, from the framework's own delivery trace

```
RenderAreaOperationTest left Observe subscriptions pending past the Quiescing budget.
Hub mesh/PAK8JykQe0GPFtQVUMlzpw: 1 pending callback(s) after 0.50s:
  SubscribeRequest@TestData/render-5bca654fe5564 (496ms)

  AWAITING SubscribeRequest → POSTED → RECEIVED runLevel=Quiescing → ENQUEUED
  → RECEIVED runLevel=Starting@TestData/render-… → ENQUEUED
  → ROUTED onTarget=False state=Forwarded
  → ROUTED onTarget=True state=Submitted (+78ms)
  → DEFERRED gates=[DataContextInit,Initialize] (+78ms)
```

`RenderArea(..., timeoutSeconds: 0)` faults the caller immediately — which is the behaviour the test asserts and it is correct. But by then `GetRemoteStream` has already posted a `SubscribeRequest` to the freshly-created target hub. That hub is still `Starting`, so it **defers** the delivery behind `DataContextInit,Initialize` and never replies.

`RenderResolvedArea` does dispose the stream (`.Finally(stream.Dispose)`), so the local side is clean. What survives is the **reply callback for the already-posted SubscribeRequest**: disposal does not resolve or cancel it, so the leak detector correctly reports it at teardown.

## Why it is load-dependent

It needs the target hub to still be initialising when the zero budget elapses. On an unloaded machine the test passes (locally: 490 ms, green). On a loaded CI runner the hub is still `Starting` at +78 ms and the delivery is deferred. That makes it read as a flake while being a real leak.

## Suggested direction

[...]

PR #1745, opened 2026-08-17
fix: RenderArea must not open a layout-area stream after its budget elapsed

rbuergi:
## What turned main red

Run [32025437472](https://github.com/Systemorph/MeshWeaver/actions/runs/32025437472) (shard 3), commit `4c0cc71d9`. One test, `MeshWeaver.AI.Test.RenderAreaOperationTest.RenderArea_Timeout_FaultsWithTimeoutException`. It is **not** a body failure — the assertions passed; the throw is from `MonolithMeshTestBase.DisposeAsync()`:

```
System.InvalidOperationException : RenderAreaOperationTest left Observe subscriptions pending past the Quiescing budget.
Hub mesh/M8jj6cm1PUG9640CTBLJ2g: 1 pending callback(s) after 0.50s:
  IvTPPAo1DE6AJvBX3VZy4A=SubscribeRequest@TestData/render-d01df4c6b4784(509ms)
  … → RECEIVED runLevel=Quiescing@mesh/M8jj… → … → ROUTED onTarget=True state=Submitted(+32ms)
    → DEFERRED gates=[DataContextInit,Initialize,MeshNodeInit](+32ms)
```

The decisive stage is `RECEIVED runLevel=**Quiescing**`: the `SubscribeRequest` was posted **after the test method had returned and teardown had begun**.

## Attribution: pre-existing flake, NOT a regression from #1733

Measured, not argued:

- The **identical signature** — same test, same leaked `SubscribeRequest`, same `DEFERRED gates=[DataContextInit,Initialize…]` — occurred on **2026-08-14, run [31832029430](https://github.com/Systemorph/MeshWeaver/actions/runs/31832029430) (shard 4)**, three days before #1733 existed. That is what issue #1613 was filed for.
- #1733's diff carries **zero product code**: two workflows, one shell script, three docs, one What's New entry, one new test file, and `Directory.Build.props` — whose 24 changed lines contain **0 lines with an XML element tag** (comment-only; verified mechanically).
- `dotnet-test.yml`'s diff contains **no shard-composition change**, so the test did not move to a busier shard.
- PR #1736's own run [32025997920](https://github.com/Systemorph/MeshWeaver/actions/runs/32025997920) — a tree **containing** #1733, with **all six shards executed** — ran this test green.

## Root cause

`RenderArea` enforced its budget with an outer `Observable.Timeout(...)` **only**. Timeout faults the *caller*; it does not stop the pipeline. Rx disposal is unsynchronised, so a path-resolution emission already in flight still runs its continuation — and that continuation is the expensive, resource-**allocating** half: `GetRemoteStream` mints a `sync/` hub and posts a `SubscribeRequest` that **wakes a cold per-node hub**.

Opened after the budget is gone, that subscribe is owned by nobody: the caller has already been faulted, so the reply is awaited by no one and the request parks on the target's init gates. In CI that is the leaked callback above. In a portal it is an orphaned sync hub + heartbeat for every abandoned render.

Why it looked load-dependent: locally the target hub answers in milliseconds, so the **owner's reply** (not disposal) closes the callback and the test is green. On a loaded runner the hub is still `Starting` and the reply never comes inside the 0.5 s quiescing budget.

## The fix

The budget already documents itself as covering *"path resolution + the first materialised frame"*. So carry it as a **deadline** and re-check it at the last instant before committing to the stream — once it is spent there is nothing left to buy by opening one.

That also makes a **non-positive budget deterministic** instead of a race between the outer Timeout and the continuation: `elapsed >= TimeSpan.Zero` always holds, so the stream is never opened and the verb always faults with `TimeoutException`. The test's old premise — *"a zero budget elapses before any frame can arrive — deterministic"* — leaned on the outer Timeout alone and was racy in **both** directions: a synchronously-arriving frame would have failed it with `OnNext`, and the abandoned `SubscribeRequest` tripped the class's leaked-callback teardown check.

No band-aid: no retry, no widened timeout, no delay, no skip. The test keeps exactly what it proved (a zero budget must FAULT with `TimeoutException`) and now proves it deterministically.

## Scope note — what this deliberately does NOT fix

Issue #1613 asked why the existing cancel-on-dispose machinery did not save this. I found the answer and left it on the issue rather than widening this PR:

`reduced.RegisterForDisposal(observeSubscription)` routes to `RegisterForDisposal` on the stream's **inner `sync/` hub**, whose `disposables` are disposed only in that hub's **ShutDown phase**. `MessageHub.Dispose()` merely posts `ShutdownRequest(Quiescing)` and returns, so cancelling the in-flight `SubscribeRequest` is gated behind a multi-turn asynchronous hub teardown (Quiescing → DisposeHostedHubs → HostedHubsDisposed → ShutDown) that nothing bounds and that routinely exceeds 0.5 s under load. `WrapWithCancelOnDispose` is correct; the stream-disposal path simply never reaches it in time. That remains #1613's to fix; this PR removes the case that was firing.

## Verification

- `dotnet build test/MeshWeaver.AI.Test -c Release -warnaserror` → **0 Warning(s), 0 Error(s)**.
- `RenderAreaOperationTest` (all 9 tests, `DOTNET_PROCESSOR_COUNT=4`) → **9 passed**.
- Flake-repro `rate` baseline dispatched on main for the measured before/after.

No What's New entry: internal resource-leak / CI-stability fix. The REST surface (`POST /api/mesh/render-area`) clamps the budget to `[1,120]`s, so the response contract is unchanged.

### DIFF
--- a/src/MeshWeaver.AI/MeshOperations.cs
+++ b/src/MeshWeaver.AI/MeshOperations.cs
@@ -1,5 +1,6 @@
 using System.Collections.Frozen;
 using System.Collections.Immutable;
+using System.Diagnostics;
 using System.IO;
 using System.IO.Compression;
 using System.Reactive.Concurrency;
@@ -513,8 +514,20 @@ public class MeshOperations
 
         var pathResolver = hub.ServiceProvider.GetRequiredService<IPathResolver>();
         var accessService = hub.ServiceProvider.GetService<AccessService>();
+        var budget = TimeSpan.FromSeconds(timeoutSeconds);
         return Observable.Defer(() =>
         {
+            // 🚨 The budget is a DEADLINE, not merely the outer `.Timeout(...)` below. Timeout
+            // faults the CALLER; it does not stop this pipeline. Rx disposal is unsynchronised, so
+            // a resolution emission already in flight still runs its continuation — and that
+            // continuation is the expensive, resource-ALLOCATING half: GetRemoteStream mints a
+            // `sync/` hub and posts a SubscribeRequest that WAKES a cold per-node hub. Opened
+            // after the budget is gone, that subscribe is owned by nobody: the caller has already
+            // been faulted, so the reply is awaited by no one and the request parks on the target's
+            // DataContextInit/Initialize/MeshNodeInit gates (issue #1613 — reported as a leaked
+            // callback, and in a portal an orphaned sync hub + heartbeat per abandoned render).
+            // So the remaining budget is re-checked at the last instant before committing to it.
+            var startedAt = Stopwatch.GetTimestamp();
             // Capture the CALLER's ambient identity at subscribe time (the transport request
             // thread / test context — same precedence the sync-stream uses). The remote-stream
             // creation below happens inside a reactive continuation on a hub scheduler where the
@@ -524,9 +537,10 @@ public class MeshOperations
             var caller = accessService?.Context ?? accessService?.CircuitContext;
             return pathResolver.ResolvePath(resolvedPath)
                 .Take(1)
-                .SelectMany(resolution => RenderResolvedArea(resolution, resolvedPath, area, id, caller));
+                .SelectMany(resolution => RenderResolvedArea(
+                    resolution, resolvedPath, area, id, caller, startedAt, budget));
         })
-            .Timeout(TimeSpan.FromSeconds(timeoutSeconds))
+            .Timeout(budget)
             .Catch((Exception ex) =>
             {
                 // A timeout stays a FAULT (transports map it to a gateway timeout); everything
@@ -541,14 +555,17 @@ public class MeshOperations
     /// <summary>
     /// Second half of <see cref="RenderArea"/>: builds the <see cref="LayoutAreaReference"/> from
     /// the resolution + explicit arguments and opens the one-shot area stream under the captured
-    /// caller identity.
+    /// caller identity — unless the budget handed down from <see cref="RenderArea"/> is already
+    /// spent, in which case it faults WITHOUT opening the stream (see the deadline note there).
     /// </summary>
     private IObservable<string> RenderResolvedArea(
         AddressResolution? resolution,
         string resolvedPath,
         string? area,
         string? id,
-        AccessContext? caller)
+        AccessContext? caller,
+        long startedAt,
+        TimeSpan budget)
     {
         if (resolution is null)
             return Observable.Return($"Not found: {resolvedPath}");
@@ -578,6 +595,18 @@ public class MeshOperations
         // and Finally guarantees disposal on every terminal path.
         return Observable.Defer(() =>
         {
+            // 🚨 LAST GATE before the only step in this verb that allocates mesh resources. The
+            // budget covers "path resolution + the first materialised frame" (see the parameter
+            // doc), so once it is spent there is nothing left to buy by opening the stream — only
+            // an orphaned `sync/` hub and a SubscribeRequest that wakes a cold per-node hub for a
+            // caller who is already gone. Faulting here is what makes a non-positive budget
+            // DETERMINISTIC (elapsed >= 0 always holds, so the stream is never opened) instead of
+            // a race between the outer Timeout and this continuation.
+            if (Stopwatch.GetElapsedTime(startedAt) >= budget)
+                return Observable.Throw<string>(new TimeoutException(
+                    $"RenderArea budget of {budget.TotalSeconds:0.###}s elapsed for '{resolvedPath}' "
+                    + "before the layout-area stream could be opened."));
+
             // This factory runs inside a reactive continuation where the ambient AsyncLocal
             // identity is NOT the caller's. GetRemoteStream captures the ambient AccessContext
             // at stream creation (it stamps the SubscribeRequest's identity), so re-apply the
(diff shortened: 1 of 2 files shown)


rbuergi:
## Measured evidence (updating the PR body's "flake-repro dispatched" line)

### The rate harness does NOT reproduce this — reporting that honestly

| run | ref | result |
|---|---|---|
| [32027663767](https://github.com/Systemorph/MeshWeaver/actions/runs/32027663767) | `main` (`a77017650`) | **0 failures / 100 iterations** |
| [32028542071](https://github.com/Systemorph/MeshWeaver/actions/runs/32028542071) | this branch | **0 failures / 100 iterations** |

So the before/after comparison is **inconclusive**, and I am not claiming 0/100 as proof the fix works. Two reasons the harness misses it, both visible in its own log line `nproc=4`:

1. It loops **one test alone**. The leak only becomes *visible* when the target per-node hub is still `Starting` at teardown, which needs a machine busy with a whole shard's worth of projects — not a 4 s isolated iteration.
2. `ubuntu-latest` reported **4 vCPU**, not the 2-vCPU runner the workflow header assumes.

What the runs do establish: the failure is rarer than 1/100 under low contention, which is consistent with the two observed occurrences being load-dependent — and is exactly why a rate is the wrong instrument here.

### The right instrument: a negative control on the assertion

The failure was never probabilistic in the "sometimes the code is wrong" sense. The `SubscribeRequest` was posted on **every** run of this test, locally included; all that varied was whether the owner replied inside the 0.50 s teardown budget. So the thing to measure is whether the request is posted at all.

`06dfe4f5` adds the assertion that measures it, and I verified it fails against the unfixed product code. With `src/MeshWeaver.AI/MeshOperations.cs` reverted to its pre-fix version and the test unchanged:

```
MeshWeaver.Reactive.Assertions.AssertionException :
  Expected "The operation has timed out." to contain "nothing was rendered"
  because a non-positive budget must be refused outright — no path resolution, no remote stream.
Failed!  - Failed: 1, Passed: 0
```

`"The operation has timed out."` is Rx's own `Timeout` message — measured confirmation that pre-fix, the caller was faulted by the **outer timer** while the render **went ahead underneath it** and opened the stream. With the fix restored: **9/9 green**, `0 Warning(s) 0 Error(s)` under `-c Release -warnaserror`.

That converts the regression gate from "the CI flake comes back on someone else's PR" into "this line fails immediately".

### Attribution recap

Not a #1733 regression. The same signature predates it by three days ([run 31832029430](https://github.com/Systemorph/MeshWeaver/actions/runs/31832029430), 2026-08-14 — the occurrence #1613 was filed on), #1733 carries no product code (its `Directory.Build.props` hunk has 0 lines containing an XML element tag) and no shard-composition change, and a tree *containing* #1733 with all six shards executed ran this test green ([32025997920](https://github.com/Systemorph/MeshWeaver/actions/runs/32025997920)).

rbuergi:
## CI red on `06dfe4f5` — attributed, not rerun blindly

Shard 0 failed on **`MeshWeaver.Hosting.Orleans.Test.OrleansThreadStreamingTest.ResponseText_StreamsToMessageNode`**, not on anything this PR touches. Attribution, in order of strength:

1. **This PR's own first run is the control.** [Run 32028464706](https://github.com/Systemorph/MeshWeaver/actions/runs/32028464706) carried the entire product change (`acfeed92f`) and was green on **all six shards**, shard 0 included. The only delta in `06dfe4f5` is an early `return` for `timeoutSeconds <= 0` plus one test assertion.
2. **The diff cannot reach it.** The change alters `MeshOperations.RenderArea` only when the budget is non-positive or already spent. `MeshWeaver.Hosting.Orleans.Test` never calls `RenderArea` — its single mention of `MeshOperations` is a doc comment about `RecycleCore`.
3. **Merging main would buy nothing.** Main is two commits ahead; `b913c3d17` touches only `ContentRouteIssuingHubTest.cs`, and #1736 is the `/api/content` routing fix. Neither is upstream of this test, so the "stale branch re-samples a flake main already fixed" case does not apply.
4. **The suite's background flake rate is currently diffuse.** Each of the last six red runs failed on a *different* single test, none repeating (table in #1753).

RCA of the failure itself is filed as **#1753** rather than lost: the wait at `OrleansThreadStreamingTest.cs:102` timed out with `Last of 3 emission(s) was: []`, and the projection one line above uses the forbidden raw `node?.Content as MeshThread`, which makes "the round never appended a message" and "the content never materialised as `MeshThread`" indistinguishable. That is a hand-off with a concrete first step, not a shrug.

Re-sampling shard 0 once on that basis; if it reds again on the same test I will stop and escalate rather than keep pulling the lever.

rbuergi:
Thanks — the negative-budget point was a real gap, and it was live against the **first** commit here: with only the in-pipeline deadline check, `RenderArea(..., -1)` composed `.Timeout(TimeSpan.FromSeconds(-1))`, and Rx's argument validation throws `ArgumentOutOfRangeException` **out of the call** rather than faulting the observable — so a caller composing this verb reactively would never see it as an Rx fault at all.

The remedy you name ("explicitly mapping `timeoutSeconds <= 0` to a `TimeoutException`") is exactly what `06dfe4f5` does: the guard is `budget <= TimeSpan.Zero` and it returns **before** `pathResolver` is resolved and before `.Timeout(budget)` is ever composed, so a negative budget can no longer reach it.

Answered with a test rather than a paragraph — `ae2e6cf8` extends `RenderArea_Timeout_FaultsWithTimeoutException` to assert that `timeoutSeconds: -1` faults the **observable** with a `TimeoutException` carrying the refusal message, explicitly *not* `ArgumentOutOfRangeException`. Green locally: 9/9, `0 Warning(s) 0 Error(s)` under `-c Release -warnaserror`.

Issue #1763, opened 2026-08-17
Bake must be compiler-driven, not mesh-driven: split the BAKE (build process, no mesh) from the GATE (mesh, consumes the bake)


rbuergi:
Maintainer-directed, and named as the **first thing to change**.

## The finding

Everything in the estate still bakes by running a mesh. `mw-plugin-test`'s `PluginGateRunner`
constructs `new MeshBuilder(...).AddGraph()`, stands up an in-process mesh, imports the repo's
content, lets the MESH compile every NodeType, and `--bake-output` then collects the assemblies the
mesh produced. That is the platform's own CI bake and every satellite's bake.

It is a headless mesh rather than a serving portal, but it is exactly *"compile through mesh
nodes"* — the thing #1707 forbids. MeshWeaver.Education's gate is the extreme form (a disposable
portal compiling lazily behind browser clicks, 23 minutes in one step), but it differs in degree,
not in kind.

## The change

**Split two concerns that `mw-plugin-test` currently fuses.**

| | today | target |
|---|---|---|
| **BAKE** — produce assemblies | in-process mesh imports content and compiles it | **build process**: read sources from git, compile with `MeshWeaver.Compiler`, emit **DLL + PDB**, write the bundle |
| **GATE** — prove it works | same mesh run, same pass | mesh **consumes** the baked output (pre-baked/seeded) and asserts render + Tests areas |

The gate legitimately needs a mesh: rendering a layout area and running a `Tests` area are runtime
behaviours. Producing an assembly is not.

## Why this is now buildable

The pieces exist and were extracted for exactly this:
- **`MeshWeaver.Compiler`** (#1712) — the toolchain is its own assembly, no Graph/hub dependency.
- **`MeshWeaver.Compiler.Cli` / `mw-compiler`** (#1714) — already packaged as a dotnet tool, already
  ships in the image, already carries the surface manifest so it resolves the same framework
  identity.
- **Per-type dependency records** (#1719) and the identity model — unchanged; a compiler-driven bake
  stamps the same records so adoption keeps working.

Nothing about the OUTPUT format should change: the same bundles, the same `_complete` sentinel, the
same identity keying. Consumers must not notice.

## The real work

Source resolution without a mesh. Today the mesh performs it: NodeType discovery, `shared=@…`
include resolution, content collections, the `configuration` lambda that lives as C# inside a JSON
string field. A build-process baker has to resolve all of that from the git tree directly. That is
[...]

Issue #1707, opened 2026-08-17
Compile cleanup: MeshWeaver.Compiler, dependency-scoped rebuilds, consume-prebuilt-everywhere (maintainer-directed)


rbuergi:
# Compile cleanup: MeshWeaver.Compiler, dependency-scoped rebuilds, consume-prebuilt-everywhere (maintainer-directed)

Follow-on program under #1660, after #1687/#1696/#1705 shipped the surface identity and the CI bake
lanes. Direction from the maintainer (2026-08-17):

1. **Rebuild key = the compiler toolchain + the referenced libraries, nothing else.** For every
   compiled NodeType we must know the full dependency set — the compiler/generator toolchain AND
   the libraries the compile references. If any of them changes, rebuild; if none changed, the
   existing DLL is valid regardless of which instance produced it.
2. **Factor out minimalistic build tools** — the generator/compile pipeline leaves
   `MeshWeaver.Graph` into a small dedicated assembly (`MeshWeaver.Compiler`), so
   `FrameworkBuildIdentity.FullMvidAssemblies` pins a low-churn assembly instead of the
   highest-churn one in the repo.
3. **Any type may have a pre-compiled lib: if yes, take it; if no, generate.**
4. **There will always be code without any CI** (DB-resident user edits) — bake-your-own stays as
   the universal fallback, via the batch driver (no per-node activations), deduped by the #1706
   work-lease.
5. **The compile cache can be populated from anywhere; at INSTALL, check whether a pre-built lib
   exists and consume it** instead of compiling.
6. **Pre-bake on a new platform release NEVER involves per-node compile** — the response to a new
   framework identity is the git-checkout batch pipeline (the `meshweaver-framework-released`
   receivers: checkout → batch compile → publish bundles); portals adopt.
7. **The git PUSH path works the same as install**: on a content push (GitSync re-import), check
   whether the default DLL is present and consume it — that should almost always be the case,
   because the content repo's CI baked the same commit. Only if not present, issue a bake.
8. **Issued bakes use the dotnet build we created** — the git-staged batch build machinery
   (`mw-plugin-test --bake-output` in CI; the batch-bake / build-protocol driver in-process, under
   a build claim) — never ad-hoc per-node compiles as a side effect of the triggering operation.
9. **The factored-out compiler ships as a dotnet tool** (`dotnet tool install`) so CI lanes —
   platform AND satellites — install the compiler matching their platform version instead of
   building or downloading the platform repo's tooling.

[...]

PR #1766, opened 2026-08-17
Bake without a mesh: mw-compiler grows a `compile` verb (#1763)

rbuergi:
Closes part 1+2 of #1763. **Piece 3 (switching a CI lane over) is deliberately NOT in this PR.**

## The finding this fixes

Everything in the estate baked by **running a mesh**. `tools/MeshWeaver.PluginTester` — which is
also what the `MeshWeaver.Compiler.Cli` / `mw-compiler` package packs today — constructs
`new MeshBuilder(...).AddGraph()`, stands up an in-process mesh, imports the checkout, lets the
**mesh** compile every NodeType behind its hub scheduler, and `--bake-output` then collects what the
mesh produced. That is "compile through mesh nodes", the thing #1707 forbids, and one binary was
wearing both names — which is precisely how the mesh-driven bake stayed invisible.

## The split

| | what it is | how it runs |
|---|---|---|
| **BAKE** | a **build step** | `mw-compiler compile <root> --output <dir> [--source-sha <sha>]` — resolve NodeType sources from the git tree, compile with `MeshWeaver.Compiler`, emit **DLL + PDB**, write the bundle. No `MeshBuilder`, no `AddGraph()`, no import, no hub. |
| **GATE** | a **runtime check** | `mw-plugin-test <root>` keeps its mesh — rendering a layout area and executing a `Tests` area are genuine runtime behaviours. Producing an assembly is not. |

- **Output format unchanged**: same `BundleWriter`, same framework-identity keying, same per-type
  dependency records (#1719). `ShippedPrebuiltBundles`, `PluginBundleClient.Adopt` and
  `PrebuiltAssemblySeeder` are untouched and cannot tell the producers apart.
- **No lane switched.** The gate's own `--bake-output` still works, so lanes migrate one at a time.
- **Emergency path preserved.** A live instance with no usable artifact still compiles its own
  (#1707). That is recovery, not a build lane.

## Where the real work was: source resolution without a mesh

`NodeSet` / `NodeSetQuery` / `NodeSetCompiler` live in **`MeshWeaver.Compiler`**, i.e. inside the
full-MVID identity boundary. That placement is load-bearing: which Code nodes a compile consumes is
part of its *generated input*, so a resolver outside the boundary could change what a bake consumes
without moving the framework identity — and every portal would adopt the changed bytes silently.

Two rules keep the second implementation honest:

1. **Expansion is not re-implemented.** `CodeQueryResolver.ExpandAll` is the same call the runtime
   makes; so are the `@@`-include walk, the dedup, the executable filter, the join order, the
   skeleton generator, the `#r "nuget:"` handling and the emit. The new code *orchestrates* the
   runtime's shaping rather than copying them.
2. **Evaluation refuses what it does not understand.** Only `path:`/`namespace:`/`scope:`/`nodeType:`
   — everything `CodeQueryResolver` can emit. Free text (vector search on a real mesh), wildcards,
   alternations and any other selector make the resolution **unestablished** and the bake declines
   to compile. Same fail-loud direction as `SourceSnapshot`, for the same reason: a SHORT source set
   compiles into completely genuine-looking `CS0246`s about code that is fine (#1218).

`CodeConventions.SanitizeNodeName` moved down from Graph's `CompilationCacheService` (which now
delegates). It names the emitted assembly, so it shapes the bytes and belonged inside the toolchain.

## The equivalence pin — the actual deliverable

`BakeEquivalenceTest` bakes one content set **both ways** and asserts the producers agree on:
framework identity · bundle + node-path sets · **resolved source set per type** · **per-type
dependency records** · **emitted type-and-member surface** · PDBs present.

The fixture exercises every rule that could fork: default `Source/`+`Test/` subtree queries
including a **nested** folder, a **cross-package** `shared=@Lib/Shared/Source`, an `@@` include of a
node **no query matches**, an **executable code cell** that must be excluded, and a
`// NodeType: Scope` source the implicit `nodeType:Code` filter must exclude.

**Bytes are deliberately not compared** — see "what differs" below.

## What differs between the two paths (the most valuable output)

1. **The mesh-driven bake is not byte-reproducible, even against itself.** The mesh folds its source
   set into an `ImmutableDictionary` and emits `dict.Values` — hash-bucket order over
   per-process-randomised string hashes — and `CombineSources` joins the files in *that* order. The
   generated skeleton also stamps `// Generated at: {UtcNow}`. The compiler path sorts (query order,
   then ordinal by path), which is strictly better. The surface comparison is what proves the
   concatenation order of independent top-level declarations is irrelevant.
2. **`sourceVersions` values were never information.** The mesh stamps
   `MeshNode.LastModified.UtcTicks`, which for a freshly imported tree is the *install clock*.
   `BundleReader` skips the property and `PrebuiltAssemblySeeder` re-stamps `CompiledSources` from
   the consumer's own `CurrentSourceVersions` on adopt. The tree bake writes `0`; the **key set** —
   which is the provenance — is asserted equal.
3. **Two different source sets exist and must not be conflated.** `sourceVersions` /
   `CompiledSources` is the **raw query match**; `CollectCompileSources` then drops executable cells
   and blank files before Roslyn. The executable cell is correctly present in the first and absent
   from the second, in **both** producers. (My first draft of the test asserted the wrong one — that
   was the only red.)
4. **`SanitizeNodeName` was outside the identity boundary** while naming the emitted assembly. Fixed
   here by moving the rule down.

## Verification (local, no CI iteration)

- `BakeEquivalenceTest` — green.
- `NodeSetQueryTest` — 16 pins on the query grammar, incl. the `namespace: + scope:subtree` →
  *descendants* degradation (the default source query's shape) and the refusal behaviour.
- Full `MeshWeaver.PluginTester.Test` suite: **56/56**. `DocumentationLinkIntegrityTest`: green.
- **Real content**: both bakes run over the platform's own staged `Doc` tree (642 nodes, 4 dynamic
  NodeTypes). Normalised manifests are **identical** (node paths, dependency records,
  source-version key sets, framework identity), and all four emitted assemblies carry **identical
  identifier sets**.
- `dotnet build -c Release -warnaserror` clean for `MeshWeaver.Compiler`, `MeshWeaver.Graph`,
  `MeshWeaver.Hosting`, `MeshWeaver.Hosting.Monolith`, the tool and the test project.

## Not in this PR

- Switching `doc-gate` / `samples-gate` / `plugin-gate` / `main-cd`'s `publish-bake` to the
  `compile` verb.
- Splitting `MeshWeaver.Compiler.Cli` into its own project. ⚠️ **Note for whoever does**: the
  surface identity is computed over `FrameworkBuildIdentity.ContentSurfaceAssemblies`, whose
  canonical list is the transitive closure of `tools/MeshWeaver.PluginTester`, and
  `FrameworkBuildIdentityTest.CanonicalList_MatchesTheTesterClosure` recomputes it from *that*
  csproj. A new CLI project with a different reference closure changes the surface manifest and
  therefore the framework identity — which would make its bundles decline everywhere, silently.
  Keeping the verb in the existing project is what guarantees the identity is unchanged today.

### DIFF
--- a/src/MeshWeaver.Compiler/CodeConventions.cs
+++ b/src/MeshWeaver.Compiler/CodeConventions.cs
@@ -23,4 +23,45 @@ public static class CodeConventions
     /// alongside (not inside) their parent NodeType.
     /// </summary>
     public const string TestSubNamespace = "Test";
+
+    /// <summary>
+    /// The node path → assembly-name-safe token every dynamic NodeType compile is keyed by:
+    /// the emitted assembly is <c>DynamicNode_{SanitizeNodeName(path)}</c> and the generated
+    /// provider class is <c>{SanitizeNodeName(path)}MeshNodeProvider</c>.
+    ///
+    /// <para>🚨 It therefore SHAPES THE EMITTED BYTES and belongs inside the toolchain identity
+    /// boundary (#1707). It lived only in <c>MeshWeaver.Graph</c>'s <c>CompilationCacheService</c>
+    /// until #1763 needed it for the build-process baker; that method now delegates here, so a
+    /// mesh-driven and a compiler-driven bake of the same node can never disagree about the
+    /// assembly's name — a disagreement no consumer could see, since the bundle keys on the node
+    /// path while the ASSEMBLY carries the name.</para>
+    /// </summary>
+    /// <param name="nodePath">The NodeType's mesh path.</param>
+    public static string SanitizeNodeName(string nodePath)
+    {
+        // Replace path separators and invalid characters with underscores
+        var sanitized = (nodePath ?? string.Empty)
+            .Replace('/', '_')
+            .Replace('\\', '_')
+            .Replace(':', '_')
+            .Replace('*', '_')
+            .Replace('?', '_')
+            .Replace('"', '_')
+            .Replace('<', '_')
+            .Replace('>', '_')
+            .Replace('|', '_')
+            .Replace(' ', '_');
+
+        // Remove leading/trailing underscores and collapse multiple underscores
+        while (sanitized.Contains("__", StringComparison.Ordinal))
+            sanitized = sanitized.Replace("__", "_", StringComparison.Ordinal);
+
+        sanitized = sanitized.Trim('_');
+
+        // Ensure it starts with a letter (for valid assembly names)
+        if (sanitized.Length > 0 && !char.IsLetter(sanitized[0]))
+            sanitized = "Node_" + sanitized;
+
+        return sanitized;
+    }
 }
--- a/src/MeshWeaver.Compiler/FrameworkBuildIdentity.cs
+++ b/src/MeshWeaver.Compiler/FrameworkBuildIdentity.cs
@@ -83,6 +83,20 @@ public static class FrameworkBuildIdentity
     /// surface. <c>FrameworkBuildIdentityTest.CanonicalList_MatchesTheTesterClosure</c> recomputes
     /// the closure from the csproj graph and fails naming the drift when the tester's references
     /// change without this list following.
+    ///
+    /// <para>🚨 <b>MOVING THE BAKE/GATE CLI INTO A NEW PROJECT SILENTLY CHANGES THE FRAMEWORK
+    /// IDENTITY.</b> This list is anchored to <c>tools/MeshWeaver.PluginTester</c>'s reference
+    /// closure, and the identity is the hash over these assemblies' surface-manifest pairs — an
+    /// assembly the PRODUCING process does not reference contributes no pair and hashes as
+    /// <see cref="AbsentMarker"/>. So a tidy-up that splits <c>mw-compiler</c> (the <c>compile</c>
+    /// verb, #1763) out of the tester into its own csproj with a leaner reference list resolves a
+    /// DIFFERENT identity, and every bundle it bakes is then declined by every portal:
+    /// <c>PrebuiltAssemblySeeder.DeclineReason</c> is doing its job, the pods simply compile
+    /// everything as though no bake existed, and nothing anywhere reports a defect. The split is
+    /// perfectly doable — it just has to keep this closure intact and extend
+    /// <c>CanonicalList_MatchesTheTesterClosure</c> to assert BOTH projects' closures, so the
+    /// equality is CHECKED rather than assumed. Keeping the verb inside the tester project is what
+    /// makes the identity provably unchanged today.</para>
     /// </summary>
     public static readonly ImmutableArray<string> ContentSurfaceAssemblies =
     [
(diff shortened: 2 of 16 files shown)


rbuergi:
## Update: baked the real trees, found three more things

The fixture proved the mechanism; running the bake over **actual content** proved the rest. I baked `samples/Graph/Data` (8 packages, 103 `.cs` files, 32 dynamic NodeTypes) both ways and diffed. Two bugs in this PR, one pre-existing defect in the repo.

### 1. Enums did not deserialize — 28 NodeTypes silently unbuildable
Enums persist as strings (`"compilationStatus": "Ok"`), and `JsonSerializerDefaults.Web` does not convert them. Every NodeType carrying a compile stamp failed to materialise — all of ACME, Cornerstone, FutuRe, Northwind. Fixed with the **framework's own** `EnumMemberJsonStringEnumConverter` (not a stock `JsonStringEnumConverter`) plus the hub's comment/trailing-comma tolerances.

This is exactly the silent class the issue warns about, and it only failed loudly because unmaterialisable files were fatal at the time — which brings us to (3).

### 2. No NuGet resolver — the one remaining divergence
`MathDemo/Matrix` declares `#r "nuget:MathNet.Numerics, 5.0.0"`. The bake refused rather than emitting an assembly missing the directive's types (safe), but refusing isn't the answer — the runtime resolves them, so the bake must. `NuGetAssemblyResolver` needs nothing from a mesh, so it's now wired. **23/24 → 24/24.**

### 3. Fatal-on-unmaterialisable-file was wrong — and it exposed a real content defect
A file that won't parse is now **reported, not fatal**, because the runtime skips it too (`PackageInstaller`: `No parser for node-repo file X; skipped`). Refusing made the bake resolve a *smaller* tree than the mesh — the same equivalence break this PR exists to prevent, in the opposite direction.

What surfaced it is worth stating on its own:

> 🚨 **`samples/Graph/Data/PensionFund/*.json` carry a UTF-8 BOM.** The mesh drops **62 of that package's 72 files** and gates **zero NodeTypes** there. The samples gate has never actually gated the PensionFund tree, and its bake ships nothing for it.

Deliberately **not** "fixed" here by stripping the BOM: that would make the bake compile content the mesh never imports. It's a content fix, and the code comment says so.

## Equivalence on real content

Measured from the **same binaries** — the framework identity moves whenever `MeshWeaver.Compiler` changes, so an earlier cross-build diff was meaningless; both bakes are re-run after every toolchain edit.

| tree | result |
|---|---|
| `samples/Graph/Data` — 7 bundles, 24 assemblies | node paths, dependency records, source-version key sets, framework identity → **identical** |
| same, emitted assemblies | CLI metadata `#Strings` name heaps → **identical, 24 of 24** |
| `src/MeshWeaver.Documentation/Data` — 642 nodes, 4 NodeTypes | normalised manifests → **identical** |

The 8 remaining RED types match the mesh's own failures (the known-debt set in `.github/samples-gate.allow`) — reproduced, not introduced.

**Note on method:** `strings(1)` is not usable for this comparison at scale — the CodeView `RSDS` signature and stray IL bytes masquerade as identifiers and produced 20 phantom diffs. The real check reads the CLI metadata `#Strings` heap.

`MeshWeaver.PluginTester.Test`: 57/57. Release `-warnaserror` clean.

Issue #1778, opened 2026-08-17
Self-update has no floor between rolls — publication frequency is now portal-restart frequency


rbuergi:
## What is missing

Since #1773 an install checks for a new image **per publication event** and rolls whenever one is
available. Nothing bounds how often that happens:

- `UpdatePolicyKind` selects a **channel** (`Continuous` / `Stable` / `None`), not a cadence.
- `SelfUpdateOptions.EventCoalesceWindow` (1 min) coalesces a *burst* of events into one check — it
  does not pace successive rolls an hour apart.
- The knob that used to pace them, `SelfUpdate__PollInterval: "1.00:00:00"` in
  `deploy/aks/values.aks.yaml`, binds to an option that no longer exists. It was **inert from the
  moment #1773 merged**, and silently: an operator reading that file sees a live daily throttle.

So **publication frequency is now portal-restart frequency, one for one.**

## Why it matters, with the numbers

That daily value was a measured decision (2026-08-10): every roll invalidated the whole NodeType
compile cache — the framework MVID changes on every CI image — costing a full sequential rebuild
(~2.4 s/type, ~10 min on the largest mesh) plus a degraded warm-up window. "One roll per day keeps
fixes flowing daily at a quarter of the compile cost."

**Half of that rationale is genuinely obsolete** and should not be used to argue for slow delivery:
adopt-before-compile removed the rebuild in prod — 80 compiles / 64.8 s became **0 compiles, 84
assemblies adopted, 32.1 s**. What remains is the part adoption cannot remove: **each roll restarts
the pod**, dropping live Blazor circuits.

## Why this is blocking something concrete

It is the reason CD's reconcile tick is hourly and not faster. The maintainer rejected the 3 h
publish tail ("3 hour window is not acceptable"), and the reconcile cadence — not the batch window —
is what sets that tail. Hourly was chosen as the fastest tick defensible while a roll is unpaced;
tightening it further multiplies restarts one-for-one.

## What a fix looks like

A floor between rolls, expressed where the cost lives (the install), not by slowing CD:

- an option on `SelfUpdateOptions` (e.g. `MinRollInterval`), bound from `SelfUpdate:*` like the rest,
  and set explicitly in `deploy/aks/values.aks.yaml` in place of the key this issue deletes;
- state that survives a restart, since a successful roll **restarts the process** — so an in-memory
  "last rolled at" is always empty exactly when it is needed. Two candidates: the age of the
[...]

PR #1779, opened 2026-08-17
fix(cd): publish within the hour, not within three — the reconcile tick was the tail, not the batch window

rbuergi:
Closes the maintainer's complaint from 2026-08-17: *"3 hour window is not acceptable"* — and the
first thing to say is that **the window was not what made it three hours.**

## The misdiagnosis the naming invites

A batched merge leaves main's HEAD without a complete image set. That is *precisely* the state the
reconciler exists to close, so it publishes on its next **tick** — whatever the window is set to.

| knob | what it actually controls |
|---|---|
| `CD_BATCH_WINDOW_MINUTES` | how many merges ride one image (coalescing) |
| the `schedule:` cron | **how long a batched merge waits** — the tail |

Both were 3 h, which is the only reason the window looked like the culprit. Raising the window does
not delay a merge by the window; slowing the tick does.

## The change

`23 */3 * * *` → **`23 * * * *`**, window 180 → **60 min** (repo variable, already applied).
Worst case merge → published image is now ~1 h + the ~20 min build, typically ~30 min.

Runner cost: a healthy tick is one ~40 s job, so 24 ticks/day ≈ 15 min/day, up from ~5. The failing
case does **not** scale with the cadence — the 3-attempts-per-commit budget is per COMMIT, so a
broken HEAD still costs 3 attempts however often we tick, and later ticks cost the ~40 s probe.

## 🚨 Why hourly and not faster — the finding that came out of this

**Since #1773, publication frequency IS portal-restart frequency, one for one.** Self-update is
event-driven with nothing pacing successive rolls: `UpdatePolicyKind` picks a *channel*
(Continuous/Stable/None), not a cadence, and `EventCoalesceWindow` (1 min) coalesces a burst into one
check rather than spacing rolls apart.

The knob that used to pace them — `SelfUpdate__PollInterval: "1.00:00:00"` in
`deploy/aks/values.aks.yaml`, holding prod to **one roll/day** — binds to an option #1773 deleted. It
has been inert since that merge, and silently: it still reads as a live throttle to anyone grepping
the file. That is the exact silent-no-op `SelfUpdateConfiguration`'s own binding comment warns about.

So this PR deletes the dead key rather than leaving it to lie, and keeps what it expressed as prose,
because half of it is now obsolete and half is not:

- **obsolete**: "every roll costs a full sequential NodeType rebuild (~10 min)" — adopt-before-compile
  measured in prod at 80 compiles / 64.8 s → **0 compiles, 84 adopted, 32.1 s**;
- **not obsolete**: each roll still **restarts the pod**, dropping live circuits, and nothing bounds
  how often that happens.

Filed as **#1778** with the mechanism sketch (the floor needs restart-surviving state, since a
successful roll restarts the process — process uptime is the tempting answer and is wrong for crash
recovery). Hourly is the fastest tick defensible until that lands; tightening further multiplies
restarts one-for-one.

## Docs

`AGENTS.md`, `ContinuousDeliveryContract`, `DeploymentOptions`, `DeployingPluginChanges`,
`ReleaseStrategy`, `CHANGE-PROPAGATION` — cadence updated, plus a correction #1773 left behind: both
AGENTS.md and the contract described installs updating "on their 6 h poll
(`SelfUpdateOptions.PollInterval`)", naming a symbol that no longer exists. They now describe the
event-driven check, with `RetryInterval` correctly framed as the faulted-watch backstop.

## Verification

- `main-cd.yml` and `values.aks.yaml` parse; cron reads `23 * * * *`; the `SelfUpdate__PollInterval`
  key is gone from the portal config map.
- `MeshWeaver.Documentation.Test` link-integrity + What's New integrity: **2/2 passed**, Release
  build `0 Warning(s) 0 Error(s)`. The new What's New node was confirmed present in the rebuilt
  `MeshWeaver.Documentation.dll` first — an embedded-resource doc edit passes vacuously against a
  stale assembly.
- Repo variable `CD_BATCH_WINDOW_MINUTES` is already `60`; merging this makes the cron agree with it.


rbuergi:
## ⏸️ Holding this merge until the roll floor (#1778) lands — deliberately

Raised by the session that owns self-update, and it is right: **hourly is not unsafe in itself, but
landing it before the floor widens an exposure that already exists.** Prod has no pacing at all right
now (since #1773), so tripling publication frequency triples portal restarts against zero throttle.
Nothing here is urgent enough to spend that window on.

#1778 is being implemented now — `SelfUpdateOptions.MinRollInterval`, with restart-surviving state as
a **Deployment annotation** stamped at patch time rather than process uptime (a pod that crashes and
comes back on an OLD image reads an OLD annotation, so the floor is already elapsed and it rolls at
once; uptime would make exactly that case wait). That PR lands first; this one rebases onto it — it
deletes the `SelfUpdate__PollInterval` block where the new key goes — and merges immediately after.

**One thing that decides whether this PR means anything**, flagged there too: if the floor is set to
restore the prior 1/day, publication being hourly changes nothing end-to-end — a merged fix still
reaches production the next day, which is *worse* than the three hours that prompted this. The floor
is choosing the AUTOMATIC cadence, not the fastest possible one (`kubectl rollout restart` and
`gh workflow run main-cd.yml --ref main` remain the immediate paths), so the trade is:

| floor | restarts/day | fix visible after merge |
|---|---|---|
| 24 h | ≤1 | next day (what prod had) |
| 4–6 h | ≤4–6 | same working day |
| 1 h | ≤24 | ~1.5 h |

That number is a product call, not an implementation detail.

rbuergi:
Correction on my own hold: I had already set the live repo variable `CD_BATCH_WINDOW_MINUTES` to 60 before agreeing to hold, which enables part of the same exposure — a smaller window means more merges publish immediately, i.e. more restarts, against no floor. **Reverted to 180.** It goes to 60 at merge time, together with the cron, so the two knobs move as one change rather than one of them leaking out ahead of the PR.

PR #1780, opened 2026-08-17
fix(#1778): a floor between automatic rolls, since a roll is a pod restart

### LINKED ISSUE #1778: Self-update has no floor between rolls — publication frequency is now portal-restart frequency

rbuergi:
Closes #1778. **This is my regression from #1773** — I am fixing it, not reporting it.

## What I broke

#1773 renamed `SelfUpdateOptions.PollInterval` and removed the recurring poll. That left `SelfUpdate__PollInterval: "1.00:00:00"` in `deploy/aks/values.aks.yaml` — a deliberate **one-roll-per-day production throttle** — binding to nothing, while still reading like a live throttle to anyone opening the file.

The cause is specific: I grepped `.cs` for the renamed option, and grepped the charts only for the option I *deleted*. **A rename is exactly as breaking as a deletion for a `__`-bound key.**

Without pacing, **publication frequency is portal-restart frequency, one for one** — and each restart drops the live circuits of everyone using the portal.

## The floor

- `SelfUpdateOptions.MinRollInterval`, bound from `SelfUpdate:*` like the rest, **default 1 hour**.
- Checked at the roll decision: inside the floor, log and defer.
- A deferred roll is re-decided by the **next publication event**, exactly like an availability hold — the floor introduces **no timer**.

## 🚨 Restart-surviving state, and why not uptime

A successful roll **restarts the process**, so an in-memory "last rolled at" is empty exactly when the floor needs it. `IDeploymentUpdater` grows `LastRolledAtAsync`, answered in `MeshWeaver.SelfUpdate.Aks` by an **annotation the patch itself stamps on the Deployment**, and "never" by the platform's `DetectOnly` fallback (an install with no patcher cannot roll anyway).

**Not process uptime** — the tempting third option, wrong for **crash recovery**: a pod that comes back on an OLD image has a young process but an old deployment, and uptime would make it wait out a floor it satisfied hours ago, precisely when it most needs the newer image. The annotation gives the right answer. `AnOldStampRollsImmediately_TheCrashRecoveryCase` pins exactly that.

## 🚨 The value is 1 h, deliberately not the 24 h prod had

Restoring the prior number is **not** the neutral default it looks like. With hourly publication, a daily floor means a merged fix reaches prod the **next day** — strictly worse than the 3 h window the maintainer rejected. "Restore prior behaviour" would silently answer their complaint with "no".

Half the 2026-08-10 rationale for 24 h is also obsolete: adopt-before-compile took a prod roll from **80 compiles / 64.8 s to 0 compiles / 32.1 s**, leaving the restart cost as the only live argument — which is what this bounds.

| floor | restarts/day | a merged fix is visible |
|---|---|---|
| 24 h | ≤1 | next day (what prod had) |
| 6 h | ≤4 | same working day |
| **1 h** | ≤24 | **~1.5 h after merge** |

`values.aks.yaml` carries that trade in prose, so the number can be changed without re-deriving it. And the floor paces only the **automatic** cadence — `kubectl rollout restart` (startup pass takes the newest image) and a `main-cd.yml` dispatch both still apply an urgent fix immediately.

## Timing, verified rather than assumed

Prod runs `3.0.0-rc4.ci.4147` (commit `324e61a`), an **ancestor of #1773** — so the portals still carry the old poller and are still on the 24 h throttle. **The unpaced behaviour is not live yet**; it begins with the first post-#1773 image prod rolls onto, which is the deadline this wants to be inside.

## Verification

- Whole solution `-c Release -p:CIRun=true -warnaserror` → `0 Error(s)`.
- **18/18** self-update tests, including four new ones: fresh roll defers, old stamp rolls immediately (crash recovery), never-rolled is not held, zero floor opts out.

### DIFF
--- a/deploy/aks/values.aks.yaml
+++ b/deploy/aks/values.aks.yaml
@@ -244,23 +244,31 @@ config:
     # cadence and your rollback headroom.
     AssemblyCache__Retention__Delete: "false"
     # ---- Platform self-update cadence --------------------------------------
-    # The poller's built-in default is 6h, which is why a freshly published image
-    # could sit unpicked for hours and look like "self-update is broken" (memex,
-    # 2026-08-03: running ci.1672 while ci.1674 had been in ACR since 07:34).
+    # The check is EVENT-DRIVEN (#1773): one pass at startup, then a check per
+    # published build of the platform or of any module this environment deploys.
+    # There is no polling interval any more — SelfUpdate__PollInterval was renamed
+    # and this key sat here inert, reading like a live daily throttle while binding
+    # to nothing (#1778).
     #
-    # DAILY, deliberately (2026-08-10, measured): every roll invalidates the whole
-    # NodeType compile cache — the framework MVID changes on every CI image — so a
-    # roll costs a full sequential rebuild (~2.4 s/type; ~10 min on the largest
-    # mesh) plus a degraded warm-up window while the 3 s enrichment probe is
-    # starved. CI publishes ~44 images/day; polling faster than the rebuild is
-    # worth just multiplies that cost. One roll per day keeps fixes flowing daily
-    # at a quarter of the compile cost and a quarter of the degraded windows.
-    # For an URGENT fix, don't shorten this — `kubectl rollout restart` the
-    # portal: the poll fires on startup (StartWith(-1L)) and picks the newest
-    # image immediately. Binding only works from the image that added
-    # SelfUpdateOptions.SectionName — before that this key is inert, because
-    # AddSelfUpdate() constructed its options directly.
-    SelfUpdate__PollInterval: "1.00:00:00"
+    # What this number now sets is the floor between two AUTOMATIC rolls. It is a
+    # RESTART budget, not a delivery speed: a roll restarts the pod and drops every
+    # live circuit on it. The trade, so this can be changed by editing one number:
+    #
+    #     24h  ->  <=1 restart/day,     a merged fix reaches prod the NEXT DAY
+    #      6h  ->  <=4 restarts/day,    same working day
+    #      1h  ->  <=24 restarts/day,   ~1.5 h after merge   <-- current
+    #
+    # 1h matches CD's reconcile tick, so hourly publication actually means hourly
+    # delivery. The former 24h is deliberately NOT restored: it predates
+    # adopt-before-compile (a prod roll went from 80 compiles / 64.8 s to 0
+    # compiles / 32.1 s), so the compile-cost half of its 2026-08-10 rationale is
+    # gone, and "tomorrow" is the answer that prompted the faster tick in the first
+    # place. What remains true is the restart cost, which is what this bounds.
+    #
+    # Neither extreme delays an URGENT fix: `kubectl rollout restart` takes the
+    # newest image immediately (the startup pass), and `gh workflow run
+    # main-cd.yml --ref main` bypasses the batch window.
+    SelfUpdate__MinRollInterval: "01:00:00"
     Mcp__BaseUrl: "http://memex-portal-service:8080"
     # ---- AI model providers (templated by the chart) ------------------------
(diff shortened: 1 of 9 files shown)


rbuergi:
All four fixed in 608d43887 — every one was substantive, and the `Task.Delay` catch is the best of them.

**The fixed delay** was me breaking a rule I have been citing all day: this repo's guidance is to wait on the condition, never the clock. Split in two — `ExpectRollAsync` waits on the `Patched` stream with a timeout, and only `ExpectNoRollAsync` keeps a bounded delay, which is the sanctioned *"confirm nothing happened"* shape since a deferral has no positive signal to wait for. Suite time **15 s → 7 s**, and the roll cases are no longer timing-sensitive under CI load.

**The zero-floor stamp read** is gone: `LastRolledAtAsync` is a Kubernetes GET in the AKS implementation, and its answer cannot change the decision when the floor is disabled — an API call and a failure point on the happy path for nothing.

**The doc comments** did have stray blank lines inside them, detaching them from the member. The build tolerated it, so thank you for catching what `-warnaserror` did not.

Whole solution `-c Release -warnaserror` → `0 Error(s)`; 18/18 self-update tests.

PR #1783, opened 2026-08-17
fix(cd): a 503 on a cosmetic label blocked publication entirely

rbuergi:
## What happened

Run [32053611859](https://github.com/Systemorph/MeshWeaver/actions/runs/32053611859), 2026-08-17. The `gate` job had resolved everything and was **about to publish**:

```
Required check 'Consolidate test results' on 94bbab80… → completed/success (commit is 19 min old)
memex-portal-ai:94bbab8 is MISSING from ACR — main 94bbab8 has an INCOMPLETE image set
   (…and portal-next, migration, mw-plugin-test)
```

Then:

```
HTTP 503: No server is currently available to service your request. (…/labels)
##[error]Process completed with exit code 1.
```

`gh label create ci-failure --force` — ensuring a **cosmetic** label, idempotently — hit the ongoing
GitHub API incident, and under `set -euo pipefail` that killed the Decide step. **No image was
published**, every self-updating install stayed on the previous one, and the failed attempt consumed
one of the 3 heal attempts for that commit.

## Why this is the right fix and not a band-aid

A label is decoration on the ledger, **not** the ledger. The ledger is the marker COMMENT, and that
write stays fatal — an attempt that is not written down is an attempt that can be repeated forever.

This is the exact test AGENTS.md already applies to the green-marker push and the TRX reporter:
*what does a failure here hide?* For a label: nothing. Losing it costs a colour, never correctness.
The failure is still visible as a `::warning::`.

Applied at **both** label sites and **not** extended to any `gh issue create` / `gh issue comment`:

| site | why |
|---|---|
| `gate` → Decide | the publish path — this is the one that cost a publication |
| the alert job | a 503 there would swallow the very alert it decorates; if the label genuinely does not exist, `gh issue create --label` still fails loudly |

## Verification

Ran the real line under `set -euo pipefail` against a stubbed `gh` that 503s on `label`:

```
--- with the fix ---
::warning::could not ensure the ci-failure label (continuing — the ledger is the comment, not the label)
REACHED THE LEDGER — publication would proceed          exit=0
--- without the fix ---
                                                        exit=1
```

Workflow YAML parses; both patched steps pass `bash -n`.

No What's New entry — CI-internal, no user-visible change.

Note this is a **separate** concern from #1779 (which changes the reconcile cadence in the same
file); they touch different regions and #1779 is currently held on the roll floor (#1778), while this
one should not wait — it blocks publication on any GitHub API blip.


rbuergi:
CI red, and **not from this diff** — which is one comment and one `||` clause in a workflow file and cannot reach a test.

Shard 3 failed on `CodeCellOutputCaptureTest.Progress_Of_Succeeded_Activity_Renders_Output_For_Anonymous_VUser`. Rather than re-run it (which would sample a race and hide it), I pulled the diagnostic a previous session added for exactly this failure, and it resolves the ambiguity it was built for:

```
FAILURE-TIME NODE STATE: Status=Running, Messages=1, End=.
Node is NOT Succeeded although this test already observed Succeeded before subscribing
⇒ a later write regressed the node (write-order bug).
```

So a Succeeded activity was written back to Running with `End` cleared — a real defect, filed as **#1784** and being root-caused now. The render pipeline is exonerated: it correctly rendered the regressed state.

**This PR is held until #1784 lands**, because merging red is forbidden and re-running to obtain a green is equally forbidden. It merges on the first genuinely green run after that.

PR #1791, opened 2026-08-17
Show which build — and which instance — is serving the session

rbuergi:
The portal header gains a chip naming the running build. Hover it and it also names the serving
process — the pod, under Kubernetes. When a newer build exists the chip switches to it, grows an
arrow, and clicking it hard-reloads the session onto whatever instance is serving now.

## Why it renders in every state

It is always on screen rather than appearing only when an update is pending, and that is the point:
half the question is *"did I arrive?"*. An indicator that shows up to announce an update can tell
you one is available but never that you are now on it. Because the running build is on screen before
and after, a refresh that lands on a new instance is visible as the value changing — which is what
you want to confirm ahead of something long-running like an agent thread round.

The tooltip names the instance as well as the build because two replicas can run the same build, so
the version alone cannot settle whether the session moved.

## A held update reads differently and offers no refresh

Per `PlatformUpdateAvailability.UpdateHeld` a hold must never be silent — but refreshing cannot
clear one, and offering the button anyway would teach the user to click at a problem they cannot fix
from the header. It says so and links to About; the reason stays on the admin-gated Updates tab.

## Two things the shape is load-bearing for

- **The auth gate is a wrapper component, not an `@if`.** `PortalLayoutBase` renders the
  `HeaderLinks` slot **outside** its `AuthorizeView` (next to the deliberately-public theme toggle),
  so anything in that slot reaches anonymous visitors — and public pages exist. Wrapping means the
  gate covers the **read** as well as the markup: an unrendered child never initialises, so an
  anonymous page load never opens a System-impersonated read of the Admin partition just to decide
  it has nothing to show.
- **The verdict is re-read on navigation, in a `SerialDisposable`.** The layout survives every SPA
  navigation, so without this the chip would read "as of when the tab was opened" for the whole
  session. Navigation is the right trigger *because* `Observe` is one-shot — a live subscription
  would hold a System-impersonated read open for the lifetime of the view, which its own docs rule
  out. Serial rather than composite so a session that navigates a hundred times holds one
  subscription, not a hundred finished ones.

## Notes

Wording lives in `PlatformUpdateChip.Describe` so it is testable without a hub or a rendered
circuit — the same reason `AboutSettingsTab.UpdateStatusMarkdown` takes its localizer as a function.
It reuses the About tab's existing vocabulary; the two genuinely new keys (`about.instance`,
`ui.updateRefreshHint`) are mirrored into `clients/react/src/i18n` per the drift guard.

I did **not** wire the grain-dispose half of "recycle and refresh" into this button — see the
comment below for the two reasons.

## Verification

| Gate | Result |
|---|---|
| `Memex.Portal.Shared` — Release, `-warnaserror` | 0 errors, 0 warnings |
| `Memex.Portal.Shared.Test` — Release, `-warnaserror` | 0 errors, 0 warnings |
| New chip tests + About/update-status neighbours | 21 passed |
| `LocalizationTest` | 56 passed |
| `WhatsNewEntryIntegrityTest` + `DocumentationLinkIntegrityTest` | 2 passed (entry confirmed embedded in the rebuilt DLL) |
| React `localize.test.ts` drift guard | 15 passed |
| Monolith boot smoke on :5099 | boots, serves HTTP 200, no startup exception |

The boot smoke cannot observe the chip itself: the portal does not prerender the header, so the
markup only exists once a SignalR circuit connects, and no Playwright browser is installed here.
The in-circuit render is unverified by test — that gap is real and I am naming it rather than
implying the smoke covered it.


rbuergi:
**On "recycle the grain" — I deliberately left that half out, and want the reason on record.**

The ask was a refresh button that recycles the grain *and* refreshes. This button only reloads. Two
independent reasons:

1. **For a platform update it is a no-op.** A roll restarts the portal (and, on the distributed
   host, the silos), so every grain is fresh by construction. The stale-grain case that
   `RecycleLayoutArea` exists for is a NodeType recompile, which is a different event from the one
   this chip reports.
2. **It would be an ungated dispose of shared state.** `DisposeRequest` to a node hub tears it down
   for *everyone* on that node, which is why the node menu's Recycle item is gated on
   `Permission.Update`. A header button visible to every signed-in user has no such gate, so wiring
   it here would let a read-only user recycle a node's hub from any page.

The gated recycle already exists in the node menu for the case that genuinely needs it. If you want
it on the chip too, the honest shape is a permission check against the current node before the
dispose — happy to add that, but it is a different change from "show me which instance I am on".

Issue #1786, opened 2026-08-17
13 NodeTypes are parked at CompileError on memex-cloud (2 on memex) — configuration lambdas reference types their source query never includes


rbuergi:
Found while stabilising prod on 2026-08-17. **Not a new regression — a standing condition** that nothing has been ticketing (see the log-watcher issue filed alongside this).

## What is broken

| portal | NodeTypes parked at CompileError |
|---|---|
| **memex-cloud** (memex.meshweaver.cloud) | **13** |
| **memex** (memex.systemorph.com) | 2 — `BinaryClickerV2/BinaryToggle`, `Edu/Course` |
| **atioz** (customer prod) | **0** ✅ |

The 13, with the first error of each:

| node | error |
|---|---|
| `Northwind/Product` | CS0246 `Supplier`, `Category` not found |
| `Northwind/AnalyticsCatalog` | CS0103 `OrderViews`, `SalesViews` do not exist |
| `FutuRe/{AmericasIns,EuropeRe,AsiaRe}/LineOfBusiness` | CS0246 `LineOfBusiness`; CS1061 `AddLineOfBusinessLayoutAreas` |
| `FutuRe/{AmericasIns,EuropeRe,AsiaRe}/TransactionMapping` | CS0246 `TransactionMapping` |
| `Doc/DataMesh/SocialMedia/{Post,Profile}` | CS0246 `SocialMediaPost`/`SocialMediaProfile`, `Platform` |
| `SocialMedia/{Post,Profile}` | same |
| `Doc/Architecture/BusinessRules/Cession` | CS0246 `CessionData`, `Cashflow` |

All are under the `MeshWeaver/samples/Graph/Data/**` import.

## One shape, three causes — and they need different fixes

**Every one of these is "the `configuration` lambda references a type its source query never included".** The compiler says so explicitly for the worst case:

```
Matched Code nodes (0):
  (none) — the configuration lambda cannot reference types because no source files were
  included. Check that your Source Code nodes exist and that the NodeType's `sources` list
  points at them.
```

But the *reason* the query comes up short differs, and lumping them would produce one wrong fix:

1. **Stale orphans.** `samples/Graph/Data/SocialMedia/**` **does not exist in the MeshWeaver repo at all** — SocialMedia moved to its own repo. The mesh still holds the NodeTypes, so the import adds and updates but never retires what the source tree dropped. These should be deleted, not fixed.
[...]

PR #1796, opened 2026-08-17
fix: red-log incidents split by the fault's own text; a truncated window is reported

### LINKED ISSUE #1787: Log-watcher collapses thousands of red lines into 1-3 fingerprints, so 13 prod compile failures were never ticketed
Found while stabilising prod on 2026-08-17: **13 NodeTypes have been parked at CompileError on memex-cloud with no issue filed for any of them** (#1786). The watcher is alive and reading Loki — it is the ticketing that never happens.

## Measured

`mw-log-watcher` is healthy and polling (`Loki: 2934 line(s) in memex-cloud for [19:27:30, 19:28:37)`, HTTP 200 throughout). Its own log shows what it does with what it reads:

```
memex-cloud: 1 distinct fingerprint(s) from 5000 red line(s)
memex-cloud: 1 distinct fingerprint(s) from 4752 red line(s)
memex-cloud: 1 distinct fingerprint(s) from 3894 red line(s)
memex-cloud: 1 distinct fingerprint(s) from 2309 red line(s)
memex-cloud: 3 distinct fingerprint(s) from 2232 red line(s)
```

Thousands of red lines per window collapse to **one to three fingerprints**. Meanwhile the same namespace contains at least these distinct, independently-actionable failures:

- `MeshWeaver.Graph.Configuration.MeshNodeCompilationService` — 13 different nodes, different compiler errors each
- `MeshWeaver.Graph.CompileWatcher` — the PARKED notices
- `crit: MeshWeaver.Hosting.Orleans.RoutingGrain` — routing back-pressure, deepest per-destination queue 33
- `fail: Polly` — recurring on a ~10 s cadence
- `fail: MeshWeaver.Graph.ActivityTracking`
- `[MonotonicWriteGuard] CONFLICT on Admin/Partition/Doc` — a `Version=0` write losing to durable `Version=4`

The stated contract is *"every `fail:`/`crit:` becomes exactly one triaged issue"*. What is happening is closer to "an entire namespace becomes one issue", so distinct defects are invisible.

## Two separate problems, and the second is worse

1. **The fingerprint is too coarse.** Whatever it keys on is collapsing unrelated failures. A fingerprint that maps 13 different nodes with 13 different compiler errors onto one ticket cannot support "one issue per error".
2. **`5000 red line(s)` is a CAP, not a count.** Several windows report exactly 5000 — so the query is truncating, and anything past the limit is not merely un-ticketed, it is unseen. A dominant noisy source can therefore starve every other error out of the sample entirely, which is the failure mode where the *quiet* defect is the one that matters.

## Why this is the highest-leverage of today's findings

Every other issue filed today was found by hand-reading pod logs. This is the component whose job is to make that unnecessary — so while it is degraded, the estate's error surface is only as good as whoever happens to look. Fixing it is what makes the rest self-reporting.

Suggested shape (not prescriptive): fingerprint on the *logger category + normalised message*, keep the per-node identity in the ticket body, and treat hitting the query limit as a reportable condition in its own right rather than a silent truncation.

Related: #1786 (the 13 unticketed compile failures), and the red-log ticketing design.

rbuergi:
Fixes #1787.

Measured on `memex-cloud`, 2026-08-17: thousands of red lines per window collapsed onto **1–3
fingerprints**, and **13 NodeTypes parked at `CompileError` were never ticketed** — #1786 had to be
filed by hand. Two separate defects.

## 1. The fingerprint was too coarse

The identity was `(top app frame | log site) + exception type`, and nothing else. Every one of the 13
compile failures throws `CompilationException` from `EmitPipeline.EmitCompilationToDirectory`, so 13
independently-actionable compiler errors were **one** ticket.

The identity now also covers **the fault's own masked text** — the *exception's* message when the
burst carries one, else the logged message:

| Part | Value |
|---|---|
| WHERE | top application frame, or `(category, eventId)` with no frame *(unchanged)* |
| WHAT | exception type, simple name *(unchanged)* |
| **WHICH** | **masked exception message, or masked log message with no exception** *(new)* |

Choosing the *exception's* message rather than the reporter's prose is what keeps #1170/#1171 folded:
two catch sites word their own messages differently ("Error during shutdown of hub …" vs "Hub …
disposal faulted") and quote the **same** exception message.

`Normalize` gains the rule that makes prose safe to hash at all: **a token the message itself uses as
a path segment is masked everywhere in that message**. That reads the subject *out of* the message
instead of guessing at its position — which is exactly how round 4 fanned out ~22 ways on
`[PluginGating] Chess: … rewrote Chess/_Access/Public_Access`.

### What it does with the two real cases

| Input | Result |
|---|---|
| **3,894 lines of the SAME error** | **one** incident, `Occurrences = 3894` — everything per-occurrence (paths, guids, counts, elapsed times) is masked before hashing |
| **13 lines of 13 DIFFERENT errors** | **one incident per distinct failure shape**, each with a distinguishable name. Two nodes failing *identically* still share one ticket and list each other in the evidence — that is one defect with two instances, not two defects. |

### The floor under "too fine"

Masking cannot anticipate every message shape. A log site that opens more than
`MaxVariantsPerSite` (default **20**) distinct fingerprints **in one window** folds onto ONE
site-level incident carrying `Variants = N`, so the ticket reads "this site produced N shapes and the
masking rule needs a case" rather than burying a human. The default deliberately sits between the two
numbers production has produced: **13 stays 13**, the **~50-way 2026-08-09 fan-out folds**. Every
fingerprint it produces is stable, so recurrences still dedup across windows.

## 2. `5000` was a CAP, not a count — and said so nowhere

Truncation was already handled *correctly on the cursor* (it resumes at the last line actually read,
so the remainder is deferred, not dropped) and reported **only as a `LogWarning` in the watcher's own
pod log** — the one place this subsystem exists to stop humans from having to grep.

Every way the watcher can fail to read a window is now a `LogIncidentReport` on the normal ingest
path (dedup per namespace, no timestamps in the fingerprint):

| Condition | Severity | Meaning |
|---|---|---|
| zero lines over a long, continuously-watched window | Critical | *(existing)* the store lost that stretch |
| **query came back at `QueryLimit`** | **Error** | the window was NOT fully read; while it lasts a noisy source crowds quieter errors out of the prefix that gets read |
| **cursor floored by `MaxCatchUp`** | **Critical** | the only path that **loses** evidence outright — that stretch will never be read |

🚨 **The limit was deliberately NOT raised.** A higher cap moves the ceiling; the finding is that one
namespace out-talks its watcher, and the actionable number the report carries is the **backlog** —
because a backlog that keeps growing ends at the `MaxCatchUp` floor, which is the row that loses data.

The per-window summary also stopped conflating counts. It read
`"1 distinct fingerprint(s) from 5000 red line(s)"` with `5000` bound to the **total** line count (the
query is unfiltered), so it looked like 5000 errors collapsing when it was 5000 lines of mostly
`info:`. Now:

```
memex-cloud: 5 distinct fingerprint(s) from 7 red burst(s) (2934 line(s) read)
memex-cloud: 3 distinct fingerprint(s) from 41 red burst(s) (5000 line(s) read — TRUNCATED at the query limit)
```

## Tested on real captured data

`ProdRedLogFixtureTest` runs on **verbatim lines pulled from the `memex-cloud` portal pod** on the day
this was measured (`test/MeshWeaver.Observability.Test/Fixtures/memex-cloud-2026-08-17-red-bursts.txt`)
— three real compile failures, the real `Polly` timeout, the real repeated health-check. Every earlier
revision of this identity function passed hand-written input and failed production, so the regression
test is the production bursts. It asserts that the three compile reports share category, event id,
exception type **and** top frame — the whole of the old key — and are still three incidents.

145/145 green in `MeshWeaver.Observability.Test`; `MeshWeaver.Documentation.Test` green (doc + What's
New entry).

## Notes

- **Existing incident nodes keep their old fingerprints.** After this deploys, the collapsed
  compile incident stays as it is and the per-error incidents open fresh. That is inherent to changing
  an identity function; the old node can be suppressed or closed by hand.
- What's New entry included (`Category: Fix`): an admin browsing `Admin/_LogIncident` goes from one
  catch-all row to one row per defect, which is user-noticeable.
- `LogTriage`'s agent instructions were updated for the two new incident fields (`normalizedDetail`,
  `variants`).

### DIFF
--- a/content/ai/Agent/LogTriage.md
+++ b/content/ai/Agent/LogTriage.md
@@ -21,10 +21,18 @@ an engineer can pick up cold — then hand it to the filer. You never open the i
 The incident node is your `MainNode` — **read it first** with `get`. It carries the evidence:
 
 - `category` — the .NET log category (e.g. `MeshWeaver.Data.MeshDataSource`)
-- `normalizedMessage` — the message with ids/guids/paths masked (this is what was fingerprinted)
+- `normalizedMessage` — the logged message with ids/guids/paths masked
+- `normalizedDetail` — **what was fingerprinted**: the exception's own message when the burst carried
+  one, otherwise the logged message. This is the text that makes this incident distinct from its
+  neighbours at the same log site — start here.
 - `exceptionType`, `topFrame` — the exception and the top application stack frame, when the burst had them
 - `occurrences`, `firstSeen`, `lastSeen`, `namespace`, `pods` — how much, how long, and where
-- `samples` — verbatim log lines
+- `samples` — verbatim log lines. These carry the **per-instance identity** the fingerprint masks out
+  (the node path, the tenant, the guid), so read them for the concrete case, not just the shape.
+- `variants` — normally `1`. Anything higher means this log site fanned out past the watcher's
+  per-site budget in one window and its bursts were **folded onto this one incident**. Then the
+  ticket is "this site reports N different things and the masking rule needs a case", not a single
+  defect — say so, and name the shapes you can see in `samples`.
 
 # How to triage
 
--- a/src/MeshWeaver.Documentation/Data/Architecture/LogWatchTriage.md
+++ b/src/MeshWeaver.Documentation/Data/Architecture/LogWatchTriage.md
@@ -12,7 +12,10 @@ agent-written description of what is probably broken, filed in the repository th
                                 ▼
                        mw-log-watcher            (ns monitoring, its own PVC)
                           │ group lines into bursts
-                          │ fingerprint  =  hash(top app frame, exception)  — or hash(category, event id, exception) with no frame
+                          │ fingerprint  =  hash(WHERE, WHAT, WHICH)
+                          │     WHERE = top app frame, or (category, event id) with no frame
+                          │     WHAT  = exception type
+                          │     WHICH = masked exception message, or masked log message with no exception
                           │ queue to disk
                           ▼  POST /api/log-incidents   (Bearer, in-cluster)
                        Portal
@@ -42,13 +45,14 @@ GitHub App credential already live.
 ## One fault, one ticket
 
 The fingerprint (`StructuralLogIncidentIdentity.Compute`) identifies **the fault, not the reporter**.
-It is a `sha256` truncated to **16 hex characters** (the first 8 bytes) over one of two payloads,
-chosen by the burst itself:
+It is a `sha256` truncated to **16 hex characters** (the first 8 bytes) over three parts — *where*
+the fault is, *what* it is, and *which* one it is:
 
-| The burst names… | The identity is | Why |
+| Part | Value | Why |
 |---|---|---|
-| an application stack frame | `("frame", topFrame, exceptionType)` | the frame names the exact method that faulted — the most specific locator available |
-| no application frame | `("site", category, eventId, exceptionType)` | there is no location to key on, so the log site the code assigns is the locator |
+| **WHERE** | the top application stack frame — or `(category, eventId)` when the burst names no frame | the frame names the exact method that faulted; with no frame, the log site the code assigns is the only locator |
+| **WHAT** | the exception type, by simple name | the same method can fail two ways, and folding those hides the second bug behind the first one's ticket |
+| **WHICH** | the **masked exception message** — or the masked log message when there is no exception | inside one site there is nothing else left that says which fault this is |
 
 - **🚨 The reporting category is NOT in the identity when a frame is present.** The category names
   the class that *caught and printed* the fault, which is not where the fault is. One exception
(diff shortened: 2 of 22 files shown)


rbuergi:
Both Copilot findings addressed in cde84d890.

1. **`MaskSubjects` regex timeout** — fixed at the root rather than caught. A `RegexMatchTimeoutException` would throw straight through the aggregator and **wedge the poll**: a failed pass leaves the cursor put, so the next tick re-reads the identical window forever. Catching it would have been the band-aid. The real problem was using a regex at all for whole-token literal matching on the ingest hot path — it was rebuilt and re-parsed per message (the pattern cache is keyed by pattern text) and only needed a timeout because of that. It is now a linear scan over word tokens, gated by a subject length window so prose tokens are rejected without allocating. It cannot time out, so there is nothing to catch, and it is faster besides.

2. **Test name** — renamed to `ParkedNodeTypes_WithDifferentCompilerErrors_AreSeparateIncidents`, and the doc comment now says the fixture carries three of the thirteen production bursts.

PR #1806, opened 2026-08-17
fix(recycle): confirm before acting, and stop racing the dispose

rbuergi:
Prerequisite for the "a newer build is available — recycle to pick it up" banner. **A banner whose button does nothing is worse than no banner**, so this lands first.

## The bug

Recycle appeared to do nothing — you picked it, the page said *"Recycling hub…"*, and stayed there. Two defects, compounding:

1. **The teardown ran during render.** `Recycle(host, _)` posted `DisposeRequest` as a *side effect of rendering the area*. Merely navigating to the URL recycled the hub — no confirmation — and any re-render did it again.
2. **It raced its own dispose.** It then relied on `Observable.Timer(100ms)` to deliver a `RedirectControl` **through the hub it had just told to die**. The 100 ms was a sleep standing in for an ordering guarantee, and it usually lost: the stream was gone before the redirect emitted.

## The fix

The area now renders a **confirmation** — explanation + Recycle/Cancel — with **no side effect**. The teardown is on the button.

In the click action the **redirect is pushed FIRST and `DisposeRequest` posted SECOND**. The click runs on the hub's own turn, so the dispose is queued *behind* it and the redirect has already left for the client before teardown begins. Nothing has to survive the shutdown — so the timer is **deleted rather than tuned**.

## A mistranslation found on the way (user-facing, and it matters)

German `menu.recycle` read **"In den Papierkorb"** — *"move to the recycle bin"*, i.e. it told German users this action **deletes the node**. It is now **"Hub neu starten"**, and the new confirmation states in both languages that nothing is deleted and no content changes.

New key `ui.mdRecycleConfirm` added to both server catalogs **and** mirrored into `clients/react/src/i18n/`.

## Verification

- `dotnet build -c Release -warnaserror` clean
- `WhatsNewEntryIntegrityTest` / localization tests pass
- i18n drift guard invariants asserted directly (react deps aren't installed locally): **967 keys, sorted-key equality and per-key value equality across all four catalogs**, and `de` covers every `en` key

## Note on what is NOT in scope here

The Recycle area has never had its own permission check — the *menu item* is gated on `Permission.Update`, but the area is reachable by URL. That was true before this change and is arguably less exposed now (it takes a click rather than a page load). Worth a follow-up; deliberately not bundled.


rbuergi:
Re-running the failed jobs only. Recording why, so the merge record carries the reasoning rather than a bare retry.

`Run tests (shard 4)` failed on `MeshWeaver.AI.Test.ThreadTokenUsageTest.ErrorRound_KeysUsageAndCellByActualModel_NotRequestedAlias`. That is **root-caused and ticketed as #1812**, not waved away:

- This diff is `RecycleLayoutArea.cs`, four i18n catalogues and two markdown files. There is **no path** from it to token accounting.
- The failure has an independent, identified mechanism: `RecordUsage` writes in two phases (create a zero-token satellite, then accumulate via the owner's stream), and the test's `WaitForUsage` reads a **lagged CQRS query** instead of the node stream — so it can observe the phase-1 intermediate. AGENTS.md forbids that reader shape explicitly.
- Measured without this change: `MeshWeaver.AI.Test` × 5, `DOTNET_PROCESSOR_COUNT=4` — 5/5 green. Recorded in #1812 as **inconclusive, not exoneration**; local green proves nothing.

The rule against re-running exists to stop flakes being buried. This one is written down, with its mechanism, its correct fix, and an explicit prohibition against the tempting wrong one (widening the timeout).

PR #1815, opened 2026-08-17
Bound each browser-install attempt so a stalled CDN cannot eat a whole shard

rbuergi:
## The failure

PR #1808, run [32071216434](https://github.com/Systemorph/MeshWeaver/actions/runs/32071216434), shard 0. Every other job succeeded — all five sibling shards, `Build solution (once)`, both content gates. Shard 0 ran for its full 20-minute `timeout-minutes` budget and was cancelled, after which `Consolidate test results` failed.

It executed **zero tests**:

```
WARNING - Could not find any files for test/**/*.trx
grep: test/test-results.log: No such file or directory
Terminate orphan process: pid (2846) (npm exec playwright-core@1.62.1 install --with-deps --only-shell chromium)
```

It never got past **setup**. `npx playwright-core install` was still alive at job cleanup.

## Root cause

The provisioning step retries on the download **failing**. A stalled CDN does not fail — it **hangs**. Because the command never returned:

- `ok` stayed `0`, so **no retry ever ran** — the loop was never reached;
- the step's own `::error::… this is an INFRA failure, not a test failure` **never printed**;
- the hang consumed the entire job budget, and the run surfaced as `CANCELLED` + a failed consolidation.

So a step written specifically to make infra problems unambiguous produced the *least* attributable possible signal: a shape that reads like a test problem and names nothing. That is what cost the cycle.

## The fix

`timeout` per attempt. The hang becomes an ordinary non-zero exit, which **the existing retry loop already handles correctly** — so one bound restores the behaviour the step was written to have, rather than adding a mechanism.

🚨 **This is not raising a bound to make something pass.** The 20-minute job cap is unchanged. It gives the unbounded operation *inside* it a bound, so a stall is reported as the infra failure it is instead of silently eating the job. AGENTS.md's rule is to find what is not completing — this names it and makes it recoverable.

`124`/`137` are called out distinctly from an ordinary non-zero exit: a **stall** (CDN/runner network) and a **failed download** (bad version pin) want different follow-up, and are otherwise indistinguishable once both read as "attempt failed".

240 s × 3 + backoff ≈ 12.5 min worst case, comfortably inside the job cap; a healthy install takes well under a minute.

## Also: `clients.yml`

It carried the same install **unbounded and with no retry at all**. A red Clients workflow blocks image publication, so a stall there is expensive to misdiagnose. It gets the bound — a failure is visible and re-runnable, a hang is neither. No retry loop added: out of scope, and the bound alone already converts a silent hang into a visible infra failure.

## Verification

Both workflows parse (`yaml.safe_load`), the step's shell parses (`bash -n`), and the retry control flow was exercised across all four modes:

| mode | result |
|---|---|
| success | ok after attempt 1 |
| **hang once, then succeed** | **HUNG on 1 → ok after attempt 2** ← previously terminal |
| hang always | 3 attempts → INFRA FAILURE, exit 1 |
| fail always | 3 attempts → INFRA FAILURE, exit 1 |

`timeout` exists on the Linux runners; AGENTS.md's warning about it is specific to the macOS dev host.

No What's New entry: CI-only, nothing a user can observe.

### DIFF
--- a/.github/workflows/clients.yml
+++ b/.github/workflows/clients.yml
@@ -272,5 +272,11 @@ jobs:
         working-directory: clients/grpc-web
         run: npm ci --no-audit --no-fund && npm run gen
       - run: npm ci --no-audit --no-fund
-      - run: npx playwright install --with-deps chromium
+      # Bounded for the same reason as dotnet-test.yml's provisioning step: a stalled CDN does
+      # not fail the download, it HANGS it, and an unbounded hang here burns the whole job's
+      # budget and reports as a timeout that names nothing. A red Clients workflow blocks image
+      # publication, so a stall that reads as "the clients broke" is expensive to misdiagnose.
+      # No retry loop here (unlike dotnet-test) — this step has none today, and a bound alone
+      # already converts a silent hang into a visible, re-runnable infra failure.
+      - run: timeout --signal=TERM --kill-after=30 240 npx playwright install --with-deps chromium
       - run: npm run e2e # expo export --platform web → serve dist/ → drive headless Chromium
--- a/.github/workflows/clients.yml
+++ b/.github/workflows/clients.yml
@@ -278,5 +278,11 @@ jobs:
       # publication, so a stall that reads as "the clients broke" is expensive to misdiagnose.
       # No retry loop here (unlike dotnet-test) — this step has none today, and a bound alone
       # already converts a silent hang into a visible, re-runnable infra failure.
-      - run: timeout --signal=TERM --kill-after=30 240 npx playwright install --with-deps chromium
+      # The apt wait is for the same reason as dotnet-test's: `--with-deps` shells out to apt-get,
+      # and the runner's own background apt makes it exit 100 "Could not get lock" — contention,
+      # not a broken download. `flock … true` waits for that holder and releases immediately, so
+      # playwright takes the lock itself. Both bounds are hard: it can wait, it cannot hang.
+      - run: |
+          sudo flock -w 60 /var/lib/apt/lists/lock true 2>/dev/null || true
+          timeout --signal=TERM --kill-after=30 180 npx playwright install --with-deps chromium
       - run: npm run e2e # expo export --platform web → serve dist/ → drive headless Chromium
--- a/.github/workflows/dotnet-test.yml
+++ b/.github/workflows/dotnet-test.yml
@@ -596,26 +596,58 @@ jobs:
     # input-shaped `if:` — a browser this step failed to install must redden the shard, not
     # silently leave the tests asserting the "no browser here" branch and calling it green.
     #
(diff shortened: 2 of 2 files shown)


rbuergi:
## Update — the first run of this PR reproduced a *different* failure in the same step, and it is now fixed too

Shard 1 of run [32077919296](https://github.com/Systemorph/MeshWeaver/actions/runs/32077919296) failed all three attempts within 30 seconds:

```
E: Could not get lock /var/lib/apt/lists/lock. It is held by process 2860 (apt-get)
Failed to install browsers
Error: Installation process exited with code: 100
```

**The timeout added by this PR was not involved.** Every attempt exited `1`, not `124` — the retry loop worked exactly as designed and reported the infra failure loudly instead of hanging for 20 minutes. What it could not do was *succeed*: `--with-deps` shells out to `apt-get`, the runner image runs its own background apt, and the backoff (10/20 s) was shorter than that other apt run, so all three attempts re-raced the same holder.

So the step had **two** independent ways to fail as infrastructure, and this PR originally addressed only one:

| failure | before | now |
|---|---|---|
| stalled download (**hang**) | ate the 20-min job cap, retry loop never reached | bounded → retried → loud |
| **apt contention** | 3 attempts inside 30 s, all lost the same race | wait for the holder, then attempt |

`flock -w 60 /var/lib/apt/lists/lock true` waits for the holder to finish and releases immediately, so playwright takes the lock itself rather than us holding it across its whole run — which would deadlock. `|| true` because a timed-out wait is not a reason to skip the attempt.

**Budget kept honest:** the install timeout comes down 240 → 180 s so the added wait does not creep the sum up on the job cap. Worst case is now 3 × (60 + 180 + 30) + 30 ≈ **14 min** against the 20-minute cap — the same figure as before, and the comment names every term with that constraint stated explicitly so the next person to touch a bound sees why it must stay balanced.

`clients.yml` gets the same wait for the same reason.

rbuergi:
## Third occurrence tonight — independent confirmation, on a different PR

While this PR was in CI, **#1816** (an unrelated Postgres-trigger change, branched before this fix) hit the identical hang:

```
run 32077430794 · Run tests (shard 5)
  8 Provision the headless browser (infra, not a test) :: cancelled  22:54:49Z -> 23:14:29Z   (19m40s)
  9 Run Tests                                          :: skipped
```

The step consumed the whole job to the 20-minute cap, zero tests ran, and the run reported `CANCELLED` — the same shape as #1808's shard 0 that started this.

So in one evening the unbounded install has cost **three** shard-jobs across **two** unrelated PRs, each burning ~20 runner-minutes and each presenting as something other than what it is. That is the frequency this is worth fixing at, and it is why the bound belongs on the step rather than being left to re-runs.

It also means #1816 cannot go green until this merges and it picks the fix up — which is the sanctioned reason to merge main into a branch (CI failing on something the diff does not touch).

PR #1825, opened 2026-08-18
fix: re-drive a NodeType that never compiled and settles at Error (#1793)

### LINKED ISSUE #1793: A NodeType that has never compiled successfully and settles at Error is unreachable by every automatic re-drive
Split out of #1786, where it is the reason a fix to the *cause* could not reach the broken nodes.

## The gap

When a compile fails, `NodeTypeCompilationHelpers.ApplyCompileFailure` stamps `CompilationStatus = Error` and **does not** stamp `CompiledFrameworkVersion`, `LatestAssemblyCollection` or `LatestAssemblyPath`. For a NodeType that has never compiled successfully on this deployment, those three are therefore null forever.

From that moment nothing can re-drive the compile:

| path | requires | a never-succeeded `Error` node |
|---|---|---|
| first-build kickoff | `CompilationStatus is null` | ✗ it is `Error` |
| recovery kickoff | `CompilationStatus == Compiling` | ✗ |
| framework-stale kickoff | `LatestAssemblyCollection` + `LatestAssemblyPath` + `CompiledFrameworkVersion` (via `HasStaleFrameworkBuild`) | ✗ none of them stamped |
| `InstallReleaseRequestWatcher` | `RequestedReleaseAt > LastReleaseRequestHandledAt` | ✗ only a user action moves it |

Only a human pressing Compile (or a Recycle) gets the node out. Nothing about a redeploy, a framework bump, a module update or a fix to the failing code does.

## Why it looks unintended rather than designed

The framework-stale kickoff's own filter already reads

```csharp
def.CompilationStatus is CompilationStatus.Ok or CompilationStatus.Error
```

so it *intends* to cover `Error`. It cannot, because the predicate it delegates to needs assembly coordinates a failed compile never writes. A type that succeeded once and then went stale is rebuilt automatically; a type that failed the first time is not — the harder case gets the weaker treatment.

The parking behaviour itself is right and should stay: `CompileWatcher` parks a failed type and serves the cached error rather than storming. The question is only whether a **new framework** (i.e. new code, possibly containing the fix) is allowed one fresh attempt.

## Evidence it has already cost us

`NodeCompileShaping.AnchorIncludePath` was written for an incident whose own comment records: *"On memex-cloud 2026-08-12 that parked 15 NodeTypes (FutuRe, Cession, SocialMedia, Northwind/AnalyticsCatalog)"*. That fix shipped — and on 2026-08-17 #1786 found near enough the same list still parked. The fix could not reach the nodes it was written for, because they were already in `Error` with no assembly coordinates.

Same shape in #1786's own data: the eight sample NodeTypes that shipped a baked `"compilationStatus": "Error"` never compiled once on any deployment, while the twenty-four that shipped a baked `Ok` self-healed on the first framework bump — purely because `Ok` came with assembly coordinates and `Error` did not.

## Sketch of a fix (not written — wants review)

Record on failure which framework produced the verdict — `CompiledFrameworkVersion`, or a distinct `FailedFrameworkVersion` / `LastVerdictFrameworkVersion` if reusing the success field would confuse `HasUsableBuild` — and let the framework-stale kickoff re-drive an `Error` node whose stamped framework differs from the live one. That gives exactly one retry per framework version: no storm (the stamp updates whether the retry succeeds or fails), and a deployed fix reaches the nodes it was written for.

Worth checking against:
* the park registry — a parked type must not be woken on every activation, only on a genuine framework change;
* the bake gate, which classifies `Error` types today;
* whether `Unavailable` (from `SourceDiscoveryUnavailableException` / `AddressRecyclingException`) wants the same treatment or already has it.

rbuergi:
Closes #1793.

## The gap

A compile that FAILS writes no assembly coordinates at all — `ApplyCompileFailure` stamps neither `LatestAssembly{Collection,Path}` nor `CompiledFrameworkVersion`. For a NodeType that never compiled successfully *on this deployment* those are null forever, and **every** automatic re-drive keys off something only a first success produces:

| path | requires | a never-succeeded `Error` node |
|---|---|---|
| first-build kickoff | `CompilationStatus is null` | ✗ it is `Error` |
| recovery kickoff | `CompilationStatus == Compiling` | ✗ |
| framework-stale kickoff | `HasStaleFrameworkBuild` → assembly coordinates + `CompiledFrameworkVersion` | ✗ a failure stamps none of them — and its own filter reads `Ok or Error`, so it *intends* to cover this |
| `InstallReleaseRequestWatcher` | `RequestedReleaseAt > LastReleaseRequestHandledAt` | ✗ only a human moves it |
| sources watcher's parked auto-retry | the **in-memory** park registry to hold the path | ✗ a failure that predates the process is not in it — so a source fix after a restart re-drives nothing either |
| enrichment self-heal (`ApplyStreamResult`) | `Ok/Error` **and** assembly coordinates | ✗ falls straight through to the "correct the code" overlay |
| pre-warm sweep (`WarmOne`) | activates the hub, then waits for `HasUsableBuild ‖ Error` | ✗ returns `CompileError` at once; the activation fires only the kickoffs above |
| `RebuildMissingBytes` | `BakeState.BytesMissing` (record claims a live-framework build) | ✗ classified `PreviouslyBroken` |

The one path that *does* re-drive it is the **batch** bake (`NodeTypeBatchBake.BakeOne` drives the compiler unconditionally) — but that is taken only on a readiness-gated bake pod / `PreWarm:BatchBake`, never on a serving pod's activation-driven sweep.

So only a human pressing **Compile** (or a Recycle) got such a node out. A redeploy, a framework bump, a module update and a fix to the failing code reached none of them — which is why `NodeCompileShaping.AnchorIncludePath`, written for the fifteen types it parked on memex-cloud 2026-08-12, could not reach them, and why #1786 found near enough the same list still parked five days later.

## The fix

A failure now records the one thing it honestly can: **the inputs the verdict was formed from** — framework identity + installed-module fingerprint + the source snapshot the compile consumed — folded into `NodeTypeDefinition.FailedBuildInputs`. A new owner-side kickoff fires exactly when the LIVE inputs differ from that stamp:

| What moved | Effect |
|---|---|
| A new framework (a redeploy, possibly carrying the fix) | one fresh attempt |
| A module update | one fresh attempt |
| An edited / added / removed source | one fresh attempt |
| Nothing — same framework, modules, sources | **no attempt**: the identical failure would reproduce |
| The stamp is `null` (every pre-existing parked node, and every `Error` baked into a node file) | one fresh attempt — the migration |

It is the strict **complement** of the framework-stale kickoff: it applies only where no assembly coordinates exist, so the two can never both claim the same state.

### Backoff / give-up

1. **STRUCTURAL** (the one that does the work) — the flip to `Pending` writes the live token **in the same `Update`**, so the trigger the re-drive fires on is false the instant it fires. A reconcile that can re-arm its own trigger is the 257,000-version `_Policy` write-storm shape; the stamp forecloses it. It is stamped at the flip as well as in `ApplyCompileFailure`, so it holds even when the compile never writes back at all (process death mid-compile, the parked re-settle, a poisoned content read).
2. **LOUD** — a process-wide ledger (`NodeTypeCompileParkRegistry.RecordFailureRedrive`) logs an **error naming the path** the moment a type is re-driven twice for the *same* inputs, i.e. the moment (1) provably did not hold.
3. **TERMINAL** — past `MaxAutomaticFailureRedrives` the kickoff gives up for the hub's lifetime and says so, naming the type, its error and the remedy. An explicit Compile refunds the budget.

And when the re-drive **declines** — a failure formed under exactly the live inputs — the hub logs one warning per activation naming the type, its error and why nothing will retry it. That state was correct and bounded before this change, and *completely silent*: nothing anywhere named a NodeType that is broken and will not be retried.

`FailedBuildInputs` is mesh-owned operational state — exports strip it, imports preserve the live node's value, and `ShippedNodeTypeStateTest` bans it from committed node files (an authored token matching the importing deployment would suppress precisely the retry it exists to grant).

`Unavailable` gets the same treatment: it records that a compile never reached a verdict at all, so it is even less of a reason to stop trying. (`EnsureCompileDispatched` already re-drives it on the next *request*; this adds the activation-time path under the same bound.)

## Tests — executed, and verified non-vacuous

`test/MeshWeaver.Graph.Test/FailedVerdictRedriveTest.cs` — 16 pure pins: the token (order-insensitive, unseeded ≠ empty), the predicate's truth table (null stamp / other framework / other modules / edited-added-removed source / `Unavailable` / only-settled-failures / left to the framework-stale twin), both stamps, the ledger, the finite budget, the operational-member contract.

`test/MeshWeaver.Hosting.Monolith.Test/NeverCompiledFailureRedriveTest.cs`:
- `NeverCompiledErrorRecord_IsReDrivenAutomatically_AndRecovers` — the #1786 record shape recovers with no user action, and is not re-driven again.
- `ABrokenSource_IsReDrivenExactlyOnce_AndThenLeftAlone` — a source that cannot compile gets exactly one attempt (ledger == 1), no further `Pending` flip in a bounded window, and re-parks.
- `AFixedSource_ReDrivesAFailureThatPredatesThisProcess` — the park is cleared first (emulating a restart), so only the durable stamp can notice the fix.

```
Passed!  - Failed: 0, Passed: 1274 - MeshWeaver.Graph.Test.dll
Passed!  - Failed: 0, Passed:   48 - MeshWeaver.Hosting.Monolith.Test.dll (compile-related filter, incl. the 3 new)
Passed!  - Failed: 0, Passed:  408 - MeshWeaver.PluginCatalog.Test.dll
Passed!  - Failed: 0, Passed:   34 - MeshWeaver.Documentation.Test.dll
```

**Control run** (AGENTS.md: a verification step that cannot fail is not a verification step) — with the new kickoff's `Where` neutered to `false`, the same three integration tests fail:

```
Failed!  - Failed: 3, Passed: 0, Total: 3, Duration: 4 m 32 s
  NeverCompiledFailureRedriveTest.ABrokenSource_IsReDrivenExactlyOnce_AndThenLeftAlone [FAIL]
  NeverCompiledFailureRedriveTest.AFixedSource_ReDrivesAFailureThatPredatesThisProcess [FAIL]
  NeverCompiledFailureRedriveTest.NeverCompiledErrorRecord_IsReDrivenAutomatically_AndRecovers [FAIL]
```

Docs: a new section in `Doc/Architecture/NodeTypeCompilation` and a What's New entry (`Category: Fix`).

https://claude.ai/code/session_016wVYniTKJ7Kuw7Hh425UPy

### DIFF
--- a/src/MeshWeaver.Documentation/Data/Architecture/NodeTypeCompilation.md
+++ b/src/MeshWeaver.Documentation/Data/Architecture/NodeTypeCompilation.md
@@ -422,6 +422,7 @@ LIVE inputs differ from that stamp:
 | An edited / added / removed source | one fresh attempt |
 | Nothing — same framework, modules and sources | **no attempt**: the identical failure would reproduce |
 | The stamp is `null` (a failure from before this field, or an `Error` baked into a node file) | one fresh attempt — the migration |
+| The source set has not been established yet (`CurrentSourceVersions` unwritten) | **no attempt — it WAITS**: "not known yet" is not "no sources", and a compile driven from a set nobody established forms a verdict from evidence the mesh does not have |
 
 It is bounded three ways, and the first is the one that does the work:
 
@@ -433,6 +434,15 @@ It is bounded three ways, and the first is the one that does the work:
    moment (1) provably did not hold. Non-convergence is never quiet.
 3. **Terminal.** Past `MaxAutomaticFailureRedrives` the kickoff gives up for the hub's lifetime and
    says so, naming the type, its error and the remedy. An explicit Compile refunds the budget.
+
+The re-drive is **owner-driven, never caller-driven**. It fires from the type's OWN hub on facts the
+node already holds; no request, and no requester's identity, is an input. The compile runs as System
+and its activity row lands in the owning partition attributed to System — exactly as the first-build,
+recovery and framework-stale kickoffs have always done — and no user's `RequestedReleaseAt` /
+`RequestedReleaseBy` is touched, so nothing is misattributed. An unauthorized caller who merely
+activates the hub therefore gains no lever: the trigger is a property of the persisted record, and
+the three inputs that can move it (framework identity, installed modules, the type's own source
+nodes) are all writable only by principals who already hold that access.
 
 And when the re-drive **declines** — a type settled at `Error`/`Unavailable` whose verdict was formed
 under exactly the live inputs — the hub logs one warning per activation naming the type, its error
--- a/src/MeshWeaver.Documentation/Data/Architecture/NodeTypeCompilation.md
+++ b/src/MeshWeaver.Documentation/Data/Architecture/NodeTypeCompilation.md
@@ -397,6 +397,53 @@ Anything else triggers a recompile. This makes a cold hub start **self-healing**
 | **MeshWeaver redeployed with a breaking change** | The cached DLL bound against the *old* framework surface (ABI-stale) | Rule 3 |
 | Module updated | The cached DLL may bind the replaced module's old ABI | Rule 4 |
 
+### A compile that FAILED is re-driven too — one attempt per set of inputs
+
+`HasUsableBuild` and its framework-stale twin both key on **assembly coordinates**, and a failed
+compile writes none: `ApplyCompileFailure` stamps neither `LatestAssembly{Collection,Path}` nor
+`CompiledFrameworkVersion`. For a NodeType that never compiled successfully *on this deployment*
+those stay null forever, so every automatic path used to skip it — the first-build kickoff needs a
+`null` status, the recovery kickoff needs `Compiling`, the framework-stale kickoff needs the
+coordinates, the release watcher needs a human, and the park registry's source-change auto-retry is
+in-memory (a failure that predates the process is not in it). Only a human pressing **Compile** got
+such a node out; a redeploy, a framework bump, a module update and a fix to the failing code reached
+none of them ([#1793](https://github.com/Systemorph/MeshWeaver/issues/1793); the fix written for
+fifteen types parked on memex-cloud could not reach the nodes it was written for).
+
+So a failure records the one thing it honestly can: **the inputs the verdict was formed from** —
+framework identity, installed-module fingerprint, and the source snapshot the compile consumed —
+folded into `NodeTypeDefinition.FailedBuildInputs`. The owner-side re-drive fires exactly when the
+LIVE inputs differ from that stamp:
+
+| What moved | Effect |
+|---|---|
+| A new framework (a redeploy, possibly carrying the fix) | one fresh attempt |
+| A module update | one fresh attempt |
+| An edited / added / removed source | one fresh attempt |
+| Nothing — same framework, modules and sources | **no attempt**: the identical failure would reproduce |
+| The stamp is `null` (a failure from before this field, or an `Error` baked into a node file) | one fresh attempt — the migration |
+
+It is bounded three ways, and the first is the one that does the work:
+
+1. **Structural.** The flip to `Pending` writes the live token **in the same update**, so the trigger
+   the re-drive fires on is false the instant it fires. A reconcile that can re-arm its own trigger
+   is the 257,000-version write-storm shape; the stamp forecloses it.
+2. **Loud.** A process-wide ledger (`NodeTypeCompileParkRegistry.RecordFailureRedrive`) logs an
+   **error naming the path** the moment a type is re-driven twice for the *same* inputs — i.e. the
+   moment (1) provably did not hold. Non-convergence is never quiet.
+3. **Terminal.** Past `MaxAutomaticFailureRedrives` the kickoff gives up for the hub's lifetime and
+   says so, naming the type, its error and the remedy. An explicit Compile refunds the budget.
+
+And when the re-drive **declines** — a type settled at `Error`/`Unavailable` whose verdict was formed
+under exactly the live inputs — the hub logs one warning per activation naming the type, its error
+and why nothing will retry it. That state is correct and bounded, and before that line it was also
+completely silent: nothing anywhere named a NodeType that is broken and will not be retried.
+
+> 🚨 `FailedBuildInputs` is **mesh-owned operational state**: exports strip it, imports preserve the
+> live node's value, and `ShippedNodeTypeStateTest` bans it from committed node files. An authored
+> token that happened to match the importing deployment's live inputs would suppress precisely the
+> retry it exists to grant.
+
 ### Framework-version freezing
 
 A compiled NodeType DLL references the MeshWeaver framework assemblies present
--- a/src/MeshWeaver.Documentation/Data/WhatsNew/2026-08-18-a-broken-node-type-can-recover-by-itself.md
+++ b/src/MeshWeaver.Documentation/Data/WhatsNew/2026-08-18-a-broken-node-type-can-recover-by-itself.md
@@ -1,0 +1,27 @@
+---
+Name: A broken NodeType can recover by itself
+Category: Fix
+Description: A NodeType whose very first build failed is now rebuilt automatically when the platform, its modules or its code change — instead of staying broken until someone presses Compile.
+Icon: ArrowSync
+Order: -20260818
+---
+
+# A broken NodeType can recover by itself
+
+When a NodeType's build fails, the page it powers shows an error card instead of its content. That
+is meant to be temporary: fix the code, or deploy a platform version that fixes it, and the type
+rebuilds. For one particular kind of failure it was not temporary at all.
+
+If a NodeType had *never* built successfully on that installation — a fresh import, a first
+deployment, or a type whose failed state came in with the content — nothing would ever try again.
+Every automatic rebuild the platform performs was looking for traces that only a successful build
+leaves behind, and a failed build leaves none of them. So a new platform release, a module update
+and even a corrected source file all went past it, and the only way out was for someone to open the
+type and press Compile. Types stayed broken for days that way, including through the very release
+that contained their fix.
+
+A failed build now records what it was built against — the platform version, the installed modules
+and the exact source files it used. When any of those change, the type gets one fresh attempt on
+its own. A type that is simply broken is retried once and then left alone, with a clear line in the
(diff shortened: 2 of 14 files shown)


rbuergi:
## Security review of the re-drive — CI red on `OrleansCompileActivityAccessTest.User_Without_Edit_Triggers_Recompile_Is_Denied_Cleanly`

The failure was real and caused by this PR: the test seeds a NodeType at `Error` to suppress every System-driven kickoff, and its stated precondition was *"the first-build kickoff suppressed by the seeded terminal status"*. A settled status alone stopped being sufficient here — a seeded `Error` carrying no `FailedBuildInputs` **is** the "imported from elsewhere, never attempted here" record that #1793 exists to re-drive.

Treated as a security question first. The three answers:

### 1. Who drives the compile — the owner, or the unauthorized caller?

**The owner, on facts the persisted record already holds.** The kickoff lives on the type's own hub and is level-triggered on its own MeshNode stream; its inputs are `CompilationStatus`, the (absent) assembly coordinates, `FailedBuildInputs`, `CurrentSourceVersions`, the process framework identity and the installed-module fingerprint. **No request, and no requester's identity, is an input.** TestUser's write was denied (`DeliveryFailureException`), nothing they sent reached the node, and `RequestedReleaseAt` / `RequestedReleaseBy` are untouched — the kickoff deliberately never sets them, exactly like the first-build kickoff, so no release is attributed to the caller.

What the caller *can* do is cause an **activation**, and activation-time System work is a pre-existing property of this design, not something this PR introduces. The test's own comments already name two members of that population: the first-build kickoff (*"INFRASTRUCTURE, orthogonal to user authorization: it would create an `_Activity` row here even though TestUser is denied"*) and `NodeTypeCompileStateMirror` (*"mirrors the SEEDED terminal status under System on activation"*, which the assertion already excludes by id). The recovery and framework-stale kickoffs are two more. This kickoff joins that population under a **tighter** bound than any existing member.

### 2. Does the row leak anything, or land where the caller should not cause writes?

No.
- Written **as System** (`AccessContextScope.AsSystem`) into `{typePath}/_Activity/compile-*` — the **owning** partition, i.e. the owner's hub writing bookkeeping about its own node. Same path, identity and partition the first-build/recovery/framework-stale kickoffs, the sources watcher and the compile-state mirror have always used.
- `CreatedBy` is System, not TestUser; nothing is misattributed.
- No new read grant: the row is visible only to principals who can already read `OtherUser/…/_Activity`.
- The park-failure notification does **not** reach the caller either: it prefers `RequestedReleaseBy` (untouched → null), so a System-driven build falls back to a satellite of the failing type — RLS-filtered to whoever can already read it.

### 3. Amplification?

Bounded, and the reasoning is explicit:

- **Structural, and the one that matters.** The flip to `Pending` writes `FailedBuildInputs = live token` in the **same `Update`**, so the trigger is false the instant it fires. Re-activating the grain a thousand times yields **zero** further re-drives, because the trigger is a property of the persisted record, not of the activation.
- **The caller cannot move the token.** Its three inputs are the framework identity (process-wide), the installed-module fingerprint (process-wide), and `CurrentSourceVersions` — written by the owner-side sources watcher from the type's own `Source/`/`Test/` nodes. Manufacturing a new input set requires write access to the owner's partition, which is precisely the access being denied.
- **Process-wide cap.** `MaxAutomaticFailureRedrives = 5` lives on the mesh-scoped `NodeTypeCompileParkRegistry`, so it survives hub recycling; only the give-up latch is per-hub. Recycling the grain does not refill the budget.
- **Net effect:** at most one compile per genuine change to inputs the caller does not control — strictly less than the pre-existing first-build kickoff already grants, and work the pre-warm sweep performs for every dynamic type at boot anyway. The caller can at most advance its schedule, never multiply it.

**Conclusion: clean.** The test's precondition was obsolete; its assertion was not.

## What changed

**Product** (one real bug the failure exposed, fixed at the root): `HasStaleFailureVerdict` now requires `CurrentSourceVersions` to be non-null. On a cold activation the sources watcher has not written a snapshot yet, so the source half of the token is not "no sources" — it is **"not known yet"**. Firing there drove a compile from a set nobody established (the #1216 lesson: *verdicts about code from evidence you do not have*) **and** burned a second attempt for free, since the watcher's first write moved the token and re-drove the type one emission later. The watcher always writes a snapshot — the empty map when the queries match nothing — for every type this predicate applies to, so this **orders** the mechanism rather than disabling it, and a mesh that cannot answer the source query degrades to *no re-drive* rather than to a wrong verdict. The stuck-type diagnostic takes the same gate so a cold activation is not reported as "stuck" while the re-drive is merely waiting.

**Test** (the precondition, never the assertion): the record is now seeded as a *locally-formed failure with nothing due* — `CurrentSourceVersions = empty` (what the watcher writes for a type with no `Source/` children, so there is no unestablished window) and `FailedBuildInputs` = the token the framework itself computes, derived through `NodeTypeCompilationHelpers.BuildInputsToken` rather than re-implemented in the test (a token two call sites compute differently either never converges or never fires — both silently). `MeshWeaver.Graph` gains `InternalsVisibleTo` for that one reason, alongside the three test projects already there. The seeding comment now enumerates all five kickoffs and why each is excluded by the record itself.

🚨 **The assertion is untouched** — the `MessageHubConfiguration.cs:328-342` pin (PostPipeline no longer stamping hub-self as principal) survives verbatim, wording included apart from naming the broader kickoff set.

## Re-run

```
Passed! - Failed: 0, Passed:    8 - MeshWeaver.Hosting.Orleans.Test.dll (OrleansCompileActivityAccessTest, incl. the failing test)
Passed! - Failed: 0, Passed: 1275 - MeshWeaver.Graph.Test.dll
Passed! - Failed: 0, Passed:   48 - MeshWeaver.Hosting.Monolith.Test.dll (compile-related filter, incl. the 3 new)
```

**Control** — reverting *only* the seeded `FailedBuildInputs` (leaving everything else) fails the access test again:

```
OrleansCompileActivityAccessTest.User_Without_Edit_Triggers_Recompile_Is_Denied_Cleanly [FAIL]
Failed! - Failed: 1, Passed: 0
```

so the new seeding is load-bearing, not a widened assertion. (The earlier control for the fix itself still stands: neutering the kickoff's `Where` fails all three `NeverCompiledFailureRedriveTest` cases.)

No collision with #1831 or #1829 — main has moved only under `deploy/aks/**` since this branch's base, and `MeshWeaver`'s `main` is not `strict`, so no rebase is required.

PR #1844, opened 2026-08-18
A sync licence with terms and a term, exchangeable for a short-lived scoped token

### LINKED ISSUE #1835: Sync licence: the registry must authorize an INSTANCE to replicate a package, and hand it a short-lived token — per-repo PATs (MW_PLUGINS_REPO_TOKEN) do not scale
## The problem

Consumers that need a plugin's content today get a **per-repo GitHub PAT**. `Systemorph/MeshWeaver.Education` holds one as the repo secret **`MW_PLUGINS_REPO_TOKEN`** (created 2026-07-13, never rotated), used at four `actions/checkout` sites against `Systemorph/MeshWeaver.Plugins` on `origin/main`:

| line | job | scope |
|---|---|---|
| 390 | `e2e-install` | full repo |
| 584 | `e2e-mesh` | full repo |
| 667 | `e2e-journeys` | full repo |
| 805 | `publish-bake` | sparse: `Edu Store Publish Video` |

(Not to be confused with **`MW_PLUGINS_REPO`**, the env var holding the *filesystem path* to that checkout — one letter apart, and a stale value of the latter has already invented a false root cause in an education e2e investigation.)

`MeshWeaver.Manufacturing` now needs `Publish`'s `Deck`/`Slide` types for its `Demo/Tour` node, and the request that arrived was "issue another PAT, same as Education". **That does not scale**: one personal credential per (consumer × source repo), each a standing whole-repo grant, none revocable centrally, none expressing which *package* is needed. Five plugin repos and growing.

**The registry exists precisely so this is unnecessary.** `Doc/Architecture/PluginRegistry`: "Onboarding a new installation is *point `PluginCatalog:RegistryUrl` at the registry*, not *provision it a GitHub App*." A consumer holding its own GitHub PAT is that principle inverted.

## What already works (do not rebuild)

1. **Consumer → registry authentication.** `mwi_` instance key: 32 random bytes, SHA-256-hashed at rest, routed through `MeshWeaverInstanceIndex`, resolved by `InstanceRegistryAuthenticator`. `Bearer`, plus `Basic` because a NuGet client cannot send `Bearer`. Kill switch via `MeshWeaverInstance.IsDisabled`.
2. **The registry brokering GitHub.** The `meshweaver-cloud` GitHub App (org-wide, `contents:write`) is the one credential; `/api/plugins` and `/api/plugins/files` re-serve, and the credential never leaves. Verified on `72cd0a53d`: both routes call `caller.Allows(source.Name, package.Id)`, and `/files` answers **404, not 403**, for an ungranted package so it cannot be used as an existence oracle.
3. **A credential-exchange endpoint, already in this shape.** `mwr_` bootstrap key → `POST /api/instances/register` → the instance's own `mwi_` returned **once**. `RegistrationKey` already carries `ExpiresAt`, revocation and use-counting.

Two things are missing.

## Gap 1 — there is no SYNC LICENCE, and it must not be borrowed from the user entitlement

The consumer here is **another MeshWeaver instance**, and what it needs permission for is **sync** — replicating a package from the registry into itself. That is its own kind of licence, and the codebase has no representation of it. Three separate concepts, currently two-plus-a-hole:

| | subject | right | today |
|---|---|---|---|
| `Package.License` | nobody — it is a *label* | what the copyright holder permits (SPDX id, `Apache-2.0`) | ✅ exists, correctly nullable ("unspecified" stays unspecified) |
| Store entitlement (`_Entitlements/{viewer}`, `PackageEntitlement.Authorize`) | a **user** | install & use in their own portal / home | ✅ exists |
| **sync licence** | a **`MeshWeaverInstance`** | pull/replicate the package from the registry | ❌ **missing** |

`PluginGrant` is currently standing in for the third row, and it is not a licence — it is a hand-written ACL at `Admin/_PluginGrant/{instanceId}` with entries and nothing else: no issuer, no term, no expiry, no commercial basis, no renewal. Today's entire live table is three nodes (`memex`, `memex-cloud`, `atioz`), each hand-edited `Plugins/*`. Every new consumer means an admin editing a node by hand — the part that cannot be done for all plugins.

🚨 **Deriving the sync right from the user-purchase entitlement would be wrong, not merely imprecise.** They differ on every axis:

- A user's purchase entitles **that user, on their portal**. It says nothing about an arbitrary instance being allowed to pull the source and hold a copy.
- A sync licence is needed for **free** packages too — `Plugins/*` to a partner's mesh, where no purchase exists at all — so purchase-derivation cannot be the mechanism.
- The commercial terms are different in kind: a per-instance replication/redistribution right versus a per-seat use right. They will be priced, termed and renewed differently.

**Ask:** introduce the sync licence as a first-class type whose subject is a `MeshWeaverInstance` and whose right is sync, with an issuer, a term and revocation. The registry authorizes against **that**. `PluginGrant` either becomes its projection or remains only as an explicit partner/override escape hatch. Preserve the current posture: fail closed, 404-not-403, and `DefaultGrants` never seeding private or paid sources.

## Gap 2 — no short-lived, scoped token

`mwi_` is a **durable static bearer**. `MeshWeaverInstance` has `KeyIssuedAt` but no expiry, no refresh, no exchange; re-issuing is a manual Settings action. A consumer that needs registry access holds a long-lived secret forever.

**Ask:** add the OAuth-style leg — exchange the durable credential for a **short-lived access token scoped to the packages the instance's sync licence actually covers**, presented per request. This is a further key type on an established pattern, not new machinery: `RegistrationKey` already has expiry + revocation + use-counting, and `InstanceKeys` already documents why the prefixes are kept disjoint (`mwi_` vs `mwr_` vs personal `mw_`) so no validator can accept another's key by accident.

## Why this dissolves the original request

**CI is already another MeshWeaver instance** — that is the part that makes this fit rather than being a special case. Education's `e2e-install` / `e2e-mesh` / `e2e-journeys` jobs literally boot a disposable mesh (`e2e/mesh/up.sh`), and the `mw-plugin-test` gate imports node repos into a fresh SQLite-backed mesh. Those meshes are consumers in exactly the sense meant here. They just get their content from a PAT-checked-out directory handed to them as `MW_PLUGINS_REPO` instead of by holding a sync licence and pulling.

With both gaps closed, such an instance registers (ephemerally), holds a sync licence for what it is entitled to replicate, exchanges its credential for a short-lived scoped token, and pulls `Plugins/Publish` via `POST /api/plugins/files` — which returns `{files:[{relativePath, content}]}`, exactly the node-repo tree the gate stages into its `/repo` mount. A one-for-one replacement for `actions/checkout`. No PAT, no GitHub access, nothing to rotate by hand.

Then `MW_PLUGINS_REPO_TOKEN` is deleted from `MeshWeaver.Education` and its four sites migrate. **All four** — `publish-bake` is easy to miss and behaves differently: it stages the four Plugins-owned modules so the courses' `requires` resolve, then deletes `Edu/Store/Publish/Video.zip` from the bake output so the same packages are not seeded from two sources.

## Notes

- Core **#1772** (bundle route authenticated but never authorized) closed 2026-08-17. That was `PluginBundleEndpoints`, a different route from the two above — but confirm the fix is in the image memex actually runs before relying on per-package scoping anywhere.
- `MeshWeaver.Manufacturing` currently holds only `ACR_USERNAME` / `ACR_PASSWORD`, so it has no such token today and nothing regresses by not issuing one.
- Requested by Roland: consumers should not hold repo credentials — memex.meshweaver.cloud checks the licence and operates the GitHub connection on their behalf; the licence permitting *sync* is its own kind, and the consumer is another MeshWeaver instance.

rbuergi:
Closes #1835.

## Why

A `PluginGrant` could say WHICH packages a registered instance may replicate — and nothing else. No term, no terms, no record of what the right was granted *for*. Two writers only: the `DefaultGrants` registration seed, and an admin typing into the Instance-grants tab.

Workable for three instances; not for a catalogue. The pressure surfaced the usual way — a consumer that needed **one package** asked for a standing GitHub credential to a **whole repository**, because there was no smaller thing to ask for.

## What changed

**The grant IS the licence.** Entries carry `ExpiresAt`, `IssuedUnderLicense`, `IssuedVia`, `IssuedAt`; the grant carries `IsRevoked`. The term sits on the *entry* because one instance routinely holds a perpetual licence for the platform repo alongside a termed one for a paid package.

Null `ExpiresAt` is perpetual — every grant written before this behaves identically, and the existing `PluginGrantTest` suite is untouched and green. `Matches()` deliberately ignores the term, so *expired* is distinguishable from *never licensed*; the two need different remedies.

`SyncLicenseService` is the one writer — idempotent issue, revoke, revoke-all, reinstate. It *records* an authorization rather than making one (the decision belongs to the caller, as `PackageEntitlement` does), but refuses an issuance with no issuing principal.

**Acceptance is enforced.** `LicenseContent.RequiresAcceptance` and `LicenseAcceptance` shipped as node types with a body hash and **no callers at all**. `LicenseAcceptanceGate` now enforces them on both install paths, beside the entitlement check, so unattended installs are gated identically to a click. The body hash is *checked*: consent given against earlier terms does not satisfy revised ones. Normalization folds line endings and trailing whitespace and nothing else.

Permissive licences gate nothing — Apache-2.0 and MIT install exactly as before.

**`POST /api/instances/token`** exchanges the durable `mwi_` key for a short-lived, scoped `mwa_` token:

- carries identity and scope but **never authority** — the live grant is re-read every request, so a revocation lands at once rather than when the token expires;
- **can only narrow** — a token minted beyond its licence grants nothing extra;
- **cannot mint its successor** — only the durable key may exchange, or a minutes-long credential becomes perpetual by renewal.

Signed rather than stored (no per-issue write amplification, no expiry sweep) — only safe *because* authority is re-checked on use. It carries the hash of the key it came from, so it routes through the same index the raw key uses and a re-issued instance key invalidates outstanding tokens for free. Unset `PluginCatalog:TokenSigningKey` → 503 naming the config, never a weaker fallback.

## Compatibility

Additive. Existing grants, the durable-key path, and permissive-licence installs are unchanged; `PluginGrantTest` and the instance-registration suites pass untouched.

## Tests — 51 new

`SyncLicenseTermTest` (term/revocation rules) · `SyncAccessTokenTest` (cryptography, forgery, clamping, prefix disjointness) · `LicenseAcceptanceGateTest` (body-hash normalization) · `SyncTokenAuthenticationTest` (end-to-end against a real mesh: a token resolves, narrows, can never widen, and loses to a revoked or expired licence).

Both new invariants were verified **non-vacuous** by neutering them and watching the suite go red — 6/12 on the term check, and the narrowing test on the token scope.

Local: `-c Release -p:CIRun=true -warnaserror` clean on Mesh.Contract, PluginCatalog, Documentation and Memex.Portal.Shared. PluginCatalog.Test 461/461, Auth.Test instance+token 18/18, Documentation.Test 34/34.

## Follow-ups (deliberately not here)

- A consumer-side client for the exchange (CI can `curl` it today).
- Deriving licences from Store orders/coupons — the writer now exists for those callers to use.
- Retiring `MW_PLUGINS_REPO_TOKEN` in `MeshWeaver.Education`, whose **four** checkout sites (`e2e-install`, `e2e-mesh`, `e2e-journeys`, `publish-bake`) are what motivated this.

### DIFF
--- a/memex/Memex.Portal.Shared/Api/InstanceTokenEndpoints.cs
+++ b/memex/Memex.Portal.Shared/Api/InstanceTokenEndpoints.cs
@@ -137,7 +137,7 @@ public static class InstanceTokenEndpoints
     /// that has already ended — the grant would refuse it on use anyway, and handing out a token
     /// that cannot work is a worse answer than saying so.</para>
     /// </summary>
-    private static IReadOnlyCollection<string> EffectiveScope(
+    internal static IReadOnlyCollection<string> EffectiveScope(
         AuthenticatedInstance caller, IReadOnlyCollection<string>? requested, DateTimeOffset now)
     {
         if (caller.Grant.IsRevoked)
@@ -150,16 +150,38 @@ public static class InstanceTokenEndpoints
         if (requested is not { Count: > 0 })
             return [.. licensed.Select(e => e.ToString()).Distinct(StringComparer.Ordinal)];
 
-        // Keep only requests the licence actually covers. Dropping rather than refusing: a token can
-        // only narrow, so an over-broad request is a stale caller, not an attack.
-        return
-        [
-            .. requested
-                .Select(PluginGrantEntry.TryParse)
-                .Where(e => e is not null)
-                .Where(e => licensed.Any(l => l.Matches(e!.Source, e.PackageId)))
-                .Select(e => e!.ToString())
-                .Distinct(StringComparer.Ordinal)
-        ];
+        // Keep only what the licence actually covers, INTERSECTING rather than filtering. Dropping
+        // what is not covered rather than refusing it: a token can only narrow, so an over-broad
+        // request is a stale caller, not an attack.
+        //
+        // 🚨 A whole-source request must EXPAND to the licensed packages in that source, not be
+        // matched against them. `Plugins/*` is the natural way to say "everything I am licensed for
+        // here", and a naive `licensed.Any(l => l.Matches(source, "*"))` answers false for an
+        // instance licensed per-package — so the effective scope came out EMPTY and the exchange
+        // returned 403 to a caller asking a perfectly reasonable question (Copilot, #1844).
+        var effective = new List<string>();
+        foreach (var entry in requested.Select(PluginGrantEntry.TryParse).Where(e => e is not null))
+        {
+            if (entry!.PackageId == PluginGrantEntry.AllPackages)
+            {
+                // Everything licensed in that source — which may itself be a whole-source licence
+                // (kept as `Source/*`) or a set of per-package ones (kept individually).
+                effective.AddRange(licensed
+                    .Where(l => string.Equals(l.Source, entry.Source, StringComparison.OrdinalIgnoreCase))
+                    .Select(l => l.ToString()));
+                continue;
+            }
+
+            // Emit the LICENCE's source casing, not the request's — the response describes what was
+            // granted rather than echoing what was typed, and the two must not disagree with the
+            // wildcard branch above. The PACKAGE stays the requested one: taking the licence's would
+            // turn an exact request under a whole-source licence back into `Source/*`, widening the
+            // very scope the caller asked to narrow.
+            var match = licensed.FirstOrDefault(l => l.Matches(entry.Source, entry.PackageId));
+            if (match is not null)
+                effective.Add($"{match.Source}/{entry.PackageId}");
+        }
+
+        return [.. effective.Distinct(StringComparer.Ordinal)];
     }
 }
--- a/memex/Memex.Portal.Shared/Api/InstanceTokenEndpoints.cs
+++ b/memex/Memex.Portal.Shared/Api/InstanceTokenEndpoints.cs
@@ -1,0 +1,170 @@
+using System.Reactive.Linq;
+using System.Reactive.Threading.Tasks;
+using System.Text;
+using MeshWeaver.Mesh.Security;
+using MeshWeaver.Messaging;
+using MeshWeaver.PluginCatalog;
+using Microsoft.AspNetCore.Builder;
+using Microsoft.AspNetCore.Http;
+using Microsoft.AspNetCore.Routing;
+using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.Logging;
+
+namespace Memex.Portal.Shared.Api;
+
+/// <summary>
+/// <c>POST /api/instances/token</c> — a registered instance exchanges its durable <c>mwi_</c> key
+/// for a short-lived, scoped <c>mwa_</c> access token.
+///
+/// <para><b>Why it exists.</b> Without it a consumer must present its long-lived credential on every
+/// call, which is the property that makes a standing per-repo PAT unacceptable. With it, a build
+/// agent mints a minutes-long token narrowed to the one package it needs, and holds nothing durable
+/// beyond the run.</para>
+///
+/// <para>🚨 <b>A token cannot be exchanged for another token.</b> The endpoint accepts ONLY the
+/// durable instance key: allowing a token to mint its successor would turn a minutes-long credential
+/// into a perpetual one by renewal, defeating the entire point of the expiry. The check is on the
+/// credential's SHAPE (<c>mwi_</c> vs <c>mwa_</c>, disjoint by construction), not on a claim the
+/// presenter could edit.</para>
+///
+/// <para>The token narrows; it never widens. The effective scope is what the caller asked for
+/// intersected with what its sync licence already allows, and the live grant is re-read on every
+/// subsequent request — so revoking a licence takes effect at once rather than when the token
+/// expires.</para>
+/// </summary>
+public static class InstanceTokenEndpoints
+{
+    /// <summary>Maps the token-exchange endpoint. Call alongside <c>MapPluginRegistry</c>.</summary>
+    /// <param name="endpoints">The route builder.</param>
+    /// <returns>The same builder, for chaining.</returns>
+    public static IEndpointRouteBuilder MapInstanceTokenExchange(this IEndpointRouteBuilder endpoints)
+    {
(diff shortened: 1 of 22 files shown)


rbuergi:
Thanks — two of the three were real and are fixed in 5c9bc88; the third I'm keeping, with the reasoning below.

**1. Whole-source scope — fixed, and it was worse than "restrictive".** `EffectiveScope` matched `Plugins/*` against the licence instead of expanding it, so for an instance licensed per-package the effective scope came out **empty** and the endpoint answered **403** to the natural way of asking "everything I hold in this source". A whole-source request now expands to the licensed entries in that source.

Fixing it surfaced a second inconsistency the new test caught: the wildcard branch emitted the *licence's* source casing while the exact branch echoed the *request's*, so `plugins/Publish` and `PLUGINS/*` disagreed about what they'd been granted. The response now always reports the licence's casing — it describes what was granted, not what was typed — while keeping the **requested** package, since taking the licence's would turn an exact request under a whole-source licence back into `Source/*` and widen the very scope the caller asked to narrow.

**2. Basic-auth parse — fixed.** You're right, and `InstanceKeys.ExtractKey` had already solved it properly; I simply hadn't copied it. Now identical: a length cap plus `TryFromBase64String` into a stack buffer, with UTF-8 replacement fallback, so the reject path neither throws nor allocates unboundedly.

**3. The 1-minute authenticator cache — keeping, deliberately.** Correct observation, and the bound is smaller than it looks: the exposure is `InstanceRegistryAuthenticator.CacheDuration` (**1 minute**), *not* the token's lifetime. The token carries identity and scope but **never authority** — the live `PluginGrant` is re-read on every subsequent request — so a revoked licence stops working within the cache window rather than when the token expires. Bypassing the cache on the mint path would shave under a minute off that while adding a second read path to the one place that must fail closed. Happy to revisit if you think the mint path specifically warrants a fresh read.

Tests added: `SyncTokenScopeTest` (13) pins the whole intersection — expansion, source boundary, whole-source licences, expired entries, revocation, malformed entries, duplicates, casing — plus 5 asserting oversized/malformed/non-UTF-8 Basic payloads return null rather than throwing.

PR #1857, opened 2026-08-18
A renamed repository silently stopped syncing every Space — and the zero-match whispered (#1856)

### LINKED ISSUE #1856: A renamed repository silently stops syncing every Space — and the zero-match whispers at Information
## What happened

`Systemorph/education` was **renamed** to `Systemorph/MeshWeaver.Education`. GitHub redirects the old
name, so git, `gh` and the API all keep working — nothing looks broken anywhere.

But every course Space's sync config on `memex.meshweaver.cloud` still stores the **old** url:

```
AdvancedBusinessRules/_GitSync → repositoryUrl: https://github.com/Systemorph/education
AgenticEngineering/_GitSync    → repositoryUrl: https://github.com/Systemorph/Education
```

The webhook arrives under the repo's **new** name. `GitHubWebhookProcessor.RepoMatches`
(`src/MeshWeaver.GitSync/GitHubWebhookProcessor.cs:526`) parses both sides to `(Owner, Repo)` and
compares them as strings:

```csharp
&& string.Equals(owner, target.Owner, StringComparison.OrdinalIgnoreCase)
&& string.Equals(repo,  target.Repo,  StringComparison.OrdinalIgnoreCase)
```

Case-insensitive, so `education` vs `Education` was fine — but `education` vs `MeshWeaver.Education`
can never match. `MatchingSpaces` / `MatchingBuildTargets` return empty and nothing is imported.

## Measured impact

**Ten course Spaces stopped autosyncing on 2026-08-14 and served content 65 commits stale for four
days.** A learner installed a course on 2026-08-18 and got a cell with a C# syntax error that had
been fixed in git on 08-14. Every other repo kept working, because none of them was renamed.

## Why it took four days to notice — the second, worse half

A push that matches **no** Space logs at **Information** and does nothing:

```csharp
if (spaces.Count == 0)
{
    logger?.LogInformation("GitHub webhook for {Repo} matched no synced Space.", repoUrl);
    return Observable.Return(0);
}
```

The green-build path is even quieter: with zero configs matching the repo, `MatchingBuildTargets`
prints its audit line as *"43 sync config(s) in the mesh, 0 selected, 0 skipped"* — a line that
starts with a healthy-looking count and reports no skips, because no config ever got as far as
being *classified*.

The file already states the principle, three lines below the bug — it was applied to the per-config
skip list but **not** to the zero-match case:

> 🚨 Classify EVERY candidate and say what happened to it. A fan-out that reports only its winners
> cannot be audited: "updated 34" and "updated 34 of 43" look identical in the log, which is how
> #1326 stayed invisible for days.

So the failure that silences **every** Space of a repository is the one that whispers.

## What to build

1. **Make matching survive a rename.** Resolve the repository's canonical identity (GitHub's own
   `full_name`, which follows the rename redirect) instead of trusting the stored string, so the old
   and the new name resolve to the same repo. Cached — no network call on the hot path — and
   self-healing where possible, so the stored url stops lying.
2. **Make a zero-match LOUD.** A delivery that matches no sync config at all must log at
   **Warning**, naming the incoming repository AND what it was compared against, so the mismatch is
   diagnosable from one line.

## Operational note

The ten configs on memex still store the old url; a `patch` from a human account is correctly
refused (`user 'rbuergi' lacks Update permission` — these are system-synced Spaces where only
`system-security` is Admin).

rbuergi:
Fixes #1856.

## The defect

`Systemorph/education` was renamed to `Systemorph/MeshWeaver.Education`. GitHub redirects the old
name, so git, `gh`, the REST API and every manual sync kept working — nothing looked broken. But a
webhook payload always carries the repository's **current** name, and every course Space's
`_GitSync` still stored the old one:

```
AdvancedBusinessRules/_GitSync → repositoryUrl: https://github.com/Systemorph/education
AgenticEngineering/_GitSync    → repositoryUrl: https://github.com/Systemorph/Education
```

`RepoMatches` reduced both sides to `(Owner, Repo)` and compared them as strings, case-insensitively.
`education` vs `Education` matched fine; `education` vs `MeshWeaver.Education` never can. Ten course
Spaces stopped autosyncing on 2026-08-14 and served content **65 commits stale for four days**, while
every delivery reported success.

## Why it took four days — the half that matters more

A zero match logged at **Information** and stopped:

```csharp
logger?.LogInformation("GitHub webhook for {Repo} matched no synced Space.", repoUrl);
```

The green-build path was quieter still: with no config matching, `MatchingBuildTargets` printed
*"43 sync config(s) in the mesh, 0 selected, 0 skipped"* — a healthy-looking count with no skips,
because nothing ever got as far as being classified.

The file already states the principle, three lines below the bug. It was applied to the per-config
skip list but not to the zero-match case:

> 🚨 Classify EVERY candidate and say what happened to it. A fan-out that reports only its winners
> cannot be audited: "updated 34" and "updated 34 of 43" look identical in the log, which is how
> #1326 stayed invisible for days.

So the failure that silences *every* Space of a repository was the one that whispered.

## What this changes

**1. Matching survives a rename — as a fallback, never on the hot path.**
`ConfigsTargeting` is now the one matching seam both fan-outs (issues/comments and green build) funnel
through, so the tolerance and the report cannot drift apart between them. It compares stored strings
first — free, and correct for every repository that was never renamed. Only when **nothing** matched
does it ask GitHub what each stored url resolves to today: `GET /repos/{owner}/{repo}` follows the
rename redirect and answers with the repository's current `full_name`
(`IGitHubRepoClient.GetCanonicalRepository`, an `IIoPool`-bridged Octokit leaf — the interface carries
a `null`-returning default so no other implementation had to change).

**Caching:** `GitHubRepoIdentityResolver` holds the answers in the sanctioned `PromiseCache`, keyed per
repository, for **one hour**. Distinct stored repositories are grouped, so a repository ten Spaces sync
costs one lookup, and the token used is the config creator's credential → GitHub App → anonymous (a
private repository is only visible to the identity that already syncs it). The timestamp lives *inside*
the cached value, so the TTL check and the answer it governs can never disagree. Every entry expires —
an "unknown" as much as a resolved name — because a repository can be renamed again and a resolution
that failed under a weak identity must get another turn.

**2. Self-healing — yes, and here is the reasoning.** A config matched only by canonical identity is
**repointed** to the current url. The stale string is the actual defect: leaving it means consulting
the resolver forever for that repository and leaving the Space's GitHub settings displaying a name the
repository no longer has. The write is safe and bounded — it touches only `RepositoryUrl`, so the
RFC 7396 merge patch cannot clobber the `LastSyncCommitSha` the same delivery's import is about to
write; it is fire-and-forget under `RunAsSystem` so a failed repair never fails the delivery (a non-2xx
would make GitHub redeliver); and once repointed the string path matches, so the branch is never
re-entered. `RepointUrl` keeps the stored url's scheme and host, so a GitHub Enterprise config is never
silently moved to github.com.

**3. A zero match is LOUD.** `Warning`, naming both sides and the denominator:

```
GitHub webhook (green build of main) for Systemorph/MeshWeaver.Education matched NONE of the 43
sync config(s) in the mesh — nothing will sync for this delivery. Compared against:
Systemorph/education → Systemorph/MeshWeaver.Education, Systemorph/MeshWeaver.Plugins, … If this
repository was RENAMED, the stored url is stale: GitHub redirects the old name so every other
operation keeps working, while the payload always carries the CURRENT name and can never
string-match the old one. Otherwise the hook is installed on a repository this mesh does not sync.
```

The two "nothing happened" outcomes are now distinguishable, which is the whole point: *no source
NEEDS updating* (already at this commit, wrong branch, export-only) stays at Information; *no source
TARGETS this repository at all* is a Warning. The audit line also gained `{Targeting}`, so
"43 in the mesh, 0 selected, 0 skipped" can no longer read as healthy. A repointed config announces
itself at Warning too — a config changing under a mesh must be readable in the log that changed it.

## Tests — and each one fails against `main`

`test/MeshWeaver.GitSync.Test/GitHubWebhookRepoRenameTest.cs`, six tests. `FakeGitHubRepoClient`
gained a `Rename(old, new)` that models GitHub's *total* redirect: `Key()` resolves through it, so both
names address one stored repository and every operation but equality keeps working — which is exactly
what makes the drift invisible.

| Test | Pins |
|---|---|
| `GreenBuild_UnderTheNewName_MatchesAConfigStoringTheOldOne_AndRepointsIt` | the defect: a green build under the new name imports into a Space whose config says the old one, and both configs are repointed |
| `NoConfigTargetsTheRepository_IsReportedAtWarning_NamingBothSides` | Warning, both sides, and the `NONE of the N` denominator |
| `CaseOnlyDifference_StillMatches_WithoutAskingGitHubAnything` | regression guard — and that the free path stays free (`CanonicalLookups` unchanged) |
| `ResolvedIdentity_IsCached_ButExpires_SoAStaleNameCannotStick` | one lookup inside the TTL, a fresh one past it, against a hand-moved clock |
| `RepointUrl_RewritesOwnerAndRepo_ButNeverTheHost` | a GHE config is never moved to github.com |
| `RepoIdentity_MatchesCaseInsensitively_ButARenameIsNotACasingDifference` | why the lookup has to exist at all |

**Non-vacuity, measured — controls run on this machine, then reverted:**

*Control A — `main`'s behaviour restored (no canonical fallback; zero match at Information):*
```
Failed GreenBuild_UnderTheNewName_MatchesAConfigStoringTheOldOne_AndRepointsIt [30 s]
Failed NoConfigTargetsTheRepository_IsReportedAtWarning_NamingBothSides [30 s]
Failed!  - Failed: 2, Passed: 4, Skipped: 0, Total: 6
```

*Control B — TTL expiry and host preservation neutered:*
```
Failed RepointUrl_RewritesOwnerAndRepo_ButNeverTheHost [84 ms] — Assert.Equal() Failure: Strings differ
Failed ResolvedIdentity_IsCached_ButExpires_SoAStaleNameCannotStick [85 ms] — Assert.Equal() Failure: Strings differ
Failed!  - Failed: 2, Passed: 4, Skipped: 0, Total: 6
```

*With the fix:* `Passed! - Failed: 0, Passed: 181, Skipped: 0, Total: 181` (whole
`MeshWeaver.GitSync.Test` suite), and `MeshWeaver.Documentation.Test` green.

## Operational note (not fixed here)

The ten configs on memex still store the old url — a `patch` from a human account is correctly refused
(`lacks Update permission`; these are system-synced Spaces where only `system-security` is Admin). No
manual repoint is needed **once this is deployed**: the next green build of MeshWeaver.Education
resolves the stored name, imports, and writes the current url back. Until then they stay stale, so a
manual "Update to latest" per Space is the stopgap. If the App identity cannot read the repository, the
new Warning says so by name instead of staying silent.


rbuergi:
**CI status: green except a CI-infrastructure failure that is not this branch.**

Shard 2 failed in `Provision the headless browser (infra, not a test)` — `playwright install` hung for 180 s on attempt 1 and then failed attempts 2 and 3, *before any test ran*. The shard therefore produced no results and `Consolidate test results` correctly refused to report success on a partially-executed suite.

Evidence it is environmental, not this diff:

- The same step is failing right now on `fix/ci-apt-lock-wait` — the branch opened to fix exactly this provisioning problem (run 32171217048).
- Re-run once (the sanctioned infrastructure exception) and it failed *identically*, so it is a reproducible environment problem, not a flake. Not re-running again.
- This diff touches `src/MeshWeaver.GitSync/*`, two Architecture docs and a What's New note. It cannot reach a browser download.

Coverage that matters here IS verified: `MeshWeaver.GitSync.Test` is assigned to **shard 5, which passed**. Shard 2 carries `MeshWeaver.Documentation.Test` (the other project this PR touches) — run locally green: 37/37, including `DocumentationLinkIntegrityTest`, `WhatsNewEntryIntegrityTest` and `ImpersonationScopeSiteRatchetGuard`. Full `MeshWeaver.GitSync.Test` locally: 181/181.

Not merging — leaving this for a maintainer once the browser-provisioning gate is healthy.

Issue #1843, opened 2026-08-18
MeshHubDisposalLeakTest / NodeTypeRecompileAlcLeakTest fail on shard 2 independently of the diff


rbuergi:
## Evidence that this is not caused by the change under test

`MeshHubDisposalLeakTest.MeshHub_IsCollected_AfterMeshAndServiceProviderDisposal` failed on **two different PRs**, and one of them cannot possibly affect it:

| PR | diff | shard 2 |
|---|---|---|
| #1838 | C# (Graph, Mesh.Contract, Social + tests) | ❌ `MeshHubDisposalLeakTest` |
| **#1841** | **two YAML workflow files, zero C#** | ❌ `MeshHubDisposalLeakTest` + `NodeTypeRecompileAlcLeakTest` |

**#1841 changes no compiled code whatsoever** — it edits `.github/workflows/node-repo-tag-modules.yml` and adds `node-repo-platform-ref-bump.yml`. A YAML-only diff cannot change GC reachability. Meanwhile `main` runs at the same time are green.

## What the failure says

```
Expected value to be false because the mesh hub is pinned by a NON-stack root
(static field / TimerQueue timer / GC handle) …, but found True

[clrmd] GC-ROOT PATH (root → … → mesh hub):
   System.Object[]  (via .ROOT[StrongHandle] System.Object[])
   System.Reactive.Concurrency.ConcurrencyAbstractionLayerImpl+PeriodicTimer
   System.Threading.Timer
   System.Threading.TimerHolder
   System.Threading.TimerQueueTimer   ← ~25 more of these
   …
```

The root is an Rx `PeriodicTimer`, and the path then walks the **process-global `TimerQueue` linked list** — every pending timer in the process is chained there. So the reported "root" depends on which unrelated timers happen to be pending when the snapshot is taken, which makes the result sensitive to co-scheduled tests and runner load rather than to the hub's own retention.

`ActivityControlPlaneResubscribeTest` already records this test as one that *"intermittently caught a hub surviving"*, and `ClrMdDacPin` documents repeated DAC load/unload cycles around it.

## Why it matters beyond the noise

This is a **leak** gate. A test that fails on unrelated PRs trains everyone to re-run it, and the next time it catches a real leak it will be re-run too. Either the chain-walk should stop at the TimerQueue boundary (a timer queue link is not a retention *cause*), or the snapshot should be taken with the queue quiesced.

[...]

PR #1862, opened 2026-08-18
One apt-lock wait, shared — clients.yml had none that worked

rbuergi:
Follow-up to #1858, from its own review.

## What was still exposed

`clients.yml` runs the same `playwright install --with-deps` with the **old** shape:

```bash
sudo flock -w 60 /var/lib/apt/lists/lock true 2>/dev/null || true
```

That acquires the lock and releases it in the same breath — **it reserves nothing**, so the runner's background apt re-takes it before playwright's `apt-get` gets there. The fix landed in `dotnet-test.yml` while this workflow stayed exposed to identical contention.

And two copies of this logic drift. The copy that drifts is the one nobody notices, until a shard reddens on a PR whose diff cannot touch tests.

## The change

Extracted to **`.github/scripts/wait-for-apt.sh`**, used by both workflows.

The script carries the *why* — that checking a lock and releasing it loses the race (measured: the very next line still failed on the same holder), and that apt locks with **fcntl**, which does not interoperate with `flock` at all. Keeping the rationale in one place is the point: it cannot go stale in one copy while the other keeps it.

Bounded exactly as before (default 240 s; `dotnet-test` passes `APT_LOCK_WAIT`), and exits 0 either way — a timed-out wait is not a reason to skip an install that is bounded on its own.

Verified: both workflows YAML-parse, every `run:` block passes `bash -n`, and the script passes `bash -n`.

Refs #1843

https://claude.ai/code/session_01LFx1SsoKDhRKHdaJbQPEWQ

### DIFF
--- a/.github/scripts/wait-for-apt.sh
+++ b/.github/scripts/wait-for-apt.sh
@@ -1,0 +1,26 @@
+#!/usr/bin/env bash
+# Wait until no apt/dpkg process is running, so a following `apt-get` (usually via
+# `playwright install --with-deps`) can take the lock instead of losing a race to the runner
+# image's own background apt (unattended-upgrades / the daily timer).
+#
+# 🚨 WAIT FOR THE HOLDER TO FINISH — do not "check the lock". The shape this replaced was
+#   sudo flock -w N /var/lib/apt/lists/lock true
+# which ACQUIRES the lock and releases it in the same breath, reserving nothing: the holder
+# re-takes it before playwright's apt-get gets there. Measured — after that wait reported the
+# lock free, the very next line still failed with
+#   E: Could not get lock /var/lib/apt/lists/lock. It is held by process 3014 (apt-get)
+# apt also locks with fcntl, which does not interoperate with flock, so that shape could not
+# have been reliable even without the release race. See #1843 / #1858.
+#
+# Bounded by the first argument (default 240s): it can wait, it cannot hang. Exits 0 either way —
+# a timed-out wait is not a reason to skip the install, which is bounded on its own.
+set -euo pipefail
+budget="${1:-240}"
+waited=0
+while [ "$waited" -lt "$budget" ]; do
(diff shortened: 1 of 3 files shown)


rbuergi:
Heads-up: **#1855 landed this same change ~20 minutes ago** (merged `f7cde965a`) — same diagnosis, same three files, even the same `.github/scripts/wait-for-apt.sh` filename. We were clearly working the same finding from #1858's review in parallel; sorry for the collision, mine merged first only by timing.

What is on `main` now:

- `.github/scripts/wait-for-apt.sh` — waits for the competing **process** to exit (`pgrep` on `apt-get` / `apt` / `dpkg` / `unattended-upgr`), bounded, and degrades **loudly** (warns when `pgrep` is missing or the holder outlives the budget, so a wait that could not run is never mistaken for "apt was free").
- Both `dotnet-test.yml` and `clients.yml` call it — `$GITHUB_WORKSPACE`-anchored, because the clients job's `working-directory: clients/react-native` breaks a repo-relative path.
- Separately: `clients.yml`'s install bound went 180s → 420s. That job was failing `exit code 124` on `main` too (run `32170659820`), because a full `--with-deps chromium` install measurably costs ~185s — the bound was below the normal path.

So this PR is likely redundant now. Worth a diff against `main` before closing, in case your version covers something mine does not — I would not want to lose it if it does.

rbuergi:
Superseded by 6755415b8 ("CI: clients.yml still had the apt `flock` no-op — one wait, two callers"), which landed the same extraction independently and went further: it also raised the clients install bound from 180s to 420s, with two measured runs showing a healthy full-chromium install dying at exactly `exit code 124` a few seconds short. That is the same lesson one layer down — a bound the normal path exceeds is not a hang detector, it is a flake generator, and it reports as the very "infra failure" wording that trains people to re-run rather than look.

Nothing here is worth carrying over: main's version is a superset. Closing rather than merging a redundant change.

Issue #1833, opened 2026-08-18
Flake: ShippedPrebuiltBundlesTest proves "the seeder skipped" with MeshNode.Version, which the NodeType's own sources watcher also bumps — 1 in 15 whole-assembly runs


rbuergi:
`ShippedPrebuiltBundlesTest.UnchangedBundle_SkipsWithoutRewriting_AndReSeedsWhenTheStoreLosesItsBytes`
fails on whole-assembly runs of `MeshWeaver.PluginCatalog.Test`, always with the same shape:

```
Expected value to be 2 because an already-adopted entry must not be re-stamped — the write is
what activates the NodeType's per-node hub, and that activation is the whole cost being removed,
but found 3.
```

`ShippedPrebuiltBundlesTest.cs:147` — `afterSecond.Version.Should().Be(adoptedVersion, …)`.

This is the boot-time bundle **adoption** path (#1660 WS1 / #1707 slice 3), i.e. the decision that
made the difference between `alreadyBaked=0` and a warm start in #1814, so 25 % of the signal
missing from that gate is worth more than a normal test flake. Part of the flake population in
#1384.

## Measured (macOS, 18 cores, Release, xUnit v3 in-process exe)

| run shape | iterations | failures |
|---|---|---|
| **whole assembly** | **15** | **1 (6.7 %)** |
| `-class ShippedPrebuiltBundlesTest` | 60 | **0** |

Whole-assembly only — the same *shape* as #1826, but (see below) **not the same mechanism**.

## What actually happens

Instrumenting the test to record every emission of the NodeType's mesh-node stream, with the
seeder's own fields, makes the sequence explicit. A **passing** run:

```
 53 EMIT  v=3 status=Ok lcv=1 succAt=…234395 csv=0 cs=null   ← the adoption lands at v3, stamping
 53 MARK  after first SeedAll                                   LastCompiledVersion = 1
 54 MARK  captured adoptedVersion=3
 55 MARK  second SeedAll -> 1
 55 MARK  afterSecond.Version=3                              ← nothing emitted in between: SKIPPED
 68 EMIT  v=4 status=Ok lcv=3 succAt=…256234 csv=0 cs=0      ← step 3's deliberate re-adoption
```

Read the numbers: `Seed` read the node at **v1** (that is what it stamps as `LastCompiledVersion`
and what it uses as the assembly-store key), yet its write committed as **v3**. Something else
[...]

PR #1869, opened 2026-08-18
The re-drive barrier was satisfied by a REPLAYED pre-forge emission

### LINKED ISSUE #1866: NeverCompiledFailureRedriveTest: the forge barrier is satisfied by a REPLAYED pre-forge emission, so the re-drive bound is unprotected
`NeverCompiledFailureRedriveTest`'s bounded assertions are anchored on a barrier that a **replayed** emission satisfies. The barrier passes whether or not the write it is a barrier for ever landed, so everything downstream of it — including the negative assertion that pins the automatic re-drive's **bound** — is anchored at a point that can precede the state it is claiming about.

That bound is load-bearing. `ABrokenSource_IsReDrivenExactlyOnce_AndThenLeftAlone` is what stops a permanently-broken NodeType from spinning in an unbounded re-drive loop on the per-NodeType hub's single-threaded action block — the recompile-storm wedge the park registry exists to prevent. An assertion that can be satisfied spuriously leaves that unprotected.

## 1. The barrier is structurally vacuous — not a race, an identity

#1850 replaced "wait to SEE the forged record" (transient by construction) with "wait for the node's **Version** to reach the forge's write":

```csharp
var forgedVersion = forged.Version;
await workspace.GetMeshNodeStream(typePath)
    .Should().Within(20.Seconds())
    .Match(n => n.Version >= forgedVersion);
```

`forged` is what `GetMeshNodeStream(path).Update(...)` returned. For a foreign path that is `UpdateRemote`, which emits **`update(current)`** — the *optimistic local snapshot* (`MeshNodeStreamExtensions.EmitTerminal` → `observer.OnNext(updated)`; the queue's own comment says it "does NOT wait for the patch's echo from the owner"). The forge lambda rewrites `Content` only, so the returned node carries the **pre-write** `Version`.

And `GetMeshNodeStream(path)` is a **`Replay(1)`** mirror (`MeshNodeStreamCache` — `new ReplaySubject<MeshNode>(1)` per entry): every new subscription is handed the snapshot the mirror already holds.

So `n.Version >= forged.Version` is a bound the stream **is already sitting on**, and the replayed pre-forge snapshot satisfies it instantly. The barrier's stated purpose — "a later `Match` still cannot match pre-forge state" (#1850) — does not hold.

Measured on the real mesh, printed by the barrier itself, twice in one run:

```
Forged the never-compiled Error record on type/RedriveBounded8e88d046… (marker FORGED-de842257…);
  the mirror was replaying v6, Update handed back v6 (the SAME revision — so the old '>= 6' bound
  was already satisfied), owner then moved to v7.
Forged the never-compiled Error record on type/RedriveRecovers31ae3494… (marker FORGED-bf2f910a…);
  the mirror was replaying v7, Update handed back v7 (the SAME revision — so the old '>= 7' bound
  was already satisfied), owner then moved to v8.
```

The consequence in `ABrokenSource_…`: the very next `Match` waits for `Error && FailedBuildInputs` — which for a type that cannot compile is **byte-identical** to the record that was already there before the forge. Satisfied by the replay, the test believes it watched the re-drive fire and re-fail having observed only pre-forge state.

## 2. …and the negative assertion has no anchor at all

```csharp
await Mesh.GetWorkspace().GetMeshNodeStream(typePath)
    .Where(n => n?.Content is NodeTypeDefinition d
        && d.CompilationStatus is CompilationStatus.Pending or CompilationStatus.Compiling)
    .Should().NotEmit(window, because);
```

Its docstring claims non-vacuity ("the sibling assertions above prove the same subscription DOES see the flip"). They do not: the siblings are different subscriptions and none of them ever asserts that a `Pending` or `Compiling` value reached a subscriber at all.

The window the test *means* opens where the caller stands. The window `NotEmit` *observes* opens when its `SubscribeOn(TaskPoolScheduler)` attach lands — and `Replay(1)` collapses everything in between into one snapshot. **A complete re-drive (`Pending → Compiling → Error`) fits in that gap**, and for a broken type the state it re-settles at is indistinguishable from the state the window opened on. The assertion then passes having watched the exact thing it forbids.

## Reproduction — the constructed replay, both halves

`RedriveObservationReplayTest` (new, hermetic — a real `ReplaySubject<MeshNode>(1)` and a real `NodeTypeCompileParkRegistry`, no mesh, 4.4 s):

```
Passed  TheOldForgeBarrier_IsSatisfiedByTheReplayedPreForgeSnapshot
  Update returned version 7; the mirror already holds version 7. The old bound is '>= 7'.
  OLD BARRIER ACCEPTED the replay: version 7,
    CompilationError='CS0103: The name 'ThisSymbolDoesNotExist' does not exist',
    FailedBuildInputs='fw=1.0|mods=abc|src=42' — no forge in it.
  NEW BARRIER REJECTED the same replay (watermark v7).
  NEW BARRIER ACCEPTED the genuine owner-minted revision v8.

Passed  TheOldNegativeAssertion_IsSatisfiedByAReplayThatHidesAWholeReDrive
  OLD NEGATIVE ASSERTION PASSED while the ledger recorded re-drive 2 and attempt 2 —
    a FALSE GREEN on the bound that keeps a permanently-broken NodeType out of an unbounded
    retry loop.
  NEW NEGATIVE ASSERTION REJECTED the same replay: Expected value to be 1 … but found 2.
```

**"It passes" proves nothing here**: `NeverCompiledFailureRedriveTest` is **10/10 green locally** on `origin/main` at 75a70c20f (class-level, 10 iterations) — which is precisely why the defect needs a constructed replay rather than a repeat run.

## The fix

**A version watermark, plus the ledger as the primary bound.**

- **Watermark** — the barrier becomes strict (`Version > watermark`), and the anchor is threaded into every observation downstream of the forge. 🚨 Whose version: the **owner's**, and it is a strict *lower bound*, not a receipt — `MeshNode.Version` is minted by the owning hub for **every** writer (`NextVersion` = `current + 1`; #1833 is a test that flaked on assuming it tracks the caller's own write). It claims exactly one thing: the owner has committed at least one revision past the state the write was diffed against, so no later emission can be that state replayed.
- **Ledger (primary)** — `NodeTypeCompileParkRegistry` counts *events*: `RecordFailureRedrive` per automatic kickoff, `RecordAttempt` per real Roslyn dispatch (`RunCompile`). A counter cannot be replayed, cannot be conflated, and — decisively — can tell "the re-drive ran and re-failed" apart from "I was handed the pre-forge snapshot back", which for a broken type are the same record. The stream observation stays as independent corroboration.

Context: #1825 added the test, #1850 fixed the *first* race in the same barrier (waiting to observe a state designed to be overwritten). This is the second one it left behind.

rbuergi:
Closes #1866.

`NeverCompiledFailureRedriveTest`'s bounded assertions were anchored on a barrier that a **replayed** emission satisfies, so the assertion pinning the automatic re-drive's **bound** — `ABrokenSource_IsReDrivenExactlyOnce_AndThenLeftAlone`, which is what keeps a permanently-broken NodeType out of an unbounded retry loop on the per-NodeType hub's single-threaded action block — could pass without observing anything.

#1825 added the test; #1850 fixed the *first* race in the same barrier (it waited to *see* a record that is transient by construction). This is the second one it left behind.

## The defect — not a race, an identity

```csharp
var forgedVersion = forged.Version;                       // ← what Update() returned
await workspace.GetMeshNodeStream(typePath)
    .Should().Within(20.Seconds())
    .Match(n => n.Version >= forgedVersion);
```

- `Update` on a foreign path is `UpdateRemote`, which emits **`update(current)`** — the *optimistic local snapshot*; the queue "does NOT wait for the patch's echo from the owner". The forge lambda rewrites `Content` only, so the returned node carries the **pre-write** `Version`.
- `GetMeshNodeStream(path)` is a **`Replay(1)`** mirror (`new ReplaySubject<MeshNode>(1)` per cache entry): a new subscription is handed the snapshot already held.

So `>= forged.Version` is a bound **the stream is already sitting on**. The barrier passes whether or not the write landed, and everything downstream is anchored before the forge. In `ABrokenSource_…` the next `Match` waits for `Error && FailedBuildInputs` — for a type that cannot compile, byte-identical to the record that was already there.

Printed by the barrier itself on the real mesh, twice in one run:

```
the mirror was replaying v6, Update handed back v6 (the SAME revision — so the old '>= 6' bound
was already satisfied), owner then moved to v7.
the mirror was replaying v7, Update handed back v7 (the SAME revision — so the old '>= 7' bound
was already satisfied), owner then moved to v8.
```

And the negative assertion had **no anchor at all**. The window it *means* opens where the caller stands; the window `NotEmit` *observes* opens when its `SubscribeOn(TaskPoolScheduler)` attach lands — and `Replay(1)` collapses everything between into one snapshot. A whole `Pending → Compiling → Error` cycle fits in that gap. Its docstring claimed non-vacuity on the strength of sibling assertions that are different subscriptions and never assert a `Pending`/`Compiling` value reached anyone.

## The fix — a version watermark, plus the ledger as the primary bound

New `RedriveObservation` / `RedriveWindow` (test-side vocabulary, shared by the mesh tests and the proof):

- **Watermark.** The barrier becomes strict (`Version > watermark`) and the anchor is threaded into every observation downstream of the forge. 🚨 **Whose version:** the **owner's**, and a strict *lower bound*, not a receipt — `MeshNode.Version` is minted by the owning hub for **every** writer (`NextVersion` = `current + 1`; #1833 flaked on assuming it tracks the caller's own write). It claims exactly one thing: the owner committed at least one revision past the state this write was diffed against, so no later emission can be that state replayed. Said in those words in the code.
- **Ledger (primary).** `NodeTypeCompileParkRegistry` counts *events* — `RecordFailureRedrive` per automatic kickoff, `RecordAttempt` per real Roslyn dispatch (`RunCompile`). A counter cannot be replayed or conflated and can tell "the re-drive ran and re-failed" apart from "I was handed the pre-forge snapshot back". The window captures both counters where the *caller* stands, not where `NotEmit` happens to attach. The stream observation stays as independent corroboration.

## Non-vacuity — both halves, because "it passes" proves nothing here

`NeverCompiledFailureRedriveTest` is **10/10 green locally on `origin/main`** (class-level, 10 iterations, 75a70c20f). A false green survives its own fix, so the evidence is a **constructed replay**, not a repeat run.

`RedriveObservationReplayTest` (new; a real `ReplaySubject<MeshNode>(1)`, a real `NodeTypeCompileParkRegistry`, no mesh — 4.4 s):

```
Passed  TheOldForgeBarrier_IsSatisfiedByTheReplayedPreForgeSnapshot
  Update returned version 7; the mirror already holds version 7. The old bound is '>= 7'.
  OLD BARRIER ACCEPTED the replay: version 7,
    CompilationError='CS0103: The name 'ThisSymbolDoesNotExist' does not exist',
    FailedBuildInputs='fw=1.0|mods=abc|src=42' — no forge in it.
  NEW BARRIER REJECTED the same replay (watermark v7).
  NEW BARRIER ACCEPTED the genuine owner-minted revision v8.

Passed  TheOldNegativeAssertion_IsSatisfiedByAReplayThatHidesAWholeReDrive
  OLD NEGATIVE ASSERTION PASSED while the ledger recorded re-drive 2 and attempt 2 —
    a FALSE GREEN on the bound that keeps a permanently-broken NodeType out of an unbounded
    retry loop.
  NEW NEGATIVE ASSERTION REJECTED the same replay: Expected value to be 1 … but found 2.

Passed  TheNewNegativeAssertion_AcceptsAQuietWindow          ← the fix is not "assert false"
Passed  TheNewNegativeAssertion_CatchesADriveThatOnlyTheMirrorSaw ← the two witnesses are independent
Passed  TheWatermarkRejectsOnlyTheReplay                     ← the anchor rejects the replay, not newer state
```

## Verification

| | |
|---|---|
| `dotnet build … -c Release -p:CIRun=true -warnaserror` | 0 Warning(s), 0 Error(s) |
| `NeverCompiledFailureRedriveTest`, class-level, BEFORE (origin/main) | **10/10 pass** |
| `NeverCompiledFailureRedriveTest` + `RedriveObservationReplayTest`, class-level, AFTER | ****10/10 pass** (80 test executions)** |

No "What's New" entry: test-only change, no user-visible effect.

### DIFF
--- a/test/MeshWeaver.Hosting.Monolith.Test/NeverCompiledFailureRedriveTest.cs
+++ b/test/MeshWeaver.Hosting.Monolith.Test/NeverCompiledFailureRedriveTest.cs
@@ -38,6 +38,15 @@ namespace MeshWeaver.Hosting.Monolith.Test;
 ///
 /// <para>The pure half — the verdict-inputs token, the predicate, the ledger — is
 /// <c>FailedVerdictRedriveTest</c> in MeshWeaver.Graph.Test.</para>
+///
+/// <para>🚨 <b>The bounded assertions here are not self-certifying.</b> Every observation runs over
+/// <c>GetMeshNodeStream</c>, a <c>Replay(1)</c> mirror that hands each new subscription the snapshot
+/// it already holds and keeps nothing before it — so an assertion can be satisfied by a REPLAY of
+/// state that predates the thing it is claiming about, and it then passes for the wrong reason
+/// forever. <see cref="RedriveObservation"/> is the vocabulary that closes it (a version watermark,
+/// with the process ledger as the primary bound); <c>RedriveObservationReplayTest</c> is what proves
+/// it, by constructing the replay and pinning BOTH halves — the shape this file used to have
+/// accepting it, and this one refusing it while still accepting the genuine case.</para>
 /// </summary>
 public class NeverCompiledFailureRedriveTest(ITestOutputHelper output) : MonolithMeshTestBase(output)
 {
@@ -70,14 +79,17 @@ public class NeverCompiledFailureRedriveTest(ITestOutputHelper output) : Monolit
         var (typePath, baselineSucceededAt) = await CompileBaselineType("RedriveRecovers", GoodSource);
         var workspace = Mesh.GetWorkspace();
 
-        await ForgeNeverCompiledFailure(typePath);
+        var forgeWatermark = await ForgeNeverCompiledFailure(typePath);
 
         // The OWNER-side re-drive must rebuild it: back to Ok, with real assembly coordinates and a
-        // STRICTLY NEWER success timestamp (a replayed old Ok would prove nothing). Before the fix
-        // nothing re-drives this record at all, so this times out.
+        // STRICTLY NEWER success timestamp (a replayed old Ok would prove nothing) at a revision
+        // STRICTLY past the forge, so the mirror's Replay(1) hand-off of the pre-forge snapshot
+        // cannot satisfy it either. Before the fix nothing re-drives this record at all, so this
+        // times out.
         var recovered = await workspace.GetMeshNodeStream(typePath)
             .Should().Within(90.Seconds())
-            .Match(n => n.Content is NodeTypeDefinition d
+            .Match(n => RedriveObservation.IsPastWatermark(n, forgeWatermark)
+                && n.Content is NodeTypeDefinition d
                 && d.CompilationStatus == CompilationStatus.Ok
                 && !string.IsNullOrEmpty(d.LatestAssemblyPath)
                 && d.CompiledFrameworkVersion == NodeTypeCompilationHelpers.FrameworkVersion
@@ -90,7 +102,7 @@ public class NeverCompiledFailureRedriveTest(ITestOutputHelper output) : Monolit
         // failure look like it had already had its attempt), and nothing re-drives a healthy type.
         ((NodeTypeDefinition)recovered.Content!).FailedBuildInputs.Should().BeNull(
             "a success retires the failure verdict, so its inputs must go with it");
-        await AssertNoFurtherCompileIsDriven(typePath, 15.Seconds(),
+        await AssertNoFurtherCompileIsDriven(typePath, recovered.Version, 15.Seconds(),
             "a recovered type must not be re-driven again — the re-drive's own bookkeeping is what "
             + "makes its trigger false, and a trigger that survives its own write is a write cycle");
     }
@@ -124,20 +136,30 @@ public class NeverCompiledFailureRedriveTest(ITestOutputHelper output) : Monolit
 
         // Now stage the pre-fix record (no stamp) so the re-drive DOES fire, and prove it fires
         // exactly once: the attempt fails again, and nothing schedules another.
-        await ForgeNeverCompiledFailure(typePath);
+        var forgeWatermark = await ForgeNeverCompiledFailure(typePath);
 
-        await workspace.GetMeshNodeStream(typePath)
+        // 🚨 ANCHORED on the forge. Without the version clause this Match is satisfied by the
+        // mirror's Replay(1) hand-off of the PRE-forge record — which, for a type that cannot
+        // compile, is byte-identical to the record the re-drive re-settles at (Error, inputs
+        // stamped). The test would then believe it had watched the re-drive fire and re-fail having
+        // observed only state that predates the forge, and every bound below it would be anchored
+        // in the past. The forge nulls FailedBuildInputs, so a record carrying it again at a
+        // revision past the forge can only be the re-drive's own verdict.
+        var reFailed = await workspace.GetMeshNodeStream(typePath)
             .Should().Within(90.Seconds())
-            .Match(n => n.Content is NodeTypeDefinition d
+            .Match(n => RedriveObservation.IsPastWatermark(n, forgeWatermark)
+                && n.Content is NodeTypeDefinition d
                 && d.CompilationStatus == CompilationStatus.Error
                 && !string.IsNullOrEmpty(d.FailedBuildInputs));
         var redrives = ParkRegistry.GetFailureRedriveCount(typePath);
         redrives.Should().Be(1,
             "exactly one automatic attempt per distinct set of compile inputs — the flip to Pending "
             + "stamps those inputs in the same write, so the trigger cannot re-arm itself");
-        Output.WriteLine($"{typePath} re-driven {redrives}× and re-failed — now bounded.");
+        Output.WriteLine(
+            $"{typePath} re-driven {redrives}× and re-failed at v{reFailed.Version} "
+            + $"(forge watermark v{forgeWatermark}) — now bounded.");
 
-        await AssertNoFurtherCompileIsDriven(typePath, 20.Seconds(),
+        await AssertNoFurtherCompileIsDriven(typePath, reFailed.Version, 20.Seconds(),
             "a type whose source cannot compile must back off after its one attempt instead of "
             + "storming Roslyn on the hub's action block");
         ParkRegistry.GetFailureRedriveCount(typePath).Should().Be(1,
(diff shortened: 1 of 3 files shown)


rbuergi:
CI run 32184462421 came back red on **`Run tests (shard 2)`** — it is **#890**, not this diff. Re-ran the failed shard.

The framework's own canary names the verdict in the failure text:

```
canary=BELOW-ROSLYN
  shared:   THREW NullReferenceException at NamedTypeSymbol.Microsoft.Cci.ITypeDefinitionMember.get_ContainingTypeDefinition
  pristine: THREW NullReferenceException at NamedTypeSymbol.Microsoft.Cci.ITypeDefinitionMember.get_ContainingTypeDefinition
— a trivial compilation with BRAND-NEW references and freshly parsed source cannot emit either,
and BOTH legs died in the same frame ⇒ nothing about the reference set explains this; the broken
state is below Roslyn (CLR heap / JIT / GC), so no reference-set change can fix it
```

That is #890 verbatim — *"one Roslyn Emit poisons the whole process — every later compile NREs in `Cci.MetadataWriter.GetConsolidatedTypeParameters`"*.

Four things say it is not the branch:

1. **11 tests across 6 unrelated classes** went down together — `CompileLeafStabilityTest`, `LinkedInProfileLayoutAreaTest`, `LinkedInTelemetryImport…`, `FrameworkStale*`, `NodeAssemblyContexts*`, `OverlaidInstance_SelfRecycles…` — every one of them compile-dependent, none of them touched here.
2. **The process was already poisoned before this class ran.** First casualty 21:03:52 (`CompileLeafStabilityTest`); `NeverCompiledFailureRedriveTest` started 21:08:20.
3. `ABrokenSource_IsReDrivenExactlyOnce_AndThenLeftAlone` — the test at the centre of this PR — **passed**, because it is the one that *expects* the compile to fail.
4. **Shard 2 also failed on `main` today** (run 32178235661, 19:43Z, before this branch existed).

And the diff cannot reach it: three test files, no `src/` change, and the new `RedriveObservationReplayTest` boots no mesh and runs no Roslyn at all.

Issue #1814, opened 2026-08-17
PROD memex: every course cover serves the 'type has not finished compilation' fallback — Store/Plugin build has not settled


rbuergi:
## Symptom, live right now

`https://memex.meshweaver.cloud/AgenticEngineering` — and, by construction, every other `Store/Plugin`-typed root — serves the compilation fallback to an **anonymous** visitor instead of the course cover:

> ⚠️ **This page can't be displayed**
> This item's type has not finished **compilation** yet, so its page couldn't be built.
> **NodeType 'Store/Plugin' build did not settle within 60s.**
> `Instance 'AgenticEngineering' is rendering this fallback until the type's build settles.`

This is the 60 s activation budget described in AGENTS.md ("a NodeType left at CompileError refuses portal readiness and parks every instance hub for the full 60 s activation budget"), reached on prod.

## Reproduced from two independent places

* **CI, on `main`, twice in a row** — MeshWeaver.Education's `Read-only smoke against a live instance` job (its only purpose is probing this deployment): [run 532 attempt 1](https://github.com/Systemorph/MeshWeaver.Education/actions/runs/32068396627) (288 s) and attempt 2 (712 s). The same job was **green at 19:33 UTC** (run 529, 71 s), so the regression window is roughly **19:33 → 20:55 UTC on 2026-08-17**.
* **From a laptop just now**, `E2E_BASE_URL=https://memex.meshweaver.cloud npx playwright test smoke.spec.ts` — same failure, screenshot above.

The test waits 30 s for *either* the hero *or* a "Sign in" heading and gets neither, so this is not an auth wall — the page renders the fallback card.

🚨 **`HTTP 200` does not reveal this.** `curl https://memex.meshweaver.cloud/AgenticEngineering` returns 200 with ~98 KB, because the Blazor catch-all answers 200 for everything and the real content arrives over SignalR. Any health check that greps the status code will call this healthy.

## Likely relation to tonight's incident

The window matches the silo rollback/roll-forward (#1800 / #1801, `AdoNetGrainStorage.Init` → `Sequence contains no matching element`). A framework-version change recompiles **every** dynamic NodeType, so a portal that came up after a roll-forward has the whole backlog to settle at once — and `Store/Plugin` is the root type of every course, Edu and Publish root, so it failing to settle takes out every cover simultaneously.

[...]

Issue #1440, opened 2026-08-13
FollowGo waits unboundedly for a GO that cannot cross processes — the PG change listener is registered but never started


rbuergi:
Found while implementing the cross-cluster build claim (#1424 / PR #1437). The doc was corrected there; the underlying defect was not fixed and needs its own home.

## The claim the doc made

`BuildCoordination.md` stated that the GO broadcast rides the Postgres `LISTEN/NOTIFY` feed and is therefore *"cross-process by construction"*.

**It is not.** `PostgreSqlChangeListener` is **registered but never started** in either partitioned-PG overload — `AddPartitionedPostgreSqlPersistence` has its `AddHostedService` commented out, and the Aspire overload never had one. So there is **no cross-process change feed at all**: a mirror can be arbitrarily far behind a rival's write and will never be told.

## The consequence

`BuildProtocolDriver.FollowGo` subscribes to `ObserveBuildGo` **with no bound**. A follower in a *different* cluster is therefore waiting for a notification that cannot reach it, with nothing to end the wait.

So the current honest description of a separate-cluster bake host is **"publishes the GO earlier"**, not *"spares its peers the bake"* — the peers never learn it happened.

Note this is not merely a slow path: an unbounded wait for an event that is structurally undeliverable never completes. That is the same class as several defects fixed this week (a request handled but never answered; a subscription whose terminal signal was discarded), where the caller has no way to distinguish "still working" from "will never happen".

## Two things to decide, and they are different

1. **Should the change listener be started?** If cross-process notification is intended, the commented-out `AddHostedService` is the gap — but starting it is a real behavioural change to every partitioned-PG deployment and wants its own justification and measurement.
2. **`FollowGo` must not wait unboundedly regardless.** Even with the feed running, an unbounded wait on a remote event is the wrong shape. The arbiter that #1437 added deliberately **reads** the witness rather than waiting to be notified, precisely because the feed cannot be relied upon — `FollowGo` should do the same, or carry a bound with an explicit outcome.

🚨 Do **not** fix this by adding a poll timer as the primary mechanism. #1366 removed a 10-minute staleness clock for exactly that reason, and #1437's arbiter is level-triggered by design: a refused claim writes nothing and re-decides on the next real change.

## Context worth carrying

[...]

Issue #1790, opened 2026-08-17
ActivityLogLogger.SealOverflowLocked latches system-security on the script thread (same shape as #1784's review finding)


rbuergi:
Split out of #1785 so it gets its own repro rather than riding along on an unproven edit.

## The shape

`ActivityLogLogger.SealOverflowLocked` writes the overflow segment like this:

```csharp
Observable.Using<MeshNode, IDisposable>(
        () => _accessService?.ImpersonateAsSystem() ?? Disposable.Empty,
        _ => meshService.CreateOrUpdateNode(segmentNode))
    .Subscribe(...);
```

Impersonation is an `AsyncLocal` scope (`AccessService.AccessContextScope`: it stores the previous
value and restores it on `Dispose`). `Observable.Using` **opens** it on the subscribing thread and
**disposes** it when the inner observable terminates — a storage/IO thread, not the subscriber's.
So two things go wrong at once:

- the subscribing thread keeps `system-security` latched after `Subscribe` returns, and
- the terminating thread has the subscriber's "previous" value written into *its* context.

## Why it matters here specifically

`SealOverflowLocked` is called from `PublishSnapshotLocked`, and the **first** publish of a script
run is issued from the script's own thread (`Console.WriteLine` → `LoggerTextWriter` →
`PublishSnapshotLocked`, inside `KernelExecutor.RunOnePass`). A latch there means the rest of the
script runs as System.

That is not hypothetical for this code path — it is exactly what happened when #1785 briefly moved
the *publish* write to `Observable.Using`:
`DocumentExportAreaAccessTest.EmbeddedArea_IsRenderedForAReaderWhoMayReadIt_AndWithheldFromOneWhoMayNot`
went red, because a `--render` export then resolved embedded areas the submitting user may not
read. Reverting to a synchronous `using` fixed it. Measured, one test, four builds of
`MeshWeaver.Kernel.Hub`:

| `Kernel.Hub` | result |
|---|---|
| `origin/main` | PASS |
| `Observable.Using` on the publish | **FAIL** |
| `RunAsSystem` on the publish | **FAIL** |
| synchronous `using` on the publish | PASS |

The seal path has the same shape on the same thread; the only reason it is not covered by that
test is that the seal fires only once a run's output exceeds `ActivityLog.MessageWindowLimit`,
which no existing test crosses on the export path.

🚨 `RunAsSystem` is **not** the fix. `ContainIdentity` restores the caller's identity around
[...]

PR #1870, opened 2026-08-18
fix(#1814): the compilation fault card re-evaluates instead of latching

rbuergi:
Closes **defect B of #1814** — the latching fault card, and the reason a ten-minute deploy window became a two-hour public outage.

## The defect, precisely

Enrichment binds a per-instance hub's `HubConfiguration` **exactly once** — the `node.HubConfiguration != null` re-enrichment short-circuit is what keeps activation cheap — so a compilation overlay applied during a bad ten seconds is served for the grain's whole lifetime unless something *revokes* it. `ArmOverlaySelfHeal` is that revocation.

Both of its heal routes subscribe `meshHub.GetWorkspace().GetMeshNodeStream(nodeType)` — **the same stream whose silence made the enrichment slow path time out in the first place**:

| route | trigger |
|---|---|
| version advance | a usable emission past the at-overlay version |
| grace (45 s) | re-subscribe and take the first usable emission |

The grace route's "re-read" is a re-subscribe to the workspace's **cached** stream, so it replays the same snapshot. Neither route reads anything. **The heal signal and the fault shared a single point of failure**: one permanently silent stream and the card is permanent.

That is exactly what was measured on 2026-08-17 (#1814): `Store/Plugin` compiled **successfully on both pods**, and **1 h 24 m later** an anonymous browser still got the card, with **no overlay and no "did not settle" event logged on either pod in the preceding 30–40 minutes**. The 120 s stuck-reporter fired (`platform admins notified`) — proving the watcher was armed and had simply never seen a usable state. Twelve roots were recycled by hand.

## The fix — a third route that owes the stream nothing

**`AuthoritativeTypeRead`** — a one-shot `path:{nodeType}` query through `IMeshQueryCore`, as System: the mesh's **query providers (storage)**, not a cached stream, so a mirror that can never learn it is stale cannot suppress it. It matches the path **exactly**; a ranked/fuzzy hit for a neighbouring node is not an answer about this type.

**A widening ladder, not a poll.** `45 s → 90 s → 3 min → 6 min`, then **10 min for ever**. One read per rung, each capped at 30 s and serialised with `Concat` so a slow read can never overlap the next rung. At the 2026-08-17 blast radius (12 latched instances) the steady state is ~72 single-node reads an hour — against the 1 h 24 m of served fault cards it replaces. The naive shape (re-probe per render) is explicitly not what this is.

**The ladder never stops.** A re-evaluation budget that ran out would restore precisely the defect: a card outliving its cause because something decided to stop looking.

**A faulted or empty read is not a verdict** — it is logged and the ladder asks again. **Non-convergence is loud**: each re-read that still finds no usable build logs the status / assembly / framework it actually saw, at `Warning` once past the last rung, beside the existing admin notification.

### Bounding the recycle, which the watcher cannot do itself

The heal disposes the hub that **owns the watcher**. The replacement hub arms a fresh watcher whose ladder starts at the first rung — so a pair whose *re-enrichment* keeps faulting (a type reporting a usable build that the instance still cannot bind: #1814's deterministic cross-hub `Conflict`) would recycle every 45 s for ever. No state inside the watcher can bound that, because the bound has to outlive the thing being bounded.

`OverlayHealBudget` — mesh-scoped singleton registered in `AddGraph`, keyed by *(instance, NodeType)*, instance maps only — is that memory. The **first** heal is never delayed (the common case is a deploy window that cleared). Each further heal inside 30 minutes waits out a widening spacing (45 s → 90 s → 3 min → 6 min → 10 min). It **defers** a recycle, **never cancels** one; a pair that heals once and stays healthy is forgotten.

### What deliberately did NOT change

A **genuinely broken type still shows its card**. The goal is "clears when the cause clears", never "never shows". `StillBrokenType_KeepsItsCard_AndTheLadderStaysBounded` pins both halves.

The overlay copy's promise — *"the page recovers automatically"* — was false for 1 h 24 m and is now true; it also says **how** it recovers, so an operator knows how long to wait before reaching for **Recycle**.

## Non-vacuity — proven by neutering, with real output

The new route removed (`_ = reEvaluated;` — i.e. `origin/main`'s push-only self-heal), same machine, same test binary:

```
  Failed …SilentTypeStream_ReEvaluatesFromStorage_AndHealsWithoutIntervention
   NSubstitute.Exceptions.ReceivedCallsException : Expected to receive exactly 1 call matching:
  Failed …StillBrokenType_KeepsItsCard_AndTheLadderStaysBounded
   Expected 0 to be greater than 3 because the ladder keeps looking — giving up is how the card latched.
  Failed …FaultedReRead_DoesNotEndTheLadder
   Expected value to be 1, but found 0.
  Failed …RepeatedlyStuckInstance_HasItsNextRecycleDeferred_NeverCancelled
   NSubstitute.Exceptions.ReceivedCallsException : Expected to receive exactly 1 call matching:
  Failed …NonConvergingInstance_RecyclesAreSpaced_NotOncePerLadderRung
   Expected 0 to be greater than 40 because the control: with no cross-recycle memory every fresh
   watcher restarts at the first rung, which is the storm the budget exists to prevent.
Total tests: 12   Passed: 7   Failed: 5
```

The first line **is** the 2026-08-17 behaviour: the instance never recycled itself. The 7 pre-existing tests stay green under the neuter, so they cover the old behaviour and the 5 new ones are non-vacuous.

With the fix restored:

```
Total tests: 12   Passed: 12       (OverlaySelfHealWatcherTest)
Passed!  - Failed: 0, Passed: 1336 (MeshWeaver.Graph.Test, 33 s)
Passed!  - Failed: 0, Passed:    4 (CompileErrorOverviewTest — overlay copy)
Passed!  - Failed: 0, Passed:    1 (OverlaySelfHealInstanceRecycleTest — the e2e self-recycle)
Passed!  - Failed: 0, Passed:    1 (WhatsNewEntryIntegrityTest)
```

All time is driven by `TestScheduler` — this is a latching bug, so every wait is bounded and virtual; nothing sleeps.

## Tests

| test | pins |
|---|---|
| `SilentTypeStream_ReEvaluatesFromStorage_AndHealsWithoutIntervention` | a permanently silent stream + a settled record ⇒ the instance recycles **itself**, inside the first rung, no intervention |
| `StillBrokenType_KeepsItsCard_AndTheLadderStaysBounded` | a never-usable record ⇒ no recycle, ever; and single-digit reads over a virtual hour |
| `FaultedReRead_DoesNotEndTheLadder` / `EmptyReRead_IsNotAHealSignal` | a non-answer is neither "broken" nor "healed" |
| `RepeatedlyStuckInstance_HasItsNextRecycleDeferred_NeverCancelled` | the budget **defers**, and the recycle still lands |
| `NonConvergingInstance_RecyclesAreSpaced_NotOncePerLadderRung` | drives the real recycle→re-overlay→recycle loop for a virtual hour; bounded, with the un-budgeted control in the same test |
| `OverlayHealBudgetTest` | first heal un-delayed, widening spacing holding at the ceiling, forget window, per-pair keying |
| `OverlayReEvaluationReadTest` | **the seam itself** — query core, System identity, exact path. Every watcher-level assertion would pass against a real read wired to something that can never answer; this is the one test that can't. |

## What remains of #1814

- **Defect A — bake/portal identity mismatch: DONE** (#1847, merged).
- **Defect B — the latching card: this PR.**
- **The `Conflict` retry** (#1814's item 3) is *partly* addressed on main already: `UpdateRemote` now includes `MeshNodeErrorCode.Conflict` in the re-enqueue set, and the recursion genuinely re-runs the lambda against a re-read initial state. But it re-reads through `AcquireRemoteStreamUnchecked` — **the same workspace mirror** — so with the notify chain broken the base is stale on every attempt and the two re-enqueues cannot converge. Nothing further to change here; it is subsumed by the item below.
- **The cross-process notify chain is still open**, all three legs: `mesh_node_notify` installed per database rather than per table (**#1816**, open), `PostgreSqlChangeListener` registered but commented out for the partitioned setup (**#1440**, open — the TODO is still in `PostgreSqlExtensions.cs`), and notifications publishing `Entity = null`, which `MeshDataSource` discards.

So **#1814 should stay open** until the notify chain lands. What this PR changes is the severity: with the card re-evaluating, a stale mirror costs a page that recovers on its own within a minute (or ten, if it keeps failing) instead of an open-ended outage that only twelve manual recycles could end.

### DIFF
--- a/src/MeshWeaver.Documentation/Data/Architecture/NodeTypeCompilation.md
+++ b/src/MeshWeaver.Documentation/Data/Architecture/NodeTypeCompilation.md
@@ -296,6 +296,66 @@ keeps running however long it needs (see the stage bounds above).
 
 ---
 
+## A fault card must not outlive its cause — the overlay RE-EVALUATES
+
+The overlay is a **degraded binding with a lifetime**, never a verdict. Enrichment binds a
+per-instance hub's configuration exactly once — the re-enrichment short-circuit
+(`node.HubConfiguration != null`) is what makes activation cheap — so a card applied during a
+bad ten seconds is served for the grain's whole lifetime unless something *revokes* it. That
+revocation is `ArmOverlaySelfHeal`, and its correctness rests on one rule:
+
+> 🚨 **The heal signal must not share a failure mode with the fault.** Both original heal routes
+> — the version-advance and the grace re-read — subscribe
+> `meshHub.GetWorkspace().GetMeshNodeStream(nodeType)`: *the same stream whose silence made the
+> enrichment slow path time out*. When that stream stops emitting, the fault and its cure vanish
+> together and the card is permanent.
+
+That is not hypothetical. On **2026-08-17** (issue #1814) a deploy left the first pod compiling
+269 types; page requests inside that window latched the card onto ~12 plugin roots
+(`Training`, `Video`, `RolePlay`, `Edu`, `Chess`, `Collaboration`, …). `Store/Plugin` then
+compiled **successfully on both pods** — and **1 h 24 m later** an anonymous browser still got
+the card, while neither pod had logged a single overlay or "did not settle" event in the
+preceding 30–40 minutes. Nothing was retrying. Recycling the twelve roots by hand was the only
+remedy, and the card's own copy ("the page recovers automatically … this instance recycles
+itself") was straightforwardly false.
+
+So the watcher has a third route that owes the stream nothing:
+
+- **`AuthoritativeTypeRead`** — a one-shot `path:{nodeType}` query through `IMeshQueryCore`, as
+  System. It reads the mesh's **query providers (storage)**, not a cached stream, so a mirror
+  that can never learn it is stale cannot suppress it.
+- **A widening ladder, not a poll** — `45 s → 90 s → 3 min → 6 min`, then **10 min for ever**
+  (`ReEvaluationLadder` / `ReEvaluationCeiling`). One read per rung, each capped by
+  `ReEvaluationReadTimeout` and serialised with `Concat`, so a slow read can never overlap the
+  next rung. At the 2026-08-17 blast radius that steady state is ~72 single-node reads an hour
+  across the whole mesh. **The ladder never stops** — a re-evaluation budget that ran out would
+  re-create precisely the defect it exists to remove.
+- **A faulted or empty read is not a verdict** — it is logged and the ladder asks again. Giving
+  up quietly is how the card latched in the first place.
+- **Loud on non-convergence** — a re-read that still finds no usable build logs the type's
+  status/assembly/framework; past the last rung it does so at `Warning`, next to the existing
+  admin notification at `StuckReportDelay`.
+
+### …and the recycle it orders is SPACED, because it destroys its own watcher
+
+The heal disposes the instance hub — taking the watcher with it. The replacement hub arms a
+**fresh** watcher whose ladder starts at the first rung, so a pair whose *re-enrichment* keeps
+faulting (a type that reports a usable build the instance still cannot bind — #1814's
+deterministic cross-hub `Conflict`) would recycle every 45 s for ever. No state inside the
+watcher can bound that, because the bound has to outlive the thing being bounded.
+
+`OverlayHealBudget` (mesh-scoped singleton, registered in `AddGraph`) is that memory, keyed by
+*(instance, NodeType)*. The **first** heal is never delayed; each further heal inside a 30-minute
+window waits out a widening spacing (45 s → 90 s → 3 min → 6 min → 10 min). It **defers** a
+recycle, never cancels one, and a pair that heals once and stays healthy is forgotten.
+
+Pinned by `OverlaySelfHealWatcherTest` (silent stream → heals unaided; still-broken type → keeps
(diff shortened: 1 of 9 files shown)


rbuergi:
Follow-up commit `81fc01c` — CI's `ImpersonationScopeSiteRatchetGuard` caught a real defect in the first push, and it is worth recording because it is the exact failure mode the guard exists for.

The authoritative re-read copied the enrichment probe's idiom, `Observable.Using(() => access.ImpersonateAsSystem(), _ => query)`. That shape opens the AsyncLocal scope on the **subscribing** thread and disposes it when the query terminates — the owning hub's response thread — so the subscriber stays latched as `system-security` (#1790). Here the subscriber is an **instance hub's long-lived overlay watcher**, which is about the worst place to leave a System latch.

Replaced with `ImpersonationScopeExtensions.RunAsSystem`, which seals both ends inside one `Subscribe`, with the whole cold pipeline composed inside the work factory so emission-time behaviour is unchanged. The ratchet's inventory does not grow — `NodeTypeEnrichmentHelpers.cs` stays at its allowed 1 (the pre-existing probe).

CI is now green end to end: all 6 shards, `Consolidate test results`, `Build solution`, `Doc content compiles and runs`, `Plugins compile against this PR`. Not merging — leaving that to the maintainer.

Issue #1752, opened 2026-08-17
Move Cosmos + Snowflake (source, tests, releases) out of the platform: ship released binaries, run their tests only when they change


rbuergi:
Maintainer-directed:

> *"can we move cosmos and snowflake to plugins?"* … *"or just create the releases and the binaries?"* … *"and move the tests over"* … *"then we don't run tests"* … *"and tests are run only when we touch the plugins"* … *"or update the platform"*

## Why this works (and why an earlier objection of mine does not)

Both backends are ALREADY free of compiled coupling. Each is a boot-pack module whose
`MeshNodeProviderAttribute` registers a keyed `IStorageAdapterFactory`; the Cosmos one states it
outright — *"everything a deployment needs for `Graph:Storage:Type = Cosmos` to resolve, **with no
compiled reference from the portal**"*. Nothing in `src/` or `memex/` references either project;
only their own test projects do.

I had objected that a lagging build would be declined by the identity gate. **That is wrong, and
the distinction matters:** there are two gates, not one.
- **Content bundles** (baked NodeType assemblies) are declined on STRICT framework identity —
  `PrebuiltAssemblySeeder.DeclineReason(FrameworkMvid)` in `PluginBundleClient`.
- **Compiled modules** gate on a **`MinMeshVersion` FLOOR** (`PluginCatalogContent.MinMeshVersion`,
  `PluginManifest.MinMeshVersion`), with MVID diagnostic-only (#1673).

So a module binary built against platform *N−1* loads into *N* by design, provided the floor is
satisfied. The residual risk is ordinary .NET binary compatibility, not an identity decline.

## The shape

1. Source + tests move to a plugin repo.
2. That repo builds and publishes **versioned releases**: the binaries, with a `MinMeshVersion`
   floor. (Releases can be modelled as `Release` nodes carrying the assembly link — see Systemorph/MeshWeaver#1751.)
3. The platform image consumes a **pinned released artifact**. No build-time source coupling, and
   no cross-repo build cycle: the platform does not trigger or wait for the plugin build.
4. **Platform CI stops running the Cosmos/Snowflake suites.** They run in the plugin repo, when
   that code changes. This is the actual prize — `MeshWeaver.Hosting.Cosmos.Test` needs the Cosmos
   emulator via Testcontainers, which is expensive to stand up on every platform commit for code
   that changes rarely.
5. If the platform changes the storage contract, the plugin is updated (or the platform re-releases
[...]

PR #1892, opened 2026-08-19
Retire NuGet packages the platform no longer builds

rbuergi:
## Why

A module that leaves the platform stops being packed — but everything it already published stays on nuget.org, **listed**. It keeps appearing in search and in `dotnet add package`, offered to consumers as current while nothing builds or patches it any more.

Nine are live today: `MeshWeaver.AI.AzureOpenAI`, `Approvals`, `Blazor.AI`, `BusinessRules`, `BusinessRules.Generator`, `Education`, `GoogleMaps`, `GridModel`, `Hosting.AzureStorage`. **#1882 is about to add eight more** (ClaudeCode, Copilot, Indexing.PostgreSql, Courses, Markdown.Export, Mcp, Notifications.Channels, Observability), and #1752 will add Snowflake.

## Derived, not listed

The set is `(published under the MeshWeaver prefix) MINUS (what this tree still packs)`. No hand-kept list, so #1882 and #1752 need no edit here — the day after they merge, their packages simply appear in the report.

## 🚨 Report by default

The derivation is a **subtraction**, so any bug in the "what do we pack" half fails toward unlisting live packages. Two such bugs appeared while writing it, both silent:

- the exclusion tested the **absolute** path, so a checkout under `.worktrees/` excluded every project — the tree looked like it packed nothing and the orphan set came out as **all 90 packages we have ever published**;
- `MeshWeaver.Compiler.Cli` is `IsPackable=false` by default and packs only under `-p:PackCompilerTool=true`, with its id declared in that same conditional group — reading `IsPackable` first hid the project, and a package we very much still ship was reported orphaned.

Both are pinned by `--self-test`, which the workflow runs **before** the apply step can be reached. Reverting either fix reproduces its own failure.

Applying additionally requires `apply: true` **and** `confirm: UNLIST`.

## Reversible

`dotnet nuget delete` **unlists**; it does not erase. Existing pins keep resolving by exact version, and a listing can be restored from the nuget.org UI.

## Verified

```
published on nuget.org : 90
packable in this tree  : 168
ORPHANED               : 9
```

`--self-test` passes; with either fix reverted it reports that fix's failure.

### DIFF
--- a/.github/workflows/unlist-orphaned-packages.yml
+++ b/.github/workflows/unlist-orphaned-packages.yml
@@ -1,0 +1,93 @@
+name: Unlist orphaned NuGet packages
+
+# A module that LEAVES the platform stops being packed — but everything it already published
+# stays on nuget.org, LISTED: it keeps appearing in search and in `dotnet add package`, offered
+# to consumers as current while nothing builds or patches it any more. This is how that set is
+# retired.
+#
+# 🚨 The set is DERIVED, never hand-kept: (published under the MeshWeaver prefix) MINUS (what
+# this tree still packs). So it stays correct on its own as further modules move out — #1882
+# (the fourteen drivers) and #1752 (Cosmos/Snowflake) need no edit here.
+#
+# 🚨 REPORT BY DEFAULT. `apply` must be set explicitly AND `confirm` must read UNLIST. The
+# derivation is a subtraction, so any bug in "what this tree packs" fails toward unlisting live
+# packages — during development one such bug computed "all 90". Report-only is what caught it.
+#
+# `dotnet nuget delete` UNLISTS on nuget.org; it does not erase. Existing pins keep resolving by
+# exact version, and a listing can be restored from the nuget.org UI. That reversibility is what
+# makes this safe to automate.
+
+on:
+  workflow_dispatch:
+    inputs:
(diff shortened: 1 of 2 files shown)


rbuergi:
Verified the one way this scan could have been wrong: a package that left core might still be actively published from a satellite repo, in which case unlisting it would be destroying a live package.

It isn't. `MeshWeaver.Plugins`' `publish-packages.yml` calls the shared `plugin-publish.yml` with **`dry-run: true`**, and it targets `nuget.pkg.github.com`, not nuget.org — the header says so explicitly ("the pack is the point"; a push "can only fail, never help" while the org is at its GitHub Packages billing limit). The other satellites publish no packages at all.

So **nuget.org is fed solely by core's tag-triggered `release-packages.yml`**, and scanning the core tree is the correct and complete scope. The nine orphans are genuinely discontinued there rather than relocated:

- `BusinessRules` / `BusinessRules.Generator` → node-native plugins in MeshWeaver.Plugins (which never pushes to nuget.org)
- `Approvals` → node-native
- `Education` → the education repo
- `AI.AzureOpenAI`, `Blazor.AI`, `GoogleMaps`, `GridModel`, `Hosting.AzureStorage` → superseded or removed

PR #1895, opened 2026-08-19
CD must queue, not mutually defer — and must never be killed mid-set

rbuergi:
## What happened

Two CD runs for the same commit both reported **success** and **neither built an image**.

Observed today on `cb33558`: a `workflow_run` CD and a manual dispatch raced, and each skipped because it saw *"a CD run for this sha is already in flight"*. Both runs are green, both summaries show a tidy ⏳ line, and the only symptom is that `main` sits with no image set while every self-updating install stays on the previous image.

That is the same family as the other CD traps in this repo: **the evidence that nothing was built is the absence of evidence.**

## Two causes, both fixed

- **The concurrency group was `main-cd-${{ github.run_id }}` — unique per run.** It grouped each run with itself and serialized nothing, so two runs could overlap in the first place. Keyed on the ref, a second run now **queues** behind the first, then finds a complete image set and exits with *"nothing to reconcile"* — the honest no-op.
- **The in-flight guard is removed, not repaired.** Once runs cannot overlap it is redundant — and worse than redundant: a run **queued behind** this one also counts as "in flight", so keeping it would re-arm the same defect from the other side.

`cancel-in-progress: false` stays deliberately: a run midway through pushing a multi-image set must never be killed, or it leaves exactly the incomplete set the reconciler exists to repair.

## Verified

`main-cd.yml` parses; the guard is gone (`inflight=` occurrences: 0) and the group is `main-cd-${{ github.ref }}`.

The immediate incident was worked around by hand — dispatching once nothing else was in flight, which built all four images — but the race stays armed until this merges.

### DIFF
--- a/.github/workflows/main-cd.yml
+++ b/.github/workflows/main-cd.yml
@@ -240,6 +240,10 @@ jobs:
       short_sha: ${{ steps.target.outputs.short_sha }}
       staging_tag: ${{ steps.target.outputs.staging_tag }}
       reason: ${{ steps.target.outputs.reason }}
+      # The three facts the delivery verdict below judges. Exposed as outputs so the check is a
+      # JOB anyone can read in the run summary, not a conclusion buried in this job's log.
+      complete: ${{ steps.acr.outputs.complete }}
+      green: ${{ steps.required.outputs.green }}
     steps:
       - uses: actions/checkout@v7
       - name: "Resolve the target commit"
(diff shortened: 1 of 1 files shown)


rbuergi:
Added the gate for the CLASS, not just this cause.

`alert-on-failure` only fires on `failure()` — and a CD run in which every image job **skipped** is a `success()`. That is exactly why the incident was invisible: two green runs, no image, nothing said so. Delivery had in fact been stuck long enough that CD reported `Newest published set: <none>`, which is also why memex has been sitting on an old commit while main moved — there was no complete set to roll to.

**`delivery-verdict`** fails when the *reconcile* path ends with `publish != true` while main is **green** and its image set is **incomplete** — the one combination that means delivery is stuck and no one is being told. The gate's inputs (`complete`, `green`) are promoted to job outputs so the verdict is a job you can read in the run summary rather than a conclusion buried in a log.

It deliberately does **not** fire on the push path: batching legitimately hands the commit to the reconciler within the hour, and a gate that cries wolf on normal behaviour is how a real red gets ignored.

Same family as the node repos' skip-detection invariant, for the same reason — a skipped job and a passed job render identically, so the absence of evidence has to be turned into evidence.

PR #1913, opened 2026-08-20
MCP: an agent can file a GitHub issue (github_issue tool)

rbuergi:
## The gap

Carson reports defects by leaving a **mesh thread**, because that is the only thing it can reach. Nothing routes a thread to a backlog, so a real finding lands somewhere nobody triages.

`IssueService` already had create / comment / sync, wired to the browser's GitHub tab. MCP exposed `git_hub_sync` and nothing else. **This is a transport gap, not a feature** — one tool over a service that already existed.

## The tool

`github_issue(space, op, title, body, number, labels, state)`

| op | what |
|---|---|
| `create` (default) | open an issue — `title`, markdown `body`, comma-separated `labels` |
| `comment` | comment on issue `number` |
| `sync` | mirror the repo's issues to `{space}/_Issue/{number}` so an agent can **search for a duplicate before filing one**, using the ordinary `search`/`get` tools |

The tool description tells the caller to `sync` and check before filing — a duplicate issue costs a human exactly the triage time this was meant to save.

## 🚨 The caller's own GitHub token is the only authority

Like every other verb on `IssueService`, this runs through the caller's stored OAuth credential (`{userId}/_Provider/GitHub`). It reaches exactly the repositories that account reaches, and no mesh grant widens it — a machine account holding Triage on one repo can file there and nowhere else.

There is deliberately **no system-identity path**. An issue filed by "the platform" names nobody you could ask about it, and such a fallback would turn *file an issue* into *write to every repository the platform can see*. `AccessContext` is re-established at the call (`SwitchAccessContext`) so `IssueService` reads the **right** user's token — the same `AsyncLocal` hazard `AsCaller` and `git_hub_sync` already handle.

## Both guards are mutation-tested, not assumed

| mutation applied | test that went red |
|---|---|
| drop the label parse | `GitHubIssue_Create_AppliesLabels` |
| fall back to `system-security` when no caller resolves | `GitHubIssue_NoIdentity_ReturnsSignInError` |

Verified by making each change and watching the named test fail, then reverting.

## Tests

11 new, driving the real `McpMeshPlugin` against the in-memory `FakeGitHubRepoClient` (whole loop offline): argument validation, the no-connected-account refusal, create → GitHub + mirrored node, labels, comment, and sync. **192/192 in `MeshWeaver.GitSync.Test`.**

Also adds `FakeGitHubRepoClient.IssuesOn(repo)` — the fake could seed and serve issues but had no accessor to assert on them, mirroring the existing `Tree(repo)`.

### DIFF
--- a/src/MeshWeaver.Mcp/McpMeshPlugin.cs
+++ b/src/MeshWeaver.Mcp/McpMeshPlugin.cs
@@ -587,6 +587,185 @@ When the sync source has two-way enabled, `update` never overwrites a node chang
     }
 
     /// <summary>
+    /// Files, comments on, and mirrors GitHub <b>issues</b> for a Space — the MCP reach of
+    /// <see cref="IssueService"/>, which the browser already exposed and MCP did not.
+    ///
+    /// <para>🚨 <b>Why this exists.</b> An agent working through MCP could read and write the mesh
+    /// but had no way to report a defect anywhere a human triages. It could only leave a mesh
+    /// thread, which nothing routes. The capability was fully built (<c>IssueService</c>: create,
+    /// comment, sync) and simply unreachable from this transport.</para>
+    ///
+    /// <para>🚨 <b>The caller's OWN GitHub token is the only authority.</b> Like every other verb on
+    /// <see cref="IssueService"/>, this runs through the caller's stored OAuth credential — so it can
+    /// reach exactly the repositories that account can reach, and nothing the mesh grants widens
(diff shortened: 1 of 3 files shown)


rbuergi:
Superseded: `MeshWeaver.Mcp` left the platform in 0b3b4266f ("Fourteen modules leave the platform") while this was in flight, so the tool belongs in MeshWeaver.Plugins now. Re-opening it there — and without the core-side `FakeGitHubRepoClient.IssuesOn` helper, which would have forced a pin bump for something the fake's public `ListIssues` already answers.

PR #1914, opened 2026-08-20
Module bundles carry their own dependency closure, derived from deps.json

rbuergi:
## The outage

When a module's source left the platform repo (#1882), its package dependencies left `/app` with it — and `module-pack` bundled the entry DLL alone (`--with` exists but nothing ever passed it). Every landed module faults at first use on its first private dependency:

- chat: `Could not load … Microsoft.Extensions.AI.OpenAI, Version=10.8.0.0`
- `3.0.0-rc5.ci.4588` crashloops at host start: `Could not load … Microsoft.Graph, Version=6.2.0.0` from `GraphMail..ctor`

Pinning each dependency into the platform (#1912) is the wrong direction — it ships 43 MB of Graph/Kiota to deployments that send no mail, the exact cost the module split exists to avoid. Fix forward: **the bundle carries the module's own closure, derived, not remembered.**

## The rule

`DepsClosure.Derive` walks the module's own `deps.json` twice from its direct references:

| reachable from | verdict |
|---|---|
| the module's **MeshWeaver.\*** references | platform-carried — in the consumer's `/app` by construction; bundling it would shadow the platform (the same trap-door `ModuleLandingService` refuses for the entry DLL) |
| ONLY the module's **own package** references | the private closure — bundled |

Diamonds (reachable from both) are platform-carried. Shared-framework assemblies never appear as package nodes, so they're excluded by construction.

## The changes

- `DepsClosure` — pure derivation, unit-tested on a synthetic graph shaped like the real one (diamond, native-assets warning, wrong-name refusal, valid-empty).
- `module-pack --deps-closure` — derived files must exist beside the entry DLL (build with `CopyLocalLockFileAssemblies=true`) or the pack **refuses**: a partial closure lands a module that faults at first use.
- `node-repo-module-pack.yml` — builds with the flag, packs with `--deps-closure`, and the inspect step asserts **no MeshWeaver.\* assembly** rides beside the entry.

## Verified against all 14 extracted modules

`Mail.MicrosoftGraph` → the ten Graph/Kiota assemblies + Std.UriTemplate (exactly what its csproj names as the module's reason to exist); `Copilot` → GitHub.Copilot.SDK; `Mcp` → ModelContextProtocol(+AspNetCore); `AzureFoundry` → 4; `ClaudeCode`/`Export` → 1; platform-only modules correctly derive empty. `Azure.Identity`/`Azure.Core` excluded wherever a MeshWeaver.* reference already reaches them.

Follow-up (separate PR): revert the #1912 platform-side pin so Microsoft.Extensions.AI.OpenAI leaves `/app` and rides in the OpenAI module's bundle instead — the derivation flips it to bundled automatically.

### DIFF
--- a/.github/workflows/node-repo-module-pack.yml
+++ b/.github/workflows/node-repo-module-pack.yml
@@ -77,9 +77,14 @@ jobs:
           echo "packing $MODULE as $PACKAGE@$version, platform floor $floor"
 
       - name: Build the module (Release, warnings-as-errors)
+        # CopyLocalLockFileAssemblies puts the module's PACKAGE dependencies into the output
+        # folder, where the pack step's --deps-closure derivation expects to find them — a module
+        # bundle must carry its own private closure (the platform does not: those packages left
+        # /app with the module's source, and an entry-DLL-only bundle faults at first use).
         run: >
           dotnet build "${{ matrix.entry.project }}" -c Release -warnaserror
           -p:MeshWeaverRoot=$GITHUB_WORKSPACE/meshweaver
+          -p:CopyLocalLockFileAssemblies=true
           -p:Version=${{ steps.identity.outputs.version }}
 
       - name: Run the module's tests, when it ships any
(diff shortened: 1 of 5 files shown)


rbuergi:
Amended: **diamonds now ride**. The closure walks the module's own package refs and stops only at MeshWeaver.* nodes — a package the platform also carries ships anyway (`/app` wins in the default load context while present; the module's copy takes over when the platform sheds it). This is what makes the follow-up #1912 revert — and any future platform slim-down — safe without coordinating bundle re-lands.

Measured: `SelfUpdate.Aks` now bundles Azure.Core/Azure.Identity + the MSAL chain (the auto-roll's broken dependency set); `OpenAI` bundles its 33 AI-SDK files; `Mail` 28; platform-only modules still derive 0.

PR #1915, opened 2026-08-20
The time-zone preference read English to every German viewer

sierragolflima:
## The defect

Settings → Preferences carries two pickers. The language one has gone through `Localize("settings.language")` since the day it was written; the display **time-zone** label was the literal `"Display time zone (IANA)"`, and the paragraph explaining the whole feature was an English literal too.

So the tab rendered half-translated for a German viewer: correct heading, correct language label, English beside it. Per AGENTS.md a hard-coded user-visible string is a bug — it renders English to every viewer regardless of their language.

## Why no existing gate caught it

- **`LocalizationTest`** only asserts that each shipped language covers the English *key* list. A string that never became a key is not a *missing* key — it is an absent one, so the catalog looks complete.
- **A render test** would pass as well, under an English viewer.

That is the part worth fixing structurally rather than just editing two strings.

## The change

The pickers now come from `UserNodeType.PreferenceFields(Func<string,string> localize)` and the prose from `PreferencesDescriptionKeys`. The localizer is a **parameter** for the same reason `PlatformUpdateChip.Describe`'s is: it makes the wording assertable without a hub, a circuit or a rendered layout area.

Two keys added to **all four** catalogs — `strings.{en,de}.json` server-side, mirrored into `clients/react/src/i18n`, where the drift guard compares values per key rather than just counts.

## Tests

Written test-first. `UserPreferencesLocalizationTest` fails on the literal — **3 of its 4 assertions**, naming the exact string — and covers both halves of the bug:

1. every label goes through the localizer (echo localizer ⇒ labels *are* keys, so a literal stands out as prose among dotted keys);
2. the real German catalog has text behind those keys — which a key added without a translation would miss.

Verified locally: `-c Release -warnaserror` clean · `MeshWeaver.Graph.Test` 1372/1372 · `MeshWeaver.Messaging.Hub.Test` 56/56 · `MeshWeaver.Documentation.Test` 39/39 · `clients/react` localize 15/15.

## Notes for review

- The React catalogs are edited by **line insertion**, not re-serialized: a full re-dump churns ~200 unrelated lines in whichever direction its escaping convention differs (the file is dominantly literal UTF-8 with a handful of pre-escaped `\uXXXX` entries).
- Noted but deliberately **not** touched: `BuildPreferencesTab`'s `var locale = access.ViewerLocale()` was already unused before this change.
- Found while investigating why a portal showed UTC timestamps. Separate and **not** addressed here: an account created before the browser auto-detect (`f31b81f0a`) has no `User.TimeZoneId`, and the write-once populate only fills an *empty* field on first sign-in — it never backfills — so such a user reads every timestamp in UTC with nothing on screen saying so.

### DIFF
--- a/clients/react/src/i18n/strings.de.json
+++ b/clients/react/src/i18n/strings.de.json
@@ -256,6 +256,8 @@
   "settings.groupIntegration": "Integration",
   "settings.language": "Sprache",
   "settings.languageDescription": "Sprache für die Benutzeroberfläche des Portals. Inhalte werden nicht übersetzt — nur die Oberfläche. Bei der ersten Anmeldung wird die Sprache aus dem Browser übernommen; hier lässt sie sich überschreiben.",
+  "settings.timeZone": "Anzeige-Zeitzone (IANA)",
+  "settings.timeZoneDescription": "Die Anzeige-Zeitzone bestimmt, wie Zeitstempel im Portal dargestellt werden — gespeichert wird weiterhin in UTC, nur die Anzeige wird umgerechnet. Bei der ersten Anmeldung wird sie aus dem Browser übernommen; hier lässt sich eine benannte IANA-Zone auswählen, falls die Erkennung falsch liegt (die Sommerzeit wird automatisch berücksichtigt). Ohne Angabe werden Zeiten in UTC angezeigt.",
   "notifications.title": "Benachrichtigungen",
   "notifications.markAllRead": "Alle als gelesen markieren",
   "notifications.allCaughtUp": "Alles erledigt.",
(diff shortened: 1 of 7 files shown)


sierragolflima:
Addressed in `06166e251` — but the stated reason does not hold, and the difference seemed worth recording rather than quietly accepting.

**The CS1574 claim is false here.** Test projects do not set `GenerateDocumentationFile` (only the src-tree `Directory.Build.props` does), so no `.xml` is produced and cref validation never runs. Verified with a clean `--no-incremental -warnaserror` build of `MeshWeaver.Graph.Test`: `0 Warning(s), 0 Error(s)`, both before and after the change.

**The underlying point is still right, for a smaller reason.** `Memex.Portal.Shared` genuinely is not referenced by this project, so the cref was a link that resolves nowhere — misleading to a reader, and a real CS1574 the day anyone enables XML docs for tests. It now names `PlatformUpdateChip.Describe` in `<c>` and says why.

PR #1916, opened 2026-08-20
Shed the OpenAI SDK cluster from the platform — modules carry it

rbuergi:
The fix-forward counterpart of #1912: instead of the platform pinning module dependencies into `/app`, the modules carry them (via #1914's derived bundle closure) and the platform sheds them.

Removes from `MeshWeaver.AI`: `Microsoft.Extensions.AI.OpenAI` (the #1912 pin), `Microsoft.Agents.AI.OpenAI`, `Azure.AI.OpenAI`, `OpenAI`. No other project references them; no MeshWeaver.AI source uses their types (the failure classifier reads exception messages; UsageTokens matches key strings). Full-solution Release build green.

**🚨 DO NOT MERGE until the fat module bundles (Plugins #539, packed under #1914's lane) are published AND re-landed on the production instances.** Once an image built from this change rolls, `/app` no longer carries the OpenAI SDKs — only the module folders do. #1914's riding-diamond rule makes the transition safe in either order *for re-packed bundles*, but the currently-landed entry-only bundles would fault.

Sequencing:
1. #1914 merges (bundle lane derives closures)
2. Plugins #539 merges → fat bundles published to the registry
3. Bundles re-landed on memex-cloud (+ restart) → prod verified green on rc5
4. **then** this merges → next image is leaner

### DIFF
--- a/src/MeshWeaver.AI/MeshWeaver.AI.csproj
+++ b/src/MeshWeaver.AI/MeshWeaver.AI.csproj
@@ -27,7 +27,6 @@
     </ItemGroup>
     <ItemGroup>
         <PackageReference Include="YamlDotNet" />
-        <PackageReference Include="Azure.Core" />
         <PackageReference Include="HtmlAgilityPack" />
         <!-- DateTime.Humanize() in ThreadMessageLayoutAreas. Declared explicitly: it used to ride
              in as a transitive of JsonPointer.Net (removed in #1231). -->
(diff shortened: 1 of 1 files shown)


rbuergi:
Widened per the sweep: `Azure.Core`, `Microsoft.Agents.AI.AzureAI` and `Microsoft.Agents.AI.Workflows` leave as well — MeshWeaver.AI's source uses only `Microsoft.Agents.AI` + `Microsoft.Extensions.AI(.Abstractions)`; zero usages of anything Azure. The AzureFoundry module already bundles its own Azure/Agents SDKs (43 files). Full-solution build green. Still gated on the fat bundles being landed on prod first.

Issue #1949, opened 2026-08-20
🚨 Generation-landed modules NEVER load at boot — activation gate checks legacy modules/<name>/ while landing writes modules/<name>@<gen>/


rbuergi:
**P0 — ships in 3.0.0-rc6.ci.4684 (a900561); every consumer that lands modules via today's generation-directory landing boots with ZERO store modules, forever.**

## Live impact (memex.systemorph.com, 2026-08-20 evening)
All 16 default modules (MeshWeaver.Mcp, every AI provider, Courses, …) landed correctly into `/data/modules/<name>@<gen>/` with activation entries carrying `directory`, yet every boot printed for each:

```
[ModuleActivation] SKIPPED store-installed module 'MeshWeaver.Mcp': its landed DLL 'modules/MeshWeaver.Mcp/MeshWeaver.Mcp.dll' does not exist (folder lost or never landed; …) — re-install the module
```

→ `/mcp` 404, `no chat-client factory resolves model` on every thread. Each boot then RE-LANDS all modules into fresh generations (the reconciler's `landed` lookup path sees them, but the boot gate doesn't) — landing and activation never converge.

## Root cause
`ModuleActivationBoot.ComputeEffectiveModuleEntries` gates each persisted entry on `landedModuleDllExists(module.Name)`, and production passes `ModuleActivationBoot.LandedModuleDllExists(root, name)` = `{root}/modules/{name}/{name}.dll` — the LEGACY fixed folder ONLY. The generation redesign (`ModuleActivationEntry.Directory`, 2026-08-20) moved the bytes to `modules/{name}@{gen}/` and nothing writes the legacy folder any more. The gate and the resolver disagree: `MemexConfiguration` resolves the load path via `landed.Directory` (correct), but the entry is dropped by the existence gate before it ever gets there.

## Fix shape
The existence check must honor the entry's generation pointer: pass the `ModuleActivationEntry` (not just the name) into the gate, and check `{root}/modules/{entry.Directory ?? entry.Name}/{entry.Name}.dll`. Keep the deliberate 'never fall back to the app closure' rule. Pin with a test: entry with `Directory` set + DLL only in the generation dir must LOAD (today it's skipped).

## Stopgap currently live on ns memex (remove once fixed)
[...]

PR #1963, opened 2026-08-21
helm: extraEnvFrom — declare CSI-synced env sources instead of losing them on upgrade

rbuergi:
The portal's `envFrom` was a fixed pair (configmap + `memex-portal-secrets`), so a deployment reading env from a Key Vault CSI-synced Secret could only attach it via a live kubectl patch — which the next `helm upgrade` silently drops. memex-cloud runs exactly that shape today (`memexcloud-portal-ai-secrets` carries its AI provider keys), so its first repo-driven helm release would detach every provider key.

- New `extraEnvFrom` values key: verbatim `EnvFromSource` objects appended after the chart's own two.
- Verified with `helm template`: renders when set, no-op when unset.
- Follow-up in Systemorph/Memex: memex-cloud's public overlay declares its CSI secret via this key.

No What's New entry — chart-internal plumbing, no user-visible change of its own.

Incident context: the 2026-08-20 memex.systemorph.com outage (#1949/#1950, Systemorph/Memex#77).

### DIFF
--- a/deploy/helm/templates/memex-portal/deployment.yaml
+++ b/deploy/helm/templates/memex-portal/deployment.yaml
@@ -111,6 +111,16 @@ spec:
                 name: "memex-portal-config"
             - secretRef:
                 name: "memex-portal-secrets"
+            # 🚨 Extra env sources the chart does not own — typically a Key Vault CSI-synced
+            # secret (secrets-store SecretProviderClass → k8s Secret). Without this key such a
+            # source can only be wired by a live kubectl patch on the Deployment, which the next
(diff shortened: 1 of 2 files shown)


rbuergi:
Extended with the volume half: `extraVolumes` + `extraVolumeMounts` (verbatim objects, same conditional shape). memex-cloud's live CSI secrets-store volume/mount is what keeps its AI-keys secret syncing — a repo deploy dropping the mount would stop rotation even with `extraEnvFrom` in place. `helm template` verified set + unset.

Issue #1580, opened 2026-08-14
Modularize importers/parsers: node types declare how their content parses — no compiled type branches


rbuergi:
## Problem (companion to #1576, which covers export)

The IMPORT side has the same hard-coding as export: the persistence parsers and importers are compiled per-format classes with per-TYPE branches baked in.

- `MarkdownFileParser` special-cases slides (`IsSlideNodeType` → build `SlideContent` instead of `MarkdownContent`, round-trip Notes/Background frontmatter) — a compiled branch for one node type's content shape.
- `CSharpFileParser` owns the `// <meshweaver>` heading contract; `.json` covers typed nodes — the extension→parser map is compiled and closed.
- A plugin cannot declare "files of my node type parse/serialize like THIS" — any plugin type with a richer file representation either degrades to the generic format or needs a new compiled branch (exactly how the slide branch got there).

## Direction

Same composition-first shape as #1576: parsing/serialization becomes a DECLARATION carried by the node type — `config.WithContentParser(...)` (or a parser descriptor on the NodeType definition) providing: file extension/format, frontmatter→content mapping, content→frontmatter round-trip. The compiled parsers keep only the format mechanics (frontmatter split, markdown pipeline); the per-type content reconstruction moves to the type's declaration, which for plugin types lives in plugin content.

## The hard edge to design first

Persistence parsing runs BEFORE node types activate (import/bake/git-sync parse files to CREATE the nodes whose types would carry the declarations) — the reason `MarkdownFileParser` stays compiled today. So the declaration must be resolvable at parse time without an activated hub: plausibly off the installed NodeType definition node (its JSON is in the store before activation), or a two-phase import (generic parse → typed re-materialization once the type is known). This bootstrapping edge is the design work; the rest is mechanical.

Suffix-aware bridge already in place: `MarkdownFileParser.IsSlideNodeType` delegates to `SlideNodeType.Matches` (PR #1579), so plugin-typed slides round-trip today. The declaration design deletes the branch.

Part of the composition-first program (WS7).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

rbuergi:
## Verdict: PARTLY DONE — the design is settled and recorded; the import half is the remaining work, and its obvious shortcut is now **measured** to be unsafe

Verified against `origin/main`. State today:

| | |
|---|---|
| The bridge (suffix-aware `SlideNodeType.Matches`) | in place — `MarkdownFileParser.cs:245` |
[...]

Issue #1576, opened 2026-08-14
Modularize the export layer: node types declare their export composers — no compiled type gates


rbuergi:
## Problem

The export subsystem is hard-coded per node type, in compiled code:

- `DeckExportMenuProvider` / `MarkdownExportMenuProvider` are separate providers, each gating on a NodeType comparison (`Deck` / `Markdown`) — a new exportable type means a new compiled provider.
- `ExportPdf.csx` / `ExportHtml.csx` branch `if (isDeck)` to pick the composition strategy (slide chapters vs single document).
- `ExportDocumentLayoutArea` decides pixel-fidelity availability off the same compiled type check.

The suffix-aware `DeckNodeType.Matches`/`SlideNodeType.Matches` gates (follow-up to the Publish-pack extraction) let plugin-typed decks/slides through these compiled gates — but they are a bridge, not the architecture.

## Direction (composition-first, same pattern as layout areas)

Exportability becomes a DECLARATION on the node type's hub configuration, contributed by the type's own configuration lambda (which lives in plugin content for plugin types):

1. A compiled public surface `config.WithExportComposer(descriptor)` — format(s), menu label key, fidelity options, and the composition strategy (how this type's content becomes an export document: e.g. deck → one chapter per slide in deck order).
2. ONE generic export menu provider that renders entries from the declarations on the node's hub config — `DeckExportMenuProvider` and `MarkdownExportMenuProvider` collapse into it and are deleted.
3. The export templates dispatch through the declared composer instead of `if (isDeck)` — the deck branch moves into the Publish pack's declaration.
4. The suffix gates from the bridge PR are then deleted.

Design constraint: the hub config freezes at hub build, so declaration changes ride the existing generation-bump → hub-recycle mechanism (same edge as the WS7 menu/composition design — solve once, use for both).

Part of the composition-first program (WS7): a deployment = base runtime + plugin composition; the export if-chains are one of the last compiled per-type switches.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

rbuergi:
## Verdict: PARTLY DONE on main → **finished** by the two PRs below

Verified against `origin/main` (not PR titles). Phase 1 (#1594) landed the declaration and the one generic menu provider. **Three compiled type gates survived it**, and the acceptance list's item 4 was still open:

| item | on main today | |
|---|---|---|
[...]

PR #1973, opened 2026-08-21
refactor(export): delete the export layer's last compiled type gate (#1576 item 4)

rbuergi:
Closes the export half of the composition-first programme. Companion to **MeshWeaver.Plugins#553**, which must merge **first**.

## What this deletes

`ExportDeclaration.Resolve(configuration, nodeType)` — the ONE transition fallback from the bridge PR (#1579): a suffix-aware `Deck`/`*/Deck` check that answered `SlideDeck` for plugin decks whose in-mesh configuration lambdas predated `WithExport`.

It has **no callers left**. MeshWeaver.Plugins#553:

- makes `Publish/Deck` DECLARE `.WithExport(ExportDeclaration.SlideDeck)`;
- moves `ExportMenuProvider` and `ExportDocumentLayoutArea` onto `Configuration.Get<ExportDeclaration>()`;
- makes `ExportPdf.csx` / `ExportHtml.csx` dispatch on a `composition` input that `ExportDocumentHandler` stamps from the type's declaration (the handler runs on the source node's own hub, so it can read the config a template cannot).

Swept for in-mesh callers too — `content/`, `samples/`, `memex/`, `clients/` — since a NodeType's `configuration` lambda is C# in a JSON string and invisible to `dotnet build`. Zero hits.

## 🚨 Merge order

**MeshWeaver.Plugins#553 first.** The fallback is what keeps a deck exportable on an install whose Publish pack has not yet synced its declaration. Removing it before that ships costs decks their Export menu until the next sync — recoverable, but pointless.

## Why a name gate must not come back

Its failure mode is **not an error**. A type it does not recognise composes as a plain document, and a deck-shaped node has no body of its own — so the export *succeeds* and writes an empty file: green activity, nothing logged, nothing in the document. That is exactly the shape `DeckNodeType.Matches` was introduced to patch, and patching by name only changes *which* types are silently excluded. Plugins#553 reproduces it as a control run.

## Test

`ExportDeclarationTest` pins the contract from the other side — an undeclared type has no exports **whatever it is called**, `Deck` and `Publish/Deck` included:

```
Passed  ExportDeclarationTest.UndeclaredType_HasNoExports_WhateverItIsCalled(nodeType: "Publish/Deck")
Passed  ExportDeclarationTest.UndeclaredType_HasNoExports_WhateverItIsCalled(nodeType: "Deck")
Passed  ExportDeclarationTest.UndeclaredType_HasNoExports_WhateverItIsCalled(nodeType: "Markdown")
Passed  ExportDeclarationTest.UndeclaredType_HasNoExports_WhateverItIsCalled(nodeType: null)
Passed  ExportDeclarationTest.BuiltInMarkdownType_DeclaresDocumentComposition
Passed  ExportDeclarationTest.DeclaredType_GetsExactlyItsDeclaration
Total tests: 6   Passed: 6
```

The *behaviour* test that fails against today's `main` lives in Plugins#553 (a declared type no suffix check can match, exporting an empty document before the fix) — a pure deletion of dead code cannot have one here.

Full suites:

```
MeshWeaver.Graph.Test          Failed: 0, Passed: 1392, Skipped: 1, Total: 1393  (1 m 34 s)
MeshWeaver.Documentation.Test  Failed: 0, Passed:   39, Skipped: 0, Total:   39
```

Release + `-warnaserror` clean on Graph, Hosting, Documentation.

## Documentation — and the measured blocker for #1580

`Doc/Architecture/DeclarativeImportExport` now records the shipped export design, how a declaration reaches a template that cannot read a hub configuration, and — for the still-open import half (#1580) — **why its obvious shortcut is refused**:

```
slide files: 235 ; WITHOUT Notes/Background: 3
```

A content-shape rule ("frontmatter carries Notes or Background ⇒ build slide content") silently empties those three, because the slide views read `SlideContent` and `ContentAs<T>` recovers only a same-short-named type. The doc also states the two conditions the registry route must settle first (a cold-boot miss must not lose the extras; a hit must not drop the body into a record with no `content` member). `MarkdownFileParser.IsSlideNodeType` — the programme's **last** compiled type branch — carries a pointer to that section.

## What's New

Skipped: nothing an end user sees changes in this PR. The user-visible half ships in Plugins#553.

### DIFF
--- a/src/MeshWeaver.Documentation/Data/Architecture/DeclarativeImportExport.md
+++ b/src/MeshWeaver.Documentation/Data/Architecture/DeclarativeImportExport.md
@@ -4,7 +4,7 @@ How a node type tells the platform what its content can DO — export to PDF, ro
 a git file — without the platform carrying a compiled `if (nodeType == …)` for it. The goal is
 composition-first: a plugin pack declares; core mechanisms read declarations.
 
-## Export: shipped (phase 1)
+## Export: SHIPPED — no compiled type gate survives (#1576)
 
 A node type declares its export behaviour on its own hub configuration:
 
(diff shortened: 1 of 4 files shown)


rbuergi:
## 🚨 The merge order in the description is NOT enforced by any gate — filed as #1980

"Plugins compile against this PR" reports **pass** on this PR, and that is a false green for the change it is being asked about: this PR deletes `ExportDeclaration.Resolve`, and MeshWeaver.Plugins' `MeshWeaver.Markdown.Export` calls it on `main` today.

The gate stages node repos only —

```yaml
```

— and drives `mw-plugin-test` over their in-mesh NodeType `Source/*.cs`. It never builds MeshWeaver.Plugins' `src/`, so the fourteen modules that left in `0b3b4266f` have no compile gate in this repo at all. Verified on run 32483032713 / job 96775118860.

So **Plugins#553 must merge first**, and today only this sentence enforces that. #1980 proposes the gate that should.

rbuergi:
Copilot's finding addressed in `68925a3de`; thread resolved.

It was right, and the criticism was the sharp kind: my theory took a `nodeType` it never used, so all four cases ran the same `Get<ExportDeclaration>()` check and would have passed unchanged had the deleted `Resolve` come back. A test that cannot fail is not a test.

What distinguishes the two designs is not the RESULT of an undeclared read — null either way — but **the shape of the surface**: the fallback existed because a reader could ask *"and what is this type called?"*, and the fix was to stop offering anywhere to ask. `NoPublicExportApi_AcceptsANodeType` asserts that directly — no public member of `ExportDeclaration`/`ExportDeclarationExtensions` takes a node-type parameter.

Proven non-vacuous by restoring the deleted method:

```
Failed  ExportDeclarationTest.NoPublicExportApi_AcceptsANodeType
  The export declaration is resolved from the type's hub CONFIGURATION alone. A member taking a
  node type is how the deleted suffix fallback was expressed … Found:
  ExportDeclaration.Resolve(configuration, nodeType)
```

and green again once removed — 4/4.

Issue #1982, opened 2026-08-21
Model configuration has no administered home — it lives in Helm values, a create-if-absent seeder and live nodes at once


rbuergi:
Split out of #1965 (part B). #1965 keeps its BUG half; this issue is the architectural one, and it is
a **problem statement first**: model configuration has no single administered home, and the symptoms
below are not four bugs — they are four faces of that.

## The problem: a model's configuration lives in three places that cannot see each other

| Where | Who writes it | Who reads it | Can it change after deploy? |
|---|---|---|---|
| Helm values / env (`Anthropic__ApiKey`, `AzureFoundry__Models__3`, `ModelTier__*`) | a chart edit + a roll | each factory's `IOptions<…Configuration>`, `BuiltInLanguageModelProvider` | only by editing the chart and rolling |
| the static seeder (`BuiltInLanguageModelProvider.GetStaticNodes`) | the platform, at boot | the static/DB catalog import | **no — it is create-if-absent** |
| `ModelProvider` / `LanguageModel` / `ModelTier` NODES | an admin, live | `ChatClientCredentialResolver`, the picker, the tier registry | yes |

Nothing reconciles them, and **the seam between the first and the third is one-way and one-shot**:
the seeder copies `{Section}:ApiKey` onto the `ModelProvider` node the first time it creates that
node, marks it `SyncBehavior.ExcludeThisAndChildren`, and never revisits it. So a key added to a
deployment's configuration AFTER the node exists reaches every factory and never reaches the node —
and the node is what the admin UI, the picker and the resolver read.

That is not hypothetical. On `memex.systemorph.com`, `Provider/Anthropic` was created by
`system-security` on **2026-08-14** with no key; `Anthropic__ApiKey` was configured in the
deployment afterwards; the node stayed keyless until a human pasted the key into it by hand on
**2026-08-21**. In between, the deployment could serve `claude-*` rounds and the platform believed
it could not.

## Four instances, all currently open

1. **#1965 (this issue's parent).** `HasUsableCredential` read only the node, so an explicitly
   pinned `claude-sonnet-4-6` was declared unusable and `ApplyStaleModelFallback` silently swapped
   it for the deployment default. Fixed by teaching the resolver the config rung — but that
   is a **compensator for the missing surface, not a cure**: it makes the platform agree with what
   runs, it does not give anyone a place to change it.
2. **#1925.** `AzureFoundry__Models__3` is set in AKS values and rendered by **no template** — a
[...]

Issue #1925, opened 2026-08-20
Three AKS config keys are set in values and rendered by no template — SelfUpdate__MinRollInterval, WebhookInbox__Targets__0, AzureFoundry__Models__3


sierragolflima:
## Three config keys are set in `deploy/aks/values.aks.yaml` and rendered by nothing

Found while fixing the same defect for `Modules__Root` (#1924). Filing rather than patching, because templating each of these **changes live AKS behaviour**, and two are recent, deliberate work by their authors.

### The mechanism

`deploy/helm/templates/memex-portal/memex-portal/config.yaml` names every ConfigMap key explicitly, there is no catch-all range over `config.memex_portal`, and the Deployment's only env path is `envFrom` on that ConfigMap plus the Secret. **A values key the template omits reaches no container, with no error anywhere.** helm is happy: the values file is syntactically perfect and the template simply never mentions it.

### The three

| key | set at | intended effect | templates that emit it |
|---|---|---|---|
| `SelfUpdate__MinRollInterval: "01:00:00"` | `values.aks.yaml:288` | the restart-budget floor from #1780 — so hourly publication does not mean hourly pod restarts | **0** (no `SelfUpdate__` key is templated at all) |
| `WebhookInbox__Targets__0: "Hosting/PlatformBuilds"` | `values.aks.yaml:88` | allowlists the platform-build webhook target (also set in Memex#52) | **0** |
| `AzureFoundry__Models__3: "Kimi-K2.6"` | `values.aks.yaml:305` | a fourth Foundry model | **0** — while `Models__0/1/2` *are* templated at config.yaml:228-230 |

`AzureFoundry__Models__3` is the clearest tell: its three siblings are templated and it is not, so this reads as an addition that missed the template rather than a deliberate omission.

`SelfUpdate__MinRollInterval` is the one worth reading twice. AGENTS.md documents it as the fix for restart churn — *"#1780 closed that with `SelfUpdate__MinRollInterval`, a restart budget set to `1h` in `values.aks.yaml` beside the 24h/6h/1h trade"* — and warns that its predecessor key `SelfUpdate__PollInterval` **"sat in the chart inert, reading as a live daily throttle, until #1778."** That is this defect, one key earlier, already fixed once.

### How to confirm

```bash
grep -rn "SelfUpdate__MinRollInterval\|WebhookInbox__Targets__0\|AzureFoundry__Models__3" deploy/helm/templates/
# → no matches

[...]

Issue #1880, opened 2026-08-19
Collapse the duplicate Azure Foundry provider data on memex.meshweaver.cloud — two nodes, plus a stale label the seeder will not overwrite


sierragolflima:
## Summary

In the AI chat "Choose a model" dialog on `memex.meshweaver.cloud`, the Azure Foundry models appear twice under two differently-titled headers — `AZURE AI FOUNDRY` (2 models) and `AZUREFOUNDRY` (6 models) — with **DeepSeek V4 Flash** and **DeepSeek V4 Pro** present in both. Two further models (`Kimi K3`, `GLM 5.2`) render inside the `AZUREFOUNDRY` block although they belong to OpenRouter, and no `OPENROUTER` header appears at all.

This is **three separate defects** that compound into one confusing dialog. Only defect 1 is data-dependent; defects 2 and 3 are pure code defects reproducible anywhere.

### Observed vs. expected

| | Observed | Expected |
|---|---|---|
| Azure Foundry | two groups, 2 overlapping models | one group |
| Seeded group title | `AZUREFOUNDRY` | `AZURE FOUNDRY` |
| OpenRouter models | listed under `AZUREFOUNDRY`, no own header | listed under `OPENROUTER` |

## How the picker groups

Verified in `src/MeshWeaver.Blazor.Portal/Chat/ThreadChatView.razor.cs`:

1. `ArrangePickerGroups` (line 2358) keys every group on `IsPickerHeaderNode(n) ? n.Path : ParentPath(n.Path)` — the **node path**, never `content.provider`.
2. `ParentPath` (line 2346) takes everything before the **last** `/`.
3. `RemoveEmptyPickerGroups` (line 2371) drops any `ModelProvider` header whose path no model's `ParentPath` matches.
4. The header text is the `ModelProvider` node's `Name` (`ThreadChatView.razor` line 904); the uppercasing is CSS (`text-transform: uppercase`), so the stored names really are `Azure AI Foundry` and `AzureFoundry`.

## Defect 1 — two `ModelProvider` nodes describe the same upstream provider

Both nodes below carry the **same** `content.provider` (`AzureFoundry`), the same endpoint and the same API key, but sit at different paths, so they render as two groups:

| Path | `Name` | `content.label` | `models` | Author | Modified (Zurich) |
|---|---|---|---|---|---|
[...]

PR #1983, opened 2026-08-21
A model the deployment can serve must not be declared unusable (#1965)

### LINKED ISSUE #1965: Store-landed Anthropic module registers NO chat-client factory — and models must be administrable from the deployment module, not kubectl
Two connected findings from the 2026-08-21 model-parity work on memex.systemorph.com (maintainer request: "see that we can administer models directly from the deployment module").

## A. The landed MeshWeaver.AI.Anthropic module registers no factory
- ns `memex` (ci.4718) loads all 16 store-landed modules (generation boot verified). `MeshWeaver.AI.AzureFoundry`'s factory works (DeepSeek rounds run). But a round pinned to `claude-sonnet-4-6` was served by the **OpenRouter factory** (failed with OpenRouter's 402) — meaning `AnthropicChatClientAgentFactory` (Order 0, `Supports(claude*)`) is absent from DI.
- Env is complete: `Features__Ai__Providers__Anthropic=true`, `Anthropic__Endpoint=https://s-meshweaver.services.ai.azure.com/anthropic/`, `Anthropic__Models__0..2`, `Anthropic__ApiKey` (endpoint+key verified working with a direct /v1/messages call).
- The landed DLL (`modules/MeshWeaver.AI.Anthropic@ffb86f44`, bundle version 1.0) DOES contain `AnthropicProvidersAttribute` (binary grep) — so either the 1.0-era bundle's `WithGlobalServiceRegistry` contract predates what the current loader applies, or the registration throws and is swallowed silently. Nothing about it in the logs at Warning+.
- Also worth checking on memex-cloud: it lists claude models (Provider/Anthropic nodes with node-held key) with `Features Anthropic=false` — do claude rounds actually work there, or does OpenRouter swallow them too?
- Related second symptom: with `Anthropic__Models__*` configured, the DB-synced Provider partition did NOT materialize the claude LanguageModel nodes at boot (kimi-k3 from `OpenRouter__Models__1` DID materialize) — created by hand for now.

## B. Administer models from the deployment module
Everything in this incident chain — endpoints, model lists, tiers, keys — was kubectl/config surgery. The deployment's admin surface (Settings → Language Models / the deployment module) should own it end to end:
- add/remove provider endpoints + keys (stored on ModelProvider nodes, enc-at-rest),
- add/remove models per provider, set order/pricing/tier labels,
- edit the tier→model mapping (the `Provider/Tier` registry exists; the `ModelTier__*` env override should become the fallback, not the primary),
- see per-model HEALTH (last round outcome — e.g. OpenRouter 402 credits-exhausted, DeepSeek 410 deprecated) so a dead model is visible in the admin UI instead of discovered by users mid-thread.
A config change through that surface must survive helm upgrades by construction (nodes, not env).

Context: OpenRouter account credits are exhausted (every full round HTTP 402, both portals) — tiers repointed to DeepSeek V4 in the meantime (Systemorph/Memex#80).

rbuergi:
Closes #1965 (its BUG half — the administration half is split out as #1982, per the maintainer's
decision to separate the two).

## First: was it a REGISTRATION failure or a LOADING failure?

**Loading — and it is already fixed.** The issue's title claim ("the landed Anthropic module
registers NO chat-client factory") does not reproduce on any current build:

- `memex.systemorph.com` right now serves a `MeshWeaver.AI.Anthropic` node of `nodeType:
  ModuleDefinition`, alongside 16 others. That node is *produced by* `AnthropicProvidersAttribute.Nodes`,
  and `MeshBuilder.InstallServices` applies each node's `GlobalServiceConfigurations` in the same
  pass that collects it — so the node existing IS the registration having run.
- The maintainer's own follow-up on #1965 (07:56Z, from a Debug-log session on ci.4730) reports
  `[AnthropicChatClientAgentFactory] Initialized with Endpoint=…/anthropic/, ApiKey=set, Models (3)`
  in the boot log, and calls the no-factory theory "an earlier boot's mis-served round".
- That earlier boot is exactly **#1949/#1954**: the activation gate looked in `modules/<name>/`
  while landing writes `modules/<name>@<gen>/`, so a generation-landed module never loaded and its
  attribute was never read. `MeshWeaver.AI.Anthropic` landed as `modules/MeshWeaver.AI.Anthropic@ffb86f44`
  — a generation directory — which is why "the DLL contains the attribute" and "no factory in DI"
  were both true at once.

The same cause explains the issue's second symptom (`Anthropic__Models__*` configured but no claude
`LanguageModel` nodes materialized): `BuiltInLanguageModelProvider` iterates the **registered
catalog sources**, and the Anthropic source ships in the same unloaded module. No source, no nodes.

`ProviderModulePackBootTest` (added here) pins the half the platform owns, so a real regression of
this shape fails a test instead of a portal: a provider module's assembly attribute must fold BOTH
its catalog source and its `IChatClientFactory` into DI through `MeshBuilder.InstallAssemblies`.
The provider modules themselves live in MeshWeaver.Plugins now, so the platform can only pin the
contract — the pack in the test is modelled line-for-line on `AnthropicProvidersAttribute`.

## What this PR actually FIXES

The defect #1965 narrowed to, and the one that cost a week of wrong model routing.

Every provider factory resolves its driver config in two steps —
`endpoint = resolution.Endpoint ?? configuration.Endpoint`,
`apiKey = resolution.ApiKey ?? configuration.ApiKey`. So a deployment that keeps a provider's key in
configuration (`Anthropic__ApiKey`) serves that provider's models perfectly well **even when the
`ModelProvider` node carries no key**. `ChatClientCredentialResolver` read only the node, so:

- `HasUsableCredential("claude-sonnet-4-6")` → **false**, for a model that runs;
- `AgentChatClient.ApplyStaleModelFallback` reads "unusable" as "stale selection" and **silently**
  swaps the user's explicit pick for the deployment default — deliberately silent for the user, and
  correctly so, which is why nobody could see it.

The resolver's own doc asserted this was impossible — *"a model whose key comes from the factory's
own config/IOptions fallback … already resolves `Missing` today and was never eligible as a catalog
default"* — written as a no-regression reassurance. **That claim was the defect**, and it is now
replaced by the opposite statement plus the incident that falsified it.

### Why the node is keyless in the first place

The catalog seeder stamps `{Section}:ApiKey` onto the `ModelProvider` node **create-if-absent** and
marks it `SyncBehavior.ExcludeThisAndChildren` — so a key added to a deployment AFTER the node
exists reaches every factory and never reaches the node. Measured, not inferred:
`Provider/Anthropic` on memex.systemorph.com was created by `system-security` on **2026-08-14** with
no key, `Anthropic__ApiKey` was configured after that, and the node stayed keyless until a human
pasted the key into it on **2026-08-21**. Everything in between resolved as unusable.

### The change

`Resolve` gains a fourth rung — the deployment's own configuration for the catalog source whose
`ProviderName` the model declares. Three deliberate constraints:

- **Last.** Consulted only when the node rungs produced no key; administered node data always wins.
- **Merged per FIELD, like the factories do it.** A node-supplied endpoint survives; only the
  missing key is filled in. Merging whole resolutions would let a config endpoint override an
  administered one.
- **Key or nothing.** A section with an endpoint and no key returns null, so an endpoint-only
  provider (the `z-ai/glm-5.2` report) stays correctly unusable — the fix must not turn every model
  green.

`Source` now names both rungs (`providerRef:Provider/Anthropic+config:Anthropic`), because "which
source answered" is the question an operator asks when a key is stale — and it is the first half of
what #1982 asks the admin surface to render.

**This is a compensator, not a cure.** It makes the platform agree with what actually runs; it does
not give anyone a place to administer model configuration. That is #1982, filed as the problem
statement (config spread across Helm values, a create-if-absent seeder and live nodes, with #1925
and #1880 as two more faces of it).

## Non-vacuity

`ModelWhoseKeyLivesOnlyInDeploymentConfig_IsUsable` was run against **unmodified `origin/main`
(ed769bf0f)** before the fix existed, and fails with the exact shape of the incident:

```
MeshWeaver.Reactive.Assertions.ObservableAssertionException :
  Expected the observable to emit a value matching the predicate within 15s because the deployment's
  configured key serves this provider's models, but it did not. Last of 300 emission(s) was:
  CredentialResolution { Endpoint = https://node.example/anthropic/v1/messages, ApiKey = ,
                         Source = providerRef:Provider/AnthropicProbeExisting }.
```

`ApiKey` empty while `AnthropicProbe:ApiKey` is configured. The negative test
(`ModelWithNoKeyAnywhere_StaysUnusable`) passed both before and after, so the fix is bounded.

## Verification

| What | Result |
|---|---|
| `MeshWeaver.AI.Test` (full, `-c Release`) | **1181 passed / 0 failed / 3 skipped**, 1 m 54 s |
| `MeshWeaver.Acme.Test` — model picker | 5 passed |
| `MeshWeaver.Hosting.Monolith.Test` — LanguageModel / AgentChatClient / AgentSyncedQuery | 5 passed |
| `MeshWeaver.Threading.Test` — `DelegationSubThreadUsageTest` | 1 passed |
| `MeshWeaver.Documentation.Test` — `WhatsNewEntryIntegrityTest` | 1 passed |
| `MeshWeaver.AI`, `MeshWeaver.Documentation`, `Memex.Portal.Shared` | `-c Release -warnaserror`, 0 warnings |

No test or portal `appsettings.json` carries a provider `ApiKey`, and the test `IConfiguration` is
built from `appsettings.json` only (no environment variables), so the new rung is inert everywhere
except where a test sets it deliberately.

What's New: `Category: Fix` — a user picking a model and getting a different one is user-visible.

### DIFF
--- a/src/MeshWeaver.AI/AgentChatClient.cs
+++ b/src/MeshWeaver.AI/AgentChatClient.cs
@@ -2129,10 +2129,15 @@ public class AgentChatClient : IAgentChat
         if (string.IsNullOrEmpty(target)
             || string.Equals(target, currentModelName, StringComparison.OrdinalIgnoreCase))
         {
-            // Nothing usable to move TO. Not fatal on its own — a deployment can keep its keys in
-            // factory config where the resolver cannot see them — so the round still tries. The
-            // latch lets CreateAgentsSync turn "no usable model AND no agent could be built" into a
-            // speaking round failure instead of a raw provider error (#476).
+            // Nothing usable to move TO. Not fatal on its own — a factory may still bind a key this
+            // chain never saw (a provider with no registered catalog source, a driver reading its own
+            // config shape) — so the round still tries. The latch lets CreateAgentsSync turn "no
+            // usable model AND no agent could be built" into a speaking round failure instead of a
+            // raw provider error (#476).
+            //
+            // 🚨 The deployment's OWN configured keys are no longer in that blind spot: the resolver
+            // reads {Section}:ApiKey for the model's provider (MeshWeaver#1965), so a model served by
+            // Anthropic__ApiKey now takes the early return above instead of being swapped away.
             modelFallbackExhausted = true;
             if (isRouter)
                 logger.LogWarning(
--- a/src/MeshWeaver.AI/ChatClientCredentialResolver.cs
+++ b/src/MeshWeaver.AI/ChatClientCredentialResolver.cs
@@ -21,7 +21,12 @@ namespace MeshWeaver.AI;
 /// </summary>
 public record CredentialResolution(string? Endpoint, string? ApiKey, string Source)
 {
-    /// <summary>Sentinel result meaning no credential could be resolved — the caller falls back to its <c>IOptions</c> configuration binding.</summary>
+    /// <summary>
+    /// Sentinel result meaning no credential could be resolved ANYWHERE — neither on a
+    /// <c>ModelProvider</c> node nor in the deployment's configuration for the model's provider
+    /// section. Factories still apply their own <c>IOptions</c> fallback afterwards, but it now
+    /// resolves the same values this chain already looked at, so <c>Missing</c> means what it says.
+    /// </summary>
     public static readonly CredentialResolution Missing = new(null, null, "missing");
 }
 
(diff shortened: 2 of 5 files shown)


rbuergi:
CI is green after a re-run of one shard. Recording why, so the re-run is auditable rather than a shrug:

The first attempt failed **shard 2** with `[CI] MeshWeaver.FutuRe.Test exit=139 SIGNAL SIGSEGV (host died on a signal, not an assertion)` — the test host segfaulted ~50 s into that project, no failing assertion. `MeshWeaver.FutuRe.Test` does not reference `MeshWeaver.AI` at all (its ProjectReferences are Fixture / Hosting / Hosting.Monolith(.TestBase) / Layout / Graph / Mesh.Contract / Messaging.Hub), and this PR touches only `src/MeshWeaver.AI`, two new files in `MeshWeaver.AI.Test`, and one Documentation data node — so the diff cannot reach it semantically.

Re-ran the failed jobs only; shard 2 passed in 9m31s and `Consolidate test results` is green. All six shard jobs report `conclusion: success` and actually EXECUTED (5–10 min each), so this is not a reused-green tree.

Not merged, and auto-merge deliberately not enabled.

Issue #1432, opened 2026-08-13
Adopting an already-durable external write MINTS a phantom revision and persists it (deterministic; root cause under #1384)


rbuergi:
Split out of #1384, where it surfaced as the surviving cross-branch flake bucket. The flake is a symptom; **the defect underneath is deterministic and has nothing to do with recycles or races.**

## The defect

A per-node hub reconciles its own node against the storage change feed, so that a write made by *someone else* (another writer, a migration, a repair job, GitSync, a second silo) is adopted into the in-RAM node. `src/MeshWeaver.Graph/MeshDataSource.cs:1394-1398`:

```csharp
// 🚨 FORWARD-ONLY refresh — never move the in-RAM node BACKWARD.
// `notification.Entity` is the node as PERSISTED. … a persisted snapshot may only
// REPLACE it when it is STRICTLY NEWER (a genuine out-of-band external write).
hub.GetWorkspace().GetMeshNodeStream()
    .Update(current =>
        current is not null && current.Version >= newNode.Version
            ? current
            : newNode)
```

The intent — stated in that comment — is to **replace** the in-RAM node with the persisted one. But `GetMeshNodeStream()` with no path is the OWN node, so this runs `UpdateOwn`, which mints **unconditionally** (`src/MeshWeaver.Mesh.Contract/MeshNodeStreamExtensions.cs:839-842`):

```csharp
updated = updated with { Version = MeshNode.NextVersion(Math.Max(current.Version, updated.Version)) };
```

So adopting a durable v5000 leaves the hub holding **v5001** — a revision that exists nowhere, with content identical to v5000. It is a fresh `with`-copy, so `OwnNodeCache.PersistedSnapshot`'s reference-identity gate does not suppress it; `.Do(pending.Track)` sits before the `Sample` (`MeshDataSource.cs:1168`), so the persistence sampler (or `FlushPendingOwnSave` at dispose) **writes the phantom revision back**; and `MonotonicWriteGuard` accepts it because 5001 > 5000.

**Adopting an already-durable external write is an observation. An observation must not mint, and must not write.**

## Reproduction — deterministic, 2.5 s, no race

Add to `test/MeshWeaver.Hosting.Monolith.Test/StaleActivationDurableFirstSeedTest.cs` (it reuses that class's `Gated` / `Storage` / `ReadDurable` helpers). The hub stays **live** — no recycle, no gate, no reactivation:

```csharp
[Fact(Timeout = 55_000)]
[...]

PR #1988, opened 2026-08-21
Add security response headers to the portal (CSP staged Report-Only; nosniff, Permissions-Policy, CORP enforced)

rbuergi:
## What
Adds a response-header middleware to the memex portal pipeline setting four security headers an OWASP ZAP full active scan of `memex.meshweaver.cloud` (2026-08-21) flagged. The scan found **zero exploitable vulnerabilities** (all injection/XSS/RCE/traversal/SSRF rules passed) — these are hardening headers.

| Header | Before | After | Mode |
|---|---|---|---|
| `Content-Security-Policy` | `frame-ancestors 'self'` only (no fallback) | complete policy: `default-src` fallback, `object-src 'none'`, `base-uri 'self'` | **Report-Only (staged)** |
| `X-Content-Type-Options` | absent | `nosniff` | enforced |
| `Permissions-Policy` | absent | denies camera/mic/geolocation/payment/usb/motion sensors | enforced |
| `Cross-Origin-Resource-Policy` | absent | `same-site` | enforced |

## Staged CSP rollout
The CSP ships as **`Content-Security-Policy-Report-Only`** first — it cannot block a single request, so it is safe to deploy to the production registry immediately while any would-be violations surface in the browser console. A **one-line follow-up** renames the header to `Content-Security-Policy` to enforce it once proven quiet.

The policy is a **superset of what the app already loads** — keeps `'unsafe-inline'`, `'unsafe-eval'`, `blob:`/`data:` and `https:`/`wss:` so the Blazor Server circuit, the Monaco editor (eval + blob workers), and embedded https content keep working. It only adds the *structural* directives ZAP wanted.

## Placement
Runs right after the `/healthz` short-circuit (probes stay header-free) and via `Response.OnStarting`, so headers land on static-file and error responses too.

## Not addressed (correctly reported by ZAP, not defects)
- **Proxy Disclosure** — the AKS ingress, working as intended
- **Dangerous JS Functions** — inside Monaco's own library
- **Backup File Disclosure** — false positive on SVG icon filenames (`/static/NodeTypeIcons/databasebackup.svg` etc.)

## Verification
- ✅ `dotnet build Memex.Portal.Shared` — 0 warnings, 0 errors (warnings-as-errors)
- After deploy: re-run the ZAP scan. The three enforced headers clear their findings immediately; the CSP finding clears after the Report-Only → enforced flip.

### DIFF
--- a/memex/Memex.Portal.Shared/MemexConfiguration.cs
+++ b/memex/Memex.Portal.Shared/MemexConfiguration.cs
@@ -1118,6 +1118,46 @@ public static class MemexConfiguration
             return next();
         });
 
+        // Security response headers — set once, early, for every real response (health
+        // probes already short-circuited above, so they stay header-free and minimal).
+        // These harden the browser surface WITHOUT changing what the app may load: the CSP
+        // is deliberately permissive — it keeps 'unsafe-inline'/'unsafe-eval', blob:/data:
+        // and https:/wss:, so the Blazor Server circuit, the Monaco editor (eval + blob
+        // workers) and embedded https content keep working. It only adds the structural
+        // directives an OWASP ZAP scan flagged as missing (a default-src fallback so every
+        // fetch directive is defined, object-src 'none', base-uri 'self', frame-ancestors
+        // 'self'). Tightening it further (per-response nonces, dropping 'unsafe-inline') is
+        // a separate hardening pass, not this change.
+        app.Use((ctx, next) =>
+        {
+            var headers = ctx.Response.Headers;
+            ctx.Response.OnStarting(() =>
+            {
+                headers["X-Content-Type-Options"] = "nosniff";
+                headers["Cross-Origin-Resource-Policy"] = "same-site";
+                headers["Permissions-Policy"] =
+                    "accelerometer=(), camera=(), geolocation=(), gyroscope=(), " +
+                    "magnetometer=(), microphone=(), payment=(), usb=()";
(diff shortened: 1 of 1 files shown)


rbuergi:
Re-running shard 1. `CrossProcessChangeFeedTest.AWriteInOneProcess_ReachesTheOtherProcessesLiveMirror_WithoutARecycle` failed with the *pre-fix* signature from #1929:

```
Expected value to be 2 because the mirror ADOPTS the durable version verbatim …
but found 3
```

i.e. the mirror ended at durable+1 — the #1432 shape the fix exists to eliminate.

**This PR's diff cannot reach it semantically.** One file, `MemexConfiguration.cs`, adding a response-header middleware (`app.Use` → `OnStarting` → set headers) to the HTTP request pipeline. `CrossProcessChangeFeedTest` builds hubs against a shared store and runs no HTTP pipeline at all. It *is* rebuilt by the change — `MeshWeaver.Hosting.Monolith.Test` references `Memex.Portal.Shared` — so it is not out of reach by dependency, only by behaviour.

Recording what I did **not** treat as evidence: the test passes locally on `main`, but in 2 s for 3 tests, which is far too fast for its 90 s bounds — the local run does not exercise the Postgres path CI uses, so it clears nothing. Re-running the shard is the control that actually discriminates.

If it fails on the same test again, that is a real intermittent in the #1929 fix and belongs to me, not to this PR.

PR #1992, opened 2026-08-21
A platform change can break the sixteen landed modules, and nothing here saw it (#1980)

### LINKED ISSUE #1980: No gate compiles the fourteen LANDED modules against a platform PR — a public-API deletion breaks them silently
Found while finishing #1576, whose platform half deletes a public API (`ExportDeclaration.Resolve`) that a landed module calls.

## The hole

`0b3b4266f` ("Fourteen modules leave the platform") moved fourteen compiled modules — `MeshWeaver.Markdown.Export`, `MeshWeaver.Observability`, the AI packs, … — into **MeshWeaver.Plugins/src/**, where they compile against the platform via `$(MeshWeaverRoot)` ProjectReferences.

The platform's cross-repo gate does **not** cover them. `dotnet-test.yml` → `plugin-gate` ("Plugins compile against this PR") stages, verbatim:

```yaml
```

…and runs `mw-plugin-test` over those **node repos** — i.e. the in-mesh NodeType `Source/*.cs` that the mesh compiles at runtime. It never builds MeshWeaver.Plugins' `src/` at all. Verified on run 32483032713, job 96775118860: it checks out `plugins-repo` at `main` and drives the tester; there is no `dotnet build` of any module project in it.

So the compiled half of every landed module is outside every gate in this repo. Its only compile check is MeshWeaver.Plugins' own CI, which builds against **`MeshWeaverRoot` = whatever platform ref that run resolves** — i.e. it discovers the break *after* the platform change is already on main.

## Why this is not theoretical

The plugins repo documents this exact failure with a body count:

> the pin predated MeshWeaver's `eafd353ed`, which deleted `AddTracking()`. Three NodeTypes kept calling it, CI stayed green against an image where the method still existed, and all three production portals hit `DynamicTypePreWarmer: REFUSING READINESS` at the next framework bump.

That was the in-mesh half, which at least has a gate. The compiled half now has none — and it fails harder: a module that does not build produces no bundle, and the capability simply disappears from the image.

**#1973 is a live instance.** It deletes `ExportDeclaration.Resolve`; MeshWeaver.Plugins' `MeshWeaver.Markdown.Export` calls it on `main` today. `plugin-gate` reports **pass**. The break is real, and the only thing preventing it is a hand-written merge-order note in the PR body (Plugins#553 first). A gate should be enforcing that, not a sentence.

## What would close it

A job that, for a platform PR, checks out MeshWeaver.Plugins and runs

```
dotnet build src/<each module>.csproj -c Release -warnaserror -p:MeshWeaverRoot=$GITHUB_WORKSPACE
```

against the PR's tree. Same shape as `plugin-gate` (same checkout, same credential, same fork guard), and it can reuse that job's existing `plugins-repo` checkout — so the marginal cost is the build, not the plumbing.

Design points worth deciding rather than defaulting:

- **Which modules.** `VITAL_MODULES` bought coverage-per-minute for the node repos; the compiled set is small enough (~14 projects) that "all of them" is probably affordable, but it should be measured before it lands in the PR critical path.
- **The reverse direction already exists** and is the precedent: MeshWeaver.Plugins pins `MW_PLATFORM_REF` and passes `-p:MeshWeaverRoot=…`. This issue is the other half of that pair.
- **The satellite repos** (`MeshWeaver.SocialMedia`, `.Reinsurance`, `.Manufacturing`, `.Education`) do not carry compiled modules today. If any gains one, it joins the same gate.
- Per this repo's own CI policy: no `if: vars.X != ''` on it, and the preflight must assert its inputs — a gate that skips reads exactly like a gate that passed.

rbuergi:
`plugin-gate` covers the **in-mesh** half of MeshWeaver.Plugins — the NodeType `Source/*.cs` the mesh Roslyn-compiles at runtime. It never builds that repo's `src/`, and since `0b3b4266f` ("Fourteen modules leave the platform") that is where **sixteen compiled modules** live: the AI packs, Observability, Markdown.Export, Mcp, Teams, the mail/blob/index adapters.

They reference this repo by **project path** (`$(MeshWeaverRoot)`) and take their NuGet pins from its **central `Directory.Packages.props`**. So a platform change reaches them directly — and their only check was the plugins repo's own CI, which builds against whatever platform ref *that* run resolves, i.e. after the break is already on `main`.

## Three instances of that coupling firing, all on 2026-08-21

| what broke | what happened |
|---|---|
| a **public API** — `eafd353ed` removed `AddTracking()` | three NodeTypes kept calling it, CI stayed green, and all three production portals hit `DynamicTypePreWarmer: REFUSING READINESS` at the next framework bump |
| a **package pin** — `Microsoft.Extensions.Options` left the central list | every landed module failed `NU1010`, on the plugins repo's `main` **and on each of its six open PRs at once**. This repo was green throughout; #1985 restored the pins reactively |
| a **test's `InternalsVisibleTo`** did not follow it out of this repo | `MeshWeaver.OgCard.Test` fails `CS1729`, reading as a signature change rather than lost visibility — plugins `main` red, fixed in #1995 |

A module that does not build produces no bundle, and the capability simply **vanishes from the image** — a failure with no error anywhere near the change that caused it.

## Non-vacuity: it reproduces the failures it is for

Deleting the `Microsoft.Extensions.Options` pin from this tree and restoring a landed module against it:

```
MeshWeaver.AI.Anthropic.csproj : error NU1010: The following PackageReference items do not
define a corresponding PackageVersion item: Microsoft.Extensions.Options.
```

And the gate's **first CI run caught the live `CS1729`** — the OgCard breakage — which is why it is currently red against a `main` that #1995 fixes.

## Design points, decided rather than defaulted

- **All 47 projects, not a vital subset.** `VITAL_MODULES` bought coverage-per-minute for the node repos, where each module is a separate tester run. Here it is **one `dotnet build` over a generated solution**, so the platform's project graph compiles once. **Measured: ~16 s for all 47** against a warm tree, versus 73 s for just the sixteen module projects built one invocation at a time. Against a 15-minute cap, and the step prints its own elapsed time so a re-scope would be made from evidence.
- **Test projects included** — not padding. The `CS1729` above *is* a test project; a module-only sweep is blind to that whole class. They are compiled here and never run: executing them needs containers and is the plugins repo's own job.
- **The set is discovered, never listed here.** A list would go stale the first time that repo adds a module — *silently*, the gate still passing while covering less. Zero projects discovered is a **failure**. So is a project that cannot be loaded.
- **Folded into `collect-results`**, the way `plugin-gate` / `doc-gate` / `clients-gate` are, rather than made a required context of its own.
- **Fork guard verbatim from `plugin-gate`.** A fork cannot read the private repo, and that is the one legitimate reason not to run — expressed as a fork check, never as an "is the credential set?" probe. Per this repo's CI policy, a missing input fails red in `preflight`; a gate that skips reads exactly like a gate that passed.

Two faults in the gate's own plumbing were found and fixed before it could mislead anyone: .NET 10 writes `.slnx`, not `.sln`; and `MeshWeaverRoot` has to cover the whole step, because `dotnet sln add` evaluates each project and GitHub's `bash -e` turns its first failure into a 4-second exit 2.

Closes #1980


rbuergi:
**Measured on CI now that the gate has actually executed: `all 47 landed projects built in 329s`** — 5.5 minutes, against the job's 15-minute cap.

That is materially more than the "~16 s for all 47" in the PR body, and the difference is real rather than noise: my measurement was against a **warm** local tree (labelled as such), while CI builds the platform's project graph cold. Recording the CI number here so nobody plans against the warm one.

It is still comfortably inside budget and runs concurrently with the test shards, so it should not extend a PR's wall clock — but "costs nothing" was too strong, and 5.5 min is the figure to re-scope against if this ever becomes the long pole. The step prints its own elapsed time on every run precisely so that judgement is made from evidence.

For the record, this gate needed three fixes to its own plumbing before it ever failed for a real reason:

1. .NET 10 writes `LandedModules.slnx`, not `.sln`;
2. `MeshWeaverRoot` has to cover the whole step, because `dotnet sln add` evaluates each project;
3. `ls A B` where one operand is missing exits **2** under GNU coreutils, which `set -e` + `pipefail` turns into a silent 4-second job failure — and that, not (2), was what produced the exit-2 runs. I attributed it to (2) first and was wrong.

Each of those looked exactly like the breakage the gate exists to catch. That is the argument for this repo's own rule applied one level in: a gate that cannot run must not look like a gate that passed — including when the reason it cannot run is its own plumbing.

It has already earned its place: its first real run caught the live `CS1729` from `MeshWeaver.OgCard.Test` losing its `InternalsVisibleTo`, which is why test projects are included rather than modules alone.

PR #2004, opened 2026-08-21
Gate the compiled half of every landed module against a platform PR (#1980)

### LINKED ISSUE #1980: No gate compiles the fourteen LANDED modules against a platform PR — a public-API deletion breaks them silently

rbuergi:
Closes #1980. `plugin-gate` covers the **node** repos (in-mesh `Source/*.cs`); it never builds MeshWeaver.Plugins' `src/`, where the landed modules compile against this repo via `$(MeshWeaverRoot)`. A public-API deletion therefore breaks them with every gate here green — held back only by a merge-order sentence in a PR body.

**This bit us twice today**: a pin prune turned nine module-bundle jobs red on the plugins trunk, and the deletion PR needed three follow-up fixes CI could only find after the fact.

- **All 28** landed modules, discovered rather than listed, so a new one is covered on arrival.
- **Finding zero projects fails the gate** — a gate that passes by finding nothing is the failure it exists to prevent.
- Reuses plugin-gate's checkout, credential and fork guard; the marginal cost is the build.

### DIFF
--- a/.github/workflows/dotnet-test.yml
+++ b/.github/workflows/dotnet-test.yml
@@ -1285,6 +1285,48 @@ jobs:
         path: plugins-repo
         ssh-key: ${{ secrets.PLUGINS_REPO_SSH_KEY }}
         fetch-depth: 1
+    # 🚨 THE COMPILED HALF of every landed module (#1980). The step below drives mw-plugin-test
+    # over the NODE repos — the in-mesh Source/*.cs the mesh compiles at runtime. It never builds
+    # MeshWeaver.Plugins' src/, where fourteen compiled modules live since they left the platform
+    # and where they compile against THIS repo through $(MeshWeaverRoot). So a public-API deletion
+    # broke them with every gate here green, and the only thing standing in its way was a
+    # merge-order sentence in a PR body. A module that does not build produces no bundle, and the
+    # capability simply disappears from the image — it fails harder than the in-mesh half, which
+    # at least had a gate.
+    #
+    # ALL of them, not a vital subset: the set is ~14 projects, they build in parallel off one
+    # restore, and a "coverage per minute" trade only makes sense where the list is long enough to
+    # need one. The reverse direction already exists (the plugins repo pins MW_PLATFORM_REF and
+    # passes -p:MeshWeaverRoot); this is the other half of that pair.
+    - name: "Landed modules compile against this PR"
+      env:
+        PLUGINS: ${{ github.workspace }}/plugins-repo
+      run: |
+        set -euo pipefail
+        mapfile -t projects < <(find "$PLUGINS/src" -maxdepth 2 -name 'MeshWeaver.*.csproj' \
+          -not -name '*.Test.csproj' | sort)
+        if [ "${#projects[@]}" -eq 0 ]; then
+          echo "::error::no compiled modules found under $PLUGINS/src — the gate would pass by "\
+               "finding nothing, which is exactly the failure it exists to prevent."
+          exit 1
+        fi
+        echo "Building ${#projects[@]} landed module(s) against this PR:"
+        printf '  %s\n' "${projects[@]##*/}"
+        failed=()
+        for project in "${projects[@]}"; do
+          name="${project##*/}"; name="${name%.csproj}"
+          if ! dotnet build "$project" -c Release -warnaserror \
+               -p:MeshWeaverRoot="$GITHUB_WORKSPACE" --nologo -v q; then
+            failed+=("$name")
+          fi
+        done
+        if [ "${#failed[@]}" -gt 0 ]; then
+          echo "::error::${#failed[@]} landed module(s) do NOT compile against this PR: ${failed[*]}."\
+               " Land the plugins-repo change first, or keep the API these modules call."
+          exit 1
+        fi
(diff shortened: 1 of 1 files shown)


rbuergi:
Closing as a duplicate of #1992, which was opened for the same issue about an hour earlier and covers strictly more. Nothing wrong with this implementation — we simply built it twice, and keeping both would mean two gates doing one job in the same file.

What #1992 has that this one does not, and why each mattered today:

- **Test projects are included.** This is the decisive one. `main`'s live failure right now *is* a test project — `MeshWeaver.OgCard.Test` moved out of the platform, silently lost its `InternalsVisibleTo`, and fails `CS1729`. A `-not -name '*.Test.csproj'` sweep is blind to that whole class, and #1992's first CI run caught it.
- **One `dotnet build` over a generated solution, not one per project.** The platform's project graph then compiles once. Measured: **~16 s for all 47 projects** in one build, versus **73 s for just the sixteen modules** built one invocation at a time — which is what this loop does.
- **A separate job, not a step inside `plugin-gate`.** That gate carries `timeout-minutes: 12` and a documented "if it ever becomes the long pole it should fail and be re-scoped, not silently stretch every PR". Adding a build to it spends that budget; a sibling job runs concurrently and keeps the two verdicts distinct.
- **Its own `outputs.failure`, folded into `collect-results`,** so the required check names the real cause rather than pointing at the plugin gate's verdict line.

Two faults in #1992's own plumbing had to be fixed before it was trustworthy, both worth knowing if this approach is ever revisited: .NET 10 writes `LandedModules.slnx`, not `.sln`; and `MeshWeaverRoot` must cover the whole step, because `dotnet sln add` evaluates each project and GitHub's `bash -e` turns its first failure into a 4-second `exit 2` that looks nothing like a compile break. This PR's per-project loop sidesteps the second by never calling `dotnet sln add` — a real advantage of the loop, just not enough to outweigh 4.5× the wall clock and the missing test coverage.

One thing from here worth keeping: reporting **every** failing module rather than the first. #1992 does this through the solution build's own diagnostics — it dedupes and prints up to 20 in the job summary — so the behaviour is equivalent.

PR #2007, opened 2026-08-21
The mesh compiled identical content into a different assembly on every process

rbuergi:
Lifted out of #1994 so it can land on its own — it is an independent bug fix with consequences well beyond the content key it was found by, and it should not wait on that PR's open decision.

## What was wrong

`CollectCompileSources` kept **"the order the snapshot delivered them"**. On the mesh path that delivery is `map.Values.ToList()` over an `ImmutableDictionary` — hash-bucket order over **per-process randomised string hashes**. The join order is in the emitted bytes.

So the mesh compiled identical content into a **different assembly on every process** — not reproducible against the compiler-driven producer, and not against itself:

| run | mesh | compiler |
|---|---|---|
| 1 | `ic920384a…` | `i3494e62e073f2a551796fa08440330ed` |
| 2 | `iead1b585…` | `i3494e62e073f2a551796fa08440330ed` |
| 3 | `i70db762d…` | `i3494e62e073f2a551796fa08440330ed` |

Now ordinal by path, inside the toolchain function that already owns the join order, so both producers agree by construction.

## What this explains

- **Cross-replica dedupe in `FileSystemAssemblyStore` was structurally impossible.** `PutWithLocation` is first-write-wins on a content-hashed file name, and its own comment reads *"The bytes are identical by construction — the hash IS the name."* With per-process source order they were not. Two silos compiling the same `(path, version)` produced different hashes, so different names — never a collision, but **never a share either**. Every replica wrote its own generation into `/data/assembly-cache`, which is retention pressure `AssemblyCacheGenerations` has been sweeping ever since. That is now genuinely dedupable and worth measuring.
- **The store's digest-of-emitted-bytes could never dedupe**, for the same reason.
- **`BakeEquivalenceTest`'s class remarks are now stale.** They document the non-determinism as a fact of life and explain why the test compares the type/member *surface* rather than bytes. That constraint is largely lifted, so byte equality between the two producers is within reach — which would prove the bake produces what the mesh would, rather than something merely surface-compatible with it. Not attempted here.

## Deploy note

It changes emitted bytes for every NodeType with **no** identity change. That is safe: existing cached DLLs stay valid for their content, and recompiles simply write new content-hashed names.

## Verification

`CompileSourceOrderTest` — 2/2, verified non-vacuous by reverting the fix. Cherry-picks onto `main` cleanly.

### DIFF
--- a/src/MeshWeaver.Compiler/NodeCompileShaping.cs
+++ b/src/MeshWeaver.Compiler/NodeCompileShaping.cs
@@ -210,12 +210,29 @@ public static class NodeCompileShaping
 
     /// <summary>
     /// The emit path's source-set fold: dedup by path (case-insensitive), keep only
-    /// non-executable Code nodes with non-blank code, in the order the snapshot delivered them.
+    /// non-executable Code nodes with non-blank code, ORDERED BY PATH (ordinal).
     /// Executable scripts run via the kernel (ExecuteScriptRequest), never folded into the parent
     /// NodeType's Roslyn unit — top-level statements would collide with class declarations from
     /// Source/ siblings; Test/ commonly mixes both shapes, and this filter lets them coexist.
     /// Returns the matched configurations together with the matched paths (activity-log
     /// material for the caller).
+    ///
+    /// <para>🚨 <b>The order is a function of the SOURCE SET, never of how it was delivered</b>
+    /// (#1707 slice 4). It used to be "the order the snapshot delivered them", and on the mesh path
+    /// that delivery is <c>ImmutableDictionary.Values</c> — hash-bucket order over string hashes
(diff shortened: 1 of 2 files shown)


rbuergi:
Closing as included: this was a cherry-pick of `c9eccc76a` out of #1994, opened while that PR still looked blocked on the full-MVID decision. That decision is settled — the rule stays, with the reason pinned as a test — so #1994 merged whole and carries this commit. Its analysis (the `FileSystemAssemblyStore` dedupe consequence, the stale `BakeEquivalenceTest` remarks, and determinism as a silent precondition of the content key) is recorded on #1994 so it lives with the commit that landed.

Issue #1974, opened 2026-08-21
Flip the Blazor view packs onto modules/ — the static-asset provider shipped and has no consumer


rbuergi:
Follow-up to #1724 (the static-asset provider, shipped in #1775) and #1644 step 2 (the flip ledger), both now closed.

## Why this exists

#1724 built the provider that unblocks the view packs — and then nothing flipped. So the provider's
loop is **empty in production**: no `MeshModule` view pack has left the host build graph, therefore
no `modules/<Name>/wwwroot` exists, therefore the mount/negotiation code has never served a byte
outside its own tests. `App.razor:42` says so in a comment. A provider with no consumer is a
provider we will discover is wrong at the worst moment.

Current state (`memex/MeshModulesPublish.targets:40-106` is the one inventory):

| pack | `MeshModule` | host `ProjectReference` | still blocked on |
|---|---|---|---|
| `MeshWeaver.Blazor.Radzen` | ✅ | `Memex.Portal.Shared.csproj:13` | **nothing** — package assets only |
| `MeshWeaver.Blazor.GoogleMaps` | ✅ | `:15` | **nothing** — collocated JS + scoped CSS, both handled |
| `MeshWeaver.Blazor.Analysis` | ✅ | `:14` | **nothing** — scoped CSS, handled |
| `MeshWeaver.Blazor.AppleMaps` | ❌ | `:20` | not even listed as a module |
| `MeshWeaver.Blazor.OpenStreetMap` | ❌ | `:19` | not even listed as a module |

The three listed ones are the flip; the two unlisted ones are the smaller gap #1724 flagged and
nobody picked up.

## The work

1. **Flip Radzen first** — package assets only (`_content/Radzen.Blazor/...`, self-loaded from
   `RadzenViewBase.OnAfterRenderAsync`), which is the mount path with no scoped-CSS or collocated-JS
   variable in it. Move its entry from `@(MeshModule)` to `@(MeshModuleClosure)`, drop the
   `ProjectReference`, and verify in a **published** container, not in dev — per #1680 the `.gz`/`.br`
   endpoint variants exist only there, and `MeshModuleStaticAssetExtensions` negotiates against them.
2. **Then GoogleMaps and Analysis**, which additionally exercise the scoped-CSS bundle
   (`{name}.styles.css` → `ModuleStaticAssetManifest.Stylesheets` → `App.razor:43-46`) and
   collocated `.razor.js`.
[...]

PR #2021, opened 2026-08-21
Flip the remaining four view packs onto the closure lane (#1974)

rbuergi:
**This completes the view-pack set.** With #2018 (Analysis) already merged, all five now ship via `modules/<Name>/` and none is referenced by `Memex.Portal.Shared`.

> Originally opened for AppleMaps + OpenStreetMap alone; #2022 (Radzen + GoogleMaps) merged into this branch, so this single PR to `main` carries all four.

## AppleMaps and OpenStreetMap were never modules at all

They rode `Memex.Portal.Shared`'s `ProjectReference` into the app closure while appearing under **no** `Modules:Assemblies` list, and no portal code names their types. Attribute discovery is by explicit path (`MeshBuilder.InstallAssemblies`), not a scan of whatever is loaded — so nothing was ever going to pick them up. Their bits shipped in every image **permanently inert**.

**This changes only where the bits live.** No pack is added to `Modules:Assemblies` here, so each stays exactly as active (or inactive) as it is today. Turning the two map packs on is a separate call — AppleMaps needs a MapKit token; OpenStreetMap needs no key.

## Radzen is the case the earlier flips could not answer

It self-loads `_content/Radzen.Blazor/{css/material-base.css, Radzen.Blazor.js}` from `RadzenViewBase`, and its chart and pivot views link a theme stylesheet from the same namespace. Those are **package** assets, not the pack's own.

Dropping the portal's reference is precisely what stops the **host** serving that namespace, so the module's own `wwwroot/_content/Radzen.Blazor/` answers instead — the same-identity prune working from the other side. Verified: **36 files** land there, including both self-loaded files and every theme; the host's copy is gone. `App.razor` carries no pack tags, which is what makes this work.

## Verified on real publishes

For all five packs: every DLL **absent** from the app publish root, every pack **dropped** from the host's `<App>.styles.css`, and assets under `modules/<Name>/wwwroot` at the paths the code actually requests — `AppleMapView.razor.js`, `OpenStreetMapView.razor.js`, and leaflet's `css`/`js` plus the marker images its css resolves beside itself.

## Also

Anchors all four packs in `Memex.Portal.Shared.Test`, which asserts on their module attributes (`ViewPackModuleTest`, `MapProviderModuleTest`) and reached them through the portal until now. Worth recording: **a publish cannot verify a flip** — it has nothing to say about a compile-time reference the tests still need, which is how #2018 went red after a green publish.

Two observations, neither acted on here:

- **Three of five** packs (Radzen, AppleMaps, OpenStreetMap) have no `.razor.css` of their own, so their aggregate is 211 bytes of nothing but dependency imports — that is what #2020 stops linking.
- `MeshWeaver.Blazor.Radzen/wwwroot/css/radzen-grid.css` is referenced by nothing in the repo. Harmless under the module's namespace; removing it is a separate call.

## MeshWeaver.Maps stays in core

Correcting an earlier plan to move it with the map packs: it has consumers of its own (`MeshWeaver.Maui`, `PluginTester`, five test projects) and is named in `FrameworkBuildIdentity`, so node source compiles against it.

### DIFF
--- a/memex/Memex.Portal.Shared/Memex.Portal.Shared.csproj
+++ b/memex/Memex.Portal.Shared/Memex.Portal.Shared.csproj
@@ -21,8 +21,14 @@
     <!-- MeshWeaver.OgCard is deliberately NOT referenced (#1644 step 2): it ships via the
          modules/<Name>/ closure layout in MeshModulesPublish.targets and activates via
          Modules:Assemblies -> OgCardModuleAttribute. -->
-    <ProjectReference Include="..\..\src\MeshWeaver.Blazor.OpenStreetMap\MeshWeaver.Blazor.OpenStreetMap.csproj" />
-    <ProjectReference Include="..\..\src\MeshWeaver.Blazor.AppleMaps\MeshWeaver.Blazor.AppleMaps.csproj" />
+    <!-- MeshWeaver.Blazor.OpenStreetMap and .AppleMaps are deliberately NOT referenced (#1974):
+         they ship via the modules/<Name>/ closure layout in MeshModulesPublish.targets. Both were
+         ships-the-bits references that shipped bits nothing could reach — no portal code names
+         their types AND neither DLL is listed under Modules:Assemblies, so until now they rode
+         the app closure permanently inert. On the closure lane the bits sit in modules/ and a
+         deployment turns each on by listing its DLL, which is what that switch is for. Their
+         collocated JS and OpenStreetMap's vendored wwwroot/leaflet/ are already addressed as
+         ./_content/<Pack>/... — exactly what the module static-asset provider serves (#1724). -->
     <ProjectReference Include="..\..\src\MeshWeaver.Blazor.Graph\MeshWeaver.Blazor.Graph.csproj" />
     <ProjectReference Include="..\..\src\MeshWeaver.Blazor.Portal\MeshWeaver.Blazor.Portal.csproj" />
     <ProjectReference Include="..\..\src\MeshWeaver.Hosting.Blazor\MeshWeaver.Hosting.Blazor.csproj" />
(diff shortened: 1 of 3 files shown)


rbuergi:
## Pushed a fix for a real defect this PR had — the framework identity

Shard 2 was **not** a flake, and it is worth recording how close it came to being waved through.

Flipping the map packs took **`MeshWeaver.Maps` out of `Memex.Portal.Shared`'s compile closure** without anyone naming it: the portal never referenced Maps directly — it arrived **transitively**, because AppleMaps, GoogleMaps and OpenStreetMap each reference it.

`MeshWeaver.Maps` is in `FrameworkBuildIdentity.ContentSurfaceAssemblies`, and losing a canonical assembly from a host's compile closure **does not fail a build**. The surface manifest is written from `@(ReferencePathWithRefAssemblies)`, so the line simply vanishes: this host hashes Maps as absent while `mw-plugin-test` hashes a real surface id, the two hosts of one commit resolve **different framework identities**, and every bake lands at an address no portal reads — publication green, pods compiling everything at boot, every cover serving a compilation-fallback card.

That is the 2026-08-17 outage (#1814). `CanonicalContentSurface_IsRecordedByEverySurfaceManifestHost` is the pin added so it could not recur, and it did its job here.

**Fix:** a direct `ProjectReference` to `MeshWeaver.Maps` from `Memex.Portal.Shared`, with the reason written on it — no portal code uses those types, so a future "remove unused reference" tidy would re-break it. Verified red without the line, green with it (23/23 in `FrameworkBuildIdentityTest`).

### On the other shard

Shard 3's `MeshWeaver.AI.Test.ThreadTokenUsageTest` is a different matter: it passes 10/10 locally on this exact branch state with all five pack DLLs confirmed absent from that test's bin, and it sits in the same assembly as the flake filed as #2025. I checked that shard first and was reasoning toward "unrelated" — the real defect was in the shard I had not read yet. Reading every failing shard before clearing any of them is the lesson.

PR #2027, opened 2026-08-21
Portal security headers: cover /api, /login, robots (outermost middleware)

rbuergi:
Follow-up to #1988.

## Problem
The post-deploy ZAP rescan of `memex.meshweaver.cloud` showed the security headers landed on content pages but **not** on `/api/`, `/login`, `/robots.txt` — those are mapped in the **Distributed portal host** outside `StartMemexApplication`'s web pipeline, where the middleware from #1988 lived, so they slipped through:

```
X-Content-Type-Options Header Missing [10021] x3 → /api/, /login, /robots.txt
Permissions Policy Header Not Set     [10063] x2 → /api/, /login
```

## Fix
- Extract the header logic into an **idempotent** `WebApplication.UseMemexSecurityHeaders()` — header values are SET (not appended), and the Report-Only CSP is written only when absent, so it is safe to call twice on one host.
- Keep the call in the web pipeline (`StartMemexApplication`) for content pages.
- **Also call it as the OUTERMOST middleware** of the deployed host (`Memex.Portal.Distributed/Program.cs`, right after `Build()`), so it wraps every response regardless of internal routing.

## Verification
- ✅ `dotnet build Memex.Portal.Distributed` — 0 warnings, 0 errors
- After deploy: re-run the baseline ZAP scan; 10021 and 10063 should drop to 0.

## Still deliberately out of scope
- CSP `[10055]` clears at the Report-Only → enforced flip (separate one-liner).
- Site-isolation `[90004]` needs COEP/COOP; COEP `require-corp` is high-breakage (embeds/workers), so not added reflexively.

### DIFF
--- a/memex/Memex.Portal.Shared/MemexConfiguration.cs
+++ b/memex/Memex.Portal.Shared/MemexConfiguration.cs
@@ -1065,6 +1065,54 @@ public static class MemexConfiguration
     /// Starts the Memex portal application with the specified App component type.
     /// Pattern taken from MeshWeaver.Portal's StartPortalApplication.
     /// </summary>
+    /// <summary>
+    /// Sets the platform's security response headers on every response, via
+    /// <see cref="Microsoft.AspNetCore.Http.HttpResponse.OnStarting(System.Func{System.Threading.Tasks.Task})"/>
+    /// so they also land on static-file, error and redirect responses. Idempotent: the header
+    /// values are SET (not appended) and the Report-Only CSP is only written when absent, so
+    /// calling this twice on one host (the web pipeline in <see cref="StartMemexApplication{TApp}"/>
+    /// plus the deployed host's outermost layer) is safe and keeps endpoints mapped outside the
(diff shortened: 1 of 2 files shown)


rbuergi:
Closing as redundant. A settled ZAP baseline rescan of memex.meshweaver.cloud (after the #1988 roll fully completed) shows `X-Content-Type-Options` [10021] and `Permissions-Policy` [10063] **PASS on every URL, including `/api/`, `/login`, `/robots.txt`** — so #1988's middleware already covers them. The "missing on those paths" finding that motivated this PR was a roll-in-progress artifact: the rescan ran ~13 min into the roll and ZAP's requests hit the still-draining old pod (no headers) via session affinity, while `/` hit the new pod. ASP.NET middleware wraps the whole request pipeline regardless of where an endpoint is *mapped*, so there was no real gap. The outermost-placement idea is harmless defense-in-depth but not worth a core change for the observed surface; reopen if we later want it for endpoints ZAP can't reach.

PR #2030, opened 2026-08-21
fix(#1786): retire six NodeTypes that type nothing and cannot compile, and guard the shape

rbuergi:
Closes the remaining half of #1786. **Measured on prod first** — of the 13 NodeTypes that issue reported parked on memex.meshweaver.cloud, **7 are already green and 6 are still Error today**; this fixes those 6 plus 3 latent duplicates, and adds the guard that would have caught all nine at authoring time.

## What is still broken, verified live today

`GetDiagnostics` against `memex.meshweaver.cloud` (serving `3.0.0-rc6.ci.4780`):

| the original 13 | today |
|---|---|
| `FutuRe/{AmericasIns,EuropeRe,AsiaRe}/LineOfBusiness` | **Error** ❌ |
| `FutuRe/{AmericasIns,EuropeRe,AsiaRe}/TransactionMapping` | **Error** ❌ |
| `Doc/DataMesh/SocialMedia/{Post,Profile}`, `SocialMedia/{Post,Profile}`, `Doc/Architecture/BusinessRules/Cession` | Ok ✅ |
| `Northwind/{Product,AnalyticsCatalog}` | node no longer exists ✅ |

## The six: copy-paste NodeTypes that type nothing

Each carries the *identical* configuration lambda as its parent but none of the parent's `Source/*.cs`:

```
CS0246 The type or namespace name 'LineOfBusiness' could not be found
--- Source discovery ---
  - namespace:…/AmericasIns/LineOfBusiness/Source scope:subtree nodeType:Code
Matched Code nodes (0):
  (none) — the configuration lambda cannot reference types because no source files were included.
```

The parent `FutuRe/LineOfBusiness` compiles **Ok** on the same mesh, because it owns its `Source/`. That contrast is the whole diagnosis.

And **nothing is typed by any of the six** — every instance beneath them names the *parent*:

| folder | instances | their `nodeType` |
|---|---|---|
| `AmericasIns/LineOfBusiness` | 8 | `FutuRe/LineOfBusiness` |
| `EuropeRe/LineOfBusiness` | 8 | `FutuRe/LineOfBusiness` |
| `AmericasIns/TransactionMapping` | 14 | `FutuRe/TransactionMapping` |
| `EuropeRe/TransactionMapping` | 13 | `FutuRe/TransactionMapping` |
| `AsiaRe/*` | 0 | — |

A repo-wide scan for `"nodeType": "<any of the six>"` returns **0 files**. So they were pure dead weight that could only ever park.

They stay as nodes — the folder must exist for its children, and the governance prose is worth keeping — but as ordinary **`Markdown`** nodes, with the description as the body. No compile, nothing lost.

> Why not a `sources` list pointing at the parent? `sources` entries are absolute mesh paths, and this tree does not mount at one fixed root: `FutuRe` is its own partition via `AddFutuRe()` locally, but on memex-cloud the same nodes live under `MeshWeaver/samples/Graph/Data/FutuRe/…`. An absolute entry would be correct in one deployment and dead in the other, and `$self` cannot address a parent's sibling. Since the types are unused, not shipping them is the honest fix.

## The three duplicates

`samples/Graph/Data/Doc/` authors ten files and **all ten duplicate `src/MeshWeaver.Documentation/Data/`**. For the three NodeTypes among them the samples copies are strictly inferior — same node path, same lambda, but **no `Source/` and no instances**, while the documentation copies have both. On a deployment importing both, the source-less copy can only clobber a good node; that is a plausible origin for the SocialMedia/Cession third of #1786 in the first place.

Removed. The `CollaborativeEditing` fixture in the same folder — the one the `MeshWeaver.Content.Test` suite actually loads via `AddDoc()` — is untouched. `CessionLayoutAreaTest` loads `AddDocumentation()`, not this tree; the comment in it claiming otherwise is stale.

## The guard

`ShippedNodeTypeSourceResolutionTest` — a shipped NodeType must not name a content type that neither its own `Source/` folder nor the framework declares.

`WithContentType<T>()` is the one reference a NodeType always makes explicitly, so this is checkable from the filesystem in **~0.4 s**. The real compile only runs on a deployment that imports the tree, and it does not fail loudly there: `CompileWatcher` **parks** the type and serves the cached error without retrying, so every page backed by it is broken while CI stays green. That is exactly why #1786 describes a *standing condition nothing was ticketing*.

Scope is stated honestly in the test: it does not type-check the whole lambda, and it skips a NodeType with an explicit `sources` list (mesh queries are not filesystem-resolvable). It is a floor, not a compiler. It excludes `src/MeshWeaver.Documentation/Data` from the "framework" set — that is a node repo that happens to live under `src/`, and its per-NodeType Source folders would otherwise vouch for types no framework assembly exports.

## Non-vacuity — proven, not asserted

Reverting **only** the data half and re-running the compiled test fails, naming all nine:

```
Failed!  - Failed: 1, Passed: 1
  A shipped NodeType names a content type that neither its own Source/ folder nor the
  framework declares … Offending files:
  samples/Graph/Data/FutuRe/EuropeRe/LineOfBusiness.json: LineOfBusiness
  samples/Graph/Data/FutuRe/EuropeRe/TransactionMapping.json: TransactionMapping
  samples/Graph/Data/FutuRe/AsiaRe/LineOfBusiness.json: LineOfBusiness
  samples/Graph/Data/FutuRe/AsiaRe/TransactionMapping.json: TransactionMapping
  samples/Graph/Data/FutuRe/AmericasIns/LineOfBusiness.json: LineOfBusiness
  samples/Graph/Data/FutuRe/AmericasIns/TransactionMapping.json: TransactionMapping
  samples/Graph/Data/Doc/DataMesh/SocialMedia/Post.json: SocialMediaPost
  samples/Graph/Data/Doc/DataMesh/SocialMedia/Profile.json: SocialMediaProfile
  samples/Graph/Data/Doc/Architecture/BusinessRules/Cession.json: CessionData
```

A second test pins both scans against vacuity — the framework scan must find the `src/` tree, and `LineOfBusiness`/`TransactionMapping` must **not** be mistaken for framework types, or the invariant would wave through the exact #1786 shape.

## Tests executed

| suite | result |
|---|---|
| `MeshWeaver.Graph.Test` | **1493 passed**, 0 failed |
| `MeshWeaver.Content.Test` (loads the `Doc` tree) | **274 passed**, 0 failed, 1 skipped |
| `CessionLayoutAreaTest` | **6 passed** |

Branched off `origin/main` at `46a4378f9`.

## Note for whoever deploys

Merging does not clear the six on memex-cloud by itself — the mesh keeps what the source tree drops, so the six parked NodeType nodes will still be there after the sync until they are retyped or removed on the instance. That is a live-mesh write, so I have not made it.


rbuergi:
Follow-up push: merged `origin/main` (branch was 183 commits behind; clean merge, `git diff origin/main...HEAD --stat` shows only the intended 10 files) and fixed the one CI failure, which was caused by this PR: `EuropeRe_LineOfBusiness_Search_ShouldReturn8LoBs` pinned the retired node's OLD NodeType presentation (Splitter admin shell + NodeType-mode search). The node is now Markdown, so `MeshNodeLayoutAreas.Search` takes the instance-catalog branch (breadcrumbs Stack + MeshSearchControl); the test now asserts that shape while keeping the original regression guard — catalog scoped to `namespace:FutuRe/EuropeRe/LineOfBusiness`, exactly 8 LoBs, each typed by the group `FutuRe/LineOfBusiness`, no Analysis/TransactionMapping siblings. Also adds the What's New entry (Category: Fix). Verified locally with CI's flags: FutuRe.Test 59/59, guard 2/2, WhatsNew+link integrity 2/2, all `-c Release -warnaserror` clean.

Issue #1754, opened 2026-08-17
Self-update must refuse to roll the platform until every DEPLOYED plugin has a build available for the target release


rbuergi:
Maintainer-directed:

> *"in deployment you must check that all plugins which are deployed are actually available in current release before proceeding with platform update"*

## The gap

`SelfUpdateHostedService` polls, `VersionSelect.IsNewer` decides a candidate is newer, and
`KubernetesDeploymentUpdater` rolls it. **Nothing asks whether the instance's installed plugins can
survive the target version.** The update proceeds on platform-version comparison alone.

That is the wrong precondition, and today's roll demonstrates why. Moving prod to
`3.0.0-rc4.ci.4049` changed the framework identity to
`s377941f549f721e01ac764e0fb8db84a`. Every satellite's previously published content bake was keyed
to the OLD identity, so until all four repos re-baked and re-published, the portal had nothing to
adopt for that content. We got the ordering right by hand — the satellites were re-baked BEFORE the
portals were rolled, and the payoff was measurable (`compiled=0 alreadyBaked=84`, warm-up 32.1 s).
Had we rolled first, the same portal would have Roslyn-compiled its way through boot, and any type
that failed to compile would have parked its hub.

Nothing in the product enforces that ordering. A 6-hourly self-update poll can fire it at any time,
unattended.

## What "available for the target release" means — two distinct things

1. **Content packages** — a published, SEALED bake under the TARGET release's framework identity
   (`prebuilt-bundles/<identity>/<source>/` with its `_complete` sentinel). Absent ⇒ the instance
   recompiles that content at boot, which is exactly the regression the CI-bake lane exists to
   prevent (#1347 → #1660).
2. **Compiled modules** — a build whose `MinMeshVersion` floor is satisfied by the target version
   (the floor gate, MVID diagnostic-only since #1673). A module whose floor EXCEEDS the target, or
   which has no build at all, cannot land.

Both are checkable BEFORE rolling: the identity of a candidate image is resolvable, and the
registry already advertises what it can serve per identity (`/api/plugins/bundles/index.json`
carries a top-level `frameworkMvid`).

## The gate

Before `KubernetesDeploymentUpdater` rolls, enumerate what THIS instance has installed and assert
each one has a usable artifact for the target release. If any does not, **do not update** — hold at
the current version, say precisely which plugin blocks it and why, and surface it where
`PlatformUpdateStatus` / the Updates settings tab already report update state. A held update is a
[...]

PR #2039, opened 2026-08-22
fix(#1754): an UNWIRED release-availability gate is a hold, not a pass

rbuergi:
Continues #1754.

## Still real? Mostly shipped — one hole left, and it is the one this cluster is about

The gate landed on `main` (`8c8cc2005` and follow-ups) and is genuinely good: `ReleaseAvailability.IsUpdatable` is pure and fails safe, `ReleaseAvailabilityService` times out into `Indeterminate` rather than hanging, and all three roll paths read one verdict — the poller, the Updates tab's manual button, `/api/plugins/is-updatable`, plus CD's post-promote `check-release-availability.sh` (unconditional, no `if:`, fails red).

What was left: **both roll paths treated "the gate is not registered on this host" as clearance to proceed.**

```csharp
// SelfUpdateHostedService.GateThenApply — before
if (gate is null)
    _logger?.LogInformation("… rolling {Tag} without checking …");
    return Apply(target);

// UpdatePolicySettingsTab — before
var decision = gate is null
    ? Observable.Return(UpdatabilityVerdict.NotEnforced("no release-availability gate is registered"))
    : gate.IsUpdatable(tag);           // NotEnforced is IsUpdatable: true
```

So the one host where nothing checked anything was also the one host that never refused, and the only trace was an Information line nobody reads. That is the trap: *a gate that cannot run must never look like a gate that passed.*

`NotEnforced` is the **single stated applicability exemption** — a deployment that consumes no CI bakes, and therefore already compiles at every boot, which holding could only freeze forever. Reusing it for a **wiring failure** is precisely how the two become indistinguishable, and it is the same shape as `if: vars.X != ''` skipping green.

Reachable today? Only by a host that registers `SelfUpdateHostedService` without `AddSelfUpdate` — which registers both. So this is a latent hazard rather than a live incident, and it is exactly the kind that surfaces later as "why has this environment not updated / why did this environment roll".

## What changed

- **`UpdatabilityVerdict.Unavailable(reason)`** — a HOLD, reported `IsIndeterminate`, so every surface that already separates *"I could not look"* from *"I looked and it is incompatible"* keeps doing so unchanged. `NotEnforcedReason` stays null; that field is what makes a verdict updatable.
- **The poller holds**, records the hold on `Admin/UpdatePolicy` where the Updates tab reads it, and logs at **Error** rather than Information. Re-evaluated every tick like every other hold, so registering the gate clears it with nothing to un-stick by hand — that is what makes refusing safe rather than a freeze.
- **The manual roll button holds too** — it is exactly the moment an operator, seeing an update that never applied, would force the roll.
- **`SelfUpdate:AllowUnverifiedRoll`** is the deliberate escape hatch, in the shape this repo already uses for `PreWarm:AllowUnprovenBake`: roll unverified *on purpose*, in configuration where it is visible, saying so at Warning on every tick. It can never waive a gate that **did** run.

## Tests

```
Passed!  - Failed: 0, Passed: 13, Skipped: 0, Total: 13  (ReleaseAvailabilityTest)
Passed!  - Failed: 0, Passed:  4, Skipped: 0, Total:  4  (SelfUpdateAvailabilityGateTest, real monolith mesh)
```

New:
- `Unavailable_IsAHold_AndReadsAsAnAvailabilityFailure_NotAnIncompatibility`
- `AnUnwiredGate_HoldsTheRoll_AndSaysItCouldNotLook` — the fail-closed proof. The assertion is the hold **written** where the Updates tab reads it, flagged indeterminate, never "no patch happened" (which would pass against a poller that had simply died).
- `AnUnwiredGate_RollsOnlyWhenTheOperatorOptedInDeliberately`

**Non-vacuity** — `GateThenApply` reverted to the pre-fix branch:

```
Failed!  - Failed: 1, Passed: 3, Skipped: 0, Total: 4

  Failed AnUnwiredGate_HoldsTheRoll_AndSaysItCouldNotLook [30 s]
```

(The hold never lands, because the poller rolled.)

## For the maintainer

One judgement call worth confirming: **the default is now HOLD.** A host that wires the poller by hand without `AddSelfUpdate` stops auto-updating until it either registers `ReleaseAvailabilityService` or sets `SelfUpdate:AllowUnverifiedRoll`. Every in-tree host goes through `AddSelfUpdate`, which registers the gate unconditionally, so nothing in this repo changes behaviour — but a downstream host might, and it will find out via an Error log plus a named hold on the Updates tab rather than silently.

### DIFF
--- a/memex/Memex.Portal.Shared/SelfUpdate/ReleaseAvailabilityService.cs
+++ b/memex/Memex.Portal.Shared/SelfUpdate/ReleaseAvailabilityService.cs
@@ -64,6 +64,30 @@ public class ReleaseAvailabilityService(
     public string? PublishedRoot => configuration[ShippedPrebuiltBundles.PublishedRootConfigKey];
 
     /// <summary>
+    /// 🚨 <b>Does this gate APPLY to this deployment at all?</b> Returns the reason it does not, or
+    /// <c>null</c> when it does.
+    ///
+    /// <para>Decided from CONFIGURATION alone, which is what makes it answerable by a caller that
+    /// has no instance of this service — and that is the whole point. "Cannot verify" and "verified
+    /// as nothing to verify" are different states, and only the first may hold. A deployment that
+    /// consumes no CI bakes already compiles its content at every boot, so a registered gate would
+    /// answer <see cref="UpdatabilityVerdict.NotEnforced"/> for it; the gate being ABSENT on such a
+    /// deployment tells you nothing new, and holding on it would freeze an environment the gate was
+    /// never going to protect.</para>
+    ///
+    /// <para>Static and shared on purpose: the poller's unwired path (#1754) and
+    /// <see cref="IsUpdatable"/> must reach the same applicability answer, and a rule that only one
+    /// caller honours is not a rule.</para>
+    /// </summary>
+    /// <param name="configuration">The host's configuration; null reads as "nothing configured".</param>
+    public static string? NotApplicableReason(IConfiguration? configuration) =>
+        string.IsNullOrWhiteSpace(configuration?[ShippedPrebuiltBundles.PublishedRootConfigKey])
+            ? $"this deployment consumes no CI bakes ({ShippedPrebuiltBundles.PublishedRootConfigKey} "
+              + "is not configured), so it already compiles its content at every boot — the "
+              + "release-availability gate has nothing to enforce here"
+            : null;
+
+    /// <summary>
     /// Is <paramref name="targetVersion"/> a release this environment may be rolled to? Cold —
     /// the file-system leaves run on the <see cref="IoPoolNames.FileSystem"/> pool, never on a hub
     /// action block — and total: every failure resolves to a HOLD carrying its reason, so a
@@ -72,12 +96,11 @@ public class ReleaseAvailabilityService(
     public virtual IObservable<UpdatabilityVerdict> IsUpdatable(string? targetVersion) =>
         Observable.Defer(() =>
         {
-            var publishedRoot = PublishedRoot;
-            if (string.IsNullOrWhiteSpace(publishedRoot))
-                return Observable.Return(UpdatabilityVerdict.NotEnforced(
-                    $"this deployment consumes no CI bakes ({ShippedPrebuiltBundles.PublishedRootConfigKey} "
-                    + "is not configured), so it already compiles its content at every boot — the "
-                    + "release-availability gate has nothing to enforce here"));
+            // ONE applicability rule, shared with the poller's unwired path — see NotApplicableReason.
+            if (NotApplicableReason(configuration) is { } notApplicable)
+                return Observable.Return(UpdatabilityVerdict.NotEnforced(notApplicable));
+
+            var publishedRoot = PublishedRoot!;
 
             return RequiredPackages(publishedRoot)
                 .SelectMany(required => PublishedBundleCatalogue
--- a/memex/Memex.Portal.Shared/SelfUpdate/SelfUpdateHostedService.cs
+++ b/memex/Memex.Portal.Shared/SelfUpdate/SelfUpdateHostedService.cs
@@ -324,15 +324,7 @@ public class SelfUpdateHostedService : IHostedService
         {
             var gate = ResolveAvailabilityGate();
             if (gate is null)
-                // No gate registered at all (a host that wires the poller without it). Say so once
-                // per tick rather than silently rolling as if it had passed.
-                return Observable.Defer(() =>
-                {
-                    _logger?.LogInformation(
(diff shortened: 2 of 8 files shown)


rbuergi:
## Review follow-up: the hold was drawn one state too wide

`SelfUpdateRollFloorTest` was right and my change was wrong. Pushed in `87301e1c0`.

### The mechanism

All three failures were **one** cause. `SelfUpdateRollFloorTest.NewService` builds a plain `SelfUpdateHostedService`, and that fixture never calls `AddSelfUpdate` — so no `ReleaseAvailabilityService` is registered, `gate is null`, and my new hold fired. It showed up in exactly the three cases that expect a roll; `AFreshRoll_DefersTheNextOne` kept passing because it expects *no* roll, i.e. for the wrong reason.

So the floor itself was never broken. What broke is that a **floor** test became sensitive to the **availability** gate.

### The distinction, now in the code

> "cannot verify" and "verified as not-ready" are different states, and only the second should hold.

Exactly right, and it turns out the codebase already knew where the line is — I just hadn't used it. `ReleaseAvailabilityService.IsUpdatable` decides its own applicability from **configuration**:

```csharp
if (string.IsNullOrWhiteSpace(PublishedRoot))
    return NotEnforced("this deployment consumes no CI bakes …");
```

That is the whole answer. When the service is absent I cannot *ask* it — but I can evaluate the same rule, because it reads configuration and nothing else. If `PreWarm:PrebuiltBundleRoot` is unset, a **registered** gate would have returned `NotEnforced` too. That is not "cannot verify"; it is "verified: there is nothing to verify against."

`ReleaseAvailabilityService.NotApplicableReason(IConfiguration?)` now states it once, and `IsUpdatable` reads the same static, so the two cannot drift — the repo's own rule that a rule only one caller honours is not a rule.

|  | bundle root configured | no bundle root |
|---|---|---|
| gate registered | verdict decides | `NotEnforced` → roll |
| **gate absent** | **`Unavailable` → HOLD** (the #1754 hole, still closed) | **`NotEnforced` → roll** (was wrongly held) |

Both roll paths — poller and manual button — consult it before holding.

### On the three tests

**No test was edited.** The product rule changed; they pass as written. They are floor tests and should never have been sensitive to the availability gate.

**One assumption did change, deliberately, and it is not one of the three.** `SelfUpdateAvailabilityGateTest`'s host now configures a published bundle root — i.e. it consumes CI bakes, like a production portal. Without that, the host has nothing to verify against, so the unwired-gate hold it claimed to pin was resting on precisely the state that turned out to be wrong. The test now says what it means.

### The new interaction test

`NeverRolled_WithTheGateUnavailable_StillHolds` — and it **holds**. The two gates answer different questions: the pacing floor asks *"has this install rolled too recently?"* (a never-rolled install passes trivially), the availability gate asks *"can the packages this environment deploys survive the target?"* (an unwired gate has not answered). Clearing the first says nothing about the second, and the tempting reading — *"no history, so nothing can be wrong, so let it through"* — would reopen the #1754 hole on the instance least likely to be watched.

It is not a brick: re-evaluated every tick, cleared the moment the gate is registered, named on `Admin/UpdatePolicy`, logged at Error, with `SelfUpdate:AllowUnverifiedRoll` as the one-line opt-out. And `AddSelfUpdate` registers the gate unconditionally, so no host in this repo can reach the state at all.

Plus `ReleaseGateApplicabilityTest` (5, pure) pinning the boundary cases — blank reads as absent, null configuration never throws, and the reason names the key.

### Two complementary non-vacuity probes

Both directions, because one alone would not show the line is in the right place:

```
# reopen the #1754 hole (unwired gate = clearance)
Failed!  - Failed: 2, Passed: 8, Total: 10
  Failed AnUnwiredGate_HoldsTheRoll_AndSaysItCouldNotLook   [30 s]
  Failed NeverRolled_WithTheGateUnavailable_StillHolds      [30 s]

# ignore applicability (hold on every unwired gate — the shape you caught)
Failed!  - Failed: 4, Passed: 6, Total: 10
  Failed AnUnwiredGate_OnADeploymentThatConsumesNoBakes_IsNotEnforced_AndRolls
  Failed SelfUpdateRollFloorTest.NeverRolled_IsNotHeld
  Failed SelfUpdateRollFloorTest.AnOldStampRollsImmediately_TheCrashRecoveryCase
  Failed SelfUpdateRollFloorTest.AZeroFloor_NeverHolds
```

Green now:

```
Passed!  - Failed: 0, Passed: 10, Total: 10   (SelfUpdateAvailabilityGateTest + SelfUpdateRollFloorTest)
Passed!  - Failed: 0, Passed: 38, Total: 38   (Memex.Portal.Shared.Test — update/release suites)
Passed!  - Failed: 0, Passed: 13, Total: 13   (ReleaseAvailabilityTest)
```

PR #2042, opened 2026-08-22
The security fold's permission-deciding queries declare completeness — and Complete() actually works now (#2011)

### LINKED ISSUE #2011: The security fold's Role/GroupMembership queries are unpinned scope:subtree — a group-derived permission vanishes silently once the mesh outgrows a page
Found while diagnosing #1962 and deliberately **not** fixed there, because it is latent today and a different defect.

## The shape

The security fold issues `scope:subtree` queries for `nodeType:Role` and `nodeType:GroupMembership` with **no `Complete()`** — the same unpinned-page family as #1960, whose measured cause was the user directory being read as a 50-row page.

Today it does not fire: `memex.meshweaver.cloud` carries **7 `GroupMembership` and 4 `Role` nodes**, far under the cross-schema page cap. It starts firing the moment either set outgrows a page.

## Why it matters more than a normal truncation

The failure is **silent and permission-shaped**. A truncated result is indistinguishable from "this group grants nothing", so a viewer simply loses a group-derived permission — and every surface gated on it disappears at once, with nothing logged and nothing failing.

That is exactly the "tab vanished" shape #1962 was reported as, which is why it turned up during that investigation. #1962's actual cause was a different mechanism (a stale permission chain that was never unsubscribed), so fixing it does nothing for this one.

## Why it is easy to miss

Growth is the trigger, so it will not reproduce on any install small enough to test on, and it will appear on the largest install first — the one where it is most expensive. Nobody will have changed anything.

## What would close it

Pin the queries, and make the pinning the kind that cannot be silently dropped — the security fold is precisely the place where "no result" and "not allowed" must not be the same value. Worth checking whether any other permission-deciding query in that fold is unpinned in the same way, rather than fixing only the two named here.

Related: #1960 (same family, user directory), #1962 (same symptom, different cause).

rbuergi:
Closes #2011.

## Verdict: still real — but not for the reason the issue gives, and the fix it asks for was itself broken

Two separate findings, both pinned by executed tests.

### 1. The fold declares no completeness — fixed, and made structural

`PermissionEvaluator`'s global reads (`$security-roles`, `$security-memberships`, the per-gated-type
queries) are path-less by necessity: a group and its members may live in a different partition than
the grant that names the group. A read of that shape which states no limit can be served as a PAGE,
and the caller cannot tell a page from the whole set.

In this fold the difference is a PERMISSION, and it moves in **both** directions:

- a group GRANT stops reaching its member — every surface gated on it disappears at once; and
- a group DENY stops reaching its member — **a revocation that fails OPEN**. That direction is not
  in the issue and is the one that leaks.

These reads go through `IMeshNodeStreamCache.GetQuery`, which takes query **strings** and builds its
own `MeshQueryRequest` — so `MeshQueryRequest.Complete()` is out of reach at every call site. This
adds its query-string spelling, `MeshQueryRequest.CompleteQualifier` = `limit:all`, and routes every
query the fold issues through `SecurityQueries.Enumeration`, which stamps it. A security query
therefore cannot come out truncatable whatever string its author wrote — including one that arrives
already carrying a limit, which is overwritten rather than honoured, because in this fold a page IS
the bug.

Anchored scope walks are stamped too: the ROOT scope's anchor (`_Access`, and the empty namespace
for `_Policy`) resolves to no partition and falls through to the same fan-out, so "it is anchored"
is not a reason to leave one unstamped.

### 2. 🚨 `Complete()` was broken on any partition-pinned read — so the issue's own remedy would have taken the fold down

`MeshQueryRequest.NoLimit` is documented as *"the storage providers translate it to an unbounded SQL
fetch rather than a literal `LIMIT -1`"*. That was **not true**:

- `PostgreSqlStorageAdapter`'s union branch emitted it verbatim → `PostgresException 2201W: LIMIT
  must not be negative`, which surfaces as an **availability failure of whatever read it was
  under**; and
- the row-count guards compared against it raw → `count >= -1` is true on the first row, so a
  completed read returned exactly **one** row, silently.

This is not hypothetical fallout of the stamp: any caller anywhere following `Complete()`'s advice on
a query that resolves to a single partition hit it. It is why the first run of the new end-to-end
test failed with `"the Create permission check … could NOT be established"` on every write.
Normalized across Postgres, Snowflake, Cosmos and the SQLite vector query (a vector search keeps its
documented default `k`, since "no cap" is not an expressible ANN parameter).

## ⚠️ Scope note the issue should be corrected with

The 50-row page the issue describes **cannot currently reach the security fold**. On this tree the
runtime fan-out (`PostgreSqlPartitionedMeshQuery.EnumerateFanOutAsync`) always takes the *table*
overload of `QueryAcrossSchemasAsync`, which applies **no default limit**; the paging form
(`search_across_schemas`, `DefaultFanOutLimit`) has **no `src/` caller at all** — only tests.

Measured, not assumed: seeding 62 `GroupMembership` nodes and reading them end to end returns 62,
stamped or not. The same control run against **#1960's** fix is the sharper result — removing
`.Complete()` from `UserIdentityCache.DirectoryQuery` leaves
`UserDirectoryCompletenessTests.AUserOutsideTheMostRecentlyModifiedPage_StillResolvesTheirProfile`
**passing**, so that test's behavioural half no longer demonstrates what its comment claims. Worth a
follow-up issue of its own; I have not touched it here.

That does not make the stamp pointless — it makes it the thing that keeps the fold indifferent to
which form answers it. The paging form is still on `ICrossSchemaQueryProvider`, still implemented on
both backends, and one routing change away from serving these reads again. So the guarantee is
pinned **at the seam where the page is actually produced**, where it is falsifiable, rather than by
an end-to-end test that would reproduce nothing.

## Tests — executed, and proven non-vacuous against the pre-fix code

`MeshWeaver.Security.Test/SecurityQueryCompletenessTests` (12 cases, no database) and
`MeshWeaver.Hosting.PostgreSql.Test/SecurityFoldEnumerationTests` (2 cases).

**Result with the fix:**

```
Passed!  - Failed: 0, Passed: 390, Total: 390 — MeshWeaver.Security.Test.dll
Passed!  - Failed: 0, Passed: 740, Skipped: 1, Total: 741 — MeshWeaver.Hosting.PostgreSql.Test.dll

  unstamped: 50 of 66 membership rows
  SecurityQueries.Memberships: 66 of 66 membership rows
  alice @ GrpSpace12f18c70/Doc   → Read, Execute, Api (group grant)
  bob   @ GrpSpace12f18c70/Doc   → Read, Execute, Api (direct grant — the control)
  bob   @ GrpSpace12f18c70/Gated → None (group deny still applies)
  mallory @ GrpSpace12f18c70/Doc → None (granted nothing, sees nothing)
```

The paging test carries its own control in the same run — the unstamped string the fold used to
issue gets **50 of 66**, the production builder's string gets **66 of 66**. So it is a measurement,
not an assumption.

**Control A — `SecurityQueries.Enumeration` neutered to return its input:**

```
Failed!  - Failed: 8, Passed: 4, Total: 12 — SecurityQueryCompletenessTests
  Failed …SecurityFoldEnumerationTests.TheFoldsOwnMembershipQuery_IsCompleteThroughThePagingFanOut
    AssertionException : Expected value to be -1, but found <null>.
```

**Control B — the provider `NoLimit` fixes reverted, stamp kept:**

```
  Failed …SecurityFoldEnumerationTests.TheStampedFold_StillGrants_StillDenies_AndWidensNothing
    errored: The Create permission check for node 'GrpSpace075914c0/Cohort' could not be
    established … (PostgresException: 2201W: LIMIT must not be negative)
```

## Never widens access

`mallory` — a viewer who was granted nothing and is a member of nothing — is asserted to see
`Permission.None` in the same test, and the group DENY assertion (`bob` keeps a direct grant, so his
`Read` is a control that does not depend on any membership) covers the leaking direction. Every
provider change is `HasValue` → `is > 0`, which can only return **more rows to the query engine** —
never more rows to a viewer, since row-level security is applied independently.

### DIFF
--- a/src/MeshWeaver.Hosting.Cosmos/CosmosMeshQuery.cs
+++ b/src/MeshWeaver.Hosting.Cosmos/CosmosMeshQuery.cs
@@ -124,6 +124,13 @@ public class CosmosMeshQuery : IMeshQueryProvider
         {
             parsedQuery = parsedQuery with { Limit = request.Limit is > 0 ? request.Limit : null };
         }
+        else if (parsedQuery.Limit is <= 0)
+        {
+            // Same encoding arriving through the QUERY STRING instead — `limit:all`
+            // (MeshQueryRequest.CompleteQualifier), which the security fold uses because it can
+            // only pass strings.
+            parsedQuery = parsedQuery with { Limit = null };
+        }
 
         // Strip $type filters — all items in the nodes container are MeshNodes,
         // so $type:MeshNode is always true and other $type values always false.
@@ -173,7 +180,8 @@ public class CosmosMeshQuery : IMeshQueryProvider
                     ? ParsedQuery.ProjectToSelect(node, parsedQuery.Select)
                     : node;
                 count++;
-                if (parsedQuery.Limit.HasValue && count >= parsedQuery.Limit.Value)
+                // `is > 0`: NoLimit is non-positive, and `count >= -1` is true on the first row.
+                if (parsedQuery.Limit is > 0 && count >= parsedQuery.Limit.Value)
                     yield break;
             }
             yield break;
@@ -202,7 +210,8 @@ public class CosmosMeshQuery : IMeshQueryProvider
 
             // Apply limit
             countOrig++;
-            if (parsedQuery.Limit.HasValue && countOrig >= parsedQuery.Limit.Value)
+            // `is > 0`: see above.
+            if (parsedQuery.Limit is > 0 && countOrig >= parsedQuery.Limit.Value)
                 yield break;
         }
     }
--- a/src/MeshWeaver.Hosting.Cosmos/CosmosSqlGenerator.cs
+++ b/src/MeshWeaver.Hosting.Cosmos/CosmosSqlGenerator.cs
@@ -68,7 +68,8 @@ public class CosmosSqlGenerator
         var sql = new StringBuilder("SELECT");
 
         // Add TOP if limit specified
-        if (query.Limit.HasValue)
+        // `is > 0`: MeshQueryRequest.NoLimit is non-positive and must not become `TOP -1`.
+        if (query.Limit is > 0)
             sql.Append($" TOP {query.Limit.Value}");
 
         // When the caller projects via `select:` and didn't ask for "content",
--- a/src/MeshWeaver.Hosting.PostgreSql/PostgreSqlMeshQuery.cs
+++ b/src/MeshWeaver.Hosting.PostgreSql/PostgreSqlMeshQuery.cs
@@ -302,7 +302,10 @@ public class PostgreSqlMeshQuery : IMeshQueryProvider, IVectorSearchProvider
                 if (string.IsNullOrEmpty(vectorUserId) || vectorUserId == WellKnownUsers.System)
                     vectorUserId = null;
                 var vectorNamespace = parsedQuery.Path ?? request.DefaultPath;
-                var topK = parsedQuery.Limit ?? request.Limit ?? 50;
+                // Positive(): a vector search always needs a real k, and NoLimit (non-positive)
+                // is "no cap stated" — which for an ANN ranking is the same thing as stating
+                // nothing, so it takes the same default rather than an impossible k.
+                var topK = Positive(parsedQuery.Limit) ?? Positive(request.Limit) ?? 50;
                 // Strip the text part from the filter so VectorSearchAsync's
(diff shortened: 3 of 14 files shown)


rbuergi:
Follow-up finding filed as #2048 — the 50-row cross-schema page has no `src/` caller, so the #1216/#1326/#1960 family's behavioural tests no longer demonstrate their claim. That is the "⚠️ Scope note" section of this PR, tracked separately so it is not lost when this merges.

PR #2053, opened 2026-08-22
The build chip is glyph-only — one label was misaligning a row of eight icons

rbuergi:
The header strip is icon buttons: each sibling is a `FluentButton` wrapping exactly one `FluentIcon`. The build chip was the same **except** it also rendered `Last deployed 08-21 22:16` — which made it wider than its siblings and pushed its own glyph off their shared centre line.

Both reported symptoms — the visible text, and one icon sitting at a different height — are that single difference.

## The text is removed, not hidden

A media query already dropped it below 720px:

```css
@media (max-width: 720px) { .platform-build-chip__deployed { display: none; } }
```

So the narrow layout was the intended one all along. It just never applied at the width people actually use.

## 🚨 The deployment time had to move, not go

`Describe`'s idle tooltip was `version · instance` and carried **no time at all**. Deleting the label would have lost *"when was this deployed?"* entirely — which the type's own documentation calls the question people actually bring to it:

> "Last deployed 08-18 15:35" answers the question people actually bring to it ("is this current?")

It is now folded into **every** state's tooltip, including the two update-pending ones, so the single surface the chip still has says everything it has to say.

## `DisplayText` is deleted, not left unrendered

The span was its only consumer. A record member nothing reads is a field that drifts out of agreement with what ships. Its three tests move to the tooltip and keep their reasoning intact — **WHEN** belongs to the reader, **WHICH BUILD** stays one hover away, and an unknown time is **omitted rather than invented** (no epoch date, no "unknown" filler).

## Verification

`Memex.Portal.Shared.Test` **358/358**, Release with `-warnaserror`.

### DIFF
--- a/memex/Memex.Portal.Shared/SelfUpdate/PlatformUpdateChip.cs
+++ b/memex/Memex.Portal.Shared/SelfUpdate/PlatformUpdateChip.cs
@@ -22,15 +22,17 @@ public enum PlatformUpdateChipAction
 /// reason <see cref="Settings.AboutSettingsTab.UpdateStatusMarkdown"/> takes its localizer as a
 /// function). The component holds only markup and wiring.
 /// </summary>
-/// <param name="DisplayText">
-/// What the header shows: when this build was last deployed, or <c>null</c> when that is unknown
-/// (the glyph still renders, so the button is never blank-but-clickable).
+/// <param name="Tooltip">
+/// The full sentence on hover — build identity in every state, and the deployment time.
+///
+/// <para>🚨 The chip renders a GLYPH ONLY. It sits in a strip of icon buttons, and a chip that also
+/// rendered text was wider than its siblings and pushed its own glyph off their shared centre line.
+/// So everything the chip has to say is here, and this string is the whole surface: if a fact is not
+/// in it, the reader cannot get at it without opening the About page.</para>
 /// </param>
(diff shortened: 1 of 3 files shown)


rbuergi:
Merging over a red `Run tests (shard 4)`, with the control run rather than asserted.

**The failure is `MeshWeaver.AI.Test.ProbeHubCostTest.GetContentSchema_DoesNotBuildTheNodeControlPlane`.** This diff is one `.razor` file, one record and its tests, all in `Memex.Portal.Shared` — it cannot reach `MeshWeaver.AI.Test`.

That alone is not evidence, so I checked the right thing: **`ProbeHubCostTest` fails on `main` itself.** Across main's last six completed *Build and Test* runs:

| run | conclusion | ProbeHubCost failures |
|---|---|---|
| 32566889549 | success | 0 |
| 32566839566 | success | 0 |
| 32566457849 | success | 0 |
| 32564482939 | success | 0 |
| **32564051012** | **failure** | **2** |
| **32562019976** | **failure** | **2** |

Two of six, on main, with no involvement from this branch. Independently, the Expo migration PR (#2056) hit the same suite on both of its runs — a diff containing **zero C# files**.

Recorded on #1384, which is where the population belongs.

Local verification for this change: `Memex.Portal.Shared.Test` **358/358**, Release with `-warnaserror`.

PR #2068, opened 2026-08-22
A transient probe hub never starts its data sources — the CD-killing flake

rbuergi:
## The flake that switched CD off twice today

`ProbeHubCostTest.ValidateContentAgainstSchema_DoesNotBuildTheNodeControlPlane` went red on shard 4 twice on 2026-08-22 — on main `1b0b834` and on #2062 — and CD builds nothing while main's required check is red, so both reds turned continuous delivery off behind a green-looking wall.

## The race

The probe's control plane is gone (that's what the test pins), but `InitializeDataSources` still eagerly started every data source on the probe's **init turn** — `GetStream → SynchronizationStream..ctor → GetHostedHub(sync/…)` — while `MeshOperations.ReadFromContentType` disposes the probe **in the same breath**. Usually creation wins. When dispose wins, `HostedHubsCollection` logs *"Rejecting hosted hub creation … during disposal"* — the exact string `IsTeardownFault` matches, by design.

## The fix

A `TransientNodeProbe` hub never starts its sources. Configuration (`Initialize`, inside `Build`) is untouched — it fills the registry maps the probe exists to read. Lazy stream creation on a real request still works; only the eager start nobody consumes is skipped. The marker moves from Graph (internal) to `MeshWeaver.Data` so the data context can read it; `AsTransientNodeProbe()` keeps its public signature.

## Non-vacuity

Measured on this branch: a validate probe creates **1 hub total** (the failing run showed 7) and **0** `sync/` streams, with **0** teardown fault records. All 3 ProbeHubCostTest cases green.

Companion: the CD workflow now goes RED when it skips a build because main's check settled red (separate PR) — so even a future flake of this class cannot switch delivery off silently.

### DIFF
--- a/src/MeshWeaver.AI/MeshOperations.cs
+++ b/src/MeshWeaver.AI/MeshOperations.cs
@@ -2749,8 +2749,14 @@ public class MeshOperations
                     var probeAddress = new Address($"{addressPrefix}/{Guid.NewGuid():N}");
                     // Stamp the NodeType path so a WithContentType reached from this probe records
                     // an EXACT registry entry rather than a bare-name one (see NodeTypePathHolder).
+                    // startDataSources: false — this probe reads NOTHING but the type registry,
+                    // which Initialize fills inside Build; starting sources would open a sync/
+                    // stream whose creation races the probeHub.Dispose() below (the CD-killing
+                    // ProbeHubCostTest flake). Probes that snapshot data keep the default.
                     var probeHub = hub.GetHostedHub(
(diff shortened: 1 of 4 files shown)


rbuergi:
Corrected: the first cut starved probes that DO read data through their sources (six tests across shards 2+4 — NodeTypeModelProbeTest ×3, StaleStampRootBindingTest ×2, IncrementalUpdateTest ×1). The opt-out is now per-probe: `TransientNodeProbe(StartDataSources = …)`, default **true** (pre-existing behaviour); only the AI schema validation/lookup probes — registry-only readers whose eager `sync/` stream races their own dispose — pass `false`. All ten tests green locally.

PR #2070, opened 2026-08-22
The jobrunner Role gets its RoleBinding — a Role without one is furniture

rbuergi:
Found during the first live operator enablement (2026-08-22): `jobrunner.yaml` ships the ServiceAccount, the token Secret and the Role — and **no RoleBinding**, so the portal's token 403s on its first CreateJob. From the mesh that reads as a misconfigured operator, not a missing binding.

Verified on the cluster with SubjectAccessReview: create-jobs-in-memex-ops `false` before, `true` after; delete-namespaces stays `false` (the boundary the three-principal design promises). Applied live already; this lands the same binding in the manifest so the next cluster doesn't rediscover it.


rbuergi:
CI triage + local verification record:

**Why the previous run was red:** shard 4 failed on `ProbeHubCostTest.{GetContentSchema,ValidateContentAgainstSchema}_DoesNotBuildTheNodeControlPlane` — a teardown race in the AI schema-validation probes (`Rejecting hosted hub creation … during disposal`). This diff (two files under `deploy/aks/manifests/hosting-operator/`) cannot reach that code. Main fixed exactly this in ed0ff1b89 (*the opt-out is per-probe*), committed one minute after this PR's run started; every main run since is green. Merged `origin/main` (f25ef49) to re-sample CI past the fix. Revert-check `git diff origin/main...HEAD --stat` shows only the two intended files.

**Verified locally (offline):** `jobrunner.yaml` parses as 4 documents; the token Secret lives in the SA's namespace (token-controller rule) and its annotation names the SA; the RoleBinding lives in the Role's namespace (`memex-ops`), its `roleRef` matches the Role in the same file, and its subject is the foreign SA `memex/hosting-jobrunner`. Cross-namespace RoleBinding subjects are legal RBAC — this is the standard shape for scoping power in one namespace to an identity mountable in another.

**Only a cluster can prove:** that the apply succeeds against the live API server and that the bound token passes a SubjectAccessReview (`create jobs` in `memex-ops` = yes, anything else = no). Per the manifest comments, that SAR was already run live on 2026-08-22 (create-jobs true, delete-namespaces false). No cluster was touched from this session.

PR #2100, opened 2026-08-23
Framework-release broadcast: memex fans releases out to the satellites

rbuergi:
## What

Makes the release→satellite-rebake wave **event-driven with memex as the broadcast hub**, instead of only the ~30-min schedule poll. When the platform releases, `FrameworkReleaseBroadcaster` (memex-side) POSTs a `repository_dispatch: meshweaver-framework-released` to each subscriber repo, whose CI already subscribes and rebakes against the new framework in seconds.

## Why this reverses "pull, not push" — and resolves its objections

`main-cd.yml` deliberately had **no** dependent-dispatch job; its comment argued the platform must not hold (a) a satellite write credential or (b) a hand-maintained subscriber list. This routes the broadcast through **memex**, which removes both from the release path:

- **No credential in the release path** — memex fans out with the GitHub App it *already* holds for GitSync (via `GitHubAppTokenService`). The platform release only signs a POST to memex (the existing `notify-platform-update` pattern).
- **No list here** — the subscriber set lives in the Hosting fleet registry on memex ("register a new repo in Hosting"); nothing is named in `main-cd.yml`.
- **Pull stays the fallback** — each repo's `schedule` run still rebakes, so a lost dispatch costs at most one delayed wave. Reporter-class.

Both comment blocks in `main-cd.yml` are updated to state the new decision.

## Shape

- `FrameworkReleaseBroadcaster` + `FrameworkBroadcastOptions` in `src/MeshWeaver.GitSync`, registered in `GitHubSyncConfiguration`; section bound in `MemexConfiguration`.
- Reactive end to end (IIoPool, no async escaping a signature), matching `GitHubAppTokenService`.
- Per-repo **isolation**: a non-2xx or a thrown network error becomes a `RepoDispatch(ok:false)`; `Observable.Concat` never short-circuits; unconfigured App or empty list → an outcome, never a throw.
- Subscriber set is **injected** (the Hosting registry is the source); `FrameworkBroadcast:Subscribers` config is the interim fallback.

## Tests / build

`dotnet build` clean (0 warnings, warnings-as-errors) on both `MeshWeaver.GitSync.Test` and `Memex.Portal.Shared`. New `FrameworkReleaseBroadcasterTest` — **6/6 pass**: subscriber normalization, dispatch-request shape, unconfigured-is-inert, no-subscribers no-op, and per-repo isolation (204 / 404 / thrown-network across three repos, all attempted).

## The other half (separate PR, Plugins repo)

The trigger + subscriber registry live in the Plugins Hosting module: `PlatformBuildInboxWatcher` (which memex already hits at `Hosting/PlatformBuilds`) calls `FrameworkReleaseBroadcaster.Broadcast(subscribers, version)`, and the registry is where a new repo registers. Also required for the broadcast to succeed: **install the memex GitHub App on the 4 satellite repos with `contents: write`** (org-admin).


rbuergi:
Addressed both review threads, and the CI red:

**1. `DispatchOne` used `IIoPool.Run`** — switched to `Invoke`. `Run` is the eager promise-cache shape: it starts the POST at call time, decoupled from the subscription, so an unsubscribe (timeout, shutdown) would leave the dispatch running and holding an HTTP pool slot. `Invoke` stays cold and propagates cancellation into the leaf's `ct`.

**2. Test options mismatch** — `Broadcast_NoSubscribers_IsANoOp` now passes ONE `GitHubAppOptions` instance to both the token service and the broadcaster (they resolve the same `IOptions` in production). And the GHES case is now genuinely pinned: `Broadcast_IsolatesPerRepoFailures_AndDispatchesTheRest` configures a non-default `ApiBaseUrl` (`https://ghes.example/api/v3`) and asserts every dispatch URL was built against it, never the default base.

**CI root cause** — shard 2's red was NodeType-compile tests dying with the "BELOW-ROSLYN" canary (`NullReferenceException` at `NamedTypeSymbol…get_ContainingTypeDefinition` on BOTH the shared and the pristine leg — the canary's own text attributes it to CLR heap/JIT state on the runner, "no reference-set change can fix this"). `MeshWeaver.GitSync.Test` itself was green on that run; this diff (GitSync + comment-only `main-cd.yml` edits + one `Configure` line in `MemexConfiguration`) cannot reach the compile pipeline. Branch is merged with today's `origin/main` for a fresh run.

No What's New entry: release/CD machinery with no user-visible change (deliberate skip per the PR checklist).

Verified locally at CI flags: `MeshWeaver.GitSync.Test` builds Release `-warnaserror` clean; `FrameworkReleaseBroadcasterTest` 6/6.

Issue #2031, opened 2026-08-21
UserDirectoryCompletenessTests times out under the full assembly — reproduces on clean main, passes in isolation


rbuergi:
`MeshWeaver.Hosting.PostgreSql.Test.UserDirectoryCompletenessTests.AUserOutsideTheMostRecentlyModifiedPage_StillResolvesTheirProfile` fails on `main`. It is **not** a network flake — it is deterministic and depends on what else ran in the process.

```
System.TimeoutException : The operation has timed out.
  at …UserDirectoryCompletenessTests.AUserOutsideTheMostRecentlyModifiedPage_StillResolvesTheirProfile()
     UserDirectoryCompletenessTests.cs:line 135
```

Line 135 is the `.Timeout(60.Seconds())` on `cache.WhenDetermined(ForgottenEmail)` — so the lookup did not complete within a minute, rather than returning a wrong answer.

## Reproduction, and the control that assigns it to main

| Run | Result |
|---|---|
| Filtered to this class, on a PR branch | **passes 2/2** |
| **Full assembly, clean `origin/main` @ `46a4378f9b`** (fresh worktree, no PR changes) | **fails** — 737 passed, 1 failed |
| Full assembly, on the PR branch that CI reddened | fails identically — 737 passed, 1 failed |

The middle row is the point: the failure reproduces with **none of the PR's changes present**, so it belongs to `main`. It surfaced on #2028 only because that PR's shard happened to run it.

A possibly useful bracket, offered as a hint rather than a bisect: the same full-assembly run passed 738/739 at `d69a20811` and fails at `46a4378f9b`. Different checkouts, so DB warmth is not controlled — worth confirming before trusting it.

## Why the shape matters

The test seeds a user, buries it under `CrowdSize` more so it sits outside the 50-row default page, then reads it back through `UserIdentityCache`. Post-v10 that read is *"an unpinned query UNIONing one schema per user"* (the test's own comment). Under the full assembly the process already holds the state of ~737 other tests, so that union has far more to do than when the class runs alone — and 60s stops being generous.

So this reads as a real scaling property of the per-user-schema union being asserted with a fixed timeout, not as a bad test. Two different fixes follow depending on which is true, which is why it is worth diagnosing rather than raising the timeout:

- if the union is genuinely O(users-in-mesh) per lookup, the production read has the same property and the test is telling us something;
- if it is test-local contamination (schemas left behind by earlier tests), the isolation is what needs fixing.

[...]

PR #2102, opened 2026-08-23
Stop OAuth re-consent on every connect; expire stale login handshake cookies

rbuergi:
Two production defects observed on `memex.meshweaver.cloud` today, both in the sign-in path.

## 1. Consent was requested on *every* connection
`/register` (RFC 7591 Dynamic Client Registration) minted a fresh `RandomNumberGenerator` `client_id` on every call and persisted nothing — the code even said so: *"The mesh does not persist client registrations — it issues a random client_id"*.

MCP clients re-register on **every reconnect**. So each reconnect presented a **brand-new application**, and there was nothing a prior consent could be matched against.

**Fix:** derive `client_id` from the client's own metadata (name + redirect URIs, order-independent) plus this deployment's origin. A re-registering client now gets the **same id** back.

Nothing else changes — the id stays opaque to the caller and is still echoed through `/authorize` and `/token`, where the code store validates `client_id`+`redirect_uri` consistency. A `client_id` is **public by construction** for these public clients (`token_endpoint_auth_method=none`); security rests on exact redirect_uri matching plus PKCE, both untouched. The origin is folded in so ids aren't shared across portals.

## 2. Login wedged by handshake cookies that never expire
Correlation/nonce cookies are written when sign-in **starts** and deleted when the callback **completes** — so every abandoned login (closed tab, error, retry) leaves one behind **permanently**, ~400 bytes each. They accumulate until the `Cookie` header crosses the server's request-header limit (~10 KB), the connection is reset mid-flow, and the user gets a **502 that retrying makes worse** — each retry adds another cookie.

**Fix:** a 15-minute `MaxAge`, so the browser evicts them itself. The handshake needs seconds.

**Evidence:** a clean browser profile signed in on the first try while an accumulated profile failed against the **same pods in the same second**; pods were healthy and no 502 appeared in any portal log.

## Verification
- ✅ `dotnet build` — Blazor.Portal and Portal.Shared, 0 warnings, 0 errors
- ✅ 6 new tests pass — derivation is stable across re-registration, order-independent, distinct per client and per deployment, URL-safe


rbuergi:
Addressed all three review threads in 33cc29032 (+ c8238067b), and the CI red:

**1. Obsolete register test** — `Register_EachCallReturnsUniqueClientId` pinned exactly the randomness that caused the re-consent bug. It is rewritten to pin the new contract: `Register_ReRegistrationReturnsTheSameClientId` (same request → same id) plus `Register_DistinctClientsStillGetDistinctClientIds` (collision-freedom kept).

**2. Handshake-cookie expiry on every provider** — the 15-minute policy is hoisted to one shared constant, `AuthenticationBuilderExtensions.LoginHandshakeCookieMaxAge`, and applied to the unified Microsoft (correlation + nonce), Apple, Google and LinkedIn handlers, and to the legacy `AddMicrosoftIdentityWebApp` branch via a scheme-scoped `Configure` of the OIDC options in `MemexConfiguration`.

**3. What's New entry** — `src/MeshWeaver.Documentation/Data/WhatsNew/2026-08-24-connect-once-consent-once.md` (`Category: Fix`, `Order: -20260824`) covers both user-visible repairs.

**CI root cause** — shard 2's three `OAuthConnectControllerTests` failures had two distinct causes: the obsolete register expectation (fixed above), and the follow-up **grace-window commit**, which broke the two #1493 supersede tests. That commit is **reverted** (c8238067b): it weakened the pinned one-live-credential invariant (with no follow-up authorization the superseded token would have stayed valid for up to a year), it was motivated by no observed failure (both supersede tests were green on the commit before it), and it would not have protected the two-install case it gestured at — the other install's token still dies two minutes later. Shard 5's `UserDirectoryCompletenessTests` red is the known main-side flake tracked in #2031 (reproduces on clean main, passes in isolation) — untouched by this diff.

Verified locally at CI flags (Release, `-warnaserror`, 0 warnings): `MeshWeaver.Auth.Test` OAuthConnectControllerTests 29/29, `Memex.Portal.Shared.Test` 366/366, WhatsNew/link-integrity 2/2. Branch is merged with today's `origin/main`.

PR #2114, opened 2026-08-24
feat(portal): ?hint=chat pulses the header chat toggle

rbuergi:
Content pages cannot reach into the portal chrome, so a course lesson that says *"the chat is always one click away in the top menu"* had no way to **show** a newcomer which button that is.

This adds a tiny, inert hint affordance: a page links **its own URL with `?hint=chat`**, and landing on it makes the header's side-panel/chat toggle **pulse for ~6 seconds** (an accent-token ring, so it reads in both themes), then self-dismisses; clicking the toggle dismisses it immediately. No navigation, no panel-state change — a pointer, not a behavior.

- `PortalLayoutBase.razor.cs`: `CheckChatHint` armed from `OnInitialized` + `LocationChanged`; `DismissChatHint` on `ToggleSidePanel`; CTS-guarded 6-second self-dismiss.
- `PortalLayoutBase.razor`: conditional `chat-hint-pulse` class on `.side-panel-toggle`.
- `PortalLayoutBase.razor.css`: the keyframes (5 × 1.2 s, `var(--accent-fill-rest)` ring).
- What's New entry included.

Built with `-warnaserror`. First consumer: the education repo's Agentic Business Primer lesson 1 (link lands after this ships).

### DIFF
--- a/src/MeshWeaver.Blazor.Portal/Layout/PortalLayoutBase.razor
+++ b/src/MeshWeaver.Blazor.Portal/Layout/PortalLayoutBase.razor
@@ -166,7 +166,7 @@
             </div>
             <AuthorizeView>
                 <Authorized>
-                    <div class="side-panel-toggle">
+                    <div class="side-panel-toggle @(ChatHintActive ? "chat-hint-pulse" : null)">
                         @* Context-aware icon: open → close-panel; closed on a thread → show-context;
                            closed elsewhere → chat (open a new conversation). *@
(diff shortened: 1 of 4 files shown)


rbuergi:
All three addressed in 991427853: `@keyframes` (single @ — .razor.css is plain CSS; note ThreadChatView.razor.css:201 carries the same `@@keyframes` typo and its skeleton pulse likely never ran — separate fix candidate), exact `hint=chat` pair matching, and the auto-dismiss flips state inside `InvokeAsync` on the renderer context. The What's New front matter was already fixed in af9782695 (Category: Feature, Order: -20260824).

PR #2126, opened 2026-08-24
Refuse to store a provider key in plaintext (fail closed on write, tolerant on read)

rbuergi:
## 🚨 Rotation is required regardless, and it is not this PR

**This stops NEW plaintext credentials. It does not un-leak the existing ones.** `Provider/_Policy`
has `publicRead: true`, and every value written in the clear is also in the node's **version
history** — so a green merge here is not "the leak is handled". Any credential that was stored
unencrypted must be **rotated at its source**, and that is the operator's action, not a code change.

---

## The contradiction this PR was stuck on

Since 2026-08-24 the branch **neither refused plaintext nor passed**:

- `ProviderKeyProtectorTest.NoMasterKey_RefusesToWrite_ButStillReads` asserted `Protect("plain")` **throws**.
- The branch's most recent commit was *"revert the protector throw for real — it was breaking the seeder"*.

The mechanism was removed and the assertion asserting it was left standing.

## Why the seeder broke — and why it is not exempted

`ProviderCredentialSeed` (the boot-time step that copies `{Section}:ApiKey` from a deployment's
configuration onto its `Provider/{name}` node) **never wanted the passthrough**. It wanted a
*capability question* — "can this deployment encrypt?" — and answered it by calling `Protect` and
inspecting what came back, which only worked *because* `Protect` degraded:

```csharp
var protectedKey = protector is null ? c.ApiKey : protector.Protect(c.ApiKey);
var outcome = Decide(current, hasConfiguredKey: true, protectionAvailable: IsProtected(protectedKey));
```

So it already refuses correctly — `ProviderSeedOutcome.RefusedUnprotected`, logged at Error, node
left keyless. It just *discovered the state from a fallback*. Making `Protect` throw turns that into
an exception inside a sequential per-provider pipeline, which aborts the seed for **every remaining
provider** instead of reporting one refusal.

### The choice, plainly

**Option 1 (give the seeder a master key) — chosen where it applies, but it cannot be the whole
answer.** The seeder's own fixture, `ProviderCredentialSeedWithoutMasterKeyTest`, *is* an
unconfigured deployment, and unconfigured deployments exist in production — that is the reported
leak. The refusal path must keep working **without** a master key, so no amount of configuration
removes the need for the seeder to decide.

Where option 1 *does* apply is the **test environment**, and that turned out to be the real finding:
`MeshWeaver.Fixture` now configures a test master key for every mesh test. The **default test
environment was the misconfigured one** — which is precisely why this survived. As the original PR
body put it, `ProviderKeyEncryptionTest` stayed green *because it configures a master key, so the
degradation only happened where nobody was testing*. It lives in `ServiceSetup` rather than
`test/appsettings.json` because eleven test projects ship their own file that **shadows** the shared
one. Fixtures that must exercise the unprotected deployment register their own `IConfiguration` and
still do.

**Option 2 (stop seeding secrets)** would delete a convergence the platform needs: measured on
`memex.systemorph.com`, `Provider/Anthropic` was created keyless on 2026-08-14, `Anthropic__ApiKey`
was configured *after* it, and the node stayed keyless until a human pasted the key in on 08-21. The
seeder exists to close that gap; removing it re-opens a real operational defect to fix a different one.

**Option 3 (a named exemption) was not needed** — so it was not taken. The seeder asks
`IMasterKeyProvider` **before** protecting:

```csharp
private static string? ProtectOrNothing(
    IProviderKeyProtector? protector, IMasterKeyProvider? masterKeys, string? configuredKey)
{
    if (protector is null || masterKeys?.GetMasterKey() is null)
        return null;                       // a DECISION NOT TO WRITE, never a fallback
    return protector.Protect(configuredKey);
}
```

`null` feeds `IsProtected(...)` → `RefusedUnprotected`. **There is no branch from here to a persist**,
the plaintext is never returned, and the produced bytes are still checked for the `enc:` tag before
the write. The refusal in `Protect` is therefore **absolute, with no exempted caller**.

## The passthrough was in five more places, all silent

This is the part the PR title understates. `IProviderKeyProtector` is not provider-key-only — it also
carries a **GitHub PAT**, the plugin registry's **sync-token signing key**, an installation's
**registry credential**, and the **Entra EA refresh token**. Every one of them wrote
`protector?.Protect(x) ?? x` *on top of* a degrading `Protect` — the same leak, in four more
subsystems, equally unreported. Leaving them would have shipped a PR titled "refuse to store a key in
plaintext" that refused in one place out of six.

| Call site | Before | Now |
|---|---|---|
| `ProviderCredentialSeed` | probe via passthrough | asks `IMasterKeyProvider`, reports `RefusedUnprotected` |
| `ModelProviderService.Protect` | duplicated master-key probe + dead null branch | delegates; one rule, one place |
| `ModelProviderLayoutAreas.SaveKey` | `protector is null ? newKey : …` | protects **inside** the chain → refusal lands on `onError`, not out of an `OnNext` |
| `GitHubCredentialService.Protect` | `protector is null ? plaintext : …` | required; refusal reaches the caller |
| `SyncTokenSigningKeyService.Protect` | `protector?.Protect(e) ?? e` | required — this key **signs** every sync token |
| `InstanceAutoRegistrationService` | `protector?.Protect(k) ?? k` | required |

`ConfigMasterKeyProvider`'s "no key configured" line moves from **Information** to **Error**. It used
to read *"keys stored as plaintext"* at Information — the log line that described the leak and was
never acted on.

## 🚨 The asymmetry is unchanged and deliberate

`Unprotect()` **stays a passthrough**. Instances already hold keys written in the clear by the old
behaviour; refusing on read would take them down on upgrade for data the platform itself produced.
**Fail on the way IN, tolerate on the way OUT.**

## Operational consequence, stated plainly

A deployment with **no** `Ai:KeyProtection:MasterKey` can no longer store a credential: entering a
provider key fails with an actionable message, a GitHub credential cannot be saved, a registry can
no longer mint a sync-token signing key, and EA token rotation fails. That is the intended reading of
"a missing master key is a configuration fault, not a degraded mode" — every one of those paths was
previously writing a live secret in cleartext. `deploy/helm/values.yaml` and `values.aks.yaml` already
carry the key; it is supplied out-of-band per environment.

## Tests

- **`ProviderKeyProtectorTest.NoMasterKey_RefusesToWrite_ButStillReads`** — the assertion that has
  been red for two days, **kept as-is** and now green. It is the test that fails without the fix.
- **`ProviderKeyNoPlaintextTest`** — refusal names the fault, never echoes the key; and the keyless
  boundary is now genuinely asserted (`null` / `""` / already-tagged still pass without a master key)
  rather than asserting a service is non-null.
- **`ProviderKeyProtectorRegistrationTest`** — asserted `Unprotect(Protect(x)) == x` *"whether or not
  this host configures a master key"*, which a passthrough satisfies. Now asserts the **stored form
  is `enc:v1:` ciphertext**, which it cannot.
- **`NoPlaintextCredentialFallbackGuard`** (new ratchet) — fails on any `Protect` call site carrying
  `??`, `is null ?` or `?.Protect(`. It caught the **first version of the seeder's own fix**, which
  is the point: the guard, not care, is what keeps the refusal from widening back.

Verified locally, Release `-warnaserror` clean on every touched project:

| Suite | Result |
|---|---|
| `MeshWeaver.AI.Test` (provider-key / seed / service) | 36 / 36 |
| `MeshWeaver.Auth.Test` (sync-token + instance registration) | 22 / 22 |
| `MeshWeaver.GitSync.Test` | 187 / 187 |
| `MeshWeaver.PluginCatalog.Test` (`ModuleAutoDiscovery`) | 14 / 14 |
| `MeshWeaver.Graph.Test` (protector registration) | 1 / 1 |
| `MeshWeaver.Documentation.Test` (all guards) | 70 / 70 |

What's New entry: `Category: Fix` — and it says the rotation part out loud, for the same reason this
body does.

---

## One follow-up, in the sibling repo — deliberately not in this PR

`MeshWeaver.Plugins` → `src/MeshWeaver.Blazor.Portal/Chat/ThreadChatView.razor.cs:1717`
(`StoreHarnessCredential`) carries the same shape this PR removes here:

```csharp
var stored = protector is null ? credential : protector.Protect(credential);
```

It compiles fine against this change (behaviour-only, no signature change) and it stops leaking as
soon as this merges — but the refusal would surface as an **unhandled Blazor exception** instead of
its own `"Couldn't store the credential: {ex.Message}"` status, because `Protect` runs *before* the
`CreateOrUpdateNode(...).Subscribe(_, ex => …)` that has the handler. The fix is the same one applied
to `ModelProviderLayoutAreas.SaveKey`: protect inside the chain so the refusal lands on the existing
`onError`. Platform first, then the dependent repo.


rbuergi:
**Merged `main` in and repaired the compile — but this PR is red for a REAL reason that predates my merge, and it is the reason the plaintext-key vulnerability is still open.**

## What I fixed (compile only)

Main had moved types out from under the test:

- `Memex.Portal.Shared.Models` no longer exists — `ModelProviderService` now lives in `MeshWeaver.AI.Portal` (the portal split).
- `IProviderKeyProtector` moved to `MeshWeaver.Mesh.Security`, so the class-doc `cref` no longer resolved — **CS1574, an error under `-warnaserror`**, not a warning.

Revert-check still shows exactly this PR's four files. No behaviour touched.

## The real blocker — the PR contradicts itself

```
ProviderKeyProtectorTest.NoMasterKey_RefusesToWrite_ButStillReads   FAIL
  Expected value to be of type InvalidOperationException, but found <null>.
```

The test asserts that `Protect("plain")` **throws** when no master key is configured. The branch's own most recent commit is:

> `fix(ai): revert the protector throw for real — it was breaking the seeder`

So the mechanism was removed and the test asserting it was left in place. **The PR currently neither refuses plaintext nor passes.** That is not merge damage — it has been true since 2026-08-24.

## Why this matters more than a red test

The test's own comment states the stakes exactly:

> *That passthrough is the defect — it PERSISTED A RAW PROVIDER KEY, and was found in production with a live OpenRouter key in cleartext in `Provider/OpenRouter` node content.*

So the fix was written, it worked, and it was reverted **because refusing to write broke the seeder**. That is the actual design question this PR is stuck on, and it needs deciding rather than patching:

1. **Give the seeder a master key** in dev/test — the refusal stays absolute, and "no master key" genuinely means misconfiguration everywhere.
2. **Make the seeder path an explicit, named exemption** — narrow, loud, and asserted (never a silent passthrough, which is the defect).
3. **Stop seeding keys at all** — seed providers without secrets and require the key to be supplied through the protected path.

🚨 I have deliberately **not** chosen. Option 2 done carelessly recreates the vulnerability with extra steps, and options 1 and 3 change how deployments are bootstrapped. Whichever is picked, the test should be updated in the SAME change so the code and its assertion never disagree again.

## And this fixes only the write side

Worth restating so it is not lost: even green, this PR stops **new** plaintext keys. It does not un-leak the existing ones. `Provider/_Policy` has `publicRead: true`, and the values are also in version history — so **rotation has to happen regardless**, and it is the first step, not the last.

PR #2128, opened 2026-08-24
Adopt the app's icon on the record hub's initialization

rbuergi:
Maintainer's placement, and it is the right one: *"we can update in the initialization of the app hubs"*.

## Why not where I first put it

The first attempt repaired icons inside `CatalogAreaView`'s reactive selector. Review caught two defects and both were real:

- its "run once" flag lived on the **subscription**, so every navigation and every reconnect re-ran a cross-partition query plus writes — the storm shape the record model exists to avoid;
- the selector runs after the layout area returns, by which point the ambient `AccessService` context is **cleared**, so the query and every remote update would have executed with no viewer identity: private targets filtered out, cross-partition writes made as nobody.

I removed it rather than patched it — patching would have relocated the defect.

## Why the record's own hub is right

It initialises **once per activation**, it **owns the node it writes**, and nothing about it is tied to a viewer's page. No render path, no ambient-identity problem, no per-subscription repetition.

`AppIconAdoption.AdoptOnInit` reads its own node, does nothing unless the icon is missing or generic, resolves exactly **one** target (its `MainNode` — the app it opens), and writes only when that target has a better icon. Three properties make it converge rather than churn:

- a target with **no** icon leaves the record untouched, instead of rewriting the same placeholder on every activation;
- the update **re-checks inside** `Update`, so a real icon written meanwhile (the Store installing over it) always wins;
- the init gate opens regardless — an unreachable target is a cosmetic miss, never a reason to hold a hub shut. Reactive end to end: no `async`, no `Task`, no `Observable.FromAsync`.

## Tests

The decision is a pure function, so every branch is pinned in `AppIconAdoptionTest` (11 cases): adopt when the icon is generic/empty/null, never overwrite a real icon, never rewrite a placeholder with a placeholder, never adopt from a record whose `MainNode` is its own path, match the placeholder case-insensitively, and confirm the node type still carries the hub configuration the hook lives on. Graph.Test **1553/1553**.

Long term the Store stamps the icon when it writes the record and this becomes a no-op; until then the platform repairs what it renders, because a placeholder grid is a broken feature regardless of which side was supposed to fill it in.

### DIFF
--- a/src/MeshWeaver.Documentation/Data/WhatsNew/2026-08-24-app-icons-adopt.md
+++ b/src/MeshWeaver.Documentation/Data/WhatsNew/2026-08-24-app-icons-adopt.md
@@ -1,0 +1,16 @@
+---
+Name: App tiles wear their own icon
+Category: Fix
+Description: An app tile still showing the generic placeholder now picks up the icon of the app it opens.
+Icon: Sparkle
+Order: -20260824
+---
+
+# App tiles wear their own icon
+
+Apps added before their icon was recorded showed a generic placeholder, and a grid of identical
+placeholders defeats the point of an icon grid — you should recognise an app before you read its
+label.
+
+Each app entry now takes the icon of the app it opens, the first time that entry is used. An app
+that already has its own icon keeps it, so nothing you or the Store set is overwritten.
--- a/src/MeshWeaver.Graph/Configuration/AppIconAdoption.cs
+++ b/src/MeshWeaver.Graph/Configuration/AppIconAdoption.cs
@@ -1,4 +1,5 @@
 using System.Reactive;
+using System.Reactive.Disposables;
 using System.Reactive.Linq;
 using MeshWeaver.Data;
 using MeshWeaver.Mesh;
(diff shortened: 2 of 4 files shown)


rbuergi:
Superseded by the logon-action framework, not abandoned.

The maintainer's direction: icon adoption "should run as user" and "when the user logs on". This PR ran it on the app record hub's INITIALIZATION, which is the wrong shape for both halves — init carries no viewer identity (the first review round on this PR was exactly that: the query and write would have failed closed and adopted nothing), and it pays the cost on every activation rather than once per person.

Adopting an app's icon is a per-user maintenance action on the user's own `{owner}/_App/{id}` records, so it belongs in the logon framework being built now, running under that user's own identity — no impersonation at all.

Carried forward: the pure decision helpers `NeedsIcon` / `TargetOf` / `IconToAdopt` and their 11 tests in `AppIconAdoptionTest`. They are the part that was worth having, and they are unchanged by the move.

Also worth recording, since it cost a CI run: the `Observable.Using(() => access.ImpersonateAsSystem(), ...)` fix I pushed here trips `ImpersonationScopeSiteRatchetGuard` — that shape is a banned defect class (#1790), and `RunAsSystem` / `RunAs` are the sanctioned forms. The branch stays for reference.

Issue #1289, opened 2026-08-12
BarePartitionPath_NoMeshNode_RespondsToPing uses 60% of its 5s budget when idle — a ping the comment says takes <1s takes ~3s


rbuergi:
`PartitionRootActivationTest.BarePartitionPath_NoMeshNode_RespondsToPing` fails intermittently on CI shard 0, and the reason is that its budget has almost no headroom left.

## Measured

The test's own comment claims *"the ping responds in < 1 s"* and that *"5 s leaves comfortable headroom"*. Measured on an **idle** dev box, the ping takes **2.9–3.1 s** — roughly **60% of `FastBudget` consumed before CI load is applied at all**.

The CI failure is that budget expiring, not the condition the test exists to catch:

```
TaskCanceledException at line 74
TEST START  11:18:25.481
TEST FAILED 11:18:30.494   = 5.013 s  = exactly FastBudget
```

Note that is *not* the 30 s Orleans grain timeout the test was written to detect — so when it fails, it fails for a reason unrelated to its purpose, which makes it a poor signal in both directions.

## Why this is worth fixing rather than tolerating

A test that consistently uses 60% of its budget on an idle machine will fail whenever CI is loaded, and CI is loaded whenever several PRs merge close together. That makes it a permanent contributor to the intermittent shard-0 Orleans red, and every occurrence costs someone a triage cycle deciding whether their diff caused it.

## 🚨 The fix is NOT raising `FastBudget`

That is the band-aid, and it would hide the actual question: **why does a bare-partition ping take ~3 s when the author expected < 1 s?** Either the expectation was always wrong (in which case the comment and the budget should be corrected together, with the real number recorded and justified), or activation has become slower since the test was written and that regression is the thing to find.

Establish which before changing any number.

## Provenance

Found during the #1250 and #1253 verification work; the measurements come from matched full-suite runs of `MeshWeaver.Hosting.Orleans.Test` (156 tests) on an otherwise idle machine, including one under `DOTNET_PROCESSOR_COUNT=2`. The test passes locally in every arm — it only fails under CI load, which is consistent with the thin-headroom explanation rather than a functional defect.

rbuergi:
## Not reproducible — the `< 1 s` comment is accurate, and the budget has ~14× headroom, not 40%

I tried to establish which of this issue's two branches was true ("the expectation was always wrong" vs "activation regressed"). **Neither is.** Five matched arms on the same Mac, and the ping never once approached 3 s:

[...]

PR #2142, opened 2026-08-24
Logon actions: per-user work at sign-in, run as the user

rbuergi:
## 🔴 Read this first: the root cause of "app icons are still super generic"

**This is independent of the framework below, and it would have made both #2128 and this PR inert.**

`UserActivityLayoutAreas.BuildAppRecord` writes a default app record with `MainNode = <app id>` —
and for a default record the node's own `Id` **is** that app id (`Store`, `Doc`, `Chess`). Then
`HandleCreateNodeRequest` step 1b' re-stamps any non-satellite node whose `MainNode == Id` to its own
full path, to stop a stale bare-id MainNode routing a thread into a phantom partition (the 42P01
bug). That guard is correct and stays.

The consequence: **every default app record arrives in the database pointing at itself.** So
`TargetOf(record)`, reading `MainNode`, returns `null` for **precisely the records wearing the
generic puzzle-piece icon** — the ones the repair exists for. Any icon-adoption pass built on
`MainNode` queries, decides "nothing to adopt", writes nothing, and reports success. Forever.

Fix: `AppIconAdoption.TargetOf` now reads **`App.Plugin` from the record's content first** (documented
as "path of the app's root node … the app's identity", and untouched by that re-stamp), with
`MainNode` as the fallback. Pinned by
`The_content_is_the_reliable_target_because_the_create_pipeline_rewrites_MainNode`.

Found only because the integration test was made to assert its preconditions instead of timing out —
the first version just hung for 30 s and said nothing.

---

## What the rest of this is

A **logon action** framework: work the platform runs *for one user, at logon, as that user*,
declaring either **run-once-per-user** or **every-logon**.

It exists because the platform had no way to give an **existing** user something new.
`INodePostCreationHandler` fires at account creation and can never fire again, so every "existing
users need this too" became a hand-written SQL backfill (`V29_PinDocsForExistingUsers`,
`V33_SeedChatInputForExistingUsers`) — raw `UPDATE`s that bypass the workspace cache, run once per
*deployment* rather than per user, and can only be written by someone shipping a `DbVersion` bump.

Docs: `Doc/Architecture/LogonActions` · Skill: `/logon-action`

## Design, and why this shape

| Decision | Why |
|---|---|
| **Ledger = `User.CompletedLogonActions`**, a dictionary on the profile | Durable and replicated for free. A **dict** because a cross-hub merge patch makes `SetItem` safe and a list append lossy (`RequestViaStreamUpdate` → Cross-Hub Patch Semantics) |
| **Effect + ledger entry in ONE `stream.Update` patch** | No window where the effect landed and the record did not (a restart re-applies over the user's later edits) or vice-versa (skipped forever) |
| **Guard *inside* the update lambda** | Two concurrent logons both reach the commit; the owner serialises, refuses the second as stale, and the rebase re-runs the lambda against state that already has the key → no-op. Same claim-before-mutate shape as `ThreadSubmissionServer`'s `Idle → StartingExecution` |
| **No `XxxRequest`/`XxxResponse`** | `stream.Update` is the only mutation API; the run happens in the caller's own flow, so no trigger message is needed at all |
| **Trigger = `UserContextMiddleware`** | The only hook where cookie/OAuth *and* Bearer both land with a resolved mesh identity. Rides the existing 5-minute session dedup — a second dedup dictionary would be a second definition of "a logon". The Blazor circuit handler fires per *tab* and misses the API surface |
| **Runs as the USER** | `access.RunAs(identity, …)`, re-established at each write (`Concat` subscribes on the previous action's completion thread). Never `Observable.Using(ImpersonateAsSystem, …)` — #1790 |
| **Actions are DATA** | Zero action nodes ship. Deployment-specific by construction |

Two ways to contribute one: a **`LogonAction` node** at `Admin/_LogonAction/{id}` (admin creates it
in-platform; no code, no image roll) — or **`ILogonAction` + `AddLogonAction<…>()`** for platform
behaviour that belongs on every portal.

## The concrete job

```
UNPIN: Doc/Architecture · Doc/DataMesh · Doc/GUI · Doc/AI
PIN:   AgenticPrimer · AgenticEngineering · AgenticBusiness
```

Doc paths confirmed from two agreeing sources (`UserOnboardingService.cs:76` and
`V29_PinDocsForExistingUsers.cs:21`). Course paths confirmed by reading each node back off the live
mesh — `path == id == mainNode` on all three.

> ⚠️ **`Store/AgenticPrimer` does not exist.** Their `nodeType` is `Store/Plugin`; a type name that
> looks like a path segment is a live trap that was written wrong once already. The failure is
> quiet — the existence check skips the bad path, the migration pins nothing, records itself as
> done, and never runs again. Called out in the doc, the skill, and the test's XML docs.

**The tests use these exact paths**, not placeholders — which is what makes them say something about
the shipped migration. This test mesh carries neither the doc sections nor the courses, so
`A_deployment_without_the_target_nodes_pins_nothing` runs the **real declaration** on a portal that
lacks all three targets: the systemorph.com case verbatim, asserting an empty pin list and no
dangling path. `A_declaration_NODE_in_the_admin_partition_is_discovered_and_run` creates a node with
the same id, description and lists as the production one — a rehearsal of the real create.

## 🚧 The production node is NOT yet created — blocked, by design

```
get  @Admin/_LogonAction/docs-to-courses   →  Not found
create …                                   →  ERROR: NodeType 'LogonAction' is not registered
get  @Admin/_LogonAction/docs-to-courses   →  Not found        (unchanged — nothing was written)
```

The `LogonAction` NodeType ships **in this PR**, so memex cannot accept the declaration until this
merges and the image rolls. That is the correct ordering and not something to work around. Once
`memex-portal-ai` carries this commit:

```
create @Admin/_LogonAction/docs-to-courses
{"nodeType":"LogonAction","name":"Swap the documentation pins for the courses","state":"Active",
 "content":{"$type":"LogonAction","mode":"RunOnce","order":0,"enabled":true,
   "description":"Existing users pinned the four documentation sections at onboarding; pin the three agentic courses instead, once per user.",
   "unpinPaths":["Doc/Architecture","Doc/DataMesh","Doc/GUI","Doc/AI"],
   "pinPaths":["AgenticPrimer","AgenticEngineering","AgenticBusiness"],
   "requireTargetsExist":true}}
```

Every user then gets it on their next sign-in, once.

## Icon adoption (closes #2128)

Rehomed from a hub-init hook to an **`EveryLogon`** action. Run-once would repair whatever was
installed the day it first ran and leave every later install on the placeholder forever; the cost of
running every time is bounded by `NeedsIcon` — one query, zero writes in the steady state. And
initialization carries no viewer context, so the hub-init shape needed `ImpersonateAsSystem` to do
anything at all — on records that belong to the user. A logon action has a real identity by
construction; `It_writes_as_the_user_never_as_system` asserts `lastModifiedBy != system-security`.

## Tests

`test/MeshWeaver.Graph.Test/{LogonActionFrameworkTest,LogonPinMigrationTest,AppIconAdoptionTest}.cs`
— `MonolithMeshTestBase`, nothing mocked, every assertion on observable state (the profile), never a
log line. **28 new tests; full project green.**

Run-once runs on the first logon and not the second · every-logon runs on both · **two concurrent
logons run a once-action exactly once** (a real race — both subscribed before either completes) · a
durable ledger survives a "restart" · a failing action is not recorded and does not stop the others ·
anonymous runs nothing · the pin swap happens once and a later hand-pin is not clobbered · **the
real declaration on a portal without the courses pins nothing** · partially-present targets pin only
what exists · a declaration NODE is discovered and run · a portal declaring nothing is a clean no-op
· icon adoption writes as the user · content-vs-`MainNode` target resolution.

## Not done, deliberately

- **Onboarding still seeds the four doc pins.** No change needed: a new user's first logon *is* a
  logon, so the same action covers them — and a second, onboarding-time source of truth for the same
  decision would be the drift this framework removes.
- **No new localization keys** — every user-visible string is a `[Description]`/`[Translation]` pair
  on a declaration, so there is nothing to mirror into `clients/react/src/i18n/`.

## Weakness worth naming

The atomic effect-plus-ledger patch only covers the **profile**. A run-once action whose real effect
is on *other* nodes gets its ledger entry written after the effect, which makes it at-least-once —
such an action must be idempotent in its own right. Stated in the doc, the skill and the XML docs.
The shipped icon action is exactly that case, which is one more reason it is `EveryLogon` and carries
no ledger at all.


rbuergi:
## CI attribution: `PartitionRootActivationTest.BarePartitionPath_NoMeshNode_RespondsToPing`

**Measured, not assumed — 30/30 pass locally** against the merged tree (branch + current `main`), `-c Release`, one process per iteration. Zero failures, so there is no failure log to attach.

I did *not* stop at "could my diff reach it". It can: that silo's `ConfigurePortalMesh` calls `AddGraph()`, so my `AddLogonActionType()` runs there. What it adds is one static NodeType node and two DI singletons that nothing in this test ever resolves (`LogonActionRunner` is only resolved by `UserContextMiddleware`; `AppIconAdoptionLogonAction` only by the runner). None of that is on the bare-partition grain-activation path this test exercises.

**Prior art matches the failure signature exactly.** [#1289](https://github.com/Systemorph/MeshWeaver/issues/1289) documents this test as *"a permanent contributor to the intermittent shard-0 Orleans red"*, failing at **exactly `FastBudget`** (measured 5.013 s) because the ping consumed ~60% of its 5 s budget on an *idle* box. Its own words: *"A test that consistently uses 60% of its budget on an idle machine will fail whenever CI is loaded."* That is a budget expiry, not the 30 s grain-strand condition the test was written to detect — so when it fails this way it fails for a reason unrelated to its purpose.

**Caveat, stated plainly:** 30/30 local green is weaker evidence than a CI re-sample. My box is not a loaded CI runner, and a load-sensitive budget is precisely what a quiet machine cannot reproduce. I have merged current `main` (carrying #2136 and #2141), so this push re-samples on CI — that is the real test of the attribution, not my loop.

I have not touched the test, its budget, or anything Orleans. Re-running to "see" would have been the wrong move in either direction.

---

## Review finding — fixed properly (`d94e40e`)

The finding was right, and there were **two** defects in it rather than one: the XML doc claimed *"one query for the whole set"* while `Select(…Query…).Merge()` fanned out one query **per target**, on the **logon path**, for every user who signs in.

Fixed the **code to match the doc**, not the reverse — path alternation (`path:a|b|c`), one round trip whatever the declaration lists, the same construction `UserActivityLayoutAreas` already uses for its shared-target and pinned-path lookups.

**Alternation introduces one risk the fan-out did not have**, and it is worth naming: it is a *match expression*, so a row for a **descendant** of a declared target lands in the same result set. Reading "the query returned something" as "the target exists" would pin a path that does not — the dangling tile the existence check exists to prevent, arrived at from the other direction. The intersection is therefore on **exact path equality**, pinned by a new test: a course present with a child under it, two courses absent → exactly one pin, and explicitly not the child.

**Absent vs. failed stay distinct**, as asked, and the doc now says so where the code decides it:

| outcome | meaning | ledger |
|---|---|---|
| empty result | this deployment carries none of them → pin nothing | **recorded** (the migration ran) |
| query throws | could not be established | **not recorded** → retried next logon |

Collapsing those would burn the single attempt a run-once action gets and silently pin nothing, forever. A path containing `|` is now declined with a warning rather than splitting into two bogus terms that could match something nobody declared.

Coverage still pins all three cases with the **real** declaration — all three present, partial (one present), none present (the systemorph.com case) — plus the new descendant case: 8 tests in `LogonPinMigrationTest`.

Also folded in: the middleware hook now handles its stream fault **fluently** per `/async` rule 1b. It previously had a `try`/`catch` around `Subscribe` *and* an `onError` arm, with a comment that muddled which covered what — a fault travelling through the stream arrives later, on another thread, and sails past the catch. `.Catch` is now the stream's handler (`Observable.Empty` — "this did not happen", so `OnNext` cannot run for a failed round) and the `try`/`catch` is scoped, in the comment, to the synchronous `Subscribe` throw it actually covers.

**Verification:** `MeshWeaver.Graph.Test` **1571/1571**, fresh `.trx`. `-c Release -warnaserror` on Graph, Blazor, Mesh.Contract, Documentation, Hosting.Blazor, and the Graph / Documentation / AI / Orleans / Hosting.Blazor **test** projects — one project per invocation.

Issue #640, opened 2026-07-24
StaticRepoImporter: incremental skip fires on an EMPTY partition (phantom manifest checkpoint)


rbuergi:
A failed/ghost-era import left a checkpoint that made subsequent real imports report 'Imported Skipped (0 node(s))' against a partition with no content. Skip must verify the partition actually holds the manifest's content (verify-before-skip).

From the 2026-07-24 plugin-install retrospective.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

rbuergi:
Fixed on main by the content-sentinel verify-before-skip that landed with #708 (StaticRepoImporter.cs — "SELF-HEAL CONTENT, not just governance"): before honoring a Succeeded fingerprint marker, the importer now queries a content sentinel node from the manifest; if it is missing (the empty-partition / phantom-checkpoint case this issue describes), it falls through to a full idempotent re-import instead of reporting 'Skipped (0 nodes)'. A forced import additionally always re-runs regardless of fingerprint.

Pinned by `StaticRepoImporterTests.Reimport_SameFingerprint_ButContentDeleted_SelfHeals` (test/MeshWeaver.Hosting.PostgreSql.Test): Succeeded marker + deleted content → re-import restores the partition; the old behavior skipped forever.

Issue #637, opened 2026-07-24
Auth: storage/read timeouts surface as 401 Unauthorized


rbuergi:
During silo degradation, ApiToken validation reads timed out (8s) and every authenticated call answered 401 — indistinguishable from an invalid token. Cost hours of misdiagnosis + a forced re-auth. 'Cannot validate (retry/503)' must be distinct from 'invalid (401)'.

From the 2026-07-24 plugin-install retrospective.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

rbuergi:
Status audit against current main (2026-08-08) — still OPEN for the client-visible half; the server-side half improved.

**What changed since filing (mitigation, not the asked fix):** validation moved off the cross-silo mesh read onto a direct IStorageAdapter.Read with absence as a terminal negative (ApiTokenService.cs:342-374), and the handler logs hash prefix + elapsed ms with an explicit 'never a silent 401' note (ApiTokenAuthenticationHandler.cs:45-53) — an operator can now tell timeout from invalid in the logs, and the 8s window is much less likely.

**What remains (the issue's actual ask):** a read TIMEOUT still collapses into the same null as 'no such token' (ApiTokenService.cs:397-406) and answers 401 'Invalid or expired API token' (ApiTokenAuthenticationHandler.cs:43-55) — identical to a forged token. Same shape on UserContextMiddleware.cs:315-325, SignalRConnectionRegistry.cs:99, GrpcConnectionRegistry.cs:71/215: infrastructure faults degrade to Anonymous. No 503/Retry-After path exists anywhere in auth, and no test asserts a retryable outcome. Clients (and their retry logic) still cannot distinguish 'cannot validate right now' from 'invalid'.

rbuergi:
Closing — both legs of this issue are now fixed, and the remainder is tracked separately as **#974** rather than left implicit here.

**The API-token leg** was fixed by **#947**: tri-state `TokenValidationStatus` at the `ApiTokenService` seam, a 503 + `Retry-After` via `HandleChallengeAsync`, and the matching changes in `UserContextMiddleware`'s bearer branch, the SignalR and gRPC registries, `ApiTokenNodeType` and `ValidateTokenResponse.IsUnavailable`.

**The identity/authorization leg** was fixed by **#970**, and it was the worse half — it is what a *browser* user hits:

[...]

PR #2192, opened 2026-08-25
Convergence opt-in: auto-recycle instances onto a newly published build

rbuergi:
## What changed

`Modules:AutoRecycleOnStaleBuild` (default **off**) turns the stale-build *offer* into an automatic self-recycle: when a NodeType publishes a usable build whose assembly path differs from the one an instance bound, the instance posts its own `DisposeRequest` instead of hanging the *"a newer build of this type is available — Recycle"* banner and waiting for a viewer's click. The banner remains the default everywhere that does not opt in.

## Why

memex 2026-08-25: a Store package update recompiled every type green while **all serving instances stayed on the previous assembly** behind the banner — mixed old/new assemblies (`$type` registration mismatches in the logs), and the store page wedged until type and instance hubs were recycled by hand. A production portal's invariant is "after an update, everything serves the new build"; this key states that policy per deployment.

The unguarded auto-recycle that once lived at this exact seam was removed because it fired per publication. The opted-in branch sits behind the guards that did not exist then — the assembly-path gate (unrelated node writes cost nothing), the settle-window throttle (a publish burst collapses to one recycle, the #1343 regression guard), and `Take(1)` per binding — so "publication frequency = restart frequency" cannot recur.

## How tested

- `StaleAssemblySelfHealWatcherTest`: two new pinned contracts — opted-in fires exactly one self-`DisposeRequest` targeting the instance's own address with **no** banner offer (and not before the settle window); same-assembly republication recycles nothing even as versions advance. All 21 self-heal watcher tests green.
- `dotnet build -c Release -p:CIRun=true -warnaserror` clean on MeshWeaver.Graph.

## Rollout

No behaviour change until a deployment sets the key. Production portals (memex-cloud, systemorph) should set `Modules:AutoRecycleOnStaleBuild: true` — flagged to the maintainer as a config/Helm change, not part of this PR.

### DIFF
--- a/src/MeshWeaver.Documentation/Data/WhatsNew/2026-08-25-auto-recycle-on-stale-build.md
+++ b/src/MeshWeaver.Documentation/Data/WhatsNew/2026-08-25-auto-recycle-on-stale-build.md
@@ -1,0 +1,24 @@
+---
+Name: Production converges onto new builds by itself
+Category: Fix
+Description: A deployment can now opt in (Modules:AutoRecycleOnStaleBuild) to have every page automatically pick up a freshly compiled module build — no more "a newer build is available" banners lingering on production while pages serve the previous code.
+Icon: ArrowSync
+Order: -20260825
+---
+
+# Production converges onto new builds by itself
+
+When a module updates on a live mesh, its types recompile green — but until now every page that
+was already running kept executing the **previous** build behind a *"a newer build of this type is
+available — Recycle"* banner, until someone clicked. On a production portal that inversion is an
+outage in waiting: after this morning's Store update, the store page served a mixture of old and
+new code until each hub was recycled by hand.
+
+Deployments can now opt into **convergence**: with `Modules:AutoRecycleOnStaleBuild: true`, an
+instance whose type published a genuinely different build recycles itself — once, only after the
+publish burst has settled, and never for node writes that carry the same assembly. The banner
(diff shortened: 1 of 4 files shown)


rbuergi:
Measured evidence for the motivation (peer session meshweaver-c0, `container_memory_working_set_bytes`, memex-cloud, 2026-08-25 UTC) — the MEMORY half of the non-convergence this PR closes:

- The two type-hosting silo pods (born 01:26 from a portal roll) sat flat at 2.7–4.5 GB for five hours.
- Inflection 06:10–06:30, right after the 05:55 scheduled bake tick — the first after #634 bumped the platform pin (new framework identity → full-catalog rebake → adoption → recompile of every type on those silos).
- Then ~2.5 GB/h climb through the #635 and #637–#640 auto-synced recompiles. Peak **20.4 GB / 17.2 GB with ~4 cores of GC**; later-born pods stayed ≤3.6 GB.
- No OOMKill (28Gi limit never reached), and liveness/readiness probe /alive (trivial), so the sick pods stayed in rotation ~3.5 h.

Mechanism: **superseded AssemblyLoadContexts stay ROOTED by serving instances that never converge onto the new build.** The mixed-assembly $type wedge was the user-visible half; this retention is the memory half. `Modules:AutoRecycleOnStaleBuild` releases the last roots of each superseded ALC as instances converge. Rollout (prod chart/values for both portals) is with the peer session; the progress-aware /alive probe is recorded as an RCA open item for the maintainer.

PR #2240, opened 2026-08-25
Chat stops guessing: fail the round on a lost history, name the provider's refusal (#2226, #2233)

### LINKED ISSUE #2226: AgentChatClient falls back to empty history when thread-history load times out
## What is failing

`MeshWeaver.AI.AgentChatClient` times out while loading a thread's prior conversation history and, after catching the `TimeoutException`, continues with an **empty** history. The agent then sees only the new user message and loses every prior turn on that thread. The failure is logged at `fail:` level as `[ThreadExec] HISTORY_LOAD_FAILED` but no exception propagates — from the caller's point of view the turn silently loses context.

## Probable cause

A plain `System.TimeoutException` with no inner exception and no application stack frame — the history-load I/O path (grain call / mesh store read) is exceeding its configured timeout. Every sample threadPath is a long-running conversation (LogTriage sub-threads under `Admin/_LogIncident/.../_Thread/...`, and a `Posts/RobertSilentCyber/_Thread/...` with substantial history), which is consistent with the load being size- or latency-bound rather than a code defect. **Confidence: moderate** on "slow I/O against a large-history thread"; low on the exact subsystem, because no stack frame was captured.

## Impact

Low-to-moderate. 7 occurrences over ~3 days (2026-08-22 → 2026-08-25) across both `memex-portal` pods in the `memex` namespace — not a burst, a steady trickle. No crash: the timeout is swallowed and the turn proceeds. The user-visible effect is quality, not availability: the agent answers without the context it should have had, so triage sub-threads in particular may repeat work or contradict earlier turns in the same thread.

## Where to look

- `MeshWeaver.AI.AgentChatClient` — the `[ThreadExec]` code path that catches `TimeoutException` around history load and logs `HISTORY_LOAD_FAILED`. The right fix is probably *not* "swallow and continue with empty" — at minimum the timeout should be surfaced to the caller, or retried, before context is dropped.
- The history-load call it wraps (thread message retrieval / grain read) — check the configured timeout and whether it scales with thread size, and whether a paged / streaming load would avoid loading the whole history synchronously on every turn.
- Worth confirming whether the affected long threads share a storage partition or grain that is slow specifically for them.

---

### Evidence

| | |
|---|---|
| Fingerprint | `dd7925893af6fa42` |
| Category | `MeshWeaver.AI.AgentChatClient` |
| Severity | Error |
| Exception | `System.TimeoutException` |
| Namespace | `memex` |
| Pods | `memex-portal-deployment-7db656d9d9-nbfzk`, `memex-portal-deployment-858f57dc6-8zlfh` |
| Occurrences | 7 |
| First seen | 2026-08-22 13:49:58Z |
| Last seen | 2026-08-25 09:07:12Z |

```
2026-08-22 13:49:58Z memex-portal-deployment-7db656d9d9-nbfzk fail: MeshWeaver.AI.AgentChatClient[0]
      [ThreadExec] HISTORY_LOAD_FAILED threadPath=Admin/_LogIncident/90a8113061a020a2/_Thread/triage-the-log-incident-at-admin-loginci-4bde — proceeding with EMPTY history; agent will see only the new user message
      System.TimeoutException: The operation has timed out.
2026-08-22 13:50:37Z memex-portal-deployment-7db656d9d9-nbfzk fail: MeshWeaver.AI.AgentChatClient[0]
      [ThreadExec] HISTORY_LOAD_FAILED threadPath=Admin/_LogIncident/90a8113061a020a2/_Thread/triage-the-log-incident-at-admin-loginci-2578 — proceeding with EMPTY history; agent will see only the new user message
      System.TimeoutException: The operation has timed out.
2026-08-22 13:50:51Z memex-portal-deployment-7db656d9d9-nbfzk fail: MeshWeaver.AI.AgentChatClient[0]
      [ThreadExec] HISTORY_LOAD_FAILED threadPath=Admin/_LogIncident/61c62838af6a2330/_Thread/triage-the-log-incident-at-admin-loginci-b399 — proceeding with EMPTY history; agent will see only the new user message
      System.TimeoutException: The operation has timed out.
2026-08-22 14:04:29Z memex-portal-deployment-7db656d9d9-nbfzk fail: MeshWeaver.AI.AgentChatClient[0]
      [ThreadExec] HISTORY_LOAD_FAILED threadPath=Admin/_LogIncident/ab4fea6c6a997d4e/_Thread/triage-the-log-incident-at-admin-loginci-82bb — proceeding with EMPTY history; agent will see only the new user message
      System.TimeoutException: The operation has timed out.
2026-08-22 14:07:18Z memex-portal-deployment-7db656d9d9-nbfzk fail: MeshWeaver.AI.AgentChatClient[0]
      [ThreadExec] HISTORY_LOAD_FAILED threadPath=Admin/_LogIncident/67d15cc4e57e97cc/_Thread/triage-the-log-incident-at-admin-loginci-922c — proceeding with EMPTY history; agent will see only the new user message
      System.TimeoutException: The operation has timed out.
2026-08-25 08:46:14Z memex-portal-deployment-858f57dc6-8zlfh fail: MeshWeaver.AI.AgentChatClient[0]
      [ThreadExec] HISTORY_LOAD_FAILED threadPath=Posts/RobertSilentCyber/_Thread/that-did-not-happen-because-the-market-v-0d27 — proceeding with EMPTY history; agent will see only the new user message
      System.TimeoutException: The operation has timed out.
2026-08-25 09:07:12Z memex-portal-deployment-858f57dc6-8zlfh fail: MeshWeaver.AI.AgentChatClient[0]
      [ThreadExec] HISTORY_LOAD_FAILED threadPath=Posts/RobertSilentCyber/_Thread/that-did-not-happen-because-the-market-v-0d27 — proceeding with EMPTY history; agent will see only the new user message
      System.TimeoutException: The operation has timed out.
```

</details>

<sub>Opened automatically from `Admin/_LogIncident/dd7925893af6fa42`. Recurrences are folded into this issue rather than opening new ones.</sub>

rbuergi:
Fixes #2226, #2227, #2228, #2233.

Two halves of one defect: `AgentChatClient` could not tell **"no history"** from **"could not load the history"**, nor **"the model said no"** from **"we crashed"**.

## #2226 / #2227 / #2228 — the empty-history fallback (a silent wrong answer)

`LoadFullConversationHistoryFromMesh` already refuses to return an empty list when cells were expected — it throws. The bug lived one level up: `ThreadExecution` **caught that refusal and continued with `Array.Empty<ChatMessage>()`**, logging `HISTORY_LOAD_FAILED … proceeding with EMPTY history`. The agent then answered a long-running thread as though it had just been created, and the round settled **`Completed`** — so neither the user, nor a delegating parent, nor any automation reading the node could tell a context-less answer from a correct one. Seven such rounds shipped wrong answers over three days in production.

The fault is now carried forward and terminates the round exactly like the existing `NO_USABLE_MODEL` branch: response cell `Status = Error` via `PushToResponseMessage`, thread → `Idle`, `NotifyParentCompletion` + `EmitCompletionNotification`, localized prose for the user and the exception on the `LogError` for the operator. No new request/response type; no new mutation surface.

### Why the load was timing out at all

The caller wrapped the loader in `.Timeout(TimeSpan.FromSeconds(5))` — while the loader's **own** staged budget is 10 s for the thread node plus 5 s per cell, fanned out in parallel. The outer bound was **strictly smaller than the budget it wrapped**, so on exactly the long/cold threads the issue reports it pre-empted a load that was still legitimately in progress, and replaced the loader's descriptive failure with `.Timeout()`'s message-less `"The operation has timed out."` — the bare `TimeoutException` with *no inner exception and no application stack frame* that #2226 was opened on.

That contradictory outer bound is **deleted**. The inner budgets are untouched — nothing was widened to make anything pass.

## #2233 — the unreadable upstream refusal

`ProviderFailureClassifier` knew only Azure's `"Status: 429 (…)"` banner. `System.ClientModel` renders **`"HTTP 402 (: )"`**, so *every* OpenAI/OpenRouter refusal fell through unclassified and pasted its raw English transport dump into the thread — the very defect #476 set out to kill, still live on the code path the portal actually uses.

The classifier now reads both banners (and rejects `"HTTP/1.1"` protocol versions in header dumps). **402** and **404** get their own conditions rather than being folded into 429, because the remedies differ: 402 needs credit, 404 needs a config change, and neither is fixed by "submit again later".

## Verified to fail without the fix

Both new round tests were run against a reverted `ThreadExecution` to prove they discriminate:

- `ClientResultRefusalRoundTest` → `Expected "*Error: HTTP 402 (: )` / `"*Error: HTTP 404 (: 404)` — the raw dump, reproduced verbatim.
- `HistoryLoadFailureRoundTest` → `Expected value to be Error … but found **Completed**` — the silent wrong answer, reproduced exactly.

## Tests

- `ProviderFailureClassifierTest` — production-verbatim 402/404 dumps classify; inner-chain walk; `HTTP/1.1` and over-long digit runs invent nothing.
- `ClientResultRefusalRoundTest` (new) — 402 and 404 each end the round `Status = Error` with the **localized** condition in cell text *and* thread summary, with the provider's body (credit figures, top-up URL) asserted **absent**, and the thread settled `Idle`.
- `HistoryLoadFailureRoundTest` (new) — a history load that fails ends the round `Status = Error` with the localized reason, and **the agent is never invoked** (invocation count `0`) — the assertion that pins "did not answer with empty history".
- Whole `MeshWeaver.Threading.Test` project green (136/136); `MeshWeaver.AI.Test` refusal + classifier suites green (26/26, including the pre-existing 429 `ProviderRefusalRoundTest`).
- `LocalizationTest` 56/56, `WhatsNewEntryIntegrityTest` + `DocumentationLinkIntegrityTest` green.

## i18n

Three new keys — `chat.historyLoadFailed`, `chat.modelQuotaExhausted`, `chat.modelNotFound` — in **en + de**, mirrored byte-identically into `clients/react/src/i18n/`. The React drift guard was actually run: `vitest src/i18n/localize.test.ts` → 15/15 passed.

Resolution is off the round's own `AccessContext.Locale` throughout — never ambient `CultureInfo`.

## Build gate

`dotnet build -c Release -warnaserror` clean (`0 Warning(s), 0 Error(s)`) for `MeshWeaver.AI`, `MeshWeaver.AI.Test`, `MeshWeaver.Threading.Test`, `MeshWeaver.Messaging.Hub.Test`, `MeshWeaver.Documentation.Test`.

## Not addressed (deliberately)

- **The `max_tokens=65536` request shaping** #2233 mentions as a code smell. Clamping it would reduce how often 402 trips, but it is a separate scope call about request budgets — not the error-handling defect, and shrinking it to dodge the refusal would be treating the symptom.
- **`ThreadExecution.LoadPriorUserMessagesFromMesh`** is dead code (zero callers) that carries the same swallow-to-empty-history shape. Left untouched to keep this diff to the live defect; worth deleting separately.


rbuergi:
Follow-up commit `f95683be`: **dispose the harness client on the new early return.**

`harnessClient` is created *above* the history load (line ~1507) and disposed only in the round's `finally` — which the new early return skips, so every failed history load would have leaked one CLI harness client (and, for the process-backed harnesses, the process behind it). Now disposed in the branch.

Worth knowing for review: the `NO_USABLE_MODEL` early return is exempt by construction (its guard *is* `harnessClient == null`), but the **`NOTHING_TO_SEND` return a few lines below is not guarded and leaks the same way**. That is pre-existing and on a path this change does not touch, so I flagged it rather than fixing it silently — happy to do it here or in a follow-up, your call.

### Local verification after the amendment
- `MeshWeaver.Threading.Test` — **136/136 pass**
- `MeshWeaver.AI.Test` (full project, not just my filter) — **1279 passed, 3 skipped, 0 failed**, which includes `ThreadTokenUsageTest.CancelledRound_KeysUsageAndCellByActualModel_NotRequestedAlias` — the test currently flaking on main (run `32876073965`). It passes locally against this branch, consistent with the flake attribution rather than anything in this diff.
- `dotnet build -c Release -warnaserror` clean on `MeshWeaver.AI` and both test projects.

### On the AI.Test parallel-safety hazard (#1040)
`ClientResultRefusalRoundTest` lives in that assembly, so it was written to add nothing to the shared-state population: **zero `static` declarations** (the only occurrences of the word are comments saying "never static"), no `Clear()`-for-isolation, and the scripted-refusal knob is a plain instance field on the test class read through a closure — one mesh per theory case. The seeded node names (`ClientResultRefusalProvider`, `clientresult-refusing-model`) are uniquely prefixed and used by no other class, so they cannot collide even if state does leak across classes.

Issue #2229, opened 2026-08-25
Write lane degraded on memex-cloud: create chain dies on a NotFound stream, a finished Store/Provision storms compile-state, plus a permanent-silent import order bug and a 2/s identity-cache loop


rbuergi:
Four defects found while investigating a write-lane degradation on memex-cloud (2026-08-25, ~14:45–15:05 UTC). Reads and renders were healthy throughout; **writes** failed. Filed together because they were found in one incident, but they are independent and separately fixable — A and B are the outage, C and D are silent bugs the incident exposed.

Ruled out first, with evidence, so nobody re-treads it: this was **not** the fattened/draining pods. Both were wedged in `Terminating` (15 min past a 90 s app drain; `terminationGracePeriodSeconds=1800`, so they hold activations for half an hour) and were force-deleted — **writes failed identically afterwards**. It was also not the freshly-armed `Modules:AutoRecycleOnStaleBuild` (disproved separately on #2194).

---

## A. `CreateNodeRequest` never replies: the create chain subscribes to the node's stream before the node exists

The primary user-visible failure. Three independent reproductions (`create rbuergi/WriteProbe2224`, different MCP session hubs), each a 60 s timeout, and the node genuinely absent afterwards. The message-flow trace names the whole chain:

```
AWAITING CreateNodeRequest → POSTED → RECEIVED runLevel=Started → ENQUEUED
→ ROUTED onTarget=True → HANDLER_ENTER
→ CREATE_CHAIN_SUBSCRIBED path=rbuergi/WriteProbe2224
→ HANDLER_EXIT state=Processed
   … a handler was entered and no reply, completion or fault has been recorded since …
[ROUTE] NotFound: No node found at 'rbuergi/WriteProbe2224'. Closest ancestor is 'rbuergi' (remainder='WriteProbe2224').
[SYNC_STREAM] OnError for 6LGxECQAzE2EKtMwdHqhOA (Reference=MeshNodeReference { Path =  }, Owner=rbuergi/WriteProbe2224)
MeshWeaver.Messaging.DeliveryFailureException: No node found at 'rbuergi/WriteProbe2224'.
System.TimeoutException: No response received in hub portal/mcp-rbuergi-04ecb166 within 00:01:00 for request CreateNodeRequest
```

[...]

Issue #2001, opened 2026-08-21
ThreadTokenUsageTest is wall-clock-bounded but missing from ConcurrencyStressCollection — reddens unrelated PRs


rbuergi:
`ThreadTokenUsageTest` judges a cancel on a 10-second wall clock while sharing a 4-vCPU runner with three other AI.Test classes. It is not in `ConcurrencyStressCollection`, and it should be — or its wait should stop being wall-clocked.

Seen on #1991's CI (run 32490339312, shard 3), where it was the one failure in that PR's shard that the PR could not have caused.

## What failed

```
MeshWeaver.AI.Test.ThreadTokenUsageTest.CancelledRound_RecordsTokensConsumedBeforeCancel_OnCellAndThread [FAIL]
  Expected the observable to emit a value matching the predicate within 10s,
  but it did not. The observable emitted nothing at all.
    at ThreadTokenUsageTest.cs:262
```

Line 262 is the usage-satellite wait after the cancel:

```csharp
await WaitForUsage(threadPath, "usage_cancel_model",
    u => u.InputTokens == InTokens && u.OutputTokens == OutTokens, 10_000);
```

It **ran out of time**. It did not read a wrong value — the preceding `WaitForThread(... Cancelled, 20_000)` had already passed, so the cancel itself landed; only the satellite had not appeared inside its own, tighter 10s budget.

## Why it is structural, not luck

`8807a6d8c` ("Give the five machine-hungry AI.Test classes the box, and parallelise the rest", 2026-08-10) turned parallelism **on** for the whole assembly:

```json
{ "parallelizeAssembly": false, "parallelizeTestCollections": true,
  "maxParallelThreads": 4, "methodTimeout": 30000 }
```

…and serialised exactly five classes into `ConcurrencyStressCollection`: `AgentChatClientDeadlockTest`, `CrossHubPatchAtomicityTest`, `DelegationDeadlockTest`, `InboxToolIntegrationTest`, `PatchWorkspaceAckTest`.

`ThreadTokenUsageTest` carries **no** `[Collection]` attribute, so since that commit it runs up to 4-wide on a 4-vCPU CI runner — while still judging a cancel-then-satellite round trip on a fixed 10 000 ms.

That commit's own message names this failure mode precisely:

> a handful of classes deliberately saturate the machine and then judge the result on a wall clock, and four of those timesharing one 4-vCPU runner blow bounds that hold with room to spare when each has the box

[...]

Issue #1812, opened 2026-08-17
ThreadTokenUsageTest reads the usage satellite through a LAGGED QUERY and can observe the zero-token phase-1 intermediate


rbuergi:
> ⚠️ **This issue has been corrected.** Its first version claimed production spend was going unaccounted. **That was wrong** — see *Correction* below. The defect is real but it is a TEST-side rule violation, not lost accounting. Corrected before anyone acted on it.

## Observed (CI run [32070434174](https://github.com/Systemorph/MeshWeaver/actions/runs/32070434174), shard 4, PR #1806)

```
MeshWeaver.AI.Test.ThreadTokenUsageTest.ErrorRound_KeysUsageAndCellByActualModel_NotRequestedAlias [FAIL]
  Expected the observable to emit a value matching the predicate within 10s.
  Last of 1 emission(s) was:
    TokenUsage { Model = usage-error-resolved, InputTokens = 0, OutputTokens = 0, … }
```

Predicate (`ThreadTokenUsageTest.cs:298`): `u.InputTokens == InTokens && u.OutputTokens == OutTokens`.

## Mechanism — the zero-token intermediate is BY DESIGN

`TokenUsageNodeType.RecordUsage` (`src/MeshWeaver.AI/TokenUsage.cs:138`) writes in **two deliberate phases**, and its own comments explain why (avoiding a point-read of an absent node, which trips the `MeshNodeStreamCache` storm breaker):

- **Phase 1** — `meshService.CreateNode(freshNode)` creates a **ZERO-token** satellite carrying the correct resolved `Model`.
- **Phase 2** — `GetMeshNodeStream(usagePath).Update(...)` accumulates the real counts off the OWNER's authoritative stream.

So `{ Model = usage-error-resolved, InputTokens = 0 }` is precisely the **phase-1 snapshot**. A reader that observes between the phases sees exactly what CI saw.

## The actual defect: the test reads through a lagged index

`WaitForUsage` (`ThreadTokenUsageTest.cs:411`) reads via **`Mesh.GetQuery(...)`** — the eventually-consistent CQRS index — to obtain **one node's content by exact path**. AGENTS.md forbids this shape outright:

> ❌ WRONG — lagged index, stale after writes … **Wrong: reading content by exact path**
> ✅ CORRECT — `workspace.GetMeshNodeStream(path)`

[...]

PR #2244, opened 2026-08-25
Read an optional node with a listing AND a stream, not one or the other

rbuergi:
## The one-line finding

`ThreadTokenUsageTest` has been declared fixed three times and failed a fourth (run [32876073965](https://github.com/Systemorph/MeshWeaver/actions/runs/32876073965) turned **main** red at `dcba079ce`). **Every failure across #1040, #1812, #2001 and today is a `WaitForUsage` — never the `WaitForThread` or `WaitForCell` sitting in the same test methods on the identical budget.** That asymmetry rules out generic runner starvation and points at the one thing the usage read does differently: it reads **one known node's CONTENT through an eventually-consistent query**.

## Why the previous fixes could not hold

Reading a node that **may not exist yet** has two rules pulling opposite ways, and each earlier fix picked one horn and got bitten by the other:

| Do this | And you hit |
|---|---|
| point `GetMeshNodeStream(exactPath)` at an **absent** node | The owner answers an authoritative routing **NotFound**, which **terminates the stream with an error** — it cannot wait for the node to appear. Worse, that NotFound opens `MeshNodeStreamCache`'s **storm-breaker** window on the path, and the breaker **fast-fails WRITES too** — so the read suppresses the write it is waiting for. This is #1040's `No node found at …/_Usage/…`: an *error*, not a timeout, so no budget could have saved it. It is why `02b851fd6` moved these tests onto a query. |
| read a known path's **`Content`** out of a query | The index is eventually consistent. The `CqrsAndContentAccess` tightening that landed hours before this failure (#2229 → #2232) puts a number on it: a query's answer for one path can be **minutes** old. Waiting on a VALUE through it is waiting on an unbounded lag. |

The irony is worth stating plainly: **the doc-only commit that turned main red is the one that tightened AGENTS.md to say a query's answer for a known path can be minutes stale — describing exactly the read shape of the test that then failed on that run.**

## The fix — stop picking a horn, compose them

```csharp
GetQuery(id, "path:{parent} scope:children nodeType:X select:path")   // EXISTENCE — empty-on-absent
    .Where(nodes => nodes.Any(n => n.Path == target))
    .Take(1)
    .SelectMany(_ => workspace.GetMeshNodeStream(target))             // CONTENT — authoritative, live
    .Select(node => node.ContentAs<X>(hub.JsonSerializerOptions));
```

The listing answers only *is it there yet* — the one query use AGENTS.md still sanctions, empty-on-absent, where a stale negative merely waits a beat longer. The value comes off the **owner's** stream, opened only once the node demonstrably exists, and left live so rounds 2+ accumulate onto the same node.

The ordering is sound rather than merely convenient: **the index TRAILS the store**, so "the index has seen it" implies "the store has it". The point read opened on that signal can never be early and can never NotFound. The same lag that disqualifies a query for CONTENT is what makes it a safe gate for the point read.

Applied to **all three** copies of the shape — `ThreadTokenUsageTest`, `ModelSubstitutionTest`, `DelegationSubThreadUsageTest`. Leaving stragglers is how this came back three times carrying three mutually contradictory "🚨 do NOT switch this to X" comments.

## `UsageSatelliteIsNeverObservableWithZeroTokens` keeps its subject

A naive port would have gutted that test **silently**. Its claim is about what a *query-bound GUI reader* can observe; the new composition's point-read leg opens only after the index has seen the node, so a regressed zero window could land and be overwritten unseen and the test would pass having watched nothing.

It now does both: collects zero-snapshots from a deliberately query-shaped reader (byte for byte what `ThreadTokenChip` binds to) while **settling** on the authoritative composition. Strictly stronger than before — the assertion keeps its subject, and the test's pass no longer hangs on that reader's lag.

## AGENTS.md is amended, because it is the artefact that made this hard

Its CQRS section states *"Never Query for a Single Node's Content"* **unqualified**, with a ❌/✅ pair that says to use the exact-path stream. The absent-node horn existed only as a parenthetical **484 lines into** `CqrsAndContentAccess.md`. An agent following AGENTS.md literally writes the point read and storm-breaks the writer — which is what happened. Both documents now carry the composition as a named pattern, and the primitives table gains a row for it.

## Also worth recording: #2001's fix was a band-aid

`3fcf79f8f` ("One wait budget in ThreadTokenUsageTest, and it is a hang-stopper") did exactly one thing: replaced every `10_000` with a `20_000` constant. Its commit message argues at length that this is *"NOT the widening AGENTS.md forbids"* — while #2001's own issue body had already written:

> **Not** raising 10 000 to 20 000. That is the band-aid `AGENTS.md` forbids.

The same assertion then failed at 20 s. Recorded so the next person does not spend the cycle re-deriving it.

## What is NOT claimed

**No local repro.** Under the documented lever (`DOTNET_PROCESSOR_COUNT=4`) over repeated full-assembly runs, the failure did not reproduce. A first attempt using four concurrent loops *did* produce failures — `AgentToolCancellationTest`, `MeshPluginTest` — but at load average 217 those are starvation artefacts, neither has ever appeared in CI, and I am not reporting them as members of #1384's population. So this rests on the read shape being wrong **by the repo's own rules**, not on catching it in the act.

**One writer-side hypothesis remains open and unproven**, and it is the inconvenient one: a cancelled round whose `RecordUsage` no-ops on zero tokens (Debug level) or whose create chain **completes without emitting** — silent, because an empty completion passes straight through the 15 s `Timeout` without tripping its deliberately-loud fail-open warning. That would mean **cancelled-round tokens go unbilled in production**. The CI log for the failing run shows no `[TokenUsage]` warning at all, which is consistent with both. It is written up on #1384 rather than guessed at here, and it is a separate product question from this test's read shape.

## Verification

- `dotnet build -c Release -warnaserror`: `MeshWeaver.AI.Test`, `MeshWeaver.Threading.Test`, `MeshWeaver.Documentation.Test` — **0 Warning(s), 0 Error(s)** each.
- `MeshWeaver.Documentation.Test` — 42/42 (covers `DocumentationLinkIntegrityTest` for the doc edits; `src/MeshWeaver.Documentation/Data/**` is `<EmbeddedResource>`, so the project was rebuilt first).
- `ThreadTokenUsageTest` + `ModelSubstitutionTest` — 12/12.
- `DelegationSubThreadUsageTest` — 1/1.
- Full `MeshWeaver.AI.Test` assembly under `DOTNET_PROCESSOR_COUNT=4`, repeated.

**No What's New entry**: tests and internal docs only, no user-visible change (per the `pullrequest` skill's pure-internal exemption — noted here deliberately).

Refs #1384, #1040, #1812, #2001.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -783,7 +783,17 @@ hub.GetQuery(id, $"path:{parent} scope:children nodeType:X select:path")   // EX
 The index **trails** the store, so "the index has seen it" implies "the store has it" — the point read
 opened on that signal can never be early, and never NotFounds. The same lag that disqualifies a query
 for CONTENT is what makes it a safe gate. Creating the node anyway? Skip the check entirely and use
-`CreateOrUpdateNodeRequest`. Full pattern: [CqrsAndContentAccess.md](src/MeshWeaver.Documentation/Data/Architecture/CqrsAndContentAccess.md) → "An OPTIONAL node".
+`CreateOrUpdateNodeRequest`.
+
+🚨 **This is about ONE known path whose value you are GATING on — not about node counts.** The
+worked counter-example is `src/MeshWeaver.Blazor.Portal/Chat/ThreadTokenChip.razor.cs:106`: it reads
+`content` out of the very same `{thread}/_Usage scope:children` query, **and it is correct** — it
+sums a SET to paint a total, and a briefly-stale total is cosmetic. Converting it to N point reads
+would mean N per-node hub activations per render, and on a legitimately EMPTY set (a thread with no
+rounds yet) every one is an absent-node read tripping the breaker **on the render path**. So: stale
+answer merely looks wrong on screen → query, `content` and all. Stale answer DECIDES whether
+something proceeds, passes, or is written → the owner's stream. Full pattern:
+[CqrsAndContentAccess.md](src/MeshWeaver.Documentation/Data/Architecture/CqrsAndContentAccess.md) → "An OPTIONAL node".
 
 **Free-floating words → vector search.** When a query contains bare text tokens (`laptop nodeType:Story`) AND PG is the backend AND an `IEmbeddingProvider` is registered, `PostgreSqlMeshQuery.QueryAsync` automatically routes through the HNSW cosine index instead of ILIKE substring scan. Structured-only queries (`nodeType:Story namespace:ACME`) stay on the regular SQL path. Full reference: [VectorSearch.md](src/MeshWeaver.Documentation/Data/Architecture/VectorSearch.md).
 
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -757,7 +757,33 @@ attempt, a stale negative produces DUPLICATE DATA, not merely duplicated work.
 **Wrong:** reading content by exact path, reading state before a write, polling for job completion,
 deciding create-or-skip for a known path.
 
-`GetMeshNodeStream(path)` + `Where(...).Take(1)` is also the right primitive for **waiting for work to finish**.
+`GetMeshNodeStream(path)` + `Where(...).Take(1)` is also the right primitive for **waiting for work to finish** — for a node that **exists**.
+
+🚨 **A node that may NOT EXIST YET is the one case where neither half is enough on its own, and each
+half alone is a defect that has shipped.** Do not pick a horn:
+
+- **A point read of an ABSENT node is forbidden by the framework**, not merely slow: the owner
+  answers an authoritative routing NotFound which **terminates the stream with an error** (it cannot
+  wait for the node to appear), and that NotFound opens `MeshNodeStreamCache`'s **storm-breaker**
+  window on the path — and the breaker **fast-fails WRITES too**, so the read suppresses the write it
+  is waiting for. The breaker says so itself: *"A point node-access to a node that does not exist is
+  a defect — read optional nodes via GetQuery (empty-on-absent), not GetMeshNodeStream(exactPath)."*
+- **Reading that node's CONTENT out of a query is forbidden above** — unbounded lag.
+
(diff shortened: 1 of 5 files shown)


rbuergi:
### Second commit: naming the counter-example (`a66dbbebb`)

Caught a defect **in this PR's own doc change** before it shipped, which is why CI restarted once.

The new section says a query cannot answer for a known path's `Content` — and the production token chip, `src/MeshWeaver.Blazor.Portal/Chat/ThreadTokenChip.razor.cs:106`, reads `content` out of the *very same* `{thread}/_Usage scope:children` query. As originally written, the doc reads as a verdict against it.

**The chip is correct, and "fixing" it would be a real defect.** It sums a SET — every `_Usage/{model}` child — to paint a total, where a briefly-stale number is cosmetic and decides nothing. Converting it to N point reads would mean N per-node hub activations on every chat render and, on a thread with no rounds yet where the set is legitimately **empty**, every one of those is an **absent-node** point read — the storm breaker firing on the render path. The threads that look most innocent would be hit hardest.

So the rule is restated as what it actually is — not about node counts, but about whether a stale answer **decides** anything:

| The value… | Read it from |
|---|---|
| is displayed, and a stale one merely looks briefly wrong | the query, `content` and all |
| **gates** something — a wait passes, a branch is taken, a write happens | the OWNER's `GetMeshNodeStream` |

`ThreadTokenUsageTest` was in the second row using the first row's primitive. The chip is in the first row using it correctly. **Same query, opposite verdicts** — and both documents now say so, with the chip named by file and line, because a rule carrying only positive examples gets over-applied, which is exactly the failure this commit prevents.

One restarted CI run is cheaper than shipping a doc that can talk someone into a render-path storm; a documentation defect outlives the test that prompted it.

`DocumentationLinkIntegrityTest` 42/42, project rebuilt first and the DLL's mtime checked against the edited `.md` (`Data/**` is `<EmbeddedResource>`, so a stale DLL would have tested the old file — that check is made, not assumed).

PR #2248, opened 2026-08-25
fix: registrations that only fail at runtime — doc images, EA consent, embeddings

### LINKED ISSUE #2218: EaConsentController fails to activate — IEaGraphAuth not registered in DI container
## What is failing

Every request that routes to `Memex.Portal.Shared.Authentication.EaConsentController` returns HTTP 500. MVC's controller activator cannot construct the controller because one of its constructor dependencies, `MeshWeaver.Mesh.IEaGraphAuth`, has no registration in the application's service collection.

## Probable cause

**High confidence — this is a DI wiring bug, not a runtime failure.** The exception message is unambiguous:

> `Unable to resolve service for type 'MeshWeaver.Mesh.IEaGraphAuth' while attempting to activate 'Memex.Portal.Shared.Authentication.EaConsentController'.`

`ActivatorUtilities.ThrowHelperUnableToResolveService` is thrown at controller construction time, before any action code runs, which means the container has never had a binding for `IEaGraphAuth` — the service was either renamed, moved to a different assembly, or the composition-root `AddXxx()` call that would have registered it is missing from the portal's `Program.cs` / startup extension. The controller itself was added (routing resolves it) but its dependency was not.

## Impact

Low volume so far — **1 occurrence, 1 pod (`memex-portal-deployment-6b8cfb8bfc-bnlzs`), single timestamp** in the `memex` namespace. However, every hit on this endpoint will 500 deterministically until the registration is added, so the EA consent flow is functionally broken for any user who reaches it. Expect the count to climb the moment a real user is directed through onboarding.

## Where to look

- **Controller:** `Memex.Portal.Shared.Authentication.EaConsentController` — inspect its constructor to see what other services it takes, so the fix registers the full set consistently.
- **Composition root:** the portal's startup / `Program.cs` (or a `AddPortalAuthentication()` / `AddEaAuth()` extension). Confirm whether an `services.AddScoped<IEaGraphAuth, …>()` (or the framework-side `AddMeshAuth()`-style call) is missing or was removed.
- **Interface origin:** `MeshWeaver.Mesh.IEaGraphAuth` — the interface is declared in the `MeshWeaver.Mesh` framework namespace but its concrete registration is the portal's responsibility. If the framework provides an extension method to register it, that call is what's missing; if not, the portal needs its own binding.
- **Middleware in the stack:** `Memex.Portal.Shared.Authentication.OnboardingMiddleware.InvokeAsync` sits above the failing MVC pipeline — worth checking whether it recently started routing to `EaConsentController` (would explain why this only just started firing).

## Routing note

The log category is the generic ASP.NET `Microsoft.AspNetCore.Diagnostics.ExceptionHandlerMiddleware`, but every application frame in the stack (`OnboardingMiddleware`, `EaConsentController`) is in `Memex.Portal.Shared.Authentication`. Leaving `repository` unset — if the default route lands this on the framework repo, please re-route to the Memex Portal repository, that's where the fix belongs.

---

### Evidence

| | |
|---|---|
| Fingerprint | `c85b85780c42acac` |
| Category | `Microsoft.AspNetCore.Diagnostics.ExceptionHandlerMiddleware` |
| Severity | Error |
| Top frame | `lambda_method2565(Closure, IServiceProvider, Object[])` |
| Namespace | `memex` |
| Pods | `memex-portal-deployment-6b8cfb8bfc-bnlzs` |
| Occurrences | 1 |
| First seen | 2026-08-25 07:58:10Z |
| Last seen | 2026-08-25 07:58:10Z |

```
2026-08-25 07:58:10Z memex-portal-deployment-6b8cfb8bfc-bnlzs fail: Microsoft.AspNetCore.Diagnostics.ExceptionHandlerMiddleware[1]
      An unhandled exception has occurred while executing the request.
      System.InvalidOperationException: Unable to resolve service for type 'MeshWeaver.Mesh.IEaGraphAuth' while attempting to activate 'Memex.Portal.Shared.Authentication.EaConsentController'.
         at Microsoft.Extensions.DependencyInjection.ActivatorUtilities.ThrowHelperUnableToResolveService(Type type, Type requiredBy)
         at lambda_method2565(Closure, IServiceProvider, Object[])
         at Microsoft.AspNetCore.Mvc.Controllers.ControllerFactoryProvider.<>c__DisplayClass6_0.<CreateControllerFactory>g__CreateController|0(ControllerContext controllerContext)
         at Microsoft.AspNetCore.Mvc.Infrastructure.ControllerActionInvoker.Next(State& next, Scope& scope, Object& state, Boolean& isCompleted)
         at Microsoft.AspNetCore.Mvc.Infrastructure.ControllerActionInvoker.InvokeInnerFilterAsync()
      --- End of stack trace from previous location ---
         at Microsoft.AspNetCore.Mvc.Infrastructure.ResourceInvoker.<InvokeNextResourceFilter>g__Awaited|25_0(ResourceInvoker invoker, Task lastTask, State next, Scope scope, Object state, Boolean isCompleted)
         at Microsoft.AspNetCore.Mvc.Infrastructure.ResourceInvoker.Rethrow(ResourceExecutedContextSealed context)
         at Microsoft.AspNetCore.Mvc.Infrastructure.ResourceInvoker.Next(State& next, Scope& scope, Object& state, Boolean& isCompleted)
         at Microsoft.AspNetCore.Mvc.Infrastructure.ResourceInvoker.InvokeFilterPipelineAsync()
      --- End of stack trace from previous location ---
         at Microsoft.AspNetCore.Mvc.Infrastructure.ResourceInvoker.<InvokeAsync>g__Logged|17_1(ResourceInvoker invoker)
         at Microsoft.AspNetCore.Mvc.Infrastructure.ResourceInvoker.<InvokeAsync>g__Logged|17_1(ResourceInvoker invoker)
         at Memex.Portal.Shared.Authentication.OnboardingMiddleware.InvokeAsync(HttpContext context) in /home/runner/work/MeshWe…[truncated]
```

</details>

<sub>Opened automatically from `Admin/_LogIncident/c85b85780c42acac`. Recurrences are folded into this issue rather than opening new ones.</sub>

rbuergi:
Four production issues that share one shape: **the code compiles, every reference resolves, and the
thing that is missing is a *registration*.** None is visible to `dotnet build`, and each surfaces far
from where it was caused.

Closes #2218, #2122, #2123 (#2123 is #2122's other face — same fingerprint `d94a160fc73a60d5`, same
two frames), and #1642.

---

## #2122 / #2123 — doc-tree images: the wire read truncated a perfectly good config

`EmbeddedResourceStreamProviderFactory.Create` threw `ArgumentException: AssemblyName required for
EmbeddedResource` for every image on a doc page, on both memex-cloud pods, over three days.

**The registration was never at fault.** `AddEmbeddedResourceContentCollection` always writes
`Settings["AssemblyName"]` and `Settings["ResourcePrefix"]`. What lost them is the **wire read**: the
content route asks the owning node hub for its collection configs with a `GetDataRequest`, the
response arrives as a `JsonElement`, and `ContentFileResolver.ReadCollectionConfigs` rebuilt each
config **by hand from five named properties** — `Settings` not among them. So the config that reached
the factory was a truncated copy of a correct registration, and a guard three frames and one assembly
away complained about a field nobody had omitted.

- The projection now carries **every** property. `Settings`, `DisplayName`, `Order`, `IsEditable` and
  `ExposeInChildren` were all being dropped.
- `EveryConfigProperty_SurvivesTheWireProjection` pins each one **and the property count**, so adding
  a tenth property to `ContentCollectionConfig` fails there with a message saying what to do.
  `IsStatic` (#587) and `Address` were lost to this same hand-written projection before `Settings`
  was — each found in production, not here.
- The guard's exception now names the collection and its address, not only the field.

## #2218 — `EaConsentController` 500s: a routed controller with a feature-flagged dependency

`AddControllers().AddApplicationPart(...)` discovers controllers by **type**, with no idea what any
of them needs — so `/auth/ea/connect` and `/auth/ea/callback` are routed on every deployment.
`IEaGraphAuth` was registered only inside `if (emailOptions.Enabled)`. Gating it did not disable the
endpoint; it left it routed and unconstructible, and MVC's activator threw before any action code ran.

`Email:Enabled` was the wrong condition anyway: it gates the **system mailbox's app-permission
sender**, while the EA's *delegated* flow needs `Authentication:Microsoft` credentials —
`EaGraphAuth.IsConfigured` already reports that honestly, and the controller already answers **400
"not configured"** on it. The registration is now unconditional (the descriptor is inert; a typed
`HttpClient` registration does no work until something resolves it), so the endpoint returns a
truthful 400 rather than a 500.

Two tests over the portal's **real** composition — which nothing did before (`EaConnectRouteTest`
asks the routing table and deliberately never instantiates a controller; the other two use a
hand-written fake):

- the specific regression, and
- `NoControllerDependency_DependsOnADeploymentFlag`, stated as a **difference** between two
  compositions. A "must all be registered" sweep cannot work here: the controllers also take
  mesh-contributed services (`IMeshService`, `IMessageHub`, `AccessService`) that live in the mesh's
  provider — the one `MeshHostApplicationBuilder` makes the ASP.NET root provider — and are
  legitimately absent from this collection. Comparing two compositions cancels those out exactly and
  leaves the property that actually broke.

## #1642 — `IEmbeddingProvider`: **optional**, and now said out loud

**Verdict: optional.** A deployment with no `Embedding:Endpoint` is supported — search falls back to
an ILIKE substring scan and content indexing resolves inert stand-ins. The crash the issue reported
(`GetRequiredService<IEmbeddingProvider>` inside the vector store's factory) was already root-caused
on main by gating the *consumer*, pinned by `ContentIndexingOffToolTest`. Making it required would be
wrong.

What was still wrong is that the decision was taken **in silence**: `TryAddEmbeddingProvider`
returned a bool one caller read, and nothing said which way it went — so "semantic search is off
here" and "semantic search is broken here" produced identical evidence, which is why #1642 had to be
triaged from a stack trace. The decision is now an `EmbeddingCapability` in DI plus exactly one
`Information` line at startup, **in both directions** (a line that appeared only when the capability
was on would be no better than the silence), naming the config key that would turn it on:

```
Semantic (vector) search DISABLED — no Embedding:Endpoint is configured. Free-text queries fall
back to an ILIKE substring scan and content indexing stays inert; this is a supported
configuration, not a fault.
```

🚨 It deliberately does **not** default to `NullEmbeddingProvider`. The *presence* of the
registration is the capability signal every consumer reads — both storage backends use
`GetService<IEmbeddingProvider>()`, and the content-indexing module's resolve-time `enabledWhen` gate
asks the same question. A Null default answers *yes* on a host that has none, and would light the
indexing pipeline up against an embedder that can never embed — #1642 from the other direction.

`EmbeddingCapability.IsEnabled` is taken from whether a provider was *actually* created, never
re-derived from the options: a second copy of `CreateEmbeddingProvider`'s branch logic could drift
and then advertise a capability the host does not have.

### …and a second, independent defect found while doing it

`PostgreSqlExtensions.cs:161` and `:206` and `SnowflakeExtensions.cs:138` each called
**`services.BuildServiceProvider()` mid-registration** to read the optional `IEmbeddingProvider`.
That builds a *second, throwaway* container: anything registered after that line is invisible to it,
and any singleton it does resolve is a **duplicate** of the one every other consumer gets. So whether
those adapters had embeddings came down to whether the host called `AddEmbeddings` before or after —
and both failure modes are invisible, presenting as the same silent degrade to lexical search. Fixed
by building the adapters from the real container through a new
`AddPersistence(Func<IServiceProvider, IStorageAdapter>)` overload (the decorator chain resolves the
last `IStorageAdapter` registration whether it is an instance or a factory).

Those three lanes currently have **no callers** (the portals use
`AddPartitionedPostgreSqlPersistence`), so the defect was latent — but it would have quietly defeated
the fix above the moment one was used. No `services.BuildServiceProvider()` remains on any
registration path in `src/` or `memex/`.

---

## Verification

**The content tests are red before the fix, by measurement.** Reverse-applied `ContentFileResolver.cs`
alone, rebuilt, re-ran: **5 failed / 0 passed**, each on `Settings` being null —

```
Expected value not to be <null> because Settings carries AssemblyName + ResourcePrefix for an
EmbeddedResource collection — dropping it on the wire is issues #2122/#2123.
```

— then restored the fix: **5 passed**.

| Test | Pins |
|---|---|
| `DocAssetWireProjectionTest` (5) | a real shipped SVG served end-to-end after the config crosses the wire, for three doc nodes at different depths, incl. `image/svg+xml` and `<svg` in the bytes; every config property survives the projection; `Doc/AI`'s collection opens |
| `EaConsentControllerActivationTest` (3) | the seam is registered with mail off; no controller dependency varies with a flag; the sweep is non-vacuous |
| `EmbeddingCapabilityTests` (9) | off / on / keyless-cloud decisions, the reported state always matches what was actually registered, and the startup line is emitted exactly once, in both directions |

`-c Release -warnaserror`, `0 Error(s) / 0 Warning(s)` on each of `MeshWeaver.ContentCollections`,
`MeshWeaver.Hosting`, `MeshWeaver.Hosting.Embeddings`, `MeshWeaver.Hosting.PostgreSql`,
`MeshWeaver.Hosting.Snowflake`, `Memex.Portal.Shared`, and all three touched test projects.

`DocumentationLinkIntegrityTest` re-run green after the `VectorSearch.md` edit, with
`MeshWeaver.Documentation.dll` rebuilt first and its mtime checked against the file's — it ships
`Data/**` as `EmbeddedResource`, so `--no-build` would have tested the old copy.

---

## 🚨 Not fixed here, and deliberately so

One of the two paths #2122 names, `Doc/AI/images/agenticai.svg`, has a **second, independent
defect** this PR does not touch: the asset ships at `Content/images/agenticai.svg` (the partition
root) while six `Doc/AI/*` pages declare `Thumbnail: "images/agenticai.svg"`, which resolves against
their own folder node. It is not confined to that file — `images/DataMesh.svg` and
`images/notifications.svg` are declared as thumbnails and ship nowhere either. That is a doc-content
question, not a registration one, and it wants its own change.

After this PR that path answers a plain "file not found" instead of a 500 — the honest answer for an
asset that genuinely is not there. `DocAiCollection_OpensWithACompleteConfig` pins exactly that much
and says so in its doc comment.

### DIFF
--- a/memex/Memex.Portal.Shared/MemexConfiguration.cs
+++ b/memex/Memex.Portal.Shared/MemexConfiguration.cs
@@ -147,18 +147,34 @@ public static class MemexConfiguration
             .GetSection(EmailOptions.SectionName)
             .Get<EmailOptions>() ?? new EmailOptions();
         services.AddSingleton(emailOptions);
-        if (emailOptions.Enabled)
-        {
-            // Executive Assistant: per-user JUST-IN-TIME delegated Graph access (the user consents to the
-            // EA touching THEIR OWN mailbox only when they first use the tool — no standing app
-            // permission). EaGraphAuth drives the consent/token flow; it is raw OAuth over HTTP, so it
-            // stays HERE with its consent controller — the EA's mailbox TOOLS (which do use the Graph
-            // SDK) ride the MeshWeaver.Mail.MicrosoftGraph module and depend only on this seam.
-            services.AddHttpClient<IEaGraphAuth, Authentication.EaGraphAuth>();
-            // The notification triage runner (escalates in-app notifications to email/Teams per each
-            // recipient's NotificationRules) rides the MeshWeaver.Notifications.Channels module
-            // (Modules:Assemblies); its hosted service self-skips unless Email:Enabled.
-        }
+        // The notification triage runner (escalates in-app notifications to email/Teams per each
+        // recipient's NotificationRules) rides the MeshWeaver.Notifications.Channels module
+        // (Modules:Assemblies); its hosted service self-skips unless Email:Enabled.
+
+        // Executive Assistant: per-user JUST-IN-TIME delegated Graph access (the user consents to the
+        // EA touching THEIR OWN mailbox only when they first use the tool — no standing app
+        // permission). EaGraphAuth drives the consent/token flow; it is raw OAuth over HTTP, so it
+        // stays HERE with its consent controller — the EA's mailbox TOOLS (which do use the Graph
+        // SDK) ride the MeshWeaver.Mail.MicrosoftGraph module and depend only on this seam.
+        //
+        // 🚨 UNCONDITIONAL, and it must stay that way. EaConsentController is registered by
+        // AddControllers().AddApplicationPart(...) below — which discovers controllers by TYPE, with
+        // no idea what any of them needs — so /auth/ea/connect and /auth/ea/callback are routed on
+        // EVERY deployment. Gating this one registration on Email:Enabled therefore did not disable
+        // the endpoint; it left it routed with an unresolvable dependency, and MVC's activator threw
+        // "Unable to resolve service for type 'MeshWeaver.Mesh.IEaGraphAuth' while attempting to
+        // activate 'EaConsentController'" — a deterministic 500 on the consent flow, in memex prod
+        // (issue #2218). A controller and its dependencies are one unit: whatever routes the one
+        // must register the other.
+        //
+        // Nothing is enabled by registering it. The descriptor is inert (a typed HttpClient factory
+        // registration does no work until something resolves it), and IEaGraphAuth carries its own
+        // honest feature probe: EaGraphAuth.IsConfigured reads Authentication:Microsoft
+        // ClientId/ClientSecret — the credentials the DELEGATED flow actually needs, which are not
+        // what Email:Enabled describes (that gates the SYSTEM mailbox's app-permission sender). The
+        // controller checks IsConfigured first and answers 400 "not configured", which is the honest
+        // answer for a deployment that has not set up the EA — instead of a 500 that reads as a bug.
+        services.AddHttpClient<IEaGraphAuth, Authentication.EaGraphAuth>();
 
         // 🚨 TryAdd, deliberately. The Graph sender lives in the MeshWeaver.Mail.MicrosoftGraph
         // module, which registers it with a plain AddSingleton. The pairing is ORDER-INDEPENDENT:
--- a/src/MeshWeaver.ContentCollections/ContentFileResolver.cs
+++ b/src/MeshWeaver.ContentCollections/ContentFileResolver.cs
@@ -1,3 +1,4 @@
+using System.Collections.Immutable;
 using System.Reactive.Linq;
 using System.Text.Json;
 using MeshWeaver.Data;
@@ -251,12 +252,27 @@ public static class ContentFileResolver
 
     /// <summary>
     /// Projects the collection configs out of the owning hub's <c>GetDataResponse</c>. The remote
-    /// form arrives as a <see cref="JsonElement"/>; <see cref="ContentCollectionConfig.IsStatic"/>
-    /// and <see cref="ContentCollectionConfig.Address"/> MUST survive that projection — without the
-    /// first, every legitimately-published remote collection fails a caller's mount check; without
-    /// the second, an inherited collection loses its owner and a child node's file resolves against
-    /// the ancestor's folder. <c>IsStatic</c> is suppressed when false, so an absent property means
-    /// false — the safe value either way.
+    /// form arrives as a <see cref="JsonElement"/>.
+    ///
+    /// <para>🚨 EVERY property of <see cref="ContentCollectionConfig"/> must survive this
+    /// projection, and <c>DocAssetWireProjectionTest.EveryConfigProperty_SurvivesTheWireProjection</c> keeps it true:
+    /// a hand-written projection that reads only the fields its author happened to need is a
+    /// SILENT truncation of the config, and the missing field surfaces far away as a failure that
+    /// names neither this method nor the property it dropped.</para>
+    ///
+    /// <para>Three of them have already cost outages. <see cref="ContentCollectionConfig.IsStatic"/>
+    /// — without it every legitimately-published remote collection fails a caller's mount check.
+    /// <see cref="ContentCollectionConfig.Address"/> — without it an inherited collection loses its
+    /// owner and a child node's file resolves against the ancestor's folder.
+    /// <see cref="ContentCollectionConfig.Settings"/> — the provider-specific payload, and for an
+    /// <c>EmbeddedResource</c> collection the ONLY thing that names the assembly and resource
+    /// prefix its files live in. Dropping it turned every doc-page image into
+    /// <c>ArgumentException: AssemblyName required for EmbeddedResource</c> raised at the far end
+    /// of the pipeline, in <see cref="EmbeddedResourceStreamProviderFactory"/> (issues #2122/#2123):
+    /// the registration DID supply the assembly, the wire read threw it away.</para>
+    ///
+    /// <para><c>IsStatic</c>/<c>IsEditable</c>/<c>ExposeInChildren</c> are suppressed when false, so
+    /// an absent property means false — the safe value for all three.</para>
     /// </summary>
     /// <param name="delivery">The response delivery, or <c>null</c>.</param>
     /// <returns>The configs, or <c>null</c> when the response carried none.</returns>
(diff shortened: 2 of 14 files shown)


rbuergi:
### CI note: shard 2 red on a pre-existing, load-dependent flake — re-ran that shard only

The first full run went **5/6 shards green**; shard 2 failed on exactly one test:
`MeshWeaver.Hosting.Monolith.Test.PandasExplorerLayoutAreaTest.Explorer_NoPythonNode_RendersToolbarAndDegradesGridWithoutHanging`.

```
Expected the observable to emit a value matching the predicate within 50s, but it did not.
Last of 4 emission(s) was: RedirectControl { … Href = /Doc/DataMesh/PythonPandasNode/PandasExplorer/LiveFrame/Explorer }
```

**Attributed by mechanism, not by "it's probably flaky".** `AreaFrameClassifier.IsTransientFrame`
defines a `RedirectControl` as a **transient** frame — *"a frame that is not the area's content and
will be REPLACED without anyone acting: the compile-progress page, and the RedirectControl it emits
once the build settles."* So the last thing that area produced was the placeholder shown **while a
dynamic NodeType build is still settling**; the 50 s budget expired before the real composition
arrived. That is in-mesh NodeType compile latency under runner load.

Corroborating evidence:

- **The runners were badly contended on this run** — the shards sat `queued` for ~20 minutes
  (`Run tests (shard 0/4/5)` queued at 18:57, shard 2 not starting until 19:05).
- **This exact test was already patched once for this exact shape**: f7b6a0532, *"Pandas Explorer
  wait matches the COMPLETE composition, not the first Stack frame"*, whose comment records a loaded
  CI runner sampling an incomplete frame on PR #1970's run. This time the sampled frame was even
  earlier in the sequence.
- **Locally it passes in 2 s** against that 50 s budget.
- **This diff cannot reach it.** The changes are: a wire projection used only by
  `ContentFileResolver.Resolve` (the `/api/content` route and the deck export — not layout-area
  rendering), an EA DI registration in `MemexConfiguration` (this test never runs
  `ConfigureMemexServices`), the embeddings capability record, and two PostgreSQL/Snowflake
  persistence lanes that have **no callers** and are not on the monolith test path.

So only shard 2 was re-run; nothing was changed to make it pass, and no bound was widened.

🚨 **Not fixed here, and worth its own issue:** this test's wait is bounded by a wall-clock budget
that a NodeType compile can legitimately exceed on a loaded runner. The durable fix is to wait on the
*settled* frame — `AreaFrameClassifier.IsTransientFrame` already names precisely the frames that mean
"keep waiting" — rather than to raise the number.

Issue #2210, opened 2026-08-25
Incident 2026-08-25: helm apply with public values half wiped the GitHub App identity from both prod ConfigMaps — GitSync + registry pulls down ~15 min


rbuergi:
## Impact

2026-08-25 ~11:13–11:30 UTC, both prod portals (memex.systemorph.com and memex.meshweaver.cloud): every GitHub sync failed with "Connect your GitHub account first … or configure the GitHub App identity". Because memex-cloud is the plugin registry and its GitHub App credential is what pulls `Systemorph/MeshWeaver.Plugins@main`, this also stalled catalog auto-updates and the in-flight `_App` migration. Restored at ~11:24Z; both portals verified (`git_hub_sync op=check space=Store` → "Branch 'main' is at b8f3523a", Succeeded).

## Root cause

**The GitHub App identity lived only as live drift on the ConfigMap, and a helm apply rendered from the public values half clobbered it.**

- `GitHub__App__{ClientId,InstallationId,InstallationOwner}` were never in any values source: I decoded the last-good helm release secrets (`sh.helm.release.v1.memex.v17`, `sh.helm.release.v1.memexcloud.v3`) and their rendered `memex-portal-config` has all three **empty** — the chart templates them as `default ""`. The real values had been `kubectl patch`ed onto the live ConfigMap (2026-07-09, the GitSync-as-App rollout) and survived every apply since only because those applies didn't touch the keys.
- At **10:54:08Z** (ns `memex`, helm rev v18) and **11:06:06Z** (ns `memex-cloud`, rev v4) — four minutes after #2203 ("render OpenRouter__Models") merged — helm upgrades were applied with the public halves only. The CM re-render emptied the three keys; the failure began when the new ReplicaSets came Ready (~11:13Z).
- The applier is unidentified (no session claimed it at the time of writing); the mechanism does not depend on who — **any** apply without the identity in a values source reproduces it.
- Not related to #2197's enablement earlier this morning: that was a single-key `kubectl patch --type merge` (cannot remove keys), and sync still worked at 11:04Z, after its rollout.

## Remediation (done)

[...]

Issue #1971, opened 2026-08-21
A roll that outlasts its grace SIGKILLs a live silo — is the departure clean, and nothing pins the drain


rbuergi:
Split out of #1340, which is otherwise superseded (see its closing comment). This is the one part of
it that is still open, and it has a sharper edge now than when it was written.

## The mechanism

`preStop` runs **before** SIGTERM and blocks until `/drain` reports the last circuit closed, with
`terminationGracePeriodSeconds: 1800` as the backstop. That is deliberate and it is what makes a roll
non-disruptive (#1342).

The consequence at the ceiling is the part nobody has looked at: if the sessions outlast the grace,
`preStop` never returns, so the process is **SIGKILLed with a live Orleans silo**. The host's 90 s
`ShutdownTimeout` (`Memex.Portal.Distributed/Program.cs`) never gets to run, `ApplicationStopping`
never fires, and the silo never departs membership gracefully.

The deployment's own annotation says what that costs:

> `cluster-autoscaler.kubernetes.io/safe-to-evict: "false"` — *"each abrupt departure left a ZOMBIE
> entry in the Orleans membership table: the cluster kept placing new grain activations on the dead
> silo, so writes timed out ('no initial state arrived within 30s') mesh-wide."*

So the roll design protects users from the *fast* teardown by guaranteeing the *abrupt* one whenever
a session outlives the grace.

## What is measured, and what is not

**Measured** (`memex`, 2026-08-21):

```
memex-portal-deployment-649ccff9f-945gf   3/3  Terminating
  deletionTimestamp = 2026-08-21T09:10:52Z    (SIGTERM scheduled 08:40:52Z, grace 1800)
  09:09:44Z  still 3/3 Running, phase=Running
  09:09:33Z  still executing application code
  09:10:52Z  gone — at the ceiling
```

and, from inside a **live** portal pod at the same time:

```
$ curl -s -w "http=%{http_code}" http://127.0.0.1:8080/drain
http=503
5 circuit(s) still open
```

So this is not an exotic path: five circuits were open on a serving pod at an arbitrary moment, and
the pod that rolled ahead of it rode the full grace. Riding to the ceiling looks like the *normal*
outcome of a roll, not the exception.

**Not measured:** whether that SIGKILL actually left a zombie membership entry. I grepped the
surviving replica's log for membership/silo-departure lines in the window after 09:10:52Z and found
nothing, which is inconclusive — Orleans membership may simply not log at the configured level. I am
[...]

Issue #1340, opened 2026-08-12
A roll kills every grain and circuit — keep the old silo, move activations only on explicit recycle


rbuergi:
## What happens today

`deploy/helm/templates/memex-portal/deployment.yaml` is `replicas: 1`, `maxSurge: 1`, `maxUnavailable: 0` — surge-first. HTTP has no outage: the old pod serves until the new one passes its probes. Then k8s **deletes the old pod**, and with it:

- every grain activation on that silo deactivates,
- every Blazor circuit bound to that pod dies (the affinity cookie pins a circuit to its pod),
- every dynamically compiled NodeType assembly is rebuilt on the new silo.

That is the "the demo crashed" phenomenon: a self-update roll lands mid-session and users see a dead circuit. Observed again on 2026-08-12 while rolling `memex-portal-ai` by hand.

`SelfUpdateHostedService` polls ACR every 6 h and PATCHes its own Deployment, so this happens unattended. `UpdatePolicyKind.None` switches auto-update off entirely, but that is all-or-nothing: it trades live sessions for never getting fixes.

## Proposed model

**Do not kill the old silo. Let its grains keep running against the code they activated with, and place only NEW activations on the new silo — an explicit recycle is what moves a grain across.**

1. **Version-aware placement.** Each silo advertises its build (image tag / informational version) as silo metadata; a placement director prefers the newest advertised build among compatible silos. Orleans' own primitives are `[Version(n)]` on grain interfaces with `VersionSelectorStrategy.LatestVersion` + `CompatibilityStrategy.BackwardCompatible` — note that **none of these appear anywhere in `src` today**, so this is greenfield, and interface-version bumps are impractical per build, which is why silo metadata + a placement director is the more workable form here.
2. **Two live deployments, not one rolling.** A k8s Deployment always reaps the old ReplicaSet's pods after a successful rollout — you cannot ask it to keep one. Keeping the old silo alive needs a version-labelled blue/green pair.
3. **A retirement policy.** Drain when the old silo's activation count reaches zero, or after a TTL. Without this, old silos accumulate forever.
4. **Explicit recycle is the migration trigger.** The existing recycle action deactivates a grain; with (1) its next activation lands on the new silo. That is the whole user-visible contract: *nothing moves until you move it.*

## What has to be settled first — these bite harder than the rollout

[...]

PR #2253, opened 2026-08-25
Deployment correctness: refuse undeliverable mail, model the migration run-once, pin the drain

### LINKED ISSUE #1788: memex-migration CrashLoopBackOff in all three prod namespaces (50/53/38 restarts) — run-once workload modelled as a Deployment
Found while stabilising prod on 2026-08-17. Present in **all three** namespaces simultaneously:

| namespace | pod | restarts |
|---|---|---|
| `memex-cloud` | `memex-migration-deployment-b6897bff4-tqrz2` | **50** |
| `memex` | `memex-migration-deployment-59486ff469-wb6z4` | **53** |
| `atioz` | `memex-migration-deployment-658896cbfb-bs8mp` | **38** |

All three sit at `0/1 CrashLoopBackOff`, restarting roughly every 4 minutes, indefinitely.

## This is "benign" only in the sense that it does not break serving

The migration container runs the migration, exits 0, and the **Deployment** restarts it — because a run-once workload is modelled as a Deployment rather than a Job. AGENTS.md already documents the symptom as expected. The costs are real anyway, and they compound:

1. **It permanently occupies a `CrashLoopBackOff` slot in every namespace**, so "is anything crash-looping in prod?" — the first question anyone asks during an incident — can never be answered `no`. A genuine migration failure looks exactly like the standing noise.
2. **It restarts ~360 times/day across three namespaces**, each restart pulling and starting a container and writing a fresh crash record.
3. It contradicts the deployment's own success criterion (AGENTS.md: *confirm its log shows `Database migration completed. Version: N`*) — you cannot distinguish "completed and exited" from "failed and will retry" by pod status.

## The fix already exists and was never applied here

The run-once Job fix is `dd49cd1bf`. The `memex` namespace has been known to be missing it; this issue records that **all three** are, so applying it to one is not enough.

Whoever takes it: verify per environment that the Job form still satisfies the portal's ordering requirement — `KubernetesDeploymentUpdater.PatchToVersionAsync` patches portal **then** migration, and the AdoNet grain-storage work makes the portal's start depend on migration-created tables, so the two must not be able to invert.

### LINKED ISSUE #2023: Email:Enabled=true with the NoOp sender stamps queued mail Sent — silent non-delivery, currently likely live on memex.meshweaver.cloud
## Summary

Since the mail stack moved into the `MeshWeaver.Mail.MicrosoftGraph` module (#1737, 2026-08-19), a deployment with `Email:Enabled=true` but **without the module landed** runs `NoOpEmailSender` — whose contract is to *report success without sending*. `OutboundEmailSender` therefore flips every queued mail `New → Sending → Sent` while nothing leaves the process. The failure is invisible by construction: the node data says delivered, no log reaches stdout (the send lines are Information under a Warning default), and the sender's own idle message fires only when `Enabled` is false.

## How this bit in practice (2026-08-21, local memex portal)

1. `Email__*` fully configured, `Enabled=true`; contact-form submissions produced enquiry records and `Email` nodes that all read `Sent`.
2. Nothing arrived; inbox/junk/quarantine hunting and a message-trace plan wasted hours because `Sent` was trusted.
3. A DI probe settled it: `IEmailSender = Memex.Portal.Shared.Email.NoOpEmailSender`. The module had no way onto the install (not in the image's `modules/` layout, no registry configured, registry bundle index needs an instance key — related: #1782, #1751, #1979).

## How to reproduce

**Fastest (unit level, no portal):**
1. Build a service collection the way `MemexConfiguration` does with `Email:Enabled=true` but **without** the `MeshWeaver.Mail.MicrosoftGraph` module in `Modules:Assemblies` (any default local checkout does this — the module ships only via the plugin bundle lane).
2. Resolve `IEmailSender` → it is `NoOpEmailSender` (the `TryAddSingleton` fallback; nothing else registered a sender).
3. Call `SendEmail("anyone@example.test", "s", "b")` → emits `true`. Any caller that records an outcome now records success for a send that never happened.

**Full portal (the live reproduction, ~10 min on a memex-local Colima install):**
1. Take a stock `memex-local` portal — `Modules:Assemblies` carries only the image baseline; `/data/modules/activation.json` has no `MeshWeaver.Mail.MicrosoftGraph` entry. (This is every local install's default state.)
2. Enable e-mail with real Graph credentials:
   `kubectl -n memex set env deployment/memex-portal-deployment Email__Enabled=true Email__MailboxAddress=<tenant mailbox> Email__TenantId=… Email__ClientId=… Email__UseManagedIdentity=false Email__ClientSecret=…`
   and wait for the rollout. (The credentials can even be **garbage** — the defect does not depend on them, which is itself the point: nothing ever tries to use them.)
3. Queue one outbound mail: create an `Email` node anywhere (e.g. `<space>/_Email/mail-1`) with `direction: Outbound`, `status: New`, a valid `to:`.
4. **Expected**: the mail is delivered, or its status becomes `Failed` with a diagnosable cause.
   **Actual**: within seconds the node reads `status: Sent` (writer `system-security`, the `New → Sending → Sent` sequence) — and no mail was sent, no error is logged, and stdout carries no send line at all (the senders' Information lines sit under the Warning default).
5. Confirm the cause with a DI probe from a kernel script (`Mesh` = the portal hub):
   ```csharp
   Console.WriteLine(Mesh.ServiceProvider.GetService<IEmailSender>()?.GetType().FullName);
   // → Memex.Portal.Shared.Email.NoOpEmailSender   (with Email:Enabled=true)
   ```
6. Cross-check that the pipeline is otherwise healthy: land the module (or run on a deployment that has it) and repeat step 3 — the same node now produces a real Graph send. Only the module's presence separates "delivered" from "silently lied about".

## Why production is probably affected right now

`Plugins/Mail` on memex.meshweaver.cloud shows a content-only install (`installedNodeCount: 1`, `installedFiles: Mail/index.json`) — byte-identical in shape to the broken local record. If the module is not in that pod's activation sidecar, **every contact-sales enquiry notification and every invitation since 2026-08-19 was stamped `Sent` and never sent.** Verification:

```
az aks command invoke -g <rg> -n <cluster> --command "kubectl -n memex exec deploy/memex-portal-deployment -- cat /data/modules/activation.json"
```

plus the human check: has any "New enquiry" mail arrived at the sales recipients since 2026-08-19?

## Root cause

`NoOpEmailSender.SendEmail` returns `Observable.Return(true)`. Its design assumption — "registered when `Email:Enabled` is false, so callers' chains complete normally" — was valid while the Graph sender was compiled in. The module extraction added a second path to NoOp (`TryAddSingleton` fallback when the module is absent) in which `Enabled` **is** true and watchers **do** run; success-faking then converts a deployment gap into silent data-plane lies. Its log line ("Email disabled (Email:Enabled=false)…") also states a falsehood on this path.

## Proposed fix

1. **Fail loudly at startup**: when `Email:Enabled=true` and the resolved `IEmailSender` is the NoOp, log an **error** naming the missing module and do **not** start the outbound/invitation watchers — queued mail then visibly stays `New` instead of lying `Sent`, and delivery resumes without data repair once the module lands.
2. **Never fake success on this path**: NoOp keeps returning `true` only for the genuine `Enabled=false` case (where no watcher calls it anyway); any invocation while `Enabled=true` should return `false` (→ `Failed`, visible and re-queueable) — or throw.
3. Correct the NoOp log text to state the actual cause (disabled vs. module absent).
4. Ops follow-up, separate from the code fix: land/adopt the module on memex.meshweaver.cloud and re-queue anything stamped `Sent` after 2026-08-19 whose delivery the recipients cannot confirm.

Related: #1737 (the extraction), #1782 / #1751 (adoption gaps), #1979 (PendingRestart surfaced to nobody), #1980 (no cross-repo module gate).

### LINKED ISSUE #2144: Deployment docs: 22 files, four overlapping clusters, two carrying stale commands
## Summary

The deployment documentation in this repo tells the same story several times, and the copies have
already drifted apart on facts that change what an operator types. A sweep of every deployment-related
markdown file found **22 files across this repo and the private ops repo**, in four clusters — two of
which contain measurable, current divergence.

This issue proposes a consolidation. It does not propose deleting anything before the open question
at the bottom is answered.

### 1. Cluster access — VPN vs `az aks command invoke`

`AGENTS.md` is absolute: on a private cluster, `kubectl` runs **only** via `az aks command invoke`.
Two large documents still lead with a point-to-site VPN and client-certificate generation:

| File | "P2S"/"VPN" | "command invoke" |
|---|---|---|
| `deploy/aks/README.md` | 40 | 3 |
| `src/MeshWeaver.Documentation/Data/Architecture/MemexCloudDeployment.md` | 21 | 4 |
| `src/MeshWeaver.Documentation/Data/Architecture/DeploymentAKS.md` | 0 | 4 |
| `src/MeshWeaver.Documentation/Data/Architecture/Instances.md` | 0 | 2 |

A reader following either of the first two starts by provisioning a VPN they may not need.

### 2. The new-environment runbook — three copies, two stale facts

`OnboardingNewEnvironment.md` (312 lines, 2026-08-17) and `.claude/skills/new-deployment/SKILL.md`
(150 lines, 2026-08-19) restate a runbook whose maintained copy lives in the private ops repo and was
updated 2026-08-24.

- **Env folder location.** `OnboardingNewEnvironment.md` says environments live under
  `deploy/aks/envs/<env>/`; they moved to the ops repo on 2026-08-08/09. `SKILL.md` mentions both
  paths. The old directory still exists here, so nothing errors — the guidance is simply no longer
  authoritative.
- **Database address.** `OnboardingNewEnvironment.md` still writes `Host=<PG_PRIVATE_IP>`. The
  servers moved to the FQDN on 2026-08-24, and the commit that made that durable warns a stale
  address "would revert the migration and re-break both instances".

### 3. The AKS story, told four times

`deploy/aks/README.md` (924) · `deploy/aks/DEPLOY-RUNBOOK.md` (446) ·
`Architecture/MemexCloudDeployment.md` (348) · `Architecture/DeploymentAKS.md` (325) cover
overlapping ground — architecture, images, bicep, helm, TLS, sign-in, operations. Two of them still
use the placeholder host `portal.example.com`, which no longer names anything.

### 4. Five release documents

`ReleaseGates.md` (475) · `ContinuousDeliveryContract.md` (364) · `ReleaseToProductionPipeline.md`
(280) · `ReleaseStrategy.md` (244) · `ReleaseProcess.md` (181) — 1,544 lines on one subject.
Flagging the cluster for review, not prejudging it: Strategy / Gates / Contract look separable;
Process and Pipeline look like the fold candidates.

## Proposed changes

**Keep and narrow**

- `Architecture/DeploymentAKS.md` — already narrow (code update only) and current; it is what
  `AGENTS.md` points at.
- `deploy/aks/README.md` — retitle as what it is, a public, self-contained deployment **sample** for
  third parties, and fold `DEPLOY-RUNBOOK.md` into it.

**Trim to a pointer**

- `Architecture/OnboardingNewEnvironment.md` — keep only what is genuinely public and
  mechanism-level (shared vs separate, the chart's config pass-through rule, the self-update model).
  Remove folder layouts, connection strings, and anything naming operational infrastructure; link
  out for the operator procedure.
- `.claude/skills/new-deployment/SKILL.md` — stop restating steps; point at the maintained runbook.
  The skill already knows the inventory is private and must not be duplicated here.

**Fold or delete**

- `Architecture/MemexCloudDeployment.md` — superseded by `deploy/aks/README.md` plus
  `DeploymentAKS.md`; whatever is unique in it should move into one of those.
- `deploy/aks/DEPLOY-RUNBOOK.md` — into `deploy/aks/README.md`.

**Leave alone** (genuinely distinct): `Architecture/Deployment.md` (route index),
`DeploymentContainerApps.md`, `DeploymentOptions.md`, `Instances.md`, `DeployingPluginChanges.md`,
`InstanceSync.md`, `CrossInstanceMirror.md`, `DatabaseBackups.md`, `PluginRegistry.md`,
`deploy/aks/SELF-UPDATE.md`, `deploy/aks/INSTANCE-LIFECYCLE.md`, `deploy/CHANGE-PROPAGATION.md`,
and the component READMEs under `deploy/`.

## Order of work

1. Correct the two stale facts in `OnboardingNewEnvironment.md` and `SKILL.md`, and repoint
   `Instances.md` at the maintained runbook. Small, no decision needed.
2. Resolve the open question below, then cut whichever document contradicts the answer.
3. Fold `DEPLOY-RUNBOOK.md` and `MemexCloudDeployment.md`.
4. Review the release cluster separately.

## Open question — blocks step 2

**Is the point-to-site VPN still a supported way to reach the cluster, or is `az aks command invoke`
the only one?** If it is still supported, the two documents need a short "when to use which" section
rather than deletion. If it is not, roughly 60 mentions across two files are actively misleading and
should go.

rbuergi:
Fixes a cluster of deployment-correctness issues. Three of the four are the same disease — **something reports success for work that did not happen**, so silence is indistinguishable from success.

Closes #1788 · Closes #2023 · Closes #2144 · advances #1971.

---

## #2023 — `Email:Enabled=true` + the NoOp sender stamped queued mail `Sent` (most urgent)

**Root cause.** Since the Graph sender moved into `MeshWeaver.Mail.MicrosoftGraph` (#1737), a deployment without the module resolves the `TryAddSingleton` NoOp fallback — but with `Enabled=true` the watchers *do* run. `OutboundEmailSender` then flipped every queued mail `New → Sending → Sent` while nothing left the process. No error, no log (the send lines are Information under a Warning default), and node data saying delivered.

**Now a refused configuration, in three places:**

| Where | What changed |
|---|---|
| `IEmailSender.DeliversMail` (new default member, `true`) | separates "a sender is registered" from "mail can be delivered" |
| `NoOpEmailSender` | takes `EmailOptions`; with `Enabled=true` it logs an **Error** and surfaces `OnError`, so `OutboundSendQueue` marks the mail **Failed** — never `Sent`. It still returns `true` for `Enabled=false`, the case it is actually for, and its log line no longer states a falsehood on the other path |
| `EmailDeliveryGuard` | `OutboundEmailSender` and `InvitationEmailSender` **do not start** on that configuration — in `StartAsync`, inside the startup window, so `StartupErrorNotifier` folds the refusal into the Admin bell |

Not starting is what keeps the mail **recoverable**: it stays at `New` and goes out by itself once the module lands. `Invitation.EmailSentAt` matters even more — that claim is permanent, and an invitation falsely stamped emailed can never be re-sent by any code path.

**Failing-first proven.** Against the pre-fix sender, 5 of the 10 new cases fail — including `Enabled_ButTheSenderCannotDeliver_QueuedMailEndsFailed_AndIsNeverStampedSent` (which ended `Sent`, exactly as reported in production) and `…TheOutboundWatcherDoesNotStart`.

**Is it live on memex.meshweaver.cloud? No.** Read-only evidence: `POST /api/email?validationToken=probe` returns **200 with the token echoed verbatim** — that is `EmailWebhookEndpoints`' Graph handshake, a route contributed only by `MailMicrosoftGraphEndpointModuleAttribute`. A non-existent `/api/` route on the same host answers 400 with an empty body. Both module attributes sit on the same assembly, so the endpoint being mapped means the mesh half's `AddMailMicrosoftGraph()` ran and `GraphEmailSender` beat the NoOp. `memex.systemorph.com` answers 200 the same way. This confirms the issue author's own withdrawal.

---

## #1788 — `memex-migration` CrashLoopBackOff in three prod namespaces

**The chart was already right** — `helm template` renders `kind: Job`, `restartPolicy: Never`, name per `.Release.Revision`, `ttlSecondsAfterFinished`. What kept the wrong model alive was the **prose**:

- `AGENTS.md` — the file every agent loads first — carried `set image` **and** `rollout restart` against a `memex-migration-deployment` the chart does not define, plus the instruction to treat the resulting crash loop as **benign**. That instruction is what made a real migration failure unreadable.
- `deploy/aks/scripts/deploy.sh` carried the same two commands; only the first was guarded with `|| true`, so a documented deploy always printed an error.
- `deploy/aks/envs/example/deploy.sh` scaled a Deployment that does not exist.

All removed. `MigrationWorkloadModelGuard` now pins the Job shape, that **no command anywhere in the repo** rolls the migration as a Deployment, and that no document teaches the crash loop as normal (a claim, not a rebuttal — the files that exist to refute the phrase keep working). Both assertions verified to have teeth by re-introducing the old lines.

**Deliberately NOT changed: the chart's self-update RBAC grant.** Removing `memex-migration-deployment` from `resourceNames` turns a 404 into a 403 while the Plugins-repo updater still patches that name — and on a namespace where the legacy Deployment is what actually applies migrations, that would stop the schema advancing while the portal rolls ahead of it, which `DbVersionGate` answers by refusing to start the portal at all. The orphan removal has to **follow** the updater, not lead it. `SELF-UPDATE.md` now says so explicitly, and also corrects its own claim that a chart-only env "logs a 404 and skips it" — `PatchDeploymentImageAsync` **throws** on any non-success status.

---

## #1971 — does the silo depart cleanly, and does anything pin the drain?

**Departure, from the rendered chart:** `DEADLINE=1680` (= `drainSeconds 1800 − shutdownMarginSeconds 120`) and `exit 0` at the deadline, so SIGTERM arrives with 120 s of grace left against the host's 90 s `ShutdownTimeout`. `ApplicationStopping` therefore fires and the silo departs membership in order. Post-#2043 the ceiling SIGKILL is structurally unreachable on the roll path, and `MapDrainEndpoint`'s `ApplicationStopping` line is the evidence — its absence after a termination is the hard-kill tell.

**The drain was NOT fully pinned.** #2043 pinned the chart's arithmetic and `DrainProgress` pins the log narrative, but nothing pinned the thing in between: that the portal maps `/drain` at all, what it answers, and that the chart probes a URL the code serves. `preStop`'s `curl -sf -m 5 -o /dev/null` cannot tell a 404 from a 503 from a refused connection — all three read as "still draining" — so a renamed route, a moved port, or a login-gated endpoint makes **every roll ride its window and then cut every open session**, with nothing in the log naming the cause. Same disease as the rest of this PR.

Now pinned:
- **`DrainEndpointTest`** — the real `MapDefaultEndpoints` composition over real HTTP: 200 `drained` with no circuits, 503 naming the count with circuits, anonymous behind a deny-by-default fallback (with a control route proving the fallback is real), and 200 on a host with no tracker at all.
- **`DrainDeadlineGuard`** gains the chart↔code contract: the probed **path** is a route the code maps and keeps `AllowAnonymous`, the probed **port** is a declared `containerPort` *and* `ASPNETCORE_HTTP_PORTS`, and the probe **tool** is installed in the portal base image — the `wget` scar itself, which shipped once and made every termination hang to the ceiling. Verified to have teeth by renaming the probed path.

**Still open** (system-changing, not done here): whether a *historical* ceiling SIGKILL left a zombie membership entry needs a live roll with a session held past the grace; and #1340's e2e — roll a disposable mesh mid-session and assert the session survives — still has no harness.

---

## #2144 — deployment docs that disagree with themselves

`DEPLOY-RUNBOOK.md` (465 lines) **folded into `deploy/aks/README.md`** and deleted; the README is retitled as what it is, a public, self-contained AKS deployment **sample**. All 14 inbound references repointed (including several in `.bicep`, `Dockerfile.logwatcher`, `main-cd.yml` and a chart template comment, none of which any test would have caught).

**The blocking open question, answered from evidence.** The P2S VPN gateway is still provisioned and **default-on** (`infra/main.bicep:79-90`), `aks.bicep` keeps local accounts enabled *for it*, and it is the only interactive route into the VNet (Grafana port-forward, `psql`). But **nothing automated depends on it**: zero VPN references in `.github/**`; every `deploy/aks` script runs its `kubectl` *inside* the cluster via `az aks command invoke --file`; `check-chart-drift.sh` takes an explicit `--via kubectl|aks-invoke` whose only CI caller passes `aks-invoke`. So: both supported, with a clear default — `az aks command invoke` is now **Step 2** with a when-to-use-which table, and the VPN moved verbatim to an optional appendix. A reader's second action is no longer provisioning a VPN they may not need.

**Stale facts corrected — each verified against the chart/code, never against other prose:**

| Was | Is | Evidence |
|---|---|---|
| env folders under `deploy/aks/envs/<env>/` | the private ops repo; `example/` here is the template only | moved 2026-08-08/09, `a69959165` |
| `Host=<PG_PRIVATE_IP>` | the server FQDN | servers moved 2026-08-24 |
| *"do not use the public FQDN form"* (`MemexCloudDeployment.md`) | **removed** — FQDN + password is correct | `AzurePostgres.UsesManagedIdentityAuth` needs the Azure suffix **and** an empty password; pinned by `AzurePostgresAuthSelectionTests` |
| `image.registry` in `values.aks.yaml` | `portal.image` / `migration.image` / `portalNext.image` | `grep '\.Values\.image' deploy/helm/` returns **nothing** — the block is dead |
| `kubectl patch … /spec/replicas 1` in `deploy.sh` | **removed** | re-created by hand the helm-vs-HPA contradiction of `check-chart-invariants.sh` invariant #1 — the 2026-08-14 production 503 |
| README's `portal-ha-patch.yaml` / `postgres-pvc-patch.yaml` step | **removed**, with why | the chart owns both halves now |

Also folded in, not lost: the OAuth `az ad app create` commands, the default-SSL-certificate fix, image-roll traps, the NodeType bake section, KEDA + the paused-scaler fence, crash dumps, the degraded-but-Ready replica, and known gaps.

---

## Verification

- **Chart change is two YAML comments.** `helm template` rendered against `origin/main` and diffed: **one changed comment line**, same 1120 lines, no key, value or structure touched. (Per the #2210 incident rule — rendered, not reasoned about.)
- `check-chart-invariants.sh`: all 3 values combinations render a self-consistent deployment.
- **Tests:** `Memex.Portal.Shared.Test` 417/417 · `MeshWeaver.Documentation.Test` 47/47 (incl. `DocumentationLinkIntegrityTest`) · `MeshWeaver.Auth.Test` 146/146.
- **Release + `-warnaserror` clean**, `0 Error(s)`, on every touched project plus `Memex.Portal.Distributed` and `Memex.Portal.Monolith`.
- **No live cluster state changed.** Production was only ever read (two `curl` probes and an anonymous `/api/version`).
- **No provider/model configuration touched.** `deploy/aks/values.aks.yaml` carries exactly one changed line — a comment at :188 repointing a doc reference — nothing `AzureFoundry__*` or provider-related.


rbuergi:
Addressed the review comment in `94b03a45f` — and it was a good catch worth naming, because it is the same defect class this PR is about.

`ScannedFiles()` filtered `deploy/` to `.md`/`.sh`/`.yaml` while `deploy/` genuinely carries `.yml` too (`deploy/whisper/docker-compose.yml`). So the guard **could not fail** for a whole class of input, and reported that as a pass — a gate narrower than its own claim, which is exactly the skip-trapdoor shape.

Now scans `.yml` as well, with a comment recording why. `MigrationWorkloadModelGuard` 3/3 green under `-c Release`, fresh `.trx`.

rbuergi:
CI red was **not this diff** — attributed by measurement, then fixed at the source rather than re-run.

Failing test: `MeshWeaver.Hosting.Monolith.Test.UserDirectoryContentShapeTest.EveryContentShape_IsIndexedAndProjects(shape: "typed")` (shard 1, run `32885660282`).

That is the `UserIdentityCache` type-confusion family — a NodeType *declaration* filed under the very type it declares, so it matched every `nodeType:User` query and handed callers a `NodeTypeDefinition` where a `User` was expected. **Root-caused and fixed by #2245**, merged as `0a0560501`.

This branch's head (`be1796fdd`) **predated that merge**, so it was re-sampling a flake main had already fixed — the case AGENTS.md names explicitly:

> *CI fails on something your diff does not touch. Merge main **before** investigating: a stale branch re-samples flakes main has ALREADY fixed, and each one looks like a defect in your change. PR #794 burned five CI runs and most of a day that way.*

So: branch updated onto main (now `903f4b1cb`) rather than re-run. I did **not** update the other in-flight PRs that also predate #2245 — being merely behind is fine here (`strict: false`), and merging main into every branch costs a full CI cycle each. They get updated only if they actually fail on this signature.

Issue #2276, opened 2026-08-25
MeshWeaver.AI leaves the platform: carve out the three non-AI surfaces, then move the engine to MeshWeaver.Plugins


rbuergi:
**Maintainer direction (2026-08-25):** *"all ai stuff deleted from core and all deployed and green."*

`MeshWeaver.AI` and `MeshWeaver.AI.Application` are the last AI in the platform repo. This issue is the program: the verified inventory of every compiled edge, the three surfaces inside `MeshWeaver.AI` that are **not AI** and must be carved out first, and the phased move to MeshWeaver.Plugins.

## What already left

`0b3b4266f` "Fourteen modules leave the platform" (2026-08-19) took the AI **provider packs** — `MeshWeaver.AI.OpenAI`, `.AzureFoundry`, `.ClaudeCode`, `.Copilot`, `.WebSearch` — to MeshWeaver.Plugins (Plugins#524). `Memex.Portal.Shared.csproj:105` documents the flip. The directories still visible under `src/MeshWeaver.AI.*/` in a dev tree are untracked `bin`/`obj` leftovers (`git ls-files` = 0 for all five); they are not in `MeshWeaver.slnx`.

## Why the engine stayed

That commit states its own criterion: *"Nothing in the platform's own `src/` referenced them. That is what made this a deletion rather than an untangling."* False for `MeshWeaver.AI` — it has **21 compiled consumers across 5 platform projects**, using ~40 distinct public types:

| Consumer | Files | Types used |
|---|---|---|
| `memex/Memex.Portal.Shared` | 10 | `ModelProviderNodeType`, `MeshOperations`, `IChatClientFactory`, `Harness`, `ConnectSessionManager`, `IConnectStrategy`/`IConnectTokenSink`/`IMcpBackConnection`, `SkillNodeType`/`SkillDefinition`/`SkillInfo`/`SkillAction`, `AiContentSources`, `AiCatalogLayoutAreas`, `TokenUsage*`, `ModelPrice*`, `IProviderKeyProtector`, `LanguageModelCatalog*`, `ProviderCredentialSeed`, `ThreadComposer`, `Thread` |
[...]

Issue #1716, opened 2026-08-17
edit_content is anchor-based but not ATOMIC — run the check-and-replace inside the owning hub's write turn


rbuergi:
`edit_content` is already ANCHOR-based — same contract as Claude Code's Edit tool (path, oldText,
newText, replaceAll; uniqueness-checked; 0-match and ambiguous-match refuse with instructive
errors; line numbers never part of the contract). What it lacks is **atomicity**: the anchor is
verified against a snapshot read OUTSIDE the owning hub's write turn, and the write that follows
ships a whole precomputed node. Scoped 2026-08-17 (maintainer asked for Claude-style anchored
editing for the mesh).

## Today's flow and the race

`MeshOperations.EditContent` (src/MeshWeaver.AI/MeshOperations.cs:1803-1892): `FetchNode` →
extract text (Markdown/Code/string) → `CountOccurrences` → `Replace` → rebuild node →
`mesh.UpdateNode(merged)` → `WaitForReadYourWrites`. The uniqueness check and replacement are
computed off the snapshot; `NodeUpdatePipeline` (src/MeshWeaver.Hosting/NodeUpdatePipeline.cs:79-107)
deliberately discards live content (`candidate = node with { Version = live.Version, ... }`).
Nothing re-checks the anchor against the state the owner actually holds. Two racing editors get
whole-content last-write-wins, softened only by the `MeshNodePatchMerge` string rebase. Worse:
`UpdateRemote` emits OPTIMISTICALLY after a 2s bound (MeshNodeStreamExtensions.cs:104) and an
exhausted retry only logs `LATE_NACK_TERMINAL` (:1377-1380) — the agent can be told "Edited: …"
for a write that never landed, and the retry re-ships the SAME stale full text.

The one correctly-atomic anchored edit already in the tree is
`CollaborationPlugin.SuggestEdit` (src/MeshWeaver.AI/Plugins/CollaborationPlugin.cs:157-216):
splice runs INSIDE `GetMeshNodeStream(path).Update(live => …)` and `Splice` THROWS when the
anchor is gone (:229-232) — but it replaces only the first occurrence and does no uniqueness
check.

## The fix (minimal — parts already ship)

**Run the anchored check-and-replace inside the write lambda**, SuggestEdit's shape:

1. Extract a pure static `ApplyAnchoredEdit(MeshNode live, string oldText, string newText,
[...]

PR #2283, opened 2026-08-25
MeshOperations is not AI — it leaves the AI assembly (#2276 step A1)

rbuergi:
First step of #2276 (*all AI deleted from core, deployed and green*).

The portal's REST API, the local-mesh sidecar's REST API and the agent/MCP tools all sit on ONE implementation of the mesh's get/search/create/update/patch/move/copy/recycle/export/upload surface — and that implementation lived in `MeshWeaver.AI`. So `Memex.Portal.Shared` compiled against the AI engine in order to serve `/api/mesh/*`. With the engine leaving for MeshWeaver.Plugins, this had to stop travelling with it.

`MeshOperations` (4,385 lines), `NodeReadOutcome` and `MeshExportManifest` move to a new `MeshWeaver.Mesh.Operations` assembly. **No behaviour change** — same instance methods, same error sentinels, same observable contract.

### The one agent-shaped member

`ResolveContextPath(IAgentChat, string)` read exactly `chat.Context?.Context` and nothing else, so it becomes `ResolveContextPath(string? contextPath, string path)`. `MeshWeaver.AI.AgentChatPaths` is now the single place that reads a context off a chat; the ~20 tool call sites go through it.

### Three members go public instead of befriending a module

`AnchoredReplace` + `AnchorNotFoundException` + `AmbiguousAnchorException` are the splice contract `CollaborationPlugin` shares with `ApplyAnchoredEdit` — #1716 unified them so the uniqueness rule cannot drift between the two edit surfaces. Consumed across an assembly boundary by a module, that is a public API, not an `InternalsVisibleTo`. The internals only the suites touch keep the same two friendships `MeshWeaver.AI` already granted.

### One namesake to disambiguate

`NodeReadOutcome` now has a namesake in `MeshWeaver.Mesh` (a different type — the read-classifier reason). They never collided while one sat in `MeshWeaver.AI`, because an enclosing-namespace match beats a using-directive. `NodeReadOutcomeClassifierTest` names which one it means with an alias.

### Verification

- Full solution build under CI's flags (`MW_ALLOW_SOLUTION_BUILD=1 dotnet build MeshWeaver.slnx -c Release -p:CIRun=true -warnaserror`) → `0 Error(s)`. A cross-assembly type move is structurally invisible to per-project builds, and this is what found the 40+ test files needing the using.
- `PackableProjectReadmeTest` passes (the new project ships a README — a packable project without one only fails at a public release tag).
- 92 unit tests over the moved code pass (`AnchoredEditContentTest`, `NodeReadOutcomeClassifierTest`, `ResolveContextPathTest`, `MeshOperationsMergePatchTest`, `PatchApplicableFieldsTest`, `DeleteErrorLineTest`, `ContentImagePathParsingTest`).

**No What's New entry**: pure-internal refactor, no user-visible effect.

Refs #2276


rbuergi:
**Design correction pushed: the namespace stays `MeshWeaver.Mesh`.** The assembly is still new; the namespace is not.

Raised by the plugins-side session while auditing what this series could break downstream. Plugin source is type-checked against **two different frameworks**:

| gate | compiles against |
|---|---|
| `compile-check.py --image` | core `main` — what the portal loads |
| the pack lane | the newest **RELEASED** framework (rc7, cut 2026-08-22) — a package's honest floor |

A brand-new `MeshWeaver.Mesh.Operations` namespace resolves on **neither** until a release moves the floor — and `publish-packages.yml` is `push: main` only, so the break would land on the plugins repo's **trunk**, not on the PR that caused it. That is live right now for an unrelated type (`FrameworkReleaseBroadcaster`, Plugins#686).

`MeshWeaver.Mesh` exists in rc7, so this move now needs **no cross-repo coordination at all**:

- The one real consumer of `MeshOperations` outside this repo is `MeshWeaver.Mcp`. It already carries `using MeshWeaver.Mesh;` and reaches the new assembly transitively through its existing `MeshWeaver.AI` reference.
- **Proven, not assumed** — built `src/MeshWeaver.Mcp` from the plugins repo against this branch (`-p:MeshWeaverRoot=/…/MW-ai-exit`): `0 Error(s)`, no plugin change.
- The other four plugin files that mention `MeshOperations` mention it in comments.

`NodeReadOutcome` → `NodeReadResult`, because in `MeshWeaver.Mesh` it would have collided with the read-classifier type already there. It is used only inside `MeshOperations`, so nothing else moves; the test that pins it is renamed with it.

Net effect on the diff: the ~40 `using MeshWeaver.Mesh.Operations;` lines the first commit added are **deleted** — consumers already import `MeshWeaver.Mesh`. Full solution build under CI's flags is green.

rbuergi:
Shard 2's red was **mine and real** — `FrameworkBuildIdentityTest.CanonicalList_MatchesTheTesterClosure`.

`FrameworkBuildIdentity.ContentSurfaceAssemblies` is the surface in-mesh CONTENT may compile against, pinned to the bake host's project closure. This PR's new `MeshWeaver.Mesh.Operations` assembly enters that closure through `MeshWeaver.AI` (which `MeshWeaver.PluginTester` references), so the canonical list had to gain it — otherwise the list and the closure drift, which is exactly what the gate recomputes from the csproj graph to prevent.

Worth recording how it was found, because none of my local verification could have: a full solution build compiles fine either way, and the thing that changed is the closure of a TOOL that the diff never names. Fixed, 23/23 in `FrameworkBuildIdentityTest`, full solution build green, and `origin/main` merged now that #2297 has landed.

rbuergi:
**Shard 0's two failures are the documented Orleans cluster, and the one that looks like mine is not.**

```
MeshWeaver.Hosting.Orleans.Test.OrleansMeshTests.HubWorksAfterDisposal(id: "Kernel")
MeshWeaver.Hosting.Orleans.Test.OrleansCompileActivityAccessTest.FrameworkStaleAssembly_SelfHealsOnInstanceActivation
```

`FrameworkStaleAssembly_SelfHealsOnInstanceActivation` deserved a real look, because this PR adds an assembly to `FrameworkBuildIdentity.ContentSurfaceAssemblies` and that changes the framework identity **value**. Read rather than assumed: the test never reads the canonical list. It compiles a NodeType, captures whatever `CompiledFrameworkVersion` the live host produced (`liveFv`), stamps a **bogus** one over it, and asserts the next activation self-heals. It is version-agnostic by construction — it compares against a value it generated itself, so a different-but-consistent framework identity is invisible to it.

`HubWorksAfterDisposal` is a disposal race, and both are on shard 0 — the shard-0 Orleans cluster that #1384 tracks as a flake population.

Also note shard 5 of this same run passed, including `MeshWeaver.Threading.Test`, which timed out on #2300 in the same window against the same main. The two reds pick different subsystems on different shards per run, which is the population's signature rather than a diff's.

`origin/main` merged (three merges landed since this branch was cut) so the re-run samples current main rather than a stale tree. The canonical-list fix from the previous push stands: `FrameworkBuildIdentityTest` is 23/23 locally.

PR #2287, opened 2026-08-25
Credential encryption is platform, not AI (#2276 step A2)

rbuergi:
Step A2 of #2276 (*all AI deleted from core, deployed and green*). Independent of #2283 — different files.

`IProviderKeyProtector` / `IMasterKeyProvider` / `ProviderKeyProtector` / `ConfigMasterKeyProvider` move from `MeshWeaver.AI` into `MeshWeaver.Mesh.Contract` (`MeshWeaver.Mesh.Security`).

They were never AI. Four unrelated subsystems already encrypt with them:

| Consumer | Secret |
|---|---|
| `ModelProviderService` | a model provider's `ApiKey` |
| `GitSync/GitHubCredential` | the committing user's GitHub PAT |
| `Memex.Portal.Shared/Authentication/EaGraphAuth` | the Entra EA credential |
| `PluginCatalog/SyncTokenSigningKeyService` | the sync-token signing key |

The last two reach the type only through `GitSync`'s transitive reference to the AI engine.

### The registration is the load-bearing half

It sat in `AddModelProviderType()` — behind `AddAI()`. Every consumer resolves the protector **optionally** (`GetService`, and on null stores the secret verbatim), so the day the engine leaves for a module a deployment without it would have written all four credentials **in plaintext and said nothing**: no exception, no warning, and the values still "work" afterwards because `Unprotect` passes untagged input straight through. `AddGraph()` registers it now, so it exists on any mesh built from the platform.

`ProviderKeyProtectorRegistrationTest` is the ratchet — and it was verified to FAIL with the two lines removed (`Expected value not to be <null>`), which also proves nothing else in the tree registers them.

### The config key does NOT get tidied

It stays `Ai:KeyProtection:MasterKey`. It is live in Key Vault and in the CSI `SecretProviderClass` mapping of every deployment, and `DeploymentOptions.md` records that deploying a *different* master key makes every stored `enc:` value undecryptable — including the live `Systemorph/Provider/AzureFoundry` key. Renaming it here would be a silent deletion of a running secret.

### Verification

- Full solution build under CI's flags (`MW_ALLOW_SOLUTION_BUILD=1 … -c Release -p:CIRun=true -warnaserror`) → `0 Error(s)`. That is what surfaced the three doc `cref`s and a fully-qualified `MeshWeaver.AI.IProviderKeyProtector` in `ThreadChatView`.
- 23 provider-key tests pass (`ProviderKeyProtectorTest`, `ProviderKeyEncryptionTest`, `ProviderCredentialSeedTest`, `ProviderCredentialSeedDecisionTest`) + the new ratchet.

### Follow-up, deliberately not in this PR

All four consumers still `GetService` + fall back to plaintext. With the registration now unconditional that fallback is unreachable in a real host — which makes it exactly the kind of fail-open worth deleting on its own, with its own reasoning about bespoke hosts. Tracked on #2276.

**No What's New entry**: pure-internal refactor, no user-visible effect.

Refs #2276

### DIFF
--- a/memex/Memex.Portal.Shared/Authentication/EaGraphAuth.cs
+++ b/memex/Memex.Portal.Shared/Authentication/EaGraphAuth.cs
@@ -1,12 +1,11 @@
+using MeshWeaver.Mesh.Security;
 using System.Reactive.Linq;
 using System.Reactive.Threading.Tasks;
 using System.Text.Json;
-using MeshWeaver.AI;                        // IProviderKeyProtector
 using MeshWeaver.Blazor.Infrastructure;     // PortalApplication
 using MeshWeaver.Data;                       // IWorkspace.GetMeshNodeStream
 using MeshWeaver.Graph.Configuration;        // EaCredentialNodeType
 using MeshWeaver.Mesh;                        // EaCredential, MeshNode
-using MeshWeaver.Mesh.Security;               // ImpersonateAsSystem
 using MeshWeaver.Mesh.Services;               // IMeshService
 using MeshWeaver.Messaging;                   // AccessService
 using Microsoft.Extensions.Configuration;
--- a/memex/Memex.Portal.Shared/Models/ModelProviderService.cs
+++ b/memex/Memex.Portal.Shared/Models/ModelProviderService.cs
@@ -1,3 +1,4 @@
+using MeshWeaver.Mesh.Security;
 using System.Collections.Immutable;
 using System.Reactive;
 using System.Reactive.Linq;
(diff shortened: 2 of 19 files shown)


rbuergi:
**The `Landed modules compile against this PR` red was stale input, not a defect — re-run, and here is the evidence rather than the assertion.**

The failure is `CS0234: IProviderKeyProtector does not exist in the namespace MeshWeaver.AI` from `plugins-repo/src/MeshWeaver.Blazor.Portal` — real, and exactly what this PR changes. Systemorph/MeshWeaver.Plugins#685 makes that file resolve the type from **either** namespace (it already carries `using MeshWeaver.Mesh.Security;`, so dropping the `MeshWeaver.AI.` qualifier compiles on both sides of the move, in either merge order).

The timestamps are the whole story:

| | |
|---|---|
| this run checked out plugins `main` | **22:26:04Z** |
| Plugins#685 merged | **23:18:36Z** |

So the gate compiled the plugins tree 52 minutes before the fix existed. The gate is right; its input was old. Re-running is what changes the input — it checks out plugins `main` fresh — rather than a retry hoping for a different roll of the dice.

Worth stating because the two look identical from the outside: **this is not "re-run to see if it was a flake"**, which this repo forbids for good reason. The input to the job provably changed between the two runs.

PR #2292, opened 2026-08-25
GitSync stops depending on the AI engine (#2276 step B2)

rbuergi:
Step B2 of #2276. **Stacked on #2287** (A2) — base retargets to `main` automatically once that merges.

A module cannot be injected into a platform that compiles against it. `MeshWeaver.GitSync` referenced `MeshWeaver.AI`, and through it so did `MeshWeaver.PluginCatalog` — **transitively**, which is why no grep of direct `ProjectReference`s ever showed that edge. This inverts it.

### What moves

`AiContentDiskWriter`, `AiContentSyncArea`, `AiContentSyncMenuProvider`, `HubAiContentSyncExtensions` and the agent-backed PR drafter → `MeshWeaver.AI`. **None of them imported a GitSync namespace**: they were AI features filed under the wrong project. `MeshWeaver.AI` now references `MeshWeaver.GitSync` for the repo client, which is the allowed direction (a module may reference the platform, never the reverse).

Their registrations move with them — the disk writer, the "Sync to repo" node-menu provider and its area now ride `AddAI()`.

### The drafter splits at an interface, and PR creation stops depending on AI

`IPullRequestDraftService` + the `PullRequestDraft` record stay in GitSync; the agent implementation goes to AI.

| | registration | result |
|---|---|---|
| GitSync | `TryAddSingleton<…, PlaceholderPullRequestDraftService>` | deterministic title + "edit this before submitting" body |
| `AddAI()` | `AddSingleton<…, PullRequestDraftService>` | the agent-drafted title/body |

Order-independent — the module wins whichever configures first (the `IEmailSender` shape). Without AI you now get the placeholder rather than a DI resolution failure, which is what the call site's `.Catch` already produced when no model was configured.

### Two consumers gain an explicit reference, and that is the point

- `MeshWeaver.PluginCatalog.Test` installs Agent/Skill packages and asserts on the pickers — a real AI dependency that had been arriving through GitSync.
- `AiContentSyncBackTest` gains `.AddAI()`. **It failed without it** (`AiContentDiskWriter has not been registered`), which is the proof the seam is real rather than nominal.

### Verification

- Full solution build under CI's flags (`-c Release -p:CIRun=true -warnaserror`) → `0 Error(s)`.
- `MeshWeaver.GitSync.Test`: 182 tests, 0 failed.

After this and #2287, `MeshWeaver.GitSync`'s dependency on the AI engine is **zero**.

**No What's New entry**: internal refactor. The one behaviour change — a PR drafted without AI gets a placeholder instead of failing — is strictly an improvement to a path that previously required a configured model.

Refs #2276

### DIFF
--- a/src/MeshWeaver.AI/AIExtensions.cs
+++ b/src/MeshWeaver.AI/AIExtensions.cs
@@ -2,6 +2,7 @@
 using MeshWeaver.AI.Persistence;
 using MeshWeaver.AI.Plugins;
 using MeshWeaver.Data;
+using MeshWeaver.GitSync;
 using MeshWeaver.Hosting.Persistence.Parsers;
 using MeshWeaver.Domain;
 using MeshWeaver.Layout;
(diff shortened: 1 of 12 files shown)


rbuergi:
Same stale input as #2287, same evidence: this run checked out plugins `main` at **22:26:05Z**, and Systemorph/MeshWeaver.Plugins#685 — which makes `ThreadChatView` resolve `IProviderKeyProtector` from either namespace — merged at **23:18:36Z**. This PR is stacked on #2287, so it inherits the namespace move and therefore the same 52-minute-old input.

Re-run triggered, which is what changes the input (a fresh checkout of plugins `main`) rather than a retry hoping for a different outcome.

Issue #2169, opened 2026-08-24
Extract the GUI implementations from the core repo: RN app + Blazor packs move to MeshWeaver.Plugins


rbuergi:
Maintainer direction (2026-08-24): client GUI implementations should not live in the platform repo — "why do we still have RN PRs in the core repo? should be factored out by now… also Blazor". The core repo keeps the platform + the SDK libraries (`clients/react` = @meshweaver/react, `clients/grpc-web`, `clients/portal-next` which bakes into the portal image); the deployable GUI implementations move to MeshWeaver.Plugins, following the lane 25+ compiled modules (AI packs, Radzen, Analysis, maps) already took.

## Phase A — the RN app → MeshWeaver.Plugins/app/react-native (unblocked, in progress)
1. **Plugins PR**: add the app tree. Cross-repo source imports (`@meshweaver/react`, `@meshweaver/client-web`) resolve via tsconfig path FALLBACK arrays (CI's nested `meshweaver/` checkout first, sibling `../MeshWeaver` dev checkout second) + dynamic resolution in metro/vitest configs. New `rn-app` CI job: nested core checkout → grpc-web gen → typecheck + vitest (parity ratchet reads the core checkout).
2. **Core PR** (after 1 merges): delete `clients/react-native`; drop the RN job + trigger entries from clients.yml; `Memex.LocalMesh`'s `BakeRnWebClient` gets an overridable `RnClientDir` probing the old path then the sibling `MeshWeaver.Plugins/app/react-native` (still fails RED when neither exists — no skip-trapdoor). No core workflow publishes the sidecar Release, so core CI gains no cross-repo dependency.

## Phase B — Blazor packs' sources follow
- **B1 (core)**: break the compiled edges — `Memex.Portal.Shared` drops its `Blazor.Graph`/`Blazor.Portal` ProjectReferences and the compiled `.AddGraphViews().AddChatViews().AddUserProfileViews()` calls; those packs plus `Blazor.Views` ship via the `modules/<Name>/` image layout (the #1644 AI-provider shape) and activate via `Modules:Assemblies`. This takes them out of the app CLOSURE, clearing the 409 constraint (closure DLL vs registry module are mutually exclusive).
[...]

PR #2293, opened 2026-08-25
Delete Blazor and the portal GUI from core

rbuergi:
**Stacked on #2285** (the infrastructure lift), which has to land first — this deletes the projects
that lift empties of platform code.

Core ends up with no GUI at all. **389 files, 54,428 lines deleted.**

```
src/MeshWeaver.Blazor                  112 files
src/MeshWeaver.Blazor.Portal           102
src/MeshWeaver.Hosting.Blazor           10
memex/Memex.Portal.Shared               29 GUI files  (77 platform files STAY)
memex/Memex.Portal.{Monolith,Distributed}, Memex.AppHost   — the hosts
test/MeshWeaver.Hosting.Blazor.Test     32  (it tests deleted code)
```

## 🚨 The test sweep was wrong three times before it was right

Selecting test files by *"imports a Blazor namespace"* is a proxy for *"is a view test"*, and the two
are **different sets**. That predicate selected 27 files. The compiler says **16** actually need a
view type. Among the 11 it would have exiled for an unused `using`:

> **`OrleansTopLevelPartitionGuardTest`** — the top-level partition guard. It imports `Blazor.Portal`
> for silo wiring whose extensions (`ConfigureMeshWeaverServer`, `UseOrleansMeshServer`) **all live in
> core**. It compiles fine without the import.

Eight `MeshWeaver.Hosting.PostgreSql.Test` files were the same shape. **Name-matching fared no
better** — it misses `MessageBubbleState`, `CompletionItem` and `MarkdownCellEditing`, view types
that don't end in `View` or `Page`.

What works is neither: **strip the import and let the compiler classify.** Every file that stays lost
one stale `using` and nothing else — which is why the diff shows `PartitionAccessTest.cs | 1 -`
rather than a deletion.

## Two checks that earned their place

- **Test-count diff** across the eight affected suites: **−244 removed against 244 expected**, exact.
  It took three attempts to measure correctly (wrong checkout, then `obj/` copies) — and each wrong
  version showed a suite *gaining* tests from a deletion, which is the only reason I caught them.
- **Guard/ratchet name check**: named the partition guard before the compiler would have.

## The ratchet that was one predicate away

`NodeTypeDeclarationSelfTypingTest` **stays in core and still runs** — verified `build exit=0`, then
2/2 passed (an earlier run reported "Passed" while the build had exited 1, i.e. against a stale DLL;
that result was discarded). It guards a defect measured at **3.2 failures/second on every serving
pod**, and its absence is silent by construction.

## Two stragglers no source-level check could see

MSBuild `<Content Include>` items in `Acme.Test` copying CSS out of the deleted projects. Not
`using`s, not `ProjectReference`s, not C# — **only a full-solution build surfaces them.**

## Verification

Full solution: `-c Release -warnaserror`, **0 warnings, 0 errors**. The final pass was incremental,
so CI's clean checkout remains the real test of a from-scratch build.

No What's New entry: no user-visible change — the portal builds from a different repo and nothing
about it changes for a viewer.


rbuergi:
🚧 **Back to draft — this must not merge yet.**

It deletes 77 test files plus `MemexConfiguration.cs` from core, and those are **not yet present in MeshWeaver.Plugins**. They are staged locally, which is not the same thing. Merging as it stands would delete the coverage rather than move it — which is precisely the failure this PR's own description argues against.

Remaining before it is mergeable:
- place the 77 staged test files in plugins (they span 11 former core suites and need homes: the `Blazor.Portal` ones, the portal-Gui ones, and `MeshWeaver.Hosting.Blazor.Test` which moves whole)
- `MemexConfiguration.cs` is now placed in `Memex.Portal.Gui` on Plugins#674, but that PR has to land first

The core-side deletion itself is verified and unchanged: full solution builds clean under `-c Release -warnaserror`.

rbuergi:
## Heads-up: this PR silently drops MeshWeaver.AI from the image

Not a review of the Blazor removal itself — a specific consequence I found while working #2276, with the fix.

`MeshWeaver.AI.dll` reaches the portal image through exactly **one** chain:

```
Memex.Portal.Distributed → Memex.Portal.Shared → src/MeshWeaver.Blazor.Portal → MeshWeaver.AI
```

`src/MeshWeaver.Blazor.Portal` is the **only** project in `src/` or `memex/` referencing `MeshWeaver.AI.csproj` (verified against `origin/main`). Deleting it takes AI.dll out of the app closure.

What happens then:

- `MeshWeaver.AI` is **not** in `MeshModule` / `MeshModuleClosure` (`memex/MeshModulesPublish.targets`), so nothing else lays it down;
- `ResolveModulePath` finds nothing, and `Modules:Assemblies` is **skip-and-continue** — one stderr line;
- it is in `Assemblies`, **not** `Required`, so the required-modules health check stays **Healthy**.

Net: a **green rollout of a portal with no AI** — chat, agents, models absent, no failing probe.

### The fix — one line, your choice of two

Both are legitimate; which one depends on where you intend AI to come from after the split:

1. **Keep it in the image** — add `MeshWeaver.AI.csproj` to `MeshModuleClosure` in `memex/MeshModulesPublish.targets`, joining `MeshWeaver.Social`. This is the documented flip for "module whose ProjectReference is GONE", which is exactly what this PR makes it.
2. **Registry-delivered** — move `MeshWeaver.AI.dll` from `Modules:Assemblies` to `Modules:Required` in both host appsettings, so an instance without the package fails readiness and the rollout stalls instead of silently losing AI.

(1) is the smaller step and keeps this PR self-contained; (2) is where #2276 ends up eventually.

### You will not have to remember this

I have opened **#2347**, a guard that fails CI naming exactly this. Once it lands, your branch goes red with the message above rather than merging quietly — so please expect it, and treat it as the checklist rather than an obstacle. I verified it genuinely fails by simulating your deletion locally: red on **both** hosts, green again when restored.

I have deliberately **not** pushed the one-line fix into your branch — where AI ships from is your call, and this PR is stacked on `refactor/lift-nonvisual-infrastructure-out-of-blazor`, so I would rather not touch the chain.

rbuergi:
## The guard is on main now (#2347 merged)

Following up on my earlier comment: the check is live, so this branch will go **red** on its next run with:

```
memex/aspire/Memex.Portal.Distributed/appsettings.json: 'MeshWeaver.AI.dll' names a project that
EXISTS in this repo (src/MeshWeaver.AI) but nothing ships it into this host's image — it is not in
the ProjectReference closure and not in the modules/ lane (memex/MeshModulesPublish.targets). Listed
under Modules:Assemblies it will be SKIPPED silently and the feature will be absent from a GREEN
rollout. Either restore the reference / add it to MeshModule(Closure), or move it to Modules:Required
so the health check stalls the rollout instead.
```

…and the same for `Memex.Portal.Monolith`.

That is the intended behaviour, not an obstacle: before #2347 this PR would have merged green and shipped a portal with **no AI at all** — chat, agents and models absent, no failing probe — because `Modules:Assemblies` is skip-and-continue by design and `MeshWeaver.AI.dll` is not declared `Required`.

**One line clears it**, whichever matches your intent:

1. **Keep AI in the image** — add `MeshWeaver.AI.csproj` to `MeshModuleClosure` in `memex/MeshModulesPublish.targets`, beside `MeshWeaver.Social`. This is the documented flip for a "module whose ProjectReference is GONE", which is exactly what this PR makes it.
2. **Registry-delivered** — move `MeshWeaver.AI.dll` from `Modules:Assemblies` to `Modules:Required` in both host appsettings, so an instance lacking the package fails readiness instead of silently losing AI.

(1) is smaller and keeps this PR self-contained. (2) is where #2276 lands eventually.

Still deliberately not pushing into your chain — this is stacked on `refactor/lift-nonvisual-infrastructure-out-of-blazor` and where AI ships from is your call.

rbuergi:
## Correct — and it is the exact class of debt this PR spent the day fixing elsewhere

Line 56 reads *"Blazor.Portal REMAINS a compiled reference for now — it is SHELL infrastructure (PortalLayoutBase, auth, viewport), tracked for extraction as #2169 Phase B3."* This PR completes that extraction, so the sentence describes a state that stops existing at merge.

That is the same failure mode as `ContinuousDeliveryContract.md`'s two wrong predictions, which this PR also corrects: **prose describing intent as if it were state**, sitting next to the code and outliving it. Worth fixing, not worth arguing about.

**Sequencing, stated plainly rather than resolved silently:** I am merging this first and fixing the comment in an immediate follow-up. The reason is that `main` is currently RED on `Landed modules compile against this PR` — because #674 landed the additive half of this move while the deletion half stayed open — and it is blocking at least three unrelated PRs from another session. This PR is what unbreaks it. Pushing a comment-only change costs a full ~25-minute CI cycle and keeps main red for the duration, to fix a sentence.

That trade is worth naming because the opposite call is also defensible: a stale comment merged knowingly is debt, and I have spent today arguing that prose nobody verified is how the next person loses an afternoon. I am taking the merge because the cost here is measured in one wrong sentence, and the cost of waiting is measured in other people's blocked work.

The follow-up lands within minutes of this merging, and I will link it here.

rbuergi:
Follow-up landed as promised: #2360 corrects the `Blazor.Portal REMAINS a compiled reference` comment in `Memex.Portal.Shared.csproj`.

Issue #595, opened 2026-07-21
Cancelled and errored rounds record no token usage — consumed tokens are lost from accounting


sierragolflima:
Follow-up to #369.

## Summary

When a round terminates as **Cancelled** or **Error**, the tokens the provider has already consumed are not recorded. No `TokenUsage` satellite is written, and an existing one is not advanced. The tokens are invisible in both the per-model roll-up and the cost total.

This was reported as the second grade of #369 (follow-up comment, 2026-07-08). That issue was closed on the first grade only — PR #519 corrected sub-thread model attribution and did not touch these paths. Grade 1 has since been verified fixed on the production portal (a delegation sub-thread carries its own `_Usage` node, keyed to a real model, resolving to a non-zero cost), so this issue is strictly about the remaining terminal paths.

## Root cause

The round's token counters (`inputTokens`, `outputTokens`, `cacheReadTokens`, `cacheWriteTokens`) are locals declared at `ThreadExecution.cs:1878`, populated inside the streaming loop from `UsageContent` blocks on streamed `ChatResponseUpdate` values.

OpenAI-compatible providers emit usage **only in the terminal chunk**. A cancellation or a transport failure raises before that chunk arrives, so all four counters remain null. `TokenUsageNodeType.RecordUsage` returns without writing when all four are zero or null (`TokenUsage.cs:147`), so nothing is persisted and nothing is logged.

The comment at `ThreadExecution.cs:2344` states the intended behaviour — *"Record tokens consumed BEFORE the cancel — the streaming loop already aggregated any UsageContent seen prior to the OperationCanceledException"*. That assumption holds only for providers that emit incremental usage during the stream. It does not hold for the OpenAI-compatible path.

**Secondary defect on the same lines:** the Completed path passes `actualModel ?? request.ModelName` (`:2281`), but Cancelled (`:2374`) and Error (`:2482`) pass bare `request.ModelName`. A delegation sub-thread carries a null `request.ModelName`, so even when these paths *do* record, they would key the entry `(unknown)` — the residual slice of #369 grade 1 that PR #519 left in place.

## Evidence — reproduced live

Local portal, build `9ee22ee1d`, provider OpenAI-compatible (Ollama), all rounds in one thread. The satellite accumulates across a thread's rounds, which makes it a direct control.

[...]

PR #2295, opened 2026-08-25
A live query's early→live handoff dropped a change notification permanently

rbuergi:
## A live query's early→live handoff dropped a change notification — permanently

Every live children/descendants query opened **two** subscriptions on the adapter's change feed: an
"early" one that buffered into a backlog while `!initialDone`, and a "live" one attached inside the
initial-results callback, after which the early one was disposed.

`IsolatedChangeFeed` snapshots its observer list **before** delivering — it must, because
subscribers attach and detach mid-publish. So a write published in the instant before the live
subscription was attached is delivered against a snapshot holding **only the early observer**, and
by the time that delivery ran the callback had already set `initialDone = true`. The early
handler's `if (!initialDone)` threw the notification away; the live observer never saw it because it
was not in the snapshot; and **nothing re-triggers the re-query**. The row stayed invisible to that
subscription for its whole life — while a point read, and any *fresh* subscription, returned it
immediately.

A permanent miss dressed as a timeout, which is why no budget could ever have fixed it.

## Measured, not inferred

This started as an investigation into whether **cancelled agent rounds go unbilled** — the writer
hypothesis #2244 pre-registered when its reader-side fix did not stop the `WaitForUsage` reds.
**The writer is exonerated.** Instrumenting the wait to read the node back at failure time, and
reproducing locally (`MeshWeaver.AI.Test` 4-wide with `DOTNET_PROCESSOR_COUNT=4`, one `dotnet test`
at a time, ~1 failure in 8 runs):

```
[USAGE-DIAG] leg1 emissions during the wait: 1
[USAGE-DIAG]   leg1 21:37:06.565 n=0 []
[USAGE-DIAG] listing(_Usage children) count=1   ← a FRESH listing, 3 ms later
[USAGE-DIAG]   listed …/_Usage/usage_model content={"inputTokens":137,"outputTokens":89}
[USAGE-DIAG]   cell e4533be4: status=Completed model=usage-model in=137 out=89 total=226
[USAGE-DIAG] POINT READ OK: …/_Usage/usage_model {"inputTokens":137,"outputTokens":89}
```

The satellite is written, with the right counts, keyed by the right model — on the Completed **and**
the Cancelled path. The test's own live subscription emitted exactly one snapshot, the empty
Initial, then nothing for the full 20 s.

Subscribing the **same shared upstream** under System (which makes `WrapWithPerUserRls`
short-circuit to the raw, unfiltered stream) exonerated the per-user RLS filter too — raw saw
exactly what filtered saw. Instrumenting the provider then named the culprit outright:

```
[QDIAG] SUBSCRIBE base=…/_Usage scope=Children
[QDIAG] INITIAL   base=…/_Usage items=0 backlog=0
[QDIAG] NOTIFY seen base=…/_Usage path=…/_Usage/usage_model match=True   ← EARLY observer, after the handoff
```

…and, for that base path, **zero** `LIVE seen` and **zero** `REQUERY` for the rest of the run.

## The fix

**One** subscription for the query's whole lifetime. The backlog-vs-live routing decision is made
inside the same critical section that publishes the buffer, so a notification lands in exactly one
of the two and can never fall between them. The same shape existed in all four providers —
`StorageAdapterMeshQueryProvider`, `PostgreSqlMeshQuery`, `PostgreSqlPartitionedMeshQuery`,
`CosmosMeshQuery` — and is corrected in all four. It also removes the double-capture window the old
comments accepted, and puts Cosmos's disposal order right as a side effect.

No budget was widened, no retry or watchdog added.

## The test

`LiveQueryHandoffDropTest` pins the window deterministically: it holds the query's scope walk open,
commits the write, takes the notification's fan-out snapshot while only the pre-Initial subscriber
exists, releases the walk, asserts the Initial is genuinely empty, and only then delivers.

- **Fails on the pre-fix code** (verified by reverting the provider: 10 s, the assertion's message).
- **Passes on the fix** in 43 ms.

`MeshWeaver.Hosting.Test` is 226/226 green.

## Scope note

Related but **not** fixed here, and deliberately kept separate: a round whose provider never emitted
usage before the cancel/error records **zero** tokens, silently (a Debug-level no-op, no tokenizer
or estimate anywhere in the tree). That is real unbilled spend and it is #595, which I reopened —
its last comment said *"Keeping this issue open, scoped to that estimate"* and it was then
auto-closed as COMPLETED by a PR that shipped only the two surgical items.

Also noticed while reading the logs, not changed here: `MeshQuery.InitialStallProbeDelay` is 20 s
and `ThreadTokenUsageTest.SettleBudgetMs` is 20 000 — so in CI run `32894972304` the one warning
that would have named a stalled query provider fired ~30 ms after `TEST END`. A stall probe armed at
exactly the consumer's budget can never explain that consumer's timeout.

Root-causes the `WaitForUsage` flake family closed prematurely as #1040, #1812 and #2001 — each of
those diagnosed a different reader shape and none reached the notification drop underneath. Reported
on #1384 (the flake-population umbrella), which stays open.

### DIFF
--- a/src/MeshWeaver.Documentation/Data/WhatsNew/2026-08-26-live-lists-stop-missing-a-just-created-node.md
+++ b/src/MeshWeaver.Documentation/Data/WhatsNew/2026-08-26-live-lists-stop-missing-a-just-created-node.md
@@ -1,0 +1,50 @@
+---
+Name: Live lists stop missing a just-created node
+Category: Fix
+Description: A view watching a folder could permanently miss a node created a fraction of a second after it opened — the token counter reading zero, a notification never appearing — while opening the same thing fresh showed it immediately. The change notification was being dropped in a handoff; now it is never dropped.
+Icon: List
+Order: -20260826
+---
+
+# Live lists stop missing a just-created node
+
+A view that watches a set of nodes — the chat token counter over a thread's usage, a notification
+bell, a folder listing — could **miss a node created moments after the view opened, and never
+recover**. Not show it late: never show it, for as long as that view stayed open, while opening the
+same list again showed the node immediately.
+
+The give-away was how strange it looked. The node was written, correct, and readable: ask for it by
+path and it came back at once. Only the view that had been watching all along failed to see it.
+
+## What was happening
+
+Every live list works the same way: it takes a snapshot, then follows the change notifications that
+follow. Until now it did that with **two** subscriptions to the change feed — one that buffered
+notifications while the snapshot was still being taken, and a second that took over once it was
+done, at which point the first was closed.
+
+The change feed decides who a notification goes to *before* it delivers it, which it must, because
+subscribers come and go while a notification is on its way. So a write published in the instant
+before the second subscription existed was addressed only to the first — and by the time it arrived,
+the handoff had happened and the first subscription had been told to stop buffering. It discarded
+the notification. The second never received it, because it had not been on the list when the
+notification set out.
+
+Nothing asks again after that. The list re-reads only when a notification arrives, so a lost
+notification means a lost row, permanently.
+
+## What changed
+
+A live list now keeps **one** subscription for its whole life. It buffers, then switches to
+following — and the decision about which of the two a notification belongs to is made in the same
+indivisible step that performs the switch. There is no longer an in-between for a notification to
+fall into.
+
+The same handoff existed in all four storage backends and was corrected in all of them.
+
+## What you will notice
+
+Nothing, most of the time — the window was small. What it removes is a rare, confusing class of
+"the page is stale and reloading fixes it": a token counter stuck at zero after a reply that plainly
+used tokens, a folder that does not show a file that is certainly there, a list that quietly stopped
+keeping up. Token usage was always recorded correctly; only the reading of it could go missing.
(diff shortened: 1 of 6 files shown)


rbuergi:
### Controlled before/after on the reproducing loop

The deterministic `LiveQueryHandoffDropTest` is the proof (it fails pre-fix, passes post-fix). This
is the corroboration: the **same** 5-class filter, `DOTNET_PROCESSOR_COUNT=4`, one `dotnet test` at
a time, pre- and post-fix on the same machine.

| build | iterations | `WaitForUsage`-family failures |
|---|---|---|
| pre-fix (first runs, machine under load ~40–100) | ~25 | **3** (reproduced at iteration 3, 8 and 14 across three runs) |
| pre-fix control (same load as the post-fix runs) | 70 | **1** — `CompletedRound_StampsTokensOnResponseCell_AndAccumulatesOnThread` |
| post-fix run 1 | 65 | **0** |
| post-fix run 2 | 70 | **0** |

**0 in 135 post-fix, 4 in ~95 pre-fix.** Note the pre-fix control's hit is a *Completed*-round test —
more evidence the flake was never cancel-specific.

### One unrelated failure, attributed rather than ignored

Post-fix run 1 also saw one
`AgentToolCancellationTest.GetVersions_ParkedOnAStalledStore_UnwindsWhenTheRoundIsCancelled`
(`Expected 2 to be greater than 2 … cancelling must dispose the version read`). It did **not** recur
in the further 70 post-fix iterations, and it is a static-counter timing assertion (`VersionStore.
Disposals` must have advanced by the instant the `OperationCanceledException` is observed) with no
path to the change-feed handoff this PR touches. Recording it here rather than leaving it unsaid:
1-in-135, independent of this change as far as the evidence goes.

### Note on the base

This branch is cut from `ed4588222`, where `AiContentPackDriftTest.EveryFile_StillHashesToItsPinned
ReconciliationPoint` is **red on `main`** (`content/ai/Skill/group.md` changed in `123f6734c` without
re-pinning `AiContentPackSync.json`). #2288 fixes it; once that lands I will merge `main` in rather
than investigate a red this diff cannot reach.

Issue #1644, opened 2026-08-15
DESIGN: the modules/ publish layout — cutting the ships-the-bits ProjectReferences


rbuergi:
The module lane is complete on the ACTIVATION side (every pack self-registers via `MeshNodeProviderAttribute`; `Modules:Assemblies` is the on/off switch), but every module still rides a **ProjectReference from Memex.Portal.Shared** whose only purpose is landing the DLL in the publish output. This design cuts those references — the step the extraction queue called *"modules-folder publish"* and the precondition for #1630's bundles becoming the distribution channel for first-party packs too.

## Goal
A deployment's image = the base portal + a `modules/` folder; which modules EXIST is a publish/composition decision, which ACTIVATE stays `Modules:Assemblies`. `ReferenceOutputAssembly=false` was already proven NOT to work (drops the DLLs) — the publish target is the missing piece.

## Design

**1. Publish layout.** A `PublishModules` MSBuild target on each host publishes every module project into `modules/<ModuleName>/` beside the app: the module DLL + the part of its transitive closure NOT in the app's own closure (`dotnet publish` per module, then prune assemblies whose identity exists in the app output — same-identity duplication is the documented trap-door class, so pruning is a MUST, and the target FAILS if a listed module is still ProjectReferenced by the host).

**2. Loading.** `Modules:Assemblies` entries resolve BaseDirectory-relative today; extend the resolution to probe `modules/<name>/<name>.dll`. `Assembly.LoadFrom` (Default ALC) probes the module's own folder for its private deps — no ALC work needed.

**3. In-mesh compile surface.** Module DLLs leave the publish closure ⇒ leave `TRUSTED_PLATFORM_ASSEMBLIES` ⇒ in-mesh consumers (e.g. Cornerstone's `MapControl`) lose them. Fix: `InstallAssemblies` records each loaded module (an `InstalledModuleAssembly` enumerable singleton), and `MeshNodeCompilationService` composes its reference list as the static TPA baseline + the mesh's installed modules. (Deliberately NOT done while modules ride the image — it becomes necessary exactly here.)

[...]

PR #2297, opened 2026-08-25
The AI application folds into the engine — one assembly, one module, one Store entry (#2276)

rbuergi:
Installing AI must install the app that maintains agents, models, skills and harnesses. **One entry in the Store, not two.** `MeshWeaver.AI.Application` becomes `MeshWeaver.AI/Application/` — namespaces unchanged (they were already sub-namespaces of the engine), only the assembly boundary goes.

## The application was dead code, and nothing said so

It carried `[assembly: AgentsApplication]`. But a `MeshNodeProviderAttribute` is read in **exactly one place** — `MeshBuilder.InstallAssemblies`, i.e. the `Modules:Assemblies` lane — and that assembly was never listed there. It rode the app closure through a `ProjectReference` from `MeshWeaver.Blazor.Portal`, so it shipped in every image contributing **nothing**: no Agents hub, no agent overview, and its two autocomplete providers registered on a container that was never built.

`ThreadChatView` carries the scar. It news up `SkillAutocompleteProvider` by hand, with a comment explaining that resolving it from DI *"returned null and typing '/' showed nothing"*.

Both portals now list `MeshWeaver.AI.dll` under `Modules:Assemblies`, so the attribute is read and the application is contributed by the same load that brings the engine. The DLL still rides the app closure, so `ResolveModulePath` falls back to the app folder — the documented **thin lane**, the same step #1644 used for every module that has since left.

## Turning it on surfaced a platform hang

`DataExtensions.HandleAutocompleteRequest` took `LastAsync()` — which emits only when the combined stream **completes**. A provider backed by mesh data never completes: it is a live subscription by design (`SkillAutocompleteProvider` composes `ObserveSkillQueries` → `ObserveSnapshot` → `Switch`). So on any hub with such a provider registered, the handler ran, returned `Processed` in 27 ms, and posted **nothing**. The caller waited out its timeout against silence — no error, nothing to grep.

Read straight off the message trace, per the `/debug` procedure:

```
:35.082  HUB_HANDLE_START | AutocompleteRequest | Hub: app/Agents
:35.110  HUB_HANDLE_END   | AutocompleteRequest | Hub: app/Agents | Result: Processed
         … 30 s, no response post from that hub …
```

A burst of work then total silence — a missed observation, not a lock.

**"Settled" is now a quiet period, never completion.** First of three wins: a 150 ms throttle; a 2 s deadline answering with the best snapshot so far; a 2.5 s floor answering empty if no provider ever produced anything. A one-shot request must produce exactly one answer — a hang is indistinguishable from "still thinking". One shared `RefCount` subscription, so the providers' queries are not issued twice.

The AI application's **duplicate** handler is deleted with it: the platform's copy has relevance scoring this one never had, and a second registration for one request type is how the two drifted.

The round trip that timed out at 30 s now answers in well under a second.

## Verification

- Full solution build under CI's flags → `0 Error(s)`.
- `MeshWeaver.Autocomplete.Test` — 155/155.
- `MeshWeaver.Data.Test` — 385/385.
- New `AgentsApplicationModuleTest` (the attribute travels with the engine and names the app) and `AgentsApplicationInstallTest` (installing the assembly the way `Modules:Assemblies` does **builds the hub and it answers**). The second one failed before the handler fix, which is what makes it a gate rather than a formality.

## What this sets up

`MeshWeaver.AI` is now a single assembly carrying the engine *and* its maintenance application, activated through the module lane. That is the shape a Store package needs: one `content.module`, one entry, `Modules:Required` once it ships from the registry.

Refs #2276


rbuergi:
**Blocked on Systemorph/MeshWeaver.Plugins#685 — by design, and the gate proved it.**

The `Landed modules compile against this PR` job is red with exactly the failure the cross-repo audit predicted:

```
error MSB9008: The referenced project src/MeshWeaver.AI.Application/MeshWeaver.AI.Application.csproj
does not exist. [plugins-repo/src/MeshWeaver.Blazor.Portal/MeshWeaver.Blazor.Portal.csproj]
```

The plugins repo keeps its own copy of `MeshWeaver.Blazor.Portal`, and its csproj carried a **ships-the-bits** `ProjectReference` to the project this PR deletes — the same reference core's copy had, and the same reason the application was dead code: it shipped in the closure while listed under no `Modules:Assemblies`, so nothing ever read its assembly attribute. Nothing there *uses* the project; the only mention is a comment.

Plugins#685 drops it, verified building that project against **both** core `main` (0 errors) and this branch with the project already deleted (0 errors) — so it is safe to land first and does not wait on this PR. **Merge order: Plugins#685 → this.**

The other red, `AiContentPackDriftTest`, was main's own — fixed by #2288, which merged after this branch was cut. `origin/main` is now merged in.

Issue #354, opened 2026-07-07
Global AI catalog partitions are absent from the search/query index — Providers/Models/Agents/Skills catalogs render empty although the nodes exist


sierragolflima:
## Summary
On affected deployed instances, the global AI catalog partitions (`Provider`, `Model`, `Agent`, `Skill`, `Harness`) are **materialized but missing from the search/query index**. Every index-based access (search, children-listing, the catalog layout areas) returns empty, while authoritative exact-path reads succeed. The AI-menu catalog areas are index-driven, so they render empty.

## How to reproduce
Precondition: a deployed instance that has at least one global catalog node (e.g. a `Provider/<name>` node exists — confirm with step 1).

1. **Exact-path read (authoritative):** `get @Provider/Anthropic` → returns the `ModelProvider` node. ✅ node exists.
2. **Children listing (index-based):** `get @Provider/*` → `[]`  ❌ (should list `Provider/Anthropic`).
3. **Search (index-based):**
   - `search nodeType:ModelProvider` → `0`  ❌
   - `search namespace:Provider scope:descendants nodeType:ModelProvider` → `0`  ❌
4. **UI:** open the top-bar AI menu → **Providers** (and Models / Agents / Skills) → the catalog renders **empty**.

**Expected:** steps 2–4 return / show the existing node(s).
**Actual:** steps 2–4 return empty, while step 1 proves the node exists.

Same pattern for the other roots: `get @Agent/*`, `@Skill/*`, `@Harness/*` → `[]`; `search nodeType:Agent`/`Skill` return only space-partition nodes, never the global catalog.

## Impact
- All four AI-menu catalogs (Providers / Models / Agents / Skills) render empty.
- The BYO-API-key UI (AI menu → Providers → create/edit a personal provider) is unusable.
- The harness picker (`namespace:Harness nodeType:Harness`) is empty ⇒ `/harness` can't switch to a CLI harness ⇒ the `/login` Connect flow can't be reached there (see #324).
- The "empty picker ⇒ no nodes exist" diagnostic in `ModelProviderSetup.md` (line 231) is misleading — the nodes exist, they're just unindexed.

## Why it's an index-registration gap, not a catalog-UI bug
[...]

PR #2315, opened 2026-08-26
The AI engine leaves the composition root — Memex.Portal.Shared no longer references it (#2276)

rbuergi:
`Memex.Portal.Shared` compiled against `MeshWeaver.AI` **in order to configure it**. That is the reference a Store-delivered module cannot coexist with: a closure DLL and a registry module of the same name are mutually exclusive, and the landing service refuses with a **409 on every instance**.

### It could not be split

If the module registered the AI content sources from configuration while the portal still called `AddAI(...)` compiled-in, `AddBuiltInAiContentSources()` would run **twice** — duplicate `IStaticRepoSource` registrations, and the importer imports every AI source twice. So the module attribute, the removal of the compiled call, and the ~13 types of wiring land together.

### `AiMeshModuleAttribute` is now the whole composition

It reads `MeshBuilder.Configuration` (#2300, added for exactly this) and does what `MemexConfiguration` did: `AddAI` with the serve-from-DB set, the AI menu seeds, `ModelProviderService`, the Connect token sink + session manager, the gated Claude strategy and its options, the skill→file sync, the AI content sources, and the provider-credential seed.

Reading configuration at **builder** time is not a preference: the serve-from-DB decision sets `MeshNode.IsDefinitionOnly`, an `init` property, and getting it wrong makes a partition root permanently unrecoverable (#902).

### Three seams the platform keeps, each because the HOST implements it

| seam | why it cannot ride the module |
|---|---|
| `IMcpBackConnection` + `McpConnectionInfo` → `MeshWeaver.Mesh.Services` | `McpBackConnectionService` needs `ApiTokenService` + ASP.NET; only the contract is platform (the `ISpeechTranscriber` shape) |
| `IPortalHubAccessor` → `MeshWeaver.Mesh.Services` | the skill sync needs the **portal's** hub, and an ambient `IMessageHub` is the router. Nullable by design — a headless host has no portal, and the consumer degrades |
| `IStaticRepoSource.AssertIndexedAfterImport` | the portal handed `ImportAll` the AI partition **names** for the #354 index assertion. A module cannot be delisted from a list the host hard-codes, so a source declares its own requirement — same shape as `SyncMode` — and `ImportAll` derives the set |

Everything else uses seams that already existed: the AI catalogs and the token-usage tab register through the per-node-hub chain and `AddSettingsMenuItems`; the portal's AI types and thread area are contributed through `PortalConfigurationRegistry`, which exists precisely so a plugin can configure portals it does not compile against.

### Tests follow their code

Five files move into the engine with their suites. `ProviderCredentialSeedWiringTest` now installs the assembly attribute the way `Modules:Assemblies` does — **who** registers the seed changed, **whether** it is registered did not, and that is what it still asserts. It failed first, which is how I knew the move was real.

### Verification

- Full solution build under CI's flags → `0 Error(s)`.
- `MeshWeaver.AI.Test` **1306**, `MeshWeaver.Threading.Test` **139**, `Memex.Portal.Shared.Test` **428** — all green.

### What this does and does not finish

`Memex.Portal.Shared` is off the list. The engine still reaches the app closure through `src/MeshWeaver.Blazor.Portal` (a view pack, deleted from core by #2293) and `tools/MeshWeaver.PluginTester` (a tool, not the platform). **The 409 clears when #2293 lands**, and then Plugins#687 can flip to `preInstalled`.

Refs #2276

### DIFF
--- a/memex/Memex.Portal.Shared/Authentication/McpBackConnectionService.cs
+++ b/memex/Memex.Portal.Shared/Authentication/McpBackConnectionService.cs
@@ -1,6 +1,6 @@
 using System.Collections.Concurrent;
 using System.Reactive.Linq;
-using MeshWeaver.AI.Connect;
+using MeshWeaver.Mesh.Services;
 using MeshWeaver.Hosting.AspNetCore;
 using Microsoft.Extensions.Logging;
 using Microsoft.Extensions.Options;
--- a/memex/Memex.Portal.Shared/Memex.Portal.Shared.csproj
+++ b/memex/Memex.Portal.Shared/Memex.Portal.Shared.csproj
@@ -101,9 +101,13 @@
          the modules/<Name>/ closure layout in MeshModulesPublish.targets and activates via
          Modules:Assemblies -> MarkdownExportProviderAttribute. Only the wire contract stays in
          the app closure (via MeshWeaver.Blazor -> MeshWeaver.Markdown.Export.Contract). -->
-    <ProjectReference Include="..\..\src\MeshWeaver.AI\MeshWeaver.AI.csproj" />
     <ProjectReference Include="..\..\src\MeshWeaver.Mesh.Operations\MeshWeaver.Mesh.Operations.csproj" />
-    <!-- The AI provider packs — MeshWeaver.AI.OpenAI, .AzureFoundry, .ClaudeCode, .Copilot — are
+    <!-- 🚨 MeshWeaver.AI is deliberately NOT referenced (#2276): the engine AND its maintenance
+         application ship as ONE Store entry and activate via Modules:Assemblies ->
+         AiMeshModuleAttribute, which reads this deployment's own configuration to decide what to
+         register. A reference here would put the DLL back in the app closure, where the module
+         landing refuses the same-identity module with a 409 on every instance.
+         The AI provider packs — MeshWeaver.AI.OpenAI, .AzureFoundry, .ClaudeCode, .Copilot — are
          deliberately NOT referenced (#1644 step 2): they ship via the modules/<Name>/ closure
          layout in MeshModulesPublish.targets and activate via Modules:Assemblies -> each pack's
          MeshNodeProviderAttribute (a deployment drops a provider by editing its module list).
--- a/memex/Memex.Portal.Shared/MemexConfiguration.cs
+++ b/memex/Memex.Portal.Shared/MemexConfiguration.cs
@@ -6,7 +6,6 @@ using Memex.Portal.Shared.SelfUpdate;
 using Memex.Portal.Shared.Settings;
 using Memex.Portal.Shared.Social;
 using Microsoft.Extensions.DependencyInjection.Extensions;
-using MeshWeaver.AI;
 using MeshWeaver.Blazor.Infrastructure;
 using MeshWeaver.Hosting.Grpc;
 using MeshWeaver.Blazor.Pages;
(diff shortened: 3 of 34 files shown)


rbuergi:
Four reds, triaged — **two were mine and are fixed, two are not.**

### 1. `Landed modules compile` — mine, and a promise I broke

```
CS0246: 'IMcpBackConnection' not found [plugins-repo/src/MeshWeaver.AI.Copilot]
```

I audited that interface's consumers **in core** and stopped there, having said two hours earlier that a cross-repo move needs the cross-repo sweep every time. Four files in the plugins repo consume it — the ClaudeCode and Copilot chat clients and harnesses.

Fixed in Systemorph/MeshWeaver.Plugins#696, order-independently: `using MeshWeaver.Mesh.Services;` alongside the existing `using MeshWeaver.AI.Connect;` resolves the simple name from either namespace. Verified by building both packs against core `main` **and** against this branch — `0 Error(s)` in all four builds. **Merge order: Plugins#696 → this.**

### 2. `ImpersonationScopeSiteRatchetGuard` — mine, and the guard was right

`ModelProviderService` moved into `src/MeshWeaver.AI/Portal/`, so its **pre-existing** `Observable.Using(() => …SwitchAccessContext(ctx), …)` stopped being accepted debt at its old path and read as a NEW site.

The failure message says explicitly: *do NOT add a line to the allow file*. That shape opens the AsyncLocal scope on the subscribing thread and disposes it on whichever thread the work terminates, leaving the subscriber latched as the impersonated identity (#1790).

So it is fixed rather than allow-listed — `accessService.RunAs(ctx, () => meshService.CreateNode(modelNode))`, which is also **smaller** than what it replaces: `RunAs` already defers when access or identity is null, which is what the ternary's other branch did by hand. The move paid the debt instead of relocating it past a path-keyed guard.

### 3–4. Not mine

- `LiveQueryHandoffDropTest` (`Hosting.Test`, shard 2) — passes locally on this branch.
- `OrleansNodeChangePropagationTest.Delegation_NodeChanges_PropagateFromSubThread` (shard 0) — passes locally on this branch; shard 0 Orleans is the population #1384 tracks, and this is the fourth *different* test from it in this series.

`origin/main` merged; full solution build under CI's flags is green.

rbuergi:
Conflicts resolved — main moved underneath this branch when **#2285** landed, relocating `PortalApplication` from `MeshWeaver.Blazor.Infrastructure` to `MeshWeaver.Hosting.AspNetCore.Portal`.

Two conflicts, both in files this PR touches from the opposite end:

- **`AgentSkillSyncService`** — main added `using MeshWeaver.Hosting.AspNetCore.Portal;` for the relocated type; this branch removed the `PortalApplication` dependency altogether in favour of `IPortalHubAccessor`. Kept the seam; the relocated using is dead here. (The seam is still the right call even though `PortalApplication` is now in a platform assembly — reaching it directly would pull ASP.NET into the engine.)
- **`MemexConfiguration`** — the resolution is main's relocation **minus** `using MeshWeaver.AI;`. Taking either side whole would have been wrong: mine would have reverted #2285's relocation, theirs would have restored the AI import this PR exists to remove.

Verified rather than assumed, per the rule that a branch-favoured hunk silently undoing merged work is the real failure mode:

- `git diff origin/main...HEAD --diff-filter=D` (three dots) → **zero** deletions introduced by this branch.
- #2285's relocation survives in `MemexConfiguration` (3 references), with no stale `MeshWeaver.Blazor.Infrastructure` import left behind.
- Full solution build under CI's flags → `0 Error(s)`.

Issue #2271, opened 2026-08-25
Second install rewrites default-valued nodes: default-suppression makes the stored node differ from the authored file, so the unchanged-skip misses


rbuergi:
## What happens

A **second** install of the same package rewrites nodes whose content is default-valued — observed on `PensionFund/Currency/{CHF,EUR,USD}` (3 nodes rewritten, expected 0). It is the deploy-flicker the idempotence pin exists to catch.

## Mechanism

`Currency.DecimalPlaces` carries a C# initializer `= 2` (`PensionFund/Currency/Source/Currency.cs`), and all three instance files author `"decimalPlaces": 2`.

1. Once the `Currency` **NodeType compiles**, its content type is registered.
2. The installer therefore materialises the instance as a **typed `Currency`** rather than an untyped `JsonElement`, and re-serialises it.
3. **Default-suppression drops the property whose value equals the type's default.**
4. The stored node now differs from the authored file — so the **unchanged-skip misses**, and the next install rewrites it.

Year/2024 and every `Position/*` round-trip clean. The three that churn are exactly the three whose content is default-valued.

## Why it is intermittent — and why that matters

It is **invisible until the NodeType compiles**. That is why the gate reported it inconsistently within one hour tonight:

- one run: `STALE allow entry (now passing — remove it): PensionFund idempotence`
- the very next: `GATE FAILED — idempotence: PensionFund — 1 new failure(s)`

Both reports are correct. I removed the allow entry on the first and had to restore it on the second — recorded at the entry so the next person does not repeat it. **A single green run is not evidence this is gone.**

## Why it is surfacing now

#2063's discovery fix means the gate now finds PensionFund's **5 NodeTypes where it previously found 0** (the BOM hid them). So the compile happens, step 1 above becomes reachable, and a defect that was always present becomes intermittently visible.

## Suggested direction

The write path should compare against **what was authored**, not against a re-serialisation of the materialised value — or default-suppression must not apply on the install round-trip. Related and possibly the same root: **#2266**, where the installer logs *"the file's authored content could not be re-read — installing the materialised value, which may carry another package's defaults."* Both are the installer preferring a materialised value over the authored one.

Currently carried as known debt in `.github/samples-gate.allow` (`PensionFund idempotence`).

rbuergi:
Fixed by #2318, merged as `187f91048` — **as a duplicate root cause of #2266, and this issue's stated mechanism was wrong.**

[...]

Issue #2266, opened 2026-08-25
PackageInstaller installs a MATERIALISED value when it cannot re-read the authored file — 'may carry another package's defaults'


rbuergi:
## What the installer says about itself

Seen 14× in one gate run (`32885213574`, `Doc content compiles and runs`), across `PensionFund/Position/*` and `PensionFund/Year/*`:

```
warn: MeshWeaver.PluginCatalog.PackageInstaller[0]
  PensionFund/Position/Payables.json materialised the runtime-compiled content type Position,
  and the file's authored content could not be re-read — installing the materialised value,
  which may carry ANOTHER PACKAGE'S DEFAULTS.
```

The installer is telling us, in its own words, that it is storing **something other than what the author wrote**, and that the substituted value may carry defaults belonging to a *different package*. It then proceeds, at `warn`.

## Why this is worth its own issue

This is the fail-closed-but-silent shape that has produced most of tonight's defects: the install **succeeds**, the node exists, nothing is red — and the stored content is not the authored content. Nobody files a bug against that, because from outside it looks like it worked.

The specific risk in the message is cross-package contamination: *"may carry another package's defaults"*. A default belonging to package B being written into package A's node is a correctness problem that would surface much later, somewhere unrelated, as a wrong value with no provenance.

## How it surfaced

Found while clearing the now-stale `PensionFund idempotence` entry from `.github/samples-gate.allow` (PR #2250). That entry became stale because #2063's discovery fix made the gate see PensionFund's **5 NodeTypes where it previously saw 0** — so this install path is only now being exercised under the gate at all. **The warning is very likely pre-existing and was simply never reached**; it is not caused by #2250.

## What to establish

1. **Why can the authored content not be re-read?** That is the actual defect — the fallback is a consequence. Is the file handle gone, is the parse failing after materialisation, is it a BOM/encoding path?
2. **Can it be right to install a materialised value at all here?** If the authored content is unavailable, refusing (and saying which file and why) is likely better than storing a value the installer itself flags as possibly wrong.
3. **Is the "another package's defaults" case reachable in practice**, or defensive wording? If reachable, it should be an error, not a warning.

Related: #2211 (an import recording a skipped Provider as `Succeeded`), #2254 (default install silently skipping a package) — same family, different lane.

rbuergi:
Fixed by #2318, merged as `187f91048`.

[...]

Issue #2264, opened 2026-08-25
SettleAsError never stamps FailedBuildInputs, so every adopt-only refusal burns one needless automatic re-drive


rbuergi:
## What

`NodeTypeCompilationHelpers`' `SettleAsError` — the compile watcher's re-settle of a Pending flip it
refuses to dispatch — writes `CompilationStatus = Error` and `CompilationError`, but never stamps
`FailedBuildInputs`.

`ApplyCompileFailure` (a real compile's failure write-back) *does* stamp it. `SettleAsError` is the
other way a NodeType reaches a terminal `Error`, and it is the ONLY way a type reaches `Error` under
the adopt-only gate (`Modules:RequirePrebuilt`, #2193 §A) — which never runs a compile at all.

## Why it matters

`HasStaleFailureVerdict` reads `FailedBuildInputs` to answer "has this type already had its attempt
under these compile inputs?". An unstamped verdict differs from every live token, so it always reads
**"never attempted"**. Consequence: the adopt-only gate's refusal is followed, on every refused type,
by one automatic failed-verdict re-drive (#1793) — a `Pending` flip, a second trip through the gate,
a second refusal — that was never needed. It burns one of the five automatic re-drives, and it puts
a needless round on the per-NodeType hub's single-threaded action block.

With the maintainer's direction on #2193 to set `Modules:RequirePrebuilt` for *"all which we mark as
such — should be almost all"*, that is one spurious round per refused type.

It converges (the re-drive's own write stamps the token, and the stamp is persisted on the node), so
it is bounded and once-per-node, not a storm.

## Not the same as #2260

#2260 was the un-park inside that re-drive; that is fixed (#2262) and this is deliberately **not**
folded into it. Two reasons to keep them apart:

- Removing this trigger would not have closed #2260's window. The first refusal settles *before*
  the sources watcher has established `CurrentSourceVersions`, so a stamp written there records
  `src=(unseeded)`; when the watcher publishes `{}` the live token moves anyway and the re-drive
  fires regardless. The window had to be closed at the un-park, which is what #2262 did.
- The two `SettleAsError` call sites need *different* answers, so this is a real design question and
  not a one-liner:
  - the **adopt-only gate** (`SettleAsError(reason)`) — the refusal genuinely was formed under the
    live inputs, so stamping is honest;
  - the **parked short-circuit** (`SettleAsError(parkRegistry.GetParkedError(hubPath))`) — the
[...]

Issue #2193, opened 2026-08-25
Design of record: modules ship as baked DLLs — no compile fallback, no source in the mesh DB


rbuergi:
## Mandate

> Never get uncompiled/stale state to prod. Error early when a pre-built DLL is missing — never fall back to compiling. Remove all module source from the mesh DB. If Store updates, all dependent packages update as well. — Roland, 2026-08-25

This issue is the design of record for that program, motivated by today's Store outage on memex.

## The incident (2026-08-25, memex)

Four merges to MeshWeaver.Plugins main (#637–#640, Store + Edu + Northwind) were picked up by the catalog auto-update (`PluginUpdateWatcher` → `PackageUpdateReconciler` → `CatalogLayoutAreas.InstallOrUpdate`) at 08:53 and 09:01 UTC. Every one of the 17 Store NodeTypes recompiled **green** (`compilationStatus: Ok`). The store page still went down:

1. **Serving never converged.** Instances keep executing the previously bound assembly behind the *"a newer build of this type is available — Recycle"* banner (`StaleBuildBanner` — the deliberate offer-not-restart design). On prod nobody clicks: `/Store` served a mixture of old and new assemblies. Tell in the logs: `Received '$type':'StoreContent' which is NOT registered in this (receiving) hub` — two hubs on different compiles of the same sources disagree about the type registry.
2. **The Store/Catalog type hub wedged.** Five portal replicas raced its recompile: `MonotonicWriteGuard` conflicts on `Store/Catalog` and `Store/Catalog/_Activity/compile-state`, `[UpdateRemote] LATE_NACK_TERMINAL … TimeoutException` on compile-state writes. `@Store/area/Catalog` timed out until the type + instance hubs were manually recycled.
3. **Dependents were never updated.** The install path releases only the installed package's own types (`PackageInstaller` → `RequestReleases` over that package's `nodeTypePaths`). Packages that compile `shared=@Store/...` INTO their assemblies (Edu, courses, satellite-repo modules) were left on assemblies built against the old Store sources — the mixed-assembly `$type` mismatch above is exactly that.

**"no dlls found"** (as observed by Roland) is the consumer bundle lane resolving nothing and silently falling back to Roslyn:
[...]

PR #2318, opened 2026-08-26
Fix plugin installer writing values it should not (#2264, #2266, #2271)

rbuergi:
## Summary

Three related defects, all in the "installer writes something other than what it should" family — investigated and fixed against the actual code, not just the issue bodies' hypotheses.

### #2264 — `SettleAsError` never stamps `FailedBuildInputs`
**Confirmed as described.** `InstallCompileWatcher`'s `SettleAsError` local function is the *only* way a NodeType reaches `CompilationStatus.Error` under the adopt-only gate (`Modules:RequirePrebuilt`, #2193 §A), which never runs a compile — so it was the only writer of `FailedBuildInputs` for such a type, and it never wrote it at all. An unstamped verdict always differs from the live compile-inputs token, so `HasStaleFailureVerdict` read every refusal as "never attempted" and drove one needless automatic failed-verdict re-drive (#1793) per refused type.

Fixed per the issue's own guidance: the two call sites need *opposite* treatment, extracted into a pure `ApplyGateSettle(parkedDef, reason, formedUnderLiveInputs, modulesHash)`:
- the **adopt-only gate's own refusal** stamps the live inputs (it genuinely was formed under them)
- the **parked short-circuit's re-settle** leaves the existing stamp untouched (it's re-serving an *already-settled* verdict — stamping live inputs there would mask a genuine input change and defeat #1793's recovery path)

### #2266 — installer installs a materialised value when it can't re-read the authored file
**Root cause found and fixed.** `PackageInstaller.AsAuthored` re-parses `file.Content` to recover the authored JSON (so a runtime-compiled content type's possibly-wrong-package materialisation never gets written). That re-parse used `JsonDocument.Parse(file.Content)` with **no options** — stricter than the primary parse (`FileFormatParserRegistry.TryParse`), which strips a leading UTF-8 BOM and applies comment/trailing-comma leniency copied from the hub's `JsonSerializerOptions` before it ever reaches `AsAuthored`.

Verified with a hex dump: `samples/Graph/Data/PensionFund/Position/*.json`, `Year/*.json`, and `Currency/{CHF,EUR,USD}.json` all open with `EF BB BF` (UTF-8 BOM). `JsonDocument.Parse` does **not** auto-skip a BOM (confirmed empirically — see scratch repro in the session), so the primary parse succeeded (it strips the BOM first) while `AsAuthored`'s re-parse threw, triggering exactly the fallback the issue describes: the materialised (possibly wrong-package) value gets installed and logged at `warn`.

Fix: `AsAuthored` now shares the exact same tolerances the primary parse already applied (`FileFormatParserRegistry.WithoutBom` + the same `JsonDocumentOptions`), threaded through from `hub.JsonSerializerOptions` at all four call sites (`ParseAll`/`ParseCanonical` for node-repo installs, `ParseNode` for regular/code package installs).

### #2271 — second install rewrites default-valued nodes
**Traced to the same root cause as #2266, not a distinct default-suppression bug.** I verified empirically that `JsonSerializerOptions.DefaultIgnoreCondition = WhenWritingDefault` compares against `default(int)` (0), **not** a property's C# initializer value — so `Currency.DecimalPlaces = 2` is never dropped by that mechanism for a value of `2` (a small repro is in the session transcript). The issue's literal mechanism description doesn't hold.

What *is* verified: `PensionFund/Currency/{CHF,EUR,USD}.json` all carry the same BOM as the #2266 files. Before the fix, the BOM made `AsAuthored` install the **materialised** value on the first install instead of the authored one — and it's *that* wrongly-installed value the second install's unchanged-skip compared the fresh (correctly re-parsed) file against, producing the observed churn. The #2266 fix removes the wrong write at its source, which removes this specific manifestation.

**Left untouched, deliberately:** `.github/samples-gate.allow`'s `PensionFund idempotence` entry — its own comment says "do not remove on a single green run," and this PR's fix is a one-off local verification, not the samples-gate's own repeated CI signal. Whoever next observes the gate green across multiple runs should remove it.

## Changes

- `src/MeshWeaver.Graph/Configuration/NodeTypeCompilationHelpers.cs` — extracted `ApplyGateSettle`, updated `SettleAsError`'s two call sites.
- `src/MeshWeaver.PluginCatalog/PackageInstaller.cs` — `AsAuthored` shares the primary parse's BOM/leniency tolerances; threaded `JsonSerializerOptions` through `ParseAll`/`ParseCanonical`/`ParseNode`.
- Tests: `test/MeshWeaver.Graph.Test/GateSettleFailedBuildInputsTest.cs` (new, pure-function pins for #2264), `test/MeshWeaver.Hosting.Monolith.Test/RequirePrebuiltParkTest.cs` (extended, asserts the stamp lands end-to-end), `test/MeshWeaver.PluginCatalog.Test/InstallAuthoredContentTest.cs` (extended, BOM regression case for #2266).
- Two What's New entries (`Category: Fix`) — both fixes are user/operator-noticeable (installer content correctness; fewer spurious compile re-drives).

## Test plan

- [x] `dotnet build src/MeshWeaver.Graph/MeshWeaver.Graph.csproj -c Release -warnaserror` — 0 warnings, 0 errors
- [x] `dotnet build src/MeshWeaver.PluginCatalog/MeshWeaver.PluginCatalog.csproj -c Release -warnaserror` — 0 warnings, 0 errors
- [x] `GateSettleFailedBuildInputsTest` + `FailedVerdictRedriveTest` (MeshWeaver.Graph.Test): 24/24 passed
- [x] `InstallAuthoredContentTest` (MeshWeaver.PluginCatalog.Test): 5/5 passed (incl. new BOM case)
- [x] `RequirePrebuiltParkTest` (MeshWeaver.Hosting.Monolith.Test, end-to-end integration): 1/1 passed, `FailedBuildInputs` confirmed stamped
- [x] `NeverCompiledFailureRedriveTest` + `NodeTypeCompileParkTest` + `RedriveObservation*` (regression check): 18/18 passed
- [x] `WhatsNewEntryIntegrityTest` + `DocumentationLinkIntegrityTest`: 2/2 passed
- [x] Full `MeshWeaver.PluginCatalog.Test` suite (regression check for the `ParseAll`/`ParseNode`/`AsAuthored` signature changes touching every install path): 647/647 passed

### DIFF
--- a/src/MeshWeaver.Documentation/Data/WhatsNew/2026-08-26-a-refused-build-stops-retrying-itself.md
+++ b/src/MeshWeaver.Documentation/Data/WhatsNew/2026-08-26-a-refused-build-stops-retrying-itself.md
@@ -1,0 +1,33 @@
+---
+Name: A refused build stops retrying itself
+Category: Fix
+Description: On a mesh that requires prebuilt assemblies, a type refused for lack of a bundle used to burn one automatic retry attempt it never needed — a second trip through the same refusal, for nothing. The refusal now records what it was decided under, so the automatic retry recognises it was already answered.
+Icon: Sparkle
+Order: -20260826
+---
+
+# A refused build stops retrying itself
+
+`Modules:RequirePrebuilt` parks a type instead of compiling it when no prebuilt assembly has landed
+for it — a named, bounded refusal instead of a silent Roslyn pass. The framework also carries a
+separate, general recovery: a type sitting at a build failure with nothing recorded about what it was
+tried under gets exactly one automatic retry, so a fix that ships later can still reach it without
+anyone pressing a button.
+
+The two mechanisms shared a gap. A require-prebuilt refusal never recorded what it was decided under,
+so the recovery mechanism read every refusal as "never attempted" — and drove its one automatic retry
+on a type that had, in fact, just been correctly and deliberately refused. The retry landed back on
+the same refusal a moment later, spending part of a budget meant for genuine recoveries on a refusal
+that needed none.
+
+## What changed
+
+A require-prebuilt refusal now records the inputs — the framework build, the installed modules, the
+source set — it was decided under, in the same write that settles it. The recovery mechanism compares
+against that record and correctly finds nothing to retry, because nothing about the refusal has
+changed since it was made.
+
+## What you will notice
+
+A type refused for lack of a bundle settles once and stays settled — no extra round through the
+compile pipeline, and the automatic-retry budget is left for the failures that actually need it.
--- a/src/MeshWeaver.Documentation/Data/WhatsNew/2026-08-26-package-installs-keep-authored-content.md
+++ b/src/MeshWeaver.Documentation/Data/WhatsNew/2026-08-26-package-installs-keep-authored-content.md
@@ -1,0 +1,44 @@
+---
+Name: Package installs keep the content you actually authored
+Category: Fix
+Description: A package file with a leading byte-order mark and a dynamically-compiled content type could install a materialised value instead of the file's own content, and rewrite that node on every later install. Both are fixed at the root — the installer's authored-content re-read now tolerates exactly what the primary parse already tolerates.
+Icon: FileCheckmark
+Order: -20260826
+---
+
+# Package installs keep the content you actually authored
+
+Installing a content package (a node-repo plugin, a course, a NodeType's instances) writes what the
+file says. For most files that was always true. For one narrow but real case it was not: an instance
+file whose content type is a NodeType that has already compiled.
+
+## What was happening
+
+Once a NodeType compiles, its content type becomes a name the mesh can resolve — but only by name.
+The same short name can exist in more than one package, so the installer's own parse of such a file
+does not trust the typed value it gets back: it re-reads the file's `content` property as written and
+installs that instead, specifically so one package's compiled defaults can never leak into another
+package's node.
+
+That re-read used stricter rules than the parse that produced the typed value in the first place. A
+file that opened with a UTF-8 byte-order mark — invisible in an editor, and common enough that one
+sample package ships several instance files this way — made the re-read fail. When it failed, the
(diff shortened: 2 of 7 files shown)


rbuergi:
## CI status: blocked on #2324, not a regression from this PR

The \`Doc content compiles and runs\` / \`Consolidate test results\` required check is red, but it is **not** a defect in this PR's changes. Confirmed with two independent runs (the original run + one rerun of the failed jobs):

\`\`\`
STALE allow entry (now passing — remove it): PensionFund idempotence
GATE FAILED — stale allow entr(ies): PensionFund idempotence — 0 new failure(s), 1 stale allow entr(ies).
\`\`\`

**\`0 new failure(s)\`** both times — every genuine check (build, all 6 test shards including shard 2, plugins build/bake, landed-modules compile) passed both runs. The only failure is the samples-gate's stale-allow-entry ratchet, which is exactly the known, already-being-fixed bug in **#2324** (*"The ratchet fails a run for obeying the allow file — teach it 'intermittent'"*): the ratchet's comment-stripping parser can't see the \`.github/samples-gate.allow\` line 32 comment that says "INTERMITTENT known debt. Do not remove on a single green run," so any run where the flapping \`PensionFund idempotence\` check happens to pass gets reported as a false "stale entry" failure.

This PR's own #2266 fix (installer BOM handling) is very likely *why* PensionFund idempotence is passing on this branch right now — but per the allow file's own documented caution (and this repo's working agreement against declaring intermittent flakes fixed off one or two green runs), **this PR does not touch `.github/samples-gate.allow`**. Removing the entry here would race/duplicate #2324, which is the correct, narrower fix (an explicit `intermittent` marker the parser can honour, rather than deleting the debt-ratchet's safety net).

**This PR is blocked on #2324 landing first.** Once #2324 merges, a rerun of this PR's CI should go green on the doc-gate step without any further change here.

(Also checked and ruled out: #2319 / `LiveQueryHandoffDropTest` on shard 2 — that shard passed cleanly on both runs, so it isn't in play for this PR.)

PR #2324, opened 2026-08-26
The ratchet fails a run for obeying the allow file — teach it "intermittent"

rbuergi:
`.github/samples-gate.allow` line 32 says, in a comment:

```
# PensionFund idempotence — INTERMITTENT known debt. Do not remove on a single green run.
```

**The parser strips comments, so the ratchet cannot read that.**

## The bug

The ratchet's rule — *a listed check that starts passing fails the run until its line is removed* — assumes the check is **deterministic**, so one green run proves the debt is paid. For a flapping check that inference is wrong, and the git history shows the loop it produces:

| commit | action |
|---|---|
| `768f9ba0b` | Drop the now-stale PensionFund idempotence allow entry |
| `f32047e0e` | **Restore** the PensionFund idempotence entry — the check is intermittent |

Since then, every run where PensionFund happens to pass fails the gate with `stale allow entr(ies)` and **0 new failures**. That is what is currently reddening **#2285**, a PR that touches nothing related.

## Why not just remove the entry

Both obvious fixes are wrong:
- **Remove it** — the file explicitly forbids it, and it was already tried and reverted one commit later.
- **Wait for the flake to fail again** — that is waiting on a coin flip, and the next PR hits the same wall.

So the intent moves out of a comment the parser discards and into a token it reads:

```
PensionFund idempotence intermittent
```

## What the marker does, and what it deliberately does not

An intermittent entry is exempt from **stale-detection** and nothing else. It still tolerates only the scope/check it names; a failure it does not name is still a new failure. What it gives up is stale-detection for that one line — which is the entire point, not a side effect.

This is a real weakening of one specific guarantee, so it is deliberately narrow: it cannot be applied by accident (a typo throws), it cannot broaden what an entry suppresses, and it leaves the `Unverifiable` bucket's meaning intact.

## Verified in both directions

A marker that cannot fail would be worse than the bug it fixes:

- an intermittent entry whose check **passes** is not stale — gate stays green
- an intermittent entry **still suppresses its own failure** — unchanged behaviour
- a **typo** in the marker (`intermitent`) throws `FormatException` rather than silently degrading to a plain entry, which would restore the stale-failure with nothing pointing at the cause

`22/22` in `MeshWeaver.PluginTester.Test`, `-c Release -warnaserror` clean.

## Scope

Found while diagnosing #2285's red. Its failure was `Doc content compiles and runs` — the same job name as the Northwind compile error fixed in #2308, but a **different check inside it**, which is why the fix that landed did not clear it.

### DIFF
--- a/.github/samples-gate.allow
+++ b/.github/samples-gate.allow
@@ -41,3 +41,17 @@ PensionFund/BalanceSheet compile
 # default-suppression drops the property whose value equals the type's default. The stored
 # node then differs from the authored file and the unchanged-skip misses.
 PensionFund idempotence intermittent
+
+# 🚨 The `intermittent` token above is a REAL WEAKENING, and its failure mode is social rather than
+# technical: the next person with a deterministic red they cannot fix will find this token and reach
+# for it. It is legitimate for exactly one thing — a check that FLAPS, where a single green run does
+# not prove the debt is paid. It is NOT for a failure that is merely inconvenient, hard, or someone
+# else's. Marking a deterministic failure intermittent silences the ratchet permanently and is the
+# band-aid the ratchet exists to prevent.
+#
+# It buys exemption from STALE-DETECTION and nothing else: the entry still tolerates only the
+# scope/check it names, and a failure it does not name is still a new failure.
+#
+# Before adding one, prove the flap — two runs of the same commit disagreeing, or the git history of
+# this file showing an entry dropped as stale and restored. That is what justified this one:
+# 768f9ba0b dropped it, f32047e0e restored it one commit later.
--- a/.github/samples-gate.allow
+++ b/.github/samples-gate.allow
@@ -40,4 +40,4 @@ PensionFund/BalanceSheet compile
 # type is registered, the installer materialises a typed `Currency` and re-serialises it, and
 # default-suppression drops the property whose value equals the type's default. The stored
 # node then differs from the authored file and the unchanged-skip misses.
-PensionFund idempotence
+PensionFund idempotence intermittent
--- a/test/MeshWeaver.PluginTester.Test/GateAllowlistTest.cs
+++ b/test/MeshWeaver.PluginTester.Test/GateAllowlistTest.cs
@@ -192,6 +192,19 @@ public class GateAllowlistTest
     }
 
     [Fact]
+    public void ToString_RendersTheEntryExactlyAsTheAllowFileHasIt()
+    {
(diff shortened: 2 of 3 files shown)


rbuergi:
Strong diagnosis of the ratchet bug — the comment-stripping parser making the file unreadable to its own enforcer is exactly right, and the drop/restore commit pair you cite was mine.

**But one input to the decision is wrong, and I wrote it.** The `PensionFund idempotence` comment in `.github/samples-gate.allow` explains the intermittency like this:

> `Currency.DecimalPlaces` carries a C# initializer `= 2` … default-suppression drops the property whose value equals the type's default.

**That mechanism is not real.** `DefaultIgnoreCondition = WhenWritingDefault` suppresses at `default(T)` — for `int` that is **0**, never a property's `= 2` initializer. `Serialization.md` states it directly, and `MeshNodeVersioning.md` uses the same fact as its worked example (`Version = 0` would be omitted, which is why versions start at 1). So `"decimalPlaces": 2` is never dropped, and my explanation for why the check flaps was wrong.

**The real cause was found while fixing #2266/#2271 in #2318:** `PackageInstaller.AsAuthored` re-parsed `file.Content` with **no `JsonDocumentOptions`** — stricter than the primary parse, which strips a leading UTF-8 BOM. `PensionFund/Currency/{CHF,EUR,USD}.json` are BOM'd (verified by hex dump: `EF BB BF`), so pre-fix the installer wrote the *materialised* value on first install, and the second install's unchanged-skip compared against that wrongly-written value. #2318 shares the primary parse's tolerances and removes the wrong write at its source.

**Why this matters for your PR:** if #2318 lands, `PensionFund idempotence` may stop flapping and start passing **deterministically** — at which point it is a genuinely stale entry that should be *removed*, not marked intermittent. Marking it intermittent would then permanently exempt a check that is actually fixed, which is the failure mode your ratchet exists to prevent, one level up.

Two things worth doing either way, and they are independent of the order the PRs land in:

1. **Your ratchet change is right regardless** — a flapping check must be expressible, and a rule readable only by humans is not a rule. Please keep it.
2. **Fix or drop that comment's mechanism** while you are in the file, so the next reader does not inherit my wrong explanation. Happy to do it in #2318 instead if you would rather not touch it — say which and I will.

Evidence for the BOM cause is in #2318 (`InstallAuthoredContentTest` gained a BOM case; full `MeshWeaver.PluginCatalog.Test` 647/647).

PR #2333, opened 2026-08-26
Restore the 8-parameter constructor of LanguageModelCatalogSource

rbuergi:
## What this fixes

`systemorph` could not roll for ~2 hours. It did not look like a code problem: the old pods kept serving `aade75d00`, every health probe returned 200, and helm just sat there. The new ReplicaSet was in `CrashLoopBackOff` the whole time — 24 restarts — on:

```
Unhandled exception. System.TypeInitializationException:
  'MeshWeaver.AI.Anthropic.AnthropicExtensions' threw
---> System.MissingMethodException: Method not found:
  'Void MeshWeaver.AI.LanguageModelCatalogSource..ctor(String, String, Int32, String,
   String, ImmutableArray`1<String>, Boolean, MeshWeaver.AI.ProviderKind)'
```

## Cause

`987da5f9e` added `Description` as a **ninth primary-constructor parameter**. It has a default, so it is source-compatible and every gate stayed green — but a record's primary constructor signature is **replaced, not extended**, and a call site that omits the optional arguments still emits a call to the *full* constructor as it stood at **its own** compile time.

The AI provider modules are compiled separately, live in another repository, and land at runtime under `/data/modules`. Per `ModuleLandingService` they *"bind by simple name, and their contract is API compatibility, not build identity"* — build identity is deliberately diagnostic-only there. So all six landed providers went on calling the 8-argument constructor, which no longer existed.

Landed **Aug 25 01:17–01:19**, all before the break:

| module | landed |
|---|---|
| `MeshWeaver.AI.Anthropic@c2916b6c` | Aug 25 01:17 |
| `MeshWeaver.AI.AzureFoundry@8f60c8c4` | Aug 25 01:17 |
| `MeshWeaver.AI.OpenAI@661a3510` | Aug 25 01:19 |
| `MeshWeaver.AI.WebSearch@8412db3f` | Aug 25 01:19 |

Modules whose *own* source changed were rebuilt and re-landed at Aug 25 20:3x. These four were untouched — only the framework under them moved — so nothing rebuilt them and, by design, nothing rejected them.

## The fix is additive — and the CI gate is why

My first cut moved `Description` to an `init` property. That restores the 8-parameter form by **removing** the 9-parameter one: the same mistake pointing the other way, repairing every module compiled before the change and breaking every one compiled after it.

`scripts/check-record-signatures.py` refused it, exactly as it should:

```
✗ src/MeshWeaver.AI/BuiltInLanguageModelProvider.cs
    record LanguageModelCatalogSource — arity
```

That gate landed in `f60dc19df` at 2026-08-26 00:21 — **eleven hours after the break it would have caught.** This PR is one of its first real customers, and it caught me committing the inverse of the bug I was fixing.

So: **the ninth parameter stays, and the eight-parameter form is restored beside it.** Both populations bind. No call site changes; `Description` keeps working positionally and through `with { … }`.

The overload has **no source-level caller by design** — its callers are separately-compiled assemblies that bind by name at runtime. It is therefore pinned by a test rather than by use, and carries a comment saying so, because the obvious next cleanup is to delete it as dead code.

## Verification

- `MeshWeaver.AI` builds clean, **0 warnings**
- `MeshWeaver.AI.Test` — **1305 passed, 0 failed**
- `check-record-signatures.py --base origin/main` — **clean** (and `--self-test` passes)
- All six provider modules in `MeshWeaver.Plugins` compile against this core
- `LanguageModelCatalogSourceBinaryContractTest` **verified by mutation in both directions**: deleting the 8-param overload turns it red (2 failures); adding a 10th parameter turns it red

## The guard

No compile-time gate inside this solution can catch this class of break, because nothing here compiles against the binaries that make the call. The gate above catches it at the diff; this test holds the same line from inside the assembly, where it still applies to a consumer that never sees the diff. It pins both arities, asserts the two forms produce equal values, and carries a negative control so it cannot pass vacuously.

Its doc comment says explicitly: **if it goes red, do not update it to match the new signature.** That one "fix" converts a caught break into a shipped one.

### DIFF
--- a/src/MeshWeaver.AI/BuiltInLanguageModelProvider.cs
+++ b/src/MeshWeaver.AI/BuiltInLanguageModelProvider.cs
@@ -478,6 +478,31 @@ public class LanguageModelCatalogOptions
 /// — there is NO central registry. Consumers (Models settings tab,
 /// <c>ModelProviderService.CreateProvider</c>) enumerate the live
 /// <see cref="LanguageModelCatalogOptions.Sources"/>.</para>
+///
+/// <para>🚨 <b>The primary constructor of this record is a SHIPPED BINARY CONTRACT — do not add a
+/// parameter to it, not even one with a default.</b> Provider modules
+/// (<c>MeshWeaver.AI.Anthropic</c>, <c>.OpenAI</c>, <c>.AzureFoundry</c>, <c>.ClaudeCode</c>,
+/// <c>.Copilot</c>, <c>.WebSearch</c>) are compiled separately, land at runtime under
+/// <c>/data/modules</c>, and — per <c>ModuleLandingService</c> — "bind by simple name, and their
+/// contract is API compatibility, not build identity". A call site that omits the optional
+/// arguments still emits a call to the FULL primary constructor as it stood at ITS compile time,
+/// so adding a parameter REPLACES that signature and every already-landed module fails with
+/// <c>MissingMethodException</c> the moment its type initializer runs — which aborts the host.
+/// Prefer an <c>init</c> property for new optional state; if a parameter really must be added,
+/// the OLD signature has to stay reachable, as the overload below keeps it.</para>
+///
+/// <para>This is not hypothetical. Adding <c>Description</c> here as a ninth parameter shipped
+/// green through every source-level gate (it had a default, so nothing failed to compile) and
+/// then crash-looped <c>systemorph</c> for 100 minutes on the next roll: the landed Anthropic
+/// module called the 8-argument constructor that no longer existed. The old pods kept serving,
+/// so every health probe stayed green while helm waited on a pod that could never start.</para>
+///
+/// <para><b>And the repair is additive, because REMOVING a signature is the same mistake pointing
+/// the other way.</b> Dropping <c>Description</c> back out of the primary constructor would fix
+/// every module compiled before the change and break any compiled after it — so the ninth
+/// parameter stays, and the eight-parameter form is restored beside it. Both populations bind.
+/// (<c>scripts/check-record-signatures.py</c> refuses either direction; it landed eleven hours
+/// after the break it would have caught.)</para>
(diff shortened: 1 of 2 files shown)


rbuergi:
### Why the existing gate didn't catch this

Worth recording, because the repo already had a gate aimed squarely at this failure and it passed the break without a murmur.

**`Landed modules compile against this PR`** was added in `3a7f8a4e3` on **2026-08-21** — *"A platform change can break the sixteen landed modules, and nothing here saw it"* — five days **before** `987da5f9e`. It ran on that PR and went green.

It had to. It builds the modules **from source** against the PR's core, and the change was *source*-compatible: `Description = null` has a default, so every provider module compiles against the 9-parameter record without edit. Nothing to report.

But production never runs that compile. It loads `MeshWeaver.AI.Anthropic.dll` **as built days earlier** — Aug 25 01:17, against the 8-parameter record — and that binary calls a constructor the new core no longer has.

> **The gate measures a different artifact than the one that runs.**

Both gates are needed and neither subsumes the other:

| gate | question it answers | sees this break? |
|---|---|---|
| `Landed modules compile against this PR` | can the module still be **built** against new core? | **no** — source-compatible |
| `check-record-signatures.py` | is the shipped **signature** still there? | **yes** |

The natural experiment on the fleet says the same thing from the other side: `memex.meshweaver.cloud` runs `b53aed0aa` — the exact commit of the image systemorph cannot boot — 3/3 healthy, because its Anthropic module was **re-landed Aug 26 07:13**. systemorph's is **Aug 25 01:17**. Same bytes in the image, opposite outcomes, one variable: when the module binary was built.

Which is also why the repair here is additive rather than a revert. Removing the ninth parameter would fix every module built before the change and break every one built after it — and on a fleet where instances re-land modules at different times, *both* populations are live simultaneously. memex-cloud is running the new one right now.

PR #2343, opened 2026-08-26
refactor(test): move the AI content sync-back test out of MeshWeaver.GitSync.Test

rbuergi:
Clears `MeshWeaver.GitSync.Test` of `MeshWeaver.AI` — one more test project off the list in #2276.

## What was coupled

Exactly one file. `AiContentSyncBackTest` asserts that `.AddAI()` registers the disk writer and that `SyncAiContentToRepo` enumerates the live Skill partition onto a `content/ai`-shaped folder. That is **AI-content behaviour that happens to target a git checkout** — it rides AI, not GitSync. (`GitHubSyncTestBase`'s `using MeshWeaver.AI;` was separately just dangling.)

## Why the move is free

`MeshWeaver.AI` already references `MeshWeaver.GitSync` — for the repo client the writer resolves its section root from:

```xml
<ProjectReference Include="..\MeshWeaver.GitSync\MeshWeaver.GitSync.csproj" />   <!-- MeshWeaver.AI.csproj:50 -->
```

so `AddGitHubSyncServices()` is reachable from `MeshWeaver.AI.Test` transitively. **No new project reference in either direction.** The test body is untouched; only its namespace changes, plus a now-explicit `using MeshWeaver.GitSync;` — it had been in the `MeshWeaver.GitSync.Test` namespace, resolving that implicitly.

## Verification

- `MeshWeaver.AI.Test` and `MeshWeaver.GitSync.Test` both build `-c Release -p:CIRun=true -warnaserror`: `0 Warning(s)`, `0 Error(s)`
- the moved test passes in its new home: **1 passed**

No What's New entry — test-only refactor, no user-visible effect.

## Note on what is left

I surveyed the remaining suites and posted the result on #2276. Short version: the rest is **not** mechanical. `Documentation.Test` and `PathResolution.Test` are platform guards that scan one hard-coded module assembly — re-pointing them would keep core green while silently dropping the coverage, so they want the `modules/<Name>/` staging lane instead. `PluginCatalog.Test` uses Agent/Skill node types as the package *under test*, which is a move rather than a re-point. Three further suites are parked behind another session's `test/` sweep.

Refs #2276


rbuergi:
## Re-synced onto the fix for its own failure — not a retry

This PR was red on `OrleansMeshTests.HubWorksAfterDisposal`, which its diff (moving one test file between projects) cannot reach.

I root-caused that failure rather than re-running it: **#2350**, fixed by **#2351**, now merged to main (`f98a96bd7`). The intake shutdown gate was returning an unclassified failure, so a disposal race reached the sender as `ErrorType.Unavailable` while saying "Hub is shutting down" — and the retry that test already contains matches `ErrorType.ShuttingDown`, so it rethrew instead of riding the race out. That is exactly the failure observed here.

So merging main is not "try again and hope": **the base now contains the fix for this PR's specific red.** Revert-check with three dots confirms the branch still contributes only its own three files —

```
test/MeshWeaver.AI.Test/AiContentSyncBackTest.cs             | 4 ++--
test/MeshWeaver.GitSync.Test/GitHubSyncTestBase.cs          | 1 -
test/MeshWeaver.GitSync.Test/MeshWeaver.GitSync.Test.csproj | 1 -
```

— so the merge has reverted nothing of anyone else's.

If it reds again on the same test, that is a genuinely new signal and I will treat it as one rather than merging main a second time.

Issue #2087, opened 2026-08-22
Regression: freshly imported course root intermittently never activates — Subscribe answers 'does not match any registered address pattern' (the #817/#824 announce-loss class)


rbuergi:
## Symptom

On education-repo CI (disposable mesh, current portal image), roughly one shard per run fails because ONE freshly-installed course's **root never activates**: the fresh-learner journey loops for ~12 minutes on

```
[e2e] not entitled yet on /{Course}: no redeem button … The surface currently reads:
"… Page not found: '{Course}/Subscribe' does not match any registered address pattern."
```

i.e. routing never learned the just-imported `Store/Plugin` root. The bootstrap retry on the same portal then often finds the Store itself wedged (`NodeType 'Store/Plugin' never came live: /Store did not render its type's own surface within 300s`).

## Evidence (education repo, 2026-08-22, PR #196 + main)

Victims rotate randomly across runs and shard meshes — each time a course **untouched** by the change under test, failing **before** the PR's new course was installed; the complementary shards go green on the next roll:

| Run | Failed shard | Victim | Rerun result |
|---|---|---|---|
| [32592276434](https://github.com/Systemorph/MeshWeaver.Education/actions/runs/32592276434) | 4/4 | AdvancedBusinessRules | — |
| [32594309666](https://github.com/Systemorph/MeshWeaver.Education/actions/runs/32594309666) | 1/4 + 2/4 | DataModeling (+1) | green on rerun |
| [32595235750](https://github.com/Systemorph/MeshWeaver.Education/actions/runs/32595235750) (main push) | — | — | failure |
| [32597272017](https://github.com/Systemorph/MeshWeaver.Education/actions/runs/32597272017) (main push) | 4/4 | AdvancedBusinessRules | rerun in flight |

No compile errors anywhere; the same commits pass fully on other rolls (e.g. the hourly schedule run [32597336268](https://github.com/Systemorph/MeshWeaver.Education/actions/runs/32597336268) on the same tip is green).

Also observed live on memex.meshweaver.cloud the same day: the first GitSync import of a brand-new course space finished `Imported ImportedWithErrors (68 node(s))` with one root-level typed instance silently missing; a forced re-update landed it.

## Why this looks like the #817 class returning

[...]

PR #2353, opened 2026-08-26
MeshNodeTypeSource's create-flush announces genuine creates on IMeshChangeFeed (#2087)

rbuergi:
Contributes to #2087 — the third and (per the issue's own audit) top-ranked remaining
recurrence of the #817/#824 announce-loss class, after #824 (bulk installer path) and #2257
(post-creation-handler + NodeType-bake paths) fixed the first two.

## The defect

`MeshNodeTypeSource.FlushPendingWrites` persists a per-node hub's own create/delete directly
against `IStorageAdapter` — the class's own doc says so: *"Adds (create) and Deletes go
straight to IStorageAdapter — insta write, no queue."* The file has **zero** references to
`IMeshChangeFeed`. A write that skips the feed leaves a node that EXISTS in storage and does
not exist to the running mesh: a path probed while it was still absent resolves to its
ancestor with a remainder — a perfectly cacheable value — and nothing ever evicts it, so
routing answers `No node found at '…'` for the life of the process. #2087's audit named this
the best remaining candidate for the reported "freshly imported course root intermittently
never activates" flake, but left it unfixed pending more care — an "add" in `UpdateImpl` is
add-RELATIVE-TO-THIS-INSTANCE, not necessarily globally new (a reactivated hub's bookkeeping
starts empty, so its own already-durable node gets reclassified as an "add" too), and
blindly announcing `Created` for every such re-add would double-fire subscribers that treat
`Created` as "this happened right now" — `AccessGrantNotifier` (re-sends the "you've been
granted access" notification) and the invite `EventSubscription`s — on every idle-recycle.

## The fix

Only the genuinely-new case announces: the incoming node's `Version` is captured as `0` (never
minted) *before* `UpdateImpl` re-stamps it, and that `IsInsert` bit rides alongside the queued
node through to the flush. A genuine insert now goes through
`StorageAdapterChangeFeedExtensions.WriteAndPublishCreated`; a re-add of already-durable
content stays a bare `Write`, exactly as before — the row was already reachable, so
re-announcing it would only cost duplicate notifications for no reachability gain.

Deletes have no such ambiguity — a path leaving the collection is always a genuine delete —
so they now always announce via `DeleteAndPublish`.

## Verification

- **Failing-first repro**: `MeshNodeTypeSourceCreateAnnounceTest` drives a brand-new node
  through the exact "add" classification `UpdateImpl` performs when a genuinely new node's
  first durable write lands here (its own per-node hub's create-flush racing `Initialize`'s
  durable-seed read — see `ApplyUpdateViaStream` in `MeshExtensions.cs`, "how imported nodes
  land when the create race is lost"), and asserts the create-flush publishes `Created` on
  the feed. Confirmed to **time out (20s)** against the unpatched code and **pass** with the
  fix.
- Full regression sweep across every existing test referencing `MeshNodeTypeSource`
  (idle-recycle version regression, disposal-pending-flush, refused-write rebase,
  stale-activation seed rollback/durable-first-seed, content-narrowing persistence,
  create-or-update, node-version-counts, path registration, hub-disposal-leak, indeterminate
  probe re-resolution) plus the sibling #2257 tests (`PostCreationNodeIsAnnouncedTest`,
  `PathResolutionCacheTest`) — all green, confirming no behavioural change to version
  stamping or the idle-recycle re-add path.
- `dotnet build -c Release -warnaserror` clean on `MeshWeaver.Graph`,
  `MeshWeaver.Graph.Test`, `MeshWeaver.Hosting.Monolith.Test`, `MeshWeaver.Documentation.Test`.
- `WhatsNewEntryIntegrityTest` + `DocumentationLinkIntegrityTest`: green.

## Honest scope

I could not reproduce the reported education-CI flake itself (no access to that mesh), so I
cannot prove this specific fix eliminates it — only that it closes a real, confirmed
announce-loss gap in the class the issue's own audit ranked as the top remaining candidate.
Not closing #2087 with this PR; leaving that to whoever can verify against the education CI
mesh, per the issue's own standing request.


rbuergi:
## CI red is main-wide and unrelated to this PR — blocked on #2293

The failing check (`Landed modules compile against this PR`) is red on `main` itself, not
because of anything in this diff. Confirmed on three consecutive main runs — `5b1901fb6`,
`b487efcc9`, `f98a96bd7` (the run before, `2699e4a05`, was green):

```
Multiple components use the tag 'MemexMobileMenu'.
  Components: Memex.Portal.Shared.Layout.MemexMobileMenu, Memex.Portal.Shared.Layout.MemexMobileMenu
The type 'MainLayout' / 'Routes' / 'Onboarding' in .../plugins-repo/src/Memex.Portal.Gui/... conflicts
  with the imported type in Memex.Portal.Shared
```

**Cause**: the GUI-extraction program landed its ADDITIVE half (plugins #674) while its DELETION
half (#2293, `+31/-135237`) is still open. `Memex.Portal.Gui` in the plugins repo now declares
`MemexMobileMenu`/`MainLayout`/`Routes`/`Onboarding`/etc., and core still holds the same components
under `memex/Memex.Portal.Shared/Layout/`, so both trees declare the same components and the
landed-modules gate correctly refuses the duplicate definition. This is a genuine duplicate-definition
state, not a flake — rerunning does not help.

This PR's own diff (`src/MeshWeaver.Graph/MeshNodeTypeSource.cs`, its test, and a What's-New entry)
never touches `memex/Memex.Portal.Shared/` or any Blazor GUI surface — zero overlap with the
duplicate. Every other check is green (all 6 test shards, build, docs, plugins compile/bake, etc.),
and there are no unresolved review threads.

**Unblock sequence** (per the session already working it): plugins #701 + #700 land →
#2293's gate goes green → #2293 merges → core's duplicate components disappear → main green again.
No unknowns left in that chain, but no fixed ETA.

Holding this PR here — not rerunning further, not touching `memex/Memex.Portal.Shared/` or
`src/MeshWeaver.Blazor*` to route around it. Will merge once main (and this PR's rebase of it) is
green.

PR #2364, opened 2026-08-26
CD builds the portal host from plugins, and asserts every delivery leg ran

rbuergi:
Raised while watching `main-cd` runs complete in **0 minutes** and report success: *"no cd run should complete in 0 minutes"* and *"we need a check that everything was actually run"*.

### What was already covered

`delivery-verdict` classifies the gate's **verdict** and fails on an empty one, a stuck reconcile, and a red-main push. Those guards are good and unchanged — they were added today after run `32914425222` passed on an empty gate output.

### What wasn't

Nothing checked that a run which **decided to publish** actually executed the nine jobs that carry out a delivery. `delivery-verdict` needed only `[gate, promote]`, so seven legs were invisible to it — and a `skipped` job renders the same grey/green tick as one that ran.

```
success   CD delivered, or had a good reason not to        3s
skipped   Resolve the target commit and decide whether to publish
skipped   Build + push portal-ai image   … and every other job
```

### The change

Every leg becomes a `needs`, and a publishing run asserts each concluded `success` **by name**. `skipped`, `cancelled`, `failure` and **empty** all fail, with the offending jobs named. No "empty means fine" branch — that assumption is what let the verdict pass on an empty gate output in the first place.

Plus a step-summary line saying plainly whether the run **PUBLISHED**, was a **NO-OP**, or **declined**. A 3-second success and a 20-minute delivery are otherwise identical at a glance, which is how a merged fix sat unshipped while its instance waited for it.

### On the 0-minute runs themselves

Those are CD firing on a *PR's* Build-and-Test — it triggers on every completed run while the gate is conditioned to push-on-main. That is a legitimate no-op, so it is **not** made to fail: erroring on every PR would train everyone to ignore the alarm, which is worse than the ambiguity. They are now labelled `⚪ NO-OP` in the summary. The missing thing was a visible difference, not a failure.

### Verified

Both directions under bash: all-success **passes**; one leg `skipped`, `cancelled`, or **empty** each **fails**, naming the leg.

No What's New entry: CI-internal.

### DIFF
--- a/.github/workflows/edge-images.yml
+++ b/.github/workflows/edge-images.yml
@@ -36,6 +36,14 @@ jobs:
     steps:
       - uses: actions/checkout@v7
       # The SDK container build overflows the runner's ~14 GB; reclaim preinstalled tooling.
+      - name: Check out MeshWeaver.Plugins (holds the portal host)
+        uses: actions/checkout@v7
+        with:
+          repository: Systemorph/MeshWeaver.Plugins
+          ref: ${{ vars.MW_PLUGINS_REF || 'main' }}
+          path: plugins-repo
+          ssh-key: ${{ secrets.PLUGINS_REPO_SSH_KEY }}
+          fetch-depth: 1
       - name: Free disk space
         uses: jlumbroso/free-disk-space@main
         with:
@@ -57,7 +65,7 @@ jobs:
         id: vars
         run: |
           set -euo pipefail
-          VERSION=$(dotnet msbuild memex/aspire/Memex.Portal.Distributed/Memex.Portal.Distributed.csproj \
+          VERSION=$(dotnet msbuild src/MeshWeaver.Mesh.Contract/MeshWeaver.Mesh.Contract.csproj \
             -getProperty:Version -p:CIRun=true -nologo | tr -d '[:space:]')
           EDGE_VERSION="${VERSION/.ci./.edge.}"          # rc:    3.0.0-rc1.ci.N -> 3.0.0-rc1.edge.N
           EDGE_VERSION="${EDGE_VERSION/-ci./-edge.}"      # clean: 3.0.0-ci.N    -> 3.0.0-edge.N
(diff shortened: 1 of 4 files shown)


rbuergi:
**A second, sharper instance — this one has a leg that FAILED, not merely skipped, while `delivery-verdict` still reported success.**

The example in the issue body is an all-skipped run. Run [32964902603](https://github.com/Systemorph/MeshWeaver/actions/runs/32964902603) (main `00cc169c5`, 11:44:58Z — the real push-triggered run) is the harder case:

```
run conclusion: failure
  success   Resolve the target commit and decide whether to publish   (gate: publish=true, reason=push)
  failure   Build + push portal-ai image
  success   CD delivered, or had a good reason not to     ← claims delivery
  skipped   Promote: tag the full set (all-or-nothing)
  success   Alert: CD failed on main
```

So on a run that **decided to publish**, **tried**, and **failed**, the job literally named *"CD delivered, or had a good reason not to"* concluded `success`. Its three guards classify the gate's *verdict* — with `publish=true reason=push green=true` none matched, and it fell through to `OK — either something was published, or not publishing was the correct answer`. With `needs: [gate, promote]` it structurally could not see the failing leg.

Two things this adds to the case for the fix as specified:

1. **Skipped is not the only shape.** The proposed remedy — every leg a `needs`, and a publishing run asserting each concluded `success` **by name** — catches this too, but only because it asserts `success` rather than merely "not skipped". A guard written as "did every leg run?" would pass here, since portal-image *did* run. Worth keeping the assertion on the conclusion, not on execution.
2. **The run-level signals were fine.** Overall conclusion was `failure` and `Alert: CD failed on main` did fire, so this was not invisible at the run level — I initially thought it was and was wrong. The defect is narrower and more specific than "CD fails silently": it is one job asserting a fact it cannot observe. That is worth stating precisely in the fix, because the run-level machinery does not need changing.

**Live effect while this is open:** no portal image has published since `00cc169c5` (11:43Z). Newest tag is `3.0.0-rc8.ci.5670` from `2699e4a`, which predates it. Verified with `az acr repository show-tags -n meshweaver --repository memex-portal-ai --orderby time_desc` — which is the check that actually distinguishes "CD is green" from "something shipped", and the one this issue exists to make unnecessary.

rbuergi:
## Review point taken — and the PR is now narrowed to the assertion only

**The NO-OP label:** correct, and it would have done the opposite of what it was for. An empty `reason` is also what an errored/cancelled gate looks like, and under `if: always()` that run would have been stamped "nothing was due". Now keyed on `needs.gate.result == 'skipped'` — the only state that genuinely means no delivery was due. `PUBLISHED` additionally requires `promote == success`, so a run that decided to publish and lost a leg reads **❌ DELIVERY FAILED** — which is exactly what this morning's run `32964902603` was, and what the old label would have called ✅.

Every branch verified under bash: PR-event → NO-OP · real delivery → PUBLISHED · publish-then-leg-failed → DELIVERY FAILED · declined → NO PUBLISH · gate errored/cancelled → INCONCLUSIVE.

**The repoint is reverted.** `PlatformBakeLaneGuard` was right: `main-cd` must check out no other repository, and that is a bright line by design (*"one `repository:` input would make it possible"*). Where the portal image gets built after the host moved to plugins is an architectural cutover — `portal-ai-image.yml` in that repo says in its own header that it is the intended successor, and it has **never run** — not a hotfix to slip in here. Separate decision, separate PR.

What ships in this PR is the part that is unambiguous: a run that decides to publish must have **run every leg**, and the summary must say what happened.

PR #2373, opened 2026-08-26
Stop mocking IMessageHub/IMeshService in 8 test files

### LINKED ISSUE #1810: Tests: 7 test files mock IMessageHub, which AGENTS.md forbids — pre-existing, and it is starting to read as precedent
## The rule

AGENTS.md, under Testing Guidelines:

> **No mocking.** Use `MonolithMeshTestBase` or `OrleansTestBase` — never mock `IMessageHub`, `IMeshService`, or core interfaces.

## The reality

`Substitute.For<IMessageHub>()` appears in **7 test files**:

- `test/MeshWeaver.Graph.Test/StaleAssemblySelfHealWatcherTest.cs`
- `test/MeshWeaver.Graph.Test/OverlaySelfHealWatcherTest.cs`
- `test/MeshWeaver.Graph.Test/NodeTypeRebindWatcherTest.cs`
- `test/MeshWeaver.Graph.Test/NodeTypeEnrichmentDoubleCallTest.cs`
- `test/MeshWeaver.Hosting.Test/PathResolutionCachePoisonTest.cs`
- `test/MeshWeaver.Hosting.Blazor.Test/NavigationServiceTest.cs`
- `test/MeshWeaver.Hosting.Blazor.Test/NavigationProgressTest.cs`

## 🚨 Pre-existing — NOT introduced by the stale-build-banner change

Raising this because the banner PR **rewrote the assertions** in `StaleAssemblySelfHealWatcherTest` (auto-recycle → offer). It did not add the mock and did not add a file to this list. Filing separately, and deliberately **not** fixing it inside that PR: an unrelated test rewrite would obscure the diff that matters.

## Why it is worth an issue rather than a shrug

This is how an absolute decays. The next person finds a mocked `IMessageHub` in a **merged, passing, carefully-commented** test, reasonably concludes the rule is soft, and the one after that stops asking. The rule then survives only in AGENTS.md, where it reads as aspiration rather than practice. An exception that is written down as an exception stays an exception; an exception nobody wrote down becomes the norm.

It also costs something concretely. A mocked hub answers whatever the test stubs, so these tests cannot observe the behaviour the real hub has and the substitute does not: turn-based delivery ordering, the single-threaded action block, disposal semantics, `Set`/`Get` lifetime tied to the hub. That is exactly the surface where this codebase's real defects live — see the wedges and self-heal races these very watchers exist for. A test that cannot see the scheduler cannot pin a scheduling bug.

There is a visible seam in one of them already: `StaleAssemblySelfHealWatcherTest.BuildInstanceHub` carries a comment explaining that `Post`'s return type **cannot** be auto-substituted (`IMessageDelivery` has an internal member → `TypeLoadException` at proxy generation), so it uses `Configure()` to record the call without invoking the auto-value provider. That workaround is the mock fighting the type system, and it is a good signal the boundary is in the wrong place.

## What to do

Migrate to the sanctioned bases — `MonolithMeshTestBase` (full integration: persistence, messaging, DI) or `HubTestBase` (message routing / layout) — one file per PR so each stays reviewable, and delete the substitute along with its workarounds.

Where a genuine unit-level contract is worth keeping (these watchers ARE pure predicate contracts, deliberately split out to be testable without building a hub), the honest fix is to make the seam take what it actually needs — an `IObservable<MeshNode>` in, an action out — rather than a whole `IMessageHub`. That keeps the fast unit test **and** removes the mock, instead of trading one for the other.

## Not urgent, but do not let it rot

No user impact and no CI failure — purely a testing-standards debt. The reason to track it is that its cost is paid later, by someone citing it as permission.

rbuergi:
## Summary

AGENTS.md's Testing Guidelines say: **"No mocking. Use `MonolithMeshTestBase` or `OrleansTestBase` — never mock `IMessageHub`, `IMeshService`, or core interfaces."** #1810 tracked 7 test files that violated this (found by the stale-build-banner PR, filed separately so an unrelated rewrite wouldn't obscure that diff).

I verified the file list myself via `git grep` rather than trusting the issue's count, per its own instruction — it had drifted in both directions:

- **2 of the 7 named files no longer exist**: `test/MeshWeaver.Hosting.Blazor.Test/NavigationServiceTest.cs` and `NavigationProgressTest.cs` — that whole project was removed when #2293 deleted Blazor/GUI from core.
- **3 additional `IMessageHub`/`IMeshService` mocks the issue didn't list**: `OverlayReEvaluationReadTest.cs`, `DeckSlidesCacheTest.cs`, `NotificationServiceTest.cs`.

Net: **8 files converted, 0 `Substitute.For<IMessageHub>()`/`Substitute.For<IMeshService>()` calls remain anywhere in `test/`** (confirmed by a final `git grep`).

## What changed, per file

- **`NodeTypeRebindWatcherTest.cs`, `StaleAssemblySelfHealWatcherTest.cs`** — real hosted hub via `HubTestBase`. Assertions moved from "was `Post` called with the right target" to the REAL effect (the hub actually disposes) — a stronger proof, since a mis-targeted post could never dispose that specific hub. `StaleAssemblySelfHealWatcherTest`'s offer-path assertions (`Get`/`Set` property bag) stayed fully synchronous since those don't touch the action block at all.
- **`OverlaySelfHealWatcherTest.cs`** — this one needed a small production change: `ArmOverlaySelfHeal` in `NodeTypeEnrichmentHelpers.cs` now takes `instancePath`/`jsonOptions`/a `recycle` action/a `reportStuck` action instead of a whole `IMessageHub`. `WithOverlaySelfHeal` (the one real caller) closes the actual hub's `Post`/`Address` into the `recycle` delegate. This was necessary because the test's non-converging-instance contract (`NonConvergingInstance_RecyclesAreSpaced_NotOncePerLadderRung`) drives a synchronous, `TestScheduler`-timed recycle→dispose→re-arm loop across a *simulated hour* — a real hub's genuinely-async `Post` cannot be driven inline with a virtual clock, so the seam had to narrow (matching the issue's own suggested resolution: "make the seam take what it actually needs... rather than a whole `IMessageHub`").
- **`NodeTypeEnrichmentDoubleCallTest.cs`** — real `MonolithMeshTestBase` hub. The mock was exploiting a missing `IWorkspace` DI registration to force a synchronous throw; with a real hub (which always has `IMeshQueryCore`), an unregistered NodeType now hits the real fast-existence-probe path (~3s budget, not the 30-60s slow path) — a more faithful repro of the actual prod failure mode.
- **`OverlayReEvaluationReadTest.cs`** — real `MonolithMeshTestBase` hub with genuine RLS (`ConfigureMeshBase`, no blanket admin grant). "Runs as System" is now proven by a circuit identity with **no read access** to the target partition still succeeding — a real access-control proof, not a captured `UserId` field that could be right while RLS still filters the row. One test (`NoQueryCore_MeansNoReRead`) needs a hub that genuinely has no `IMeshQueryCore` registered — a bare `HubTestBase` router hub (no `AddGraph()`) rather than a mock standing in for one.
- **`DeckSlidesCacheTest.cs`** — a `CountingMeshService` **decorator** wraps the real `IMeshService` (forwards every call, counts `Query` subscriptions) instead of substituting the interface. Fixture slides are seeded via `ConfigureMesh`'s `AddMeshNodes` rather than `CreateNode`, because `SlideNodeType.NodeType` ("Slide") is a **retired** type — `CreateNode` correctly refuses it now ("NodeType 'Slide' is not registered"), which the old mock never caught since it never validated anything.
- **`NotificationServiceTest.cs`** — real `IMeshService` via `MonolithMeshTestBase`; reads back `CreateNode`'s own materialized return value instead of a captured pre-write object.

## Verification

- `dotnet build src/MeshWeaver.Graph -c Release -warnaserror` and both touched test projects: clean, `--no-incremental`.
- Every converted test run **individually** (fresh `.trx` each): `OverlaySelfHealWatcherTest` + `NodeTypeRebindWatcherTest` (21/21), `StaleAssemblySelfHealWatcherTest` (9/9), `NodeTypeEnrichmentDoubleCallTest` + `OverlayReEvaluationReadTest` (6/6), `DeckSlidesCacheTest` (4/4), `NotificationServiceTest` (4/4), `PathResolutionCachePoisonTest` (7/7).
- No test failed once converted to real infrastructure — the mocks weren't hiding a live defect, but two of them (`DeckSlidesCacheTest`'s retired `SlideNodeType`, `NodeTypeEnrichmentDoubleCallTest`'s slow-path-vs-probe-path mismatch) were exercising something the real system no longer does the way the mock assumed.

## Out of scope

- `test/MeshWeaver.Security.Test/SourceQueryJsonOptionsTest.cs` mocks `ITypeRegistry`/`ITypeDefinition`, not `IMessageHub`/`IMeshService` — it already extends `MonolithMeshTestBase` for the hub itself; the narrow type-registry doubles are outside what #1810 and AGENTS.md's rule are about.
- Per the task's stay-out list: nothing touched under `memex/Memex.Portal.Shared/`, `test/MeshWeaver.Hosting.Orleans.Test/`, `test/MeshWeaver.Query.Test/`, `test/MeshWeaver.Hosting.PostgreSql.Test/`.

## What's New

Skipped — test-only change with one internal, behavior-preserving production signature narrowing (`ArmOverlaySelfHeal`, `internal static`, one caller, already updated in this PR). No user-visible effect.

Closes #1810.

### DIFF
--- a/src/MeshWeaver.Graph/Configuration/NodeTypeEnrichmentHelpers.cs
+++ b/src/MeshWeaver.Graph/Configuration/NodeTypeEnrichmentHelpers.cs
@@ -1546,7 +1546,12 @@ internal static class NodeTypeEnrichmentHelpers
                     {
                         instanceHub.RegisterForDisposal(ArmOverlaySelfHeal(
                             meshHub.GetWorkspace().GetMeshNodeStream(nodeType),
-                            instanceHub, nodeType, typeVersionAtOverlay, logger,
+                            instanceHub.Address.ToString(),
+                            instanceHub.JsonSerializerOptions,
+                            recycle: () => instanceHub.Post(
+                                new DisposeRequest(), o => o.WithTarget(instanceHub.Address)),
+                            reportStuck: () => ReportStuckOverlayToAdmins(instanceHub, nodeType, logger),
+                            nodeType, typeVersionAtOverlay, logger,
                             guards: NodeTypeCompilationHelpers.GuardsOf(meshHub),
                             // The escape hatch out of a permanently silent type stream — see
                             // ArmOverlaySelfHeal's re-evaluation route (issue #1814 defect B).
@@ -1799,9 +1804,23 @@ internal static class NodeTypeEnrichmentHelpers
     /// hub that holds this watcher, so a pair whose re-enrichment faults again would otherwise
     /// re-arm at the first rung forever. Null applies no spacing.
     /// </param>
+    /// <remarks>
+    /// 🚨 Takes ONLY what the firing contract needs — <paramref name="instancePath"/> (logging),
+    /// <paramref name="jsonOptions"/> (ContentAs), <paramref name="recycle"/> (the self-heal
+    /// action) and <paramref name="reportStuck"/> (best-effort admin notify) — deliberately NOT a
+    /// whole <c>IMessageHub</c>. <see cref="WithOverlaySelfHeal"/> is the one real caller and
+    /// closes over the actual hub to build these; <c>OverlaySelfHealWatcherTest</c> is the reason
+    /// the seam is this narrow: its non-converging-instance contract drives a synchronous,
+    /// <c>TestScheduler</c>-timed recycle→dispose→re-arm loop across a SIMULATED hour, which a real
+    /// hub's genuinely-async <c>Post</c> cannot do inline with a virtual clock, and AGENTS.md
+    /// forbids standing in a mocked <c>IMessageHub</c> for it (Systemorph/MeshWeaver#1810).
+    /// </remarks>
     internal static IDisposable ArmOverlaySelfHeal(
         IObservable<MeshNode> typeStream,
-        IMessageHub instanceHub,
+        string instancePath,
+        JsonSerializerOptions jsonOptions,
+        Action recycle,
+        Action reportStuck,
         string nodeType,
         long? typeVersionAtOverlay,
         ILogger? logger,
@@ -1811,7 +1830,6 @@ internal static class NodeTypeEnrichmentHelpers
         OverlayHealBudget? budget = null)
     {
         var clock = scheduler ?? Scheduler.Default;
-        var instancePath = instanceHub.Address.ToString();
 
         // ContentAs, never `is NodeTypeDefinition` — the #1669 blindness on un-materialized JSON
         // emissions; see ArmStaleAssemblySelfHeal's predicate note. Deserialize ONCE per node and
@@ -1819,7 +1837,7 @@ internal static class NodeTypeEnrichmentHelpers
         // degraded-content warning logs.
         (NodeTypeDefinition? Def, bool Usable) Evaluate(MeshNode? t)
         {
-            var def = t?.ContentAs<NodeTypeDefinition>(instanceHub.JsonSerializerOptions, logger);
+            var def = t?.ContentAs<NodeTypeDefinition>(jsonOptions, logger);
             return (def, def is not null
                 && NodeTypeCompilationHelpers.HasUsableBuild(t!, def, guards));
         }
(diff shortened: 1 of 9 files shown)


rbuergi:
Good catch — fixed in 395d05e19: bracketed `CreatedAt` to the actual `[before-call, after-call]` window instead of a fixed tolerance.

Issue #2367, opened 2026-08-26
A local portal on current main cannot sign in or render: the Blazor login UI is deleted and the default views arrive only as registry bundles


sierragolflima:
## Summary

Two independent consequences of the GUI migration leave a freshly-updated local portal unusable, and
each one presents as something other than its cause.

1. **No login UI.** `83356b3d5` ("Delete Blazor and the portal GUI from core", #2293 — 389 files,
   −54,420 lines) removed `Login.razor`, `DevLogin.razor` and `DevLoginConfirm.razor`.
   `DevAuthController` survives, so `POST /dev/signin` still works — but nothing renders a way to
   reach it.
2. **No control views.** `DefaultViews` and `GraphViews` were deliberately removed from
   `Modules:Assemblies` (#2169 Phase B2) and now reach a deployment **only** as registry bundles. An
   install that has them uninstalled renders every control as its `ToString()`.

Both together mean a developer following `LocalColimaMac.md` gets a portal they cannot log into,
and — if they get in another way — cannot read.

## How to reproduce

```bash
# 1. A local install on current main
memex-local up && memex-local update

# 2. No login page is served by the Blazor shell
curl -sk -o /dev/null -w "%{http_code}\n" https://memex.localhost:8443/login
#    on a build after 83356b3d5 the Blazor login route is gone

# 3. …but the sign-in endpoint still works, proving only the UI is missing
curl -sk -D - -o /dev/null -X POST https://memex.localhost:8443/dev/signin \
     -d "personId=<an existing user>"
#    HTTP 302, location: /, set-cookie: MemexAuth=…

# 4. Once signed in, controls render as text rather than UI
open https://memex.localhost:8443/Agent
#    NamedAreaControl { Id = , Style = , Readonly = , Skins = …, Area = Details, … }

# 5. The cause: no view pack is loaded
kubectl -n memex logs deploy/memex-portal-deployment | grep ModuleLoad
#    MeshWeaver.AI, MeshWeaver.Social, MeshWeaver.Hosting.Grpc, Mcp, Mail — and no views

# 6. The packages are visible but not installed
kubectl -n memex exec deploy/memex-portal-deployment -- ls /plugin-repos/Plugins/DefaultViews
#    index.json, Gallery, manifest.lock …
kubectl -n memex exec deploy/memex-portal-deployment -- ls /data/modules
[...]

PR #2380, opened 2026-08-26
Local-dev docs: the portal hosts live in MeshWeaver.Plugins now (#2367)

rbuergi:
Fixes the docs half of #2367.

A developer following the local-dev docs from a core checkout got a portal with **no login UI and no control views**. The docs said `dotnet run --project memex/Memex.Portal.Monolith` / `memex/aspire/Memex.AppHost` — paths #2293 deleted, together with the login pages, when the GUI extraction moved the hosts to plugins.

**Twelve files** carried those commands (`AGENTS.md`, `Readme.md`, `Deployment.md`, `LocalDevWorkflow.md`, `MacLocalStack.md`, the two sample GettingStarted pages, the AI provider docs, the E2E README, +2). Every occurrence now points at `../MeshWeaver.Plugins/src/Memex.Portal.Monolith` / `…/Memex.AppHost`, which runs **unchanged from the documented cwd** under the standard sibling layout — plugins' `src/Directory.Build.props` defaults `MeshWeaverRoot` to `../MeshWeaver`. A note at the top of `LocalDevWorkflow.md` explains where the hosts went and how to override the root.

The second half of #2367 (`DefaultViews`/`GraphViews` reaching a deployment only as registry bundles) is #2169 Phase B2's decision, not a docs matter — left open there.

`DocumentationLinkIntegrityTest` passes. No What's New entry: developer docs only.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -465,7 +465,7 @@ The `memex` portal runs on the shared **AKS cluster** `<aks-cluster>` (RG `<aks-
 ```bash
 az acr login -n meshweaver
 # Portal (custom base) AND migration (the migration is what creates schema + the matview):
-dotnet publish memex/aspire/Memex.Portal.Distributed/Memex.Portal.Distributed.csproj -c Release \
+dotnet publish ../MeshWeaver.Plugins/src/Memex.Portal.Distributed/Memex.Portal.Distributed.csproj -c Release \
   -t:PublishContainer -p:ContainerRegistry=meshweaver.azurecr.io \
   -p:ContainerRepository=memex-portal-ai -p:ContainerImageTag=<tag> \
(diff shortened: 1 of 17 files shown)


rbuergi:
Both points correct, and the second pass found more of the first: my rewrite matched the `run`/`watch` forms and not `build`/`publish`, so step 3's rebuild plus the AKS publish commands in `DeploymentAKS.md` and `LocalColimaMac.md` (and a `DebuggingMessageFlow.md` row) still named the deleted path. All now point at the sibling plugins checkout, with step 3 saying so in prose. The README link is an absolute URL — a relative link leaving the repo is broken on GitHub, as you said. Prose citing old `src/MeshWeaver.Blazor.Portal/…` paths as where *past* fixes landed is left as history; only runnable commands were rewritten. Link-integrity gate passes.

Issue #2235, opened 2026-08-25
Release broadcast has never delivered: notify-platform-update succeeds, Hosting/PlatformBuilds stays empty, no satellite dispatch since 08-22


rbuergi:
## Summary

The memex-side release broadcast introduced on 2026-08-23 (`FrameworkReleaseBroadcaster`, `src/MeshWeaver.GitSync`) **has not delivered a single `repository_dispatch` since it shipped.** The fleet still updates only because the `schedule` poll fallback is doing all the work — which is exactly the state invariant #8 exists to prevent, and it is invisible because every part reports success.

## Evidence

**1. The sender works.** `main-cd.yml` → `Notify platform-update registry` fires precisely when `promote` succeeds, and it did so today:

```
run 32869293853   created 2026-08-25T16:00:19Z   sha 8c7e290b1
  Promote: tag the full set (all-or-nothing)   success
  Notify platform-update registry              success
```

That is the promotion of `8c7e290b1` — the build now serving production. (Runs where `notify` shows `skipped` are correct: `if: needs.gate.outputs.publish == 'true' && needs.promote.result == 'success'`, and those runs did not promote.)

**2. The target is configured.** Repo variable `PLATFORM_WEBHOOK_URL` = `https://memex.meshweaver.cloud/api/hooks/Hosting/PlatformBuilds`, and `PLATFORM_WEBHOOK_SECRET` is present.

**3. Nothing lands on the mesh.** On memex, `Hosting/PlatformBuilds` is a **`Markdown` node from 2026-08-19, version 1, with no children at all** — no `main`, no `_Inbox`, nothing:

```
get @Hosting/PlatformBuilds/*          → []
get @Hosting/PlatformBuilds/_Inbox/*   → []
```

`Hosting/Deployment/Source/PlatformBuildInboxWatcher` exists, so the watcher is deployed; it simply has nothing to watch.

**4. No satellite has been dispatched since 08-22.** Every receiver is armed — all four repos carry an *active* (uncommented) `repository_dispatch` trigger with `meshweaver-framework-released` — yet:

| Repo | Most recent `repository_dispatch` runs |
|---|---|
| MeshWeaver.Plugins | 08-22 09:51, 08-22 08:48, 08-22 08:37 |
| MeshWeaver.Education | 08-22 08:37, 08-19 09:50, 08-19 09:43 |
| MeshWeaver.Reinsurance | 08-22 08:37, 08-19 09:50, 08-19 09:43 |
[...]

PR #2382, opened 2026-08-26
main-cd builds and publishes EVERY plugin in the same run as the platform

rbuergi:
**Rebased onto main after #2364 merged** (`557862521`); exactly one commit.

Maintainer decision 2026-08-26: *"run full plugin build in main memex build"*, *"plugins should not have a lane in this sense"*, *"only single plugins on pr, and dependencies"*. This is the main-cd half; the plugins-PR half already exists (`affected-modules.py` narrows to the changed modules + dependency closure).

### What `publish-bake` now does, after the platform's own bake
1. checks out MeshWeaver.Plugins (`vars.MW_PLUGINS_REF || main` — the same steerable pin as portal-image; `PlatformBakeLaneGuard` asserts it is the only foreign repo)
2. **bakes every plugin against the same `mw-plugin-test` image this run built** — identity file + ≥1 bundle, or red
3. gates the bake (renders + executes every Tests area on the baked bytes)
4. **asserts the plugin bake's identity equals the platform bake's** — equal by construction; asserted because two bakes that disagree would publish bundles no portal can adopt together (#1814's class)
5. publishes under source key `plugins`, sealed, with **`RELEASE_VERSION` required and equal to the version this run computed**

### Why 5 matters beyond wiring
A bundle's version used to come from `manifest.lock` while its bytes came from a bake against a pinned image dozens of publishes old — Plugins#695, the route by which an already-fixed defect kept shipping (#2356). Here the version is minted by the run that produced the bytes, and the publish refuses without it. The property, from the review that named it: *"the version this bundle is stamped with was computed by the run that produced its bytes."*

### Retires
plugins' own publish lane (`portal-ai-image.yml` — never run; the publish half of `publish-bake`), the `repository_dispatch` rebake (dead since 08-22, #2235), and the `MW_IMAGE_DIGEST` pin's drift. Plugins PRs still gate only what they touch.

### Verified
`PlatformBakeLaneGuard` 9/9 against this file; YAML parses; `publish-bake` is 15 steps. **Not exercised end to end until it runs on main** — there is no local way to run `publish-bake`. Every piece it composes is already exercised in this job or in `dotnet-test.yml`'s plugin-gate.

No What's New entry: CI-internal.

### DIFF
--- a/.github/workflows/main-cd.yml
+++ b/.github/workflows/main-cd.yml
@@ -1300,7 +1300,6 @@ jobs:
           PLUGINS="$GITHUB_WORKSPACE/plugins-repo"
           [ -f "$PLUGINS/plugin-gate.allow" ] || { echo "::error::plugins-repo/plugin-gate.allow is missing — the known-debt ratchet the plugin bake is configured against. Refusing to bake without it."; exit 1; }
           PLUGINS_SHA="$(git -C "$PLUGINS" rev-parse HEAD)"
-          echo "plugins_sha=$PLUGINS_SHA" >> "$GITHUB_OUTPUT"
           BAKE_PLUGINS="$RUNNER_TEMP/bake-plugins"
           rm -rf "$BAKE_PLUGINS"; mkdir -p "$BAKE_PLUGINS"; chmod 777 "$BAKE_PLUGINS"
           echo "── baking /plugins (every module) against $IMAGE at plugins $PLUGINS_SHA"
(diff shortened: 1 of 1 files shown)


rbuergi:
Correct — removed. `PLUGINS_SHA` reaches the identity-assert and publish steps via `$GITHUB_ENV`; the `$GITHUB_OUTPUT` line had no `id:` and no consumer and read like a contract. `PlatformBakeLaneGuard` still 9/9, YAML parses.

Issue #2383, opened 2026-08-26
_Policy satellites are minted with MainNode = themselves, so they list as content on every package cover


rbuergi:
Found by MeshWeaver.Plugins' new covers sweep (#706) and measured by that session; root cause located in core.

### Symptom
Every package cover's `@@("area/Search")` contents catalog lists `Access Policy` (the `_Policy` satellite), and on packages with children `_Entitlements` too. On the thirteen packages that ship no nodes of their own (`Anthropic`, `AzureBlob`, `AzureFoundry`, `ClaudeCode`, `Copilot`, `Import`, `Indexing`, `Mail`, `Mcp`, `Notifications`, `OpenAI`, `Teams`, `WebSearch`) it is the **only** entry — a public Store cover ending in a "Contents" section whose sole item is an internal security node.

### Root cause — the mint, not the listing
Five `_Policy` nodes read off memex:

```
Chess     mainNode: "Chess"            ← partition ROOT   NOT listed   ✓
RolePlay  mainNode: "RolePlay/_Policy" ← itself           LISTED
Teams     mainNode: "Teams/_Policy"    ← itself           LISTED
Mail      mainNode: "Mail/_Policy"     ← itself           LISTED
```

- `MeshNode.MainNode` defaults to the node's own path (`MeshNode.cs:96`); line 329 states the contract: *"Satellite nodes' MainNode is set to their primary node path at creation time."*
- The catalog's `is:main` filter is exactly `node.MainNode != node.Path` (`StaticNodeQueryProvider.cs:348/367`). **The listing is correct.** A self-referential `_Policy` is, by the framework's own definition, a main node.
- Six built-in providers mint `_Policy` as `new MeshNode("_Policy", RootNamespace)` and never set `MainNode`: `BuiltInAgentProvider:55`, `BuiltInHarnessProvider:24`, `BuiltInLanguageModelProvider:284,322`, `BuiltInSkillProvider:41`, `DocumentationExtensions:135`. **They violate the contract on the type they construct.** Chess is correct only because the gating reconcile (plugins side) rewrites the field when it runs.

### Not an access defect
`MeshWeaver.Security` contains no reference to `MainNode`; `PartitionAccessPolicy` resolves by path (`{partition}/_Policy`). The AGENTS.md note about a wrong `mainNode` being silently ignored is about `AccessAssignment`. `publicRead` on the affected packages is scoped correctly. Cosmetic — but on the one page whose job is to sell.

### Fix shape
[...]

PR #2388, opened 2026-08-26
Built-in _Policy nodes carry their MainNode, so they stop listing as content (#2383)

rbuergi:
Core half of #2383 (the Store-package half is a converging-predicate change in MeshWeaver.Plugins' `PluginGate.cs`, owned there).

Six built-in providers mint a partition's `_Policy` as `new MeshNode("_Policy", ns) { … }` without setting `MainNode`. The plain constructor defaults it to the node's **own** path (`MeshNode.cs:96`), while line 329 of the same type states the satellite contract. The catalog's `is:main` is literally `MainNode != Path`, so a `_Policy` minted this way is a main node by definition — and **"Access Policy" appears under Contents** on the Agent, Skill, LanguageModel, Harness and Doc covers. The listing was never wrong; the mint was.

### Change
- each of the six initializers sets `MainNode = <partition root>` — static seeds, re-registered on boot, so it takes effect on the next deploy with no repair step
- `MeshNode.Satellite(id, mainNode)` for new satellites, with a test pinning both that it points at its main node **and** that the plain constructor does not (the trap this closes)

### Not an access defect
`MeshWeaver.Security` never reads `MainNode`; `PartitionAccessPolicy` resolves by path.

### Verified
Mesh.Contract, AI, Documentation: 0 errors under `-warnaserror`; 2/2 tests. What's New: `Fix`.

### DIFF
--- a/src/MeshWeaver.AI/BuiltInAgentProvider.cs
+++ b/src/MeshWeaver.AI/BuiltInAgentProvider.cs
@@ -54,6 +54,11 @@ public class BuiltInAgentProvider : IStaticNodeProvider
         // The write caps keep the built-in agents unmodifiable.
         yield return new MeshNode("_Policy", RootNamespace)
         {
+            // 🚨 A satellite must point at its MAIN node, not at itself (#2383): the plain
+            // constructor defaults MainNode to the node's own path, which makes _Policy a main
+            // node by the catalog's definition (is:main == MainNode != Path) and lists "Access
+            // Policy" as content on every cover. New satellites: MeshNode.Satellite(id, main).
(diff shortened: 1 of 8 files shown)


rbuergi:
Both correct, both fixed in the follow-up commit: the test now sits in `MeshWeaver.Data.Test` like its siblings, and the `Satellite` doc no longer cites `_Access` — it states the distinction instead (access assignments live under a `_Access/` namespace, not as a sibling id, so they are not what the factory is for). Rebuilt with `-warnaserror`; 2/2 tests under the corrected namespace.

PR #2390, opened 2026-08-26
A test's Catch(...).Repeat() was a 54 kHz busy-wait that killed the whole test host (#2341)

### LINKED ISSUE #2341: MeshWeaver.Threading.Test wedged with exit=124 and is still unexplained — two static ManualResetEventSlim signals the #2013 sweep reported as absent
## The unexplained half of #2013

#2013 is closed by #2335 (`0ba9ba8d8`), which correctly fixed the `MeshWeaver.Hosting.Monolith.Test` / `MeshWeaver.Persistence.Test` residue — including the wedge that hit **`main`** and blocked CD.

**But the second wedge of 2026-08-26 is still unexplained**, and closing #2013 absorbs it:

```
[CI] MeshWeaver.Threading.Test exit=124 TIMEOUT (8m wall-clock cap hit — likely fixture/init hang)
```
PR #2318, shard 5, run [32943789488](https://github.com/Systemorph/MeshWeaver/actions/runs/32943789488), ~08:00Z — identical signature to #2013 (host dies, **no failing test named**), different assembly.

## Correcting the sweep's negative

#2335's report concluded that `MeshWeaver.Threading.Test` contains none of the blocking markers — *"swept for all three ratcheted markers plus other blocking primitives (`ManualResetEventSlim`, `SemaphoreSlim`, `lock`, `Task.Result`). None exist in that assembly."*

That is not correct. On `origin/main`:

```
test/MeshWeaver.Threading.Test/AiPoolDrainJoinsRoundTest.cs:181
    internal static readonly ManualResetEventSlim Entered = new(false);
test/MeshWeaver.Threading.Test/AiPoolDrainJoinsRoundTest.cs:107
    ParkingChatClient.Entered.Wait(TimeSpan.FromSeconds(30), TestContext.Current.CancellationToken)

test/MeshWeaver.Threading.Test/DelegationDrainJoinsParkedToolCallTest.cs:151
    internal static readonly ManualResetEventSlim DelegationEntered = new(false);
test/MeshWeaver.Threading.Test/DelegationDrainJoinsParkedToolCallTest.cs:105
    .Wait(TimeSpan.FromSeconds(60), TestContext.Current.CancellationToken)
```

The negative finding was load-bearing for closing #2013 — it is why the Threading wedge was left without an explanation or an owner.

## What is PROVEN vs what is HYPOTHESIS

**Proven:** both signals are **`static readonly ManualResetEventSlim`**, which AGENTS.md forbids outright — *"A `static` field that is a collection or cache is FORBIDDEN … `[ThreadStatic]` … Process-wide static state survives mesh disposal, so it bleeds across tests"* — and `ManualResetEventSlim` is named explicitly in the hand-woven-async ban. Neither is reset between tests; `Entered` is only ever `Set`.

**Hypothesis, not yet measured:** a static signal that is never reset makes its own guard **vacuous on every run after the first**. `Entered.Wait(30s)` returns `true` immediately because a *previous* test set it, so the assertion *"the streaming stub must be executing before the drain proves anything"* passes without the precondition holding. The test then drives `DrainAll()` against a round that is not actually parked — and the drain is the part with the real blocking behaviour. That would produce an unbounded hang rather than a clean 30 s timeout, which matches `exit=124` with no test named.

I have not proven that chain, which is exactly why this is a filed issue rather than a patch.

## Suggested first step

Make both signals per-instance rather than static (the fixture already scopes per test), then re-run `MeshWeaver.Threading.Test` repeatedly under `DOTNET_PROCESSOR_COUNT=4`. If the vacuous-guard theory holds, the pre-fix version should wedge more readily when the class runs after another that sets the same signal — i.e. only under the FULL assembly, never in isolation (the #2031/#2185 family shape).

🚨 Do not raise the 8 m cap. Over budget means stuck.

rbuergi:
Closes #2341.

`MeshWeaver.Threading.Test` died with `exit=124 TIMEOUT (8m wall-clock cap hit)` and no `.trx` — **12 times in ~15 h across five branches AND `main`** (first `2026-08-25T23:24Z`, latest `2026-08-26T12:35Z`; 400 runs scanned back to 2026-08-22, none before). It reddened shards on PRs whose diffs never touched the threading library.

## The issue's hypothesis is disproved, with evidence

#2341 named the two static `ManualResetEventSlim` signals in that assembly, on the theory that a signal set by one test makes a later test's guard vacuous. That cannot happen: `ParkingChatClient.Entered` and `ParkedDelegationFactory.DelegationEntered` are each referenced **only** from inside their own single-`[Fact]` class, so each is `Set()` and `Wait()`ed exactly once per process, by the same test; both waits are bounded (`Wait(30 s/60 s, ct)`); and in every wedged run's phase trace those two drain tests complete in **0.7–2 s**. They are still wrong per AGENTS.md's no-static-state rule — but they are not this defect, and this PR leaves them alone.

## What it actually is

The failing shards' own artifact carries the answer. In `collected-logs/_meshweaver-test-trace.log` the wedged host logs **112 `INIT_DONE` and 111 `DISPOSE_DONE`** — exactly one test method entered and never returned — and in all three failing runs whose artifacts survive it is the same one: **`SubThreadHangRepro`**. Everything else finishes ~36 s in; the process then looks idle for 7.5 min until CI kills it.

Reproduced locally **2 attempts out of 2** by running six hosts of the assembly at once under `DOTNET_PROCESSOR_COUNT=4` (12 sequential single-host runs on the same box were all green — the repro needs CPU contention). Both local wedges show the same trace imbalance, the same class, and one thread at 100 % CPU whose managed stack names the defect:

```
SubThreadHangRepro.<HungSubThread_WithoutUserCancel_StaysExecuting>d__4.MoveNext()
  ObservableAssertions.Emit -> WaitForFirst -> ToTask(...)          <- SYNCHRONOUS subscribe
    Timeout.Relative.Run -> ToList.Run -> Take.Count.Run
      ObservableAssertions.SubscribeHereOffSyncContext.b__0
        Do.OnNext.Run -> Concat.Run -> TailRecursiveSink.Drain()    <- Repeat()'s engine, spinning
          Producer.SubscribeRaw (Catch) -> Producer.SubscribeRaw (Defer) -> Defer.Run()
            SubThreadHangRepro+<>c__DisplayClass4_0.b__6()          <- the Defer lambda
              AutofacServiceProvider.GetService -> ResolveOperation.ExecuteOperation
```

The site:

```csharp
Observable.Defer(() => workspace.GetMeshNodeStream(subThreadPath) ... .Take(1))
    .Catch<MeshThread?, Exception>(_ => Observable.Empty<MeshThread?>().Delay(200.Milliseconds()))
    .Repeat()
```

`Catch` intercepts **OnError only**. A point read of the not-yet-created sub-thread node **completes without emitting**, so the 200 ms backoff never runs and `Repeat` re-subscribes immediately. Measured with a temporary probe on the real repro: **460 000 re-subscribes in 8.45 s (~54 kHz) and ZERO `OnError`** — the delay everyone believed was there had never executed once, on any run.

Two things turn that from wasteful into fatal:

* **Rx subscribes synchronously**, so the loop runs inside the assertion's own `.ToTask()` call and the `await` is never reached. That puts it out of reach of *every* timeout the harness has: `.Within(15 s)`, `[Fact(Timeout)]`, `methodTimeout`, and `MonolithMeshTestBase`'s hard-deadline watchdog all need the method to return or park at an await. Nothing can fail this test.
* **It starves what it waits for.** The spin pins a core and hammers Autofac + the node-stream cache; on a 4-vCPU runner that is a quarter of the machine plus lock traffic against the very hub that must create the sub-thread node. The condition stops arriving, so the loop that prevents it cannot end. On an idle 18-core box the node lands after ~170 ms and it exits — which is why this stayed invisible locally for the seven weeks the construct has existed.

Why it started on 2026-08-25: the construct dates from 2026-07-02 and its comment records the then-true behaviour (*"GetMeshNodeStream surfaces missing satellite as OnError almost immediately"*). The absent-node answer is now an empty **completion**, which silently made the `Catch` guard vacuous and flipped a 200 ms poll into a 54 kHz spin. Nothing in `src/` shares the shape — its only two `.Repeat()` sites are over sources that carry their own cadence.

## The fix

Both waits move to the composition AGENTS.md prescribes for a node that may not exist yet: the keyed `GetQuery` listing for **existence** (empty-on-absent, live — the index trails the store, so "the index has seen it" implies "the store has it"), then the owner's stream for **content**. No polling, no retry, no resubscribe; every wait is bounded by the assertion's own timeout again because nothing blocks the subscribe. The `n.Content as MeshThread` trap-door casts on the touched lines become `ContentAs<MeshThread>(...)`.

`SubThreadColdStartRepro` and `SubThreadStallRepro` carried the same construct in two more places (same assembly, same delegation wait) and are fixed identically — otherwise the wedge simply moves.

## The control

`UnthrottledResubscribeInTestRatchetGuard` refuses an argument-less `.Repeat()` / `.Retry()` anywhere under `test/`. Its allow file is deliberately **empty**: the three sites that existed are the three fixed here. `.Repeat(count)` / `.Retry(count)` / `.RetryWhen(handler)` are not matched — a bounded count terminates, and `RetryWhen` is the sanctioned replacement (OnError only, its own timer, bounded).

Proved non-vacuous two ways in the same run: a fixture asserts the scanner counts the shape in **code** and ignores it in comments and string literals (load-bearing — the guard's own prose quotes `.Repeat()` repeatedly), and it was **mutation-tested** by reintroducing one call site, which turns it red with `NEW SITE`.

## Verification

| | result |
|---|---|
| BEFORE — 6 concurrent hosts, `DOTNET_PROCESSOR_COUNT=4` | **2 wedges in 2 rounds**; stack + full dump captured for both |
| BEFORE — 12 sequential single-host runs, same box | all green (the repro needs contention) |
| BEFORE — CI | 12 × `exit=124` in the last 400 runs |
| **AFTER — 12 rounds × 6 concurrent hosts = 72 full-assembly runs** | **0 wedges** |
| `MeshWeaver.Threading.Test` | 139/139 green |
| `MeshWeaver.Documentation.Test` | 64/64 green (both ratchets + `DocumentationLinkIntegrityTest`) |
| `dotnet build -c Release -p:CIRun=true -warnaserror` | clean on both touched test projects |

## Follow-up worth a separate look (not changed here)

`MeshNodeStreamCache`'s cached entry tracks a terminal **OnError** (`MarkFaulted` → re-probe) but not a terminal **OnCompleted**, and the absent-node read now answers with an empty completion rather than the NotFound error the 2026-07 comment assumed. No other caller depends on that assumption today (checked: the only `.Repeat()` sites in `src/` are cadence-carrying), so this PR does not touch it — but it is the behaviour change that made the guard above vacuous, and worth a decision of its own.

### DIFF
--- a/src/MeshWeaver.Documentation/Data/WhatsNew/2026-08-26-test-suite-busy-wait.md
+++ b/src/MeshWeaver.Documentation/Data/WhatsNew/2026-08-26-test-suite-busy-wait.md
@@ -1,0 +1,16 @@
+---
+Name: Test suite no longer stalls a release
+Category: Fix
+Description: A busy-wait in three delegation tests could hang a whole test run, blocking releases from shipping.
+Icon: Sparkle
+Order: -20260826
+---
+
+# Test suite no longer stalls a release
+
+Three tests that wait for a delegated sub-thread used a retry loop that only paused between
+attempts when the underlying read failed. When the read instead returned "nothing yet", the loop
+retried tens of thousands of times a second, pinned a processor and never finished — taking the
+whole test run down with it and, on twelve occasions in a single day, blocking the build that ships
+new versions. The three tests now wait for the sub-thread to appear instead of polling for it, and a
+new build check refuses any future test that reintroduces the same loop.
--- a/test/MeshWeaver.Documentation.Test/UnthrottledResubscribeInTestRatchetGuard.cs
+++ b/test/MeshWeaver.Documentation.Test/UnthrottledResubscribeInTestRatchetGuard.cs
@@ -1,0 +1,179 @@
+using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Linq;
+using Xunit;
+
+namespace MeshWeaver.Documentation.Test;
+
+/// <summary>
+/// Governance ratchet for the unthrottled-resubscribe defect class (#2341): a test may not wait on
+/// a mesh stream with Rx's argument-less <c>.Repeat()</c> / <c>.Retry()</c>.
+///
+/// <para><b>Why this is a defect and not a style preference.</b> Both operators re-subscribe to the
+/// source the instant it terminates, with NO delay of their own. The delay people believe they have
+/// added lives in a <c>.Catch(_ =&gt; Observable.Empty&lt;T&gt;().Delay(200.Milliseconds()))</c>
+/// handler — and <c>Catch</c> intercepts <b>OnError only</b>. A point read of a node that does not
+/// exist yet <b>completes without emitting</b>, so that handler never runs and the pair degenerates
+/// into a busy-wait. Measured on the #2341 site: <b>460 000 re-subscribes in 8.4 s (~54 kHz), zero
+/// OnError</b> — the 200 ms backoff never executed once.</para>
+///
+/// <para><b>And it costs the whole shard, not one test.</b> 🚨 Rx subscribes SYNCHRONOUSLY, so the
+/// loop runs inside the assertion's own <c>.ToTask()</c> call and the <c>await</c> is never reached.
+/// That puts it out of reach of every timeout the harness has — <c>.Within(...)</c>,
+/// <c>[Fact(Timeout)]</c>, <c>methodTimeout</c>, and <c>MonolithMeshTestBase</c>'s hard-deadline
+/// watchdog all need the test method to return or to park at an await. One xUnit worker pins a core
+/// forever; on a 4-vCPU runner that also starves the hub that has to create the very node the loop
+/// is waiting for, so it can never terminate. CI reports <c>exit=124 TIMEOUT</c> with no failing
+/// test named and no <c>.trx</c> — 12 occurrences in ~15 h across five branches and <c>main</c>
+/// (2026-08-25/26) before it was root-caused.
+/// This is the same signature as <see cref="BlockingBridgeInTestRatchetGuard"/>'s defect class and
+/// the reason that guard could not see this one: nothing here blocks, it spins.</para>
+///
+/// <para><b>The fix at a site</b> is the composition AGENTS.md prescribes for a node that may not
+/// exist yet — the keyed <c>GetQuery</c> listing for EXISTENCE (empty-on-absent, live) and only then
+/// the owner's stream for CONTENT:
+/// <code>
+/// hub.GetQuery(id, $"path:{parent} scope:children select:path")
+///    .Where(nodes =&gt; nodes.Any(n =&gt; string.Equals(n.Path, target, StringComparison.OrdinalIgnoreCase)))
+///    .Take(1)
+///    .SelectMany(_ =&gt; workspace.GetMeshNodeStream(target))
+///    .Select(n =&gt; n.ContentAs&lt;T&gt;(hub.JsonSerializerOptions))
+///    .Should().Within(...).Match(predicate);
+/// </code>
+/// Where a retry genuinely belongs (an owner that can transiently reject), use
+/// <c>.RetryWhen(errors =&gt; errors.Select((_, i) =&gt; i).TakeWhile(i =&gt; i &lt; n)
+/// .SelectMany(_ =&gt; Observable.Timer(...)))</c> — it re-subscribes on OnError ONLY, carries its own
+/// timer, and is bounded, so a completing source can never turn it into a spin.</para>
+///
+/// <para><b>The allow file is EMPTY, and must stay that way.</b> Unlike the blocking-bridge ratchet
+/// there was no inventory to seed: the three sites that existed (<c>SubThreadHangRepro</c> ×1,
+/// <c>SubThreadColdStartRepro</c> ×2 — all three waiting on a delegation sub-thread) are fixed in
+/// the change that adds this guard, so the correct count is zero. Adding a line here is not a fix.</para>
+/// </summary>
+public class UnthrottledResubscribeInTestRatchetGuard(ITestOutputHelper output)
+{
+    /// <summary>
+    /// <c>test/</c> only. <c>src/</c> has two deliberate <c>.Repeat()</c> sites over sources that
+    /// carry their own cadence (<c>InstanceSyncWorker</c>'s sweep timer,
+    /// <c>PackageInstaller</c>'s), where the operator is doing exactly what it says. The defect is
+    /// specific to a TEST waiting on a mesh read, where the source can terminate immediately and
+    /// nothing bounds the loop — and where the blast radius is the whole test host.
+    /// </summary>
+    private static readonly string[] ScannedRoots = ["test"];
+
+    /// <summary>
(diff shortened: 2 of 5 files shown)


rbuergi:
### The first CI run was red on something this PR cannot cause — evidence, not assertion

`Run tests (shard 2)` failed with **`MeshWeaver.Hosting.Monolith.Test (part 2/2) exit=124 TIMEOUT`**. That is a *different* assembly from the one this PR fixes, and a different mechanism.

**This PR's own assembly passed**, in the same run, on the first attempt:

```
MeshWeaver.Threading.Test  Total: 139, Errors: 0, Failed: 0, Skipped: 0, Not Run: 0, Time: 41.650s
[CI] MeshWeaver.Threading.Test exit=0
```

**The failure is a different defect.** Applying the same phase-trace technique to shard 2's artifact: pid 3520 logs 92 `INIT_DONE` against 91 `DISPOSE_DONE`, and the unbalanced class is `MeshNodeLanguageServiceTest`. It hangs in **teardown**, not in the test body — `DISPOSE_START` and `DISPOSE_INVOKED` both fired, `DISPOSE_DONE` never did — which is the opposite half of the lifecycle from #2341 (where the test method never returns and `DISPOSE_START` is never reached). `teardown-stragglers.log` names the straggler:

```
System.ObjectDisposedException: Instances cannot be resolved ... has already been disposed.
   at Autofac.Extensions.DependencyInjection.AutofacServiceProvider.GetService(Type)
   at MeshWeaver.Graph.MeshDataSourceExtensions.FlushPendingOwnSave(...)  MeshDataSource.cs:1659
   at MeshWeaver.Messaging.MessageHub.DisposeImpl()                       MessageHub.cs:1853
```

Both frames are framework teardown code this PR does not touch, and `main` has not changed either file since this branch's base.

**It is a live, branch-independent flake, and it predates this branch.** The byte-identical marker on two other runs:

| run | branch | when |
|---|---|---|
| [32939560960](https://github.com/Systemorph/MeshWeaver/actions/runs/32939560960) | **`main`** | 2026-08-26T06:45Z |
| [32965944797](https://github.com/Systemorph/MeshWeaver/actions/runs/32965944797) | `refactor/test-suites-drop-ai-batch` | 2026-08-26T11:56Z |

The second is ~2 h before this branch existed **and after #2335 merged (08:58Z)**, so it is a post-fix recurrence of the `Hosting.Monolith.Test` wedge rather than residue #2335 already closed. It deserves its own issue; it is not in scope here.

**What I did about it** — the step AGENTS.md prescribes for a red outside the diff, not a bare re-run: merged current `main` (the branch was 4 merges behind) so CI re-samples on a current base. Revert-checked with `git diff origin/main...HEAD --stat` — still exactly the five intended files — and re-verified locally after the merge: `dotnet build -c Release -p:CIRun=true -warnaserror` clean on both touched test projects, `MeshWeaver.Documentation.Test` 64/64 green.

I will not merge this PR on a red or pending suite.

Issue #2305, opened 2026-08-26
AutoExecute: a Completed response cell still holds the "Generating response..." placeholder — user sees a finished message stuck on the placeholder


rbuergi:
## Symptom

`MeshWeaver.Hosting.Orleans.Test.OrleansAutoExecuteTest.AutoExecute_UpdateThreadMessageContent_RoutesToResponseGrain` fails with:

```
Expected "Generating response..." to contain "Echo:"
```

The response cell reached `Status = Completed` **while still holding the placeholder text**. User-visible as a finished message stuck on *"Generating response…"* with the answer missing.

## It is NOT #2290 — three independent reasons, checked rather than assumed

#2290 (the history loader's `Content is ThreadMessage` cast, fixed by #2294) was the obvious suspect. It is excluded:

1. **The loader is unreachable from this test.** It builds a fresh thread via `BuildThreadWithMessages`, so `Messages` holds only the current user cell and the response cell — both excluded → `cellIds.Count == 0` → immediate empty history, **zero cell reads**.
2. **The write path never had the cast.** `ThreadExecution.cs:1192` already used `ContentAs`, and post-#2294 `git grep` finds **zero** `Content is`/`Content as` `ThreadMessage` sites anywhere in `src/`.
3. **The observed state rules degradation out.** That lambda's `if (node?.Content is not null && current is null) return node;` means an unreadable cell drops *every* write — `Status` could never have reached `Completed`, which the test observed.

## What it looks like instead

A **write-side text-ordering race**, in a region whose own comments already describe a late `"Generating response..."` placeholder push (from a continuation on the history/context read) and the **`terminalLocked` guard** built for exactly that. Worth checking whether that guard is the mechanism — i.e. whether the placeholder push wins against the terminal write rather than being locked out.

Note #2291 records a second failure sharing the `OrleansChatHistoryTest` name with the same signature — *a Completed round whose `Text` is still the placeholder while `Summary` holds the answer* — so this is likely **one defect visible from two tests**, not two.

## Adjacent, not a fix

#2294 makes the history read faster and non-faulting in the degraded case, so it **narrows** this window slightly. It does not close it, and this issue should not be closed against it.

## Acceptance

A round that reaches a terminal `Status` must never still carry the placeholder — assert on the **final** cell state, and pin the ordering so a late placeholder push cannot land after the terminal write.

rbuergi:
## Investigated as part of PR #2326 — hypothesis disproven, issue remains open

[...]

Issue #2291, opened 2026-08-25
Flake: OrleansChatHistoryTest.ColdStart_AgentSeesAllPreviousMessages — TWO different failures share this name (a Completed cell stuck on 'Generating response...', and a fixture deploy that took 13 tests down)


rbuergi:
`MeshWeaver.Hosting.Orleans.Test.OrleansChatHistoryTest.ColdStart_AgentSeesAllPreviousMessages` is in
the flake population with **no issue** (searched `--state all`). Filing one — and 🚨 **two completely
different failures wear this test's name**, so the first job of anyone picking it up is to read which
one they got. Conflating them will waste a cycle.

Both from a sweep of the last 300 `MeshWeaver Build and Test` runs, reading `[FAIL]` lines out of
**every attempt** (a re-run-to-green otherwise hides the attempt that failed).

---

## (A) The REAL one — a COMPLETED round whose `Text` is still the placeholder

Run [32716111656](https://github.com/Systemorph/MeshWeaver/actions/runs/32716111656), attempt 1,
shard 0, 2026-08-24 10:18 — branch `dependabot/npm_and_yarn/clients/react-native/multi-aee9a37cb4`,
i.e. an npm bump that cannot reach the AI thread pipeline.

The wait at `OrleansChatHistoryTest.cs:83` is for the response cell's **`Text`** to contain
"received". It times out at 30 s, and the reported last emission is the giveaway:

```
Text     = Generating response...
Status   = Completed
Summary  = I received 6 messages. [0:system:Echo agent. … | [5:user:Third question - can you see h...]
Timestamp   = 08/24/2026 10:29:36
CompletedAt = 08/24/2026 10:29:43
```

So the round **finished, correctly, seeing all six messages** — the answer is right there in
`Summary` — while `Text` never moved off the framework's own placeholder. The test is not flaky about
history; it is reporting that a terminal cell kept its placeholder.

**This exact state is already named in the code as the thing to prevent.**
`ThreadExecution.cs` (~line 1250), on the `terminalLocked` guard:

> *"Without it that late push writes its own stale `text` verbatim and the user is left staring at
> 'Generating response...' on a cell that is Completed/Error and whose Summary already carries the
> real answer."*

The guard is `terminalLocked = nextStatus != requestedStatus` → a `Streaming` push arriving at an
already-terminal cell keeps `current.Text`. That is right for a late STALE push, but it cannot tell a
[...]

PR #2393, opened 2026-08-26
The Orleans test host allocated a production-sized grain directory per silo (#2346)

rbuergi:
Refs #2346 — and deliberately does **not** close it. This fixes one of the conditions behind the
rotating cast, with a measurement; several named members of that cast have their own, different
mechanisms and are still open (list at the bottom). Auto-closing the umbrella would bury them, which
is the exact failure mode the issue warns about.

## The unit of work is the PROCESS, not the shard

First correction, because it changes what to look at: **the cast is not "shard 0"**. Reading the
shard artifacts of the red runs, `Acme.Test.NodeTypeProgressAreaTest` ran in shard 4,
`Hosting.Test.LiveQueryHandoffDropTest` in shard 2, `AI.Test.*` in shards 3/4 — different runners,
different processes, nothing shared. What the *Orleans* members of the cast share is one thing:
they all live in **`MeshWeaver.Hosting.Orleans.Test`**, which is ONE test host process, and which
opts out of the repo-wide serial policy (`parallelizeTestCollections: true`, `maxParallelThreads: 4`)
so four test classes — four full `TestCluster`s — are live at once.

Position in the run does not predict the failure either. Across six red runs the victim's offset was
41 s, 50 s, 65 s, 106 s, 137 s, 149 s (index 43, 63, 73, 83, 160, 192 of ~198), and the concurrently
running classes had no member in common. So: not accumulation-by-position, not a specific predecessor.

## The mechanism, from CI's own logs

The trx of the run that failed `OrleansGrainTeardownStragglerTest` on main
([32950493525](https://github.com/Systemorph/MeshWeaver/actions/runs/32950493525)) contains Orleans
diagnosing the host it is running in:

```
LocalSiloHealthMonitor: .NET Thread Pool is exhibiting delays of 1.9080052s.
                        Self-monitoring determined that local health is degraded.
Watchdog:               .NET Runtime Platform stalled for 00:00:03.5537093.
                        Total GC Pause duration during that period: 00:00:02.7441830.
                        We are now using a total of 3775MB memory.
                        Collection counts per generation: 0: 1709, 1: 735, 2: 143
```

A **whole-process stall of ~3 s** expires whichever in-test budget happens to be open at that moment.
That is the rotating cast: one condition, a different named victim every run. It also explains the
one member nothing else can — `RoutingBackpressureShapeTest`, which failed a 10 s budget while
having **no cluster and no network at all**: 128 in-memory `Subject`s draining through an `IoPool`.
Only a process-wide stall can break that test.

## Where the 3775 MB went

Reproduced locally (whole assembly, `DOTNET_PROCESSOR_COUNT=4`): RSS climbs strictly monotonically,
86 MB → 4.4 GB in one 124 s run, no sawtooth. A heap dump 95 s in, `dumpheap -stat`:

```
MT           Count      TotalSize  Class Name
00010ddc6b20   194  1,804,494,880  ConcurrentDictionary<GrainId, ConcurrentLruCache<GrainId,
                                    (GrainAddress, int)>.LruItem>.VolatileNode[]
Total 15,294,993 objects, 2,886,214,986 bytes
```

**194 objects, 1.80 GB — 62% of the entire heap**, and `dumpheap -live` confirms all 194 owning
`LruGrainDirectoryCache` instances are live. 194 is one per silo this process ever started.

`GrainDirectoryOptions.CacheSize` defaults to **1,000,000** and Orleans pre-allocates the LRU's
backing `ConcurrentDictionary` at that capacity — a ~9.3 MB Large Object Heap array **per silo**.
Correct for a production silo tracking a million grains; this assembly starts ~194 silos in one
process, each tracking dozens.

## The fix

**`TestGrainDirectorySizing`** — one `ISiloConfigurator` registered centrally in
`OrleansTestCluster.DeployAsync` (the single path every cluster in the assembly is built through, so
new test classes inherit it), setting `CacheSize = 50_000`. That is 20× below the production default
— where ~95% of the saving already is — and still orders of magnitude above any test cluster's grain
count, so the cache can never evict and no behaviour changes. It is registered *before* the caller's
own configurators, so a test that needs a different size can still set one.

This is not a bound being raised to make a test pass; it stops a test fixture from allocating a
production-scale buffer 194 times in one process.

Ten classes build their own `TestClusterBuilder` instead of going through `DeployAsync`, so the
central registration missed them; they opt in on the line after they create the builder.
**`TestClusterSizingGuard`** makes that structural rather than remembered — any file in this project
that constructs a `TestClusterBuilder` must also name `TestGrainDirectorySizing`. The cost it guards
is invisible at the call site (`new TestClusterBuilder()` reads as completely ordinary; the price
shows up as some *unrelated* test failing its budget during a GC pause), which is the shape a control
catches and review does not. It asserts it found the sites before checking them, so it cannot pass
vacuously, and it is verified in both directions: green as committed, red naming the file when the
line is stripped from one.

## Measured, both arms, back to back

Same machine, same half-hour, whole assembly per iteration, `DOTNET_PROCESSOR_COUNT=4`, single
variable (the one `AddSiloBuilderConfigurator` line):

| | peak RSS per run (MB) | median | wall clock | failures |
|---|---|---|---|---|
| without | 4204 / 4248 / 4320 / 4357 / 4423 / 4460 | 4.34 GB | 74.9–98.8 s | 2 / 17 runs |
| with | 2041 / 2246 / 2267 / 2273 / 2352 / 2561 / 2853 / 3080 | 2.31 GB | 73.2–92.4 s | 6 / 24 runs |

The ~2 GB that disappears is the 1.80 GB the dump named — prediction and measurement agree, which is
the reason to trust the mechanism rather than the story.

**Wall clock is unchanged** on an 18-core dev box; an earlier "30 % faster" reading of mine was
machine load and is retracted rather than left standing.

**Flake rate is not distinguishable at this sample size** — 12 % vs 25 %, Fisher p ≈ 0.4 across 41
runs — and the families OVERLAP, which is the part that actually attributes:
`OrleansGrainTeardownStragglerTest` failed in BOTH arms, and the delegation family failed in both too
(`OrleansThreadStreamingTest` without, `OrleansNodeChangePropagationTest` with; plus one
`PodHubTransportTest`). Six named tests, no two arms sharing a majority member — that is the #2346
population itself, which this change neither causes nor cures. Reported as measured rather than
rounded off; and the bound on the worry is that at 50,000 the cache **cannot evict**, so the only way
it could touch these races at all is timing (less GC, faster silo start), never a changed answer.

## Second change: the teardown registry rooted every cluster

`OrleansClusterDisposal` backgrounds each class's cluster teardown and registered the **drain itself**
— a connected `Replay(1)` — in a `ConcurrentBag` nothing ever removed from. A connected `Replay` holds
its source; the source's lambdas close over the `TestCluster` and the Orleans client `IHost`. So every
silo the assembly ever built stayed reachable from a static field for the life of the process, long
after its disposal completed.

Fixed by registering a completion **signal** (`AsyncSubject<Unit>`) that references nothing, and
removing the entry when the drain settles. Guarded by `ClusterDisposalRetentionTest`, which is
deterministic (a `Subject` the test completes itself — no cluster, no timing, no sleep) and asserts
both that the entry leaves the registry and that what the drain captured becomes collectable.

**Stated honestly: this one does not move the number on its own,** and the file now says so. The dump
above was taken *with* this fix already applied and still found all 194 caches live — the silos are
also rooted by undisposed timers in the process-wide `TimerQueue` (`gcroot`: `TimerQueueTimer` →
`GrainTimer` → `PersistentStreamPullingManager` → … → `Orleans.Runtime.Silo`, plus a second family
through an Rx `Sample(…)` periodic timer). Removing one of two roots frees nothing. It is still a real
root, it is cheap, and it now cannot come back; the timer roots are recorded on #2346 as an open,
separate finding.

## What this does NOT explain — deliberately

- **`OrleansMeshTests.HubWorksAfterDisposal`** (the issue's title) — a 2.3 s assertion failure, a
  shutdown NACK classification bug. Already fixed by #2351.
- **`OrleansChatHistoryTest.ColdStart` (#2291) and `OrleansAutoExecuteTest` (#2305)** — the
  Completed-round-still-holding-`"Generating response…"` write-ordering defect. A separate mechanism
  with its own investigation; nothing here touches it.
- **`OrleansGrainTeardownStragglerTest`'s 30 s "activation never leaves the catalog"** — reproduced
  locally in BOTH arms of the A/B, so it is independent of this change.
- **The 45 s silo-silent window** documented in `OrleansClusterDisposal` — untouched, and the file's
  correction note says so.
- **`Acme.Test.NodeTypeProgressAreaTest`, `Hosting.Test.LiveQueryHandoffDropTest`, `AI.Test.*`** —
  different shards, different processes.

## No What's New entry

Per the `pullrequest` skill: pure-internal change (test fixtures only, `test/` exclusively), with no
user-visible effect.

### DIFF
--- a/test/MeshWeaver.Hosting.Orleans.Test/ClusterDisposalRetentionTest.cs
+++ b/test/MeshWeaver.Hosting.Orleans.Test/ClusterDisposalRetentionTest.cs
@@ -40,22 +40,18 @@ public class ClusterDisposalRetentionTest
     [Fact]
     public void ASettledTeardown_RetainsNothingItCapturedSoItsClusterCanBeCollected()
     {
-        // Other test classes are tearing their clusters down concurrently, so the registry is not
-        // empty — the invariant is that it RETURNS to its prior depth, not that it is ever zero.
-        var before = OrleansClusterDisposal.PendingCount;
-
         var gate = new Subject<Unit>();
-        var captured = Enqueue(gate);
+        var (id, captured) = Enqueue(gate);
 
         // Settle the teardown, exactly as the real ordered stop→dispose chain settles when the last
         // leg completes on the I/O pool.
         gate.OnNext(Unit.Default);
         gate.OnCompleted();
 
-        // Deterministic half of the guard: the entry is gone. (The old bag never removed one, so the
-        // registry — and everything each entry reached — grew for the whole run.)
-        OrleansClusterDisposal.PendingCount.Should().BeLessThanOrEqualTo(
-            before,
+        // Deterministic half of the guard: THIS entry is gone. Asserted per-id, never as a total
+        // count — other test classes tear their clusters down concurrently, so a count is not a
+        // property this test could own.
+        OrleansClusterDisposal.IsPending(id).Should().BeFalse(
             "a settled teardown must leave the registry, or it accumulates for the life of the process");
 
         Collect();
@@ -68,21 +64,21 @@ public class ClusterDisposalRetentionTest
     }
 
     /// <summary>
-    /// Registers a drain that captures a fresh object, and hands back only a weak reference to it.
-    /// <see cref="MethodImplOptions.NoInlining"/> so the closure's display class is unreachable from
-    /// the caller's frame once this returns — otherwise a live stack slot, not the registry, would
-    /// decide the outcome.
+    /// Registers a drain that captures a fresh object, and hands back its registry id plus only a
+    /// WEAK reference to what it captured. <see cref="MethodImplOptions.NoInlining"/> so the
+    /// closure's display class is unreachable from the caller's frame once this returns — otherwise
+    /// a live stack slot, not the registry, would decide the outcome.
     /// </summary>
     [MethodImpl(MethodImplOptions.NoInlining)]
-    private static WeakReference Enqueue(IObservable<Unit> gate)
+    private static (long Id, WeakReference Captured) Enqueue(IObservable<Unit> gate)
     {
         var captured = new object();
-        OrleansClusterDisposal.Enqueue(gate.Select(_ =>
+        var id = OrleansClusterDisposal.Enqueue(gate.Select(_ =>
         {
             GC.KeepAlive(captured);
             return Unit.Default;
         }));
-        return new WeakReference(captured);
+        return (id, new WeakReference(captured));
     }
 
     /// <summary>Two forced, blocking, compacting full collections — enough for a plain object with
--- a/test/MeshWeaver.Hosting.Orleans.Test/ClusterDisposalRetentionTest.cs
+++ b/test/MeshWeaver.Hosting.Orleans.Test/ClusterDisposalRetentionTest.cs
@@ -1,0 +1,98 @@
+using System;
+using System.Reactive;
+using System.Reactive.Linq;
+using System.Reactive.Subjects;
+using System.Runtime.CompilerServices;
+using Xunit;
+
+namespace MeshWeaver.Hosting.Orleans.Test;
+
+/// <summary>
+/// Regression guard for issue #2346 — the shard-0 Orleans "rotating cast" of flakes.
+///
+/// <para><b>The defect.</b> <see cref="OrleansClusterDisposal"/> backgrounds each test class's
+/// <c>TestCluster</c> teardown and registers it so the assembly fixture can await every one of them
+/// at the end of the run. It used to register the DRAIN itself, as a connected
+/// <c>Replay(1)</c>, in a bag nothing ever removed from. A connected <c>Replay</c> holds its source;
+/// the source is a <c>SelectMany</c> chain whose lambdas close over the <c>TestCluster</c> and the
+/// Orleans client <c>IHost</c>. So every silo this assembly ever built — ~90 of them, each with its
+/// own Autofac container, grain catalog, serializer codecs, mesh hubs, workspaces, seeded MeshNodes
+/// and collectible NodeType load contexts — stayed reachable until the process exited, even though
+/// its disposal had long since completed.</para>
+///
+/// <para><b>Why it fails TESTS rather than merely wasting memory.</b> The xUnit test host runs
+/// workstation GC (Orleans logs <c>ServerGC=False</c> at every silo start), so a multi-GB gen-2 heap
+/// buys multi-second blocking pauses. CI caught Orleans' own health monitor naming them —
+/// <c>".NET Thread Pool is exhibiting delays of 1.9080052s"</c> and <c>".NET Runtime Platform
+/// stalled for 00:00:03.55 … We are now using a total of 3775MB memory"</c> — and a whole-process
+/// stall of that size expires whichever in-test budget is open at that moment. That is the rotating
+/// victim: one condition, a different named test every run. Measured on this repo before the fix,
+/// RSS across one 124 s run of this assembly climbed strictly monotonically from 86 MB to 4.6 GB
+/// with no sawtooth at all — retention, not the working set of the concurrently-running clusters.</para>
+///
+/// <para><b>What this test pins.</b> Once a backgrounded teardown has SETTLED, nothing the drain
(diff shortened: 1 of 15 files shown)


rbuergi:
Copilot review — three threads, addressed and resolved:

**1. `Interlocked` needs `using System.Threading;`** — it does not: `Directory.Build.props` sets
`<ImplicitUsings>enable</ImplicitUsings>`, and the SDK's implicit set for a library includes
`System.Threading`. Verified rather than assumed:
`dotnet build test/MeshWeaver.Hosting.Orleans.Test -c Release -warnaserror` → `0 Warning(s),
0 Error(s)`, including a `--no-incremental` build so it is a real compile and not an up-to-date
no-op. No change.

**2. `PendingCount` is racy under `maxParallelThreads: 4`** — correct, and fixed in `74fdd43b`,
which is exactly the suggested shape: `Enqueue` now returns its registry id and the guard asserts
`IsPending(id)` on its own entry. A total count is not a property any single class in this assembly
can own. Thanks — that one would have shipped a flake into the PR that exists to remove flakes.

**3. `Pending` is a `static` mutable collection.** Deliberate, and it is the narrow case AGENTS.md's
rule is not aimed at — but the objection deserves the specific answer rather than a wave-off:

- The thing this registry coordinates *is* process-scoped. `OrleansDisposalDrainFixture` (an xUnit
  **assembly** fixture) must block on every backgrounded cluster teardown after the last test class
  has run and before the native runner's foreground-thread check, or the shard force-exits non-zero
  with "Foreground threads were left running" on an otherwise-green run. `DisposeInBackground` is
  called from `OrleansTestBase.DisposeAsync` and `SharedOrleansFixture.DisposeAsync` — neither of
  which xUnit hands an assembly-fixture instance (the fixture is constructed by
  `OrleansSharedTestBase` itself, not injected), so an instance registry would mean threading a
  reference through both base classes and every derived class's constructor. The static `IoPool` two
  lines below exists for the same reason.
- The harm the rule names — *"the moment you add a `Clear()` for test isolation, that method IS the
  proof of the bug"* — is precisely what this PR **removes**. The old shape accumulated forever and
  nothing could ever be removed from it; the new one drops each entry when its own drain settles,
  guarded by `ClusterDisposalRetentionTest`, which fails with the #2346 message if retention comes
  back (verified in both directions: it passes on the fixed shape, and fails with that exact message
  when a temporary probe reintroduces the old bag-of-connected-`Replay`).

Happy to take the instance-registry refactor as a follow-up if the maintainer prefers it; it is a
mechanical but wide change (~90 test classes' constructors) and would be noise inside a PR whose
point is one measured line.

rbuergi:
**Parked, not merged — GitHub Actions is in a declared major outage** (opened 15:11 UTC, throttled
inbound traffic / upstream Vitess). This PR's run `32984555673` shows `Build solution (once)` failing
with *"The job was not acquired by Runner of type hosted even after multiple attempts"* — runner
starvation, not this diff. Per the outage guidance I am **not** re-running and **not** pushing to
retrigger (run creation is what is failing), and I will not merge on a green I cannot trust.

**Nothing in this PR's evidence comes from the outage window.** Every CI artifact I analysed predates
15:11 UTC — runs 32950493525 (08:58), 32951361671 (09:07), 32951676794 (09:11), 32956130892 (10:01),
32970091147 (12:43) and 32834479916 (yesterday) — and the entire before/after measurement is local,
41 whole-assembly runs on this machine. The heap dump likewise. So no conclusion here rests on
outage collateral.

**Local verification stands as the evidence until CI is trustworthy again:**
`dotnet build test/MeshWeaver.Hosting.Orleans.Test -c Release -warnaserror` → `0 Warning(s),
0 Error(s)` (including a `--no-incremental` run); the whole assembly green on the committed tree
(208 tests); and both new guards proven in BOTH directions — green as committed, red with the exact
#2346 message when a temporary probe reintroduces the old retention / when the sizing line is
stripped from one file.

Ready to merge on a green CI run once Actions recovers.

PR #2399, opened 2026-08-26
Sibling satellites derive their MainNode at the write boundary (#2383)

### LINKED ISSUE #2383: _Policy satellites are minted with MainNode = themselves, so they list as content on every package cover

rbuergi:
Closes #2383. Completes the core half — #2388 fixed six of the fifteen mint sites; this fixes the root and the remaining nine.

## Root cause

`MeshNode.MainNode` self-defaults to the node's own path. The framework already has a compensating normalizer — `MeshExtensions.NormalizeSatelliteMainNode`, applied on **both** the `CreateNode` and `UpsertNode` paths — but it was gated on `meshConfig.IsSatelliteNodeType(node.NodeType)`, i.e. only the types whose definition node carries `IsSatelliteType` and which therefore also live in their own satellite **table**.

The `_`-prefixed **siblings** that share `mesh_nodes` — `_Policy`, `_Provider`, `_GitSync`, `_DefaultInstallLedger` — are deliberately *not* such types (`SatelliteTableMapping` line 71 records why: routing them to a satellite table hid `{ns}/_Policy` from the permission evaluator). So they were normalised by **nothing**, and every writer had to remember the field by hand. Fifteen writers exist in core; none did.

`MainNode == Path` is the framework's literal definition of a MAIN node, so such a satellite becomes content:
- the catalog's `is:main` is exactly `MainNode != Path` (`PostgreSqlSqlGenerator` emits `n.main_node = n.path`);
- `search_across_schemas` **hard-filters every union branch** on the same predicate — that one is not opt-in, so the row reached mesh-wide search as well as the Contents section.

## Why not just fix the mint sites again

#2388 corrected six by hand. Its grep — `new MeshNode("_` — cannot match the target-typed spelling `new("_Policy", ns)`, and two writers are raw SQL. Nine were missed. Hand-sweeping is the wrong instrument here, so:

- **`NormalizeSatelliteMainNode` now also fires on a satellite-shaped ID** (`SatelliteTableMapping.IsSiblingSatellite`), so `CreateNode`/`UpsertNode` derive the field for `_Policy` and every future sibling. Flagging the types instead was not an option — the flag also *relocates* the node (`CreateLayoutArea` files a flagged type under `{ns}/_{Type}/{id}`) and routes it to a satellite table, which is the regression `SatelliteTableMapping` warns about.
- **The nine missed writers are corrected**: `DocumentationNodeProvider` (the **second** Doc policy mint), `RoleNodeType`, `LicenseNodeType`, `GlobalSettingsNodeType`, `AccessControlLayoutArea`, `InstanceAutoRegistrationService`, `AssignmentNodeFactory`, plus `PostgreSqlAccessControl` and `SnowflakeAccessControl`. Static seeds bypass the handler entirely, so the mint fixes are needed *alongside* the framework fix, not instead of it.
- **A guard sweeps the static seeds the mesh actually serves**, so the next writer that forgets is caught by a control rather than by a grep.

## Two traps the classifier has to avoid

Both are real nodes in this repo, and both are pinned by tests:

1. **A partition may legitimately be named with a leading underscore.** `GlobalSettingsNodeType.SettingsPath` is literally `_Setting`. The owner is therefore still derived by `OwnerOfSatellitePath` over the **namespace**, never by scanning the path — a path scan resolves `_Setting/_Policy` to `""`, the empty prefix that `COALESCE(main_node, namespace)` projects as a **root-scope grant**.
2. **`Admin/Partition/_Access` is a partition DECLARATION**, not a satellite: its id is the partition's *name*, and it is a main node of the catalog that lists it. Re-pointing its MainNode would delete it from that listing. **The guard caught this on its first run** — my own predicate was wrong, and the control found it.

`SnowflakeAccessControl` also had `main_node` and `path` sharing one variable, so the INSERT wrote the policy's own path into **both** columns. They are now separate; `path` is unchanged.

## Is this an access defect? No — but the issue's reasoning for that was wrong

The issue concluded "not an access defect" from *"`MeshWeaver.Security` contains no reference to `MainNode`"*. **`src/MeshWeaver.Security/` does not exist** — the security rules live in `src/MeshWeaver.Graph/Security/`, and they do read `MainNode`. The conclusion happens to hold, but for a different reason, so it is worth stating properly:

- **Both policy readers resolve by `id = '_Policy'` keyed on `namespace`** — `rebuild_user_effective_permissions` (`PostgreSqlSchemaInitializer.cs:1894`, `:1921`) and the in-memory `PermissionEvaluator` (`:443`, `:764`) — **never by `main_node`**. So what a policy grants or denies is unaffected by this change, in both directions.
- `MainNode` **is** security-relevant in general: `SatelliteAccessRule` delegates a satellite's permissions to it and has an explicit degenerate branch that abandons delegation the moment `MainNode == Path`; grants project at `COALESCE(main_node, namespace)`; and the SQL read clause matches `n.main_node LIKE prefix || '%'`. But `PartitionAccessPolicy` registers no `SatelliteAccessRule` and is not an `AccessAssignment`, so none of those apply to the population here.

**Verdict: disclosure-shaped, not an access-control bypass** — an internal governance node listed as page content and returned by mesh-wide search. Real, worth fixing, not urgent.

**No StackOverflow risk either** (the self-referential-`record` failure mode): `MainNode` is a `string` path, `MeshNode` has no object-valued back-reference (`Content` is `object?` but nothing in `src/` ever puts a `MeshNode` in it), `GetHashCode` deliberately excludes `Content`, and nothing anywhere recurses by following `MainNode` — every read is a single dereference.

## Already-minted rows

- **DB-synced partitions repair themselves, with no migration.** `PartitionSourceFingerprint` frames `MainNode` into the per-partition hash, so changing the static seeds changes the fingerprint; the static-repo importer then takes the full re-import path instead of the content-skip and upserts the corrected node. Worth flagging that #2388's "static seeds … no repair step" was not quite right on its own: `SkillStaticRepoSource`/`HarnessStaticRepoSource` explicitly include the `_Policy` in `EnumerateSourceNodes`, so those rows *are* durable — it is the fingerprint change that heals them.
- **Rows written by `AccessControlLayoutArea` or the raw-SQL writers before this change keep their old value.** They are cosmetic per the section above, so no `LogonAction` and no migration: a `RunOnce` action would cost a per-user write across every partition to correct a listing artefact, and a raw `UPDATE` on a live portal bypasses the workspace cache (AGENTS.md). New writes are correct from here.
- **Store packages are owned in MeshWeaver.Plugins** (`PluginGate.cs:936/956`, which assigns the self-reference *explicitly*, and whose `PolicyMatches` compares content only) — per the issue's own split.

## Verification

Release + `-warnaserror`, 0 errors: Mesh.Contract, Graph, Documentation, PluginCatalog, Hosting.PostgreSql, Hosting.Snowflake, Hosting.Monolith.TestBase.

**2692 tests green** — Data 390, Security 376, Graph 1626, PathResolution 155, NodeOperations 76, Documentation 69.

**Both directions proven by reverting the fix and re-running:**

| reverted | result |
|---|---|
| the four static mint fixes | sweep fails naming `Doc/_Policy`, `License/_Policy`, `Role/_Policy`, `_Setting/_Policy` |
| the normalization widening | `CreatingASiblingSatellite_DerivesItsMainNode` fails, `Actual: "TestData/_Policy"` |

The two negative controls — ordinary content stays a main node, an explicit `MainNode` is preserved — pass in **every** configuration, which is what bounds the blast radius. The guard also asserts a named inventory floor, so it fails loudly rather than passing vacuously if the composition ever stops registering a seed.

What's New: `Category: Fix`.

### DIFF
--- a/src/MeshWeaver.AI/BuiltInAgentProvider.cs
+++ b/src/MeshWeaver.AI/BuiltInAgentProvider.cs
@@ -56,8 +56,9 @@ public class BuiltInAgentProvider : IStaticNodeProvider
         {
             // 🚨 A satellite must point at its MAIN node, not at itself (#2383): the plain
             // constructor defaults MainNode to the node's own path, which makes _Policy a main
-            // node by the catalog's definition (is:main == MainNode != Path) and lists "Access
-            // Policy" as content on every cover. New satellites: MeshNode.Satellite(id, main).
+            // node by the catalog's definition (is:main KEEPS exactly the rows where MainNode ==
+            // Path) and lists "Access Policy" as content on every cover. New satellites:
+            // MeshNode.Satellite(id, main).
             MainNode = RootNamespace,
             NodeType = "PartitionAccessPolicy",
             Name = "Access Policy",
--- a/src/MeshWeaver.AI/BuiltInHarnessProvider.cs
+++ b/src/MeshWeaver.AI/BuiltInHarnessProvider.cs
@@ -25,8 +25,9 @@ public sealed class BuiltInHarnessProvider(IEnumerable<IHarness> harnesses) : IS
         {
             // 🚨 A satellite must point at its MAIN node, not at itself (#2383): the plain
             // constructor defaults MainNode to the node's own path, which makes _Policy a main
-            // node by the catalog's definition (is:main == MainNode != Path) and lists "Access
-            // Policy" as content on every cover. New satellites: MeshNode.Satellite(id, main).
+            // node by the catalog's definition (is:main KEEPS exactly the rows where MainNode ==
+            // Path) and lists "Access Policy" as content on every cover. New satellites:
+            // MeshNode.Satellite(id, main).
             MainNode = HarnessNodeType.RootNamespace,
             NodeType = "PartitionAccessPolicy",
             Name = "Access Policy",
--- a/src/MeshWeaver.AI/BuiltInLanguageModelProvider.cs
+++ b/src/MeshWeaver.AI/BuiltInLanguageModelProvider.cs
@@ -285,8 +285,9 @@ public class BuiltInLanguageModelProvider : IStaticNodeProvider
         {
             // 🚨 A satellite must point at its MAIN node, not at itself (#2383): the plain
             // constructor defaults MainNode to the node's own path, which makes _Policy a main
-            // node by the catalog's definition (is:main == MainNode != Path) and lists "Access
-            // Policy" as content on every cover. New satellites: MeshNode.Satellite(id, main).
+            // node by the catalog's definition (is:main KEEPS exactly the rows where MainNode ==
+            // Path) and lists "Access Policy" as content on every cover. New satellites:
+            // MeshNode.Satellite(id, main).
             MainNode = ModelProviderNodeType.RootNamespace,
             NodeType = "PartitionAccessPolicy",
             Name = "Access Policy",
@@ -328,8 +329,9 @@ public class BuiltInLanguageModelProvider : IStaticNodeProvider
             {
                 // 🚨 A satellite must point at its MAIN node, not at itself (#2383): the plain
                 // constructor defaults MainNode to the node's own path, which makes _Policy a main
-                // node by the catalog's definition (is:main == MainNode != Path) and lists "Access
-                // Policy" as content on every cover. New satellites: MeshNode.Satellite(id, main).
+                // node by the catalog's definition (is:main KEEPS exactly the rows where MainNode ==
+                // Path) and lists "Access Policy" as content on every cover. New satellites:
+                // MeshNode.Satellite(id, main).
                 MainNode = LanguageModelNodeType.RootNamespace,
                 NodeType = "PartitionAccessPolicy",
                 Name = "Access Policy",
--- a/src/MeshWeaver.AI/BuiltInSkillProvider.cs
+++ b/src/MeshWeaver.AI/BuiltInSkillProvider.cs
@@ -42,8 +42,9 @@ public class BuiltInSkillProvider : IStaticNodeProvider
         {
             // 🚨 A satellite must point at its MAIN node, not at itself (#2383): the plain
             // constructor defaults MainNode to the node's own path, which makes _Policy a main
-            // node by the catalog's definition (is:main == MainNode != Path) and lists "Access
-            // Policy" as content on every cover. New satellites: MeshNode.Satellite(id, main).
+            // node by the catalog's definition (is:main KEEPS exactly the rows where MainNode ==
+            // Path) and lists "Access Policy" as content on every cover. New satellites:
+            // MeshNode.Satellite(id, main).
             MainNode = SkillNodeType.RootNamespace,
             NodeType = "PartitionAccessPolicy",
             Name = "Access Policy",
--- a/src/MeshWeaver.Documentation/Data/WhatsNew/2026-08-26-governance-nodes-stop-listing-as-page-content.md
+++ b/src/MeshWeaver.Documentation/Data/WhatsNew/2026-08-26-governance-nodes-stop-listing-as-page-content.md
@@ -1,0 +1,23 @@
+---
+Name: Governance nodes stop listing as page content
+Category: Fix
+Description: Access policies, install ledgers and other internal bookkeeping no longer appear under Contents or in mesh-wide search on the pages they belong to.
+Icon: ShieldTask
+Order: -20260826
+---
+
+# Governance nodes stop listing as page content
+
+Internal bookkeeping filed alongside a page — its access policy, an install ledger, a sync
+configuration — no longer shows up as if it were content of that page. It had been appearing under
+**Contents**, and turning up in search results across the whole workspace, because each of these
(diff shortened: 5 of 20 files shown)


rbuergi:
## Parked before merge — GitHub Actions major outage (opened 15:11Z)

Not merging on a green from this window, and not re-running anything: run *creation* is among what is failing, so a red here would be evidence about the outage, not about this change. Any CI result on this PR from the last hour should be read as untrustworthy in both directions.

**Local verification stands on its own** and is the instrument for this change (see the commit message for the full list): 8 src projects at `0 Error(s)` under `-c Release -warnaserror`, and **2722 tests green** — Data 390, Security 376, Graph 1626, PathResolution 155, NodeOperations 76, Documentation 69, Hosting.PostgreSql 26 (the container-backed access-control and permission-projection suites that exercise the changed SQL writer), Hosting.Monolith 4. Both directions were proven by reverting each half and re-running, not by assertion.

## Copilot review addressed

Both threads were **right**, and the error was wider than the two lines they flagged. I had written `is:main == MainNode != Path` — that is the **exclusion** test, not the selection test. The code keeps exactly the rows where `MainNode == Path` (`PostgreSqlSqlGenerator` emits `n.main_node = n.path`; `StaticNodeQueryProvider` *skips* a node when `MainNode != Path`). A satellite is normally excluded **precisely because** its MainNode differs from its path; one pointing at itself passes instead.

The conclusion is unchanged, but the stated predicate was backwards — so I fixed it in **all nine places**, not just the two flagged. Seven of them came from #2388 and are already on `main`: `MeshNode.Satellite`'s own doc comment, `DocumentationExtensions`, and the four `BuiltIn*Provider` files. Worth naming how it happened: I inherited the phrasing from the issue text instead of reading `StaticNodeQueryProvider` underneath it.

Amended rather than stacked, so the commit message no longer carries the backwards predicate either.

## Ready to merge when Actions recovers

Preconditions for whoever picks this up: the `MeshWeaver Build and Test` suite `COMPLETED/SUCCESS`, and — because this PR **adds** test files — a run whose shard jobs actually **executed** rather than reusing an already-green tree:

```
gh run view <run-id> --repo Systemorph/MeshWeaver --json jobs \
  --jq '[.jobs[] | select(.name | startswith("Run tests")) | .conclusion]'
```

All `skipped` ⇒ that run proves nothing about the new tests.

PR #2408, opened 2026-08-26
🚨 main does not compile — restore the MeshOperations using I deleted as a phantom

rbuergi:
**main is broken and it is my regression.** Merging this restores it.

## The break

`48b3e31a3` (#2349) removed `using MeshWeaver.AI;` from `memex/Memex.Portal.Shared/Api/MeshApiEndpoints.cs` as the "fourth phantom" of a repo-wide sweep. Line 335 takes a `Func<MeshOperations, IObservable<string>>`, and `MeshOperations` lives in namespace `MeshWeaver.AI`.

Control, on **pristine main `c753ad2` with no edits**:

```
error CS0246: The type or namespace name 'MeshOperations' could not be found
Memex.Portal.Shared → Build FAILED, 1 Error(s)
```

With the using restored: `0 Error(s)`.

## Why the sweep was wrong

It classified imports by asking whether the project can reach the **MeshWeaver.AI assembly** through `ProjectReference`. `Memex.Portal.Shared` cannot — and the import was still load-bearing, because:

> `MeshOperations` **ships** in `MeshWeaver.Mesh.Operations` while **declaring** `namespace MeshWeaver.AI`.

That namespace is a **frozen binary contract** (#2370) — `ec56366a1` kept it deliberately, because a plugin compiled against it breaks if it moves.

The right question is *"does any referenced assembly contribute a type to this namespace"*, not *"can we reach the assembly of the same name"*.

🚨 **This is the `FuzzyScorer` pattern with the opposite intent.** `FuzzyScorer` declaring `MeshWeaver.AI.Completion` from `MeshWeaver.Data` was an accident that manufactured phantom imports, and #2349 correctly fixed it. `MeshOperations` declaring `MeshWeaver.AI` from `MeshWeaver.Mesh.Operations` is deliberate and load-bearing. Same shape, opposite meaning — I documented the first and then failed to recognise the second in the same sweep.

The using is restored **with the reason it exists**, so the next sweep does not delete it again.

## It merged green, which is a second finding

The required check passed while `Memex.Portal.Shared` did not compile. My final verification was a solution build on the **pre-merge branch**; nothing rebuilt that project against the merged tree. Worth a look at whether the green-tree reuse path can mark a suite green for a tree that was never built in that combination.

## Verification

- pristine main: **FAILS** (`CS0246`)
- this commit: `dotnet build memex/Memex.Portal.Shared -c Release -p:CIRun=true -warnaserror` → **0 Error(s)**

Refs #2344, #2370

### DIFF
--- a/memex/Memex.Portal.Shared/Api/MeshApiEndpoints.cs
+++ b/memex/Memex.Portal.Shared/Api/MeshApiEndpoints.cs
@@ -6,6 +6,11 @@ using MeshWeaver.Hosting.AspNetCore.Portal; // PortalApplication
 using MeshWeaver.Hosting.AspNetCore;    // SessionHubResolver, McpConfiguration
 using MeshWeaver.Mesh.Security;         // WellKnownUsers
 using MeshWeaver.Mesh.Services;
+using MeshWeaver.AI;                    // MeshOperations — the TYPE ships in
+                                        // MeshWeaver.Mesh.Operations, but its namespace is a
+                                        // FROZEN binary contract (#2370), so this import is
+                                        // load-bearing and is NOT an AI-assembly dependency.
+                                        // Removed as a 'phantom' in #2349 and restored here.
 using MeshWeaver.Messaging;             // AccessService
 using Memex.Portal.Shared.Authentication;
 using Microsoft.AspNetCore.Builder;


rbuergi:
## Control confirms it: main genuinely does not compile

I nearly withdrew this PR on the strength of `memex-portal-ai:71bd288` existing in ACR — a published image seemed impossible if `Memex.Portal.Shared` were broken. That reasoning was wrong: `71bd288` published **before** `c753ad2`, so the image says nothing about the current tip.

The clean control — a detached worktree pinned to `c753ad2`, nothing modified, full solution build with CI's flags:

```
HEAD: c753ad2f5
using count: 0
PRISTINE SOLUTION EXIT=1
MeshApiEndpoints.cs(335,14): error CS0246: 'MeshOperations' could not be found
Build FAILED — 1 Error(s)
```

My two earlier attempts at this control were both invalid, and worth recording so the next reader does not trust them:

1. a **single-project** build, where CI does a solution build;
2. a "pristine" run where I used `git checkout -- .` — which reverts the working tree but **not HEAD**, so it built this very hotfix commit and reported success.

Only the third attempt actually tested pristine main.

So the regression stands as described, this fix is required, and everything merged after `48b3e31a3` is unpublishable until it lands — CD reports `green=false` and skips every build leg.

Auto-merge armed; it will land the moment CI is green and not otherwise.

PR #2415, opened 2026-08-26
🚨 main is red — two parallel hotfixes both added the MeshOperations using (CS0105)

rbuergi:
**main does not compile.** `using MeshWeaver.AI;` appears twice in `MeshApiEndpoints.cs`, and `CS0105` is an error under `-warnaserror`.

## How

My #2349 removed that import and broke main. **Two sessions then fixed it independently:**

| commit | what it added |
|---|---|
| `c01ed6946` | "HOTFIX: main is red — MeshApiEndpoints lost sight of MeshOperations" — bare import |
| `3a6f95553` (#2408) | annotated import explaining why it must never be deleted |

Both merged. Neither collided in git — they touched different lines — so the duplicate only appears at COMPILE time.

## The fix

Keep the **annotated** import (it carries the explanation of why the import exists, which is exactly what was missing when I deleted it), drop the bare duplicate.

```
removed the bare duplicate at line 6: using MeshWeaver.AI;
remaining: 1
Memex.Portal.Shared → 0 Error(s)
```

## 🚨 The lesson, which I had already been taught today

A defect can exist **only in the combination**. I verified main green at `831e63b` and said so publicly; the other hotfix merged afterwards and my statement was stale the moment I made it.

A clean build on my own branch says nothing about main once main has moved — **only main's own post-merge run is authoritative**. That is precisely what caught the original regression, and precisely what I failed to consult before declaring main healthy.

Refs #2344, #2370, #2408

### DIFF
--- a/memex/Memex.Portal.Shared/Api/MeshApiEndpoints.cs
+++ b/memex/Memex.Portal.Shared/Api/MeshApiEndpoints.cs
@@ -3,7 +3,6 @@
 // modules through [assembly: TypeForwardedTo], and a forwarder cannot rename — so the namespace is
 // frozen at whatever it was when those modules were built. Do not "tidy" this using to match the
 // assembly name; that is what broke main at 17:26.
-using MeshWeaver.AI;
 using MeshWeaver.Mesh;
 using System.Reactive.Linq;
 using System.Reactive.Threading.Tasks;


rbuergi:
**Closing as already fixed — `main` is green. This is not a rejection of the diagnosis, which was correct.**

The CS0105 landed in **#2414** (merged 18:2x), which removed the duplicate block. `origin/main` now carries **exactly one** `using MeshWeaver.AI;`, and `Memex.Portal.Shared` builds clean under `-c Release -warnaserror` — I verified that locally against the tip rather than inferring it from the merge.

That is also why this PR shows conflicts: the change it makes is already applied.

## What actually happened, since it is worth recording

**Three sessions independently diagnosed the same break and wrote the same fix.** The original `CS0246` was fixed by #2408 at 17:44 (main green at 17:44:57). I diagnosed it independently, wrote #2411, and merged it at 18:01 **without re-checking whether main was still red** — producing two identical usings and CS0105. Then this PR and its sibling were both opened to fix *that*.

So the duplicate-using bug was itself caused by duplicated effort, and then attracted more of it. The root cause is not the `using` at all: **a red main is exactly when several sessions converge on the same fix, and none of us checked whether the failure still reproduced on the current tip before landing.** That check is seconds; skipping it cost three PRs and a second red main.

## One thing from here worth keeping

#2416's consolidated comment is **better than what is on main** — it folds the banner and the inline note into one directive and records the double-restore history, which is precisely the context that would stop a fourth person "tidying" the import away again. I have not cherry-picked it, because main is green and sits directly in front of a blocked deploy chain, and churning it for comment prose would cost a CI cycle at the wrong moment. If you want it, land it as a standalone docs-only change once the manifest image has rolled — it has my support.

PR #2416, opened 2026-08-26
fix: main does not build — the MeshWeaver.AI using was restored TWICE

rbuergi:
## main is red

`main` at `7d7bf903e` fails **Build solution (once)** with:

```
error CS0105: The using directive for 'MeshWeaver.AI' appeared previously in this namespace
```

`memex/Memex.Portal.Shared/Api/MeshApiEndpoints.cs` carries `using MeshWeaver.AI;` **twice** — once at line 6 and again at line 15.

## How it got there

1. **#2349** removed the using as a "phantom". main went red with `CS0246: MeshOperations could not be found`.
2. **#2408** restored it at the top of the file, under a banner comment.
3. **#2411** restored it *again* further down, inline, with its own explanation.

Both landed. The second restore could not see the first, because the first was fifty lines away behind a different comment style.

## The fix

Keep **one** directive, and fold both explanations into the banner that sits directly above it — including the part each author had to rediscover: `MeshOperations` ships in assembly `MeshWeaver.Mesh.Operations` but keeps namespace `MeshWeaver.AI`, frozen by the `TypeForwardedTo` contract from #2370. The import looks removable and is load-bearing, and removing it has now cost three red-main incidents.

Splitting the directive from its rationale is what let this happen twice. One directive, one explanation, kept together.

## Verified

`dotnet build memex/Memex.Portal.Shared/Memex.Portal.Shared.csproj` → **Build succeeded**, one `using MeshWeaver.AI;` remaining.

This also unblocks #2413 and #2395, which are byte-identical to main on this file and inherit the failure.

### DIFF
--- a/memex/Memex.Portal.Shared/Api/MeshApiEndpoints.cs
+++ b/memex/Memex.Portal.Shared/Api/MeshApiEndpoints.cs
@@ -3,6 +3,11 @@
 // modules through [assembly: TypeForwardedTo], and a forwarder cannot rename — so the namespace is
 // frozen at whatever it was when those modules were built. Do not "tidy" this using to match the
 // assembly name; that is what broke main at 17:26.
+//
+// This import is therefore load-bearing and is NOT an AI-assembly dependency. It was removed as a
+// "phantom" in #2349 and restored twice independently (#2408 and #2411) — once with this banner and
+// once inline further down, which is how main ended up with the SAME using twice and CS0105 under
+// warnings-as-errors. One directive, one explanation: keep them together.
 using MeshWeaver.AI;
 using MeshWeaver.Mesh;
 using System.Reactive.Linq;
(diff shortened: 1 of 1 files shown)


rbuergi:
**Closing as already fixed — `main` is green. This is not a rejection of the diagnosis, which was correct.**

The CS0105 landed in **#2414** (merged 18:2x), which removed the duplicate block. `origin/main` now carries **exactly one** `using MeshWeaver.AI;`, and `Memex.Portal.Shared` builds clean under `-c Release -warnaserror` — I verified that locally against the tip rather than inferring it from the merge.

That is also why this PR shows conflicts: the change it makes is already applied.

## What actually happened, since it is worth recording

**Three sessions independently diagnosed the same break and wrote the same fix.** The original `CS0246` was fixed by #2408 at 17:44 (main green at 17:44:57). I diagnosed it independently, wrote #2411, and merged it at 18:01 **without re-checking whether main was still red** — producing two identical usings and CS0105. Then this PR and its sibling were both opened to fix *that*.

So the duplicate-using bug was itself caused by duplicated effort, and then attracted more of it. The root cause is not the `using` at all: **a red main is exactly when several sessions converge on the same fix, and none of us checked whether the failure still reproduced on the current tip before landing.** That check is seconds; skipping it cost three PRs and a second red main.

## One thing from here worth keeping

#2416's consolidated comment is **better than what is on main** — it folds the banner and the inline note into one directive and records the double-restore history, which is precisely the context that would stop a fourth person "tidying" the import away again. I have not cherry-picked it, because main is green and sits directly in front of a blocked deploy chain, and churning it for comment prose would cost a CI cycle at the wrong moment. If you want it, land it as a standalone docs-only change once the manifest image has rolled — it has my support.

Issue #2127, opened 2026-08-24
Build race: concurrent LayoutMeshModuleClosures restores clobber a shared project.assets.json


rbuergi:
## Signature

`Build solution (once)` (run 32578666522, PR #2078, 2026-08-22T14:33) failed with:

```
/usr/share/dotnet/sdk/10.0.400/NuGet.targets(198,5): error : The file '/home/runner/work/MeshWeaver/MeshWeaver/src/MeshWeaver.AI/obj/project.assets.json' already exists. [/home/runner/work/MeshWeaver/MeshWeaver/src/MeshWeaver.Social/MeshWeaver.Social.csproj]
memex/MeshModulesPublish.targets(279,5): error MSB4181: The "MSBuild" task returned false but did not log an error.  (x2)
```

The PR's diff was ONE markdown file (`src/MeshWeaver.Documentation/Data/Architecture/PlatformAndContent.md`), and main's run on the identical base tree (865dc1794, run 32577873729) was green minutes earlier — this is a nondeterministic race in the build itself, not a change-induced break.

## Mechanism

`LayoutMeshModuleClosuresAfterBuild` (memex/MeshModulesPublish.targets) fires `AfterTargets="Build"` on every host that imports it. In the failing run, three hosts finished Build within ~7 s of each other and each launched its own closure lane:

- 14:33:16.5 `Memex.Portal.Monolith` → `Determining projects to restore...` at 14:33:16.88
- 14:33:20.5 `Memex.Portal.Distributed`
- 14:33:23.8 `test/Memex.Portal.Shared.Test` (subset lane, Cosmos+Snowflake) → `Determining projects to restore...` at 14:33:23.88, and a third restore at 14:33:25.69

[...]

Issue #2234, opened 2026-08-25
A landed module built against a newer platform aborted every new pod on the registry: optional-parameter ctor change is binary-breaking and raised no minMeshVersion floor


rbuergi:
## Impact

memex-cloud (the fleet's plugin registry) **could not start a new portal pod for ~90 minutes** on 2026-08-25. Every replacement pod died ~2 s into boot (`exit 139`, SIGABRT, a 178 MB core dump per attempt). Only two pods that had booted *before* the offending module landed kept the portal alive, and they could not be replaced — one eviction from an outage.

The knock-on effects consumed most of an afternoon before the cause was found, because none of them looked like a boot failure:

- **No image adoption.** The deployment kept trying to roll, the new pod died, the rollout never completed → the self-updater never finished a pass → `Admin/UpdatePolicy.checkedAt` froze at 14:44:36 with `latestAvailableTag: ci.5337` while four newer image sets had published.
- **No package content.** Five registry-delivered packages (DefaultViews, GraphViews, EntityViews, Import, Northwind — the ones with no `_GitSync`) never received their gallery nodes. Packages that landed before the stall did: `Analysis/Examples` 13:44:06, `Radzen/Examples` 14:13:13, `DefaultViews/Examples` absent.
- It masqueraded as "the same image behaves differently on different pods", which sent the investigation toward pod state rather than version skew.

## Root cause: an optional parameter on a record's primary constructor is BINARY-breaking, and no floor rose

```
Unhandled exception. System.MissingMethodException: Method not found:
'Void MeshWeaver.AI.LanguageModelCatalogSource..ctor(String, String, Int32, String, String,
 ImmutableArray`1<String>, Boolean, MeshWeaver.AI.ProviderKind, String)'
   at MeshWeaver.AI.AzureFoundry.AzureFoundryProvidersAttribute.<get_Nodes>b__1_0(IServiceCollection)
   at MeshWeaver.Mesh.MeshBuilder.InstallAssemblies(String[])
   at Memex.Portal.Shared.MemexConfiguration.ConfigureMemexMesh(...)
```

- `987da5f9e` (2026-08-25 13:32) added a 9th parameter — `string? Description = null` — to `LanguageModelCatalogSource`'s primary constructor.
- The running image was `ci.5337` = commit `5544f9f6b`, which carries the **8-parameter** version.
- The landed `MeshWeaver.AI.AzureFoundry` module was built against the **9-parameter** version.
[...]

PR #2420, opened 2026-08-26
Don't put the AI engine on core's SHARED closure lane — it lands in the portal image

rbuergi:
## What this became

This branch started as the flip you'd expect: `Memex.LocalMesh` off its ships-the-bits `ProjectReference` for the AI engine and onto the `modules/` **closure lane**. That worked — verified on a real publish, correct layout, no double-ship, subset honoured.

**It was still the wrong change, and the PR now reverts it**, keeping the one gate that is independently right.

## Why the flip was wrong

`@(MeshModuleClosure)` is **shared**. The portal host in MeshWeaver.Plugins imports `memex/MeshModulesPublish.targets` through `$(MeshWeaverRoot)` and sets **no** `MeshModulesClosureSubset` — an image host must lay out the complete inventory. So a row added here for a *sidecar* also puts `modules/MeshWeaver.AI/` into the **portal image**.

A closure module and a registry module of the same name are mutually exclusive: the landing service answers **409 on every instance**. That is precisely the gate the AI Store entry (Plugins#687) is waiting on — and the portal is clean today, with zero `MeshWeaver.AI` mentions in its csproj or its `Modules:Assemblies`.

I claimed in this PR's original description that the flip was *"a precondition for #2276"*. It is the reverse: it **re-arms** that gate.

Host-scoping the row is not a cheap escape either. The closure **restore** is deliberately invoked with a host-*independent* property set (#2127) so MSBuild dedupes every host's lane into one build configuration; a per-host module forks it and reopens the `project.assets.json` race documented there — whose victim was `src/MeshWeaver.AI` itself.

## What ships instead

The sidecar keeps the `ProjectReference`, and **the reason now lives in the csproj at the point of temptation** — the double-ship looks like obvious cleanup, and the cost of "fixing" it is invisible from this repo. The double-ship costs bits in a sidecar binary; it is the portal image that must stay clean.

The endgame is neither lane: when Plugins#687 lands, the engine ships from the **registry** and this host lists it under `Modules:Required`, exactly as `MeshWeaver.Speech` does since it left for MeshWeaver.Plugins.

## Kept — a real gate, independent of the above

`ModuleDeclarationIntegrityTest` now fails when a host declares a module that only the **closure lane** ships but does not import the lane worker.

> `Memex.LocalMesh.csproj: modules are declared on the CLOSURE lane for this host, but the project does not import memex/MeshModulesPublish.targets — so the lane never runs, modules/ is never laid out, and every lane-only entry above is SKIPPED at load on a green build.`

**Proven to fail** by removing the `Import`. This closes a real hole: **`Memex.LocalMesh` never imported that file at all** — it left with the portal hosts for MeshWeaver.Plugins (83356b3d5) — so its `modules/` folder was never created and *any* module it declared beyond a compiled-in one would have been skipped silently. Inventory membership is not shipping.

## Reverted along with the flip

The `Microsoft.Extensions.AI.Abstractions` pin (the `ProjectReference` carries the 10.8.3 pin again, so the version skew it fixed cannot occur), `AiEngineModuleLayoutTest`, and the What's New entry — all three describe a layout this host no longer has.


rbuergi:
## Found and fixed a real defect in this change before merge — the flip downgraded a shared package

Publishing the sidecar (not just building it) surfaced this:

```
LayoutMeshModuleClosures: KEPT Microsoft.Extensions.AI.Abstractions.dll in modules/
  — the app root ships a same-named file with DIFFERENT bytes …
```

| | version |
|---|---|
| app root | `10.4.1` (10.400.126.16803) |
| `modules/MeshWeaver.AI/` | `10.8.3` (10.800.326.37702) |

**Cause.** While the engine was a `ProjectReference`, *its* direct `PackageReference` (which the central pin does reach) unified this host's graph at 10.8.3. Dropping the reference floated the app root down to what `ModelContextProtocol.Core` asks for transitively — and `CentralPackageTransitivePinningEnabled` is not set repo-wide, so the bare `PackageVersion` pin cannot reach a pure-transitive dependency. `Directory.Packages.props` documents exactly this inertness about `Microsoft.AspNetCore.DataProtection`.

**Why it would have broken chat, not just looked untidy.** `MeshBuilder.InstallAssemblies` loads modules with `Assembly.LoadFrom` — the **default `AssemblyLoadContext`**, no per-module isolation. First assembly of a simple name wins, and the app base directory is probed ahead of a module's load-from path. The sidecar hosts MCP, so 10.4.1 loads first and the engine binds against it despite compiling against 10.8.3. A 10.8-only member then surfaces as the `MissingMethodException` that #2234's per-module isolation catches: the module drops out reported incompatible, and every send is refused with *"NodeType 'Thread' is not registered"* — the exact regression 8dd8eeecf fixed, on a green build and a green rollout.

🚨 **The prune's `KEPT …` line does not save this.** *"the module must carry the one it binds to"* only wins the binding under ALC isolation, and there is none here. It should be read as a **skew warning to investigate**, not as a fix. Worth knowing for every future module flip, since a `dotnet build` never lays out `modules/` — only a publish shows these lines.

**Fix** (30c72e15d): the narrow remedy `Directory.Packages.props` itself names — *"adding the direct reference where the package flows"* — a versionless `<PackageReference>` in `Memex.LocalMesh`, so the central pin bites for that project only. Deliberately **not** repo-wide transitive pinning, whose blast radius is resolution-wide and belongs in its own PR.

**Verified on a real publish**: app root now carries 10.800.326.37702, the module's copy prunes away as a same-identity duplicate (`modules/` 60 → 59 files), and the skew warning is gone. Solution build `0 Warning(s) 0 Error(s)` — which also rules out an `NU1504` duplicate-reference collision.

### Left alone deliberately

`Google.Protobuf` stays skewed (app `3.31.1`, module `3.30.2`) and is **not** a regression: the app root already carried 3.31.1 before this change, so the engine bound to it then and binds to it now. Unchanged behaviour, and the safe direction — newer serving older-compiled — unlike the downgrade above. Direction is what decides severity here.

### Two smaller self-corrections

- The comment marking the removed reference wrote `MeshWeaver.AI.csproj` literally, so `git grep` still listed this host among projects referencing the engine — a false positive on the exact sweep #2276 is driven by, in the one file whose point is that the reference is gone.
- The closure-lane comment named YamlDotNet and Humanizer as the engine's private deps. Both prune away; replaced with what the built layout actually holds.

PR #2427, opened 2026-08-26
The gRPC Kestrel test owns its ports — closes the FreePort() TOCTOU race (#2379)

### LINKED ISSUE #2379: MeshGrpcTrustedKestrelTest: FreePort() is a TOCTOU race — Kestrel binds a port another process took in between
`MeshWeaver.Hosting.Grpc.Test.MeshGrpcTrustedKestrelTest.Trusted_port_is_detected_on_real_kestrel_with_config_bound_options` failed on #2364 (shard 5), whose diff is three workflow YAML files and one Documentation.Test guard — it cannot reach a gRPC Kestrel port binding.

```
System.IO.IOException : Failed to bind to address http://127.0.0.1:35145: address already in use.
```

4.6 s — a real failure, not a hang.

### Mechanism (`MeshGrpcTrustedKestrelTest.cs:55-74`)

```csharp
var publicPort  = FreePort();
var trustedPort = FreePort();
...
kestrel.ListenLocalhost(publicPort,  l => l.Protocols = HttpProtocols.Http2);
kestrel.ListenLocalhost(trustedPort, l => l.Protocols = HttpProtocols.Http2);
```

`FreePort()` binds a socket to `:0`, reads the assigned port, and **releases it**. Kestrel binds that number later. Anything else on the runner — another test, another shard's process, an ephemeral outbound connection — can take it in the gap. Classic time-of-check/time-of-use. It is rarer on a quiet box and common on a loaded CI runner, which is why it reads as a flake.

### Why the obvious fix is not trivial

The clean cure for TOCTOU is `ListenLocalhost(0)` and reading the bound port back from `IServerAddressesFeature`. But this test's whole subject is `Grpc:TrustedPort` **bound from configuration before the host starts** (the chart injects `Grpc__TrustedPort`), so the trusted port number must be known up front — the product code under test reads it at startup. Binding `:0` removes the number the test exists to inject. Options, none of which are a retry:

1. Make trust detection consult `IServerAddressesFeature` after start (product change; then the test can bind `:0` and configure post-start).
2. Keep the socket from `FreePort()` **open** with `SO_REUSEADDR` and let Kestrel bind alongside — platform-dependent; works on Linux, not reliably elsewhere.
3. Reserve the port range for this test from the OS ephemeral range (`net.ipv4.ip_local_port_range`) — runner-config, not a code fix.

**Not** acceptable: retry-on-`AddressInUse` (a bound raised to hide a race — AGENTS.md "No band-aids").

Filed rather than fixed in #2364 because it is a real race in a test that PR does not touch, and it deserves the right fix rather than a fast one.

rbuergi:
Closes #2379.

## The race

`MeshGrpcTrustedKestrelTest.FreePort()` bound a socket to `:0`, read the port the kernel assigned, and then **released it**. Kestrel bound that number later. Between the release and the bind the port was owned by nobody, so anything else on the machine — another test, another CI shard's process, an ephemeral outbound connection — could take it, and Kestrel then failed to start:

```
System.IO.IOException : Failed to bind to address http://127.0.0.1:35145: address already in use.
```

Narrow window on an idle box, wide one on a loaded runner — which is why it reads as a flake while being a plain ownership bug. It redded #2364, whose diff is three workflow YAMLs and one Documentation.Test guard.

## The fix — structural, not a retry

`LoopbackPort.Reserve()` binds **and listens** and never closes: the kernel refuses every rival bind for as long as it lives. Kestrel does not open a second socket for that port either — it is handed *this* one via `SocketTransportOptions.CreateBoundListenSocket`, so **discovery and use are the same socket** and there is nothing in between to lose.

No retry loop, no wider range, no sleep — each of those only makes the window smaller (AGENTS.md → "No band-aids": never raise a bound to make a race less likely).

The issue's option 1 — bind `:0` and read the port back from `IServerAddressesFeature` — is not available here: the test's whole subject is `Grpc:TrustedPort` bound from configuration **before** the host starts, so the number has to exist up front. Option 2 (`SO_REUSEADDR` alongside) and option 3 (reserving from the OS ephemeral range) are both declined for the reasons the issue gives.

`ListenLocalhost` → `Listen(IPAddress.Loopback, …)`: the former binds the IPv4 **and** IPv6 loopback, i.e. two sockets for one port number, and only one of them can be the reserved one. The clients now dial `127.0.0.1` explicitly — also the address `GrpcOptions.TrustedPort` documents (`http://127.0.0.1:{TrustedPort}`).

## The test fails without the fix — deterministically

`LoopbackPortReservationTest.Kestrel_binds_a_reserved_port_no_rival_can_take` runs the losing interleaving on purpose rather than waiting for a loaded runner to produce it. Both halves are load-bearing:

1. the rival must be **refused** — with a released port it wins;
2. Kestrel must bind that port **while the reservation is still held** — only possible because it adopts the reserved socket. Drop the handover and this half fails, because the reservation itself would be what blocks Kestrel.

**Evidence** (macOS 25.5.0, .NET 10.0.301, `-c Release -warnaserror`, 0 warnings / 0 errors):

- `Reserve()` temporarily regressed to the old release-after-discovery shape ⇒ the new test fails in **17 ms**: `Assert.Equal() Failure — Expected: AddressAlreadyInUse, Actual: null` (the rival took the port). No load needed.
- A standalone probe of the same regressed shape, with the rival *holding* the port, reproduces the issue's exact failure: `IOException: Failed to bind to address http://127.0.0.1:56067: address already in use.`
- With the fix: **20/20** iterations of the new test and **5/5** of the full 12-test project pass (68 s wall clock).

## Not in scope

Two other sites reserve-then-release the same way, and neither can use this cure:

- `test/MeshWeaver.Graph.Test/TestOgServer.cs:43` — `HttpListener` cannot be handed a socket;
- `test/MeshWeaver.Persistence.Test/StorageAndDb.cs:10` — hands the port to Azurite, an external process.

Both need a different fix; filing separately rather than widening this one. (`MeshGrpcLiveTest` already does the right thing — it binds `:0`, because it does not need the number up front.)

### DIFF
--- a/src/MeshWeaver.Documentation/Data/WhatsNew/2026-08-26-grpc-test-no-longer-loses-its-port.md
+++ b/src/MeshWeaver.Documentation/Data/WhatsNew/2026-08-26-grpc-test-no-longer-loses-its-port.md
@@ -1,0 +1,21 @@
+---
+Name: A gRPC test no longer loses its port to another process
+Category: Fix
+Description: A test that picked a free port and released it before use could red an unrelated pull request when anything else on the machine took that port first.
+Icon: Sparkle
+Order: -20260826
+---
+
+# A gRPC test no longer loses its port to another process
+
+The trusted-endpoint test needed two free network ports before it could start its server, and it
+found them the usual way: open a socket, ask the operating system which port it assigned, then close
+it again. Between that close and the server's own bind, the port belonged to nobody — so anything
+else on the machine could take it, and the server then refused to start with *address already in
+use*. On a quiet machine the gap is invisible; on a loaded build agent it is wide enough to catch,
+which is how the failure kept appearing on pull requests that had not touched the code at all.
+
+The test now keeps the socket it opened and hands that same socket to the server, so the port is
+owned continuously and there is no moment for anything else to claim it. A new check runs the losing
+interleaving on purpose — a rival process grabbing the port at exactly the wrong instant — and fails
+if the port is ever left unowned again.
--- a/test/MeshWeaver.Hosting.Grpc.Test/MeshGrpcTrustedKestrelTest.cs
+++ b/test/MeshWeaver.Hosting.Grpc.Test/MeshGrpcTrustedKestrelTest.cs
@@ -15,6 +15,7 @@ using MeshWeaver.Messaging;
 using Microsoft.AspNetCore.Builder;
 using Microsoft.AspNetCore.Hosting;
 using Microsoft.AspNetCore.Server.Kestrel.Core;
+using Microsoft.AspNetCore.Server.Kestrel.Transport.Sockets;
 using Microsoft.Extensions.Configuration;
 using Microsoft.Extensions.DependencyInjection;
 using Microsoft.Extensions.Logging;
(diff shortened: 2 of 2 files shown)


rbuergi:
Filed the two sibling sites named under **Not in scope** as #2436 — both are the same bug class, both need a *different* structural fix (`HttpListener` cannot be handed a socket; Azurite is an external process), so neither belongs in this PR.

Issue #2398, opened 2026-08-26
7 public types that left GitSync/Hosting for MeshWeaver.AI CANNOT be forwarded — move back, or republish atomically


rbuergi:
## 13 more public types have left their assembly since `v3.0.0-rc7` — same mechanism as #2370

#2370 was one instance of a class: **a public type that moves assembly (or namespace) breaks every
module compiled earlier**, because a module's IL binds `Namespace.Type` scoped to an *AssemblyRef*,
and the module lane's gate is a semver floor that cannot see a type at all.

The fix for #2370 restored the four `MeshOperations`-family names behind `[TypeForwardedTo]` and
added `scripts/check-type-forwards.py`, which refuses the next such move. Running that gate across
the window between the last release and today's `main` says how much of the same exposure is already
out there:

```
python3 scripts/check-type-forwards.py --base v3.0.0-rc7 --head origin/main
```

**17 public types moved without a forwarder. Four are fixed by #2370's PR. Thirteen remain:**

| type (old full name) | left | landed in |
|---|---|---|
| `MeshWeaver.AI.Connect.IMcpBackConnection` | `MeshWeaver.AI` | `MeshWeaver.Mesh.Contract` (`MeshWeaver.Mesh.Services.*`) |
| `MeshWeaver.AI.Connect.McpConnectionInfo` | `MeshWeaver.AI` | `MeshWeaver.Mesh.Contract` |
| `MeshWeaver.AI.IMasterKeyProvider` | `MeshWeaver.AI` | `MeshWeaver.Mesh.Contract` (`MeshWeaver.Mesh.Security.*`) |
| `MeshWeaver.AI.ConfigMasterKeyProvider` | `MeshWeaver.AI` | `MeshWeaver.Mesh.Contract` |
| `MeshWeaver.AI.IProviderKeyProtector` | `MeshWeaver.AI` | `MeshWeaver.Mesh.Contract` |
| `MeshWeaver.AI.ProviderKeyProtector` | `MeshWeaver.AI` | `MeshWeaver.Mesh.Contract` |
| `MeshWeaver.GitSync.AiContentDiskWriter` | `MeshWeaver.GitSync` | `MeshWeaver.AI` |
| `MeshWeaver.GitSync.AiContentSyncArea` | `MeshWeaver.GitSync` | `MeshWeaver.AI` |
[...]

PR #2428, opened 2026-08-26
Forward the six #2276 types that left MeshWeaver.AI without one (#2398)

rbuergi:
Fixes the forwardable part of #2398. **#2398 stays open**, narrowed to the two groups that cannot be forwarded at all — those need a maintainer decision, not a clever workaround.

## What this is

#2370 was one instance of a class, not a one-off. A module's IL binds `Namespace.Type` scoped to an **AssemblyRef**, and the module lane's gate is a **semver floor** that cannot see a type at all — so a public type that changes assembly or namespace breaks every module compiled earlier, at the next platform roll, with no warning anywhere. **No source build in either repo can observe it**: building the plugin's source against the branch answers *source* compatibility, which is a different question. That is precisely why #2283 shipped a `TypeLoadException` on every MCP tool call having been verified exactly that way.

Replaying #2370's gate across `v3.0.0-rc7 → main` found **17** unguarded moves. #2370 fixed four. This fixes **six** more — the ones #2276 made when it moved the credential-protection and MCP-back-connection contracts into `MeshWeaver.Mesh.Contract`:

| type (frozen full name) | left | now in |
|---|---|---|
| `MeshWeaver.AI.IMasterKeyProvider` | `MeshWeaver.AI` | `MeshWeaver.Mesh.Contract` |
| `MeshWeaver.AI.ConfigMasterKeyProvider` | `MeshWeaver.AI` | `MeshWeaver.Mesh.Contract` |
| `MeshWeaver.AI.IProviderKeyProtector` | `MeshWeaver.AI` | `MeshWeaver.Mesh.Contract` |
| `MeshWeaver.AI.ProviderKeyProtector` | `MeshWeaver.AI` | `MeshWeaver.Mesh.Contract` |
| `MeshWeaver.AI.Connect.IMcpBackConnection` | `MeshWeaver.AI` | `MeshWeaver.Mesh.Contract` |
| `MeshWeaver.AI.Connect.McpConnectionInfo` | `MeshWeaver.AI` | `MeshWeaver.Mesh.Contract` |

**Three have a proven module consumer** on `Systemorph/MeshWeaver.Plugins@main` — confirmed by grepping the repo, not assumed: `IMcpBackConnection` + `McpConnectionInfo` in `MeshWeaver.AI.ClaudeCode` / `MeshWeaver.AI.Copilot`, `IProviderKeyProtector` in `MeshWeaver.Blazor.Portal/Chat/ThreadChatView.razor.cs` and `Memex.Portal.Gui/MemexPortalComposition.cs`.

Their *source* was updated in the same wave, so `landed-modules-gate` is green and a **rebuilt** bundle is correct. What that says nothing about is a deployment still running a bundle published **before** the wave: it holds the old TypeRef and dies exactly like #2370 the moment its platform rolls. **A forwarder makes that ordering irrelevant** — which is the whole point, since the ordering is not something we control.

## The two non-obvious constraints, both already load-bearing here

**A forwarder cannot rename.** So each type keeps its **original** full name in its new home: five files under `src/MeshWeaver.Mesh.Contract` now declare `namespace MeshWeaver.AI` / `MeshWeaver.AI.Connect`. That mismatch is the contract, not a leftover — the same shape `MeshOperations` already has in `MeshWeaver.Mesh.Operations`. Each file carries a comment saying so, because a "tidy-up" of exactly that broke `main` today. Consumers gain a `using`.

**No shim types.** A forwarder keeps **one** type identity via the CLR's ExportedType table. A shim class left behind in `MeshWeaver.AI` would satisfy a naive resolution check while minting a *second* identity — reintroducing the `is`/`as` trap-door that reads as a silent null. `MovedTypeBinaryContractTest.TheForwardersAreForwarders_NotDuplicatedTypes` reads the ExportedType table specifically to refuse that.

## Proven, not asserted

- **The gate.** `python3 scripts/check-type-forwards.py --base v3.0.0-rc7` reports **13** findings without the forwarders — exactly the list in #2398 — and **7** with them, all of which are the cycle-blocked group below. Against `origin/main` (what CI runs) it is green.
- **The runtime test.** `MovedTypeBinaryContractTest` extended from 4 frozen names to 10. With the six forwarder lines deleted it fails **7/12**; with them it passes **12/12**. The frozen list became a `name → hosting assembly` map, because the two waves landed in different assemblies.
- Touched projects build clean with CI's flags (`-c Release -warnaserror`, one project per invocation): `Mesh.Contract`, `AI`, `PluginCatalog`, `Memex.Portal.Shared`, `AI.Test`, `GitSync.Test`, `Graph.Test`, `Documentation.Test`. Tests run: 40 in `AI.Test` (forwarder + provider-key + credential-seed), `ProviderKeyProtectorRegistrationTest`, `WhatsNewEntryIntegrityTest`, `DocumentationLinkIntegrityTest` — all green.

## Two gate defects found while doing it

**A commented-out forwarder counted.** `check-type-forwards.py`'s `FORWARD_RE` ran over raw file text, so `// [assembly: TypeForwardedTo(typeof(X))]` satisfied the gate. That is the gate's own **false-PASS** shape, and the likeliest way a forwarder gets disabled in a refactor (comment first, delete later) — it is literally what made the first negative control here look green. `parse_forwards` now truncates each line at `//`. Three self-test rows pin it, including that a **real** forwarder with a trailing comment still counts (without that row, the fix would look correct while silencing live forwarders). 20 → 23 self-tests, all passing. Line comments only; a `/* … */`-buried forwarder would still count, and the docstring says so rather than implying coverage that is not there.

**`check-record-signatures.py`'s step anchored on the base branch TIP** while the script's own docstring says merge base (recorded on #2398). The half that was wrong is *not* the file list — the three-dot `git diff --name-only {base}...HEAD` already resolves the merge base. It is the **before content**: `git show {base}:{file}` reads the literal ref, so a record whose primary constructor changes on main while a PR is open reads as *that PR reverting it* — a red on somebody else's commit. Now passes the merge-base SHA, matching what the type-forwards step already did.

## What is NOT fixed, and why it is a decision rather than a task

Two groups **cannot be forwarded at all**. The forwarder must live in the assembly being *left*, so that assembly must reference the type's new home — and `MeshWeaver.AI` already references both of these, so the forwarder would be a **reference cycle**:

- `MeshWeaver.GitSync → MeshWeaver.AI`: `AiContentDiskWriter`, `AiContentSyncArea`, `AiContentSyncMenuProvider`, `AiContentSyncResult`, `HubAiContentSyncExtensions`, `PullRequestDraftService`
- `MeshWeaver.Hosting → MeshWeaver.AI`: `MeshWeaver.Hosting.Persistence.Parsers.AgentFileParser`

**Exposure, confirmed rather than assumed:** zero consumers of those seven names in `MeshWeaver.Plugins`, `MeshWeaver.Reinsurance` or `MeshWeaver.SocialMedia`. So it does look core-only — but "no consumer in the repos I can see" is not the same as "no published bundle holds the TypeRef", which is the question that actually matters and the one #2370 answered the expensive way.

The two honest options, both of them a maintainer call: **move the types back**, or accept the break and do the **atomic republish** (#2234) — rebuild and republish every affected bundle, then roll the image, so no deployment is ever running an old bundle against a new platform — recording it in `scripts/type-forwards.allow`. Full detail on #2398.

~~Also worth a separate decision, unchanged here on purpose: the *Public surface (binary compatibility)* job is **not** in `collect-results`' `needs`.~~ **Superseded while this PR was open** — #2407 (`5ce64c984`) landed on `main` and wired `record-signatures` into `collect-results`' `needs` plus the explicit fail step, so both binary-compat gates can now block a merge. Nothing to do here; this branch predates it and the two edits touch different hunks of `dotnet-test.yml`, so they merge cleanly (`mergeStateStatus` is `BLOCKED` on checks, not `DIRTY`).

### DIFF
--- a/.github/workflows/dotnet-test.yml
+++ b/.github/workflows/dotnet-test.yml
@@ -227,8 +227,22 @@ jobs:
     - name: "Refuse a binary-breaking primary-constructor change"
       run: |
         set -euo pipefail
-        base="origin/${{ github.base_ref }}"
         git fetch --no-tags --depth=200 origin "${{ github.base_ref }}"
+        # 🚨 The MERGE BASE, not the base branch tip — same reasoning as the type-move gate below.
+        # This step disagreed with its own script until #2398: check-record-signatures.py's
+        # docstring says it compares "the merge base against the working tree", while the step
+        # handed it `origin/<base_ref>`.
+        #
+        # The half that was wrong is NOT the file list — that uses the three-dot
+        # `git diff --name-only {base}...HEAD`, which resolves the merge base itself. It is the
+        # BEFORE CONTENT: `git show {base}:{file}` reads the literal ref, so with the tip, a record
+        # whose primary constructor changed on main while this PR is open gets compared against
+        # THAT change rather than against what this PR branched from — and the PR reads as
+        # reverting it. A red on somebody else's commit, on a gate whose whole value is that people
+        # believe it. Passing the merge-base SHA fixes both halves consistently (`sha...HEAD`
+        # degenerates to `sha`, since it is an ancestor of HEAD).
+        base=$(git merge-base "origin/${{ github.base_ref }}" HEAD)
+        echo "Comparing against merge base $base"
         python3 scripts/check-record-signatures.py --base "$base"
     # The OTHER half of the same class (#2370): a public TYPE moving assemblies, which no repo-local
     # build can see. `landed-modules-gate` compiles the plugins repo's module SOURCE against this PR
--- a/memex/Memex.Portal.Shared/Authentication/EaGraphAuth.cs
+++ b/memex/Memex.Portal.Shared/Authentication/EaGraphAuth.cs
@@ -1,3 +1,4 @@
+using MeshWeaver.AI;   // IProviderKeyProtector & co keep their ORIGINAL namespace in MeshWeaver.Mesh.Contract (#2398 forwarders)
 using MeshWeaver.Mesh.Security;
 using System.Reactive.Linq;
 using System.Reactive.Threading.Tasks;
--- a/memex/Memex.Portal.Shared/Authentication/McpBackConnectionService.cs
+++ b/memex/Memex.Portal.Shared/Authentication/McpBackConnectionService.cs
@@ -1,3 +1,4 @@
+using MeshWeaver.AI.Connect;   // IMcpBackConnection/McpConnectionInfo keep their ORIGINAL namespace in MeshWeaver.Mesh.Contract (#2398 forwarders)
 using System.Collections.Concurrent;
 using System.Reactive.Linq;
 using MeshWeaver.Mesh.Services;
--- a/scripts/check-type-forwards.py
+++ b/scripts/check-type-forwards.py
@@ -170,7 +170,23 @@ def parse_declarations(path: str, text: str) -> list[Decl]:
 
 
 def parse_forwards(text: str) -> set[str]:
-    return {m.group("type") for m in FORWARD_RE.finditer(text)}
+    # 🚨 A COMMENTED-OUT forwarder must not count. FORWARD_RE is a plain regex over the file text,
+    # so before this line `// [assembly: TypeForwardedTo(typeof(X))]` satisfied the gate — the exact
+    # false-PASS shape this gate exists to prevent, and the likeliest way a forwarder gets disabled
+    # during a refactor (comment first, delete later). Found by running the gate's own negative
+    # control while fixing #2398: commenting the six new lines out left it GREEN, deleting them
+    # turned it red, so "the forwarder is disabled" and "the forwarder is there" read identically.
+    #
+    # Line comments only. A forwarder buried in a /* ... */ block would still count; closing that
+    # needs a real C# tokenizer rather than a regex, which is out of proportion for a CI script and
+    # a far less likely way to disable one line. Truncating at `//` can also cut a string literal
+    # containing a URL; harmless here, because the two regexes match declarations and assembly
+    # attributes, neither of which lives inside a string.
+    return {
+        m.group("type")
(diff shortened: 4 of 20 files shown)


rbuergi:
🔗 **Cross-repo half: Systemorph/MeshWeaver.Plugins#718 must merge FIRST.**

`landed-modules-gate` builds every `plugins-repo/src/*/*.csproj`, and one line there names the moved type **fully qualified** — the single shape no `using` can rescue:

```csharp
// src/Memex.Portal.Gui/MemexPortalComposition.cs:291
services.AddSingleton<MeshWeaver.Mesh.Services.IMcpBackConnection, McpBackConnectionService>();
```

So this PR will go red on that gate until #718 lands. It was the **only** such reference — swept every `*.cs`/`*.razor`/`*.json` in that repo including the in-mesh node trees. Everything else already names these types unqualified with both usings present (`MeshWeaver.AI.ClaudeCode`, `MeshWeaver.AI.Copilot`, `MeshWeaver.Blazor.Portal`, and the in-mesh `Providers/ProvidersApp/Source/*.cs` since plugins#685).

**#718 is tolerant of both platforms, so plugins-first is the safer order here rather than the usual platform-first**: `MeshWeaver.AI.Connect` already exists on the live image, so #718 compiles against today's platform *and* against this one. Going first closes the window in which the plugins repo's `main` is broken against core's; the reverse order would open one. Verified by building `src/Memex.Portal.Gui` at `-c Release -warnaserror` against this branch's tree.

One more datapoint from the sweep, in the reassuring direction: the private `Systemorph/Memex` repo already names these types by their **pre-#2276** spelling — `using MeshWeaver.AI.Connect;` and `MeshWeaver.AI.Connect.IMcpBackConnection` — so this PR restores what that repo expects rather than breaking it.

PR #2430, opened 2026-08-26
Publish NodeType bakes per architecture (arm64 + amd64), with the guard that makes it safe

rbuergi:
## What

Publish NodeType bakes **per architecture** — one lane each for `linux-x64` and `linux-arm64` — plus the guard that makes publishing two lanes safe at all.

Until now only amd64 was ever baked, so every arm64 install (memex-local, any Apple Silicon host) resolved the other framework identity, found nothing, and **compiled every NodeType at boot**.

## Everything it needs already exists

| precondition | evidence |
|---|---|
| arm64 CI runners | repo is **public** ⇒ `ubuntu-24.04-arm` is free (`gh api …/repos/Systemorph/MeshWeaver --jq .visibility`) |
| an arm64 bake image | `mw-plugin-test` publishes `ContainerRuntimeIdentifiers='"linux-x64;linux-arm64"'` |
| multi-arch portal images | `check-image-set.sh` already asserts `linux/amd64 + linux/arm64` |
| an architecture-sensitive identity | the bake host ships a surface manifest ⇒ resolves `s<hash>`, not `g<sha>` |

So nothing new gets built; the second leg just runs where its bytes belong.

## The guard, which had to come first

A bake publishes to `…/prebuilt-bundles/<framework-identity>/<source-name>/`, and the sealed-skip keys on **content × framework**. Neither carries the architecture — and the identity only *sometimes* does. `FrameworkBuildIdentity` resolves **surface `s<hash>` → commit `g<sha>` → MVID set**; the first differs per architecture, `g<sha>` does not.

Under a `g<sha>` identity a second architecture fails two ways, both silent: same source ⇒ the sealed-skip says *"already published"* and the second lane ships **nothing** while its pods adopt the other's bytes; different source ⇒ it **unseals and overwrites**. Adopting bytes from an identity you did not resolve is exactly what `CiContentBake.md` forbids, and it lands as a `TypeLoadException` inside a collectible ALC at activation — no overlay, nothing to grep.

`publish-bake-bundles.sh` now records the producing architecture (`architecture.txt`) and **refuses** a mismatch rather than skipping or overwriting. Adding the lane without this does not produce two lanes — it corrupts one.

### Both review findings fixed

- **Fail-open probe** — reading the marker with `download … || echo ""` made every transient/auth/CLI error indistinguishable from "no marker", the one answer that lets a publish through. Existence is established first; an unusable answer, an unreadable marker, or an empty one all refuse.
- **A deadlock between the two rules** — an unmarked incumbent made arm64 refuse and point at the amd64 lane to stamp it, but that lane hits the sealed-skip and `continue`s, never stamping. amd64 now backfills the marker **before** the skip, so the remediation is actually reachable.

## Verification

Offline against a stubbed `az`, **real exit codes** (not a piped `tail`, which reports the pager's status):

```
marked x64        + linux-x64     exit 0   normal skip — the control, no false trip
marked x64        + linux-arm64   exit 1   cross-architecture refusal
probe unusable    + linux-arm64   exit 1   fail closed
marker unreadable + linux-arm64   exit 1   fail closed
unmarked          + linux-x64     exit 0   stamps the marker
unmarked          + linux-arm64   exit 1   refuses, guidance now reachable
```

And the amd64 leg resolves **exactly as before**: `ubuntu-latest`, `--platform linux/amd64`, `BAKE_ARCHITECTURE=linux-x64` (the script's own default), `continue-on-error: false`, zero hard-coded amd64 pins left in the job.

## 🚨 arm64 ships OFF, behind `BAKE_ARM64`

`delivery-verdict` reads `needs.publish-bake.result` — for a matrix, the **aggregate**. Whether a `continue-on-error` leg reports `success` *there* is a GitHub semantic I could not verify from here, and guessing wrong would red-line delivery of the architecture the entire fleet runs.

So the default leaves this workflow behaviourally identical to today (a skipped leg cannot fail an aggregate), and enabling arm64 is one deliberate repo-variable flip whose first run someone watches. This is a feature flag on a **new publication lane**, not a gate skipping its own inputs: with it off nothing claims arm64 is covered, and the boot compile remains the honest signal. Delete the flag once a green arm64 publication has been observed.


rbuergi:
## Two defects fixed, and the second one was the serious one

### 1. The guard demanded the literal this PR removed (the shard-1 red)

`PlatformBakeLaneGuard.PlatformBake_RunsInsideTheShippedImage` asserted the job text contains `--platform linux/amd64`. Splitting the lane replaced every such literal with `--platform ${{ matrix.docker_platform }}`, so the assertion had become the *opposite* of the invariant — it now demanded the arm64 leg pull amd64 bytes.

The invariant was never "amd64", it is **pinned explicitly, never inherited from the runner**, so the check is now structural in two independently-falsifiable halves: the lanes must name both platforms as literals, and every `docker run` / `pull` / `create` in `publish-bake` must carry the per-leg pin. Both were negative-tested against a real regression, not assumed.

### 2. 🚨 The arm64 flag made `main-cd.yml` an INVALID WORKFLOW — merged, CD would never have started

```yaml
if: … && (matrix.arch == 'linux-x64' || vars.BAKE_ARM64 == 'true')
```

`matrix` is **not an available context in `jobs.<id>.if`** — only `github`, `inputs`, `needs`, `vars` are. GitHub does not evaluate that false; it rejects the whole file with a **startup failure**.

The branch was already showing it and it reads as nothing: run [33007848017](https://github.com/Systemorph/MeshWeaver/actions/runs/33007848017), path `.github/workflows/main-cd.yml`, conclusion `failure`, **zero jobs, zero steps, no log** — a red suite named after the workflow with nothing inside it. `main-cd.yml` on `main` lints clean, so it is this diff. `actionlint` names it exactly:

```
main-cd.yml:1079:87: context "matrix" is not allowed here.
  available contexts are "github", "inputs", "needs", "vars"
```

On merge this would not have been "the arm64 leg is off". It would have been Continuous Delivery refusing to start: no portal image, no migration image, no bake, **nothing published**, on every commit until someone reverted it.

**The flag is kept** — the reason for it (not guessing how a `continue-on-error` leg reports into an aggregate `needs.publish-bake.result`) is still right. It is now applied where a matrix can legally be filtered: when it is **built**. `gate` — which already owns every other publish decision here and was already a dependency — emits `bake_arches`, and `publish-bake` consumes `include: ${{ fromJSON(needs.gate.outputs.bake_arches) }}`.

That is also strictly better than a skip: **a disabled leg does not exist**, so it cannot contribute a grey tick to an aggregate that something downstream reads as evidence. (A `matrix.arch` list plus a static `include` would *not* have worked — an include entry whose value for an original matrix key matches no combination **adds** a combination rather than being dropped.)

### The control, because nothing here could have caught it

There is no `actionlint` step in CI and no test that reads workflow context usage — this defect class had **no control at all**, and its symptom is a red that looks like any other red. `WorkflowJobIfContextGuard` now fails on any job-level `if:` reading `matrix`, in any workflow, **block scalars included** — the continuation-line form is precisely the one that shipped.

Verified by reinstating the exact defect: the guard fails naming `main-cd.yml:1079` and quoting the folded expression, and actionlint agrees at the same line.

### Also

The two `CiContentBake.md` paragraphs still stating the bake is pinned to `--platform linux/amd64` (§ identity, § the portal-identity comparison) now describe the per-lane pin, and the comment above the identity check no longer says "the bake above already pins amd64" — it pins this leg's architecture, which is what makes the two sides of that comparison describe the same bytes.

**Verification:** 70/70 in `MeshWeaver.Documentation.Test` (fresh `.trx`), `actionlint` clean on `main-cd.yml`, and each new assertion negative-tested against a real regression rather than assumed.

rbuergi:
### Empirical confirmation of the fix

Every push of this branch while the invalid `if:` was present minted a startup-failure run of `main-cd.yml` — no jobs, no steps, just a red suite:

| pushed SHA | main-cd run | conclusion |
|---|---|---|
| `f0a72032` | [33007564272](https://github.com/Systemorph/MeshWeaver/actions/runs/33007564272) | failure (0 jobs) |
| `2b56aad7` | [33007848017](https://github.com/Systemorph/MeshWeaver/actions/runs/33007848017) | failure (0 jobs) |
| `7e1e8050` | [33015316652](https://github.com/Systemorph/MeshWeaver/actions/runs/33015316652) | failure (0 jobs) |
| **`676cd008`** | **none** | — |

The push of `676cd008` produced **no such run at all**, because there is no longer an invalid workflow for GitHub to reject. `main-cd.yml` has no `push` trigger (only `workflow_run`, `schedule`, `workflow_dispatch`), so a run appearing on a feature-branch push was never anything but the startup failure — which is exactly why it was easy to read past.

PR #2442, opened 2026-08-26
refactor: delete AsyncDisposeQueue — nothing ever enqueued anything

rbuergi:
## The finding

`AsyncDisposeQueue` exists for one purpose, stated in its own summary:

> resources **Enqueue** their async cleanup onto this mesh-scoped queue instead of running or leaking it

**No resource ever did.** There is not one call to `Enqueue` anywhere in `src/` — only its own tests construct a queue and post to it.

```
$ grep -rn "AsyncDisposeQueue[^;]*\.Enqueue\|asyncDisposeQueue\??\.Enqueue" src/
(nothing)
```

In production it is resolved in three teardown paths and **drained empty**.

## Why that is worse than merely unused

The cancel-then-join, the bounded quiesce budget, the *"no async continuation runs against a disposed scope"* guarantee — all of it is a barrier over nothing. And it **reads as protection**: comments and test messages across teardown cite it as one of the phases standing between a live continuation and an unloading node ALC, i.e. the use-after-unload SIGSEGV.

That is the same shape as a CI gate that skips and still shows a tick. The appearance of a guarantee is the harm, because it stops anyone from looking for the real one.

## Why the concept goes rather than getting an `IObservable` API

Disposal in MeshWeaver is reactive and **synchronous**, and stays that way. A hub tears itself down through the ShutDown phase, cancels and joins its offloaded I/O through `IoPoolRegistry.DrainAll`, and — as of #2438 — disposes the container it owns. None of that needs a Task-based background queue carrying its own cancellation token, and having one on the teardown path invites exactly the async-inside-teardown shape that *wedges* a dispose instead of finishing it.

## Removed

- the class and its test file
- the DI registration
- the drain phase in `MeshTeardownExtensions` (three phases → two)
- `TeardownReport.AsyncDisposeClean` — a teardown report can no longer be "dirty" for a reason that was never real
- comments across five files that cited the queue as a teardown phase, now describing the two that actually run

`TeardownReport` keeps `LeakedIoLeaves`, which **is** measured and does gate: a pooled I/O leaf still running when the scope closes is live code surviving teardown, and that assertion is untouched.

## Verified

- every touched project builds
- `MeshWeaver.Hosting.Test` **235/235** — including the teardown-ordering and signal tests
- `MeshWeaver.Messaging.Hub.Test` **280/280**

Stacked conceptually on #2438 (the hub now disposing its own container) but touching disjoint files.


rbuergi:
## The binary-compat gate was right, and its suggested remedy was not the right one

The gate refused this correctly:

```
record TeardownReport — arity
    primary constructor went from 2 to 1 parameter(s)
    before: (int LeakedIoLeaves, bool AsyncDisposeClean)
    after:  (int LeakedIoLeaves)
```

That is the failure AGENTS.md documents as **aborting a host at boot**: a record's primary constructor *replaces* its signature rather than extending it, so any consumer compiled against the old arity throws `MissingMethodException` — in both directions, with no image able to serve a mixed set of module builds.

**I did not take the allow-file route the gate offers.** That file's own header defines what an entry means:

> An entry is a **STATEMENT THAT THE ATOMIC MOVE IS PLANNED** — every bundle rebuilt and republished for the new signature, then the image rolled — not a way to make the gate quiet.

A fleet-wide republish to delete one always-`true` boolean is wildly out of proportion, and it is a system-changing act that is not mine to trigger.

## What changed instead

`TeardownReport` keeps its shape. `AsyncDisposeClean` stays, documented as exactly what it now is — residue that is always `true`, retained only because dropping it breaks the arity — with the removal condition written down so it can ride a planned republish later. Its two construction sites pass `true` with an inline note rather than silently.

I also **restored the assertion I had deleted**. While the field exists, `Clean` must still honour it: a report that ignored one of its own inputs is the next silent green, which is the shape this whole change set exists to remove.

Everything else stands — the class, its tests, the DI registration and the drain phase are still gone.

Build clean · `MeshWeaver.Hosting.Test` **235/235**.

rbuergi:
## Closing: the premise is false as of 18 minutes before this branch was written

This PR rests on one claim — *"There is not one call to `Enqueue` anywhere in `src/`"*. That was true of the tree this branch was cut from, and false of `main` by the time the branch existed.

**#2334 — "Stream teardown is enqueued, not fire-and-forget — the exit-139 shape" — merged at `19:45:52Z` today.** This branch's first commit is dated `20:03:25Z`, and its merge-base is `23eeebd0b` (#2407), which predates #2334. So the grep in the description was run against a base `main` had already moved off.

`main` today, code only:

```
src/MeshWeaver.Connection.Orleans/OrleansRoutingService.cs:734:
    asyncDisposeQueue.Enqueue(_ => teardown.DefaultIfEmpty(Unit.Default).LastAsync().ToTask());
```

CI said the same thing louder — `Build solution (once)` on the merge commit:

```
src/MeshWeaver.Connection.Orleans/OrleansRoutingService.cs(79,22): error CS0246:
  The type or namespace name 'AsyncDisposeQueue' could not be found
```

That is not a stale reference this PR forgot to sweep. It is a **deliberate, load-bearing caller added as a fix**, and #2334's own comment at the call site says exactly why the queue and not a pool subscription:

> 🚨 **ENQUEUED, not fire-and-forget.** This used to be `ioPool.Invoke(...).Subscribe(_ => {})` — the handle dropped on the floor — so `Dispose()` reported "torn down" while the Orleans unsubscribe was still running on a pooled thread. […] `DisposalCompleted` covers a hub's action blocks and message round-trips but NOT mesh-shared pooled I/O leaves, so `UnloadContextIfSafe` could unload this hub's collectible context while this leaf was still executing types from it. A leaf touching unloaded code is a native SIGSEGV (exit 139), not an exception anything can catch.
>
> `AsyncDisposeQueue` exists for exactly this shape […] **Enqueuing is what makes the drain WAIT for this work instead of racing it.**

So the two things this PR removes are the two things #2334 needs:

- the queue itself — the only teardown phase that *joins* genuinely-async cleanup, and
- the drain phase in `MeshTeardownExtensions` (three phases → two) — the join that makes an enqueued unsubscribe finish before the DI scope and the collectible ALC go away.

Merging this would silently reinstate the use-after-unload window #2334 closed ~two hours ago. The PR's own framing applies to itself with the sign flipped: it argued the queue was *the appearance of a guarantee*; as of #2334 it is the guarantee, and deleting it would leave the appearance of a teardown that has finished.

Nothing here needs porting forward — `TeardownReport.LeakedIoLeaves` and the two surviving phases are untouched on `main`. Closing rather than rebasing, because after the rebase there is no change left to make.

*(Verified in an isolated worktree at `65bb6034f`; the falsifying commit is `17c8a0c73` / #2334.)*

rbuergi:
Superseded by #2334 landing 18 minutes before this branch was written — the queue now has a live, deliberate caller. Details in the comment above.

PR #2452, opened 2026-08-26
fix(cd): the reconciler measured images, so the bake could never run again

rbuergi:
## Policy: always bake on CI, never bake on a silo. It was neither.

The delivery reconciler decides "nothing to reconcile" from the **image set alone**:

```bash
if [ "$COMPLETE" = "true" ]; then
  decision false "main $SHORT already has a complete image set"
```

But the bake — the thing that stops every pod Roslyn-compiling shipped content at boot — is a **consequence of promoting an image**. So once the image set was complete, CD was permanently satisfied and the bake could never run again.

**Measured 2026-08-26: no successful platform bake in the last 60 CD runs.** Every one skipped, every run green.

## It was already written down

This job's own ACR-retry comment describes the hole exactly, and was never closed:

> the release ships **WITHOUT its content bake**, and the CD reconciler **keys on the IMAGE set (complete) rather than on the publication, so nothing re-attempts it** … until then pods compile shipped content at boot exactly as they did before the bake lane existed.

## The change

A reconcile whose images are complete now still re-asserts the **content bake** — never an image build. New `bake_only` gate output; `publish-bake` gated on it as well as on the promote path.

- **`always()` is required**, because `promote` is *skipped* on that path. Without it the job inherits the skip and the fix is inert — which is the exact shape being fixed.
- **It bakes in the PROMOTED image**, not this run's staging tag: a run that built nothing has no staging tag.

**Re-publication stays honest rather than hourly.** `publish-bake-bundles.sh` already skips a sealed directory whose content × framework identity match, so a reconcile resolving the same identity uploads nothing. Bundles are re-published only when genuinely absent for the identity the running images resolve.

## 🚨 What is deliberately NOT on this path

The version-keyed **plugin bundle** publish. `RELEASE_VERSION` is `3.0.0-ci.<build>` — it embeds the run number, so a reconcile would mint a *different* version from the one the promoted image carries and publish bytes under a version no run produced. That is exactly what the existing guard refuses:

```
[ -n "$RELEASE_VERSION" ] || refusing to publish plugin bundles under an unknown version (Plugins#695)
```

Weakening that to make the reconcile look "complete" would trade one silent wrong for another. The platform content bake has no such coupling — its key is the framework identity, which the promoted image resolves identically however often it is asked.

## What this does and does not fix

It restores the platform's own content bake, so silos stop recompiling ~80 shipped types at every boot (`compiled=80 alreadyBaked=4`, 64.8 s warm-up).

It does **not** by itself prove the course CTA stops recompiling: `Store/Plugin` is owned by MeshWeaver.Plugins and published by that repo's own bake, which is green and publishes under `s306f5384964b0bb5e58f2242367651fb`. Whether the running pods resolve that same identity still needs checking against the share — I could not reach the cluster (private endpoint, DNS unresolvable from my shell) to confirm it.


rbuergi:
## The bake ran for the first time tonight — and failed on its publish step

CD run [33024861072](https://github.com/Systemorph/MeshWeaver/actions/runs/33024861072) (commit `139fde8`) is the first run in at least the last 25 where `Bake platform content…` was **not skipped** — consistent with this issue's measurement of *"no successful platform bake in the last 60 CD runs"*. It got a real publish because main finally went green after ~4 hours of starvation.

The bake itself succeeded. The **publish step** failed:

```
failed step: Publish the plugin bundles to every portal target (same identity, same version)
::error::no release version from portal-image — refusing to publish plugin bundles
         under an unknown version (Plugins#695).
```

i.e. `${{ needs.portal-image.outputs.version }}` evaluated **empty** in `publish-bake`.

**What is verifiably fine:**

- `portal-image` ran and succeeded; its `Compute image version` step succeeded and logged `Computed image version: 3.0.0-rc8.ci.6007`;
- that value is correct — it is the tag that actually got promoted;
- the step writes it properly (`echo "version=$VERSION" >> "$GITHUB_OUTPUT"`, `id: vars`), and the job declares `outputs.version: ${{ steps.vars.outputs.version }}`;
- `publish-bake` still carries `needs: [gate, plugin-test-image, portal-image, promote]`.

So producer, output declaration and consumer wiring all look correct, and the consumer still saw empty. **I do not have a mechanism for that yet** and am not going to guess at one.

**Why this matters for this issue:** it is evidence that the bake lane has rotted while it was never running — which is the hazard a permanently-skipped job creates. Fixing the reconciler so the bake runs again (this issue) will surface whatever else decayed in the same period, so the two fixes want to land together rather than the reconciler alone.

**Disclosure:** #2430 (merged earlier today) made `publish-bake` a per-architecture matrix. It did not touch `portal-image`, the `needs` list, or this step. But because the bake was skipped in every recent run, there is **no prior non-skipped execution to compare against**, so I cannot rule it out by evidence — only by inspection. Worth checking first if someone picks this up: a matrix consumer reading a non-matrix producer's outputs is the only thing that changed about this job.

The failed leg was `linux-x64`; the `linux-arm64` leg is gated OFF behind `BAKE_ARM64` and did not run.

Issue #2457, opened 2026-08-26
Build process: bake off CI onto a dedicated worker — git sync + compile, all repos, dependency-ordered cascade


rbuergi:
Design captured from a working session on 2026-08-26, after an incident where CI starvation left production ~11 hours stale with a crashlooping rollout.

## The shape

**CI's job shrinks to: build the plain image, push it, finish.** Everything else is built *from* that image on a dedicated bake worker.

```
CI            build plain image → push → promote          (fast, few runners)
bake worker   git sync → C# compile → bundles → disk      (CPU-heavy, off the CI pool)
envs          seed bundles matching their own identity
```

## 1. The bake never stands up a mesh

**Bake = git sync, then C# compile.** Standing up hubs to produce a bake crashes, and it is the wrong shape regardless — `#2064` already settled it: *"producing an assembly is a build step; the mesh's job is to CONSUME a bake, not to be the thing that makes one."*

Status: the platform's two lanes were split by `#2064`. The reusable node-repo lane was still fused; **#2456** converts it. That is the precondition for everything below — while the bake needs a mesh, it cannot leave a test harness.

## 2. A dedicated bake worker, not CI runners

An AKS node pool rather than a VM, because the write path collapses:

```
az aks nodepool add -g memex-aks-rg --cluster-name memexaks-cluster \
  --name bake --node-vm-size Standard_D48ads_v5 \
  --enable-cluster-autoscaler --min-count 0 --max-count 2 \
  --node-taints workload=bake:NoSchedule --mode User
```

- A bake pod **mounts the same PVC the portal mounts**, so publishing is a local write to `prebuilt-bundles/<identity>/<source>/` — no `az storage file upload`, no cross-account storage credentials, no `BAKE_PUBLISH_TARGETS` to provision.
- **min-count 0** means no node exists between bakes: ~$160/mo at ~2 h/day versus **$1,927/mo** for an always-on D48 VM.
- The taint keeps bakes off the `silos` pool, which runs the portal. Today user capacity is 3 × D16s_v5 (48 vCPU, autoscale 2–4) and all of it serves production — a bake there would compete with the portal.
- The migration `Job` in the chart is the pattern to copy (`restartPolicy: Never`, `ttlSecondsAfterFinished`).

Input contract: **the image, the git to sync, and which packages to run.** One Job per env namespace, each mounting its own claim.

## 3. All repos in one build process

[...]

Issue #2449, opened 2026-08-26
Per-module isolation (#2234) does not cover HOSTED SERVICES — one incompatible module still kills the portal


rbuergi:
## What happened (memex-cloud, 2026-08-26 ~20:38 UTC onward)

`memex.meshweaver.cloud` stopped rolling. 7 pods served on `ci.5670` while the new ReplicaSet on `ci.5876` sat in **CrashLoopBackOff, 12 restarts**:

```
fail: Microsoft.Extensions.Hosting.Internal.Host[11]
      Hosting failed to start
System.InvalidOperationException: An exception was thrown while activating
  λ:Microsoft.Extensions.Hosting.IHostedService[] -> MeshWeaver.AI.OpenAI.OpenAICompatibleModelSync
 ---> Autofac.Core.DependencyResolutionException: None of the constructors found on type
      'MeshWeaver.AI.OpenAI.OpenAICompatibleModelSync' can be invoked with the available services and parameters
```

Process aborted with **SIGABRT** (`signo 6` in the createdump trailer — a managed exception, not a segfault).

The crashing pod's own log names the module and proves the skew:

```
[ModuleLoad] MeshWeaver.AI.OpenAI ← /data/modules/MeshWeaver.AI.OpenAI@637569ba/... written=2026-08-26 20:38:28Z
```

One of a wave republished 20:37–21:10, built against a core **newer** than image `ci.5876` (≈ commit `4327030`, 19:01). The trigger was an atomicity violation — modules moved ahead of the image while core CD was starved — and that half is operational.

## The defect this issue is about

**`#2234`'s per-module isolation protects INSTALLATION, not ACTIVATION.**

`MeshBuilder.InstallAssemblies` isolates, per module:

1. `Assembly.LoadFrom` + attribute materialisation (`Nodes` / `AddressTypes` / `HubConfigurations` getters);
2. invoking `GlobalServiceConfigurations` against the live `IServiceCollection`;
3. the `BuilderConfigurations` fold.

Its comment states the goal exactly: *"One module that cannot install against THIS build must cost that module's contribution and nothing else. It used to cost the process."*

But **registering a service is not resolving it.** A module adding `OpenAICompatibleModelSync` as an `IHostedService` registers a *descriptor* — nothing throws. The constructor is invoked later, when the generic host resolves `IHostedService[]` during startup, **outside `InstallAssemblies` entirely**. One unsatisfiable constructor there aborts the whole host.

[...]

Issue #2412, opened 2026-08-26
No CI run builds the combination that lands: a merge burst ships semantic conflicts untested


rbuergi:
## The gap

**No CI run ever builds the combination that lands.** With `strict: false` (correct, and not what I am proposing to change) a PR is tested against the `main` it branched from. Merge several in quick succession and the tree that reaches `main` is a combination **no run has ever compiled** — while every individual PR was legitimately green.

## It shipped today

Five PRs merged in fifteen seconds at 17:05:04–17:05:11. Main went red at 17:26:

```
memex/Memex.Portal.Shared/Api/MeshApiEndpoints.cs(335):
error CS0246: The type or namespace name 'MeshOperations' could not be found
```

A **semantic** conflict, so git had nothing to report and no merge was `DIRTY`:

- **#2283** moved `MeshOperations` to namespace `MeshWeaver.Mesh` and updated that call site's `using`.
- **#2397** restored the ORIGINAL namespace `MeshWeaver.AI`, because published modules reach the type through `[assembly: TypeForwardedTo]` and **a forwarder cannot rename**.

Each is correct. Each compiles against its own base. Only together do they fail. Fixed in #2411 (one `using`), but the *detection* is the issue: it surfaced from a scheduled main build ~20 minutes later, and it hard-blocks CD, which refuses a red tip. Today that sat directly in front of the surface-manifest image.

## Why the existing machinery does not cover it

- **`strict: false` is deliberate** and worth keeping — requiring every branch to re-sync before merge costs a full CI cycle per PR and, with several in flight, most of the throughput.
- **The green-tree marker / skip-reuse** makes it likelier, not less: a merge that reuses an already-green tree does not re-compile the combination either.
- **#2407** (binary-compat gate now blocking) does **not** catch this. That gate is about an already-published module binding a name; this is two source changes that disagree. Same family — *a green solo run proves less than it looks* — different mechanism. Worth stating plainly so nobody assumes it is covered.

## Options, roughly by cost

1. **A post-merge tripwire on `main` that is fast and loud.** Main already builds; the gap is *latency and attribution*. A compile-only job on every push to `main`, alerting with the failing file and `git log -S <symbol>` already run, turns a 20-minute silent window into a named alert. Cheapest, does not slow any PR, does not prevent the break.
[...]

PR #2458, opened 2026-08-26
main groups by SHA — cancel-in-progress:false does not protect a QUEUED run

rbuergi:
## What #2447 fixed, and what it did not

#2447 exempted `main` from cancel-in-progress. Correct and necessary — **and it does not hold when runners are scarce.**

`cancel-in-progress: false` protects a run that is **already running**. GitHub still keeps only **one pending run per concurrency group**, and a newer run supersedes the older *pending* one regardless of that setting.

While runners are free this is invisible: a run starts within seconds, so "queued" barely exists. When the pool is saturated it is decisive.

## Observed today, after #2447 merged

The shared 60-job pool was saturated by a sibling repo (12 concurrent runs in MeshWeaver.Plugins). Main's runs never reached in-progress:

```
7faac22   pending → cancelled   (never ran — and this IS the #2447 merge commit)
9e427c0   pending → cancelled   (never ran)
667b74d   pending → cancelled
08d82de / 444c6a9 / fe12f91 / 71e3ec2   cancelled
```

Six consecutive main commits, no completed Build-and-Test between them.

## What that cost

`main-cd.yml` gates delivery on `Consolidate test results` succeeding **for that SHA**. With no completed run its gate job stayed `skipped`:

```
GATE_RESULT: skipped
gate=skipped reason= publish= complete= green= sha=
```

Every CD run then reported **success** having published nothing — every image job simply `skipped`. Meanwhile plugin modules kept republishing against a newer core, so `memex-cloud` ran an older image against newer landed modules and every new pod aborted at boot (`IHostedService[] -> OpenAICompatibleModelSync` — blast radius in #2449). Production served **~11-hour-old code** with a wedged rollout, and the first symptom was a user noticing.

## The change

```diff
- group: build-test-${{ … workflow_dispatch && run_id || github.ref }}
+ group: build-test-${{ … workflow_dispatch && run_id
+                      || github.ref == 'refs/heads/main' && github.sha
+                      || github.ref }}
```

Each main commit gets its own group, so nothing can supersede it — queued or running — and CD always has a completed run to gate on.

**PR runs are untouched**: they keep grouping by ref, where superseding is correct and is where #2316's ~28% runner saving comes from.

## Cost

During a burst, several main runs proceed in parallel instead of one surviving. That is the point: runner minutes are what buys a publishable image, and the alternative is what happened today.

Worth pairing with the underlying capacity problem (#2457): the bake is the heavy part of CI and belongs on a dedicated worker, not on the shared pool.

### DIFF
--- a/.github/workflows/dotnet-test.yml
+++ b/.github/workflows/dotnet-test.yml
@@ -157,7 +157,19 @@ concurrency:
   # folded (`>-`) scalar, continuation lines indented deeper than the first are preserved
   # literally rather than folded, so a "tidier" multi-line version silently changes the
   # string rather than just its layout.
-  group: build-test-${{ github.event_name == 'workflow_dispatch' && github.run_id || github.ref }}
+  # 🚨 `main` groups by SHA, not by ref (#2412 follow-up). `cancel-in-progress: false` below
+  # protects a run that is ALREADY RUNNING — it does not protect a QUEUED one: GitHub keeps only
+  # ONE pending run per group and supersedes the older pending run whatever this setting says. That
+  # distinction is invisible while runners are free (a run starts within seconds, so "queued" barely
+  # exists) and decisive when they are not: on 2026-08-26 the shared 60-job pool was saturated by a
+  # sibling repo, main's runs never reached in-progress, and each merge cancelled the previously
+  # PENDING one — including `7faac22`, the very commit that merged the never-cancel fix. Six
+  # consecutive main commits produced no completed run, so CD's gate stayed `skipped` and no image
+  # published for hours while production ran ~11-hour-old code.
+  # Grouping by SHA gives each main commit its own group, so nothing can supersede it, queued or
+  # running. PR runs keep grouping by ref — superseding there is correct and is where #2316's
+  # runner saving comes from.
+  group: build-test-${{ github.event_name == 'workflow_dispatch' && github.run_id || github.ref == 'refs/heads/main' && github.sha || github.ref }}
   # 🚨 NEVER cancel a run on `main` (#2412). Superseding is safe on a PR branch — the later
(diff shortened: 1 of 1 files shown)


rbuergi:
Closing in favour of #2457 (move the bake off the shared pool), and recording the gap on #2412 rather than fixing it this way.

**The mechanism I found is real:** `cancel-in-progress: false` protects a run that is already RUNNING; GitHub still keeps only ONE pending run per concurrency group and supersedes the older *pending* one regardless. While runners are free that is invisible — a run starts in seconds, so "queued" barely exists. Tonight the shared 60-job pool was saturated by a sibling repo, main's runs never reached in-progress, and each merge cancelled the previously PENDING one — including `7faac22`, the commit that merged the never-cancel fix itself, which went `pending → cancelled` without ever running. Seven main commits, no completed run, `gate=skipped` on every CD run.

**Why the fix is still wrong.** Grouping main by SHA puts back exactly the demand #2316 deliberately removed — *"≈1,199 runner-minutes, ~28% of ALL demand in that window"* — and runner capacity is the binding constraint that caused the starvation in the first place. Trading more demand for guaranteed completion, silently, against an explicit dated decision, is the wrong call. The right lever is removing the bake from the shared pool (#2457): the bake is the heavy part of CI and the part that does not need a runner at all.

Noted for #2412's record, since the comment in `dotnet-test.yml` states it slightly too strongly:

> *The newest run is NEVER the one cancelled — cancel-in-progress cancels the older members of the group. main's tip therefore always runs to completion.*

True when merges pause. Under sustained merges **plus** a saturated pool, the tip keeps being replaced before it can start, so no tip completes. The hourly reconciler cannot heal it either, because it gates on `Consolidate test results` being green for the commit it resolves — and there is no completed run to produce that. Worth a sentence in that comment so the next person does not have to re-derive it at 22:00.

Issue #2301, opened 2026-08-25
Flake: OrleansGrainTeardownStragglerTest — the activation never leaves the silo catalog (4 hits on 2026-08-25, all unrelated diffs)


rbuergi:
`MeshWeaver.Hosting.Orleans.Test.OrleansGrainTeardownStragglerTest.GrainLifetimeCallbacks_AfterActivationIsInvalid_AreGracefulNoOps_NotThrows`
has no issue of its own (only the umbrella #1384). It is currently one of the most frequent members
of the population, and it is accelerating.

## Occurrences — four, all on diffs that cannot reach Orleans grain teardown

From a sweep of the last 300 `MeshWeaver Build and Test` runs, reading `[FAIL]` lines out of **every
attempt** (a re-run-to-green hides the attempt that failed):

| run | when | branch |
|---|---|---|
| [32841922930](https://github.com/Systemorph/MeshWeaver/actions/runs/32841922930) | 2026-08-25 11:21 | — |
| [32883602310](https://github.com/Systemorph/MeshWeaver/actions/runs/32883602310) | 2026-08-25 18:24 | — |
| [32886635560](https://github.com/Systemorph/MeshWeaver/actions/runs/32886635560) | 2026-08-25 18:55 | — |
| [32905773749](https://github.com/Systemorph/MeshWeaver/actions/runs/32905773749) | 2026-08-25 23:09 | `fix/2281-listen-readiness-signal` (a PostgreSQL LISTEN readiness signal — it cannot reach a grain catalog) |

All four on 2026-08-25, all attempt 1. That is a rate worth a name of its own.

## What the failure says

```
System.TimeoutException : The operation has timed out.
  at OrleansGrainTeardownStragglerTest.GrainLifetimeCallbacks_AfterActivationIsInvalid_AreGracefulNoOps_NotThrows()
     OrleansGrainTeardownStragglerTest.cs(98)
```

Line 98 is the `.Timeout(30s)` on step 4 — **"wait for the activation to leave the silo catalog"**:

```csharp
deactivate!.Invoke();
await hub.DisposalCompleted…;               // this part completed — the log line after it prints
…
await Observable.Interval(100ms).StartWith(0L)
    .SelectMany(_ => mgmt.GetDetailedGrainStatistics().ToObservable())
    .Where(stats => stats.All(s => !string.Equals(s.GrainId.ToString(), grainId, …)))
    .FirstAsync()
[...]

PR #2476, opened 2026-08-27
fix(#2423): the test harness stops driving the mesh through the ROUTER

### LINKED ISSUE #2423: Test harness drives the mesh through the ROOT MESH HUB — two ROUTER_TRAFFIC errors per affected class (58 call sites)
Split out of #2421, where two `ROUTER_TRAFFIC` errors were logged during
`ScriptExecutionInUserHomeTest.Run_StreamsProgressTimely_NotBuffered` (run 32999531439, attempt 1,
job 98280485356):

```
ROUTER_TRAFFIC: RawJson has the mesh hub as sender
  (sender: mesh/gQFBIAYIq0y5H9VgPWusgQ, target: rbuergi/demo-fd170c42…)
ROUTER_TRAFFIC: ExecuteScriptResponse has the mesh hub as target
  (sender: rbuergi/demo-fd170c42…, target: mesh/gQFBIAYIq0y5H9VgPWusgQ)
```

**Not the cause of #2421** — established by measurement, not by elimination: the same two lines
appear in local runs where every progress write reached the subscriber as its own emission within
1–3 ms of its timestamp. Recorded here so it is not lost with that issue's close.

## What it is

`MonolithMeshTestBase.Mesh` **is the root mesh hub** (`ServiceProvider.GetRequiredService<IMessageHub>()`,
MonolithMeshTestBase.cs:918). Tests use it as their request client:

```csharp
var execMessage = (await Mesh.Observe(
    new ExecuteScriptRequest(), o => o.WithTarget(new Address(codePath)))
    .Should().Within(60.Seconds()).Emit()).Message;
```

So the router really is the request's sender and the response's target, which is exactly what
`RouterTrafficRule.RoleOf` reports — and the second line is the one its own doc calls out: *"A
response ADDRESSED AT the router still reports: it proves the router ISSUED a request, which is the
violation the issuing seam (`NodeOperationIssuingHub`) exists to remove."*

The production callers of `ExecuteScriptRequest` look fine by comparison — `MeshOperations.ExecuteScript`
(MCP), `CodeLayoutAreas`, `EventSubscriptionRunner`, `ScriptDispatch` all post from a non-router hub.
This is the **test harness** violating the never-root-mesh-hub rule.

## Why it is worth fixing rather than muting

1. **Fidelity** — a test that drives the mesh through the router is not exercising the shape any
   production caller uses, and it loads the router's single action block with work while the same
   test measures latency through it. That is a plausible (unproven) contributor to the 914 ms
   subscribe that #2421 turned out to be.
2. **Signal hygiene** — the detector is deliberately once-per-role+type, and every affected test
   class emits two `[Error]` lines. Errors that everyone learns to ignore are how a real report gets
   missed (the detector's own docstring makes this argument).

## Scope

Repo-wide, not local to one test: **58 `Mesh.Observe(` call sites across 32 test files**.

The fix is not a mechanical find-and-replace — some of those genuinely test mesh-hub-addressed
operations (`CreateNodeRequest` / `MoveNodeRequest` are legitimately routed there). The shape to aim
for is a client hub (`GetClient()`) as the default request origin for tests that address a *node*,
with `Mesh` reserved for tests that address the *router*. Possibly a test-base helper so the choice
is made once rather than per call site.

rbuergi:
Closes #2423.

`MonolithMeshTestBase.Mesh` **is** the root mesh hub, so a test using it as its
request client made the ROUTER genuinely both ends of the delivery — request out
stamped `Sender = mesh/{id}`, response addressed straight back at `mesh/{id}`.
That is exactly what `RouterTrafficRule.RoleOf` reports, and because the detector
fires once per role+type, every affected class emitted two `[Error]` lines.

## Classified by what each site ADDRESSES, not by how it is spelled

| Bucket | Count | Change |
|---|---:|---|
| Addresses a **NODE** | 58 | → `RequestHub` (new: one `GetClient()` per test method, disposed with the rest) |
| **Node CRUD** (aimed at `Mesh.Address`, or target-less) | 21 | → `ObserveNodeOperation(...)` — a client origin **and** `NodeOperationTarget()` |
| `AwaitResponseAsync` default hub | 7 | `Mesh` → `RequestHub`; the sites relying on the old default got an explicit target |
| Addresses the **ROUTER** (kept) | 6 | unchanged, each with a 🚨 comment saying why |

**Why node CRUD needed more than "not the router":** a **target-less** delivery is
handled by the POSTING hub, so those requests *executed* on the router's action
block — the starvation shape from prod 2026-06-11. `ObserveNodeOperation` aims
them at `portal/nodeops-{meshId}`, so neither end is the router. That is the seam
`MeshService` already uses and `NodeOperationOriginTest` pins.

**The 6 kept sites are not debt** — in each, the router is the *subject*:

- `RouterTrafficDetectorTest` must **produce** the shape (it asserts the "sender"
  role still fires; migrating it would delete the only guard against a future
  narrowing of `RouterTrafficRule` silently switching the detector off).
- `StreamRouteSelfParentTerminationTest` asserts the mesh hub's **own drain** stays
  responsive after a poison `StreamMessage`.
- `UpsertInnerCreateObservationTest` (arm 1) pins #981's literal
  `CreateNodeRequest@mesh/<self>` — a self-targeted create on the router whose
  reply the router's own `HandleCallbacks` drops.
- `ShutdownRoutingRejectClassificationTest` depends on a **synchronous enqueue
  ordering** on the router's action block (stall → probe → `Dispose()`); from a
  client hub the enqueue becomes a routed hop and `Dispose()` could beat the
  probe in, so the reject would come from a different branch — a flake, not a
  stricter test.

## The issue's inventory was an undercount — now measured, and ratcheted

`Mesh.Observe(` as a **substring** finds **56** of the **85** sites in `test/`. It
misses **29**: `Mesh.Observe<TResponse>(` is a different string, and eighteen sites
wrap as `await Mesh` ⏎ `.Observe<…>(`. Eleven of the misses were target-less node
CRUD executing on the router. So the new `RouterAsTestRequestOriginRatchetGuard`
scans with a **whitespace-tolerant regex** over comment/string-masked source
(`SourceScan`), seeded at 6 in `test/RouterRequestOriginSites.allow`, and may only
shrink. It carries the usual non-vacuity test plus two masking canaries (its own
file, and `siloMesh.Observe<…>` which the identifier-boundary look-behind must
exclude).

**The guard was shown FAILING before it was trusted.** Injecting one wrapped site
into a file with no allowance:

```
  NEW SITE   test/MeshWeaver.AI.Test/CodeCellLayoutTest.cs (1) — issue the request from
  MonolithMeshTestBase.RequestHub (a client hub), and for node CRUD aim it at
  RequestHub.NodeOperationTarget(). Do NOT add a line to RouterRequestOriginSites.allow.
Failed!  - Failed: 1, Passed: 1
```

(injection reverted; the tree is the committed one.)

## Evidence — measured, not asserted

`ScriptExecutionInUserHomeTest.Run_StreamsProgressTimely_NotBuffered`, the issue's
own example, run **before** the change (file restored from `origin/main`, rebuilt,
same command):

```
[Error] ROUTER_TRAFFIC: RawJson has the mesh hub as sender
  (sender: mesh/XnvduYALJ0adLXeUOWdNGQ, target: rbuergi/demo-42cdbbc4898f4d129ffe029d9b719587)
[Error] ROUTER_TRAFFIC: ExecuteScriptResponse has the mesh hub as target
  (sender: rbuergi/demo-42cdbbc4898f4d129ffe029d9b719587, target: mesh/XnvduYALJ0adLXeUOWdNGQ)
```

**After: 0 lines** — and 0 across all 9 tests of that class.

- **155 tests green locally**, 13 projects, 0 failures.
- **17 projects build clean** at `dotnet build -c Release -warnaserror`.
- Every site reconciled: 85 before, 85 after (`Mesh.Observe` + `RequestHub.Observe`
  + `ObserveNodeOperation`), so nothing was dropped rather than migrated.

## Deliberately out of scope — filed as #2477

1. **A production `ROUTER_TRAFFIC` source**, visible only once the harness stopped
   generating noise: `MeshExtensions.PreValidateDescendantsObs(meshHub, …)` posts
   `ValidateDeleteRequest` from the **root mesh hub**, so every recursive delete
   logs two lines. Reproduced independently by
   `NodeOperationsTest.DeleteNode_WithChildren_Recursive_ShouldSucceed` (production
   APIs only) and `DeleteTimeoutStageDiagnosticsTest`. Fix is
   `meshHub.NodeOperationIssuingHub()`, but it moves a permission-sensitive
   pipeline and belongs in its own reviewed change.
2. **The TARGET half in tests** — 61 `client.Observe(req, o => o.WithTarget(Mesh.Address))`
   sites (27 `RlsIntegrationTests`, 22 `MeshWeaver.Threading.Test`, 6
   `MeshNodeAuditingTest`, 6 others). Origin is already off the router but the
   request still *executes* on it; retargeting moves execution onto a hub with a
   different permission surface — a judgement call per security assertion.
3. **`Mesh.Post(...)`** — 12 sites mixing the router's genuine duties
   (`HeartBeatEvent`, `DisposeRequest` to a hub it hosts) with real violations
   (target-less `TrackActivityRequest` in `Query.Test`/`Auth.Test`). One marker
   cannot tell them apart, which is why the ratchet does not match it.

No What's New entry: test-harness only, no user-visible effect.

### DIFF
--- a/src/MeshWeaver.Hosting.Monolith.TestBase/MonolithMeshTestBase.cs
+++ b/src/MeshWeaver.Hosting.Monolith.TestBase/MonolithMeshTestBase.cs
@@ -974,9 +974,19 @@ public abstract class MonolithMeshTestBase : Fixture.TestBase
 
     /// <summary>
     /// Test-only Task wrapper around <see cref="MessageHubExtensions.Observe{TResponse}"/>:
-    /// posts <paramref name="request"/> via <paramref name="hub"/> (defaults to <see cref="Mesh"/>)
-    /// and awaits the typed response, propagating <see cref="DeliveryFailureException"/> /
-    /// <see cref="TimeoutException"/> as the awaited Task's exception.
+    /// posts <paramref name="request"/> via <paramref name="hub"/> (defaults to
+    /// <see cref="RequestHub"/>, NOT <see cref="Mesh"/> — see that property for why the router must
+    /// not be an end of the delivery) and awaits the typed response, propagating
+    /// <see cref="DeliveryFailureException"/> / <see cref="TimeoutException"/> as the awaited Task's
+    /// exception.
+    /// <para>
+    /// 🚨 The default hub is a CLIENT, so a TARGET-LESS request is delivered to that client, which
+    /// has no handlers. Node CRUD must therefore name its target —
+    /// <c>o =&gt; o.WithTarget(RequestHub.NodeOperationTarget())</c>, or use
+    /// <see cref="ObserveNodeOperation{TResponse}"/>. Before #2423 the default was <see cref="Mesh"/>
+    /// and a target-less request silently EXECUTED on the router's action block, which is the
+    /// starvation shape the <c>ROUTER_TRAFFIC</c> detector exists to surface.
+    /// </para>
     /// <para>
     /// Use ONLY in test code — production hub handlers / click actions / services MUST stay
     /// on the observable form (<c>hub.Observe(request).Subscribe(...)</c>). The Task return
@@ -1009,7 +1019,7 @@ public abstract class MonolithMeshTestBase : Fixture.TestBase
     {
         try
         {
-            return await (hub ?? Mesh).Observe(request, options)
+            return await (hub ?? RequestHub).Observe(request, options)
                 .FirstAsync()
                 .ToTask(ct ?? TestContext.Current.CancellationToken);
         }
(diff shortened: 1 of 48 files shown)


rbuergi:
Follow-ups filed as #2477 (the production `PreValidateDescendantsObs` router-origin defect, plus the target-half and `Mesh.Post` inventories).

rbuergi:
CI attempt 1 red on `OrleansGrainTeardownStragglerTest.GrainLifetimeCallbacks_AfterActivationIsInvalid_AreGracefulNoOps_NotThrows` (shard 3) — the known flake #2301, recurring after the #2270 fix that closed it (recurrence reported there with evidence).

Not reachable by this diff: zero files under `test/MeshWeaver.Hosting.Orleans.Test/`, and that project has no reference to `MeshWeaver.Hosting.Monolith.TestBase` — the only `src/` file this PR touches. `OrleansGrainTeardownStragglerTest : OrleansSharedTestBase : TestBase` never passes through it. The runner also logged thread-pool starvation, a 2 s runtime stall and Orleans self-health degradation inside the same 30 s window.

Re-running the failed jobs only.

PR #2478, opened 2026-08-27
Plugin build contract: the bake ships the package, the install seed is bounded by its own work

rbuergi:
Three pieces that together make step 2 of the plugin build contract — **install my dependencies** — mean *install* rather than *rebuild*. The doc lands first so the code has something to be measured against.

## 1. `Doc/Architecture/PluginBuildContract`

Four steps: take the image, install dependencies as built artifacts, build your plugin, test it. One rule inside them: **step 2 is an install, never a build.** Plus the measured state of every repo against it, and why mounting bundles alone is not enough.

## 2. The bake emits node definitions

`BundleReader.Manifest.Content` has said all along that a consumer using a package as an upstream needs them — *"the bytes only stamp nodes that already exist, so without the definitions there is nothing to stamp."* The writer accepted them. **Both bake call sites passed nothing.** So every published bundle was assemblies-only, which is exactly why `node-repo-publish-bake`'s `stage-repo` carries *"staging cannot be dropped until a published artifact carries the upstream's NODE DEFINITIONS"* — and why every plugin repo stages upstreams as **source** and recompiles them.

| | `--bake-output` | `--seed` |
|---|---|---|
| core | 21 | **9** |
| every plugin repo | 1–2 | **0** |

`BakeOutput.NodeDefinitionsOf` collects the package's whole tree relative to its source folder; both `BakeOutput` and `TreeBake` pass it, `SourceIncluded` declared `true`.

**Pinned in the Education shape** (`BundleCarriesTheUpstreamTest`): repo 1 bakes a base package; repo 2 reconstructs and *rebuilds* it from repo 1's bundle with **no checkout of repo 1 at all**. That is what "install my dependencies" means, and it was impossible before this.

## 3. The install-time seed is bounded by its own work

It was a flat 60 s over a seed that is a sequential `Concat` with a **30 s budget per assembly** — a 17-assembly package (Store) was legitimately allowed minutes while the cap fired at one. And `Timeout` abandons the *result*, not the *work*. Measured on Education e2e shard 1 this morning:

```
06:38  Store: "⏳ Compiling… Running Roslyn against 9 source binaries"
06:51  Install: prebuilt adoption attempt failed — the installed types compile instead
06:51  27 prebuilt assembly(ies) … 27 already current
```

When the cap fired the system did **both** — the seed and the compile it exists to avoid. Same inversion #1317 removed one layer down. Bound is now `floor + perType × types` where `perType` *is* the inner budget, so the outer bound cannot expire while the inner work is inside its own. Pure function, pinned with no hub (`InstallSeedBoundTest`).

## Not in this PR

Wiring `--seed` into each plugin repo's gate and dropping `stage-repo` — that is one PR per satellite once this is on main, and each is now a mechanical change rather than a design one.


rbuergi:
Shard 2 went red with ten Monolith compile tests failing. Diagnosed before re-running, not after:

- Every failure is the same shape — `NullReferenceException` **inside Roslyn's emitter**: `Microsoft.Cci.MetadataWriter.GetConsolidatedTypeParameters → NamedTypeSymbol.ContainingTypeDefinition`. Not a MeshWeaver frame.
- **Control:** the four failing classes are 7/7 green on unmodified `origin/main` locally.
- **Repro:** the same four are 7/7 green on this branch locally.
- The runtime surface of this PR is three files; `Hosting.Monolith.Test` references none of `PluginTester`, and `SeedPrebuiltAssemblies` returns 0 immediately when no `IPrebuiltAssemblyConsumer` is registered — which the Monolith host does not register. There is no path from this diff into Roslyn emit in that suite.
- Same `Microsoft.CodeAnalysis.CSharp` 5.6.0 on both. Main's own run on `397db998` (post-#2479, which this branch merged) was green on shard 2.

Environmental — Roslyn's metadata writer under runner pressure on the shard that carries the Monolith floor. Re-ran the failed jobs; nothing pushed.

PR #2481, opened 2026-08-27
The shared gate INSTALLS its upstreams: upstream-seed fetches the sealed publication and hands it to the tester as --seed

rbuergi:
**Draft until #2478 lands** — it works only once published bundles carry node definitions; before that the seed stamps nothing and the tester's `Shortfall` verdict correctly refuses the run.

Step 2 of the plugin build contract (`Doc/Architecture/PluginBuildContract`, on #2478): **install my dependencies** — as built artifacts, served by MeshWeaver.

## The gap

The reusable gate had eight inputs and **no way to consume a bake**. `stage-repo` copied the upstream's *source* into the mount and the tester recompiled it. Measured today: core passes `--seed` in 9 places, every plugin repo in **0**.

## What `upstream-seed: plugins` does

1. **Resolves the framework identity its own image resolves** — composed exactly as the tester step composes it (digest-pinned), so the seed is addressed to the identity that will consume it.
2. **Asserts each upstream has a sealed publication under that identity** (`_complete`, existence-first, fail-closed — the publish script's own rule). Fails **red** if not: a gate that silently compiled its upstream is indistinguishable from one that installed it, which is `BakeSeed`'s own warning.
3. **Downloads the bundles** into a seed dir with `framework-mvid.txt` beside them, mounts it read-only, passes `--seed /seed`. The tester checks the seed *before the mesh boots* (usage error, exit 2 — never green) and refuses a run that compiled what it was handed.

Reads storage with the **same OIDC identity publish-bake writes with**, declared the same way: three optional secrets, `id-token: write`, a preflight that names what is missing rather than skipping.

Empty `upstream-seed` is today's behaviour byte for byte.

## Verified

- YAML parses; every `steps.*` / `inputs.*` / `secrets.*` reference resolves to a declaration (the checker caught a phantom `steps.pin` on the first pass — fixed to compose the image as the tester step does).
- `bash -n` clean on every `run:`.

## Then, per satellite (one small PR each)

```yaml
with:
  upstream-seed: plugins
  bake-publish-targets: ${{ vars.BAKE_PUBLISH_TARGETS }}
secrets:
  azure-client-id: …   # the three already provisioned for publish-bake
permissions: { contents: read, id-token: write }
```

and drop `stage-repo` / `stage-modules`.

### DIFF
--- a/.github/workflows/node-repo-gate.yml
+++ b/.github/workflows/node-repo-gate.yml
@@ -111,6 +111,14 @@ on:
         required: false
         type: string
         default: ''
+      registry-url:
+        description: >-
+          The registry that SERVES sealed publications (MeshWeaver#2487), e.g.
+          https://memex.meshweaver.cloud. With registry-key set, upstream-seed fetches over
+          HTTPS with an instance key and never touches Azure — and it works on pull_request.
+        required: false
+        type: string
+        default: ''
       bake-publish-targets:
         description: >-
           Where upstream-seed READS the upstream publication from — the same
@@ -137,6 +145,9 @@ on:
       # The same OIDC identity publish-bake uses, read-only here: the gate FETCHES the caller's
       # upstream publication to seed from. Required only when upstream-seed is set; the preflight
       # below fails RED naming whichever is missing rather than letting the step skip.
+      registry-key:
+        description: An instance key (mwi_…) holding a whole-source grant on each upstream (upstream-seed via registry-url).
+        required: false
       azure-client-id:
         description: Client id of the Azure identity that can read the portals' storage (upstream-seed only).
         required: false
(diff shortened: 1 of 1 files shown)


rbuergi:
## ⚠️ Cannot go green on a PR until a federated credential exists for it — maintainer decision

The first run of `upstream-seed` on a PR failed at the OIDC login, and the reason is provisioning, not code:

```
AADSTS700213: No matching federated identity record found for presented assertion subject
  'repo:Systemorph@77832550/MeshWeaver.SocialMedia@1320607286:pull_request'
```

`github-actions-bake` holds **four** federated credentials, every one scoped to `ref:refs/heads/main`:

```
gh-meshweaver-reinsurance-main   repo:Systemorph@77832550/MeshWeaver.Reinsurance@881295217:ref:refs/heads/main
gh-meshweaver-socialmedia-main   repo:Systemorph@77832550/MeshWeaver.SocialMedia@1320607286:ref:refs/heads/main
gh-reinsurance-main-classic      repo:Systemorph/MeshWeaver.Reinsurance:ref:refs/heads/main
gh-socialmedia-main-classic      repo:Systemorph/MeshWeaver.SocialMedia:ref:refs/heads/main
```

That is exactly right for publish-bake, which runs on main. The **gate** runs on pull requests, whose subject is `…:pull_request`, and nothing matches it — so the gate can install its upstream on main and never on the PR that most needs it.

**What closes it** — one credential per repo for the PR subject, in **both** subject formats (AGENTS.md: never normalise to one; the format varies per repo within the org):

```
repo:Systemorph@77832550/MeshWeaver.SocialMedia@1320607286:pull_request
repo:Systemorph/MeshWeaver.SocialMedia:pull_request
repo:Systemorph@77832550/MeshWeaver.Reinsurance@881295217:pull_request
repo:Systemorph/MeshWeaver.Reinsurance:pull_request
```

Read-only on storage, same role the identity already holds. Adding a credential to a live identity is a system change, so I have **not** done it and have **not** armed this PR. Until it exists this PR is correct code that cannot be verified on its own CI.

rbuergi:
**Direction change, per maintainer:** the Azure-OIDC fetch in this PR is the *interim*. The target is the mesh verifying GitHub's OIDC identity directly — no secret, a `Store/BuildPrincipal` trust node per repo, and the registry serving publications so the gate never touches Azure. Scoped in Systemorph/MeshWeaver#2483; design on #2478. This PR stays draft; the `pull_request` credential gap above becomes moot once that lands.

rbuergi:
**Registry mode added** (`registry-url` input + `registry-key` secret): `upstream-seed` now fetches the sealed publication over HTTPS from `GET /api/plugins/bundles/prebuilt/<identity>/<source>` (MeshWeaver#2487) with an instance key holding a whole-source grant — no Azure, no federated credential, works on `pull_request`. The Azure mode stays as the fallback when `registry-url` is empty. The `AADSTS700213` gap above is moot on this path; what each repo needs instead is one `mwi_` key with `plugins/*`.

PR #2493, opened 2026-08-27
The gate BAKES ONCE and tests the baked bytes; publish-bake REUSES that bake (bake-run-id)

rbuergi:
Steps 3 and 4 of the plugin build contract: build once, test what you built.

The reusable gate compiled every package in-mesh and DISCARDED the result; on
merge, publish-bake compiled the identical source a second time. Two full Roslyn
passes per change, and the PR gate judged bytes that never shipped.

The gate now mirrors publish-bake's own shape: `compile /repo --output /bake`
(mesh-free, the producer) writes this repo's OWN bundles; the gate then stands
its mesh up on THOSE bytes with `--seed`, so what is tested is what would ship;
and the bake is uploaded as `bake-<sha>`. An upstream seed (upstream-seed) is
merged into the same seed directory — consumed alongside, never re-emitted.

publish-bake gains `bake-run-id`: download `bake-<sha>` from that run and skip
the compile. The reused bake's framework-mvid.txt must EQUAL the identity this
publish targets — a mismatch fails the job, never falls back to a silent
recompile, because publishing a bake addressed to another identity would seal
bundles no portal here can adopt.

Floor: the `compile` verb landed in mw-plugin-test at 77bc5f1 (2026-08-17). A
caller pinned below it fails the gate loudly at the identity check, naming the
floor. Reinsurance's pin (a0a0288, rc3) is below it and is bumped alongside.

Co-Authored-By: Claude Opus 5 (1M context) <noreply@anthropic.com>


rbuergi:
Shard 5 red was **real and correct**: `PlatformBakeLaneGuard.PlatformBake_NeverAdoptsAnotherBuildsBundles` scans every workflow for one that both calls `publish-bake-bundles.sh` and uses `download-artifact` — exactly the `bake-run-id` reuse I added to `node-repo-publish-bake.yml`. Control on main: 9/9 green, so it was mine.

The guard is right and I have removed the reuse. Its premise holds even with my identity check: a `FOLLOW_RELEASE` publish bakes against the *released* image, the PR gate against the *pinned* one — two compilations that never share an identity — and a structural guard should not be argued with from inside the PR it is guarding.

What stays, and is the actual win: the **gate** bakes once (`compile`, mesh-free) and tests **its own baked bytes** (`--seed`), so the PR verdict is on what would ship, and the bake is uploaded for in-run diagnosis. publish-bake compiles from the promoted image, unconditionally, as the guard requires. The `compile`-verb floor and Reinsurance's re-pin are unchanged.

PR #2497, opened 2026-08-27
ci(gate): a 200 from the registry is not a publication index until it parses as one

rbuergi:
The portal answers 200 + HTML (the app shell's catch-all route) for ANY unknown path, /api/…
included, so a registry that does not serve /api/plugins/bundles/prebuilt yet answers the
upstream-seed index request with a web page. Measured 2026-08-27 on MeshWeaver.SocialMedia#84
against a memex one build behind #2487: jq's parse error was swallowed, `fetched` stayed 0, and
the step died as "sealed but empty" — the wrong condition, pointing at the wrong fix.

The registry branch now requires the index to be a JSON object with a `bundles` array; anything
else fails loudly naming the real condition (the registry runs a build without #2487). The
portal answering 200 for unknown API routes is the root cause and gets its own change.


rbuergi:
Root cause (the portal answering 200 HTML for unmapped `/api/…` paths) is fixed in #2498; this PR stays as the gate-side defence — a registry that is merely behind must still fail with the true reason.

PR #2560, opened 2026-08-27
The self-update trigger could never see a build record — satellites need a scoped query

rbuergi:
**This is why memex-cloud stopped self-updating.** Not the webhook, not ACR, not the update policy — the trigger query itself could never match.

## The defect

`SelfUpdateHostedService.BuildCompletionTicks()` watched an **unscoped** index query, `nodeType:BuildCompletion`. Build records are **satellites** (`Admin/_Build/{owner}.{repo}`), and satellites are excluded from an unscoped query. So the watch returned empty however many records existed and however often they were written — it never ticked, and after its single startup pass the install simply stopped updating, with every component reporting success.

## Measured on memex-cloud, 2026-08-27

```
search nodeType:BuildCompletion                                   →  0 results
search path:Admin/_Build scope:children nodeType:BuildCompletion  →  5 results
get @Admin/_Build/Systemorph.MeshWeaver   →  version 1746, written minutes earlier
```

Meanwhile the install served `ci.6145` while `ci.6158`, `ci.6178` and `ci.6183` were published, `Admin/UpdatePolicy` frozen at `checkedAt: 20:45Z` — and every other link healthy: hook subscribed to `workflow_run`, **15/15 recent deliveries `200 OK`**, `GitHubWebhookProcessor` writing the node **1,746 times**.

## The fix, and why this shape

The sibling consumer already had it right, and says why — `PluginUpdateWatcher`:

```csharp
// The canonical single-node read: one shared handle per path, live, authoritative.
// Never QueryAsync — that index lags a write by design.
var sub = hub.GetMeshNodeStream(buildPath)
```

It watches **one** repository, so a node stream fits. The self-updater needs the **set** across repositories, so it gets a **path-scoped query** instead — the form proven above to actually return the records.

`BuildCompletion.WatchQuery` becomes the single definition of that query, sitting next to `PathFor`, so the watch scope and the write path cannot drift apart. That pairing is the real guard: a scope that does not contain the records is exactly as blind as no scope, and looks just as correct.

## Tests

`BuildCompletionWatchQueryTest` (2) pins the scope, the namespace, and the exact query string — so a well-meaning simplification back to the bare form fails here rather than in production weeks later. The 22 existing self-update tests are unchanged and green.

Context: Systemorph/Memex#138 (where I first mis-diagnosed this as missing webhook wiring, then as missing nodes — both wrong, corrected there).

🚨 Worth generalising in review: **an unscoped `nodeType:` query is not a valid watch for anything satellite-shaped**, and it fails silently. Other watchers should be checked for the same shape.

https://claude.ai/code/session_01Tz1i2fFcJhvR1Eje8yFH5F

### DIFF
--- a/memex/Memex.Portal.Shared/SelfUpdate/SelfUpdateHostedService.cs
+++ b/memex/Memex.Portal.Shared/SelfUpdate/SelfUpdateHostedService.cs
@@ -201,11 +201,12 @@ public class SelfUpdateHostedService : IHostedService
         Observable
             .Defer(() => NewBuildEventsAcross(_hub
                 .GetQuery($"SelfUpdate.BuildTrigger:{_hub.Address}",
-                    // 🚨 PATH-SCOPED (BuildCompletion.WatchQuery). Build records are SATELLITES
-                    // under Admin/_Build, and satellites are excluded from an unscoped query — the
-                    // bare `nodeType:BuildCompletion` this used to pass returned EMPTY however many
+                    // 🚨 PATH-SCOPED (BuildCompletion.WatchQuery). Build records live in the ADMIN
+                    // partition, which an UNSCOPED query does not reach — the bare
+                    // `nodeType:BuildCompletion` this used to pass returned EMPTY however many
                     // records existed and however often they were written, so this watch never
-                    // ticked and the install silently stopped self-updating.
+                    // ticked and the install silently stopped self-updating. (Not a satellite rule:
+                    // GitHubSyncConfig satellites in ordinary partitions ARE returned unscoped.)
                     BuildCompletion.WatchQuery)))
             .Throttle(_options.EventCoalesceWindow)
             .Select(_ => -2L)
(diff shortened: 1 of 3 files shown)


rbuergi:
## Correction pushed: the reason is the ADMIN PARTITION, not satellites

The fix is unchanged and still correct. The **explanation** I shipped with it was wrong, and since it sits in a 🚨 comment that the next reader will generalise from — exactly as I did — it needed correcting before merge rather than after.

I claimed satellites are excluded from unscoped queries. They are not. Measured as one viewer on memex-cloud:

| query | result | shape |
|---|---|---|
| `nodeType:BuildCompletion` | **0** | `Admin/_Build/…` — satellite |
| `nodeType:UpdatePolicy` | **0** | `Admin/UpdatePolicy` — **not** a satellite |
| `nodeType:Store/Coupon` | **0** | `Admin/Coupons/…` — **not** a satellite |
| `nodeType:GitHubSyncConfig` | **5** | `{Space}/_GitSync` — satellite, and **visible** |

So the discriminator is the **partition**: the Admin partition is not reachable from an unscoped query, whatever the node's shape. Satellite-ness has nothing to do with it.

That also changes the generalisation I offered for review. It is not "sweep for satellite watchers" — it is **anything that reads the Admin partition BY TYPE needs a path scope**. From the sweep, the unscoped `nodeType:` call sites in `PluginCatalog`/`GitSync` target types in ordinary partitions (`Package`, `GitHubSyncConfig`, `MeshWeaverInstance`, `ModuleDiscovery`), so they are not affected — but that is now a claim someone should confirm against where each type actually lands, using the four-way measurement above as the discriminator rather than the node's shape.

Worth stating plainly for whoever reviews: this is the third wrong explanation I gave for the same symptom tonight (missing webhook wiring → missing nodes → satellites), each corrected on measurement. The fix survived all three because it was derived from the working query, not from the theory.

Issue #2563, opened 2026-08-28
Prod: every prebuilt bundle referencing MeshWeaver.AI is DECLINED after the cutover — 218 NodeTypes compile on first access


rbuergi:
# 🚨 Prod: every bake referencing MeshWeaver.AI is DECLINED after the cutover — 218 NodeTypes now compile on first access

Measured on memex-cloud, 2026-08-28 ~00:55Z, after tonight's AI delivery cutover (#2276: the engine left `ContentSurfaceAssemblies` in Systemorph/MeshWeaver#2524 and now ships as a registry module).

```
Prebuilt assembly for Store/Plugin DECLINED: dependency record mismatch —
  'MeshWeaver.AI' built against ref:5A915A0238FA8A94E884C655628A3A53445CA61D84C9517A980BE2206E624E66,
  live is mvid:4fcf729f140d47fe8…
```

Same for `Store/Order`, `Store/Maintenance`, `Store/Installer`, `Store/Catalog`, `RolePlay/Story`, … and the aggregate:

```
DynamicTypePreWarmer: ADOPT-ONLY boot complete — adopted=133 uncovered=218 of 351 dynamic NodeType(s)
DynamicTypePreWarmer: 218 dynamic NodeType(s) were NOT covered by the prebuilt bundles and will each
  pay a Roslyn compile on FIRST ACCESS  …  framework=s8929555
```

## What changed and why it declines

The AI engine moved from being a **content-surface assembly** (recorded by REFERENCE hash, `ref:…`) to an **installed module** (matched by **MVID**, `mvid:…`). Existing bundles carry the `ref:` record; the live process presents an `mvid:` one. The dependency records are not comparable, so every bundle that references AI is declined — correctly, by a gate doing its job.

## Impact, and what it is not

- **Not an outage.** Pages work; each uncovered NodeType pays one Roslyn compile the first time it is opened. 218 of 351 are uncovered.
- **This is the degraded-start cost of an identity flip**, which was expected — but it is worth confirming whether it *converges*. The next CI bake runs with AI-as-a-module and should emit `mvid:` records that match. **If it does not converge, this is permanent**, and every pod restart re-pays 218 compiles.

## What to check next

1. Does the bake published **after** the cutover (`c3ba6a8` / `ci.6215`, whose CD bake step succeeded) produce `mvid:`-shaped records for `MeshWeaver.AI`? If yes, this clears on the next roll and needs nothing.
[...]

Issue #764, opened 2026-08-03
Plugin Catalog: progress and bulk-update affordances (Update all, loading state, install-in-progress)


sierragolflima:
Three low-priority usability improvements for the Plugin Catalog settings page (`/{partition}/Settings/PluginCatalog`). **None of these is a bug** — the catalog lists, installs and updates correctly in every case below. What is missing is feedback: the page does not tell the user what it is doing while it is doing it, and bulk updating is manual.

Grouped into one issue because all three sit in the same view (`src/MeshWeaver.PluginCatalog/CatalogLayoutAreas.cs`) and would be addressed together. They can be split if it is easier to schedule.

---

## 1. Add an "Update all" button

Each package whose installed `ModuleVersion` differs from the catalog's renders its own `Update to v{version}` button (`CatalogLayoutAreas.cs:206`). With many installed plugins, bringing everything current means clicking through the list card by card and keeping track of which ones were already started.

**Suggestion:** render a single **Update all** action in the page header when at least one package has an update available. It would run the same `InstallPackage(...)` path for every card currently in the "update available" state — the set already computed by the `upToDate` check at `CatalogLayoutAreas.cs:194-197`. A count in the label (for example `Update all (4)`) would also make it visible at a glance how many packages are behind, which today requires scrolling the whole list.

Sequential rather than concurrent execution would be preferable, so the updates do not contend for the same I/O budget.

---

## 2. The empty-state message is shown while the catalog is still loading

For the first few seconds after opening the page, the view shows:

> *No installable packages found.*
>
> Source: registry — 0 package(s) available.

The packages then appear normally. The wording states a definite negative result at a moment when the result is not yet known, so the natural reading is that the catalog is empty or the source is misconfigured — and a user who navigates away during that window concludes exactly that.

**Cause:** `ObserveAvailable` seeds its stream with an empty list (`.StartWith((IReadOnlyList<PackageManifest>)[])`, `CatalogLayoutAreas.cs:101`), and `ObserveInstalled` does the same (`:132`). The `CombineLatest` therefore emits immediately with `available.Count == 0`, which satisfies the empty-state branch at `:160-161`. The `*Loading catalog…*` placeholder at `:86` is displaced by that first real emission almost instantly, so it is never actually seen. The seeded empty list and a genuinely empty source are indistinguishable at the render site.

[...]

PR #2565, opened 2026-08-28
The bake compiles against modules, like the portal it bakes for (#2563)

rbuergi:
## The fleet was held back by a reference list

`memex-cloud` has been sitting three published images behind on `ci.6145`, and every component
reported success. The self-updater was working perfectly — it was **correctly declining** an image
whose content had no bake.

Here is why there was no bake.

A module is published into `modules/<name>/`, so it is by construction **absent from
`TRUSTED_PLATFORM_ASSEMBLIES`**. It reaches a compile only by being composed in
(`CompileReferences.ComposeWithModules`). The portal does that from its installed-module set. The
bake used the bare TPA baseline and could not resolve a single module type.

So when the AI engine left the content surface to become a module (#2276), five `Store/*` NodeTypes
stopped resolving `AiSettings` **in the bake and nowhere else**.

## Why nobody saw it

The two lanes of the *same tool* disagreed, silently:

| lane | mesh? | modules composed | verdict |
|---|---|---|---|
| `compile-check` (gate) | yes | yes, for free | 🟢 green |
| `publish-bake` | no | **no** | 🔴 red |

Same content, same image, same commit. A reader sees a green compile gate beside a red bake and
concludes the *bake* is broken — not that a reference is missing. And nothing in the failure names a
module; it arrives as `CS0246: The type or namespace name 'AiSettings' could not be found`, which
reads as a content defect.

## The fix

The module set now lives in **one place both lanes read** (`TesterModules`). That is the actual
fix — a module added for the gate can no longer go missing from the bake's reference set.

The bake also records module ids by **exact-build MVID** instead of leaving the map empty, so the
dependency record a consumer checks names the build it will bind. That is the second half of #2563,
where a bundle's `ref:` entry could never match a module-installing portal's `mvid:`.

Resolution happens at the **CLI boundary** deliberately: resolving names `MeshBuilder`, and
`MeshFreeBakePathTest` fails the build if anything reachable from `TreeBake.Run` so much as *names*
a mesh type. That guard is right — a bake is a build step, not a mesh — so the bake receives
resolved paths and never meets a mesh type. **The guard caught this during development**, which is
why the shape changed.

No workflow changes: no caller passes `--module` today, and the image-shipped set is resolved
beside the binary exactly as the gate resolves it.

## Verification

- **End to end on the real content that failed.** The actual `MeshWeaver.Plugins` tree —
  the one CI could not compile — bakes **75/75 NodeTypes into 33 bundles, exit 0**, logging
  `bake: module MeshWeaver.AI mvid=… — composed into the reference set`.
- **The causal chain is closed, not inferred.** `MeshWeaver.AI.dll` appears in **zero** `deps.json`
  entries and exists only at `modules/MeshWeaver.AI/`, so `CompileReferences.Default` could never
  have contained it.
- **The new test has a control.** It bakes the same tree with no modules and *requires* it to fail
  on the missing type — so the test cannot pass against the behaviour it pins.
- Full `MeshWeaver.PluginTester.Test` suite: **96/96 pass**. `MeshFreeBakePathTest` still green.
- Release + `-warnaserror` clean.

The framework identity is unchanged — no `ProjectReference` was added or removed, so this does not
flip the identity again.

## What this unblocks

Once this is in an image, the satellites' scheduled runs (which follow the released image) bake
against the current identity, bundles get sealed, and installs resume updating on their own.

`MeshWeaver.Plugins` has a **second, unrelated** blocker in its `test-repos` gate —
`ProvisionPlan 'Analysis' FAILED in phase 2/7 'Create home root': timed out`. That one compiles
fine (`compile=Ok`) and is tracked separately.

### DIFF
--- a/src/MeshWeaver.Documentation/Data/WhatsNew/2026-08-28-portals-update-again-after-a-module-moves.md
+++ b/src/MeshWeaver.Documentation/Data/WhatsNew/2026-08-28-portals-update-again-after-a-module-moves.md
@@ -1,0 +1,23 @@
+---
+Name: Portals update again after a module moves
+Category: Fix
+Description: Content that uses a module now builds ahead of time as well as at run time, so installs no longer stop updating when a feature moves into a module.
+Icon: Sparkle
+Order: -20260828
+---
+
+# Portals update again after a module moves
+
+Content can use the features a portal has installed — an installer that adds a plugin's skills to
+your AI settings, a page that draws a map. When one of those features moved out of the platform and
+became a module you install, that content still worked in the portal but could no longer be built
+ahead of time. The prepared build is what an install checks for before it updates itself, so
+installs correctly declined every new version and stayed on the last one they had a build for.
+
+Everything reported success while that happened, which is what made it hard to see: the checks that
+run against a live portal passed, because a live portal has its modules. Only the ahead-of-time
+build was missing them, and its failure looked like the content was broken rather than like a
+feature had moved.
+
+The ahead-of-time build now uses exactly the modules a portal does, so content that uses an
+installed feature builds the same way in both places, and updates resume on their own.
--- a/test/MeshWeaver.PluginTester.Test/BakeCompilesAgainstModulesTest.cs
+++ b/test/MeshWeaver.PluginTester.Test/BakeCompilesAgainstModulesTest.cs
@@ -1,0 +1,152 @@
+using System;
+using System.IO;
+using System.Text;
+using MeshWeaver.PluginTester;
+using Xunit;
+
+namespace MeshWeaver.PluginTester.Test;
+
+/// <summary>
+/// 🚨 <b>The bake compiles against the same modules a portal does.</b> A module is published into
+/// <c>modules/&lt;name&gt;/</c> and is therefore, by construction, NOT in
+/// <c>TRUSTED_PLATFORM_ASSEMBLIES</c> — so it reaches a compile only by being composed in
+/// (<c>CompileReferences.ComposeWithModules</c>). The portal does that from its installed-module
+/// set; until this test the bake used the bare TPA baseline and could not resolve a single module
+/// type.
+///
+/// <para><b>How that stayed invisible, and why it needed a test rather than care.</b> The two lanes
+/// of this same tool disagreed silently. The GATE stands up a mesh, so it composed modules for
+/// free and its <c>compile-check</c> went GREEN; only <c>publish-bake</c> went red — on the same
+/// content, the same image and the same commit. A reader sees a green compile gate beside a red
+/// bake and concludes the BAKE is broken, not that a reference is missing. Nothing in the failure
+/// names a module: it arrives as <c>CS0246: The type or namespace name 'AiSettings' could not be
+/// found</c>, which reads as a content defect.</para>
+///
+/// <para><b>What it cost, measured.</b> When the AI engine left the content surface to become a
+/// module (#2276), five <c>Store/*</c> NodeTypes stopped resolving <c>AiSettings</c> in the bake
+/// and nowhere else. No bundle was sealed for the new framework identity; every install then
+/// CORRECTLY declined to self-update onto an image whose content had no bake, and the fleet sat
+/// three published images behind with every component reporting success (#2563). The self-updater
+/// was working perfectly — it was refusing an image the bake had never covered.</para>
+///
+/// <para><b>The control is the point.</b> The same tree is baked twice: once with the module set
+/// composed, once with none. If the second did NOT fail, this test would be asserting nothing —
+/// it would pass just as happily against the broken behaviour it exists to pin.</para>
+/// </summary>
+public class BakeCompilesAgainstModulesTest(ITestOutputHelper output)
+{
+    private const string IndexJson =
+        """{"$type":"MeshNode","id":"ModuleBound","namespace":"","path":"ModuleBound","mainNode":"ModuleBound","name":"Module Bound","nodeType":"Space","state":"Active","content":{"$type":"PluginManifest","description":"Content that binds a module type."}}""";
+
+    private const string TypeJson =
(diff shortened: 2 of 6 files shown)


rbuergi:
## Independent reproduction, with a clean control — three runs, one unchanged commit

Confirming this from the other end. `MeshWeaver.Reinsurance` `main` has now baked commit
`7f454af0` **three times**, and the only variable is the framework:

| run | time | framework | result |
|---|---|---|---|
| [`33114326364`](https://github.com/Systemorph/MeshWeaver.Reinsurance/actions/runs/33114326364) | 20:38 | `s81a9b60ed905b4f5263f55ff8369976d` | **179/179** |
| [`33119003918`](https://github.com/Systemorph/MeshWeaver.Reinsurance/actions/runs/33119003918) | 21:37 | `s96fb96913046dff7a1d6773286ad3fc8` | **167/179 — 12 new** |
| [`33130950060`](https://github.com/Systemorph/MeshWeaver.Reinsurance/actions/runs/33130950060) | 00:49 | `s7a169a5a075f8ddbe4f8c5e3c3ff62c8` | **167/179 — the same 12** |

Same source, byte for byte. The 20:38 green is the control, and it rules out content: nothing in
`7f454af0` changed. The third run also settles that this is **deterministic, not a flake** — a
different framework hash again, the identical twelve.

The twelve, which match your "five `Store/*` stopped resolving `AiSettings`" exactly and add the two
neighbours it took with it:

```
Edu/PromptCell
Store/Catalog · Store/Installer · Store/Maintenance · Store/Order · Store/Plugin
UWDeepfield/IndustryNewsFeed · TreatyCommunication · TreatyDecision
                            · TreatyIteration · TreatyMemento · TreatySubmission
```

with `CS0246` on `ThreadMessage`, `AiSettings`, `ChatClientCredentialResolver`, `CS0103` on
`AiSettingsNodeType`, `CS1061` on `IMessageHub.StartThread`.

## Your diagnosis corrects mine, and the difference matters for what gets changed

I filed the compile half as MeshWeaver#2561 and got as far as "the reference set is TPA + installed
modules, the module IS loading, so the registration must not be reaching the compile service". That
was one step short.

> The bake used the bare TPA baseline and could not resolve a single module type.

That is the piece I did not have, and it changes the remedy rather than refining it. On my reading
the fix looked like a **content** change — a contract assembly carrying `AiSettings` and friends, or
porting twelve NodeTypes off them — and MeshWeaver.Plugins#764 independently proposed the same, which
I took as corroboration. **Two readings of an incomplete picture is not corroboration.** If the bake
simply needs to compose modules the way the portal already does, then no content moves and no
contract assembly is needed.

Worth saying plainly on #764 as well, since its stated assumption — *"the gate lands the engine at
boot, so CI covers the composed state"* — is right about the gate and wrong about the bake, which is
a narrower correction than "the composed state failed" that I left there.

## One thing I would keep from the wreckage of my version

The 20:38-vs-21:37 pair is worth preserving as a regression check in whatever form suits: **the same
commit, two frameworks, and a count.** It is the only signal in this whole chain that was unambiguous
— every other component reported success, including the self-updater, which as you say was declining
correctly.

rbuergi:
### Both satellite trees bake clean with the fix

Run locally against the real checkouts, with the built binary:

| repo | before (CI, released image) | after (this branch) |
|---|---|---|
| `MeshWeaver.Plugins` | *bake never ran — gate red* | **75/75 NodeTypes, 33 bundles, exit 0** |
| `MeshWeaver.SocialMedia` | **38/43, 5 new failures** (`CS0246 'AiSettings'`) | **26/26 NodeTypes, 5 bundles, exit 0** |

Both log the composition line that was missing:

```
bake: module MeshWeaver.AI mvid=… — composed into the reference set
```

### Scope note, so the remaining work is not mistaken for this

`SocialMedia`'s **gate was already green** — only its `publish-bake` was red, and that is what this
PR fixes. So this unblocks SocialMedia end to end.

`Plugins` is blocked by a *separate* defect: its gate wedges on
`CreateOrUpdateNodeRequest@portal/nodeops-*` (59.5 s unanswered → 60 s timeout), which skips its
bake entirely. Filed as Systemorph/MeshWeaver.Plugins#785 — intermittent, tests-phase only,
`compile=Ok`. Prod needs **all three** bake sources sealed (`IsUpdatable` is true only when every
installed package is available), so that one still has to land before the fleet rolls.

Core's own `meshweaver-content` bake is already green for the current identity (`ci.6219`), because
core's content does not reference the AI types — which is exactly why only the satellites broke.

rbuergi:
### Correcting my own scoping: this is all four node repos, not one

I originally described this as SocialMedia's bake. It is not — **every node repo went green→red in
the same window**, which is the identity flip:

| repo | last green scheduled run | first red | state |
|---|---|---|---|
| `MeshWeaver.SocialMedia` | 19:11Z | 22:54Z | 5 RED NodeTypes |
| `MeshWeaver.Education` | 15:11Z | 19:07Z | red |
| `MeshWeaver.Reinsurance` | 18:06Z | 21:37Z | **12 RED NodeTypes** |
| `MeshWeaver.Plugins` | — | — | bake skipped behind a separate gate wedge (#785) |

Reinsurance is the worst hit and shows it is not confined to shared `Store/*` content — its own
domain types are down too: `UWDeepfield/{TreatySubmission,TreatyDecision,TreatyIteration,
TreatyMemento,TreatyCommunication,IndustryNewsFeed}` and `Edu/PromptCell`.

And it is not one type. Unresolved symbols in that run, all from `MeshWeaver.AI`:

```
102 × AiSettings
  8 × ThreadMessage
  2 × ChatClientCredentialResolver
```

Which is the point of the fix rather than a complication: nothing here is per-type. The bake could
not see **the module**, so every symbol in it vanished at once — and composing the module restores
all of them together. The two trees I could bake locally (Plugins 75/75, SocialMedia 26/26) both go
fully green, `Store/*` included.

So the accurate statement of the blast radius: when a feature moved into a module, **all
ahead-of-time content builds across the fleet stopped**, and every install then correctly declined
to update. Core was the only one unaffected, because core's own content does not reference the AI
types.

PR #2581, opened 2026-08-28
Pin the two temporal guarantees of UntilDetermined (#2302 finding 10)

rbuergi:
Closes the last actionable item on #2302 (finding 10).

`UserIdentityLookup.UntilDetermined` had **no test anywhere** — `grep -rln UntilDetermined --include='*.cs' test/ src/` returns exactly one hit, the implementation. (The finding's path is also stale: the file moved from `MeshWeaver.Blazor/Infrastructure` to `MeshWeaver.Hosting.AspNetCore/Portal`.)

Three cases, all pure — a `Subject<Unit>` for the hot feed and a lambda for the lookup, no mesh, hub or query subscription, exactly as the method's own remarks promise. That is the point rather than a convenience: the unavailable window is one a live-mesh test cannot hold open.

| Test | Guarantee |
|---|---|
| `An_emission_during_the_initial_lookup_is_still_delivered` | An index snapshot applied *while* the initial reading is taken is still seen. The lookup pushes the feed on its first call, placing the emission precisely in the gap the old order leaves open. |
| `Unknown_is_an_answer_and_ends_the_subscription` | `Unknown` terminates rather than listening on for every never-onboarded visitor. |
| `A_throwing_initial_lookup_disposes_the_feed_subscription` | The feed subscription is torn down before the fault surfaces. |

## Red-proof

Both mutations were applied to the implementation, built, and run — and they fail **disjoint** tests, so each test pins a distinct defect rather than all three riding on one:

- **read-then-subscribe** (the old `Defer`/`StartWith` order) → fails only `An_emission_during_the_initial_lookup_is_still_delivered`, with the assert message naming the cause: *"published to a feed nobody had subscribed to yet … a visitor in that window waits out the full request budget."*
- **dropping `changes.Dispose()`** from the catch → fails only `A_throwing_initial_lookup_disposes_the_feed_subscription`.

That second mutation is why the throwing case asserts `feed.HasObservers` and not merely the `OnError`. Subscribing first is what *creates* the teardown obligation; asserting the fault alone would pass while a subscription to a hot, process-wide stream leaked and re-invoked the lookup for the life of the cache. Note the ordering revert leaves that test green — with read-first, nothing is subscribed when the throw happens — which is correct, and is the reason it needed its own proof.

Implementation restored byte-identical to `main` afterwards: the diff is the one new test file.

**No What's New entry** — test-only, no user-visible effect.

### DIFF
--- a/test/MeshWeaver.Hosting.Monolith.Test/UntilDeterminedTest.cs
+++ b/test/MeshWeaver.Hosting.Monolith.Test/UntilDeterminedTest.cs
@@ -1,0 +1,111 @@
+#pragma warning disable CS1591
+
+using System;
+using System.Collections.Generic;
+using System.Reactive;
+using System.Reactive.Linq;
+using System.Reactive.Subjects;
+using MeshWeaver.Hosting.AspNetCore.Portal;
+using Xunit;
+
+namespace MeshWeaver.Hosting.Monolith.Test;
+
+/// <summary>
+/// The two temporal guarantees of <see cref="UserIdentityLookup.UntilDetermined"/>, which were the
+/// point of the subscribe-before-read change and had no test (MeshWeaver#2302, finding 10).
+///
+/// <para>Both are pure: a <see cref="Subject{T}"/> for the hot feed and a lambda for the lookup —
+/// no mesh, hub or query subscription, exactly as that method's own remarks promise. That matters
+/// because the window under test is one a live-mesh test cannot hold open.</para>
+///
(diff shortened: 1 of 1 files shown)


rbuergi:
Correction to this PR's body: it calls finding 10 "the last actionable item on #2302". That was wrong — finding 8 (the `limit:10` role truncation) was also still live, and is fixed in #2583. This PR covers finding 10 only.

PR #2586, opened 2026-08-28
Factor AGENTS.md's bulk into on-demand skills (103.9 KB → 41.7 KB)

rbuergi:
## 🚨 Do not merge yet — sequencing

This PR **must merge AFTER core #2568** ("MeshWeaver.AI leaves core"). It rewrites AGENTS.md for the
post-move world: agent/skill node definitions, the thread-submission extensions, the AI Orleans rig
and the `Agent`/`Skill`/`Thread`/`ThreadMessage` node types are described as living in
MeshWeaver.Plugins. Those statements are **not true until #2568 lands**. Where a sentence could not
be made correct for both worlds, the post-merge wording was chosen rather than a hedge — the list is
at the bottom.

It also touches `MigrationWorkloadModelGuard.cs`, which #2568 modifies too. The conflict is one
collection-expression tail and resolves by keeping #2568's removal of the `content` root plus this
PR's addition of the `.claude/skills` root.

## Why

`AGENTS.md` is loaded into **every** agent session's initial context (via `CLAUDE.md`, which is just
`@AGENTS.md`). At ~104 KB that was ~26K tokens of standing cost per session, spent before the model
had seen the question. Most of it was *evidence* — worked examples, incident dates, command
transcripts, tables of past failures — not *instruction*.

| | before (`main` @ 27f1ca74f) | after |
|---|---|---|
| `AGENTS.md` | 103,914 B / 1,081 lines | **42,022 B / 283 lines** (−59.6%) |

## What moved, and what did not

**The rules stay in AGENTS.md.** Every ABSOLUTE keeps its heading, its 🚨 markers and its imperative
stated in full, followed by a pointer to the skill that now carries the detail. A heading-by-heading
diff of old vs new confirms nothing was dropped; the four sub-sections that lost their own heading
(*Test Triage*, *MCP Mutations*, *Satellite CI*, *Key Dependencies*) had their imperatives folded
into the parent section rather than relocated.

A skill index table at the top maps topic → skill, so an agent knows what to load.

**New skills** (`.claude/skills/<name>/SKILL.md`):

| Skill | Receives |
|---|---|
| `worktree` | Git Workflow, the primary-checkout rule, per-repo `strict`, CI flags, the seven false-pass traps |
| `ci` | The gate/skip-trapdoor rule, in-mesh source invisible to `dotnet build`, `cancel-in-progress` on main, satellite-repo reusable workflows |
| `testing` | Testing Guidelines, test triage, reading a CI shard artifact, test bases, DevLogin, node types |
| `mesh-data` | `stream.Update` as the only mutation API, CQRS reads, `.As<T>()`/`.ContentAs<T>()`, the data-access table, Postgres one-schema-per-partition |
| `i18n` | The whole internationalization section |
| `deployment` | The AKS + Container Apps routes, the migration Job, `DbVersionGate`, image verification |

**Extended rather than duplicated** — these already covered the ground, so the missing detail was
merged into them and the AGENTS.md section collapsed:

- **`async`** ← the `SemaphoreSlim` ban and its hub/`IIoPool` alternatives, the `Observable.FromAsync`
  prohibition, the `IIoPool` method table, `PromiseCache` vs a bare `ConcurrentDictionary`,
  `ImpersonateAsHub`.
- **`gui`** ← the thread-submission surface, the `RenderHtml` ban's provenance.
- **`pullrequest`** ← finishing means merged, capability is CREDENTIAL × REPO, review-thread
  resolution, filtering by workflow (the permanently-`queued` Azure Boards suite), the persistent
  `Monitor` subscription, merge coordination and holds.
- **`release`** ← the CD batching window, the reconciler, `SelfUpdate__MinRollInterval`, the
  `main-cd.yml` trigger gate and its three silent consequences, the ACR/AKS image assertions.

## The guard had to follow its subject

`MigrationWorkloadModelGuard` exists because AGENTS.md once told operators to roll the migration as a
Deployment the chart does not define (#1788). Moving the runbook into `.claude/skills/` would have
left it scanning nothing while still reporting green — the exact skip-trapdoor shape this repo's own
CI rule forbids. Its roots now include `.claude/skills/**/*.md`, and that root is **required**:
`Directory.EnumerateFiles` throws if it is renamed, so the guard fails loudly rather than checking
less. `content/` becomes the one *optional* root (it holds only `content/ai` and leaves with the AI
engine), so the scan works before and after #2568.

**Verified negatively, not just positively:** injecting
`kubectl -n memex rollout restart deployment/memex-migration-deployment` into the new deployment
skill makes the guard fail RED and name `.claude/skills/deployment/SKILL.md`. Reverted, green again.

**Naming:** the skill is `deployment`, not `deploy`. `.gitignore`'s Visual Studio `[Dd]eploy/` rule
silently swallows `.claude/skills/deploy/` (re-included only for `/deploy/**`), which would have left
the runbook untracked — and the freshly re-rooted guard scanning nothing on CI. Caught by
`git status` not listing the directory.

## Merged `main` (this PR was `DIRTY`, and a dirty PR runs zero CI)

Three things landed on main while this was in progress, and all three are folded in rather than
reverted (`git diff origin/main...HEAD --stat` shows only this PR's own files):

- **`343642a5b` added `SkillFrontMatterGuard` and quoted `gui`/`playwright`/`storm`.** This PR had
  independently found and fixed the same three unparseable descriptions; the duplicate quoting
  collapsed cleanly on merge, and **all six new skills pass the new guard** (its two tests are in the
  100 below).
- **`b0b57e01c` corrected the i18n drift-guard claim** — the guard DOES run (the plugins repo's
  `RN app + web clients (typecheck + test)` job), it compares VALUES not just keys, and core merges
  first. This PR had carried the stale "runs nowhere" wording into both AGENTS.md and the new `i18n`
  skill; both are corrected, including the skill's front-matter `description`.
- **`f7eaae66c` added the `/sigsegv` skill** — now in the index table.

## Post-#2276 corrections

- `content/ai/Agent/`, `content/ai/Skill/code.md`, `content/ai/Skill/logon-action.md` → the AI
  engine's shipped Agent/Skill nodes in MeshWeaver.Plugins.
- `src/MeshWeaver.AI/HubThreadExtensions.cs` → `MeshWeaver.Plugins/src/MeshWeaver.AI/`.
- "AI agents use plugins (MeshPlugin, LayoutAreaPlugin)" → the AI engine is a module hosted in
  MeshWeaver.Plugins.
- The node-type list splits: `Agent`/`Skill`/`Thread`/`ThreadMessage` are declared by the AI module,
  not `AddGraph()`.
- `OrleansTestBase` (in `MeshWeaver.AI.Orleans.TestBase`, moving) is distinguished from
  **`MeshWeaver.Hosting.Orleans.TestBase`, which stays in core** — the two were being conflated.
- `DocumentationLinkIntegrityTest` is noted as travelling with the Agent/Skill partitions it also
  validates.
- The "'it's only markdown' is not a check" example no longer cites `content/ai/**`; it cites the
  core-resident embedded docs and the guards over them, plus `.claude/skills/**` now that the
  migration guard scans it.
- Dropped as already-stale: `src/MeshWeaver.Blazor.Portal/Chat/ThreadTokenChip.razor.cs`, cited as a
  worked example — `src/MeshWeaver.Blazor*` no longer exists in this repo. The rule it illustrated is
  kept, described generically.

**Statements that cannot be true in both worlds** (post-merge wording chosen, flagged as requested):
the Documentation section's "Agent and Skill node definitions ship with the AI engine, which lives in
`MeshWeaver.Plugins`", the Architecture Overview's placement of the AI engine, and the Testing
section's split of the Orleans test bases. Until #2568 merges, all three describe the intended state
rather than the current one.

## Verification

```
dotnet build test/MeshWeaver.Documentation.Test/MeshWeaver.Documentation.Test.csproj -c Release -warnaserror
  → Build succeeded. 0 Warning(s) 0 Error(s)   (77 s — a real compile, not an up-to-date no-op)
dotnet test test/MeshWeaver.Documentation.Test -c Release --no-build --logger trx
  → Passed! Failed: 0, Passed: 100, Skipped: 0
```

Covers `MigrationWorkloadModelGuard` (all three facts), `SkillFrontMatterGuard` (both),
`DocumentationLinkIntegrityTest` and `WhatsNewEntryIntegrityTest`. The build was re-run **after**
adding the What's New entry and the embedded resource confirmed present in the rebuilt DLL — a
`--no-build` run would have validated a stale copy. Every non-URL link target in AGENTS.md and every
`../../../` link in the new skills was resolved against the filesystem.

## Note on the size target

The brief asked for ≲30 KB. It landed at 42 KB because what remains is essentially all imperative:
reaching 30 KB would mean deleting rules rather than evidence, and the brief's own tie-breaker is to
keep a rule when in doubt. About 3.5 KB of what remains is the skill-index table and the
`Full reference:` pointers — the cost of the indirection itself. Deliberately kept in AGENTS.md
despite looking movable: the *No band-aids* bullet list (five distinct prohibitions, no owning
skill), the cold-observable ❌/✅ block (the highest-frequency mistake in the repo, hard to state
without the two lines), and the `Development Commands` block plus the `Data Access Patterns` table
(pure reference, but consulted constantly — a skill round-trip for "how do I run one test project"
costs more than the bytes).


rbuergi:
Superseded by #2588, which carries this change plus the other two (one CI run instead of three). The only conflict — MigrationWorkloadModelGuard's `content/` root — is resolved there keeping BOTH the optional root and the note that the subject moved. Branch kept; reopen if #2588 has trouble.

PR #2605, opened 2026-08-28
Heal durable self-typed NodeType declaration rows at startup

### LINKED ISSUE #2425: UserIdentityCache resolves the User TYPE node instead of a user instance: As<User> gets NodeTypeDefinition — ~600/h errors on prod
**Loki triage 2026-08-26 (~17–20Z), both prod portals:**

```
fail: MeshWeaver.Hosting.AspNetCore.Portal.UserIdentityCache[0]
      As<User> for User: value is MeshWeaver.Graph.Configuration.NodeTypeDefinition (MeshWeaver.Graph), not convertible
```

Counts over 3 h: **memex 1,790**, **memex-cloud 518** (kubectl grep -c on the deployments; Loki query capped at its limit). That is ~10/min sustained on memex — this is a hot path, not an edge case.

The message says it outright: the cache asked for a `User` and got the **`User` NodeType definition node** — i.e. the lookup resolved the TYPE's own node (path `User`) where a user INSTANCE was intended. `As<User>` correctly refuses the differently-named type and returns null (the probe-dispatch contract), so whatever consumes this silently proceeds without the user — worth knowing what that degrades (identity fallback? anonymous rendering of a signed-in view?).

Fix direction: find the `UserIdentityCache` read path and its key/path construction — a bare `User` path where `{partition}/User/{id}` (or a query-shaped lookup) was meant. Per the As-degradation rules, the durable fix is the right path/hub, not tolerating the wrong node. Possibly related to the uncapped-read findings batch on `Memex.Portal.Shared` (#2302) only by file adjacency — this is a distinct live defect.

Found by the Loki error-triage pass; the loop is being installed as `Hosting/Agent/loki-error-triage` on the mesh.

### LINKED ISSUE #2506: UserIdentityCache: As<User> resolves the NodeTypeDefinition named 'User' — identity lookups failing in bulk
Prod, 2026-08-27 (image gen ci.6096): `fail: MeshWeaver.Hosting.AspNetCore.Portal.UserIdentityCache — As<User> for User: value is MeshWeaver.Graph.Configuration.NodeTypeDefinition (MeshWeaver.Graph), not convertible` — 599 occurrences in 4h on memex.systemorph.com, 69 on memex.meshweaver.cloud. The cache's lookup for a user resolves to the NodeType DEFINITION node named 'User' instead of a user instance — a name-collision in the lookup path. Whatever consumes the cache (display identity, zones, roles) is degraded on every miss.

https://claude.ai/code/session_01QVbVvrHeU1gm8oq4PuMVjF

rbuergi:
Fixes #2425. Fixes #2506.

## The defect that survived two fixes

`UserIdentityCache` logged `As<User> for User: value is NodeTypeDefinition (MeshWeaver.Graph), not convertible` ~600 times/hour on memex.systemorph.com (and ~70/4h on memex.meshweaver.cloud) — on image `ci.6096`, which already carried BOTH prior fixes:

- **#2245** retyped the three self-typed built-in declarations (`User`, `VUser`, `Partition`) in their STATIC registrations (`NodeType = "NodeType"`), and
- **#2378** added `NodeTypeDeclarationSelfTypingValidator`, refusing any NEW write that re-introduces the collision.

Neither touches a row that was **persisted** self-typed before they landed. Verified live: `search nodeType:User` on prod still returns `{path: "User", nodeType: "User", version: 2, lastModified: 2026-06-02}` — the durable declaration fossil, answering the user-directory query beside real accounts. Reads are bad-data tolerant by design, and the row is unreachable through the serve path (the static claim shadows it — the #2534 mechanism), so "until it is next written" was **never**: the defect had moved from the code into the data, and no code-only guard could end it.

## The fix — heal the data, once, at startup

**`SelfTypedDeclarationDurableRepair`** (hosted service, registered by `AddGraph` beside the validator) runs once at startup, at the storage seam where `LegacyUserPartitionRepair` already writes (the only seam that can reach a static-shadowed row):

- one batched `IStorageAdapter.ReadMany` over the statically-registered declaration paths (the same multi-path probe the URL resolver runs constantly — absent partitions answer "absent", never fault: the PG routing adapter's reads catch 42P01 by contract);
- rewrites **exactly** the rows matching the write guard's own predicate — extracted as `NodeTypeDeclarationSelfTypingValidator.IsSelfTypedDeclaration` (public, shared), so what the guard refuses and what the repair heals can never drift apart;
- the rewrite is #2245's correction applied to the row: `NodeType = "NodeType"`, version minted forward, content/identity untouched;
- everything else is left alone — already-correct declarations, real user instances, package roots typed as an unrelated type (the UWDeepfield shape). A healthy store reads one batch and writes nothing;
- fire-and-forget and failure-tolerant like `InstalledPackageRepairService`: a skipped pass heals on the next boot; concurrent replicas write the same value under the monotonic-version guard.

The adapter change feed fires on the heal, so live `nodeType:User` subscriptions (`UserIdentityCache`'s directory query) converge without a restart.

## Repro test — red on main for the right reason

`SelfTypedDeclarationDurableRepairTest` boots a mesh over a store **already carrying the fossils** (seeded straight into the adapter before boot — exactly how production got them):

- `FossilSelfTypedDeclarationRows_AreRetypedAtStartup` — the `User` and `VUser` fossils flip to `NodeType = "NodeType"` (content preserved, version forward); an already-correct `Partition` declaration row and a real user instance are untouched.
- `TheUserDirectory_NoLongerReturnsTheFossilDeclaration` — `UserIdentityCache.DirectoryQuery` (verbatim) returns the real account and NOT the declaration; every returned row reads as a `User`.

On main both fail with `TimeoutException` waiting for a heal that never happens (`Failed: 2, Passed: 0`); with the fix, `Passed: 2` in <1s.

## Verification

- `MeshWeaver.Graph` and `MeshWeaver.Hosting.Monolith.Test` build clean `-c Release -warnaserror`.
- Adjacent guards stay green: `NodeTypeDeclarationSelfTypingValidatorTest` (9 — the predicate extraction is behavior-preserving), `NodeTypeDeclarationSelfTypingTest` (2), `WhatsNewEntryIntegrityTest`.
- What's New entry included (`Category: Fix`).

### DIFF
--- a/src/MeshWeaver.Documentation/Data/WhatsNew/2026-08-28-user-directory-declaration-repair.md
+++ b/src/MeshWeaver.Documentation/Data/WhatsNew/2026-08-28-user-directory-declaration-repair.md
@@ -1,0 +1,15 @@
+---
+Name: User lookups no longer trip over the User type definition
+Category: Fix
+Description: A leftover database row that made the "User" type definition answer user searches is now repaired automatically at startup, so identity lookups stop failing in bulk.
+Icon: Sparkle
+Order: -20260828
+---
+
+# User lookups no longer trip over the User type definition
+
+On long-lived installations, the built-in "User" type definition had once been saved to the
+database claiming to be a user itself. Every user lookup then received that definition row
+alongside real accounts, which degraded identity resolution (display name, time zone, language)
+and produced a steady stream of background errors. The platform now detects and corrects such
+rows automatically at startup, so the user directory returns only real accounts.
--- a/src/MeshWeaver.Graph/Configuration/GraphConfigurationExtensions.cs
+++ b/src/MeshWeaver.Graph/Configuration/GraphConfigurationExtensions.cs
@@ -240,6 +240,17 @@ public static class GraphConfigurationExtensions
                 // CodeNodeSegmentNameValidator below.
                 services.AddScoped<INodeValidator, Security.NodeTypeDeclarationSelfTypingValidator>();
 
+                // The RETROACTIVE half of the same fix (#2425/#2506): the validator above only
+                // refuses NEW writes, and #2245 only fixed the STATIC registrations — a
+                // declaration row PERSISTED self-typed before either landed keeps answering its
+                // own instance query forever (prod: ~600 UserIdentityCache errors/hour on an
+                // image already carrying both). This startup pass retypes such durable rows at
+                // the storage seam (the serve path cannot reach them — the static claim shadows
+                // the row, #2534), using the validator's own predicate so guard and repair can
+                // never drift apart. One batched read per boot; writes nothing on a healthy
+                // store.
+                services.AddHostedService<Security.SelfTypedDeclarationDurableRepair>();
+
                 // Keeps the batch bake's "the union is every Code node" claim true BY
                 // CONSTRUCTION (issue #1235): a Code node named after a code-table routing
                 // segment (Source/Test) lands in the code table but has a namespace that
--- a/src/MeshWeaver.Graph/Security/NodeTypeDeclarationSelfTypingValidator.cs
+++ b/src/MeshWeaver.Graph/Security/NodeTypeDeclarationSelfTypingValidator.cs
@@ -49,8 +49,11 @@ namespace MeshWeaver.Graph.Security;
 ///
 /// <para>Bad-data TOLERANCE is unchanged: this guard never rejects a READ. An existing row that
 /// already carries the collision keeps loading (degraded, tolerated by
-/// <see cref="ObjectAsExtensions.As{T}"/>) until it is next written; only a NEW create/update that
-/// would (re)introduce the collision is refused.</para>
+/// <see cref="ObjectAsExtensions.As{T}"/>); only a NEW create/update that would (re)introduce the
+/// collision is refused. Rows PERSISTED with the collision before this guard existed are healed at
+/// startup by <see cref="SelfTypedDeclarationDurableRepair"/> (#2425/#2506) using
+/// <see cref="IsSelfTypedDeclaration"/> — the shared predicate that keeps the two halves exact
+/// mirrors.</para>
 /// </summary>
 public sealed class NodeTypeDeclarationSelfTypingValidator : INodeValidator
 {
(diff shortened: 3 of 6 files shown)


rbuergi:
## Root cause of the CI red: this hosted service was rooting the mesh hub for the process lifetime

Picked this up from the session that filed it (its triage subagent died mid-run on a 429, so there was no live hand). Handing the finding forward, because the diagnosis is the useful part.

`MeshHubDisposalLeakTest.MeshHub_IsCollected_AfterMeshAndServiceProviderDisposal` failed on the first run, and its ClrMD GC-root analysis named the chain outright — no inference needed:

```
IHostedService[]
  -> MeshWeaver.Graph.Security.SelfTypedDeclarationDurableRepair
    -> MeshWeaver.Messaging.MessageHub  [Address=mesh/v6J4wXmBGUWYem-6gzlChw]
```

**The generic host keeps its `IHostedService[]` for the whole process lifetime**, so a hub held in a field on a one-shot startup service outlives every mesh it was built for. That is not a one-mesh problem — it ACCUMULATES: a process that builds many meshes (this test suite; a portal recycling silos) keeps all of them, and everything they own, alive through that array.

### The fix — both fields released the moment they stop being needed

- **The hub** goes on the first line of `StartAsync`, via `Interlocked.Exchange`. That makes the field null before the method returns *however* it returns — including the two early exits (no storage adapter, no declarations) — and makes the pass one-shot for free.
- **The sweep handle** goes when the sweep terminates. A `SingleAssignmentDisposable` is published to the field **before** it is armed, so on a synchronously-completing store `Finally` nulls the field first and arming an already-disposed gate disposes the inner subscription on the spot. Either order leaves the instance referencing nothing.
- `StopAsync` and the `Finally` share one `Release()`, because the sweep completing and the host stopping race by construction.

### 🚨 Pinned locally, not only by the leak test

`MeshHubDisposalLeakTest` is the outcome check, but its ClrMD snapshot-attach is unavailable on macOS — it `Assert.SkipWhen`s as *inconclusive* there (#674). So on every developer machine in this repo the only signal is CI, hours later on a Linux runner. `SelfTypedDeclarationRepairRetainsNothingTest` asserts the **mechanism** deterministically and locally:

1. the hub field is null after `StartAsync` (the direct edge of the chain), with the premise asserted first so it cannot pass vacuously;
2. a second `StartAsync` is an inert no-op;
3. the subscription handle is dropped when the sweep terminates — waited on the condition, never a sleep;
4. `StopAsync` is safe before any start and twice.

**Red-proof:** reverting the exchange to a plain field read fails two of the four, with *"...but found MeshWeaver.Messaging.MessageHub"*. All 6 tests across both classes pass on the fix, and `MeshWeaver.Graph` + the test project build `Release -warnaserror` clean.

rbuergi:
## The red is #890, not this diff — and it is the second recurrence since that issue was closed

Shard 1 failed **seven** tests on run `33179258773`. Every one is a compile, and nothing else in the shard failed:

`NodeTypeAlcSharedWithInstancesTest`, three `NodeTypeReleaseGateTest` cases, two `MeshNodeCompilationIntegrationTest` cases, and `CompileSourceSnapshotWedgeTest` — the "one Emit poisons the process, every LATER compile NREs" shape #890 is named for.

The attribution is not mine to argue; the instrumentation makes it. The emit canary's verdict, verbatim:

```
canary=BELOW-ROSLYN
  shared:   THREW NullReferenceException at NamedTypeSymbol.…get_ContainingTypeDefinition
  pristine: THREW NullReferenceException at NamedTypeSymbol.…get_ContainingTypeDefinition
⇒ a trivial compilation with BRAND-NEW references and freshly parsed source cannot emit either,
  and BOTH legs died in the same frame ⇒ nothing about the reference set explains this;
  the broken state is below Roslyn (CLR heap / JIT / GC)
```

The **pristine** leg — brand-new references, freshly parsed source — died in the same frame as the shared one. That is exactly the discriminator built to separate "a PR changed the reference set" from "the process is poisoned", and it says the latter.

This PR adds a startup hosted service that retypes durable self-typed NodeType **declaration rows** at the storage layer. It has no path to Roslyn's emit.

**Reopened Systemorph/MeshWeaver#890** with this occurrence: it was closed 2026-08-18, recurred 2026-08-24 on main (`32784957424`, shard 4) where the note said *"the canary's dump-capture + tiering-off advice stands if it recurs"*, and has now recurred again. Leaving it closed means every occurrence gets re-diagnosed and pinned on whichever PR was running — which is what the canary exists to prevent.

Merged `main` for a fresh run (AGENTS.md sanctions this for a red the diff cannot reach). Revert-checked: the diff is still the same 6 files. **If the same seven compile tests fail again, that is #890 and not this change** — the fix belongs there, and this PR should not be held for it.

Issue #99, opened 2026-05-11
AI Chat in MeshWeaver stops working after 3-4 interactions


sierragolflima:
The AI Chat in MeshWeaver stops working after 3-4 prompt/response interactions.

The chat window shows 'generating response' but nothing happens.

The 'Stop' button to stop this process does not work.

Issue #2622, opened 2026-08-28
A plugins-side change to the portal HOST (appsettings, csproj) cannot mint an image until core moves — CD is keyed on core HEAD only


rbuergi:
## Observed 2026-08-28

Plugins#814 fixed fresh-install engine activation by editing `src/Memex.Portal.Distributed/appsettings.json` (`Modules:Assemblies`) — a file that ships **in the portal image**. It merged at 14:54Z. The newest image, `ci.6327`, was built at ~14:4x from core `bf1ae3a` and therefore predates the fix. Nothing will build a newer image until **core** main moves:

- `main-cd.yml` checks out `Systemorph/MeshWeaver.Plugins@${{ vars.MW_PLUGINS_REF || 'main' }}` for the host (line ~747), but its trigger and its image-set key are **core's HEAD SHA**.
- The hourly reconciler asks "does core HEAD have a complete image set?" — yes → `nothing to reconcile`.
- Plugins' own `portal-ai-image.yml` is `workflow_dispatch`-only and has never run (its header explains why its version scheme must not reach the fleet).

So a host-affecting plugins merge is **invisible to delivery** until an unrelated core merge happens to rebuild. Today that stalled the deploy tail: the FOLLOW_RELEASE wave's bake failed on a fresh mesh (engine never landed), the fix merged in plugins, and the image that would carry it has no producer.

## Options (not deciding here)

1. Core CD subscribes to plugins' main push too (a `repository_dispatch` from plugins → core `main-cd`, with the image set keyed on the PAIR of SHAs).
2. Plugins' `portal-ai-image.yml` becomes the sanctioned producer for host-only changes (its version-scheme concerns in the header would need solving first).
3. Accept the coupling and document "a host change in plugins ships on the next core merge" — cheapest, but it is what bit today.

Related: Systemorph/MeshWeaver#2276 (the move that made the host live in plugins), plugins#814, the two-webhook-chains note.

rbuergi:
[...]

Issue #833, opened 2026-08-06
Auto-discover new modules in a configured plugin repo and start syncing them — a per-repo setting, no per-Space interaction


rbuergi:
## The problem this solves

On our instances modules arrive through **per-Space `{Space}/_GitSync` entries**, not the plugin catalog: `memex.meshweaver.cloud` carries **37** sync configs and **zero** install records. The consequence is that a module with no entry is simply **not on the mesh** — and nothing says so. Two live examples today: `Planning` (merged this morning) and `Ifrs17` are both absent from memex purely because nobody created their entries, while their nine sibling modules from the same repo are present.

Worse, the entry cannot be made by hand: a `_GitSync` needs its Space to exist, and creating a Space makes the creator **Admin on a system-owned partition** — forbidden by the access rules, and re-minted by every human-run sync. So today "add a module to an instance" has no safe self-service path at all.

## Requested feature (maintainer, 2026-08-06)

> Detect new plugins on git automatically and start syncing without any interaction — as a **setting on the plugin repo**, in settings.

Make the **repo** the unit of configuration instead of the Space:

```
Plugin repo setting (per configured source)
  ├─ repo + branch                     (exists today as a catalog source)
  ├─ ☑ auto-discover new modules       ← new
  └─ ☑ auto-sync discovered modules    ← new
```

When enabled, on each build event (or poll) for that repo:
1. **Discover** its top-level modules — the same node-repo layout the catalog already understands (a module folder with a root node; `manifest.lock` gives its content identity).
2. For any module with **no Space yet**, provision it **as SYSTEM** through the existing install/provision path — never as the acting user, so no personal grant is minted.
3. Write its `{Space}/_GitSync` (repo + subdirectory + branch) and run the first `update`.
4. Recompile the affected types, in dependency order, exactly as the existing update path does.
5. Record what was created, so an instance can answer "what appeared, and when?".

Removed modules should **not** be auto-deleted — surface them as orphaned rather than destroying content.

## Interactions to respect

- **Permissions**: gate discovery/creation by the free-vs-commercial rule in #830 — an unpriced module may be auto-provisioned; a priced one must not appear without an entitled admin.
- **Update policy**: newly created records inherit the instance default (auto), with the per-record Auto/OnDemand/Ignore override.
- **Idempotence**: re-running must be a no-op; the module's content hash is the identity.
[...]

PR #2628, opened 2026-08-28
ci(bake): external modules ride --module — same-run artifacts and registry-fetched bundles

rbuergi:
## Why (the fleet freeze, third diagnosis, this one from the code)
Since #2276 the tester image ships **no modules** (`TesterModules.ImageShipped` is empty by design — "a node repo that needs the AI node types mounts its OWN build with `--module`", #2541), and **no workflow passes `--module`**. So no node repo's bake or gate can compose `MeshWeaver.AI.dll`: plugins `MyAi/Panel`, SocialMedia/Reinsurance `Store/*` (via `Store/Installer/Source/Localizer.cs` → `AiSettings`) and `Edu/PromptCell` (`ThreadMessage`) fail CS0103/CS0246 on every post-exit wave, no publication seals for the new identities, and `ReleaseAvailability.IsUpdatable` correctly holds the fleet (both portals pinned on ci.6263 all day). My earlier attempts (#827 ordering — closed; #829 `requires` — kept for install semantics only) treated symptoms.

## What
`node-repo-publish-bake.yml` gains:
- `module-artifacts` — a download-artifact pattern for module bundles built by the SAME run (plugins: `needs: modules` + `module-bundle-MeshWeaver.AI`).
- `registry-modules` + `registry-url` (input) + `registry-key` (secret) — fetch the named packages' bundles from the registry (`GET /api/plugins/bundles/index.json` → advertised version → `GET /{plugin}/{version}`). The module lane is **floor-gated, not identity-gated** (`PluginBundleClient.AdoptModule` doc), so the registry's current bundle is the right bytes for any image satisfying the module's `minMeshVersion`.
- Preflight asserts `registry-url`/`registry-key` RED when `registry-modules` is set — no skip-trapdoors.
- The unpack step refuses ambiguity (no manifest `module.assemblyName` and ≠1 dll) and requires `meshweaver/modules/<Name>.dll`; both docker runs mount `/ext` and pass `--module /ext/<Name>/<Name>.dll`; the tester already validates each path and prints `external module: …` per entry.

## Wiring (follow-up PRs, dependency order)
1. this PR (platform)
2. plugins ci.yml: `publish-bake` → `needs: [... , modules]` + `module-artifacts: module-bundle-MeshWeaver.AI`
3. SocialMedia/Reinsurance: `registry-modules: AI` + `registry-url: ${{ vars.MW_REGISTRY_URL }}` + `registry-key: ${{ secrets.MW_REGISTRY_KEY }}` + bump their `uses:` pin to a ref containing this change (their #99/#109 keep staging AI **content** for the package install; the DLL comes from here)

## Not covered here
The PR-lane gates (`node-repo-gate.yml`, plugins' inline `test-repos`) still compose nothing — they narrow by affected closure, so full-gate reds only appear on release waves; same mechanism can be added there next. If the registry serves a shape without `meshweaver/modules/`, the step fails RED naming the file — adjust then, never skip.

Refs: Plugins#831 (the AI-bundle test flake that can still red a wave), MeshWeaver#2235 (two-phase broadcast), Plugins#826.

### DIFF
--- a/.github/workflows/node-repo-publish-bake.yml
+++ b/.github/workflows/node-repo-publish-bake.yml
@@ -219,6 +219,36 @@ on:
         required: false
         type: string
         default: 'main'
+      module-artifacts:
+        description: >-
+          Artifact-name pattern (actions/download-artifact `pattern`) of module bundles UPLOADED
+          BY THIS RUN (each a *.module.nupkg) whose module DLLs the bake and the gate must
+          compose. The tester image ships NO modules since MeshWeaver#2276
+          (TesterModules.ImageShipped is empty by design; a repo mounts its own build with
+          --module, #2541) — so a repo whose NodeType source binds a module's types names that
+          module here or in registry-modules, or its bake fails CS0103/CS0246 on every wave.
+          Empty = none.
+        required: false
+        type: string
+        default: ''
+      registry-modules:
+        description: >-
+          Package ids (whitespace-separated, e.g. `AI`) whose module bundle to fetch from
+          registry-url and compose into the bake/gate with --module — for a repo that stages the
+          package's CONTENT but does not build its module. The module lane is FLOOR-gated, not
+          identity-gated (PluginBundleClient.AdoptModule), so the registry's current bundle is
+          the right bytes for any image satisfying the module's minMeshVersion. Requires
+          registry-url and the registry-key secret; asserted RED in preflight, never skipped.
+        required: false
+        type: string
+        default: ''
+      registry-url:
+        description: >-
+          Base URL of the registry serving /api/plugins/bundles (the caller's MW_REGISTRY_URL
+          variable). Needed only when registry-modules is set.
+        required: false
+        type: string
+        default: ''
     secrets:
       acr-username:
         description: Registry user able to pull the test image.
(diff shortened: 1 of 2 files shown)


rbuergi:
Auto-merge intentionally disarmed for ORDERING, not for a defect: this PR's merge is the next core commit, and its CD run mints the image whose host build checks out plugins@main — that image must carry Systemorph/MeshWeaver.Plugins#833 (Blazor.Views baseline activation; without it every page on a fresh/rolled portal renders raw JSON). Re-arming the moment #833 is merged. (#2622 is why a plugins-side host fix cannot mint an image on its own.)

rbuergi:
Shard-3 red attributed and fixed in-branch, not re-rolled: `PlatformBakeLaneGuard.PlatformBake_NeverAdoptsAnotherBuildsBundles` flags any workflow with `publish-bake-bundles.sh` + `download-artifact`. #1725's invariant is **cross-run** adoption; `actions/download-artifact` without `run-id:` sees only the current run — exactly the property `module-artifacts` relies on (and the module bytes go to `/ext` for `--module`, never into the published $BAKE). The guard now keys on `gh run download` or `download-artifact`+`run-id:`; 12/12 green locally.

PR #2644, opened 2026-08-29
Join hub disposal before a harness proceeds — the SIGSEGV rule, plus a ratchet

rbuergi:
## The failure mode: SIGSEGV mid-run, from use-after-dispose

`IMessageHub.Dispose()` **starts** the disposal state machine and returns. The action blocks drain, the hosted hubs tear down, the sync streams unregister and the registrants run **afterwards, on other threads**. A test base, an xUnit fixture or a run harness that disposes and then proceeds is running concurrently with all of that — and the next thing a harness always does is one of the three that live teardown cannot survive:

* dispose the service scope — a continuation resolves from a dead Autofac scope;
* unload a collectible node ALC — a live thread dereferences freed metadata, a native use-after-unload **SIGSEGV**;
* return to xUnit, so the next test class starts on top of the previous one's teardown.

The result is not a red test. It is a dead process, naming nothing. **MeshWeaver.Plugins run 33236823482, job 99060770662**: the plugin gate exited **139** mid-run while installing packages, immediately after a burst of `[SYNC_STREAM] Not setting … — stream is disposed` and `Stream …: resubscribe failed … TargetInvocationException`.

The instructive part is *where* it was. The gate's **final** teardown already joined the mesh correctly. What did not join was the render **client**, disposed two statements earlier. The file reads as disciplined right up until you check which hub the join actually names.

## The rule

> In a teardown or run-boundary path, every hub disposal joins **that hub's** `DisposalCompleted` before the caller proceeds.

`MeshWeaver.Messaging.HubDisposalJoin` (new) is the small form for every hub a harness owns:

| | |
|---|---|
| `await hub.DisposeAndJoinAsync(report, timeout?)` | an async teardown (`DisposeAsync`, `IAsyncLifetime`). **Suspends** the caller rather than parking its thread, so it cannot self-deadlock against the hub's own scheduler. |
| `hub.DisposeAndJoin(report, timeout?)` | a genuinely synchronous run boundary — a harness's `IDisposable.Dispose`, a tool's exit path. The sanctioned block-join, and nothing else; never from hub- or view-reachable code. |

`MeshTeardownExtensions.TeardownAsync` / `WaitForDisposalAndIoDrainAsync` stay the richer form for a **mesh root** (they also cancel+join the `IIoPool` leaves and quiesce the `AsyncDisposeQueue`).

Both new helpers are **bounded** and both **say it** when the bound is hit — a silent hang is not an improvement over a crash, and a budget that fires is a hub that is not finishing, not a number to raise. Neither splices a `Timeout` into the signal: the bound is a `CancellationTokenSource` (async) or a `ManualResetEventSlim` wait (sync), and the subscription stays attached afterwards, so a fault arriving after the wait gave up is still reported instead of becoming the unobserved exception xUnit v3 escalates to a Catastrophic failure (#2301/#2488). The report sink is wrapped so a logger or `ITestOutputHelper` whose scope is already gone cannot become the thing that fails a teardown.

## Enumeration — every hub disposal in the covered surface

The surface is `src/*.TestBase`, `src/*.Fixture`, `tools/`, and fixtures under `test/`. Receiver-anchored scan (the join must name the hub being disposed):

| site | joined before? | verdict |
|---|---|---|
| `src/MeshWeaver.Fixture/HubTestBase.cs:182` `Mesh.Dispose()` | ✅ `Mesh.WaitForDisposalAndIoDrainAsync` | already correct |
| `src/MeshWeaver.Hosting.Monolith.TestBase/MonolithMeshTestBase.cs:1445` `Mesh.Dispose()` | ✅ `WaitWithProgressAsync` (its own wait on `Mesh.DisposalCompleted`) | already correct |
| `src/MeshWeaver.Hosting.Monolith.TestBase/MonolithMeshTestBase.cs:1253` `client.Dispose()` | ❌ | **FIXED** |
| `src/MeshWeaver.Hosting.Orleans.TestBase/OrleansTestBase.cs:132` `client.Dispose()` | ❌ | **FIXED** |
| `src/MeshWeaver.Hosting.Orleans.TestBase/SharedOrleansFixture.cs:211` `reg.Hub.Dispose()` | ✅ `reg.Hub.DisposalCompleted.ObserveCompletion` | already correct |
| `src/MeshWeaver.Hosting.Orleans.TestBase/SharedOrleansFixture.cs:295` `hub.Dispose()` | ❌ | **FIXED** |
| `tools/MeshWeaver.PluginTester/PluginGateRunner.cs:843` `Client.Dispose()` | ❌ | **FIXED** — the incident above |
| `tools/MeshWeaver.PluginTester/PluginGateRunner.cs:851` `Mesh.Dispose()` | ✅ (spliced `.Timeout`) | moved onto the helper |

**Can the process/test exit before teardown finishes?** Yes at all four:

* **`MonolithMeshTestBase.DisposeTestClients`** — on the `SharesMeshAcrossTests` path this is the *last* thing that touches the mesh: `DisposeAsync` returns straight after it and the next `[Fact]` begins on a mesh whose client teardowns are still draining. On the non-shared path the very next statements stop the hosted services and dispose the Mesh, which is the same race one level up. Now `async Task` and joins each client, bounded by the existing `DisposeTimeout`, reporting through `FileOutput` (safe after the test has been reported; a raw `ITestOutputHelper` throws there).
* **`OrleansTestBase.DisposeAsync`** — the next statement hands the cluster to `OrleansClusterDisposal.DisposeInBackground`, stopping the silos those client teardowns are still talking to. Its `catch { /* best-effort */ }` also discarded disposal faults outright; they are now written out.
* **`SharedOrleansFixture.CleanupSiloHubsWithPrefix`** — silo-side hosted hubs whose teardown deactivates the owning grain (`MessageHubGrain`); returning early lets the next test's grain calls race a deactivation in flight on a hub it can still address. Synchronous block-join (10s), reported through a logger captured **before** disposal began.
* **`PluginGateRunner.GateMesh.Dispose`** — its own doc comment already claimed "dispose the hubs and JOIN their disposal" (plural); the client half was never implemented. The mesh half moved onto the same helper, dropping the `.Timeout(30s).Catch(...)` spliced into the signal.

### Deliberately out of scope

**Product code.** 15 hub disposals across `src/` (`SynchronizationStream` unsubscribing its sync hub, `Workspace` evicting a client subscription, `OrleansRoutingService` releasing a pod claim, `Activity` dropping its hosted hub, `MeshOperations`' probe hub, …). Every one is a *live-system* disposal: nothing is about to tear the scope down or exit, the state machine finishes on its own, and blocking a hub thread to watch it would be the deadlock the reactive rule forbids.

**Test bodies under `test/`.** A per-test client disposed inside a `[Fact]` is bounded by the class teardown: the base class joins the mesh in `DisposeAsync`, and the mesh's disposal joins its hosted hubs collectively. 10 such sites exist; none can outlive the class.

## The ratchet

`test/MeshWeaver.Documentation.Test/HubDisposalJoinRatchetGuard.cs` + `test/HubDisposalJoinSites.allow`, in the idiom this repo already uses (`BlockingBridgeInTestRatchetGuard`, `ImpersonationScopeSiteRatchetGuard`) and sharing their `SourceScan` masker so prose quoting the shape is not counted as a call site.

**The allow file is seeded EMPTY**, and `TotalBudget = 0`. There was no debt left to seed it with — all four sites are fixed here — and an allow list seeded with today's debt makes the debt permanent. Adding a line to it fails the PR.

Two design points worth the review:

* **The join must NAME the hub being joined.** An unanchored scan reads a *neighbouring* hub's join as this one's — which is exactly the plugin-gate miss. `AnchoredJoins` are matched as `<receiver>.<marker>`, against the full dotted chain (`reg.Hub.DisposalCompleted`) as well as the tail.
* **Two marker lists are documented as LOAD-BEARING**: `HubReceiver` (a hub variable named after none of `hub`/`mesh`/`client` is invisible — the identity-scope ratchet learned this the hard way) and `UnanchoredJoins` (today exactly one entry, `WaitWithProgressAsync`, `MonolithMeshTestBase`'s private wait on `Mesh.DisposalCompleted`).

### Mutation evidence — the guard is not vacuous

Three mutations, each run against the built guard, then restored:

**1. Revert the plugin gate's client to the un-joined shape** → red:

```
NEW SITE   tools/MeshWeaver.PluginTester/PluginGateRunner.cs (1) — the hub(s) disposed here are never joined.
Sites:
  tools/MeshWeaver.PluginTester/PluginGateRunner.cs:852  Client.Dispose()
Failed!  - Failed: 1, Passed: 2
```

**2. Reconstruct the EXACT pre-fix file** — bare `Client.Dispose()` with the *mesh's* `Mesh.DisposalCompleted…Wait()` two statements below it, i.e. the state that shipped and crashed → still red, and it flags **only** the client, not the correctly-joined mesh below it. This is the receiver-anchoring proof:

```
Sites:
  tools/MeshWeaver.PluginTester/PluginGateRunner.cs:852  Client.Dispose()
Failed!  - Failed: 1, Passed: 2
```

**3. Revert `OrleansTestBase.DisposeAsync`** to `try { client.Dispose(); } catch { }` → red:

```
NEW SITE   src/MeshWeaver.Hosting.Orleans.TestBase/OrleansTestBase.cs (1)
Sites:
  src/MeshWeaver.Hosting.Orleans.TestBase/OrleansTestBase.cs:141  client.Dispose()
Failed!  - Failed: 1, Passed: 2
```

**Restored → `Passed! - Failed: 0, Passed: 3`.**

Two further non-vacuity facts are pinned *as tests*, so they cannot rot silently:

* `TheClassifierTellsAJoinedDisposalFromABareOne` proves the discrimination on **synthetic** source (bare vs. joined vs. via-helper vs. the neighbour's-join trap vs. dotted receiver vs. non-hubs `hostedHubs`/`clientHost`/`_portalProcess` vs. a commented occurrence) — independent of the tree, so it stays meaningful once every real site is clean.
* `TheScannerStillSeesTheHarnessSurface` proves the **walk** still selects files (≥10) and still sees hub teardowns in them (≥6). It counts `DisposeAndJoin` calls **as well as** literal `.Dispose()` on purpose: counting only the literal form would make the assertion *fall* as sites get converted — a gate that goes red for doing what the gate asks is how a ratchet gets deleted.

## Tests actually run

All on this branch, built first (full solution restore+compile, 0 warnings / 0 errors under warnings-as-errors):

| suite | result |
|---|---|
| `MeshWeaver.Documentation.Test` (holds the new guard) | **100 passed** |
| `MeshWeaver.Messaging.Hub.Test` (the helper's home assembly; the disposal suites) | **317 passed** |
| `MeshWeaver.Hosting.Orleans.Test` (`OrleansTestBase` + `SharedOrleansFixture`) | **177 passed** |
| `MeshWeaver.Content.Test` (exercises the `ShareMeshAcrossTests` path) | **223 passed, 1 skipped** |
| `MeshWeaver.PluginTester.Test` (the gate harness) | **96 passed** |
| `MeshWeaver.NodeOperations.Test` (`MonolithMeshTestBase`) | **76 passed** |
| `MeshWeaver.AccessControl.Test` (`MonolithMeshTestBase`) | **29 passed** |
| `MeshWeaver.Hosting.Monolith.Test` (158 files on `MonolithMeshTestBase`) | started locally — result posted as a comment below |

Not run locally: the remaining ~30 suites, the E2E/portal projects, and anything needing external infrastructure. CI covers those.

## 🚨 Do not merge yet

`main` CD is red for an unrelated reason (a promote/image issue on `aaf95afc3`, owned by another worker). This PR is prepared and green on its own gates; the maintainer decides when it lands.

### DIFF
--- a/src/MeshWeaver.Hosting.Monolith.TestBase/MonolithMeshTestBase.cs
+++ b/src/MeshWeaver.Hosting.Monolith.TestBase/MonolithMeshTestBase.cs
@@ -1253,7 +1253,7 @@ public abstract class MonolithMeshTestBase : Fixture.TestBase
     /// suite. The waits are sequential on purpose — a client that will not finish is named
     /// individually, which is what the trace file needs to be attributable.</para>
     /// </summary>
-    private async Task DisposeTestClients(string testName, Stopwatch sw)
+    private async Task DisposeTestClientsAsync(string testName, Stopwatch sw)
     {
         lock (_requestHubLock) _requestHub = null;
         IMessageHub[] snapshot;
@@ -1399,7 +1399,7 @@ public abstract class MonolithMeshTestBase : Fixture.TestBase
             // intermediate tests were already slow from the action-block
             // congestion). See ConcurrentRequests deadlock (commit 02dd88f37)
             // and the AI/Threading suite 6-min CI timeout.
-            await DisposeTestClients(testName, sw);
+            await DisposeTestClientsAsync(testName, sw);
             TestPhaseTrace(testName, "DISPOSE_SHARED_SKIP", sw.ElapsedMilliseconds);
             try { await base.DisposeAsync(); }
             catch (Exception ex)
@@ -1416,7 +1416,7 @@ public abstract class MonolithMeshTestBase : Fixture.TestBase
         // Non-shared path also benefits — Mesh.Dispose disposes all hosted hubs
         // including clients, but doing it via the tracked list is faster and
         // more deterministic (no race against the Mesh's own dispose).
-        await DisposeTestClients(testName, sw);
+        await DisposeTestClientsAsync(testName, sw);
 
         try
         {
--- a/src/MeshWeaver.Hosting.Monolith.TestBase/MonolithMeshTestBase.cs
+++ b/src/MeshWeaver.Hosting.Monolith.TestBase/MonolithMeshTestBase.cs
@@ -1236,7 +1236,24 @@ public abstract class MonolithMeshTestBase : Fixture.TestBase
         });
     }
 
-    private void DisposeTestClients(string testName, Stopwatch sw)
+    /// <summary>
+    /// Disposes every client hub this test created AND JOINS each one's teardown before returning.
+    ///
+    /// <para>🚨 The join is not tidiness. On the <see cref="SharesMeshAcrossTests"/> path this method
+    /// is the LAST thing that touches the mesh — <c>DisposeAsync</c> returns straight after it and
+    /// the next <c>[Fact]</c> begins — so a bare <c>Dispose()</c> here starts the client's teardown
+    /// (action-block drain, sync-stream unregistration, registrant callbacks) and then hands the
+    /// mesh to the next test while all of that is still running on other threads. The observed
+    /// result is a burst of <c>[SYNC_STREAM] Not setting … — stream is disposed</c> followed by a
+    /// use-after-dispose that takes the host down with exit 139, naming no test. On the non-shared
+    /// path the very next statements stop the hosted services and dispose the Mesh, which is the
+    /// same race one level up.</para>
+    ///
+    /// <para>Bounded per client and LOUD on expiry: a wedged client must fail visibly, not hang the
+    /// suite. The waits are sequential on purpose — a client that will not finish is named
+    /// individually, which is what the trace file needs to be attributable.</para>
+    /// </summary>
+    private async Task DisposeTestClients(string testName, Stopwatch sw)
     {
         lock (_requestHubLock) _requestHub = null;
         IMessageHub[] snapshot;
@@ -1250,12 +1267,9 @@ public abstract class MonolithMeshTestBase : Fixture.TestBase
         TestPhaseTrace(testName, "DISPOSE_CLIENTS_START", sw.ElapsedMilliseconds, $"count={snapshot.Length}");
         foreach (var client in snapshot)
         {
-            try { client.Dispose(); }
-            catch (Exception ex)
-            {
-                FileOutput.WriteLine(
-                    $"[DISPOSE] {testName}: client {client.Address} dispose failed: {ex.GetType().Name}: {ex.Message}");
-            }
+            await client.DisposeAndJoinAsync(
+                message => FileOutput.WriteLine($"[DISPOSE] {testName}: {message}"),
+                DisposeTimeout);
         }
         TestPhaseTrace(testName, "DISPOSE_CLIENTS_DONE", sw.ElapsedMilliseconds);
     }
@@ -1385,7 +1399,7 @@ public abstract class MonolithMeshTestBase : Fixture.TestBase
             // intermediate tests were already slow from the action-block
             // congestion). See ConcurrentRequests deadlock (commit 02dd88f37)
             // and the AI/Threading suite 6-min CI timeout.
-            DisposeTestClients(testName, sw);
+            await DisposeTestClients(testName, sw);
             TestPhaseTrace(testName, "DISPOSE_SHARED_SKIP", sw.ElapsedMilliseconds);
             try { await base.DisposeAsync(); }
             catch (Exception ex)
@@ -1402,7 +1416,7 @@ public abstract class MonolithMeshTestBase : Fixture.TestBase
         // Non-shared path also benefits — Mesh.Dispose disposes all hosted hubs
         // including clients, but doing it via the tracked list is faster and
         // more deterministic (no race against the Mesh's own dispose).
-        DisposeTestClients(testName, sw);
+        await DisposeTestClients(testName, sw);
 
         try
         {
--- a/src/MeshWeaver.Hosting.Orleans.TestBase/OrleansTestBase.cs
+++ b/src/MeshWeaver.Hosting.Orleans.TestBase/OrleansTestBase.cs
@@ -127,11 +127,17 @@ public abstract class OrleansTestBase<TSiloConfigurator>(ITestOutputHelper outpu
             snapshot = _clientsCreated.ToArray();
             _clientsCreated.Clear();
         }
+        // 🚨 JOIN each client's teardown before the cluster goes down. A bare Dispose() only STARTS
+        // it: the action-block drain, the routing-stream unregistration and the registrant callbacks
+        // all run afterwards, on other threads — and the very next statement hands the cluster to
+        // OrleansClusterDisposal, which stops the silos those callbacks are still talking to. The
+        // outcome is a use-after-dispose (grain calls onto a stopping silo, continuations resolving
+        // from a torn-down scope) that kills the host mid-run rather than failing a test. The
+        // previous `catch { /* best-effort */ }` also discarded a disposal fault entirely; it is now
+        // written to the test output (via FileOutput, which is safe to call after the test has been
+        // reported — a raw ITestOutputHelper throws there).
         foreach (var client in snapshot)
-        {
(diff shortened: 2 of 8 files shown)


rbuergi:
## Local suite results, and one flake I could not clear — reported rather than waved through

**CI on this head is green**: 21 checks pass (all six test shards, the binary-compatibility gate, the doc gate), one still consolidating.

### Suites executed locally on this branch

| suite | result |
|---|---|
| `MeshWeaver.Documentation.Test` (holds the new guard) | 100 passed |
| `MeshWeaver.Messaging.Hub.Test` (the helper's home; 14 of these are the new `HubDisposalJoinTest`) | 329 passed |
| `MeshWeaver.Hosting.Orleans.Test` (`OrleansTestBase` + `SharedOrleansFixture`) | 177 passed |
| `MeshWeaver.Content.Test` (exercises the `ShareMeshAcrossTests` path) | 223 passed, 1 skipped |
| `MeshWeaver.PluginTester.Test` (the gate harness) | 96 passed |
| `MeshWeaver.NodeOperations.Test` / `MeshWeaver.AccessControl.Test` (`MonolithMeshTestBase`) | 76 / 29 passed |
| `MeshWeaver.Hosting.Monolith.Test` (full, ×2) | 751 passed, 3 skipped, **1 failed each time — a different test each time** |

### The flake, and what measurement says about it

Two full local runs of `MeshWeaver.Hosting.Monolith.Test` each failed one test, and not the same one:
`BuildCoordinationTest.Follower_StandsDown_…` (TimeoutException) the first time,
`WorkspaceCacheEvictionTest.NewSubscriber_AfterRecreate_GetsFreshSnapshot` the second. Both runs were on a
laptop at load average 10–20 with another session running the identical suite concurrently, and CI ran the
whole assembly green.

I ran the control anyway rather than calling it load, because "my PR makes a test flake" is not something to
assume away. Interleaved runs of `WorkspaceCacheEvictionTest` (branch and a detached worktree at the merge
base `aaf95afc3`, alternating, so load drift hits both equally):

| tree | failures |
|---|---|
| merge base | **0 / 87** |
| merge base **+ `HubDisposalJoin.cs` only** (the new file, no call site uses it) | **0 / 30** |
| this branch, with `MonolithMeshTestBase`'s client join **reverted** | 1 / 30 |
| this branch | 5 / 55 |

So the perturbation is not the new file, and it survives reverting the one changed site that can reach this
assembly — the branch flakes at ~9%, the ablation at ~3%, the merge base at 0%.

**What the failure actually is.** `No node found at 'TestData/cache-recreate' — the node was deleted, so this
address will not reactivate.` That string has exactly one home: `MessageService.DeletedAddressMessage`, on the
**teardown** path. The test deletes the node, waits for the change feed's `Created` event, then subscribes a
fresh client — and the change-feed event fires independently of the OLD per-node hub finishing its disposal.
When the new subscriber lands on that hub while its delete tombstone is still live, it is NACKed permanently
rather than reactivating.

That is the same defect class this PR is about — someone proceeding past a disposal that has not finished —
one level down, in the mesh's own per-node hub recycle, and neither the test nor the product joins that
disposal. It is a pre-existing race that timing perturbs, not something the join introduces; the single test
run in isolation passes 15/15 on this branch, and in the captured failure it was the FIRST test in its class,
so none of the changed teardown code had executed yet (`xunit.runner.json` here is
`maxParallelThreads: 1`, no parallelization).

I have deliberately **not** fixed it here: it is a product/test race in `WorkspaceCacheEviction` +
`MessageService`'s tombstone, adjacent to #2647, and folding it into this change would mix a governance
ratchet with a mesh recycle fix. Flagging it for whoever owns that cluster.

Issue #2556, opened 2026-08-27
6,902 'NodeType is not registered' failures in 90 minutes on memex-cloud — content is imported before its type, and it is the same edge as the FutuRe render flake


rbuergi:
Found by querying Loki for the class behind the CI FutuRe flake (#2533). It is not only a test problem — **production is failing content upserts in bulk.**

## Measured, memex-cloud, 90-minute window (2026-08-27 ~21:28Z)

```
sum(count_over_time({namespace="memex-cloud"} |~ "is not registered" [90m]))  →  6902
```

Two shapes, same root:

**1. Upserts REFUSED because the type is not registered yet**
```
System.InvalidOperationException: Upsert of
  'MeshWeaver/samples/Graph/Data/FutuRe/EuropeRe/TransactionMapping/EUR-COMM_FIRE-PROP'
  failed: NodeType 'FutuRe/TransactionMapping' is not registered
System.InvalidOperationException: Upsert of
  'MeshWeaver/samples/Graph/Data/Systemorph/Marketing/Posts/MeshWeaverLaunch'
  failed: NodeType 'Systemorph/Marketing/Post' is not registered
```
At least `FutuRe/TransactionMapping` and `Systemorph/Marketing/Post`; several distinct nodes each.

**2. Payloads degrading to untyped `JsonElement`**
```
Received '$type':'Organization' which is NOT registered in this (receiving) hub — it can only be
read as an untyped JsonElement (the value renders empty / reactive waits time out).
```
Also `NBodyContent`, `DataModelExplorer`.

## Why this matters beyond the noise

The warning text states the consequence itself: **"the value renders empty / reactive waits time out."** That is exactly the shape of the CI flake in Systemorph/MeshWeaver.Plugins#1141 — `FutuReAnalysisTest.EuropeRe_AnnualReport_EmbeddedCharts` waits 50 s for `KeyMetrics` at `FutuRe/EuropeRe/Analysis`, receives six emissions that are never the control, and times out. Same partition, same class, and the CI job log carries the same `$type … NOT registered` wall (`AnalysisContent`, `BusinessUnit`, `Section`, `Space`).

So Systemorph/MeshWeaver.Plugins#1141 is likely not an isolated test race but the test-visible face of this ordering defect.

## The shape of the defect

[...]

PR #2646, opened 2026-08-29
fix(#2636, #2637): an install that CLAIMS mail it cannot send is refused at BOOT, by key

rbuergi:
Fixes the outbound-email outage reported in Systemorph/MeshWeaver#2636 and Systemorph/MeshWeaver#2637 — **two views of one failure, one fix**. Both tickets are the same memex install at the same host start: `Email:Enabled=true` with `Email:TenantId` and `Email:ClientId` unset, seen once through `OutboundEmailSender` and once through `InvitationEmailSender`. Same fingerprint pod, same minute.

## The defect

Not the missing key — that is ordinary. **Nothing surfaced it.**

The portal came up perfectly healthy with mail dark: `/health` 200, every page served, and every invitation, notification and document share queued as `New` and stayed there. `EmailDeliveryGuard` did exactly its job — the watchers refused, so nothing was ever falsely stamped `Sent` (#2023) — but its entire output is **one `Error` line per host start**, and a log line nobody reads is not a signal. The install said mail was on and dropped every message, for as long as it took a human to notice mail had not arrived.

## The fix — the established shape, not a new one

`EmailConfigurationGuard` (`memex/Memex.Portal.Shared/Email/`) is a pure startup decision — no container, no credential object, no I/O — called from `MemexConfiguration.ConfigureMemexMesh` beside `ValidateContentStorageDurability` (#435) and `MicrosoftTenant.Validate` (#2634). **Two rules, and they are deliberately different** — the same distinction the OIDC tenant guard encodes:

| The `Email` section is… | Verdict | Why |
|---|---|---|
| **absent, or `Email:Enabled=false`** | **starts** — never refused, however blank or however full | Blank is what "no mail on this install" looks like: every local dev run, every test host, every deployment that never wanted mail. Aborting a portal for an unconfigured **optional** integration is Systemorph/MeshWeaver#2510 verbatim. |
| **`Email:Enabled=true` and complete** (either flow) | **starts** | Nothing to refuse. Managed identity needs no credential keys at all. |
| **`Email:Enabled=true` and INCOMPLETE** | **refuses at startup**, naming every missing key | Someone *meant* to enable mail. Such an install sends nothing today, so refusing cannot take working mail down — it converts a silent drop into a named configuration error an operator can act on. |

The refusal names each key in **both** forms — `Email:TenantId` (binder, what the docs and the code speak) **and** `Email__TenantId` (what an operator actually types into a Helm value, a configMap or `kubectl set env`) — plus every way out: complete the flow, `Email:UseManagedIdentity=true`, or `Email:Enabled=false`.

### 🚨 This does NOT re-open Systemorph/MeshWeaver#2510

Systemorph/MeshWeaver#2510 was not "email configuration stopped the host" in the abstract. It was `EmailDeliveryGuard` **activating** the Graph sender from `IHostedService.StartAsync`, whose `ClientSecretCredential` constructor threw `ArgumentException: Invalid tenant id provided` → `Hosting failed to start` — an unactionable message, from a container resolution, naming Azure's validator instead of the key to set. Two properties keep that fixed and are asserted here:

- the verdict is reached from **inert configuration data only**, so it cannot throw for any reason other than the one it reports; and
- the **unconfigured** install (blank/disabled) is never refused.

What changes is only the *severity* of the enabled-but-incomplete case: from a log line nobody reads to a named startup refusal. **`EmailDeliveryGuard` is untouched** — it still refuses the watchers on hosts that never call this guard (LocalMesh, test hosts, anything not going through `ConfigureMemexMesh`), and it is still what guarantees queued mail is never stamped `Sent`. All 8 of its Systemorph/MeshWeaver#2510/#2023 pins stay green.

### One definition, not a second copy

The credential half delegates to `EmailOptions.MissingCredentialKeys()` — the same answer `EmailDeliveryGuard` and `NoOpEmailSender` already report — rather than re-deriving it, so "which keys does the selected flow need" keeps exactly one definition and the two can never tell the same operator different stories. `Email:MailboxAddress` is added on top because it is required by **both** flows and by no credential: the system send path is `/users/{MailboxAddress}/sendMail`, so a blank value composes a Graph request for no mailbox. Scope is **outbound**; the inbound keys are deliberately not guarded (same shape, no incident, and widening a boot-time refusal past the failure it fixes is how a guard acquires a blast radius nobody signed off on).

## Which repo, and why

**Core (`Systemorph/MeshWeaver`)**, not `MeshWeaver.Plugins`. The portal *hosts* did move to the plugins repo, but everything this touches stayed here: `ConfigureMemexMesh` (`memex/Memex.Portal.Shared/MemexConfiguration.cs`) is the single boot path **both** hosts call — `Memex.Portal.Monolith/Program.cs:70` and `Memex.Portal.Distributed/Program.cs:372`, both via `$(MeshWeaverRoot)/memex/Memex.Portal.Shared` — and `EmailOptions`, `EmailDeliveryGuard`, `NoOpEmailSender`, both watchers and their tests all live here too. Landing it here gives one guard for both hosts, next to the code it is about and next to the two precedents it copies. Nothing in the plugins repo changes; there is no signature change to `ConfigureMemexMesh`.

## Tests — `test/Memex.Portal.Shared.Test/EmailStartupConfigurationGuardTest.cs`

**29 cases. Against the unfixed code (guard neutered to pre-fix behaviour — nothing refuses at startup): `Failed: 13, Passed: 16`.** With the fix: **29/29 green**. The 16 that already passed are the "must not refuse" direction — the Systemorph/MeshWeaver#2510 protection — which is exactly what stops the fix from being satisfied by a guard that refuses everything.

| Direction | Cases |
|---|---|
| **Never configured must NOT refuse** (#2510) | absent section (`null`); `Enabled=false` over three real shapes (all blank / half-filled / fully filled but switched off); bound configuration with no `Email__*` at all and with `Email__Enabled` `false`/`False` |
| **Enabled + incomplete REFUSES, by name** | the incident shape (tenant + client id missing, secret set) — names both in both forms and **not** the key that is set; each of the 4 required keys blanked on its own; whitespace-is-unset ×3; enabled-with-nothing-else names all four at once; managed identity refused for the mailbox **alone**, never for credential keys its flow does not use |
| **Enabled + complete starts** | both credential flows; a complete section written in `Email__*` environment form |
| **Message quality** | every way out named (`UseManagedIdentity`, `Enabled=false`, both forms); binder→environment mapping |
| **One definition** | the credential half agrees with `EmailDeliveryGuard.MissingConfiguration` across 4 flow shapes |
| **The guard is WIRED, not merely present** | the real `ConfigureMemexMesh` boot path refuses a half-configured section, and is unaffected by an install that never configured mail — a pure function nobody calls refuses nothing, and its unit tests stay green forever |

Full suites: `Memex.Portal.Shared.Test` **454/454**, `MeshWeaver.Documentation.Test` **97/97**. Build clean, 0 warnings (warnings-as-errors).

## ⚠️ What this does NOT do — the maintainer's half

**This does not restore mail on memex.** It converts a silent drop into a named startup error; the `Email__*` values are deployment configuration and are deliberately untouched here — exactly as Systemorph/MeshWeaver#2634 left the OIDC tenant to the deployment. No Helm values, no ConfigMap, no cluster change is in this diff. The default chart already ships `Email__Enabled: "false"`, so only the memex overlay is affected.

🚨 **Ordering matters, and it is a live-install decision, which is why this is not self-merged.** An install that is enabled-but-incomplete **today** will refuse to start once it rolls onto a build carrying this guard. On memex, set `Email__TenantId` + `Email__ClientId` (the secret is already set — the refusal named only the other two), **or** `Email__UseManagedIdentity=true` with the `Mail.Send` app role, **or** `Email__Enabled=false` — **before** the image carrying this merges and rolls. Once the keys land and the portal restarts, the queued `New` mail drains by itself, with no data repair and nothing to re-queue.

That trade is the intent, not a side effect: a portal that says mail is enabled and drops every message is the defect this replaces.

## Docs

`Doc/Architecture/SendingEmail` gains a *"An INCOMPLETE section refuses at STARTUP, by key"* section (the verdict table, the required keys, why Systemorph/MeshWeaver#2510 stays fixed, and the roll-ordering warning). A What's New entry ships for 2026-08-29, and the 2026-08-27 Systemorph/MeshWeaver#2510 entry is qualified so the feed does not carry two adjacent notes saying opposite things.

### DIFF
--- a/memex/Memex.Portal.Shared/Email/EmailConfigurationGuard.cs
+++ b/memex/Memex.Portal.Shared/Email/EmailConfigurationGuard.cs
@@ -1,0 +1,172 @@
+using System.Collections.Immutable;
+using MeshWeaver.Mesh;
+using Microsoft.Extensions.Configuration;
+
+namespace Memex.Portal.Shared.Email;
+
+/// <summary>
+/// The BOOT-TIME guard for the <c>Email</c> section: an install that CLAIMS mail
+/// (<c>Email:Enabled=true</c>) and cannot possibly deliver it is refused at startup, by key
+/// (MeshWeaver#2636, #2637).
+///
+/// <para>🚨 <b>The defect this closes.</b> On memex the section was half-set — enabled, with
+/// <c>Email:TenantId</c> and <c>Email:ClientId</c> unset — and the portal came up perfectly
+/// healthy with mail DARK. <see cref="EmailDeliveryGuard"/> did its job (the watchers refused, so
+/// nothing was ever falsely stamped <c>Sent</c>), but a single <c>Error</c> line in a pod log is
+/// not a signal anyone reads: <c>/health</c> was 200, the site served, and every invitation,
+/// notification and document share queued as <c>New</c> and stayed there. The install said mail
+/// was on and dropped every message. That is the failure — not the missing key, which is
+/// ordinary, but the fact that nothing surfaced it.</para>
+///
+/// <para><b>Two rules, and they are deliberately different</b> — the same distinction
+/// <see cref="Authentication.MicrosoftTenant"/> encodes for the OIDC tenant (#2621), and it is the
+/// whole of this guard:</para>
+/// <list type="number">
+/// <item><description><b>Absent or <c>Email:Enabled=false</c> = the integration was NEVER
+/// configured, and it must NOT refuse.</b> A blank section is exactly what "no mail on this
+/// install" looks like — it is every local dev run, every test host, every deployment that never
+/// wanted mail — and aborting a portal because an OPTIONAL integration is unconfigured is the
+/// #2510 failure verbatim (a guard that fails worse than the thing it guards). Nothing here is
+/// even looked at in that case.</description></item>
+/// <item><description><b><c>Email:Enabled=true</c> with the section INCOMPLETE = a real
+/// misconfiguration, refused at startup, naming the exact keys.</b> Someone MEANT to enable mail.
+/// An install in this state is not sending anything today, so refusing it cannot take a working
+/// portal's mail down — it converts a silent drop into a named configuration error at boot, which
+/// is the one thing an operator can act on.</description></item>
+/// </list>
+///
+/// <para>🚨 <b>Why this does NOT re-open #2510.</b> #2510 was not "email configuration stopped the
+/// host" in the abstract; it was <c>EmailDeliveryGuard</c> ACTIVATING the Graph sender from
+/// <c>IHostedService.StartAsync</c>, whose <c>ClientSecretCredential</c> constructor threw
+/// <c>ArgumentException: Invalid tenant id provided</c> → <c>Hosting failed to start</c> — an
+/// unactionable message, from a container resolution, that named Azure's validator instead of the
+/// key to set. Two properties keep that fixed and are not negotiable here:
+/// <list type="bullet">
+/// <item><description>this verdict is reached from <b>inert configuration data only</b> — no
+/// container, no credential object, no I/O — so it cannot throw for any reason other than the one
+/// it is reporting; and</description></item>
+/// <item><description>the <b>unconfigured</b> install (blank/disabled) is never refused.</description></item>
+/// </list>
+/// What changes is only the SEVERITY of the enabled-but-incomplete case: from a log line nobody
+/// reads to a named startup refusal. <see cref="EmailDeliveryGuard"/> stays exactly as it is — it
+/// still refuses the watchers on hosts that never call this guard (LocalMesh, test hosts, any host
+/// not going through <c>ConfigureMemexMesh</c>), and it is still what guarantees queued mail is
+/// never stamped <c>Sent</c>.</para>
+///
+/// <para>Pure decision, no I/O and no container, so it is unit-testable without spinning a host —
+/// the same shape as <c>MemexConfiguration.ValidateContentStorageDurability</c> and
+/// <see cref="Authentication.MicrosoftTenant"/>.</para>
+///
+/// <para>Scope is OUTBOUND, deliberately. The inbound channel (<c>Email:InboundEnabled</c>,
+/// <c>Email:WebhookBaseUrl</c>, <c>Email:SubscriptionClientState</c>) has the same shape but no
+/// incident behind it, and widening a boot-time refusal past the failure it is fixing is how a
+/// guard acquires a blast radius nobody signed off on.</para>
+/// </summary>
+public static class EmailConfigurationGuard
+{
+    /// <summary>The configuration section this guard reads, in binder form.</summary>
+    public const string SectionName = EmailOptions.SectionName;
+
+    /// <summary>The switch that turns a blank section into a CLAIM, in binder form.</summary>
+    public const string EnabledKey = $"{SectionName}:{nameof(EmailOptions.Enabled)}";
+
+    /// <summary>The mailbox the system path sends AS — required by both credential flows.</summary>
+    public const string MailboxAddressKey = $"{SectionName}:{nameof(EmailOptions.MailboxAddress)}";
+
+    /// <summary>The alternative to the client-secret flow, named in every refusal as a way out.</summary>
+    public const string UseManagedIdentityKey = $"{SectionName}:{nameof(EmailOptions.UseManagedIdentity)}";
+
+    /// <summary>
+    /// A binder key (<c>Email:TenantId</c>) as the environment-variable / configMap entry an
+    /// operator actually sets (<c>Email__TenantId</c>).
+    ///
+    /// <para>Both forms appear in every refusal on purpose. The binder form is what the code, the
+    /// docs and <see cref="EmailOptions.MissingCredentialKeys"/> speak; the environment form is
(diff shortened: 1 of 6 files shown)


rbuergi:
## Sequencing decided — land Systemorph/Memex#144 FIRST

The maintainer's call: **email stays ON; the credentials get fixed.** So the ordering risk this PR carries is removed by completing the config rather than by disabling mail.

**[Systemorph/Memex#144](https://github.com/Systemorph/Memex/pull/144)** does that. What I measured while writing it:

| | |
|---|---|
| `memexcloud-Email-ClientSecret` in the `Systemorph` vault | **already present** — the secret half was provisioned |
| `Email__*` in any memex-cloud values file (public / ha / replica / gate) | **absent, all four** |
| chart default | `Email__Enabled: "false"` |

So the section was never wired at all — someone provisioned the secret and stopped. That also refines this issue's own premise: it is not that two keys were lost in a rollout, it is that the overlay never carried any of them.

**The trap worth recording here**, because it would have produced a config that looks complete and still delivers nothing:

| app | granted roles | |
|---|---|---|
| `Memex Mail (Graph sender)` `e18a4307-…` | Graph **`Mail.Send`** (`b633e1c5-…`) + `Mail.ReadWrite`, admin-consented **application** roles | ✅ the sender; atioz already uses it |
| `Memex Portal (memex.systemorph.com)` `eed68df1-…` | **delegated scopes only, ZERO app-role assignments** | ❌ cannot send in the client-credentials flow |

The second is the one that matches the install by name. `EmailOptions.MissingCredentialKeys()` would have been perfectly satisfied by it.

**Order:** Memex#144 → then this. With the section complete, this guard has nothing to refuse on that install, and it starts protecting against the next half-configuration instead of firing on the current one.

rbuergi:
The shard-3 red is **not this diff** — it is the documented FutuRe flake.

```
MeshWeaver.FutuRe.Test.FutuReAnalysisTest.EuropeRe_AnnualReport_EmbeddedCharts_ShouldRenderViaPathResolution [FAIL]
ObservableAssertionException : Expected the observable to emit a value matching the predicate
within 50s because KeyMetrics should render …
```

That exact test is Systemorph/MeshWeaver.Plugins#1141 (closed), and Systemorph/MeshWeaver#2556 names it again verbatim: *"exactly the shape of the CI flake in Systemorph/MeshWeaver.Plugins#1141 — `FutuReAnalysisTest.EuropeRe_AnnualReport_EmbeddedCharts` waits 50 s for `KeyMetrics`"*, attributing it to `NodeType is not registered` upserts leaving a value that renders empty while reactive waits time out.

**This PR cannot reach it.** The whole diff is:

```
memex/Memex.Portal.Shared/Email/EmailConfigurationGuard.cs
memex/Memex.Portal.Shared/MemexConfiguration.cs
src/MeshWeaver.Documentation/Data/Architecture/SendingEmail.md
src/MeshWeaver.Documentation/Data/WhatsNew/…  (×2)
test/Memex.Portal.Shared.Test/EmailStartupConfigurationGuardTest.cs
```

A startup configuration guard and its tests. There is no path from that to chart rendering via path resolution.

Re-running the failed shard. Per this repo's own rule I am saying *why* it is unrelated rather than just asserting it — a red waved through on a feeling is the one thing the policy does not license, and the evidence here is the file list plus a named prior issue for the same test.

⚠️ Reminder for whoever lands this: **Systemorph/Memex#144 must merge FIRST.** memex-cloud is `Email:Enabled=true` with the credentials incomplete today, so this guard would refuse startup on its next roll. Systemorph/MeshWeaver#144 completes the config; after it, this guard has nothing to refuse there.

rbuergi:
The red is **not reachable from this diff** — closure analysis plus a local run, rather than a
re-run.

**The failing test** (1 fail / 8,564 pass; extracted from the `testResults-shard3` trx, since the
check summary carries only counts):

```
MeshWeaver.FutuRe.Test.FutuReAnalysisTest.EuropeRe_AnnualReport_EmbeddedCharts_ShouldRenderViaPathResolution
```

Chart rendering via path resolution. This PR is an email-configuration guard.

**Why it cannot be this diff.** The two production files changed both live in
`memex/Memex.Portal.Shared/`:

- `Email/EmailConfigurationGuard.cs` (new)
- `MemexConfiguration.cs` (+9)

`MeshWeaver.FutuRe.Test`'s transitive project closure is **25 projects, and `Memex.Portal.Shared`
is not one of them.** The changed assembly cannot load in the process under test, so no behaviour
of it is observable there. (It does reach `MeshWeaver.Hosting.Monolith`, so this is not a
"nothing in common" claim — the specific changed assembly is absent.)

I checked this rather than assuming, because unlike a pure-Orleans change this diff touches
`MemexConfiguration`, which portal-booting tests legitimately use. It just is not on that test's
path.

**And the PR's own tests pass.** Built and ran the new suite locally on this branch, Release:

```
Passed!  -  Failed: 0, Passed: 454, Skipped: 0, Total: 454  — Memex.Portal.Shared.Test
```

So the 396 lines of new test added here are green; the one shard-3 failure is a different project
entirely.

Re-running to watch it go green would be a re-roll rather than evidence, so I am recording the
closure instead. Same treatment as Systemorph/MeshWeaver#2645, whose red was also unreachable from its diff.

Issue #2640, opened 2026-08-29
Every page render fires unanchored cross-schema queries (access, notifications, threads, mesh_nodes) — ~2 s each over 186 partition schemas


rbuergi:
Measured from Loki on memex-cloud (namespace `memex-cloud`, 05:32–06:02Z, 2026-08-29, probe script from the mesh):

```
716 × [CrossSchema] SLOW … across 186 schema(s)
  mesh_nodes     454 ×  avg 1710 ms  max 7827 ms
  notifications  117 ×  avg 1997 ms  max 4533 ms
  access         113 ×  avg 2055 ms  max 5643 ms
  threads         32 ×  avg 1929 ms  max 3825 ms
```

`PostgreSqlCrossSchemaQueryProvider` warns each time: "An unanchored query UNIONs every partition; add a path:/namespace: first segment to pin it to one schema." These fire on ordinary page loads (browser timings: server 40–250 ms, DOM < 1 s, first painted layout area 1.5–2.2 s) — the `access` and `notifications` ones look like the permission evaluation and the notification bell, `threads` the chat thread list, and a large share of `mesh_nodes` are 0/0-row lookups (misses).

The per-page floor is therefore ~2 s and grows with the partition count. Candidates: anchor the permission/notification/thread queries to the viewer's partition (`{user}/…`) + the group-grant partitions, cache the access fold per circuit, and stop the `Admin/Menu/{X}` route misses (`[ROUTE] NotFound: No node found at 'Admin/Menu/AI'` ×7/30 min per menu key) which each pay a routed lookup.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01ETgRRMRVzYMg1WU4pMywed

rbuergi:
Design direction (Roland, 2026-08-29: "can we issue events so we don't have to poll?"): these three reads should not be per-render queries at all. The mesh is reactive — establish ONE viewer-anchored subscription per circuit at sign-in (notifications, open threads, effective permissions, anchored to `{user}/…` plus the partitions the viewer's group grants name) and let the change feed update it; a page render then reads a warm stream instead of UNIONing 186 schemas. The `Admin/Menu/{X}` route misses belong to the same change: resolve the menu keys once per circuit, not per render.

rbuergi:
## 🚨 First: the log line this issue is built on cannot tell fan-out from slowness

Before the enumeration, a correction to the evidence — because it changes what the 716 lines mean.

[...]

PR #2651, opened 2026-08-29
perf(menu): read the menu catalog with the storm-safe existence query, not a point probe

rbuergi:
Addresses the `Admin/Menu/{X}` half of #2640 — deliberately NOT a closing reference; the rest of that issue's enumeration is unaffected by this change.

## The defect

`MenuPresentationOverlay.CatalogStream` point-read `Admin/Menu/{context}` with `GetMeshNodeStream`. That node is absent **by design**, as the same file says four lines above the call:

> There is deliberately NO seeded catalog. "No entry for this area" is the normal resting state, not drift to be repaired.

A point read of a node that does not exist is neither free nor quiet. It resolves through `PathResolutionService` → `RoutingGrain.BuildGrainRoute`, finds nothing, and answers NotFound — one `[ROUTE] NotFound: No node found at 'Admin/Menu/{context}'` warning plus a DeliveryFailure NACK per probe. `PathResolutionService` deliberately caches **no** negative resolution (`if (resolution is null) return;`), so the only damping is `MeshNodeStreamCache`'s storm breaker, whose window expires (2s → 4s → … → 5min) and re-probes for the life of the process.

**And it is not once per circuit.** `RenderMenus` is a global predicate renderer:

```csharp
.AddLayout(layout => layout.WithRenderer(_ => true, RenderMenus));
```

Its own comment says so — *"RenderMenus is a GLOBAL predicate renderer (`WithRenderer(_ => true, …)`), so it runs on EVERY area render"* — and it installs this stream with `ReplaceDisposable`, i.e. dispose the old subscription and open a new one, **once per menu CONTEXT per area render**. Contexts are data-driven (`UiContributionCatalog.Current`, read at render time), so installing a plugin that declares a TopBar menu adds one more permanently-missing routed lookup to every area render on the mesh. #2640 counts these misses directly.

## The rule this broke is already written down

`PlatformUpdateStatus.Observe`, in this repo:

> Absence is detected with the storm-safe existence `GetQuery` (empty-on-absent) — **NEVER a point `GetMeshNodeStream` probe of a maybe-absent path**, which NotFound-resubscribe-storms on an install that never wired self-update.

`NotificationService.ReadSettings` states the same thing for the same shape, and names the same precedents (`NotificationSettingsNodeType.EnsureExists`, the AiSettings/UpdatePolicy nodes). So this is an **existing rule applied where it was missed**, not a new one — and it was missed on a far hotter path than the one the rule was written for: About-page-a-few-times-a-day versus once-per-context-per-area-render.

## The change

`CatalogStream` reads through the synced query surface instead:

```csharp
workspace.GetQuery(CatalogQueryId(context), CatalogQuery(context))
```

with `CatalogQuery(context)` = `path:Admin/Menu/{context} select:path,id,namespace,name,nodeType,content` and `CatalogQueryId(context)` = `MenuPresentation|Admin/Menu/{context}` — the `{type}|{path}` convention `UpdatePolicyNodeType` and `NotificationSettingsNodeType` already register under, so a future seeding path shares this registration rather than opening a second synced query for the same node.

Four properties, each load-bearing:

- **Empty-on-absent.** A query answers "no rows"; no NotFound, no NACK, no negative-cache window to expire.
- **Still anchored, and anchored the only way that works.** `path:Admin/…` routes by first segment to the `admin` schema. A `namespace:`-shaped read could not substitute: `admin` is deliberately excluded from cross-schema search, so that form would fan out over every partition **and still miss the catalog**.
- **One upstream per context, for the process.** `GetQuery` caches under the stable id with `Replay(1).AutoConnect(1)`, so the per-render re-subscribe lands on a warm replay. This is the "resolve the menu keys once, not per render" in @rbuergi's design comment on #2640, at silo scope rather than circuit scope.
- **Liveness unchanged**, which is the point. It is still a synced query fed by the change feed, so an admin creating `Admin/Menu/Node` updates every open menu with no recycle — the "Seconds, no build, no image, no rollout" contract in `Doc/Architecture/MenuAsData`.

Deliberately **not** filtered by `nodeType:` — `MenuPresentation` is a registered CONTENT type (`.WithTypes(…)`), not a NodeType, so the node's own `nodeType` is whatever its author chose. The path is already unique and `ContentAs<MenuPresentation>` is what decides whether it is a catalog. Adding a `nodeType:` filter would silently drop a legitimately-authored catalog.

Access is unchanged in both directions: the point read was gated by `MeshNodeStreamCache.ProbeEffectivePermissions`, and `GetQuery` reads upstream as System with `WrapWithPerUserRls` re-applying `CheckPermission(path, Read)` per subscriber. A viewer without Read on `Admin/Menu/{X}` sees no overlay either way.

## Tests

Five new cases in `MenuPresentationOverlayTest` (19 in the class, all green):

- `CatalogQuery_IsAnExactPathRead_SoAnAbsentCatalogIsNoRowsRatherThanNotFound`
- `CatalogQuery_IsAnchoredToTheAdminPartition_ByPathNotNamespace`
- `CatalogQuery_ProjectsContent_OrAnExistingCatalogReadsAsAbsent`
- `CatalogQueryId_IsStablePerContext_AndDistinctAcrossContexts`
- `CatalogStream_UsesTheStormSafeExistenceQuery_NeverAPointReadOfAMaybeAbsentPath`

The last one is a **source guard**, and the choice is deliberate: the string tests cannot see the thing that actually changed, because a point read and a query produce the same catalog when the node exists and differ only when it does **not** — which here is the normal state. Reproducing that behaviourally needs a live hub, a routing grain and an absent node, and a test that heavy is the first one silenced. The guard fails the moment the primitive is swapped back, which is the only regression that matters; it ignores comment lines, so the rule can keep explaining itself in prose.

**Mutation-verified**: restoring `workspace.GetMeshNodeStream(path)` in the body fails exactly that guard (1 of 19) and nothing else — which is also the honest reading of the other four, since they pin the query shape rather than the primitive.

Full `MeshWeaver.Graph.Test` green: **1717 passed, 0 failed**.

## What this does not claim

I cannot measure the saving. What is measured is the mechanism — a global predicate renderer, `ReplaceDisposable` per context per area render, a path that resolves to nothing, and a resolver that caches no negative — and that the repo already forbids this exact combination by name. The DB cost of one probe was modest (a pinned single-schema read against `admin`); the removed cost is mostly the routed round trip, the warning volume and the NACKs, so **do not expect this alone to move Postgres CPU**.

### DIFF
--- a/src/MeshWeaver.Graph/Configuration/MenuPresentationOverlay.cs
+++ b/src/MeshWeaver.Graph/Configuration/MenuPresentationOverlay.cs
@@ -22,6 +22,11 @@ namespace MeshWeaver.Graph.Configuration;
 /// </summary>
 internal static class MenuPresentationOverlay
 {
+    /// <summary>
+    /// The projection a catalog read needs. <c>content</c> is named deliberately: a <c>select:</c>
+    /// that omits it yields a node whose <c>Content</c> is silently null, which would read as
+    /// "no catalog" for a catalog that exists.
+    /// </summary>
     internal const string CatalogProjection = "select:path,id,namespace,name,nodeType,content";
 
     /// <summary>
@@ -63,11 +68,6 @@ internal static class MenuPresentationOverlay
     /// renders its compiled presentation until someone chooses to override it. Seeding a copy of the
     /// compiled presentation would instead freeze it, and silently outrank every later change to it.
     /// </para>
-    /// </summary>
-    /// <summary>
-    /// The projection a catalog read needs. <c>content</c> is named deliberately: a <c>select:</c>
-    /// that omits it yields a node whose <c>Content</c> is silently null, which would read as
-    /// "no catalog" for a catalog that exists.
     /// </summary>
     public static IObservable<MenuPresentation?> CatalogStream(
         IWorkspace workspace, JsonSerializerOptions options, string context, ILogger? logger)
--- a/src/MeshWeaver.Graph/Configuration/MenuPresentationOverlay.cs
+++ b/src/MeshWeaver.Graph/Configuration/MenuPresentationOverlay.cs
@@ -22,6 +22,34 @@ namespace MeshWeaver.Graph.Configuration;
 /// </summary>
 internal static class MenuPresentationOverlay
 {
+    internal const string CatalogProjection = "select:path,id,namespace,name,nodeType,content";
+
+    /// <summary>
+    /// The synced-query id for <paramref name="context"/>'s catalog — <c>{type}|{path}</c>, the same
+    /// convention <c>UpdatePolicyNodeType</c> and <c>NotificationSettingsNodeType</c> register under,
+    /// so a future seeding path can share this registration instead of opening a second synced query
+    /// for the same node. Stable per context and never composed per call, so the process-wide query
+    /// cache holds ONE upstream per menu context for the life of the process — see
+    /// <see cref="CatalogQuery"/> for why that matters here.
+    /// </summary>
+    internal static string CatalogQueryId(string context)
+        => $"{nameof(MenuPresentation)}|{MenuPresentation.PathFor(context)}";
+
+    /// <summary>
+    /// The catalog read for <paramref name="context"/>: an EXACT, partition-anchored read of the one
+    /// node at <see cref="MenuPresentation.PathFor"/>.
+    ///
+    /// <para>Anchored by construction — <c>path:Admin/Menu/{context}</c> carries a concrete first
+    /// segment, so every backend routes it to the <c>admin</c> partition alone rather than fanning
+    /// out. (A <c>namespace:Admin</c>-shaped read could not: <c>admin</c> is deliberately excluded
+    /// from cross-schema search, so a path is the only way in.) It is deliberately NOT filtered by
+    /// <c>nodeType:</c> — <c>MenuPresentation</c> is a registered CONTENT type, not a NodeType, so
+    /// the node's own <c>nodeType</c> is whatever its author chose; the path is already unique and
+    /// <c>ContentAs&lt;MenuPresentation&gt;</c> is what decides whether it is a catalog.</para>
+    /// </summary>
+    internal static string CatalogQuery(string context)
+        => $"path:{MenuPresentation.PathFor(context)} {CatalogProjection}";
+
     /// <summary>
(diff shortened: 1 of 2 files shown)


rbuergi:
Independent corroboration turned up by accident, from a **test harness** rather than production logs.

A `MeshWeaver.Markdown.Export.Test` job on an unrelated PR (Plugins run `33239564587`, job `99067366801`) captured this while `PixelFidelityOfferedOnlyWhenAvailableTest` sat waiting:

```
[SYNC_STREAM] OnError for PUm0ZN1HP0qSA5ilh-C4Kw (Reference=MeshNodeReference { Path =  }, Owner=Admin/Menu/Default)
 Exception: MeshWeaver.Messaging.DeliveryFailureException: No node found at 'Admin/Menu/Default'.
[SYNC_STREAM] OnError for iMMJvuHg5kiOBgVIQpTklA (Reference=MeshNodeReference { Path =  }, Owner=Admin/Menu/Node)
 Exception: MeshWeaver.Messaging.DeliveryFailureException: No node found at 'Admin/Menu/Node'.
[SYNC_STREAM] OnError for SqnJ0phrqUWlCumt6NVa9g (Reference=MeshNodeReference { Path =  }, Owner=Admin/Menu/Mesh)
 Exception: MeshWeaver.Messaging.DeliveryFailureException: No node found at 'Admin/Menu/Mesh'.
```

Three contexts, three faulted sync streams, in a single monolith test — on core `main`, i.e. **without** this PR. So the point-probe of the deliberately-absent catalog is not only a production-log inference: it reproduces in the standard harness, it faults a `SynchronizationStream` per context rather than answering empty, and the cost scales with the number of registered menu contexts exactly as the PR describes.

Worth being precise about what this does **not** show: I am **not** claiming the storm caused that test's timeout. The same window also carries `No node found at 'Roland'` (a user-root point read), so there is at least one other absent-path probe in flight, and the test passes locally in 864 ms — it looks like a load flake on a contended runner. The claim is only that the mechanism is real and observable here, which is stronger evidence than the Loki counts alone.

It also suggests a cheap regression test for later, if anyone wants one beyond this PR's source guard: assert that a monolith harness render produces **no** `DeliveryFailureException` naming `Admin/Menu/*`. That would fail on `main` today and pass with this change.

PR #92, opened 2026-04-19
refactor: reactive persistence — IMeshStorage writes return IObservable

rbuergi:
## Summary
- Flip `IMeshStorage` write ops (`SaveNode`, `DeleteNode`, `MoveNode`, `AddComment`, `DeleteComment`, `SavePartitionObjects`, `DeletePartitionObjects`) from `Task`-returning to `IObservable`-returning. Subscribe to drive — no more `await` in hub handlers.
- `HandleCreateNodeRequest`, `HandleMoveNodeRequest`, `DeleteSelfFromStorage`, post-creation save-extras, `ActivityLogBundler`, `MeshDataSourceLayoutAreas`, and `MeshService.CreateTransient` consume the observable directly.
- `MeshCatalog.CreateTransientNode` returns `IObservable<MeshNode>`; dead `UpdateAsync` / `ConfirmNodeAsync` / `IMeshCatalog.DeleteNodeAsync` removed.
- `DeleteNode` returns `IObservable<string>` (path) — no read-then-delete TOCTOU under concurrent writers.
- Companion fix (prior commit on this branch): `AccessControlLayoutArea` now builds from `host.Workspace.GetStream(new MeshNodeReference())` instead of `meshQuery.QueryAsync`, clearing the prod eternal-spinner on `Settings/AccessControl`.

Why: follows the `CLAUDE.md` "no await in hub flows / UI flows" rule. Every write to the mesh must compose as an `IObservable<T>` chain so the hub dispatcher never blocks and the click/handler path never deadlocks under load.

## Test plan
- [x] `dotnet test test/MeshWeaver.Graph.Test` — 236/236 pass
- [x] `dotnet test test/MeshWeaver.NodeOperations.Test` — 73/73 pass (2 skipped by design)
- [ ] Manual: open `/Settings/AccessControl` in prod after deploy — spinner should clear
- [ ] Manual: create / move / delete nodes through Memex portal

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #2658, opened 2026-08-29
Key the permission-probe cache by the CONTEXT it evaluated, not just (path, user)

rbuergi:
Found while investigating [MeshWeaver.SocialMedia#92](https://github.com/Systemorph/MeshWeaver.SocialMedia/issues/92) — *"MCP token cannot read Profiles nodes although the portal renders them for the same user"*.

## The defect

`PermissionEvaluator.GetEffectivePermissions` is **not** a function of `(path, userId)`. It also reads the caller's `AccessContext`:

- **`IsApiToken`** + the token's **claim roles** drive the API-token capability clamp (`PermissionEvaluator.cs`, the `IsApiToken` branch), which sets `p = Permission.None` — zeroing the **whole** permission set, not merely the `Api` bit — when a bearer context lacks `Api`;
- **`IsHub`** drives the hub-credential early return.

`MeshNodeStreamCache.ProbeEffectivePermissions` cached that result under `(path, userId)` **alone**. And `IMeshNodeStreamCache` is a process-wide **singleton** (`PersistenceExtensions.cs`: `TryAddSingleton`), so a person's browser session and their MCP token genuinely meet in the same entry for the `AccessTtl` (30 s) window.

The comment directly above the cache already notes that the evaluation depends on the captured context — *"so claim-based (Bearer-token) roles on `AccessContext.Roles` resolve"* — while the key ignores it.

## Both directions are wrong ANSWERS, not errors

That is why nothing logged, and why neither surface could reproduce the other's result:

| Direction | What happens |
|---|---|
| **Too permissive — a capability bypass** | A portal read caches the *unclamped* permissions. A bearer read of the same node inside the TTL takes that entry and **never runs the clamp at all**, admitting a token to the API surface the clamp exists to keep it off. |
| **Too restrictive — the reported symptom** | A bearer read the clamp zeroed caches `Permission.None`. The same user's **portal** read inside the TTL is then refused with `lacks Read permission` on a node they are plainly granted. |

## The fix

The key now carries every context dimension the evaluation reads — **and nothing else**, so the cache stays useful rather than being quietly disabled:

- claim roles are folded **order-independently** (they are a set), so two equivalent tokens still share an entry;
- fields the evaluator never reads (name, email) do not split entries.

## Tests

7 new cases in `PermissionProbeCacheKeyTests`, and they are **proven non-vacuous**: restoring the pre-fix `(path, userId)` key fails **exactly the 3** that pin the context dimensions, while the **4** that pin *"the cache must still work / must not over-split"* pass in both states — which is the shape you want, since a fix that merely disabled the cache would also make the first three pass.

| Suite | Result |
|---|---|
| `MeshWeaver.Hosting.Test` | 259/259 ✓ |
| `MeshWeaver.AccessControl.Test` | 29/29 ✓ |
| `MeshWeaver.Hosting.Monolith.Test` (access + permission filter) | 29/29 ✓ |

## ⚠️ Scope — please read before closing anything

This is a real defect on the reported code path, but it is **probably not the whole of SocialMedia#92**: a 30-second TTL cannot explain a denial that persisted for a week. That issue stays open. Two live reads are still needed to attribute the sustained denial, and I could not make them (the `memex`/`systemorph` MCP servers have failed DNS all session):

1. **`get @Profiles/_Policy` and `@Profiles/RobertLinkedIn/_Policy`** — does either set `api: false`? A policy that caps `Api` out makes `p` lack the `Api` bit, and the clamp then zeroes the entire set for *every* MCP token on that subtree, permanently. That would be a sustained, non-cache explanation with exactly this symptom.
2. **carson's API-token `Roles` claim snapshot** — `ClaimsCarryApi` is the clamp's only escape hatch, and the claim is snapshotted at token-creation time, so a token minted before a role change can be permanently short.

What the investigation *did* settle, against the issue's own stated theory: **the MCP path does not "evaluate only the node's own scope"**. Both surfaces provably go through the same evaluator, and it folds the full ancestor chain (`GetScopeHierarchy` → `ComputeRoleState`, `Inherited ∪ Local, deny overrides`). So @rbuergi's reading on the issue is right — the portal is correct and the denial is the bug — and the remaining suspects are the two above plus this cache.

### DIFF
--- a/src/MeshWeaver.Hosting/MeshNodeStreamCache.cs
+++ b/src/MeshWeaver.Hosting/MeshNodeStreamCache.cs
@@ -227,7 +227,49 @@ internal sealed class MeshNodeStreamCache : IMeshNodeStreamCache, IDisposable
     // Lazy<T>(ThreadSafety.ExecutionAndPublication) guarantees the factory
     // runs at most once per key, even when multiple GetOrAdd calls race.
     private readonly ConcurrentDictionary<string, Lazy<Entry>> _streams = new();
-    private readonly ConcurrentDictionary<(string Path, string UserId), AccessEntry> _access = new();
+    private readonly ConcurrentDictionary<AccessCacheKey, AccessEntry> _access = new();
+
+    /// <summary>
+    /// The key of the permission-probe cache below.
+    ///
+    /// <para>🚨 <b>It carries every <see cref="AccessContext"/> dimension the evaluation READS,
+    /// not just the identity.</b> <c>PermissionEvaluator.GetEffectivePermissions</c> is not a
+    /// function of <c>(path, userId)</c> alone: it also reads <see cref="AccessContext.IsApiToken"/>
+    /// and the token's claim roles (the API-token capability clamp, which zeroes the WHOLE
+    /// permission set when a bearer context lacks <c>Api</c>), and <see cref="AccessContext.IsHub"/>
+    /// (the hub-credential early return). Keying on <c>(path, userId)</c> alone made two contexts
+    /// for the SAME person share one verdict, and the cache is a process-wide singleton, so the
+    /// two surfaces really do meet in it:</para>
+    ///
+    /// <list type="bullet">
+    /// <item><b>Too permissive (a capability bypass).</b> A portal read caches the unclamped
+    /// permissions; an MCP/bearer read of the same node within the TTL takes that entry and
+    /// never runs the API-token clamp at all — the token is admitted to the API surface the
+    /// clamp exists to keep it off.</item>
+    /// <item><b>Too restrictive (the reported symptom).</b> A bearer read the clamp zeroed caches
+    /// <c>Permission.None</c>; the same user's PORTAL read within the TTL then fails with
+    /// "lacks Read permission", on a node they are plainly granted.</item>
+    /// </list>
+    ///
+    /// <para>Both directions are wrong ANSWERS rather than errors, which is why nothing logged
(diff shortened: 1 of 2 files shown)


rbuergi:
## Flagging the direction this PR fixes that #92 never reported

MeshWeaver.SocialMedia#92 reports only the **denial** half — a token that cannot read what the portal
shows. The cache defect is symmetric, and the other direction is the serious one:

| order of arrival | cached verdict | effect |
|---|---|---|
| bearer read first, then portal | clamped (`Permission.None`) | the user's own portal is denied — **#92's symptom** |
| portal read first, then bearer | **unclamped** | **the API token inherits the portal's verdict — a capability bypass** |

Because the cache is a **process-wide singleton** keyed `(path, userId)`, which of the two a caller
gets is decided by *whoever happened to read that path first within the TTL* — not by the caller's own
credentials. The API-token clamp exists precisely so a bearer token is not simply "the user"; a cache
hit lets it become exactly that.

So this is not only a false-negative annoyance. **It is an access-control decision that can be
answered from another principal's context**, and the benign-looking direction is the one users
report while the dangerous one is silent by construction — nobody files a ticket saying "I could
read more than I should have".

Worth carrying in the title/labels as a security fix rather than a caching fix, and worth noting for
whoever reviews: the 30 s TTL bounds the *window*, not the severity — one hit inside it is enough.

(#92 itself should stay open regardless: a 30 s TTL cannot explain a week-long denial, so there is a
second cause there still to find — the two live reads that would settle it are named on that issue.)

Issue #2656, opened 2026-08-29
Inbound mail to the memex mailbox should create a Thread — the third instance of the triage-lane pattern, and the only one not built


rbuergi:
## What

Mail sent to the memex mailbox should be **received by memex and turned into a Thread** — the same
shape as the log-incident triage lane: an inbound artefact creates a thread, and an agent works it.
Plus an option on that thread to **forward to `info@…`**.

## What exists today

The **outbound** half is complete. `memex/Memex.Portal.Shared/Email/` holds:

```
OutboundEmailSender.cs · OutboundSendQueue.cs · InvitationEmailSender.cs
NoOpEmailSender.cs · EmailDeliveryGuard.cs
```

The **inbound plumbing is declared but nothing consumes it**:

- `EmailOptions.InboundEnabled`, `WebhookBaseUrl`, `SubscriptionClientState` (`src/MeshWeaver.Mesh.Contract/EmailOptions.cs`)
- `GraphSubscriptionState` (`src/MeshWeaver.Mesh.Contract/GraphSubscriptionState.cs`)
- the `/api/email` webhook, which validates `SubscriptionClientState` and rejects mismatches

So a Graph change-notification can arrive and be authenticated, and then there is no handler that
turns a delivered message into a node.

## The shape to follow — deliberately not a new one

This is the **third** instance of one pattern, and it should look like the other two rather than
invent a third vocabulary:

| lane | inbound artefact | becomes |
|---|---|---|
| log incidents | a crash/error report | an incident node → triage thread → agent |
| GitHub issues | an issue/PR event | a thread an agent works |
| **email (this)** | a delivered message | **a thread on the memex inbox** |

Worth reading `LogIncidentControlPlane` and the issue-triage lane first and matching their claim
lifecycle (claim → perform → record failure ON the artefact), rather than writing a bespoke one.

## Requirements

1. An inbound message creates a **Thread** node under the memex inbox, carrying sender, subject,
   body and message id.
2. **Idempotent on Graph's delivery guarantees.** Change notifications are at-least-once; the same
   message id must not create a second thread. (The install/reconcile code in this repo has the
   same property and states it explicitly — follow that.)
3. **Threading**: a reply to an existing conversation joins that thread rather than starting one.
4. A **forward-to-`info@…` action** on the thread.
5. The write must be a **registered NodeType** through the owning hub's canonical verbs — an
   invented `nodeType` is refused at the store, and only a test that EXECUTES the write proves the
   registration.

[...]

PR #2668, opened 2026-08-29
An outbound forward stops going out as "Re: Fwd: …" (#2656)

rbuergi:
Follow-up to #2656 (implemented in Systemorph/MeshWeaver.Plugins#873). Independent of it — that PR
does not compile against this one.

## The defect

`OutboundEmailSender` adds a reply marker unless the subject already has one, and "already has one"
was `StartsWith("Re:")`:

```csharp
var subject = email.Subject.StartsWith("Re:", StringComparison.OrdinalIgnoreCase)
    ? email.Subject : $"Re: {email.Subject}";
```

That is right for the case this lane was built for — the Email Router's reply, whose subject the
agent writes as `Re: <original>` anyway — and wrong for every outbound mail that is **not** a reply.
The forward action of #2656 is what made it visible: the inbox lane queues
`Fwd: Quote for 200 seats` and it leaves the building as `Re: Fwd: Quote for 200 seats`, which tells
the recipient the opposite of what happened to their mail.

## The fix

The check now recognises the same leading reply/forward markers `EmailInboundProcessor.ThreadKey`
strips when it normalises a subject into a conversation key — the two ends of one conversation should
agree about what a marker looks like. Only a **leading** marker counts: this decides whether to add
one, not what the subject means, so `"Question about the RE: header"` still gets its `Re:`.

Extracted as `OutboundSubject`, so the rule is a pure function with its own test rather than a lambda
inside a hosted service.

## Non-vacuity, measured

`OutboundSubjectTest` — 12 cases. Reverting the implementation to `StartsWith("Re:")` turns **6 of
the 12** red (every forward marker, in both languages, plus the leading-whitespace case). The whole
`Memex.Portal.Shared.Test` suite is 437/437 green with the fix in place.

## Also

`Doc/Architecture/EmailIngestionAndNotifications` is brought up to date. It still described the
non-user lane as "notify admins (in-app)" with no agent — which #2656 changes — and its moving-parts
table still named `Memex.Portal.Shared.Email.GraphMail` and `EmailWebhookController`, both of which
moved into the `MeshWeaver.Mail.MicrosoftGraph` module with #2276. The doc now also states the
at-least-once delivery contract and the two claims that make the lane idempotent.

### DIFF
--- a/memex/Memex.Portal.Shared/Email/OutboundEmailSender.cs
+++ b/memex/Memex.Portal.Shared/Email/OutboundEmailSender.cs
@@ -165,20 +165,36 @@ public sealed class OutboundEmailSender(
     /// left the building as <c>Re: Fwd: Quote for 200 seats</c>, which tells the recipient the
     /// opposite of what happened.</para>
     ///
-    /// <para>The markers are the same set <c>EmailInboundProcessor.ThreadKey</c> strips when it
-    /// normalises a subject into a conversation key — the two ends of the same conversation should
-    /// agree about what a reply/forward prefix looks like. Only the LEADING marker matters: this
-    /// decides whether to add one, not what the subject means.</para>
+    /// <para>Only the LEADING marker matters: this decides whether to ADD one, not what the subject
+    /// means. "Question about the RE: header" is not a reply.</para>
+    ///
+    /// <para>🚨 <b>The marker set is a deliberate SECOND COPY of the one
+    /// <c>EmailInboundProcessor.ThreadKey</c> strips</b>, and the duplication cannot be removed from
+    /// here today. That processor is in the <c>MeshWeaver.Mail.MicrosoftGraph</c> MODULE, in the
+    /// plugins repository, which depends on this one — the reference cannot run the other way, so a
+    /// shared constant would have to live in the mesh CONTRACT. Putting it there is a new framework
+    /// API, and a module may not call an API newer than the newest RELEASED framework without
+    /// raising its <c>minMeshVersion</c> floor; so "de-duplicate it" today means shipping a floor
+    /// bump for a list of natural-language prefixes that changes about never. The copies are
+    /// therefore left standing and each NAMES the other, so whoever edits one is told where the twin
+    /// is. Fold them into the contract when a floor bump is being paid for anyway.</para>
     /// </summary>
-    /// <param name="subject">The subject as queued.</param>
+    /// <param name="subject">The subject as queued; null is treated as empty.</param>
     /// <returns>The subject to send.</returns>
     internal static string OutboundSubject(string? subject)
     {
-        var trimmed = (subject ?? "").TrimStart();
-        return ReplyOrForwardPrefix.IsMatch(trimmed) ? subject! : $"Re: {subject}";
+        // One normalisation, not two: the pattern's own `^\s*` handles the leading whitespace, so
+        // trimming here as well would be a second answer to the same question — and returning the
+        // TRIMMED string would silently rewrite a subject this method was only asked to classify.
+        var queued = subject ?? "";
+        return ReplyOrForwardPrefix.IsMatch(queued) ? queued : $"Re: {queued}";
     }
 
-    /// <summary>Leading reply/forward markers, several languages — the set <c>ThreadKey</c> strips.</summary>
+    /// <summary>
+    /// Leading reply/forward markers, several languages. 🚨 Twin of the pattern in
+    /// <c>EmailInboundProcessor.ThreadKey</c> (plugins repo) — change both; see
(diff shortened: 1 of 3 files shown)


rbuergi:
Thanks — four of the five taken, pushed as `efc32c7c`.

**Double normalisation + the null-forgiving operator** (comments 2 and 3) — one defect, and you are
right that they were the same one. `OutboundSubject` trimmed the subject *and* used a pattern that
already begins `^\s*`; the `subject!` existed only to paper over the local the trimming had
introduced. It now classifies the queued string directly.

One subtlety worth recording, because it is the reason not to "fix" this by returning the trimmed
value: this method is asked to **classify** a subject, not to clean it. Returning the trimmed copy
would silently rewrite a subject the caller only asked about. Behaviour is unchanged and pinned by
the leading-whitespace case (`"  Fwd: Quote for 200 seats"` comes back byte-identical).

**The duplicated marker set** (comment 4) — agreed that it is a genuine drift risk: a marker added at
one end and not the other splits a conversation, silently. It cannot be made one constant today, and
the reason is structural rather than a preference. `EmailInboundProcessor` is in the
`MeshWeaver.Mail.MicrosoftGraph` **module**, in the plugins repository, which depends on this one —
the reference cannot run the other way, so a shared constant would have to live in the mesh
**contract**. That is a new framework API, and a module may not call an API newer than the newest
*released* framework without raising its `minMeshVersion` floor. Shipping a fleet-wide floor bump for
a list of natural-language prefixes that changes about never is the worse trade.

So both copies now **name the other and say why they are two** — the reciprocal comment is in
Systemorph/MeshWeaver.Plugins#873 — with a note to fold them into the contract when a floor bump is
being paid for anyway. That makes the next edit find its twin instead of discovering it in production.

**The contradictory heading** (comment 5) — a fair catch, and it said the opposite of what it meant.
Delivery stays at-least-once; that is the transport's contract and nothing here changes it. What the
lane guarantees is that the second and later deliveries of one message have no *effect*. Now reads
**"Processed once, because Graph delivers at least once."**

**Not taken: the `<b>` tags and 🚨** (comment 1). That is this repository's house style for a doc
comment carrying a war story — `EmailOptions`, `GraphSubscriptionState` and `LogIncidentControlPlane`
next door all read the same way, and the emoji is a deliberate scanning marker for "this paragraph is
load-bearing". A lone file written in a different register would be the actual inconsistency. Happy
to revisit if the house style changes, but not one file at a time.

Non-vacuity re-measured after the refactor, since it changed the branch under test: reverting to
`StartsWith("Re:")` still turns **6 of the 12** cases red. Full `Memex.Portal.Shared.Test` 437/437.

rbuergi:
## `Run tests (shard 2)` is red on a test this diff cannot reach

Failing test: `MeshWeaver.PathResolution.Test.HierarchicalBrowsingTests.Query_WithSkipAndLimit_ReturnsPaginatedResults`.

**The dependency direction settles it.** `test/MeshWeaver.PathResolution.Test` references exactly two
projects:

```
src/MeshWeaver.Hosting.Monolith.TestBase
src/MeshWeaver.Graph
```

This PR's only production change is `memex/Memex.Portal.Shared/Email/OutboundEmailSender.cs`.
`Memex.Portal.Shared` is *downstream* of the framework — nothing under `src/` references it — so it
cannot affect that suite's compilation or its behaviour. The whole diff is three files:

```
memex/Memex.Portal.Shared/Email/OutboundEmailSender.cs
src/MeshWeaver.Documentation/Data/Architecture/EmailIngestionAndNotifications.md
test/Memex.Portal.Shared.Test/OutboundSubjectTest.cs
```

A hosted service's subject-prefix helper, a markdown doc, and a new test in a different test project.

**The failure has a flake shape.** The test boots a mesh, writes a 7-node hierarchy, then asserts
`Skip`/`Limit` pagination and cross-page distinctness over a query whose index trails the store — the
classic eventual-consistency window. It asserts exact counts (`HaveCount(7)`, then 3/3/1) with no
wait-until, so a snapshot taken before the last write is visible fails it.

**What I could not establish:** whether it is currently failing on `main` too. Core `main` is moving
fast enough that the recent `MeshWeaver Build and Test` runs are all `cancelled` (superseded), so there
is no clean trunk comparison to point at right now. I am relying on the "subsystem this change cannot
reach" half of the evidence, not on a matching trunk red.

**Next step:** re-run the failed shard once the workflow finishes — `gh run rerun --failed` is refused
while it is still running. If it recurs on the re-run, it is not a flake and I will stop and say so
rather than re-running it again.

Note this *does* gate the PR: `main`'s ruleset requires exactly one context, `Consolidate test results`,
which aggregates the shards — so a red shard 2 keeps this blocked even though the shard itself is not a
listed required check.

Issue #1701, opened 2026-08-17
Package-root hub recycle wedges dispose for 8s and every NodeType compile in that window fails terminally — 'Retry the read' is advice nothing follows


rbuergi:
Found while wiring task #74 (satellite CI bakes); deterministic local repro via `mw-plugin-test`
over the MeshWeaver.Reinsurance tree WITHOUT Store staged (all timestamps one run, idle machine,
framework = main + #1700):

1. `Ifrs17` (the repo's one `Store/Plugin`-typed root; Store not in the mount) lands on the
   unresolvable-NodeType error overlay.
2. Its install-time recycle then hits a DISPOSAL DEADLOCK:
   ```
   fail: DISPOSAL DEADLOCK DETECTED: Hub Ifrs17 did not complete shutdown within 00:00:08.
         RunLevel=DisposeHostedHubs. Forcing out-of-band teardown …
   warn: [FORCE-TEARDOWN] Ifrs17: out-of-band teardown complete after 8001ms
   ```
3. During that 8+ s window the compile sweep for the package's 33 NodeTypes reads the root and
   every read fails TERMINALLY:
   ```
   warn: CompileWatcher: Compile failure for Ifrs17/EconomicBasis: InvalidOperationException:
         GetMeshNode('Ifrs17'): the owning hub answered ShuttingDown twice — the address is
         recycling, NOT absent. … Retry the read.
   ```
   `GetMeshNode` deliberately surfaces "recycling ≠ absent — Retry the read", but
   `CompileWatcher` does not retry: all 33 types settle `compile=FAILED(Error)` and the gate is
   red. On a portal the same race parks a whole package's types on the error overlay after any
   root-hub recycle that loses this race.
4. Adjacent (same run): the overlay self-heal watcher on a type node that does not exist ANYWHERE
   faults immediately (`DeliveryFailureException: No node found at 'Store/Plugin'`) — the stream
   errors instead of waiting for the node to appear, so the watcher is dead from birth and the
   "self-heal on type arrival" contract silently does not exist for the absent-type overlay case.

Two root-cause directions (no band-aids):
- the disposal deadlock itself — what in the package-root hub's teardown parks for 8 s
  (`RunLevel=DisposeHostedHubs`); the force-teardown is the backstop, not the fix;
- `CompileWatcher` (and any reader of `GetMeshNode`'s recycling verdict) must treat
  "recycling" as transient — resubscribe/wait for the address to reactivate, never settle a
[...]

Issue #1599, opened 2026-08-14
SilentReadNackTest.OwnerDisposedWithReadOutstanding is 35% flaky on main — two disposal arms race, and the test asserts a deterministic outcome over it


rbuergi:
`MeshWeaver.Hosting.Monolith.Test.SilentReadNackTest.OwnerDisposedWithReadOutstanding_IsNacked_NotLeftHanging` fails **21 of 60 times** on `main`, unmodified, on the 2-vCPU runner.

## The measurement

`flake-repro.yml`, `rate` mode, `--ref main` — [run 31818285657](https://github.com/Systemorph/MeshWeaver/actions/runs/31818285657):

```
RESULT: 21 failure(s) in 60 iteration(s) run (requested 60);
failing iterations: 3 7 14 15 18 20 22 23 25 29 30 35 36 37 40 41 42 43 53 56 58
```

That is **35%** — by a wide margin the highest rate measured on this repo (#1557's route-collapse was ~4%, and that was considered severe enough to fix).

## What it fails on

The test pins `HandleGetDataRequest`'s DISPOSAL arm specifically, and its own doc-comment says why that matters:

> this pins HandleGetDataRequest's DISPOSAL arm specifically; the routing-level NACK for an abandoned delivery carries different wording, so a passing assertion here cannot be satisfied by the pre-existing path instead of the one under test

But on a failing iteration the OTHER arm answers, and the assertion sees:

```
Expected "…the read faulted because the owning hub is shutting down — its hosted-hub creation is
frozen, so the data stream for this reference could not be created (HubDisposingException: …)"
to contain "still outstanding"
```

So **two disposal arms race**, and which one answers is not ordered: the hosted-hub-creation freeze can win over the outstanding-read NACK. The test asserts a deterministic outcome over that race, which is the same shape as #1557 — a guard that is flaky by construction because the mechanism under it is only probabilistically ordered.

🚨 And, exactly as with #1557, the interesting half is not the test. If the two arms can answer in either order, then a consumer whose read is outstanding when the owner is disposed gets **one of two different failures depending on timing** — and #1470's whole point was that distinguishable failures must stay distinguishable. Worth establishing whether the hosted-hub-frozen answer is merely a differently-worded NACK or a materially worse one for the caller (retryability, whether the read is even reported as failed).

## Why it does not redden main constantly

[...]

PR #2670, opened 2026-08-29
fix(mesh): a cross-hub update reports the owner's COMMIT VERDICT, never a bound expiring (#2661)

### LINKED ISSUE #2661: Write path fails OPEN: UpdateRemote's timeout emits an optimistic SUCCESS, and a late DeliveryFailure(Unauthorized) is never surfaced
## What

`MeshNodeStreamHandle.UpdateRemote` posts the `PatchDataRequest` and waits a bounded time for the owner's ack. On `TimeoutException` it **emits the optimistic snapshot and completes the caller's observable as a success**:

```csharp
else if (ex is TimeoutException)
{
    // No ack within the bound — the owner is busy (a round is running). Emit the
    // optimistic snapshot now rather than wait for the activity to finish; the
    // patch is posted and applies when the owner drains its queue.
    // The late watch STAYS armed — a late OwnerDisposing NACK proves the patch
    // will never apply and triggers the re-enqueue above.
```

The late watch is `LatePatchResponseRegistry`, which handles a late **`PatchDataResponse`**. An RLS denial is not a `PatchDataResponse` — `AccessControlPipeline` posts a **`DeliveryFailure{ErrorType.Unauthorized}`** ahead of the owner's action block. So a denial that loses the race against that bound reaches nothing, and the caller keeps an optimistic **success** for a write that was refused.

## Why it matters

The data is safe — the owner denies and the node is unchanged. The defect is that **the caller is told the write succeeded**. That is the silent-failure shape: a UI renders the optimistic value, a workflow proceeds on a write that never landed, and nothing anywhere reports a fault.

`MeshNodeStreamCache.Update` already documents the assumption this rests on:

> the denial is posted by the pipeline gate ahead of the owner's action block, so it returns within `UpdateRemote`'s short response bound even while a round is running

That assumption is not guaranteed — the bound also covers owner-hub **activation**, which can dominate under load.

## Evidence

`RlsIntegrationTests.StreamUpdate_WithoutUpdateRights_IsDeniedAndErrors` is the test that pins this. Observed failing:

```
[Warning] [MeshWeaver.AccessContext] AccessControlPipeline: Access denied: user 'viewer_noupd'
  lacks Update permission on 'rls/noupdate/NoUpdateStream'
  (triggered by message=PatchDataRequest sender=cache/... hub=rls/noupdate/NoUpdateStream)
```

The owner denied; the node was unchanged (the test's first assertion passes); the caller's `updateError` was `null`. It reproduces only under the full 376-test suite and is timing-sensitive — it passed on a later full-suite run with no code change, which is the signature of the race rather than a reason to dismiss it.

`INodeValidator`'s own remarks already name this area: *"the cause of the cache-only-write-gate CI flakes"*.

## The fix is a design decision, not a patch

The timeout path cannot currently distinguish **"the owner is busy, the patch is queued"** (safe to emit optimistically) from **"we do not yet know whether you are allowed"** (must not). Options, none of them free:

1. Keep the caller's observable open for the authorization verdict specifically, and fault on a late `Unauthorized`. Requires routing `DeliveryFailure` into the late-watch registry, and the caller must not have completed.
2. Establish authorization before the optimistic emit is permitted (a positive "authorized" signal, which does not exist today — only a denial).
3. Treat a late `Unauthorized` as a mirror-invalidation plus a loud error, accepting that the caller was already told it succeeded.

Deliberately **not** in scope for #2660 (the collaboration carve-out), which only made the race easier to hit by changing what a Markdown node hub registers.

## Not a fix

Raising the bound, retrying, or skipping the test. The bound is not the bug — the fail-open on bound expiry is.

rbuergi:
Fixes #2661.

## The diagnosis first — why `update` was the odd one out

`add` and `delete` settle on a real commit verdict: `workspace.RequestChange(...)` →
`DataChangeStatus.Committed`, else a genuine failure (`UpdateDataPath` / `DeleteDataPath`,
`MeshWeaver.Data/DataExtensions.cs`). `update` could not use that call — a cross-hub writer holds
only a **mirror** of the node, so its own workspace's `RequestChange` would commit into the mirror
and the owner would never see the write. So it ships an RFC 7396 `PatchDataRequest` to the owner and
takes the correlated `PatchDataResponse` as the verdict.

**The verdict machinery was never the problem.** The owner's ack is if anything *stronger* than
`Committed`: `ApplyMeshNodePatchInTurn` merges in-turn, waits for a post-commit emission that
provably contains this write (`ChangeContainsStampedWrite`), runs `IPostCommitFlush.Flush` for
durability, and only then acks — and it commits through the very same
`hub.GetWorkspace().RequestChange(...)` add/delete use, just executed on the owner.

What was wrong was only this: **the caller's terminal was decoupled from that verdict.**
`UpdateResponseWaitBound` (2 s) capped the wait, and on expiry `UpdateRemote` emitted the optimistic
snapshot and **completed the caller as a success**. A bound elapsing is not a commit. And a refusal
that lost the race against it reached *nothing*: an RLS denial is a
`DeliveryFailure{ErrorType.Unauthorized}`, not a `PatchDataResponse`, so
`LatePatchResponseRegistry` did not watch for it — `HandleCallbacks` found no live subject, logged
*"No subject found for response message"*, and marked the delivery processed. The caller kept
"saved" for a write the owner had refused.

## The fix

**`update` now settles on the owner's commit verdict, wherever it arrives.**

- **`UpdateResponseWaitBound` expiry emits nothing and completes nothing.** Its meaning is now what
  it was always structurally for: the point at which the wait moves off a pending `hub.Observe`
  callback (bounded by the hub's quiescing budget) and onto `LatePatchResponseRegistry`.
- **The registry watches `DeliveryFailure` too** (`DispatchFailure`, fed by a second handler on the
  cache hub). A late `Unauthorized` faults the caller with the *same* `UnauthorizedAccessException`
  the early-denial arm raises — "then error properly", on both paths.
- **Every late branch now settles the caller**: a late ack completes it as a success; a late
  terminal NACK faults it; a late re-enqueueable NACK chains the re-attempt's verdict to it instead
  of swallowing it into a log line.
- **Silence past `LateResponseWatchBound` faults** (`VERDICT_TIMEOUT`), it does not succeed. See
  "The one judgement call" below.
- **The stale comment is fixed.** `MeshNodeStreamExtensions.cs` claimed *"This runs before
  EmitTerminal, which is what releases the queue slot"*. Since #2346 the slot is released by the
  **hand-off** (`onLocalState`), never by the caller's terminal —
  `MeshNodeStreamCache.BuildUpdateQueueObservable` says so explicitly at the advance site. That
  stale sentence is what made this issue's own analysis fear the fix would stall the successor. It
  does not: holding the caller's terminal open leaves the queue untouched.

## The one judgement call — silence FAULTS

A write still unanswered at `LateResponseWatchBound` + 1 s is **not** reported as saved. This is
deliberate and it is the part to push back on if you disagree:

- Silence here is not "the owner is busy" — a busy owner acks late, and the late-ack arm completes
  the caller as a **success**. Silence here means the owner produced **no terminal at all** inside a
  window explicitly built to dominate every owner-side terminal path: the identity-gated ack
  watcher's 20 s `Timeout`, the 10 s post-commit flush, and `RegisterOwnerDisposingNack` for
  teardown — every one of which posts through `AckOnce`.
- Completing optimistically there would be the same fail-open one bound later, and #2661 says
  plainly that raising the bound is not the fix.
- The error is honest rather than absolute: *"The patch was posted and may still apply, but it is
  NOT confirmed — re-read the node and re-apply if the change is required."* The update lambda
  re-diffs idempotently, so a retry costs nothing.

**The caveat, so you can price it:** the owner's 20 s clock starts when its *handler* runs. An owner
whose action block is backed up more than ~10 s before that could exceed the window and get a write
faulted that would later have landed. If that turns out to be common, the correct response is to
widen `LateResponseWatchBound` (protocol, documented, and now also the caller's bound) — **not** to
reinstate the optimistic completion.

## The cost, priced

**MEASURED — healthy owners are unaffected, in aggregate and not merely in theory.** The early-ack
path is byte-identical, and two controls run against this branch's base commit say so:

| Suite | Base commit (no fix) | With the fix |
|---|---|---|
| `MeshWeaver.Security.Test` (write-heavy RLS) | 376/376 in **36 s / 32 s** (two runs) | 376/376 in **32 s** |
| `MeshWeaver.Hosting.Monolith.Test` | 752 passed / 3 skipped / **755** in **9 m 53 s** | 754 passed / 3 skipped / **757** in **9 m 54 s** |

The monolith delta is +2 tests and **+1 second** — and the +2 tests are *this PR's own*, which
contain ~12 s of deliberate "nothing should happen yet" waits by construction. The pre-existing 755
did not get slower. (An earlier reading of 10 m 19 s for the same suite was inflated by a
concurrently-running build; the 9 m 54 s figure is the clean one, taken against the same machine
state as the baseline.)

**MEASURED — busy owners now wait.** In the new test, with the owner's merge turn parked, the caller
is still open at 6 s where it previously completed with an unconfirmed value at ~2 s, and settles
when the owner commits.

**INFERRED — worst case.** A caller's maximum wait for a write goes from ~2 s to ~31 s. A caller
inside a request with its own 30 s budget now times out with its own error rather than being handed
a false success. That is the trade the maintainer called: *"for update, it has to wait until queue
drains."*

## Test evidence

New: `test/MeshWeaver.Hosting.Monolith.Test/WriteWaitsForCommitVerdictTest.cs`, two facts, both
**deterministic** — the interleaving is constructed (the owner's merge executor is parked behind a
gated turn, the same device `LateNackReenqueueTest` uses), not raced for. No cluster, no load, no
376-test-suite luck.

1. `Update_WhileTheOwnerIsBusy_ReportsSuccessOnlyWhenTheOwnerCommits` — the caller must still be
   waiting at 6 s, and must settle as a success only when the owner commits (asserted against
   durable storage).
2. `Update_WhenTheDenialArrivesAfterTheResponseBound_FaultsTheCaller` — the literal #2661 shape.
   The `DeliveryFailure{Unauthorized}` is injected onto the **real seam** it travels (posted to the
   cache hub, correlated by the in-flight patch's request id) after the response bound, and must
   fault the caller with `UnauthorizedAccessException` carrying the denial's own message.

**Non-vacuity, MEASURED.** With the single line `EmitTerminal()` restored in the `TimeoutException`
branch and nothing else changed, both fail on the assertion that matters — *"Expected value to be
false because a write must not be reported as saved before the owner has committed it, but found
True"*. With the fix, both pass.

## Suites run (all green, MEASURED)

| Suite | Result |
|---|---|
| `MeshWeaver.Security.Test` — the 376-test suite the issue names | 376/376 |
| `MeshWeaver.Hosting.Monolith.Test` | 754 passed / 3 skipped / 757 |
| `MeshWeaver.Data.Test` | 393/393 |
| `MeshWeaver.Persistence.Test` | 1037/1037 |
| `MeshWeaver.Query.Test` | 351/351 |
| `MeshWeaver.Content.Test` | 223 passed / 1 skipped |
| `MeshWeaver.NodeOperations.Test` | 76/76 |
| `MeshWeaver.AccessControl.Test` | 29/29 |
| `Memex.Portal.Shared.Test` | 425/425 |
| `MeshWeaver.Acme.Test` | 65/65 |
| `MeshWeaver.Hosting.Orleans.Test` (cross-silo write path) | 182/182 |
| `MeshWeaver.Markdown.Collaboration.Test` | 292/292 |
| `MeshWeaver.Documentation.Test` (doc links + What's New integrity) | 100/100 |
| `MeshWeaver.Todo.Test` | 6/6 |

Two existing tests pinned the *optimistic* terminal and were adapted — their invariants are
untouched, only the narration and the await/subscribe shape:
`QueuedWriteAdvancesOnHandoffTest` (which now also asserts the predecessor's caller settles on the
owner's verdict) and `LateNackReenqueueTest` (which now also asserts the re-enqueued attempt's
verdict reaches the original caller).

## Deliberately left

- **`Overwrite`** still writes through the sync stream and emits optimistically. It is a different
  protocol (full-node assertion for static-repo import, no correlated response) and #2661 names
  `UpdateRemote`. Worth its own issue, not this one.
- **The registry-less escape hatch.** On a hub with no `LatePatchResponseRegistry` (the
  `_bypassCache` path) there is no seam a late verdict could arrive on, so waiting could only hang.
  That branch keeps the optimistic completion and now says so at `Warning`.


rbuergi:
All three review findings were correct and are fixed in 6c0e659.

**1. The race in the verdict bound — a real bug, and the most valuable of the three.** The timer read `callerSettled` before acting, but `Dispatch` removes the registry entry and *then* runs the late callback: between those two steps a verdict is provably in flight while nothing has been claimed yet. A deadline firing in that window ran its side effects and could win the claim against a concurrently-arriving ack, faulting a write the owner had actually answered.

The fix is to arbitrate on the thing that actually carries the fact. The registry entry is the only object whose removal proves no verdict has taken this patch, so the timer now gates on `TryComplete(requestId)` and does nothing at all unless it wins — no `onLocalState`, no claim, no error. Losing the race means the verdict wins, which is the outcome we want. (`TryComplete` is additive next to `Complete`, so no signature changes.)

The concern about clearing the successor's base does not apply, for the record: the cache's hand-off callback only writes `_pendingSelfWrites` when `local is not null`, so `onLocalState(null)` releases the slot without disturbing a published base. But the claim race was real on its own.

**2. The bound reported in the diagnostic.** Correct — it named `LateResponseWatchBound` (30s) while the deadline fires at that plus `VerdictBoundGrace` (~31s). The effective bound is now computed once and reported verbatim in both the log and the error message. A diagnostic naming a deadline different from the one that fired is exactly the kind of thing that sends the next reader down the wrong path.

**3. The stale opening doc in `LatePatchWriteWatch`.** Correct — I updated the later sections and missed the first paragraph, which still described the pre-#2661 fallback. It now says what the 2s bound is actually for (the quiescing budget on a pending `hub.Observe` callback) and that the caller's own terminal now arrives on this registry.

Re-verified: the four write-path tests (`WriteWaitsForCommitVerdictTest`, `LateNackReenqueueTest`, `QueuedWriteAdvancesOnHandoffTest`) pass on the new binaries; `Security`, `Orleans` and `Monolith` are re-running locally.

rbuergi:
One local full-suite run of `MeshWeaver.Hosting.Monolith.Test` on the review-fix commit failed on `SilentReadNackTest.GetMeshNode_WhenOwnerIsDisposedMidRead_RecoversOnTheReProbe`. Recording it rather than re-running quietly, with the evidence for calling it a pre-existing flake this change cannot reach.

**Full-suite tally for this suite:**

| Run | Tree | Result |
|---|---|---|
| 1 | this branch (pre-detach-fix) | 754 passed / 3 skipped / 757 — green |
| 2 | this branch (post-detach-fix) | 754 passed / 3 skipped / 757 — green |
| 3 | this branch (post-review-fix) | 753 / 757 — **`SilentReadNackTest` failed** |
| 4 | this branch (post-review-fix, **same binaries as run 3**) | 754 passed / 3 skipped / 757 — green |
| — | base commit `82ab9f9` | 752 passed / 3 skipped / 755 — green |

Run 4 is the one that matters: identical bytes to the run that failed, green. The test also passes 5/5 in isolation.

**Why this change cannot reach it.** The failure is on the READ path. The whole diff touches `UpdateRemote` (the write path), `LatePatchResponseRegistry` (the write-side late watch), and one added `WithHandler<DeliveryFailure>` on the **`cache/{meshId}`** hub. This test reads through `GetClient(c => c.AddData())` — a **`client/*`** hub, which never carries that handler — so the NACK it depends on is not routed anywhere my diff observes. The only occurrences of `GetMeshNode` anywhere in the diff are two *deleted* comment lines.

**What the failure actually was.** The read resolved in **11 ms with a null** — it never reached a re-probe at all. That is a terminal-ordering race between `primary.Dispose()` and the read subscribing, which is precisely the subsystem this file is about: its own history carries `fix: stop asserting which of three racing terminals answered (#1599)` and `fix(#1701): GetMeshNode rides out a recycling address`.

Filed separately so it does not get buried here — it deserves a real fix in the read path, not an allow-list entry.

rbuergi:
Flake filed as #2673 (with the analysis of why answering `null` for a node that exists is the read-side shape of the same silent-failure class, and an explicit "do not allow-list it").

Issue #2088, opened 2026-08-23
Module-bundle lane publishes only on pushes — a release-follow schedule rebuild is thrown away


rbuergi:
Observed in the 2026-08-23 ci.5050 wedge: after a framework release, both prod portals' health said modules "FrameworkDeclined (built against s451d9a5, live see0c273c)" — the module-bundle lane (registry-served C# packages) had rebuilt on the release-follow schedule but its `publish:` condition is `github.event_name == 'push'`, so the build was discarded. Every framework release therefore breaks module adoption until an unrelated push to the plugin repo republishes (here: Plugins #599's merge republished 30/31 and unwedged it).

The node-repo bundle bake already publishes on the schedule lane; the module lane should too, with the same publication rules (complete set, carry-forward, sealed last). Same 'the evidence a lane did nothing is the absence of evidence' family as the bake-follows-release invariant (#8 in the satellite CI policy).

🤖 Filed from the 2026-08-23 fleet session.

rbuergi:
**Platform half fixed by #2256; the caller half is MeshWeaver.Plugins#668** (open). Leaving this open until both land.

Root cause is a documentation defect that propagated into every caller: `node-repo-module-pack`'s `publish` input said *"Trunk only"*, which taught callers to write `publish: ${{ github.event_name == 'push' }}`. So a **release-follow schedule run packed, versioned and inspected the bundles and handed them to nobody** — and module adoption stayed `FrameworkDeclined` until an unrelated push republished by accident.

The input is redocumented as **trunk-only, not push-only**, and the workflow gains the caller-contract block its sibling reusable workflows carry (modelled on `node-repo-publish-bake`): the three-way `if:`, the release-follow cron, and the shared `framework-release` concurrency lane.

rbuergi:
**Confirmed — and there is a prerequisite. Enabling schedule-publish today would turn the lane RED fleet-wide, not fix it.**

## The report is accurate

`MeshWeaver.Plugins/.github/workflows/ci.yml:201`:

```yaml
publish: ${{ github.event_name == 'push' && github.ref == 'refs/heads/main' }}
```

Observed directly rather than inferred — on a `schedule` run ([32938184228](https://github.com/Systemorph/MeshWeaver.Plugins/actions/runs/32938184228)) the module job reports:

```
13. Hand the bundle to the registry -> skipped
```

versus the same job on a `push` run, where it executes. So a release-follow rebuild is genuinely discarded, exactly as filed.

[...]

Issue #2550, opened 2026-08-27
New-instance setup wizard: an empty image boots into a setup GUI that chooses storage, boot modules, provisioned packages and per-user pre-installs


rbuergi:
## The idea

Install an **empty image**. It comes up with a setup GUI that asks:

1. **Which database** — pick from what this image actually ships, then enter the connection string.
2. **Which modules BOOT** — loaded into the process (`Modules:Assemblies` / `Modules:Required`).
3. **Which packages are PROVISIONED** — installed into the mesh from the registry at first boot.
4. **Which are PRE-INSTALLED FOR USERS** — landed per user on first logon.

And the same path creates a new instance in a **deployment**, rather than being a second mechanism that drifts from it.

## The constraint that shapes everything

Today an empty image **cannot boot at all**: `MemexConfiguration.ConfigureMemexMesh` throws

```
InvalidOperationException: Graph:Storage configuration is required.
Configure it in appsettings.json with Type and BasePath/ConnectionString.
```

So step one is not a GUI — it is making "no storage configured yet" a *legitimate boot state* instead of a fatal one.

## The artifact: an instance manifest, read before DI exists

The wizard's output is a durable **instance manifest** on the writable root (`ModuleRoot.Resolve`, the same volume as `modules/activation.d/`), not a mesh node — for exactly the reason the module activation sidecar is not one, quoting `Modules.md`:

> the list is consumed before any storage provider, hub, or connection string exists, and it must move with the DLLs it describes

A manifest carries: the storage selection (type + connection string, or a Key Vault secret NAME), the boot module set, the provision set, and the per-user pre-install set.

**Restart applies it** — restart-as-activation, the mechanism modules already use. Wizard writes the manifest → restart → `ConfigureMemexMesh` finds storage and boots for real.

## Boot states

| state | condition | behaviour |
|---|---|---|
| **Setup** | no manifest AND no `Graph:Storage` | in-memory mesh, no modules, serves ONLY the wizard |
| **Configured** | manifest or appsettings present | today's boot, unchanged |

An instance configured by appsettings (every deployment that exists now) never enters setup mode — the manifest is additive, and its absence is not an error.

## The three module questions are three EXISTING mechanisms

The wizard composes what exists; it must not invent a parallel notion of "installed".

| wizard question | mechanism today |
|---|---|
| boots | `Modules:Assemblies` + `Modules:Required` → `InstallConfiguredModules` |
[...]

Issue #886, opened 2026-08-07
TypeSource.WithUpdate documents a transform but the return value is discarded — contract is side-effect-only


rbuergi:
Split out of #678's "also noted" section so it isn't lost when that issue closes.

## 1. `WithUpdate`'s documented transform contract is dead code

`ITypeSource.Update(ChangeItem<EntityStore>)` returns the transformed `InstanceCollection`, and `TypeSourceWithType<T>.WithUpdate` documents *"Applies the configured update action to the incoming collection before it is stored"* — but both call sites (`TypeSourceBasedUnpartitionedDataSource.Synchronize` and the partitioned twin, src/MeshWeaver.Data/GenericUnpartitionedDataSource.cs) discard the return:

```csharp
foreach (var typeSource in TypeSources.Values)
    typeSource.Update(item);   // ← InstanceCollection return dropped
```

In practice every consumer (e.g. `DataTest`'s `SaveMyData`, `TypeSourceWithTypeWithDataStorage` which persists inside `UpdateImpl`) works by SIDE EFFECT, so nothing is broken today — but a user who writes a pure transform ("normalize before store") gets a silent no-op. Either make the return value meaningful (write the transformed collection back through the stream, with the source's own `Id` as `ChangedBy` so the echo filter drops it) or change the signature/docs to be honestly side-effect-only (`void`). The second is the smaller, truthful fix.

## 2. `DataChangeStatus.Committed` is acked before the write-back runs

`UpdateStream` completes (and `DataChangeResponse` maps to `Committed`) when the in-memory store applied the change; persistence is dispatched to a separate hub afterwards. Consistent with the design, but the name over-promises — callers reasonably read "Committed" as "durable". Consider `Applied`, or document the semantics on the enum.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #2671, opened 2026-08-29
CI: a PR builds only what it changed; a publish or a release-follow builds everything

rbuergi:
## The measurement

Two lanes fanned out over the whole repo on every push of every PR. Measured on `MeshWeaver.Plugins` this morning, they were **29 of the 31 queued runs** on that repo, while two clean, zero-failure PRs sat on six pending checks each:

| lane | what it did per run | cost |
|---|---|---|
| `Plugin Catalog CI` → `modules` → `node-repo-module-pack.yml` | a **hardcoded literal 29-entry matrix**, on every event | 29 jobs per PR push |
| `Publish Plugin Packages` → `plugin-publish.yml` | `for plugin in plugins/*/` | **53 packages** in one job (`Store` alone = 17 compilation units) |

`scripts/affected-modules.py` already computed a closure — and was wired only to the mesh gate and the bake.

## What this changes

Both lanes now ask one script, `.github/scripts/node-repo-scope.py`, what the diff can reach.

**Only `pull_request` / `merge_group` narrow.** `push`, `repository_dispatch`, `schedule` and `workflow_dispatch` build the FULL set — a module is built against a platform PIN, and when that pin moves every bundle must be rebuilt and republished or every portal reads `FrameworkDeclined` and adopts nothing (#2088). The pack lane's value is that same property from the compile side: a package that did not change can stop compiling when the published framework advances.

**`push` is full for a second, separate reason.** The baseline a *publishing* run must diff against is its own publication, never `github.event.before` (which silently under-builds after any cancelled, superseded, red or re-run commit). `bake-scope.sh` has that marker — `source-commit.txt`, sealed beside the bundles. **The module lane has none**: the registry is keyed `{package}@{version}` from `manifest.lock`, and that version identifies only the CONTENT half, so a `src/`-only change repacks the same version with different bytes. Narrowing a push needs that marker first; noted in the script.

## Two halves, because a bundle has two

- **content** — the caller's own affected closure, `scripts/affected-modules.py`, **invoked rather than reimplemented**.
- **compiled** — the caller's `scripts/project-closure.py`: the csproj, its transitive in-repo `ProjectReference`s, and the sibling `.Test` project this lane runs. `affected-modules.py` cannot answer this — it says ALL modules for any `src/` path, which is right for the mesh gate and useless for a matrix whose entries ARE projects.

Both graphs live in the CALLER repo because the edges are the caller's content — the split `bake-scope.sh` already uses. A repo shipping neither is simply not narrowable: the scope script answers FULL and says why. Nothing to opt into, nothing that can silently skip.

`project-closure.py` is deliberately a **shared** answer, not a private helper: Systemorph/MeshWeaver.Plugins#878 (a mixed package's `src/`-only change moves no version, so the rebuilt bundle is shelved and no existing portal ever asks for it) needs the same graph for its version gate.

## The count is asserted — and it is the context to require

🚨 A narrowed fan-out **cannot** be a required status check: the per-module contexts appear and disappear with the diff, so requiring one means requiring a context that will legitimately never report — which is how a lane ends up with no required gate at all.

So the SELECTION is what gets checked. Every pack leg drops a receipt as its **last** step; a new final job — **`All selected bundles built`**, present on every run whatever the diff, `if: always()` — fails when the receipts and the selection disagree: a selected bundle with no receipt, a receipt nobody selected, a non-empty selection whose pack job never ran, a failed or failing `select`. The pack lane asserts the same thing inline.

## Evidence

Both new scripts carry `--self-test` and **both run it in CI, in the job that produces the answer**:

- `node-repo-scope.py --self-test` — 23 full-run fallbacks, 8 narrowing assertions, 2 explicit-empty answers, 5 caller-graph refusals.
- `node-repo-pack-verify.py --self-test` — 2 green-run assertions and **7 mutations of that green run** (remove a receipt, add an unexpected one, skip the pack job, fail it, corrupt a receipt, mismatch its name, lose the whole directory) that must each turn red *and name what they found*, plus 2 empty-selection cases.

Measured narrowing on `MeshWeaver.Plugins`' real 29-entry matrix, over real git ranges:

| changed path | before | after |
|---|---|---|
| `src/MeshWeaver.OgCard/OgCard.cs` | 29 | **2** (OgCard, and Export whose `.Test` references it) |
| `src/MeshWeaver.Markdown.Export.Test/**` | 29 | **1** |
| `src/MeshWeaver.AI/**` | 29 | 15 |
| `src/MeshWeaver.Blazor/**` | 29 | 13 |
| `Chess/index.json` | 29 | **0** |
| `AI/index.json` | 29 | 2 |
| `Store/Plugin/Source/**` | 29 | 29 — Store really is a universal dependency |
| `src/Directory.Packages.props`, `.github/**`, `scripts/**`, a repo-root file | 29 | 29, with the fallback **named** in the log and the job summary |

## Blast radius

- `node-repo-module-pack.yml` — called by `MeshWeaver.Plugins` (`@main`, 29 entries) and `MeshWeaver.SocialMedia` (pinned at `89f8862`, so unaffected until its pins are bumped together).
- `plugin-publish.yml` — called by `MeshWeaver.Plugins` only (`@main`).
- No other satellite calls either lane; each falls back to FULL if it ships neither graph.

**Required contexts:** neither repo requires any module-bundle context today (measured: `MeshWeaver.Plugins` requires `Validate node repos`, `Compile every NodeType (vs core)`, `Compile + render node repos (MeshWeaver from ACR)`; SocialMedia the same three, path-prefixed). So nothing breaks. The follow-up ask is that `Module bundles / All selected bundles built` be **added** — that is a branch-protection edit and therefore the maintainer's.

### DIFF
--- a/.github/scripts/node-repo-pack-verify.py
+++ b/.github/scripts/node-repo-pack-verify.py
@@ -62,7 +62,8 @@ def read_receipts(directory: Path) -> tuple[dict[str, dict], list[str]]:
 
 
 def verify(expected: list[str], receipts: dict[str, dict], broken: list[str],
-           pack_result: str, receipts_dir_exists: bool) -> tuple[int, list[str], list[str]]:
+           pack_result: str, receipts_dir_exists: bool,
+           scope: str = "") -> tuple[int, list[str], list[str]]:
     """(exit code, error lines, note lines)."""
     errors: list[str] = []
     notes: list[str] = []
@@ -72,6 +73,17 @@ def verify(expected: list[str], receipts: dict[str, dict], broken: list[str],
     if broken:
         errors.append(f"{len(broken)} receipt file(s) could not be read as a receipt: "
                       + ", ".join(broken))
+
+    # 🚨 "BUILD EVERYTHING" AND "EVERYTHING IS NOTHING" CANNOT BOTH BE TRUE. A `full` scope with
+    # an empty expectation is an internal contradiction, and the shape it takes is a green tick
+    # over a lane that built nothing — a bad --root, an empty matrix input, a tree that could not
+    # be listed. The selector refuses it now; this is the second lock, because the two are
+    # computed in different jobs and only one of them has to be wrong.
+    if scope == "full" and not want:
+        return 1, [f"the scope is '{scope}' but the selection is EMPTY — 'build everything' and "
+                   "'everything is nothing' cannot both be true. Something upstream (a --root "
+                   "that is not the checkout, an empty matrix input) produced a contradiction, "
+                   "and nothing was built."], []
 
     if not want:
         # An empty selection is legitimate (a diff reaching no package and no module project) —
@@ -134,9 +146,9 @@ def self_test() -> int:
         if not ok:
             failures.append(name)
 
-    def run(directory: Path, expected: list[str], result: str = "success"):
+    def run(directory: Path, expected: list[str], result: str = "success", scope: str = ""):
         receipts, broken = read_receipts(directory)
-        return verify(expected, receipts, broken, result, directory.is_dir())
+        return verify(expected, receipts, broken, result, directory.is_dir(), scope)
 
     with tempfile.TemporaryDirectory() as tmp:
         d = Path(tmp) / "receipts"
@@ -199,6 +211,17 @@ def self_test() -> int:
         code, errors, notes = run(empty, [], "skipped")
         check("zero selected + zero receipts + a skipped job ⇒ pass, and says so",
               code == 0 and bool(notes), f"{errors}")
+        code, errors, notes = run(empty, [], "success")
+        check("zero selected + zero receipts + a job GitHub reported success ⇒ pass",
+              code == 0 and bool(notes), f"{errors}")
+        # 🚨 The contradiction lock. Without the scope, this is indistinguishable from the case
+        # above — which is why the count alone was not enough.
+        code, errors, _ = run(empty, [], "skipped", scope="full")
+        check("scope=FULL with an empty selection ⇒ fail (a contradiction, not a scope)",
+              code == 1 and any("cannot both be true" in e for e in errors), f"{errors}")
+        code, errors, _ = run(empty, [], "skipped", scope="narrowed")
+        check("…while scope=narrowed with an empty selection stays legitimate",
+              code == 0, f"{errors}")
         (empty / "MeshWeaver.AI.json").write_text(json.dumps({"module": "MeshWeaver.AI"}),
                                                   encoding="utf-8")
         code, errors, _ = run(empty, [], "skipped")
@@ -212,7 +235,8 @@ def self_test() -> int:
               "skipped module bundle.")
         return 1
     print("\n✓ node-repo-pack-verify self-test: 2 green-run assertions, 7 mutations that must go "
-          "red, 2 empty-selection cases — all green.")
+          "red, and 5 empty-selection cases including the scope=full contradiction lock — all "
+          "green.")
     return 0
 
 
@@ -223,6 +247,9 @@ def main() -> int:
     p.add_argument("--receipts", default="", help="directory the receipt artifacts were merged into")
     p.add_argument("--pack-result", default="", dest="pack_result",
                    help="the pack job's aggregated result (needs.pack.result)")
+    p.add_argument("--scope", default="",
+                   help="the selection's own scope (full|narrowed). A 'full' scope with an empty "
+                        "selection is a contradiction and is refused.")
     p.add_argument("--self-test", action="store_true", dest="self_test")
     args = p.parse_args()
     if args.self_test:
@@ -233,7 +260,8 @@ def main() -> int:
     directory = Path(args.receipts)
     expected = [m for m in args.expected.split() if m]
     receipts, broken = read_receipts(directory)
-    code, errors, notes = verify(expected, receipts, broken, args.pack_result, directory.is_dir())
+    code, errors, notes = verify(expected, receipts, broken, args.pack_result,
+                                 directory.is_dir(), args.scope)
 
     for note in notes:
         print(f"✓ {note}")
--- a/.github/scripts/node-repo-pack-verify.py
+++ b/.github/scripts/node-repo-pack-verify.py
@@ -1,0 +1,253 @@
+#!/usr/bin/env python3
+"""node-repo-pack-verify.py — assert that EXACTLY the selected module bundles were built.
+
+    "we can estimate the required count and validate it"
+
+A dynamic matrix cannot be a fixed list of required status checks: the per-module contexts appear
+and disappear with the diff, so requiring one means requiring a context that will legitimately
+never report. That is not a detail — it is the same trap as an `if:` on a gate, and it is how a
+gate ends up not required at all and PRs merge past it.
+
+The answer is to make the SELECTION itself checkable: `node-repo-scope.py` computes how many
+bundles must be built, every pack job drops a RECEIPT, and this script — one final job, one
+STABLE context, present on every run whatever the diff — asserts the two agree and names the
+discrepancy. That is CI invariant #4 (a final job that fails iff a real gate skipped for an unsafe
+reason) applied to a matrix that is allowed to change size, and it is what makes narrowing safe to
+require.
+
+🚨 THE FAILURE IT EXISTS TO CATCH IS A SKIP. GitHub renders a skipped matrix leg with the same
+tick as a passed one, so "this module was never packed" and "this module packed fine" look
+identical on the checks wall. A receipt is positive evidence; its absence is the finding.
+
+    node-repo-pack-verify.py --expected "MeshWeaver.AI MeshWeaver.Mcp" \\
+                             --receipts "$RUNNER_TEMP/receipts" --pack-result success
+    node-repo-pack-verify.py --self-test
+
+EXIT 0 only when: every expected module has a well-formed receipt, no unexpected receipt exists,
+and the pack job's own result agrees with the expectation (an empty selection is the ONLY case in
+which the pack job may be `skipped`).
+"""
+from __future__ import annotations
+
+import argparse
+import json
+import os
+import sys
+from pathlib import Path
+
+
+def read_receipts(directory: Path) -> tuple[dict[str, dict], list[str]]:
+    """{module → receipt} plus the files that could not be read as one."""
+    found: dict[str, dict] = {}
+    broken: list[str] = []
+    if not directory.is_dir():
+        return found, broken
+    for path in sorted(directory.rglob("*.json")):
+        try:
+            doc = json.loads(path.read_text(encoding="utf-8"))
+        except (OSError, json.JSONDecodeError):
+            broken.append(path.name)
+            continue
+        module = doc.get("module") if isinstance(doc, dict) else None
+        if not isinstance(module, str) or not module:
+            broken.append(path.name)
+            continue
+        # The receipt must be self-describing: a file called X.json claiming to be module Y means
+        # the matrix and the artifact names have drifted, and the count would then prove nothing.
+        if path.stem != module:
+            broken.append(f"{path.name} (claims module '{module}')")
+            continue
+        found[module] = doc
+    return found, broken
+
+
+def verify(expected: list[str], receipts: dict[str, dict], broken: list[str],
+           pack_result: str, receipts_dir_exists: bool) -> tuple[int, list[str], list[str]]:
+    """(exit code, error lines, note lines)."""
+    errors: list[str] = []
+    notes: list[str] = []
+    got = set(receipts)
+    want = set(expected)
+
+    if broken:
+        errors.append(f"{len(broken)} receipt file(s) could not be read as a receipt: "
+                      + ", ".join(broken))
+
+    if not want:
+        # An empty selection is legitimate (a diff reaching no package and no module project) —
+        # GitHub skips a job whose matrix is empty. What must NOT happen is a bundle being built
+        # that the selection did not name.
+        if pack_result not in ("skipped", "success"):
+            errors.append(f"the selection named ZERO module bundles, but the pack job reported "
+                          f"'{pack_result}' — a job that should not have run did something.")
+        if got:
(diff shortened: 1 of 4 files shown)


rbuergi:
Two things landed alongside this PR that reviewers should have in view:

**Executed before merge.** This changes the lane that publishes module bundles to the registry, and MeshWeaver.Plugins consumes both changed workflows at `@main` — so they go live the moment this merges. Five throwaway draft PRs on that repo point at this branch and execute the wiring first (each says what it expects, so a wrong answer is visible rather than inferred):

| PR | lane | expected |
|---|---|---|
| Systemorph/MeshWeaver.Plugins#883 | module | `full` — 2 of 2 (its own diff touches `.github/`) |
| Systemorph/MeshWeaver.Plugins#884 | module | `narrowed` — **1 of 2**, OgCard must not appear |
| Systemorph/MeshWeaver.Plugins#885 | module | `narrowed` — **0 of 2**, pack skipped, verify green *for the right reason* |
| Systemorph/MeshWeaver.Plugins#886 | pack | `full` — 53 of 53, loop count assertion agrees |
| Systemorph/MeshWeaver.Plugins#887 | pack | `narrowed` — **1 of 53** |

**The one deliberate gap**, filed as Systemorph/MeshWeaver.Plugins#889 so it is not mistaken for an omission: `push` builds everything because the module lane records no publication baseline to diff against. `bake-scope.sh` has `source-commit.txt`; this lane has only `{package}@{version}`, which identifies the content half alone — a `src/`-only change repacks the same version with different bytes. The issue lists the three ways to close it; the cheapest also fixes Systemorph/MeshWeaver.Plugins#878.

rbuergi:
## What this costs today, in runner-minutes

Measured from a real successful `Plugin Catalog CI` run (`33240892860`) by summing the actual per-job durations of the module-bundle matrix, and from the last eight `Publish Plugin Packages` runs:

| lane | per PR push |
|---|---|
| `Module bundles` (29 jobs) | **197 runner-minutes** |
| `Publish Plugin Packages` (1 job, 53 packages) | **25–40 minutes** |

Applying the closure to the same durations:

| changed path | jobs | module-bundle minutes | saved |
|---|---|---|---|
| `src/MeshWeaver.Markdown.Export.Test/**` — the flaky suite's own PRs | 1/29 | 9m | **95%** |
| `src/MeshWeaver.OgCard/**` — a leaf module | 2/29 | 12m | **94%** |
| `AI/index.json` — a mixed package's content | 2/29 | 15m | **93%** |
| `src/MeshWeaver.Blazor/**` — a shared view library | 13/29 | 69m | 65% |
| `src/MeshWeaver.AI/**` — the AI engine | 15/29 | 157m | 20% |
| `Chess/index.json` — a node-only package | 0/29 | 0m | **100%** |
| `src/Memex.Portal.Gui/**` — a portal host, no module | 0/29 | 0m | **100%** |

The AI-engine row is the honest one: that closure really is 15 modules wide, and it should be.

**Why it matters right now.** `MeshWeaver.Plugins` measured **43 queued runs / 5 in progress** while this was being written, and **39 of the 43 were these two lanes** (20 `Plugin Catalog CI` + 19 `Publish Plugin Packages`). Clean PRs with zero failures were sitting on pending checks for hours. The pool is starved, and this is what is eating it.

rbuergi:
## Executed. The matrix is genuinely dynamic in real CI.

Both module-lane integration PRs reached their verdict, and the **check names themselves are the evidence** — a narrowed matrix expands fewer legs, so the contexts that appear differ:

**Systemorph/MeshWeaver.Plugins#883** — diff touches `.github/`, so the scope must be FULL:
```
pass  Module bundles / Select the affected bundles
pass  Validate node repos
      Module bundles / Module bundle (MeshWeaver.OgCard)             ← 2 of 2
      Module bundles / Module bundle (MeshWeaver.Markdown.Export)
```

**Systemorph/MeshWeaver.Plugins#884** — diff is one file under `src/MeshWeaver.Markdown.Export`:
```
pass  Module bundles / Select the affected bundles
pass  Validate node repos
      Module bundles / Module bundle (MeshWeaver.Markdown.Export)    ← 1 of 2
```
`MeshWeaver.OgCard` **does not appear at all.** That is the narrowing, in GitHub, not in a script.

#885 (zero-selection) and #886/#887 (pack lane) are still queued behind the pool.

## …and the review found two things worth having

An independent review pass caught two **silent-wrong-answer** holes — both of which reported SUCCESS having built nothing, which is the exact failure this design exists to prevent. Fixed in the commits above, each with a self-test verified by mutation:

1. **`scope: full` with `count: 0` exited 0.** An empty universe took the `full` branch, which sets `count = len(universe) = 0`, and every consumer keys on the count alone. Reachable via a `--root` that does not match the checkout, a `modules` input of `[]`, or an unreadable tree (`node_packages` caught `OSError` and turned *"I could not read this"* into *"there is nothing here"*). Now a **refusal** — exit 2, named — not a scope; the `OSError` is no longer caught; and `verify` carries the same lock independently, because scope and count are computed in different jobs and only one has to be wrong.
2. **A stale matrix entry was silently never selected.** The guard tested only the `src/` prefix, never existence, so a renamed csproj or a vanished package simply could not be reached by any closure. Before narrowing, every PR built it and went red at once; after, the failure deferred to the next push on main — where this repo's policy says the cost must not land. Two filesystem checks, run **unconditionally** (not behind `if src_files`, or a diff touching no `src/` path would never look at the entry list).

Plus five smaller ones: `NOOP_DIRS` claimed to be verbatim and was missing `WhatsNew` (a note-only PR ran the whole fleet — measured 29, now **0**, and the script now reads the caller's set and refuses to narrow on drift); both count gates could only catch a lost output, never a wrong selection, so on a FULL scope they now cross-check against an independent source (the caller's declared matrix / the plugins in the tree); `merge_group` + `dry-run: false` could have **published** a narrowed subset, so `--publishing` forces FULL on the reusable side; `pull_request_target` dropped from `NARROWABLE`; `set -f` around the deliberate word-split and the job summary written before the failure branches.

Self-tests: **50** cases (scope, was 41) and **14** (verify, was 11).

rbuergi:
## The pack lane is executed too — and it produced a clean natural experiment

Both halves ran, on the same tree, minutes apart:

| PR | scope | result |
|---|---|---|
| Systemorph/MeshWeaver.Plugins#886 | `full` | `scope=full — the selection names 53 plugin(s)` → `packed 52 plugin(s) of 53 selected` → **red** |
| Systemorph/MeshWeaver.Plugins#887 | `narrowed` | `scope=narrowed — the selection names 1 plugin(s)` → `packed 1 plugin(s) of 1 selected` → **green** |

The FULL run went red on a **pre-existing break in `main`** — `Hosting/FleetConsole/Test/FleetConsoleTests.cs(697,93): error CS1513: } expected`, from the hand-resolved `bc9f20a5` SetupDialog UNION merge. `Publish Plugin Packages` had failed on `main` four times running because of it, and since the lane packs every plugin before it pushes any, **it was holding back the entire catalog**. Fixed in Systemorph/MeshWeaver.Plugins#894, filed separately.

Three things that experiment settles:

1. **The loop really iterates the selection** — 53 groups in package order for FULL, exactly one (`Publish`) for narrowed.
2. **The count assertion did not false-positive on a genuine failure.** `packed 52 … of 53` gives `seen = 52 + 1 = 53`, which matches, so the gate stayed quiet and the run failed as `packing failed for: Hosting` — the real cause, reported as itself rather than as a count discrepancy. That is the behaviour `--self-test` pins and this is it happening for real.
3. **The trade-off is visible and is the intended one.** A PR touching only `Publish/` no longer surfaces a break in `Hosting/` — and that is precisely why `push` is never narrowed. The break did surface, on `main`, exactly as designed. Nothing that reaches the trunk is skipped.

## Module lane, for completeness

| PR | expected | contexts that appeared |
|---|---|---|
| Systemorph/MeshWeaver.Plugins#883 | `full`, 2 of 2 | `Module bundle (MeshWeaver.OgCard)` **and** `Module bundle (MeshWeaver.Markdown.Export)` |
| Systemorph/MeshWeaver.Plugins#884 | `narrowed`, 1 of 2 | `Module bundle (MeshWeaver.Markdown.Export)` only |

All four verdicts match the expectation each PR stated before it ran.

rbuergi:
## Armed. All three module-lane paths are proven against the POST-review code.

The earlier verdicts were on the pre-review commit, so I re-ran them after the guards landed. Both post-review runs behave identically:

| PR | scope | contexts that appeared | on |
|---|---|---|---|
| Systemorph/MeshWeaver.Plugins#883 | `full` 2/2 | `Module bundle (MeshWeaver.OgCard)` + `(MeshWeaver.Markdown.Export)` | pre-review |
| Systemorph/MeshWeaver.Plugins#884 | `narrowed` 1/2 | `(MeshWeaver.Markdown.Export)` only | pre-review **and re-run post-review** |
| Systemorph/MeshWeaver.Plugins#885 | `narrowed` 0/2 | matrix un-expanded (`skipping`), **`All selected bundles built` PASS** | post-review |

#885 is the one I most wanted: an empty matrix means GitHub skips `pack`, and the count gate still **runs and passes for the right reason** rather than vanishing alongside the thing it watches. That is invariant #4 holding for a matrix that is allowed to change size.

Pack lane, from the same session: `scope=full — 53 plugin(s)` / `packed 52 of 53` (red on a real pre-existing break) versus `scope=narrowed — 1 plugin(s)` / `packed 1 of 1` (green).

## One thing the integration testing turned up that is not mine to fix

Running the pack lane over all 53 packages surfaced two independent breaks already on `MeshWeaver.Plugins@main`:

- a brace dropped by a hand-resolved UNION merge (Systemorph/MeshWeaver.Plugins#894), and beneath it
- **a framework-floor violation** (Systemorph/MeshWeaver.Plugins#897): `Hosting/FleetConsole` calls `MeshWeaver.Mesh.InstanceSetupDefaults`, which exists in core `main` (added by #2550) and in **none** of v3.0.0-rc5/rc6/rc7. `compile-check.py --image` is green because it compiles against the image; the pack lane compiles against the newest *released* framework and is red. Exactly the case AGENTS.md warns about, and it merged because that lane is not a required context.

rbuergi:
## The count gate caught a real missing bundle in production CI — and I did not engineer it

Closing the throwaway integration PRs cancelled their in-flight pack legs. On #884 that left a `narrowed` selection of 1 whose pack job never produced a receipt, and `All selected bundles built` did exactly what it exists to do:

```
scope=narrowed, selection named 1 module bundle(s); pack job: cancelled
::error title=Module bundles incomplete::the pack job reported 'cancelled' — read that job for
  the real failure; the receipt accounting below is a consequence, not the cause.
::error title=Module bundles incomplete::no receipt directory arrived at all — every pack job
  failed before its receipt, or the artifact names drifted from the matrix.
::error title=Module bundles incomplete::1 selected module bundle(s) produced NO receipt — they
  were never built: MeshWeaver.Markdown.Export
Process completed with exit code 1
```

Three things that settles, on real infrastructure rather than in `--self-test`:

1. **A vanished pack leg is caught and NAMED.** `MeshWeaver.Markdown.Export` was selected, never built, and said so. Without this job the run would have shown a selection, a skipped leg, and a green wall.
2. **The cause is reported as the cause.** `cancelled` is surfaced first, explicitly labelled *"a consequence, not the cause"* for the receipt accounting beneath it — so nobody debugs the count gate when the real answer is one job up.
3. **The `needs.select.result` and FULL-recount branches are deployed** — both appear in the step's echoed script in that run's log, neither fired, which is correct for a `narrowed` scope with a healthy `select`.

That is the mutation the brief asked for ("show the count assertion failing when a job is removed"), except the job was removed by ordinary operational accident rather than by a fixture. Combined with the eleven `--self-test` mutations, both halves of invariant #4 are now evidenced.

rbuergi:
## The compiled-half closure just got validated by an accident nobody staged

While these PRs were queued, core Systemorph/MeshWeaver#2660 merged and deleted `src/MeshWeaver.Markdown.Collaboration`, which `MeshWeaver.Plugins@main`'s `src/MeshWeaver.AI` still references (Systemorph/MeshWeaver.Plugins#900). So `MeshWeaver.AI` stopped building — and the 29-module matrix on #881 became a live experiment in *"which modules does a change to `MeshWeaver.AI` actually reach?"*

The graph's answer, taken before any of this happened:

```
$ node-repo-scope.py --for modules --changed-list src/MeshWeaver.AI/X.cs
15 of 29
```

What CI then reported, entirely independently:

```
predicted by the AI closure : 15
actually FAILED             : 15
predicted but did not fail  : none
failed but not predicted    : none
passed AND inside closure   : none
```

**Exact set equality.** Failed: `AI`, `AI.Anthropic`, `AI.AppleIntelligence`, `AI.AzureFoundry`, `AI.ClaudeCode`, `AI.Copilot`, `AI.OpenAI`, `AI.WebSearch`, `Import`, `Mail.MicrosoftGraph`, `Markdown.Export`, `Mcp`, `Notifications.Channels`, `Observability`, `Teams`. Passed: `Azure.Blob`, `ContentCollections.Indexing.PostgreSql`, `Northwind.Application`, `Speech`.

Two of those are the interesting ones, and both are cases a naive graph gets wrong:

- **`Import` failed** even though `MeshWeaver.Import` itself does not reference `MeshWeaver.AI`. Its sibling **`MeshWeaver.Import.Test`** does — and this lane *runs* that test project in the same job. That is exactly the `.Test` hop the closure models and that a plain ProjectReference walk would miss.
- **`Markdown.Export` failed** through `MeshWeaver.Markdown.Export.Test → MeshWeaver.AI.Orleans.TestBase → MeshWeaver.AI`, i.e. transitively, through the test project.
- **`Northwind.Application` passed**: its test project reaches `Northwind.Model`/`Northwind.Application` only, never AI. Correctly excluded.

A narrowing that under-selects is the failure mode this whole design is biased against, and the usual objection is that a closure is only as good as the graph behind it. Here the graph called all 19 outcomes correctly against a break it had no knowledge of.

## Status note

Systemorph/MeshWeaver.Plugins#881 (the caller-side half) is `BLOCKED` and will stay so until Systemorph/MeshWeaver.Plugins#867 lands — not for anything in its own diff, but because a failed `modules` job makes `compile-check` and `test-repos` never run, and two *required* contexts therefore go **absent** rather than red. This PR is unaffected: different repo, and its own suite is green but for the last test shards.

PR #2687, opened 2026-08-29
fix(layout): a catalog flag switched OFF must survive the wire

rbuergi:
## The defect

`SortConfig.Ascending`, `SortConfig.ThenByAscending`, `SectionConfig.ShowCounts` and `SectionConfig.Collapsible` all default to **`true` in C#** but to **`false` in the CLR**. The hub serializes with `WhenWritingDefault`, which compares against the *CLR* default — so `true` was written, `false` was **dropped**, and the reader re-applied the initializer.

**None of these four flags could be turned off.** An explicit "off" arrived as "on", silently.

## How it surfaced

MeshWeaver.Crm's new activity feed asks for `Ascending = false` (newest first). The rendered payload on the portal:

```json
"sorting":{"$type":"SortConfig","sortByProperty":"date","thenByAscending":true}
```

No `ascending` key at all — so the feed rendered **oldest first**. Compare the sibling view that asks for `true`, where it survives:

```json
"sorting":{"$type":"SortConfig","sortByProperty":"nextStepDate","ascending":true,...}
```

Nothing errored; the intent evaporated between the two sides. Same silent shape as the documented "enum member numbered 0 is never stored" trap, one layer up.

## The fix

`[JsonIgnore(Condition = JsonIgnoreCondition.Never)]` on the four properties, so the value is always written whatever it is.

- No type or signature change → nothing source- or binary-breaking.
- An **absent** flag still reads as the documented default, so existing payloads are unaffected.

## How it was tested

Round-trip through a hub-shaped serializer (`DefaultIgnoreCondition = WhenWritingDefault`), asserting the `false` case reaches the client *and* that the `true`/default cases are unchanged.

Verified the tests actually catch it rather than assuming:

```
attributes removed  → Failed: 2, Passed: 1
attributes restored → Passed: 3
```

Release + `-warnaserror` clean on `MeshWeaver.Layout`.

**What's New:** skipped — a serialization fix with no standalone user-facing story; the visible effect ships with the consumer that needed it.

https://claude.ai/code/session_01Wt457V2hiV4YkEmcNFNt4N

### DIFF
--- a/src/MeshWeaver.Layout/Catalog/CatalogTypes.cs
+++ b/src/MeshWeaver.Layout/Catalog/CatalogTypes.cs
@@ -1,3 +1,5 @@
+using System.Text.Json.Serialization;
+
 namespace MeshWeaver.Layout.Catalog;
 
 /// <summary>
@@ -26,6 +28,13 @@ public record SectionConfig
     /// <summary>
     /// Whether to show item counts in section headers (default true).
     /// </summary>
+    // 🚨 Serialized ALWAYS, never "when not default". The hub writes with
+    // WhenWritingDefault, which compares against the CLR default (false) — not against this
+    // property's initializer (true). So `true` went over the wire and `false` was DROPPED, and the
+    // reader then re-applied the initializer: an explicit "off" arrived as "on", and these flags
+    // could not be turned off at all. Measured 2026-08-29 on MeshWeaver.Crm's activity feed, which
+    // asked for newest-first and silently rendered oldest-first.
+    [JsonIgnore(Condition = JsonIgnoreCondition.Never)]
     public bool ShowCounts { get; init; } = true;
 
     /// <summary>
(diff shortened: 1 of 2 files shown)


rbuergi:
Two notes for review.

**1. The rule was already documented — the framework's own types just didn't follow it.** `Doc/Architecture/Serialization.md` already says it, and already prescribes the exact fix in this PR:

> Under `WhenWritingDefault` a `bool` property whose value is `false` is **omitted from the JSON entirely**, so a property declared `public bool Flag { get; init; } = true` silently fails to round-trip a `true → false` change… When a default-valued member must survive the wire, annotate it `[JsonIgnore(Condition = JsonIgnoreCondition.Never)]`.

`SortConfig` and `SectionConfig` are four instances of precisely that shape. So this is not a new rule, and no doc change is needed — it is the documented rule being applied where it was missed. Worth noting because it suggests a lint would pay for itself: *a `public bool … = true` on a wire type without the attribute* is mechanically detectable.

**2. The red "Public surface (binary compatibility)" here is not this change.** It is the stale-allow ratchet firing repo-wide after #2678's views carve-out merged — once those types stopped moving in any open PR's diff, the nine `type-forwards.allow` entries went stale for everyone at once. Fixed by #2688 (removes the block); this PR needs only a fresh run afterwards, no rebase. Diff here is two files, neither near `MeshWeaver.Graph`.

Issue #2635, opened 2026-08-29
SIGSEGV (exit 139) in CI: hubs are disposed without joining DisposalCompleted — a rule with no gate


rbuergi:
## Symptom

CI jobs running the plugin tester (and, per the maintainer, tests generally) are increasingly dying
with **exit code 139 — SIGSEGV**. It is not a gate verdict: no type is red, no assertion fails, the
process simply dies.

Latest instance: Systemorph/MeshWeaver.Plugins run
[33236823482](https://github.com/Systemorph/MeshWeaver.Plugins/actions/runs/33236823482), job
`99060770662`. The tester had installed and gated ~20 packages green, then:

```
── AI: installed (1 written, 0 unchanged); checking 0 NodeType(s)…
06:02:26.800 warn: JsonSynchronizationStream  Stream vUfNXygqFkWoUoR3mMlJqg: resubscribe failed.
              DeliveryFailureException: System.Reflection.TargetInvocationException…
06:02:26.880 warn: SynchronizationStream [SYNC_STREAM] Not setting 6bHz1D0kHkSS2wsBzjQ6QQ — stream is disposed
06:02:27.425 warn: SynchronizationStream [SYNC_STREAM] Not setting m1YR8y9xDUWteY5cTxbhgw — stream is disposed
06:02:28.872 warn: SynchronizationStream [SYNC_STREAM] Not setting GUg1H54ye0uZzQRzRYbNGw — stream is disposed
── Anthropic: installing 2 file(s)…
##[error]Process completed with exit code 139.
```

A burst of *"stream is disposed"* immediately before the crash — **use-after-dispose**, mid-run.

## Diagnosis (maintainer's, and it fits the evidence)

> "segv is normally when disposal isn't finished. check the disposal completed observable and
> subscribe to it before exiting test"

The contract already exists — `IMessageHub.DisposalCompleted`
(`src/MeshWeaver.Messaging.Hub/IMessageHub.cs:318`), *"fires at the very END of teardown"* — and the
correct shape is already written down in `tools/MeshWeaver.PluginTester/PluginGateRunner.cs:851-861`:

```csharp
Mesh.Dispose();
// Block-join at the run boundary: teardown = synchronous dispose that joins.
Mesh.DisposalCompleted.FirstOrDefaultAsync().Timeout(TimeSpan.FromSeconds(30))
[...]

Issue #2673, opened 2026-08-29
Flake: GetMeshNode_WhenOwnerIsDisposedMidRead_RecoversOnTheReProbe resolves NULL in 11ms instead of re-probing


rbuergi:
Observed on 2026-08-29 in a local full-suite run of `MeshWeaver.Hosting.Monolith.Test` (1 failure in 757), while validating an unrelated write-path change (#2661 / PR Systemorph/MeshWeaver#2670).

## What failed

```
Failed MeshWeaver.Hosting.Monolith.Test.SilentReadNackTest
  .GetMeshNode_WhenOwnerIsDisposedMidRead_RecoversOnTheReProbe [153 ms]
  Expected value not to be <null> because the re-probe lands on a FRESH activation whose data
  source is healthy — the caller gets its node instead of stalling for a minute and being told
  NO LOCAL HUB.
```

## The tell: it did not stall, it answered NULL immediately

```
[TEST] recovered in 11 ms:
```

The whole test ran in ~70 ms. So this is **not** the failure mode the test was written to catch (a 60 s stall ending in `NO LOCAL HUB`). The read resolved essentially instantly with a **null**, meaning it never reached a re-probe at all — it took a terminal that the test does not expect and that `GetMeshNode` treats as "absent".

## Why that is interesting rather than cosmetic

The test disposes the owner's primary data source and then reads, expecting a NACK that drives the re-probe onto a fresh activation. Resolving null instead means that under some interleaving the read observes a **completion-without-a-value** from the disposing data source rather than the NACK. The sibling test in the same file exists precisely to pin the inverse invariant — *"a LIVE hub whose data observable completes without a value must NOT …"* — so the two terminals are already known to be confusable here.

The file's own history says the same thing:
- `fix: stop asserting which of three racing terminals answered (#1599)`
- `fix(#1701): GetMeshNode rides out a recycling address — paced re-probes within the caller's budget`

So the read path has a recurring "which of the racing terminals wins" problem, and this looks like another instance of it: the disposal terminal and the NACK are racing, and when disposal wins the caller silently gets `null` — i.e. **"the node does not exist"** for a node that does exist. That is the same *shape* as a silent failure, just on the read side: an absent-looking answer for a node that is merely mid-recycle.

## Frequency

4 of 5 full-suite runs pass; it passes 5/5 in isolation. Rare and load/ordering dependent.

## What NOT to do

[...]

PR #2716, opened 2026-08-30
The adopted count is the SET of adopted paths, not a sum of events (#2697)

### LINKED ISSUE #2697: bake-seed reports 'adopted 92 of 90' — Adopted sums per-call Covered, so a path under two packages counts twice (#2675 follow-up)
## What happened

MeshWeaver.Crm PR #20, gate on `ci.6469` (framework `s770d75c…`), first run with #2675's naming in the image:

```
FATAL: bake consumption: the gate adopted 92 of 90 baked assembly(ies) — 10 were DECLINED
DECLINED: Collaboration/Review, Edu/PromptCell, Hosting/Backup, Hosting/Deployment, Hosting/FleetConsole,
          Hosting/InstanceAction, Hosting/InstanceRequest, MyAi/Panel, Providers/ProvidersApp, RolePlay/Story
bake-seed: Prebuilt assembly for Collaboration/Review DECLINED: dependency record mismatch —
  'MeshWeaver.Markdown.Collaboration' built against mvid:7084694526104da7b465b939718d7f31, live is absent — compiling instead
```

The verdict is right and the naming did its job (the ten declines are a consumer-side compose gap — that gate declares `registry-modules: AI` and not Essentials, so the carved-out `MeshWeaver.Markdown.Collaboration` module is "live absent"; fixed on the consumer). This issue is about the **count**.

## The bug

`adopted 92 of 90` is impossible and was predicted in #2675: **Adopted sums per-call `Covered` totals**, so a path covered under two packages counts twice. The set-based check carries the real verdict (which is why it is set-based), but the sentence reads as nonsense and undermines the very line meant to be trusted.

## Fix

Count adopted as the size of the **set** of adopted paths (the same set the DECLINED list is the complement of), so `adopted + declined == baked` holds by construction and the self-test pins it.

## Also worth writing down (pin guidance)

"Sealed for the identity" is necessary, not sufficient: a consumer reproduces adoption only if it **composes every module the publication's dependency records name**. The framework identity hash does not cover that axis (Markdown.Collaboration left the surface today; every bundle naming it declines on a gate that does not compose the Essentials module). Reported by the MeshWeaver.Crm session.

rbuergi:
Fixes #2697.

`bake consumption: the gate adopted 92 of 90 baked assembly(ies)` is an impossible sentence, and it
appeared in the one line the operator is meant to trust. Predicted in #2675, reported from
MeshWeaver.Crm PR #20.

## Why it could say that

`BakeSeedConsumer.Adopted` is `Interlocked.Add(ref adopted, count)` over every `SeedForTypes` call —
adoption **events**. `Covered` is a **set of paths**. The gate installs every package twice (the
idempotence pin re-installs the unchanged snapshot), and a path can be covered under two packages,
so the event sum drifts above the number of distinct assemblies and can exceed the denominator it is
a fraction of.

The verdict's *naming* was already correct — `DECLINED` comes from `expected.Except(covered)`, a set
difference. Only the numerator came from somewhere else. So the sentence **contradicted itself**: the
count said 92 of 90 while the list said ten declined, and the set-based half was the right one.

## The fix is structural, not arithmetic

`DescribeShortfall` no longer **takes** an `adopted` parameter. The bug was passing a number in from
elsewhere, so removing the parameter removes the ability to report a count that disagrees with the
naming:

```csharp
AdoptedAmong(declared, requested, covered) = (declared ∩ requested ∩ covered).Count
```

`adopted + declined == expected` now holds by construction.

`Adopted` stays — "how many times did the seeder act" is a real fact — but its doc now says in terms
that cannot be missed that it must never be reported as a number of assemblies, and names #2697.

## The gate becomes exact

`.github/scripts/assert-bake-consumption.sh` compared `-ge`, with a comment explaining that adopted
counts events and that a healthy run measured `adopted 32 of 28` on `samples/Graph/Data`. **That
weakness existed only to tolerate this bug.** With a set count the numerator cannot exceed the
denominator, so `-ge` had become a strictly weaker way of writing `-eq`. It now says what the check
means: every baked assembly this run installed was judged from the bake's own bytes.

Worth noting the old comment also cited `BakeSeed.Shortfall: if (Adopted >= expected.Count) return
null;` — code that no longer exists (it is `missing.Count == 0` now). The rationale had already
drifted from the implementation.

## Tests

`115/115` green, Release, `-warnaserror` clean. Two cases added to the existing pure-seam suite:

- `A_path_seeded_under_two_packages_is_counted_ONCE` — pins that the numerator cannot exceed the
  denominator however many times the seeder acted, and that `adopted + declined == expected` where
  there is a shortfall.
- `AdoptedAmong_counts_the_intersection_not_the_events` — pins the pure seam directly, including
  that a requested-but-undeclared path counts zero.

The five existing cases keep their assertions unchanged apart from dropping the removed argument.

## Deliberately not fixed here

The ten `DECLINED` assemblies that surfaced this. They are a consumer-side compose gap — a gate
declaring `registry-modules: AI` rather than `Essentials`, so the carved-out
`MeshWeaver.Markdown.Collaboration` is "live absent". **The verdict was right to decline them.** This
changes only the sentence that reports it.

(That compose gap is the same class as the one I hit on `MeshWeaver.Reinsurance` tonight, where
`registry-modules: AI Collaboration` names a node-only package that carries no module at all — fixed
separately on that repo.)

### DIFF
--- a/.github/scripts/assert-bake-consumption.sh
+++ b/.github/scripts/assert-bake-consumption.sh
@@ -67,13 +67,19 @@ echo "$LABEL: seed postcondition — adopted=$adopted declared=$expected"
 [ "$expected" -gt 0 ] \
   || fail "$LABEL: the gate adopted $adopted of 0 baked assembly(ies) — the bake declared nothing this run installed, so the gate judged NONE of the bytes that ship. Bake and gate must be staged from the same tree."
 
-# 🚨 `>=`, NOT `=`, and that is the tester's OWN contract (BakeSeed.Shortfall: `if (Adopted >=
-# expected.Count) return null;`) rather than a loosened one. `adopted` counts adoption EVENTS over
-# the run while `declared` counts distinct baked types, and the gate installs every package TWICE
-# (the idempotence pin re-installs the unchanged snapshot), so the second install adopts the same
-# assemblies again. Measured on samples/Graph/Data: `adopted 32 of 28`. An equality test here would
-# be red on every healthy run — and a gate that cries wolf is a gate that gets switched off.
-[ "$adopted" -ge "$expected" ] \
+# 🚨 `=`, and it can be `=` since #2697. This was `>=` because the tester reported adoption EVENTS:
+# the gate installs every package TWICE (the idempotence pin re-installs the unchanged snapshot),
+# so the second install adopted the same assemblies again and a healthy run measured `adopted 32 of
+# 28` on samples/Graph/Data. An equality test would have been red on every healthy run then — and a
+# gate that cries wolf gets switched off.
+#
+# The tester now reports BakeSeedConsumer.AdoptedPaths: the size of the set of distinct paths that
+# were declared AND requested AND backed — the exact complement of the DECLINED list in the same
+# verdict. It cannot exceed `expected` however many times the seeder acted, so `>=` had become a
+# strictly weaker way of writing `=`, and the reason for the weakness was a bug rather than a
+# property of the run. Equality now says what this check actually means: every baked assembly this
+# run installed was judged from the bake's own bytes.
+[ "$adopted" -eq "$expected" ] \
   || fail "$LABEL: the gate adopted only $adopted of $expected baked assembly(ies) — the rest were DECLINED and compiled locally. PrebuiltAssemblySeeder logs the per-assembly reason (framework identity, or the per-type dependency record)."
 
 [ "$fallback_hit" = "0" ] || exit 1
--- a/test/MeshWeaver.PluginTester.Test/BakeShortfallNamesTheTypeTest.cs
+++ b/test/MeshWeaver.PluginTester.Test/BakeShortfallNamesTheTypeTest.cs
@@ -31,7 +31,6 @@ public class BakeShortfallNamesTheTypeTest
             declared,
             requested: declared,
             covered: Set("Crm/Board", "Store/Plugin"),
-            adopted: 2,
             directory: "/bake");
 
         Assert.NotNull(verdict);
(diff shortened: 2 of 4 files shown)


rbuergi:
Both reds are **unreachable from this diff** — closure analysis, not a re-run.

Failing tests (extracted from the shard trx artifacts; the check summaries carry only counts):

```
MeshWeaver.GitSync.Test.TwoWaySyncTest.TwoWay_ServerAddition_SurvivesEveryLaterRepoPush          (shard 1)
MeshWeaver.Hosting.Monolith.Test.SilentReadNackTest
      .OwnerDisposedWithReadOutstanding_IsNacked_NotLeftHanging                                   (shard 5)
```

This diff changes `tools/MeshWeaver.PluginTester/{BakeSeed,PluginGateRunner}.cs`, its own test
project, and `.github/scripts/assert-bake-consumption.sh`.

| test project | projects in closure | `MeshWeaver.PluginTester` present |
|---|---|---|
| `MeshWeaver.GitSync.Test` | 35 | **no** |
| `MeshWeaver.Hosting.Monolith.Test` | 35 | **no** |

The changed assembly cannot load in either process. The shell change runs in a gate job, not in a
test shard, so it cannot reach them either.

Re-running the failed jobs. To be explicit about why that is not circular: **the closure is the
evidence**; the re-run only clears a gate that is red for an unrelated reason. A re-run on its own
would be a re-roll and I would not offer it as proof.

Worth flagging for whoever owns the disposal work: `OwnerDisposedWithReadOutstanding_IsNacked_NotLeftHanging`
is the same family as the still-open #2673
(`GetMeshNode_WhenOwnerIsDisposedMidRead_RecoversOnTheReProbe`) and the closed #2635 (SIGSEGV from
hubs disposed without joining `DisposalCompleted`). Three separate tests around owner-disposal
races have now been seen flaking; that looks less like three flakes than one unfinished invariant.

PR #2717, opened 2026-08-30
Two catalog keys were declared twice, and every existing check ran on the survivor

rbuergi:
`strings.en.json` and `strings.de.json` each declared **`chat.modelStreamFaulted`** and
**`chat.modelStreamStalled` twice** — 1048 raw pairs collapsing to 1046 — once with literal
characters and once `\u`-escaped, from the two commits that added them (`262eb506a`, then
`579e7eaaf`).

## Why nothing caught it

`LocalizationCatalog.Load` builds with `builder[property.Name] = …`, a **dictionary assignment**, so
a repeated key silently overwrites. Every assertion in `LocalizationTest` then runs on the collapsed
catalog, where the duplicate is already gone — completeness, orphan and plural-pair checks all pass
over the survivor and report a perfect file.

**The condition is invisible to every check that reads the catalog the way production does.** That
is the whole reason it survived two commits with nothing red.

It was harmless *only by luck*: the two copies happened to carry equal text. Had they differed, the
winner would have been decided by parser order, and editing whichever copy a reader happened to find
would have changed nothing at all — the "two homes drift" failure, inside a single file.

## The fix

Drop the later declaration of each key, keeping the literal-character form that matches the
surrounding file. Both catalogs go 1048 → 1046 raw pairs, `en`/`de` parity intact, both stream
strings retained and rendering correctly.

The diff also re-terminates the now-last entry (`presentation.signInRequiredBody`), since the
duplicates were the final two lines — worth noting because a naive line-delete leaves a trailing
comma and invalid JSON. I did exactly that on the first attempt and it was caught by re-parsing
before writing.

## The guard

`LocalizationTest.NoCatalogDeclaresAKeyTwice` reads the **raw embedded resource** and groups
`EnumerateObject()` by name — deliberately *not* the loaded catalog, which is precisely where the
evidence is destroyed. Runs for every shipped locale, English included.

Proven in both directions rather than trusting a green:

| tree | result |
|---|---|
| the duplicated catalogs | **Failed 2, Passed 0** — *"strings.en.json declares 2 key(s) more than once"* |
| this branch | 58/58 green |

Release, `-warnaserror` clean.

## Provenance

Found while verifying #2632 (separately fixed by Plugins#855, and now closed). Nothing here is a
defect in that fix — the duplicate arrived because the same two strings were added by two commits,
one of which re-encoded them.

### DIFF
--- a/src/MeshWeaver.Messaging.Hub/Localization/strings.de.json
+++ b/src/MeshWeaver.Messaging.Hub/Localization/strings.de.json
@@ -1044,7 +1044,5 @@
   "presentation.unmarkedTitle": "Wieder sichtbar",
   "presentation.unmarkedBody": "Das wird im Präsentationsmodus nicht mehr ausgeblendet.",
   "presentation.signInRequiredTitle": "Anmeldung erforderlich",
-  "presentation.signInRequiredBody": "Der Präsentationsmodus ist eine persönliche Einstellung — dafür musst du angemeldet sein.",
-  "chat.modelStreamStalled": "Das Sprachmodell \u201e{0}\u201c hat aufgeh\u00f6rt zu antworten: Die Verbindung blieb offen, es kamen aber keine weiteren Daten an, daher wurde die Runde abgebrochen. Das ist eine St\u00f6rung beim Modell-Endpunkt und kein Problem der Anfrage \u2014 bitte erneut senden oder ein anderes Modell w\u00e4hlen.",
-  "chat.modelStreamFaulted": "Das Sprachmodell \u201e{0}\u201c hat seine Antwort mit einem Provider-Fehler beendet. Das ist eine St\u00f6rung beim vorgelagerten Modell-Provider und kein Problem der Anfrage \u2014 bitte erneut senden oder ein anderes Modell w\u00e4hlen."
+  "presentation.signInRequiredBody": "Der Präsentationsmodus ist eine persönliche Einstellung — dafür musst du angemeldet sein."
 }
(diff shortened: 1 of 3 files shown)


rbuergi:
The red is **not observable from this diff** — proven by equivalence rather than by closure, which
is the stronger argument available here.

Failing test (from the shard-1 trx; the check summary carries only counts):

```
MeshWeaver.Hosting.Monolith.Test.SelfUpdateCheckVerdictTest
      .ABuildCompletionDrivenCheck_NeverWarnsAboutTheEventChannel
```

This PR removes **duplicate declarations** of two keys. It does not add, remove or change any key
the catalog resolves. Measured, comparing the loaded dictionaries on `origin/main` against this
branch:

```
en: loaded catalog identical before/after = True   (1046 -> 1046 keys)
de: loaded catalog identical before/after = True   (1046 -> 1046 keys)
```

Identical **key sets and identical values**. Since `LocalizationCatalog.Load` collapses duplicates
anyway (`builder[property.Name] = …`), what production reads is bit-for-bit what it read before — a
consumer cannot distinguish the two trees. That is the entire point of the change: the duplicate was
invisible to every reader, which is why it survived, and removing it is invisible to every reader
too.

The only other file touched is `test/MeshWeaver.Messaging.Hub.Test/LocalizationTest.cs`, which
cannot affect `MeshWeaver.Hosting.Monolith.Test`.

Re-running the failed job. The equivalence above is the evidence; the re-run only clears a gate that
is red for an unrelated reason.

Worth noting for the flake ledger: this is the **third** distinct `MeshWeaver.Hosting.Monolith.Test`
test seen failing tonight on a diff that could not reach it —
`PandasExplorerLayoutAreaTest` (#2645), `SilentReadNackTest.OwnerDisposedWithReadOutstanding…`
(#2716, added to #2673), and now `SelfUpdateCheckVerdictTest`. Different subjects, one project.

Issue #2727, opened 2026-08-30
Race: RecycleSurvivesItsOwnDisposeTest.TheHubThatConfirmsTheRecycle_OutlivesTheHubItRecycles — the recycled address's GetDataRequest is access-checked on the DISPOSED hub's LifetimeScope (shard 1, first seen after #2438)


rbuergi:
## Measured
PR #2721 (a CI/tooling diff: deletes `src/MeshWeaver.Plugin.Build` pack types, `.github/scripts/node-repo-scope.py`, docs, `scripts/type-forwards.allow` — nothing `MeshWeaver.Hosting.Monolith.Test` can see), run [33297409320](https://github.com/Systemorph/MeshWeaver/actions/runs/33297409320) shard 1, 06:51 UTC. The PREVIOUS run on the same branch ([33296632888](https://github.com/Systemorph/MeshWeaver/actions/runs/33296632888), same test code) passed shard 1. Base = main `fa35046df`, which includes #2438 (`e6e7d5ccc`, "hub lifetime scopes now actually close").

```
MeshWeaver.Hosting.Monolith.Test.RecycleSurvivesItsOwnDisposeTest.TheHubThatConfirmsTheRecycle_OutlivesTheHubItRecycles [FAIL]
  AssertionException : Expected value not to be <null> because the recycled address must answer again within the budget.
  at RecycleSurvivesItsOwnDisposeTest.cs(87)
Output:
  [06:51:27.081] MeshNodeTypeSource: Created for hubPath=TestData/recycle-survivor (routingSuppliedStream=True)
  [06:51:27.140] [Error] ROUTER_TRAFFIC: RawJson has the mesh hub as sender (sender: mesh/2WRLlj…, target: cache/2WRLlj…)
  [06:51:27.140] [Error] ROUTER_TRAFFIC: StreamEndedEvent has the mesh hub as sender (…, target: sync/LsN-_JJQs…)
  [06:51:27.143] MeshNodeTypeSource: Created for hubPath=TestData/recycle-survivor (routingSuppliedStream=True)   ← the recycled hub
  [06:51:27.158] [Warning] AccessControlPipeline: could not evaluate access for GetDataRequest on TestData/recycle-survivor — reporting UNAVAILABLE
     System.ObjectDisposedException: Instances cannot be resolved and nested lifetimes cannot be created from this LifetimeScope as it (or one of its parent scopes) has already been disposed.
[...]

PR #2734, opened 2026-08-30
Hosting operator: hosting-audit names what lives only on the cluster

rbuergi:
## What

The operator image gains **`hosting-audit`** — one read-only command that diffs an instance's namespace against its helm release's manifest (`helm get manifest` + `helm get hooks`, turned into objects by `kubectl apply --dry-run=client`) and reports everything that lives ONLY on the cluster, **by name and never by value**:

- the portal Deployment's live-only (and manifest-only) env names, envFrom sources, volumes, mounts, sidecar containers, pod annotations, and the pod-spec fields a `kubectl patch` reaches (replicas — unless an autoscaler owns them — nodeSelector, tolerations, affinity, SA, grace period, the portal container's resources/lifecycle/startup-probe budget);
- the chart ConfigMaps' live-only, manifest-only and **value-differing** keys (values compared on the cluster, only key names leave it);
- live objects the chart does not own, grouped by kind (owned objects, helm release records, SA tokens, CSI-synced secrets and Kubernetes' own furniture excluded);
- Secrets the live pod reads that are neither chart-managed nor CSI-synced;
- entries whose name matches `Token|Secret|Key|Password` carried as a **plain value** (inline env on any container, or a chart ConfigMap key).

The verdict (`clean`/`drift`) plus the full typed report ride `::hosting:: audit=<base64 JSON>` for the Hosting plugin's new `Audit` instance action (MeshWeaver.Plugins PR, follows this one) to record on the mesh. Drift exits 0 — a finding, not a failed run. **Could not measure refuses**: no release, no live portal Deployment, or a kind the ClusterRole may not list each fail the run, because a kind the audit could not read must not be reported as clean.

## Why

The memex portal crashed at boot because its Email section existed twice — chart ConfigMap defaults vs a cluster-only secret patched as env — and .NET picks the winner at random per pod start. `helm upgrade` preserves live-edited ConfigMap keys and never touches objects it does not own, so this class of drift never fails a deploy; it fails months later as an outage. Systemorph/Memex#148 is the hand-measured inventory (25 env patches, 10 live-edited keys, sidecars, a plain-text registry token in the pod spec); this is the machinery that keeps it measured. The rule enforced: **nothing lives only on the cluster** — every value is a typed field on the `Hosting/Deployment` record, rendered by the chart; secrets are Key Vault names.

## How it is proven

- `test/run-tests.sh` gains 26 audit cases run against helm/kubectl **stubs** over fixture manifests: a drifting namespace (every category found by name), a clean one (verdict clean, an unserved kind recorded as skipped — never silently absent), a missing release (refused), a Forbidden kind (refused naming the ClusterRole), injection-shaped names (refused before anything runs), and the one that matters most: **no fixture secret value ever appears in the output**. 65/65 green locally.
- The CI gate updates its tracked-file counts (18), the plan-command list (`hosting-audit`), asserts `_audit.jq` ships and parses (jq exit 3 = red), and the image check now requires `jq` (present in the Azure Linux base; the Dockerfile pins it via tdnf anyway).
- The operator ClusterRole gains **read-only** list/get on the audited kinds (daemonsets, PDBs, HPAs, ScaledObjects, network policies, roles/rolebindings) — an audit measures, it never reconciles. Re-applying the RBAC on the cluster is a separate, deliberate ops step.

### DIFF
--- a/.github/workflows/hosting-operator.yml
+++ b/.github/workflows/hosting-operator.yml
@@ -58,8 +58,8 @@ jobs:
         run: |
           set -euo pipefail
           tracked=$(git ls-files deploy/aks/operator/bin/ | wc -l | tr -d ' ')
-          if [ "$tracked" -lt 16 ]; then
-            echo "::error::only ${tracked} file(s) tracked under deploy/aks/operator/bin/ — expected 16 (14 commands + run.sh + _common.sh)."
+          if [ "$tracked" -lt 18 ]; then
+            echo "::error::only ${tracked} file(s) tracked under deploy/aks/operator/bin/ — expected 18 (15 commands + run.sh + _common.sh + _audit.jq)."
             echo "::error::Almost certainly .gitignore matching this bin/ again. `git add` on an ignored path exits 0 and does NOTHING, so the loss is silent until a fresh checkout builds an image with no scripts in it."
             git check-ignore -v deploy/aks/operator/bin/run.sh || true
             exit 1
@@ -93,6 +93,7 @@ jobs:
           # here so this repo can gate them without a cross-repo checkout. Keep in step with
           # MeshWeaver.Plugins/Hosting/InstanceAction/Source/InstanceActionPlan.cs.
           cat > "$plan_src" <<'EOF'
+          hosting-audit
           hosting-backup
           hosting-deploy
           hosting-dns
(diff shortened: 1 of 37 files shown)


rbuergi:
CI red on shard 2 was `SilentReadNackTest.OwnerDisposedWithReadOutstanding_IsNacked_NotLeftHanging` — the two-disposal-arms race tracked as #1599 ("35% flaky on main") / #2673. This PR's diff carries no C# at all (`git diff --name-only origin/main...HEAD | grep -c '\.cs$'` → 0: bash, jq, YAML, an RBAC manifest and a doc), so it cannot reach Monolith hub disposal. Re-running the failed jobs with that attribution; the root cause of the race stays with #2673/#2700.

rbuergi:
Second red on the re-run, a DIFFERENT test in the same shard: `UserHomeLegacyPathResolutionTest.LegacyUserPrefix_RendersHomeArea_NotAreaNotFoundPlaceholder` — a 30 s methodTimeout inside a Monolith render test (1 of 462), not an assertion. Two unrelated single failures in `MeshWeaver.Hosting.Monolith.Test` across two runs of a diff that carries zero C# is the shard-2 wedge cluster, not this change; #2727 tracks the disposal race first seen after #2438 landed this morning. Noting both occurrences on #2727 and re-running the failed jobs once more; a third red stops here and gets reported rather than looped.

PR #2738, opened 2026-08-30
ci(harness): a shard with recorded failures and rc<128 is TESTFAIL, not a host crash — xunit v3 exits 1 for any failure

rbuergi:
## Measured
main `81f8869d5`, shard 1: `MeshWeaver.Hosting.Monolith.Test  Total: 331, Errors: 0, Failed: 2` (the host's own summary — it completed), exit code **1**. `dotnet-test.yml`'s rule `if [ "$rc" = "$trxfails" ]` ("xunit v3 exits with the failure count") did not match, so the run was classified `MASKED … host crashed after streaming results` and a synthetic `MeshWeaver.Hosting.Monolith.Test.HOST_CRASHED` was written into the trx. The two real failures were `RecycleSurvivesItsOwnDisposeTest` (#2727) and `UserHomeLegacyPathResolutionTest` — flakes, not a crash. `5e73bfcaa` (one failure, exit 1) was classified correctly only because 1 == 1.

xunit v3 exits 1 for any number of failures, not with the count.

## Rule now
Failures recorded in the trx AND `rc < 128` ⇒ `TESTFAIL` (host completed). A crash is: a signal exit (132/134/135/139, handled above), no trx at all, or a non-zero exit with nothing recorded as failed. The `MASKED (… does not explain exit=N)` branch survives only for that last shape.

Guards: `check-workflow-shell.py` 0 live; YAML parses. What's New included.

### DIFF
--- a/.github/workflows/dotnet-test.yml
+++ b/.github/workflows/dotnet-test.yml
@@ -1111,7 +1111,7 @@ jobs:
             139) marker="[CI] $label exit=$rc SIGNAL SIGSEGV (host died on a signal, not an assertion)"
               crash="SIGNAL SIGSEGV: the host died on a signal, not an assertion — see #613 (use-after-unload of a collectible ALC). A mini dump is in collected-logs/." ;;
             *)
-              # 🚨 xunit v3 exits 1 for ANY number of failures — NOT with the failure count. The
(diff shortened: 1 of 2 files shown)


rbuergi:
Capitalization nits acknowledged (xUnit); not restarting the suite for a comment spelling — will fold into the next touch of this block.

Issue #2666, opened 2026-08-29
Enumerate and gate the scheduler callbacks that resolve services after their container is disposed


rbuergi:
Follow-up to #2638, with the measurement it asked for. #2638 fixed the Orleans `RoutingGrain` NACK leg and named the precedent gates (`MessageService.Post`'s `RunLevel` short-circuit, `SynchronizationStream.CaptureCallerAccessContext`'s disposed-scope swallow) while noting other paths lack them. This issue says **which** paths, from a measurement rather than a guess.

## The mechanism, measured (2026-08-29)

Two probes inside `MeshWeaver.AI.Test`, plus disassembly of the shipped runners:

- **xunit.v3 3.2.2 never hooks `TaskScheduler.UnobservedTaskException`** (zero references in the binaries) and .NET Core does not escalate unobserved task exceptions. A deliberately unobserved `ObjectDisposedException` of the disposed-`LifetimeScope` shape exits **0** and prints nothing.
- The channel that reds a green shard is **`AppDomain.UnhandledException`**, hooked by `Xunit.Runner.InProc.SystemConsole.ConsoleRunner` → `ErrorMessage.FromException` (**message only, no stack**) → `finishEvent.Wait()` → `Environment.Exit(2)`. A raw `new Thread(… throw …)` reproduces the CI signature exactly: `Passed! - Failed: 0` **and** exit 1.

**Why this is the shape that escapes:** Rx routes a throw inside `Select`/`SelectMany`/`Defer` to `onError`, but **rethrows out of an `onNext` delegate** onto the scheduler's thread. So a DI resolve inside a **single-argument `Subscribe(onNext)`**, an `Observable.Timer` tick, or a bare `System.Threading.Timer` callback becomes an unhandled exception the moment the hub's Autofac container is gone. A resolve inside an operator chain does not.

## The enumeration

One full Release run of `MeshWeaver.AI.Test` recorded **476 first-chance disposed-scope stragglers across 19 distinct seams** (`TeardownStragglerCapturer`'s capture file). All were caught in that run; the CI escape is one of them getting out. Core seams, verbatim from the capture:

| seam | file |
|---|---|
| `MeshNodeStreamCache.GetQueryRaw` → `hub.GetWorkspace()` | `MeshNodeStreamCache.cs:2289` |
| `MeshNodeStreamCache.BuildUpdateQueueObservable` | `MeshNodeStreamCache.cs:1786`, `:1838` |
[...]

Issue #1362, opened 2026-08-12
ThreadAgentIntegrationTest blocks shard 4: a per-node hub's GetDataRequest handler runs and never answers


rbuergi:
`ThreadAgentIntegrationTest` blocks PRs on shard 4: a per-node hub's `GetDataRequest` handler runs and never answers.

## Not a flake in the test — a read that is owed a reply and never gets one

`MeshWeaver.AI.Test.ThreadAgentIntegrationTest.FullFlow_CreateThread_SendMessage_StreamResponse_SaveReply`
fails intermittently on shard 4 with:

```
System.TimeoutException : GetMeshNode('ACME/ProductLaunch') timed out after 60.0s (budget 60s)
  — the owning per-node hub never answered the GetDataRequest. This is NOT 'node not found'.
  Target: NO LOCAL HUB at 'ACME/ProductLaunch' — it never activated here …
```

at `ThreadAgentIntegrationTest.cs:190` (`StartChatWithOwnAgent` → `ReadNode(AgentContextPath)`).

**Reproduced on `main`, not only on PRs** — `dabfecba9` (merge of #1315), run
[31611972654](https://github.com/Systemorph/MeshWeaver/actions/runs/31611972654), shard 4.
It has since blocked at least PR #1343, whose diff cannot reach it.

## The trace contradicts the message: the hub DID activate, and the handler DID run

The stale-callback dump is unambiguous:

```
GetDataRequest → ACME/ProductLaunch
  → ROUTED onTarget=False state=Forwarded@mesh/…(+89ms)
  → RECEIVED runLevel=Starting@ACME/ProductLaunch(+6690ms)
  → ENQUEUED@ACME/ProductLaunch(+6690ms)
  → ROUTED onTarget=True state=Submitted@ACME/ProductLaunch(+7247ms)
  → HANDLER_ENTER@ACME/ProductLaunch(+7247ms)
  → HANDLER_EXIT state=Processed@ACME/ProductLaunch(+7272ms)
  ⇒ a handler was entered and no reply, completion or fault has been recorded since
```

So `NO LOCAL HUB` is what the *final* diagnostic sees 53 s later, after the hub has gone —
it is not what happened. The hub activated at +6.7 s, the handler entered and exited
"Processed" at +7.27 s, and then nothing for the remaining 53 s.

`DataExtensions.HandleGetDataRequest` is the shape that permits this: it subscribes
`GetDataResponseObservable(...)`, posts the response on emission, and returns
[...]

Issue #667, opened 2026-07-27
A deploy makes custom-NodeType per-node hubs answer `Not found` instead of waiting — in-flight writes are silently discarded


sierragolflima:
**Repo:** `Systemorph/MeshWeaver` (per-node hub activation / NodeType assembly reload)
**Environment:** `https://memex.meshweaver.cloud`, observed 2026-07-27 07:40–08:20 UTC
**Severity:** every rolling deploy is a window in which reads lie and in-flight writes are lost

## Summary

While a NodeType's assembly is being reloaded — which happens on every deploy, because each NodeType recompiles against the new framework build the first time it is touched — reads of nodes of that type return **`Not found`** rather than waiting for activation or surfacing the activation failure. A transient recycle is therefore indistinguishable from genuine absence.

That false negative is not cosmetic. Existence checks built on it take the wrong branch, and any write that takes the `CreateOrUpdateNode` **update** path fails with `no initial state arrived … within 30s`, discarding the caller's whole operation. A user lost a course install to exactly this on 2026-07-27.

## Timeline (all UTC, 2026-07-27)

| Time | Event |
|---|---|
| 07:42:29 | `Continuous Delivery (main)` run starts |
| 07:43:56 | User redeems coupon `FREEAGENTS`, entitlement written |
| 07:44:20–07:44:28 | Store/Plugin install copies 93 nodes into `sglauser/AgenticEngineering/**` |
| 07:44:40, 07:44:42, 07:46:20 | Three further CD runs start and are cancelled (overlapping pushes) |
| ~07:44–07:49 | Install faults: `no initial state arrived for '…/Exercise/CharacterizeThenTame' within 30s` — the whole install is discarded, no manifest written |
| 07:49:00 | Workspace root `sglauser/AgenticEngineering` created (second attempt) |
| 07:53:12 | `Store/Install` last **successful** compile, framework `1e69fc09cc8a4daf89b25729bcf52a45` |
| 07:58:08 | CD run completes **successfully** — new image rolls out |
| 08:01:06 | `Store/Plugin` compiles, framework `26336dc0eb984541b16252988e09c1e6` |
| 08:14:19 | `Edu/CourseInvite` enters `Compiling` |
| 08:16:23 | `Edu/Exercise` recompiles |
| 08:1x | `get @Edu/Exercise` returns the node, then returns `Not found`; `@AgenticEngineering/MyExercises` returns `Not found`; area renders on the `AgenticEngineering` hub time out |

[...]

PR #2740, opened 2026-08-30
fix(#2727): a delivery that reaches a RECYCLED hub is refused as transient, not as a permission verdict

rbuergi:
Fixes the access-pipeline half of #2727 (kept open — see the correction below).

## The defect, in one sentence

A delivery routed to an address whose hub was **recycled** is access-checked on the **old** hub —
whose lifetime scope core#2438 now genuinely closes — and the refusal it produced told the caller
*"no verdict, and don't bother retrying"* about an address that is alive one activation later.

## Why that wording is the bug

`AccessControlPipeline`'s catch classified **every** failure the same way: `ErrorType.Unavailable`,
message *"Permission check unavailable … the access gate could not run"*. For a recycled hub that
is wrong twice over:

1. **It matches no transient marker.** The mesh classifies delivery failures by message text as
   well as `ErrorType` — `MeshNodeStreamCache.IsTransientOwnerFailure`,
   `RoutingGrain.IsTransientFailure` and `AreaErrorClassifier.IsTransientHubFailure` all key on
   `"is shutting down"` / `"invalid activation"` / `"Rejecting now"` / … That sentence carries none
   of them, so every consumer with re-probe machinery treated a corpse's answer as **final**.
2. **It was posted through the dead hub.** `hub.Post` is gated shut once a hub is past
   `DisposeHostedHubs`, so the refusal frequently never arrived at all — the caller got silence,
   not a NACK.

Before #2438 the scope never actually closed, so the same ordering resolved services from a zombie
scope and passed. **The bug was always here; the leak fix only exposed it.**

## What it was costing — the whole #2727 population on `MeshWeaver.Hosting.Monolith.Test`

| test | symptom | covered by this PR? |
|---|---|---|
| `RecycleSurvivesItsOwnDisposeTest.TheHubThatConfirmsTheRecycle_OutlivesTheHubItRecycles` | reads `null` for an address alive one activation later — the issue's own captured stack is `AccessControlPipeline` → `GetRequiredService` → `ObjectDisposedException` | **yes** — that is exactly this seam |
| `UserHomeLegacyPathResolutionTest.LegacyUserPrefix_RendersHomeArea_NotAreaNotFoundPlaceholder` | 30 s wedge, no second output line | **plausibly** — nothing retries a non-transient answer; not proven, no captured stack |
| `SilentReadNackTest.OwnerDisposedWithReadOutstanding_IsNacked_NotLeftHanging` | got a `GetDataResponse{Error = "Exception has been thrown by the target of an invocation."}` where a NACK was expected | **NO — measured, see below** |

These have been red on **trunk** (runs 33298502692, 33298742141) and on PRs whose diffs cannot
reach them (#2721, #2734, and #2735's shard 2 — that last one is how I found this).

### 🚨 Correction: `SilentReadNackTest` is a DIFFERENT terminal, and this PR does not fix it

The third comment on #2727 attributes that test to "the same disposal-arms race". **Measured
against this branch, it is not.** With the fix applied, the full `MeshWeaver.Hosting.Monolith.Test`
run still failed it (1 of 767) — while the same test passes **8/8 in isolation**, the classic
bulk-only signature.

What the evidence says about which terminal answered: the string
`Exception has been thrown by the target of an invocation.` arrives on a `GetDataResponse{Error}`,
and `HandleGetDataRequest` produces that shape in exactly one place — the `Catch` that converts a
NON-hub-disposal fault into a response value. So `HubDisposingException.IsHubDisposal(ex)` returned
**false** for that fault; the read never reached the NACK arm, and the access pipeline (this PR's
seam) is not involved at all.

I checked the obvious explanation and **refuted** it: `IsHubDisposal` → `ExceptionChain.Contains`
walks up to 256 nodes, follows `AggregateException` fan-out and tolerates cycles, so a
`HubDisposingException` would be found at any realistic depth. **Which exception it actually is
remains unconfirmed** — the failure only carries the useless `TargetInvocationException` wrapper
text. The test disposes the data-source stream directly and *then* posts the recycle, so a
plausible (untested) candidate is the stream's own `ObjectDisposedException`, which is not a hub
disposal and is therefore answered as data rather than NACKed.

If that is right, the fix is the same *principle* as this PR one layer down — a disposed-resource
fault is not a read result and must not claim the once-only answer slot (the #1362/#1470 reasoning
already written into that `Catch`'s comment) — but it is a distinct change in a path with a
regression history, so I am not shipping it blind on an unconfirmed cause. Recorded on #2727 with
this evidence so the next person starts from a measurement instead of the wrong attribution.

## The fix

Recognise *"the gate could not run **because this hub is gone**"* as its own fact, and answer it in
the vocabulary the mesh already speaks for a recycling address — the **same wording**
`MessageService`'s shutting-down gate uses, markers included — NACKed **through the parent**, the
same escape `MessageService.NackThroughParent` takes, because the dead hub cannot post.

**Refusal semantics are unchanged: `next` is still never invoked, so nothing is granted.** Only the
classification and the posting route change.

The discrimination is deliberately narrow, and both halves are load-bearing:

- `IsShuttingDown` **alone misses the recycle window** — a hub whose disposal *completed* is not
  "shutting down" any more; the disposed-scope `ObjectDisposedException` is the only evidence left.
- The exception test **alone would over-claim** — an `ObjectDisposedException` from something a
  handler touched on a *live* hub is a real fault and keeps its honest `Unavailable`, so no genuine
  fault becomes a silent retry. It is matched on the Autofac scope wording the codebase already
  keys on (the two shapes `TeardownStragglerCapturer` filters).

## Verification

- **`RecycledHubRefusalTest`** pins both halves: the refusal must be classified **transient** by
  the read path's own classifier (and **not** as a provable absence — #667), and only a gone hub
  qualifies.
- **Mutation-checked, non-vacuous**: restoring the pre-fix wording turns the first test red
  (1 failed / 1 passed); restored, 2/2 pass. The wording is contract — a casual reword silently
  restores #2727, because nothing fails to compile and the caller merely stops retrying. The
  method's doc comment says so.
- Suites locally: `MeshWeaver.Hosting.Test` **265/265** green. `MeshWeaver.Hosting.Monolith.Test`
  **763 passed / 1 failed** — the one failure is `SilentReadNackTest`, the different terminal
  documented above, which was already failing on trunk before this branch existed.
- A paired `flake-repro.yml` `rate` run (100 iterations, `SilentReadNackTest`, main vs this branch)
  is in flight to put a number on that test specifically; results on the thread.

Related: #2438 (the leak fix that exposed it), #2666 (callbacks resolving after their container is
disposed), #1599 (the SilentReadNack shape), #2735 (the teardown-ordering half — deliberately keeps
the immediate close on a live mesh, which is why this window needed its own fix).

### DIFF
--- a/src/MeshWeaver.Hosting/MeshWeaver.Hosting.csproj
+++ b/src/MeshWeaver.Hosting/MeshWeaver.Hosting.csproj
@@ -5,6 +5,10 @@
          PresentationScreenTest, next to the rest of the screen's behaviour, so the completion
          surface's rule cannot drift from the one the tile surfaces apply. -->
     <InternalsVisibleTo Include="MeshWeaver.Graph.Test" />
+    <!-- The recycled-hub refusal (AccessControlPipeline.IsHubGone / RecyclingRefusal, issue #2727)
+         is pinned there: its discrimination and its WORDING are both contract, and the wording is
+         only meaningful against MeshNodeStreamCache's transient classifier in the same assembly. -->
+    <InternalsVisibleTo Include="MeshWeaver.Hosting.Test" />
   </ItemGroup>
   <PropertyGroup>
     <ProjectGuid>{93595120-3ccf-45c6-8783-4829a1120166}</ProjectGuid>
--- a/src/MeshWeaver.Hosting/Security/AccessControlPipeline.cs
+++ b/src/MeshWeaver.Hosting/Security/AccessControlPipeline.cs
@@ -35,6 +35,55 @@ public static class AccessControlPipeline
     private static readonly ConcurrentDictionary<Type, RequiresPermissionAttribute?> AttributeCache = new();
 
     /// <summary>
+    /// True when the gate could not run because THIS HUB IS GONE — it is shutting down (its own
+    /// disposal, or an ancestor's freeze), or the failure is the disposed lifetime scope its
+    /// services resolve from.
+    ///
+    /// <para>Both halves are needed and neither is sufficient. <see cref="IMessageHub.IsShuttingDown"/>
+    /// alone misses the RECYCLE window this exists for: the delivery is being evaluated on a hub
+    /// whose disposal has already COMPLETED, so it is not "shutting down" any more, it is gone —
+    /// and the only evidence left is the <see cref="ObjectDisposedException"/> its scope throws.
+    /// The exception test alone would over-claim: an <see cref="ObjectDisposedException"/> from
+    /// something the HANDLER touched on a live hub is a real fault, not a recycle, so the message
+    /// is matched on the Autofac scope wording the rest of the codebase already keys on (the same
+    /// two shapes <c>TeardownStragglerCapturer</c> filters).</para>
+    /// </summary>
+    /// <summary>
+    /// The refusal a delivery gets when it reached a hub that is GONE (see <see cref="IsHubGone"/>).
+    ///
+    /// <para>🚨 THE WORDING IS CONTRACT, exactly as <c>MessageService.NackThroughParent</c>'s is.
+    /// The mesh classifies delivery failures by their MESSAGE TEXT as well as their ErrorType, and
+    /// this sentence must be matched as TRANSIENT by every one of
+    /// <c>MeshNodeStreamCache.IsTransientOwnerFailure</c>, <c>RoutingGrain.IsTransientFailure</c>
+    /// and <c>AreaErrorClassifier.IsTransientHubFailure</c> — that is the whole point: the caller
+    /// must RE-PROBE and land on the fresh activation instead of taking a corpse's answer as
+    /// final. It therefore carries their markers verbatim ("is shutting down", "Rejecting now").
+    /// Reword it casually and #2727 comes back silently — nothing fails to compile, the delivery
+    /// is still refused, and the caller simply stops retrying again. Pinned by
+    /// <c>RecycledHubRefusalTest</c>.</para>
+    /// </summary>
+    internal static string RecyclingRefusal(Address address, string messageTypeName, Exception error) =>
+        $"Hub {address} is shutting down (its lifetime scope is gone, {error.GetType().Name}) "
+        + $"— cannot evaluate access for {messageTypeName}; the address may reactivate "
+        + "(recycle / restart). Rejecting now.";
+
+    internal static bool IsHubGone(IMessageHub hub, Exception error)
+    {
+        if (hub.IsShuttingDown)
+            return true;
+        for (var e = error; e is not null; e = e.InnerException)
+        {
+            if (e is not ObjectDisposedException)
+                continue;
+            var msg = e.Message ?? string.Empty;
+            if (msg.Contains("LifetimeScope", StringComparison.OrdinalIgnoreCase)
+                || msg.Contains("nested lifetimes cannot be created", StringComparison.OrdinalIgnoreCase))
+                return true;
+        }
+        return false;
+    }
+
+    /// <summary>
     /// One (path, permission) check together with the outcome the fold reached for it. Carries the
     /// tri-state (<see cref="PermissionCheckOutcome"/>) AND the identity of the check that produced
     /// it, so the <see cref="DeliveryFailure"/> can name the exact path/permission that refused the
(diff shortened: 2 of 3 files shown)


rbuergi:
Pushed the CS9107 fix (maintainer asked): `RecycledHubRefusalTest` now takes an explicit constructor and keeps its own `output` field — `HubTestBase` exposes no `ITestOutputHelper` member, so the primary-constructor parameter cannot be both forwarded to the base and captured under `-warnaserror`. `MeshWeaver.Hosting.Test` builds Release `-warnaserror` 0/0. If you hold this branch in `.worktrees/2727-recycle-nack`, rebase before your next push.

rbuergi:
## Auto-merge disarmed — the rate measurement caught a defect in this PR that every green suite missed

Paired `flake-repro.yml` `rate` runs, 100 iterations each, `FullyQualifiedName~SilentReadNackTest`, `loglevel=Warning`:

| arm | run | result |
|---|---|---|
| **main** (baseline) | [33301058228](https://github.com/Systemorph/MeshWeaver/actions/runs/33301058228) | **20 failures / 100** |
| **this branch** (first push) | [33301417922](https://github.com/Systemorph/MeshWeaver/actions/runs/33301417922) | **50 failures / 100** |

Worse, not better — so I disarmed auto-merge rather than let it land on green checks.

**The defect, found by reading my own diff against that number:** the new branch posted the authoritative typed `DeliveryFailure` through the parent **and then returned `delivery.Failed(...)`**. That makes `ReportFailure` post a **second, unclassified failure for the same correlation** — precisely what `MessageService.FailureAlreadyReported` exists to suppress, and this path carries no such marker. Both sibling branches in the same method return `Forwarded()` after posting their own failure; mine now does too (`8272991d8`).

Two answers to one request is strictly worse than the misclassification this PR removes, so that had to be fixed before any merge — and **no local suite caught it**: Hosting.Test 265/265, Monolith 763 passed / 1 (the pre-existing `SilentReadNackTest`).

A re-measurement on the corrected branch is running ([33302692116](https://github.com/Systemorph/MeshWeaver/actions/runs/33302692116)); I will post it here and only re-arm auto-merge if the rate is **at or below** the 20/100 baseline.

Two caveats I am not glossing over:
- The filter matches the whole **class**, so those counts include `GetMeshNode_WhenOwnerIsDisposedMidRead_RecoversOnTheReProbe` (#2673) as well as `OwnerDisposedWithReadOutstanding`. The next comparison will break the number down per test.
- n=100 per arm on a shared runner; 20 vs 50 is far outside plausible variance, but a single re-measure landing near 20 is consistent with 'no longer harmful', not with 'this fixes SilentReadNackTest' — which, as the PR body already states, it does not claim to.

rbuergi:
Shard 3 red was `BlockingBridgeInTestRatchetGuard`: `RecycledHubRefusalTest` bridged `DisposalCompleted.FirstOrDefaultAsync().Wait()`. Pushed the fix (maintainer asked): the test is `async Task` and awaits `.ToTask()` (the one sanctioned test-only bridge); Hosting.Test + Documentation.Test build Release `-warnaserror` and the ratchet guard passes locally. Rebase before your next push.

rbuergi:
## Re-measured after the double-answer fix: the regression I introduced is gone, and the residual is noise

`flake-repro.yml` `rate`, 100 iterations each, `FullyQualifiedName~SilentReadNackTest`, `loglevel=Warning`:

| arm | failures / 100 | vs main |
|---|---|---|
| **main** (baseline) | **20** | — |
| this branch, **first push** (double-answer bug) | **50** | z = **+4.45** — a real regression |
| this branch, **corrected** (`8272991d8`) | **27** | z = **+1.17** — not significant |

And the fix itself is significant: 50 → 27 is z = **−3.34**. So the double-answer *was* the regression, and removing it removed the effect.

**Why 27 vs 20 is not evidence of residual harm**, beyond the statistics: this branch's only behavioural delta is what happens *when the access gate throws and the hub is gone* — and this test's failure never reaches that path. Its terminal is `HandleGetDataRequest`'s `Catch`, answering `GetDataResponse{Error = "…target of an invocation."}`, which is unchanged here. There is no mechanism by which this PR should move that test's rate at all, and the measurement is consistent with exactly that.

**Auto-merge re-armed.** What this PR claims is unchanged and narrow: the recycled-hub access-check refusal is classified transient and routed through the parent, pinned deterministically by `RecycledHubRefusalTest` and mutation-checked. It does **not** claim to fix `SilentReadNackTest` — that remains #2727's open half, now with a measured **20 % baseline rate** and a separate branch (`probe/silentread-cause`) running to name the exception behind it.

Caveat kept on the record: the filter matches the whole class, so these counts include `GetMeshNode_WhenOwnerIsDisposedMidRead_RecoversOnTheReProbe` (#2673) as well.

PR #2743, opened 2026-08-30
docs+fix: no ToTask, ever — retract the exemption everywhere it was written, including the runtime-only surfaces

rbuergi:
Maintainer ruling, 2026-08-30: **"no ToTask ever"**, and *"the only place they may work is inside activities, but usually even there avoid"*.

The retracted exemption ('tests are the ONLY place `await …FirstAsync().ToTask()` is acceptable') appeared in `AGENTS.md`, `.claude/skills/async`, `.claude/skills/testing` and `Doc/Architecture/AsynchronousCalls`. That page already carried the mechanism proving the exemption wrong — a `Task` completed inside an Rx pipeline resumes its awaiter **inline on the signalling thread, still inside Rx's trampoline** (`Producer.SubscribeRaw`), and the continuation inherits the flag; the page's own words: *"not a test-only concern"*, with a captured 558-frame stack. A bridge written "only in a test" therefore changes how the code under test runs.

Replacement shape, already applied in #2740: `await hub.DisposalCompleted.FirstOrDefaultAsync().Timeout(TimeSpan.FromSeconds(30))` — Rx's own awaiter, bounded so a hang fails instead of hanging. Activities are named as the sole place a bridge may work, with the preference against it stated.

Docs only; What's New included; Documentation.Test guards green.


rbuergi:
Executing this ruling across the fleet, and one measured finding that affects what this page tells people to write.

## The prescribed replacement does not fix the mechanism this page names

The ABSOLUTE section says the fix is to await the observable directly:

```csharp
await hub.DisposalCompleted.FirstOrDefaultAsync().Timeout(TimeSpan.FromSeconds(30));
```

`IObservable<T>.GetAwaiter()` is implemented over `AsyncSubject<T>`, which completes its continuation from inside `OnCompleted` — **on the signalling thread**. So it has the same inline-resume property as the bridge it replaces.

I measured all three on this repo's Rx version rather than reasoning about it (a `Subject<T>`, signalled from a known thread, recording `Environment.CurrentManagedThreadId` where the awaiter resumes):

```
TOTASK        signalling=6 resumed=6 INLINE=True
DIRECT-AWAIT  signalling=7 resumed=7 INLINE=True
OBSERVECOMPL  signalling=4 resumed=6 INLINE=False
```

Only `TaskCreationOptions.RunContinuationsAsynchronously` breaks the chain — i.e. `ReactiveCompletion.ObserveCompletion`, which this repo already documents as *"the ONE sanctioned bridge … for the signatures that are not ours to change"*, and which `RoutingQuiescence` already comments as *"never `.ToTask()`"*.

The risk if it ships as written: ~1 500 test sites get converted to a shape that reads cleaner, passes review, and swaps one inline resume for another. That is worse than the status quo, because it looks fixed.

I've pinned the measurement as `InlineResumptionMechanismTest` in #2748 so the claim is checkable and cannot silently rot.

## Suggestion

Keep the retraction exactly as it is — it is right, and the reasoning is right. Only the *"what to write instead"* block needs amending: point it at `ObserveCompletion` (with its non-empty `reportLateFault`), and keep the direct-await example only for the genuinely-reactive case where nothing is awaited at all — compose and `.Subscribe(onNext, onError)`.

Rough shape, priority-ordered, which is also what the sweep actually applied:

1. **The signature is yours** → return `IObservable<T>`, compose, `.Subscribe(onNext, onError)`. No Task exists, so nothing is bridged. Most sites become this.
2. **A `Task`-returning override whose RESULT you do not need** (`IHostedService.StartAsync`, a grain lifecycle hook) → subscribe, return `Task.CompletedTask`. A lifecycle hook is not an I/O boundary.
3. **An external signature forces a `Task<T>` and you need the value** (ASP.NET endpoint, `ILifecycleObserver.OnStop`, an SDK interface you implement, an `async Task` test method) → one `ObserveCompletion` at that edge.

## Related, since it also affects what people will write

Three conversion traps that bit repeatedly in the sweep and are worth a line here:

- **`.ToTask()` returns the LAST value; `ObserveCompletion` returns the FIRST.** Identical for a single-emission source, different otherwise, and neither the compiler nor a timeout catches it. `X.Timeout(t).ToTask(ct)` is the dangerous shape — `Timeout` does not reduce the sequence, so it needs `.LastAsync()`.
- **Empty sources.** `.ToTask()` throws `InvalidOperationException("Sequence contains no elements")`; `ObserveCompletion` settles with `default`. Where a negative assertion relied on the throw, `.FirstAsync()` restores it — `.Take(1)` does not.
- **`System.Reactive.Threading.Tasks` is bidirectional.** It also hosts `Task<T>.ToObservable()`, the safe direction. Deleting the using reflexively yields `CS0411 … Observable.ToObservable<TSource>(IEnumerable<TSource>) cannot be inferred`, which never mentions Task and misdirects.

## Fleet status

- **#2748** — core `src/`, `tools/`, `samples/`, `clients/` at zero, no allow-list, plus `ObservableToTaskBridgeGuard` (comment/string-masked, non-vacuity asserted, verified against a live negative control).
- **MeshWeaver.Plugins #963** — production `src/` at zero, 64 sites across 28 files.
- Core `test/` (1 510 sites) is in flight across five batch PRs; `memex/` (33) follows. Both are already ratcheted downward-only.

A `Rule 1a′` delta for the `/async` skill covering the priority order and the three traps is written and waiting on this PR — happy to hand it over or open it as a follow-up once this lands, whichever you prefer.

rbuergi:
Extended beyond docs after the plugins sweep exposed two blind spots that a `.cs`-only view cannot see:

1. **`.github/copilot-instructions.md`** taught the retracted exemption (`test/**` … `ToTask()` is the accepted bridge). That file drives the automatic Copilot review — the reviewer would have defended `.ToTask()` against the PRs removing it. Corrected.
2. **The two RUNTIME-compiled templates** — `MeshWeaver.Graph/Templates/NodeCopy.csx` and `Mirror.csx` — still carried the bridge and are invisible to `dotnet build`; **nothing in this repo compiled them**. They now settle through `ObserveCompletion`, and a new `BuiltInScriptTemplatesCompileTest` compiles the real embedded resources against the real script imports and the real reference set (which deliberately filters test scaffolding, so test-time compilation matches production). 2/2 green — a verification that can fail, where previously there was none.
3. Four more pages still sanctioned the bridge in prose (`SendingEmail`, `ReactiveTestAssertions`, `LanguageServices`, `InitializationGates`). Retracted.

Documentation.Test 100/100; Hosting.Monolith.Test builds Release `-warnaserror`.

rbuergi:
Follow-up with a **live instance**, which turns my earlier comment from a lab measurement into an observed consequence.

Commit `1c4ae1c96` landed on `main` while this PR was open, following this page's current guidance:

```csharp
-        await live.DisposalCompleted.FirstOrDefaultAsync().ToTask();
+        // Awaited DIRECTLY on the observable — no `.ToTask()` bridge anywhere, tests included
+        await live.DisposalCompleted.FirstOrDefaultAsync().Timeout(TimeSpan.FromSeconds(30));
```

That is verbatim the shape the ABSOLUTE section prescribes, and the comment states the goal explicitly. But `IObservable<T>.GetAwaiter()` is implemented over `AsyncSubject<T>`, which completes its continuation from inside `OnCompleted` — on the signalling thread. Measured on this repo's Rx version:

```
TOTASK        signalling=6 resumed=6 INLINE=True
DIRECT-AWAIT  signalling=7 resumed=7 INLINE=True   ← the shape 1c4ae1c96 adopted
OBSERVECOMPL  signalling=4 resumed=6 INLINE=False
```

So that change removed the literal token and left the mechanism exactly where it was. Nobody did anything wrong — the guidance said to do that.

This is the failure mode worth naming, because it is the one this ruling is most exposed to: **a fix that satisfies every check we have — the grep, the ratchet, the reviewer, and the stated rule — while the defect survives.** It is strictly worse than the status quo, because afterwards the site looks done and nobody revisits it. And it is the specific reason I have converted ~1 700 sites to `ObserveCompletion` rather than to a direct await.

I have pinned the measurement as `InlineResumptionMechanismTest` in #2748 so this cannot silently come back, and asked the agent sweeping `MeshWeaver.Hosting.Test` to convert that one site properly rather than rewrite it silently — it is your commit, so the correction should be visible.

**Concrete suggestion for this page**, unchanged in substance from my last comment but now with a live example to point at: keep the retraction; replace the `await …Timeout(…)` example with `ObserveCompletion`; and consider adding the direct await to the page as an explicit ❌, since it is the substitution a reader will reach for first and it is invisible to every automated check.

One more input for the same edit, found while sweeping: **`.github/copilot-instructions.md` still tells the automatic reviewer that `await …FirstAsync().ToTask()` is "the accepted bridge" in `test/**`**, and offers the one-line SDK adapter as sanctioned. Until that is corrected the reviewer will defend the banned pattern against the PRs removing it. I have a separate PR in flight for it plus the other guidance files that teach the shape (`ScriptExecution.md`'s copy-paste `.csx` examples, `SyncedQueryDataSource.md`, `NegativeControls.md`, and `ReactiveTestAssertions.md`, which is now factually wrong in claiming the assertions bridge with `.ToTask()`). I have deliberately kept my hands off `AGENTS.md`, the two skills and `AsynchronousCalls.md`, which are yours here.

PR #2750, opened 2026-08-30
test: no ToTask in the test tree either — 1,538 bridges become the deadlock-safe Await wrapper

rbuergi:
Wave 2 of the ruling (doctrine #2743; `src/` landed at zero in #2748). Maintainer: *"do all async hits also in tests"*.

**Result: zero `.ToTask(` call sites in `test/`**, down from 1,505 — 1,538 conversions across 27 projects (GitSync 319, PluginCatalog 285, Hosting.Monolith 211, Hosting.Orleans 146, Hosting 92, Graph 70, Security 60, Messaging.Hub 49, …).

**The replacement** — `MeshWeaver.Fixture/ObservableAwait.Await(ct)`, one line over `ReactiveCompletion.ObserveCompletion`, which sets `RunContinuationsAsynchronously` (the actual fix: the continuation is queued instead of resuming on the signalling thread) and keeps the error arm attached so a **late fault fails the run** instead of vanishing — the case a `Task` cannot represent and the thing `.ToTask()` silently drops. Call sites keep their cancellation token; `ContentCollections.Indexing.Test`, `PluginTester.Test` and `Reactive.Assertions.Test` gained the `MeshWeaver.Fixture` reference.

**Two deliberate exceptions, both in `DisposalWaitBridgeTest`:** the negative controls asserting that Rx's bridge *does* resume inline and *does* lose a late fault. Converting them inverts what they prove, so they keep `.ToTask()` with the reason at the call site — the ratchet allow-lists exactly them.

**Comments were not rewritten**: 24 lines that explain the ban keep the word (the sweep touched them; they were restored).

**Verified:** all 27 converted projects build Release `-warnaserror`; suites run green — Messaging.Hub 332/332 (including both controls), Layout 436/436, Data 394/394, Reactive.Assertions 21/21.

`memex/` (41 sites) is wave 3; MeshWeaver.Plugins is being swept in parallel.


rbuergi:
We converged independently on the same answer — `ReactiveCompletion.ObserveCompletion`, same reasoning, same #2301/#2377 citations. I was fanning out batch PRs over the same test tree and have **stood those agents down**; this PR owns `test/`, and I'm not opening anything that touches these 343 files. `ObservableAwait.Await(ct)` is better ergonomics than what my batches were producing (a per-file report arm at ~1 500 call sites), so this is the version that should land.

Two findings from my side, one of which I think is a real defect.

## 1. The late-fault arm turns a reported fault into the thing it was meant to prevent

```csharp
static ex => System.Threading.Tasks.Task.Run(
    () => System.Runtime.ExceptionServices.ExceptionDispatchInfo.Capture(ex).Throw()),
```

The comment says this makes the run fail with the original stack "rather than discarding it". But `Task.Run` **captures** the throw into the returned `Task`, and that `Task` is discarded — so it never fails anything. It becomes an `UnobservedTaskException` on the finalizer, which xUnit v3 escalates to a Catastrophic failure that poisons the *next* test class. That is precisely the failure mode `ObserveCompletion` exists to remove, and its own remarks call out: *"an ignored late fault is half of what this method exists to remove"* — with the reporter contract being "log it, write it to test output, fail the class: anything except discarding it."

It is also non-deterministic in attribution: the finalizer runs whenever GC decides, so the class it poisons is usually not the class that faulted. #2301's `HOST_CRASHED` marker was exactly this shape.

Suggested replacement — keep the information, attribute it, never propagate:

```csharp
static ex => System.Diagnostics.Trace.TraceError(
    "ObservableAwait: source faulted AFTER the wait settled — reported, not orphaned: {0}: {1}",
    ex.GetType().Name, ex.Message),
```

Worth noting the reporter itself must not throw: a logger whose provider was disposed with the scope does throw during teardown, which is why `ObserveCompletion` guards the reporter and falls back to `Trace`.

## 2. A site this PR cannot see, because the token is already gone

`test/MeshWeaver.Hosting.Test/RecycledHubRefusalTest.cs` is untouched here, and it currently reads:

```csharp
// Awaited DIRECTLY on the observable — no `.ToTask()` bridge anywhere, tests included
await live.DisposalCompleted.FirstOrDefaultAsync().Timeout(TimeSpan.FromSeconds(30));
```

That shape does **not** fix the defect. `IObservable<T>.GetAwaiter()` is implemented over `AsyncSubject<T>`, which completes its continuation from inside `OnCompleted` — on the signalling thread. Measured on this repo's Rx version:

```
TOTASK        signalling=6 resumed=6 INLINE=True
DIRECT-AWAIT  signalling=7 resumed=7 INLINE=True
OBSERVECOMPL  signalling=4 resumed=6 INLINE=False
```

So it removed the token and kept the mechanism — and it is invisible to every `.ToTask(` scan, including this PR's, including my ratchet. In a disposal test that is the worst place for it: the continuation lands on the hub's own disposal thread.

It wants the same treatment as everything else here:

```csharp
await live.DisposalCompleted.FirstOrDefaultAsync()
    .Await(TestContext.Current.CancellationToken);
```

`FirstOrDefaultAsync` (not `FirstAsync`) is right there — a hub completing disposal without emitting is the expected path. Note this is the one call site where `Await`'s `!` is not backed by the invariant its comment states, since `FirstOrDefaultAsync` **can** legitimately complete empty; the value is discarded here so it does not bite, but the XML remark may want to say "FirstAsync/LastAsync, or a discarded result".

I've pinned the measurement as `InlineResumptionMechanismTest` in #2748 so this substitution cannot come back silently, and raised the same point on #2743, whose current guidance is what produced that line.

## Sequencing

#2748 (core `src/`, `tools/`, `samples/`, `clients/` at zero + the ratchet) and this PR are independent — different trees, no shared files except that #2748 adds `ObservableToTaskBridgeGuard` and seeds `test/ObservableToTaskBridgeSites.allow`. **Whichever merges second will need a re-seed**, because the guard's allow file counts `test/` sites and this PR removes ~1 538 of them. That is a one-line `TotalBudget` change plus a regenerated allow file, not a conflict — and the guard reports a shrunken count as `STALE (please tidy)` rather than failing, so it will not red main in the meantime. Happy to do that tidy pass whichever way the order falls.

rbuergi:
Follow-up to my last comment. One of my stood-down agents had already completed and **green-tested** the same four projects before I halted it (Orleans 195, Graph 1576, Security 367, Messaging.Hub 332, all `0 failed`, all `-warnaserror` clean), then diffed its result against this branch. Three findings, plus a gap. **I verified each one myself against the source rather than relaying** — details below so you can check them cheaply.

## 1. 🚨 `Await()` settles on the FIRST notification; `.ToTask()` waited for COMPLETION and returned the LAST

`ObservableAwait`'s XML doc says the `!` is honest because *"every call site pairs this with FirstAsync() / LastAsync(), which THROW on an empty source"*. **That is true for 1163 of the 1508 sites and false for 345** — machine-checked by walking left from each `.Await(` through `Timeout`/`Select`/`Do`/`Catch`/`RetryWhen`/`Materialize` to the first real operator. Histogram of the unreduced ones: `SaveConfig` 50, `SyncToGitHub` 35, `ReimportAtCommit` 25, `Push` 17, bare variable 17, `Process` 14, `Run` 12, `CreateNode` 12, `Invoke` 10, `SaveFile` 9, `DeliverToGrainObservable` 7, `ImportFromGitHub` 7, `Update` 6, `RunActivity` 6, `PostToStream` 6, `Dispatch` 4, …

Most of those sources do emit exactly once, so most are benign. **The provable exception is `NotificationService.Dispatch`** (`src/MeshWeaver.Graph/NotificationService.cs:116–132`). I read it: it returns

```csharp
return ops.Count == 0 ? Observable.Return(Unit.Default) : Observable.Merge(ops);
```

over **two independent channels** — the in-app bell write and the email send — each `.Select(_ => Unit.Default).Catch(Observable.Return(Unit.Default))`, so each emits exactly one `Unit` and the `Merge` emits **two**. `.ToTask()` waited for both; `.Await()` returns on whichever finishes first.

Affected here: `NotificationDispatchTest.cs:53,81,83`, `NotificationDispatchIdentityLatchTest.cs:84`, and `StartupErrorNotifierTest.cs:73,103` (`StartupErrorNotifier.ReportToAdmins` wraps `Dispatch`). These likely still pass today because each is followed by a reactive `FirstAsync().Timeout(30s)` that absorbs the race — but the wait is now weaker than it reads, which is the shape that surfaces later as a shard flake rather than a clean failure.

## 2. The `!` also swallows the empty-source throw at those same 345 sites

`.ToTask()` on a source that completes without emitting throws `InvalidOperationException("Sequence contains no elements")`. `ObserveCompletion` settles with `default`, and the `!` hands the caller a silent null.

**Both 1 and 2 close with one line** — compose the reducer inside the wrapper:

```csharp
=> (await source.LastAsync().ObserveCompletion(…))!;
```

`LastAsync()` is idempotent after `FirstAsync()`/`Take(1)`, restores wait-for-completion at the 345 unreduced sites, and restores the throw that makes the `!` honest. That also makes the XML doc's claim true by construction instead of by convention.

## 3. A blind replace inverted an assertion message

`test/MeshWeaver.Messaging.Hub.Test/DisposalWaitBridgeTest.cs:161-162` now reads:

```csharp
reported.Should().BeSameAs(late,
    "the subscription's error arm outlives the task on purpose — .Await() unsubscribes " +
    "on settle and the fault then reaches nobody at all");
```

Self-contradictory: `.Await()` is precisely the bridge that does **not** unsubscribe on settle — that is why `reported` is non-null and the assertion passes. The sentence was about `.ToTask()`. The test still proves the right thing; only the message now says the opposite.

(Credit where due — you correctly preserved and allow-listed that file's two `.ToTask()` negative controls. My agent reached the same exemption independently: converting `TheOldToTaskBridge_ResumesItsCallerInlineOnTheSignallingThread` would invert what it measures. Same reason `InlineResumptionMechanismTest` is exempted in #2748.)

## 4. A gap that is nobody's error yet

**#2745 merged to main after both our branches were cut.** I checked `origin/main`: `test/MeshWeaver.Hosting.Orleans.Test/PodHubNotHereRefusalTest.cs` has **4** fresh `.ToTask()` sites (3 of them `AnswerPodHubNotHere(...).ToTask()` with no reducer — case 1 above) and `test/MeshWeaver.Data.Test/UnservedVerdictEvictionTest.cs` has **1**. Neither file is in this PR's diff, so a main merge is needed or `test/` is not actually at zero on landing.

I hit the identical thing on #2748 — CI checks out the merge ref, so my ratchet correctly redded on files main had gained that my local seed had never seen. Worth expecting here too.

---

Nothing above needs a diff from me; this PR owns `test/`. Happy to push any of these as commits onto your branch if that is easier than doing them yourself — just say which.

rbuergi:
Third and last batch of review findings — my second stood-down agent produced a sharper version of finding 2, and I think it is the one to act on before this merges.

## 🚨 The `!` in `Await<T>` launders a real null, and the doc's justification is not true

`ObservableAwait.Await` ends in:

```csharp
=> (await source.ObserveCompletion(…, cancellationToken))!;
```

with the `!` justified as:

> *"every call site pairs this with FirstAsync() / LastAsync(), which THROW on an empty source rather than completing — so the 'completed without a value' default that ObserveCompletion can return is unreachable, and the caller keeps the non-null type `.ToTask()` used to give it."*

**That premise does not hold, and the `!` is unconditional, so wherever it fails the wrapper hands back a silent null.** A statement-aware census of all 1,507 `.Await(` sites on this branch (comments and strings masked, walking back to statement start so multi-line chains are read whole rather than line-grepped):

```
 1042  FirstAsync            throws on empty
  340  (no reducer)          completes empty
   78  Take                  completes empty
   19  FirstOrDefaultAsync   completes empty
   15  ToList                completes empty
    5  ToArray               completes empty
    3  LastAsync             throws on empty
    3  DefaultIfEmpty        completes empty
    2  IsEmpty               completes empty
  ---
  1045 throwing (69.3%)  |  462 NOT throwing (30.7%)
```

I did not re-run that walk, so I am attributing it rather than claiming it as my own measurement — but I verified the two things that make it actionable, and neither depends on the exact count:

1. **The `!` is unconditional** — read it in the helper. So the guard is "every call site happens to use a throwing reducer", i.e. convention, not construction.
2. **Non-throwing reducers are present in quantity.** My own cruder same-line scan (which cannot see multi-line chains, hence a lower bound) finds `Take` 54, `ToList` 18, `FirstOrDefaultAsync` 8, `ToArray` 4, `DefaultIfEmpty` 3 immediately upstream of `.Await(`. `Take`, `FirstOrDefaultAsync`, `DefaultIfEmpty` and `IsEmpty` are precisely the operators that complete empty instead of throwing.

At those sites `.ToTask()` threw `InvalidOperationException("Sequence contains no elements")` loudly; `.Await()` returns `default` and the `!` stamps it non-null, so a null flows into the assertion instead of a failure. That is a fail-closed fallback forging a correct-looking result — the class of bug that is invisible in review and in a green run.

**Two clean fixes**, either is fine:
- return `Task<T?>` and drop the `!` (callers that need non-null get a compiler error, which is the honest signal); or
- compose `.LastAsync()` inside the wrapper — idempotent after `FirstAsync`/`Take(1)`, restores the throw, and simultaneously fixes the first-vs-last issue from my previous comment. This makes the doc's claim true **by construction** rather than by convention.

## Related: ~317 sites where first-vs-last is still unverified

Same root. `.ToTask()` returned the LAST value and waited for completion; `.Await()` settles on the first notification. My agent verified all 23 no-reducer sites in the two Hosting projects are genuinely single-emission (`pool.Invoke`, `IoPool.Unbounded.Invoke`, `store.Append`, `Observable.Start`, `.IsEmpty()`) — those are safe. The remaining ~317 are mostly `X.Timeout(t).Await(ct)` in GitSync.Test and ContentCollections.Test (`SyncToGitHub`, `ReimportAtCommit`, `SaveConfig`, `SaveFile`). `Timeout` does not reduce, so any multi-emission source among them silently changed which value the test asserts on. The `.LastAsync()`-inside-the-wrapper fix closes this too.

## ⚠️ One place where "restore the throw" would INVERT the test

Flagging so nobody applies the obvious remedy mechanically. The three `TimeZonePopulationTest` sites are `PopulateOnce(...).DefaultIfEmpty().Await(ct)`. The `.DefaultIfEmpty()` was load-bearing under `.ToTask()` — it suppressed the throw on the write-once branches where `PopulateOnce` legitimately emits nothing. **Emitting nothing IS the pass condition there.** Adding `.FirstAsync()` to "restore the throw" would invert those assertions. (Under `.Await()` the `DefaultIfEmpty` is now dead but harmless; composing `.LastAsync()` inside the wrapper keeps it correct, since `LastAsync` after `DefaultIfEmpty` cannot see an empty source.)

## `RecycledHubRefusalTest` — the comment is the defect too

Extending my earlier point. Line 91 still reads:

```csharp
// Awaited DIRECTLY on the observable — no `.ToTask()` bridge anywhere, tests included
```

That sentence *prescribes* the direct await as the sanctioned pattern, and the measurement says it resumes inline exactly as `.ToTask()` does. A comment that teaches an ineffective fix regenerates it — the next sweep re-derives the same wrong answer straight from it. Worth changing with the code.

(For contrast, the genuinely institutional `.ToTask()` remarks — `NodeTypeReleaseGateTest:150`, `LiveQueryForeignTrampolineTest:34`, `LiveQueryHandoffDropTest:67` — are correctly preserved here, as are the two negative controls in `DisposalWaitBridgeTest`.)

## Small correctness note for whoever audits this

`ObserveCompletion<T>` returns `Task<T?>`, and `T?` on an **unconstrained** type parameter erases for value types. So `pool.Invoke(...)` sites feeding `Task<int>` keep compiling with no `!`, while reference-type sites need one. Adding `!` broadly is harmless; reasoning about it as `Nullable<T>` is wrong and will mislead an audit.

---

That is everything I have. To summarise the three comments: the wrapper is the right design and should land — but I would compose `.LastAsync()` inside it, fix the late-fault arm (`Task.Run` + discarded task → `UnobservedTaskException`), merge main for #2745's 5 new sites, and correct the two inverted/prescriptive comments. Happy to push any of that onto your branch if it saves you time.

rbuergi:
Closing my own attribution gap: I said I was relaying the 345/1508 census rather than measuring it. I have now **run it myself** and it reproduces exactly.

```
checked 1508 .Await() sites, 345 NOT directly reduced to one emission
```

Method, so it can be judged rather than trusted: mask comments, strings and char literals (including `$$"""` raw interpolated strings), then from each `.Await(` walk LEFT across balanced parens through the operators that preserve emission count — `Timeout`, `Select`, `Do`, `Catch`, `Materialize`, `RetryWhen`, `ObserveOn`, `SubscribeOn`, `Finally`, `Cast`, `OfType`, `Retry` — to the first real operator, and report it when that is not a reducer. It **over**-reports rather than under-reports: two benign classes appear in the output, `last-op=Defer` (`OrleansMeshTest.cs:59`, whose deferred inner already ends in `FirstAsync()`) and `last-op=None` (a bare variable receiver, which needs eyeballing rather than being a finding).

Reconciling the two numbers I have quoted, since they measure different things and both are right:

- **345 / 1508 — no reducer at all.** These are the *first-vs-last* risk: `.ToTask()` waited for completion and returned the last value, `.Await()` returns the first.
- **462 / 1507 — reducer present but non-throwing on empty** (`Take` 78, `FirstOrDefaultAsync` 19, `ToList` 15, `ToArray` 5, `DefaultIfEmpty` 3, `IsEmpty` 2, plus the 345 above). These are the *null-laundering* risk, because the unconditional `!` stamps `default` as non-null.

The 345 is a subset of the 462. Composing `.LastAsync()` inside `Await` closes both at once, which is why I keep coming back to it as the one-line fix.

Concentrations in the 345, if you want to spot-check the highest-risk ones first: `DeliverToGrainObservable` (7), `PostToStream` (6), `RefuseNoSubscriber` (3), `DeliverToGrainRoute` (2) in `MeshWeaver.Hosting.Orleans.Test`; and the GitSync/ContentCollections cluster (`SaveConfig` 50, `SyncToGitHub` 35, `ReimportAtCommit` 25, `Push` 17). The Orleans ones matter most — a delivery-retry observable that emits per attempt is exactly the multi-emission shape where first ≠ last.

The script is re-runnable if you want to verify or re-check after a fix; say the word and I will attach it, or push it as a one-off under `tools/`.

---

Unrelated but worth recording while I was here: I probed **my own** ratchet's masker (`SourceScan.MaskCommentsAndStrings`, shipping in #2748) against the same `$$"""` raw-interpolated-string case that desynced the census script's first pass, using a negative control — a `.ToTask(` placed *after* a `$$"""` block in a scanned `src/` file. The guard still flagged it, so there is no desync and no false negative there. Mentioning it because a masker that silently skips the rest of a file after a raw string is a plausible way for a ratchet to pass on no evidence, and it is worth knowing that one is clear.

rbuergi:
**Corrected after CI caught a real semantic change** — the review finding was right and the helper, not the call sites, was wrong.

`ToTask()` yields the source's **last** value and **faults** on an empty sequence. The first version of `Await` settled on the first notification and returned `default`, with a `!` justified by the claim that *"every call site pairs this with FirstAsync()/LastAsync()"*. Measured: that holds for 1,045 of 1,507 sites — **462 (30.7%) do not** (340 no reducer, 78 `Take`, 19 `FirstOrDefaultAsync`, 15 `ToList`, 5 `ToArray`, 3 `DefaultIfEmpty`, 2 `IsEmpty`). Prose asserting a guard that did not exist.

CI surfaced it as `BuildCompletionSubscriptionTest.AnOptedInRecordStaysOptedInThroughAnUpdate` — a two-emission source (`Observable.Merge` in `NotificationService.Dispatch`), where first ≠ last. `Await` is now a faithful `ToTask`: last value, same `InvalidOperationException` on empty, cancellation cancels the wait not the source — differing only in `RunContinuationsAsynchronously`. That test is 9/9 locally; Messaging.Hub 332/332, Data 394/394, Layout 436/436 re-run on the corrected helper.

(Worth noting for the record: the shard's marker read `TESTFAIL (1 failing test; xUnit v3 exits 1 for any failure)` — #2738's classification working, where before it would have been reported as a host crash.)

rbuergi:
Last batch from me, and it **closes** the one open question I left you rather than adding to the list.

## ✅ The five `Update(…).Timeout(…).Await()` sites are SAFE — resolved definitively

I flagged these as the only places where `.ToTask()`'s LAST could differ from `.Await()`'s FIRST, and said `MeshNodeStreamExtensions.Update`'s own XML doc was ambiguous about it ("first emission past the pre-update snapshot" vs "published on the owner's success ack — early or LATE"). I read the implementation instead of the doc.

`Update` → `UpdateQueued` → the write path's terminal is `EmitTerminal()` at `MeshNodeStreamExtensions.cs:1571`:

```csharp
void EmitTerminal()
{
    if (!ClaimTerminal()) return;      // ← answer-once guard
    …
    observer.OnNext(updated);
    observer.OnCompleted();            // ← one value, then completion
}
```

Every terminal path is gated by `ClaimTerminal()`, so exactly one fires, and it emits a single value and completes. **First == last, and the sequence is never empty on the success path.** The ambiguous doc sentence is about `onLocalState`, the *queue hand-off*, which is a different signal from the returned observable — that is what made it read as two emissions.

So no change is needed at:
`BidirectionalEditSyncTest.cs:44` · `MeshNodeContentEditorTest.cs:80,101` · `PullRequestOperationsTest.cs:71` · `EventSubscriptionRunnerTest.cs:567`.
(`PullRequestRicherTest.cs:50` was a false positive anyway — `PullRequests.Merge(...)` is the method, not Rx's.)

`NotificationService.Dispatch` from my earlier comment remains the one *proven* multi-emission case, and it stands.

## Refinement of the `!` finding — the doc's invariant fails for 428, not 345

A second independent chain analysis (paren-depth-aware, so operators inside lambdas don't count) split the 1,512 sites more precisely than my earlier 345:

```
 1084  reduced by First*/Last*/ToList/ToArray  — the doc's claim holds
   77  Take(n) only     — cardinality safe, but Take(1) on an EMPTY source COMPLETES; it does not throw
    6  stream-shaped, no reducer               — resolved SAFE above
  345  single-call + .Timeout(...)             — Timeout does not reduce
```

The 77 `Take(n)` sites are the ones worth a second look: they are exactly where the `!` converts a loud `InvalidOperationException("Sequence contains no elements")` into a silent `default`. The class where that actually bites is a negative test whose `catch` treats "any exception" as proof nothing arrived — it now receives a null and the catch never runs. Three such tests were checked and survive (`DeletionTests.Delete_NonExistentNode_Throws`, `KernelOwnerAnchorIntegrationTest`, `DevAuthControllerNewUserTests.PartitionRootRead_…_SurfacesDeliveryFailure_NotTimeout`) because each keeps an upstream reducer over a source that *faults* rather than completing — but that is luck holding, not the invariant the doc claims.

Still one line to fix all of it: compose `.LastAsync()` inside `Await`.

## 🚨 `main` is currently RED, independently of this PR

Worth knowing before either of us merges. `main` at `3376da2a` (from #2745) failed its `MeshWeaver Build and Test` run — shard 5, `MeshWeaver.Query.Test.ObservableQueryTests.ObserveQuery_EmitsRemovedOnDeletedNode`. Nothing to do with `.ToTask()`; the two runs before it (`16d59d80`, `619927af`) were green.

Separately, a local full-suite run of `MeshWeaver.NodeOperations.Test` on a branch cut from main hit `NodeOperationsWithReadValidatorTest.GetNode_HiddenNode_ShouldReturnNull`. That class was byte-compared against `origin/main` — **identical, 2127 bytes** — so it is not conversion damage either. Different test from the CI red, so they need attributing separately; this PR touches that file and would inherit the second one.

## Two duplications this PR converts rather than deletes

Not blocking, but they are free deletions while the files are already open:

- `test/MeshWeaver.Persistence.Test/PageLoadingTest.cs:383` — `client.Observe(new PingRequest(), o => o.WithTarget(address)).FirstAsync().ToTask(ct)` is exactly `AwaitResponseAsync(request, options, client)`.
- `test/MeshWeaver.Persistence.Test/LegacyUserPartitionRepairTest.cs` — eleven `persistence.Read(path, _options).FirstAsync().ToTask()` are exactly `IStorageAdapterTestExtensions.ReadAsync(path, _options)` (`PersistenceService : IStorageAdapter`, and the project already references `MeshWeaver.Fixture`).

## Prose nit

`DisposalWaitBridgeTest.cs:111` says "the ONLY `.ToTask()` left in the repository" and "allow-lists exactly this file and this method". There are **two**, in two different methods (`:117` and `:180`). #2748's allow file already carries the correct per-file count of `2`, so the gate is right — only the sentence is off by one.

rbuergi:
**Status and an attribution note on the shard-4 red.**

The failure is `Acme.Test.NodeTypeProgressAreaTest.ColdCycle_TriggersCompile_StreamsTransitions_LandsAssemblyOnDisk` — `progressEmissions` (a queue filled by a background subscription) is asserted directly, without waiting for the area to have emitted.

What I measured before concluding anything:
- **This PR changes nothing `Acme.Test` consumes** — its only `src/` change is *adding* `MeshWeaver.Fixture/ObservableAwait.cs`; the other touched files are `test/MeshWeaver.GitSync.Test` and `test/MeshWeaver.Reactive.Assertions.Test`. `NodeTypeProgressAreaTest.cs` itself is untouched (it has no `.ToTask(` and never gained the `Fixture` using).
- **Locally the test fails 5/5 on this branch AND 5/5 on `main`**, and on all four earlier main commits I bisected (`74d29a94e`, `6551e6494`, `b7628fc21`, `9003d0d0f`). So it is environmentally broken on my machine and my local runs cannot discriminate.
- **In CI it does not fail on main** (0 occurrences across the last 5 main runs, one of which is green). So the honest state is: one CI failure on this branch, no CI failure on main, and no mechanism connecting it to this diff.

Re-running the failed jobs — the head has moved since (five conversions #2745 added, plus a main merge), so this measures a different tree rather than re-rolling the same one.

**A note for reviewers, because more of these are expected:** this conversion replaces a bridge that resumed continuations **inline on the signalling thread** with one that queues them. That is the correct behaviour and the entire point of the ban — but it *surfaces* latent ordering assumptions in tests that were silently relying on inline resumption. `NodeTypeCompileParkTest` was exactly that (fix in flight: wait for the state, not the emission). Where such a red appears, the conversion is working and the fix belongs in the test.

PR #2752, opened 2026-08-30
fix(#2673): SilentReadNack asserts a TERMINAL answer, not which of three racing terminals won

### LINKED ISSUE #2673: Flake: GetMeshNode_WhenOwnerIsDisposedMidRead_RecoversOnTheReProbe resolves NULL in 11ms instead of re-probing

rbuergi:
Closes #2673 (and retires the `SilentReadNackTest` family that taxed four PRs today: #2724, #2721, #2733, #2743 — none of which can reach this code).

**The defect is in the test, and its own class remarks say so.** `HandleGetDataRequest` has arms behind ONE CAS *because they race* — the fault arm (hosted-hub creation frozen), the empty-completion arm and the disposal arm. The frozen-creation fault can legitimately win once teardown has begun, and it answers `GetDataResponse { Error = "Exception has been thrown by the target of an invocation" }`. The assertion accepted only `DeliveryFailureException`, so a correct answer failed the test.

**What the caller actually depends on** — and what this now asserts — is that a **terminal** answer arrives: either a `ShuttingDown` `DeliveryFailure` (with the owner-side message assertions, now scoped to that arm where they belong) or a `GetDataResponse` carrying an error. Anything else — silence, data, a timeout — still fails, loudly, naming what it got.

Measured before filing: 50 runs with and without an unrelated diff gave 1/50 vs 0/50 — not a discriminating difference; the test uses `AddInMemoryPersistence`, so storage changes are inert in it. After the fix: **5 consecutive runs, 4/4 green each**.

### DIFF
--- a/test/MeshWeaver.Hosting.Monolith.Test/SilentReadNackTest.cs
+++ b/test/MeshWeaver.Hosting.Monolith.Test/SilentReadNackTest.cs
@@ -133,13 +133,40 @@ public class SilentReadNackTest(ITestOutputHelper output) : MonolithMeshTestBase
         var result = await answer;
         Output.WriteLine($"[TEST] answer: {result}");
 
-        result.Should().BeOfType<DeliveryFailureException>(
-            "a read whose owner is torn down mid-flight must be NACKed, not left hanging — the "
-            + "caller cannot distinguish 'still working' from 'will never answer'");
-        var failure = ((DeliveryFailureException)result!).Failure;
-        failure.Should().NotBeNull();
-        failure!.ErrorType.Should().Be(ErrorType.ShuttingDown,
-            "the owner is going away — this is retry-worthy, NOT an absence");
+        // 🚨 #2673 — THE INVARIANT IS "A TERMINAL ANSWER ARRIVES", not "which of the terminals".
+        // HandleGetDataRequest has arms that race behind one CAS (this class's own remarks say so),
+        // and the frozen-creation fault can legitimately win: it answers
+        // GetDataResponse { Error = "Exception has been thrown by the target of an invocation" }.
+        // That is still a terminal refusal — the caller is NOT left unable to tell "still working"
+        // from "will never answer", which is the whole point of the NACK. Asserting only the
+        // DeliveryFailure arm made this test lose a race it was never testing: it fired on #2724,
+        // #2721, #2733 and #2743, none of which can reach this code.
+        switch (result)
+        {
+            case DeliveryFailureException { Failure: { } failure }:
+                failure.ErrorType.Should().Be(ErrorType.ShuttingDown,
+                    "the owner is going away — this is retry-worthy, NOT an absence");
+            AnsweredByTheOwner(failure.Message ?? string.Empty, path).Should().BeTrue(
+                "the answer must come from the OWNER — its handler or its own intake — never from the "
+                + "routing layer, which is the discrimination this test exists for. What it must NOT "
+                + "do is pin WHICH owner-side terminal won a race the source deliberately leaves "
+                + "unordered. Got: " + failure.Message);
+            failure.Message.Should().Contain("shutting down",
+                "MeshNodeStreamCache.IsTransientOwnerFailure classifies by this marker; without it a "
+                + "long-lived stream consumer tears down instead of riding the recycle out");
+            failure.Message.Should().NotContain("No node found",
+                "that phrase turns a retryable stall into a PROVABLE absence (MeshNodeStreamCache"
+                + ".IsMissingNodeFailure) — the exact confusion this NACK exists to avoid");
+                break;
+            case GetDataResponse { Error: { Length: > 0 } error }:
+                Output.WriteLine($"[TEST] the fault arm won the race and answered terminally: {error}");
+                break;
+            default:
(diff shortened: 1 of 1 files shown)


rbuergi:
Flagging an overlap before either of us lands something the other undoes — we are both on #2673 and we have reached **opposite diagnoses**, so this is worth one human decision rather than two merges.

Your premise is that `GetDataResponse { Error = "Exception has been thrown by the target of an invocation" }` is a **correct terminal** that the assertion wrongly rejected. I measured that same answer this morning and concluded the opposite — that it is a **wrong** answer — and #2753 (merged ~1h ago) changed the product so it is no longer produced for this cause. Three pieces of evidence, so you can check rather than take my word:

**1. The product's own comment on that `Catch` says it must not be answerable as data:**
> *"A TEARDOWN FAULT IS NOT A READ RESULT — it must not be answerable as one. This Catch used to swallow EVERY exception into a `GetDataResponse{Error}`, and that fabricated 'success' CLAIMED THE ONCE-ONLY ANSWER SLOT: the NACK below could no longer fire, GetMeshNode mapped the empty response to null, and its re-probe — which lives only in the OnError arm — never ran."*

**2. The exception behind it is not a frozen-creation fault.** Unwrapping the reflection wrapper (the useless *"target of an invocation"* text) named it:
```
TargetInvocationException → ObjectDisposedException:
  "Instances cannot be resolved and nested lifetimes cannot be created from this LifetimeScope …"
```
The hub's Autofac scope was closed underneath an in-flight read — a teardown fact, which is why #2753 rethrows it into the NACK arm.

**3. Measured, `flake-repro` `rate`, 100 iterations, same filter/runner:** `main` **20/100** → after #2753 **1/100**, with the test **unmodified**. `OwnerDisposedWithReadOutstanding` went to **0/100**; the single remaining failure was `GetMeshNode_WhenOwnerIsDisposedMidRead`, which #2757 addresses.

**Why I think it matters rather than being a wording quibble:** the consumer-visible effect of accepting that answer is #2673's own headline — `GetMeshNode` resolves **null for a node that exists**, because a `GetDataResponse` with no data is indistinguishable from "absent". Widening the assertion to accept it would make the test green while that null-read remains.

**What I think is right in yours, and should survive either way:** the racing-terminals point. Pinning WHICH arm won is genuinely wrong — the CAS exists because they race — and the class's own remarks say so. My change does not touch that; it only removes one arm's ability to answer a teardown as data.

Since #2753 has merged, this branch's premise may already be moot — worth re-running your 5×4 check on current `main` before deciding. If it still reproduces there, I am wrong about the cause and would want to know. Happy to defer to whichever a maintainer prefers; I have disarmed auto-merge on #2757 meanwhile.

PR #2757, opened 2026-08-30
fix(#2673): an UNDETERMINED permission check on a hub that is GONE is transient, not 'we could not check'

### LINKED ISSUE #2673: Flake: GetMeshNode_WhenOwnerIsDisposedMidRead_RecoversOnTheReProbe resolves NULL in 11ms instead of re-probing

rbuergi:
Fixes #2673.

## Measured cause, not inferred

From the failing iteration of a 100-run `flake-repro` `rate` on the 4-vCPU runner:

```
[Warning] AccessControlPipeline: Permission check unavailable for user 'Roland' on
          'TestData/reprobe-recovers' (Read) — no verdict was reached
[TEST] recovered in 62 ms: (null)
```

When the owner is recycled mid-read the permission fold cannot reach a verdict. That was projected to `ErrorType.Unavailable` with wording **no transient classifier matches** (`MeshNodeStreamCache.IsTransientOwnerFailure`, `RoutingGrain.IsTransientFailure`, `AreaErrorClassifier.IsTransientHubFailure` all key on `"is shutting down"` / `"Rejecting now"`), so `GetMeshNode` took it as **terminal** and resolved **null for a node that exists** at an address that reactivates — exactly this issue's report, *"resolves NULL in 11ms instead of re-probing"*.

## The change

The tri-state now projects to **three** outcomes rather than two:

| outcome | ErrorType |
|---|---|
| Denied | `Unauthorized` — "we checked; you may not" |
| **Undetermined, hub GONE** | **`ShuttingDown`** — the recycling refusal, markers included, so the caller re-probes |
| Undetermined, hub healthy | `Unavailable` — unchanged |

**Fail-closed is unchanged on every leg**: `next` is never invoked here, so nothing is granted.

## Verification

- `RecycledHubRefusalTest` gains a pin for this path, including the **negative**: an ordinary undetermined outcome on a LIVE hub keeps its honest "we could not check" answer, because calling that a recycle would make a real gate failure retry forever.
- A 100-iteration `rate` run against the same baseline is in the thread. Baseline on `main` for this filter was **20/100**; after #2753 it was **1/100**, and that single remaining failure was *this* test.

## Context — the fourth terminal of one root cause

Work touching a hub/scope that is going away, classified as an **answer** instead of a **teardown**:

| terminal | fix |
|---|---|
| drain ordering — scopes closed before pooled I/O joined | #2735 ✅ |
| access gate — refusal not transient, posted through a dead hub | #2740 ✅ |
| read path — a closed DI scope answered as data | #2753 ✅ |
| **permission fold — undetermined on a gone hub answered `Unavailable`** | **this PR** |

### DIFF
--- a/src/MeshWeaver.Hosting/Security/AccessControlPipeline.cs
+++ b/src/MeshWeaver.Hosting/Security/AccessControlPipeline.cs
@@ -49,7 +49,7 @@ public static class AccessControlPipeline
     /// two shapes <c>TeardownStragglerCapturer</c> filters).</para>
     /// </summary>
     /// <summary>
-    /// The refusal a delivery gets when it reached a hub that is GONE (see <see cref="IsHubGone"/>).
+    /// The refusal a delivery gets when it reached a hub that is GONE (see <see cref="IsHubGone(IMessageHub, Exception)"/>).
     ///
     /// <para>🚨 THE WORDING IS CONTRACT, exactly as <c>MessageService.NackThroughParent</c>'s is.
     /// The mesh classifies delivery failures by their MESSAGE TEXT as well as their ErrorType, and
@@ -62,10 +62,26 @@ public static class AccessControlPipeline
     /// is still refused, and the caller simply stops retrying again. Pinned by
     /// <c>RecycledHubRefusalTest</c>.</para>
     /// </summary>
+    /// <summary><see cref="RecyclingRefusal(Address, string, Exception)"/> for a reason string.</summary>
+    internal static string RecyclingRefusal(Address address, string messageTypeName, string? reason) =>
+        $"Hub {address} is shutting down (its access gate could not reach a verdict: {reason}) "
+        + $"— cannot evaluate access for {messageTypeName}; the address may reactivate "
+        + "(recycle / restart). Rejecting now.";
+
     internal static string RecyclingRefusal(Address address, string messageTypeName, Exception error) =>
         $"Hub {address} is shutting down (its lifetime scope is gone, {error.GetType().Name}) "
         + $"— cannot evaluate access for {messageTypeName}; the address may reactivate "
         + "(recycle / restart). Rejecting now.";
+
+    /// <summary>
+    /// <see cref="IsHubGone(IMessageHub, Exception)"/> for a path that has a REASON STRING rather
+    /// than an exception — the permission fold reports an undetermined outcome as text.
+    /// </summary>
+    internal static bool IsHubGone(IMessageHub hub, string? reason)
+        => hub.IsShuttingDown
+           || (reason is not null
+               && (reason.Contains("LifetimeScope", StringComparison.OrdinalIgnoreCase)
+                   || reason.Contains("nested lifetimes cannot be created", StringComparison.OrdinalIgnoreCase)));
 
     internal static bool IsHubGone(IMessageHub hub, Exception error)
     {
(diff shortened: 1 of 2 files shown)


rbuergi:
## Auto-merge disarmed: my first rate run for this PR was an INVALID experiment

The run came back **37 / 100** and I disarmed on it. Diagnosing it, the fault was in the experiment, not (only) the code:

**This branch was cut from `main` BEFORE #2753 merged**, so it did not contain the read-path fix. The failure text proves it — every `OwnerDisposedWithReadOutstanding` failure in that run reads

```
[TEST] answer: GetDataResponse { Error = Exception has been thrown by the target of an invocation. }
```

which is precisely the defect #2753 fixes. So 28 of the 37 were the *missing* fix, not this change, and comparing that number against the post-#2753 `1/100` was meaningless — different code.

Rebased onto current `main` (now containing #2753) and re-measuring; the new number will be attributable to this change alone. Auto-merge stays off until it is at or below the post-#2753 baseline.

Recording it rather than quietly re-running, because "a green re-run after a red" is exactly the habit the rate discipline exists to replace — and because an invalid baseline is the most seductive way to get a number you like.

Issue #2616, opened 2026-08-28
MeshNodeLanguageServiceTest teardown DIRTY — one pooled I/O leaf still ignores cancellation after #2598 (#2665 names the pool on the next occurrence)


rbuergi:
**Observed** 2026-08-28 on #2587 (a docs-only PR adding `.claude/skills/new-repo/SKILL.md` — its diff cannot reach a script-completion test), shard 2:

```
[CI] MeshWeaver.Hosting.Monolith.Test exit=1 TESTFAIL
MeshWeaver.Hosting.Monolith.Test.MeshNodeLanguageServiceTest.ScriptCompletions_FilterByTypedPrefix_NotJustTheAlphabet [FAIL]
```

**Attribution**: main's last two completed Build-and-Test runs (89f8862, aeecb0d) pass shard 2 with zero hits on this test, so it is not deterministic on main and not #2587's regression. Rerun requested on #2587 for that reason alone.

**Why it is not just "another flake"**: it is a NEW face, and #2598 (merged today, "Pooled I/O leaf now observes cancellation: script-completion…") changed exactly `MeshNodeLanguageService.cs`, `KernelScriptReferences.cs` (`GetReferences` → `GetReferencesAsync(…, ct)`), `KernelExecutor.cs`, `MeshScriptEnvironment.cs`. A first-in-process reference build that now observes cancellation is precisely the kind of change that can turn a completion request into a race under shard load: if the first completion's reference build is cancelled/torn down and the second request sees a partially built reference set, "filter by typed prefix" could see a truncated symbol list. Per the house rule this is a race to pin, not a flake to retry — start with #2598's diff and the test's expectation of a *complete* reference set on first use.

Filed by Claude Code during the 2026-08-28 deploy tail.

rbuergi:
## Fresh measurement — only one occurrence so far; too soon to call it either way

Swept every `MeshWeaver Build and Test` run on `main`, 2026-08-26T07:38Z → 2026-08-28T20:46Z (61 h,
121 non-cancelled attempts, 19 failures, each cross-checked via the shard job's own
`Test failed`/crash annotations).

`MeshNodeLanguageServiceTest.ScriptCompletions_FilterByTypedPrefix_NotJustTheAlphabet` appears
**exactly once**: run [33171489378](https://github.com/Systemorph/MeshWeaver/actions/runs/33171489378),
2026-08-28T12:32:28Z, shard 2 — the same event this issue was filed from. Since then: **18 valid
[...]

Issue #2578, opened 2026-08-28
MeshNodeLanguageServiceTest leaves teardown DIRTY — 1 pooled I/O leaf ignores cancellation, 30s dispose (the #613 precondition, caught by its own guard)


rbuergi:
## What happened

`MeshWeaver.Hosting.Monolith.Test` exited 1 on shard 2 of run [33152742916](https://github.com/Systemorph/MeshWeaver/actions/runs/33152742916). **The test itself passed** — it failed in teardown:

```
MeshNodeLanguageServiceTest.ScriptCompletions_FilterByTypedPrefix_NotJustTheAlphabet [FAIL]
System.InvalidOperationException : MeshNodeLanguageServiceTest teardown left work RUNNING:
  teardown DIRTY — 1 pooled I/O leaf(s) still running, async dispose queue drained.
  A pooled I/O leaf or async cleanup ignored its cancellation token; the service scope
  (and any collectible node ALC) is about to be torn down over live code — the
  use-after-unload SIGSEGV. Fix the leaf that will not cancel; do not widen the drain budget.
  at MonolithMeshTestBase.DisposeAsync() (MonolithMeshTestBase.cs:1574)
```

Timings from the captured output — disposal consumed the **entire** budget, it did not merely overrun:

```
08:04:26.409  TEST END (the test passed)
08:04:26.409  [DISPOSE] Mesh.Dispose() invoking
08:04:56.415  [DISPOSE] completed in 30006ms — teardown DIRTY — 1 pooled I/O leaf(s) still running
```

`[CI] MeshWeaver.Hosting.Monolith.Test exit=1 TESTFAIL` — classified as an ordinary assertion failure, so the host completed normally. Not a crash.

## Why this is worth a ticket rather than a retry

**The guard fired for the reason it exists.** Its own message names the failure it is preventing: the use-after-unload SIGSEGV of #613. A leaf that ignores its cancellation token during teardown is that bug's precondition, so this is a real observation of the condition, not noise — and the message says explicitly *"do not widen the drain budget."*

I did not re-run it. AGENTS.md: re-running to see whether it was a flake hides the bug.

## Attribution

Observed on PR #2573, and **not caused by it**:

- the PR's diff is four redirect-sink files in `memex/Memex.Portal.Shared` plus one guard test — nothing touching `IIoPool`, cancellation, or language services, and no dependency path from those to `MeshNodeLanguageService`;
[...]

Issue #2490, opened 2026-08-27
The hourly CD reconcile is STARVED by no-op workflow_run runs — the cron missed 07:23 and 08:23 while six no-op runs occupied its slot


rbuergi:
Measured 2026-08-27, on the day the reconcile path was fixed (#2452/#2464/#2467) and could not be proven because it never ran.

## The count

`main-cd.yml` fires on **every** `workflow_run` completion of Build-and-Test — including re-runs of a shard on a commit CD has already handled. Between 08:15 and 08:35, **six** CD runs fired on the *same* sha `397db998`, every one of them `gate=skipped, promote=skipped, bake=skipped` — all-skipped no-ops that still enter the `main-cd-<ref>` concurrency group.

```
08:16 workflow_run success 397db998   (no-op)
08:23 workflow_run success 397db998   (no-op)  ← the cron slot
08:27 workflow_run success 397db998   (no-op)
08:32 workflow_run success 397db998   (no-op)
08:32 workflow_run success 397db998   (no-op)
08:33 workflow_run success 397db998   (no-op)
```

Scheduled runs in the same window: **none.** The `23 * * * *` cron did not fire at 07:23 or 08:23. Earlier, at 06:34, a scheduled run *did* start — and was cancelled with zero jobs 3½ min later, squeezed out by the one-pending-slot rule the workflow's own comment at line 117 describes ("CANCELLED queued CD runs during merge bursts").

## Why it matters

The reconciler is **the delivery path** — `dotnet-test.yml` says so where it justifies `cancel-in-progress` on main: "main-cd no longer depends on a per-commit workflow_run: its hourly reconciler …". A push path that starves the reconciler is a push path that can silently stop delivering bundles the moment merge activity is high — which is exactly when a fresh identity most needs a bake.

## What would close it

- The `workflow_run` trigger should **exit before entering the concurrency group** when the sha has already been handled (the gate already knows this — it is what makes the run a no-op). A no-op that holds the slot is the defect.
- Or: give `schedule` its own concurrency group so a reconcile is never queued behind push-path runs.
- A `workflow_dispatch` takes the same reconcile branch (anything not `workflow_run` → `reason=reconcile`), so it can force the proof — but a republish is fleet-wide, which is why nobody dispatched it.

Refs #2452, #2464, #2467.

rbuergi:
## Recurrence 2026-08-30 — same starvation, different driver: MERGE VELOCITY, not no-op runs

[...]

PR #2760, opened 2026-08-30
fix(#2742): a permission fold that terminates without a verdict is Undetermined, not an ALLOW (both gates)

### LINKED ISSUE #2742: A starved permission fold parks every [RequiresPermission] delivery until the caller's own timeout — AccessControlPipeline has no terminal, and the bounded twin attributes the same event correctly
## What is failing

`PermissionEvaluator.GetEffectivePermissions` is a `CombineLatest` fold over the grant and policy reads of the target's scope **and every ancestor scope**. Per [Access Control](/Doc/Architecture/AccessControl):

> A leg that **starves** never emits, never completes and never errors … The fold therefore has **no terminal at all**, and a `.Take(1)` around it bounds the number of emissions rather than the wait. This is not hypothetical — it is the ordinary cross-silo shape, where the owning activation lives on a peer silo that is busy or has just gone away.

There are **two** gates in front of that fold and they behave differently:

| gate | bounded? | what the caller sees |
|---|---|---|
| `RlsNodeValidator` (node operations) | ✅ `PermissionEstablishmentBudget`, 20 s | `Unavailable` — **attributed to the starved read** |
| `AccessControlPipeline` (every `[RequiresPermission]` message, incl. `SubscribeRequest`) | ❌ none | the **caller's** timeout, 60 s later, naming the caller rather than the cause |

So one event produces two different-looking failures, and the more common one is the uninformative one.

## Evidence

Reported from a satellite deployment (private repo, so the detail stays there): intermittent first-read-after-idle failures, some surfacing as an attributed `Unavailable`, others as `No response received in hub cache/… within 00:01:00 for request SubscribeRequest → target …`, plus portal pages hanging on *"Subscribing to …"*. The two shapes are the two rows above. A cold fold has legs that must reach a peer activation; a warm one emits synchronously from the process-wide cached queries — which is exactly a first-read-starves / retry-succeeds shape, and it needs no replica straddle to explain it.

## 🚨 The fix is NOT a timeout on the gate

`AccessControlPipeline.cs:233` says so outright:

```csharp
// No Timeout here: the access cache must always be a reactive Subscribe … A 10s wait was a
// workaround for a wedged cache — fix the cache, don't ceiling-block here.
```

and that is the house rule (AGENTS.md → *No band-aids*): raising or adding a bound to make a symptom pass is not the fix. **The defect is that a leg of the fold can starve** — never emitting, completing, or erroring — so the work is to give those legs a terminal (an owner that cannot answer must answer *something*: the `TargetUnserved`/`ShuttingDown` verdicts the routing layer now produces are the model), not to ceiling-block the gate that waits on them.

Note the doc's own caveat, which is why this is filed as a decision and not just a patch: giving the shared gate a terminal *"changes the behaviour of every gated message and wants its own argument"*. If the conclusion is that the gate should also carry a terminal as defence-in-depth, that is a deliberate platform decision to be argued here — not something to slip in under a symptom fix.

## Where to look

- `src/MeshWeaver.Mesh.Contract/Security/PermissionEvaluator.cs` — `GetEffectivePermissions`, the `CombineLatest` fold and its legs. (#2679 / PR #2728 reworked this fold today so every service resolves once at entry rather than per emission, and a scope-teardown fault now terminates it as a classified benign teardown. That removes a *different* hazard on the same path — it does **not** address leg starvation, but any measurement should be taken on a build containing it.)
- `src/MeshWeaver.Hosting/Security/AccessControlPipeline.cs:233` — the deliberate absence of a terminal, and `:80-105` for the synchronous prologue that IS guarded (#974) and answers `Unavailable` on a disposing hub.
- `RlsNodeValidator`'s `PermissionEstablishmentBudget` — the bounded twin, and the model for what an attributed answer looks like.

## The measurement that discriminates

On a current build, hit an idle node and check which of the two shapes comes back: the **attributed `Unavailable`** (bounded path) or the **60 s `SubscribeRequest` timeout** (unbounded path). If it is always the latter, the caller-side symptom is the gate's missing terminal; either way the starving leg is the root.

rbuergi:
Closes #2742

## What this fixes, and what it deliberately does not

The issue asked for the fold's **legs** to be given a terminal, following the `StartWith(empty)` pattern already on `ObserveGatedNodes`. Working through it leg by leg produced two results — one shippable fix, and one refusal with a proof.

### ✅ Fixed: a fold that terminates without a verdict was an **ALLOW**

The fold has **three** terminals, not two: it can emit, it can fault, and it can **complete without ever emitting** — which is exactly what one silent leg does to a `CombineLatest` (it completes the moment any source completes having produced no value). Only the first two had a representation, so the third left `CheckPermissionOutcome`'s observable **empty**.

An empty check is not a refusal anywhere downstream. `AccessControlPipeline` ends its decision chain in `.Take(1).Select(…).DefaultIfEmpty()`, whose `null` means *"no check refused ⇒ every check was granted"* — so it invoked `next` and **delivered the message**.

Measured on this tree before the fix (`PermissionFoldSilentTerminalTest`):

```
Expected a DeliveryFailureException, but no exception was thrown.
```

A `[RequiresPermission(Read)]` `GetDataRequest` was answered normally, with no failure, no log line, nothing — a full authorization bypass, reachable through the public `WithPermissionEvaluator` seam. It is the **fail-open half of the same defect class** the issue is about: a leg of the fold that produces no value.

`CheckPermissionOutcome` now materialises that terminal as `Undetermined` → `IsGranted=false` → the delivery is **refused** and reported as `ErrorType.Unavailable` (retryable), never "Access denied" — exactly the #974 rule. Nothing is bounded, nothing is defaulted to a verdict, and a healthy fold never reaches the new leg.

### ✅ Fixed: the SAME hole in the node-operation gate — a silent fold let an ungranted caller **WRITE**

`RlsNodeValidator` ends its chain in `TakeDecisionOutsideGate().Timeout(budget).Catch(…)` — a value and a fault, again not the third terminal. `Take(1)` on an emptied fold completes empty, `Timeout` forwards that completion unchanged (it bounds silence *before* a terminal, not an empty one), the `Catch` never fires, and the validator emits **no** `NodeValidationResult`. `RunCreationValidatorsObs`'s `Concat` skips a validator that yields nothing.

Measured before the fix — a caller holding **no grant at all**, under a silent evaluator:

```
RESULT: MeshNode { Path = SilentRls/Child, …, State = Active, CreatedBy = probeuser }
ERROR:  (none)
```

Not slow, not denied — **written**. The chain now ends in `.DefaultIfEmpty(UnestablishedCheck(…, cause: null))`, and `UnestablishedCheck` takes a *nullable* cause so its message names the silent case apart from the stalled and the faulted one (three roads, three causes — "did not answer within 20s" about a fold that never started sends the operator to the wrong place). Answered `Unavailable`, never a denial; fail-**closed**.

> **The rule, stated once in the doc:** wherever a permission answer is consumed, **"no outcome" is not consent.** A chain that can end without emitting must materialise that ending as a refusal — never leave it empty for a downstream `DefaultIfEmpty` / `Concat` / `Where` to read as "nothing objected".

### ❌ Refused: seeding the other three legs — every one of them is fail-OPEN

The rule that decides it is **monotonicity**:

> A leg may carry an empty seed **iff** its contribution is purely **ADDITIVE**.

`ObserveGatedNodes` is the only leg that qualifies — `GateGrant` never subtracts, and says so. The other three all subtract, so an empty seed drops their subtractions and the fold's first emission is **more permissive than the truth**:

| leg | why an empty seed is unsafe |
|---|---|
| `ObserveEffectiveAssignments` | `ComputeScopeRoles` derives `Denied` from the **same** nodes as `Granted`, so a seed drops runtime denials of roles that survive it (a static grant, the self-partition Admin) → **fail-OPEN**. It also drops every *runtime* grant — which is what almost all real grants are — so a cold scope's first emission is `Permission.None`. |
| `ObserveScopePolicies` | `GetPermissionCap()` and `BreaksInheritance` are **restrictions**, so their *absence widens*: the seed falls back to the static policy map and the verdict ignores every runtime cap and inheritance boundary → **fail-OPEN**. |
| `ObserveAllMembershipNodes` | the subject set gates **denials** as much as grants, so a seed keeps a viewer's direct grant while dropping the group deny that revokes it → **fail-OPEN**. |

The brief's premise for the assignment and membership legs ("empty ⇒ denied ⇒ stricter") is the half that does not hold: both legs carry denials.

**And the fail-closed direction is not acceptable either.** `AccessControlPipeline` takes the fold's **first** emission as its verdict (`CheckPermissionOutcome(…).TakeDecisionOutsideGate()`, and `TakeDecisionOutsideGate` is a `Take(1)`). A seed is not a "brief pre-load window that settles a moment later" — from the gate's point of view it *is* the answer. Seeding the assignment leg therefore answers an entitled user on a cold scope with **"Access denied"**: the false, actionable-looking verdict #974 exists to prevent, now fired on every cold scope. Measured, not argued — `TheAssignmentLegCarriesEVERYRuntimeGrant_…` goes RED under an assignment-only seed.

So: **there is no permissive seed that is not a hole and no conservative seed that is not a spurious denial.** The fold genuinely needs its inputs; a starving leg's only sound terminal is an **error**, which the fold already propagates as `Undetermined` → `Unavailable`.

### 🚦 The residual, and where its fix belongs

The *silent-starvation hang* (a leg that never emits, completes, or errors) is **not fixed here** — deliberately, rather than traded for a wrong answer. `AccessControlPipeline` is untouched: **no terminal was added to the shared gate**, per the maintainer decision and the file's own "No Timeout here" comment.

The detection already exists one layer down and only logs: `MeshQuery.InitialStallProbe` (20 s) warns *"the query is silently stalled on its all-providers Initial gate and its consumer hangs with no error. Fix the stalled provider; **never bump the consumer's timeout**."* Turning that probe into a terminal would give every consumer — the permission fold included — the attributed `Unavailable` the bounded twin (`RlsNodeValidator`) already produces. That changes every query in the mesh, so it is a platform decision in its own right and is written up rather than slipped in under a symptom fix.

## Tests

`test/MeshWeaver.Security.Test/`

- **`PermissionFoldSilentTerminalTest`** — the regression pin. RED before the fix (no exception at all: the request was delivered). Plus an over-reach control on the same mesh proving a healthy fold still answers, so a fix that refused everything could not pass.
- **`RlsSilentFoldWriteTest`** — the same pin for the write path. RED before the fix (the node was created). Plus its own over-reach control proving a healthy fold on the same mesh still writes.
- **`PermissionFoldLegSeedGuardTest`** — three **cold-fold** cases pinning the seeding rule. Each seeds its fixture under `ImpersonateAsSystem` (the System short-circuit means no fold runs, so the scope's `$security-*` queries stay cold) and asserts the fold's *first* emission.
  - Non-vacuity, measured by applying `.StartWith(empty)` and re-running: the **assignment** case is RED under an assignment-only seed and under an all-legs seed — a true seed detector. The **policy** and **membership** cases stay green under a single-leg seed, because the assignment leg answers later; they are therefore honest **semantic pins** of the subtractivity the rule rests on, and the class doc says exactly that. Catching those two directly needs fault injection into `IMeshNodeStreamCache.GetQuery`, which this suite does not have.

Local: `MeshWeaver.Security.Test` **374/374 passed** (32 s) — covering `PermissionFoldAfterScopeTeardownTest`, `PermissionFoldUnavailableTest`, `SecurityServiceTests`, `GroupPermissionTests`, `NodeTypeGateTests`, `AdminPartitionAdminTests`, `UserPublicReadTest`, `HubCredentialReadAccessTest`. Every touched project builds `-c Release -warnaserror` with `0 Error(s)`.

## Docs

- `Doc/Architecture/AccessControl` → new **"The convergence contract"**: the monotonicity rule, the per-leg fail-open argument, why the gate stays unbounded, the "silence is not consent" finding for **both** consumers, and where a real fix for silent starvation would have to live.
- `Doc/Architecture/PermissionApi` → the exactly-one-outcome contract on `CheckPermissionOutcome`.
- What's New entry, `Category: Fix`.

### DIFF
--- a/src/MeshWeaver.Documentation/Data/Architecture/AccessControl.md
+++ b/src/MeshWeaver.Documentation/Data/Architecture/AccessControl.md
@@ -408,11 +408,11 @@ fold therefore has **no terminal at all**, and a `.Take(1)` around it bounds the
 rather than the wait. This is not hypothetical — it is the ordinary cross-silo shape, where the
 owning activation lives on a peer silo that is busy or has just gone away.
 
-Not every leg is like that, and the difference is deliberate. `ObserveGatedNodes` starts with
-`.StartWith(empty)` precisely so a slow leg cannot stall the fold — safe because a gate only ever
-*adds* `Read`. The grant and policy legs **must not** be seeded: "no grants yet" reads as a denial,
-which would be a silent wrong answer. So an unanswerable read cannot be projected onto the yes/no
-axis at all. It needs a third outcome:
+Not every leg is like that, and the difference is deliberate — see
+[the convergence contract](#-the-convergence-contract--which-legs-may-be-seeded-and-why-the-gate-is-not-bounded)
+below for the rule that decides it. `ObserveGatedNodes` starts with `.StartWith(empty)` precisely so a
+slow leg cannot stall the fold; the grant, policy and membership legs **must not** be seeded. So an
+unanswerable read cannot be projected onto the yes/no axis at all. It needs a third outcome:
 
 | Outcome | Means | Reported as |
 |---|---|---|
@@ -443,6 +443,75 @@ deadline, and it must re-emit when the grants finally land.
 Before this existed, `CreateNodeRequest` simply sat `Executing` — 33 s in the reported case — until
 its *caller's* `RequestTimeout` ended it, which names the caller's impatience rather than the read
 that starved (#1446).
+
+### 🚨 The convergence contract — which legs may be seeded, and why the gate is not bounded
+
+The fold's liveness problem has one obvious cure — give every leg a `.StartWith(empty)` so
+`CombineLatest` can never park on a slow source — and it is **wrong for three of the four legs**. The
+rule that decides it is **monotonicity**:
+
+> A leg of the permission fold may carry an empty seed **if and only if** its contribution is purely
+> **ADDITIVE**. A leg that also SUBTRACTS must never be seeded: the seed drops its subtractions, and
+> the fold's first emission is then more permissive than the truth.
+
+| leg | additive contribution | subtractive contribution | seeded? |
+|---|---|---|---|
+| `ObserveGatedNodes` | `GateGrant` ORs in `Read` on a declared public surface | **none** — "a gate NEVER subtracts" | ✅ **yes**, `StartWith(empty)` |
+| `ObserveEffectiveAssignments` | every runtime `AccessAssignment` grant | `Denied` role assignments, from the *same* nodes (`ComputeScopeRoles` returns `Granted` **and** `Denied`) | ❌ no |
+| `ObserveScopePolicies` | `PublicRead` | `GetPermissionCap()`, `BreaksInheritance` — and their **absence widens** | ❌ no |
+| `ObserveAllMembershipNodes` | group grants reach the viewer | the same subject set decides which **denials** match | ❌ no |
+
+**Why the first emission is the whole story.** `AccessControlPipeline` runs
+`hub.CheckPermissionOutcome(…).TakeDecisionOutsideGate()`, and `TakeDecisionOutsideGate` is a
+`Take(1)`. The fold's **first** emission *is* the verdict for every `[RequiresPermission]` delivery. A
+seed is not a "brief pre-load window that settles a moment later" — from the gate's point of view it
+is the answer, permanently.
+
+Each refusal, concretely:
+
+- **`ObserveEffectiveAssignments`** is fail-**OPEN** in the narrow direction (an empty seed drops a
+  runtime *deny* of a role that survives the seed — a static grant, the self-partition Admin) and
+  catastrophically wrong in the wide one: it drops every *runtime* grant, which is what almost all
+  real grants are. The first emission on a cold scope is `Permission.None`, so an entitled user is
+  answered **"Access denied"** — the false, actionable-looking verdict #974 exists to prevent, now
+  fired on every cold scope. A hang is a bad answer; a wrong answer is worse.
+- **`ObserveScopePolicies`** is fail-**OPEN**. `ComputeRoleState` falls back to the *static* policy
+  map when a scope has no runtime policy, and a missing policy means `cap = (Permission)~0` with
+  inheritance unbroken. An empty seed therefore hands the gate a verdict that ignores every runtime
+  permission cap and every `BreaksInheritance` boundary.
+- **`ObserveAllMembershipNodes`** is fail-**OPEN**. The viewer's subject set gates *denials* as much
+  as grants, so an empty seed keeps a viewer's direct grant while dropping the group deny that
+  revokes it.
+
+**The conclusion, and it is a strong one:** there is **no permissive seed that is not a hole and no
+conservative seed that is not a spurious denial**. The fold genuinely needs its inputs, so a starving
+leg's only sound terminal is an **error** — which the fold already propagates as
+`PermissionCheckOutcome.Undetermined` → `ErrorType.Unavailable` (retryable, and attributed to the read
+rather than to the caller's impatience). Making a *silent* starvation produce that error is a
+query-layer change, not a fold change: `MeshQuery` already **detects** it (`InitialStallProbe`, 20 s,
+whose own warning says *"Fix the stalled provider; never bump the consumer's timeout"*) and
+deliberately only logs. Turning that probe into a terminal would change every query in the mesh and is
+a platform decision in its own right — it is tracked, not slipped in under a symptom fix.
+
+**The gate stays unbounded, deliberately.** `AccessControlPipeline` carries no `Timeout` (see its "No
+Timeout here" comment) and this change does not add one. Bounding the shared gate would change the
+behaviour of every `[RequiresPermission]` message; more importantly a ceiling there cannot attribute
+anything — it would only say "the gate ran out of time", never *which* read starved. That attribution
+is exactly what `RlsNodeValidator`'s `PermissionEstablishmentBudget` provides on the node-operation
+path, and it is why the bounded twin's `Unavailable` is a *better* answer than the unbounded path's
+60 s caller timeout even though both describe the same event.
+
+**Silence is not consent (#2742).** The fold has **three** terminals, not two — emit, fault, and
+*complete without ever emitting* — and only the first two were represented. `CombineLatest` completes
+the moment any source completes having produced no value, so one silent leg empties the whole fold;
+`CheckPermissionOutcome` passed that empty completion straight through; and an empty check is not a
+refusal anywhere downstream — the pipeline's `.Take(1).Select(…).DefaultIfEmpty()` reads `null` as
+"no check refused ⇒ every check was granted" and **invokes the handler**. Measured before the fix: a
+`[RequiresPermission(Read)]` `GetDataRequest` was delivered and answered normally with no failure at
+all — a full authorization bypass, reachable through the public `WithPermissionEvaluator` seam.
+`CheckPermissionOutcome` now materialises that terminal as `Undetermined`, so it fails **closed** and
+reports `Unavailable` rather than vanishing. Pinned by `PermissionFoldSilentTerminalTest`; the seeding
+rule above is pinned by `PermissionFoldLegSeedGuardTest`.
 
 ### 🚨 The budget is DERIVED, never configured beside the bound it sits inside (#1198)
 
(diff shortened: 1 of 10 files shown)


rbuergi:
## 🛑 MERGE HOLD — do not merge this PR on green

Not a CI problem: this PR is fine and `main` is not red. **CD on `main` is being starved by merge velocity**, and merging now would evict the queued CD run again.

Measured on `main-cd.yml` at 2026-08-30T11:15Z:

| signal | value |
|---|---|
| last CD run to actually complete | `success` at **10:28:25Z**, `de321a440` |
| conclusions over the last 60 runs | **42 cancelled**, 14 success, 2 failure, 2 in flight |
| last 12 runs | all `cancelled` bar the newest, head SHA cycling `6551e649` → `b7628fc2` → `9003d0d0` |

`main-cd.yml` declares `concurrency: group: main-cd-${{ github.ref }}` with `cancel-in-progress: false`, so GitHub holds exactly **one** pending slot and each new CD arrival evicts the previously queued one. With merges landing faster than the ~18-minute multi-arch publish, CD never gets a clear window — the starvation the workflow's own header documents (#2490, 2026-08-27).

This is AGENTS.md's *"merging is a shared action"* rule: a merge supersedes the run queued behind it, including a deploy another session is gating on.

**State of this PR while held:** green-and-ready, watches armed, **no auto-merge enabled**. It will be merged only on an explicit go from the session coordinating the CD window. If it goes `DIRTY` in the meantime it will be re-synced — but still not merged.

rbuergi:
## 🔴 CI red on `e0cd6e9f5` — attributed to #2616, NOT to this diff. Not merging; not re-running.

`Run tests (shard 2)` → `MeshWeaver.Hosting.Monolith.Test.MeshNodeLanguageServiceTest.ScriptCompletions_FilterByTypedPrefix_NotJustTheAlphabet`

```
System.InvalidOperationException : MeshNodeLanguageServiceTest teardown left work RUNNING:
teardown DIRTY — 1 pooled I/O leaf(s) still running, async dispose queue drained.
  at MonolithMeshTestBase.DisposeAsync() : MonolithMeshTestBase.cs(1621,0)
```

This is **not** the `NodeTypeCompileParkTest` intermittent, so it does not fall under the standing exception. It was measured rather than pattern-matched, because "it's a known flake" is exactly the claim that must not be assumed.

### 1. The same src ran GREEN on this shard 20 minutes earlier

| commit | contains every src change in this PR | `Run tests (shard 2)` |
|---|---|---|
| `eace590e` | ✅ yes | **success** |
| `e0cd6e9f5` | ✅ yes (identical src) | failure |

`git diff --stat eace590ea e0cd6e9f5` is **three files, all in `test/MeshWeaver.Security.Test/`** — a different assembly from `MeshWeaver.Hosting.Monolith.Test`. The delta cannot reach this test, and the src half was already exercised green on this exact shard.

### 2. The diff is structurally incapable of leaking a pooled I/O leaf

The entire **code** delta in `src/` (comments excluded) is:

```csharp
.Select(result  => (NodeValidationResult?)result).DefaultIfEmpty().Select(result  => result  ?? UnestablishedCheck(…))
.Select(outcome => (PermissionCheckOutcome?)outcome).DefaultIfEmpty().Select(outcome => outcome ?? Undetermined(…))
```

plus a nullable `Exception?` parameter and one `why`-string branch. No IO, no `IIoPool`, no new subscription, no lifetime change — `Select`/`DefaultIfEmpty` are stateless pass-through operators that dispose with the subscription they wrap.

### 3. The failure shape is wrong for the plausible mechanism

The hypothesis worth taking seriously was that a fail-**closed** change refuses a read the test needs. It does not fit: **the test itself PASSED.** The log shows `=== TEST END ===` with no assertion failure; only the *class* `DisposeAsync` failed, 30 s later, on a leaf that ignored its cancellation token. A refused permission surfaces as an assertion failure / `Unauthorized` / `Unavailable` **inside** the test — never as an I/O leaf that will not cancel.

### 4. Local distribution — no branch-only signal

| set | branch (`e0cd6e9f5`) | main (`be39fc607`) |
|---|---|---|
| single test, ×5 | 5 pass / 0 fail | 5 pass / 0 fail |
| whole class, `DOTNET_PROCESSOR_COUNT=4`, ×5 | 5 pass / 0 fail | 5 pass / 0 fail |

20 runs, zero reproductions on either side — consistent with the CI-load-dependent race, and with no branch-only component.

### 5. Prior art: #2616 is OPEN with this exact signature

[#2616](https://github.com/Systemorph/MeshWeaver/issues/2616) — *"MeshNodeLanguageServiceTest teardown DIRTY — one pooled I/O leaf still ignores cancellation after #2598"* — records this same test and same message, first observed **on a docs-only PR whose diff could not reach a script-completion test**. (#2578 is the earlier closed instance of the same shape.) A new occurrence has been added there.

### Verdict

Pre-existing and ambient, owned by #2616. **This PR is not merged and the run is not re-triggered** — re-running to obtain a green wall would hide a live race, and the house rule is to attribute it, not retry it. Handing the merge decision back to the session coordinating the CD window.

PR #2767, opened 2026-08-30
fix(ci): give every main commit its own concurrency group — cancel-in-progress: false was not enough

rbuergi:
## Three commits landed on main today without ever being built or shipped

They report `cancelled`, which reads like someone cancelled them. Nobody did — they were **evicted while queued**.

`cancel-in-progress: false` (#2412) protects the run that is already **executing**. GitHub *also* keeps exactly **one pending run per concurrency group**, and every main push shared `build-test-refs/heads/main`. So a third arrival evicted the second's queued run, which reported `cancelled` without running a step.

Measured 2026-08-30, one merge burst — note `run_started_at == created_at` even while a run is only queued, which is what makes this look like a mid-run cancellation:

| commit | queued | cancelled |
|---|---|---|
| `dca02bdd7` | 11:42:10 | **11:53:02** — 1 s after the next push was created |
| `57de73c45` | 11:54:57 | **12:08:54** — 2 s after the next push |
| `637ee3921` | 12:08:52 | **12:09:16** — 1 s after the next push |

## Both failures #2412 exists to prevent happened anyway

1. **Nothing built the tree that landed.** With `strict: false` each PR is tested against the main it branched from, so the *merged* combination is first compiled by main's own run. Three landed commits never got one — this is the shape that put `CS0246: MeshOperations` on main on 2026-08-26.
2. **Nothing published.** `main-cd` gates delivery on `Consolidate test results` succeeding **for that SHA**, which an evicted run never produces. Three CD runs reported green with **every job `skipped`** while ACR's newest image stayed several commits behind.

## The fix

A push to `main` groups by `github.sha`, so no two main runs share a group and none can be evicted — which is exactly the cost #2412's comment says is deliberately accepted (*"one full run per merge to main"*). PRs still group by ref and still supersede, so #2316's ~28% saving is untouched.

## The guard was green the whole time

`MainRunsAreNeverCancelledGuard` asserted that the `cancel-in-progress` expression evaluates false on main. That was **true, and beside the point** — the invariant is *"a main run is never cancelled"*, and eviction violates it without touching that setting. A guard that checks the setting instead of the property is the same defect class this repo keeps finding.

It now also evaluates the **`group`** expression for two different main commits and requires the results to differ (and for two commits on one PR to match, so superseding is not silently disabled). Its evaluator gains value-faithful `&&`/`||` — Actions returns the **operand**, not a bool, which is precisely how `group:` selects a string.

## Verified by mutation, not by passing

| probe | result |
|---|---|
| restore the ref-grouped line that was live during the evictions | 🔴 *"puts every main commit in ONE concurrency group … yields `build-test-refs/heads/main` for two different commits"* |
| restore the fix | 🟢 11/11 |

(My first mutation attempt silently failed to apply and the suite passed — that "green" proved nothing and was redone. Noting it because it is the same false-pass shape this PR is about.)

## Not in scope

`main-cd.yml`'s push lane still groups by ref and can evict its own runs the same way; it already carries a partial fix for this (#2490, the reconcile lane split). Flagging rather than widening.

### DIFF
--- a/.github/workflows/dotnet-test.yml
+++ b/.github/workflows/dotnet-test.yml
@@ -152,12 +152,28 @@ on:
 # deliberate act — usually the hand-bisect above — and must never be cancelled by an
 # unrelated push landing on main.
 concurrency:
-  # PR runs group by `refs/pull/N/merge`, main pushes by `refs/heads/main` — i.e. both
-  # group by ref, and only a manual dispatch needs its own group. Kept on ONE line: in a
-  # folded (`>-`) scalar, continuation lines indented deeper than the first are preserved
-  # literally rather than folded, so a "tidier" multi-line version silently changes the
-  # string rather than just its layout.
-  group: build-test-${{ github.event_name == 'workflow_dispatch' && github.run_id || github.ref }}
+  # PR runs group by `refs/pull/N/merge`; a manual dispatch gets its own group; and a push to
+  # `main` groups by its own COMMIT. Kept on ONE line: in a folded (`>-`) scalar, continuation
+  # lines indented deeper than the first are preserved literally rather than folded, so a
+  # "tidier" multi-line version silently changes the string rather than just its layout.
+  #
+  # 🚨 MAIN GROUPS BY SHA, NOT BY REF — `cancel-in-progress: false` was NOT enough (#2412 was
+  # incomplete). GitHub keeps exactly ONE PENDING run per concurrency group: `false` protects the
+  # run that is already EXECUTING, but a third arrival still EVICTS the queued one, and an evicted
+  # run reports `cancelled` having never run a step. Grouping every main commit under one ref
+  # therefore reintroduced both failures the note below exists to prevent, silently.
+  #
+  # Measured 2026-08-30 (one merge burst, and `run_started_at` == `created_at` even while a run
+  # is only queued, which is what makes this read as "cancelled mid-run"):
+  #   dca02bdd7 queued 11:42:10 → cancelled 11:53:02, 1s after dfdd2d01c was created
+  #   57de73c45 queued 11:54:57 → cancelled 12:08:54, 2s after 637ee3921 was created
+  #   637ee3921 queued 12:08:52 → cancelled 12:09:16, 1s after 0c584a2d5 was created
+  # Three landed commits untested, and CD published nothing for any of them: ACR's newest image
+  # stayed several commits behind while three CD runs reported green with every job `skipped`.
+  #
+  # A per-SHA group cannot contend with anything, so every commit that lands on main gets its own
+  # full run — which is precisely the cost the note below says is deliberately accepted.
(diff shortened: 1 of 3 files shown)


rbuergi:
**Correction to the record: this was not a novel finding — `MeshWeaver.Plugins` fixed the same defect a day earlier, and I did not check the fleet for precedent before writing it.**

`MeshWeaver.Plugins/.github/workflows/ci.yml` already keys its main concurrency group on `github.sha`, and its comment states the mechanism in the same terms:

> 🚨 A MAIN PUSH GETS ITS OWN GROUP, KEYED ON THE COMMIT — never `github.ref`.
> `cancel-in-progress: false` (below) does NOT mean "nothing is cancelled". GitHub keeps at most ONE PENDING run per concurrency group: when a new run enters a group that already has a QUEUED run, the queued one is CANCELLED, whatever `cancel-in-progress` says. That flag only ever protects the run that is already IN PROGRESS.

That landed as Plugins #888 on 2026-08-29, driven by its own measurement — 22 of 25 main runs cancelled with `jobs=0, started=0`. Same defect, same fix, one repo over, one day before core's three commits were evicted.

Two things worth taking from that:

1. **The fix here is right and is now proven in production** (three core merges inside 2½ minutes, all three runs concurrent). Nothing about it changes.
2. **The independent rediscovery is the cost of not looking.** A fleet-shared CI defect solved in one repo should be swept across the others in the same change; instead core kept the hole for a day and paid three untested commits and a stalled publish for it. When a CI defect is found in one repo of this fleet, the next step is to grep the sibling repos for the same shape before writing the fix — the answer may already be written, with the measurement attached.

Filing that as the durable lesson rather than leaving this PR implying it was first.

rbuergi:
**Follow-up on the "main-cd's push lane still groups by ref" note above: it stays that way, deliberately, and here is why it must NOT get this PR's fix.**

Observed today: four consecutive `Continuous Delivery (main)` runs `cancelled` between 15:29 and 15:38 — the same single-pending-slot eviction this PR removed from Build-and-Test.

But CD's serialization is load-bearing in a way Build-and-Test's was not. `main-cd.yml` (lines 65–67) records the invariant:

> *"Re-publishing a superseded commit is not merely low-value, it is UNSOUND: the version tag comes from the BUILDING run's number, so republishing older code would mint a higher `-ci.<n>` for it and every install would roll "forward" onto older code — breaking the monotonic-build-number invariant VersionSelect documents in red."*

Group CD runs by SHA and two runs for consecutive commits can promote **concurrently**; whichever finishes last wins `main`/`ci.<n>`, and if that is the older commit, every self-updating install rolls backwards. The single ref-keyed lane is what makes "the last promote is the newest commit" true.

So the eviction of a *queued* CD run is a cost the design accepts, and the reconciler is its remedy — it heals HEAD on the hour, never a superseded commit. Measured today: three complete promoted sets (`ci.6691`, `ci.6714`, `ci.6748`) despite ~10 evicted runs. Nothing to fix here; recording it so the next reader does not apply this PR's fix to CD by analogy.

PR #2770, opened 2026-08-30
fix: a pooled-I/O residual NAMES the leaf that would not unwind

rbuergi:
## Why

A dirty teardown reports `teardown DIRTY — 1 pooled I/O leaf(s) still running` and, since #2480, **which pool**. It does not report **which leaf** — so the failure is unactionable.

Measured 2026-08-30, `MeshWeaver.AI.Test` (1,630 tests) under load (`DOTNET_PROCESSOR_COUNT=4` + CPU burners), 20 runs:

| run | test | residual | drain |
|---|---|---|---|
| 7 | `MeshNodeAgentFileStoreTest` | `AgentStore=1` | 30 081 ms |
| 12 | `PluginCatalogTest` | `Query=1` | 30 004 ms |
| 16 | `MeshPluginTest` | `FileSystem=1` | 30 043 ms |

Three pools, three tests, always the full budget — and nothing to act on. #2480 added the pool name for this exact reason and stopped one level short.

## How

`IoPool` keeps each in-flight leaf's work **by reference** — no stack capture, no string built on the hot path — and formats only when a drain reports a residual. A lambda's compiler-generated name carries its enclosing method, so the residual reads:

```
AgentStore=1 [MeshWeaver.AI.Stores.MeshNodeAgentFileStore+<>c__DisplayClass…<WriteCore>b__0]
```

All four entry points instrumented (`Invoke`, `InvokeStream`, `InvokeBlocking`, `SubscribeThroughPool` — the last has no delegate, so it reports the source observable's type). `PoolResidual.Sites` flows into both the drain warning and the `pools=[…]` phase trace `MonolithMeshTestBase` already prints, so a CI `test-evidence-*` artifact carries it with no further change.

## Verification

`IoPoolTest` 25/25, including two new pins: a leaf that ignores its token **must** be nameable, and a clean pool reports **no** sites (the diagnostic must not become noise).

Note: on current `main` this class did not reproduce in 24 further loaded runs (P ≈ 0.02 under the measured 15 % rate), so something in the intervening commits — most plausibly the continuing `ToTask` sweep — has largely fixed it. The diagnostic is worth landing regardless: the next occurrence names itself instead of restarting this investigation.

### DIFF
--- a/src/MeshWeaver.Mesh.Contract/Threading/IoPool.cs
+++ b/src/MeshWeaver.Mesh.Contract/Threading/IoPool.cs
@@ -1,3 +1,4 @@
+using System.Collections.Concurrent;
 using System.Reactive.Concurrency;
 using System.Reactive.Disposables;
 using System.Reactive.Linq;
@@ -155,6 +156,64 @@ public sealed class IoPool : IIoPool, IDisposable
     /// <summary>Number of operations currently executing through this pool.</summary>
     public int CurrentInFlight => Volatile.Read(ref _inFlight);
 
+    // 🚨 WHICH leaf, not just how many. The residual a dirty teardown reports has been an
+    // anonymous "AgentStore=1" — enough to know a pool did not drain, never enough to fix it.
+    // #2480 added the POOL NAME for exactly this reason and stopped one level short: measured
+    // 2026-08-30 on MeshWeaver.AI.Test under load, 3 of 20 runs ended in a dirty teardown naming
+    // AgentStore, Query and FileSystem on different runs — three pools, no leaf, nothing to act on.
+    //
+    // The delegate is kept BY REFERENCE while the leaf is in flight (no stack capture, no string
+    // built on the hot path) and formatted only when a residual is reported, so a pool that drains
+    // cleanly pays one dictionary insert and one removal per leaf and nothing else.
+    private readonly ConcurrentDictionary<long, object> _inFlightLeaves = new();
+    private long _leafSeq;
+
+    /// <summary>
+    /// Records an in-flight leaf so a residual can NAME it; returns its ticket. Takes the work
+    /// ITSELF — a delegate on the three functional entry points, the source observable on
+    /// <see cref="SubscribeThroughPool{T}"/>, which has no delegate to name.
+    /// </summary>
+    private long EnterLeaf(object io)
+    {
+        var id = Interlocked.Increment(ref _leafSeq);
+        _inFlightLeaves[id] = io;
+        return id;
+    }
+
+    /// <summary>Retires a leaf's ticket. Safe to call for an id that was never added.</summary>
+    private void LeaveLeaf(long id) => _inFlightLeaves.TryRemove(id, out _);
+
+    /// <summary>
+    /// The call sites of the leaves still in flight, most useful form first: a lambda's
+    /// compiler-generated method name carries its ENCLOSING method, so
+    /// <c>MeshQuery+&lt;&gt;c__DisplayClass22_0.&lt;MergeProviderObservables&gt;b__1</c> points at the
+    /// exact operation that did not unwind. Empty when the pool drained clean.
+    /// </summary>
+    internal IReadOnlyList<string> PendingLeafSites =>
+        _inFlightLeaves.Values
+            .Select(d =>
+            {
+                try
+                {
+                    // A delegate names its enclosing method through the compiler-generated
+                    // lambda; anything else (the subscribe path's observable) can only offer
+                    // its type, which still says WHICH chain is stuck.
+                    if (d is Delegate del)
+                    {
+                        var m = del.Method;
+                        return $"{m.DeclaringType?.FullName ?? "?"}.{m.Name}";
+                    }
+                    return d.GetType().FullName ?? d.GetType().Name;
+                }
+                catch
+                {
+                    // A diagnostic must never be the reason a teardown fails.
+                    return "(unavailable)";
+                }
+            })
+            .Distinct(StringComparer.Ordinal)
+            .ToArray();
+
     /// <summary>
     /// Runs an async I/O leaf off the calling scheduler under the pool's concurrency gate.
     /// </summary>
@@ -190,12 +249,14 @@ public sealed class IoPool : IIoPool, IDisposable
                 // the inner await, so the gate caps in-flight ops, not threads.
                 await _gate.WaitAsync(ct).ConfigureAwait(false);
                 Interlocked.Increment(ref _inFlight);
+                var leaf = EnterLeaf(io);
                 try
                 {
                     return await io(ct).ConfigureAwait(false);
                 }
                 finally
                 {
+                    LeaveLeaf(leaf);
                     // 🚨 RELEASE THE PERMIT FIRST. TryFinishDisposal disposes _gate the moment the
                     // last leaf's decrement takes _inFlight to zero — so with the decrement first,
                     // THIS leaf disposed the gate and then called Release() on it, throwing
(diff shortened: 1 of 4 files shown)


rbuergi:
The red is **this PR's own new test**, not an inherited or flaky one — and it is a **hang, not an assertion failure**, which changes what to look for.

```
MeshWeaver.Hosting.Test.IoPoolResidualNamesItsPoolTest
  .DrainAll_reportsTheResidual_againstTheNameOfThePoolThatLeaked
```

The Test Results annotation reports the trx as `[took 30s]`, and `test/xunit.runner.json` sets `methodTimeout: 30000`. So it ran to **exactly** its budget and was killed. There is no assertion message because it never reached one.

Two consequences worth stating explicitly:

1. **Nothing here is flaky-shaped.** Do not re-run it to see. A wait that consumes precisely the whole budget is a wait that never settles, and it will do so every time.
2. 🚨 **`methodTimeout` only bounds time spent INSIDE the test method.** If the hang were in fixture construction, class init or teardown it would run unbounded and present as a killed host instead. That it stopped cleanly at 30 s is itself evidence the wait is inside the method body — so the thing that never completes is something the test itself awaits.

Given the subject, the most likely shapes, in order:

- **The drain is waiting on the residual it is meant to report.** `IoPool.Dispose` must not block — the wait is an observable, not a `Wait()` — so if `DrainAll` is being awaited synchronously while the deliberately-leaked leaf never unwinds, the test is asserting its own deadlock.
- **The residual leaf is never cancelled.** Related, and now tracked as #2772: `ObserveCompletion` cancels the *wait*, not the source, so a pooled leaf reached through it is not cancelled and keeps its gate permit until it finishes naturally. If the test's leaf is meant to be released by cancellation, it will not be.

Worth checking whether the test's wait has its own bound and a positive success signal, rather than relying on the 30 s method timeout to end it — a timeout is not an assertion, and a test that can only fail by timing out reports the same thing whether the feature works or the harness stalled.

rbuergi:
## Root cause — the test cannot fit inside the harness's method timeout, and its `await` leaks the leaf when it doesn't

**1. The arithmetic makes the hang inevitable.**

```
IoPool.DrainTimeout            = 30 s          (IoPool.cs:81)
test/xunit.runner.json         methodTimeout: 30000   (= 30 s)
```

`Drain()` spends that budget **twice** on the path this test forces:

```csharp
var cancelResidual = StartCancelOffCallerThread().Join(DrainTimeout) ? 0 : 1;   // up to 30 s
for (var i = 0; i < _maxConcurrency; i++)
    if (_gate.Wait(DrainTimeout)) acquired++;  else break;                      // up to 30 s more
```

The test's leaf **deliberately ignores its token** and holds the permit for two minutes — that is the residual condition being pinned. So the drain is *required* to burn its full budget. A test that needs ≥30 s of wall clock cannot pass under a 30 s `methodTimeout`, and the trx confirms it: `[took 30s]`, killed at exactly the budget, with no assertion message because it never reached one.

`[Fact(Timeout = 180_000)]` did not save it — the observed kill is at 30 s, so the runner-level `methodTimeout` is what applied.

**2. The `await` is what makes the failure dirty rather than merely red.**

```csharp
var drain = Task.Run(() => total = registry.DrainAll(out byPool), TestContext.Current.CancellationToken);
await drain.WaitAsync(TimeSpan.FromSeconds(120), TestContext.Current.CancellationToken);
release.Set();          // ← unreachable once the await throws
```

`TestContext.Current.CancellationToken` *is* the method-timeout token. At 30 s it cancels, `WaitAsync` throws, and **`release.Set()` never runs**. So:

- the leaf keeps blocking a pool thread for its full 2 minutes,
- `using var registry` disposes with live work still inside it,
- and `DrainAll` is still blocking a ThreadPool thread after the test method is gone.

That is the residual escaping into the next test, from the test written to prove residuals are reported.

## Two fixes, and I'd do both

**a. `release.Set()` must be unconditional** — `try { await … } finally { release.Set(); }`. A signal that frees a blocked worker cannot sit after an `await` that is designed to be cancelled. This is worth doing even if the timing is fixed, because it is the difference between a failing test and a failing test that poisons the shard.

**b. The budget has to be injectable.** The docstring argues *"the budget is a contract, so shortening it for the test would be testing something else."* I'd push back: the subject of this test is the **residual is reported and NAMES its pool** — not the numeric value of `DrainTimeout`. Making the budget a constructor/optional parameter and driving it at, say, 200 ms tests the reporting exactly as written, deterministically, in under a second. Keep a separate one-line assertion that the *default* is 30 s if the constant is genuinely contractual.

As it stands the test can only ever report the same thing whether the feature works or the harness stalls — and it costs 30 s of shard time each run to say it.

rbuergi:
## Measured, from the trx — 30 s → 0.28 s

You asked whether CI actually waits that long. It did, and the trx is the place the per-test number lives. Branch `fix/iopool-drain-budget-injectable` (cut from this one) makes it:

```
Failed   00:00:00.2816728  DrainAll_reportsTheResidual_againstTheNameOfThePoolThatLeaked
Passed   00:00:00.0005564  TheDefaultDrainBudget_IsTheProductionContract
Passed   00:00:00.0013094  DrainAll_reportsNoPool_whenEveryLeafUnwound
whole class: 300 ms
```

Against the current branch, CI reports the same test as `[took 30s]` — killed at exactly `methodTimeout`, no assertion message.

**What changed.** `IoPool` gains an `(int maxConcurrency, TimeSpan drainTimeout)` overload and `IoPoolOptions` a `DrainTimeout`, threaded through `IoPoolRegistry.Get`. The default is **unchanged at 30 s** and now pinned by its own test, so making it injectable cannot quietly move the teardown contract. The test drives 250 ms and still calls the **real** `registry.DrainAll(out byPool)` — the subject is *"the residual is reported and NAMES its pool"*, not the numeric budget.

`release.Set()` also moves into a `finally`. It was the statement after `await drain.WaitAsync(…, TestContext.Current.CancellationToken)`, and that token **is** the method-timeout token — so on the timeout the await threw, the release never ran, and the deliberately-blocked leaf held a pool thread for its full two minutes while `using` disposed over live work. The test written to prove residuals are reported was leaking one into the next test.

## 🚨 The speed-up exposed a real defect the timeout was hiding

It now fails in 0.28 s on a genuine mismatch **in this PR's own diff**:

```
expected "Query=1"
found    "Query=1 [MeshWeaver.Hosting.Test.IoPoolResidualNamesItsPoolTest+<>c__DisplayClass0_0.<DrainAll…>b__0]"
```

`PoolResidual.ToString()` appends the pending leaf **sites**, but the test asserts `{pool}={n}` and calls it *"the wire format the teardown traces embed, so it is part of the contract"*. Both cannot be true. Either `ToString()` should stay the stable wire format with sites exposed as a separate member, or the contract assertion needs updating to the new shape — **that is your call, so I left it failing rather than papering over it.**

Worth noting what this says about the original: a 30 s timeout kill and a one-line format mismatch are indistinguishable from the outside. The test could only ever report "something went wrong", at a cost of 30 s of shard time per run.

Cherry-pick `fix/iopool-drain-budget-injectable` if you want it; I have not pushed to your branch.

rbuergi:
Shard 2 red was this PR's own test pinning the OLD wire format (`Query=1` exactly) while the PR extends it to `Query=1 [site]`. The assertion now checks the prefix unchanged, that the site names this test's own lambda, and one site per leaked leaf. Verified locally on this branch: 2/2 pass.

Process note for the record: my first push of this fix landed on `fix/iopool-drain-budget-injectable` — a stacked branch another session created from this PR's head, with no PR of its own. I restored that branch to exactly its prior tip (`b0a60eab4`, exact-lease force push of my one commit only) and applied the fix here instead. Nothing of the other session's work was touched.

PR #2771, opened 2026-08-30
fix: the Await bridge moves to a PRODUCT assembly — product code deadlocks too

rbuergi:
## The gap

`ObservableAwait.Await<T>` is the sanctioned replacement for `.ToTask()` — faithful semantics (last value, faults on empty) but `RunContinuationsAsynchronously`, so the continuation is **queued** instead of resuming the caller inline on the signalling thread.

It lives in **`MeshWeaver.Fixture`** — a *test* assembly.

So the bridges that actually deadlock a hub had nowhere to go. An SDK boundary we must implement — a Microsoft.Agent.Framework `…Async` override, `IHostedService`, ASP.NET middleware — has a signature that is not ours to change, cannot reference the test fixture, and was therefore left on Rx's `.ToTask()`.

## Measured

| tree | `.ToTask(` in product `src/` | ratchet |
|---|---|---|
| **MeshWeaver** (core) | **0** | `ObservableToTaskBridgeGuard` |
| **MeshWeaver.Plugins** | **65** | none |

Seven of those 65 are in `MeshNodeAgentFileStore` — the class whose test produced the `AgentStore=1` dirty-teardown residual I measured under load (3 of 20 runs, each burning the full 30 s drain budget).

## The change

Moved to `MeshWeaver.Messaging.Hub`, beside `ReactiveCompletion` (the other sanctioned bridge). `MeshWeaver.Fixture.ObservableAwait` keeps its name and **forwards**, so:

- the ~1,500 existing test call sites need no churn, and
- there is exactly **one** implementation of the semantics the sweep depends on — no drift between a test copy and a product copy.

Follow-up (not in this PR): sweep `MeshWeaver.Plugins` product `src/` onto this bridge, starting with `MeshNodeAgentFileStore`, and port core's ratchet so the count can only shrink.


rbuergi:
## The CI red was mine, and the fix changes the shape of this PR

The first push kept a `MeshWeaver.Fixture.ObservableAwait` forward "so ~1,500 call sites need no churn". That forward **was** the bug: the test tree imports **both** `MeshWeaver.Fixture` and `MeshWeaver.Messaging`, so two identical extension signatures made every `.Await(ct)` **ambiguous** — `CS0121`, 1,174 errors.

None of it was visible locally, because I built only the two projects I changed rather than their consumers. (And my first attempt to check that produced a *false pass*: I ran `dotnet build MeshWeaver.sln`, which does not exist here — the repo uses `.slnx` — so the run failed with `MSB1009` while my `grep 'error CS'` found nothing.)

**Now:** the forward is deleted, so there is exactly one implementation, and the namespace arrives the way the test tree already gets `MeshWeaver.Reactive.Assertions` — a single global `<Using>` in `test/Directory.Build.props`. Three test projects (`Json`, `PluginImage`, `Portal.E2E`) have no messaging dependency and are excluded by name, with the reason in the props file: a global using cannot resolve a namespace the project does not reference, and adding a `ProjectReference` just to satisfy a using would be the wrong direction.

**Verified the way it should have been the first time:** full `MeshWeaver.slnx` build, 158 projects, `MW_ALLOW_SOLUTION_BUILD=1` — **exit 0, zero errors**.

rbuergi:
The red on `74ab2e09c` is **live, not stale** — worth saying because the previous head's red (`CS0121 ambiguous Await<T>`) *was* stale and is already fixed by deleting the Fixture copy. This is a different failure:

```
A PUBLIC TYPE MOVED ASSEMBLIES WITHOUT A TYPE FORWARDER — binary-breaking (#2370):
  MeshWeaver.Fixture.ObservableAwait
      left   MeshWeaver.Fixture
      now in MeshWeaver.Messaging.Hub (MeshWeaver.Messaging.ObservableAwait)
```

The gate offers two remedies. **The second one is the correct one here, and it is provable rather than a judgement call.**

## Why the forwarder is the wrong remedy

The gate says *"Restore `namespace MeshWeaver.Fixture;` on the type in its new assembly, then add the forwarder."* That would undo the point of this PR — and your own `test/Directory.Build.props` comment already records that a forward in the old namespace made every `.Await(ct)` ambiguous (CS0121) in files importing both. So a forwarder trades this red for the previous one.

## Why the allow entry is justified

The gate's condition is *"If no shipped module can hold this TypeRef"*. For this type, **no module can, structurally** — `MeshWeaver.Fixture` is excluded from the script reference set by name:

```csharp
// src/MeshWeaver.Kernel.Hub/MeshScriptEnvironment.cs:125
public static bool IsTestScaffolding(string assemblyName)
    => assemblyName.Equals("MeshWeaver.Fixture", StringComparison.Ordinal)
       || assemblyName.EndsWith(".TestBase", …) || .EndsWith(".Test", …) || .EndsWith(".Tests", …);
```

`KernelScriptReferences.IsProductionReferenceable` applies that exclusion to every script compilation, deliberately — it is the fix for the 2026-08-12 "exports dead in prod" defect, where `MeshWeaver.Fixture` being loaded in a test process made test compilation *more permissive* than production. So in-mesh source has never been able to see `MeshWeaver.Fixture.ObservableAwait`, and cannot acquire a TypeRef to it in any process.

Two supporting facts: the type is **~3 hours old** (I added it in #2750, merged 11:54 today), and only one image has ever carried it (`3.0.0-rc8.ci.6714`, promoted 12:44).

## Suggested entry

```
# MeshWeaver.Fixture.ObservableAwait -> MeshWeaver.Messaging.ObservableAwait (#2771)
# No shipped module can hold this TypeRef: MeshWeaver.Fixture is excluded from the script
# reference set BY NAME (MeshScriptEnvironment.IsTestScaffolding), so in-mesh source has never
# been able to bind it. A forwarder is not an option — it cannot rename, and restoring the old
# namespace re-introduces the CS0121 ambiguity this PR exists to avoid.
```

🚨 And per the fleet rule this entry goes **stale the moment this merges** — it describes a move that will already have happened. Delete it in the very next change rather than leaving it as free budget; that is the exact pattern that left seven stale entries behind #2678→#2688.

rbuergi:
## The new shard-3 red is the bridge guard working correctly, and the fix is one register entry

```
ObservableToTaskBridgeGuard.EveryProductionBridgeIsASanctionedImplementation
ObservableToTaskBridgeGuard.TheSanctionedBridgeRegisterIsStillAccurate
```

This PR moves `ObservableAwait` into `MeshWeaver.Messaging.Hub` — a **production** assembly — and `ObservableAwait.Await` is a hand-rolled bridge (`TaskCompletionSource` settled from inside a `Subscribe`). #2762's structural detector finds bridges by *shape*, not by the `.ToTask(` spelling, so it sees the new one immediately. It is supposed to.

`SanctionedBridges` is not an allow file: every entry must still exist, still contain a bridge, and still construct with `RunContinuationsAsynchronously` — an entry whose subject moved or was fixed **fails**. So the remedy is to register the new implementation, not to widen a list.

**It qualifies as the safe form.** `ObservableAwait.Await` creates its completion source as
`new TaskCompletionSource<T>(TaskCreationOptions.RunContinuationsAsynchronously)` and disposes its subscription when the task settles, so it is exactly the shape the register exists to admit.

Suggested entry, in the register's own idiom:

```csharp
("src/MeshWeaver.Messaging.Hub/ObservableAwait.cs",
    "The await-side bridge, moved here from MeshWeaver.Fixture because the bridges that actually "
    + "deadlock are in PRODUCT code: an SDK boundary we must implement (a Microsoft.Agent.Framework "
    + "…Async override, IHostedService, ASP.NET middleware) cannot reference a test fixture, so those "
    + "sites had nowhere to go and stayed on Rx's .ToTask(). Safe form: constructs with "
    + "RunContinuationsAsynchronously and disposes its subscription when the task settles."),
```

## Both reds on this head are the same root cause

The binary-compat red and this one are two consequences of one move:

| red | remedy |
|---|---|
| `Public surface (binary compatibility)` | an entry in `scripts/type-forwards.allow` — no shipped module can hold that TypeRef, because `MeshScriptEnvironment.IsTestScaffolding` excludes `MeshWeaver.Fixture` **by name** from the script reference set (details in my earlier comment) |
| `ObservableToTaskBridgeGuard` ×2 | the `SanctionedBridges` entry above |

Neither is a defect in the move itself — both are gates correctly noticing that a public bridge changed assemblies, which is exactly what they were built for. Worth saying plainly, because this is the second PR today where a guard firing was the guard being right.

PR #2781, opened 2026-08-30
docs(homebrew): brew trust systemorph/memex before install

rbuergi:
First install from the published tap (v0.2.11427) on this Mac: `Error: Refusing to load formula systemorph/memex/memex-local from untrusted tap systemorph/memex.` — current Homebrew refuses a third-party tap's formulae until `brew trust <tap>` has been run. The README, LocalColimaMac §1 and the formula's header comment now carry the step (the tap's own README already does). Docs + a comment only.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01C1SquDKJNKZxEUbuaPjUgu

Issue #2778, opened 2026-08-30
OwnerDisposing NACK is dropped exactly when the parent is disposing hosted hubs — the caller hangs for its full budget


rbuergi:
## The hole

`RegisterOwnerDisposingNack` (`src/MeshWeaver.Data/DataExtensions.cs`) mints the transient `OwnerDisposing` NACK for an in-flight `PatchDataRequest`, then delivers it like this:

```csharp
var parent = hub.Configuration.ParentHub;
if (parent is not null && parent.RunLevel < MessageHubRunLevel.DisposeHostedHubs)
    parent.Post(resp, o => o.ResponseFor(request));
```

With `MessageHubRunLevel` ordered `Starting, Started, Quiescing, DisposeHostedHubs, HostedHubsDisposed, ShutDown, Disposed`, the NACK is delivered **only** while the parent is `Starting`/`Started`/`Quiescing`. Two cases yield silence:

- `parent is null`
- `parent.RunLevel >= DisposeHostedHubs` — i.e. **precisely when the parent is disposing its hosted child hubs**, which is the very moment a batch of child streams goes down with patches in flight.

So the caller gets exactly the silence the NACK exists to replace, and waits out its whole budget.

## The assumption, and why it is false

The guard's rationale is stated in the sibling `NackSilentRead` docs from the #1362 fix (0bf4456c2):

> *"Posts through this hub while it can still post, and through the PARENT once `RunLevel >= DisposeHostedHubs` closes its own gate … **During a whole-mesh teardown the parent is past that mark too, the post is skipped, and nobody is waiting.**"*

**"Nobody is waiting" is an assumption the code cannot verify, and it is wrong here.** A caller whose wait outlives the start of teardown is still waiting, and it is the caller who most needs the answer that is guaranteed not to get it.

## Evidence

`MeshWeaver.Plugins` `StreamingCellWriteByteCountTest.StreamingACellsText_CostsTheAnswerAboutOnce_NotOncePerTick`, intermittent on Plugins `main` (green `7e0750c27`, `42604ce8b`; red `6b995913`):

```
[SYNC_STREAM] Not setting HJHG_KBlWkOilS_4bxOMMA — stream is disposed
[SYNC_STREAM] Not setting lf6CBIvxx0-gduhIkD7wzA — stream is disposed
… ~12 more, ~13 DISTINCT stream ids over ~5 s …
System.TimeoutException : The operation has timed out.
[...]

Issue #2773, opened 2026-08-30
ObserveCompletion does not cancel a pooled IIoPool leaf — cancelling the wait no longer releases the gate permit


rbuergi:
## The regression

`IoPool.InvokeStreamCore` / `InvokeObservable` build their work with `Observable.Create<T>(async (observer, subscriberCt) => …)` and link the leaf's token to `subscriberCt`:

```csharp
using var linked = CancellationTokenSource.CreateLinkedTokenSource(subscriberCt, _poolCts.Token);
var ct = linked.Token;
await _gate.WaitAsync(ct).ConfigureAwait(false);
```

`subscriberCt` fires when the **subscription is disposed**. So disposing the subscription is what cancels the pooled leaf and releases its gate permit.

Rx's `.ToTask(ct)` disposed the subscription on cancellation. **`ReactiveCompletion.ObserveCompletion` deliberately does not** — its own doc says *"Cancels the WAIT, not the source. The error arm stays [attached]"*, and that non-disposal is the designed behaviour that lets a late fault still be reported instead of vanishing.

The consequence: at an `ObserveCompletion` boundary wrapping a pooled leaf, cancelling the caller no longer cancels the I/O or frees the pool slot. The leaf runs to completion holding its permit. **PG pools are capped at 1** (`pg:{adapter}`), so the gate *is* the single Npgsql connection — an aborted request can leave the next caller waiting for a query nobody wants any more.

## Measured blast radius — smaller than first reported

A first report put this at "core's own ~1,538 sites". That is **not right**, and the correction matters for prioritisation:

| surface | passes a real `ct` | disposes on cancel? |
|---|---|---|
| `ObservableAwait.Await` (the 1,538 test conversions, #2750) | yes | ✅ **yes** — `sub.Dispose()` in its settle continuation, matching `.ToTask()` |
| `ObserveCompletion` in core `src/` | 20 of 37 sites | ❌ no (by design) |
| …of those 20, in **production** code | **2** | ❌ no |

The two production sites are both `GenericUnpartitionedDataSource` initial-store builds (`.FirstAsync().ObserveCompletion(ReportLateInitFault, cancellationToken)`). Their token is an initialization/lifetime token, so the realistic symptom is a **delayed shutdown or a held slot during teardown**, not steady-state exhaustion.

## Proposed fix

An **opt-in** overload — the default must keep today's behaviour, because non-disposal is what preserves late-fault reporting:

```csharp
[...]

PR #2788, opened 2026-08-30
fix(test): the inline-resumption probe compared thread IDs, which proves nothing

rbuergi:
## The intermittent

`InlineResumptionMechanismTest.ObserveCompletion_DoesNotResumeOnTheSignallingThread` failed on **#2781 — a documentation-only PR** (a Homebrew formula, a README, one doc page). Nothing in that diff can reach a threading primitive.

## Why it fails, and why re-running would have buried it

It asserted `NotEqual(signalling, resumed)` on `Environment.CurrentManagedThreadId`. **That proves nothing in the negative direction:** a correctly *queued* continuation runs on a pool thread, and the pool is free to hand it the very thread that just finished signalling. The assertion fails while the behaviour under test is entirely correct.

🚨 The same unsound shape is already written down two files away, in `ActivityLiveProgressTest`:

> *"`createdThread != commandThread` — the runtime makes no such [guarantee]"*

Rediscovered rather than reused. This is my own test family from the ToTask sweep, so the miss is mine.

## What it measures now

The continuation ran **on the signalling thread AND while the signaller was still inside `OnNext`/`OnCompleted`**.

Neither half suffices, and I proved both insufficient rather than assuming:

| probe | verdict |
|---|---|
| thread ids differ | the intermittent this replaces |
| "the waiter had completed when `OnCompleted` returned" | **my first attempt** — reported INLINE on **5 of 5** runs against `ObserveCompletion`, because the pool runs a queued continuation inside that window and measurably does |
| same thread **and** inside the call | rigorous |

Together they are rigorous because **a thread cannot be in two places at once**: if the continuation ran on the signalling thread while that thread was still inside the signalling call, it was on the signaller's stack. A queued continuation that later lands on that recycled thread necessarily finds the flag cleared — the thread was occupied until the signaller left.

## The mutation proof is built in

The two **positive controls** — Rx's `.ToTask()` bridge, and awaiting the observable directly — must report `TRUE`. They are the negative controls for this measurement: if it stopped discriminating, they would go red. So the proof lives in the file rather than being performed once and forgotten.

Verified: **6 consecutive runs, 3/3 green each.**

Note the two positive assertions were always sound — inline resumption *does* imply the same thread — so only the negative direction needed replacing.

### DIFF
--- a/test/MeshWeaver.Messaging.Hub.Test/InlineResumptionMechanismTest.cs
+++ b/test/MeshWeaver.Messaging.Hub.Test/InlineResumptionMechanismTest.cs
@@ -33,16 +33,38 @@ namespace MeshWeaver.Messaging.Hub.Test;
 /// </summary>
 public class InlineResumptionMechanismTest
 {
-    /// <summary>Signals from a known thread and reports where the awaiter resumed.</summary>
-    private static async Task<(int Signalling, int Resumed)> Measure(Func<IObservable<int>, Task> wait)
+    /// <summary>
+    /// Signals from a known thread and reports whether the awaiter resumed INLINE — that is,
+    /// whether the wait had already completed by the time the signalling call returned.
+    ///
+    /// <para>🚨 <b>This used to compare managed thread IDs, and that probe is unsound.</b>
+    /// <c>Environment.CurrentManagedThreadId</c> differing proves nothing: a correctly QUEUED
+    /// continuation runs on a pool thread, and the pool is free to hand it the very thread that
+    /// just finished signalling. So <c>NotEqual(signalling, resumed)</c> fails intermittently on a
+    /// loaded shard while the behaviour under test is perfectly correct — observed 2026-08-30 on a
+    /// documentation-only PR. The same unsound shape is recorded two files away, in
+    /// <c>ActivityLiveProgressTest</c>: <i>"createdThread != commandThread — the runtime makes no
+    /// such [guarantee]"</i>.</para>
+    ///
+    /// <para>Inline resumption has a DIRECT observable: the continuation runs on the signaller's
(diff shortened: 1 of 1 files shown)


rbuergi:
**Superseded by #2791, which landed first and is sound.** Closing rather than forcing a competing rewrite of the same test — that would be churn on a file someone has just fixed.

#2791 removes the confound instead of measuring around it: the signal is sent from a **dedicated `Thread`**, not a pool worker. A live thread's managed ID is unique, so a queued continuation running on a pool thread can never share it — which makes `NotEqual(signalling, resumed)` sound in the direction that was previously unsound, while the two `Equal` assertions keep holding for the inline cases. Simpler than what I had.

**Third time today I have duplicated a sibling's work** (also #2773/#2772 and #2778/#2779). The tell each time was the same and I keep noticing it too late: a conflict in a file I did not expect to be contested.

Worth keeping from this branch's investigation, since the reasoning is not obvious and the next person may reach for one of these:

- **Comparing managed thread IDs is unsound in the negative direction** unless the signaller is off-pool — the pool may hand a queued continuation the very thread that just finished signalling. That is what #2791 fixes by construction.
- **"The waiter had completed by the time `OnCompleted` returned" is also unsound.** I tried it as the replacement and measured it: it reported *inline* on **5 of 5** runs against `ObserveCompletion`, because the pool can and does run a queued continuation inside that window.
- The rigorous direct form, if a future change ever removes the dedicated producer thread, is **same thread AND still inside the signalling call** — a thread cannot be in two places at once, so that combination can only mean the continuation ran on the signaller's stack.

The file's two positive controls (Rx's bridge, and awaiting the observable directly) remain the thing that keeps any of these honest: if the measurement stopped discriminating, they would go green when they must be red.

PR #2800, opened 2026-08-30
fix: four silent defects from the issue sweep (#2729, #2731, #2793, #2797)

### LINKED ISSUE #2729: memex: Admin/_GraphSubscription/inbox 'stayed an untyped JsonElement' at boot — the inbound-mail subscription state is unreadable on the portal hub (Email__InboundEnabled=true)
Seen on memex (memex.systemorph.com) right after today's roll to the record-driven Email section (helm revision 28, then the env-patch removal; pod `memex-portal-deployment-56699545bf-jzbvt`, image 3.0.0-rc8.ci.6576), first lines of the portal log:

```
MeshNodeStreamCache.GetStream: Content for Admin/_GraphSubscription/inbox stayed an untyped JsonElement after deserialization (TypeRegistry lacks the $type discriminator) — downstream 'Content is X'/'as X' consumers …
```

`Email__InboundEnabled=true`, `Email__WebhookBaseUrl=https://memex.systemorph.com` — so the inbound Graph change-notification lane is on for this install, and the node that carries its subscription state (`GraphSubscriptionState`, `src/MeshWeaver.Mesh.Contract/GraphSubscriptionState.cs`) is read on a hub whose TypeRegistry does not know its `$type`. Per AGENTS.md that warning means the reader gets a silent null: the subscription renewal / inbox drain cannot see its own state, so inbound mail on memex is dead without an error.

Fix per the `.As<T>()` rule: register the type where the read happens (`WithType(typeof(GraphSubscriptionState), …)` on the hub that owns `Admin/_GraphSubscription`), or move the read to the owning hub — then a test that reads the node back typed on that hub. Outbound is unaffected (the guard passed and the section is complete). Related: #2656 (inbound lane), Plugins#873.

### LINKED ISSUE #2731: Self-update rolls under Admin/UpdatePolicy = None: the policy source is seeded with the platform DEFAULT before the persisted policy arrives, and every (re)subscription runs one check under it (memex-cloud rolled ci.6576 at 07:04:57 with policy None)
## Measured (memex-cloud, 2026-08-30)

`Admin/UpdatePolicy` has carried `policy: "None"` since the maintainer set it on 2026-08-28 10:51:56 (v1704; v1702 had no `policy` field = default). Today:

| when (UTC) | source | what |
|---|---|---|
| 06:48–07:04 | helm rev 20 upgrade | new pod wedged in boot-time modules GC (#2684); helm timed out, rev 20 `failed` |
| **07:04:57.266** | `Admin/UpdatePolicy` **v1852** | `lastCheckTrigger: PolicyChange`, verdict **"updates are disabled on this install (Admin/UpdatePolicy = None); the registry was not listed."** |
| **07:04:57.419** | Deployment annotation `meshweaver.io/self-update-rolled-at` | **rolled to `3.0.0-rc8.ci.6576`** (new ReplicaSet `554545c8cc`, rev 709) |
| 07:06:07 | v1854 | `SafetyNet` → "disabled (None)" |
| 07:06:54 | v1855/1856 | `PolicyChange` → "disabled (None)" |
| 07:08:16 | v1863 (the NEW ci.6576 pod) | `Startup` → "6576 is available but this install rolled 00:03:18 ago, inside the 01:00:00 floor — deferring" |

No version records a "rolling to ci.6576" verdict, yet the roll happened 150 ms after a check that concluded *disabled*, on an install whose persisted policy has been `None` for two days. The last line shows the new pod's `Startup` check ALSO evaluated a non-None policy (a `None` policy says "disabled", never "inside the floor").

## Mechanism (from `memex/Memex.Portal.Shared/SelfUpdate/SelfUpdateHostedService.cs`)

```csharp
// ~281
.StartWith(new UpdatePolicyContent { Policy = _options.DefaultPolicy }))   // Continuous
.DistinctUntilChanged(c => (c.Policy, c.RequireCiGreen));               // re-switch only on a REAL policy change
```
The policy source is SEEDED with the platform default and the persisted node value arrives afterwards as a `PolicyChange`. Every subscription — startup, and any re-subscription of the node stream (a hub recycle, a stream re-attach; here, while the wedged rev-20 pod was torn down) — therefore runs one check under **Continuous** before it has read `None`. That first check is the one that rolls; the `PolicyChange` check that follows dutifully records "disabled". Also visible in the log noise earlier today: `OWNER_NACK_REENQUEUE … target=Admin/UpdatePolicy code=Conflict … rebased on state newer than 1848` — two writers (the seed-check and the change-check) racing to record verdicts on the same node.

## Fix
- The seed must not be a *policy*: no check runs until the persisted `UpdatePolicyContent` has been read (wait for the first real emission; treat "node absent" as the explicit default, not "not yet read" as Continuous). A `Replay(1)` of the node stream without a `StartWith` seed, and the check pipeline gated on it.
- The verdict recorder must not race itself: one writer per install.
- A test: policy `None` persisted, then a re-subscription of the policy source (simulate the stream re-attach) — assert zero roll attempts and a single "disabled" verdict.

Consequence today: the roll was benign (it unstuck the failed helm upgrade by scaling the rev-20 ReplicaSet away), but `UpdatePolicy = None` is not a freeze. Related: #2553 (silent updater), #2684 (the wedge that triggered the re-subscription), #2494.

### LINKED ISSUE #2793: StreamAttachTransientRetryTest reads attempts=1 on a loaded shard — the settle-by-silence poll (25 ms) returns before retry 2 is scheduled
## What failed

`MeshWeaver.Hosting.Orleans.Test.StreamAttachTransientRetryTest.TransientAttachFailure_IsRetried_UpToTheBudget` on **main** at `f52ff52e3` — run [33318685970](https://github.com/Systemorph/MeshWeaver/actions/runs/33318685970), shard 3 (`MeshWeaver.Hosting.Orleans.Test exit=1 TESTFAIL`, elapsed 63 s):

```
Expected value to be 6 because a grain-directory rejection is TRANSIENT … but found 1.
   at StreamAttachTransientRetryTest.cs(159,0)
```

The merge that produced `f52ff52e3` is #2791 (a `Messaging.Hub.Test` change; cannot reach Orleans). The two preceding main runs (`fdfbbed78`, `d46f08d2e`) did not fail this test — this is intermittent, and "found 1" is exactly the value the test documents as THE regression signature (#2633), so a false positive here would be read as the retry latch coming back.

## Why it reads 1 on a loaded shard — the measurement, not the retry

`AttachAttemptsAsync` polls the fake provider's attempt count and returns as soon as **two consecutive readings 25 ms apart are equal** (`test/MeshWeaver.Hosting.Orleans.Test/StreamAttachTransientRetryTest.cs`, the `Task.Delay(25)` loop):

```csharp
for (var i = 0; i < 200; i++)
{
    await Task.Delay(25);
    var now = provider.GetStreamCalls;
    if (now > 0 && now == last) return now;   // "settled"
    last = now;
}
```

The production retry (`OrleansRoutingService`, `RetryWhen(errors => … Observable.Timer(wait, Scheduler.Default))`, with `AttachBackoff = _ => TimeSpan.Zero` in the test) hops through the thread-pool scheduler between attempts. A zero-delay timer still needs a pool thread; on a CI shard running the Orleans suite that hop can exceed 25 ms. Then the loop reads `1`, sleeps 25 ms, reads `1` again — before attempt 2 has been scheduled — and returns `1`. The retry is intact; the poll declared "settled" on a pause.

"Unchanged for 25 ms" is not the condition under test. The condition is **"the attach gave up"** — the retry budget exhausted (`SubscribeAttachRetries + 1` attempts made and the terminal failure surfaced). That is the positive signal to wait on (a stream/observable from the routing service, or the fake provider completing a promise on its *last* permitted call), with a `Timeout` bounding a regression — per the testing rule: never `Task.Delay` to wait for propagation, wait on the actual condition.

## Suggested fix

Expose the terminal outcome of the detached attach (the same seam the readiness-gate test probes) and await THAT; count attempts after it, not before. Any settle-by-silence poll will produce this failure again under load, and at exactly the value that looks like #2633 returning.

### LINKED ISSUE #2797: Self-update: a pinned (UpdatePolicy=None) install rolls on every pod restart — the Startup pass runs under the configured DEFAULT policy before the persisted policy arrives
## What I saw on memex.meshweaver.cloud (2026-08-30)

`Admin/UpdatePolicy` has read `policy: None` all day (every version from v1896 @ 07:08Z through v2088 @ 15:19Z). Yet the install **rolled twice** — to `be39fc607` (ci.6664) at ~11:54Z and to `41dd85a24` (ci.6739) at ~15:03Z (ready 15:57Z) — with `/api/version` proving both.

The node's own history explains it. Every pass whose `lastCheckTrigger` is `Startup` evaluates a candidate as if the policy were Continuous, and only the 1 h floor decides whether it rolls:

| version | time (Z) | trigger | verdict |
|---|---|---|---|
| 1898 | 08:14 | Startup | `check FAILED: HttpRequestException: PATCH memex-migration-deployment → …ci.6585 failed: 404 Not Found` — **it was rolling** |
| 1899 | 08:15 | PolicyChange | `updates are disabled on this install (Admin/UpdatePolicy = None)` |
| 2000 | 11:55 | Startup | `ci.6664 is available but this install rolled 00:01:25 ago, inside the 01:00:00 floor — deferring` |
| 2036/2039 | 13:30 | Startup | `ci.6691 is available but this install rolled 00:03:08 ago … deferring` |
| 2040 | 13:33 | PolicyChange | `updates are disabled …` |

`None` → "Startup evaluates a candidate" → `PolicyChange` back to disabled, on every pod start.

## Root cause

`SelfUpdateHostedService.CreatePolicySource()`:

```csharp
.SelectMany(_ => Observable
    .Defer(ReadPolicyStream)
    .RetryWhen(ResubscribeAfterRetryInterval("policy stream"))
    .StartWith(new UpdatePolicyContent { Policy = _options.DefaultPolicy }))   // Continuous
.DistinctUntilChanged(c => (c.Policy, c.RequireCiGreen));
```

The `StartWith` seeds the pipeline with the **configured default (Continuous)** before the persisted node's value arrives, and the Switch downstream runs a full `RunOnce` on that seed — ACR listing, candidate selection, and the Deployment PATCH — as the *Startup* pass. The persisted `None` arrives a second later as a "PolicyChange" and disables the updater — after the roll has already been issued.

So on a pinned (`None`) install, **every pod (re)start that happens more than `MinRollInterval` after the last roll rolls the portal to the newest rollable tag.** A restart for any reason (OOM, node drain, a probe failure, a `Restart` InstanceAction) becomes an unrequested upgrade. Pinning is not a pin. This is the exact evidence-confusion #2553 was about, with the opposite sign: an install the admin pinned advances, and the node says it never does.

## Second finding, same log line

v1898's failed roll tried to `PATCH memex-migration-deployment`, which no longer exists — the migration is a run-once Job the chart runs on `helm upgrade` (#1788). The portal Deployment patch had already been issued by then, so the roll went ahead with **no migration for the new image**, which is exactly the state `DbVersionGate` exists to refuse. The updater's roll step still assumes the pre-#1788 topology.

## Fix

- Do not seed the policy pipeline with the default. Read the persisted node first (`EnsureExists` already guarantees it exists — the seed was written with `DefaultPolicy` at creation, so nothing is lost) and run the first pass only on the value actually read. If the read must fall back, fall back to `None` (fail closed), never to Continuous.
- Make the roll step match the chart: no `memex-migration-deployment` patch; run the migration the way the chart does, or refuse to roll when it cannot.
- A guard test: a `None` node + a fresh service start must produce ZERO ACR calls and zero patches (`SelfUpdateVerdict.UpdatesDisabled()` as the Startup verdict, not a candidate evaluation).

rbuergi:
Two silent production defects from the issue sweep, both of the shape *"the code acted on a value nobody wrote"*.

## 1. A pinned install rolled on every pod restart — #2731, #2797

`CreatePolicySource` prepended the deployment's **configured default** to the policy stream before the persisted `Admin/UpdatePolicy` value arrived, and the startup pass ran a full evaluation on that seed: registry listing, candidate selection, Deployment PATCH.

memex-cloud has read `policy: None` since 2026-08-28 and **still rolled twice on 08-30** (`ci.6664` 11:54Z, `ci.6739` 15:03Z). The node's own history records the sequence — a `Startup` verdict that evaluated a candidate, then seconds later a `PolicyChange` verdict reading *"updates are disabled on this install"*, after the roll had been issued.

The seed is deleted. Stage 1 (`EnsureExists`) already guarantees the node exists, so the live read is the first emission on its own. `ReadPolicyStream` also drops a null node instead of parsing it — `UpdatePolicyContent.Policy` defaults to `Continuous` and `Continuous` is enum `0`, so an absent node parsed to exactly the verdict a pinned install must never reach.

**Pinned by a test that asserts on the registry listing, not a log-line spelling**: the `None` branch of `RunOnce` returns its verdict without listing, so `acr.Calls == 0` proves which policy decided. Mutation-checked — restoring the seed fails it with `but found 1`.

`ABuildCompletionDrivenCheck_NeverWarnsAboutTheEventChannel` published its build before the watch was subscribed, so the publication was absorbed as the watch's **baseline**; it passed only because the synthetic seed made that subscription fast. It now waits for the startup check line first — the same positive signal the resilience test already documents.

## 2. Node content read as raw JSON on any hub but its own — #2729

A node type's `WithContentType<T>()` configures its own hub. **Seven** built-in types declared one and registered its `$type` discriminator nowhere else, so a reader elsewhere in the mesh got an untyped `JsonElement` and a **silent null** — no exception, no log; the value renders empty and reactive waits time out.

memex logged it at boot for `Admin/_GraphSubscription/inbox`. That node carries the inbound-mail subscription state, so the renewal could not see its own record and **inbound mail was dead** on an install whose every signal was green.

Registered: `EaCredential`, `GraphSubscriptionState`, `Invitation`, `MemexClientContent`, `NotificationChannel`, `NotificationRule`, `TeamsConversation`.

`NodeTypeContentDiscriminatorGuard` keeps them registered. The two halves of this contract are joined by nothing, and omitting one compiles, boots, serves and passes every test. The guard reads both halves out of the source, carries a floor so it cannot pass on a broken scan, and self-tests its own classifier. Mutation-checked — deleting one registration fails it by name.

## Verification

```
dotnet build src/MeshWeaver.Graph            -c Release -warnaserror   0 Error(s)
dotnet build memex/Memex.Portal.Shared       -c Release -warnaserror   0 Error(s)
dotnet build test/MeshWeaver.Documentation.Test  -c Release -warnaserror   0 Error(s)
dotnet build test/MeshWeaver.Hosting.Monolith.Test -c Release -warnaserror 0 Error(s)

dotnet test test/MeshWeaver.Hosting.Monolith.Test --filter SelfUpdate        29 passed, 0 failed
dotnet test test/MeshWeaver.Documentation.Test  --filter <the three guards>   3 passed, 0 failed
```

Fixes #2729
Fixes #2731
Fixes #2797

https://claude.ai/code/session_014FhP7Jg65q8cAATnFfeaTK

---

## 3. A retry test read a scheduler pause as the regression — #2793

`StreamAttachTransientRetryTest` polled the attempt counter and treated two equal readings 25 ms apart as *settled*. The attach retry hops through the thread-pool scheduler between attempts, so on a loaded shard that hop exceeds 25 ms — the poll reads mid-hop and returns **1**, which is exactly the #2633 regression signature the test documents. Observed on main at `f52ff52e3` (run 33318685970, shard 3), on a merge that cannot reach Orleans.

`RegisterStream` already records the real condition: `subscriptionReady[address]` completes when the attach has attached, given up, or been cancelled, and never faults. Exposed as the internal `AttachSettled(address)` seam (same `InternalsVisibleTo` discipline as `AttachBackoff`), awaited with a 30 s **backstop** — real elapsed time is milliseconds. Captured before the readiness gate opens so the attach cannot finish unobserved, and a missing seam throws rather than silently falling back to a poll.

```
dotnet build test/MeshWeaver.Hosting.Orleans.Test -c Release -warnaserror   0 Error(s)
dotnet test --filter StreamAttachTransientRetryTest    4 passed, 0 failed, 39 ms
```

Fixes #2793

### DIFF
--- a/.github/scripts/shard-assign.sh
+++ b/.github/scripts/shard-assign.sh
@@ -42,14 +42,38 @@
 # drifted the same way (Monolith 200→345, Data 25→73, PluginTester 60→8).
 #
 # The floor for ANY sharding is the heaviest single SCHEDULABLE UNIT — a project,
-# or one part of a split one. Today that is Hosting.Monolith.Test's two halves at
-# ~331 s each. Adding shards below that buys nothing.
+# or one part of a split one. Today that is Hosting.Monolith.Test's three thirds
+# at ~221 s each. Adding shards below that buys nothing.
 #
-# SPLIT RULE: split a project only when its solo weight exceeds the ideal shard
-# load (sum ÷ SHARD_TOTAL, currently 2526 ÷ 6 ≈ 421 s), because only then is it
-# the binding floor. Monolith (662 s) qualifies; nothing else does. (AI.Test's
-# split was dropped for the same reason before the suite itself moved to
-# MeshWeaver.Plugins beside the engine, #2276.)
+# SPLIT RULE — TWO independent triggers. Either one is sufficient.
+#
+#  (1) BALANCE. Solo weight exceeds the ideal shard load (sum ÷ SHARD_TOTAL,
+#      currently 2526 ÷ 6 ≈ 421 s), because only then is the project the binding
+#      floor. Monolith (663 s) qualifies. (AI.Test's split was dropped under this
+#      rule before the suite itself moved to MeshWeaver.Plugins, #2276.)
+#
+#  (2) 🚨 HEADROOM AGAINST THE PER-PROJECT CAP (#2747). Solo weight exceeds ~60%
+#      of the `timeout 8m` = 480 s wall-clock cap each project runs under in
+#      dotnet-test.yml. This trigger is INDEPENDENT of balance and it is the one
+#      that bites in production: a project can sit comfortably under the ideal
+#      shard load and still be one slow runner away from exit=124.
+#
+#      PluginCatalog was exactly that. Measured across four consecutive runs it
+#      ran 315/320/315 s — 66-72% of the cap — and then 480 s exit=124 TIMEOUT on
+#      e12697ebd (run 33302269352, shard 0), with the SAME tree passing at 320 s
+#      on the very next run. No change was responsible. The five "test failures"
+#      that came with the kill were timing casualties of it, not defects, so every
+#      occurrence costs a full investigation before it can be dismissed — and this
+#      repo's rules (correctly) forbid simply re-running it.
+#
+#      Raising the cap is NOT the alternative: the cap is what turns a wedge into a
+#      bounded, attributable failure instead of a 20-minute shard. Splitting lowers
+#      the numerator instead — PluginCatalog's two parts run ~174 s each, ~36% of
+#      the cap — and it costs nothing, because the parts land on different runners.
+#
+# 🚨 The two triggers must BOTH be checked when re-measuring. A project that grows
+# past 288 s (60% of 480) needs splitting even while the LPT loop still reports
+# six balanced shards, which is precisely why the balance rule alone missed this.
 #
 # Unlisted projects get DEFAULT_WEIGHT — a deliberate over-estimate, since an
 # unlisted project is a NEW one whose cost nobody has measured yet.
@@ -86,8 +110,8 @@ fi
 
 # "<seconds> <project-name>", heaviest first.
 WEIGHTS=$(cat <<'EOF'
-663 MeshWeaver.Hosting.Monolith.Test 2
-348 MeshWeaver.PluginCatalog.Test
+663 MeshWeaver.Hosting.Monolith.Test 3
+348 MeshWeaver.PluginCatalog.Test 2
 269 MeshWeaver.Hosting.Orleans.Test
 91 MeshWeaver.Data.Test
 76 MeshWeaver.Messaging.Hub.Test
--- a/memex/Memex.Portal.Shared/SelfUpdate/SelfUpdateHostedService.cs
+++ b/memex/Memex.Portal.Shared/SelfUpdate/SelfUpdateHostedService.cs
@@ -258,11 +258,16 @@ public class SelfUpdateHostedService : IHostedService
     /// <list type="number">
     /// <item>Seed <c>Admin/UpdatePolicy</c> if absent (storm-safe, via a query — never a point-read
     /// of a maybe-absent node); a seeding fault retries at the polling cadence.</item>
-    /// <item>The live node stream. The default policy is prepended exactly ONCE — after the seed
-    /// (so the first tick can never point-write a not-yet-existing node), before the first live
-    /// emission. A stream fault retries INSIDE the StartWith, so a retry re-establishes the read
-    /// SILENTLY: it never re-emits the default, and therefore can never flip a Stable/None install
-    /// back to default-policy polling (Copilot review on #611).</item>
+    /// <item>The live node stream, and NOTHING ELSE. 🚨 The configured default is never emitted
+    /// here (MeshWeaver#2731/#2797): it used to be prepended once, before the first live emission,
+    /// and that one synthetic value drove a full <see cref="RunOnce"/> — ACR listing, candidate
+    /// selection and the Deployment PATCH — under a policy the install had never set. On a pinned
+    /// (<c>None</c>) install, EVERY pod (re)start therefore rolled the portal to the newest tag,
+    /// with the persisted <c>None</c> arriving a second later as a <c>PolicyChange</c> that
+    /// dutifully logged "updates are disabled" AFTER the roll had been issued. Stage 1 guarantees
+    /// the node EXISTS, so the live read is the first emission on its own — the poller now waits
+    /// for a policy it has actually read instead of guessing one. A stream fault retries inside
+    /// the RetryWhen, so a retry re-establishes the read silently (Copilot review on #611).</item>
     /// </list>
     /// Retries are delayed, Rx-composed resubscribes at the polling cadence — not a hot retry loop,
     /// not a watchdog. <c>DistinctUntilChanged</c> sits outermost so only a REAL policy change (or
@@ -277,8 +282,7 @@ public class SelfUpdateHostedService : IHostedService
             .Take(1)
             .SelectMany(_ => Observable
                 .Defer(ReadPolicyStream)
-                .RetryWhen(ResubscribeAfterRetryInterval("policy stream"))
-                .StartWith(new UpdatePolicyContent { Policy = _options.DefaultPolicy }))
+                .RetryWhen(ResubscribeAfterRetryInterval("policy stream")))
             .DistinctUntilChanged(c => (c.Policy, c.RequireCiGreen)); // <-- re-switch only on a REAL policy change
     }
 
@@ -292,6 +296,12 @@ public class SelfUpdateHostedService : IHostedService
         var workspace = _hub.GetWorkspace();
         var jsonOptions = _hub.JsonSerializerOptions;
         return workspace.GetMeshNodeStream(UpdatePolicyNodeType.NodePath)
+            // 🚨 An ABSENT node is not a policy (MeshWeaver#2731/#2797). UpdatePolicyContent.Policy
+            // defaults to Continuous and UpdatePolicyKind.Continuous is enum 0, so parsing a null
+            // node yields "roll to the newest tag" — the exact verdict a pinned install must never
+            // reach. Stage 1 has already guaranteed the node exists, so a null here is the stream
+            // not having loaded it yet: wait for the real one rather than deciding without it.
+            .Where(node => node is not null)
             .Select(node => UpdatePolicyNodeType.Parse(node, jsonOptions));
     }
 
--- a/src/MeshWeaver.Connection.Orleans/OrleansRoutingService.cs
+++ b/src/MeshWeaver.Connection.Orleans/OrleansRoutingService.cs
@@ -5,6 +5,7 @@ using System.Reactive;
 using System.Reactive.Concurrency;
 using System.Reactive.Disposables;
 using System.Reactive.Linq;
+using System.Reactive.Subjects;
 using System.Reactive.Threading.Tasks;  // Task<T>.ToObservable() — the SAFE direction; nothing here bridges the other way.
 using MeshWeaver.Mesh;
 using MeshWeaver.Mesh.Services;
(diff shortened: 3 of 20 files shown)


rbuergi:
## Folded in a fifth defect: the OTHER settle-by-silence poll (`555679425`)

This PR's own merge-queue entry ejected at **16:48Z** — not on anything in its diff, but on the sibling of the very test it fixes:

```
MeshWeaver.Hosting.Orleans.Test.PodHubClaimLifetimeTest
  .AClaimThatCannotLand_WhereNoGrainCanBeHosted_StopsAtTheInitialBudget_AtDebug [FAIL]
  Expected value to be 6 … but found 1
```

Same shape, same *value*, same cause as #2793: `WaitUntilSettled` polled `factory.AttachCalls` every 25 ms and called two equal readings "settled", while the claim's `RetryWhen` hops through the thread-pool scheduler between attempts. Its own doc comment said the quiet part out loud — *"there is no completion signal to await for a claim that ends"*.

**There is one now.** `OrleansRoutingService.PodHubClaimSettled(address)` is the sibling of this PR's `AttachSettled`: a per-address `AsyncSubject<Unit>`, an **instance** field (never static), armed *before* the claim is subscribed so it cannot terminate unobserved, completed in **both** terminal arms — it landed, or it hit the terminal that is impossibility rather than a budget. A claim that is still retrying completes nothing, which is the honest answer for a lifetime with no give-up; that test keeps its existing positive `WaitForAttempts`.

### Why it had to go in *this* PR, not its own

The queue builds each entry on the entries ahead of it, so the two fixes were ejecting each other:

| entry | failed on | fixed by |
|---|---|---|
| `pr-2800` (16:48Z) | `PodHubClaimLifetimeTest` | only this commit |
| `pr-2803` (17:00Z) | `StreamAttachTransientRetryTest` | only this PR |

Neither sibling could land alone. One entry, both fixes.

### Verified

- **5/5 green in 1 s** (the poll took seconds per test).
- **Mutation-proved, not merely observed green:** collapsing the budget (`attempt >= PodHubAttachRetries` → `attempt >= 0`) makes the test **FAIL** with *found 1* — the new wait still detects the regression it exists for. Source restored and re-verified afterwards.
- The What's New entry now covers both siblings rather than one.

I dequeued this PR to push (GitHub declines a push to a branch with a live queue entry) — **please re-queue it**; it is `CLEAN` and the required check was green on the previous head.

Issue #2802, opened 2026-08-30
Hosting: RotateRegistryKey — instance keys rotated through the Hosting plugin, never by hand (first verb of the record-driven deployment process)


rbuergi:
Tracking issue for the maintainer directive of 2026-08-30: **a clean deployment process, administered through the Hosting plugin, with instance records in `Systemorph/Memex`** — starting with the verb the day's incident demands, registry-key rotation.

## Why now

Two live `mwi_` instance keys (`memex`, `memex-cloud`) need rotating. Today that cannot be done cleanly: `MeshWeaverInstanceService.ReissueKey` exists and has **no caller** (no endpoint, no UI, no action), the key is stored as a **literal** in memex's helm values and as a **cluster-only patch** on memex-cloud (`helm get values` / `helm get manifest`: 0, live deployment: 1 — a `helm upgrade` would drop it), and the operator's `hosting-kv-ensure` marks `PluginCatalog-RegistryToken` as `REQUIRED` = *"supplied out-of-band"*. Rotation by hand would be a workaround; the verb is the fix.

## Measured state (2026-08-30)

| fact | value |
|---|---|
| Deployment records | committed in `Systemorph/Memex` `mesh/Deployments/{memex,memex-cloud,pearl,atioz}.json` (Memex #146, core #2686, Plugins #918 — all merged) |
| …in the registry mesh | **zero** — `nodeType:Hosting/Deployment` returns 0, no GitSync node references `Systemorph/Memex`; the Hosting plugin cannot act on records it cannot see |
| record secret model | both records still use the LEGACY `secretMounts` hand-made SPC; `keyVaultSecrets` (typed) is `null`/empty, so the registry token is declared nowhere |
| vault `Systemorph` | access-policy model (`enableRbacAuthorization=false`); the operator identity's *Key Vault Secrets Officer* role assignment was **inert**; policy `get/set/list` granted today to `hosting-operator` (oid `bd994331…`) and read back |
| operator ↔ Azure | Jobs run as SA `hosting-operator` with workload identity (`azure.workload.identity/use`), image `azure-cli:2.77` |
| operator ↔ mesh | one-way: the mesh reads `::hosting::` marker lines out of the pod log (`OperatorOutput`, the audit line); the Job holds NO mesh credential |

## The verb: `InstanceActionKind.RotateRegistryKey`

Designed so the raw key **never leaves the operator job**, and only its hash crosses back — the same channel the backup verifier and the audit already use.

[...]

PR #2803, opened 2026-08-30
feat(hosting): instance keys rotate through the operator — AdoptKeyHash + hosting-kv-rotate (#2802)

rbuergi:
The **platform half** of #2802. The Hosting plugin's `RotateRegistryKey` verb follows in `MeshWeaver.Plugins` once this is on main (cross-repo halves land platform first).

## Why

Two live `mwi_` instance keys need rotating. `MeshWeaverInstanceService.ReissueKey` exists with **no caller**, and the only way to land a new value in a running portal was a terminal that would have to see the key — which is how today's exposure happened in the first place (an audit selected `.value`).

## The design: the raw key exists in exactly two places, ever

| where | what it holds |
|---|---|
| the operator job's memory | the raw key, until `unset` |
| Key Vault | the raw key |
| the job log / the mesh / this repo | **the hash only** |

**`hosting-kv-rotate`** — the one operator script allowed to overwrite a vault secret (`hosting-kv-ensure` never will: the master key it guards seals stored credentials, which is why this is a separate verb and `kv-ensure` must never grow a `--force`). Mints `mwi_` + base64url(32 random bytes) — the `InstanceKeys` scheme exactly — sets it in the vault from a variable that is never echoed, discards it, reports `::hosting:: key_hash=<sha256>`, then **waits for the synced Kubernetes Secret to carry the new value, compared by hash** (DeploymentAKS → *"KeyVault CSI env timing"*: a restart that reads the old synced Secret puts the *old* key into the new pods) and refuses with a message naming the SPC object if it never does. Dry-run aware; shellcheck clean.

**`IInstanceKeyRegistry`** (contract assembly) + **`MeshWeaverInstanceService.AdoptKeyHash`** — the registry-side half: the instance takes the new hash and issued-at, a fresh index entry points at it, and the **previous index entry is deleted** so the old key stops authenticating the moment adoption lands (`ReissueKey` left that "to be cleaned up separately"; a lifecycle verb must not). Refuses anything that is not 64-char lowercase hex — a raw key arriving here by mistake is precisely what must never happen. On the contract assembly so the Hosting plugin's in-mesh control plane can resolve it from the hub without referencing the portal host.

## Verified

`InstanceKeyAdoptionTest` — **2/2 on the real Monolith mesh**: register → adopt *by instance id through the contract* → new hash + later `KeyIssuedAt` on the node, new index entry resolves to the instance, **old index entry gone**; plus the shape refusal. Index nodes are read as System through `GetMeshNode`, the way the registry's own tests read fresh nodes (the first draft point-read them and hit the absent-node NotFound — fixed by reading the way the house rules say, not by widening a wait).

## Already done outside the repo

The `hosting-operator` identity had a *Key Vault Secrets Officer* role assignment that was **inert** — the `Systemorph` vault uses access policies (`enableRbacAuthorization=false`). It now holds `get/set/list` as a policy, verified by read-back. Recording it here because it is the precondition the script depends on and it lives nowhere else yet.

### DIFF
--- a/deploy/aks/operator/bin/hosting-kv-rotate
+++ b/deploy/aks/operator/bin/hosting-kv-rotate
@@ -1,0 +1,88 @@
+#!/usr/bin/env bash
+# hosting-kv-rotate --vault <name> --prefix <prefix> --namespace <ns> --synced-secret <k8s secret> [--key <config key>]
+#
+# Rotate this instance's plugin-registry key: mint a fresh `mwi_` key, put it in Key Vault, and
+# report ONLY ITS HASH back to the mesh. The registry adopts the hash (the control plane reacts to
+# the `::hosting:: key_hash=` line below); the restart that follows in the plan puts the new value
+# into the pods. The raw key exists in exactly two places, ever: this process's memory and the vault.
+#
+# 🚨 IT NEVER PRINTS THE KEY. Not on success, not on failure, not in a `set -x` (deliberately not
+# enabled), not in a `::hosting::` fact. The 2026-08-30 exposure this verb exists to repair came
+# from a CHECK that selected `.value` while asking "is this stored in the clear?" — the audit was
+# the leak. Every read below compares hashes and names.
+#
+# 🚨 IT DOES OVERWRITE — the one operator script that does. `hosting-kv-ensure` refuses to touch
+# an existing secret because the master key seals stored credentials; a registry key seals
+# nothing and is meant to be replaced. That is why this is a separate verb with its own name, and
+# why `kv-ensure` must never grow a --force.
+# shellcheck source=_common.sh
+source "$(dirname "$(readlink -f "${BASH_SOURCE[0]}")")/_common.sh"
+# shellcheck disable=SC2034  # read by hosting::die in _common.sh, which shellcheck does not follow here
+HOSTING_CMD="hosting-kv-rotate"
+vault="" prefix="" namespace="" synced="" key="PluginCatalog__RegistryToken"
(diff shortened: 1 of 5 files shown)


rbuergi:
**Ejected from the merge queue at 17:00Z (`failed_checks`) — not this PR's diff.** Its entry (`gh-readonly-queue/main/pr-2803-afb5a5ec9`) failed shard 3 on:

```
MeshWeaver.Hosting.Orleans.Test.StreamAttachTransientRetryTest.TransientAttachFailure_IsRetried_UpToTheBudget [FAIL]
```

That is **#2793**, and #2800 fixes it. This PR touches `MeshWeaver.Mesh.Contract/Security`, `Memex.Portal.Shared/Authentication` and an operator shell script; it cannot reach an Orleans stream attach.

## Why the queue currently cannot land anything

Both settle-by-silence tests false-RED on a queue-entry runner, and each is unfixed on the branch the other builds on:

| entry | failed on | fixed by |
|---|---|---|
| `pr-2803` (this) | `StreamAttachTransientRetryTest` | #2800 |
| `pr-2800` | `PodHubClaimLifetimeTest.AClaimThatCannotLand_WhereNoGrainCanBeHosted_StopsAtTheInitialBudget_AtDebug` (expected 6, found 1) | *nothing yet* |

The queue builds each entry on the ones ahead of it, so #2800 alone still meets the unfixed pod-hub test and ejects before this one is ever built. **They have to be in the same entry**, so I am folding the pod-hub fix into #2800 rather than opening a second PR that would eject it.

Re-queuing this once #2800 lands. No change needed here.

Issue #2772, opened 2026-08-30
ObserveCompletion cancels the wait but not the source — a cancelled caller keeps its IoPool slot


rbuergi:
## What

`ReactiveCompletion.ObserveCompletion(reportLateFault, ct)` cancels the **wait** and deliberately never disposes the **source** subscription. That is documented and intentional — it is what keeps the error arm attached so a *late* fault is still reported. But it silently changes cancellation semantics for every site converted from Rx's `.ToTask(ct)`, and for pooled work the cost is a retained pool slot.

## Evidence, not inference

`IoPool.InvokeObservable` builds its work as:

```csharp
Observable.Create<T>(async (observer, subscriberCt) =>
{
    …
    using var linked = CancellationTokenSource.CreateLinkedTokenSource(subscriberCt, _poolCts.Token);
```

(`src/MeshWeaver.Mesh.Contract/Threading/IoPool.cs`, ~244-250.)

`subscriberCt` is cancelled by Rx **when the subscription is disposed**. So:

| bridge | on cancellation |
|---|---|
| `.ToTask(ct)` | disposes the subscription → cancels `subscriberCt` → **cancels the pooled operation and releases its slot** |
| `ObserveCompletion(…, ct)` | settles the task only → subscription lives on → **pooled work runs to completion, holding its slot** |

So a caller that has already given up keeps an IoPool slot for the full duration of work nobody is waiting for. Under load that is the shape that turns into a wedge, and it is invisible: nothing faults, nothing logs, the slot is simply not available.

## Why this is a core issue and not a plugins one

The property belongs to the bridge, and the bridge is core's. #2750 applied it to ~1,538 call sites in this repo; the plugins repos have their own. If the semantics are wrong for a caller holding a pooled or request-scoped resource, they are wrong everywhere — so the fix belongs here, not as N local Rx wrappers per repo that would drift from each other and from this one.

Found by the automatic review on `MeshWeaver.Plugins#963` (four threads: `MeshNodeAgentFileStore` and `LocalMeshApiEndpoints` ×2, plus two late-fault-logger findings that were separate bugs and are fixed). Two threads were left **open on purpose** pending this decision — they should not be read as accepted.

## Suggested shape

[...]

PR #2828, opened 2026-08-30
fix(cd): a broken pipe mid-push no longer throws away the whole release (#2810)

rbuergi:
Delivery has been blocked for hours. **Across the last 20 CD runs there is not one non-skipped `Promote` job**, so every self-updating install is still on the previous image. There are two independent causes; this PR fixes the one nobody owns.

## The measurement

Run 33328213655 (target `84bf70b17`):

```
18:48:05  linux-arm64 pushed cleanly
18:52:58  [Registry] Encountered a HttpRequestException Unknown with message "Broken pipe".
          Pausing before retry.
18:53:07  CONTAINER1013 → CONTAINER1001: Failed to upload blob using
          PATCH https://meshweaver.azurecr.io/v2/memex-portal-ai/blobs/uploads/…
```

Same pair on run 33322756606 (target `8cbf69ac9`) at 16:49Z. **The SDK's own retry did not recover either.**

And `ci.6846` promoted successfully at **18:14Z — between the two failures.** So this is **intermittent**: not credentials, not quota, not a purge policy. The registry is Premium SKU and healthy throughout.

Publication is all-or-nothing by design — `promote` applies the real tags only once every leg succeeded, so an install can never pull a half-shipped set. The cost of that guarantee is that **one broken pipe on one architecture throws away the entire release.**

## The change

Up to three attempts on the portal-ai push, with an `az acr login` between them (a token can expire during a multi-arch push whose legs take minutes each).

🚨 **Not a skip-trapdoor**, and the distinction is the whole point:

- a compile error, a missing project or a rejected credential fails **identically on all three attempts** and the step ends **RED**, naming the last attempt's output;
- only a connection dying mid-upload gets a second chance;
- three consecutive failures emit an `::error::` saying it is no longer a transient and wants investigating rather than another retry, pointing at #2810.

It mirrors the idiom already in `node-repo-{compile-check,gate,publish-bake}.yml`, which carry the same *"Bounded retry, and deliberately NOT a skip-trapdoor"* comment for their registry logins — added after three ACR logins failed on 2026-08-28 with the registry healthy throughout.

## Verified

| | |
|---|---|
| YAML parses | `yaml.safe_load` |
| the failure path **exits 1** | **proved, not assumed** — the loop shape simulated with an always-failing command ends `::error::` + `exit 1` |
| no guard's parse broken | `MeshWeaver.Documentation.Test` **123/123**, which is every guard that reads this workflow: `ReleaseDeliveryChainGuard`, `MainRunsAreNeverCancelledGuard`, `PlatformBakeLaneGuard`, `PlatformReleaseNotifyGuard`, `WorkflowJobIfContextGuard` |

## The other blocker, so they are not conflated

A `CS0419` ambiguous cref in **MeshWeaver.Plugins** (`MeshStoreAccess.cs:72` and two more) — core CD builds the portal hosts from `plugins-repo`, and #2772's new `ObserveCompletion` overload made three bare `<see cref>`s ambiguous there. Fixed by MeshWeaver.Plugins#993.

**Different cause, different commits:** both ACR failures are on commits that **predate** the overload. If `Promote` is still skipped after both land, read *which* error the portal-ai job carries — `CS0419` means the cref fix did not take; `CONTAINER1001` means the connection died three times running and this is no longer transient.

### DIFF
--- a/.github/workflows/main-cd.yml
+++ b/.github/workflows/main-cd.yml
@@ -848,16 +848,59 @@ jobs:
       # that IS what pushing costs) but under a name no consumer can select, so this leg succeeding
       # ships NOTHING on its own. `promote` applies version/sha/main once every leg has succeeded.
       - name: Build + push portal-ai (staging tag)
-        run: >
-          dotnet publish plugins-repo/src/Memex.Portal.Distributed/Memex.Portal.Distributed.csproj
-          -c Release --no-self-contained -t:PublishContainer -p:PublishProfile=
-          -p:ContainerRuntimeIdentifiers='"linux-x64;linux-arm64"'
-          -p:CIRun=true
-          -p:MeshWeaverRoot=${{ github.workspace }}
-          -p:ContainerRegistry=${{ env.ACR }}
-          -p:ContainerRepository=memex-portal-ai
-          -p:ContainerImageTags=${{ needs.gate.outputs.staging_tag }}
-          -p:ContainerBaseImage=${{ env.ACR }}/memex-portal-ai-base:latest
+        run: |
+          # 🚨 Bounded retry, and deliberately NOT a skip-trapdoor: after the last attempt this
+          # still fails RED. It exists because the BLOB UPLOAD is the failure, not the build.
+          #
+          # Measured 2026-08-30, run 33328213655 (target 84bf70b17): the linux-arm64 leg pushed
+          # cleanly at 18:48:05, then at 18:52:58 the x64 leg logged
+          #     [Registry] Encountered a HttpRequestException Unknown with message "Broken pipe".
+          #     Pausing before retry.
+          # and nine seconds later
+          #     CONTAINER1013: Failed to push to the output registry: CONTAINER1001: Failed to
+          #     upload blob using PATCH https://meshweaver.azurecr.io/v2/memex-portal-ai/blobs/uploads/...
+          # The SDK's own retry did not recover it. Same pair on run 33322756606 (target
+          # 8cbf69ac9) at 16:49Z. Between those two, ci.6846 promoted successfully at 18:14Z — so
+          # the fault is INTERMITTENT, not a credential, quota or purge problem, and the registry
+          # is Premium SKU and healthy throughout.
+          #
+          # Publication is ALL-OR-NOTHING (the promote job tags only when every leg succeeded), so
+          # one broken pipe on one architecture costs the whole delivery: no image, and every
+          # self-updating install stays on the previous one until a later run happens to get a
+          # clean connection. That is the cost this retry removes.
+          #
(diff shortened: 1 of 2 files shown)


rbuergi:
## 🚨 Correcting my own justification: `CONTAINER1013` is a WRAPPER, and the two failures are not the same

I grouped these as *"the same `CONTAINER1013/1001` pair, twice"*. **That grouping is wrong**, and I verified it against the logs rather than accepting the correction:

| run | commit | image | **inner** cause |
|---|---|---|---|
| 33322756606 | `8cbf69a` 16:35 | **migration** | `Connection refused (meshweaver.azurecr.io)` — **no `CONTAINER1001` at all** |
| 33328213655 | `84bf70b` 18:53 | **portal-ai** | `CONTAINER1001 … PATCH … 'RequestedRangeNotSatisfiable'` (HTTP **416**) |

`CONTAINER1013` is the SDK's outer *"failed to push"* wrapper; it wraps different inner causes. I compared wrappers. That is the same *same-symptom-⇒-same-cause* shortcut this repo has paid for three times today — and the countermeasure took thirty seconds: **open the error string per run.**

## What that means for this PR — it is narrower than I claimed

**Still exactly right for the connection-refused class.** Nothing was uploaded; the connection never opened. A bounded retry with a re-login is the correct answer, and it is the same fault that put the retry idiom into `node-repo-*.yml` in the first place.

**Unproven for the 416, and I will not claim otherwise.** The upload session's `_state` token decodes to `{"Offset":0,"StartedAt":"…18:47:50Z"}` and the PATCH was rejected at 18:53:07 — **5m17s in, still at offset 0**. A 416 on `blobs/uploads/` means the registry's offset no longer agrees with the client's. So:

- this retry does give it a **fresh upload session** each attempt, which the SDK's in-place resume does not — that is a real improvement over dying on an invalidated session;
- but if the mechanism is *a single long-lived upload of a very large layer outliving the registry's session window*, then attempts 2 and 3 take the same five minutes and hit the same wall, and this costs 3× the CI time to fail anyway.

The step's final message already says the right thing for that case — three consecutive failures emit *"no longer a transient … wants investigating rather than another retry"* — so the PR degrades honestly. But the **body oversold one cause**, and I have corrected it.

## Keeping it, with the scope stated

I am not withdrawing this: one of the two blockers is squarely in scope, the other gets a materially better shot than the SDK's resume, and the failure mode is loud rather than silent. What it is **not** is a fix for the 416 mechanism — that wants its own investigation (layer size, chunked upload, or ACR's actual session window, none of which is measured), and belongs on #2827 rather than being quietly absorbed here.

*(The wrapper-vs-inner-cause finding and the `_state` decode are the reviewing session's; the verification and this correction are mine.)*

PR #2838, opened 2026-08-30
feat(doc): the documentation is world-readable — /Doc links unfurl, docs get indexed

rbuergi:
Requested by the maintainer: mesh links pasted into Teams show no card. The outbound SEO pipeline (SeoHead + `/api/og/{path}.png`) is live and complete — measured today, **all 81 store covers on memex serve full OG cards anonymously** — but `SeoResolver` fails closed through the `AnonymousGate`, so only anonymous-readable pages unfurl. `/Doc` wasn't one: its seeded policy granted Viewer to `Public` (all *authenticated* users) and never carried `PublicRead`.

## Change

`Doc/_Policy` (code-seeded in `DocumentationExtensions`) gains `PublicRead = true` — exactly the shape `PartitionAccessPolicy.PublicRead`'s own docs name Documentation as the intended user of. The docs GitSync from this public repository, so an anonymous reader learns nothing that is not already on github.com. CUD stays false; the `Public→Viewer` grant stays for the signed-in comment/thread surface.

Everything downstream keys off this one bit: the crawler-facing head (`og:*`, JSON-LD), the `/api/og` share card, `<noscript>` content for indexing.

## Also: `Doc/Architecture/LinkPreviews`

A new doc page conserving today's investigation — the two features both called "OG card" (outbound SEO head vs the inbound `OgCard` layout-area package), the anonymous-read rule and why a private page must never unfurl, the `PublicRead` opt-in shape, and how to verify an unfurl with curl instead of pasting into chat. Pairs with MeshWeaver.Plugins (OgCard package description fix + `og:site_name` empty-string fix).

## Tests

`SyncedPartitionReadOnlyTest` gains `Doc_IsAnonymouslyReadable` (root + a page): anonymous holds Read and never C/U/D. Suite: **6/6 passed.**

https://claude.ai/code/session_01QoG4USP5fd8wsxCD7e9q4x

### DIFF
--- a/src/MeshWeaver.Documentation/Data/Architecture/LinkPreviews.md
+++ b/src/MeshWeaver.Documentation/Data/Architecture/LinkPreviews.md
@@ -1,0 +1,104 @@
+---
+Name: Link Previews
+Category: Architecture
+Description: How a mesh link unfurls in Teams, Slack or LinkedIn — the crawler-facing SEO head, the generated share card, and the one access bit that decides whether a page unfurls at all. Plus the OTHER feature called "OG card", which points the opposite way.
+Icon: <svg xmlns="http://www.w3.org/2000/svg" width="24" height="24" viewBox="0 0 24 24" fill="none" stroke="currentColor" stroke-width="2" stroke-linecap="round" stroke-linejoin="round"><rect x="2" y="4" width="20" height="14" rx="2"/><path d="M2 9h20"/><circle cx="6" cy="6.5" r=".5"/><path d="M6 13l3 3 5-5 4 4"/></svg>
+---
+
+# Link Previews
+
+Paste a mesh link into Teams, Slack or LinkedIn and one of two things happens: a **card** — title,
+description, picture — or a bare URL. This page explains what decides which, and how to get the
+card. It also disentangles the two features that both answer to the name "OG card" and point in
+opposite directions.
+
+## Two features, two directions
+
+| | Direction | What it does | Where it lives |
+|---|---|---|---|
+| **SEO head + share card** | **outbound** — our pages unfurling *elsewhere* | serves `og:*` meta tags and a share image to crawlers | `Memex.Portal.Shared/Seo` + `SeoHead` in the portal GUI shell |
+| **`OgCard` package** | **inbound** — *other* pages rendering *inside ours* | a layout area drawing link-preview cards for external URLs or mesh nodes, embedded from markdown | the `OgCard` store package (`MeshWeaver.OgCard` module) |
+
+When someone says "the link doesn't show a card in Teams", that is the **outbound** feature —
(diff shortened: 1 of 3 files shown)


rbuergi:
## This red is not yours — and it is a RECURRENCE of a flake the tree believes it fixed

Triaged while draining the merge queue. Shard 2's failure is:

```
##[error] MeshWeaver.Hosting.Monolith.Test.MeshNodeLanguageServiceTest
          .ScriptCompletions_FilterByTypedPrefix_NotJustTheAlphabet
##[error] Shard 2: failing tests in trx — failing the shard.
```

A documentation PR cannot reach the kernel script language service, so this is not attributable to the diff.

### 🚨 Why it is worth more than a re-run

That test is named **verbatim** in `KernelScriptReferences.cs`, as the symptom of the defect #2616/#2769 was written to remove:

> *"It used to be the answer: a `Lazy<Task<ImmutableArray<…>>>` whose value `GetReferencesAsync` returned directly. That FROZE `AppDomain.CurrentDomain.GetAssemblies()` as of the process's FIRST kernel session — and assemblies load LAZILY … Under parallel shard load the lottery is decided by whichever unrelated test ran first, which is why it presented as a flake reproducing on completely unrelated diffs (**`ScriptCompletions_FilterByTypedPrefix_NotJustTheAlphabet`** losing "Mesh")."*

**#2769 is already merged** (`ead524e99`), so the freeze is gone — `GetReferencesAsync` now composes the set fresh from the current assembly list on every call. And the test failed anyway, on 2026-08-30 20:33.

That is informative rather than disappointing, because it narrows the remaining cause. The fix removed the **permanent** freeze; it cannot remove the **transient** one. A fresh composition still only sees *what is loaded at that instant*, so a completion computed before the assembly carrying "Mesh" has lazily loaded still shorts the symbol. Under parallel shard load that window is real.

So the residue is a **startup-ordering race**, not a caching bug — and the fix that would close it is different in kind: the completion path has to wait for (or force) the assembly set it depends on, rather than sample whatever is loaded.

There is **no open issue** for it — #2616 is closed and was about `MeshNodeLanguageServiceTest`'s *teardown*, a different aspect. So this recurrence is currently untracked, which is how it will be re-diagnosed from scratch next time.

### One more thing in the same log, unexplained

```
digest-mismatch: error        ← 20:26:16 (shard 2) and again 20:34:12 (consolidate)
```

I have not chased it and it may be benign, but it appears in both jobs and it is the kind of line that matters given tonight's registry trouble. Flagging rather than asserting.

### What I would do with this PR

Nothing to fix here. The shard is red on an unrelated, pre-existing flake with a named mechanism — so a re-run is **diagnosis-backed**, not the hope-based re-run AGENTS.md forbids. Worth saying so explicitly in the PR record, and worth filing the recurrence separately so the next occurrence is the second data point rather than the first again.

Issue #2463, opened 2026-08-26
Adopted prebuilt build's source stamp is refused by MergeGuard — IsDirty never converges (the gate now names it compile-status-unrecorded, #2649)


rbuergi:
## What happened

`2fe6590` on main failed the plugin gate:

```
GATE FAILED — compile: RolePlay/Scenery — 1 new failure(s), 0 stale allow entr(ies).
```

with the standard explanation — *"A failing 'compile' check means a public API change here broke plugin node source that is Roslyn-compiled at runtime from the plugins repo."*

**No public API changed, and the compile succeeded.** The run's own log, in order:

```
22:56:03  ok   RolePlay/Scenery [4 source(s), 7 dependency record entr(ies)]   ← COMPILED FINE
22:56:14  bake: RolePlay → 4 assembly(ies) → …/bake-plugins/RolePlay.zip       ← bytes produced

23:02:03  warn MeshWeaver.Data.MergeGuard  RolePlay/Scenery: refused stale/reordered
               cross-hub write to 'lastCompileSucceededAt' (changed since the writer's base)
               — kept the newer live value.        …same for 'lastCompiledVersion',
               'latestAssemblyPath', 'lastModified'

23:02:04  fail MeshWeaver.Graph.CompileWatcher [AdoptedSourceStamp]  RolePlay/Scenery:
               the adopted build's source stamp was committed for request … and the request
               is still standing with IsDirty=true — THE WRITE DID NOT CONVERGE

23:07:22  RED  RolePlay/Scenery [compile:Failed render:Skipped tests:Skipped]
23:10:30  GATE FAILED — compile: RolePlay/Scenery
```

The compiler produced the assemblies. The **mesh phase** then tried to record the outcome on the node, `MergeGuard` correctly refused the cross-hub write as stale/reordered, `CompileWatcher` said so explicitly — and the gate read a status that was never written and called it a compile failure.

## Why it matters

- **It is a false RED on the required check.** `Consolidate test results` fails, so the PR (or main) cannot merge or deliver.
- **It stops delivery.** `main-cd` gates on that check for the SHA: `gate=success reason=push publish=false green=false`. On 2026-08-26 this blocked the recovery of a production rollout that was already wedged for an unrelated reason — the portal stayed ~11 hours behind while the gate reported a compile failure that had not happened.
[...]

PR #2842, opened 2026-08-30
fix: a successful in-mesh compile's diagnostics reach its ACTIVITY

rbuergi:
> *"also doc safety etc. same standard as with normal build … see that you tune it to same level. nullability etc … and doc references and comment references etc … diagnostic must go to activity."*

## Measured from the outside first

A deliberate **`CS0219`** — an unused local — added to an in-mesh source compiled **`ok` with "0 warning(s)"**. That is not a clean build; it is the *absence of a report*. `emitResult.Diagnostics` is read only when `Success` is **false**, so every warning a green compile produced was discarded.

That is why in-mesh C# was not held to the standard the compiled half is held to under `-warnaserror`: no unused-code warnings, and therefore no doc-comment and no **cref** ones either — even though `CreateParseOptions` has *always* asked for `DocumentationMode.Diagnose`. The diagnostics were being produced and thrown away.

## They now go to the compile activity

Which is where a diagnostic belongs: the activity is the record of that compile, what a reader opens, what the Overview/Tests areas render, and what a runner can stream. **A warning that only reached a server log reached nobody.**

The route respects the log-once contract rather than working around it. `EmitPipeline` deliberately does not log — logging failures there once double-counted every compile failure in production (~150 ERROR lines/24h were ~72 real failures logged twice, the duplicate arriving first and context-free, which is what made *"your C# does not compile"* read like an emit/IO defect). So the warnings are **returned**, on `EmittedArtifact`, and appended by the same single funnel a failure takes — at **Warning** severity, which `ActivityLog.Finish` leaves alone, so surfacing them cannot turn a green compile red on its own.

**Both emit paths report the same thing.** The in-memory path reads `compilation.GetDiagnostics()` explicitly: a warning that appeared only when the disk cache was on would be a warning that depends on a deployment's caching mode — the kind of difference nobody finds.

**Deduped, ordered, capped at 50 with the remainder counted** in the last entry. One bad using-directive produces hundreds of identical diagnostics, and an activity that is 400 lines of the same warning is one nobody reads. A truncation a reader cannot see is a lie about how much was wrong.

## Verification

`CompileWarningsReachTheActivityTest` — **4/4**: the unused local (with its line, the only locator a concatenated tree gives), an **unresolvable cref** (`CS1574`/`CS1584` — exactly the breakage an assembly move causes), a clean compile reporting nothing, and the cap naming its remainder.

## Deliberately not in this PR

The compilation **options**. Adding a nullable context is the other half of "the same level" — and it moves `EmitPipeline.OptionsFingerprint`, which is *reflected* over the options, so **every cached assembly in the fleet is invalidated and recompiles**. That is an operational event to schedule, not a rider on a diagnostics fix.

This change is **fingerprint-neutral**: it reads diagnostics that were already being computed.

https://claude.ai/code/session_011vRaDpW29KDS9NX1gvCjw3


rbuergi:
## RCA: this PR is what ejected the merge queue all evening (2026-08-30) — mechanism proven, fix shape below

**Symptom.** `CompileFinishAndDisposeTest.TriggeredCompile_FinishesGeneration_AndActivityReachesTerminalDisposedState` failed `Expected Succeeded … but found Warning` in ≥5 merge-group builds (21:07, 21:17, 22:32, 22:49 ×2), ejecting every stack that contained this PR. Nothing landed on main between 21:51 and 23:22. The 21:07 failure was this PR's own entry built on `main@3421b001e` + this diff ONLY — no other queued PR in the tree.

**Why its own CI is green.** The PR's Build-and-Test run (33334872930, 20:53) tested against merge-base `aab572a7d` (20:22) — six commits stale. Its green attests to `#2842 + main@20:22`; the queue builds `#2842 + main@now`. Only the second has ever failed. Re-queuing on the PR-page green is how the loop closed five times.

**Mechanism, each step measured:**

1. This diff routes a *successful* compile's warnings into the compile activity (`discoveryLog = emit.Warnings.Aggregate(discoveryLog, AppendWarning)`), and `CreateParseOptions` has always asked for `DocumentationMode.Diagnose` — so the test's `public record FinishStory { … }` (no XML doc comment) now lands four `CS1591` entries in the activity log. Transcript from an instrumented run on `main`+this PR:

   ```
   [compile-activity transcript] Id=compile-20260830235720… Status=Succeeded MaxSeverity=Warning MessageCount=14
     Warning: CS1591: Missing XML comment … 'FinishStory' (line 31)
     Warning: CS1591: Missing XML comment … 'FinishStory.Title' (line 31)
     Warning: CS1591: Missing XML comment … 'type_FinishStoryMeshNodeProviderAttribute' (line 35)
     Warning: CS1591: Missing XML comment … 'type_FinishStoryMeshNodeProviderAttribute.Nodes' (line 37)
   ```

2. The in-code comment "Warning severity never flips ActivityLog.Finish's terminal status" is **false**. `ActivityLog.Finish(version, overrideStatus)` computes `MAX(override, GetFinalStatus())` — the override is a floor, not a pin — and `GetFinalStatus()` folds `MaxSeverity=Warning → ActivityStatus.Warning`. So every `.Finish(v, Succeeded)` in `MeshNodeCompilationService` now yields a `result.Log.Status = Warning` for any compile with a warning.

3. There are **two terminal writers that now disagree** on the same activity: `NodeTypeCompilationActivity.Update` PINS the activity node's status (`log with { Status = Succeeded }`), while the pipeline's `result.Log` carries the FOLDED status (Warning). Note `Status=Succeeded` + `MaxSeverity=Warning` in the transcript above — that is the disagreement, live.

4. Which one the response surfaces is a race. `GetCompilationPathRequest` takes the published-release short-circuit (reads the activity NODE → pinned Succeeded → green) only when the compile's write-back stamps (`LatestReleasePath`/`LatestAssembly*` — the same four fields as #2463) have landed; when the request wins that race it falls to the inline path, gets the compile-cache-hit `NodeCompilationResult`, and `BuildResponse` uses the FOLDED `result.Log` → Warning → red. CI load flips the race; a quiet macOS run doesn't (isolated AND full-project bulk runs on `main`+this PR: green, 776/778).

**Fix shape (root cause, in this PR's scope — not a test relax):**
- Make the stated contract true and the writers agree: the compile lane finishes by OUTCOME — `Failed` when the log holds errors (the re-Finish contract at MeshNodeCompilationService.cs:1171-1176 must survive: a post-emit load error still flips to Failed), else `Succeeded`, with warnings present as entries — applied to BOTH the `result.Log` finishes and the activity-node terminal write. (If the desired policy is instead "warning compile ⇒ Warning status", apply THAT to both writers and update the two tests + this PR's comment; either policy is defensible, split-brain is not.)
- Correct the false comment either way.
- Keep a transcript print in `CompileFinishAndDisposeTest` (instrumentation exists on branch `debug/compile-finish-warning-dump`, commit f4f8f768e — take it) so the next anonymous status names its entry.
- Rebase onto current main and let CI run on the ACTUAL tree before this returns to the queue.

**Process note for whoever re-queues anything:** an ejection outranks the PR page's green — the green attests to the merge base, the queue tests the combination. `behind_by` is one API call.

Separate, lower-rate failures tonight (Acme ColdCycle 21:50, GitSync Bidirectional 21:51, Pandas 23:09, ScopeTeardown 22:31/23:22, HOST_CRASHED 22:17/22:32) occurred on trees WITHOUT this PR — same write-back-race neighborhood, different defect(s); not addressed here.

rbuergi:
The writer-consistency half is now its own PR: #2859 (`FinishByOutcome` — both terminal writers, one answer; makes this PR's stated premise true). Once #2859 lands: rebase this onto current main, re-run CI on the actual tree (your last green tested a 6-commits-stale base — see the RCA above), and it should be safe to queue as-written. Your `CompileWarningsReachTheActivityTest` then becomes the end-to-end pin for warnings-surface-without-demoting.

rbuergi:
#2859 is MERGED (main `52e48c807`): `FinishByOutcome` is on main and this PR's premise ('surfacing warnings cannot turn a green compile red on its own') is now true. Ready for you: rebase onto current main, let CI run on the actual tree, then queue. Note your merge base is many commits stale — the full re-run on the rebased head is the point, not a formality (see the RCA above).

rbuergi:
## Merged main in and verified on the real tree — this was 24 commits stale

Acting on the request already on this thread:

> *"rebase onto current main, let CI run on the actual tree, then queue. Note your merge base is many commits stale — the full re-run on the rebased head is the point, not a formality"*

**It read `CLEAN` while being `behind_by = 24`.** That is the stale-green shape exactly: the PR page attests to a merge base that no longer exists, which is what had this re-queued repeatedly before. So it was not safe to queue on that green, however green it looked.

**Merged `origin/main` in cleanly** (no conflicts), then verified on the merged tree rather than trusting it:

```
MeshWeaver.Compiler      -c Release -warnaserror   0 Warning(s) 0 Error(s)
MeshWeaver.Graph         -c Release -warnaserror   0 Warning(s) 0 Error(s)
MeshWeaver.Graph.Test    -c Release -warnaserror   0 Warning(s) 0 Error(s)
MeshWeaver.Graph.Test    1607 / 1607               1 m 1 s
```

That includes this PR's own `CompileWarningsReachTheActivityTest`, and it is green **because #2859 (`FinishByOutcome`) is now on main** — the fix this PR's premise depends on. The interlock that made `CompileFinishAndDisposeTest` fail deterministically when this was batched no longer exists.

CI is now running on the actual tree. **I have not queued it** — that is yours to do once you have read the fresh run, since the whole point of this exercise is that the run, not the rebase, is the evidence.

*(Touched only by merging main and pushing; no change to your commit. Flagging because this branch is checked out in a worktree — pull before you continue.)*

Issue #1369, opened 2026-08-13
A transient fault poisons a promise-cache entry for the life of the process — one blip and a partition can never be provisioned again


rbuergi:
The sanctioned promise-cache recipe caches **failures forever**. Eight live sites hold a `pool.Run(...)` (or a hand-rolled `ReplaySubject`) in a field that is never evicted, so **one transient fault poisons the entry for the life of the process** — the pod must be restarted before that partition / container / connection ever works again.

This is the **fourth** appearance of this exact class in a week, which is what makes it a defect in the *recipe*, not in eight authors:

- `MeshNodeStreamCache`'s per-path entry cache — fixed by 2776a0c86 (`EvictFaultedEntry`), whose comment already names the error class: *"an Npgsql connect failure … replayed forever until a manual recycle."*
- `MeshNodeStreamCache._queries` — #1316, fixed by #1367: one slow Postgres handshake made `nodetype-sources:Edu/Module` permanently unreadable, and the resulting null snapshot classified as a gating `PreWarmStatus.CompileError` that blocked a rollout.

## Why it latches

`IoPoolExtensions.Run` pumps the leaf into a `ReplaySubject<T>` and hands back `subject.AsObservable()`. A `ReplaySubject` latches **terminals**, `OnError` included — so once the cached observable faults, every later subscriber is replayed that same exception, immediately, forever. `Replay(1).AutoConnect(1)` / `.RefCount()` behave identically: the connectable is backed by one subject instance that has already terminated.

Nothing in the recipe removes the entry, so nothing ever re-attempts.

## The worst instance

`PostgreSqlPartitionStorageProvider.EnsurePartitionProvisioned`:

```csharp
return _provisioned.GetOrAdd(def.Schema!, _ =>
    _ioPool.Run(ct => EnsureSchemaAsync(def, ct)).Select(_ => Unit.Default));
```

One transient DDL failure — a connect blip, a `lock_timeout`, a momentary permission problem — **permanently blocks provisioning for that partition**. `EnsurePartitionProvisioned` is the documented gate every write to a not-yet-provisioned partition composes on (`EnsurePartitionProvisioned(p).SelectMany(_ => write…)`), so from that moment every write to the partition fails `42P01 undefined_table`, for the process lifetime, with no self-heal. A partition that is silently and permanently broken until someone restarts the pod.

## Every site

| Site | Shape | Evicts a fault? |
|---|---|---|
[...]

PR #2846, opened 2026-08-30
fix(#2844): an EMPTY scan is a broken scan, not a clean tree

### LINKED ISSUE #2844: 30 of 34 source-tree guards would pass on an EMPTY scan — a planted-tree self-test proves the scanner works, not that it found the tree
## The gap

`test/MeshWeaver.Documentation.Test/` holds **34** source-tree guards and ratchets. They locate the repo by walking up from the test binary (`SourceScan.FindRepoRoot()`) and scan roots like `src`, `test`, `memex`.

**Only 4 assert that the scan found anything.** The other 30 report **green on an empty scan**, because for a zero-tolerance guard "no offenders found" and "nothing was scanned" produce the identical result.

| has a vacuity floor | |
|---|---|
| `NodeTypeContentDiscriminatorGuard` | `MinimumDeclarationsExpected = 25` |
| `TestTimeoutLiteralRatchetGuard` | `TheBaselineStaysCloseToTheTree` (slack ≤ 100) |
| `SubscribeErrorArmRatchetGuard` | `TheBaselineStaysCloseToTheTree` (slack ≤ 15) |
| `ShardWeightHeadroomGuard` | reads a real cap and compares |

Everything else — `HandWovenGateRatchetGuard`, `BlockingBridgeInTestRatchetGuard`, `ObservableToTaskBridgeGuard`, `NoPlaintextCredentialFallbackGuard`, `IdentityReadsRunAsSystemGuard`, `MeshServiceHasNoTaskShimGuard`, `NoLiteralNulInSourceGuard`, and ~23 more — passes if handed an empty directory.

## 🚨 A planted-tree self-test is NOT this protection

Several of these carry an excellent self-test (`TheScannerSeesWhatItClaimsTo`) that plants a temp tree and runs the **real** scanner over it, pinning both directions. That is genuinely good and should stay — but it proves **the scanner works**, not **that the scanner found the production tree**. The two failures are independent:

- *scanner broken* → the planted-tree test catches it.
- *scanner pointed at the wrong or empty tree* → the planted-tree test still passes, and so does the guard.

The second is the one that produces a wall of green while the rule it enforces is unenforced, and nothing in this suite currently detects it.

## How it surfaced

A session preparing to migrate test projects out of core into MeshWeaver.Plugins considered moving `MeshWeaver.Documentation.Test` with them, and stopped on realising the guards would then *"scan the wrong tree and pass on an empty scan, which is worse than deleting them"*. That is exactly right, and it is worth noting the near-miss: **relocating the project is a one-line `.csproj` move that would have silently disarmed ~30 rules**, with every one of them still reporting green forever after.

The same hazard does not need a repo migration to fire. Anything that changes what `FindRepoRoot()` resolves to — a different runner layout, a shadow-copied test binary, a future `bin/` restructure — has the same effect, and the signature is *"all guards green"*, which is indistinguishable from success.

## The fix

Three lines per guard: assert the scan found a plausible number of FILES before evaluating offenders.

```csharp
var files = SourceScan.SourceFiles(root, ProductionRoots).ToArray();
Assert.True(files.Length >= MinimumFilesExpected,
    $"the scan found only {files.Length} file(s) — below the floor. The scan itself has broken "
    + "(a moved tree, a renamed root, a relocated test project); this guard must fail LOUDLY "
    + "rather than report a clean tree it never looked at.");
```

`NodeTypeContentDiscriminatorGuard`'s wording is the model and can be lifted verbatim: *"a count below the floor means the scan broke, not that the tree is clean."*

Suggest a shared helper on `SourceScan` so the floor is one call rather than 30 copies — `SourceScan.SourceFilesWithFloor(root, roots, minimum)` — since a rule copied 30 times is how the 30 came to lack it in the first place.

## Related

This is the in-repo form of the AGENTS.md rule that **a gate must never be unable to fail**, and of `Doc/Architecture/GuardsAndUnknownStates`. Also adjacent to the CI-side version already recorded: a skipped job and a passed job are painted the same colour, so the gate has to assert its own inputs.

rbuergi:
34 source-tree guards and ratchets share one seam — `SourceScan.SourceFiles`. **Only 4 of them asserted that the scan found anything.**

For the other 30, every one a zero-tolerance guard, *"no offenders found"* and *"nothing was scanned"* produce the **identical** result. An empty scan reports green while enforcing nothing.

## The hole is one line

```csharp
.Where(Directory.Exists)      // silently drops a root that is not there
```

That is deliberate — a caller may name an optional root — but it means pointing the scanner at the **wrong tree** drops *every* root and yields an empty sequence.

## Not hypothetical

A session preparing a test migration considered relocating `MeshWeaver.Documentation.Test` into MeshWeaver.Plugins — **a one-line csproj move** — and stopped only on realising the guards would then *"scan the wrong tree and pass on an empty scan, which is worse than deleting them."*

That move would have disarmed ~30 rules **silently and permanently**, every one of them still reporting green. Deleting them would at least have been visible.

## Fixed at the seam, not as 30 floors

A rule copied 30 times is how 30 came to lack it. `SourceFiles` now throws on an empty result, naming the roots and which of them do not exist.

🚨 **The check is on the RESULT, not on every root existing.** A caller naming an optional root that is absent keeps working. That matters: a floor that false-positives gets removed, taking the protection with it — so `TheRealTreeScansFine_AndAnAbsentOptionalRootIsStillTolerated` pins it explicitly.

## 🚨 This is not what the existing self-tests cover

Several guards carry an excellent `TheScannerSeesWhatItClaimsTo` that plants a temp tree and runs the **real** scanner. Those should stay — but they prove **the scanner works**, not **that the scanner found the production tree**. The two fail independently:

| failure | caught by planted-tree self-test? |
|---|---|
| scanner logic broken | ✅ |
| scanner pointed at the wrong/empty tree | ❌ — self-test passes, guard passes, rule unenforced |

Only the second produces a wall of green over an unenforced rule, which is why it needs its own test.

## Mutation-proven

With the throw disabled behind a runtime-false condition the compiler cannot fold (so `--no-build` cannot run stale binaries):

```
AnEmptyScanThrows_RatherThanReportingACleanTree   [FAIL]
AMissingRootThrows_AndNamesWhatWasNotThere        [FAIL]
Failed: 2, Passed: 1
```

The real-tree test **still passes**, so the mutation is isolated to the new behaviour rather than breaking the scanner.

## Verification

`MeshWeaver.Documentation.Test` **132/132 green** — the whole project, since this changes the seam all 34 guards run through. Release `-warnaserror` clean.

Fixes #2844

### DIFF
--- a/test/MeshWeaver.Documentation.Test/SourceScan.cs
+++ b/test/MeshWeaver.Documentation.Test/SourceScan.cs
@@ -52,13 +52,53 @@ internal static class SourceScan
     /// only <c>.cs</c> would report a tree as clean while the same defect sat in a file no compiler
     /// ever looks at. Two <c>.ToTask(</c> sites were found in exactly that position.</para>
     /// </summary>
-    public static IEnumerable<string> SourceFiles(string root, IEnumerable<string> scannedRoots) =>
-        scannedRoots
+    /// <remarks>
+    /// 🚨 <b>An EMPTY result is treated as a BROKEN SCAN, not as a clean tree (#2844).</b>
+    ///
+    /// <para>The <c>Where(Directory.Exists)</c> below silently drops a root that is not there. That
+    /// is deliberate — a caller may name an optional root — but it means that pointing this scanner
+    /// at the wrong tree drops EVERY root and yields an empty sequence. For the ~30 zero-tolerance
+    /// guards built on it, "no offenders found" and "nothing was scanned" are then the same result,
+    /// so all of them report green while enforcing nothing.</para>
+    ///
+    /// <para>That is not hypothetical: relocating <c>MeshWeaver.Documentation.Test</c> into another
+    /// repository — a one-line csproj move considered on 2026-08-30 — would have disarmed the lot,
+    /// silently and permanently. Only 4 of 34 guards asserted that their scan found anything.</para>
+    ///
+    /// <para>🚨 A planted-tree self-test does NOT cover this. Running the real scanner over a temp
+    /// directory proves the SCANNER works; it cannot prove the scanner found the PRODUCTION tree.
+    /// The two fail independently, and only the second produces a wall of green over an unenforced
+    /// rule. Hence the check here, at the one place every guard passes through, rather than 30
+    /// individually-tuned floors — a rule copied 30 times is how 30 came to lack it.</para>
+    /// </remarks>
+    public static IEnumerable<string> SourceFiles(string root, IEnumerable<string> scannedRoots)
+    {
+        var roots = scannedRoots as string[] ?? scannedRoots.ToArray();
+        var files = roots
             .Select(r => Path.Combine(root, r))
             .Where(Directory.Exists)
             .SelectMany(dir => Directory.EnumerateFiles(dir, "*", SearchOption.AllDirectories))
             .Where(f => Path.GetExtension(f) is ".cs" or ".razor" or ".csx")
-            .Where(f => !IsExcluded(root, f));
+            .Where(f => !IsExcluded(root, f))
+            .ToArray();
+
+        if (files.Length == 0)
+        {
+            var missing = roots.Where(r => !Directory.Exists(Path.Combine(root, r))).ToArray();
+            throw new InvalidOperationException(
+                $"SourceScan found NO source files under '{root}' for root(s) "
+                + $"[{string.Join(", ", roots)}]"
+                + (missing.Length == 0
+                    ? " — every root exists but contains no .cs/.razor/.csx. "
+                    : $" — these do not exist: [{string.Join(", ", missing)}]. ")
+                + "🚨 This is a BROKEN SCAN, not a clean tree. A guard that reports green here has "
+                + "checked nothing: the repo root resolved somewhere unexpected (a relocated test "
+                + "project, a shadow-copied binary, a renamed root). Fix the scan — never soften "
+                + "the guard that surfaced it (#2844).");
+        }
+
+        return files;
+    }
 
     /// <summary>
     /// Blanks comment and string-literal characters, preserving offsets and newlines, so the scan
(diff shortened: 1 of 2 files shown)


rbuergi:
## Shard 1's red is not this diff — proven locally, not argued from file paths

```
##[error] MeshWeaver.Hosting.Monolith.Test
          .MeshNodeStreamCacheFaultedEntryReprobeTest.FaultedEntry_IsNotServedTwice_NextRea…
```

This PR's entire diff is two files in `test/MeshWeaver.Documentation.Test/` (`SourceScan.cs` + a new `SourceScanFloorTest.cs`). That *argues* it cannot reach a `MeshNodeStreamCache` test, but the argument is not the evidence, so I checked:

```
$ git log --oneline -1
2ad6b0924 fix(#2844): an EMPTY scan is a broken scan, not a clean tree

$ dotnet test …Hosting.Monolith.Test --filter MeshNodeStreamCacheFaultedEntryReprobeTest
Passed!  -  Failed: 0, Passed: 3, Skipped: 0, Total: 3, Duration: 668 ms
```

**3/3 on this exact head.** So the change is not the cause.

## But it is CI-only, and that is not nothing

A CI-only failure is a real timing/state signal, not noise — local green does not clear it. Recording rather than re-running it away:

- **First occurrence.** Of the last six failed `MeshWeaver Build and Test` runs, only this one carries it, and there is no open issue naming the test.
- **Shape.** `FaultedEntry_IsNotServedTwice_NextRead…` is the faulted-entry reprobe path, i.e. the family where *a `ReplaySubject` latches `OnError` and the cache then serves the corpse* — a defect class this repo has hit before (#1369, the transient connect blip that left a partition permanently un-provisionable). If it is that, "not served twice" is exactly the property that would fail intermittently under load.
- **Not filing an issue on one occurrence**, but a **second** occurrence should get one immediately, with this comment as the first data point rather than starting the count again.

## Why re-running is legitimate here

The run's head (`2ad6b0924`) is the PR's current head — checked before acting, because re-running a run whose head has been superseded re-enters the concurrency group and cancels the live one. And the re-run is backed by a local reproduction attempt that *passed*, so it is diagnosis, not hope: the question it answers is "does this recur", which is the thing worth knowing.

Issue #1741, opened 2026-08-17
mw-plugin-test crashes at startup on a missing --allow file and then SPINS at 100% CPU forever instead of exiting


rbuergi:
## What happened

Two `mw-plugin-test` containers on this machine each burned a **full CPU core for over half an hour** having done nothing at all. Their entire log output was a startup crash, written in the first second:

```
Unhandled exception. System.IO.FileNotFoundException: Could not find file '/repo/plugin-gate.allow'.
   at System.IO.File.ReadLines(String path)
   at MeshWeaver.PluginTester.GateAllowlist.Load(String path) … GateAllowlist.cs:line 51
   at Program.<Main>$(String[] args) … Program.cs:line 48
```

After that: no further output, `docker stats` showing **103% and 101% CPU**, and the containers still "Up" 36 and 57 minutes later. Removing them dropped the colima VM from **239% to 48% CPU**.

## Two defects, both worth fixing

1. **The process does not exit after the unhandled exception.** Something keeps it alive — a foreground thread, a started host, or a hub that outlives `Main` — and it spins rather than idling. On CI this burns a runner until the per-project `timeout 8m` fires, where it reads as a *hang* rather than a *bad argument*, which is a materially worse diagnosis. A tester that cannot start must exit non-zero, immediately.
2. **A missing allowlist is reported as a raw `FileNotFoundException`.** `--allow` names a known-debt ratchet file; when the caller points at a path that does not exist, the correct answer is a one-line message naming the flag, the resolved path, and the fact that the file is optional — not a stack trace through `File.ReadLines`. (Whether a missing allow file should be fatal at all is worth deciding: an absent ratchet is "no known debt", which is a legitimate state — the reusable workflow `node-repo-gate.yml` already omits the flag entirely when no allow file is configured.)

## Why it matters beyond one machine

This is the exact shape of the "orphaned run pegs a core" incident this repo has hit before: a starved colima VM (128 MB free, load average 195) OOM-killed unrelated containers. A crashed process that spins is far more damaging than one that dies, because nothing looks broken — no error is emitted after the first second, and the container reports "Up".

## Fix direction

Fail fast and loud at argument-validation time (before any host is built), print an actionable message, and `return` a non-zero exit code from `Main` so the container exits. Then confirm the process actually terminates — the spin suggests something is holding it open, and that root cause matters more than the message.

Issue #1412, opened 2026-08-13
"format is invalid" on a fresh NodeType assembly survived #1387 — a second root in the compilation-cache path


rbuergi:
Follow-up to **#1387** (closed, fix merged as `f978f2384` / `fc948bf1e`, PR #1393). Filed separately rather than reopening: #1387's fix is correct and complete **for the site it names** — this is a second site, and it must not vanish with the closed issue.

## The claim being corrected

#1387 attributed `ReflectionTypeLoadException … because the format is invalid` on a freshly-emitted NodeType assembly to a non-atomic publish in **both `IAssemblyStore` implementations** (`File.WriteAllBytes` / `DownloadToAsync` straight onto the name readers discover by).

**The signature recurred on #1387's own merge commit.**

| | |
|---|---|
| run | [31683641381](https://github.com/Systemorph/MeshWeaver/actions/runs/31683641381) |
| commit | `e45466ee6` — the merge of PR #1393, i.e. the commit that *landed* the fix |
| when | 2026-08-13T08:47Z, `main`, shard 4 |
| test | `MeshWeaver.PluginCatalog.Test.CodePackageInstallTest.CodePackage_InstallsNodeTypeAndSource_CompilesLive` |

```
Failed to load the compiled assembly — Could not load type 'Widget' from assembly
'DynamicNode_type_widget-type' because the format is invalid.
… Could not load type 'MeshWeaver.Graph.Generated.type_widget_typeMeshNodeProviderAttribute' …

Failed to extract NodeTypeConfigurations from
  /tmp/meshweaver-test-mesh-cache-4693-CodePackageInstallTest/
    type_widget-type_8def91993854d3e_48a950f45c2c4dad89d6b122a275c6f1/type_widget-type.dll

NodeType 'type/widget-type' PARKED after compile failure — further activations serve the
cached error without recompiling
```

## Why this is not #1387's defect

The path is the **compilation cache**, not the assembly store: `{cacheDir}/{nodeName}_{ticks}_{guid}/` is the shape `EmitToDiskWithRetry` publishes, and **that publish is already atomic** — it emits into a `.staging-…` directory whose name cannot match the `{nodeName}_*` discovery glob, verifies the DLL, then `Directory.Move`s it into place. I checked the lambda wiring too: `CompileToDiskAsync` really does pass the staging directory through to `EmitCompilationToDirectory`, so the emit is not bypassing staging.

So neither #1387's defect nor a torn write at this site explains it.

[...]

PR #2850, opened 2026-08-30
feat(cli): memex build project — compile a csproj against the container, no SDK, no NuGet

rbuergi:
> **Maintainer directive, 2026-08-30:** *"the platform builds dll completely without any external dotnet kit or nuget."*

`memex build project <csproj|dir> --image <image>` compiles a .NET project with **no dotnet SDK, no NuGet restore and no platform source checkout** — against the assemblies of a MeshWeaver image, resolved from its own `/app`, its `.deps.json` and the shared frameworks installed in it.

The CLI half is only the trip into the container: it shares `ImageRunner` (pull retry, digest pin, process launch) with `build plugin` rather than carrying a second copy of it. The work is `mw-plugin-test build-project`, which is already in every image.

## What is supported

**`ProjectFile` — the `.csproj`, evaluated without MSBuild**
- Properties: `TargetFramework`, `Nullable`, `LangVersion`, `TreatWarningsAsErrors`, `NoWarn`, `WarningsAsErrors`, `WarningsNotAsErrors`, `DefineConstants`, `GenerateDocumentationFile`, `AllowUnsafeBlocks`, `AssemblyName`, `RootNamespace`, `OutputType`, `ImplicitUsings`, `EnableDefaultItems`/`EnableDefaultCompileItems`, `DefaultItemExcludes`.
- Items: the SDK default glob (`**/*.cs` minus `bin`/`obj` and dot-directories), `Compile Include/Remove` with MSBuild glob semantics, `ProjectReference` (incl. `Remove`, and Windows `\` separators), `PackageReference` (incl. `Remove` and central versions), `PackageVersion`, `Using`.
- The **nearest** `Directory.Build.props`, `Directory.Build.targets` and `Directory.Packages.props` on the way up (MSBuild imports the nearest, not every one), plus every `<Import>` whose condition holds — recursively.
- Condition grammar: `'a' == 'b'`, `!=`, `Exists('p')`, `!`, `AND`, `OR`, parentheses, `$(Prop.EndsWith/StartsWith/Contains('x'))`, `$([MSBuild]::IsOSPlatform('OSX'))`, `$([System.Runtime.InteropServices.RuntimeInformation]::ProcessArchitecture)`.
- The SDK's synthesized inputs: implicit `global using`s (`Microsoft.NET.Sdk` and `.Web`), the target-framework preprocessor ladder (`NET`, `NET10_0`, `NET8_0_OR_GREATER`, …), and its default `NoWarn` (`1701;1702`).

**`ContainerReferenceSet` — the C# port of `MeshWeaver.Plugins/scripts/container-refs.py`**, read from `/app` directly instead of from an extracted image: package versions and the `assemblyVersion` binding identity from the image's own `.deps.json`; assemblies from `/app` ∪ the process TPA ∪ the container's shared frameworks (`Microsoft.AspNetCore.App` is not on the builder's TPA — this is the C# form of `Directory.PlatformRefs.targets`' `<FrameworkReference>`).

**`ProjectBuild`** sequences the `ProjectReference` graph on the existing `Cascade` (a cycle is refused up front and named), compiles with Roslyn, and emits through the platform's own `EmitPipeline` — the verified emit-to-memory-then-write, so the file on disk is provably the image that was emitted (#1412).

## What it refuses, and why nothing is silent

🚨 **A construct the evaluator cannot reproduce FAILS the load, naming the construct and the file** — an unknown element or item type, a `Condition` outside the grammar, an `<Import>` of a missing file (MSB4019's behaviour, deliberately), a `<Target>` (it cannot execute one), an `<EmbeddedResource>` (it embeds nothing), a raw `<Reference>`. `--accept <construct>` acknowledges one deliberately. The alternative is worse than no build: a silently dropped `Nullable`, `NoWarn` or `DefineConstants` is a green build that is **not the build the SDK would have produced**, and nothing downstream can tell.

🚨 **The reference set fails closed** on an unreadable `/app`, a missing or ambiguous `.deps.json`, a deps.json listing no packages, or MeshWeaver assemblies that disagree on their binding identity — MeshWeaver#143's failure, caught in the image instead of at run time.

🚨 **A package is matched by the ASSEMBLY FILE on disk, never by its id alone.** A metapackage whose version the image records but whose assembly is not there is *not* supplied. A `PackageReference` the container does not supply is an **additional library** — additional to the platform — reported by name and refused; `--extra-refs <dir>` is the one way to supply one, so what a build needed beyond the platform is visible in the command that ran it.

🚨 **A run that compiles nothing is a failure**, never a silent success.

## The diagnostic standard

Nullable reference analysis follows the project. `DocumentationMode.Diagnose` is **always** on, so doc-QUALITY defects surface (CS1574 unresolved cref, CS0419 ambiguous cref, CS1570 malformed XML) — while the doc-COMPLETENESS family (CS1591/CS1573/CS1712) is suppressed exactly when the project asked for no doc file, which is when csc would not raise it either. Warnings fail the build by default; `--no-warn=false` / `--allow-warnings` is the deliberate opt-out.

🚨 **`OptionsFingerprint` is untouched.** This verb builds its **own** `CSharpCompilationOptions`; it never calls or mutates `EmitPipeline.CreateCompilationOptions`, which feeds `GeneratedInputIdentity.OptionsFingerprint` — the key every cached NodeType assembly on every mesh is filed under. Only the EMIT is shared.

## What is explicitly NOT supported yet

| | why, and what it would take |
|---|---|
| **Razor / Blazor** | `.razor` and `.cshtml` are not compiled at all. Doing so needs `Microsoft.CodeAnalysis.Razor.Compiler`, `Microsoft.AspNetCore.Razor.Utilities.Shared` and `Microsoft.Extensions.ObjectPool` present in the image — **today it ships none of them**. 15 projects fail with CS0115 ("no suitable method found to override") because their components are missing. |
| **SDK source generators** | `[GeneratedRegex]`, `[LoggerMessage]`, `[JsonSerializable]` live in the .NET **SDK**; a MeshWeaver image ships the **runtime**. `--generators <dir\|dll>` loads them when supplied (through the platform's `GeneratorPipeline`); with none, the builder NAMES the generator that did not run rather than leaving a wall of CS8795 that reads like broken source. 15 projects. |
| **Embedded resources** | Not embedded — so the assembly would differ from the SDK's. A named failure with `--accept embedded-resource`. Supporting it needs faithful reproduction of the SDK's manifest-name mangling, and a half-right name is worse than a refusal. |
| **`<Protobuf>` / gRPC codegen** | protoc is a build task, not a compile. 3 projects. |
| **MSBuild `<Target>`s** | Not executed. Named per target; `--accept target:<Name>` or `--accept targets`. |
| **`<Sdk>` ELEMENT** (Aspire AppHost) | Only the `Sdk` ATTRIBUTE on `<Project>` is understood. |
| **Multi-targeting, `Choose`/`When`, `UsingTask`** | Refused by name. |

## How it was verified

Every command below was run; the numbers are measured, not estimated.

**1. `mw-plugin-test build-project` inside the real pinned tester image**, reading the real `/app`:

```
$ docker run --rm --init -v $BUILDER:/builder:ro -v $PLUGINS/src:/repo:ro -v $OUT:/out \
    --entrypoint dotnet meshweaver.azurecr.io/mw-plugin-test@sha256:74e26e31… \
    /builder/mw-plugin-test.dll build-project /repo/MeshWeaver.Speech --app /app --output /out \
    --accept target:VerifyAssemblyVersionMatchesPlatform
reference set: 272 assembl(y|ies) from /app + the shared framework; 73 package version(s) from the image's deps.json; platform AssemblyVersion 3.0.0.0
MeshWeaver.Speech: ProjectReference MeshWeaver.Mesh.Contract → the container's assembly
[MeshWeaver.Speech.Contract] OK — 1 source file(s), 0 warning(s), 6144 bytes in 557 ms
[MeshWeaver.Speech] package Microsoft.Extensions.Configuration.Binder → the container (10.0.10)
[MeshWeaver.Speech] OK — 4 source file(s), 0 warning(s), 18944 bytes in 339 ms
build-project: GREEN
```

**2. The full `memex build project` verb**, against a local image built `FROM` the pinned tester image with this branch's builder laid over `/app`:

```
$ memex build project $PLUGINS/src/MeshWeaver.Speech --image mw-plugin-test-buildproject:local \
    --no-pull --output ./cli-out --accept target:VerifyAssemblyVersionMatchesPlatform
image: mw-plugin-test-buildproject:local
source root: …/MeshWeaver.Plugins/src → /repo; project: /repo/MeshWeaver.Speech
$ docker run --rm --init -v …/src:/repo:ro -v …/cli-out:/out --entrypoint /app/mw-plugin-test …
build-project: GREEN
assemblies: …/cli-out
```
→ `MeshWeaver.Speech.dll` + `.pdb` + `.xml`, and `MeshWeaver.Speech.Contract.dll` + `.pdb` + `.xml`.

**3. The biggest real project that builds — `MeshWeaver.Import`, 90 source files**, against the **portal** image with two additional libraries:

```
[MeshWeaver.Import] package ClosedXML → --extra-refs
[MeshWeaver.Import] package DocumentFormat.OpenXml → the container (3.5.1)
[MeshWeaver.Import] package CsvHelper → --extra-refs
[MeshWeaver.Import] OK — 90 source file(s), 0 warning(s), 202240 bytes in 1421 ms
build-project: GREEN
```

**4. Every non-test project in `MeshWeaver.Plugins/src` (54 of them)**, against `memex-portal-ai@sha256:15c49ee…`, with `--accept targets --accept embedded-resource` and ClosedXML + CsvHelper as `--extra-refs`:

| outcome | count |
|---|---|
| **green** | **12** — Azure.Blob, ContentCollections.Indexing.PostgreSql, Graph.Views, Hosting.PostgreSql, Hosting.SignalR, **Import**, InstanceSync, Northwind.Domain, Northwind.Model, OgCard, Speech, Speech.Contract |
| source generators (CS8795) | 15 |
| Razor/Blazor (CS0115) | 15 |
| `<Protobuf>` | 3 |
| additional libraries | 5 |
| portal hosts (`<Import>` above the mount, Aspire `<Sdk>`) | 3 |
| other | 1 |

**5. Tests — 51 new, executed** (`dotnet test test/MeshWeaver.PluginTester.Test -c Release` → `Passed! - Failed: 0, Passed: 175`, the whole suite, no regressions):
- `ProjectFileTest` (20) — the default glob and its bin/obj/dot exclusions, `Compile Remove` globs, explicit `Include`, `Directory.Build.props` inheritance and precedence, nearest-only import, `.targets` after the body, the condition grammar incl. a skipped conditional import, `NoWarn` numeric normalisation *and* the SDK defaults, implicit usings, central package versions, the framework symbol ladder, `IsOSPlatform`, and every loud refusal (target, `Choose`, embedded resource, unparseable condition, MSB4019, empty compile set).
- `ContainerReferenceSetTest` (12) — all five fail-closed paths (missing dir, no assemblies, no/ambiguous deps.json, no packages, split binding identity), assembly-on-disk matching, the RENAMED-assembly case, the metapackage case, the additional-library case, the shared-framework scan, and the REAL shipped `mw-plugin-test.deps.json`.
- `ProjectBuildTest` (19) — end to end with the emitted assembly **loaded and its method invoked**; implicit usings; a platform type from the container with nothing declared; dependency ordering asserted on the cascade's own timestamps; cycle refusal; the outside-the-source-root ProjectReference; unresolved packages named; `--extra-refs`; **an unresolvable cref is reported (CS1574)**; **an unused local is reported (CS0219) and fails the no-warn policy**; `--allow-warnings`; project `NoWarn`; nullable on/off; three fail-closed paths; and the streaming contract (every diagnostic reaches the observer, the `ActivityLog` is the same stream sealed).

`FrameworkBuildIdentityTest` (23) still passes — no `ProjectReference` was added to `tools/MeshWeaver.PluginTester`, whose closure IS the framework identity.

## Two side effects worth flagging in review

1. **`BuildPluginCommand`'s docker plumbing moved to `ImageRunner`** and its `docker run` gained **`--init`** — the convention every workflow already passes, and the guard against the SIGABRT spin (#1741). Behaviour otherwise byte-identical; verified by running `memex build plugin` against the pinned tester image and reading the echoed argv.
2. **`MeshWeaver.Compiler` grants `InternalsVisibleTo("mw-plugin-test")`** so the emit is SHARED rather than re-implemented. (The ASSEMBLY name, not the project name — an IVT naming the project grants nothing.)


rbuergi:
Heads-up from #2854, which wires `build-project` into the shared module-pack lane (draft, blocked on this PR + an image + the digest bump).

🚨 **`build-project` does not stamp `<AssemblyVersion>`.** `ProjectBuild` runs no MSBuild targets, so the SDK's `GenerateAssemblyInfo` never happens and `CSharpCompilation.Create` emits Roslyn's default identity. `ProjectFile` never reads the property either (grep for `AssemblyVersion` / `assembly:` hits nothing but `ContainerReferenceSet.PlatformAssemblyVersion`, which is only used to *refuse* a disagreeing image).

That matters for exactly the projects this verb is aimed at. `MeshWeaver.Plugins/src/Directory.Build.props` pins `<AssemblyVersion>3.0.0.0</AssemblyVersion>` with a long comment explaining why: a module loads into the same process as the platform assemblies and is bound BY them, so 0.0.0.0 is MeshWeaver#143 — it does not fail at build, it fails at runtime binding, in another repo, as `FileNotFoundException: Could not load file or assembly 'MeshWeaver.DataStructures, Version=3.0.0.0'`. The measured 12-green sweep would not have surfaced it, because nothing in that sweep loads the emitted bytes.

Two things follow, and the first is free:

1. The 12 "green" projects are green as *compiles*, not yet as *shippable assemblies*. #2854 therefore compares the emitted `AssemblyVersion` against the `platform AssemblyVersion` line this builder already prints, and fails on drift — so a container-built bundle cannot reach the registry with the wrong binding identity.
2. The real fix belongs here: synthesize the assembly-info attributes the SDK's `GenerateAssemblyInfo` emits (`AssemblyVersion`, `AssemblyFileVersion`, `AssemblyInformationalVersion`, and the title/product/company set) from the evaluated properties, unless `GenerateAssemblyInfo` is false. That is one more synthesized syntax tree beside the implicit-usings one, and it closes the gap for every consumer rather than for one lane.

No action needed for this PR to land — just recording it where the next person looks.

rbuergi:
## Heads-up: your branch moved — #2856 was merged into it by me, not by its author

While draining green PRs I merged #2856 (*"the container compiles Razor"*), whose base is **this PR's branch**, `feat/cli-build-project`. So `feat/cli-build-project` has advanced by that change without you asking.

**It was a stacked PR, so this is where its content was always headed** — but the timing was my doing, and you should know before you look at a diff that grew.

This PR reads `DIRTY`, so it runs no CI at all until it is merged with `main`; that was true before my merge and is still true now.

I will revert if you would rather land the stack yourself — say the word. Full explanation on #2856.

rbuergi:
## Merged main in and resolved the conflict — it was a two-sided change, and the obvious resolutions are both wrong

This was `DIRTY`, which means it was running **no CI at all**. One conflict, in `src/MeshWeaver.Cli/BuildPluginCommand.cs`:

- **this branch** extracted `PullImage` / `Exec` / `Capture` / `PullAttempts` out into the new `ImageRunner.cs`;
- **main** added `ComposeSealedModules` + `StageModuleBundle` to the same file (#2851, #2858 — modules come from the publication's sealed module set).

Git presented both as **one 159-line block with an empty "ours" side**, which makes each easy resolution wrong in a different direction:

| resolution | what it does |
|---|---|
| take **theirs** | reinstates `PullImage` here *and* leaves it in `ImageRunner` — duplicate definition |
| take **ours** | drops the whole block, **silently deleting main's `ComposeSealedModules`** |

I did take "ours" first, and the build caught it — `CS0103: ComposeSealedModules does not exist`. That is precisely why the build ran before the push rather than after: a merge resolution that reverts main is invisible in the diff of the PR that does it.

**Resolved by splitting the block at its real seam:** dropped the 39 lines of relocated `PullImage`, kept the 120 lines of main's new code. The compiler then named two follow-ons — main's auto-merged code called `Capture()` unqualified (it is now a `static` on `ImageRunner`), and the split orphaned `PullImage`'s `/// <summary>` opener, which `CS1570`'d against the next member's doc.

**Verified that neither side was lost, not merely that it compiles:**

```
ComposeSealedModules / StageModuleBundle present   5 references
PullImage defined exactly once   ImageRunner 1, BuildPluginCommand 0

MeshWeaver.Cli / PluginTester / PluginTester.Test   -c Release -warnaserror, 0/0
MeshWeaver.PluginTester.Test                        188/188
```

CI is running now for the first time on this PR since it went dirty. **Not queued** — the resolution touched your refactor's seam, so you should read the run and confirm the split matches what you intended.

*(This branch is checked out in a worktree — pull before continuing. #2853, the batch that contained this PR, is now closed: its other three members landed individually.)*

PR #2853, opened 2026-08-30
chore: land the core batch as one merge — four green PRs, one queue cycle

rbuergi:
Batches four individually-green PRs into a single merge so they cost **one** queue cycle instead of four.

| PR | what |
|---|---|
| #2848 | docs: the fleet test shape |
| #2849 | the install no longer waits 30s for a node documented as optional |
| #2850 | `memex build project` — compile a csproj against the container, no SDK, no NuGet |
| #2852 | a FATAL is never silent; the repo sweep skips unreadable files loudly |

## 🚨 #2842 was REMOVED from this batch — it is the cause, not a passenger

The first cut of this batch included #2842 and failed `CompileFinishAndDisposeTest` on shard 2 **three times, deterministically**, including on a freshly-rebased base with the merge queue empty. I misread that as base-dependence. It was not.

meshweaver-61 proved the mechanism (core #2842, `issuecomment-5472066551`) — a four-part interlock:

1. #2842 funnels **successful**-compile warnings into the activity log (`DocumentationMode.Diagnose` was always on), so an undocumented `public record` yields CS1591s.
2. Its own comment — *"Warning never flips Finish's terminal status"* — is **false**: `Finish` is `MAX(floor, fold)`.
3. Dual-terminal-writer split-brain: the activity-node writer pins `Succeeded` while `result.Log` carries the folded `Warning`.
4. Which of the two the response surfaces depends on #2463's stamp-write-back race — that is the coin-flipper; #2842's inconsistency is the two-headed coin.

Hence: green in isolation on a quiet machine, red under CI load. #2842 returns once it carries the writer-consistency fix and a rebase.

## The #2850 / #2851 conflict, resolved deliberately

#2851 landed on main while #2850 sat queued, and both touch `src/MeshWeaver.Cli/BuildPluginCommand.cs`. The resolution keeps **#2851's `ComposeModulesFromSeed`** (now trunk content) and drops the **`PullImage`** that #2850 moved into the new `ImageRunner.cs` — verified the move had actually landed (`ImageRunner.PullImage` exists; `BuildPluginCommand` calls `_runner.PullImage`) before deleting anything.

Two defects in the first resolution attempt, both caught by building rather than reading: a duplicated `/// <summary>` (the original tag sat above the conflict marker as shared context) and a `Capture` call needing qualification to `ImageRunner.Capture`.

## Verification

Full solution build on the merged tree: **90 projects, 0 Warning(s), 0 Error(s)** under `-warnaserror`.

Each constituent PR carries its own tests and evidence; this changes none of them.


rbuergi:
## Superseded — three of the four members landed individually, so the batch no longer saves a cycle

This existed to spend **one** queue cycle instead of four. That rationale is gone:

| member | state |
|---|---|
| #2848 docs: the fleet test shape | **MERGED** |
| #2849 the install no longer waits 30 s | **MERGED** |
| #2852 a FATAL is never silent | **MERGED** |
| #2850 `memex build project` | still open |

Only #2850 remains, so there is nothing left to batch — and keeping this open is actively risky rather than merely redundant. It is `DIRTY`, so it runs **no CI at all**, and its branch now carries three commits whose content is already on `main`. A branch holding an already-merged commit reads `CLEAN` on the PR page and comes back `UNMERGEABLE` from the queue, because the speculative merge sees that content arriving twice — which is a confusing hour to spend for no benefit.

**#2850 should land on its own.** It is `DIRTY` too and needs its conflicts with `main` resolved before it runs any CI; I am looking at that next.

Closing rather than deleting the analysis: the RCA in the body — why #2842 was removed from this batch, and the four-part interlock behind `CompileFinishAndDisposeTest` — is worth keeping readable, and #2842 is now rebased onto a main that carries #2859, which is the fix that interlock depended on.

PR #2854, opened 2026-08-30
ci(module-pack): the platform image compiles the module, not the runner's SDK

rbuergi:
> **DRAFT — do not merge, do not arm auto-merge.** This lane cannot be switched on until the
> pinned platform image carries the verb it calls. The dependency chain is at the bottom.

## What this does

> *"we want to use memex build plugin and not dotnet build"* — maintainer, 2026-08-31

A module does not run against a source tree and it does not run against a feed: it is loaded into
the platform **image** and bound by the assemblies in there. So the honest compiler is that image's
own. `node-repo-module-pack.yml` — the reusable lane every node-plugin repo calls — now takes the
compiler **per matrix entry**:

```jsonc
{"package":"Import","module":"MeshWeaver.Import",
 "project":"src/MeshWeaver.Import/MeshWeaver.Import.csproj",
 "build":"container","accept":"targets"}
```

* `"build": "container"` → `memex build project`, i.e. `mw-plugin-test build-project` **inside the
  pinned platform image**: no .NET SDK, no NuGet restore, no platform source checkout. The lane runs
  the container form directly (`docker run --entrypoint /app/mw-plugin-test … build-project`) rather
  than through the `memex` CLI, because CI has already pulled and digest-pinned the image — going
  through the CLI would add an SDK build in order to remove one.
* omitted (or `"sdk"`) → today's `dotnet build` + `dotnet publish`, **byte for byte unchanged**.

**No caller changes behaviour from this PR.** Every satellite that declares nothing keeps the SDK
path with the identical commands, the identical `--deps-closure`, and the identical assertions.

## The mechanism for the unsupported majority

#2850's own sweep over all 54 non-test projects in `MeshWeaver.Plugins/src` compiled **12** green.
The rest need what a *runtime* image does not carry: SDK source generators (15), Razor/Blazor (15),
`<Protobuf>` (3), additional libraries (5), portal hosts (3). A blanket swap would break most module
bundles, so the split is **declared per module** — never guessed, never auto-detected.

🚨 **And there is no fallback between the two.** A lane that tries the container and quietly drops
back to the SDK makes *"the container built it"* and *"the SDK built it"* indistinguishable in a
green log. A declared mode that cannot run **fails**; an unknown mode **fails**; `container` without
a platform image to build against **fails**.

It is also **one step, not two `if:`-guarded ones** — two guards can both skip, and a skipped step
renders with the same tick as a passed one. The single step always runs and always says which
compiler it used.

## How the log makes the path unmistakable

| where | what it says |
|---|---|
| build step, every run | `compiler: CONTAINER — memex build project ⇒ mw-plugin-test build-project inside <image>@<digest>` … or `compiler: dotnet SDK <ver> on the runner — the LEGACY path` plus a 🚧 line naming the entry field that would convert it |
| build step, container path | every ride-along decision (`closure: X.dll RIDES — module-owned` / `does NOT ride — the image ships it`), why there is no `--deps-closure`, and the binding-identity comparison |
| job summary | a per-module block: mode, pack input, closure flags |
| built marker + receipt JSON | `"compiler": "container" \| "sdk"` |
| `All selected bundles built` | **`N of M` module bundles built by `memex build project`**, and the two module lists — a split nobody counts is a split that becomes permanent |

## The closure, with no `deps.json` to derive one from

`--deps-closure` reads the `deps.json` the SDK emits, and there is no SDK on the container path.
What replaces it is a stronger fact, not a weaker one: `build-project` **refuses by name** every
`PackageReference` the image does not supply, and this lane passes **no `--extra-refs`** — so a
container build that succeeded proves every assembly the module binds is one the image it lands
into already carries.

That claim is **asserted, not asserted-in-prose**: the bundle inspection refuses any
non-`MeshWeaver.*` assembly in a container-built bundle. And the in-root `ProjectReference`s the
builder compiled from source are handled explicitly — **module-owned** siblings ride via `--with`
(derived from what the builder actually emitted, classified by the same `module-owned-platform.sh`
the packer and the inspection already use), **image-shipped** ones (`src/platform-shipped.txt`)
deliberately do not, because that would land a same-identity duplicate beside `/app`'s copy.

An entry that would need `--extra-refs` simply stays on the `sdk` path — the builder refuses it by
name, which is the right answer rather than a bundle whose closure had to be guessed (the
2026-08-19/20 outage).

## 🚨 A finding: `build-project` does not stamp `<AssemblyVersion>`

`ProjectBuild` runs **no MSBuild targets**, so the SDK's `GenerateAssemblyInfo` never happens and
`CSharpCompilation.Create` emits Roslyn's default identity. `MeshWeaver.Plugins/src/Directory.Build.props`
pins `<AssemblyVersion>3.0.0.0</AssemblyVersion>` precisely because a mismatch is MeshWeaver#143: it
does not fail at build, it fails at **runtime binding**, in another repo, as a `FileNotFoundException`
naming a version nobody wrote down.

So the lane **compares the emitted `AssemblyVersion` against the one the builder itself read out of
the image** (`platform AssemblyVersion …`, printed by `ContainerReferenceSet`) and fails on drift.
Until `build-project` honours the property, that check is what a `container` entry hits — by design:
loud and early, rather than a 0.0.0.0 assembly on the registry.

## 🚨 Dependency chain — why this is a draft

1. **#2850 merged** (`feat/cli-build-project` — `memex build project` + the `build-project` verb).
2. **An image built from ≥ that commit.** `mw-plugin-test` ships *inside* the platform image; before
   that image exists the container run dies on an unknown argument.
3. **`platform-image-digest` moved to that image** in each caller. `meshweaver-plugins-08` is
   already carrying that bump — not duplicated here.
4. **`build-project` honouring `<AssemblyVersion>`** (the finding above), or the identity check
   fails every `container` entry.
5. Only then: a follow-up PR **in each caller repo** flips the converted entries to
   `"build": "container"`. Nothing in this PR arms anything.

## What was verified, and what could not be

Verified locally:

* `actionlint` clean; `actionlint` + `shellcheck` reports **fewer** findings than the base file (no
  new ones).
* Every `run:` script `bash -n` clean (GitHub expressions neutralised first).
* `node-repo-pack-verify.py --self-test` and `node-repo-scope.py --self-test` green — the receipt's
  new `compiler` field is tolerated, and unknown matrix-entry fields already pass through the scope
  script's `{**e, …}`.
* The `jq` split expressions, the portable array/`while read` handling (no `mapfile`), the
  own-platform `;name;` membership test, and the `AssemblyName.GetAssemblyName` one-liner run as a
  .NET 10 file-based app — each exercised against real inputs.

**Not verified, and cannot be until the image carries the verb:** no `container`-mode run has ever
executed. The container command line, `build-project`'s `<output>/<AssemblyName>/` layout, the
`platform AssemblyVersion` log line this lane greps, and the `--accept targets` requirement are all
read off #2850's source and docs, not observed. The first real proof will be the first caller entry
that declares `container` — which is exactly why no entry declares one here.

---

## Coverage — the statement of record

🚨 **Quote the measurement basis whenever this number comes up.** Three different figures have circulated tonight and they are not comparable:

| figure | basis |
|---|---|
| 12 of 54 | the **tester** image (`mw-plugin-test`, ~100 assemblies) — its `/app` is the PluginTester tool's own closure, NOT the platform |
| 9 of 54 | the **portal** image (`memex-portal-ai`, 209 assemblies) — the correct reference set |
| **10 of 54** | portal image, **after** `<EmbeddedResource>` support | 

**The current measured floor is 10 of 54, portal basis.** Size any adoption plan from that.

### A ranking of FIRST refusals is not a ranking of independent gaps

The builder stops at its first refusal, so a refusal masks every gap behind it. Adding `<EmbeddedResource>` support was expected to take 9 → ~24 because 19 projects were refusing on it. The measured result was **9 → 10**:

| | before | after |
|---|---|---|
| green | 9 | **10** |
| `<EmbeddedResource>` refused | 19 | **0** |
| Razor (CS0115) | 12 | **15** |
| source generators (CS8795) | 3 | **14** |
| additional libraries | 5 | 8 |

Those 19 were never "19 projects from green" — they were 19 projects about which **nothing further was known**. Closing the gap turned one green and revealed what was behind it; source generators went 3 → 14. No project regressed.

The same caveat applies to the numbers still on the board: Razor (15) and generators (14) are *first*-refusal counts too, so closing Razor will likely reveal more generators rather than yield 15 greens.

**Source generators are the honest unknown** — they live in the SDK, not the runtime. Nobody should project a number for them before someone runs one generator on one project in-container.

This is why `"build": "container"` is opt-in per matrix entry in this PR and never a fallback: the opt-in list grows on measured evidence, one module at a time.

### DIFF
--- a/.github/workflows/node-repo-module-pack.yml
+++ b/.github/workflows/node-repo-module-pack.yml
@@ -72,6 +72,66 @@ name: node-repo module pack
 # A caller that ships neither scripts/affected-modules.py nor scripts/project-closure.py is simply
 # not narrowable: the scope script says FULL and says why. Nothing to opt into, nothing to skip.
 #
+# ─────────────────── Which compiler builds a module (`build`) ───────────────────
+# 🚨 "we want to use memex build plugin and not dotnet build" (maintainer, 2026-08-31).
+#
+# A module does not run against a source tree and it does not run against a feed: it is loaded into
+# the platform IMAGE and bound by the assemblies in there. So the honest compiler is that image's
+# own — `memex build project`, whose work is `mw-plugin-test build-project` INSIDE the pinned image
+# (MeshWeaver#2850). This lane runs the container form directly, because CI has already pulled and
+# digest-pinned the image; going through the `memex` CLI would add an SDK build in order to remove
+# one.
+#
+# It is DECLARED PER ENTRY, and it is deliberately not yet the default:
+#
+#     {"package":"Import","module":"MeshWeaver.Import",
+#      "project":"src/MeshWeaver.Import/MeshWeaver.Import.csproj",
+#      "build":"container","accept":"targets"}
+#
+# #2850's own sweep over all 54 non-test projects in MeshWeaver.Plugins/src compiled 12 green. The
+# rest need what a RUNTIME image does not carry: SDK source generators (15 — they live in the SDK),
+# Razor/Blazor (15 — the image ships none of the three Razor assemblies), `<Protobuf>` (3),
+# additional libraries (5), portal hosts (3). A blanket swap would therefore break most bundles, so
+# the split is deliberate, per module, visible, and expected to SHRINK — `verify` prints the count
+# every run for exactly that reason.
+#
+# 🚨 THERE IS NO FALLBACK BETWEEN THE TWO, and that is the point. A lane that tries the container
+# and quietly drops back to the SDK makes "the container built it" and "the SDK built it"
+# indistinguishable in a green log. A declared mode that cannot run FAILS; an unknown mode FAILS.
+# The mode is printed by the build step, written into the built marker and the receipt, and
+# summarised by `verify`.
+#
+# What the container path needs from the caller:
+#   * `platform-image` + `platform-image-digest` — the image IS the reference set, so an entry that
+#     asks for `container` without them fails RED rather than building against anything else.
+#   * the `acr-username`/`acr-password` secrets, to pull it.
+#   * `accept` for every construct the evaluator refuses. MeshWeaver.Plugins' own
+#     `src/Directory.Build.props` carries a `<Target>` (the AssemblyVersion contract check), so its
+#     entries need at least `"accept": "targets"`.
+#
+# What it deliberately does NOT support:
+#   * `--extra-refs`. An "additional library" is one the image does not carry; with no deps.json to
+#     derive a private closure from, bundling it would be guesswork and NOT bundling it would land a
+#     module that faults at first use (the 2026-08-19/20 outage). build-project refuses such a
+#     PackageReference BY NAME, which is the right answer: that module stays on the `sdk` path.
+#   * `--deps-closure`, for the same reason there is nothing to derive it from. The container
+#     bundle's closure is complete BY CONSTRUCTION — every reference resolved from the image the
+#     bundle lands into — and the bundle inspection ASSERTS that claim by refusing any
+#     non-MeshWeaver assembly in a container-built bundle.
+#
+# 🚨 PRECONDITIONS BEFORE ANY ENTRY MAY DECLARE `container` — both are checked at runtime, loudly,
+# but they are cheaper to read here:
+#   1. The PINNED image must carry the `build-project` verb (MeshWeaver#2850 merged AND an image
+#      built from it AND `platform-image-digest` moved to that image). Before that, the container
+#      run exits non-zero on an unknown argument.
+#   2. `build-project` must stamp the project's <AssemblyVersion>. It runs no MSBuild targets, so
+#      the SDK's GenerateAssemblyInfo does not happen and Roslyn's default identity is emitted —
+#      shipping 0.0.0.0 into a 3.0.0.0 process is MeshWeaver#143 exactly: it does not fail at build,
+#      it fails at runtime binding, in another repo, naming a version nobody wrote down. The build
+#      step compares the emitted identity against the one the builder itself read out of the image
+#      and FAILS on drift, so this cannot ship silently — but until the builder honours the
+#      property, that check is what a `container` entry will hit.
+#
 # Republishing is SAFE to repeat, which is why the schedule may simply publish rather than
 # first proving something changed: the registry is keyed by `{package}@{version}` from
 # `manifest.lock`, so a rebuild of an unchanged version overwrites the same slot with equivalent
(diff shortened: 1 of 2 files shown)


rbuergi:
## Fresh measurement: 18 of 49 green — the floor DOUBLED today, and every input is now on main

Full native sweep over `MeshWeaver.Plugins/src` (49 non-test projects), builder = **main as of `9fb74fd9`** (#2892: embedded resources + assembly identity, union-merged with #2856's Razor), reference set = a fresh portal publish from current main, plus the SDK's own regex generator handed via `--generators`:

```
=== 18 green / 31 red ===
```

### The green list — the seed for this lane's opt-in entries

```
MeshWeaver.AI                MeshWeaver.AI.Anthropic       MeshWeaver.AI.AppleIntelligence
MeshWeaver.AI.WebSearch      MeshWeaver.ContentCollections.Indexing.PostgreSql
MeshWeaver.Graph.Views       MeshWeaver.Hosting.PostgreSql MeshWeaver.Hosting.SignalR
MeshWeaver.InstanceSync      MeshWeaver.Markdown.Collaboration  MeshWeaver.Mcp
MeshWeaver.Northwind.Domain  MeshWeaver.Notifications.Channels  MeshWeaver.Observability
MeshWeaver.Observability.Contract  MeshWeaver.OgCard       MeshWeaver.Speech  MeshWeaver.Speech.Contract
```

`MeshWeaver.AI` itself is the headline: **169 sources, 0 errors, 0 warnings** — the whole family was unbuildable this morning behind one `<EmbeddedResource>` refusal.

### What closed, measured one blocker at a time

| blocker | closed by | evidence |
|---|---|---|
| `<EmbeddedResource>` (blocked the 12-project AI family transitively) | #2892's resource support | refusal gone; AI compiles |
| CS0436 type conflicts | a FRESH reference set | 22 errors → 7 on `Markdown.Collaboration` |
| CS8795 `[GeneratedRegex]` — this PR's *"source generators are the honest unknown"* | `--generators <SDK's System.Text.RegularExpressions.Generator.dll>` | 7 → 0; **the unknown now has a measured answer: they run** |

That third row matters most for this lane: the generator DLL ships in `packs/Microsoft.NETCore.App.Ref/<ver>/analyzers/` — the lane can mount it beside the razor-generators the image already carries.

### The remaining 31, by first-refusal class (samples verified verbatim)

- **additional libraries** — provider SDKs the portal image does not carry (`Azure.AI.Inference`, `Microsoft.Agents.AI.*`, Sqlite, Snowflake…): the `--extra-refs` bucket, and the honest split between "image should carry it" and "module brings it".
- **scoped CSS** — the Blazor family (`*.razor.css` bundling): the next builder layer after Razor.
- **`<Protobuf>`** — 2 projects; likely stays SDK-built.

First-refusal caveat from this PR's own text still applies: closing scoped CSS will reveal what is behind it.

### Two sweep artefacts worth recording

- The per-satellite cost (~2.7 min each) is the SWEEP re-building `MeshWeaver.AI` per dependent — this lane builds each module once, so it does not inherit that.
- In-container on Apple Silicon segfaults (exit 139, Roslyn under amd64 emulation): local measurement is native against a published `/app`; the authoritative sweep is CI's, exactly as this lane runs it.

**With #2892 merged, everything this draft depends on is on main.** The opt-in list above is measured, not guessed — ready to seed the matrix entries, and the Plugins / SocialMedia adoption is a per-entry `"build": "container"` flip in their own repos.

Issue #143, opened 2026-07-01
Database schema issue on the backend (sync_behavior column missing) preventing Space creation


rosalie1984:
In the thread "can you create a new space called ILS Deepfield" I got the error message that there is a database schema issue on the backend (sync_behavior column missing) preventing Space creation. This isn't something I can fix from the mesh — it's an infrastructure error.

rbuergi:
**Root cause found (2026-07-01).** The `sync_behavior` column is missing because the **database migration never runs** — the `memex-migration` container on the current image (`3.0.0-ci.122` / `6943991`) crashes on startup:

```
Current DB version: 39
Unhandled exception. System.IO.FileNotFoundException: Could not load file or assembly
'MeshWeaver.Documentation, Version=3.0.0.280, Culture=neutral, PublicKeyToken=null'.
```

So the `memex-migration-deployment` is in **CrashLoopBackOff** on atioz + memex (verified), the schema stays at **v39**, and the migration that adds `sync_behavior` (PR #119, `MeshNode.SyncBehavior` persistence) is never applied → Space creation fails exactly as reported.

Note this is **separate from the portal**: the portals on `ci.122` run fine (all 3 rolled and 1/1 Running) — only the *migration image* is broken. It's a packaging/assembly-version mismatch: the migration binary references `MeshWeaver.Documentation` v`3.0.0.280` but that assembly isn't present at that version in the migration image.

**Fix required:** correct the `memex-migration` image build so `MeshWeaver.Documentation` is packaged at the referenced version (CD build), then let the migration run to apply v40 (`sync_behavior`). Until then the schema can't advance past v39. Investigating the migration image build next.

PR #2855, opened 2026-08-30
fix(build-project): the emitted assembly carries the platform's binding identity

rbuergi:
## The defect: a wrong binding identity that nothing turns red

`GenerateAssemblyInfo` is an MSBuild **target**, and `build-project` (#2850) deliberately runs no
targets. So the SDK's `[assembly: AssemblyVersion("…")]` was never written, and Roslyn emitted its
own default: **`0.0.0.0`**, where the whole fleet binds **`3.0.0.0`**.

Measured first-hand, not inferred — `MeshWeaver.Plugins/src/MeshWeaver.Speech.Contract` built
through a tester image from `feat/cli-build-project`:

```
MeshWeaver.Speech.Contract AssemblyVersion=0.0.0.0
```

**The build was GREEN.** That repo's `src/Directory.Build.props` pins `<AssemblyVersion>3.0.0.0`
precisely because #143 shipped 1.0.0.0 assemblies into a 3.0.0.0 process and CrashLoopBackOff'd a
migration; its own comment records why the class is nasty — *"a target that runs
`BeforeTargets="CoreGenerateAssemblyInfo"` still builds 1.0.0.0, silently."* The failure lands at
**runtime binding, in a different repo**, as `FileNotFoundException: Could not load file or assembly
'…, Version=3.0.0.0'`.

> 🚨 **This is a prerequisite for any module adopting `"build": "container"` in the pack lane
> (#2854).** A bundle built without it ships the wrong binding identity, passes every gate it is
> subjected to, and fails only at runtime — in a different repo, in a different process, as a
> missing-file error naming a version nobody wrote down. Nothing in the pack lane can detect it.

## The fix

`ProjectFile` **evaluates** the identity; `ProjectBuild` **synthesizes** it into the compilation as
one more source document — exactly as the SDK's generated `<Project>.AssemblyInfo.cs` is one more
Compile item.

🚨 `EmitPipeline.CreateCompilationOptions` and everything feeding `OptionsFingerprint` are
**untouched**; #2850's own `CSharpCompilationOptions` is where this lives.

### The SDK semantics, MEASURED rather than recalled

Every rule below was produced by building the same project with SDK 10.0.400 and reading the
generated `*.AssemblyInfo.cs`:

| `$(Version)` | SDK `AssemblyVersion` | here |
|---|---|---|
| `1.2.3-beta.4` | `1.2.3.0` (label dropped) | ✅ |
| `1.2.3+meta` | `1.2.3.0` (metadata dropped) | ✅ |
| `1.2.3.4` | `1.2.3.4` | ✅ |
| `4.5` / `1` | `4.5.0.0` / `1.0.0.0` (padded) | ✅ |
| `01.2.3` | `1.2.3.0` (zeros normalise) | ✅ |
| *(unset)* | `1.0.0.0` via `VersionPrefix` | ✅ |
| `1.2.3.4.5` | rejected | named refusal |

Plus the fallback ORDER, which is the easy thing to get backwards: an explicit `AssemblyVersion`
wins, **`FileVersion` follows `AssemblyVersion`** (not `$(Version)`), `InformationalVersion` follows
`$(Version)`, and `Company`←`Authors`←`AssemblyName` / `Product`,`Title`←`AssemblyName`.

`GenerateAssemblyInfo=false` synthesizes **nothing** — the project supplies its own and a second set
is CS0579. Each individual `Generate…Attribute` switch is honoured.

### Three item types that were being dropped

`InternalsVisibleTo`, `AssemblyMetadata` and `AssemblyAttribute` were on this evaluator's *"metadata
that does not change which sources compile against what"* list. True of the **compile**, false of the
**assembly**: an `InternalsVisibleTo` dropped here makes the friend project fail to compile in a
later, unrelated run. Sixty-odd projects across these repos declare one.

One consequence worth flagging: now that `AssemblyMetadata` **Value** metadata is read, a project
whose value uses a property function outside this evaluator's grammar becomes a **named refusal**
where it used to be a silent drop. Exactly one project in either repo is affected —
`MeshWeaver.Plugins/src/MeshWeaver.AI.Test`, whose
`<AssemblyMetadata Include="MeshWeaverRoot" Value="$([System.IO.Path]::GetFullPath('$(MeshWeaverRoot)'))" />`
now fails the load by name. That is this file's stated doctrine, and **no CI lane calls
`build-project` yet**, so nothing regresses.

### Where the SDK cannot be matched, it FAILS BY NAME

An unparseable `$(Version)`; `PublishRepositoryUrl=true` with no `$(RepositoryUrl)` (the SDK reads
that off the git remote). No plausible-looking substitute — a wrong binding identity is the entire
defect.

### Two divergences, both said out loud

- **`$(SourceRevisionId)`** — the SDK appends `+<sha>` from git. There is no git here, so *every
  build reports the absent suffix by name*, and the new `-p:Name=Value` gives exact parity.
- **`TargetFrameworkAttribute`** — a different target (`GenerateTargetFrameworkMonikerAttribute`)
  with its own switches and its own `TargetPlatform`/`SupportedOSPlatform` companions for a
  platform-suffixed TFM this evaluator cannot compute. Emitting half of that set would be worse than
  the gap, so the gap is written down in the code and the README instead.

### One bug found on the way

`ProjectFile`'s `globalProperties` were documented to *"win over anything the files set
unconditionally"* and did not — a `<PropertyGroup>` overwrote them. Fixed (a global property is now
immutable during evaluation, as under MSBuild), because otherwise `-p:AssemblyVersion=…` would mean
nothing for exactly the property a caller passes it for.

## Verified by RUNNING

**Unit** — 23 new cases in `test/MeshWeaver.PluginTester.Test/ProjectAssemblyIdentityTest.cs`; suite
**198/198 green** (175 before, +23). Every assertion reads the emitted **file's metadata** (`PEReader`
— not `Assembly.LoadFrom`, which refuses a second assembly of the same simple name and would turn an
identity suite into a load-order suite).

**Mutation proof** — with the synthesized tree disabled, the key test fails exactly as the defect
predicts:

```
Expected value to be 3.0.0.0 because the fleet binds 3.0.0.0 …, but found 0.0.0.0.
```

**End to end**, the reproduction that found the defect — tester image built from this branch,
`MeshWeaver.Plugins/src/MeshWeaver.Speech.Contract` built inside it:

```
[MeshWeaver.Speech.Contract] assembly identity — AssemblyVersion 3.0.0.0, FileVersion 3.0.0.0,
  InformationalVersion 1.0.0 (no SourceRevisionId: this builder runs no git, so the SDK's +<sha>
  suffix on InformationalVersion is absent — pass --property SourceRevisionId=<sha> to reproduce it)
build-project: GREEN

$ AssemblyName.GetAssemblyName(…/MeshWeaver.Speech.Contract.dll).Version
3.0.0.0
```

**Parity against a real `dotnet build` of the same project**, attribute for attribute — Company,
Configuration, Description, FileVersion, InformationalVersion, Product, Title, AssemblyVersion all
identical; the only difference is `TargetFrameworkAttribute` (above). With
`-p:SourceRevisionId=7e0750c…` the `InformationalVersion` matches the SDK's exactly too:
`1.0.0+7e0750c274faa2ecf4a918ed6891ef96e326c8b2`.

## A recommendation, not a change: `--accept targets`

`build-project` refuses to run against `MeshWeaver.Plugins/src` without `--accept targets`, because
of `<Target Name="VerifyAssemblyVersionMatchesPlatform">` — the guard that exists to catch exactly
this drift. So today that flag is what lets a wrong binding identity through. **I have not changed
it.** What I'd suggest instead, for a maintainer call:

`ContainerReferenceSet` already knows `PlatformAssemblyVersion` (`3.0.0.0`, read from the image).
`build-project` could **WARN when the identity it is about to emit differs from the platform's** —
naming both numbers. That is the same check the skipped target performs, done from data the builder
already has, and it needs no MSBuild. A warning rather than an error because a standalone tool may
legitimately differ; under the default no-warn policy a module build would still go red. That would
make `--accept targets` safe again, rather than an accepted blind spot.

## Base — read this before reviewing the diff

The branch is cut from **`integration/core-batch`** (#2853), because #2850 has not landed on `main`
yet and the `build-project` files this changes **do not exist there**.

The PR nonetheless targets **`main`**, not the batch: a PR merged into a feature branch looks merged
while never reaching `main`, and this has to land on the trunk. The cost is that GitHub currently
shows the batch's files in the diff as well. **The five files this PR actually changes are:**

- `tools/MeshWeaver.PluginTester/ProjectFile.cs` — evaluate the identity
- `tools/MeshWeaver.PluginTester/ProjectBuild.cs` — synthesize it into the compilation
- `tools/MeshWeaver.PluginTester/Program.cs` — `-p:Name=Value`
- `tools/MeshWeaver.PluginTester/README.md`
- `test/MeshWeaver.PluginTester.Test/ProjectAssemblyIdentityTest.cs` *(new, 23 cases)*

The clean diff is [`integration/core-batch...fix/build-project-assembly-version`](https://github.com/Systemorph/MeshWeaver/compare/integration/core-batch...fix/build-project-assembly-version).
Once the batch merges, this diff collapses to those five files on its own; no rebase needed.

**Not merged, no auto-merge armed**, as requested.


rbuergi:
## Removed from the merge queue — `CLEAN` here, `UNMERGEABLE` there, and the difference is real

This PR reads **CLEAN** on its own page and came back **UNMERGEABLE** from the queue. That is not a glitch, and re-queueing it will reproduce it exactly.

**Why.** This branch descends from `integration/core-batch` (#2853) and still carries its merge commits:

```
Merge remote-tracking branch 'origin/docs/tests-carry-no-dependencies'  -> #2848  MERGED
Merge remote-tracking branch 'origin/fix/gating-settle-stall'          -> #2849  MERGED
Merge remote-tracking branch 'origin/fix/cascade-build-silent-fatal'   -> #2852  MERGED
Merge remote-tracking branch 'origin/feat/cli-build-project'           -> #2850  in queue
```

Three of those landed on `main` as **squash** merges. So this branch reintroduces their content as separate commits, and the queue's speculative merge — which builds this PR on top of `main@now` — sees it arriving twice. **The PR page cannot show this**, because it compares against a base where the duplication does not exist. That gap between the two answers is the whole failure mode, and an ejection outranks the page.

It was also occupying a queue slot and would be ejected each time the stack reached it, costing everyone a cycle — hence removing it rather than leaving it to fail repeatedly.

## What it needs, and when

It is genuinely stacked on **#2850**, which is `MERGEABLE` in the queue right now. Once that lands:

```bash
git fetch origin main
git reset --hard origin/main
git cherry-pick <this PR's own commits only>   # NOT the integration/core-batch merges
```

That drops the stale batch ancestry entirely and produces a branch whose diff is only what this PR contributes. It cannot be done correctly before #2850 is on main, because this work genuinely builds on it.

I have not done that reset myself: it rewrites your branch, and the "own commits only" selection is a judgement about your work rather than a mechanical merge. Happy to do it once #2850 lands if you would rather — say the word.

*(#2853 itself is now closed — three of its four members landed individually, so the batch no longer saves a cycle. The main-merge and conflict resolution already pushed here stay valid; only the stale ancestry needs dropping.)*

rbuergi:
Superseded by **#2892**, now merged as `9fb74fd9`. This PR's unique commits are on main byte-for-byte (cherry-picked and union-merged with the Razor work); only this branch's history — the `integration/core-batch` ancestry that made it permanently `UNMERGEABLE` in the queue — was dropped. The arity-21 fix the binary-compat gate demanded (`AssemblyInfo` as an init property rather than a 22nd positional parameter) is included; this PR's own green had attested to a stale base where that break was invisible.

PR #2856, opened 2026-08-30
feat(image): the container compiles Razor — the SDK-free build covers Blazor

rbuergi:
> **Stacks on #2850** (`feat/cli-build-project`, the SDK-free `memex build project`). This PR is
> based on that branch and targets it, so the diff here is only the Razor work; GitHub retargets it
> to `main` when #2850 lands. **Do not merge this into the feature branch by hand** — a stacked
> sub-PR that merges into its parent is merged and *not delivered*.

Razor was the **biggest single gap** in #2850's first no-SDK sweep of `MeshWeaver.Plugins/src`:
**15 of 42 failures were CS0115**, more than any other category. The cause was one missing file, not
a missing feature — **Razor compilation in the .NET SDK is a Roslyn source generator**
(`Microsoft.CodeAnalysis.Razor.Compiler`) that turns each `.razor` into the partial class carrying
its `BuildRenderTree` override. A runtime image ships no SDK, so every component compiled to a class
with nothing to override.

The image now carries the generator in `razor-generators/<rid>/` beside the builder, and
`build-project` finds and runs it automatically for any project whose `Sdk` processes Razor items.

## The dependency closure, established rather than assumed

The brief named three assemblies. **Two are needed; the third is not.**

| assembly | verdict |
|---|---|
| `Microsoft.CodeAnalysis.Razor.Compiler` | needed — carries `RazorSourceGenerator` |
| `Microsoft.AspNetCore.Razor.Utilities.Shared` | needed — its one private dependency |
| `Microsoft.Extensions.ObjectPool` | **not referenced at all** by this compiler build |

**How it was established.** Read the compiler's own `AssemblyRef` table (`netstandard`,
`Microsoft.CodeAnalysis` 5.9.0.0, `Microsoft.CodeAnalysis.CSharp` 5.9.0.0,
`System.Collections.Immutable`, `System.Memory`, `System.Buffers`,
`Microsoft.AspNetCore.Razor.Utilities.Shared` — no ObjectPool), then **proved it by deleting**:
with `Utilities.Shared` removed the compiler assembly still loads and its types still enumerate, and
every call into it throws — the exact "generator silently produces nothing" shape this design exists
to prevent. `Microsoft.Extensions.ObjectPool` is referenced by the older *NuGet* build of the Razor
compiler (`10.0.0-preview.25277.114`), which is presumably where the belief came from.

## Two traps that fail at run time and nowhere else

🚨 **The generator is built against the SDK's Roslyn, not the image's.** SDK 10.0.400's copy wants
`Microsoft.CodeAnalysis` **5.9.0.0**; the image carries this repo's pin, **5.6.0**. The default load
context binds by name *and refuses a lower version*, so a plain `Assembly.LoadFrom` fails with
`Could not load … Version=5.9.0.0` — and `SourceGeneratorLoader` correctly reads that as "not a
usable generator" and moves on, which is how a missing Razor compiler ends up indistinguishable from
a `.razor` file nobody asked to compile. Generators therefore load into a context that binds **every
assembly the host already has to the host's copy, version ignored** (the same thing Roslyn's own
analyzer loader does). That is also what keeps `ISourceGenerator` **one** type: a second Roslyn in
that context would hand the generator a different interface than the driver expects.

🚨 **The generator is ReadyToRun-compiled for the SDK's own RID.** Measured on SDK 10.0.400, the same
`Microsoft.CodeAnalysis.Razor.Compiler.dll` carries PE machine **`0xFD1D` on linux-x64**, **`0xD11D`
on linux-arm64** and **`0xEC20` on osx-arm64** (the target machine XOR'd with the OS's R2R marker),
and the wrong one throws `BadImageFormatException`. `mw-plugin-test` publishes **both** architectures
from **one x64 host**, so "copy the build machine's SDK" would have shipped an arm64 image that
cannot compile a single Blazor project. CD now stages one directory per RID — the arm64 copy read out
of `mcr.microsoft.com/dotnet/sdk:<the runner's own SDK version>` with `docker create` + `docker cp`,
so no emulation runs and there is no second pin to drift.

## The `.razor` item glob, and what is refused

`ProjectFile` now evaluates the Razor SDK's default items (`Content Include="**\*.razor"` /
`"**\*.cshtml"`, promoted to `RazorComponent`/`RazorGenerate`) plus the project's own
`Content`/`RazorComponent`/`RazorGenerate` `Include`/`Remove`, and assigns each the `TargetPath`
`AssignTargetPath` would have (relative to the project directory — that is what decides the generated
type's namespace, and getting it wrong compiles into a namespace nothing references).

**Nothing is skipped in silence.** Four named failures, all tested:

| refusal | why |
|---|---|
| Razor files + no generator | says what the CS0115 wall *would* have been, instead of printing it — **one** error, not a wall |
| generator ran, emitted nothing | not a compile error; a generator that did not recognise its input |
| `*.razor.css` CSS isolation | the `b-…` scope comes from the SDK's `ComputeCssScope`/`ApplyCssScopes` tasks; this builder runs no MSBuild task, so the components would compile without their scope attributes. `--accept razor-css-scope` builds them anyway. Reproducing the scope hash from memory is the guess this evaluator exists to avoid |
| `.razor` under a non-Razor `Sdk` | the SDK ignores them too — but this says so. `--accept razor-not-compiled` |

## Measured, 2026-08-31 — 10 of the 11 Razor projects

Every `Microsoft.NET.Sdk.Razor` project in `MeshWeaver.Plugins/src`, built inside
`memex-portal-ai@sha256:6f38db08…` with this branch's builder mounted:

| | count | |
|---|---|---|
| **green, no extra help** | **7** | `MeshWeaver.Blazor` (**31 `.cs` + 42 `.razor`**, 0 warnings under warnings-as-errors), `.EntityViews`, `.Graph`, `.Analysis`, `.OpenStreetMap`, `.GoogleMaps`, `.AppleMaps` |
| green with `--generators` | +2 | `.Views`, `.Portal` — blocked by `[GeneratedRegex]` in `MeshWeaver.Markdown.Collaboration`, **not** by Razor |
| green with `--extra-refs` | +1 | `.Radzen` — `Radzen.Blazor` is an additional library |
| still red | 1 | `Memex.Portal.Gui` — a transitive `MeshWeaver.Hosting.Grpc` carries `<Protobuf>`; protoc is a build task, not a compile |

**Razor blocks none of them.** The two remaining categories are the pre-existing ones.

⚠️ The sweep ran on **`linux/arm64`**, natively on an Apple Silicon host. The `linux/amd64` runs on
that host crash under Rosetta — and the **control proves it is the emulator, not this change**: the
same emulated container dies with `AccessViolationException` building `MeshWeaver.Speech`, a project
with no Razor in it that #2850 verified green. CI runs on real x64 and is the x64 evidence.

## The framework build identity is untouched

No `ProjectReference` was added to `tools/MeshWeaver.PluginTester` — whose closure **is** the
framework identity. The assemblies ride along as `Content` with `CopyToPublishDirectory` and are
loaded at run time. `FrameworkBuildIdentityTest` (23 cases) still passes. `OptionsFingerprint` and
`EmitPipeline.CreateCompilationOptions` are not touched either; `MeshWeaver.Compiler` is not
modified at all by this PR.

## Tests — 13 new, EXECUTED

`dotnet test test/MeshWeaver.PluginTester.Test -c Release` → **Passed! Failed: 0, Passed: 188**
(175 before, no regressions).

- the shipped generator is found beside the builder and **loads** against this repo's Roslyn, and the
  loaded generator is `RazorSourceGenerator`;
- the staged closure is the compiler + its one private dependency, in a **per-RID** directory, with a
  provenance manifest;
- a **real** `.razor` (with `@inherits`, markup and `@code`) compiles and the emitted assembly's
  component type is **loaded** and its generated `BuildRenderTree` override found on it;
- a sub-directory component lands in the namespace the `TargetPath` dictates;
- `_Imports.razor` contributes its usings to a sibling (it would not compile otherwise);
- **the failure path names the missing generator** and reports **one** error rather than CS0115 noise;
- scoped CSS, and `.razor` under a plain `Sdk`, are each refused by name and build under their
  `--accept` token;
- `Content Remove` takes a component out of the build; `--razor-generators` **replaces** the search
  path rather than heading it (no silent fallback to the image's own copy);
- a missing directory, and a directory of non-generator assemblies, are each a named failure.

## Docs

`Doc/Architecture/InMeshBuildAndTest.md` gains a *"The container compiles Razor"* section — the
measured closure, both traps, the refusals and the sweep table — and the two READMEs
(`MeshWeaver.Cli`, `MeshWeaver.PluginTester`) are updated.

### DIFF
--- a/.github/workflows/main-cd.yml
+++ b/.github/workflows/main-cd.yml
@@ -1055,12 +1055,55 @@ jobs:
       # csproj (project-local — a global -p:RuntimeIdentifiers breaks referenced libs, see above).
       # `latest` (the tag the plugins repo pins as MW_TEST_IMAGE) and the GHCR mirror both moved
       # to `promote`: they are consumer-visible, so they must not appear until every leg is green.
+      # 🚨 THE RAZOR COMPILER, STAGED FOR BOTH IMAGE ARCHITECTURES.
+      # `mw-plugin-test build-project` compiles .razor/.cshtml, and Razor compilation in the .NET
+      # SDK is a Roslyn SOURCE GENERATOR. This image ships the runtime and no SDK, so the generator
+      # has to travel with the builder — and it CANNOT be copied once: the SDK crossgens it for the
+      # SDK's own RID, so the same 10.0.400 file carries PE machine 0xFD1D on linux-x64 and 0xD11D
+      # on linux-arm64, and the wrong one throws BadImageFormatException at run time and nowhere
+      # else. This image publishes BOTH architectures from ONE x64 host, so both copies are staged
+      # and the builder picks the one for the RID it is running on.
+      #
+      # The arm64 copy comes from the dotnet/sdk image of the EXACT SAME SDK VERSION the runner is
+      # using — one source of truth, no second pin to drift — and `docker create` + `docker cp`
+      # reads it out without ever executing an arm64 binary, so no qemu is needed.
+      - name: Stage the Razor source generator for linux-x64 and linux-arm64
+        id: razor
+        run: |
+          set -euo pipefail
+          test "$(uname -m)" = "x86_64" \
+            || { echo "::error::this step assumes an x64 runner (it takes the host SDK's copy as the linux-x64 one)"; exit 1; }
+          SDK_VERSION="$(dotnet --version)"
+          SDK_ROOT="$(dotnet --list-sdks | sed -n "s|^${SDK_VERSION} \[\(.*\)\]$|\1|p" | head -1)"
+          HOST_SOURCE="$SDK_ROOT/$SDK_VERSION/Sdks/Microsoft.NET.Sdk.Razor/source-generators"
+          ROOT="$RUNNER_TEMP/razor-generators"
+          echo "SDK $SDK_VERSION; host source-generators: $HOST_SOURCE"
+
+          mkdir -p "$ROOT/linux-x64" "$ROOT/linux-arm64"
+          cp "$HOST_SOURCE"/*.dll "$ROOT/linux-x64/"
+
+          IMAGE="mcr.microsoft.com/dotnet/sdk:$SDK_VERSION"
+          docker pull --platform linux/arm64 "$IMAGE"
+          CID="$(docker create --platform linux/arm64 "$IMAGE")"
+          trap 'docker rm -f "$CID" >/dev/null 2>&1 || true' EXIT
+          docker cp "$CID:/usr/share/dotnet/sdk/$SDK_VERSION/Sdks/Microsoft.NET.Sdk.Razor/source-generators/." \
+            "$ROOT/linux-arm64/"
+
+          # 🚨 Fail closed. An empty or half-staged RID ships an image whose Razor builds fail on
+          # exactly one architecture — which nothing in this pipeline would notice.
+          for rid in linux-x64 linux-arm64; do
+            test -f "$ROOT/$rid/Microsoft.CodeAnalysis.Razor.Compiler.dll" \
+              || { echo "::error::no Razor compiler staged for $rid"; exit 1; }
+            echo "$rid: $(ls "$ROOT/$rid" | tr '\n' ' ')"
+          done
+          echo "root=$ROOT" >> "$GITHUB_OUTPUT"
       - name: Build + push mw-plugin-test (staging tag, ACR)
         run: >
           dotnet publish tools/MeshWeaver.PluginTester/MeshWeaver.PluginTester.csproj
           -c Release --no-self-contained -t:PublishContainer -p:PublishProfile=
           -p:ContainerRuntimeIdentifiers='"linux-x64;linux-arm64"'
           -p:CIRun=true
+          -p:MeshWeaverRazorGeneratorRoot=${{ steps.razor.outputs.root }}
           -p:ContainerRegistry=${{ env.ACR }}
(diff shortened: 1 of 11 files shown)


rbuergi:
## I merged this, and I should say plainly that I did not check its base first

Merged at 04:51:25Z by an automated drain of the green PR backlog. **This PR's base is `feat/cli-build-project` (the branch of #2850), not `main`** — so the merge landed here into that branch, which is what the PR declared it would do, but *not* what I was doing when I ran it: I was draining PRs to main and treated this as one of them.

The check I ran before merging was `mergeStateStatus == CLEAN` and `Consolidate test results == SUCCESS`. Both were true. **Neither says anything about which branch the merge targets**, and I did not look.

## Consequence

Small, and not a loss: this is a deliberately stacked PR, so its content was always going to reach `main` via #2850. It is there now instead of waiting, and #2850's branch has moved without its owner asking.

What it is *not* is progress toward main — `c700412783f3` is `diverged` from `main` (ahead 3, behind 10), so nothing here has shipped.

## If this was not what you wanted

Say so and I will revert the merge on `feat/cli-build-project`. I have deliberately not done anything further to that branch.

Every other PR I queued in the same batch (#2848, #2849, #2852, and my own #2867/#2868/#2872/#2873) has `base=main`, verified explicitly after this.

PR #2860, opened 2026-08-31
feat(build-project): embedded resources, with the SDK's manifest names

rbuergi:
> **Cut from `origin/integration/core-batch`, targeting `main`.** #2850 (`feat/cli-build-project`) introduced this builder and is not on `main` yet — it sits inside batch PR #2853, which is blocked on an unrelated regression. This branch needs #2850's three files to work against, so it is based on the batch and **must land after it**. Nothing here was pushed to #2850's branch or to the batch branch.

`<EmbeddedResource>` items are **embedded now**, each under the manifest name the SDK's own `CreateCSharpManifestResourceName` would have produced.

Measured over the 54 non-test projects in `MeshWeaver.Plugins/src`, this was the single biggest coverage gap: **19 failed on nothing else**, most on one line — `<EmbeddedResource Include="Data\**\*.md">`.

## Why #2850 refused it, and why that was right

Its PR body says a wrong manifest-resource name *"is exactly the silent difference the design exists to prevent"*. That judgement was correct and this PR does not relax it. **A wrong resource name fails nothing:** the assembly compiles, the emit is verified, the DLL loads — and `Assembly.GetManifestResourceStream(name)` returns `null` at run time, in another process, later. Core's own `MeshWeaver.Messaging.Hub.csproj` already carries a comment describing that exact outcome for its `strings.*.json`: *"the build succeeds, the main assembly carries ZERO manifest resources, every lookup falls through to the key-fallback path, and the UI renders raw `chat.new` tokens."*

So **every rule here was established by measurement, not recall** — 15 probe projects built with the real .NET SDK, the names read back out of the emitted PE's `ManifestResource` table. Where a rule could not be matched, the construct is **refused by name**.

## The rules, and how each was measured

The SDK's pipeline, reproduced step for step: `AssignTargetPath` assigns `%(TargetPath)` (Microsoft.Common.CurrentVersion.targets says so out loud — *"AssignTargetPath generates TargetPath metadata that is consumed by CreateManifestResourceNames target for manifest name generation"*), `AssignCulture` splits culture-carrying items off towards satellites, `CreateCSharpManifestResourceName` turns the survivors into `$(RootNamespace).<mangled dir>.<file name>`.

| rule | measured |
|---|---|
| the DIRECTORY is mangled, the FILE NAME is **not** | `Data\with-dash\Three.md` → `…Data.with_dash.Three.md`; `Weird-File.Name.md` keeps hyphen **and** dot |
| a leading digit is PREFIXED, not replaced | `9digits` → `_9digits` |
| a dot in a directory is a SEPARATOR | `Dot.9Dir` → `Dot._9Dir` — each half mangled alone |
| a segment reducing to one `_` DOUBLES | `--` and `_` both → `__`; a project with both sibling dirs fails the real SDK with `CS1508` — that failure *is* the measurement |
| `$(RootNamespace)` defaults to the PROJECT name | not `$(AssemblyName)` — proven with a project whose two differ; empty ⇒ no prefix at all |
| `%(TargetPath)` > `%(Link)` > item spec | a `Link` renames a file already inside the project |
| a file OUTSIDE the project loses its directory | `..\shared\Shared.md` → `<ns>.Shared.md` — no `..`, no `shared` |
| resources are emitted PUBLIC | `ManifestResourceAttributes.Public`, every one |
| the SDK default glob is `**/*.resx` | reproduced, so a stray `.resx` is a refusal not a silent omission |
| a missing LITERAL include is `CS1566`; an empty glob is legal | measured both ways |
| `.resources` needs no special case | the SDK's strip-and-re-append branch is arithmetically identical, and mangles the directory too |

## Refused by name, each with its own `--accept` token

| construct | why a name cannot be promised |
|---|---|
| `.resx` / `.restext` | the NAME is reproducible; the CONTENT needs resgen. Embedding the XML under the `.resources` name compiles green and throws at run time. **Refusing loudly is the honest outcome** |
| a CULTURE in the file name | the SDK routes it to a SATELLITE assembly and out of the main one — and an explicit `LogicalName` does **not** rescue it (measured). `WithCulture="false"` is the project-side fix and needs no acceptance |
| `DependentUpon` | the name becomes the first CLASS in that file, fully qualified, extracted by MSBuild's own tokenizer — which skips `struct`/`interface`/`enum`, takes `record`, drops generic arity |
| `ManifestResourceName` | makes the SDK skip its own naming task, which also sets `%(LogicalName)`, so csc falls back to the bare file name — the metadata does not do what it appears to |
| an outside-the-project GLOB | this evaluator expands from the project down, so it matched **nothing**, silently — an empty glob is legal |
| `LinkBase` outside the project | synthesizes a `%(Link)` this evaluator does not compute. Ignored *inside* the project, so only the outside case is refused |
| the build's OWN OUTPUT | `Include="bin\…\$(AssemblyName).xml"` builds green under the real SDK from clean (csc writes `/doc:` and reads `/resource:` in one invocation); named and skipped rather than failing the project |

Under invariant globalization the runtime reports no predefined culture even for `de`, so the culture question cannot be answered — the capability is **probed**, and any dotted-basename resource is refused rather than guessed.

## Three findings worth calling out

1. **An explicit `%(Culture)` beats `%(WithCulture)='false'`** — the opposite of what the names suggest. My first implementation had the order backwards and would have embedded, under a name the SDK never produces, a resource the SDK puts in a satellite.
2. **`$(AssemblyName)` and `$(RootNamespace)` were never seeded**, so `$(AssemblyName)` expanded to the empty string. `MeshWeaver.Northwind.Domain` reads it in an `<EmbeddedResource Include="bin\$(Configuration)\$(TargetFramework)\$(AssemblyName).xml">`, which resolved to a file called `.xml`. Both now default to the project name, exactly where `Microsoft.NET.Sdk.props` sets them.
3. **A glob reaching outside the project silently matched nothing.** The same is true of `Compile` globs today (pre-existing, not touched here) — worth a follow-up.

## Verified by RUNNING

- **229 tests green** (34 new; the suite was 225). `MeshWeaver.Graph.Test` **1602 green** — the `EmitPipeline` consumers are unaffected.
- **Every resource assertion reads the emitted PE**, never this repo's source.
- **Parity test**: one probe project carrying every hard case is built by the **real SDK** and by `build-project`, and the manifest name sets must be identical. It does **not** skip when the SDK is missing — a parity test that passes because it could not find its oracle is worse than none.
- **Mutation proofs** — three, each reverted: dropping the leading-digit prefix rule → **5 tests red incl. parity**; mangling the file name too → **3 red**; disabling culture detection → **4 red**.
- **Real-world parity, end to end in the container**: `MeshWeaver.Northwind.Model` (11 glob-matched CSVs) built by the real SDK and by `build-project` inside the combined tester+portal image — **all 11 names AND their contents identical, byte for byte** (SHA-256 per resource).

## The sweep: 9 → 10 of 54, and why that is the honest number

Both columns from the **same** image (`memex-portal-ai@sha256:6f38db08…`, the pin in `MeshWeaver.Plugins/.github/workflows/ci.yml`), `--accept targets`, no `--extra-refs`. The "before" column was produced by building the **pre-change** builder from `origin/integration/core-batch` into its own image and sweeping it — it reproduces the reported baseline of **9 exactly**.

| | before | after |
|---|---|---|
| **green** | **9** | **10** |
| `<EmbeddedResource>` refused | **19** | **0** |
| Razor/Blazor (CS0115) | 12 | 15 |
| source generators (CS8795) | 3 | **14** |
| additional libraries | 5 | 8 |
| `<Protobuf>` | 2 | 3 |
| `<Import>` above the mount | 2 | 2 |
| Aspire `<Sdk>` element | 1 | 1 |
| other (pre-existing errors) | 1 | 1 |

🚨 **This is not the 9 → ~24 the gap size suggested, and the reason matters.** A refusal **stops the load**, so it MASKS every gap behind it. Those 19 projects were never "19 projects away from green" — they were 19 projects about which nothing further was known. Closing the gap turned one green and revealed the real blocker for the other eighteen: **ten want SDK source generators**, three want Razor, three want an additional library, one wants protoc. That is why the old sweep counted 3 source-generator failures when the true figure is 14.

**In a builder designed to refuse rather than guess, a failure ranking is a ranking of FIRST refusals.** Removing the top entry redistributes its count; it does not add it to the green column. Razor (#2856) is now the largest remaining gap at 15, with source generators right behind at 14.

**No project regressed**, and none fails on an embedded resource any more.

## Notes for review

- **`EmitPipeline` gains an OVERLOAD, not an optional parameter** — adding a parameter replaces a method's signature, the same binary break `scripts/check-record-signatures.py` refuses for records. The four-argument entry point is byte-identical for `MeshNodeCompilationService` and `NodeSetCompiler`, which pass no resources and never will.
- **`CreateCompilationOptions` and `OptionsFingerprint` are untouched.**
- **New model members are `init` properties, not primary-constructor parameters** — following #2856's discipline. `check-record-signatures.py` passes.
- **Stacked-PR contact points.** #2855 appends an `AssemblyInfo` positional parameter to `ProjectFile.Model` and #2856 adds `init` properties to it; this PR only adds `init` properties, so the conflicts should be mechanical. All three also touch `ProjectFile.EvaluateItem` and `ProjectBuild`'s compile path.
- `ProjectFileTest.AnEmbeddedResourceFailsUntilAcknowledged_…` asserted the old refusal and is replaced by one asserting the resolved SDK manifest name.


rbuergi:
## Removed from the merge queue — `CLEAN` here, `UNMERGEABLE` there, and the difference is real

This PR reads **CLEAN** on its own page and came back **UNMERGEABLE** from the queue. That is not a glitch, and re-queueing it will reproduce it exactly.

**Why.** This branch descends from `integration/core-batch` (#2853) and still carries its merge commits:

```
Merge remote-tracking branch 'origin/docs/tests-carry-no-dependencies'  -> #2848  MERGED
Merge remote-tracking branch 'origin/fix/gating-settle-stall'          -> #2849  MERGED
Merge remote-tracking branch 'origin/fix/cascade-build-silent-fatal'   -> #2852  MERGED
Merge remote-tracking branch 'origin/feat/cli-build-project'           -> #2850  in queue
```

Three of those landed on `main` as **squash** merges. So this branch reintroduces their content as separate commits, and the queue's speculative merge — which builds this PR on top of `main@now` — sees it arriving twice. **The PR page cannot show this**, because it compares against a base where the duplication does not exist. That gap between the two answers is the whole failure mode, and an ejection outranks the page.

It was also occupying a queue slot and would be ejected each time the stack reached it, costing everyone a cycle — hence removing it rather than leaving it to fail repeatedly.

## What it needs, and when

It is genuinely stacked on **#2850**, which is `MERGEABLE` in the queue right now. Once that lands:

```bash
git fetch origin main
git reset --hard origin/main
git cherry-pick <this PR's own commits only>   # NOT the integration/core-batch merges
```

That drops the stale batch ancestry entirely and produces a branch whose diff is only what this PR contributes. It cannot be done correctly before #2850 is on main, because this work genuinely builds on it.

I have not done that reset myself: it rewrites your branch, and the "own commits only" selection is a judgement about your work rather than a mechanical merge. Happy to do it once #2850 lands if you would rather — say the word.

*(#2853 itself is now closed — three of its four members landed individually, so the batch no longer saves a cycle. The main-merge and conflict resolution already pushed here stay valid; only the stale ancestry needs dropping.)*

rbuergi:
Superseded by **#2892**, now merged as `9fb74fd9`. This PR's unique commits are on main byte-for-byte (cherry-picked and union-merged with the Razor work); only this branch's history — the `integration/core-batch` ancestry that made it permanently `UNMERGEABLE` in the queue — was dropped. The arity-21 fix the binary-compat gate demanded (`AssemblyInfo` as an init property rather than a 22nd positional parameter) is included; this PR's own green had attested to a stale base where that break was invisible.

PR #2863, opened 2026-08-31
fix(cd): the three script-less ACR login sites get a working tree — promote is broken on main

rbuergi:
🚨 **`main` is currently broken for delivery.** CD run [33344774738](https://github.com/Systemorph/MeshWeaver/actions/runs/33344774738) on `74a13086a` built and pushed **all three images**, then died in `promote`:

```
Run bash .github/scripts/acr-login.sh
bash: .github/scripts/acr-login.sh: No such file or directory
##[error]Process completed with exit code 127
```

The set was pushed to staging and **never tagged**. `Verify every image shipped` then correctly reported all three MISSING, and every self-updating install stayed on the previous image.

## What happened

#2857 replaced every inline `az acr login` with the shared `acr-login.sh`, closing a real hole — the bare login sat **outside** the publish retry, and a `context deadline exceeded` on it blocked delivery at 23:43. **That reasoning is right and stays.**

But **a script needs a working tree and the inline one-liner did not**, and three of the seven call sites live in jobs that never check out.

## Audited all seven, not just the one that failed

| job | checkout | |
|---|---|---|
| `portal-image` | ✅ | |
| `migration-image` | ✅ | |
| `plugin-test-image` | ✅ | |
| **`promote`** | ❌ | ← the observed failure |
| **`notify-platform-update`** | ❌ | two call sites |
| **`plugins-bake-image`** | ❌ | |

The other two would have failed **identically** the first time they were reached — on paths that do not run every time, so they would have surfaced much later and looked like a new problem.

## The fix

A **sparse** checkout of `.github/scripts` only, added to each of the three. Negligible cost, and the login keeps **one implementation** across all seven sites rather than reverting some to an unretried inline call — which would reopen exactly what #2857 closed.

## 🚨 Why #2857's own green could not have caught this

`promote`, `notify-platform-update` and `plugins-bake-image` **only run on a publishing CD run, never on a pull request.** So #2857's checks attested to code paths that never executed.

That is worth naming as a class: **a job that only runs in production is not covered by anything that gates the change to it.** It is the same shape as a PR's green attesting only to its merge base, one level further out — and it is why the audit above matters more than the single fix.

## Verification

- `yaml.safe_load` parses the workflow (15 jobs) — a malformed workflow is a startup failure that publishes nothing, so this is not optional.
- Re-audit: every `acr-login.sh` call site is now in a job carrying a checkout.

Credit to the #2857 author for the diagnosis and the shared-script direction; this is the missing half, not a disagreement.

### DIFF
--- a/.github/workflows/main-cd.yml
+++ b/.github/workflows/main-cd.yml
@@ -1106,6 +1106,15 @@ jobs:
       V_MIGRATION: ${{ needs.migration-image.outputs.version }}
       V_PLUGIN: ${{ needs.plugin-test-image.outputs.version }}
     steps:
+      # 🚨 The ACR login below is a SCRIPT now (#2857), so this job needs a working tree — the
+      # inline `az acr login` it replaced needed none. Sparse: only the scripts directory, so this
+      # costs nothing. Without it the step dies `bash: .github/scripts/acr-login.sh: No such file
+      # or directory` with exit 127, which is what broke `promote` on 2026-08-31 00:54 and left a
+      # published image set untagged.
+      - uses: actions/checkout@v7
+        with:
+          sparse-checkout: .github/scripts
+          sparse-checkout-cone-mode: false
       - name: Ensure Azure CLI
         run: command -v az >/dev/null || curl -sL https://aka.ms/InstallAzureCLIDeb | sudo bash
       - name: Azure login (OIDC)
@@ -1844,6 +1853,15 @@ jobs:
     runs-on: ubuntu-latest
     timeout-minutes: 10
     steps:
+      # 🚨 The ACR login below is a SCRIPT now (#2857), so this job needs a working tree — the
+      # inline `az acr login` it replaced needed none. Sparse: only the scripts directory, so this
+      # costs nothing. Without it the step dies `bash: .github/scripts/acr-login.sh: No such file
+      # or directory` with exit 127, which is what broke `promote` on 2026-08-31 00:54 and left a
+      # published image set untagged.
+      - uses: actions/checkout@v7
+        with:
+          sparse-checkout: .github/scripts
+          sparse-checkout-cone-mode: false
       - name: Ensure Azure CLI
         run: command -v az >/dev/null || curl -sL https://aka.ms/InstallAzureCLIDeb | sudo bash
       - name: Azure login (OIDC)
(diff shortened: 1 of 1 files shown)


rbuergi:
## Closing in favour of #2864 — same fix, plus the guard that stops it recurring

#2864 carries this fix and adds what this PR does not: a **static check** in `check-workflow-shell.py` (`script-needs-tree`) that fails any job running a `.github/scripts/` file without an `actions/checkout` step. Self-tested both directions, and a full sweep found no other workflow hiding the shape.

That is strictly better, and for the reason that matters here: **it closes the class statically, on a check that CAN run on a pull request.** The whole problem was that `promote`, `notify-platform-update` and `plugins-bake-image` only execute on a publishing CD run, so nothing gating the change to them ever exercised them. A guard that runs on every PR is the only honest answer to "but it was green".

Two PRs touching `main-cd.yml` would also conflict in the merge queue — the mutual-ejection shape that cost core `main` two and a half hours tonight. One entry is the right number.

**Both audits independently found the same three jobs** (`promote`, `notify-platform-update` — the build-fact job, which would have broken every consumer pin update the first time it fired — and `plugins-bake-image`), which is a good cross-check on the count.

Nothing here is lost: the reasoning lives in #2864, and the class note — *a job that only runs in production is not covered by anything that gates the change to it* — is in its guard's docstring, which is a better home than a closed PR body.

rbuergi:
## Reopened — the guard cannot land before the fix

I closed this in favour of #2864 (same fix, plus a static `script-needs-tree` guard) on the grounds that theirs was strictly better. **That was premature, for a reason I had not considered:**

> **The guard cannot land before the fix — it would red its own gate on the very defect it detects.**

`script-needs-tree` fails any job running a `.github/scripts/` file without a checkout. Landed against current `main`, its first run finds `promote`, `notify-platform-update` and `plugins-bake-image` — three live findings — and goes red on itself. **The order is forced: fix first, guard second.**

So this PR lands the trunk fix (fix-only, so it clears checks fastest, and delivery is currently down), and #2864 rebases to keep only the guard. Not a disagreement about which is better — #2864 *is* better, and it is the half that stops this recurring. It simply cannot go first.

**Both audits independently found the same three jobs**, including `notify-platform-update` — the build-fact job, which would have broken every consumer pin update the first time it fired. Independent agreement on the count is worth more than either audit alone.

Reopened and unchanged: sparse checkout of `.github/scripts` in each of the three, `yaml.safe_load` clean at 15 jobs, re-audit showing every `acr-login.sh` call site now in a job that checks out.

rbuergi:
## Closing — #2864 landed with both halves

`#2864` merged, carrying the same three sparse checkouts **plus** the `script-needs-tree` guard that stops this recurring. This PR was kept open only as a fallback in case that one ejected; it did not, so there is nothing left for this to do.

**The guard is mutation-proven, which is the part worth recording.** On #2864's branch, deleting the `promote` job's checkout made the gate exit 1 and **name** `promote` + `acr-login.sh`; restoring it returned exit 0 / "0 live". So zero findings on the merged tree means *"the checkouts exist"*, not *"the scan never reached those jobs"* — the same distinction #2844 closed for `SourceScan`, and the one that makes a green guard worth anything.

**Both audits independently found the same three jobs** — `promote`, `notify-platform-update` (the build-fact job, which would have broken every consumer pin update the first time it fired), and `plugins-bake-image`. Agreement between two separately-derived counts is a better check on the number than either audit alone.

Nothing lost by closing: the fix and the audit are on main, and the class note — *a job that only runs in production is not covered by anything that gates the change to it* — lives in the guard's docstring, which is where the next person will meet it rather than in a closed PR body.

Issue #2471, opened 2026-08-27
A portal can serve STALE compiled code while reporting Ok/compiled — recycles, forced compile and the stale-build banner are all inert


rbuergi:
## The claim

A portal can serve **stale compiled code while reporting `Ok`/compiled**, and survive every remedy the product offers.

Observed on `memex` on 2026-08-26 after MeshWeaver.Plugins#732 merged (22:57Z):

- the deployed **source node** contained the change — `get` confirmed the new registration, **v6 at 23:04:54**;
- `get_diagnostics` reported **`Ok` / compiled**;
- but `ReinsuranceDemo/Walkthrough`'s in-node `Tests` area still rendered the **old 15-case suite** instead of 16;
- this survived **2 NodeType recycles, 4 instance recycles and a forced `compile`**, over **30+ minutes**;
- the **`$Banner` stale-build adornment was empty throughout** — the one adornment whose job is to say "you are looking at an old build" did not fire.

The sibling PR in the same batch (#731's Store half) landed cleanly at 22:45:59, so this is a **race, not a broken pipeline**.

## 🚨 Provenance — which parts are measured and by whom

Filing this because it was recorded only as a comment on a *content* repo's talk-prep issue (MeshWeaver.SocialMedia#75) and would be lost there. It is a platform defect and belongs here.

- **Measured by the session that hit it** (2026-08-26): every bullet above.
- **Re-verified by me now**: the node paths exist and are live — `ReinsuranceDemo` (`Store/Plugin`, v1698) and `ReinsuranceDemo/Walkthrough` (`Training/Tour`, v312, `lastModified 2026-08-26T19:25:00Z`).
- **NOT re-derived by me**: the 15-vs-16 case count and the recycle attempts. I am not restating those as my own observation.

## Why this is the worst class of defect in this codebase

Every remedy the product offers a user — recycle the NodeType, recycle the instance, press Compile — **reports success and changes nothing**, and the diagnostic that exists to reveal it says `Ok`. So:

- a merged, published, correct fix is **indistinguishable from one that is live**;
- the only honest check is **rendering the surface and counting what it produced**, which is exactly what a compile status is supposed to save you from;
- it silently invalidates "I fixed it and verified the compile is green" for every fix on this portal.

That is the same shape as the MCP write that reports `Patched:` and does not land (#2469), and as a cancelled CI run reading as a settled-red one (#2470) — a success signal that is not measuring the thing it names.

## What would settle it

[...]

PR #2866, opened 2026-08-31
fix(#2471): a stale bind is REFUSED at activation — recompile instead of serving bytes the type did not publish

rbuergi:
Closes the reopened half of #2471 with tonight's production evidence (Plugins e2e, 2-of-2 on `1dc5fc41`): an instance hub bound `Store/Plugin` bytes with MVID `810f4775…` while the type published `7f5df496…` — and from that single stale bind, the whole paid-fulfilment failure followed (family-wide `As<T>` case-3 degradation across StoreContent/TierContent/OrderContent, the fulfilment watcher never firing, Plugins #455's "completed without an outcome" blocking the #2802 rotation chain).

## Why the existing fix couldn't help here

The #2655 fix made the same-path mismatch **detectable** (the banner + `LogError` fired, on schedule) but deliberately **report-only** — its own comment: *"a recycle re-binds the same local copy and will not change it."* Correct reasoning, wrong conclusion for the bind moment: once bound, nothing downstream can fix it. But **at activation, before the bind, both MVIDs are already in hand** — and there a refusal is cheap.

## Change

At the enrichment seam, after `ResolveAssembly` + `ServedBuildIdentity.OfFile`, a mismatch with retry budget left routes into the **existing** `TriggerRecompileAndRetry` machinery — the same bounded path the bytes-missing sibling case has used all along. A fresh compile mints new bytes **and a new published MVID**, and the retry re-enriches against them; the store's own `Put` contract ("same version → same output") is exactly the assumption this state violates, so recompiling is the honest reconciliation. On budget exhaustion (`MaxRecompileAttempts`), the activation falls through to today's bind-and-banner — degraded but functional beats parked.

The decision is extracted pure (`ShouldRefuseStaleBind`) with the evidence rules pinned: refuses only on **evidence** (both MVIDs present and different — a legacy node without a published MVID never refuses; an unreadable local file never refuses) and only within budget.

## Tests

`ServedBuildIsVerifiableTest` + 5 new theory cases (refuse on mismatch-with-budget; bind on exhaustion / identical / no-evidence both ways). Graph suite: **1603/1603 passed**.

Diagnosis trail: Plugins #455 (comment 5472886155) · #2471 reopen comment with the MVID pair and blast radius.

https://claude.ai/code/session_01QoG4USP5fd8wsxCD7e9q4x

### DIFF
--- a/src/MeshWeaver.Graph/Configuration/NodeTypeEnrichmentHelpers.cs
+++ b/src/MeshWeaver.Graph/Configuration/NodeTypeEnrichmentHelpers.cs
@@ -1181,6 +1181,28 @@ internal static class NodeTypeEnrichmentHelpers
                     // publication-driven watcher stayed silent while memex served old code.
                     var boundMvid = ServedBuildIdentity.OfFile(localPath);
 
+                    // 🚨 #2471, the SAME-PATH half, healed where it is FIRST true. The path is a
+                    // store key resolved through a per-pod local cache, so the bytes behind it can
+                    // differ from the bytes the type published — and once BOUND, nothing downstream
+                    // can fix it: the banner is report-only ("a recycle re-binds the same local
+                    // copy"), and the served instance splits the type family across two collectible
+                    // assemblies, degrading every cross-hub read of its content (As<T> case 3 — the
+                    // 2026-08-31 paid-fulfilment outage: StoreContent, all TierContent and the
+                    // order degraded in one run, the fulfilment watcher never fired). So a
+                    // mismatch is REFUSED here, before the bind, by the same bounded machinery the
+                    // bytes-missing sibling uses: a fresh compile mints new bytes AND a new
+                    // published MVID, and the retry re-enriches against them. On exhaustion the
+                    // activation falls through to today's bind-and-banner — degraded but
+                    // functional beats parked.
+                    if (ShouldRefuseStaleBind(def.LatestAssemblyMvid, boundMvid, nodeType, recompileAttempts))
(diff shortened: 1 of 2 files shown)


rbuergi:
Core-side review (requested):

**The shape is right, and it is not a band-aid.** The refusal changes the violated state rather than retrying around it: a fresh compile mints new bytes AND a new published MVID, and — the load-bearing detail — a new VERSION, so the retry resolves a NEW store key and bypasses the poisoned per-pod cache entry entirely. That is why recompile beats store-cache invalidation here, and I would state it in exactly those terms: **invalidation re-fetches the SAME key, and the Put contract violation (same-version⇒same-bytes broken) means the authoritative bytes for that key are ambiguous — re-fetching cannot decide which bytes are right; minting a new key can.** Your instinct on the design question was correct.

Evidence-only refusal with both null-sides never refusing, budget-bounded, exhaustion falling to today's bind+banner (degraded beats parked) — all correct, and the pure `ShouldRefuseStaleBind` + theory is the right guard shape.

Two minor notes, neither blocking:
1. The theory row `(PublishedMvid, BoundMvid, 1, false)` hard-codes the budget boundary as the literal 1. If `MaxRecompileAttempts` is a `const`, reference it in the InlineData so a future budget change cannot silently turn this row into a still-under-budget case that asserts the wrong thing.
2. Worth one comment line at the refusal site: the recompile path is deliberately chosen over deleting the local cache entry + re-resolving — considered and rejected for the same-key-ambiguity reason above — so the next reader doesn't 'simplify' it into an invalidation.

Merge on green from my side.

Issue #2865, opened 2026-08-31
Ambient test-host hangs and crashes are the merge queue's residual risk — an unowned family that ejects entries at random


rbuergi:
Filing so this stops living in session chat. It was named and deliberately **not** absorbed by the session that found it, and it is currently **unowned**.

## What it is

After Systemorph/MeshWeaver#2842's dual-terminal-writer split-brain was root-caused and fixed (#2859), a **second, lower-rate family** remains. It ejects merge-queue entries at random and is unrelated to any particular PR:

| symptom | where |
|---|---|
| `HOST_CRASHED`, `exit=124` | `MeshWeaver.Hosting.Monolith.Test` (seen during the `NodeTypeCompileParkTest` window) |
| shard-3 crash | `MeshWeaver.FutuRe.Test` |
| flakes | Acme ColdCycle, GitSync Bidirectional, Pandas, ScopeTeardown |

Occurrences on 2026-08-30/31 include **22:17** and **22:32** on trees *without* Systemorph/MeshWeaver#2842, and the **first Systemorph/MeshWeaver#2859 queue entry**, which ejected on the Monolith hang before a clean second attempt.

## Why it needs an owner rather than tolerance

**In a merge queue, an ambient hang is not a flaky test — it is a throughput tax.** Every merge-group build contains every entry, so one hang **ejects an entry and restarts every remaining build**. A defect that would be a shrug in a PR run becomes the thing that stops `main` moving, and it is indistinguishable from a real regression until someone reads the `.trx`.

That is not hypothetical: tonight core `main` landed nothing for over two and a half hours, and the *first* explanation everyone reached for was load. It was not load. See [Merge Queue Mechanics](../Architecture/MergeQueueMechanics) (#2862) for why the obvious readings mislead.

## What is already excluded

- **Not Systemorph/MeshWeaver#2842.** These occurred on trees without it.
- **Not Systemorph/MeshWeaver#2859.** Its park suite passes 10/10 locally and the change is behaviourally inert on non-#2842 trees.
- **Not the queue's parallelism alone.** Members of this family have appeared on plain `pull_request` runs.

## Where to start

🚨 **`exit=124` is a TIMEOUT KILL, so there is no assertion and no stack** — the shard's cap fired and the host was killed. Per `/debug`, read the **duration** first and the **last `MESSAGE_FLOW:` tag** that fired, then ask what should have fired next. A multi-minute failure with no assertion text is a hang, never a wrong answer.

[...]

Issue #2533, opened 2026-08-27
FutuReAnalysisTest.EuropeRe_AnnualReport_EmbeddedCharts times out on a RedirectControl — new flake population, distinct from #2500


rbuergi:
One sample, recorded rather than re-run.

**Run** 33108923187, job 98648497799 (shard 3), branch `chore/2276-moved-test-suites-leave-core`, 2026-08-27 19:46Z.

```
MeshWeaver.FutuRe.Test.FutuReAnalysisTest.EuropeRe_AnnualReport_EmbeddedCharts_ShouldRenderViaPathResolution [FAIL]
Expected the observable to emit a value matching the predicate within 50s because KeyMetrics should
render at FutuRe/EuropeRe/Analysis (not the 'Area not found' placeholder and not the mid-compile
progress page), but it did not. Last of 6 emission(s) was:
RedirectControl { … Href = /FutuRe/EuropeRe/Analysis/KeyMetrics }
```

The area resolved to a **redirect to itself** and stayed there for the full 50 s — six emissions, none of them the control. The log carries a wall of `'$type':'X' is NOT registered in this (receiving) hub` warnings (`AnalysisContent`, `BusinessUnit`, `Section`, `Space`) — per AGENTS.md that class of warning means the value degrades to an untyped `JsonElement`, the view renders empty and reactive waits time out, which matches the symptom exactly. Whether the warnings are the cause or ambient noise for this suite is the thing to determine.

**Attribution work already done** (so the next reader does not repeat it):

- The branch's diff deletes three test projects and edits shard weights. **Shard 3's membership is byte-identical before and after** the weight change (verified by running `shard-assign.sh` both ways) — the diff cannot reach this test through co-tenancy, the only plausible mechanism.
- Not #2524 (`MeshWeaver.AI` leaving `ContentSurfaceAssemblies`): `CompileReferences.Default` derives from `TRUSTED_PLATFORM_ASSEMBLIES`, not from the identity list, so that change cannot alter what in-mesh content compiles against in a test host. `MeshWeaver.FutuRe.Test` also has zero AI references, and the branch predates #2524 anyway.
- **Distinct from #2500**, whose population is Orleans-specific (`PartitionRootActivationTest`, `OrleansScheduledPostTest`, `OrleansNodeChangePropagationTest`, `OrleansGrainTeardownStragglerTest`). This is a FutuRe layout-area render.

[...]

PR #2867, opened 2026-08-31
fix(#2836, #2787): a transient 503 is not a refusal — on the portal and in the gate

### LINKED ISSUE #2787: Platform promotion rolls the portal, and dependent repos' CI reading the plugin registry mid-roll fails on 503
## Promotion rolls the portal; a dependent repo's CI that needs the registry during that window fails

`MeshWeaver.Manufacturing` main, run [`33316402646`](https://github.com/Systemorph/MeshWeaver.Manufacturing/actions/runs/33316402646):

```
registry answered 503 for
https://memex.meshweaver.cloud/api/plugins/bundles/prebuilt/sa152a8f90e56acedee26b2ac44d7bfc3/plugins
 — refusing.
```

**The gate behaved correctly** — it refused rather than proceeding on an unreadable upstream, which is the right design. The problem is upstream of it.

## The chain, measured

| time (UTC) | event |
|---|---|
| 14:12:41 | core image `3.0.0-rc8.ci.6748` promoted (digest `c4b2bb5c7607`, tags `ci.6748` / `4213425` / `main`) |
| ~14:16 | Manufacturing CI reads the registry → **503** → refuses |
| 14:18:22 | a `memex-portal-deployment` pod starts (`memex-cloud`) |

Installs self-update **on the publication event**, so promoting the platform rolls the portal — and the registry endpoint is served by that portal. A dependent repo whose CI reads the registry inside the roll window gets a 503 through no fault of its own diff.

Verified the portal is healthy either side of the window: `/` → 200, the registry endpoint → 401 (auth required, i.e. serving), three pods ready with 0 restarts.

## Why this is worth fixing rather than re-running

It is **structural, not incidental**: every platform promotion creates the window, and the fleet has six repos whose CI reads that endpoint. Today three images were promoted (`ci.6691`, `ci.6714`, `ci.6748`), so the window opened three times. The more reliably promotion works, the more often this fires.

Two directions, not exclusive:

1. **Tolerate it at the consumer.** A 503 from the registry is transient by nature; the gate could retry with backoff and refuse only if it stays unreadable. Distinguish "unreadable right now" from "answered something wrong" — the gate already refuses correctly for the second.
2. **Close it at the source.** Three replicas roll one at a time, so a 503 suggests traffic reaching a pod before it can serve the registry route, or the roll not being surgical. Worth checking the readiness probe actually gates *that* route — `/api/og` returning 200 is not proof the registry endpoint is ready (the negative-control lesson from `api-og-is-the-negative-control-not-a-probe`).

## Not the same as #2778

#2778 is a NACK suppressed during hub teardown, inside the mesh. This is an HTTP 503 during a Kubernetes rollout, between repos. Filed separately so neither gets buried in the other.

Re-run requested on that run **because the input changed** (the portal is serving again), not to see whether it was a flake.

### LINKED ISSUE #2836: RegistryUpdateReconciler strands package reconciliation for the whole boot when the registry feed answers a transient 503
At boot, `RegistryUpdateReconciler` pulls the package feed of the Plugins registry source (https://memex.meshweaver.cloud) to reconcile the installed packages against it. The feed answered HTTP 503 — "Instance-key resolution is temporarily unavailable — retry shortly" — and after the built-in retry budget (4 attempts) the reconciler gave up for the remainder of the process lifetime.

**Probable cause** — confidence high on the mechanism, medium on the trigger. The registry's instance-key resolution was transiently unavailable at the moment of these boots; the reconciler's policy — a fixed retry budget at boot and nothing behind it, exactly as the log message itself states ("no poll behind this") — turns a server-side blip the error explicitly marks as retryable into a whole-boot stall of package reconciliation. A background re-poll with backoff would very likely have recovered within minutes. The registry-side availability question (why instance-key resolution was unavailable at all) is real but secondary; the client-side give-up is the actionable defect.

**Impact** — each occurrence leaves that portal pod with an unreconciled installed set: no package updates and no grant changes land until the next restart or until a human opens the catalog page. The portal itself stays up, so the staleness is silent. Low volume, but zero self-healing, and the exposure is growing rather than fixed: registry-consumer mode and un-keyed open registration shipped the same day (`Doc/WhatsNew/2026-08-30-a-local-install-consumes-the-cloud-registry`, `Doc/WhatsNew/2026-08-30-a-local-install-registers-without-a-key`), which puts instance-key resolution on the request path of every install's boot — any install that comes up during a brief registry hiccup now lands in this state.

**Where to look** — start at `src/MeshWeaver.PluginCatalog/RegistryPackageSource.cs`, `ListPackages` (line 88 in the captured stack), where the exhausted retry throws `InvalidOperationException`; the boot-time orchestration and the give-up decision live in `MeshWeaver.PluginCatalog.RegistryUpdateReconciler`. The likely fix is a background re-poll (or extending the retry budget with backoff for 503 responses that announce themselves as retryable), not anything about the key or grant — the error text is explicit that neither is in question.

No existing incident or issue found in the mesh for this component and symptom — not a duplicate.

---

### Evidence

| | |
|---|---|
| Fingerprint | `e21cccb65138f43b` |
| Category | `MeshWeaver.PluginCatalog.RegistryUpdateReconciler` |
| Severity | Error |
| Top frame | `MeshWeaver.PluginCatalog.RegistryPackageSource.<>c__DisplayClass15_0.<<ListPackages>b__0>d.MoveNext()` |
| Namespace | `memex` |
| Pods | `memex-portal-deployment-658bbfd875-xpf9w`, `memex-portal-deployment-8f976f958-j28fg`, `memex-portal-deployment-b74d87c5c-5qjxx` |
| Occurrences | 3 |
| First seen | 2026-08-30 12:04:04Z |
| Last seen | 2026-08-30 19:54:17Z |

```
2026-08-30 12:04:04Z memex-portal-deployment-658bbfd875-xpf9w fail: MeshWeaver.PluginCatalog.RegistryUpdateReconciler[0]
      [RegistryUpdate] could not read the package feed of Plugins (https://memex.meshweaver.cloud) after up to 4 attempt(s) — installed packages from it are NOT reconciled this boot, and there is no poll behind this: the next chance is the next restart or a human opening the catalog page.
      System.InvalidOperationException: Registry catalog list failed (503): {"error":"Instance-key resolution is temporarily unavailable — retry shortly. This is NOT a statement about your key or your grant."}
         at MeshWeaver.PluginCatalog.RegistryPackageSource.<>c__DisplayClass15_0.<<ListPackages>b__0>d.MoveNext() in /home/runner/work/MeshWeaver/MeshWeaver/src/MeshWeaver.PluginCatalog/RegistryPackageSource.cs:line 88
      --- End of stack trace from previous location ---
         at MeshWeaver.Mesh.Threading.IoPool.<>c__DisplayClass23_0`1.<<InvokeCore>b__0>d.MoveNext()
2026-08-30 15:02:00Z memex-portal-deployment-8f976f958-j28fg fail: MeshWeaver.PluginCatalog.RegistryUpdateReconciler[0]
      [RegistryUpdate] could not read the package feed of Plugins (https://memex.meshweaver.cloud) after up to 4 attempt(s) — installed packages from it are NOT reconciled this boot, and there is no poll behind this: the next chance is the next restart or a human opening the catalog page.
      System.InvalidOperationException: Registry catalog list failed (503): {"error":"Instance-key resolution is temporarily unavailable — retry shortly. This is NOT a statement about your key or your grant."}
         at MeshWeaver.PluginCatalog.RegistryPackageSource.<>c__DisplayClass15_0.<<ListPackages>b__0>d.MoveNext() in /home/runner/work/MeshWeaver/MeshWeaver/src/MeshWeaver.PluginCatalog/RegistryPackageSource.cs:line 88
      --- End of stack trace from previous location ---
         at MeshWeaver.Mesh.Threading.IoPool.<>c__DisplayClass23_0`1.<<InvokeCore>b__0>d.MoveNext()
2026-08-30 19:54:17Z memex-portal-deployment-b74d87c5c-5qjxx fail: MeshWeaver.PluginCatalog.RegistryUpdateReconciler[0]
      [RegistryUpdate] could not read the package feed of Plugins (https://memex.meshweaver.cloud) after up to 4 attempt(s) — installed packages from it are NOT reconciled this boot, and there is no poll behind this: the next chance is the next restart or a human opening the catalog page.
      System.InvalidOperationException: Registry catalog list failed (503): {"error":"Instance-key resolution is temporarily unavailable — retry shortly. This is NOT a statement about your key or your grant."}
         at MeshWeaver.PluginCatalog.RegistryPackageSource.<>c__DisplayClass15_0.<<ListPackages>b__0>d.MoveNext() in /home/runner/work/MeshWeaver/MeshWeaver/src/MeshWeaver.PluginCatalog/RegistryPackageSource.cs:line 88
      --- End of stack trace from previous location ---
         at MeshWeaver.Mesh.Threading.IoPool.<>c__DisplayClass31_0`1.<<InvokeCore>b__0>d.MoveNext()
```

</details>

<sub>Opened automatically from `Admin/_LogIncident/e21cccb65138f43b`. Recurrences are folded into this issue rather than opening new ones.</sub>

rbuergi:
Closes Systemorph/MeshWeaver#2836. Closes Systemorph/MeshWeaver#2787.

Two faces of one mistake, found a day apart in different layers: **a registry that has not answered yet was treated as a registry that answered *no*.**

Both readers collapsed every unsuccessful HTTP status into one failure, so the policy that (correctly) declines to re-ask a refusal also declined to re-ask a `503 … retry shortly`. In both places the fix is the same, and it is not a longer budget or a poll behind it: **make the status survive as data, then branch on what the server actually said.**

---

## Systemorph/MeshWeaver#2836 — the portal stopped reconciling its own packages

`RegistryUpdateReconciler` has no poll timer *by design* — its promise is "a consumer learns on its **next boot**". That makes the boot's bounded retry the only thing between a hiccup and a pod that never reconciles again. It skipped `InvalidOperationException`, reading it as *"a definite HTTP answer (401/403/404) — the same answer will come back in two seconds"*.

**That reasoning was right about the statuses it named and wrong about the type.** `RegistryPackageSource` threw that one bare type for *every* non-2xx, keeping the status only inside the message text:

```csharp
if (!resp.IsSuccessStatusCode)
    throw new InvalidOperationException(
        $"Registry catalog list failed ({(int)resp.StatusCode}): {json}");   // 401, 404, 503 — all the same type
```

So the single condition a boot retry exists for was the single condition it skipped. The pod's installed set then went stale for the rest of its life — no package updates, no grant changes — and **silently**, because the portal itself stays up.

- `RegistryResponseException(HttpStatusCode, message) : InvalidOperationException` — the same shape, and for the same stated reason, as the `InstanceRegistrationException` already in this file's neighbour (*"so callers can branch … without string matching"*). Deriving from `InvalidOperationException` keeps every existing `catch` behaving exactly as before.
- Thrown from the three sites that discarded the status: `ListPackages`, `FetchPackageFiles`, `PluginBundleClient.FetchIndex`.
- `ShouldRetryFeedRead` replaces the type test. Transient (any 5xx, 408, 429 — the conventional set, so the boundary is a published convention rather than a per-call-site judgement) is re-asked; definite still fails at once; a transport fault is re-asked exactly as before (#1500).

## Systemorph/MeshWeaver#2787 — and a platform promotion redded other repos' builds

Promoting the platform **rolls the portal**, and that portal *is* what serves the registry. On Manufacturing run `33316402646`: image promoted `14:12:41` → gate read 503 at `~14:16` → a portal pod started `14:18:22`. Six repos read that endpoint and three images were promoted that day, so the window opened three times. The only recourse was re-running a red build — precisely the habit worth removing.

Both gate readers now re-ask a transient answer with a widening pause (15/30/60/90 s, five attempts, ~3 min — sized to the *measured* window):

- `node-repo-gate.yml` — the upstream-seed index read **and** the per-bundle fetches. The latter were `curl -sSf`, which discards the very status that says whether re-asking would help.
- `compose-sealed-modules.sh` — the module-set index read and `fetch_module`, the same shape one lane over.

🚨 **A retry, not a skip-trapdoor.** When the attempts are spent the last status is returned and every caller still fails **RED**. A registry that is genuinely down fails the gate exactly as before, about three minutes later.

The helper is **duplicated on purpose**. That script is fetched from core at a pinned `platform-ref` while the workflow is pinned by the caller's `uses:`, so a shared third file would be a *third* independently-pinned artefact — the exact skew that once let an old workflow drive a new script and seal an empty module set. Each copy names the other.

---

## Verification

**#2836 — mutation-proved.** Restoring the pre-fix code (bare throw + type test) fails **8 of 15**, including the composed test. The 7 that still pass are the pre-existing behaviours the change preserves — which is the point of listing them.

> The two halves are tested **together** deliberately. A policy that branches on a status the throw site never records is still broken, and *each half passes its own test while the system fails*. So the pinning test feeds a **real** exception, from a real HTTP round trip through a stub handler, into the **real** predicate.

- `RegistryTransientFailureRetryTest` — 15/15 green.
- Full `MeshWeaver.PluginCatalog.Test`: **716 passed, 0 failed** (5 m 21 s).
- `MeshWeaver.PluginCatalog`, `Memex.Portal.Shared`, `MeshWeaver.Documentation.Test`, `MeshWeaver.PluginCatalog.Test` — each `-c Release -warnaserror`, `0 Warning(s) 0 Error(s)`.
- All ratchet guards + `WhatsNewEntryIntegrityTest` green.

**#2787 — behaviour-tested, not read.** Both copies driven through a stubbed `curl` across 13 scripted status sequences:

```
── Systemorph/MeshWeaver#2787: a roll window, then the portal is back ──
  ok   503 then 200          code=200 calls=2 slept=[15]
  ok   503 503 200           code=200 calls=3 slept=[15 30]
── a definite answer is NOT re-asked (behaviour unchanged) ──
  ok   404 refuses immediately   code=404 calls=1 slept=[]
  ok   401 refuses immediately   code=401 calls=1 slept=[]
  ok   403 refuses immediately   code=403 calls=1 slept=[]
  ok   200 first time            code=200 calls=1 slept=[]
── a registry genuinely down still fails RED, just later ──
  ok   503 forever           code=503 calls=5 slept=[15 30 60 90]
  ok   500 forever           code=500 calls=5 slept=[15 30 60 90]
── the rest of the transient set ──
  ok   429/502/504/408 then 200, and a doubled connection failure — all recover
```

`ALL PASS` for both copies; YAML parses; `bash -n` clean.

> A note on that harness: my first version reported **13 failures**, and every one was the harness — it read a counter across a `$( )` boundary. The function was correct the whole time. Worth recording because a broken harness and a broken subject look identical from the summary line.

Two What's New entries (`Category: Fix`), one per user-visible behaviour.

### DIFF
--- a/.github/scripts/compose-sealed-modules.sh
+++ b/.github/scripts/compose-sealed-modules.sh
@@ -53,6 +53,48 @@ case "$identity" in */*|*..*|"") echo "::error::compose-sealed-modules.sh: ident
 mkdir -p "$out"
 work=$(mktemp -d); trap 'rm -rf "$work"' EXIT
 
+# ── re-ask a registry that has not answered yet ───────────────────────────────────────────────
+# 🚨 A 5xx is not an answer — it is the ABSENCE of one. Promoting the platform ROLLS the portal
+# that serves this registry, so every promotion opens a window in which a dependent repo's gate
+# reads a 503 through no fault of its own diff (MeshWeaver#2787: three promotions in one day, six
+# repos reading this endpoint). Refusing on it makes the fleet's CI a function of deploy timing.
+#
+# A DEFINITE answer is untouched and still refuses at once — 404 (unsealed / predates module
+# sealing) and 401/403 (no grant) are decisions the registry has already made, and waiting on one
+# only delays the message naming it. Same distinction, same reason, as MeshWeaver#2836 on the
+# portal side: branch on what the server SAID, never on "the read failed".
+#
+# 🚨 A retry, NOT a skip-trapdoor: when the attempts are spent the last status is returned and
+# every caller still fails RED — a registry that is genuinely down fails the gate exactly as it
+# did before, about three minutes later.
+#
+# 🚨 There is a SECOND copy of this function, in .github/workflows/node-repo-gate.yml (same
+# policy, REGISTRY_KEY spelled upper-case there). It is duplicated ON PURPOSE rather than
+# factored into a third file: THIS script is fetched from core at a pinned `platform-ref` while
+# that workflow is pinned by the caller's `uses:`, so a shared file would be a THIRD
+# independently-pinned artefact — the exact skew that once let an old workflow drive a new
+# script and seal an empty module set. Change one, change both.
+registry_get() { # <url> <out-file> → echoes the final HTTP status
+  local url="$1" out="$2" code="" attempt=0
+  local -r delays="15 30 60 90"
+  while :; do
+    code=$(curl -sS -o "$out" -w '%{http_code}' \
+             -H "Authorization: Bearer $registry_key" "$url") || code="000"
+    case "$code" in
+      # Transient: a roll window, a gateway blip, or a connection that never landed (000).
+      408|429|5??|000) ;;
+      *) printf '%s' "$code"; return 0 ;;
+    esac
+    attempt=$((attempt+1))
+    local delay
+    delay=$(printf '%s\n' $delays | sed -n "${attempt}p")
+    [ -n "$delay" ] || { printf '%s' "$code"; return 0; }
+    echo "registry answered $code for $url — transient (a platform roll looks like this);" \
+         "re-asking in ${delay}s, attempt $((attempt+1))" >&2
+    sleep "$delay"
+  done
+}
+
 # ── one upstream's sealed module set (an index), or a RED reason ──────────────────────────────
 # Writes the listed bundle names, one per line, into <list-file>. Exit 1 with ::error when the
 # upstream has no seal for this identity or the seal predates module sealing — both are stop
@@ -65,7 +107,7 @@ sealed_modules_of() { # <source> <list-file>
   if [ -n "$registry_url" ]; then
     local base="${registry_url%/}/api/plugins/bundles/prebuilt/$identity/$src"
     local code
-    code=$(curl -sS -o "$work/$src.modules.json" -w '%{http_code}' -H "Authorization: Bearer $registry_key" "$base/modules")
+    code=$(registry_get "$base/modules" "$work/$src.modules.json")
     case "$code" in
       200) ;;
       404)
@@ -102,9 +144,10 @@ sealed_modules_of() { # <source> <list-file>
 fetch_module() { # <source> <bundle-name> <dest>
   local src="$1" name="$2" dest="$3"
   if [ -n "$registry_url" ]; then
-    curl -sSf -H "Authorization: Bearer $registry_key" -o "$dest" \
-      "${registry_url%/}/api/plugins/bundles/prebuilt/$identity/$src/modules/$name" \
-      || { echo "::error::could not fetch sealed module $name of '$src' for identity $identity"; return 1; }
+    local code
+    code=$(registry_get \
+      "${registry_url%/}/api/plugins/bundles/prebuilt/$identity/$src/modules/$name" "$dest")
+    [ "$code" = 200 ] || { echo "::error::could not fetch sealed module $name of '$src' for identity $identity — registry answered $code."; return 1; }
   else
     local account="${storage_target%%/*}" rest="${storage_target#*/}"
     local share="${rest%%/*}" base=""
--- a/.github/workflows/node-repo-gate.yml
+++ b/.github/workflows/node-repo-gate.yml
@@ -416,13 +416,56 @@ jobs:
           set -euo pipefail
           SEED_DIR="${RUNNER_TEMP:-/tmp}/upstream-seed"
           mkdir -p "$SEED_DIR"
+
+          # 🚨 A 5xx is not an answer — it is the ABSENCE of one, and refusing on it makes every
+          # platform promotion a source of red CI across the fleet (MeshWeaver#2787). Promotion
+          # rolls the portal that SERVES this registry, so the window is STRUCTURAL: the more
+          # reliably promotion works, the more often a dependent repo's gate lands inside it. One
+          # day's three promotions opened it three times, and six repos read this endpoint.
+          #
+          # A DEFINITE answer still refuses immediately and unchanged — 404 (unsealed), 401/403
+          # (no grant) are decisions the registry has already made, and waiting on one only
+          # delays the message that names it. Same distinction, same reason, as MeshWeaver#2836
+          # on the portal side: branch on what the server SAID, never on "the read failed".
+          #
+          # 🚨 This is a retry, NOT a skip-trapdoor. Exhausting the attempts returns the last
+          # status to the caller, which still refuses RED — a registry that is genuinely down
+          # fails the gate exactly as before, roughly three minutes later.
+          #
+          # 🚨 There is a SECOND copy of this function, in .github/scripts/compose-sealed-modules.sh
+          # (same policy, `registry_key` spelled lower-case there). It is duplicated ON PURPOSE
+          # rather than factored into a third file: that script is fetched from core at a pinned
+          # `platform-ref` while this workflow is pinned by the caller's `uses:`, so a shared file
+          # would be a THIRD independently-pinned artefact — the exact skew that once let an old
+          # workflow drive a new script and seal an empty module set. Change one, change both.
+          registry_get() { # <url> <out-file> → echoes the final HTTP status
+            local url="$1" out="$2" code="" attempt=0
+            local -r delays="15 30 60 90"
+            while :; do
+              code=$(curl -sS -o "$out" -w '%{http_code}' \
+                       -H "Authorization: Bearer $REGISTRY_KEY" "$url") || code="000"
+              case "$code" in
+                # Transient: a roll window, a gateway blip, or a connection that never landed
+                # (curl reports 000). Everything else is the registry's decision — return it.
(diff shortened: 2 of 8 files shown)


rbuergi:
## The red is Systemorph/MeshWeaver.Plugins#1141, and this diff cannot reach it

Shard 3 failed on a single test out of 1 186:

```
FutuReAnalysisTest.EuropeRe_AnnualReport_EmbeddedCharts_ShouldRenderViaPathResolution  [took 50s]
```

That is the tracked flake **#2533** — *"times out on a RedirectControl — new flake population"* — by name and by shape (a 50 s timeout, not an assertion).

This PR's diff is three files in `MeshWeaver.PluginCatalog`, two CI YAML/shell files, one new test, and two What's New entries. None of it is on the path of a FutuRe chart render, and the CI files do not execute in the test job at all.

Re-running the failed job **on that attribution**. Recorded a fresh occurrence on Systemorph/MeshWeaver.Plugins#1141.

*(#2868, open in the same window with a completely different diff, went red simultaneously on a different test and different shard — the ambient-population signature rather than either change. Its failure carried the `Dropping StreamEndedEvent` + `ADVANCE_WITHOUT_HANDOFF` pair inside the hang's own window, which resolved a discriminator I had left open on Systemorph/MeshWeaver#2865.)*

Issue #2776, opened 2026-08-30
Flake: ActivityLiveProgressTest.EveryProgressLine_LandsWhileTheCommandIsStillRunning — the activity node's owner ends its streams ~5 s in, the next progress write gets no verdict (OwnerUnreachable after 31 s)


rbuergi:
## Observed

2026-08-30 13:28Z on #2774 (a one-line change to `deploy/homebrew/scripts/make-tarball.sh` — its diff cannot reach a GitSync activity test), shard 5, run https://github.com/Systemorph/MeshWeaver/actions/runs/33313654005/job/99263942522, commit `2abb6a60`. main's three preceding Build-and-Test runs (`41dd85a2`, `ead524e9`, `2ead7832`) are green.

```
MeshWeaver.GitSync.Test.ActivityLiveProgressTest.EveryProgressLine_LandsWhileTheCommandIsStillRunning [FAIL]
Assert.Equal() Failure: Expected: Succeeded  Actual: Failed
[CI] MeshWeaver.GitSync.Test exit=1 TESTFAIL (the host completed normally)
```

## Mechanics, from the test's own output (activity `GhLive22d6da4e6/_Activity/b55c9a35`, cache hub `cache/6D99IQwI1U23HtKNOUtrNw`)

| t | what |
|---|---|
| 13:28:27.815 | TEST START |
| 13:28:32.936 | `[DataExtensions] Dropping StreamEndedEvent for stream uZzjE4MWi0KBrSXTSLNN9g … no synchronization hub found on this hub or any parent — the target stream is gone` (sender: the activity node) |
| 13:28:32.940 | `[MeshNodeStreamCache] [UpdateQueue] ADVANCE_WITHOUT_HANDOFF path=…/_Activity/b55c9a35 seq=3 bound=5000ms — the owner never acknowledged this write inside the bound` |
| 13:28:32.946 | a second `Dropping StreamEndedEvent` (stream dPb18nIfvUCsJEhA8RgoLQ), same sender |
| 13:28:58.951 | `[MeshNodeStreamHandle] [UpdateRemote] VERDICT_TIMEOUT … bound=31s — the owner produced no terminal for this patch` → `MeshNodeStreamException: MeshNode OwnerUnreachable` |
| 13:28:58.952 | `[ActivityRunner] Activity … failed` → status `Failed` → the assertion |

[...]

PR #2868, opened 2026-08-31
fix(#2778): the OwnerDisposing NACK is handed to the waiter, not posted and hoped for

### LINKED ISSUE #2778: OwnerDisposing NACK is dropped exactly when the parent is disposing hosted hubs — the caller hangs for its full budget

rbuergi:
Closes Systemorph/MeshWeaver#2778.

The owner mints an explicit *"this patch was NOT applied — safe to retry"* NACK when it goes down mid-write. Delivering it is the entire point of minting it. It was delivered like this:

```csharp
var parent = hub.Configuration.ParentHub;
if (parent is not null && parent.RunLevel < MessageHubRunLevel.DisposeHostedHubs)
    parent.Post(resp, o => o.ResponseFor(request));
```

With the levels ordered `Starting, Started, Quiescing, DisposeHostedHubs, …`, that guard is open for the first three and shut for the rest — **so the one moment a whole batch of owner hubs goes down with patches in flight is the one moment none of their NACKs is delivered.** The caller then burns the full 31 s `WriteVerdictBound` waiting for an answer that had already been written and thrown away.

The guard justified itself with *"during a whole-mesh teardown the parent is past that mark too, the post is skipped, and **nobody is waiting**."* That is not something the code could verify, and it is false exactly when it matters.

## The post was never the seam this NACK was designed to arrive on

A verdict that lands after `UpdateRemote`'s ~2 s bounded wait has **no pending `Observe` callback left**. So posting it is only a way of reaching `LatePatchResponseRegistry` the long way round — routed to the caller, into the cache hub's `PatchDataResponse` handler, and finally `Dispatch`. That handler's own comment already names this NACK as the thing it exists for:

> *"the ONLY place a post-optimistic-emit verdict (**above all the OwnerDisposing disposal NACK**) can still be observed"*

Within one mesh the registry is a singleton both hubs already share, so the owner can hand the verdict over directly: same waiter, no hub woken, no message routed. And `Dispatch` **returns whether a caller was armed** — so "nobody is waiting" stops being an assumption and becomes a checked fact.

- **`ILatePatchVerdictSink`**, new in `MeshWeaver.Data.Contract`. The owner side lives in `MeshWeaver.Data`, which cannot reference `MeshWeaver.Mesh.Contract` — that closes the cycle `Data → Mesh.Contract → Layout → Data`. `PatchDataResponse` already lives in `Data.Contract`, so the verb belongs there too.
- `LatePatchResponseRegistry` implements it — its existing `Dispatch` already has the signature. Registered as a **factory over the concrete singleton**, never a second `AddSingleton`: two instances would mean the owner dispatching into a registry nobody armed, which reads exactly like *"nobody was waiting"* and would restore the bug in a form no test could see.
- The parent post **stays, unchanged**, as the only route to a caller in another process, which this registry cannot see.

## Cost — why the obvious fix was reverted, and why this one is affordable

A previous attempt walked the parent chain to the first ancestor that could still post. It is correct about the caller and was measured at **~10× on teardown**, so it was reverted. This shape is a dictionary lookup that misses, so the common teardown — nobody waiting — pays nothing.

Measured on one box, one variable at a time, same 166 tests, `DOTNET_PROCESSOR_COUNT=4`:

| variant | duration | tests |
|---|---:|---|
| baseline (main-equivalent worktree, measured now) | **24 s** | 166/166 |
| **this branch** | **25 s** | 166/166 |
| the rejected ancestor walk | 176 s | *(prior session)* |

## The test, and two things it taught me

`NackReachesTheWaiterDuringTeardownTest` disposes the **mesh**, not just the node hub. Disposing only the owner leaves the parent at `Started`, where the old guard was **open** — which is why `OwnerDisposalNackTest` and `LateNackReenqueueTest` both pass with the defect fully present, and why it survived this long.

🚨 **It asserts the armed watch is consumed, not that the caller's callback runs.** My first draft asserted the caller's terminal and **failed with the fix in place**. The reason is sound: a mesh teardown takes the caller's own subscription down along with the owner, so `observer.OnError` lands on a disposed subscription and is a silent no-op. That is a property of the scenario, not of the framework — asserting it would have meant asserting something no implementation could satisfy. The registry entry is armed while a caller waits and is removed by `Dispatch`, so it returning to zero **is** "the verdict reached the waiter". The bound is well under the 30 s watch bound, because an entry that merely *expires* also ends at zero.

The failing draft is also what proved the fix works end to end: `LATE_NACK_REENQUEUE … code=OwnerDisposing` appears in its log, which the pre-fix code cannot produce at all.

**Mutation-proved:** with the direct hand-off removed, the new test fails on a timeout. All **9** sibling NACK / late-watch tests stay green — and `LateNackReenqueueTest` now exercises the new path, since its caller is armed.

`MeshWeaver.Data.Contract`, `MeshWeaver.Data`, `MeshWeaver.Mesh.Contract`, `MeshWeaver.Hosting`, `MeshWeaver.Hosting.Monolith.Test`, `MeshWeaver.Content.Test` — each `-c Release -warnaserror`, `0 Warning(s) 0 Error(s)`.

One What's New entry (`Category: Fix`).

### DIFF
--- a/src/MeshWeaver.Data.Contract/ILatePatchVerdictSink.cs
+++ b/src/MeshWeaver.Data.Contract/ILatePatchVerdictSink.cs
@@ -1,0 +1,42 @@
+namespace MeshWeaver.Data;
+
+/// <summary>
+/// 🚨 The seam a LATE owner verdict for a cross-hub MeshNode write is handed to, declared here so
+/// the hub that MINTS the verdict can reach the watch that is waiting for it.
+///
+/// <para><b>Why the interface exists at all.</b> The implementation is
+/// <c>MeshWeaver.Mesh.LatePatchResponseRegistry</c>, in <c>MeshWeaver.Mesh.Contract</c>. The owner
+/// side that mints the <c>OwnerDisposing</c> NACK is
+/// <c>DataExtensions.RegisterOwnerDisposingNack</c>, in <c>MeshWeaver.Data</c> — and
+/// <c>MeshWeaver.Data</c> cannot reference <c>MeshWeaver.Mesh.Contract</c>, because that would
+/// close the cycle <c>Data → Mesh.Contract → Layout → Data</c>. <see cref="PatchDataResponse"/>
+/// already lives here, in the assembly both sides reference, so this is where the verb belongs
+/// too.</para>
+///
+/// <para><b>What it is for.</b> A verdict that lands after <c>UpdateRemote</c>'s ~2 s bounded wait
+/// has no pending <c>Observe</c> callback left, so posting it is only a way of reaching this same
+/// registry the long way round — routed to the caller, into the cache hub's
+/// <see cref="PatchDataResponse"/> handler, and finally to <c>Dispatch</c>. Within one mesh the
+/// registry is a singleton both hubs already share, so the owner can hand the verdict over
+/// directly: same waiter, no hub woken, no message routed. That matters most exactly when the post
+/// is unavailable — a parent already past <c>DisposeHostedHubs</c> (#2778).</para>
+///
+/// <para>🚨 It also turns an assumption into a fact. The dropped-NACK guard justified itself with
+/// "nobody is waiting", which no code could verify and which was false precisely when it mattered.
+/// <see cref="Dispatch"/> REPORTS whether a caller was armed, so the question is answered rather
+/// than assumed — and a miss costs one dictionary lookup, which is what makes it affordable to ask
+/// on every teardown.</para>
+/// </summary>
+public interface ILatePatchVerdictSink
+{
+    /// <summary>
+    /// Delivers a late owner response to its armed watch, if one is still armed and unexpired.
+    /// </summary>
+    /// <param name="requestId">The originating <c>PatchDataRequest</c> delivery id — the same value
+    /// carried back as the response's <c>RequestId</c> correlation property.</param>
+    /// <param name="response">The owner's verdict.</param>
+    /// <returns><c>true</c> when an armed, unexpired watch consumed it; <c>false</c> when nobody in
+    /// this mesh was waiting — which the caller may then treat as a checked fact rather than a
+    /// guess.</returns>
+    bool Dispatch(string requestId, PatchDataResponse response);
+}
--- a/src/MeshWeaver.Data/DataExtensions.cs
+++ b/src/MeshWeaver.Data/DataExtensions.cs
@@ -892,6 +892,40 @@ public static class DataExtensions
                 Error = nodeErr.Message,
                 NodeError = nodeErr,
             };
+            // 🚨 Hand the verdict to the waiting caller DIRECTLY, before considering a post.
+            //
+            // The transport below is not the seam this NACK was designed to arrive on. A verdict
+            // that lands after UpdateRemote's ~2 s bounded wait has NO pending Observe callback,
+            // so the post is only ever a way to reach LatePatchResponseRegistry the long way round
+            // — routed to the caller, into the cache hub's PatchDataResponse handler, and finally
+            // Dispatch. That handler's own comment names this NACK as the thing it exists for.
+            // Inside one mesh the registry is a singleton both hubs already share, so asking it
+            // first reaches the same waiter with no hub woken and no message routed.
+            //
+            // 🚨 This is what makes the fix affordable. The obvious repair — walk the parent chain
+            // and post through the first ancestor that can still post — is correct about the
+            // caller and was MEASURED at ~10x on teardown: MeshWeaver.Content.Test went 29 s → 176 s,
+            // a uniform ~0.9 s per test, because every hub going down with a delivery outstanding
+            // now wakes callers mid-drain. It was implemented and reverted for that reason. A
+            // dictionary lookup that misses costs nothing, so the common teardown — nobody waiting
+            // — pays nothing at all.
+            //
+            // 🚨 And it replaces an assumption with a fact. The old guard's rationale was "during a
+            // whole-mesh teardown the parent is past that mark too, the post is skipped, and NOBODY
+            // IS WAITING". Nobody-is-waiting is not something that code could verify, and it was
+            // false precisely when it mattered: the caller whose wait outlives the start of
+            // teardown is still waiting, and it is the one guaranteed to get silence instead of its
+            // NACK, then to burn the full 31 s WriteVerdictBound (#2778). Dispatch RETURNS whether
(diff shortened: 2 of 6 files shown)


rbuergi:
## The red is attributed, not assumed — the changed path never ran in the failing test

Shard 2 failed on `LogonActionFrameworkTest.An_every_logon_action_runs_on_every_logon`. That test lives in `MeshWeaver.Graph.Test`, which **does** depend on all four assemblies this PR touches, so "unrelated" was not something I could claim by inspection. There is even a plausible mechanism: this change makes the late-watch callback run inline on the owner's disposal thread rather than on the cache hub's action block, and that callback does real work.

So I checked whether the modified path executed at all.

```
LATE_NACK / OwnerDisposing / LATE_ACK lines in the failing test's own output:   0
the same tags elsewhere in shard 2:                                           30
```

**The path this PR changes produced no output in that test.** `RegisterOwnerDisposingNack` runs only when an owner hub disposes with a patch in flight; nothing in that test entered it. A change cannot hang a test through a branch the test never takes.

## What it actually is

The hang's own window:

```
02:46:36.206  TEST START
02:46:36.292  Logon action test.failing failed for logonuser   ← the test's own deliberate failure
02:46:41.286  Dropping StreamEndedEvent for stream _FIj1jTenEynt12CpiSiJA
02:46:41.298  [UpdateQueue] ADVANCE_WITHOUT_HANDOFF path=logonuser
02:46:41.300  Dropping StreamEndedEvent for stream P3XyG-XfKkKUCVWDB_Yu6w
              ── 25 s of complete silence ──
02:47:06.218  TEST FAILED: The operation has timed out
```

Burst, then *total* silence to the deadline — `/debug`'s signature for a dropped reactive emission rather than a lock. `ADVANCE_WITHOUT_HANDOFF` is the queue side of that same absence: it advanced because the owner never acknowledged. This is the ambient family of Systemorph/MeshWeaver#2865 / Systemorph/MeshWeaver#2776 / Systemorph/MeshWeaver#2778, whose signature predates tonight (it is recorded in Systemorph/MeshWeaver#2776 and Plugins#993).

Note the direction: this PR makes **more** owner verdicts reach their waiter, so if it moves this family at all it moves it the other way.

I have posted the fuller finding on Systemorph/MeshWeaver#2865 — this instance is a better witness than the ones investigated so far, because a **method-timeout** kill preserves the hung test's transcript while a **host-cap** kill destroys it.

## Corroboration from an unrelated PR

Systemorph/MeshWeaver#2867 went red in the same window on a *different* test and a *different* shard (`FutuReAnalysisTest`, the tracked flake Systemorph/MeshWeaver.Plugins#1141). Its diff is PluginCatalog plus two CI YAML files and cannot reach the mesh write path or a chart render. Two unrelated diffs failing simultaneously on two unrelated tests is the signature of an ambient population, not of either change.

Re-running the failed jobs **on that attribution**, not to see whether it passes.

PR #2875, opened 2026-08-31
fix(#2446): a wedged root teardown no longer costs 30 silent seconds

### LINKED ISSUE #2446: Module bundle installs: writing 31 MB takes <1 s, but 1–7 minutes of idle sit BETWEEN packages — a full pull takes ~30–40 min instead of ~1
Measured live on memex-cloud during the 2026-08-26 post-bake bundle pull (the 4327030-baked generation landing on the volume):

- Inside `MeshWeaver.SelfUpdate.Aks@6df52936` every file carries an mtime within a ~200 ms window (20:15:06.03–20:15:06.23) — the download+verify+write of a 31 MB package is **sub-second**.
- The previous package (`MeshWeaver.AI.Anthropic@e0f09bc8`) finished at 20:08:15 — a **7-minute gap** before the next package started. Earlier stretch: ~9 packages in 9 minutes (~1/min pacing).

So a fleet-critical pull (the one a platform roll waits on when the identity flips and every bundle re-mints) takes 30–40 minutes when the actual I/O totals well under a minute. Sequential-in-dependency-order is correct and is not the cost; the cost is whatever runs or waits BETWEEN packages — candidates: a timer-paced queue drain (~1/min tick shape), per-package post-install verification, or the idempotence re-install probe. The variable 1→7 min spread suggests at least one wait is event/timer-shaped rather than work-shaped.

Why it matters: tonight the fleet roll to `ci.5876` is gated on exactly this pull (the image↔landed-modules atomic move; rolling early = the OpenAICompatibleModelSync/ProviderModelLister constructor break). Every identity-flipping platform build pays this ~40 min tail per portal.

Needed: instrument the installer's inter-package interval (what completes at package N end vs what starts package N+1), name the wait, and make it work-shaped (drain the queue as fast as verification allows). Per no-band-aids: not a smaller timer — find what the timer is papering over.

rbuergi:
Toward #2446 (does **not** close it — see the last section).

#2446 measures a bundle pull taking 30–40 minutes when the I/O totals under a minute: *"31 MB written in under a second, then minutes of idle"*. Its triage established the shape — the installer's eight bounds are **ceilings, not sleeps**, so a gap means one of them **expired**, and which one is the whole diagnosis. Its prescribed next step:

> *"Each expiry is observable. The next time a pull is watched, grep the portal log across the gap for the messages these bounds emit — whichever one appears names the wait."*

I checked that premise across every `.Timeout(<named bound>)` in `PackageInstaller` rather than assuming it. **Seven of the eight report their own expiry. One does not.**

```
line   bound                        reports on expiry?
  521  GatingSettleTimeout          yes
  636  DeclaredAccessVerifyBudget   yes
 1172  RootRecycleTimeout           🚨 NO — silent
 1274  RootTypeSettleTimeout        yes
 1399  RootReadyTimeout             yes
 1453  RootTypeProbeTimeout         yes
 1481  WarmTimeout                  yes
 3405  SeedBound(typeCount)         yes
```

```csharp
.Timeout(RootRecycleTimeout)                                  // 30 s
.Catch<Unit, Exception>(_ => Observable.Return(Unit.Default)) // ← swallowed, no log
```

**Continuing is right; continuing in silence is not.** The bound's own remark says it *"is only ever reached when a hub's teardown itself wedges"* — so expiring it is not a slow path, it **is** the wedge, and that is the single most interesting event on the path. Thirty seconds per occurrence, appearing nowhere.

## Why this blocked the diagnosis rather than merely being untidy

The technique #2446 depends on is *read the log across the gap and see which bound names itself*. For this bound the answer would always have been "none of them" — a gap it caused is **unattributable by construction**, and a reader would conclude the stall came from somewhere else. A wait nobody can see is a wait nobody can fix.

It is also the `AGENTS.md` swallow-and-continue shape, narrowly applied: the recovery is correct, only the silence is wrong.

## The change

The arm logs the root, the bound, and what it was waiting for, then returns exactly as before — **the recovery path is byte-identical, no behaviour change**. `logger` is threaded from the caller, which already has it.

`InstallerBoundsNameThemselvesGuard` keeps it that way: every bounded wait in the installer must report its own expiry, plus a self-test that the scanner actually finds the bounds — a regex matching nothing would satisfy the first test **vacuously**, which is the same defect one level up.

**Mutation-proved:** restoring the silent swallow fails the guard with

```
src/MeshWeaver.PluginCatalog/PackageInstaller.cs:1173 — .Timeout(RootRecycleTimeout) expires without a log
```

## Verification

- `MeshWeaver.PluginCatalog.Test`: **716/716**
- `MeshWeaver.Documentation.Test`: **134/134**
- `MeshWeaver.PluginCatalog`, `MeshWeaver.PluginCatalog.Test`, `MeshWeaver.Documentation.Test` — each `-c Release -warnaserror`, `0 Warning(s) 0 Error(s)`

## 🚨 This does not close #2446

It makes the next occurrence diagnosable; it does not say which bound is expiring. That still needs a live pull, exactly as the issue says. What changes is that the pull no longer has to be *watched* — the log will name whichever bound it was, **including this one**.

One What's New entry (`Category: Fix`).


rbuergi:
## CI caught my own guard being brittle — fixed, and the failure is worth recording

Shard 3 went red on `InstallerBoundsNameThemselvesGuard.TheScannerSeesTheBoundsItClaimsTo` — **this PR's own self-test**, not the thing it guards:

```
Set:       ["GatingDetectorBudget", "DeclaredAccessVerifyBudget", "RootRecycleTimeout", …]
Not found: "GatingSettleTimeout"
```

**#2849 renamed `GatingSettleTimeout` → `GatingDetectorBudget` while this branch was being written**, and my self-test listed three bound names verbatim. So a guard about *silence on expiry* failed for a reason that had nothing to do with silence.

That is exactly the brittleness I noted to myself while writing it and introduced anyway. **A guard that names a symbol inherits every rename of that symbol as a false failure — and a guard that cries wolf gets deleted, taking the real check with it.**

The self-test's only job is to prove the scanner is not vacuous. It now asserts a **count** (≥ 6 bounded waits, so an empty match cannot satisfy the real assertion) and a **shape** (each bound ends `Timeout` / `Budget` / `Bound`, proving it matches budgets rather than incidental tokens). Both survive renaming. No hard-coded identifiers remain.

Main is merged in, which is where the rename came from.

**Re-proved after the change:** restoring the silent swallow still fails the primary assertion, naming file, line and bound. `Documentation.Test` 134/134; both touched projects `-c Release -warnaserror` at `0 Warning(s) 0 Error(s)`.

Worth noting the guard did its job in the end — it caught a real defect, which was its own.

PR #2879, opened 2026-08-31
fix: a leaf both sides already agree on is not a conflict — and the audit stamp never was one

rbuergi:
## The measurement

`OrleansResubmitDeadlockTest.Resubmit_AfterExecution_DoesNotDeadlock` on `MeshWeaver.Plugins` fails **3 of 12 unloaded runs**, each in about a second. A failure in milliseconds is state, not timing, and the thread's `Messages` list says which state — it oscillates and loses cells:

```
[u,r] → [] → [u,r'] → [r'] → [r',u,r''] → …
```

## What the owner was actually refusing

Instrumenting `MeshNodePatchMerge.Apply` at the refusal point, on a failing run:

```
[PROBE-LEAF] key=status             live=<null> base="Executing" patch=<null> patchEqLive=True
[PROBE-LEAF] key=activeMessageId    live=<null> base="d2a7479e"  patch=<null> patchEqLive=True
[PROBE-LEAF] key=executionStartedAt live=<null> base="…44068Z"   patch=<null> patchEqLive=True
```

The writer's patch asked for **exactly the state the owner was already in** — an RFC 7396 removal of three keys the owner had already removed — and all three were refused.

`Apply` classified each leaf on `live != base` alone. That says the owner moved since the writer read it; it does **not** say the two disagree. When `patch == live` there is nothing to resolve: applying it is a no-op, and refusing it invents a conflict.

## Why inventing one is expensive

Since *"a PARTIAL merge refusal is still a refusal"* (#2463 / #2840) a refusal NACKs `Conflict` on a write that **landed**. `Conflict`'s prescribed remedy is *"re-read and re-apply"*, which `MeshNodeStreamHandle.UpdateRemote` and `MeshNodeStreamCache` both implement by **re-running the caller's update lambda** — up to three attempts each. Sound for a lambda that SETS a value; unsound for one that performs a RELATIVE mutation. The chat resubmit truncates a thread at an index, so each re-run truncated away the round the submission watcher had dispatched in between.

The re-enqueue arm still documents the invariant it was written under, and #2840 ended it:

> *"The owner emits Conflict ONLY when the merge refused keys AND the node is byte-identical to its pre-merge state — i.e. nothing landed"*

The measured sequence on one failing run: 3 lambda invocations ~100 ms apart (`UpdateRemote`'s re-enqueue budget), then `[PROBE-CONFLICT-RETRY] … attempt=1 … PARTIALLY refused: 2 field(s)` from the cache's own retry on top.

## The second guaranteed refusal, same reason

`LastModified` / `LastModifiedBy` are not caller data — `UpdateRemote` stamps them onto **every** cross-hub patch itself. So the second of two concurrent writers always carries a base stamp the owner has moved past, and the generic scalar rule refused it **every time**, i.e. every concurrent cross-hub write was a `Conflict` and a re-run. Observed in the same trace:

```
[PROBE-REFUSE] key=lastModified live="…20.513404" base="…20.398719" patch="…20.614317"
```

The stamp is monotonic by construction, so this is the `RequestedReleaseAt` case one field along: `RebaseAuditStamp` drops a superseded stamp and rebases a newer one, with `LastModifiedBy` travelling with whichever wins.

## Scope

Both changes are exact. A leaf whose values genuinely differ is still refused (`NoOpLeaf_DoesNotMaskAGenuineScalarConflict`), and a conflicting content scalar carried in the same patch as an audit stamp is still refused (`AuditStampMerge_DoesNotSoftenAConflictingContentScalar`).

## Verification

- 6 new unit tests in `MeshNodePatchMergeTest` / `MonotonicTriggerMergeTest`. **Mutation-proved**: with the two source changes reverted, all 6 fail and the "does not mask a genuine conflict" case correctly still passes.
- `MeshWeaver.Data.Test` full suite: **413 passed, 0 failed**.
- `OrleansResubmitDeadlockTest` against this branch, with the Plugins tree **unchanged**: **30 runs, 0 failures** (baseline 3 in 12).

## Not fixed here, and worth its own issue

A **partial** refusal and a **total** refusal both NACK `Conflict`, and the two retry layers cannot tell them apart. With the spurious refusals gone this is much rarer, but a genuine partial refusal still answers a landed write with "re-run your mutation", which is unsound for any relative mutation. The companion PR on `MeshWeaver.Plugins` hardens the one caller this bit; the general contract is a maintainer call.

Fixes Systemorph/MeshWeaver.Plugins#1009

### DIFF
--- a/src/MeshWeaver.Data/DataExtensions.cs
+++ b/src/MeshWeaver.Data/DataExtensions.cs
@@ -1052,6 +1052,16 @@ public static class DataExtensions
                 // request LOST (memex-cloud 2026-07-20 GitSync burst: every explicit Store/Plugin
                 // compile trigger was dropped while mirrors lagged the owner).
                 RebaseMonotonicTriggers(currentNode, patchNode, baseValues, jsonOpts, logger, hubPath);
+                // 🚨 …and the same resolution for the node's AUDIT STAMP, which is not caller data at
+                // all: MeshNodeStreamHandle stamps LastModified/LastModifiedBy onto EVERY cross-hub
+                // patch itself. So the moment two writers touch one node the second one's base stamp
+                // is stale by construction, the generic scalar rule refuses it, and — since a PARTIAL
+                // refusal nacks Conflict (#2463/#2840) — a write that fully landed is answered with
+                // "re-read and re-apply", which re-runs the caller's mutation. For a RELATIVE mutation
+                // that is a second application, not a no-op. LastModified is monotonic by construction
+                // (every writer sets it to UtcNow), so newest-wins is the correct merge, exactly as for
+                // RequestedReleaseAt — never a refusal.
+                RebaseAuditStamp(currentNode, patchNode, baseValues, jsonOpts, logger, hubPath);
                 var refused = 0;
                 Serialization.MeshNodePatchMerge.Apply(currentNode, patchNode, baseValues,
                     onRefuse: key =>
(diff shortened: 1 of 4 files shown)


rbuergi:
## The shard-2 red on the first run was not this diff — evidence, not a feeling

`Run tests (shard 2)` failed as `MeshWeaver.Hosting.Monolith.Test.HOST_CRASHED (exit=124)` — the 8 m per-project wall-clock cap. Chasing it rather than re-running it:

**Which test was stuck.** Walking the shard log for a `TEST_START` with no matching `TEST_END` leaves exactly one, out of 1 111:

```
ReleaseRequestWatcherHighWaterTest.BurstOfTwoReleaseTriggers_SecondTriggerIsNotLost
```

**Why it was stuck — its own log says so, and it is the assembly-load path:**

```
[CompileWatcher] Compile failure for TestData/HighWaterType: Failed to load the compiled assembly —
  Could not load type 'HighWaterLayoutAreas' from assembly 'DynamicNode_TestData_HighWaterType…'
  because the format is invalid.
[CompileWatcher] NodeType TestData/HighWaterType is STUCK at Error with no compiled assembly: its verdict
  was formed under the compile inputs that are live now …, so the automatic re-drive correctly declines
  and nothing will retry it until … someone requests a build.
```

The type parks at `Error`, so the test's closing `Match(… CompilationStatus == Ok)` can never be satisfied, it burns its 45 s wait, and the project runs past the cap. That is #1387's signature — an assembly read while it is being written — not a merge outcome.

**Why this diff cannot reach it:**

- It changes two things only: one leaf comparison in `MeshNodePatchMerge.Apply`, and a new top-level `lastModified`/`lastModifiedBy` rebase. Neither participates in Roslyn emit or assembly load.
- All three writes that test makes go through the **owner's own stream** (`nodeHub.GetWorkspace().GetMeshNodeStream()` → `IsOwn` → `UpdateOwn`), which never calls `ApplyMeshNodeMerge` at all.

**Measured on this branch:**

| | |
|---|---|
| `MeshWeaver.Hosting.Monolith.Test`, locally, whole project | **777 passed, 0 failed**, 2 skipped by design (10 m 36 s) |
| `MeshWeaver.Data.Test` — the suite that owns the changed code | **413 passed, 0 failed** |
| `Run tests (shard 2)` re-run | **pass, 7 m 43 s** |

The control is also worth recording: on the neighbouring PR #2880 the same shard ran `MeshWeaver.Hosting.Monolith.Test` as ONE project in 268 s, while this run partitioned it (`part 2/3`, 124 of 187 classes) — a different partitioning and ordering, which is the usual precondition for this class of wedge.

PR #2883, opened 2026-08-31
ci(module-pack): installing the .NET SDK is a DECLARED need, not a reflex

rbuergi:
> **DO NOT MERGE, DO NOT ARM AUTO-MERGE.** No caller changes behaviour from this PR — the
> default is today's behaviour, byte for byte — but the flag has never had a caller that
> declares anything other than the default, and that is deliberate (see *Which lanes can opt
> out today*, below: none can).

## What this does

> *"make installing dotnet sdk a feature flag"* — maintainer, 2026-08-31

`actions/setup-dotnet` ran in the `pack` job the way it runs in every job in the fleet: as a
reflex, before anyone asked whether the job compiles anything with it. That is the wrong shape
the moment the platform **image** can be the compiler (`"build": "container"`, #2854) — "the SDK
is here" then silently answers a question nobody put, and an unconverted step keeps working on a
runner SDK the conversion was supposed to have removed.

So the install becomes a **declaration**, with a safe default:

| `dotnet-sdk:` | behaviour |
|---|---|
| **`install`** (the **DEFAULT**) | install it, unconditionally. A caller that says nothing gets exactly today's behaviour; a lane added tomorrow inherits it. |
| `by-declaration` | install it only if a consumer in **this** job still needs one. |
| anything else, **the empty string included** | **RED**. A flag that cannot be read must never resolve to "skip". |

Opting out is a deliberate edit. Drifting out is impossible.

## Why it cannot become a skip-trapdoor (invariants #3 / #4)

Three things, structural rather than hoped-for:

1. **The DECISION step is unconditional** and fails on an unreadable policy. The only `if:` the
   flag introduces is on `actions/setup-dotnet` itself — and skipping an *installer* is not
   skipping a check: every build, pack, inspect and test step below still runs and still gates.
   **The flag chooses a COMPILER; it never chooses whether to verify.**
2. **"No SDK" is ENFORCED, not merely declined.** GitHub's ubuntu runners ship an SDK of their
   own, so "we skipped setup-dotnet" would otherwise mean "we quietly built against whatever
   .NET the runner image happens to carry". When the answer is no, the job puts a **failing
   `dotnet` on PATH** (exit 97, `::error::` naming the flag and what to do). Verified by
   execution — see below.
3. **The path taken is printed every run**, in the log *and* the job summary, with the consumers
   enumerated by name — including when the answer is *install*. The two outcomes are never told
   apart only by what is missing.

## The consumer inventory is complete, not convenient

An omission here is exactly what would let a converted lane skip an install it needs and then
blame the shim four steps later. Evaluated per entry, against the tree:

| consumer | when |
|---|---|
| the compiler for the module | unless the entry declares `build: container` |
| the **module-pack tool** (`meshweaver/src/MeshWeaver.Plugin.Build`) | **always** — `dotnet build` + `dotnet run` in every mode |
| the binding-identity probe (`dotnet run` file-based app) | on the `container` path (#2854) |
| the module's own suite (`dotnet test <project>.Test`) | when `matrix.entry.test != false` **and** the directory exists — resolved the same way the test step resolves it |

## 🚨 Which lanes can opt out today: **none**, and this PR says so every run

**No entry can drop the SDK, container or not** — `MeshWeaver.Plugin.Build` alone keeps every
pack job on it, because the packer is `dotnet build` + `dotnet run` in every mode and the
platform image does not carry it. A module that ships a test suite is blocked a second time
(`dotnet test` is not a verb `build-project` has at all).

That is the **measured state**, not a placeholder, and it is stated in the header, in the input
description, and in the log line every run — rather than left as an inert flag someone later
mistakes for a working opt-out. **The lane converts when the packer moves inside the platform
image, not before.**

## Relationship to #2854 (`ci/pack-lane-memex-build`)

Same idea, one level up, and deliberately **independent**: this PR is based on `main` and touches
none of the lines #2854 rewrites (it inserts a step *before* `setup-dotnet`; #2854's hunks start
after it). It reads `matrix.entry.build` defensively — absent on `main` ⇒ `sdk` ⇒ install — so it
is correct today and composes with #2854 the moment that lands, with no merge conflict either way.

## What was verified by RUNNING, and what was not

**Run locally** (the step script extracted from the YAML and executed with a fabricated runner
environment):

* `install` / `sdk` → `install=true`, consumers listed ✔
* `by-declaration` / `sdk` / suite present → 3 consumers, `install=true` ✔
* `by-declaration` / `container` / suite present → compiler dropped from the list, 3 consumers,
  `install=true` ✔
* `by-declaration` / `container` / `test: false` → 2 consumers, `install=true` ✔
* a fully-converted job (both remaining consumers removed by hand, to reach the branch the real
  lane cannot reach yet) → `install=false`, shim written, `$GITHUB_PATH` appended ✔ — and
  `dotnet build foo.csproj` through that PATH exited **97** with the `::error::` naming the flag ✔
* `dotnet-sdk: instal` (typo) → exit 1, `::error::` naming the two legal values ✔
* `dotnet-sdk: ''` → exit 1, same ✔
* the `$GITHUB_STEP_SUMMARY` block rendered and inspected ✔

**Static**: `actionlint` clean (exit 0); the file reports the **same five** pre-existing
shellcheck findings as `origin/main` and no new ones; `.github/scripts/check-workflow-shell.py`
reports `0 live` findings; the extracted script is `bash -n` and `shellcheck` clean.

**NOT verified — stated rather than implied:** no `container`-mode run has ever executed
anywhere, here or in #2854. The pinned image does not yet carry `build-project` (it is in
unmerged #2850), so the `container` branch of the consumer list is exercised only by the local
runs above, with `MODE=container` supplied by hand. Nothing in this PR makes a container entry
possible; it only makes sure the accounting is right when one becomes possible.

### DIFF
--- a/.github/workflows/node-repo-module-pack.yml
+++ b/.github/workflows/node-repo-module-pack.yml
@@ -72,6 +72,43 @@ name: node-repo module pack
 # A caller that ships neither scripts/affected-modules.py nor scripts/project-closure.py is simply
 # not narrowable: the scope script says FULL and says why. Nothing to opt into, nothing to skip.
 #
+# ──────────── The .NET SDK is a DECLARED need (`dotnet-sdk`) ────────────
+# 🚨 "make installing dotnet sdk a feature flag" (maintainer, 2026-08-31).
+#
+# `actions/setup-dotnet` used to run in this job the way it runs in every job in the fleet: as a
+# reflex, before anyone asked whether the job compiles anything. That is the wrong shape the moment
+# a module can be compiled by the platform IMAGE instead (`"build": "container"`, see above) —
+# because "the SDK is here" then silently answers a question nobody put: an unconverted step keeps
+# working on a runner SDK that the converted lane was supposed to have removed.
+#
+# So the install is now DECLARED, in two halves that must both be read to know what happened:
+#
+#   * `dotnet-sdk: install` (the DEFAULT) — install it, unconditionally. A caller that says nothing
+#     gets exactly today's behaviour, and a lane ADDED tomorrow inherits it. Opting out is a
+#     deliberate edit; drifting out is impossible.
+#   * `dotnet-sdk: by-declaration` — install it only if this job's own declared consumers still
+#     need it. Today they are: the compiler for any entry that is not `build: container`; the
+#     module-pack tool, which is `dotnet build` + `dotnet run` in EVERY mode; a `build: container`
+#     entry's binding-identity probe; and the module's own `dotnet test` suite when it ships one.
+#     🚨 So NO entry can drop the SDK yet, container or not — the packer alone keeps every pack job
+#     on it. That is not a placeholder, it is the measured state, and the log says it every run.
+#     The lane converts when the packer moves inside the platform image, not before.
+#   * anything else, the empty string included — RED. A flag that cannot be read resolving to
+#     "skip" is precisely the trapdoor this whole family of gates exists to forbid.
+#
+# 🚨 IT IS NOT A SKIP-TRAPDOOR, and three things make that structural rather than hoped-for:
+#   1. The DECISION step is unconditional — it runs on every event, in every mode, and it FAILS on
+#      an unreadable policy. Only `actions/setup-dotnet` itself carries an `if:`, and skipping an
+#      installer is not skipping a check: every build, pack, inspect and test step below still runs
+#      and still gates. The flag chooses a COMPILER; it never chooses whether to verify.
+#   2. When the answer is "no SDK", the job does not merely decline to install one — it PUTS A
+#      FAILING `dotnet` ON PATH. GitHub's ubuntu runners ship an SDK of their own, so "we skipped
+#      setup-dotnet" would otherwise mean "we quietly built against whatever .NET the image happens
+#      to carry", which is the same class of lie as a skipped gate rendering a green tick.
+#   3. The path taken is printed in the log AND in the job summary on EVERY run, with the consumers
+#      enumerated by name — including when the answer is "install", so the two outcomes cannot be
+#      told apart only by what is missing.
+#
 # Republishing is SAFE to repeat, which is why the schedule may simply publish rather than
 # first proving something changed: the registry is keyed by `{package}@{version}` from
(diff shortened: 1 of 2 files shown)


rbuergi:
## The same flag, proven on a real runner — in the caller repo

The shared lane's `container` path still cannot run anywhere (the pinned image has no `build-project`; #2850 is unmerged), so the flag's mechanism was proven on the **caller** side instead, where the identical decision step is wired into two ordinary jobs: **Systemorph/MeshWeaver.Plugins#1026**.

`memex-template` there — run [33362837011](https://github.com/Systemorph/MeshWeaver.Plugins/actions/runs/33362837011), **completed/success**:

```
declared consumers of a runner SDK in this job:
  • MeshWeaver.MemexTemplate.Test — 'dotnet test', a verb the image's build-project does not have at all (core #2850)
  • the generated template itself — the test's subject is a 'dotnet new' template that only an SDK can instantiate and build
.NET SDK on the runner: INSTALLED — policy 'install' (the default) — installed whatever the consumers say
##[group]Run actions/setup-dotnet@v5
```

The step ran unconditionally, the flag was read, the consumers were enumerated by name, and `setup-dotnet` then ran rather than being skipped.

## Design conserved

The second commit records the flag's shape, the three anti-trapdoor properties and the **per-lane blocker table** in [`InMeshBuildAndTest.md`](src/MeshWeaver.Documentation/Data/Architecture/InMeshBuildAndTest.md) — beside the builder it is about, and deliberately not in `PluginPackaging.md`, so there is no conflict with #2854.

The table's point is the one thing that must not be lost: **nothing can opt out today.** The packer keeps every pack job on the SDK, container entries included. Written down as a measurement rather than left for someone to mistake this flag for a working opt-out.

rbuergi:
The caller this flag was waiting for now exists: Plugins#1032 flips ALL 31 module entries to "build":"container" (acceptance green on the shelf image ci.7159; landing pending a strict-base lap + one ACR-transient rerun). Once it merges, the maintainer's standing directive is to remove the SDK install from the module lane — plan: land this flag, then declare `dotnet-sdk: by-declaration` in Plugins' two module-pack calls with the remaining consumers named explicitly (module `dotnet test` until the in-image test leg ships, and the runner-side asmver read). Will drive unless the branch owner objects.

Issue #2698, opened 2026-08-29
Module MVID skew: a publication's dependency records name one build of AI/Collaboration, the registry serves another under the same version — every consumer gate declines (root cause of the bake-shortfall saga)


rbuergi:
## Measured (MeshWeaver.Crm gate on `ci.6469`, identity `s770d75c…`, with `registry-modules: AI Essentials` composed)

```
fetched AI@1.1.0   fetched Essentials@1.0.18
Collaboration/Review  'MeshWeaver.Markdown.Collaboration' built against mvid:7084694526104da7b465b939718d7f31,
                      live is mvid:99611c85bc6b46febdea3b8def87160a
Edu/PromptCell        'MeshWeaver.AI' built against mvid:4ca2020e188d43c9995ff61eec737a65,
                      live is mvid:900d4208d4d44c0498ab090e56ea4563
(+ Hosting/Backup, Hosting/Deployment, Hosting/FleetConsole, Hosting/InstanceAction, Hosting/InstanceRequest,
   MyAi/Panel, Providers/ProvidersApp, RolePlay/Story — same pairs)
```

The sealed Plugins publication under `s770d75c…` was baked with module bundles from **its own run** (`module-artifacts`), whose MVIDs its dependency records name. A consumer gate composes the same modules by **version** from the registry's `/api/plugins/bundles/<pkg>/<ver>` — a **different build** of `AI@1.1.0` / `Essentials@1.0.18` (each Plugins run re-packs them; version strings do not move on a rebuild, MVIDs do). Same name, same version, different bytes ⇒ every bundle naming either module is declined and compiled locally.

## Why it looked like everything else first

- On the old pin (`ci.6353`) `MeshWeaver.AI` was composed all along and already skewed, so a run declined **1** assembly whenever its installed content happened to include an AI-dependent type — "random", but actually a function of which types the run held. (Reported earlier as a flake; it never was.)
- Composing Essentials moved the reason from `live is absent` to a genuine MVID mismatch — necessary, not sufficient.
- Not fixable from a consumer: composing the right module **name** is all a gate can do.

## Why a rebake is not the fix

A rebake against the currently-served builds is consistent until the next module publish re-packs them. The two channels — identity-keyed prebuilt bundles (the bake) and version-keyed module bundles (module-pack publish) — are produced by different runs and drift by construction.

## Durable fix: one source of module bytes, keyed by identity

[...]

PR #2889, opened 2026-08-31
feat(build-project): stage the Regex and Orleans generators — the container runs them

rbuergi:
> **Stacks on #2850** (`feat/cli-build-project`, the SDK-free `memex build project`) **and #2856**
> (the Razor generator, already merged INTO that branch). This branch was cut from
> `origin/integration/core-batch` — which carries #2850 plus `f63d19b9` — and then merged
> `origin/feat/cli-build-project` so the Razor work could be **generalised rather than duplicated**.
> It targets `main`, so the diff shown here includes those parents until they land. **The work in
> this PR is the last five commits**, from `6b7eb527` onward.
>
> 🚨 **This PR reads `CONFLICTING` against `main`, and the conflict is not in this change.** The only
> conflicted file is `src/MeshWeaver.Cli/BuildPluginCommand.cs`, which this PR **does not touch**
> (its twelve files are listed at the bottom). It is a real divergence between the *parent stack* and
> main: `main` has since refactored that command onto an `ImageRunner` and replaced
> `ComposeModulesFromSeed` with `ComposeSealedModules` (#2698/#2707), while the batch still carries
> the seed-scanning version from `0ebc13b4`. Choosing between those two module-composition designs
> belongs to the owner of #2850/#2851 — picking wrong there silently reintroduces the mvid-decline
> failure several rounds were spent on — so it is deliberately left alone. It resolves when the
> parents land. **Until then GitHub runs no check suite on this PR**, so the evidence below is the
> local evidence, all of it executed.
>
> Expect further mechanical conflicts with **#2855** (appends a positional parameter to `ProjectFile.Model`)
> and **#2860** (adds `init` properties) in `ProjectFile.cs` / `ProjectBuild.cs` / `Program.cs`. This
> PR only ever adds `init` properties, per `check-record-signatures.py`.

With Razor closed in #2856, **source generators were the largest remaining block** in the no-SDK
sweep of `MeshWeaver.Plugins/src` — and the diagnosis is the same one twice. A generator is not part
of the runtime: `[GeneratedRegex]`'s ships in the **.NET SDK's targeting pack**
(`Microsoft.NETCore.App.Ref/<version>/analyzers/dotnet/cs`), Orleans' in the
**`Microsoft.Orleans.CodeGenerator`** NuGet package. A runtime image has neither, so a project using
one compiled to an assembly missing exactly what the generator would have written.

The image now stages both in `generators/` beside the builder, and `build-project` runs them.

## The two failures are NOT the same shape, and that decides the design

| | what a missing generator looks like |
|---|---|
| **`[GeneratedRegex]`** | a wall of **CS8795** *"partial method must have an implementation part"* — loud, but it reads like broken source |
| **Orleans** | **nothing.** The project compiles GREEN and the assembly simply has no serializers, copiers or grain proxies. The silo throws at grain activation, days later, nowhere near the build |

So Orleans gets a rule regex does not need: a project referencing a package in
`StagedGenerators.CodegenRequiredPackages` whose generator is **not** staged **fails the build**,
naming what the green build would have hidden. `--accept generators-missing` is the recorded escape
and really does emit the incomplete assembly — asserted both ways.

## Which project gets which generator — the SDK's own rule

- **`generators/sdk/`** applies to **every** project, exactly as the targeting pack's analyzers do.
- **`generators/packages/<package id>/`** applies to a project whose `PackageReference` set names
  that id, exactly as a NuGet analyzer does.

🚨 The rule is deliberately **not** *"does the compilation resolve an Orleans type"*. This builder
hands every project the container's **whole** reference set — Orleans included — so a
reference-shaped test would run Orleans codegen over every project in the repo. Assemblies are
de-duplicated by file name across activations, because Orleans' codegen is reachable through more
than one package id and a generator that runs twice emits every type twice (CS0101, against code
nobody wrote).

## What was staged, and how each was obtained

| generator | from | staged as |
|---|---|---|
| `System.Text.RegularExpressions.Generator.dll` | the SDK's targeting pack on the build host — `$(NetCoreRoot)packs/Microsoft.NETCore.App.Ref/$(BundledNETCoreAppPackageVersion)/analyzers/dotnet/cs` | `generators/sdk/` |
| `Orleans.CodeGenerator.dll` | `PackageDownload` of `Microsoft.Orleans.CodeGenerator`, version read from the **central `Microsoft.Orleans.Sdk` `PackageVersion`** so there is ONE pin | `generators/packages/microsoft.orleans.sdk/` |

**The closure is one assembly each — established, not assumed.** Read from the AssemblyRef tables and
checked against the portal image's 209 libraries: the regex generator wants Roslyn, Immutable, Memory,
Buffers, **`Microsoft.CodeAnalysis.Workspaces`**, **`System.Composition.AttributedModel`** and
`CompilerServices.Unsafe`; `Orleans.CodeGenerator` wants Roslyn, Immutable, Memory, Buffers. The image
ships **every one of them**, Workspaces and Composition included — so unlike the Razor compiler
neither needs a private dependency beside it. **`Orleans.Analyzers` is deliberately NOT staged**: it
carries diagnostics and code fixes, not a `[Generator]`, and this builder runs no analyzers.

## 🚨 Trap 2 does not apply here, and that is a measurement

The SDK's Razor compiler is ReadyToRun per RID and had to be staged twice. **These two are plain
MSIL** — PE machine **`0x014C`**, measured inside the **`linux/amd64` and `linux/arm64`
`mcr.microsoft.com/dotnet/sdk:10.0.400` images** (the same source CD stages Razor from), against
`0xFD1D`/`0xD11D`/`0xEC20` for Razor. **One copy therefore serves both image architectures, and CD
needs no change at all** — no `docker create`/`docker cp` step, no second pin.

That is asserted rather than believed: `StagedGeneratorsTest` reads the PE machine of every staged
assembly and requires `0x014C`, so an SDK that starts crossgenning its analyzers turns **this** PR's
successor red instead of shipping an arm64 image that silently drops every generated regex. The
search path still probes `<rid>/` first, so re-splitting later is a staging change, not a code change.

## Trap 1 DOES apply — and it applied to `--generators` too

`--generators` used to route through the platform's node-compile discovery (`SourceGeneratorLoader`),
which reads a failed load as *"not a generator"*, logs it at **debug**, and
`RunSourceGenerators` then returns the compilation **unchanged**. An operator who supplied a
generator built against a different Roslyn got a **green build that ran none of it**. It now runs
through the same loud loader as everything else, with the project's own parse options.

**Measured, both builders, same command** (`--generators <dir of one non-generator assembly>` on
`MeshWeaver.Observability.Contract`):

| | what the operator sees |
|---|---|
| baseline | `source generators: 1 candidate assembl(y\|ies)` … then **11 raw CS8795 errors and no explanation** — the "a SOURCE GENERATOR did not run" sentence was *suppressed* precisely because `--generators` was non-empty |
| this PR | one error: *"--generators supplied assemblies holding no [Generator] type — they compile nothing, and the project would look merely broken."* |

Loud, and each with a test:

| refusal | why |
|---|---|
| a staged generator that fails to LOAD | it produces nothing, and the build would emit an assembly missing exactly that |
| a staged directory holding no `[Generator]` | staged by mistake; every project it served would look merely broken |
| a staged directory that is **empty** | a mis-staged image — not the same thing as an assembly already loaded via another package id |
| `--staged-generators` naming a non-root | falling back to the image's own copy would report on a generator nobody asked for |
| `--generators` that does not load | the regression above |
| Orleans referenced, generator unstaged | the green lie; `--accept generators-missing` |

#2856's host-first `AssemblyLoadContext` is now **shared** by both paths (`GeneratorLoader`) rather
than copied — `RazorGenerators` lost ~90 lines to it and behaves identically.

## An analyzer-only package is "supplied" when its generator is staged

`Microsoft.Orleans.Sdk` is **absent from the portal image's `deps.json`** (measured: 209 libraries,
`Microsoft.Orleans.Core` and its siblings among them, no `Orleans.Sdk`) because the publish prunes a
package with no runtime assets. Reading that as *"the container does not supply it"* would refuse the
one project in the fleet that authors grains, for a file that has never existed.

## Measured, 2026-08-31 — the sweep, honestly

Every non-test project in `MeshWeaver.Plugins/src` (**55** with a `.csproj`), inside
`memex-portal-ai@sha256:6f38db08…` on `linux/arm64`, with this branch's builder mounted at
`/builder` and the linux-arm64 Razor compiler staged as CD stages it. **Baseline is the same harness
run against this branch's parent** (`aa98f6f1`), not a remembered number.

| | baseline (`aa98f6f1`) | this PR |
|---|---|---|
| green, `--accept targets --accept embedded-resource` | **10** | **19** |
| green, `--accept razor-css-scope` as well | 20 | **32** |
| **CS8795 — a generator that did not run** | 18 | **0**, in either flag set, in no log |
| regressions | — | **0** |

The first row reproduces the previously reported **10** exactly, which is what makes the second
column comparable. No project that was green went red, in either row.

**The nine that flipped, all of them CS8795:**

`MeshWeaver.Markdown.Collaboration` (7 × `[GeneratedRegex]`, and the dependency that was blocking the
rest) · `MeshWeaver.Observability.Contract` (11 ×) · `MeshWeaver.Observability` ·
**`MeshWeaver.AI`** · `MeshWeaver.AI.Anthropic` · `MeshWeaver.AI.AppleIntelligence` ·
`MeshWeaver.AI.WebSearch` · `MeshWeaver.Mcp` · `MeshWeaver.Notifications.Channels`

**`MeshWeaver.AI` — the AI engine, 169 source files, 0 warnings under warnings-as-errors — now
compiles in the container.** So does every provider whose additional libraries the image happens to
carry.

🚨 **No ceiling is projected, and the count is not a gap count.** A refusal stops the load, so
closing generators reveals whatever sat behind it — five of the AI providers moved from "CS8795" to
"`--extra-refs` needed for `Azure.*` / `ClaudeAgentSdk` / `GitHub.*`", which is a *different, older*
category rather than a new problem. What can be said cleanly is that **CS8795 is now absent from the
sweep entirely**.

The remaining 36, by cause — every one of them pre-existing:

| cause | count | note |
|---|---|---|
| `--accept razor-css-scope` not passed | 13 | a flag, not a defect — that is the second row of the table above |
| additional libraries (`--extra-refs`) | 14 | ClosedXML, CsvHelper, Snowflake.Data, Microsoft.Data.Sqlite, `Azure.*`, ClaudeAgentSdk, `GitHub.*`, Microsoft.Graph, Radzen |
| constructs the evaluator refuses by name | 7 | `<Sdk>` element (AppHost), `<Protobuf>` (Grpc + 2 transitively), an `<Import>` above the mount (both portal hosts), a `ProjectReference` outside the source root |
| genuine compile errors | 2 | `Memex.Database.Migration` — CS0436, its own `DbVersion` against the copy baked into `Memex.Portal.Distributed` in the image; `MeshWeaver.LogWatcher` — one CS1061 on `IConfigurationSection.Get<T>` (it was 11 errors, 10 of them CS8795) |

*(counts are for the second, Razor-inclusive row — 23 remaining of 55)*

### The sweep got SLOWER, and that is the result working

`MeshWeaver.AI.ClaudeCode` used to finish in **0.8 s** — because `MeshWeaver.Markdown.Collaboration`,
its dependency, died on 7 × CS8795 and the cascade stopped. Nothing behind it was ever compiled. With
the generator staged, the same command compiles **`MeshWeaver.AI` — 169 source files, 0 warnings
under warnings-as-errors, a 1.1 MB assembly — in 282 s**, of which the generator itself is 240 ms.
The old speed was the speed of failing early.

## The framework build identity is untouched

No `ProjectReference` was added to `tools/MeshWeaver.PluginTester`, whose closure **is** the framework
identity. `PackageDownload` puts the Orleans analyzer in the package cache and contributes **nothing**
to this project's assets; the assemblies ride along as `Content` and are loaded at run time. Proved
three ways: `FrameworkBuildIdentityTest` (23 cases) green; the published `mw-plugin-test.deps.json` is
**byte-identical** to the baseline's; and the published tree differs by **exactly the two staged
files** and nothing else. `EmitPipeline.CreateCompilationOptions` and `OptionsFingerprint` are not
touched, and `MeshWeaver.Compiler` is not modified at all.

## Tests — 14 new, EXECUTED

`dotnet test test/MeshWeaver.PluginTester.Test -c Release` → **Passed! Failed: 0, Passed: 202**
(188 before, no regressions).

The positives assert against the **emitted assembly**, never against this PR's own source:

- the staged generator is found beside the builder and **loads** against this repo's Roslyn 5.6.0
  (the regex generator asks for 4.14.0.0), and the loaded generator is `RegexGenerator`;
- a real `[GeneratedRegex]` partial compiles, the emitted assembly is **loaded**, the implementing
  half of the partial method is found by reflection, and the regex is **called** and matched — a stub
  would pass a "does the method exist" test and fail this one;
- an Orleans `[GenerateSerializer]` type compiles and the emitted assembly carries **`OrleansCodeGen`
  types** and the assembly-level **`TypeManifestProviderAttribute`** — the only observable difference,
  since the compile is green either way;
- a project with nothing to generate emits **0** documents (staging must not move every assembly in
  the repo);
- the staged closure is one assembly each, in the expected directories, with a provenance manifest;
- every staged assembly is PE machine `0x014C`;
- the same generator reached through two package ids loads **once**;
- and each of the six refusals above names itself.

**Mutation proof.** With `generators/` moved aside in the test bin, **8 of the 14 go red** — and the
build-level one goes red asserting the message *names the generator and where it looked*, not that a
CS8795 wall appeared. Independently, the same project built against the baseline builder reports
`7 of those errors are unimplemented partial members — the signature of a SOURCE GENERATOR that did
not run`, while this branch reports `OK — 15 source file(s) + 1 generated document(s)`.

## Docs

`Doc/Architecture/InMeshBuildAndTest.md` gains *"The container runs the SDK and Orleans generators"* —
the two failure shapes, the activation rule, the measured closure, the `0x014C` finding, the
analyzer-only package rule and the refusal table — and both READMEs (`MeshWeaver.Cli`,
`MeshWeaver.PluginTester`) are updated.

## The twelve files this PR changes

`tools/MeshWeaver.PluginTester/` — `StagedGenerators.cs` (new), `GeneratorLoader.cs` (new),
`ProjectBuild.cs`, `ProjectFile.cs`, `Program.cs`, `RazorGenerators.cs` (refactored onto the shared
loader), `MeshWeaver.PluginTester.csproj`, `README.md` ·
`test/MeshWeaver.PluginTester.Test/` — `StagedGeneratorsTest.cs` (new), the `.csproj` ·
`src/MeshWeaver.Cli/README.md` · `Doc/Architecture/InMeshBuildAndTest.md`.

No workflow file changes: the staged generators are architecture-neutral, so **CD needs no step**.


rbuergi:
Heads-up from the queue drain (maintainer directive): merging current main into this branch now — it reads DIRTY against main after today's container-build wave landed (#2905 staged the Regex generator into the image, #2907/#2910/#2912/#2914 reshaped the same lane files). If the merge shows this branch's generator staging is fully superseded by what shipped, I'll say so here rather than resolve blindly; the branch owner has the final call on closing vs continuing.

rbuergi:
Merge attempted, aborted rather than resolved blindly: the conflicts are whole-file add/add (`ProjectFile.cs`, `RazorGenerators.cs`, `main-cd.yml`) — this branch predates the 2026-08-28 history rewrite, so there is no usable merge base.

Substance check against what shipped today:
- **The Regex-generator half is superseded**: #2905 stages the SDK's built-in generators (`System.Text.RegularExpressions.Generator` + `System.Text.Json` + the three interop generators) into the image as `sdk-generators/`, auto-discovered by `build-project` with no flag, gated by an executing test (`SdkGeneratorBuildTest`). All 31 Plugins module entries have since compiled green in-container on that image.
- **The Orleans-generator half is still novel**: `Microsoft.Orleans.CodeGenerator` ships via NuGet, not the ref pack, so #2905 does not cover it. If a module needs Orleans codegen under `build-project`, the clean move now is small: stage that one generator beside the others (the `sdk-generators/` landing spot and the staging pattern both exist), cut fresh from current main.

Recommend: close this branch and re-cut the Orleans piece (happy to do it on request) — a conflict resolution across unrelated histories would risk reverting today's shipped work.

rbuergi:
Unrebaseable after tonight's builder rewrite (16 conflicted files: single body pass, multi-entry workspace, shared reference universe, closure manifests — #2925/#2931/#2932/#2933 — all touch the same code). The FEATURE remains wanted: staging the Regex/Orleans generators beside razor-generators/+sdk-generators/ closes a real gap the CS8795 refusal documents. Re-raise it as a fresh, small change against the new ProjectBuild — the staging seams (GeneratorPipeline, sdk-generators discovery) survived the rewrite intact.

PR #2917, opened 2026-08-31
feat(layout): NavigateToSidePanel — a click can open a thread beside the page

sierragolflima:
## The defect this enables fixing

In AgenticBusiness lesson 3 (and every TryIt embed), sending a prompt to the agent **navigates the whole page** to the newly created thread — "the AI chat takes over the entire screen" — losing the exercise page that told the learner to come back and compare results. Reported 2026-08-31.

The embed had no alternative. The framework carries the complete side-panel navigation chain end to end — `NavigationRequest.Target`, the portal handler forwarding it into `NavigationOptions`, `NavigationService` firing `SidePanelNavigationRequested` on `Target == "SidePanel"`, and `PortalLayoutBase` opening the chat panel on the path — but `LayoutAreaHost` exposed only `NavigateTo(uri, forceLoad, replace)`, which posts the request with **no Target**. Server-side area code could reach every link of the chain except the first one.

## The change

1. `LayoutAreaHost.NavigateToSidePanel(uri)` — posts the `NavigationRequest` with `Target = "SidePanel"`; the existing `NavigateTo` is refactored onto a shared private `PostNavigation`, behaviour unchanged.
2. A matching `UiActionContext.NavigateToSidePanel` extension beside the existing `NavigateTo`.

**Deliberately a new method, not an optional parameter on the existing one:** adding a parameter replaces the method's signature, and module DLLs compiled against `NavigateTo(string, bool, bool)` would throw `MissingMethodException` at runtime — the image/module atomicity hazard the plugins repo documents ("the image and the full set of landed modules must move atomically").

## Verification

`NavigateToSidePanelTest` pins the behaviour with a **real hub round-trip** (no mocking): a rendered button whose click calls the new API, a real `ClickedEvent` posted from the client, and the assertion that the `NavigationRequest` arriving at the subscriber carries `Target = "SidePanel"`.

TDD, all three colours:
- **RED** before the API existed (`CS1061`).
- **GREEN** with it.
- **RED under mutation** — dropping `Target` from the implementation fails the test with its own message: *"an untargeted request navigates the whole page — the take-over this API exists to avoid"*.

Full `MeshWeaver.Layout.Test` suite: **437/437** under `-c Release -warnaserror`, 0 warnings.

## Follow-up (dependency order)

The consumer lands in `MeshWeaver.Education` after this merges: `TryItAreas.Send` switches `click.NavigateTo("/" + created.Path)` → `click.NavigateToSidePanel(...)`, fixing every TryIt embed in one call site. That PR carries the user-facing What's New entry; this one is API-only.

### DIFF
--- a/src/MeshWeaver.Layout/Composition/LayoutAreaHost.cs
+++ b/src/MeshWeaver.Layout/Composition/LayoutAreaHost.cs
@@ -1533,13 +1533,34 @@ public record LayoutAreaHost : IDisposable
     /// <param name="forceLoad">Whether to force a full page reload.</param>
     /// <param name="replace">Whether to replace the current history entry instead of adding a new one.</param>
     public void NavigateTo(string uri, bool forceLoad = false, bool replace = false)
+        => PostNavigation(new NavigationRequest(uri) { ForceLoad = forceLoad, Replace = replace });
+
+    /// <summary>
+    /// Opens the specified URI in the portal's side panel — the page the viewer is on stays where
+    /// it is. This is the missing first link of a chain the framework already carries end to end:
+    /// <see cref="NavigationRequest.Target"/> → the portal's handler → the navigation service's
+    /// side-panel event → the layout opens the chat panel on the path. Use it wherever a click
+    /// creates or references a thread the viewer should see BESIDE their work — an embedded
+    /// exercise composer navigating the whole page to the thread it just opened loses the very
+    /// page the exercise told the viewer to come back to.
+    ///
+    /// <para>A NEW method rather than an optional parameter on <see cref="NavigateTo"/>: adding a
+    /// parameter replaces that method's signature, and module DLLs compiled against
+    /// <c>NavigateTo(string, bool, bool)</c> would throw <c>MissingMethodException</c> at runtime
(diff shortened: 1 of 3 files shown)


sierragolflima:
CI failure triage for run 33426192391 — both failures are outside this diff (two additive files in MeshWeaver.Layout + one new test):

1. **`MeshWeaver.FutuRe.Test` — `exit=139 SIGNAL SIGSEGV`** (host died on a signal). This is the documented CoreCLR GC crash family (/sigsegv skill: FutuRe, eight sightings, "Not ours"; the collectible-ALC hypothesis falsified three ways). A crashed harness is an explicit re-run exception in AGENTS.md.
2. **`MeshWeaver.Autocomplete.Test.AutocompleteMultiSourceTest.ScoreOrdering_MergedResults_SortableByPriority`** — `TESTFAIL`. Passes locally against this branch (`--filter`, 2 s, after a full Release build of the project). A multi-source merge assertion with a 30 s reactive window, in a project this diff does not reference; `main` is green on the five most recent runs. The race pre-exists this PR and deserves its own investigation rather than being masked here — noting it so the rerun is on the record, not silent.

Re-running the failed jobs only.

PR #2941, opened 2026-09-01
The indexing pipeline, the map control and Cornerstone leave the content surface

rbuergi:
Wave B of the carve-out, platform half. 🔗 Paired with **Systemorph/MeshWeaver.Plugins#1057 — that one merges FIRST**, or the assemblies exist in neither place.

## One identity flip, three projects

`FrameworkBuildIdentity.ContentSurfaceAssemblies` is the boundary that matters. Wave A's two projects were **outside** it, which is why they cost a `.csproj` edit. `MeshWeaver.Maps`, `ContentCollections.Indexing` and `.Indexing.Graph` are **inside** it, so each is the `MeshWeaver.AI` / `Markdown.Collaboration` move — and leaving the surface costs a **framework identity flip plus a fleet re-bake**.

They therefore go in one change. The flip is the cost, not the count.

## Why Maps could not leave alone

`samples/Graph/Data/Cornerstone/Pricing/Source/PricingLayoutAreas.cs` binds `MapControl` — and it is a `Source/*.cs` **mesh node**, compiled at runtime in the portal, invisible to `dotnet build`, and baked into the shipped image.

🚨 That is why **five test projects referenced `MeshWeaver.Maps` while not one line of their own C# mentioned it**: their test meshes compile that sample. The `.csproj` graph misleads in both directions here — four references turned out to carry no code dependency at all (`Memex.Portal.Shared` builds `-warnaserror` clean without `Maps`, `Indexing`, `.Indexing.Graph` *and* `Observability.Contract`), while the consumer that actually pinned Maps appears in no project file.

So the content travels with the module, exactly as `Markdown.Collaboration` took its Review sources. Cornerstone is a `Store/Plugin` package in the plugins repo now, and its six suites pass **114/114** there.

## The contract split

`Mesh.Operations` stays here and needs exactly one type. `DocumentPaths` moves to **`MeshWeaver.Mesh.Contract`**, keeping its `MeshWeaver.ContentCollections.Indexing` namespace — every `using` still resolves and no consumer changed.

`Mesh.Contract` rather than `ContentCollections` because the indexing base uses `DocumentPaths` too (`ContentChunkSearch`), so `ContentCollections` would have handed the pipeline a new dependency on Layout and Markdown — precisely what its own "storage-agnostic core" note rules out. Both remaining consumers already reference `Mesh.Contract`, so the split adds **no edge to the graph**.

Leaving the surface is safe for Indexing because **nothing in-mesh binds it** — verified across every gated sample tree and the doc data, where the only occurrences are prose in `.md` files.

## 🚨 A guard that would have checked nothing

`FrameworkBuildIdentityTest.SurfaceIdentity_RecordsAbsence` proved "a host missing a canonical assembly is a different surface" by removing `"MeshWeaver.Maps"` from the baseline. With Maps out of the baseline that `Remove` removes nothing, and the test would have compared two identical dictionaries — **passing while proving nothing about absence.**

It now names an assembly that is staying *and* asserts the removal actually happened, so the next name to go stale fails loudly instead of silently.

## Also

`AddCornerstone` leaves `SampleDataExtensions` (declaration + its `AddAllSampleData` entry) — a one-line `IncludePartition` wrapper the moved suites call directly now. `TestUsers` keeps its Cornerstone grant on purpose: it is the shared fixture **both** repos build against, and the suites needing it run in the other one.

## Verification

| | |
|---|---|
| `Mesh.Contract`, `Mesh.Operations`, `Portal.Shared` | Release `-warnaserror`, **0 Warning(s) 0 Error(s)** |
| `Content.Test`, `Persistence.Test`, `Hosting.Monolith.Test`, `PluginTester` | **0 Error(s)** after the deletions |
| `FrameworkBuildIdentity` | **23/23** with all three out of the surface |
| Dangling references | none — 0 tracked files left, 0 `.csproj`/`.slnx` hits |

The survey behind both waves, including how a sample Space becomes a package, is in `Doc/Architecture/CarvingProjectsOutOfCore`.

https://claude.ai/code/session_01XM3bKbVAFCCDH9ZyYCxKQy


rbuergi:
🚨 **The merge order changed — this PR now has a precondition: #2943.**

Plugins#1057's CI surfaced a circular dependency this pair could not resolve on its own. The plugins copy of the indexing pipeline *binds* `DocumentPaths` (`ContentChunkSearch.cs`), so it needs the type in `Mesh.Contract` before it can compile — while this PR cannot delete the pipeline until plugins ships it. Each was the other's precondition, and neither would ever have gone green.

**#2943 breaks the cycle** by moving `DocumentPaths` to `Mesh.Contract` and leaving a `TypeForwardedTo` behind, so every tree compiles at every point.

New order: **#2943 → Plugins#1057 → this PR.**

This branch still carries the `DocumentPaths` relocation itself; once #2943 lands I'll rebase, which drops it here and leaves this PR as what it should be — the deletions plus the content-surface change.

Worth recording why local verification missed it: I built the plugins copy against my own core worktree, which already had the move. CI builds against `origin/main`, which did not. A throwaway worktree pinned at `origin/main` is the honest reference, and is what caught it.

Nothing here is landing today regardless — held pending a decision on flipping framework identity before Friday's talk.

rbuergi:
Holding this out of the merge queue — it would turn MeshWeaver.Plugins red the moment it lands, and its counterpart does not prevent that. Re-arm as soon as the Plugins side is rewired; nothing here is wrong, the two halves are just not safe in this order.

## What happens if this merges now

This PR removes `src/MeshWeaver.Maps` from core (373 deletions, the directory is gone on the branch). Plugins CI checks out `Systemorph/MeshWeaver` in three places with **no `ref:`** — so it builds against core `main`, live, not a pin:

```
.github/workflows/ci.yml:529,1066,1279   repository: Systemorph/MeshWeaver
                                          path: meshweaver          # no ref: → default branch
```

And Plugins `main` carries **11 unconditional** references to the project this PR deletes:

```
$ git grep -c "MeshWeaverRoot)/src/MeshWeaver.Maps" -- "*.csproj"   # on Plugins main
11 files, 0 guarded by Exists()
```

`Memex.Portal.Gui` (the portal host), `MeshWeaver.AI.Test`, `Blazor.AppleMaps`, `Blazor.GoogleMaps`, `Blazor.OpenStreetMap`, `Northwind.Application`, and five migrated test suites. Every one fails at restore, so Plugins `main` and all 21 open Plugins PRs go red together.

## The counterpart does not cover it

Plugins #1057 brings `src/MeshWeaver.Maps` in and adds **one** local `ProjectReference`. Measured on its head: **11 files still hold the `$(MeshWeaverRoot)` form**, so landing #1057 does not fix the other ten either.

## Ordering

`MeshWeaver.Maps` has to exist in Plugins *before* it stops existing in core, and all 11 references have to move with it. Core CI cannot see this — it does not build Plugins; the coupling only shows up in CD (core builds `portal-ai` from `plugins@main`), which is after the fact.

I am fixing the Plugins side now: the 11 references get the dual-path shape this repo already uses for the `MeshWeaver.Fixture` move, so they resolve whichever half has landed and the ordering stops mattering. I will re-arm this PR here when that is in.

Dequeuing costs the three PRs ahead of this one a rebuild, which is the cheaper mistake.

rbuergi:
Correcting a number in my comment above: **on Plugins `main` it is 6 unconditional references, not 11.**

The 11 was measured on #1057's head, which adds the reference to five migrated test suites as well. On `main` today:

```
$ grep -rn "MeshWeaverRoot)/src/MeshWeaver.Maps" src --include="*.csproj" | grep -c ProjectReference
6      # Memex.Portal.Gui, MeshWeaver.AI.Test, Blazor.AppleMaps,
       # Blazor.GoogleMaps, Blazor.OpenStreetMap, Northwind.Application
```

The conclusion is unchanged — six unconditional `ProjectReference`s to a project this PR deletes, none guarded by `Exists()`, so Plugins `main` still fails at restore the moment this lands. But the blast radius I quoted was too large and the five test suites were the wrong half of the evidence.

Also worth noting for whoever lands #1057: on its branch the count goes to 11, so the rewiring has to cover the five it adds too, not just the six here.

rbuergi:
Retracting my own correction — **the original number, 11, was right.** Apologies for the churn; here is the cause, because it matters more than the number.

My "it is 6, not 11" comment was measured with `grep` in my primary Plugins checkout, which turns out to be **216 commits behind `origin/main`**. Five of the eleven csproj files gained the reference in commits I did not have locally, so the stale tree honestly showed 6. Measured against the ref instead of the working directory:

```
$ git grep -l "MeshWeaverRoot)/src/MeshWeaver.Maps" origin/main -- "src/**/*.csproj" | wc -l
11
```

Memex.Portal.Gui · MeshWeaver.AI.Test · Blazor.AppleMaps · Blazor.GoogleMaps · Blazor.OpenStreetMap · Northwind.Application · Content.Test · Hosting.Monolith.Test · Persistence.Test · Query.Test · Security.Test.

So the blast radius is the larger one, and the five migrated test suites were the right half of the evidence after all.

**The fix is in flight and covers all eleven** — each reference becomes the dual-path shape this repo already uses for the `MeshWeaver.Fixture` move, with `Exists()` anchored to `$(MSBuildThisFileDirectory)` so it does not depend on the working directory:

```xml
<ProjectReference Include="../MeshWeaver.Maps/MeshWeaver.Maps.csproj"
                  Condition="Exists(...)" />
<ProjectReference Include="$(MeshWeaverRoot)/src/MeshWeaver.Maps/MeshWeaver.Maps.csproj"
                  Condition="!Exists(...)" />
```

That resolves against whichever half has landed, so this PR and plugins#1057 stop needing a specific merge order across two repos. I will re-arm this one once it is in.

rbuergi:
Unblocked — re-arming. plugins#1092 has merged, so all 11 Plugins references to `MeshWeaver.Maps` now resolve through the dual-path `Exists()` guard and pick up the local copy or core's, whichever is present. The cross-repo ordering constraint that made this unsafe is gone; this can land in any order relative to plugins#1057.

rbuergi:
### Downstream check: two of the three deleted projects still have **unconditional** references in `MeshWeaver.Plugins`

Measured just now by the new `Hosting/RepoHealth` cross-repo rule (Systemorph/MeshWeaver.Plugins#1101) over `MeshWeaver.Plugins@main` — 567 `$(MeshWeaverRoot)` references across 116 project files, against all 14 open PRs here:

| This PR deletes | Unconditional refs in Plugins | |
|---|---|---|
| `src/MeshWeaver.Maps/MeshWeaver.Maps.csproj` | 0 (11 total) | ✅ already hardened with the `Exists()` fallback |
| `src/MeshWeaver.ContentCollections.Indexing/…csproj` | **2** | 🔴 |
| `src/MeshWeaver.ContentCollections.Indexing.Graph/…csproj` | **2** | 🔴 |

The two red rows are:

- `MeshWeaver.Plugins` → `src/MeshWeaver.AI/MeshWeaver.AI.csproj` lines **51–52**
- `MeshWeaver.Plugins` → `src/Memex.Portal.Gui/Memex.Portal.Gui.csproj` lines **103–104**

Both reach for `$(MeshWeaverRoot)/src/MeshWeaver.ContentCollections.Indexing{,.Graph}` with no condition, while `MeshWeaver.Plugins` **already carries its own copy** of both projects. Neither repo's CI can see this: this repo does not build Plugins, and Plugins checks this repo out with no `ref:` at four sites in its `ci.yml`, so it builds against this repo's moving `main`.

When this lands, restore fails for the AI engine and for `Memex.Portal.Gui` — i.e. the `Build + test the portal hosts` job on the Plugins trunk.

The Maps references in that repo already carry the pattern that solves it (prefer the local `../` copy when it exists, fall back to `$(MeshWeaverRoot)` otherwise, with a comment saying to drop the fallback once this PR has merged). Applying the same two-line treatment to those four references ahead of the merge should close the window — but the sequencing is yours, so flagging rather than changing.

Not blocking anything from here; just making the downstream visible before it becomes a red trunk.

rbuergi:
Now safe in either order — all three of this PR's deletions are covered on the Plugins side.

My earlier comments only covered `MeshWeaver.Maps`. **This PR deletes three projects**, and I had audited for the one name in front of me. `Hosting/RepoHealth`'s cross-repo rule (plugins#1101) found the other two on its first real run:

| project | unconditional refs on plugins main | covered by |
|---|---|---|
| `MeshWeaver.Maps` | 11 | plugins#1092 (merged) |
| `MeshWeaver.ContentCollections.Indexing` | 2 | **plugins#1104 (merged)** |
| `MeshWeaver.ContentCollections.Indexing.Graph` | 2 | **plugins#1104 (merged)** |

The two I missed were `src/MeshWeaver.AI/MeshWeaver.AI.csproj:51-52` and `src/Memex.Portal.Gui/Memex.Portal.Gui.csproj:103-104` — the AI engine and the portal host, so the failure would have landed on `Build + test the portal hosts`, which is now a required context on that repo.

All four now use the dual-path `Exists()` shape, anchored to `$(MSBuildThisFileDirectory)`. Both projects already exist in Plugins, so they resolve locally the moment this lands.

Nothing further blocks this from my side.

Issue #1446, opened 2026-08-13
A starved security-policy read leaves CreateNodeRequest hanging for the full timeout instead of failing fast


rbuergi:
Found while investigating a CI failure in `SourceDiscoveryUnavailableTest` (fixed test-side in #1427). The **test artifact** is fixed; the **product behaviour underneath is not**, and it is worth a look on its own.

## What happens

Creating a node runs the security layer's policy and grant reads, which are `namespace:`-shaped queries over the target subtree:

```
namespace:{…}/Source          id:_Policy   nodeType:PartitionAccessPolicy
namespace:{…}/Source/_Access               nodeType:AccessAssignment
```

When those reads **starve** — the cross-silo case this suite exists for (#1218) — the create does not fail fast. It sits waiting. Observed in the hub snapshot at teardown:

```
Hub portal/nodeops-… RunLevel=Started Disposal=Pending
  Queue(buffer=2, …) Executing(CreateNodeRequest, 33014ms)
```

33 seconds, still executing, on reads that could never answer.

## Why it matters beyond the test

The faulted-query eviction already works — the log shows both queries evicted so the next `GetQuery` opens a fresh upstream:

> its Replay(1) had latched the terminal error and would have replayed it to every future subscriber for the life of the process

So the **query** recovers. The **in-flight caller does not** — it is still blocked on the read that faulted. In a real cross-silo starvation, node creation into the affected namespace would hang this way for every caller until something above it times out.

## Suggested shape

A starved security read is an availability failure, and the codebase already has the vocabulary for exactly this distinction — `SourceDiscoveryUnavailableException` / `CompilationStatus.Unavailable` were introduced (#1218) so a starved *source* read is not misreported as a compile verdict. The same reasoning applies here: a create whose permission check could not be **established** should fail fast and say so, rather than block.

## Repro

Not currently reproducible from the suite — #1427 narrowed the stand-in to source discovery precisely so the security reads stop starving. To see it again, widen `StarveMarker` in `SourceDiscoveryUnavailableTest` to match the namespace alone (drop the `nodeType:Code` clause) and watch the `CreateNode` at the top of the test.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #2945, opened 2026-09-01
An Editor can now delete the satellite it was allowed to publish — one rule, one answer (#2913)

### LINKED ISSUE #2913: An Editor can publish a satellite it cannot delete: the delete pipeline bypasses INodeTypeAccessRule
Surfaced by **executing** the writer while building session presence (MeshWeaver.Plugins#976 / #1031), not by reading the code. Separable from that work, which routes around it.

## The inconsistency

`SatelliteAccessRule` maps write-class operations on a satellite to **`Permission.Update`** on its `MainNode`. But the delete pipeline never consults the rule: `MeshExtensions.CheckDeletePermissionForNode` demands **`Permission.Delete`** on `MainNode`.

`Role.Editor` has `Update` and **not** `Delete`.

## Why it matters beyond a permission asymmetry

For anything that publishes a satellite representing *ongoing state*, this creates a state a consent model must never allow: **you may start it and be refused when stopping it.**

In the presence case an Editor could publish their live location onto a node and then be denied when erasing that row. "I can turn it on but not off" is the exact failure a revocable-consent feature exists to prevent, and it is reachable today by an ordinary Editor.

## How #1031 avoids it (and why that is a workaround, not a fix)

`Start` now pre-flights `CanWithdraw` and **refuses to start where it could not revoke**. That is correct and conservative — better to decline the feature than to offer irrevocable publication — but it costs Editors a capability they should have, purely because two code paths disagree about which permission governs a satellite write.

If the delete decision routed through `INodeTypeAccessRule` like the other write-class operations, that pre-flight relaxes to `Update` and the asymmetry disappears.

## Suggested direction

Route `CheckDeletePermissionForNode` through `INodeTypeAccessRule` so a NodeType's own rule decides, with the current `Permission.Delete` demand as the default when no rule applies. That keeps today's behaviour for plain nodes while letting a satellite's rule govern its own lifecycle — which is what `SatelliteAccessRule` already claims to do for every other write.

Whatever the resolution, the two paths should agree, and a satellite rule that governs create/update but silently not delete is the kind of split that only shows up when someone runs it.

## Provenance

Found by `SessionPresenceAccessTest` driving the real `MeshPresenceWriter`/`RunAs` on a closed-by-default monolith mesh, under Editor, Viewer, an unentitled signed-in account and Anonymous.

rbuergi:
Closes #2913

## The defect: one question, two answers, both on the same request

`Role.Editor` holds `Update` and **not** `Delete`. `SatelliteAccessRule` maps a satellite's Delete to `Permission.Update` on its `MainNode` — creating a satellite is a modification of its main node, and so is removing it. So an Editor may publish a satellite. But it could not erase one:

- `RlsNodeValidator` routes Read / Create / Update / **Delete** through the node type's registered `INodeTypeAccessRule`.
- `HandleDeleteNodeRequest`'s **pre-flight** — `MeshExtensions.CheckDeletePermissionForNode`, the first gate a `DeleteNodeRequest` meets on the off-router node-operation hub where `IMeshService.DeleteNode` lands — demanded `Permission.Delete` on `MainNode ?? Path` outright and **never looked at a rule**.

Both are correct in isolation. The effective policy was their **conjunction**, so wherever a rule was *more permissive* than `Permission.Delete` the rule silently did not apply. "I can turn it on but not off" is the exact state a revocable-consent feature exists to prevent, and it was reachable by an ordinary Editor. It surfaced only when someone *ran* the writer (session presence, MeshWeaver.Plugins#1031) — never by reading either file.

## The fix is structural, not a second copy of the lookup

`NodeTypeAccessRuleSet` (`src/MeshWeaver.Mesh.Contract/Services/NodeTypeAccessRuleSet.cs`) is now **the one index** — a mesh-scoped singleton registered in `MeshBuilder`, holding the selection semantics in one place (keyed by node type, case-insensitive, last registration wins, `SupportedOperations` honoured). Both `RlsNodeValidator` and the delete pre-flight resolve their rule from it, so the two paths cannot drift apart again. Two code paths answering one question was the actual defect; fixing only the symptom would have left it live.

## What widened — precisely — and what did not

| Case | Before | After |
|---|---|---|
| Type with a Delete-supporting rule that **allows** | rule ∧ `Permission.Delete` | the rule — **this is the fix** |
| Type with **no rule** | `Permission.Delete` | `Permission.Delete` — byte-for-byte, closed by default |
| Rule that **denies** | refused | refused (now at the pre-flight, so `Unauthorized` rather than `ValidationFailed`) |
| Rule that **faults** / reaches no verdict | n/a | refused as `Unavailable` — **fail-closed** |
| `WellKnownUsers.System` | allowed via the fold's `Permission.All` | allowed before any rule is consulted (unchanged outcome) |

The only types that gained anything are those whose rule *chose* a lighter demand: the `SatelliteAccessRule` family — `Comment`, `TrackedChange`, `Kernel`, `Activity`, `UserActivity`, `ActivityLogSegment`, and in MeshWeaver.Plugins `Thread`, `ThreadMessage`, `TokenUsage`, `Portal`.

🚨 **`Space` and `Partition` were the trap.** Their rules also said `Update` for Delete — unreachable, because the pre-flight's own `Permission.Delete` demand sat in front of it. Making the rule authoritative would have let anyone with `Update` on a Space **delete the Space and drop its backing store**. Both rules now state `Permission.Delete` for Delete explicitly, which is exactly the policy that was already enforced. **A disagreement between two gates resolves to the STRICTER of the two unless there is a stated reason for the looser one** — and for satellites the issue states it.

## Fail-closed, deliberately

A rule reaches the same starve-able permission fold every other check does, so it can fault or complete without emitting. Neither is permission to proceed: both yield `DeletePreflight.Unestablished`, the delete is refused, and the response says `NodeDeletionRejectionReason.Unavailable` — an availability failure, not a statement about the caller's rights (#1446). There is deliberately **no `.Catch(_ => true)`** on this path; the identical instinct on the security-fold twin is what made a group deny fail OPEN (#2011). The same third terminal is added to the no-rule leg, where an empty fold previously posted **no response at all** and the caller waited out its own timeout (the #2742 shape).

## Not changed, on purpose

`DeleteNodeRequest` / `ValidateDeleteRequest` carry `[RequiresPermission(Permission.Delete)]`, so on a route through a hub with the `AccessControlPipeline` (a per-node hub) the delivery gate still demands `Permission.Delete` on the receiving hub's own path. Widening the message-level gate is a separate decision with a much larger blast radius; it has not been made.

## Verification

`test/MeshWeaver.Security.Test/DeleteHonoursNodeTypeAccessRuleTest.cs` pins all four rows of the table, using the **real** `SatelliteAccessRule`. **Each was proven falsifiable**:

- rule lookup removed from the pre-flight (the pre-fix behaviour) → tests 1, 3, 4 red;
- the no-rule fallback relaxed from `Delete` to `Update` (an over-permissive fix) → test 2 red — the no-widening test, the one that matters most.

Green locally, Release + `-warnaserror`, `0 Error(s)` on every touched project: Graph 1578/1578 · Security 378/378 · NodeOperations 76/76 · Hosting 269/269 · AccessControl 21/21 · Documentation 147/147 (link integrity + the timeout ratchet — the new test adopts `TestTimeouts`, it does not re-seed the baseline) · Hosting.Monolith delete + node-operation-target suite 37/37.

## Work product conserved

The two-paths-one-question finding, the widening table and the still-unrouted message gate are committed to `Doc/Architecture/AccessControl` → *"One question, ONE answer"*. What's New entry: `Category: Fix`.

https://claude.ai/code/session_01Kwar1ZGt5FMEpHiXmkzKCm

### DIFF
--- a/src/MeshWeaver.Documentation/Data/Architecture/AccessControl.md
+++ b/src/MeshWeaver.Documentation/Data/Architecture/AccessControl.md
@@ -1215,6 +1215,11 @@ statement about the caller's rights (`Unavailable`, above). There is deliberatel
 `.Catch(_ => true)` on this path: the identical instinct on the security-fold twin is what made a
 group deny fail OPEN.
 
+The caller-visible detail names the exception **type** and never its **message**. A rule is arbitrary
+code, and its exception text can carry an internal path, a connection string or another tenant's
+identifier — while the type is all a caller can act on ("retry, or ask an operator"). The full
+exception, message and stack, goes to the log where only an operator sees it.
+
 **What is still NOT routed through the rule, on purpose.** `DeleteNodeRequest` and
 `ValidateDeleteRequest` carry `[RequiresPermission(Permission.Delete)]`, so on a route that passes
 through a hub with the `AccessControlPipeline` (a per-node hub) the delivery gate still demands
--- a/src/MeshWeaver.Documentation/Data/Architecture/AccessControl.md
+++ b/src/MeshWeaver.Documentation/Data/Architecture/AccessControl.md
@@ -1159,6 +1159,74 @@ For the common "predicate over `(context, userId)`" case you don't write a class
 - Other identities are denied — the standard AccessAssignment check is **not** performed for VUser nodes.
 - Delete operations are not covered by this rule and fall through to standard RLS.
 
+## 🚨 One question, ONE answer — every gate resolves the rule through `NodeTypeAccessRuleSet`
+
+**A rule only governs a node type if EVERY path that decides that node type's access consults it.**
+Until #2913 one did not, and the gap was invisible from either side.
+
+- `RlsNodeValidator` built its own `nodeType → rule` dictionary and routed Read / Create / Update /
+  **Delete** through it.
+- The delete handler's **pre-flight** — `MeshExtensions.CheckDeletePermissionForNode`, the first gate
+  a `DeleteNodeRequest` meets — demanded `Permission.Delete` on `MainNode ?? Path` outright and never
+  looked at a rule at all.
+
+Both are correct in isolation and both are on the same request. The effective policy was their
+**conjunction**, so wherever a rule was *more permissive* than `Permission.Delete` the rule silently
+did not apply. `SatelliteAccessRule` maps a satellite's Delete to `Permission.Update` on its
+`MainNode` — creating a satellite is a modification of its main node, and so is removing it — and
+`Role.Editor` holds Update and **not** Delete. So an Editor could **publish a satellite onto a node
+and then be refused when erasing it**: "I can turn it on but not off", the exact state a
+revocable-consent feature exists to prevent, reachable by an ordinary Editor. It surfaced only when
+someone *ran* the writer (session presence, MeshWeaver.Plugins#1031), never by reading either file.
+
+The fix is structural, not a second copy of the lookup: **`NodeTypeAccessRuleSet`
+(`src/MeshWeaver.Mesh.Contract/Services/NodeTypeAccessRuleSet.cs`) is the one index**, a mesh-scoped
+singleton registered in `MeshBuilder`, and both `RlsNodeValidator` and the delete pre-flight resolve
+their rule from it. Selection semantics live in one place: keyed by node type, case-insensitive,
+last registration wins, and a rule applies only when its `SupportedOperations` is empty or contains
+the operation.
+
+**What that widened, precisely — and what it did not:**
+
+| Case | Before | After |
+|---|---|---|
+| Node type with a Delete-supporting rule that ALLOWS | rule ∧ `Permission.Delete` | the rule (this is the fix) |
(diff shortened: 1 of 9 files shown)


rbuergi:
Addressed the Copilot finding in 4e5ef7b9e: the caller-visible detail now names the exception **type** only (`the 'X' access rule failed (InvalidOperationException)`) — the full exception, message and stack, stays on the `LogWarning` where only an operator sees it.

Turned it into a guard rather than just a fix: the faulting test rule throws a distinctive marker string, and `AFaultingRule_RefusesTheDelete_AndReportsUnavailableNotDenied` now asserts the response contains `InvalidOperationException` and does **not** contain that marker. Proven falsifiable — restoring `{ex.Message}` reddens it.

Note on scope: the neighbouring `RlsNodeValidator.UnestablishedCheck` still includes `cause.Message` in its result. That is pre-existing behaviour on an error contract other code may match on, so tightening it is a separate decision; this change keeps the new path the stricter of the two.

PR #2963, opened 2026-09-01
test(orleans): one Orleans test base, not two — the cluster becomes a DECLARATION

rbuergi:
Facility 3 of the decentralise-tests programme: **eliminate `OrleansTestBase` and `OrleansSharedTestBase`.**

> 🧱 **Stacks on #2961** (the `IMeshBootstrap` seam), which is open and green. Its two commits ride along in this diff until it merges; GitHub drops them from the diff automatically once it does. The seam is **not** duplicated here.

## What they actually were

Not two designs — the **same machinery written twice**. Both deployed a `TestCluster` through `OrleansTestCluster.DeployAsync`; both created a client hub with a seeded `AccessContext` and a routing-stream registration; both tracked those hubs and joined their teardown before the silos went down. What differed was three **values** — which silo configurator, how many silos, leased or dedicated — and values do not need a second base class. Measured: **23 subclasses, 20 of which overrode nothing at all.**

Both are gone, replaced by one `OrleansMeshTestBase` that says which mesh it wants through the same `IMeshBootstrap` seam `MonolithMeshTestBase` uses, so the declaration reads identically on both sides:

```csharp
protected override IMeshBootstrap Bootstrap => MeshBootstrap.Orleans(o => o.WithSilos(2));
protected override Type SiloConfiguratorType => typeof(MySiloConfigurator);
```

`OrleansBootstrap.Applicator` is armed here from a `[ModuleInitializer]`. That hook is the whole reason `MeshWeaver.Hosting.Monolith.TestBase` can define the seam without ever referencing Orleans — 23 in-process suites would otherwise drag a cluster's worth of packages to boot a mesh that has no silos.

## 🚨 One thing in the plan does not reproduce

`DecentralisedTests.md` step 3 says "convert the 23 Orleans subclasses to `Bootstrap => MeshBootstrap.Orleans(…)`", which reads as *make them `MonolithMeshTestBase` subclasses*. **They cannot be**, for two independent reasons:

1. A cluster is deployed **asynchronously**, in `InitializeAsync`. `ConfigureMeshBase` — where `Bootstrap.Bootstrap(builder)` is called — runs in the **constructor**, before any cluster exists.
2. **18 of these suites assert on `Cluster` or `SiloServices`** — surfaces a base that must not reference Orleans cannot expose.

So the seam unifies the **declaration**, not the boot. The doc is updated to say this rather than leaving the next reader to rediscover it.

## Numbers, re-measured by running

The survey's figures were counted from `[Fact]` declarations. Running the assembly says otherwise, and the doc is corrected:

| | survey | measured |
|---|---:|---:|
| suites on the two bases | 24 | **23** |
| cases they carry | 47 | **50** |
| cases in the assembly | 192 | **207** |

`OrleansSharedTestBase` reaches 27 cases, not 24; the `OrleansTestBase<T>` row's twelfth "class" is the non-generic `OrleansTestBase` declaration itself — a base, not a suite.

**Verdicts:** `MeshWeaver.Hosting.Orleans.Test` **207 passed / 0 failed** before → **215 passed / 0 failed** after (207 + 8 new seam cases), 30 s → 26 s. `MeshWeaver.Documentation.Test` 150/150 (the disposal-join ratchet and the doc guards).

## The pool now leases on the cluster SHAPE

`OrleansMeshPool` keyed on the fixture **type**. That was sound only while *"a different cluster"* and *"a different fixture subclass"* were the same statement. Once a suite can **describe** its cluster they are not, and a lease across two shapes hands a suite a cluster built by someone else's silo configurator — which surfaces as a missing registration two files from the cause. The key is now the whole `OrleansClusterShape` (silo configurator, client configurator, silo count, shared store, fixture type), with a test pinning that any difference makes two clusters non-interchangeable.

## Why the two names still exist

`OrleansTestBaseCompat.cs` carries **no implementation** — three declarations forwarding to the new base. It exists because **MeshWeaver.Plugins derives from them in 11 places** and its CI checks this repository out at `main`, **unpinned** (`.github/workflows/ci.yml`), so deleting them here reds that trunk the moment this merges, before its own conversion can land. The order with no red window:

1. **this PR** — `OrleansMeshTestBase` lands; core stops using the old names.
2. **MeshWeaver.Plugins PR** — its 10 `OrleansTestBase<T>` suites and `AiOrleansSharedTestBase` move over.
3. **a two-line PR here** — the bridge file goes.

A guard test (`NoSuiteInThisRepo_DerivesFromTheRetiredBases`) fails if any suite in this repo derives from them again, so the bridge cannot quietly become permanent. `InitialSilosCount` is kept on the bridge alone because Plugins' `OrleansPostgresLifecycleTest` overrides it.

Grep, this repo, excluding the bridge and the guard that names them: **zero derivations.**

---

*No What's New entry: pure-internal test infrastructure, no user-visible effect.*


rbuergi:
## ⛔ Auto-merge is deliberately DISARMED until Systemorph/MeshWeaver.Plugins#1091 merges

Building MeshWeaver.Plugins@main against this branch — the check that actually matters, since its CI checks this repo out at `main`, **unpinned** — found that the compat bridge is necessary but **not sufficient**. Two Plugins suites declare private helpers whose names the new base grows:

| Plugins file | member | against this branch |
|---|---|---|
| `OrleansHarnessPartitionPublicReadTest` | `SiloServices` | **CS0108** |
| `OrleansMeshChangeFeedTest` | `ClientMesh` | **CS0108** |

CS0108 is a warning that repo compiles as an **error**, so merging this first reds that trunk — and that is the run which publishes its module bundles.

**Systemorph/MeshWeaver.Plugins#1091** removes both helpers (they re-derive `Cluster.SiloServices()` and `Fixture.ClientMesh` verbatim). It is deliberately compatible with **both** cores — measured `Build succeeded.` against `main` (b54fb5f16) *and* against this branch — so it can land first and leave no red window in either repo.

**Order:** Plugins#1091 → this PR. Re-arm with `gh pr merge 2963 --auto` once #1091 is merged.

### Everything else in MeshWeaver.Plugins@main builds green against this branch

`MeshWeaver.AI.Orleans.TestBase`, `MeshWeaver.AI.Orleans.Test` (with #1091), `MeshWeaver.Hosting.PostgreSql.Test`, `MeshWeaver.Hosting.Grpc.Test`, `MeshWeaver.Markdown.Collaboration.Orleans.Test`, `MeshWeaver.Markdown.Export.Test` — all six, `Build succeeded.`

PR #2996, opened 2026-09-01
The ExecuteScript pre-flight keeps #841's Warning trace

rbuergi:
## What happened

#2918's pre-flight (`DescribeUnrunnableTarget` → `PreflightError`) answers a guaranteed-unrunnable target **before** dispatching. That is the right change and it stays. But it answers before the two sinks that carried the *other half* of #841.

> #841's guarantee is not "the caller is told" — it is **"the caller is told AND the pod leaves a Warning+ trace"**, because a dispatch that creates no Activity node produces no evidence anywhere afterwards.

Both branches the pre-flight now front-runs log it:

* `CodeNodeType.RefuseDispatch` → `ExecuteScript refused on {Hub}: {Error} No Activity node was created.`
* `MeshOperations.DispatchScript` → `ExecuteScript dispatch refused for {Path}: {Error}`

`PreflightError` logged nothing. So making the guaranteed failure **fast** silently made it **silent**, for exactly the three conditions the pre-flight recognises (`NodeNotFound`, and both `NotExecutable` shapes) — the pre-#841 condition, restored by a fix for something else. The caller-visible verdict was correct throughout, which is why nothing else noticed.

## How it surfaced

From the satellite side. MeshWeaver.Plugins' `ExecuteScriptDispatchDiagnosticsTest.Dispatch_NonExecutableNode_SurfacesRefusalToCaller` asserts the **trace**, not the verdict:

```
Expected value not to be <null> because a dispatch failure must leave a Warning+ trace containing
'ExecuteScript refused' — the whole point of #841 is that the pod emitted NOTHING at Warning or above.
Captured: [Warning HostedHubsCollection: … | Warning HostedHubsCollection: … | Error KernelContainer: …]
```

while the verdict came back perfectly:

```json
{"status":"Error","errorType":"NotExecutable",
 "message":"Not executable: '…' has CodeConfiguration.IsExecutable = false. Set it to true…"}
```

That message text exists only in `MeshOperations.DescribeUnrunnableTarget` — it is the pre-flight answering, not the hub.

**It is also load-bearing right now.** That suite runs inside Plugins' `Module bundle (MeshWeaver.AI)` job, which packs and inspects the bundle and then **refuses to hand it to the registry while the suite is red**. So this failure is one of the two things keeping every satellite on yesterday's module bytes (the other is a stale content pin, fixed in MeshWeaver.Plugins#1124) — including MeshWeaver.Reinsurance's trunk, which is waiting on the module-closure fix in #2977.

## The fix

`PreflightError` writes the same warning the sinks it replaced do, naming the path and the `errorType`. One place, so all three refusal shapes carry it.

## The guard

`APreflightRefusalLeavesAWarningTrace…` in `ExecuteScriptPreflightTest`, beside the two elapsed-time tests, using the same instance-owned `ILoggerProvider` capture pattern the Plugins suite uses.

**Verified it can fail**: with the `LogWarning` removed the test is `Failed: 1, Passed: 0`; with it, `Passed: 3` for the file. A gate that cannot fail is not a gate.

## Verification

- `dotnet build -c Release -warnaserror` clean on `src/MeshWeaver.Mesh.Operations` and `test/Memex.Portal.Shared.Test` (`0 Error(s)`, `0 Warning(s)`).
- `ExecuteScriptPreflightTest` — 3 passed, 0 failed; falsification run as above.
- `DocumentationLinkIntegrityTest` green (What's New entry added).
- Core tip at branch point: `861f1a015`.

### DIFF
--- a/src/MeshWeaver.Documentation/Data/WhatsNew/2026-09-01-a-refused-script-still-leaves-a-trace.md
+++ b/src/MeshWeaver.Documentation/Data/WhatsNew/2026-09-01-a-refused-script-still-leaves-a-trace.md
@@ -1,0 +1,23 @@
+---
+Name: A script refused before it starts still leaves a trace
+Category: Fix
+Description: The fast refusal for a script that cannot run now writes the same warning the slow path always did, so an operator can still find out why a run never happened.
+Icon: Sparkle
+Order: -20260901
+---
+
+# A script refused before it starts still leaves a trace
+
+Asking to run a script on something that cannot run one — a page that is not a script, a path with
+nothing behind it, a script switched off — now answers immediately instead of waiting out the whole
+dispatch budget. That is a good change, and it stays.
+
+It had one cost nobody could see. The slow path used to leave a **warning in the log** on its way
+past, and that warning is the only evidence anywhere that a run was asked for and never happened:
+no run means no activity to look at afterwards. Answering earlier skipped the place the warning was
+written, so the caller got a clear message and the server went quiet — which is precisely the
+situation the warning was added for in the first place.
+
+The fast refusal now writes the same warning, naming the path and what was wrong with it. Nothing
+changes for the person who made the call; what changes is that whoever looks at the logs afterwards
+can still see it happened.
--- a/src/MeshWeaver.Mesh.Operations/MeshOperations.cs
+++ b/src/MeshWeaver.Mesh.Operations/MeshOperations.cs
@@ -4609,10 +4609,26 @@ public class MeshOperations
     /// to read regardless of which side answered. <c>errorType</c> is a stable token here rather
     /// than an exception type name — there is no exception, and inventing one would be a lie.
     /// </summary>
-    private string PreflightError(string resolvedPath, string errorType, string message) =>
-        JsonSerializer.Serialize(
+    private string PreflightError(string resolvedPath, string errorType, string message)
+    {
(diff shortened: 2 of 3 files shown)


rbuergi:
**Closing as a duplicate — #2991 landed the identical fix at 16:22:28Z (`667c42cb`), 40 minutes before I opened this.** Same one-line `LogWarning` in `PreflightError`, same guard in `ExecuteScriptPreflightTest` (theirs is `ExecuteScript_PreflightRefusal_IsVisibleToTheOperator_NotJustToTheCaller`, parameterised over the refusal shapes — stronger than my single case), same What's New entry. Nothing here is worth carrying forward.

My miss, and worth writing down: I traced the failure from the SATELLITE side (MeshWeaver.Plugins' `ExecuteScriptDispatchDiagnosticsTest` red in the module lane), reached the pre-flight through a **local core checkout that was hours stale**, and did not check for a sibling PR on the core issue before writing the fix. `gh pr list --search "ExecuteScript"` would have cost seconds. Two sessions fixing one thing in parallel is the recurring waste here.

Recording the cross-repo link, which does still matter: that Plugins test is inside the `Module bundle (MeshWeaver.AI)` job, which packs and inspects the bundle and then **refuses to hand it to the registry while its suite is red**. So until a promoted wave carries `667c42cb`, that job stays red and no satellite picks up new module bytes — including the closure fix in #2977 that MeshWeaver.Reinsurance's trunk is waiting on. Tracked in MeshWeaver.Plugins#1121.

Issue #2813, opened 2026-08-30
GitSync adopts a stale prebuilt assembly over the source it just synced — silently running last week's code, and it destroyed client data


rbuergi:
## What happened

A `git_hub_sync update` pulled new source for `Crm/Migration`, then **adopted a prebuilt assembly instead of compiling what it had just pulled**. The node reported success. The next run executed the *old* code and destroyed live client data.

Sync activity `Crm/_Activity/0c389edf`, 2026-08-30T17:06:04Z:

```
Fetching the branch HEAD from GitHub and importing the deltas…
Recompiling 2 NodeType(s): Crm/Migration, Crm/Opportunity
Adopted 2 prebuilt assembly(ies) — the rest compile.
Imported Imported (6 node(s)).
```

Status `Succeeded`. Nothing in that log tells an operator that the assembly now serving `Crm/Migration` was **not** built from the source in the line above it.

## Why it was wrong, provably

The synced commit (`904854f`) contained three fixes merged ~40 minutes earlier. A migration run at 17:06:37 exhibited **all three pre-fix behaviours**:

| fixed in the synced source | what the adopted assembly did |
|---|---|
| read the document body shape-tolerantly | stripped the body from 4 promoted documents |
| a bullet whose bold text is a link is not a person | promoted 2 documents to `Crm/Contact` |
| never promote/suggest a node already typed `Crm/*` | suggested the client's own `Crm/Opportunity` |

I then forced a compile via the `compile` verb. Afterwards `compiledSources` **matched** `currentSourceVersions` and the MVID changed (`ed21acce…` → `07d5ca73…`), confirming the adopted assembly had been built from different source than the node was holding.

## The damage

Four client documents lost their entire text — one a 7,000-word commercial offer. Recovered from version history.

**One did not recover.** `PartnerRe/EslLifeCycleProcess` has exactly one version in history — the one the stale code wrote. `restore_from_point_in_time` returns *"No version found at or before…"*, and content indexing is off on that deployment, so there is no chunk copy either. Its markdown body is permanently gone.

## Why this is worse than a normal staleness bug

The operator did everything right: merged a green PR, ran `check`, ran `update`, saw `Succeeded`, saw the NodeType at `compilationStatus: Ok` with `compiledSources == currentSourceVersions`. **Every signal available said the fix was live.** The adopted assembly is what made all of those true while being false in the only way that mattered.

## What I think the fix is

[...]

PR #3000, opened 2026-09-01
fix(compile): the adopted-source fingerprint now covers @@ includes

### LINKED ISSUE #2948: The adopted-source fingerprint does not cover `@@` includes — a change to an included-only snippet is invisible to staleness detection
Split out of #2813 rather than folded into it, because it is a *pre-existing* gap with a real cost to close, and burying it in the fix would have made the fix's coverage look larger than it is.

## What is not covered

`NodeTypeSourceFingerprint.Compute` (added in #2947) hashes the compile input as
`NodeCompileShaping.CollectCompileSources` yields it: the deduplicated, non-executable, non-blank
Code nodes a source/test query matched, reduced to `(node path, SHA-256 of the code text)`.

An `@@Path/To/Snippet` include pulls a Code node that **no source query matches**. It reaches the
emitted assembly — `NodeSetCompiler.ResolveInputs` resolves includes *after* the fold — but it is
absent from the fold, so:

- editing an included-only snippet does not move `AdoptedSourceFingerprint` / `CurrentSourceFingerprint`, and
- a prebuilt assembly built before that edit still adopts as `AdoptedVerified`.

`BakeEquivalenceTest` already pins this shape explicitly: `Widget/Snippets/Greeting` is asserted
absent from `sourceVersions` while `Greeting` is asserted present in the emitted surface.

## Why it is not a regression

It is the **same set** `NodeTypeDefinition.CompiledSources` records, so `IsDirty` has always had
this hole too — an included-only edit never dirtied the type and never triggered a recompile. #2947
does not widen the gap; it just makes it worth naming, because the fingerprint now carries a
*verified* claim where before there was only an unverified one.

## What closing it would take

Both sides must resolve includes before hashing:

- the bake already does (`NodeSetCompiler.ResolveInputs` → `NodeCompileShaping.ResolveCodeIncludes`
  against the tree's `NodeSet`, a dictionary lookup);
- the live side does **not**. `InstallSourcesWatcher` holds only the query-matched nodes, and
  resolving includes there means a mesh read per include, on a path that runs at every hub
  activation and every source edit. That is a real cost and a real reactive-composition change
  (the watcher's `Update` lambda is pure and synchronous today), which is why it is not smuggled
  into a data-loss fix.

An alternative worth weighing first: fold the include-resolved text in only where it is already
computed — i.e. have the compile stamp the resolved set — so the live half never has to do the
mesh reads.

## Acceptance

A test in the shape of `AdoptedBuildSourceStampTest`: a type whose only change is to an
included-only snippet, whose bundle was baked before that change, must be **refused**, not adopted
as verified.

Related: #2813 (the fingerprint), #2947 (the PR that made it reachable), #2820 (the execute-time
interlock that consumes `BuildProvenance`).

rbuergi:
Closes #2948

## The gap

`NodeTypeSourceFingerprint.Compute` hashed the compile input as
`NodeCompileShaping.CollectCompileSources` yields it. An `@@` include pulls a Code node that **no
source query matches**: it reaches the emitted assembly (`NodeSetCompiler.ResolveInputs` substitutes
it *after* the fold) and was absent from the hash. So editing an included-only snippet moved neither
`AdoptedSourceFingerprint` nor `CurrentSourceFingerprint`, and a prebuilt assembly baked **before**
that edit still adopted as `AdoptedVerified`.

That is the worst of the three rows to be wrong on. `AdoptedUnverified` says "nobody established
where these bytes came from" and an operator reads it as the warning it is. `AdoptedVerified` is an
**assertion** that the shipped bytes match the source this mesh holds — standing over source that
was never hashed, it is a *false verification*: the #2813 incident one layer in.

## The fix — both sides resolve includes before hashing

- **Producer.** `NodeCompileShaping.ResolveCodeIncludes` takes an optional collector and records
  `resolved path → code text` at the point it consumes each include, so the closure is a by-product
  of the **one walk that already substitutes them** — there is no second traversal to disagree with
  the first. `CompileInputs.ResolvedIncludes` carries it; `TreeBake` / `CascadeBuild` fold it
  straight in. **No second walk, no second read, and no new blocking bridge** at a synchronous build
  step (the `ProductionBlockingBridgeSites` ratchet does not move).
- **Consumer.** The sources watcher resolves it through the mesh. Reads cannot happen inside a pure
  `Update` lambda, so the fingerprint moved into an observable step ahead of it, composed with
  `Switch()` so a newer source set supersedes an in-flight resolution rather than racing it.
  `CollectIncludeClosure` scans each source's text for `@@` and **issues no read at all** when there
  is none — almost every type — so only a type that actually has includes pays, and it pays the same
  reads its own compile would.

### Deterministic and cycle-safe

An order-dependent fingerprint manufactures phantom staleness (endless recompiles, or a false
refusal), which is worse than the gap being closed. Three properties:

1. **Order cannot reach the result.** The closure is keyed by *resolved path* and returned sorted
   (ordinal); `PartitionSourceFingerprint` sorts again. Roots are walked in `CollectCompileSources`
   order — itself ordinal by node path — and **serially** (one `SelectMany` chain, never a `Merge`),
   so there is no read interleaving to observe either.
2. **Suppression is result-preserving.** The per-root visited set skips a path that root already
   walked; skipping a re-read of the *same* anchored path can only re-derive an entry that is
   already present, so which parent reached a shared snippet first cannot change the closure.
3. **A cycle terminates.** `A → B → A` adds each path once; the second visit takes the already-added
   branch, reads nothing, recurses no further. A self-include is the same case with one hop.

### An unreadable include is INCONCLUSIVE, never absent

The producer's lookup is in-memory and never stalls. A consumer that quietly treated a stalled read
as "the include is gone" would hash a *shorter* closure — indistinguishable from a stale bundle — so
a perfectly good adoption is **refused**, and on a `Modules:RequirePrebuilt` mesh that is terminal.
`SourceFingerprintIncludeReader` reads via `GetMeshNodeOutcome` and keeps the three states apart:
`Present` contributes, `Absent` contributes nothing (it contributes nothing to the bytes either —
the directive stays verbatim, so both sides agree), and `Unavailable` / `DeleteInProgress` / a
timeout raise `SourceIncludeUnavailableException`. The watcher catches exactly that and **leaves the
previously published fingerprint standing**, so the judgement takes the "nothing has been compared"
branch. Same #890 rule: a probe must not answer its scariest branch on its own inability to run.

**`AdoptedUnverified` is still never a refusal.** Pinned by a test in this PR.

## What was NOT changed, deliberately

The include closure is **not** added to `CompiledSources` / `CurrentSourceVersions`. That is the
recompile-*trigger* half — a change feed, not a hash — and `sourceVersions` is a producer/consumer
contract mirroring the RAW query match on both bakes. `#2948` closes the **verified claim**, which
is decided by the two fingerprints and nothing else. Stated in the doc rather than assumed.

## `BakeEquivalenceTest` — reconciled, not loosened

It pinned `Widget/Snippets/Greeting` **absent** from `sourceVersions`. That assertion **stays** and
is now explained: `sourceVersions` mirrors `CompiledSources` (the raw query match), which is a
different field answering a different question. A **new** assertion pins the include **present** in
the bake's resolved include set — because without it, assertion 3b's cross-producer hash equality
would be satisfiable by *both* producers ignoring the include, which is exactly the state #2948
found them in. `MeshDrivenAndCompilerDrivenBakes_ProduceTheSameArtifacts` passing is now the proof
that the mesh bake (real workspace, mesh reads) and the tree bake (`NodeSet`) resolve the same
closure.

## Binary compatibility

`CompileInputs.ResolvedIncludes` is a `required init` member, **not** a positional parameter:
positional would have changed the primary constructor's arity, which
`scripts/check-record-signatures.py` correctly rejects as a binary break between a module and the
platform it loads into. `required` because forgetting it is silent — an empty closure for a type
with includes is a refusal, not "the fingerprint minus the includes". For the same reason the
include-less `Compute` overload now **refuses** a set that has an include rather than under-covering
it: a convenience overload that can under-cover is exactly how #2813 sat inert (the 7-arg `Seed`
hard-coded `sourceFingerprint: null` and both production callers took it).

Types with no includes fold to the **identical** value as before, so there is no fleet-wide
fingerprint churn; and this file is inside `FrameworkBuildIdentity.FullMvidAssemblies`, so a producer
and a consumer that can adopt across each other necessarily run the same implementation.

## Verification

Every new test was proven falsifiable — inverted, seen red, restored (output in the PR thread).

- `MeshWeaver.Compiler.Pipeline.Test` — **583 passed** (whole project, no `--filter`)
- `MeshWeaver.PluginTester.Test` — **312 passed** (whole project; includes `BakeEquivalenceTest`)
- `MeshWeaver.Documentation.Test` — **166 passed** (whole project), plus a broken-link probe proving
  `DocumentationLinkIntegrityTest` names *both* edited pages
- `dotnet build -c Release -warnaserror` — 61 projects, `0 Warning(s) 0 Error(s)`
- `scripts/check-record-signatures.py` — clean

Docs: `Doc/Architecture/NodeTypeCompilation` gains the mechanism (determinism argument, cycle brake,
the inconclusive rule) and its "deliberately does NOT do" list is rewritten to state what is still
open. What's New entry, `Category: Fix`.

### DIFF
--- a/src/MeshWeaver.Compiler.Pipeline/NodeTypeCompilationHelpers.cs
+++ b/src/MeshWeaver.Compiler.Pipeline/NodeTypeCompilationHelpers.cs
@@ -1306,6 +1306,9 @@ internal static class NodeTypeCompilationHelpers
         // changed" path). Same mesh-scoped singleton the compile watcher's parked short-circuit
         // uses; null in the rare host that does not register it (no auto-retry, unchanged before).
         var parkRegistry = hub.ServiceProvider.GetService<NodeTypeCompileParkRegistry>();
+        // #2948 — the mesh read the source fingerprint's `@@`-include walk needs. Built once per
+        // watcher; it issues nothing at all for a source set with no `@@` in it.
+        var includeReader = SourceFingerprintIncludeReader.For(hub, logger);
 
         // Outer subscription: discover the source path set via the shared
         // synced query (NodeSources.GetSources). When the path set changes
@@ -1400,7 +1403,36 @@ internal static class NodeTypeCompilationHelpers
                 // adopted fingerprint to compare it against. The nodes are already materialised
                 // here; the hash is what costs, and it is paid only where it is used.
                 return (Snapshot: (IReadOnlyDictionary<string, long>)snap, Nodes: sources);
-            }),
+            })
+            // 🚨 #2948 — the CONTENT fingerprint now covers the `@@`-include closure, and resolving
+            // that needs mesh READS, so it happens HERE (an observable step) rather than inside the
+            // pure Update lambda below. Cost: `CollectIncludeClosure` scans each source's text for
+            // "@@" and issues NOTHING when there is none — which is nearly every type — so the
+            // reads are paid only by the types that actually have includes.
+            //
+            // 🚨 Switch(), not SelectMany: a newer source set must SUPERSEDE an in-flight include
+            // resolution, or a slow read could publish a fingerprint for a set that is already gone.
+            //
+            // 🚨 An include the mesh could not READ (a stall, an unavailable owner) is INCONCLUSIVE,
+            // never "absent". Degrading it to absence shortens the hashed set, which is
+            // indistinguishable from a stale bundle and refuses a perfectly good adoption — on a
+            // Modules:RequirePrebuilt mesh that is terminal. So the fingerprint is dropped for this
+            // emission (null) and the previously published value stands; the judgement then takes
+            // ApplyAdoptedSourceStamp's "nothing has been compared" branch (AdoptedUnverified),
+            // which is the honest answer and the same #890 rule the emit canary follows.
+            .Select(published => NodeTypeSourceFingerprint
+                .Compute(published.Nodes, hubPath, includeReader, logger)
+                .Select(fingerprint => (published.Snapshot, Fingerprint: (string?)fingerprint))
+                .Catch((SourceIncludeUnavailableException ex) =>
+                {
+                    logger?.LogWarning(ex,
+                        "SourcesWatcher: the @@-include closure for {HubPath} could not be "
+                        + "established, so CurrentSourceFingerprint is left at its previous value "
+                        + "for this emission — an unreadable include must never read as an absent "
+                        + "one", hubPath);
+                    return Observable.Return((published.Snapshot, Fingerprint: (string?)null));
+                }))
+            .Switch(),
                 published =>
                 {
                     var snapshot = published.Snapshot;
@@ -1479,8 +1511,13 @@ internal static class NodeTypeCompilationHelpers
                         // in memory, on a path that runs at hub activation and per source edit —
                         // i.e. exactly where a Roslyn compile is about to cost four orders of
                         // magnitude more.
-                        var fingerprint = NodeTypeSourceFingerprint.Compute(
-                            published.Nodes, hubPath, logger);
+                        //
+                        // 🚨 #2948 — it is now resolved UPSTREAM (the include closure needs mesh
+                        // reads, which cannot happen inside this pure lambda), and it can be NULL:
+                        // null means the closure could not be established, NOT that there is
+                        // nothing to hash. The previous value stands in that case — inventing one
+                        // from a short closure is a false refusal.
+                        var fingerprint = published.Fingerprint;
 
                         // Idempotent: no-op when CurrentSourceVersions already
                         // matches the just-computed snapshot. IsDirty is a
@@ -1492,18 +1529,22 @@ internal static class NodeTypeCompilationHelpers
                         // before this field existed — snapshot already current, fingerprint null —
                         // would no-op forever and never acquire one, so the type could never be
                         // judged again. The condition is what makes the field SELF-HEALING on the
-                        // first activation after an upgrade.
+                        // first activation after an upgrade. An INCONCLUSIVE emission (fingerprint
+                        // null) drops out of the comparison entirely: it has nothing to say about
+                        // the field, so it must neither force a write nor block the snapshot's.
                         if (!pendingStamp
                             && def.CurrentSourceVersions is not null
                             && DictEquals(def.CurrentSourceVersions, snapshot)
-                            && string.Equals(
-                                def.CurrentSourceFingerprint, fingerprint, StringComparison.Ordinal))
+                            && (fingerprint is null
+                                || string.Equals(
+                                    def.CurrentSourceFingerprint, fingerprint,
+                                    StringComparison.Ordinal)))
                             return curr;
 
                         var updated = def with
                         {
                             CurrentSourceVersions = snapshot,
-                            CurrentSourceFingerprint = fingerprint,
+                            CurrentSourceFingerprint = fingerprint ?? def.CurrentSourceFingerprint,
                         };
 
                         if (pendingStamp)
--- a/src/MeshWeaver.Compiler.Pipeline/SourceFingerprintIncludeReader.cs
+++ b/src/MeshWeaver.Compiler.Pipeline/SourceFingerprintIncludeReader.cs
@@ -1,0 +1,153 @@
+using System.Reactive.Linq;
+using MeshWeaver.Compiler;
+using MeshWeaver.Mesh;
+using MeshWeaver.Mesh.Services;
+using MeshWeaver.Messaging;
+using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.Logging;
+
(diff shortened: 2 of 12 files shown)


rbuergi:
## Falsification output — every new test proven able to fail

A test that passes after a change proves nothing unless it would have failed before it. Four
mutations, each applied to the working tree, built, run, then reverted. All 14 tests in
`SourceFingerprintIncludeClosureTest` pass unmutated.

### Probe 1 — `CollectIncludeClosure` always returns the empty closure (i.e. #2948 never fixed)

```
Failed ...EditingAnIncludedOnlySnippet_MovesTheFingerprint
  Did not expect value to be "b59c18eb06dc5846" because the snippet is INSIDE the emitted bytes,
  so a change to it must move the hash — otherwise an assembly compiled before the edit still
  claims AdoptedVerified against source it was never built from.

Failed ...ABundleBakedBeforeAnIncludedOnlyEdit_IsRefused_NotVerified
  Expected value to be AdoptionRefused because the only difference between the two builds is an
  included-only snippet, and that is the whole of #2948 — before the fix these fingerprints were
  EQUAL and the verdict was AdoptedVerified, but found AdoptedVerified.

Failed ...ACyclicIncludeTerminates_AndIsStable
  Did not expect value to be "7d43d876b1693e92" because B is in the closure — the cycle brake
  stops the walk, not the coverage.

Failed ...ANestedIncludeIsCovered_Transitively
Failed ...AnUnreadableInclude_IsInconclusive_NeverAbsent
Failed ...ThePreResolvedOverload_AgreesWithTheReaderOverload

Failed!  - Failed: 6, Passed: 8, Total: 14
```

That second one is the issue's acceptance criterion, reproduced exactly: on `main` the verdict for
an included-only edit is `AdoptedVerified`.

### Probe 2 — an UNSTABLE token (`SHA256(code + Guid.NewGuid())`): the hash moves when nothing changed

```
Failed ...TheFingerprintIsIndependentOfDeliveryOrder
Failed ...ACyclicIncludeTerminates_AndIsStable
Failed ...AnEditToANodeTheCompileNeverSees_LeavesTheFingerprintAlone
Failed ...AnIncludeThatResolvesToNothing_ContributesNothing
Failed ...ASetWithoutIncludes_FoldsIdenticallyThroughEveryOverload
Failed ...ThePreResolvedOverload_AgreesWithTheReaderOverload

Failed!  - Failed: 6, Passed: 8, Total: 14
```

This is the phantom-staleness control arm: the no-move / order-stability / cross-overload-agreement
family all bite.

### Probe 3 — the include-less `Compute` overload stops refusing an include-bearing set

```
Failed ...TheIncludeLessOverload_RefusesASetThatHasAnInclude
  Expected a InvalidOperationException, but no exception was thrown.

Failed!  - Failed: 1, Passed: 13, Total: 14
```

### Probe 4 — `ApplyAdoptedSourceStamp` refuses EVERY row

```
Failed ...ALegacyBundleWithNoFingerprint_StillAdopts
Failed ...ABundleWhoseIncludeStillMatches_AdoptsAsVerified

Failed!  - Failed: 2, Passed: 12, Total: 14
```

Note what stays **green** under probe 4: the refusal test itself. That is the point of carrying the
two positive controls — "widen the refusal" would satisfy the headline test while breaking the
anti-outage property that `AdoptedUnverified` is never refused.

### Doc probe — `DocumentationLinkIntegrityTest` actually covered the edited pages

A green trx that names no pages proves nothing, so a deliberately broken link was inserted into each
page this PR touches:

```
Failed MeshWeaver.Documentation.Test.DocumentationLinkIntegrityTest.AllInternalDocLinks_ResolveToExistingNodes
  Doc/WhatsNew/2026-09-02-a-shared-snippet-no-longer-hides-behind-a-verified-build:
      (/Doc/Architecture/AlsoDoesNotExist2948) — resolves to /Doc/Architecture/AlsoDoesNotExist2948,
      which does not exist
  Doc/Architecture/NodeTypeCompilation:
      (/Doc/Architecture/ThisPageDoesNotExist2948) — resolves to
      /Doc/Architecture/ThisPageDoesNotExist2948, which does not exist, but found 2 item(s).
```

Both pages are named, so the suite is genuinely reading them. Restored, and the whole project is
green (166 passed).

### Unmutated

```
MeshWeaver.Compiler.Pipeline.Test   Passed! - Failed: 0, Passed: 583, Total: 583
MeshWeaver.PluginTester.Test        Passed! - Failed: 0, Passed: 312, Total: 312
MeshWeaver.Documentation.Test       Passed! - Failed: 0, Passed: 166, Total: 166
dotnet build -c Release -warnaserror   61 projects, 0 Warning(s), 0 Error(s)
scripts/check-record-signatures.py     ✓ 10 changed C# file(s): no binary-breaking changes
```

Issue #2468, opened 2026-08-27
The Doc content gate CRASHES on a `$model-probe` hub-activation timeout — no verdict, on unrelated PRs


rbuergi:
Two PRs whose diffs cannot reach the Doc content at all — #2467 (one workflow file) and #2461 (CI) — both went red on **Doc content compiles and runs**, with the same shape:

```
warn: MeshWeaver.Mesh.GetMeshNode[0] GetMeshNode('$model-probe/a75e76b9d2fa4ebbaca5e979f648ee2d') timed out
      after 10.0s (budget 10s) — the owning per-node hub never answered the GetDataRequest.
Unhandled exception. System.AggregateException: One or more errors occurred.
 ---> System.TimeoutException: GetMeshNode('$model-probe/…') timed out after 10.0s …
      at System.Reactive.PlatformServices.ExceptionServicesImpl.Rethrow(Exception)
      …
      at MeshWeaver.Mesh.MeshNodeStreamExtensions.<…>g__EmitError|2(Exception)
      at System.Threading.CancellationTokenSource.ExecuteCallbackHandlers(Boolean)
##[error]The content gate failed before it produced a verdict — no check was judged.
```

- #2467 → [job 98407873831](https://github.com/Systemorph/MeshWeaver/actions/runs/33038570991/job/98407873831), probe `a75e76b9…`
- #2461 → [job 98355928593](https://github.com/Systemorph/MeshWeaver/actions/runs/33021741071/job/98355928593), probe `47eef047…`

Same 10 s budget, different probe guid, unrelated diffs. #2456 / #2455 / #2451 passed the same check in the same window, so it is intermittent rather than a trunk break.

## Two defects, and the second is the worse one

**1. A `$model-probe/<guid>` per-node hub does not answer within 10 s.** The message already says this is *not* "node not found" — the hub never answered its `GetDataRequest`. A transient probe address that never activates is a real bug or a budget that is too tight under CI load; either way it is diagnosable from the address family.

**2. The timeout escapes as an UNHANDLED EXCEPTION and takes the process down.** The gate then reports *"failed before it produced a verdict — no check was judged"* — so a run in which the gate learned **nothing** is presented in the same red as a run in which it judged content and found it wrong. Those are different facts and want different responses: the first is "re-run", the second is "fix your content". Worse, the failure is in the family this repo keeps re-learning — a check whose outcome does not mean what its colour says.

[...]

Issue #991, opened 2026-08-09
A pending Rx timer in ActivityControlPlaneExtensions pins the AccessAssignment hub past Mesh.Dispose()


rbuergi:
## Summary

`MeshHubDisposalLeakTest.MeshHub_IsCollected_AfterMeshAndServiceProviderDisposal` intermittently fails with a **real** GC-root detection: a single-shot `Observable.Timer` closure keeps a `TimerQueueTimer` alive on the process-wide `TimerQueue`, and that chain pins a disposed mesh's `MessageHub`.

This is filed to preserve the root chain — CI logs expire, and the guard already did the hard part of naming the culprit.

## The chain (ClrMD, run `31300373896`, shard 1, Linux)

```
ROOT[StrongHandle] System.Object[]
  → System.Threading.TimerQueue[] → TimerQueue → TimerQueueTimer
  → System.Reactive.Concurrency.ConcurrencyAbstractionLayerImpl+Timer
  → UserWorkItem<ValueTuple<ObservableImpl.Timer+Single+_, …>>
  → System.Reactive.Linq.ObservableImpl.Timer+Single+_
  → System.Reactive.AnonymousSafeObserver<Int64> → System.Action<Int64>
  → MeshWeaver.Mesh.ActivityControlPlaneExtensions+<>c__DisplayClass2_1<MeshNode>
  → System.Action
  → MeshWeaver.Mesh.ActivityControlPlaneExtensions+<>c__DisplayClass2_0<MeshNode>
  → System.Func<IObservable<MeshNode>>
  → MeshWeaver.Graph.Configuration.NodeTypeCompilationHelpers+<>c__DisplayClass5_0
  → MeshWeaver.Messaging.MessageHub  [RunLevel=6] [Address=AccessAssignment]
```

`visited=253524 hubFound=True`, i.e. the analysis ran for real (not the `Unavailable` skip path from #674).

## Reading it

An `Observable.Timer` subscription created in `ActivityControlPlaneExtensions` — reached through `NodeTypeCompilationHelpers` — is **not disposed** when the mesh is disposed. Rx registers the timer on the shared `TimerQueue`, which is a strong-handle root, so for as long as the timer is pending it transitively holds the two display-class closures, the `Func<IObservable<MeshNode>>`, and finally the `AccessAssignment` `MessageHub`.

[...]

PR #3004, opened 2026-09-01
fix(mesh): a registration probe no longer reports itself as a transient ERROR

### LINKED ISSUE #2990: ActivityControlPlane watcher reports the boot-time content-type probe as a transient ERROR and arms a re-establish
## Symptom

Every mesh start writes one **Error**-level line per static NodeType whose `HubConfiguration`
installs an Activity Control Plane:

```
Error MeshWeaver.Kernel.Hub.KernelContainer: ActivityControlPlane subscription faulted on
content-type-registration/65936dcbcfa348bd921eb49885dea848 — re-establishing
```

The address is the giveaway: `content-type-registration/{guid}` is minted **only** by
`ContentTypeRegistration.ProbeRegister` (`src/MeshWeaver.Graph/ContentTypeRegistrationSweep.cs:70`).
Nothing is actually wrong with the mesh — the line reports the boot-time content-type sweep against
itself.

## Evidence

Reproduced locally on `origin/main` (`24254ad65`), captured off the mesh's own `ILoggerFactory` in
`Memex.Portal.Shared.Test` (an `ILoggerProvider` registered through `MeshBuilder.ConfigureServices`,
so this is what a deployment's log sink receives, not test output). The exception attached to the
record:

```
System.InvalidOperationException: Failed to create stream
   at MeshWeaver.Data.Workspace.ReduceLocalStream[TReduced](WorkspaceReference`1 reference, Func`2 configuration) in src/MeshWeaver.Data/Workspace.cs:line 925
   at MeshWeaver.Data.Workspace.<>c__DisplayClass33_0`1.<GetStream>b__1() in src/MeshWeaver.Data/Workspace.cs:line 874
   at System.Lazy`1.ViaFactory(LazyThreadSafetyMode mode)
   at System.Lazy`1.ExecutionAndPublication(LazyHelper executionAndPublication, Boolean useDefaultConstructor)
   at System.Lazy`1.CreateValue()
   at MeshWeaver.Data.Workspace.GetStream[TReduced](WorkspaceReference`1 reference, Func`2 configuration) in src/MeshWeaver.Data/Workspace.cs:line 881
   at MeshWeaver.Mesh.MeshNodeStreamHandle.AcquireStream() in src/MeshWeaver.Mesh.Contract/MeshNodeStreamExtensions.cs:line 311
   at MeshWeaver.Mesh.MeshNodeStreamHandle.Subscribe(IObserver`1 observer) in src/MeshWeaver.Mesh.Contract/MeshNodeStreamExtensions.cs:line 406
```

It also shows up in CI: `MeshWeaver.Plugins` job `Module bundles (floor) / Module bundle
(MeshWeaver.AI)`, run `33523800964` / job `99911922886`, captured inside an unrelated assertion's
log dump.

## Mechanism

1. `ContentTypeRegistrationSweep.StartAsync` walks every static node definition carrying a
   `HubConfiguration` and calls `ProbeRegister`, which builds a probe hub with
   `AsTransientNodeProbe(startDataSources: false)` and disposes it immediately. Registration is a
   *side effect of building the configuration* — nothing is meant to run.
2. For an Activity-shaped NodeType the configuration includes `KernelContainer`'s
   `WithInitialization(hub => … StartActivityControlPlane(hub))`, which runs on the
   `InitializeHubRequest` turn and calls `hub.WatchControlPlane(...)` →
   `SubscribeWithReEstablish(() => workspace.GetMeshNodeStream() …)`.
3. The probe has **no data sources**, so `ReduceManager.ReduceStream` finds no matching reducer and
   returns `null`. `Workspace.ReduceLocalStream` turns that into a bare
   `InvalidOperationException("Failed to create stream")` — no reference, no owner, no inner cause.
4. `SubscribeWithReEstablish`'s classifier (`ActivityControlPlaneExtensions.cs`) tests for
   `HubDisposingException` naming this address (`IsOwnHubDisposing`), a deserialization
   `MeshNodeStreamException` (`IsPoisonedContent`) and a routing `NotFound`
   (`IsOwnNodeGone`). A bare `InvalidOperationException` matches **none** of them, so it falls
   through to "everything else = genuinely transient": **`LogError` + a 1 s re-establish timer**.

That fourth branch exists for a hub that is alive and briefly unreachable. Here the hub is a
throwaway probe that was never going to have a stream, so "re-establishing" is a promise about a
state that cannot improve.

## Impact

- **Error-level log noise at every portal boot**, once per swept Activity-shaped NodeType.
  `Information` and above ship to Loki; `Error` is what operators alert on. This is the same
  *"routine lifecycle reported as a prod ERROR"* drip the `IsOwnHubDisposing` branch was written to
  remove (see its own remarks) — the guard is present and simply cannot classify this path.
- **A brief re-establish timer armed on a hub that is already being disposed.** This is #991's
  TimerQueue-root shape. It does *not* leak today: `ProbeRegister` disposes the probe, and
  `MessageHub.RegisterForDisposal` disposes a late registrant immediately, which cancels the
  pending schedule. But the ordering is incidental — `Establish()` arms the timer *before*
  `WatchControlPlane` returns the disposable the caller registers — so the safety here is timing,
  not design.

## What the fix is (and is not)

Not a level change and not a `catch`. Two candidates, both root-cause:

1. **`AsTransientNodeProbe(startDataSources: false)` should mean the control plane does not start.**
   A probe built explicitly without data sources should not install a watcher that requires one —
   this is the narrower and more honest fix.
2. **`ReduceLocalStream` must not throw a causeless `InvalidOperationException`.** "Failed to create
   stream" names no reference, no owner and no reason, which is precisely why three deliberate
   classifiers all miss it. Whatever it throws should carry the reference and the owner, and the
   no-data-sources case should be distinguishable from a real failure.

Either way the classifier keeps its meaning: transient means *the node is alive and will come back*.

## Not in scope

Found while restoring #841's Warning trace to `ExecuteScript`'s up-front refusal (the noise appears
in that test's captured-log dump). Deliberately **not** fixed there — a re-establish that recovers
from a state that cannot improve is a design question about the probe, not a line to patch inside an
unrelated change.

rbuergi:
Fixes #2990.

## The symptom

Every mesh start wrote one **Error**-level line per static NodeType whose `HubConfiguration`
installs an Activity Control Plane:

```
Error MeshWeaver.Kernel.Hub.KernelContainer: ActivityControlPlane subscription faulted on
content-type-registration/65936dcbcfa348bd921eb49885dea848 — re-establishing
   << InvalidOperationException: Failed to create stream
```

`content-type-registration/{guid}` is the boot-time content-type sweep's own throwaway probe, so
the line was that sweep reporting itself — at the severity operators alert on, once per swept type,
on every boot.

## Verified root cause — issue candidate **(a)**: the probe was running the node control plane

`AsTransientNodeProbe` states outright that a probe must not run the per-node control plane, and
`MeshDataSource` honours it (`SubscribeToOwnDeletionInit` returns early on a probe). But the ACP is
installed by the **adopter** — `KernelContainer` and every Activity-shaped NodeType, from their own
`WithInitialization` — so a guard placed per-adopter is a guard the next adopter will not have, and
the ACP went on being installed on every probe.

There it subscribes to the hub's own MeshNode stream. The probe is built `startDataSources: false`,
so that reduce has no started data source and `Workspace.ReduceLocalStream` threw a causeless
`InvalidOperationException("Failed to create stream")`. `SubscribeWithReEstablish`'s three terminal
classifiers all miss a bare `InvalidOperationException`, so it fell through to *genuinely
transient*: `LogError` + a 1 s re-establish timer armed against a hub already being disposed (the
#991 `TimerQueue`-root shape, safe today only by the ordering accident that `RegisterForDisposal`
disposes a late registrant immediately).

**Same root, second face.** An Activity NodeType that *also* declares a content type got far enough
for the reduce to construct a `SynchronizationStream`, whose constructor always calls
`GetHostedHub(sync/{id}, Always)` — into the probe's own disposal, re-creating exactly the
`Rejecting hosted hub creation … during disposal` warning that `startDataSources: false` was
introduced to remove.

## The fix — at the seams, never per adopter, and never a quieter log level

| # | Change | Why here |
|---|---|---|
| 1 | `WatchControlPlane` / `WatchSubmission` install **nothing** on a hub carrying the `TransientNodeProbe` marker | the shared seam every ACP adopter goes through |
| 2 | `MeshNodeStreamHandle.Subscribe` answers a probe's **own-node** stream with an **empty stream** | the THIRD own-node read seam, and the last one not saying "there is no node here" — `GetMeshNodeOutcome` answers `Absent` (#2468), `MeshNodeStreamCache.GetStreamRaw` answers empty (#2894). Covers watchers that compose their own source rather than going through (1) — `BuildNodeType`'s claim arbiter is the live example, and was the remaining `sync/` sub-hub after (1) alone |
| 3 | `ProbeRegister` mints its address from `TransientProbeAddresses`; `IsProbeAddress` learns the prefix | the registration probe was a transient probe by every other measure while the one guard keyed off the ADDRESS did not know it existed — a read of one of its paths died on `No node found at 'content-type-registration/…'`, i.e. #2894 re-created by one string literal |
| 4 | `ReduceLocalStream`'s exception names the reference and the owning hub | issue candidate **(b)**: a diagnostic naming no reference, no owner and no cause is *why* three deliberate classifiers missed it |

🚨 **Deliberately not the fix:** downgrading the Error. That would have left the re-establish timer
armed and the sub-hub still being created. "Transient" keeps meaning exactly what it meant — *the
node is alive and will come back*.

## Test that fails before and passes after

`test/MeshWeaver.Graph.Test/ContentTypeProbeControlPlaneTest.cs` starts a **real** monolith mesh
(`MonolithMeshTestBase`, no mocking), registers two Activity-shaped static NodeTypes — one bare like
the platform's own `kernel` type, one declaring a content type — and captures the mesh's **own
`ILoggerFactory`** through an `ILoggerProvider` registered via `MeshBuilder.ConfigureServices`: what
a deployment's log sink receives, not test output.

Ordering is a positive signal, not a hope. Each NodeType reports into an `AsyncSubject` from inside
its own `WithInitialization`, once `WatchControlPlane` has **returned** — and
`SubscribeWithReEstablish` establishes synchronously, with `MeshNodeStreamHandle.Subscribe`
reporting a failed `AcquireStream` through `OnError` on the calling thread. By the time the signal
arrives the line has been written or provably will not be. (An assertion on a log's absence with no
such signal passes whenever it runs first — which is most of the time.)

| Build | `…LogsNoControlPlaneFault` | `…OpensNoSyncSubHubOnAProbe` | `StreamCache_ReadOfARegistrationProbeAddress_CompletesEmpty` |
|---|---|---|---|
| `origin/main` | **FAIL** — 2 Error records, the reported line verbatim (one from the platform's `kernel` type, one from the test's) | **FAIL** — 3 `sync/`-into-disposal warnings | **FAIL** — `DeliveryFailureException: No node found at 'content-type-registration/…'` |
| with (1) only | PASS | **FAIL** — 1 warning left (the claim arbiter) | — |
| with (1)+(2)+(3) | PASS | PASS | PASS |

The third row's third column was falsified independently by reverting **only** (3) and re-running.

## Docs

`Doc/Architecture/TransientNodeProbes` gains the registration probe in its producer table, the
**third** own-address read seam, and a section on the own-node watchers with the boot-time Error it
removed; `Doc/Architecture/ActivityControlPlane` gains the probe rule, a cross-link, and the missing
fourth fault class its table never listed. What's New entry: `Category: Fix`.

https://claude.ai/code/session_01SD7aiLSo59xng2TzU32xEa

### DIFF
--- a/src/MeshWeaver.Data/TransientNodeProbe.cs
+++ b/src/MeshWeaver.Data/TransientNodeProbe.cs
@@ -84,6 +84,20 @@ public static class TransientProbeAddresses
     public const string SchemaLookupProbePrefix = "_schema_lookup/";
 
     /// <summary>
+    /// Prefix of the boot-time content-type registration probe's address —
+    /// <c>ContentTypeRegistration.ProbeRegister</c> (MeshWeaver.Graph).
+    ///
+    /// <para>🚨 That producer minted the literal itself, which is precisely the drift this class
+    /// exists to prevent: it is a transient probe by every other measure — same
+    /// <c>AsTransientNodeProbe</c> marker, same create-and-dispose-in-one-breath lifetime, same
+    /// impossibility of ever carrying a node — yet <see cref="IsProbeAddress"/> did not recognise
+    /// it, so the one guard keyed off the ADDRESS rather than the configuration marker
+    /// (<c>MeshNodeStreamCache.GetStreamRaw</c>, #2894) silently did not fire for it. Added with
+    /// Systemorph/MeshWeaver#2990.</para>
+    /// </summary>
+    public const string ContentTypeRegistrationProbePrefix = "content-type-registration/";
+
+    /// <summary>
     /// True when <paramref name="path"/> is a transient probe hub's own synthetic address — a
     /// path that is not, and can never become, a mesh node.
     /// </summary>
@@ -94,5 +108,6 @@ public static class TransientProbeAddresses
            && (path.StartsWith(ModelProbePrefix, StringComparison.Ordinal)
                || path.StartsWith(SchemaProbePrefix, StringComparison.Ordinal)
                || path.StartsWith(SchemaValidationProbePrefix, StringComparison.Ordinal)
-               || path.StartsWith(SchemaLookupProbePrefix, StringComparison.Ordinal));
+               || path.StartsWith(SchemaLookupProbePrefix, StringComparison.Ordinal)
+               || path.StartsWith(ContentTypeRegistrationProbePrefix, StringComparison.Ordinal));
 }
--- a/src/MeshWeaver.Data/Workspace.cs
+++ b/src/MeshWeaver.Data/Workspace.cs
@@ -919,6 +919,14 @@ public class Workspace : IWorkspace
                 .Remove(new KeyValuePair<WorkspaceReference, Lazy<ISynchronizationStream>>(key, entry));
     }
 
+    /// <summary>
+    /// 🚨 The failure NAMES the reference and the owning hub. It used to throw a bare
+    /// <c>InvalidOperationException("Failed to create stream")</c> — no reference, no owner, no
+    /// inner cause — and that is exactly why three deliberate fault classifiers in
+    /// <c>SubscribeWithReEstablish</c> all missed it and reported a routine boot-time probe as a
+    /// transient prod ERROR (Systemorph/MeshWeaver#2990). A diagnostic that cannot be classified
+    /// costs more than the line it saves.
+    /// </summary>
     private ISynchronizationStream<TReduced> ReduceLocalStream<TReduced>(
         WorkspaceReference<TReduced> reference,
         Func<StreamConfiguration<TReduced>, StreamConfiguration<TReduced>>? configuration)
@@ -926,7 +934,10 @@ public class Workspace : IWorkspace
                this,
                reference,
                configuration
-           ) ?? throw new InvalidOperationException("Failed to create stream");
+           ) ?? throw new InvalidOperationException(
+               $"Failed to create stream for {reference} on {Hub.Address}: the workspace's "
+               + $"ReduceManager has no reducer producing {typeof(TReduced).Name} from this "
+               + "reference, or the data source it would reduce from has not been started.");
 
     /// <inheritdoc />
     public ISynchronizationStream<EntityStore> GetStream(params Type[] types)
--- a/src/MeshWeaver.Documentation/Data/Architecture/ActivityControlPlane.md
+++ b/src/MeshWeaver.Documentation/Data/Architecture/ActivityControlPlane.md
@@ -158,11 +158,20 @@ Faults are handled by `SubscribeWithReEstablish`, not by a bare log-and-die `OnE
 
 | Fault class | Behaviour |
 |---|---|
+| Own hub tearing down (`HubDisposingException` naming this watcher's own address) | **Terminal, and not an error.** From the first instant of `Dispose` the hub refuses hosted-hub creation, so the own-node stream cannot build its `sync/` sub-hub. The watcher is meant to die with its hub — a fresh activation installs a fresh one. Logged `Debug`; re-establishing here reported routine teardown as a prod `Error` and armed a timer rooted at the hub being collected. |
 | Own-node content cannot be deserialized (`MeshNodeStreamException` / `Deserialization`) | **Terminal.** The faulted stream would replay the same poisoned emission, so re-establishing is a 1 Hz poison loop. Logged `Critical` and surfaced through the optional `onPoisonedContent` sink; the watcher stays down until the content is repaired and the hub re-activates. |
 | Own node gone (routing `NotFound`) | **Terminal.** Re-subscribing re-issues a doomed cross-hub `SubscribeRequest` forever — the 2026-06-10 prod storm. Logged and stopped; the orphaned hub idle-disposes. |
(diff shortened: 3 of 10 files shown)


rbuergi:
Review finding acted on in #3007 — it arrived after auto-merge had already taken this PR, so it lands as a follow-up rather than a push here. `GetStream(params Type[] types)` now names the resolved collections and the hub, matching `ReduceLocalStream`'s contract.

PR #3071, opened 2026-09-02
fix(bake): the node-repo bake + gate run AS the platform image — portal supplies the reference set, tester executes (#3022)

### LINKED ISSUE #3022: main-cd: the bake's framework identity does not match the promoted portal image, so the publication is never sealed and every satellite is starved
`main-cd`'s **`Bake platform content in the shipped image + publish (linux-x64)`** fails at its last step, `Assert the bake's identity is the one the PROMOTED PORTAL IMAGE resolves`, and because it does, the two downstream jobs that satellites depend on are **skipped**:

```
skipped  Plugins: resolve this run's tester image digest
skipped  Plugins: pack the module bundles the bake composes
skipped  Plugins: bake + seal the publication for this identity
```

## The measurement

Run [33566326021](https://github.com/Systemorph/MeshWeaver/actions/runs/33566326021), head `30dda849`:

```
the bake published under: s8f66b561b1622d91d37350bbc2c8e407
meshweaver.azurecr.io/memex-portal-ai:30dda84
portal /app: 46 MeshWeaver assemblies, 46 manifest lines
sd9d6036adc810f598506ef4c930452fd
framework-identity: MISMATCH — the bake published under 's8f66b561b1622d91d37350bbc2c8e407'
but '/portal' resolves 'sd9d6036adc810f598506ef4c930452fd'.
  Both manifests record every canonical content-surface assembly, so the difference is in the
  recorded HASHES — different binaries (a different architecture, or hosts built from different
  commits). Check that both were taken with the same --platform and from the same release.
```

Both images carry the **same commit tag `30dda84`** and both manifests list **46 of 46** canonical assemblies — so this is not a missing manifest (#1699) and not a missing assembly. The two images simply contain **different bytes for the same sources**:

| image | identity |
|---|---|
| `mw-plugin-test:30dda84` (what the bake ran in) | `s8f66b561b1622d91d37350bbc2c8e407` |
| `memex-portal-ai:30dda84` (what is promoted, and what asks) | `sd9d6036adc810f598506ef4c930452fd` |

The gate is doing its job — this is exactly the #1814 shape it exists to catch, and publishing anyway would put bundles on the shelf under an identity no pod ever asks for, so every pod would compile at boot as if no bake had run, silently.

## Why it matters beyond this run

**Nothing downstream of the seal can move while this stands.** The seal is what every satellite consumes; MeshWeaver.Education has already measured the consequence from the other side — its `schedule` and `meshweaver-framework-released` lanes fail on a 404 because *no publication exists under the identity they resolve*, and its maintainer note records the fix as "upstream: Plugins must publish for that identity".

This is **not** the previously-reported cause. Systemorph/MeshWeaver.Plugins#1121 attributed the skipped seal to two failing `MeshWeaver.AI.Test` cases in the pack job. Both are now fixed — `AiContentPackDriftTest` is 15/15 and `ExecuteScriptDispatchDiagnosticsTest` 5/5 locally, and `Module bundle (MeshWeaver.AI)` reports **success** in Plugins run 33551532882. The pack job is no longer the blocker; this identity mismatch is, and it sits one job earlier.

## Where to look

Two candidate causes, and they need distinguishing rather than guessing:

1. **The two images are built by different jobs and are not byte-reproducible.** Same sources, same commit, different assembly hashes is what you get when the builds differ in something the compiler records — normalized paths (`ContinuousIntegrationBuild`), SDK patch level, or `--platform`. If so, the fix is to make the two builds agree, or to have the bake run inside the image it is certifying.
2. **The two tags do not actually contain the same build.** `mw-plugin-test:30dda84` and `memex-portal-ai:30dda84` are separate pushes; if one is a cache hit from an earlier build of the same tag, the commit tag would agree while the bytes do not.

Worth resolving before treating this as a build-flag problem: compare the two manifests line by line and see *which* assemblies differ — if it is all 46, it is a global input (path/SDK/platform); if it is a handful, it is those projects' inputs.

🚨 Please do not "fix" this by relaxing the assert. Publishing under an identity the promoted image cannot resolve is the failure mode this gate was added for, and it fails silently everywhere else in the pipeline.

rbuergi:
Closes #3022 (the reference-set half; #3041 was the identity half). Neighbours: Plugins #1043, Plugins #1061.

## Root cause

The node-repo bake and gate (`node-repo-publish-bake.yml`, `node-repo-gate.yml`) ran `mw-plugin-test` against **the tester image's own `/app`** — reference set, framework identity and the environment the per-type dependency records are computed against all came from the process the bake happened to run in — while the bundles are adopted by the **portal**. The tester's `/app` is a strict subset of the portal's. Content binding a portal-only assembly compiles in every portal and failed the bake with a CS0234/CS0246 that names the content. Doctrine (`ModuleBuildArchitecture`: "the platform image is the compiler and the reference set") held for module compiles since #2907 and never reached the bake.

## Measurement (both images of `3.0.0-rc9.ci.7534`, `linux/amd64`, extracted with `docker cp`)

| | `mw-plugin-test` | `memex-portal-ai` |
|---|---|---|
| assemblies in `/app` | 88 | 219 |
| `MeshWeaver.*` assemblies | 26 | 46 |
| `MeshWeaver.*` only here | 1 (`Hosting.Monolith`) | **21**: `Maps`, `AI`, `Markdown.Collaboration`, `ContentCollections.Indexing[.Graph]`, `Blazor[.Portal,.Views]`, `Hosting.{AspNetCore,Blazor,Orleans,PostgreSql,SignalR,Grpc,Embeddings}`, `Connection.Orleans`, `InstanceSync`, `Documentation`, `{Speech,Observability,Markdown.Export}.Contract` |
| the 25 shared `MeshWeaver.*` | byte-identical 25/25 |
| manifest lines | 26 | 46 (shared names: identical hashes) |
| identity | `s8fe4902c0b2f5974f824be2867221dbd` | same (which is why #3041's gate is green) |

(a) `FrameworkBuildIdentity.ContentSurfaceAssemblies` on `origin/main` does **not** contain `MeshWeaver.Maps` (removed with #2941; the comment in Plugins' `Memex.Portal.Gui.csproj` calling it canonical is stale). The portal's direct reference is therefore invisible to the identity and the identity gate cannot see this class of gap at all.
(b) Plugins NodeType sources binding portal-only assemblies: `using MeshWeaver.Maps` in `Cornerstone/Pricing/{Source,Test}`, `AppleMaps/Gallery`, `GoogleMaps/Gallery`, `OpenStreetMap/Gallery` (CD run 33587509969 shows exactly these RED). `MeshWeaver.AI` (11 sources) is composed via `--module`. No content binds Indexing/Blazor/Hosting.* (`Indexing/index.json` only *declares* a module).

## Fix — the bake and its gate run AS the platform image (the module lane's shape since #2907)

* `node-repo-publish-bake.yml` / `node-repo-gate.yml`: new inputs **`platform-image` (required)** + **`platform-image-digest`** (resolved branch-for-branch like `image-digest`: release version tag / upstream payload `platform_image` / pin / `allow-unpinned` / RED — never a fallback to the tester's `/app`; the tester passed as the platform is refused). The lane pulls both images, extracts both `/app`, asserts with the tester's `framework-identity /portal --expect <tester identity>` that the two images are **one build**, composes the **gate host** (`.github/scripts/compose-gate-host.sh`, with `--self-test` wired into `dotnet-test.yml`: the portal's `/app` + the tester CLI; portal bytes win; the tester's manifest and `deps.json` never ride — the latter so the dotnet host probes the directory and every portal assembly is in the TPA), and runs `compile … --app /app --shared-frameworks /usr/share/dotnet/shared` and the gate `… --app /app` from that host on the portal image's own `dotnet` (`--user`, `HOME=/tmp` as the module lane). The dispatch payload carries `platform_image` so a woken dependent bakes against the same portal. After the compile the lane asserts `framework-mvid.txt` equals the portal's identity (true by construction; kept as the cheap invariant).
* `main-cd.yml`: `plugins-bake-image` resolves the promoted portal's digest too; `plugins-bake` passes `platform-image`/`platform-image-digest`.
* Tester: `BakeHost` (one directory → identity via `ResolveIdentityForDirectory`, surface pairs, MVIDs, reference set from `ContainerReferenceSet.Read(app, tpa: "", shared)`; **refuses** a host without a manifest and a host whose toolchain closure MVIDs differ from the running process's — the invariant that makes recording the portal's identity honest), used by `compile` and `build` (`--app`/`--shared-frameworks`, both-or-neither); `GateHostCheck` (`--app` on the gate = the process must resolve the host's identity, exit 2 otherwise); `ReferenceGapAttribution` (CS0234/CS0246 → `reference set lacks <assembly> (portal-shipped, not composed: modules/…)` or `no assembly in the reference set declares namespace '…'`; a referenced namespace adds nothing). `FrameworkBuildIdentity` gains public `ToolchainClosureOf(dir)` / `ImplMvidInDirectory(dir, name)`.
* Docs (same PR, minimum sections): `ModuleBuildArchitecture.md` (new section with the table), `CiContentBake.md` (identity-rule subsection + lane table), `ModuleClosureAccounting.md` (mirror-rule paragraph). Internal/CI change → **no What's New**.

## Identity gate

For the node-repo lanes "bake identity == portal identity" is now tautological; the comparison that can lose moved in front of the composition ("are the two images one build?"). `main-cd`'s own Doc bake keeps the original comparison (still meaningful). `check-image-build-identity.sh` (image-config version) is unrelated and untouched.

## Verification

* `dotnet build -c Release -warnaserror`: `tools/MeshWeaver.PluginTester`, `test/MeshWeaver.PluginTester.Test`, `test/MeshWeaver.Documentation.Test`, `test/MeshWeaver.Compiler.Pipeline.Test` — 0 warnings, 0 errors.
* Tests: `BakeAgainstPlatformHostTest` (bake keyed to a synthesized host, not the process; manifest-less host refused; toolchain drift refused by name; missing shared root refused; gap attribution incl. the composed-module control; gate host check), `NodeRepoLaneHostGuard` (workflow text), `PlatformBakeLaneGuard`, `DocumentationLinkIntegrityTest`, `FrameworkBuildIdentityTest`, whole `MeshWeaver.PluginTester.Test` — see the run log summary in the last commit / below. `compose-gate-host.sh --self-test`, `check-workflow-shell.py` (0 live findings).
* Not provable locally: the composed-host lane end to end (the images are `linux/amd64`; x64 .NET under emulation on this arm64 host crashes in both the tester image and the composed host alike — `AccessViolationException` in `List.Add`, so no evidence either way). The identity verbs did run from the composed host inside the portal image: process identity `s8fe4902…` = portal's `/app` = tester's. First real proof is the next `main-cd` run's `plugins-bake`.

## Cost (45-minute wall)

One extra image pull + `/app` extraction per bake/gate job (~1 min; the portal's `/app` is ~300 MB) and a larger, lazily-mapped reference set. Recent publish-bake runs are ~20 min; expect ~21–22.

## Plugins edits for the coordinator (after this merges; pin bump in the same commit)

`.github/workflows/ci.yml`: (1) `test-repos` → `with:` add `platform-image: meshweaver.azurecr.io/memex-portal-ai` and `platform-image-digest: <MW_PORTAL_IMAGE_DIGEST literal, = the sha256 already on modules-floor/modules-rest>`; (2) `preflight` step `bake-target` → also resolve the portal digest on FOLLOW_RELEASE (`docker manifest inspect -v meshweaver.azurecr.io/memex-portal-ai:<same tag as MW_TEST_IMAGE>`), export `bake-platform-digest`, else the pin; (3) `publish-bake` → `with:` add `platform-image: meshweaver.azurecr.io/memex-portal-ai` and `platform-image-digest: ${{ needs.preflight.outputs.bake-platform-digest }}`; (4) bump both `uses:` SHAs (`node-repo-gate.yml@…`, `node-repo-publish-bake.yml@…`) to this PR's merge commit. Education/Reinsurance/SocialMedia/Manufacturing: same two inputs on their `test-repos`/`publish-bake` calls when they bump. `node-repo-compile-check.yml` (compile-check.py) still extracts the reference set from `test-image`; passing the portal image there is a caller-side choice today — follow-up.


rbuergi:
Both Copilot threads are help-text nits on the tester CLI (a missing-value message for `--app`/`--shared-frameworks` on `build`, and `CompileUsage()` printing `mw-compiler` instead of `mw-plugin-test`). Neither affects the bake/gate behaviour this PR fixes; merging now (the seal is the live fleet blocker) and leaving both as follow-ups for the tester CLI.

Issue #1962, opened 2026-08-21
Preferences ▸ display time zone tab is absent on memex.meshweaver.cloud — split from #1936


rbuergi:
Split out of **#1936**, whose actual cause — the user directory being read as a 50-row page — is fixed in **#1960**. The missing tab was reported alongside it but is **a separate defect**: the investigation established the tab gates on `Permission.Update` over `ObjectId`, which was correct on both installs it was checked against.

## What is known

- Reported on `memex.meshweaver.cloud`: **Preferences → display time zone** does not render.
- The local install **shows the tab** while still rendering UTC — so the tab's absence and the UTC rendering are independent symptoms. That table is in #1936's body.
- The permission gate itself evaluated correctly where it was measured.

## The cheapest thing to eliminate first

From #1960's investigation, worth checking before anything else:

> an un-indexed user falls back to the **email-local-part guess** for `ObjectId`, so any user whose node id differs from that guess loses **every self-permission at once** — exactly a "tab vanished" shape.

That is **not** the reporter's case (`sglauser` matches the guess, so it cannot be the cloud symptom), but it is a real failure mode of the same shape and costs nothing to rule out. If a user's mesh node id ever diverges from their email local part, they lose self-permissions silently and every self-gated surface disappears at once.

Note that #1960 removes the *cause* of un-indexed users, so this fallback should now fire far less often — which makes it both less likely as the explanation here and more important to pin before it becomes rare enough to forget.

## Suggested next steps

1. Reproduce on memex as an affected user and capture what the gate actually evaluates — `ObjectId`, the resolved permissions, and whether the `User` node id matches the email local part.
2. If the gate is correct, the defect is in tab registration/contribution rather than permissions — check whether the tab is contributed at all on that install (the settings tabs ride the `UiContribution` lane, #1645/#1648/#1659).
3. Whatever the cause, add a test at the layer it lives in. #1960 added `ViewerTimeZoneRenderTest`, which passes on main — a reminder that the render path is not where this lives.

*Filed to keep #1936 closed on its measured cause rather than left open on an unrelated symptom.*

rbuergi:
Fixed in #2010 (merged).

**The cause was a race, not a data difference between installs** — which is why the investigation kept finding the permission gate evaluating correctly wherever it was measured.

[...]

Issue #1645, opened 2026-08-15
DESIGN WS7: menus and standard views as plugin content — the UiContribution mechanism


rbuergi:
The composition-first end state ("deployment = base runtime + plugin composition; ALL menus + ALL standard views become content of a pre-installed platform-UI plugin"). Wave 2–4 modularized the compiled packs; this designs the CONTENT lane for the UI shell itself.

## Inventory (measured on current main)
- **Node/mesh menus**: a reactive provider pipeline (`INodeMenuProvider` DI + `AddNodeMenuItems` delegates per context) with the defaults compiled in `DefaultNodeMenuProvider`/`DefaultMeshMenuProvider`; presentation is already editable data (`MenuPresentation` catalog) but **deliberately cosmetic/subtractive only**.
- **Settings tabs**: `GlobalSettingsMenuItemProvider` collections; ~20 compiled `Add*SettingsTab` calls in the portal composition (the ContentIndex tab already rides its module).
- **Top-bar AI menu / side panel**: compiled `AddNodeMenuItems(AiMenuContext…)` / `("SidePanel"…)`.
- **Home tabs**: `HomeTab` NODES — already pure data; the precedent this design generalizes.
- **Per-NodeType menu items**: plugins already have FULL power via in-mesh `AddNodeMenuItems` delegates in their NodeTypes' configuration lambdas (plugin-gate CI'd). Not a gap.
- **The gaps**: plugins cannot contribute (a) every-node menu items, (b) settings tabs, (c) shell menus — and the defaults are compiled.

## The crux: the security boundary
`MenuPresentationOverlay` is cosmetic-only for a documented reason: permission filtering happens by providers *not emitting* an item, so **an item conjured from data would carry no permission gate**. The mechanism therefore moves enforcement into the compiled aggregator with a CLOSED declarative gate vocabulary:

- `RequiredPermission` — enforced by the aggregator against the live permission stream (per-viewer reactive, exactly like today);
- node-shape gates: `nodeTypes` (suffix-aware `Matches` semantics), `syncedOnly`, `excludePartitionRoot`, `excludeViewerHome`, `adminOnly` (`IsGlobalAdmin`).

Anything beyond the vocabulary stays in code (in-mesh per-NodeType delegates, or compiled providers). A contribution can only ever NARROW visibility relative to what the aggregator enforces — never widen it.

## The contribution node
nodeType **`UiContribution`** (lives in the contributing plugin's partition):
[...]

PR #3089, opened 2026-09-02
The per-node settings surface gets its own UiContribution lane (#3055)

### LINKED ISSUE #3055: WS7 step 4, the core half: the per-node settings surface has NO UiContribution lane (and two gate words are still missing)
Follow-up to the closed design **#1645**, and the core half of
Systemorph/MeshWeaver.Plugins#960. The Plugins-repo half shipped as
Systemorph/MeshWeaver.Plugins#1162 (a `preInstalled` **PlatformUI** pack that declares the shell's
data-driven top-bar menus, plus a gate for the whole silent class). Everything below is
`MeshWeaver.Graph` / `Memex.Portal.Shared` and can only be done here.

Measured on `MeshWeaver@71869075f`.

## Where WS7 actually stands

Slices 1–3 and settings-tab tranche 2 all landed: `UiContributionNodeType` + the closed gate
vocabulary + the compiled `UiContributionProjection`, the `TopBar` menu-declaration context, the
mesh-scoped live `UiContributionCatalog` (which **superseded** the design's `Admin/UiGeneration`
bump — it cannot go stale and needs no hub recycle), and thirteen platform entries now seeded as
`Admin/UiContribution/*` nodes (7 settings tabs + 6 AI-menu entries; verified on the live mesh).

## 🚨 The real blocker for the remaining settings tabs is NOT the gate vocabulary

#960 records it as *"most of the thirteen carry visibility the closed gate vocabulary cannot express
yet"*. That is not what stops them. There are **two settings surfaces**, and only one has a
contribution lane:

| surface | registration | definition type | contribution lane |
|---|---|---|---|
| **Global settings** | `AddGlobalSettingsMenuItems` | `GlobalSettingsMenuItemDefinition` | ✅ `ContributedSettingsTabs` in `GlobalSettingsMenuItemsExtensions.EvaluateGlobalSettingsMenuItems` |
| **Per-node settings page** | `AddSettingsMenuItems` | `SettingsMenuItemDefinition` | ❌ none — `SettingsMenuItemsExtensions.cs` does not mention `UiContribution` at all |

`UiContribution.SettingsContext` projects into `GlobalSettingsMenuItemDefinition` only
(`UiContributionProjection.ProjectSettingsTabs`). **All ten remaining tabs register on the per-node
surface**, so there is nowhere for them to go regardless of gating:

| tab | file |
|---|---|
| `AddApiTokensSettingsTab` · `AddInstancesSettingsTab` · `AddNotificationsSettingsTab` | `memex/Memex.Portal.Shared/Settings/` |
| `AddGitHubSyncSettingsTab` · `AddGitHubIssuesTab` | `src/MeshWeaver.GitSync/` |
| `AddCouponAdminSettingsTab` · `AddInstanceGrantAdminSettingsTab` · `AddCompositionAdminSettingsTab` | `src/MeshWeaver.PluginCatalog/` |
| `AddGlobalAdminSettingsTab` · `AddUserPreferencesSettingsTab` | `src/MeshWeaver.Graph/Configuration/UserNodeType.cs` |

(The other two of the thirteen are no longer core's: `AddPartitionSyncSettingsTab` moved to
`MeshWeaver.Plugins/src/MeshWeaver.Graph.Views` and `AddTokenUsageSettingsTab` to
`MeshWeaver.Plugins/src/MeshWeaver.AI` with their modules. Both register on the per-node surface too,
so they are blocked by the same gap even though the code lives elsewhere.)

## The precise asks

1. **Give the per-node settings surface the same lane.** Add a `ContributedSettingsTabs` equivalent
   to `SettingsMenuItemsExtensions`, projecting `UiContribution` into `SettingsMenuItemDefinition`.
   Open question worth deciding here rather than in the migration: does it reuse the `Settings`
   context, or does the per-node surface get its own context key? A single key means every seeded
   tab appears on both surfaces.

2. **`SettingsMenuItemDefinition.Keywords` has no `UiContribution` equivalent.** It backs settings
   SEARCH, and it is not decoration — `PartitionSyncAdminLayoutArea` ships fifteen keywords
   (`"partitions"`, `"sync source"`, `"decouple"`, `"delete space"`, …). Migrating a tab without a
   `keywords` field silently removes it from settings search. Add `Keywords` to `UiContribution`, or
   accept and record the loss.

3. **The two gate words #1645 specified and `UiContributionGates` still lacks** — the record today
   carries only `NodeTypes`, `ExcludePartitionRoot`, `AdminOnly`:
   - `syncedOnly` — StopSync only on synced nodes;
   - `excludeViewerHome` — Edit/Move/Copy/Delete suppressed on a viewer's own home.
   These block the **node-menu** defaults (step 4 item 2), not the settings tabs.

4. **The minimal compiled fallback** (step 4 item 3): Data + Versions + Delete, enough to operate a
   zero-plugin mesh. Note the design names `DefaultNodeMenuProvider` / `DefaultMeshMenuProvider` —
   neither exists; the defaults are `AddNodeMenuItems` delegates scattered across the layout areas,
   so this is a consolidation before it is a retirement.

## One thing worth copying back

A contributed entry naming a context nobody declares renders **nowhere**, and nothing anywhere says
so — no error, no warning, not even an area-not-found placeholder. Six shipped entries were dark for
nine days that way (Plugins#1162). `MeshWeaver.Plugins/scripts/check-menu-contexts.py` catches it
statically, along with an empty `Area` (dropped before any gate runs) and a non-portal-internal
`Href` (rejected by `ResolveHref`, so the entry quietly opens the derived area URL). Core seeds
contributions from compiled code (`PlatformSettingsTabAreas`, `AiMenuContributions`) and has no such
check; a pinning test over `Seeds` would be the cheap equivalent.

rbuergi:
Closes #3055.

WS7 step 4, the core half. The global settings surface has had a `ContributedSettingsTabs` lane
since slice 2; the per-node one had **none** — `SettingsMenuItemsExtensions` did not mention
`UiContribution` at all, and `UiContributionProjection.ProjectSettingsTabs` projected into
`GlobalSettingsMenuItemDefinition` only. All ten remaining tabs register on the per-node surface,
so gating was never what blocked them: there was nowhere for them to go.

## Ask 1 — the lane, with its OWN context key

`UiContribution.NodeSettingsContext` (`"NodeSettings"`) → `UiContributionProjection.ProjectNodeSettingsTabs`
→ `SettingsMenuItemsExtensions.ContributedSettingsTabs`, mirroring the global lane's shape rather
than inventing a parallel one.

**The decision, and why.** The issue leaves open whether the per-node surface reuses `Settings` or
gets its own key. It gets its **own key**:

* A single key means every seeded tab appears on **both** surfaces. Seven platform tabs are already
  seeded as `Admin/UiContribution/*` for the global page (What's New, About, Privacy, Invitations,
  Inbox, Updates, Published to the web) — reusing `Settings` would list all seven on *every node's*
  settings page. That is a visible regression shipped in the name of a refactor.
* The two surfaces are different pages with different definition types, routes and content-builder
  signatures (the per-node builder receives the anchoring `MeshNode`; the global one has no node).
* Two keys make the surfaces **independently gateable**, which is the property the contribution lane
  exists for. A tab that genuinely belongs on both is two contributions, and says so.

The cost is one `const string`. `UiContributionProjectionTest.NodeSettingsTabs_AreASeparateSurface_FromTheGlobalSettingsTabs`
pins the decision in both directions.

🚨 **The per-node lane takes no permission and does not filter on one.** It stamps
`RequiredPermission` (floored at `Read`) onto the definition and lets `FilterByPermission` apply it
at the render fold, against the *latest* value. Filtering inside the provider stream is precisely
the #1962 defect — a long-lived chain frozen on an early `Permission.None` seed that later
re-renders the menu with every entitled tab silently missing. The Read floor is what still makes an
anonymous viewer see nothing.

## Ask 2 — `Keywords`

Added to `UiContribution` and projected into `SettingsMenuItemDefinition.Keywords`, which backs the
settings search box (`SettingsLayoutArea.FilterMenuItems` matches Label ∪ Group ∪ Keywords).
Migrating a tab without it removes the tab from search with no other symptom —
`PartitionSyncAdminLayoutArea` alone ships fifteen terms. Consumed by `NodeSettings` only: the
global settings page has no search box and `GlobalSettingsMenuItemDefinition` has no keyword slot.

## Ask 3 — the two missing gate words

* **`SyncedOnly`** — the node still participates in sync (`SyncBehavior.Include`), the shape the
  "Stop synchronization" entry needs. A node that could not be resolved fails the gate: narrowing,
  never widening, on missing evidence. The inverse ("only on claimed nodes", the *Resume* branch) is
  deliberately **not** added — a second gate word is its own decision and the vocabulary stays closed.
* **`ExcludeViewerHome`** — the anchoring path IS the viewer's partition key, the same comparison
  `PinLayoutArea` and `PresentationLayoutArea` already make. **Strictly narrower** than
  `ExcludePartitionRoot` and not a replacement for it: `ExcludePartitionRoot` suppresses on *any*
  user's home (so an admin browsing someone else's home still gets no Delete entry), this one only
  on the viewer's own. Declare both when both apply — there is a test for exactly that distinction.

All gate words now evaluate in **one** place (`UiContributionProjection.PassesNodeGates`), so the
node menu and the per-node settings page cannot drift on what a gate word means.

## The seed-integrity check (the issue's last section)

`UiContributionSeedValidation` — core's equivalent of `MeshWeaver.Plugins/scripts/check-menu-contexts.py`,
as a pure function so a *test* enforces it rather than a script. It catches the whole silent class:

| defect | symptom without the check |
|---|---|
| a `Context` nobody declares | renders **nowhere** — no error, no warning, no placeholder. Six shipped entries were dark for nine days that way |
| an empty `Area` | dropped before any gate runs, so relaxing gates does not find it |
| a non-portal-internal `Href` | discarded; the entry quietly opens the *derived* area URL instead |
| a `NodeType` that is not `UiContribution` | the catalog query never returns the node |
| a `Label`/`Group` with no key | English for every German viewer |
| two seeds at one path | the catalog is keyed on path — one silently replaces the other |

It folds in the context keys a `TopBar` declaration introduces (an entry targeting a menu declared
in the same set is legitimate) and takes optional `additionalContexts` / `registeredAreas`, so the
Plugins repo can call the same function over `AiMenuContributions.Seeds`.

**Control arms**, because a check that cannot fail reproduces the silence it exists to break:

* `UiContributionSeedValidationTest` — 13 tests, one per defect, each asserting the validator
  *reports* a deliberately broken seed, plus the positive cases and an assertion that
  `PlatformContexts` actually contains the keys the platform projects (a set that drifted empty
  would make the negative tests pass for the wrong reason).
* `PlatformSettingsTabSeedTest.Every_Seed_Passes_The_Contribution_Integrity_Check` runs it over the
  **real** seed list, then (arm 1) asserts the seed count so an emptied `Seeds` cannot pass having
  checked nothing, and (arm 2) breaks one real seed six ways and requires the validator to reject
  each — a validator degenerated into "no problems" reds there.

## 🚫 Ask 4 is NOT included

The minimal compiled fallback (Data + Versions + Delete) is not in this PR, and it is not a matter
of time. Three things block it, and two of them are outside this repo:

1. **The PlatformUI pack must carry the retired defaults first.** It lives in `MeshWeaver.Plugins`
   and today declares the shell's top-bar menus, not the node-menu default item set. Shrinking
   `DefaultNodeMenuProvider` here before that lands removes Edit/Pin/Presentation/Move/Copy/Files/
   StopSync/Recycle from every mesh that does not have the pack — a shipped regression, not a
   migration.
2. **Two defaults cannot be expressed by the closed vocabulary at all.** The Presentation entry's
   *label* flips (Hide/Show) on the viewer's live screen, and StopSync's flips (Stop/Resume) on the
   node's `SyncBehavior`. `SyncedOnly` (this PR) migrates StopSync's *Stop* half; the *Resume* half
   needs the inverse gate that #1645 deliberately excludes. A label that depends on viewer or node
   state is behaviour, which the vocabulary is designed not to carry — so these stay compiled, and
   saying "the defaults retire" without saying that would be inaccurate.
3. **The separators are computed, not declared.** The `_separator` entries at Order 20/40 exist only
   where both adjacent sections are non-empty — pure compiled logic over the *result*.

The issue calls item 4 "a consolidation before it is a retirement", and notes the design's
`DefaultNodeMenuProvider`/`DefaultMeshMenuProvider` do not exist. They do, as private methods in
`NodeMenuItemsExtensions` — so the consolidation the design imagined is already done; what remains
is the retirement, which is cross-repo. What this PR contributes to item 4 is the vocabulary half:
`SyncedOnly` and `ExcludeViewerHome` were the two gate words the node-menu defaults were missing.

## No What's New entry

Deliberate, per the step-0.5 rule: this is a mechanism plus a test with **no user-visible effect** —
no tab has been migrated onto the lane, no gate word is declared by any shipped contribution yet.
The first PR that moves a tab onto the per-node lane is the user-facing one and gets the entry.

## One finding recorded for the migration: `RequiredPermission = Permission.None`

Found while checking the lane against the ten blocked tabs. `FilterByPermission` passes a
`Permission.None` tab **unconditionally** — it is chrome every viewer sees — but a contribution can
never demand less than `Read`, and that floor is exactly what makes the lane fail closed. So a
compiled per-node tab registered as `None` (Notifications is one of the ten) becomes `Read`-gated
the moment it moves onto the lane.

Same viewers in practice — you cannot open a node's settings page without reading the node — with
one real difference: a viewer holding `Permission.None` on the node loses the tab. That is the
closed vocabulary working as designed (a contribution narrows, never widens), but it IS a behaviour
change, so it is now recorded on the doc page for whoever plans the migration and on
`RequiredPermissionFloor` for whoever reads the code, rather than being discovered afterwards.

## Verification

* `dotnet build … -c Release -warnaserror` → `0 Error(s)`: `MeshWeaver.Graph`,
  `Memex.Portal.Shared`, `MeshWeaver.Graph.Test`, `Memex.Portal.Shared.Test`,
* Full, **unfiltered** test projects: `MeshWeaver.Graph.Test` (1202), `Memex.Portal.Shared.Test`
  (592), `MeshWeaver.Documentation.Test` (189 — the ratchet-guard home, so the new files are
  scanned) — all green.
* `scripts/check-record-signatures.py` → `✓ 8 changed C# file(s): no binary-breaking
  primary-constructor changes`. `UiContribution` / `UiContributionGates` carry init-only
  properties, not primary constructors, so the new fields are additive;
  `SettingsMenuItemDefinition`'s primary constructor is untouched.

The design decision is committed as a doc page, not left in this description:
`Doc/Architecture/UiExtensibility` gains the two-surfaces table with the reasoning, the two gate
words, `Keywords`, and a "a contribution nobody consumes is DARK" section pointing at the validator.

https://claude.ai/code/session_01PPe8CH4mFU9E74HafWppqL

### DIFF
--- a/src/MeshWeaver.Documentation/Data/Architecture/UiExtensibility.md
+++ b/src/MeshWeaver.Documentation/Data/Architecture/UiExtensibility.md
@@ -220,6 +220,17 @@ long-lived chain — the #1962 defect, where an early `Permission.None` seed sta
 later re-renders the menu with every entitled tab silently missing. The Read floor is what still
 makes an anonymous viewer see nothing.
 
+🚨 **One compiled tab shape does not survive migration unchanged: `RequiredPermission =
+Permission.None`.** `FilterByPermission` treats a `None` tab as chrome *every* viewer sees, but a
+contribution can never demand less than `Read` (that floor is what makes the lane fail closed), so
+a `None` tab becomes `Read`-gated the moment it moves onto the lane. In practice that is the same
+set of viewers — you cannot open a node's settings page without reading the node — with one real
+difference: a viewer holding `Permission.None` on the node loses the tab. Per-node tabs registered
+as `None` today (Notifications is one) are therefore a deliberate decision at migration time, not a
+copy-paste: either accept the narrowing, or leave that tab compiled. This is the closed vocabulary
+working as designed — a contribution narrows, never widens — but it is a behaviour change, so it
+belongs in the migration PR's description rather than being discovered afterwards.
+
 **Whole top-bar menus** (`Context: TopBar`): the contribution declares a NEW dropdown — its `Area`
 names the menu's own context key, `Label`/`Icon`/`Order`/`Tooltip` style the button, and its
 entries are ordinary contributions targeting that key. The gates apply to the declaration itself
--- a/src/MeshWeaver.Documentation/Data/Architecture/UiExtensibility.md
+++ b/src/MeshWeaver.Documentation/Data/Architecture/UiExtensibility.md
@@ -146,7 +146,8 @@ in its `TypeRegistry` or it degrades to an untyped `JsonElement` and the area re
 | A named area on a node type | `LayoutDefinition.WithView(name, generator)` | the NodeType's configuration — in-mesh or compiled |
 | Default areas on every node | `AddDefaultLayoutAreas()` composition | core only — extending it puts your area on *every node in the mesh*; prefer a NodeType-scoped area |
 | A left-rail navigation for a node family | `INodeNavigationProvider` | DI singleton (pack) — claimed by node shape at render time |
-| A settings tab | a `UiContribution` node (`Context: Settings`) pointing at a layout area — or compiled `AddGlobalSettingsMenuItems(...)` for content that needs code | mesh data / hub configuration |
+| A GLOBAL settings tab (`/_Setting/GlobalSettings/{id}`) | a `UiContribution` node (`Context: Settings`) pointing at a layout area — or compiled `AddGlobalSettingsMenuItems(...)` for content that needs code | mesh data / hub configuration |
+| A PER-NODE settings tab (`/{nodePath}/Settings/{id}`) | a `UiContribution` node (`Context: NodeSettings`) — or compiled `AddSettingsMenuItems(...)` | mesh data / hub configuration |
 | Node-menu entries / presentation | a `UiContribution` node (`Context: Node`/`Mesh`/`AI`/`SidePanel`/…) — or compiled `AddNodeMenuItems(...)`; presentation stays editable data (`MenuPresentationOverlay`) | mesh data / hub configuration |
 | A whole NEW top-bar menu | a `UiContribution` node (`Context: TopBar`) declaring the dropdown; entries target its key | mesh data — no code at all |
 | A home-screen tab | a `HomeTab` node | mesh data — no code at all |
@@ -171,7 +172,7 @@ The full field surface:
 
 | Field | Meaning |
 |---|---|
-| `Context` | Which menu: `Node` (default), `Mesh`, `Settings`, `AI`, `SidePanel`, `GitHub`, any registered context — or `TopBar` (below) |
+| `Context` | Which menu: `Node` (default), `Mesh`, `Settings` (GLOBAL settings page), `NodeSettings` (PER-NODE settings page), `AI`, `GitHub`, any registered context — or `TopBar` (below) |
 | `Area` | The layout area the entry opens (settings: embedded into the pane; menus: the link target) |
 | `Href` | Optional explicit link overriding the derived area URL — **portal-internal only** (a single-slash-rooted path like `/search?…`; schemes and `//host` forms are discarded by the compiled gate and the entry falls back to its area link) |
 | `Label` / `LabelKey` | Display text; the key resolves against the shared localization catalog |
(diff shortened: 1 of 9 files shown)


rbuergi:
**Held as a draft deliberately — this is not incomplete work.**

The change set is finished and locally green (see the Verification section of the description).
It is a draft because the requester asked explicitly that it not be merged, and this repo's
auto-arm lane had already enabled auto-merge on it (`app/meshweaver-cloud`, 09:59Z) — so leaving it
"ready" would have merged it on green regardless of that instruction. Converting to draft is the
sanctioned way to hold a PR here; it disarmed auto-merge.

Mark it ready for review to release it. Note that doing so **re-arms auto-merge**, so it will merge
itself once CI is green — that is the intended behaviour, just worth knowing before clicking.

Issue #2938, opened 2026-09-01
Cross-silo reply to a pod-local hub is refused — authenticated bundle requests 503 on whichever replica does not own the MeshWeaverInstance hub


rbuergi:
## Symptom

Authenticated requests to the plugin registry's sealed-bundle endpoints intermittently answer **503 + Retry-After**, at roughly 50% on `memex.meshweaver.cloud`. The bundle *index* fetch succeeds; the individual blob download fails, on a **different bundle each attempt** (`AppleMaps.zip`, `Analysis.zip`, `Chess.zip`, `Approvals.zip`).

It is not content-related, not the module wave, and not the stale image fixed in Memex#154 / #2935:

- `Systemorph/education` main flapped all night on the **unchanged** commit `1ca4beb3`: fail 23:55, pass 02:04, pass 03:46, fail 05:35, pass 06:44.
- It still reproduces **after** memex-cloud was rolled to `rc9.ci.7231` (re-runs at 07:16 and 07:23 today).
- `MeshWeaver.Reinsurance` PRs died on the same signature overnight.

## Mechanism — the request arrives, the REPLY cannot get back

Two pods, `memex-cloud`, same build `rc9.ci.7231`. The instance-key resolution that gates every authenticated bundle request is a `GetDataRequest` to a `MeshWeaverInstance/<id>` per-node hub. Both of those hubs are activated on pod **q5zcm**. Pod **7sqbk** hosts the pod-local hub `portal/nodeops-0IalraLccE-vUcjDU8yXeQ`.

**On 7sqbk (the requester)** — 24 occurrences in 90 minutes:

> `Instance-key resolution UNAVAILABLE for /api/plugins/bundles/prebuilt/…/plugins/Analysis.zip (MeshWeaverInstance/a014d18ff324: GetMeshNode(…) timed out after 10.0s (budget 10s) — the owning per-node hub never answered the GetDataRequest. This is NOT 'node not found'. Reader: Hub portal/nodeops-0IalraLccE-vUcjDU8yXeQ … Target: NO LOCAL HUB at 'MeshWeaverInstance/a014d18ff324' — it never activated here (or is owned by another silo)) — answering 503 + Retry-After, NOT 401: nothing was established about the presented key`

**On q5zcm (the owner), the same window** — the reply exists and is refused:

[...]

Issue #2901, opened 2026-08-31
A freshly uploaded content file 503s for minutes (10s timeout per request) before it starts serving — /api/content settle window


rbuergi:
## Measurement (memex.systemorph.com, 2026-08-31 09:10–09:30Z)

After `upload` of 32 course binaries (videos + posters, 0.3–9 MB) via MCP, `/api/content/{Course}/content/…` behaved per file, not per route:

```
AgenticPrimer/content/og.png                       200  0.9s   (immediately)
AgenticPrimer/content/videos/skills.mp4            503 10.3s → later 200 2.7s  (healed untouched)
AgenticBusiness/content/og.png                     503 10.4s ×2 over ~15 min → 200 6.4s
DataModeling/content/videos/datamodeling.poster    503 10.3s ×3 → 200 1.7s after re-upload + ~60s
DataModeling/content/videos/datamodeling.mp4       503 10.3s repeatedly; timed out (>15s) under a
                                                   concurrent bulk upload
```

Every failure is ~10.2–10.8 s then **503** — an internal 10 s timeout, repeated on every probe until the file "settles" minutes later. The same files on memex.meshweaver.cloud (uploaded weeks ago) serve in <1 s including range requests (206).

## Why it matters

- An editor who uploads a video and immediately previews it sees a broken player for minutes, with no signal whether the upload failed or is settling — and each preview attempt burns a 10 s server-side wait (these stack: under the bulk upload the same read exceeded 15 s).
- 503 is also what a genuinely broken route returns (#2876 family), so monitoring cannot tell "settling" from "down".

## Asks

- [ ] Root-cause the settle window: what takes minutes between an acknowledged content write and the first successful serve (chunking? collection-config propagation? per-node hub warmup? a breaker latched by the first early read)? The fix is at that edge, not a longer timeout.
- [ ] While settling, answer honestly: 409/425-style "not ready" (or block until ready under a bound), never a generic 503 after a silent 10 s.

Repro: upload any ≥400 KB binary into a content collection on the ACA deployment, GET it via `/api/content/...` within a few seconds, repeat until 200.

rbuergi:
**Diagnosis correction — this is NOT a settle window.** A longer probe (7 min, 4 URLs, every ~60 s) shows the same URL flapping 200↔503 randomly per request, uncorrelated with upload time:

```
12:06:23 poster=200 mp4=206 og=200 dm-mp4=503
[...]

PR #3090, opened 2026-09-02
docs(#2901): the /api/content 503 is the pod-hub claim loss, measured end to end

rbuergi:
Conserves a root-cause investigation that would otherwise have lived only in an issue thread.

## The finding

#2901 has carried **two** root-cause attributions and neither survives measurement — a post-upload "settle window" (already retracted in the thread) and the memory PubSubStore / streams-retirement programme (#1729/#1742, still recorded as the cause).

Measured on `memex-cloud` today, it is the **pod-hub claim loss** — #2938/#2915 — already fixed in core by `f37f3fc87` ("fix(orleans): re-assert the pod-hub claim on every membership change", merged 2026-09-02 02:46). **Production does not have it.**

## Why the evidence is worth keeping

The issue's own probe was re-run first and came back **18/18 `206`, all sub-2.5s** — which proves nothing about a per-replica defect, and the investigation says so rather than declaring victory. The log distribution is what discriminates:

```
sum by (namespace,pod) (count_over_time({namespace=~"memex|memex-cloud"} |= "Content read timed out" [24h]))
  memex-cloud / …-rfjgc   33
  every other pod, both namespaces   0
```

**33 occurrences in 24h, all on ONE pod**, zero across six sibling replicas and zero in the other namespace. That alone excludes a settle window and excludes a cluster-wide transport fault.

The terminal then names the link:

```
Reader: Hub portal/nodeops-… RunLevel=Started Queue(buffer=0,deferred=0,exec=0)
        PendingCallbacks=1[…=GetDataRequest@DoublePendulum(10003ms)]
Target: NO LOCAL HUB at 'DoublePendulum' — it never activated in this process
        (or it is owned by another silo and the reply was lost in transit).
```

Reader `Started`, queue empty, one callback outstanding — **not** a wedged caller, **not** a busy caller, **not** the request leg. The reply never came, because the reader's `portal/nodeops-{meshId}` address is not claimed in the cluster's grain directory.

## Disposition

This makes #2901 **fixed-and-undeployed**, not open-and-unexplained — which is a different action for whoever picks it up. It is the same root as this morning's observation that both portals are behind and were not self-updating; the CD stalls fixed today (#3075, #3078, #3080) are what let a fix reach them at all.

Doc-only. `MeshWeaver.Documentation.Test`: **189 passed, 0 failed** (full project, unfiltered, including `DocumentationLinkIntegrityTest`).

### DIFF
--- a/src/MeshWeaver.Documentation/Data/Architecture/PodHubClaimReassertion.md
+++ b/src/MeshWeaver.Documentation/Data/Architecture/PodHubClaimReassertion.md
@@ -59,6 +59,69 @@ Put those together and a mapping that is lost once is lost for the life of the p
 The startup population and the lost-after-landing population are two different triggers of one
 cause. Neither recovers, for the same reason: **there is no second assertion.**
 
+## The user-visible face of it: `/api/content` answers 503 (#2901)
+
+The measurement above is stated in routing terms. This is what the same fault looks like from a
+browser, and it is worth writing down because the shape was mis-attributed twice before the routing
+cause was found — once to a post-upload *settle window*, once to the memory `PubSubStore` and the
+streams-retirement program (#1729/#1742).
+
+`/api/content/{node}/{collection}/{file}` resolves the owning node's collection config with **one
+`GetDataRequest` per request** — deliberately uncacheable, because that round trip IS the permission
+check (`ContentFileResolver.Resolve`). It is issued from `portal/nodeops-{meshId}` and bounded by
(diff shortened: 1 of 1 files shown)


rbuergi:
Closing as a duplicate of #3088, which is the same work and got there first (09:57 vs 10:00) and is already green.

My mistake: I checked for an existing PR with a `createdAt > 10:00` filter, which excluded #3088 by three minutes, and concluded the agent had opened none. The work was already pushed; I just could not see it through my own query.

#3088 is the one to take — same doc page, same measurement, currently draft and CLEAN.

Issue #3094, opened 2026-09-02
AutocompleteRequest answers a 150 ms settle-window snapshot — under load the cross-partition rows miss it, and the orchestrator treats the truncated Nearby batch as final


rbuergi:
## Measured

Plugins#1135 item 1 (`AutocompleteIntegrationTest.ChatAutocomplete_GlobalFanOut_ReachesOtherPartitions`) was attributed to "the query index had not caught up". That is falsified: with a precondition poll establishing that `IMeshService.Autocomplete("", "ACM")` already returns the `ACME` root, the very same failure reproduced locally on the first full run (`Release`, `DOTNET_PROCESSOR_COUNT=4`, core at `0a4334a6e`):

```
Expected the observable to emit a value matching the predicate within 30s because broadening should find ACME
from Systemorph context, but it completed without one. Last of 1 emission(s) was: CompletionBatch { Category = Nearby … }
```

Run alone (no parallel classes) the test passes in 2.1 s with `First ACME-bearing batch 'Nearby': 34 items` — and the same `EnrichWithNodeType … error overlay to 'Systemorph'` warning in both runs, so the overlay is not the discriminator either. The discriminator is **load**.

## Mechanism

`DataExtensions.HandleAutocompleteRequest` answers ONE snapshot of a progressively-filling merge:

```csharp
Observable.Amb(
    combined.Throttle(AutocompleteSettleWindow /* 150 ms */).Take(1),
    combined.Sample(AutocompleteAnswerDeadline /* 2 s */).Take(1),
    Observable.Timer(AutocompleteSilenceFloor /* 2.5 s */).Select(_ => Empty))
```

`combined` is `AutocompleteSnapshots.Combine` — every provider seeded with `Empty`, so it emits immediately and grows as each provider lands. The ACME row for a bare `@ACM` comes from `UnifiedReferenceAutocompleteProvider.GetTopLevelSuggestions` → `meshQuery.Autocomplete("", "ACM")`, itself a CombineLatest over the mesh query providers that emits partials. Whenever the fast in-memory providers have landed and the cross-partition rows arrive more than 150 ms after the previous emission, the throttle fires and the response is posted WITHOUT them. In the failing run the request reached `Systemorph` at 12:38:02.397 and the response left at 12:38:02.612 — the settle window, not the deadline.

[...]

Issue #3017, opened 2026-09-01
SilentReadNackTest rejects a FIFTH owner-side terminal — the access pipeline's shutdown refusal — while documenting that it must not pin which terminal won


rbuergi:
## The test says it must not pin which terminal won — and then pins it

`SilentReadNackTest.OwnerDisposedWithReadOutstanding_IsNacked_NotLeftHanging` failed on a terminal
that **is** owner-side, because `AnsweredByTheOwner` enumerates four owner terminals and there is a
fifth.

```
MeshWeaver.Reactive.Assertions.AssertionException :
Expected value to be true because the answer must come from the OWNER — its handler or its own
intake — never from the routing layer, which is the discrimination this test exists for. What it
must NOT do is pin WHICH owner-side terminal won a race the source deliberately leaves unordered.
Got: Hub TestData/silent-read is shutting down (its lifetime scope is gone,
ObjectDisposedException) — cannot evaluate access for GetDataRequest; the address may reactivate
(recycle / restart). Rejecting now.
```

The message is `AccessControlPipeline`'s shutdown refusal
(`src/MeshWeaver.Hosting/Security/AccessControlPipeline.cs`, the *"cannot evaluate access for
{messageTypeName}"* pair). That pipeline runs **inside the owner hub** — it is the owner's own
intake, not the routing layer — so the invariant the test exists for (*"a terminal answer arrives,
from the owner"*) held. Only the predicate disagreed.

## Why it slips through today

`SilentReadNackTest.AnsweredByTheOwner` (line 213) accepts exactly two shapes:

```csharp
   (message.Contains($"GetDataRequest({new MeshNodeReference()}) at '{path}'")
    && message.Contains("Retry against the fresh activation."))          // NackSilentRead's arms
|| (message.Contains($"Hub {path} is shutting down")
    && message.Contains("retry to get the authoritative answer."))       // the hub's own intake
```

The observed message satisfies the **first half** of the second arm (`Hub {path} is shutting down`)
and fails the second: it ends *"Rejecting now."*, because it is refused by the access pipeline before
it reaches the hub's message intake. So it is rejected as "not from the owner".

🚨 The class's own guard test — `AnsweredByTheOwner_AcceptsEveryOwnerTerminal_AndRefusesRoutings` —
**passes**, because it asserts the four terminals somebody thought of. It cannot discover a fifth.
That is the "a guard asserting config cannot see the outcome" shape: the widening it protects is
[...]

Issue #3008, opened 2026-09-01
A subscriber after delete+recreate can still get the tombstone's non-retryable 'will not reactivate' — after Created has already been published


rbuergi:
## A fresh subscriber after delete+recreate gets the TOMBSTONE, not the recreated node

Found while enumerating the 48 suites behind MeshWeaver.Plugins' `Build + test the portal hosts`
(Plugins#1119). **Not on any existing list**, and it is in `Monolith.Test` — the suite that only
started running in CI today, so it has had almost no exposure.

```
MeshWeaver.Hosting.Monolith.Test.WorkspaceCacheEvictionTest.NewSubscriber_AfterRecreate_GetsFreshSnapshot [FAIL]
  MeshWeaver.Reactive.Assertions.ObservableAssertionException :
  Expected the observable to emit a value within 10s, but it errored:
  No node found at 'TestData/cache-recreate' — the node was deleted,
  so this address will not reactivate.
```

## Why this one is worth a look rather than a re-run

The test does exactly what the name says, and it does **not** race the write — it waits on positive
signals throughout (`MeshWeaver.Hosting.Monolith.Test/WorkspaceCacheEvictionTest.cs`):

1. create `TestData/cache-recreate` (`Name = "First"`), warm the cache with a subscriber;
2. subscribe to `IMeshChangeFeed` **before** mutating;
3. delete, then **wait for the `Deleted` event** to have fanned out;
4. recreate (`Name = "Second"`), then **wait for the `Created` event**;
5. only then, a **new** client subscribes and must see `"Second"`.

So by step 5 the node exists and the platform has already published `Created` for it. The new
subscriber still received the delete tombstone's terminal error — and that error carries the
**non-retryable** classification (*"the node was deleted, so this address will not reactivate"*),
which terminates the stream rather than being ridden out the way a `ShuttingDown` NACK is. A caller
that follows the documented contract will treat it as authoritative and stop.

That makes it a correctness question, not just a flake: **for how long after a recreate can a reader
be told, authoritatively, that the path is gone forever?** In production the same shape is
delete-then-recreate at a path (a re-import, a Repair migration, a node moved back), and the reader
that hits the window gets a terminal answer it is designed not to retry.

## Frequency, honestly

Seen **once in three full runs** of `Monolith.Test` on the same tree:

| run | result |
|---|---|
| full 48-suite pass #1 | passed (the suite's only red was the `NodeTypePathRegistrationTest` contract conflict) |
[...]

PR #3106, opened 2026-09-02
fix(Plugins#1135): Skip/Limit paged over an order that did not exist

rbuergi:
## Outcome: **reproduced under load, cause identified, fixed** — red-then-green under identical conditions.

`MeshWeaver.Plugins#1135` item 7 — `HierarchicalBrowsingSuite` *"Skip and Limit paginate without
overlap"* — was filed rather than patched because a race nobody can reproduce cannot be falsified.
It reproduces on demand under CPU load, and what it exposes is not a test problem.

## The recorded mechanism does not survive the failure's own shape

The issue proposed that the underlying set GREW between the page queries (core's new
`ContentTypeRegistrationSweep` writing activity nodes). It cannot have: the case checks the page
counts BEFORE the overlap, `MeshTestContext.Expect` throws on the first false check, and
`page1=3 / page2=3 / page3=1` all passed. Seven rows at every page, one of them served twice —
that is a **re-ordering**, not a growth. (The sweep is also outside the query: it is scoped to the
case's own subtree and filtered to `nodeType:Markdown`.)

## Skip/Limit was paging over an order that did not exist

Neither of the two clip sites defined one.

* **`StorageAdapterMeshQueryProvider.RunQueryNodes`** left `sorted = matchedNodes` for any query
  with no `sort:` and no free text — then clipped to the load cap, `Skip + Limit`.
* **`MeshQuery.ClipMergedInitial`** sorted by score and named *"insertion order as the final
  tiebreaker"* before applying `Skip`/`Take`.

Insertion order is not an order, and the store is not what makes it vary. The scope walk reads
every path with `SelectMany(path => persistence.Read(path))` — a **merge** — so rows arrive in the
order the pooled reads COMPLETE. Idle, that is the walk order and three successive page queries
agree. Under CPU pressure the reads interleave differently each time, so page 2 skips past a row
page 1 never served and re-serves one it did.

Both sites needed it: the provider clips to `Skip + Limit` **before the merge sees a row**, so a
total order in the merge alone would still be paging over three different subsets.

## The fix

Every ordering branch now ends in `Path` ascending, ordinal — the one field that is present,
unique, and stable across the three independent queries a caller issues for three pages. An
explicit `sort:` still wins; relevance still wins for a free-text query; only the ties, which were
arbitrary, are now decided. `RunQueryNodes` gains the `else` branch that was missing entirely.

## Measurements (this worktree, Release, 18-core M-series)

| tree | conditions | failures |
|---|---|---|
| unfixed | idle (machine quiet, load avg ~2) | **0 / 60** |
| unfixed | no burners of mine, machine busy at load avg 60–100 | **2 / 60** |
| unfixed | 36 CPU burners | **11 / 120** |
| **fixed** | 36 CPU burners (load avg ~140) | **0 / 120** |

Monotone in load, and zero at both ends for the two reasons that matter: idle it does not happen,
and with the total order it does not happen under the load that made it happen 11 times in 120.

## The committed test asserts the CAUSE, not the race

`PagedQueryTotalOrderTest` has **no timing, no repetition and no load** in it. Both cases drive a
source that reorders DETERMINISTICALLY — a storage walk rotated one position per call, and two
providers whose canned rows arrive forward and reversed — which is the same input the runner
produces by accident. Neither can flake, and neither can go green by getting lucky.

* unfixed → **RED**: `Expected collection to have 7 item(s) … but found 6` (the overlap), and the
  merge case's rows come back out of order;
* fixed → **GREEN**.

Nothing was relaxed: no timeout widened, no retry, no assertion softened, no allow-file touched.

## Suites run locally on the fixed tree

Core (`-c Release -warnaserror`):

* `MeshWeaver.Hosting.Test` — **300/300** (including the two new cases)
* `MeshWeaver.Graph.Test` — **1205/1205**

MeshWeaver.Plugins' moved suites, built against this branch with `-p:MeshWeaverRoot`:

| suite | result |
|---|---|
| `PathResolution.Test` | 168/168 |
| `Persistence.Test` | 1213/1213 |
| `Query.Test` | 354/354 † |
| `NodeOperations.Test` | 76/76 |
| `Search.Test` | 5/5 |
| `Acme.Test` | 65/65 (includes Plugins#1135 item 8, the ColdCycle case) |
| `Content.Test` | 142/142 |
| `Hosting.Monolith.Test` | 786/790, 2 skipped, **2 failed** ‡ |
| `Autocomplete.Test` | 145/146, **1 failed** ‡ |

† **with MeshWeaver.Plugins#1183**, which lands first. That case
(`GlobalSearchPartitionTest.GlobalSearch_WithAccessAssignment_ReturnsGrantedNodes`) asserted through
`scope:descendants limit:50` with **no `sort:`** — a fifty-row window over the very order this
change defines. It was passing on luck (the mesh it shares holds more than fifty rows by the time it
runs), and it now asks for `sort:LastModified-desc` like its sibling in the same class. Verified
green **both** with and without this change, so the two land in either order without a red minute on
either trunk. That is the one behavioural consequence of this change and it is deliberate: a clip
over an undefined order is a sample, not a window.

‡ **Pre-existing, measured on the control tree** (this branch with the two source files reverted to
`main`, everything else identical):

* `LateNackReenqueueTest.LateOwnerDisposingNack_AfterOptimisticEmit_ReenqueuesAndLands` and
  `NackReachesTheWaiterDuringTeardownTest.OwnerDisposingUnderMeshTeardown_StillAnswersTheWaitingCaller`
  — **2/2 fail on the control** as well. Teardown-NACK races, the family already tracked as #3008 /
  #3017; nothing in them reaches a query.
* `AutocompleteIntegrationTest.ChatAutocomplete_GlobalFanOut_ReachesOtherPartitions` — Plugins#1135
  item 6, filed as **#3094**: a 150 ms settle window in `HandleAutocompleteRequest` answering a
  partially-filled merge. It failed here in 412 ms with the recorded signature, i.e. the settle
  window, not a clip.

## Relationship to MeshWeaver.Plugins#1179

That PR (open, in the Plugins repo) reaches the same diagnosis from the other side and fixes it in
the TEST, by adding `sort:path` to the paging query. The two are complementary and do not conflict
— with an explicit `sort:` the provider takes its `OrderBy` branch and this change is a no-op for
that query. This one closes the same hole for every OTHER caller: `MeshQueryRequest.Skip` is a
public paging surface, and the provider's load cap already clipped a plain `limit:N` to an
arbitrary N rows, which is the shape the code's own comment records as the earlier
`SearchQueryTests.TextSearch_CaseInsensitive` CI flake.

Refs Systemorph/MeshWeaver.Plugins#1135

### DIFF
--- a/src/MeshWeaver.Hosting/Persistence/Query/MeshQuery.cs
+++ b/src/MeshWeaver.Hosting/Persistence/Query/MeshQuery.cs
@@ -760,6 +760,17 @@ public class MeshQuery : IMeshQueryCore
     }
 
     /// <summary>
+    /// The FINAL sort dimension of the merged result set: <see cref="MeshNode.Path"/>, ascending
+    /// and ordinal. It is what turns the score ordering into a TOTAL order, so that the
+    /// <see cref="MeshQueryRequest.Skip"/>/<see cref="MeshQueryRequest.Limit"/> clip below is a
+    /// partition of the result set rather than three independent samples of an arbitrary one.
+    /// A non-<see cref="MeshNode"/> hit sorts under the empty key and so keeps its arrival order
+    /// among its peers — LINQ's sorts are stable.
+    /// </summary>
+    private static string PathTiebreak<T>((T Item, double Score) hit)
+        => hit.Item is MeshNode node ? node.Path ?? "" : "";
+
+    /// <summary>
     /// Sort + skip + clip the merged initial set. The authoritative ordering
     /// pass for every multi-provider Initial emission. Mirrors the post-collect
     /// pipeline that <c>StorageAdapterMeshQueryProvider.RunQueryNodes</c> runs per-provider.
@@ -782,9 +793,12 @@ public class MeshQuery : IMeshQueryCore
     ///     scores and hands them here. Higher score = stronger match.
     ///     See <see cref="QueryResultChange{T}.Scores"/> for the per-provider
     ///     scoring conventions.</item>
-    ///   <item>Insertion order as the final tiebreaker — preserves the
-    ///     provider's own deterministic ordering for two items at the same
-    ///     score (e.g. two PG rows tied on the prefix bonus).</item>
+    ///   <item><b>Path ascending</b> as the FINAL tiebreaker
+    ///     (<see cref="PathTiebreak{T}"/>) — the dimension that makes this a TOTAL order, which
+    ///     is what <see cref="MeshQueryRequest.Skip"/> needs to be paging rather than sampling.
+    ///     It replaced "insertion order", which reads as a tiebreaker and is not one: nothing
+    ///     orders two providers' Initial emissions against each other, and even a single
+    ///     provider's scope walk emits in read-COMPLETION order.</item>
     /// </list>
     ///
     /// <para><b>Why score sort lives here, not in each provider.</b> A single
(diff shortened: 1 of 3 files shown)


rbuergi:
🚦 **Auto-merge deliberately disarmed until Systemorph/MeshWeaver.Plugins#1183 has merged.**

The Plugins gates check core out **unpinned at `main`**, and `Build + test the portal hosts` is one
of that repo's three required contexts. This change makes one clip in `Query.Test` deterministic
(`GlobalSearch_WithAccessAssignment_ReturnsGrantedNodes`, which asserted through a fifty-row window
with no `sort:`), so landing it first would redden Plugins `main` for every open PR there until
#1183 catches up. #1183 is green on `main` with and without this change, so the safe order is
strictly: **#1183, then this**.

Re-arm with `gh pr merge 3106 --merge --auto` once #1183 shows merged.

PR #3131, opened 2026-09-02
fix: a build grant committed after its candidate stood down no longer wedges the build

rbuergi:
## The defect

`BuildNodeType.ArbitrateDurably` runs one arbitration pass as four steps:

1. read the candidate set (`RequestedClaims`) off **this cluster's mirror**,
2. read the claim **LOCK** (`Admin/Build/_Claim`),
3. commit the grant with `WriteIfVersion` (the compare-and-set that makes it exclusive across clusters),
4. publish the grant onto the Build node so `ObserveBuildClaim` emits.

**Two storage round-trips separate step 1 from step 4, and a candidate is free to stand down inside that window.** A follower that sees the GO calls `WithdrawBuildClaim`, whose two halves are *both* conditional on state this pass is about to change:

- it removes its registration — one the pass captured in step 1;
- it then reads the lock to release it, and finds no lock of its own, because step 3 has not happened yet.

Conditional writes make each half safe on its own. **They cannot order two writers.** The pass then commits the build to a process that has already finished with it and stopped listening.

Nothing converges afterwards, and which record is left behind depends on the interleaving:

| interleaving | debris |
|---|---|
| the withdrawal's mirror write lands first | the **mirror's claim projection** names the withdrawn candidate at `Planning` — and on a `GrantOnMirror` host (no durable store, or the fail-open path) that projection **is** the claim |
| the withdrawal's lock read lands before the CAS | the **durable lock** names it — the one witness every cluster's arbiter decides on |

Either way `HolderStillHoldsIt` then defends it **by design**, because the process really is alive (#1355 — a stopped heartbeat on a running process means busy, not dead). With cluster membership the claim is never taken over at all; without it, every later candidate waits out `ClaimStaleAfter` and is handed the same dead claim again. **No builder elected, no NodeType baked, no pod ready** — the readiness stall #1440 fixed from the candidate's side, arriving instead through the arbiter's own commit.

The doc asserted the opposite (`BuildCoordination.md`: *"both halves are conditional, so they are safe against a concurrent grant"*). That sentence is corrected here.

## Evidence

Reproduced from `MeshWeaver.Hosting.Monolith.Test.BuildCoordinationTest.Follower_StandsDown_SoTheNextBuildCanStillBeClaimed` (MeshWeaver.Plugins), which reddens the required `Build + test the portal hosts` gate — Systemorph/MeshWeaver.Plugins#1193.

**Control first: this is not a load flake.**

| condition | result |
|---|---|
| idle, unfixed | **2 failures / 15 runs** |
| idle, unfixed (second batch, with a state dump) | **1 failure / 10 runs** |
| idle, fixed | **0 failures / 25 runs** |

The state dumped at a failure — the mechanism, not an inference:

```
MIRROR claimedBy=Rolands-MacBook-Pro/3283624a…  status=Planning  requested=[]  ready=[fp-stand-down]
LOCK   present=False
```

A build node locked, at `Planning`, to a holder whose driver had already completed, with the GO recorded and no candidate queued. The CI signature matches exactly: `TimeoutException` at 32.1 s (`Admin/Build` never becomes free), on shard 2, per-method evidence log.

## The fix

The arbiter re-checks at the one point that can still see the stand-down.

- **`ApplyGrant` refuses** to publish a grant whose winner is *neither the holder the mirror already names nor still a candidate*. It runs inside the node's own serialised write lambda, so the check and the publication are one step — no residual window.
- **A refused publication hands the lock straight back** (`HandBackAStoodDownGrant`). The CAS already took it, and the arbiter is the lock's **only** writer, so undoing its own commit is the only place this can be closed. The delete rides the store's change feed, which is the arbiter's own trigger, so a queued candidate is elected immediately rather than at the 2-minute tick.

🚨 **The guard's first clause is load-bearing.** A holder *mid-bake* has no registration left either — its own grant consumed it — so a bare "is it still queued?" test would read a running bake as a candidate that stood down and pull the lock out from under it. `AHolderMidBake_IsNotMistakenForACandidateThatStoodDown` pins that.

Both writes stay conditional, like every other write to the lock: a pass that lost a later race removes nothing.

## Tests

`test/MeshWeaver.Graph.Test/BuildGrantPublicationTest.cs` — 7 cases, **no wall-clock and no concurrency anywhere**: the interleaving is *constructed*, not raced for, in the same family as the existing `BuildNodeType.Arbitrate` tests.

- refusal for a candidate that stood down (**fails on `main`**: verified by deleting the guard — 1 failed / 6 passed);
- publication for a candidate still registered, consuming only its own registration;
- the mid-bake holder guard;
- a redundant pass writes nothing;
- the lock is released on refusal, kept on publication, and **left alone when it has already moved on**.

`Fixes` is deliberately omitted: the plugins-side PR closes Systemorph/MeshWeaver.Plugins#1193, because it also adds the assertion this failure could otherwise hide from — the durable **lock** being free, not just the projection.

Pairs-with: none — no public type is removed; two private members become `internal` for the tests.

### DIFF
--- a/src/MeshWeaver.Documentation/Data/Architecture/BuildCoordination.md
+++ b/src/MeshWeaver.Documentation/Data/Architecture/BuildCoordination.md
@@ -306,7 +306,26 @@ never run — holding the claim at `Planning`, never heartbeating, and never tak
 process is alive and the takeover rule below defends a live holder by design (a stopped heartbeat on
 a live process means busy, not dead). The *next* image's fingerprint could then never be claimed and never get a GO, holding
 every silo's readiness down. `WithdrawBuildClaim` removes the registration and gives back a claim
-that raced it; both halves are conditional, so they are safe against a concurrent grant.
+that raced it.
+
+🚨 **Standing down cannot be made safe from the candidate's side alone (#1193).** Both halves of
+`WithdrawBuildClaim` are conditional on state the arbiter is about to change, and an arbitration
+pass spans two storage round-trips: it reads the candidate set off the mirror, reads the lock,
+commits the compare-and-set, and only then publishes the grant. A follower that stands down inside
+that window removes a registration the pass has already captured, and finds no lock of its own to
+release — so the pass commits a grant to a process that has finished with the build. **Conditional
+writes make each half safe on its own; they cannot order two writers.** Depending on the
+interleaving the debris is either the mirror's claim projection or the durable lock, and the
+takeover rule then defends it *because the process really is alive*. Measured on
+`BuildCoordinationTest.Follower_StandsDown_SoTheNextBuildCanStillBeClaimed`: 2 failures in 15 runs
+on an **idle** machine, the build node left at `ClaimedBy=<the follower>, Status=Planning`.
+
+So the arbiter re-checks at the point where it can: **`ApplyGrant` refuses to publish a grant whose
+winner is neither the holder the mirror already names nor still a candidate**, and a refused
+publication **hands the lock straight back** (`HandBackAStoodDownGrant`) — the arbiter is the lock's
+only writer, so undoing its own commit is the only place this can be closed. The guard's first
+clause is load-bearing: a holder *mid-bake* has no registration left either, its own grant having
+consumed it.
 
 There is deliberately **no timer and no bound bolted onto the wait**. A bound would end the wait by
 guessing, and a follower that guesses "the build finished" certifies a share it never probed — a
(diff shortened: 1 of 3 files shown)


rbuergi:
Marking draft: the body's "0 failures / 25 runs" line is **wrong** and I am retracting it. A longer run against this change measured **2 failures in 13** — the same rate as before it. The refusal + hand-back closes one interleaving (its unit tests fail on `main`), but it is not the whole of #1193. Investigating the residual now; the body will be corrected with the real numbers before this is ready.

PR #3152, opened 2026-09-02
fix(gitsync): a green build of the default branch is a publish signal however it was triggered

rbuergi:
## What

The `workflow_run` webhook gate tested `event == "push"` — **one value, not a set** — and so
discarded green builds that *were* the default branch's own tree. It becomes an **allow-list** of
triggers that mean "a build of the default branch's own tree", with unknown event names refused.

Origin: investigation of Systemorph/MeshWeaver.Plugins#1194 (no closing keyword — that issue's
*primary* cause was a genuinely red push lane on `MeshWeaver.Reinsurance`, where this gate behaved
correctly and refused to ship a red tree. This PR fixes the **secondary**, separate defect.)

## The measurement (2026-09-02)

`MeshWeaver.Reinsurance`'s `main` built **green three times at `636ebd5`** — 11:17, 12:27, 12:55Z —
every one with `event: repository_dispatch`, from the release-follow lane, which rebuilds every
module against a new platform pin **without a commit to push**. All three were discarded by the
`runEvent != "push"` early return.

`Underwriting/_GitSync` on `memex.systemorph.com` then sat **38 hours** behind a merged main — hook
664846394 armed and healthy, every delivery 200 OK, nothing red anywhere. There is deliberately no
scheduled poll behind this, so a dropped publish signal has no symptom except content that quietly
stops arriving.

## The gate now

Two independent guards, both fail-closed:

| Trigger | Admitted | Why |
|---|---|---|
| `push` | ✅ | The branch moved and its CI ran — the original case. |
| `repository_dispatch` | ✅ | GitHub only ever runs a dispatched workflow from the **default branch**; `head_sha` is that branch's tip. This is how a platform release re-verifies every satellite repo. **The bug.** |
| `schedule` | ✅ | Same — a cron run only ever exists on the default branch. |
| `workflow_dispatch` | ✅ | May target any ref, so the head_branch guard discriminates. On the default branch it is a manual re-verification, and the only recovery lever when a merge burst cancelled the push run. |
| `pull_request` / `pull_request_target` | ❌ | Green **unmerged** code. Both can report `head_branch=main`, so the *trigger* guard is what rejects them. |
| `dynamic` | ❌ | GitHub's Copilot reviewer — completes green on the default branch, not a build at all. |
| anything unknown | ❌ | Fail closed: an allow-list means the next trigger GitHub invents cannot publish by accident. |

**`head_branch == default_branch` is unchanged** and stays the second, independent guard.

### Why `merge_group` is deliberately absent

A merge-queue run's `head_branch` is the temporary
`gh-readonly-queue/{base}/pr-{n}-{sha}` ref, so the branch guard already rejects it. Measured on this
repository's own queue (2026-09-02):

```
event: merge_group
head_branch: gh-readonly-queue/main/pr-3143-508c1ee373c66aba2999622fa21c395ffc0a9480
```

Listing it would be a line no test could ever reach. The test asserts the mechanism that rejects it
instead.

### Widening cannot cause churn

A sync source already sitting on the built sha is skipped by `SkipReason` ("already at this commit"),
so a scheduled or dispatched re-verification of an unchanged default branch triggers **no import at
all**.

## Tests

`test/MeshWeaver.Documentation.Test/GreenBuildPublishSignalTest.cs` — 22 cases. The end-to-end suite
(`MeshWeaver.GitSync.Test`) lives in MeshWeaver.Plugins since the boots-a-mesh split and pins the
*wiring*; this policy is a pure function of a payload, so it is pinned here next to the code. Both
guards are extracted as pure statics (`IsPublishSignalTrigger`, `IsDefaultBranchBuild`) reached via a
new `InternalsVisibleTo`.

The table is stated in **both directions** — a test that only listed the admitted triggers would go
green against a gate that admits everything — plus a **control arm** asserting `ProcessWorkflowRun`
still calls both guards, so a refactor that drops one cannot leave every other assertion green.

**Red-then-green, run locally:**

- allow-list narrowed back to `{"push"}` → `Failed: 4, Passed: 18` (`repository_dispatch`,
  `schedule`, `workflow_dispatch`, case-insensitivity)
- with the fix → `Failed: 0, Passed: 22`
- full project → `Failed: 0, Passed: 233` (includes `DocumentationLinkIntegrityTest` and
  `WhatsNewEntryIntegrityTest`)

Release + `-warnaserror` builds clean for `MeshWeaver.GitSync`, `MeshWeaver.Documentation.Test`,
`MeshWeaver.PluginCatalog` and `Memex.Portal.Shared`.

## Docs

- `Doc/Architecture/PluginUpdateOnGreenBuild` — new section *Which green runs count as a publish
  signal* (the decision table + the 2026-09-02 incident); the failure-modes row no longer claims
  "not triggered by a push".
- What's New entry (`Category: Fix`): *A rebuilt repository syncs again*.

## Follow-up (not in this PR)

`MeshWeaver.Plugins/src/MeshWeaver.GitSync.Test/GitHubWebhookProcessorTest.cs` still passes
unchanged (every one of its assertions covers behaviour this PR preserves), but its docstring for
`Process_WorkflowRun_RecordsOnlyDefaultBranchPushTriggeredGreenBuilds` says "push-triggered" and
would be worth widening with a `repository_dispatch` case in a satellite PR.

Pairs-with: none — no public top-level type is removed.

### DIFF
--- a/src/MeshWeaver.Documentation/Data/Architecture/PluginUpdateOnGreenBuild.md
+++ b/src/MeshWeaver.Documentation/Data/Architecture/PluginUpdateOnGreenBuild.md
@@ -31,6 +31,47 @@ per installed module:  ModuleVersion changed?
    └── yes ──▶ opted in ? install the delta : raise an "Update available" reminder
 ```
 
+## 🚨 Which green runs count as a publish signal
+
+`workflow_run` fires for far more than a repo's content CI, so the webhook applies **two independent
+guards** before a delivery becomes a `BuildCompletion` (`GitHubWebhookProcessor.ProcessWorkflowRun`):
+
+1. **The trigger must be on the allow-list** — `push`, `repository_dispatch`, `schedule`,
+   `workflow_dispatch`. Each of those means *a build of the default branch's own tree*.
+2. **`head_branch` must BE the repository's default branch.**
+
+Both fail closed: an unknown trigger is refused, and a payload whose branch cannot be read records
+nothing.
+
+| Trigger | Admitted | Why |
+|---|---|---|
+| `push` | ✅ | The branch moved and its CI ran — the original case. |
+| `repository_dispatch` | ✅ | GitHub only ever runs a dispatched workflow from the **default branch**, and `head_sha` is that branch's tip. This is how a platform release re-verifies every satellite repo: no commit to push, same tree, a genuine green verdict on it. |
+| `schedule` | ✅ | Same — a cron run only ever exists on the default branch. |
+| `workflow_dispatch` | ✅ | May target any ref, so guard 2 does the discriminating. On the default branch it is a manual re-verification of that tree, and the only recovery lever when a merge burst cancelled the push-triggered run. |
+| `pull_request` / `pull_request_target` | ❌ | Green **unmerged** code. Note both can report `head_branch=main`, so guard 1 — not guard 2 — is what rejects them. |
+| `dynamic` | ❌ | GitHub's Copilot reviewer. Completes green on the default branch and is not a build at all. |
+| `merge_group` | ❌ | A merge-queue run's `head_branch` is the temporary `gh-readonly-queue/{base}/pr-{n}-{sha}` ref, so guard 2 already rejects it. Listing it would be unreachable. |
+| anything else | ❌ | Fail closed. An allow-list means the next trigger GitHub invents does not publish by accident. |
+
+**Widening the list cannot cause churn.** A sync source already sitting on the built sha is skipped
+("already at this commit"), so a scheduled or dispatched re-verification of an unchanged default
+branch triggers no import at all.
+
+### 🚨 The single-value test that dropped real signals (2026-09-02)
+
+Guard 1 was `event == "push"` — one value, not a set — and it discarded green builds that *were* the
+default branch's tree. `Systemorph/MeshWeaver.Reinsurance`'s `main` built green three times at
+`636ebd5` that day (11:17, 12:27, 12:55Z), every one with `event=repository_dispatch` from the
+release-follow lane, which rebuilds every module against a new platform pin **without a commit to
+push**. All three were dropped, and `Underwriting/_GitSync` on `memex.systemorph.com` sat 38 hours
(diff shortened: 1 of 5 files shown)


rbuergi:
Held as a **draft on purpose**, not because it is unfinished: this repo's `auto-arm.yml` arms every non-draft PR for the merge queue, and the session that produced this change was asked not to merge it or arm auto-merge. Everything is done and verified — mark it ready for review to let it land normally.

rbuergi:
Marking ready. I reviewed the diff before releasing it, because this changes what every mesh in the fleet accepts as a publish signal.

Three things make it safe to land rather than to sit on:

1. **It is an allow-list, so it fails closed.** An event name nobody has considered is refused, not admitted. A deny-list here would fail open on whatever trigger GitHub invents next.
2. **Both guards survive.** The trigger check states the requirement; the `head_branch == default_branch` check discriminates the one admitted trigger (`workflow_dispatch`) that can target any ref. Neither subsumes the other, and the control arm asserting `ProcessWorkflowRun` still calls both is the right way to keep a refactor from quietly dropping one.
3. **Widening cannot cause churn**, which was my main worry. A sync source already sitting on the built sha is skipped by `SkipReason`, so a scheduled or dispatched re-verification of an unchanged default branch imports nothing.

The `merge_group` conclusion is right and the evidence is the good kind: a queue run's `head_branch` is `gh-readonly-queue/{base}/pr-{n}-{sha}`, so an allow-list entry would have been unreachable by any test.

**Follow-up, deliberately not in this PR and sequenced after it:** `MeshWeaver.Plugins/src/MeshWeaver.GitSync.Test/GitHubWebhookProcessorTest.cs` still describes the rule as "push-triggered" and has no `repository_dispatch` case. Adding one before this lands would fail, since it asserts behaviour this PR introduces. It should follow, not accompany.

Context on the origin: Systemorph/MeshWeaver.Plugins#1194. That incident had two causes and this is the second one. The first — a genuinely red push lane on MeshWeaver.Reinsurance `main` — is where this gate behaved correctly by refusing to ship a red tree, and it is not addressed here.

PR #3157, opened 2026-09-02
ci: take a module's own test suite off its consumers' critical path

rbuergi:
## The measurement

A pull request in `MeshWeaver.Plugins` takes **50 minutes median**; core takes 13. The critical
path is **not** the test shards. Traced on Plugins run 33656010754, a 39-minute run:

```
 0.0 ->  2   Module bundles (floor) / Warm the shared build environment (once)
 2    ->  8   Module bundles (floor) / Build the workspace (one global build)
19.1 -> 32.5  Module bundles (floor) / Module bundle (MeshWeaver.AI)      <-- 13.4 min
32.5 -> 32.7  Module bundles (floor) / All selected bundles built
32.8 -> 37.0  Compile every NodeType (vs core)          (REQUIRED)
32.9 -> 38.6  test-repos / Compile + render node repos  (the Tests-area gate)
38.9 -> 39.0  Every gate executed
```

Portal-host shards finished at 30.9 min, in parallel, off the critical path.

Step breakdown of that 13.4-minute `Module bundle (MeshWeaver.AI)` job:

```
12.7 min  Run the module's tests, when it ships any
 0.2 min  Build the module — and say which compiler produced it
 0.1 min  Set up job
```

**The bundle artifact is ready after 0.2 min.** The job then spends 12.7 minutes running a suite
its consumers do not read: `compile-check` and the Tests-area gate `needs:` the floor call and
consume only `module-bundle-MeshWeaver.{AI,Markdown.Collaboration,Maps}` — but a `needs:` on a
`uses:` job waits for the **whole** called workflow. The floor call exists precisely so those
gates do not wait for the other 30 bundles (Plugins#892); the suite put the wait straight back.

### It costs a second thing: the recovery loop

A run cannot be re-run until it fully completes — `gh run rerun --failed` answers *"cannot be
rerun; This workflow is already running"*. Measured 2026-09-02 on four Plugins PRs whose required
`Build + test the portal hosts` context had **already** gone red:

```
run 33676248730 (PR #1214)  status=queued
  queued  Module bundles (floor) / Module bundle (MeshWeaver.AI)
  queued  Module bundles (floor) / Module bundle (MeshWeaver.Markdown.Collaboration)

run 33675701470 (PR #1200)  status=queued
  in_progress  Module bundles (floor) / Module bundle (MeshWeaver.AI)
  queued       Module bundles / All selected bundles built
```

So after a shard flakes, the fix is blocked for a further ~13 minutes by a suite whose verdict
nobody is waiting on. That is the compounding half of the cost: every flake pays it.

## The shape

`publish` — the input that already means *trunk* in this lane's caller contract — decides where
the suite runs. No new input; every caller inherits it with no change.

| `publish` | where the suite runs | what the position buys |
|---|---|---|
| `true` | **inline in `pack`**, after the bundle upload, before the hand-over (unchanged, byte for byte) | *a failing suite never publishes* — the registry serves what every installation reads |
| `false` | the new **`tests` job**, in parallel with select → prepare → build-workspace → pack | nothing — see below |

```
select ─┬─► prepare (ONCE) ──► build (ONE workspace) ──► pack ─┬─► verify
        │                                                      │
        └─► tests  (a lane of its own — publish: false) ────────┘
```

The `tests` job `needs: select` **only**: the suite is a `dotnet test` against the content
checkout with `-p:MeshWeaverRoot=<the platform checkout>`; it consumes no platform image, no
workspace build and no bundle. Depending on any of them would put it back on the chain it exists
to run beside.

## Pull request vs trunk push — what I concluded

**Your reading holds, and the workflow proves it rather than suggesting it.** The hand-over step's
own condition is `if: ${{ inputs.publish && steps.plan.outputs.need_publish == 'true' }}`. On a
run with `publish: false` that step **cannot fire at all**, so "the bundle was built before its
suite passed" has no consequence: there is nothing for the ordering to protect, and the ordering
instead costs every consumer of the call the whole suite. On a publishing run the ordering
protects the registry, which every installation reads — so there the suite stays inline and the
`tests` job does not run.

Two refinements to the framing:

* **The switch is the INPUT, not the event.** Core's own `main-cd` calls this lane with
  `publish: false` for the bake's compose set (`plugins-modules`), and that call is not a pull
  request — it wants the fast path for exactly the same reason. Keying on `github.event_name`
  would have missed it and would have duplicated a meaning `publish` already carries.
* **Splitting the suite into a job that runs *after* pack would have bought nothing**, because a
  consumer's `needs:` waits for every job of the called workflow. The win comes from running the
  suite *beside* the pack chain, not merely outside the pack job.

## How a failing module suite still reds the PR

Three mechanisms, none of which is a comment:

1. **The `tests` job is part of this called workflow**, so a red suite fails the caller's `uses:`
   job (`Module bundles (floor)`) exactly as a red inline suite did.
2. **`verify` now `needs: [select, pack, tests]`** and goes RED when the tests lane did not
   succeed — under the same name it already has, `All selected bundles built`. That is the lane's
   one stable context and the one a repo's protection can require. The verifier's new
   `--tests-result` turns `failure` and `cancelled` into a named error.
3. **A suite that ran in NEITHER lane is caught by receipts, not by re-reading two conditions.**
   Two mutually exclusive `if:`s can both be false, and a suite that ran nowhere renders exactly
   like one that passed. So every pack receipt now records which lane owns its suite
   (`tests: none | inline | lane`), the tests lane drops a lane-stamped receipt of its own, and
   `node-repo-pack-verify.py` fails when a delegated module has no test receipt, when a test
   receipt arrives for a module whose pack receipt says `inline`, when a receipt is corrupt, or
   when the tests lane SKIPPED while modules delegated to it.

`bundles-built` is deliberately untouched — it still answers from the built markers dropped
*before* any suite, so a red suite still does not read as "bundle missing" to a gate that only
COMPOSES the bundle (#2710, Plugins#937). A new guard asserts that `bundles_built()` never learns
about tests.

## Status-check contexts

**No context is renamed or removed. No repo's branch protection needs updating.**

* `<caller job> / All selected bundles built` — unchanged name, unchanged meaning (still red on a
  red suite), now additionally asserting the tests lane.
* `<caller job> / Module bundle (<module>)` — unchanged.
* **New, and deliberately not requireable:** `<caller job> / Module tests (<module>)`. Like
  `Module bundle (<name>)`, it is a *dynamic* matrix context that appears and disappears with the
  diff, so it must never be added to branch protection — requiring one means requiring a context
  that will legitimately never report. That is exactly why `verify` carries the verdict too.

One honest note the maintainer should see: I measured Plugins' protection today —
`["Validate node repos", "Compile every NodeType (vs core)", "Build + test the portal hosts"]`,
`strict: false`. `Module bundles (floor) / All selected bundles built` is **not** among them, so a
red module suite did not block the merge button before this change either. This PR preserves the
redness exactly; whether to add that context to protection is a separate, system-changing call.

## Verification run locally

* `python3 -c "import yaml; yaml.safe_load(...)"` — the lane parses; jobs are
  `select, prepare, build-workspace, pack, tests, verify`.
* `node-repo-pack-verify.py --self-test` — green, now **11 tests-lane cases**: a failed lane, a
  cancelled lane, a skipped lane with delegated suites, a missing receipt, a corrupt receipt and
  an inline/lane disagreement each go RED and name the module; the publishing, nothing-owed,
  older-caller and pack-already-failed shapes stay green.
* `node-repo-scope.py --self-test` — 44 cases green (unchanged).
* `check-workflow-timeouts.py` — 65 jobs, 0 violations (the new job carries a literal
  `timeout-minutes: 45`).
* `dotnet build test/MeshWeaver.Documentation.Test -c Release -warnaserror` — 0 warnings, 0 errors.
* `ModuleSuiteLaneGuard` (new, 5 facts) + `ModuleBuildLedgerLaneGuard` (extended with the tests
  lane's Tested / test-phase-Failed records) — 11 passed. **Mutation-tested**: reverting
  `needs: [select, pack, tests]` and the inline step's `inputs.publish &&` turns 2 of the 5 new
  facts red; restoring them makes all 5 green again.
* `DocumentationLinkIntegrityTest` — green after the doc edit.

## Compatibility

The two new verifier arguments are optional, so a repo still pinned to an older
`node-repo-module-pack.yml` while resolving `platform-ref` to core's tip (which is what Plugins
does) keeps working: the old workflow simply does not pass them and the accounting stands down,
saying so. The doc — `Doc/Architecture/ModuleBuildArchitecture` — is updated in the same change;
its pipeline diagram already drew `tests` as a lane beside `pack`, which is only now true.

Pairs-with: none — no public type is removed by this diff.

### DIFF
--- a/.github/scripts/node-repo-pack-verify.py
+++ b/.github/scripts/node-repo-pack-verify.py
@@ -52,6 +52,18 @@ carries the bundle it attests and the closure evidence its build resolved, and a
 either is refused here rather than trusted because of where its step happened to sit in the job.
 🚨 None of this touches the module's TESTS: the marker is still dropped BEFORE them (#2710,
 Plugins#937) so a red suite cannot read as "bundle missing" to a gate that only composes.
+
+🚨 …BUT THE SUITE ITSELF IS ACCOUNTED FOR HERE TOO, because it no longer always runs in the pack
+job. On a run that PUBLISHES, the suite stays inline in `pack` so a red suite can never publish;
+on a run that does not, it runs in a `tests` job beside the pack matrix, so a caller whose gates
+consume only the bundle artifact does not wait 12.7 minutes for a verdict they never read
+(MeshWeaver.Plugins run 33656010754). Two mutually exclusive conditions can BOTH be false, and a
+suite that ran in neither lane renders exactly like one that passed — so each pack receipt records
+which lane OWNS its suite (`tests`: `none` | `inline` | `lane`), the tests lane drops a receipt of
+its own, and `--test-receipts` / `--tests-result` make the pairing checkable: a delegated module
+with no test receipt, a test receipt for a module that delegated nothing, and a tests lane that
+did not succeed are each RED, under this job's own unchanged context name. `bundles-built` is
+deliberately untouched by all of it.
 """
 from __future__ import annotations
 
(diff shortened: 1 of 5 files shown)


rbuergi:
The work here is **finished** — implemented, guarded and verified locally (self-tests, the workflow parse, the timeout gate, the new `ModuleSuiteLaneGuard` mutation-tested to prove it can fail). Marking it **ready for review** is all that is needed; it is a draft only so it cannot be armed into the merge queue unreviewed.

Two things worth a maintainer's eye before it lands:

1. **The publish-vs-PR distinction is the whole safety argument.** On a publishing run the suite stays inline in `pack`, before the hand-over, exactly as today — `a failing suite never publishes` is preserved bit for bit. Only a run with `publish: false` moves the suite beside the pack chain, and on such a run the hand-over step's own `if:` cannot fire at all.

2. **Measured, and not fixed by this PR:** `MeshWeaver.Plugins` `main` currently requires only `Validate node repos`, `Compile every NodeType (vs core)` and `Build + test the portal hosts` (`strict: false`). `Module bundles (floor) / All selected bundles built` is **not** required, so a red module suite did not block the merge button before this change either. This PR preserves that context's redness exactly; adding it to branch protection is a separate, system-changing call for you.

rbuergi:
Marking ready. I reviewed the risky part myself before releasing it, because this workflow is shared by six repos and the failure mode to fear is a red suite becoming invisible rather than a slow run.

Four things make it safe:

1. **The two conditions are exact complements on the same input.** The inline step is `inputs.publish && …`; the `tests` job is `!inputs.publish && …`. Keying both on `publish` rather than on the event name is the right call, and it matters more than it looks: core's own `main-cd` calls this lane with `publish: false` for the bake's compose set and is not a pull request, so it earns the fast path too.
2. **`verify` is `if: always()` with `needs: [select, pack, tests]`**, so a failed or cancelled tests lane still reaches it and still reds `All selected bundles built` under its existing name.
3. **The both-false trapdoor is closed by accounting, not by hope.** Two mutually exclusive conditions can both be false, which is exactly the shape this repo's rules warn about. Every pack receipt recording which lane owns its suite, plus a verifier that reds a delegated module with no test receipt, a disagreeing one, a corrupt one, or a skipped lane, is the right answer, and mutation-testing the eleven cases is what makes it believable.
4. **No context is renamed**, so no repo's branch protection needs touching. The new `Module tests (<module>)` is a dynamic matrix context and must never be required, like `Module bundle (<name>)`.

`needs: select` alone for the tests job is the whole win. Depending on the workspace build or the pack would have put the suite back on the chain it exists to run beside.

**On the finding about protection** — that `All selected bundles built` is not in Plugins' required contexts, so a red module suite did not block the merge button before this change either. That is worth raising separately and I am filing it rather than folding it in here. Editing branch protection is system-changing and belongs to the maintainer, and this PR preserving that context's redness exactly is the correct scope.

Measured motivation, for the record. On Plugins run 33656010754 the AI bundle job was 13.4 min: 0.2 min to build the module, 12.7 min to run its suite, while both required gates consume only the artifact. It costs a second time too — a run cannot be re-run until it completes, so after a shard flakes the fix waits a further ~13 min on a suite nobody is waiting for.

Issue #3103, opened 2026-09-02
Cross-repo pairs: four shapes the surface gate cannot see — the answer is triggering the dependent's suite, not a wider detector


rbuergi:
Follow-up to #2689, which closed once the pair gate shipped (#3091). That gate covers **one** of the five shapes the issue named. These are the other four, recorded so they are not mistaken for solved.

## What the gate does cover

A core PR that **removes or moves a `public` top-level type** must declare its counterpart, and the job refuses to pass while that counterpart is open, draft, or unmerged-to-`main`. Mutation-proven against #2678's real diff, including the case that matters most: a counterpart PR that *is* merged, but into a feature branch rather than `main`.

## 🚨 The four it does not cover

1. **An added overload breaking a `<see cref>` in a dependent.** Nothing is removed, so the surface detector sees nothing. This is how #2678 also broke `MeshWeaver.AI` — a dangling `<see cref="ApiTokenLayoutAreas"/>` where `CS1574` is an error — a second, independent break from one merge.

2. **A JSON envelope's field names changing.** No C# surface moves at all; the contract is the serialized shape, and the dependent parses it.

3. **A comment another repo's regex parses as data.** The platform sees prose; the dependent sees a schema.

4. **The i18n mirror shape (#2650).** A *value* change, not a surface deletion — every plugin-repo PR went red on a job those diffs could not possibly reach, until the mirror PR landed.

## Why a wider detector is the wrong instinct

Each of these is detectable only by knowing what the *dependent* consumes, and core cannot read the private repos — that constraint is what made `check-type-forwards.py --sibling` unusable in CI in the first place, and it has not changed.

The issue's own criterion is the structural answer, and it is worth restating as the target rather than a bigger regex:

> **compile-and-run the dependent's suite.**

That is what the node-repo compile-check lanes already do for content. The gap is that a *core* PR does not trigger them, so the failure lands in another repository, on unrelated PRs, minutes to hours later — and the people who see the red are never the people who caused it.

## Suggested next step

Rather than extending the pair gate, the higher-value change is to make a core PR that touches public surface able to *trigger* a dependent's compile-check against its merge ref. That is a credential and orchestration problem, not a detection one — and unlike a wider regex, it covers all five shapes at once, including ones nobody has thought of yet.

rbuergi:
## A sixth shape, and it fired for real an hour after the gate shipped

[...]

PR #3178, opened 2026-09-03
Dependent suites: a manual run resolves the merge commit through the API — refs/pull/N/merge is materialized lazily (#3103 follow-up)

rbuergi:
## `Dependent suites`: a manual run resolves the merge commit through the API — `refs/pull/N/merge` is materialized lazily

Follow-up to #3171 / #3173 (#3103). The first manual run — [33730055183](https://github.com/Systemorph/MeshWeaver/actions/runs/33730055183), against #3170 minutes after two merges to `main` — died in `actions/checkout`:

```
fatal: couldn't find remote ref refs/pull/3170/merge
```

`git ls-remote` at that moment advertised only `refs/pull/3170/head`; the API reported `mergeable: null, mergeable_state: unknown`. GitHub computes a pull request's merge commit in the background, on demand, and drops it whenever the base moves — between those two moments the ref does not exist. Relying on it was the defect, not the timing.

A manual run now `GET`s the pull request (which is what triggers the recomputation), polls up to a minute until `mergeable` is known, and checks out the **merge commit by SHA** — a real object in the repository whether or not a ref advertises it, which is exactly how `actions/checkout` fetches `github.sha` on a `pull_request` event. A conflicting (`mergeable: false`) or non-open pull request is refused with its reason; a merge commit that never resolves is a RED, not a guess. The job gains `pull-requests: read` for that read. The `pull_request` path is unchanged.

Verification: `actionlint` clean; `check-workflow-shell.py` 0 live findings; `check-workflow-timeouts.py` 0 violations. Pure CI change, no What's New.

Claude-Session: https://claude.ai/code/session_01G4xPnGYEbdu8AVvR5jytyj

### DIFF
--- a/.github/workflows/dependent-suites.yml
+++ b/.github/workflows/dependent-suites.yml
@@ -92,16 +92,49 @@ jobs:
     timeout-minutes: 45
     permissions:
       contents: read
+      pull-requests: read
     steps:
+      # 🚨 `refs/pull/<n>/merge` is MATERIALIZED LAZILY. GitHub computes a pull request's merge
+      # commit in the background, on demand, and drops it whenever the base moves; between those
+      # two moments the ref is simply absent — `git fetch` answers "couldn't find remote ref". The
+      # first manual run of this workflow (33730055183, against #3170 minutes after two merges to
+      # main) died in checkout on exactly that. So a manual run resolves the pull request through
+      # the API, whose GET is what triggers the recomputation, polls until `mergeable` is known,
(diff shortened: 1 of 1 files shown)


rbuergi:
Withdrawn: the maintainer asked to remove the Dependent suites check from core PRs (red on every PR while no receiver exists). The dispatcher workflow is being removed in a follow-up PR; the member-level surface detector and the searched:false red from #3171 stay.

PR #3220, opened 2026-09-03
First-run setup: an empty image boots into a wizard for database, sign-in, model keys and modules

rbuergi:
An instance with **no `Graph:Storage` and no completed `instance.json`** now serves a first-run
setup wizard and nothing else, writes the operator's answers to the instance manifest, and restarts
configured. This is **slice 3 of [Instance Identity and Setup](https://github.com/Systemorph/MeshWeaver/blob/main/src/MeshWeaver.Documentation/Data/Architecture/InstanceIdentityAndSetup.md)** — `MeshBuilder.IsAwaitingSetup` finally has a production reader.

Every deployment that exists today states its storage in configuration and is **byte-identical**.

## What it asks

| Step | Default | Written to | Read next boot as |
|---|---|---|---|
| Database | **SQLite** — a single file, nothing else to run | `Storage` | `Graph:Storage:*` |
| Sign-in | developer login on | `SignIn` | `Authentication:*`, `GitHub:OAuth:*` |
| Model keys + embeddings | none | `Ai` | `{section}:ApiKey`, `Embedding:*` |
| Modules + packages | gRPC, `Plugins/*` | `BootModules` / `ProvisionPackages` | `Modules:Assemblies` |

Every list is **discovered** — keyed `IStorageAdapterFactory` registrations, `SignInProviderCatalog`,
the provider sections — so an operator cannot record a backend the image cannot open (which fails at
the *next* boot with `Unknown storage type`, after the wizard is gone).

`AddInstanceManifest` inserts the projection at **index 0**, so appsettings / ConfigMap / env still win.

## 🚨 Where the secrets live

- **Master key** — a deploy secret, else `instance.key` (`0600`) *beside* the manifest, **never in it**: a manifest gets copied, backed up and pasted into issues, and a key stored beside its ciphertext is plaintext with extra steps.
- **Sign-in secrets** — in the manifest, `enc:v1:`. They cannot be mesh nodes: auth schemes register before any storage is open.
- **Model keys** — in the manifest as a *bootstrap envelope*, until the built-in provider imports them into the `Provider` partition on the first configured boot.

A secret that cannot be encrypted is **refused**, never stored in the clear (the 2026-08-24 leak's
shape). Ciphertext is never projected as a value — a handler handed ciphertext registers, renders
its button, and fails at the token exchange far from its cause.

## 🚨 Why a separate host, not a branch

Measured, not chosen. Two attempts to serve the wizard from inside the ordinary pipeline **died
before reaching it**: `MapMeshWeaver` asserts a permission evaluator that `ConfigureMemexMesh` never
registered on a setup-mode host; past that, `EventSubscriptionRunner` — an `IHostedService` a module
registers *before* the storage decision — could not resolve `IMeshService`. A portal's startup
assumes a configured mesh from very early on.

## 🚨 The chart had to stop answering these questions

The portal ConfigMap emitted `Graph__Storage__Type=""` for a values file that stated nothing — and an
**empty section is not an absent one**: `GraphStorageConfig.Type` defaults to `FileSystem`, so the
instance boots past setup onto container-ephemeral disk. Wizard-owned keys are now emitted only when
stated (`helm template` with defaults is unchanged; verified both ways). The sign-in keys matter more
sharply: an always-emitted `Authentication__EnableDevLogin: "false"` **outranks the manifest** and
would silently discard the operator's answer.

## Homebrew

`values.local.yaml` stops making the identity decision for the operator (dev login on,
`<your-username>` as sole admin, a commented Entra block). `memex-local setup` prints the wizard's
URL and its one-time token and opens it. Pinning an answer in the overlay still works and still wins
— that is how an unattended install skips the wizard.

`LocalOverlayDefaultsToDevLoginTest` **follows its subject** rather than being deleted: the overlay
must now *ask*, and the dev-login default is pinned where it moved to — the wizard's own
pre-selection (`SetupSurfaceTest.TheDeveloperLogin_IsOnByDefault`).

## Verification

- `Memex.Portal.Shared.Test` — **694/694 pass**, including 30 new ones (projection rules, master-key derivation and round-trip, composition refusals, the surface over a real HTTP pipeline).
- `MeshWeaver.Documentation.Test` — link integrity + the rewritten overlay guard, 14/14.
- `deploy/homebrew/test/run-tests.sh` — 53/53.
- All six touched projects build `-c Release -warnaserror`.
- **End to end in a browser**: a Monolith booted with no `Graph:Storage` served the wizard; Playwright filled and submitted it; the restart came up on SQLite with a real 86 KB `memex.db`, 6 mesh nodes and the `embedding` column present, `/dev/login` answering 200 from the manifest's answer.

## Pairs-with

`Pairs-with: none — this PR removes no public type or member; it only adds (`InstanceManifest.SignIn`/`.Ai`, `InstanceManifestProjection`, `InstanceMasterKey`, `LiteralMasterKeyProvider`, the `Memex.Portal.Shared.Setup` namespace). The MeshWeaver.Plugins half (SQLite keyed factory, the catalog provider, the host branch) is additive on top of this and lands after it.

https://claude.ai/code/session_0155oPabmCjpdBEoTz5bXQRJ


rbuergi:
Both review findings landed after auto-merge had already fired, and both were correct. They are
addressed in #3224:

- **the connection string was echoed back into the form** — inconsistent with the sign-in secrets in
  the same file, and it carries a password. Withheld now, like the others.
- **`AlreadyConfigured` documented behaviour the markup did not implement** — the fields are
  `disabled` (not `readonly`: a disabled control is not submitted at all), and `SetupComposition`
  refuses such an answer too, so the guarantee does not rest on the markup.

Both pinned by tests.

PR #3248, opened 2026-09-03
Shard the node-repo gate: a cap cut reports as `cancelled`, so the verdict is unreadable

rbuergi:
## The defect: a cap cut reports as `cancelled`, so the gate's verdict is unreadable

`node-repo-gate.yml` installs every package into a fresh mesh, **serially**. On MeshWeaver.Plugins that is 59 installs, each paying the `portal/nodeops` tax of #2543, against `timeout-minutes: 30`.

Measured across **28 consecutive gate jobs** on MeshWeaver.Plugins (2026-09-03, 12:32-16:29Z), each classified by reading its **duration and its verdict text** — never its colour:

| what the job actually was | n | durations |
|---|---|---|
| green, full set | 12 | 17.9 – 27.6 min (median 21.9) |
| green, narrowed PR closure | 3 | 7.7 – 9.9 min |
| **cut at the 30-minute cap** | **4** | 30.3 – 30.4 min |
| real content verdict (`GATE FAILED — install: Hosting`) | 3 | 28.1, 29.4, 29.7 min |
| superseded (the whole run cancelled) | 1 | 22.9 min |
| red on its inputs (image / bundle preflight) | 5 | 0.3 – 2.2 min |

**Of the 20 runs that gated the full set, 4 (20%) were cut at the cap and 7 (35%) came within two minutes of it.** A cap cut reports as `cancelled`, byte-identical to a real cancellation, so:

- **it blocked nothing** — MeshWeaver.Plugins#1253 merged with this gate `cancelled`;
- **it took the Tests-area evidence with it** — the caller's ratchet is `if: always() && needs.test-repos.result != 'cancelled'`. On cap-cut run [33776132232](https://github.com/Systemorph/MeshWeaver.Plugins/actions/runs/33776132232) every other one of the 61 jobs finished; only the gate was cancelled, and both `Tests-area ratchet over the gate log` and `publish-bake` were skipped.

🚨 **A `failure` on this gate is NOT evidence of a cap cut.** Three of the four longest reds in that sample were genuine content verdicts. Attributing them to the cap inflates the "before" and overstates what sharding buys.

**Raising the cap is not the alternative.** The fleet ceiling is 45 minutes, so it buys 15 against a moving tail — and a 45-minute cut still reports `cancelled`.

## What lands

**1. `mw-plugin-test --shard <i>/<n>`** (`GateShardPlan`). A shard gates its slice and **installs that slice's forward dependency closure**, reusing the installed-but-not-gated shape an upstream package already has (`PackageResult.Upstream`) — so every package is gated **exactly once** across the fan-out. The closure comes from `LocalNodeRepo.DependencyMap`, the *same* map the install order is sorted by; deliberately not a second reader of those three edge kinds (a second reader is how `affected-modules.py` had 51 of 52 `requires` edges dead).

**2. Equal package COUNTS, and no weight table.** Measured per package from run 33758875713's gate log:

| package | files | NodeTypes | measured |
|---|---|---|---|
| Store | 185 | 17 | 56 s |
| Hosting | 141 | 15 | 109 s |
| RolePlay | 38 | 4 | **129 s** |
| Chess | 37 | 3 | **111 s** |

Every structural proxy is anti-correlated enough to lose to plain equal counts. Simulated makespan at 4 shards: equal counts **8.8 min**, files 13.4, NodeType count 14.1, test bytes 16.6, perfect oracle 8.4 — equal counts is within 0.4 min of the oracle at every shard count from 2 to 6. So there is **nothing here to re-measure and nothing to go stale** (contrast `shard-assign.sh`, whose table drifted for three weeks).

**3. The lane becomes `plan` → `gate` (matrix, `fail-fast: false`) → `verify`.** `verify` keeps the name **`Compile + render node repos (MeshWeaver from ACR)`**, so the required context on the three satellites that require it is untouched. The per-shard contexts (`Gate shard i/n`) deliberately are **not** requirable — they appear and disappear with the input, exactly why `All selected bundles built` exists in the module-pack lane.

**4. `gate-shard-merge.py` FOLDS the shard logs and refuses a fan-out whose slices are not a disjoint cover.** It reads each shard's own plan receipt and fails on: a missing shard log, a package claimed by two shards, a package that fell between the slices, shards disagreeing about the discovered total, a shard that printed no plan line (it ran without `--shard`, so it judged everything), a truncated summary. Seven cases, all self-tested (`--self-test`).

### Why a fold and not a concatenation — measured

`check-test-suites.py` reads **the LAST** `=== mw-plugin-test summary ===` block. Splitting run 33758875713's real gate log into 4 synthetic shard logs and running MeshWeaver.Plugins' own ratchet over each shape:

| input to the ratchet | what it reported | exit |
|---|---|---|
| the original single log | 87 NodeType(s) across 59 package(s), 0 skipped | 0 |
| the four logs **concatenated** | **10 NodeType(s) across 15 package(s)**, 0 skipped | **0** |
| the four logs **folded** | 87 NodeType(s) across 59 package(s), 0 skipped | 0 |

The concatenation is a false pass — a green tick over one sixth of the evidence. The fold is byte-compatible with the caller's ratchet and is re-emitted under the canonical `gate-log-<sha>` name, so no caller changes.

## Expected effect, and how it will be verified

Modelled from the phase breakdown above (fixed per shard ≈ 2.5 min: image pull, `/app` extract, the compile pass; variable = the shard's own installs + its closure):

| shards | wall-clock | runner-minutes |
|---|---|---|
| 1 (today) | 21.9 min | 21.9 |
| 3 | 11.1 | 31.6 |
| **4** | **10.8** | **35.5** |
| 6 | 8.3 | 43.0 |

The model reproduces the measured 1-shard baseline to 0.1 min, which is the only reason to trust the rest of the column — but **it is a model, not a measurement.** The measured after/before, with sample sizes, follows on the MeshWeaver.Plugins pin-bump PR that opts in; this PR does not claim the improvement is verified.

## 🚨 This reaches a satellite only when that satellite bumps its pin

This is a `workflow_call` lane pinned by each caller with a 40-char commit sha, so **nothing changes for MeshWeaver.Plugins / .SocialMedia / .Reinsurance / .Manufacturing until they bump `uses:` and `platform-ref`.** And with `shards: 1` (the default) the tester command line is byte-for-byte the pre-fan-out one, so a caller that bumps the pin *without* opting in sees no behaviour change at all.

`--shard` ships in the **tester image**, so a caller setting `shards: > 1` also needs `image-digest` / `platform-image-digest` from a CD wave carrying it. The shard job asserts the flag by name against the composed host before it gates, so a stale pin fails with a message naming the pin rather than `Unknown argument '--shard'` naming the tester's argv.

## Verified locally

- `dotnet build -c Release -warnaserror`: `tools/MeshWeaver.PluginTester`, `test/MeshWeaver.PluginTester.Test`, `src/MeshWeaver.Documentation`, `test/MeshWeaver.Documentation.Test` — 0 warnings, 0 errors each.
- `MeshWeaver.PluginTester.Test` — 339 passed (15 of them the new `GateShardPlanTest`).
- `MeshWeaver.Documentation.Test` — 246 passed, including `NodeRepoLaneHostGuard` and `DocumentationLinkIntegrityTest`.
- `check-workflow-timeouts.py --root .` — 66 jobs, 0 violations (`plan` 10, `gate` 30, `verify` 15).
- `check-workflow-shell.py` — 0 live findings.
- `gate-shard-merge.py --self-test` — 0 failures.
- The fold/concatenation table above, against the real `gate-log` artifact of run 33758875713.

Doc: `Doc/Architecture/NodeRepoGateSharding`. What's New: *The package gate no longer runs out of time* (`Category: Fix`).

Pairs-with: none — this PR removes no public type and no public member; it only adds (`GateShard`, `GateShardPlan`, `LocalNodeRepo.DependencyMap`, `PackageResult.Support`) and changes a CI lane that every caller pins by sha.


rbuergi:
## Verified end-to-end against the pinned portal image, before any CI run

This lane is `workflow_call`-only, so **core's own CI never exercises it** — the first honest test would otherwise be a satellite's pin bump. So it was run locally, composing the gate host exactly as the lane does (`compose-gate-host.sh` over `memex-portal-ai@sha256:c4467638…` + the tester CLI, `234 assemblies at the root`) and composing the same three module bundles the caller passes as `--module`, fetched from Plugins run 33758875713. The one deviation is that the CLI is this branch's build, because that is where `--shard` lives.

Over `Store, Training, Video, Northwind, Chess` from the real MeshWeaver.Plugins tree (Chess depends on Training; everything depends on Store):

```text
shard 1/2: gating 3 of 5 discovered package(s) — Store, Training, Video; installing 0 support package(s) …
shard 2/2: gating 2 of 5 discovered package(s) — Northwind, Chess; installing 2 support package(s)
           gated on another shard: Store, Training
```

Shard 2 **installed** Store and Training — it could not install Chess otherwise — and did **not judge** them:

```text
shard 2's summary: [PASS] Store (184 node(s), 0 type(s)) [support: installed, gated on another shard]
shard 1's summary: [PASS] Store (184 node(s), 17 type(s))   ← all seventeen verdicts
```

**Folding the two real logs and comparing the verdict SET against a `--shard 1/1` run of the same five packages: 27 verdict lines each, IDENTICAL.** Every package once, all 21 NodeTypes (17 Store + 1 Training + 3 Chess) compiled, rendered and Tests-executed, `ALL GREEN.`

Both earlier attempts are worth recording, because each failure was the design working rather than a surprise: a mount missing `Training` failed Chess's install with `NodeType(s) not registered: Training/Tour`, and a run without the `MeshWeaver.AI` module failed with `NodeType(s) not registered: Skill` — the content-shaped-error-with-an-infrastructure-cause the lane's own comments warn about.

Also verified against the real `gate-log` artifact of run 33758875713 (59 packages, 87 NodeTypes), running **MeshWeaver.Plugins' own `check-test-suites.py`** over three shapes of the same evidence:

| input to the caller's ratchet | reported | exit |
|---|---|---|
| the original single log | 87 NodeType(s) across 59 package(s) | 0 |
| the four shard logs **concatenated** | **10 NodeType(s) across 15 package(s)** | **0** |
| the four shard logs **folded** | 87 NodeType(s) across 59 package(s) | 0 |

Recorded in `Doc/Architecture/NodeRepoGateSharding`.

Issue #1270, opened 2026-08-12
A gate that can never reopen, and an unbounded nested create, can strand a request forever


rbuergi:
Two independent holes surfaced while tracing the `SkillAutocompleteTest` teardown leak (PR #1269). Neither is reachable from *that* path once the AiSettings seed can no longer outlive its reader, but both are live in their own right.

## 1. `DataContextInit` can never reopen after shutdown starts

`src/MeshWeaver.Data/DataContext.cs` → `OpenInitializationGate`:

```csharp
if (Hub.IsShuttingDown)
{
    logger.LogDebug("DataContext initialization for {Address} ended by hub disposal — recognized "
        + "shutdown outcome, no failure state recorded.");
    return;                              // ← returns WITHOUT opening the gate
}
...
Hub.OpenGate(InitializationGateName);
```

Every hub built with `AddData()` starts with `DataContextInit` **shut** (`DataExtensions.cs`: `.WithInitializationGate(DataContext.InitializationGateName, d => d.Message is PingRequest)`). If data-source init resolves while the hub is shutting down, the gate stays shut permanently and **anything already deferred behind it is stranded** — the only escape is the 30 s `DeferralTimeout` in `MessageService`.

Observed directly: a `CreateNodeRequest` arriving at `runLevel=Quiescing` recorded `DEFERRED gates=[DataContextInit]` and never drained.

A gate that will never open should **fail** its deferred messages, not park them — per the repo's own "wedges to zero, never a silent hang" rule. A caller getting a prompt `DeliveryFailure` can react; a caller parked for 30 s cannot.

## 2. `MeshService`'s advertised operation timeout is dead code

`src/MeshWeaver.Hosting/MeshService.cs` class doc:

> Each call is bounded by `MeshOperationOptions.Timeout` so a lost/slow response surfaces as `TimeoutException` … never a hang

`OpTimeout` is defined — and **`grep` finds no `.Timeout(` anywhere in the file**. `CreateNode` is an unbounded `hub.Observe<CreateNodeResponse>` round trip.

This matters because `EnsurePartitionBootstrap` issues **nested** creates through it (`ProvisionAndCreateRoot`, `CreateCreatorGrant`). A nested create that is never answered makes the *outer* create chain neither emit, fault, nor complete — so the #981 `CREATE_CHAIN_COMPLETED_EMPTY` backstop, which only runs on `OnCompleted`, cannot fire. The caller waits forever.

[...]

PR #3261, opened 2026-09-03
A hosted hub's leaked callback is reported on its owner, and a subtree drains under one Quiescing budget

rbuergi:
Measured 2026-09-03 on a MeshWeaver.Plugins suite (`MeshWeaver.Hosting.Monolith.Test`, local run with the per-test phase trace and `MESHWEAVER_MSG_TRACE`, load 5–8 on 12 cores): **72 of 690 disposals took ≈2.0 s — 145 s of an 11-minute run — and the leak detector reported zero for every one of them.** The distribution is bimodal: 617 disposals under 100 ms, 72 at 2,002–2,168 ms, one in between. The 2 s is the hosted-hub `QuiesceTimeout` default.

## Two defects, both by construction

**1. A departed child's verdict left with it.** A hosted hub removes itself from its owner's `HostedHubsCollection` in its own `DisposeImpl` (ShutDown) — *after* its Quiescing phase has set `QuiescingTimedOut`. The owner's `AnyHubQuiescingTimedOut()` / `GetQuiescingTimeoutSummary()` walk `hostedHubs.Hubs`, and a caller can only ask after the owner's own `DisposalCompleted`, i.e. after every child has finished and departed. So a child's timed-out quiesce was unreachable. `HostedHubsCollection` now snapshots a departing child's summary in that same removal callback (`DepartedQuiesceTimeouts`) and both walkers consult it. The report names the hub and the request:

```
[QUIESCE-TIMEOUT] hosted/leaky-child: 1 callback(s) still pending after 0.3s — forcibly cancelling.
  Pending: …=SwallowedRequest@hosted/leaky-child(307ms)
```

**2. Children drained under a different clock than their owner.** `CreateMessageHub` starts every hub from a fresh `MessageHubConfiguration` whose `QuiesceTimeout` is the production default (2 s). A mesh that configured a different budget on its root — the test base's 500 ms — had hosted hubs paying 2 s per leaked child regardless. The owner now seeds its collection (`InheritedQuiesceTimeout`); a hosted hub inherits it unless its own configuration says otherwise. One policy per subtree; a hub-specific override still wins; **production keeps the default everywhere**, so nothing changes there.

## What this does not yet do

It makes the class *reportable*; it does not remove the callbacks. With this merged, a Plugins suite whose hosted hubs leak will start **failing** on `DISPOSE_QUIESCE_LEAK` with the request named — which is the point, and also the blast radius: those are real leaks that were silently costing 2 s each. The first specimen is already in hand from the message trace (a `SubscribeRequest` parked behind a hub's `DataContextInit`/`MeshNodeInit` gates at the instant of teardown); the local re-run against this branch is what names the rest.

## 🚨 A detector that cannot fail

The number worth less than the shape: for as long as this code has existed, the leak report was **structurally unable to see the thing it exists to find**, and reported clean. It is the same class as an assertion satisfied by the wrong field, a query projection that returns silent nulls, a green CI whose required contexts were absent, and a watcher rendering a throttled empty response as a verdict — a check whose failure path was unreachable. The pin here is written so it fails on the unfixed code first; the ticker change (second commit) is the same idea one layer up: a diagnostics snapshot at 3 s can never describe a 2 s wait.

## Third commit: teardown answers every gate that is still shut

With the report reachable, the biggest single mechanism became visible and had a clean root cause.
An initialization gate promises to release what is parked behind it when it opens; **shutdown is the
outcome after which no gate can ever open**. `DataContext` learned this for its own gate (#1270) and
fails `DataContextInit` on shutdown — but that fix lives in the gate's **owner**, so it did not
generalise. `MeshNodeInit` (armed in `MeshDataSource`, opened in five places) has no such hook, and
the entire `src/` tree contained exactly **one** `FailGate` call. A delivery deferred behind
`gates=[DataContextInit,MeshNodeInit]` had one gate answered and stayed parked on the other.

`MessageHub.Dispose()` now calls `messageService.FailAllGatesOnShutdown()` right after
`SignalShuttingDown()` — before the Quiescing poll starts counting, because a delivery parked behind
an unanswered gate is exactly a pending callback that poll can only wait out. In the framework
rather than in each owner: the difference between fixing two gates and making a third impossible to
get wrong.

Measured on the Plugins Monolith suite (801 tests), same machine, same load band:

| leaked-callback last stage | before | after |
|---|---|---|
| `DEFERRED gates=[DataContextInit,MeshNodeInit]` | 12 | **0** |
| `DEFERRED gates=[DataContextInit]` | 3 | **0** |
| `DROPPED_SHUTTING_DOWN runLevel=Dead` | 8 | 3 |
| **total leaked callbacks** | **69** | **63** |

🚨 **Read the failing-test count honestly:** 38 → 37, because most failing tests leak by more than one
mechanism — killing one class does not clear a test that also leaks another way. The callback counts
above are the measurement that matters.

## What is left, named

Three mechanisms remain and are NOT fixed here: a handler that exits `Processed` whose async reply
never follows (18), a handler fault whose error arm never answers the requester (11 —
`MessageService`'s generic branch reports through `ReportFailure`, which returns early once
`RunLevel >= DisposeHostedHubs`, and `NackThroughParent` also declines when the parent is past that
level), and a reply posted to a hub that had already stopped listening. Each deserves its own change.

## Not an `InternalsVisibleTo`

`QuiesceTimeout` becomes public (`WithQuiesceTimeout` already was) so the pin can read it. An `InternalsVisibleTo` on Messaging.Hub was tried first and broke two unrelated test files — it shifts overload resolution of the internal `Observe(object, …)` across the whole test assembly, exactly as `MessageStormBreaker.cs` records.

## Verification

Release, `-warnaserror`, one project per invocation: `MeshWeaver.Messaging.Hub.Test`, `MeshWeaver.Documentation.Test` — `0 Error(s)`.

- `HostedHubQuiesceLeakVisibilityTest` (3/3): a self-addressed request the child swallows, posted as the hub (the post pipeline fails closed on a missing `AccessContext`), the owner disposed, and the owner's report naming `hosted/leaky-child` + `SwallowedRequest` after the child departed — in 0.3 s, not 2. A request aimed at the *owner* is deliberately not the shape: the owner's teardown NACKs it, which settles the callback and is exactly the case that does not leak.
- `QuiescingTimeoutNamesHandlerSideTest` (4/4) unchanged.
- `DocumentationLinkIntegrityTest`, `TestTimeoutLiteralRatchetGuard` green.

Doc: `HubDisposalModel.md` gains the section between the DisposeHostedHubs and ShutDown phases. What's New (Fix).

Pairs-with: none — an added default and two added members; no public type or member is removed.

https://claude.ai/code/session_01UxxLJxS5SWxuwP4JK1NTNY

### DIFF
--- a/src/MeshWeaver.Documentation/Data/Architecture/HubDisposalModel.md
+++ b/src/MeshWeaver.Documentation/Data/Architecture/HubDisposalModel.md
@@ -247,6 +247,41 @@ Pinned by `HostedHubQuiesceLeakVisibilityTest`: a self-addressed request the chi
 the case that settles), the owner disposed, and the owner's report naming the child and the
 request after the child has departed.
 
+### 🚨 Teardown answers every gate that is still shut
+
+An initialization gate is a promise: the delivery parked behind it is released when the gate opens.
+**Shutdown is the outcome after which no gate can ever open**, so every gate its owner did not
+answer breaks that promise, and the deferred caller hears nothing until an unrelated deadline
+expires — its own request timeout, or whenever teardown reaches `messageService.Dispose()`.
+
+`DataContext` learned this for its own gate (#1270) and fails `DataContextInit` explicitly on
+shutdown. That fix could not generalise, because it lives in the gate's **owner**: the
+`MeshNodeInit` gate — armed in `MeshDataSource`, opened in five places across `MeshDataSource` and
+`MeshNodeTypeSource` — had no such hook at all. The whole `src/` tree contained exactly **one**
+`FailGate` call. A delivery deferred behind `gates=[DataContextInit,MeshNodeInit]` therefore had one
+of its two gates answered and stayed parked on the other, for ever.
+
+`MessageHub.Dispose()` now calls `messageService.FailAllGatesOnShutdown()` immediately after
+`SignalShuttingDown()` — before the Quiescing poll starts counting, because a delivery parked behind
+an unanswered gate is a pending callback that poll can only wait out, at the full budget per hub.
+Doing it in the FRAMEWORK rather than in each owner is the difference between fixing two gates and
+making a third one impossible to get wrong.
+
+Measured on the MeshWeaver.Plugins Monolith suite (801 tests), with the leak report made reachable
+by the section above:
+
+| leaked-callback last stage | before | after |
+|---|---|---|
+| `DEFERRED gates=[DataContextInit,MeshNodeInit]` | 12 | **0** |
+| `DEFERRED gates=[DataContextInit]` | 3 | **0** |
+| `DROPPED_SHUTTING_DOWN runLevel=Dead` | 8 | 3 |
+| total leaked callbacks | 69 | 63 |
+
+The remaining leaks are three other mechanisms — a handler that exits `Processed` without its async
+reply ever following, a handler fault whose error arm never answers the requester, and a reply
+posted to a hub that had already stopped listening. They are real, they are now NAMED, and they are
+not this section's subject.
+
 ### ShutDown — tear down and signal
 
 Runs on the action block, fully synchronous: `CancelCallbacks()` (push
--- a/src/MeshWeaver.Documentation/Data/Architecture/HubDisposalModel.md
+++ b/src/MeshWeaver.Documentation/Data/Architecture/HubDisposalModel.md
@@ -214,6 +214,39 @@ late construction produced, so a hub built during the teardown window is never l
 outside the snapshot. On completion **or** the cap, the owner advances to ShutDown — a
 hung child never blocks the parent.
 
+### 🚨 A departed child's verdict stays with the owner — and a subtree drains under one budget
+
+Two facts about hosted hubs that were each correct on their own and, together, made a whole class
+of leaks unreportable (measured 2026-09-03 on a MeshWeaver.Plugins suite: **72 of 690 disposals
+took exactly the 2 s hosted-hub Quiescing budget — 145 s of an 11-minute run — and the leak
+detector reported zero**).
+
+1. A hosted hub **removes itself from its owner's `HostedHubsCollection`** in its own ShutDown
+   phase (`DisposeImpl` runs the `RegisterForDisposal` callbacks, one of which is
+   `messageHubs.TryRemove`). That is *after* its Quiescing phase has set `QuiescingTimedOut`.
+2. The owner's `AnyHubQuiescingTimedOut()` / `GetQuiescingTimeoutSummary()` walk
+   `hostedHubs.Hubs` — and a caller can only ask after the owner's *own* `DisposalCompleted`,
+   which is after every child has finished and left. So a child's timed-out quiesce was
+   **unreachable by construction**: the verdict left with the child.
+
+`HostedHubsCollection` now snapshots a departing child's summary
+(`DepartedQuiesceTimeouts`) in that same removal callback, and both walkers consult it. The
+report names the hub and the request, which is the line someone greps for.
+
+The second half is the budget itself. `CreateMessageHub` starts every hub from a fresh
+`MessageHubConfiguration`, whose `QuiesceTimeout` is the production default (2 s) — so a mesh that
+configured a different budget on its root (the test base's 500 ms; an operator's longer one) had
+children draining under a different clock than their owner, and a test base that thought it had
+bounded leaks at 500 ms was paying 2 s per leaked child. The owner now seeds its collection with
+its own budget (`InheritedQuiesceTimeout`); a hosted hub inherits it unless its configuration
+says otherwise, so **one policy per subtree** and a hub-specific override still wins. In
+production nothing changes — every hub already carried the default.
+
+Pinned by `HostedHubQuiesceLeakVisibilityTest`: a self-addressed request the child swallows
+(a request aimed at the *owner* does not leak — the owner's teardown NACKs it, which is exactly
+the case that settles), the owner disposed, and the owner's report naming the child and the
+request after the child has departed.
+
 ### ShutDown — tear down and signal
 
 Runs on the action block, fully synchronous: `CancelCallbacks()` (push
--- a/src/MeshWeaver.Documentation/Data/WhatsNew/2026-09-03-a-leaked-callback-in-a-hosted-hub-is-now-reported.md
+++ b/src/MeshWeaver.Documentation/Data/WhatsNew/2026-09-03-a-leaked-callback-in-a-hosted-hub-is-now-reported.md
@@ -1,0 +1,22 @@
+---
+Name: A leaked callback inside a hosted hub is now reported
+Category: Fix
+Description: A request that was never answered inside a hosted hub used to cost every shutdown a silent two-second wait and was invisible to the leak report; the report now names the hub and the request, and hosted hubs drain under the same budget as their owner.
+Icon: Sparkle
+Order: -20260903
+---
+
+# A leaked callback inside a hosted hub is now reported
+
+When the platform shuts a hub down it first waits, briefly, for any request that hub is still
+expecting an answer to. If an answer never comes, that wait runs to its budget and the leak is
+supposed to be reported so someone can fix the request that was abandoned.
+
+For hubs hosted inside another hub, the report never fired. By the time the owning hub asked its
+children whether anything had leaked, the children had already finished and left — taking the
+answer with them. Measured on one test suite: 52 of 320 classes were paying that silent two-second
+wait, and the report said nothing.
+
+A child now hands its verdict to its owner as it leaves, so the report names the hub and the
+request. And a hosted hub is created with its owner's waiting budget rather than a fixed default,
+so a whole tree of hubs drains under one policy.
--- a/src/MeshWeaver.Messaging.Hub/HostedHubsCollection.cs
+++ b/src/MeshWeaver.Messaging.Hub/HostedHubsCollection.cs
@@ -1,4 +1,5 @@
 ﻿using System.Collections.Concurrent;
+using System.Collections.Immutable;
 using System.Diagnostics;
 using System.Reactive;
 using System.Reactive.Linq;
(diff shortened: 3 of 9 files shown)


rbuergi:
Converting to **draft** — the code is finished and green locally; what is open is a policy call, not a defect.

## The red is this PR working

All 10 core failures are `left Observe subscriptions pending past the Quiescing budget` — pre-existing leaks that were invisible until this PR made the report reachable. Shards 0–2 are clean; shard 3's two apparent failures are `PluginGateRunnerTest` printing its own **expected** `[FAIL] Broken` output.

| leaking class | n |
|---|---|
| `MeshWeaver.Compiler.Pipeline.Test.LeavingHubAdoptionSweepTest` | 2 |
| `Memex.Portal.Shared.Test.PackageInstallRebuildsDependentsTest` | 2 |
| `MeshWeaver.Graph.Test.NodeTypeCompileStateMirrorTest` | 1 |
| `MeshWeaver.Graph.Test.DanglingNodeTypeUpdateTest` | 1 |
| `MeshWeaver.Graph.Test.NackReachesTheWaiterDuringTeardownTest` | 1 |
| `Memex.Portal.Shared.Test.SessionDenialIsAnAnswerTest` | 1 |
| `Memex.Portal.Shared.Test.InstanceKeyAdoptionTest` | 1 |
| `Memex.Portal.Shared.Test.InstanceKeyLiveResolutionTest` | 1 |

8 × `SubscribeRequest`, 5 × `CreateOrUpdateNodeRequest`, 1 × `PatchDataRequest`; 9 of the 10 held on the `cache/<id>` hub. This is the blast radius the PR body predicted, now enumerated.

## The call to make

1. **Ratchet (recommended).** Record these 10 in a shrink-only allow file; any NEW leak fails. It is this repo's own idiom — `BlockingBridgeSites.allow`, `HubDisposalJoinSites.allow`, `UnthrottledResubscribeSites.allow`, and `TestTimeoutLiteralRatchetGuard`'s baseline that "only ever goes down". The list IS the inventory: visible, countable, shrinking — and the failure path stays live for everything not on it.
2. **Fix all 10 first.** Each is its own investigation into a request whose reply never comes.
3. **Report-only.** Argued against: that rebuilds a detector that cannot fail, which is the defect this PR removes.

Nothing here is a band-aid either way — option 1 keeps the gate armed for every case except a named, shrinking list.

rbuergi:
Measured this draft on a merge with #3291 (the teardown change: no forced teardown, stall detector, grace-before-cancel drain, progress-based activity quiesce). One conflict, in `HostedHubsCollection.CreateHostedHub`: this branch's inherited quiesce budget at creation vs. main's `HostedHubResult` + the #3243 teardown-race catch — resolved by keeping both (create with the inherited budget, return `HostedHubResult`). Merged branch: `fix/3261-on-robust-disposal` (local worktree only, not pushed).

Results on that tree, Release, the classes this PR's own runs reddened:

| set | result |
|---|---|
| `HostedHubQuiesceLeakVisibilityTest`, `DisposalStallWatchdogTest`, `HostedHubDisposalJoinTest` | 9 passed |
| `DanglingNodeTypeUpdateTest`, `NodeTypeCompileStateMirrorTest`, `NackReachesTheWaiterDuringTeardownTest` | 8 passed, 1 failed |
| `LeavingHubAdoptionSweepTest` | 2 failed |
| the five `Memex.Portal.Shared.Test` classes | 18 passed, 5 failed |

Every remaining red is the same finding this PR exists to surface, and all eight have one shape: a hub at teardown holds ONE pending callback for a request to a **sibling that is itself disposing**, and the 0.50 s test quiesce expires before the sibling answers —

- `SubscribeRequest@…` and `PatchDataRequest@Plugins/_Policy` pending on a `cache/*` hub (5×);
- `CreateOrUpdateNodeRequest@portal/nodeops-*` pending on a type hub (3×, the same shape Loki shows on `memex-cloud` at the 2 s production budget: `[QUIESCE-TIMEOUT] … CreateNodeRequest@portal/nodeops-*`).

So these are not leaks in the tests; they are the requester's **fixed** quiesce budget racing the sibling's **shutdown-time** answer (the sibling NACKs what it cannot serve only when its own `messageService.Dispose()` runs, several phases later). #3291 makes every other teardown wait progress-based; Quiescing is the one phase still on a duration. Making it stall-based — keep waiting while replies keep landing, give up only after a budget with none — is the follow-up that would turn these eight into passes without widening anything; until then this detector will report them, which is what it is for.

rbuergi:
## Measured on this branch ALONE — the eight reds are NOT caused by #3291

I merged this branch with `fix/robust-disposal` (#3291) to see whether the new stall-based
teardown removed them. It does not, and it is not implicated either: **running
`origin/fix/hosted-hub-quiesce-visibility` on its own reproduces every one of them**, at the same
tests, with `#3291` nowhere in the tree.

| suite | red | on this branch alone |
|---|---|---|
| `MeshWeaver.Graph.Test` | `DanglingNodeTypeUpdateTest.Upsert_WithTheNamedImportEscapeHatch_StillLands` | ✅ reproduces |
| `MeshWeaver.Compiler.Pipeline.Test` | `LeavingHubAdoptionSweepTest` ×2 | ✅ reproduces |
| `Memex.Portal.Shared.Test` | `PackageInstallRebuildsDependentsTest` ×2, `SessionDenialIsAnAnswerTest`, `InstanceKeyAdoptionTest` | (measured on the merged tree) |
| `MeshWeaver.Messaging.Hub.Test` | `HostedHubQuiesceLeakVisibilityTest`, `DisposalStallWatchdogTest`, `HostedHubDisposalJoinTest` | ✅ 9/9 GREEN — these pass with #3291 |

So this is exactly the blast radius the PR body predicts: real leaks, previously costing 2 s each
in silence. They are **four distinct shapes**, and the ledger names each one:

- `⇒ a reply WAS posted … the reply was lost between the responder and the requester` (graph, 1)
- `⇒ the chain FAULTED and no reply was posted` (compiler 1, portal 1)
- `⇒ a handler was entered and no reply, completion or fault has been recorded since` (compiler 2)
- `⇒ the delivery reached a hub but no handler was ever entered` (portal 2)

## The specimen, traced end to end

`DanglingNodeTypeUpdateTest`, `MESHWEAVER_MSG_TRACE=1` plus `MeshWeaver.Messaging: Debug`. The
whole test lives **980 ms** and its teardown starts at **+393 ms**. The leaked callback is a
`SubscribeRequest` from `cache/iIN4…` to `TestData/dnt289b340b`, and the decisive lines are:

```
11.928 [QUIESCE-START] TestData/dnt289b340b: 0 pending callbacks at dispose entry: <none>
11.928 [QUIESCE-OK]    TestData/dnt289b340b: drained 0 callback(s) in 0ms
11.968 Start processing SubscribeRequest (ID: p2qsFux3…, cache/iIN4… -> TestData/dnt289b340b)
11.969 Start processing SubscribeRequest (ID: pglEBn_z…, cache/iIN4… -> TestData/dnt289b340b)
11.969 Creating new hosted hub for address sync/4wzOafyEDUCs6r3rq9b-ag in host TestData/dnt289b340b
```

The node hub **quiesced clean and then accepted three more `SubscribeRequest`s**, spinning up
fresh `sync/*` children for them, while the mesh was already tearing down. Exactly one of the
three was answered (`DeliveryFailure Am9fMOOc…`, dispatched to the cache hub's subject at 11.969);
the others died with their target, and the cache hub's own Quiescing then expired with one of them
pending. In the earlier run the same test took the other door — `GATE_FAILED
gates=[DataContextInit,MeshNodeInit]` at +14 ms, whose `AnswerUnreleasableDelivery` NACK also never
reached the requester.

**So the shared root cause of at least the graph and portal specimens is not a slow reply — it is
that a hub past `[QUIESCE-OK]` still ACCEPTS new correlated work.** The intake gate at
`ScheduleNotify` only refuses from `DisposeHostedHubs` on, so the window between "quiesce finished"
and "children disposed" takes requests it can never answer, and can still create hosted hubs to
serve them. Every such request is a callback the requester will hold until ITS budget expires —
which is the 2 s × 72 disposals this PR measured.

## Suggested split

The detector here is right and should land; the callbacks are a separate change set (the intake
window above, plus the two "handler entered, nothing recorded" shapes, which need
`hub.NoteRequestStage` at the terminal arms the ledger names). I have not touched either — this is
measurement, so the PR's owner keeps the wheel. #3291 is unaffected either way: its own teardown
suites are green here.

Full trace and per-shape ledger output: session `meshweaver-bd`, worktrees `/private/tmp/MW-3261`
(merged) and `/private/tmp/MW-3261-solo` (this branch alone).

rbuergi:
## Another instance of this class, in a different suite — and it reddens unrelated PRs

`MeshWeaver.PluginTester.Test.BakeEquivalenceTest.MeshDrivenAndCompilerDrivenBakes_ProduceTheSameArtifacts` failed shard 3 on **#3332**, whose diff is `deploy/**` and health-probe endpoints — it cannot reach PluginTester. The log shows the same shape this issue is about:

```
ok  Widget/Thing [compile:Passed render:Passed tests:Skipped]          ← the bake SUCCEEDED
…
[QUIESCE-TIMEOUT] Widget/Thing: 2 callback(s) still pending after 2s — forcibly cancelling.
  Pending: HYUEK0RZB0ePe-zHDHwSjw=CreateOrUpdateNodeRequest@portal/nodeops-O0ixdCgnbkawTtQz-bUBQw
[CompileStateMirror] Widget/Thing: satellite write failed … HubDisposedBeforeResponseException
CompileWatcher: failed to create Release node at Widget/Thing/Release/… HubDisposedBeforeResponseException
[ReleasePostCondition] Widget/Thing: a release request … was consumed and this compile succeeded, yet the no[de …]
```

`CreateOrUpdateNodeRequest@portal/nodeops-*` pending at quiesce is verbatim one of the four shapes in my measurement above. Here the consequence is visible rather than silent: the compile's follow-on writes (CompileStateMirror satellite, Release node) die with the hub, the release post-condition correctly reports a consumed request that produced no node, and an unrelated PR goes red.

**Why this raises the priority.** The eight reds on this branch are contained — they are this PR's own detector doing its job. But the same leak is now demonstrably reddening PRs on `main` that have nothing to do with it, and it presents as a bake failure rather than as a leak, so the next person to see it will debug the baker.

Note this is on a core that already carries #3291: the forced-teardown watchdog is gone, but the Quiescing budget still cancels pending callbacks at expiry, so a write that has not been answered by then is still lost. #3291 made teardown honest about what it stops; it did not make these particular writes finish.

PR #3267, opened 2026-09-03
First-run setup asks WHO before HOW: name it, register it, then choose from your plan

rbuergi:
Reorders the first-run wizard around the flow the maintainer specified:

> *a hello dialog asking for instance name. the id (guid) is used to register it in
> memex.meshweaver.cloud. then it gets access to free plugins. from plugin list, we choose postgres
> and some connection string to some local pg* — then *lead through main sections to select auth,
> database, llm settings*.

**The order is the substance, not presentation.** Registration is what gives the instance a plan,
the plan is what lists the plugins, and the plugins are where the remaining answers — the database
above all — come from. Asking "which database?" first, as this wizard did, is asking before there is
anything to choose from.

## Phase one — who

Asks for a **name**. The id is a **guid the instance mints for itself**, rendered read-only: ids are
claimed globally and never re-issued even after deletion, so an operator must not be able to type one
they might type again on another install. A guid already satisfies the registry's alphabet.

`InstanceIdRulesMatchTheRegistryTest` pins our copy of that rule against
`MeshWeaverInstanceService`'s — two copies of one rule is exactly the shape that drifts, and here the
drift costs a real, permanent id.

Registration is **open** by default (no bootstrap key), which is what enrols the free plan.

🚨 **The issued key is encrypted into the manifest before anything else can fail.** The registry
returns it exactly once and never re-issues it, so a crash between "registered" and "wrote it down"
permanently burns that id. For the same reason a missing name is refused *before* the registry call —
an id claimed for a form that was going to be rejected is an id burnt for nothing. And if the key
cannot be stored, the refusal says the id is *already claimed* and to fix the directory rather than
retry under a new one.

## Phase two — how

Lists what the plan entitles this instance to, then leads through plugins → database → sign-in →
models → modules. A package declaring a `StorageType` becomes a database option, and choosing it
provisions that package too.

🚨 **A storage package for a backend the image cannot open is listed as a plugin but never offered as
a database.** Landing such a module would have to happen *before* persistence selection reads
`Graph:Storage`, and package provisioning runs after the mesh is up — so offering it would record a
backend that never resolves and fail at the next boot with the wizard gone. Tested explicitly.

A registry that cannot be reached is **reported, not swallowed**: an instance that registered but
cannot see its plan is a state worth naming, where an empty list would read as "your plan includes
nothing".

## Verification

Against an in-process `FakeRegistry` serving the real wire contract — the client under test is the
**real** one, exercising real serialization and every refusal path (taken id → "already claimed",
rejected key, unreachable host, accepted-but-no-key).

Deliberately **not** the live registry: a suite pointed there would burn a permanent id per run on
shared infrastructure. The single live registration is the maintainer's to make.

**758/758** in `Memex.Portal.Shared.Test`; all four touched projects build `-c Release -warnaserror`.

`Pairs-with: none — additive. Builds on #3266 (a Graph:Storage section without a Type means no
storage), which should land first.`


rbuergi:
**Dependent note — the i18n mirror must follow this PR, core-first.**

This PR adds **34 localization keys** in both languages (`setup.hello.*`, `setup.identity.*`,
`setup.packages.*`, `setup.problem.*`, the `plural.package.*` pair). Those keys have a second home in
`MeshWeaver.Plugins/clients/react/src/i18n/`, whose drift guard compares **values**, not just keys.

Measured against the Plugins mirror as it stands: **34 missing, 0 value drift** — so nothing is
broken, but the mirror is now behind by exactly this PR's additions.

Sequence, per the localization rule: **core lands first**, then the mirror syncs
(`npm run sync:i18n -- --ref <this merge sha>`). The mirror PR stays red until then, and that red
must not be "fixed" by reverting the mirror.

Flagged explicitly because `AGENTS.md` now states there is **no `Dependent suites (MeshWeaver.Plugins)`
check**, and the `Cross-repo pair (public surface)` gate sees only two of the six break shapes — a
public type leaving `src/`, and a public member leaving a type that stays. An **i18n value change is
one of the four it cannot see**, so a green core PR here is not evidence the dependent is fine.

Checked this PR against the other invisible shapes too:
- **new overloads → `CS0419` in a dependent's `<see cref>`**: none added in `src/`.
- **JSON envelope field names**: no serialized member removed or renamed.
- **a comment another repo parses as data**: none touched.

The one genuinely new public surface is `PackageManifest.StorageType` — an **added** optional field,
consumed by the Plugins half of the setup wizard (#1284), so it is additive in the safe direction.

rbuergi:
## The red is `Public surface (binary compatibility)`, and there is an additive fix that needs no allow entry

Diagnosing rather than claiming the PR — it has been red since 2026-09-03T21:11Z and is now `dirty`, so it is running no fresh CI at all and the results below are frozen at yesterday's head (`4d4671f45`).

**Both findings are the same shape** — a parameter added *positionally* to a record's primary constructor:

```
✗ memex/Memex.Portal.Shared/Setup/SetupCatalog.cs
    record SetupCatalog — arity: 4 → 7
      + ImmutableList<SetupPackageOption> Packages = …
      + MeshWeaver.Mesh.InstanceIdentitySelection? Identity = …
      + string? RegistryProblem = …

✗ memex/Memex.Portal.Shared/Setup/SetupCatalog.cs
    record SetupStorageOption — arity: 5 → 6
      + string? PackageId = …

🚨 2 binary-breaking record change(s), 0 stale allow entr(ies).
```

The gate's reasoning is right and worth restating, because a default argument makes this look safe when it is not: **adding a parameter — even with a default — replaces the signature.** The default is a *compile-time* courtesy; the emitted constructor is a different method, so every assembly already compiled against the old arity calls one that no longer exists. A module and the platform it loads into must agree exactly, and no image can serve a mixed set.

## The fix: `init` properties, not constructor parameters

All four additions are optional, so none of them needs to be positional. Converting them to `init` properties is **binary-additive** — old callers keep working, new callers use an object initialiser — and it needs no `record-signatures.allow` entry at all.

There is precedent in this repo, with the reasoning already written down. `InstanceRegistrationPayloads.Response`:

```csharp
public record Response(string InstanceId, string InstanceKey)
{
    /// <summary>The plan the instance was registered ON (#2804) … An init property, not a
    /// constructor parameter, so the wire shape stays binary-compatible.</summary>
    public string? Plan { get; init; }
}
```

Applied here that is:

```csharp
public record SetupCatalog(
    ImmutableList<SetupStorageOption> Storage,
    ImmutableList<SetupSignInOption> SignIn,
    ImmutableList<SetupAiOption> Ai,
    ImmutableList<SetupModuleOption> Modules)
{
    public ImmutableList<SetupPackageOption> Packages { get; init; } = ImmutableList<SetupPackageOption>.Empty;
    public InstanceIdentitySelection? Identity { get; init; }
    public string? RegistryProblem { get; init; }
}

public record SetupStorageOption(
    string Type, string DisplayName, bool NeedsConnectionString = …, bool NeedsBasePath = …,
    string? ConnectionStringHint = …)
{
    public string? PackageId { get; init; }
}
```

Construction sites change from trailing positional arguments to `with { … }` / an object initialiser; JSON round-tripping is unaffected (`init` setters deserialize normally).

**The allow-file route is the wrong tool here.** `scripts/record-signatures.allow` exists for a break that is genuinely necessary and gets an atomically-planned move; taking it for four optional additions would spend a real escape hatch on a change that does not need one, and would leave a live arity mismatch to coordinate across the module boundary.

## Also worth knowing before the next run

The PR is **`dirty`** — it needs `main` merged before CI will run again, and `main` has moved a lot today (15 PRs). Two of those touch nearby ground: #3252 changed notification addressing across `MeshWeaver.Graph`, and #3287 added a `rest-not-graphql` shared-rule block plus its manifest entry. Worth merging main first and re-reading the reds on the fresh head rather than fixing yesterday's list — `Run tests (shard 5)` in particular may have a different cause now.

Not touching the branch. Happy to implement the above if its owner would rather hand it over.

rbuergi:
Pushed 4d432b4a1 for the one remaining red (run 33855809520, shard 5): `ObservableToTaskBridgeGuard.NoNewBlockingBridgeInProductionCode` named a new site in `memex/Memex.Portal.Shared/Setup/SetupEndpoints.cs` — `CatalogFor` blocked the request thread on `ListPackagesAsync(...).GetAwaiter().GetResult()`. The helper now returns `Task<SetupCatalog>` and its two callers (the GET page and the phase-two `Apply`) await it on the request's own continuation, the shape `RegisterIdentity` already uses. No behaviour change otherwise.

Verified locally in Release (`-warnaserror`): `Memex.Portal.Shared.Test` setup classes 70 passed; `ObservableToTaskBridgeGuard` 9 passed (the site count is back to the inventory).

Issue #3276, opened 2026-09-04
A core change in b88be8a05..c41a34fd breaks the plugin-catalog card render — caught only by a test in another repo


rbuergi:
## A core change in `b88be8a05..c41a34fd` breaks the plugin-catalog card render — and core CI cannot see it

Two tests go red when the core commit under test advances across this range, and are **6/6 green** at the old end:

```
MeshWeaver.PluginCatalog.Test.CatalogRenderTest.Catalog_ListsPackagesFromGitSource_AsCards
MeshWeaver.PluginCatalog.Test.CatalogRenderTest.ALandedModule_PutsTheRestartNoteOnItsPackageCard
```

| | |
|---|---|
| Good | core `b88be8a058689e216f0df20e795cde121da0a365` |
| Bad | core `c41a34fd` |
| Range | **48 commits** (`git rev-list --count b88be8a05..c41a34fd` = 48) |
| Subject under test | `src/MeshWeaver.PluginCatalog/` — **this repo's code** |
| Where the test lives | `Systemorph/MeshWeaver.Plugins`, `src/MeshWeaver.PluginCatalog.Test/CatalogRenderTest.cs` |

## 🚨 Why this is worth a ticket rather than a note

**The test that catches it is in another repository, so nothing in this repo's CI runs it.** `MeshWeaver.PluginCatalog` is compiled and shipped from here; its card-rendering behaviour is asserted only over in MeshWeaver.Plugins, by a job (`Portal hosts`) that checks core out at whatever `MW_PLATFORM_REF` names. So:

- every one of these 48 commits was merged green here;
- the breakage becomes visible only when a satellite deliberately advances its pin;
- and it then presents as *"my pin bump is red"*, which reads like the bumper's problem and is not.

This is the shape recorded in `Doc/Architecture/` and in the fleet's own history as a test that is never executed before merge. The `Dependent suites (MeshWeaver.Plugins)` gate exists for exactly this class — worth checking whether it covers `MeshWeaver.PluginCatalog.Test` and, if it does, why this got through; if it does not, that is the actionable gap and it is bigger than the two tests.

## How it surfaced

[...]

PR #3285, opened 2026-09-04
docs(#3276): shape 7 — a behaviour change behind an unchanged signature, and the pin that hides it

rbuergi:
## The bisect for #3276, and the shape no detector can see

`c521f929e` (#3258, merged `97ad19cc1`) made the plugin catalog open on its **categories**, so
`CatalogLayoutAreas.RenderFromSource` renders tiles and no package card on the parameterless area —
deliberately, and with core-side coverage of the new shape shipped in the same commit
(`test/Memex.Portal.Shared.Test/CatalogCategoryFirstTest.cs`). Two render tests over in
MeshWeaver.Plugins require a card there, and went red. They were found **48 commits later**, by a pin
bump, presenting as *"my pin bump is red"*.

Bisected — both ends reproduced first, then the 13-merge first-parent path narrowed by sandwiching
the candidate merge:

| core commit | | the two tests |
|---|---|---|
| `b88be8a05` (Plugins' `MW_PLATFORM_REF`) | good end | **2 passed** |
| `a2b312d9f` = `97ad19cc1^1` | main immediately before | **2 passed** |
| **`97ad19cc1`** (the merge of #3258) | | **2 FAILED** |
| `c41a34fd` | bad end | **2 FAILED** |

`c521f929e` is also the only commit of the 48 that touches `src/MeshWeaver.PluginCatalog/` at all.
**The dependent's assertions are the half that had to catch up**, and do in
Systemorph/MeshWeaver.Plugins#1315 — on the `?all=true` page, with nothing relaxed. This PR is the
other half: **why nothing told the dependent.**

## What this adds, after #3280

#3280 landed the AGENTS.md half — *"there is no `Dependent suites (MeshWeaver.Plugins)` check"* —
while this was in flight, and it says it better than my version did, so **main's wording is taken
wholesale**. Two things it could not know, because they come out of this bisect:

### 1 · Shape 7 — a BEHAVIOUR change behind an UNCHANGED signature

`RenderFromSource` has the same name and the same parameters before and after; it answers
differently. The `Cross-repo pair (public surface)` gate fires on public types and members
**leaving**, and `c521f929e` removed nothing — it passed on #3258, correctly. Neither an API-surface
diff nor a binary-compatibility check can see this class, by construction. Added to
`CrossRepoPairGate.md` as **shape 7**; "the six shapes" becomes seven, and the two counts that
quantify the gate's reach (*"TWO of the six … the other four"*) are corrected with it — in the doc
and in the sentence #3280 just wrote, so the two do not drift apart on their first day.

### 2 · The event-based fallback does NOT fire for a dependent that PINS

`CrossRepoPairGate.md` said a *"shape-1…6 break therefore surfaces as a red release-follow run in the
dependent's repository, minutes after the platform published"*. That holds only if the dependent's
source lane resolves to the released commit — and MeshWeaver.Plugins pins core (`MW_PLATFORM_REF`,
its own #1255, bought with a real incident). Its `platform-ref` job applies the **same pin** on a
`repository_dispatch` release event as on a pull request; only the *bake target* follows the release.
Its `ci.yml` says so in its own words:

> *"The PIN exists so two runs of identical code agree … Every GATE reads it, unchanged. … on the two
> triggers that mean 'a release happened' … the bake targets the CURRENT released image."*

So the release-follow run rebuilds the dependent against the commit it already used, and a break
introduced after the pin is invisible to it until a human bumps the pin. **For a dependent that pins,
the pin bump IS the integration test**, and the cost of a break scales with how long the pin sat —
48 commits here. That is not an argument for unpinning; it is an argument for saying it out loud.
`RepositoryDependencyDirection.md` already states the accurate version; this makes
`CrossRepoPairGate.md` agree with it.

## No new gate, deliberately

Rebuilding the `Dependent suites` dispatcher walks straight back into the directive that removed it
(`7c27f3e95`, 2026-09-03 — *"none of the top-level repos should have any dependency to anyone
else"*), and making Plugins' release-follow lane resolve the released core instead of its pin would
undo #1255, in a file another session owns. So this records the boundary rather than moving it.

## Verification

```
dotnet build test/MeshWeaver.Documentation.Test/... -c Release -warnaserror   0 Warning(s)  0 Error(s)
dotnet test  test/MeshWeaver.Documentation.Test/... -c Release --no-build     Passed!  250/250
```

Re-run after merging `origin/main` (the PR was `DIRTY` on AGENTS.md against #3280).
`DocumentationLinkIntegrityTest`, `WhatsNewEntryIntegrityTest` and `MigrationWorkloadModelGuard` all
live in that project and all ran. Revert-check on the merge: `git diff origin/main...HEAD --stat`
touches **only** `AGENTS.md` (one line) and `CrossRepoPairGate.md`.

No What's New entry: documentation and agent guidance only.

Refs #3276 · pairs with Systemorph/MeshWeaver.Plugins#1315 (test-only, in that repo) · builds on #3280.


rbuergi:
## Dequeued by the steward — for a failure this PR cannot reach

The merge-queue steward left this out with `queue-rejected`, correctly: the failing job was
`AGENTS.md shared rule blocks`, and a gate failure is never treated as a flake. But the failure is
not this PR's:

```
[rest-not-graphql] Systemorph/MeshWeaver.Education:    AGENTS.md opens a shared-rule block that
[rest-not-graphql] Systemorph/MeshWeaver.SocialMedia:  .github/shared-rules.json does not declare.
[rest-not-graphql] Systemorph/MeshWeaver.Manufacturing: Nothing compares it, so it reads as
[rest-not-graphql] Systemorph/Memex:                    protected while being unchecked.
```

Four satellites opened a **`rest-not-graphql`** shared-rule block whose counterpart entry is missing
from core's `.github/shared-rules.json`. This PR touches neither the manifest nor any `shared-rule:`
marker — its AGENTS.md edit is one line inside the un-marked, core-only paragraph 92.

**It is red on `main` itself**, so it is blocking every core merge, not just this one:

| run on `main` @ `192a60d8` | conclusion |
|---|---|
| 07:21:09Z | success |
| **07:37:06Z** | **failure** |

The drift landed between those two runs. Per the gate's own rule — *"the remedy is always a one-PR
fix in the repo the error NAMES; there is no bypass"* — this needs either the manifest entry in core
or the markers removed in the four satellites, and that belongs to whoever is rolling
`rest-not-graphql` (a `docs/rest-not-graphql` branch was building in MeshWeaver.Plugins at 07:26Z).
I am deliberately not guessing at the intended slot semantics from the outside.

So this PR stays out of the queue until that lands. Its own checks were **27/27 green** on
`9967de1e`; nothing here needs changing. Re-arm it with `gh pr merge --auto` once `main` is green
again — not by re-queueing or re-running the rejected queue build.

The half of #3276 that actually unblocks the pin — Systemorph/MeshWeaver.Plugins#1315 — **is merged**
(`df7950df`).

Issue #3211, opened 2026-09-03
A module bundle that states no framework identity is still publishable, so the #3154 comparison has nothing to compare — part 3 of Plugins#931


rbuergi:
The consumer half of Plugins#931 landed in #3154 (merged `d51b2971`, 2026-09-03): `ModuleUpdateDecision.Decide` now compares **(version, framework identity)** before answering `SkipUpToDate`, and the registry advertises a per-bundle `frameworkMvid` on `/api/plugins/bundles/index.json`.

That change deliberately left one blind spot open, and named where it must be closed rather than papering over it at the consumer:

> an unstated served identity SKIPS — and the verdict SAYS the identity could not be checked rather than implying the two agree. […] the place to remove that blind spot is where it is created: part 3 of the agreed shape makes a bundle that cannot state what it was built against unpublishable
> — `ModuleUpdateDecision.Decide`, `bundleFrameworkMvid` parameter doc

The reasoning for skipping there is sound and should not be revisited at the consumer: answering `Land` on an unstated served identity re-downloads every module on every reconcile, forever, on every deployment pointed at a registry that predates the field. Landing can heal an unknown on the *landed* side (it writes the identity back); it can never heal one on the *served* side.

## What is still open

**A module can still be published with no framework identity.** `ModulePublish.Accepted.FrameworkMvid` is `string?` and `ModulePublish.Validate` does not refuse a manifest that states none — so the shelf entry carries null, the index advertises `frameworkMvid: null`, and every consumer of that bundle is permanently in the skip-and-say-so branch. The comparison exists but has nothing to compare.

Concretely:

- `src/MeshWeaver.PluginCatalog/ModulePublish.cs:45` — `string? FrameworkMvid` on `Accepted`, never validated.
- `memex/Memex.Portal.Shared/Api/PluginBundleEndpoints.cs` — `ShelveModule(frameworkMvid: accepted.FrameworkMvid, …)` shelves whatever arrived, including null.

## Also stale from #3154

`src/MeshWeaver.PluginCatalog/ModulePublish.cs:39` still reads:

> `<param name="FrameworkMvid">The framework build the producer compiled against — diagnostic only.</param>`

[...]

PR #3293, opened 2026-09-04
fix(ci): the bundle identity anchor is a platform property, not a module's reference graph

rbuergi:
## What broke

Core CD has delivered nothing since **2026-09-04 10:23Z**. Eight red checks on `main`, all one cause:

```
Plugins: pack the module bundles the bake composes / Module bundle (MeshWeaver.Maps)              failure
Plugins: pack the module bundles the bake composes / Module bundle (MeshWeaver.Payments.Stripe)   failure
Plugins: pack the module bundles the bake composes / All selected bundles built                   failure
CD delivered, or had a good reason not to                                                         failure
```

Both bundle jobs die identically:

```
##[error]no identity anchor for MeshWeaver.Maps — expected MeshWeaver.Compiler.dll at
'…/src/MeshWeaver.Maps/bin/Release/net10.0/publish/MeshWeaver.Compiler.dll', i.e. in the
module's own build output (no platform image is pinned, so the platform was built from source).
```

## Why

Measured across `main`'s last 40 commits, the lane built **two** entries and both were green:

| commit | time | bundles |
|---|---|---|
| `41a00a50d` | 10:09 | `MeshWeaver.AI`=success `MeshWeaver.Markdown.Collaboration`=success |
| `1318ca66c` | 10:23 | + `MeshWeaver.Maps`=**failure** `MeshWeaver.Payments.Stripe`=**failure** |

`1318ca66c` is #3290, which correctly grew the compose set to every module the Plugins content
binds. It did not break the packer — it **falsified an assumption the packer had been carrying**.

The from-source arm read the anchor out of `$PACKDIR`, the module's own publish output, on this
stated ground:

> platform built from SOURCE (`REFS` empty — core's own main-cd call) — the platform
> ProjectReferences are real, so MeshWeaver.Compiler.dll IS beside the module in the publish output.

`dotnet publish` puts it there only when the module **transitively ProjectReferences it**:

- `MeshWeaver.AI` → `MeshWeaver.Graph` → **`MeshWeaver.Compiler`** ✅
- `MeshWeaver.Markdown.Collaboration` → `MeshWeaver.Graph` → **`MeshWeaver.Compiler`** ✅
- `MeshWeaver.Maps` → `MeshWeaver.Layout` → `MeshWeaver.Data{,.Contract}` → **no compiler** ❌
- `MeshWeaver.Payments.Stripe` — a leaf → **no compiler** ❌

So a property of two reference graphs was written down as a property of the arm. The population
that could have falsified it was 2, and both members agreed.

## The fix

The anchor is a property of the **platform** the call pins, not of a module's reference graph — so
derive it from that platform, identically for every module, with the same global properties the
module build passes (`MeshWeaverRoot`, `Version`), which is what makes it the assembly `publish`
would have copied.

The `[ ! -f "$anchor" ]` refusal is untouched: an anchor that cannot be built is still RED. This
adds no fallback and no second opinion — it removes an accident.

## Verified

Clean `bin/Release`, then the exact command the arm now runs:

```
$ dotnet build src/MeshWeaver.Compiler/MeshWeaver.Compiler.csproj -c Release \
    -p:MeshWeaverRoot=… -p:Version=9.9.9-anchortest
Build succeeded.  0 Warning(s)  0 Error(s)
$ ls src/MeshWeaver.Compiler/bin/Release/net10.0/MeshWeaver.Compiler.dll
-rw-r--r--  136192  MeshWeaver.Compiler.dll
```

The `container` arm and the pinned-image arm are unchanged.

Refs #3211


rbuergi:
We converged on the same CD break from opposite ends — you at 12:32, me at 12:53 (#3296) — and the two fixes differ in a way I think decides it, so here is the measurement rather than an opinion.

**Building the compiler per module leaves the identity per-module.** This PR builds `MeshWeaver.Compiler.csproj` with `-p:Version="$VERSION"`, which is the *module package's* version — different for every matrix entry. `Version` reaches the assembly attributes, and the attributes are compilation input, so `MeshWeaver.Compiler.dll` still gets a different MVID per module. That is not hypothetical: it is what CD run **33867996265** already recorded, one run, one platform commit, from the copies publish happened to drop —

|---|---|

So the "no identity anchor" error goes away and the field stays a property of the module.

**And neither value can ever match a consumer.** A portal image is built `-p:CIRun=true` on GitHub Actions, so `AddCommitHashMetadata` stamps `AssemblyMetadata("MeshWeaverFrameworkIdentity", "g<sha>")` on its `MeshWeaver.Compiler`, and `FrameworkIdentity.ReadIdentity` prefers that stamp over the MVID (#1660 WS3). A consumer therefore compares against `g<sha>`; a bundle stating an MVID is never equal to it. The field #3154 reads on every update decision has been unmatchable in every source-built bundle the fleet has published — which is the part I think is worth fixing while we are both in here.

**#3296 states it instead of deriving it:** in the source arm, `--framework-mvid "g$(git -C meshweaver rev-parse HEAD)"` — the packer's own flag for "the platform's own reading", `g<sha>` being one of the three tokens it documents (`s<hash>`, `g<sha>`, 32-hex MVID). One identity per platform commit, identical for every module in the run, and the same token the image stamps. It also needs no extra compiler build per matrix entry.

Both PRs are green (28/28) and they touch the same block, so they cannot both land. I have no attachment to mine being the one — but if this one lands as-is, CD goes green with the identity still per-module and still unmatchable, and the next person to look will be measuring the same thing again. Happy to fold your `-warnaserror` reasoning and the reference-graph explanation (which is better written than mine) into #3296, or to hand you the `g<sha>` change to apply here — your call, whichever gets us one green CD and one identity.

Cross-referencing #3296 so neither of us merges over the other.

rbuergi:
Enqueuing #3296 — CD has now delivered nothing since 09:31Z, and both PRs have been sitting green and unarmed for ~50 minutes with no answer on the discriminator.

Not a claim that mine is better work: yours has the better prose and the `-warnaserror` reasoning, and I said so. The reason I am landing this one rather than yours is the single measured difference — building `MeshWeaver.Compiler` per matrix entry with `-p:Version=$VERSION` keeps the recorded identity a property of the MODULE (run 33867996265: `fc70b6e1…` for AI, `0a96ab4c…` for Markdown.Collaboration, one run, one platform commit), and an MVID can never match a portal image, which stamps `g<sha>` under `-p:CIRun=true`. #3296 states `g<commit>`: one identity per platform commit, and the token the consumer actually reads.

Please close this as superseded — or if you disagree with that measurement, say so and I will revert #3296 rather than argue; the evidence is reproducible from the run above either way. If you want your comment text preserved (it explains the reference-graph cause better than mine), send it over and I will fold it in as a follow-up.

Issue #3175, opened 2026-09-03
Release availability must confirm a CONSISTENT sealed set (dependency records match), not bundle presence — memex-cloud rolled to ci.7621 and then DECLINED SocialMedia at adoption


rbuergi:
## The release availability gate accepted a platform whose sealed bundles were mutually inconsistent — the portal rolled to ci.7621 and then DECLINED SocialMedia at adoption ("dependency record mismatch")

**Maintainer requirement (2026-09-03):** "we must have a clear confirmation that all plugins deployed to an instance are available for the correct platform version. If not the case ⇒ nothing goes."

**Measured on memex.meshweaver.cloud:**
- 2026-09-02 ~20:16Z the self-updater rolled ci.7616 → ci.7621 after `ReleaseAvailabilityService` answered "available" (every installed package had a sealed bundle for identity of 7621).
- 2026-09-03 06:50–07:20Z, image ci.7621, Loki `namespace=memex-cloud`:
  ```
  ShippedPrebuiltBundles: bundle SocialMedia.zip: adopted 0/4 prebuilt assembly(ies); 0 were already current
  Prebuilt assembly for SocialMedia/Post DECLINED: dependency record mismatch — 'MeshWeaver.Markdown.Collaboration' built against mvid:A, live is mvid:B
  Prebuilt assembly for SocialMedia/PostTarget DECLINED before writing (#2813): the bundle records source fingerprint X but the live sources are Y
  MeshNodeStreamCache.GetQuery: Content for Posts/… stayed an untyped JsonElement (TypeRegistry lacks the $type discriminator)
  ```
  Result: `/Posts` renders empty and slow (types compile on first access), the user sees "Something went wrong on this page" (#3174). A half-broken state the gate exists to prevent.
- Since 22:41Z every satellite release-follow bake fails on one upstream Store test (Plugins Systemorph/MeshWeaver#1224/#1225), so nothing is sealed for ci.7647 and the gate HOLDS — that half is correct.

[...]

PR #3296, opened 2026-09-04
CD: a source-built platform states its identity as the COMMIT — CD has been red since 09:31Z

rbuergi:
## Why

**Core CD has not concluded green since 09:31Z today.** Every run from 7751 on dies in the same place:

```
##[error]no identity anchor for MeshWeaver.Maps — expected MeshWeaver.Compiler.dll at
'…/src/MeshWeaver.Maps/bin/Release/net10.0/publish/MeshWeaver.Compiler.dll', i.e. in the
module's own build output (no platform image is pinned, so the platform was built from source).
```

#3290 made the Plugins bake compose **every** module its content binds — four instead of two — and that exposed two defects in the *source* arm of the identity anchor. Both make `frameworkMvid`, the field #3154 compares on every consumer's update decision, useless:

1. **A module that never references `MeshWeaver.Compiler` has no copy of it.** Only `MeshWeaver.Graph` and `MeshWeaver.Compiler.Pipeline` reference it, so `AI` and `Markdown.Collaboration` (→ Graph) carry one while `Maps` (→ Layout) and `Payments.Stripe` (→ Mesh.Contract) do not. Nothing about a module's own reference graph should decide whether the *platform* can state its identity.

2. **Where a copy does exist it is a per-MODULE value.** The source arm rebuilds the platform ProjectReferences with that module's `-p:Version=$VERSION`, and the assembly attributes take it into the MVID. Measured in **one** run (33867996265, core `7d644de95`):

   |---|---|

   Two identities, one platform commit, one run.

And **neither value could ever match a consumer.** A portal image is built with `-p:CIRun=true` on GitHub Actions, so its `MeshWeaver.Compiler` carries the *stamped* commit identity `g<sha>` (`Directory.Build.props` → `AddCommitHashMetadata`, #1660 WS3) — never an MVID. So the comparison had nothing it could match, in every source-built bundle the fleet has published.

## What changes

The source arm **states** the identity instead of reading it off an incidental copy:

```bash
sha="$(git -C "$GITHUB_WORKSPACE/meshweaver" rev-parse HEAD)"
identity=(--framework-mvid "g$sha")
```

`--framework-mvid` is the packer's own flag for "the platform's own reading", and `g<sha>` is one of the three tokens it documents (`s<hash>`, `g<sha>`, 32-hex MVID). The **image arm is untouched** — there the anchor really is a file in the extracted `/app`, and reading it *is* the platform's own reading.

Both arms still end **RED** with nothing to state: the branch picks where the identity comes from, never whether to have one. The source arm's refusal is new (`::error::no identity for $MODULE …`) and fires when `git rev-parse` cannot answer.

## Guards moved with the subject

- **`ModuleIdentityPublishGuard`** pins both arms by shape, asserts the source arm does **not** go back to `$PACKDIR/MeshWeaver.Compiler.dll`, and requires a RED on each. Without the update it would have kept asserting the old shape — a guard passing on prose nobody executes.
- **`module-build-key.py` `RECIPE_VERSION` 2 → 3.** The bytes for the same source genuinely differ now, and a reused recipe-2 bundle attests an identity no consumer can match — those must be repacked, not reused. (This is the lever the script documents for exactly this, and the guard asserts it moved.)

## Verification

```
MeshWeaver.Documentation.Test          257/257 (Release, the governance-guard suite)
module-build-key.py --self-test        36 cases green
python3 -c "yaml.safe_load(node-repo-module-pack.yml)"   parses
```

The other CD failure — `Plugins: bake + seal`, the #3175 double producer — was fixed independently by MeshWeaver.Plugins#1268, which merged a few minutes ago. With both in, CD's remaining legs have nothing known blocking them.

Pairs-with: none — CI lane + guard + script; no public surface changes.

### DIFF
--- a/.github/scripts/module-build-key.py
+++ b/.github/scripts/module-build-key.py
@@ -73,7 +73,15 @@ from pathlib import Path
 # genuinely differ — and a REUSED pre-#3211 bundle would be refused at the publish step, which is the
 # right verdict for those bytes and the wrong thing to spend a run discovering. One rebuild wave, and
 # every recorded key afterwards attests bytes that state what they were built against.
-RECIPE_VERSION = "2"
+# "2" → "3": when the platform is built from SOURCE the identity is now STATED as the commit
+# (`--framework-mvid g<sha>`), not read off a copy of MeshWeaver.Compiler.dll in the module's own
+# publish output. Recipe 2 recorded a per-MODULE value there — the module's own rebuild of the
+# platform carries its own `-p:Version`, which the assembly attributes take into the MVID: measured
+# in ONE run (33867996265) AI packed fc70b6e19be34b65983eb5fa15ce5243 and Markdown.Collaboration
+# 0a96ab4c9eb744558431b2a018dc4699 for the same platform commit, and neither could ever match a
+# portal image, which stamps `g<sha>`. Bundles recorded under recipe 2 therefore attest an identity
+# no consumer can use; they must be repacked, not reused.
+RECIPE_VERSION = "3"
 
 # Repo-level files that change what EVERY project compiles or tests. Presence and content both
 # count; a file the repo does not have hashes as null so adding it later changes the key.
--- a/.github/workflows/node-repo-module-pack.yml
+++ b/.github/workflows/node-repo-module-pack.yml
@@ -1739,18 +1739,55 @@ jobs:
           #     publish output. Measured green on CD run 33779812466:
           #     "built-against MVID ce81fd4e3f5146c7a6ebaa6dc72da3f2".
           #
-          # Both arms end in a RED with no anchor — neither is a quiet fallback, and the packer
-          # refuses again behind them, so a bundle whose identity is a guess cannot exist.
+          # 🚨 …and the SOURCE arm above is wrong in two ways that only showed once core CD composed
+          # more than two modules (#3290, 2026-09-04). Reading the anchor out of the MODULE's own
+          # publish output makes the identity a property of the MODULE, not of the platform:
+          #
+          #   * a module whose reference chain never reaches MeshWeaver.Compiler simply has no copy.
+          #     Only MeshWeaver.Graph and MeshWeaver.Compiler.Pipeline reference it, so AI and
+          #     Markdown.Collaboration (→ Graph) carry one while Maps (→ Layout) and
+          #     Payments.Stripe (→ Mesh.Contract) do not — CD run 33867996265 died there, and every
+          #     run since 7751 with it;
+          #   * and where a copy DOES exist it is the module's OWN rebuild of the platform, made
+          #     with that module's `-p:Version=$VERSION`, which the assembly attributes carry into
+          #     the MVID. Measured in ONE run (33867996265, core 7d644de95): AI recorded
+          #     frameworkMvid fc70b6e19be34b65983eb5fa15ce5243 and Markdown.Collaboration recorded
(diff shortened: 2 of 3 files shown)


rbuergi:
🚨 **Do not merge over #3293** — it fixes the same CD break, opened 21 minutes before this one, and both are green and touch the same block.

I have posted the discriminator there rather than racing: #3293 builds `MeshWeaver.Compiler` per matrix entry with `-p:Version=$VERSION`, so the recorded identity stays a property of the MODULE (that is measured, not predicted — run 33867996265 recorded `fc70b6e1…` for AI and `0a96ab4c…` for Markdown.Collaboration in one run on one platform commit), and an MVID can never match a portal image, which stamps `g<sha>` under `-p:CIRun=true`. This PR states `g<commit>` instead: one identity per platform commit, and the token the consumer actually reads.

Either PR makes CD green. Only this one makes the identity comparable. Whoever picks it up: fold #3293's `-warnaserror` reasoning and its reference-graph explanation in — both are better written than mine — and close the other.

rbuergi:
**Why this was ejected twice, and what is being done about it — both causes now have fixes open.**

The steward's two comments name them:

| ejected at | assertion | fix |
|---|---|---|
| 13:19 | `CascadeTest.IndependentNodesRunInParallel_BoundedByTheSlotCount` | **#3304** |
| 13:40 | `NackReachesTheWaiterDuringTeardownTest.OwnerDisposingUnderMeshTeardown_StillAnswersTheWaitingCaller` | **#3302** |

Neither is this PR's doing — this diff is a CI workflow, a guard test and a script constant, and it cannot reach either subject. Both are real defects rather than noise, so neither is going into `known-flakes.json`:

- **Cascade** asserts that independent nodes overlap by sleeping 80 ms in each body and trusting two land inside it together. Nothing enforces that; on a loaded runner the scheduler can admit, run and release one node before the next is dispatched. #3304 makes each body WAIT, bounded, for a second node to join it — the green path costs nothing, and genuine serialisation still fails, for the right reason.
- **NackReachesTheWaiter** is the owner answering correctly and `HierarchicalRouting` dropping the reply because the parent is past `DisposeHostedHubs`, while the waiter sits armed in the same process. #3302 hands such a reply to the local waiter instead of the floor.

I have cleared the `queue-rejected` label. Not re-queueing yet: the queue is currently jammed by exactly those two flakes, so anything enqueued ahead of them draws the same lottery. Order is #3304 → #3302 → re-queue this one.

(#3293 fixes the same CD break as this PR and is still open — the discriminator is in my earlier comment; whoever picks it up should close one.)

rbuergi:
Closing in favour of #3293, which landed as `fc4adc666` and ends the CD outage. Reverting a landed fix to swap in mine would be churn for no gain — CD is green either way.

The one thing #3293 does not do is make the recorded identity comparable: it builds `MeshWeaver.Compiler` per matrix entry with `-p:Version=$VERSION`, so the MVID stays a property of the MODULE (measured: run 33867996265 recorded `fc70b6e1…` for AI and `0a96ab4c…` for Markdown.Collaboration, one run, one platform commit), and an MVID can never equal the `g<sha>` a portal image stamps under `-p:CIRun=true`, which is what `FrameworkIdentity.ReadIdentity` prefers and `ModuleUpdateDecision` (#3154) compares.

That is filed as #3308 with the measurement, the three-line change, and the `RECIPE_VERSION` bump that must move with it — so the finding outlives this PR rather than closing with it.

PR #3302, opened 2026-09-04
A reply routing cannot deliver goes to the waiter, not the floor — and a quiescing hub stops taking work it can never answer

rbuergi:
## Why — this is currently reddening every open PR

`MeshWeaver.Graph.Test.NackReachesTheWaiterDuringTeardownTest.OwnerDisposingUnderMeshTeardown_StillAnswersTheWaitingCaller` is failing **shard 4 on #3297, #3299 and #3300 simultaneously**, and failed core CI runs 33861078925 and 33865385033 earlier today. It is not three problems and it is not a flake to retry — it is one broken edge, and the test's own log names it:

```
[fence] caller is armed on the late watch (ArmedCount=1)
[dispose] mesh disposal invoked — parent is past DisposeHostedHubs
Message delivery failed for PatchDataResponse (ID: l9ret6X…) in TestData/teardown-nack-node:
  Hub TestData/teardown-nack-node cannot route PatchDataResponse to cache/VLe3ZpQ…
```

The owner **answered correctly**. Its post was accepted — its own run level is still open, so `PostImplGeneric` does not refuse — and one turn later `HierarchicalRouting` finds the parent past `DisposeHostedHubs` and **drops the reply: classified, logged, and told to nobody**. The waiter is in the *same process*, with its registry entry armed and able to take the answer directly, but the routing layer had no way to reach it. So the caller burns its whole budget in silence.

## What changes

**1. An undeliverable correlated reply is handed over, not dropped.** New `IUndeliverableReplySink` in `MeshWeaver.Messaging.Hub`; `HierarchicalRouting` offers a dropped reply to a local waiter on the two arms where it gives up (parent past `DisposeHostedHubs`, and the disposing no-route arm). `LatePatchResponseRegistry` implements it, registered as a **factory over the same singleton** — a second instance would dispatch into a registry nobody armed, which reads exactly like "nobody was waiting". A sink that throws cannot turn a classified drop into a routing fault.

🚨 **Only where the delivery is being dropped.** Serving an armed caller *alongside* a healthy post was tried twice at the patch-verdict seam and reverted both times — the sink dispatches synchronously on the owner's turn, ahead of the state the ack is about (`ComboGateRollTest` run 33863349195, `ImportTypeBeforeInstanceTest` run 33865385033). Here there is no post left to race, which is what makes it safe.

**2. A quiescing hub refuses new work it can never answer.** Quiescing exists so the replies a hub is still *owed* can come in, and the intake gate only refused from `DisposeHostedHubs` on. Between the two, a hub whose quiesce had finished still admitted brand-new work — and built hosted hubs to serve it. Measured on #3261's branch (`DanglingNodeTypeUpdateTest`):

```
[QUIESCE-START] TestData/dnt…: 0 pending callbacks at dispose entry: <none>
[QUIESCE-OK]    TestData/dnt…: drained 0 callback(s) in 0ms
Start processing SubscribeRequest (cache/… -> TestData/dnt…)          ← 40 ms later
Creating new hosted hub for address sync/4wzOafy… in host TestData/dnt…
```

Now refused at the door with the same transient `ShuttingDown` NACK the `DisposeHostedHubs` arm mints. Responses stay admitted — the `RequestId` is read off the **envelope**, so a cross-hub `RawJson` still counts — and so does lifecycle traffic.

🚨 Scoped to what is addressed **here, by somebody else**. `ScheduleNotify` is not an inbox: a hub's own *outbound* posts pass through it too. The first version refused those and `SourcesWatcherStopsAtTeardownTest` went red — the defect in miniature, since `NackThroughParent` declines to answer a hub's own delivery and the callback it had just armed then held Quiescing for the whole budget.

## Tests

- `QuiescingRefusesNewRequestsTest`: a request arriving after Quiescing began is refused at the door (the victim genuinely owes a withheld reply, so the phase is really waiting, and the late request's handler *would* answer — a pass can only come from the refusal), plus an **anti-vacuity anchor** that a reply the hub is owed is still admitted. Verified by reverting the production change: the first fails with *"No exception was thrown"*, the anchor passes either way.
- `DisposalRaceNackTest` door one now **observes** the acceptance it always assumed, waiting on the fate ledger for `ENQUEUED@<victim>` before disposing. It had been decided by whether the host's routing turn beat `Dispose()`; adding a class to the suite shifted the scheduling and it lost. (First attempt at that wait used `o.WithMessageId(...)`, which `Observe<TResponse>` overrides — so it polled a key that never existed and timed out having proved nothing. The `Observe(request, options, messageId)` overload is the right one.)

## Verification

Release, `DOTNET_PROCESSOR_COUNT=2` (CI's shape): `Graph` **1324/1324** (including the failing test above), `Hosting` **350/350**, `Data` **475/475**, `Memex.Portal.Shared` **727/727**, `DisposalRaceNackTest` + `QuiescingRefusesNewRequestsTest` **4/4**.

## What this does NOT do

**It does not close #3261.** I re-measured its eight reds on a tree carrying this change: still 7 of 7 failing (Graph 1, Compiler.Pipeline 2, Portal.Shared 4). The graph specimen's ledger verdict *moved* — from "a reply WAS posted … lost" to "no handler was ever entered" — so the reply-loss path it took is closed and a different leak now dominates. The remaining shapes, counted off the ledger: *"a reply WAS posted … still pending"* ×6, *"no handler was ever entered"* ×3 (a request that cannot be routed during disposal, classified and still unanswered — a third hole, on the request leg), *"handler entered, nothing recorded since"* ×2, *"chain FAULTED, no reply"* ×1. Those need a further change set; this one is scoped to the edge that is breaking CI today.

Pairs-with: none — one new public interface, no removals.


rbuergi:
**Dropping the quiescing intake gate from this PR — CI found a documented contract it contradicts, and the gate had no measured benefit anyway.**

`MeshWeaver.Layout.Test.SubscribeDuringRecycleTest.SubscribeRacingDisposal_IsRejectedAsTransient_NotWithANullReference` went red on shard 5, and its own comments state the contract my gate broke:

> Waits until the hub has demonstrably entered the disposal window (`Quiescing` or later) — **creation frozen, message intake still open**.

> …proves the test exercised the STREAM-CREATION refusal and was not answered by MessageService's intake / deferred-queue NACKs, which carry a different banner and are a different, already-fixed defect.

So a quiescing hub *can* answer new work: it admits the SubscribeRequest, `LayoutAreaHost` fails to build its stream, and the handler NACKs with `HubDisposingException` — a real refusal path with a real diagnosis. My premise ("a hub past Quiescing can never answer a new request, so it must not accept one") is false, and the gate pre-empted that path with a generic door NACK.

Weighed against what the gate actually bought: **nothing measurable.** I re-ran #3261's eight reds on a tree carrying it — still 7 of 7 failing, same counts. The late-admission leak I traced is real, but it is not what those tests fail on, and answering it at the intake gate breaks a case that was working.

So this PR is now the routing hand-over alone — `IUndeliverableReplySink`, the two arms where `HierarchicalRouting` gives up, the registry implementation and its factory registration, plus the `DisposalRaceNackTest` determinism fix. That is the part with direct evidence: it cures `NackReachesTheWaiterDuringTeardownTest`, currently red on #3297, #3299 and #3300 simultaneously.

Re-verifying now over Layout (which I had not run locally — that is how this got past me), Graph, Messaging.Hub, Hosting and Data.

rbuergi:
Re-verified with the quiescing gate removed — the change is now the routing hand-over alone, and every suite I can run locally is green at `DOTNET_PROCESSOR_COUNT=2` (CI's shape):

| suite | result |
|---|---|
| `MeshWeaver.Layout.Test` | **445/445** — the suite whose `SubscribeDuringRecycleTest` caught the dropped gate |
| `MeshWeaver.Graph.Test` | **1324/1324** — includes `NackReachesTheWaiterDuringTeardownTest`, the failure this PR exists for |
| `MeshWeaver.Messaging.Hub.Test` | **367/367** |
| `MeshWeaver.Hosting.Test` | **350/350** |
| `MeshWeaver.Data.Test` | **475/475** |

2,861 tests, zero failures, including both suites that previously found defects in this change.

The diff is now: `IUndeliverableReplySink` (new), the two `HierarchicalRouting` arms that were dropping correlated replies, `LatePatchResponseRegistry` implementing it, the factory registration over the same singleton, and the `DisposalRaceNackTest` determinism fix. No `MessageService` intake change.

rbuergi:
**Folding #3304 in here — the two flakes were cross-blocking each other's fixes.**

Measured deadlock:

| PR | fixes | red on |
|---|---|---|
| #3302 (this) | `NackReachesTheWaiterDuringTeardownTest` | `CascadeTest` (shard 3) |
| #3304 | `CascadeTest` | `NackReachesTheWaiterDuringTeardownTest` (shard 4) |

Each one's CI trips over the flake the OTHER one fixes, so neither can go green alone while both are live — and the merge queue was ejecting everything for the same two reasons (steward comments on #3296 at 13:19 and 13:40, one per flake). Re-running either is a lottery, not a fix.

So #3304's commit is cherry-picked here unchanged and I will close it. This PR now carries both cures, which is the only combination whose CI can actually be green:

- **the routing hand-over** — a correlated reply `HierarchicalRouting` must drop is offered to a waiter armed in the same process (`IUndeliverableReplySink`), instead of being dropped classified-and-silent;
- **the cascade overlap** — the parallelism assertion OBSERVES a second node joining rather than sleeping 80 ms and trusting the scheduler, which on a loaded runner measures the runner.

Local verification of the combined branch, Release at `DOTNET_PROCESSOR_COUNT=2`: `PluginTester` CascadeTest 10/10, plus the earlier full runs — Layout 445/445, Graph 1324/1324 (including the test this PR exists for), Messaging.Hub 367/367, Hosting 350/350, Data 475/475.

rbuergi:
Third CI-blocking failure covered by this PR, found on #3305 (opened after this one):

| PR / shard | failing test | fixed here by |
|---|---|---|
| #3305 shard 4, #3297/#3299/#3300 shard 4, #3296 queue build | `NackReachesTheWaiterDuringTeardownTest` | the routing hand-over |
| #3296 queue build, #3302 shard 3 | `CascadeTest.IndependentNodesRunInParallel_BoundedByTheSlotCount` | observing the overlap |
| **#3305 shard 2** | **`DisposalRaceNackTest.FaultingRequest_IsNacked_WhenTheHandlerRacesDisposal`** | **observing the acceptance** |

That third one is door TWO of the same test whose door one I made deterministic here — both doors go through `RunRace`, and #3305 shows door two racing on main exactly the same way: `Observe` posts on the HOST, whose turn then routes down to the victim, and `Dispose()` runs on the test thread in between, so whether the request got in was decided by which turn won. The fate-ledger wait covers both.

So this PR is currently the fix for every distinct assertion blocking the core queue: three tests, one of them a real framework defect (a reply routing must drop, dropped classified-and-silent while the waiter sits armed) and two tests that inferred a concurrent condition from elapsed time instead of observing it.

Issue #3281, opened 2026-09-04
Activity transcripts: migrate the 21 remaining unkeyed LogMessage sites, and mirror the 41 activity.* keys


rbuergi:
Follow-up to #3236, which landed the seam (`LogMessage.MessageKey` / `MessageArgs`,
`LocalizationCatalog.GetNamed`, `message.Localize(locale)` / `host.Localize(message)`) plus 37 of 58
migrated write sites and the `UnkeyedActivityLogMessageRatchetGuard` inventory.

## What is left, and why each is where it is

`test/UnkeyedActivityLogMessages.allow` carries the 21 remaining unkeyed `new LogMessage(` sites
under `src/`. They split three ways, and only the first group is actual work.

### 1. MIGRATABLE — needs a key threaded through a public signature (9 sites)

| File | sites | what it needs |
|---|---:|---|
| `MeshWeaver.GitSync/ActivityRunner.cs` | 3 | The activity TITLE, each step's `message` and the terminal `finalMessage` arrive as `string` parameters of `RunActivity` / `Log` / `Finish`. The eight callers in `GitHubActivityExtensions` compose real sentences ("Commit {space} to GitHub", "Re-import {space} at {commitish}", "Open pull request", …), so these ARE translatable — they need an optional key parameter (or a `LogMessage` overload) on the runner. |
| `MeshWeaver.Compiler.Pipeline/NodeTypeCompilationActivity.cs` | 2 | `AppendLog(…, string message, …)` and `Update(…, string? error, …)`. The batched `AppendLogs(…, LogMessage[] …)` already accepts a keyed message, so the migration is at the ~40 CALLERS of `AppendLog`, one sentence at a time — no signature change here. |
| `MeshWeaver.Compiler.Pipeline/MeshNodeCompilationService.cs` | 3 | `AppendInfo` / `AppendWarning` / `AppendError(log, string message)`. Same shape: migrate the callers, or give the helpers a key parameter. |
| `MeshWeaver.Graph/StaticRepoImporter.cs` | 1 of 4 | The content-sync ledger's `string message` parameter. |

The two composed import SUMMARIES in `StaticRepoImporter` (the lock stamp and the terminal line) are
a judgement call rather than a mechanical migration: each folds 6+ conditional notes carrying #604 /
#3101 incident wording. Keying them well means keying each note and composing the translated pieces —
worth doing, but it is its own change with its own review.

### 2. GENERIC PLUMBING — nothing to key at the site

[...]

Issue #3236, opened 2026-09-03
ActivityLog messages are persisted English strings — a rendered surface that cannot be localized (53 sites, 0 localizable)


rbuergi:
## `ActivityLog.Messages` is a rendered surface that CANNOT be localized

`Doc/Architecture/Localization` states the rule as *"The portal renders its **chrome** in the
viewer's language"*, and `LocalizationTest` + the plugins-repo drift guard hold the catalogs
together. Activity transcripts fall outside both, not by exemption but **by construction**, and the
gap is invisible because nothing fails.

`LogMessage` is:

```csharp
public record LogMessage(string Message, LogLevel LogLevel)
```

A **rendered English string**, persisted on the node. There is no key, no argument list, and no
render-time resolution hook — so `AccessService.Localize` / `host.Localize` has nothing to resolve
and no place to run.

**Measured on `main`, 2026-09-03: 53 `new LogMessage(...)` sites across `src/`, of which 0 are
localized** — because none of them *can* be:

| File | sites |
|---|---:|
| `MeshWeaver.Mesh.Contract/MeshExtensions.cs` | 14 |
| `MeshWeaver.Graph/StaticRepoImporter.cs` | 10 |
| `MeshWeaver.Compiler.Pipeline/NodeTypeCompilationHelpers.cs` | 7 |
| `MeshWeaver.Data/WorkspaceOperations.cs` | 5 |
| `MeshWeaver.Hosting/Persistence/MonotonicWriteGuardStorageAdapter.cs` | 4 |
| `MeshWeaver.GitSync/ActivityRunner.cs` · `MeshWeaver.Compiler.Pipeline/MeshNodeCompilationService.cs` | 3 each |
| 7 more files | 1–2 each |

So a German viewer opening an import activity, a compile activity, a write-conflict activity or a
GitSync activity reads English today — every line of it.

## Why the write site cannot fix it

An activity message is written **server-side, at the moment the work happens, with no viewer in
scope**: the static-repo import runs as System at boot; a compile runs on a node hub; a write-conflict
record is raised inside a storage adapter. `AccessContext.Locale` belongs to whoever is *looking*,
and at write time there is nobody looking — and there may later be several viewers with different
locales reading the same stored row.

🚨 So this is not "somebody forgot a `Localize(...)` call at 53 sites". Resolving a locale at write
time would be **wrong even if it were possible** — it would freeze one viewer's language into a
shared record.

## What the fix has to be

[...]

PR #3305, opened 2026-09-04
i18n: key the activity transcripts a viewer actually reads (#3281)

rbuergi:
Part of #3281 — **the core half, which is all of the migration work**. Deliberately NOT `Closes`:
two items on that issue live in MeshWeaver.Plugins and are each blocked on something this PR cannot
do (see *What is left on #3281* at the bottom).

#3236 landed the `LogMessage` key+args seam and migrated 37 of 58 write sites. This finishes the
migratable remainder, so the ratchet's inventory drops **21 → 18** and every line left in it is
there because it *cannot* carry a key — not because nobody got to it.

## What a German viewer now reads in German

| Surface | Lines |
|---|---|
| All eight GitHub operations — commit, update, re-import, create branch, open PR, check branch, sync issues, merge | title **and** progress lines |
| The compile pipeline | timings, source-discovery report, cache hit, assembly outcome, failure |
| The static-repo import | progress lines and the content-sync ledger |

```
Branch 'main' steht auf a1b2c3d4 — Ihr Space ist aktuell.
Die Quellensuche fand 14 Code-Node(s): …
✔ Content-Assets sind synchron — 3 Datei(en) übertragen.
```

## How the string-plumbing sites were opened up

Each was a helper whose whole input is a `string message`, so the site never saw a *sentence* and had
nothing to key. Three shapes, three answers:

- **`ActivityRunner`** grew `RunActivity(…, LogMessage title, …)` and `ActivityContext.Log(LogMessage)`.
  The `string` spellings stay, as the deliberate un-keyed form.
- **`MeshNodeCompilationService.Append{Info,Warning,Error}`** take a **required** key. An optional one
  would leave the helper looking migrated while the next caller silently re-opens the hole — and the
  ratchet cannot see past a helper to the callers behind it. The one caller that genuinely cannot
  supply a key (Roslyn's own diagnostics) calls a separately named `AppendVerbatimWarning`, so the
  un-keyed case is a decision somebody wrote down rather than an omission.
- **`NodeTypeCompilationActivity.AppendLog`** needed no signature change at all: its three
  `StaticRepoImporter` callers moved to the batched `AppendLogs`, which already takes a `LogMessage`
  the caller built. (This is what the issue predicted, and it held.)

## Two things deliberately not done

- **A conditional clause gets its own key, never a `{suffix}` argument.** `", repository created"`
  spliced into a translated sentence is untranslated English sitting inside German word order — hence
  `commit.done` / `commit.doneRepoCreated`, and `storeUploadTimedOut` / `storeUploadFailed` rather
  than one key taking a composed `{reason}`.
- **Verbatim upstream text stays English** — `ex.Message`, a Roslyn diagnostic, GitHub's own refusal.
  Where it is spliced into a sentence the platform owns, the lead is keyed and the detail rides as
  `{detail}`. This is issue group 3, unchanged.

**Every English fallback is the sentence it was before, character for character**, so rows already in
the database and any un-migrated writer render exactly as they did.

## The cross-repo half — already landed first

Adding the `RunActivity` overload makes an unqualified `<see cref="ActivityRunner.RunActivity"/>`
**CS0419** in any dependent, which under `-warnaserror` is an error. That is break shape 3 in
`CrossRepoPairGate.md`: the pair gate cannot see it (it detects departures, not additions), and
because MeshWeaver.Plugins **pins** core, it would not have gone red on this merge — it would have
reddened whichever later PR happened to move the pin. MeshWeaver.SocialMedia hit exactly this on
2026-09-04.

So the three affected crefs were disambiguated in the dependent **first**: Systemorph/MeshWeaver.Plugins#1338.
A signature-qualified cref resolves against one overload as well as two, so that change is correct
both before and after this one.

`Pairs-with: none` — this diff removes **no** public type and **no** public member; it only adds
overloads. #1338 is a compatibility counterpart, not a paired departure.

## Handing over the i18n mirror

56 keys are added to `strings.{en,de}.json`. Per `Localization.md`, the client mirror's drift guard
compares against a **pinned** core commit, so nothing in MeshWeaver.Plugins goes red — and that is
exactly why the sync has to be handed over rather than waited for:

```
npm run sync:i18n -- --ref <this PR's merge commit>
```

## Verification

Locally, `-c Release -warnaserror`, one project per invocation:

- `MeshWeaver.GitSync`, `MeshWeaver.Compiler.Pipeline`, `MeshWeaver.Graph` — `0 Error(s)`
- dependents `MeshWeaver.Hosting`, `MeshWeaver.Mesh.Operations`, `MeshWeaver.Plugin.Build`, `MeshWeaver.PluginCatalog` — `0 Error(s)`
- `UnkeyedActivityLogMessageRatchetGuard` + `DocumentationLinkIntegrityTest` + What's New — **9 passed**
- `LocalizationTest` — **58 passed**

The ratchet earned its keep on the way: it caught a site where I had built the `LogMessage` on one
line and chained `.WithKey` on the next, which reads as unkeyed — correctly, because the English and
its key are then free to drift.

## What is left on #3281, and what each is blocked on

Neither is deferred by choice; both are gated on something outside this diff, so the issue stays
open with these named on it rather than being auto-closed with live work in it.

1. **The client i18n mirror** (`clients/react/src/i18n/`) — blocked on *this PR's merge commit*,
   because the sync takes it as an argument: `npm run sync:i18n -- --ref <sha>`. Nothing goes red in
   the meantime; the drift guard compares against a pinned core commit, which is exactly why the
   hand-over has to be explicit.

2. **The Plugins-side activity renderers** — blocked on a **core pin bump**. Three surfaces still
   render `m.Message` raw, so the keys added here will not reach them:

   | File | Line |
   |---|---|
   | `MeshWeaver.Blazor/Components/CompileProgressIndicator.razor.cs` | `msgs[^1].Message` |
   | `MeshWeaver.ContentCollections.Indexing.Graph/ContentIndexSettingsTab.cs` | `Esc(m.Message)` |
   | `MeshWeaver.Blazor.EntityViews/EditFormView.razor` | `@logMessage` — renders the record's `ToString()`, so this one is a display bug already |

   The fix is `m.Localize(viewerLocale)`, and it **cannot compile there yet**: measured against
   MeshWeaver.Plugins' `MW_PLATFORM_REF` (`b88be8a`, 2026-09-03), `src/MeshWeaver.Data/LogMessageLocalizationExtensions.cs`
   — the `Localize` seam #3236 introduced — **does not exist at that commit**. Core's own renderer
   (`GitHubSyncSettingsTab.ActivityHtml`) already calls `m.Localize(host.ViewerLocale())`, so the
   GitSync transcript this PR keys is live on merge; the three above follow the pin.


rbuergi:
## CI red, attributed — not this diff, and not re-run to find out

Two tests failed, in two shards:

| Shard | Test | Assertion |
|---|---|---|
| 4 | `MeshWeaver.Graph.Test.NackReachesTheWaiterDuringTeardownTest.OwnerDisposingUnderMeshTeardown_StillAnswersTheWaitingCaller` | — |
| 2 | `MeshWeaver.Messaging.Hub.Test.DisposalRaceNackTest.FaultingRequest_IsNacked_WhenTheHandlerRacesDisposal` | `Expected value to be ShuttingDown … but found NotFound` |

Both are teardown/NACK **classification** races. This diff adds catalog keys, doc comments and
`LogMessage` plumbing in GitSync / the compile pipeline / the importer; it cannot reach how a router
classifies a NACK against a disposing hub.

**Attribution, measured rather than assumed:**

- **#3300** — an unrelated diff (`fix/3249-stamp-failure-observable`) — fails on the **identical**
  shard-4 test. Whatever this is, it is not made by either diff.
- Both test files were last touched by the **same** main commit, `f92305ae1` *("Teardown lets accepted
  work finish and names what it had to stop — no forced teardown", 2026-09-04 11:49)*, which is in
  this branch's merge base.
- **#3302** — *"A reply routing cannot deliver goes to the waiter, not the floor — and a quiescing hub
  stops taking work it can never answer"* — is **green**. That is the fix for exactly this family,
  and the shard-2 assertion (`ShuttingDown` vs `NotFound`) is its subject sentence.

**So the plan is to wait, not to retry.** Per AGENTS.md a re-run would only hide a real race, and the
sanctioned response to *"CI fails on something your diff cannot reach"* is to merge `main` — once the
fix is actually in it. When #3302 lands I will merge main here and let CI re-measure.

Locally, `-c Release -warnaserror`, one project per invocation, all `0 Error(s)`:
`MeshWeaver.GitSync`, `MeshWeaver.Compiler.Pipeline`, `MeshWeaver.Graph`, `MeshWeaver.Hosting`,
`MeshWeaver.Mesh.Operations`, `MeshWeaver.Plugin.Build`, `MeshWeaver.PluginCatalog`,
`MeshWeaver.Graph.Test`, `MeshWeaver.Compiler.Pipeline.Test`. Guard + doc tests 9 passed;
`LocalizationTest` 58 passed.

🤖 Filed from Claude Code.

Issue #3249, opened 2026-09-03
CodeNodeType's last-execution stamp swallows its failure, and nothing can tell that it did


rbuergi:
Split out of Systemorph/MeshWeaver.Plugins#1256 (ask 3), which invited exactly this: *"Core seam (split out if preferred)"*. Asks 1 and 2 are fixed in Plugins (#1266, #1288); this is the remainder, and it belongs here.

## The seam

`CodeNodeType`, `src/MeshWeaver.Graph/Configuration/CodeNodeType.cs` ~line 380: after dispatching a run, the handler stamps `Last{ExecutedAt,ExecutedBy,ActivityPath,ExecutedCodeHash}` onto the Code node. The write is **fire-and-forget**, and both of its failure paths only log:

```csharp
.Subscribe(_ => { },
    ex => stampLogger?.LogWarning(ex, "CodeNodeType: stamp UpdateMeshNode failed for {Hub}", hub.Address));
…
catch (Exception stampEx)
{
    // The workspace might not be ready (cold-start race) — the missing stamp is a UI nicety,
    // not a correctness invariant, and the Activity log IS written.
    logger?.LogWarning(stampEx, "CodeNodeType: could not stamp last-execution fields on {Hub} …");
}
```

Then `ExecuteScriptResponse { Success = true }` is posted regardless — and it is posted **before** the stamp can have failed, so it could not report the failure even if it wanted to.

## Why the "UI nicety" reasoning no longer holds — and why it is now smaller, not larger

When that comment was written the stamp was the pane's **only** pointer to the run: `LastActivityPath` is what the cell's output pane keys on. A swallowed stamp therefore meant a `Succeeded` run that was **invisible until a reload** — the symptom Plugins#1256 reported (two runs, both Succeeded, nothing on screen, output present four minutes later).

Plugins#1288 removed that dependency: the pane now prefers the activity path carried on `ExecuteScriptResponse.ActivityLog`, falling back to the stamp. So the stamp is genuinely a nicety again — **for the pane**. What remains is narrower and still real:

- **"Last executed: `<when>` by `<who>`"** silently goes stale or missing;
- `LastExecutedCodeHash` is what drives the **stale-cell** indicator (`IsOutputStale`), so a swallowed stamp can leave a cell reading "up to date" against code it did not run — a wrong claim rather than a missing one;
- nothing anywhere can distinguish "this cell has never been run" from "this cell ran and we failed to record it".

[...]

PR #3318, opened 2026-09-04
A cell finds its own runs without the last-execution stamp (#3301)

### LINKED ISSUE #3301: A cell cannot find its own runs without the last-execution stamp
Split out of #3249, which is otherwise finished (#3270 made the currency verdict fail closed; #3300
made a swallowed stamp reported rather than silent, and argued why the two "put a marker on the
cell" shapes are unsound).

## What is left

A cell reads its own run history from **one denormalised pointer**: `CodeConfiguration.LastActivityPath`,
written by the last-execution stamp. When that stamp does not land, the pointer is not there, and on
the next page load the cell is indistinguishable from a cell nobody has ever run — even though the
run happened and its record is durable.

That last part is the point. **The run is not lost.** `CodeNodeType.HandleExecuteScript` creates the
Activity node *before* it dispatches, and stamps the originating cell onto it:

```csharp
Content = new ActivityLog("ScriptExecution")
{
    Id = activityId,
    HubPath = hub.Address.Path,   // ← the Code node
    Status = ActivityStatus.Running
}
```

So the cell → run edge exists in the data; it is only missing in the direction the view reads it.
Nothing today follows it back.

## Why this is not closed by #3300

#3300 deliberately reports the stamp failure to an **operator** (one `Error`, event id `3249`) and
argues that recovering it on the cell is not possible from the dispatching hub: every scheme that
writes a marker onto the cell is a second write to the node whose first write just failed. That
reasoning is written up in `Doc/Architecture/ScriptExecution` → *"When the stamp does not land"*.

It says nothing about the **viewer**, because the fix for the viewer is not a write at all — it is a
read that does not depend on the stamp.

## Shape of a fix

A surface that answers *"has this cell been run?"* from a listing over `_Activity` filtered on
`ActivityLog.HubPath`, falling back to the stamp rather than depending on it. That is a valid CQRS
use — a listing by predicate, where a stale negative is harmless — and it is the same fallback
shape MeshWeaver.Plugins#1288 already applied to the output pane, which prefers
`ExecuteScriptResponse.ActivityLog` over the stamp for the *current* session.

Two halves, and they are in different repos:

- **core** — `ActivityLog.HubPath` and `CodeConfiguration` live here, and if the lookup is worth
  making canonical (rather than re-derived per view) it belongs beside
  `CodeOutputCurrencyExtensions.OutputCurrency`, which is already the one rule every surface uses.
- **MeshWeaver.Plugins** — `CodeViews.IsOutputStale` is the toolbar that renders the verdict.

## Cost, stated

A per-cell activity listing is not free on a notebook with many cells. Whether this is worth a query
per cell, a batched one per page, or only a lazy check when a cell would otherwise render
`NeverRun`, is the design question — and "only when it would say NeverRun" is probably the answer,
since that is the one state where the lookup can change the outcome.

## Not in scope

The stamp's own failure reporting (#3300) and the currency rule (#3270) are done.

rbuergi:
Closes #3301.

`Pairs-with: none` — this PR removes no public type and no public member. It **adds** one type
(`CodeRunHistory`) and two constants; nothing outside core is asked to change, and the two names are
free in MeshWeaver.Plugins (`gh api search/code … repo:Systemorph/MeshWeaver.Plugins` → `total_count 0`,
measured 2026-09-05). No `<see cref>` becomes ambiguous: `ResolveOutputCurrency` is a new NAME, not a
new overload — deliberately, because #3305's `RunActivity(…, LogMessage)` overload turned an
unqualified cref into `CS0419` in three Plugins files (Plugins#1338) and that break shape has no gate.

## The defect, and why #3300 did not close it

A cell reads its own run history from ONE denormalised pointer: the last-execution stamp
(`LastExecutedAt` / `LastExecutedBy` / `LastActivityPath` / `LastExecutedCodeHash`), written
fire-and-forget after the dispatch. It is one write to one node — it lands whole or not at all — so
when it fails (#3249: workspace not ready, a refused partition write, content unreadable) the cell is,
after the next page load, **indistinguishable from a cell nobody has ever run**.

#3300 made that failure *reported* (one `Error`, event id `3249`) and wrote up why recovering it on
the cell is impossible from the dispatching hub: all three candidate shapes are a **second write to
the node whose first write just failed**, i.e. recovery in the same failure domain. That argument is
correct and stands. It simply says nothing about the **reader** — because the reader's fix is not a
write.

**The run is not lost.** `CodeNodeType.HandleExecuteScript` creates the Activity node *before* it
dispatches and stamps the originating cell onto it (`ActivityLog.HubPath = hub.Address.Path`). The
cell → run edge is durable; it was only missing in the direction the view reads it. Nothing followed
it back.

## The fix — one read, along the edge that already exists

`CodeRunHistory` (`src/MeshWeaver.Mesh.Contract/CodeOutputCurrency.cs`, beside the currency rule it
extends, as the issue asked):

| step | | cost |
|---|---|---|
| 1 | `OutputCurrency()` — the stamp. Anything but `NeverRun` returns unchanged. | **no query** |
| 2 | Only on `NeverRun`: one listing for an Activity naming this cell on `HubPath` → `Unverified`. | one row |
| 3 | No hit → `NeverRun`, exactly as today. | — |

- **A listing by predicate, not a point read.** A stale negative is harmless by construction — the
  worst outcome is the answer the cell already gives today. Reading a *specific known node* stays
  `GetMeshNodeStream`'s job.
- **A recovered run is `Unverified`, never `Current`.** The Activity records *that* the cell ran,
  never *what* it ran. Same evidence as a stamp that landed without its fingerprint, so the same
  fail-closed verdict — claiming currency here would be the wrong claim the #3249 rule exists to
  prevent.
- **It runs only where it can change the outcome.** The issue's cost paragraph asked for exactly
  this: a notebook of normally-stamped cells costs **zero** queries, because step 1 answers first.
- Reactive end to end (`IObservable<CodeOutputCurrency>`, one verdict then completion) — no `async`,
  no `Task<T>`, no `.Result`. Immutable collections throughout (`ImmutableArray<string>`).

**Where it looks.** `ActivityNamespaces` reconstructs the layers visible from the cell alone — its own
`ActivityParentPath` (with `{viewer}` expanded), the partition root, the reading viewer's home —
deduplicated into ONE union query, usually one namespace. Stated in the code and the doc: a
partition-level `DefaultActivityParentPath` pointing at a fourth place is **not** covered (that lookup
is a live partition-registry query, which `Mesh.Contract` sits below). Where it applies the listing
finds nothing and the verdict falls back to `NeverRun` — today's answer — so the gap costs nothing
that was already working.

## Two cross-assembly agreements, both pinned — because no compiler sees them

The lookup filters on `nodeType:Activity` and expands `{viewer}`, and both are declared in assemblies
it sits *below*. A rename on either side would leave the query looking for something nothing writes:
zero rows, every recovered cell silently back to `NeverRun`, **no compiler error anywhere**.

- `CodeRunHistoryTest.TheActivityNodeTypeNameMatchesTheGraphVocabulary` — equality against
  `GraphNodeTypeNames.Activity`.
- `CodeRunHistoryTest.TheViewerSentinelExpandsTheSameWayOnBothSides` — pinned *behaviourally*: it
  drives `CodeNodeType.ResolveActivityParent` and requires the answer the lookup assumes. The
  dispatcher now spells the sentinel through `CodeRunHistory.ViewerHomeSentinel` instead of a second
  literal.

## Verification

**Deterministic repro, through a real mesh** —
`CodeCellCurrencyThroughTheMeshTest.ARunWhoseStampNeverLandedIsStillFoundByTheActivityItWrote`. Two
cells really run through `ExecuteScriptRequest`, so two real Activities share one `_Activity`
namespace; one cell's stamp is then WIPED, reproducing exactly what #3249's failure paths leave
behind. Four arms, so the test cannot pass vacuously: the wiped cell reads `NeverRun` from its own
content (the defect, asserted); it resolves to `Unverified` (the fix); a cell that genuinely never ran
still resolves to `NeverRun` (non-vacuity — and the neighbour's activity in the same namespace must
not be mistaken for it); the still-stamped neighbour resolves to `Current` without a query.

**Mutation-proved, twice — the guard was made to fail on purpose and quoted:**

1. Drop `content.hubPath:{cellPath}` from `RunsQuery` → 2 failures:
   > Expected value to be NeverRun because no Activity anywhere names this cell — and the neighbour's
   > run, which lives in the very same `_Activity` namespace, must not be mistaken for it. … **but
   > found Unverified.**

   > Expected "namespace:acme/_Activity nodeType:Activity sort:LastModified-desc limit:1" to contain
   > "content.hubPath:acme/notebook/cell-7" because HubPath is the cell → run edge the dispatcher
   > stamps onto the Activity node BEFORE it dispatches…

2. Make `ResolveOutputCurrency` never consult the mesh (the pre-fix behaviour) → 1 failure:
   > Expected value to be Unverified because the Activity node created before the dispatch still names
   > this cell on HubPath, so the run is not lost — only the cell's pointer to it is. … **but found
   > NeverRun.**

Both mutations reverted; the suite is green again.

**Builds — `dotnet build -c Release -warnaserror`, one project per invocation, all `0 Error(s)` / `0 Warning(s)`:**
`MeshWeaver.Mesh.Contract` · `MeshWeaver.Graph` · `MeshWeaver.Documentation` · `MeshWeaver.Graph.Test` ·

**Tests:** `MeshWeaver.Graph.Test` currency + run-history suites — 30/30 passed.
`DocumentationLinkIntegrityTest`, `WhatsNewEntryIntegrityTest`, `UnkeyedActivityLogMessageRatchetGuard`
— 9/9 passed. Full `MeshWeaver.Documentation.Test` guard suite green.

## Work products committed

- `Doc/Architecture/ScriptExecution` → new subsection *"Recovering the run for the READER — follow the
  edge, do not write another marker"*, replacing the "what is still not covered" paragraph that
  pointed at this issue. It states the cost model, the namespace derivation, the two cross-assembly
  agreements, and what remains uncovered.
- What's New entry (`Category: Fix`): *A cell that ran no longer claims it never did*.

## Not in scope

- The output pane is deliberately **not** re-pointed at a recovered activity — this answers *whether*
  the cell ran, not *what it produced*.
- The operator still gets the `3249` `Error`. The stamp failing is a real fault; a reader who can no
  longer see the consequence is not a reason to stop reporting it.
- The Plugins half (`CodeViews.IsOutputStale` rendering the recovered verdict) follows a core pin
  bump — `MW_PLATFORM_REF` is `7d644de95` and this change is not in it yet.

### DIFF
--- a/src/MeshWeaver.Documentation/Data/Architecture/ScriptExecution.md
+++ b/src/MeshWeaver.Documentation/Data/Architecture/ScriptExecution.md
@@ -161,7 +161,36 @@ Two things make that an honest answer rather than a shrug.
 
 `Error` rather than `Warning` is deliberate and costs nothing extra: both levels ship to Loki and the path is exceptional, so the only thing that changes is that it trips error-rate alerting. That is the right classification — nothing retries the write, the loss is permanent, and the visible consequence is a cell that tells a human it has never run. "Degraded but self-correcting" is what `Warning` claims, and none of it holds here.
 
-What is still **not** covered: a viewer looking at a cell after a reload cannot tell `NeverRun` from "ran, not recorded". Closing that needs a surface that finds a cell's runs by `ActivityLog.HubPath` rather than by the stamp — a read that does not depend on the stamp, not another write. See [issue #3301](https://github.com/Systemorph/MeshWeaver/issues/3301).
+#### Recovering the run for the READER — follow the edge, do not write another marker
+
+The three candidates above all fail for the same reason: they are **writes**. What the reader needs is a **read**, and the edge to read along already exists. `CodeRunHistory.ResolveOutputCurrency` (`MeshWeaver.Mesh.Contract`, beside the currency rule it extends) walks it:
+
+```csharp
+// The verdict a cell can substantiate when its own stamp cannot supply one.
+// Reactive like every other read — one verdict, then completion.
+code.ResolveOutputCurrency(cellPath, viewerHome, meshService)
+    .Subscribe(verdict => { /* NeverRun · Unverified · Stale · Current */ });
+```
+
+| Step | What happens | Cost |
+|---|---|---|
+| 1 | `OutputCurrency()` — the stamp. Anything but `NeverRun` is returned unchanged. | **no query** |
+| 2 | Only on `NeverRun`: one listing for an Activity naming this cell on `HubPath`. A hit ⇒ `Unverified`. | one row |
+| 3 | No hit ⇒ `NeverRun`, exactly as today. | — |
+
+Three properties are what make this the right shape rather than merely a working one.
+
+**It is a listing by predicate, not a point read.** `Query` is eventually consistent, so it may never decide what a *specific known node* contains — that stays `GetMeshNodeStream`'s job. Here a stale negative is harmless by construction: the worst outcome is the answer the cell already gives today.
+
+**A recovered run is `Unverified`, never `Current`.** The Activity records *that* the cell ran, never *what* it ran — no fingerprint travels with it. That is the same evidence a stamp which landed without its hash leaves, so it gets the same fail-closed verdict, for the same reason.
+
+**It runs only where it can change the outcome.** A notebook of cells that ran normally costs zero queries, because step 1 answers first. This is the whole reason the lookup hangs off `NeverRun` rather than being a general "find this cell's runs" call every view makes.
+
+**Where it looks.** The run lands under `{activityParent}/_Activity`, and `CodeNodeType.ResolveActivityParent` picks that parent in three layers: the Code node's own `CodeConfiguration.ActivityParentPath`, then the partition's `PartitionDefinition.DefaultActivityParentPath`, then the partition root — with the `{viewer}` sentinel at either configured layer expanding to the calling user's home. `CodeRunHistory.ActivityNamespaces` reconstructs the layers it can see from the cell alone — the node's own `ActivityParentPath` with `{viewer}` expanded, the partition root, and the reading viewer's home — deduplicated into a single union query, usually one namespace. A partition-level `DefaultActivityParentPath` pointing somewhere else is **not** covered: that lookup is a live query on the partition registry, which `MeshWeaver.Mesh.Contract` sits below. Where it applies the listing finds nothing and the verdict falls back to `NeverRun` — the answer the cell gives today, so the gap costs nothing that was already working. A caller that has already resolved the parent (the dispatcher has, via `CodeNodeType.ResolveActivityParent`) skips the derivation and calls `CodeRunHistory.RunsQuery` with the namespace it resolved.
+
+🚨 **Two cross-assembly agreements hold this together, and neither is visible to a compiler.** The lookup filters on `nodeType:Activity` and expands the `{viewer}` sentinel — both declared in assemblies it sits *below* (`GraphNodeTypeNames.Activity`, `CodeNodeType.ResolveActivityParent`). A rename on either side would leave the query looking for something nothing writes: zero rows, every recovered cell silently back to `NeverRun`, no compiler error anywhere. `CodeRunHistoryTest` pins both — the node-type name by equality, the sentinel by *driving the dispatcher's own resolver* and requiring the answer the lookup assumes.
+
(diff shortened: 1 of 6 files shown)


rbuergi:
The Plugins half — `CodeViews.IsOutputStale` rendering the recovered verdict — is tracked as Systemorph/MeshWeaver.Plugins#1345. It follows a core pin bump past this PR (`MW_PLATFORM_REF` = `7d644de95`, core 2026-09-04 11:07Z), not effort. That issue also records a second drift found while measuring: the toolbar still re-derives the currency rule by hand instead of calling `OutputCurrency()`, which #3270 introduced so that no surface would.

Issue #2194, opened 2026-08-25
RCA 2026-08-25: sync/bake recompile waves fatten type-hosting silos to 20 GB — superseded builds stay rooted behind the Recycle banner, and a degraded pod never leaves rotation


rbuergi:
## Impact

memex.meshweaver.cloud (ns `memex-cloud`), 2026-08-25 ~06:10–~09:50 UTC: intermittent page hangs (Store and others) whenever a request landed on one of two degraded portal pods. Reported by the maintainer as "something must be fundamentally fucked up — like someone bakes on the mesh or it's wedged". Both halves of that guess were right: the mesh WAS repeatedly recompiling, and the pods it happened on WERE wedged — while reporting Ready the whole time.

## What happened (measured)

`container_memory_working_set_bytes`, the two sick pods (`wns9j`, `8zlfh`, born 01:26 UTC on the ci.5285 roll):

```
8zlfh 01:30..06:10 flat 2.7G │ 06:30=3.8 07:10=7.0 07:30=9.4 08:30=12.8 09:30=17.2
wns9j 01:30..05:50 2.7→4.5G │ 06:30=7.9 07:10=9.7 08:10=18.5 08:50=20.4 09:30=18.1
```

- **Five flat hours, then a hard inflection at 06:10–06:30** — immediately after the 05:55–06:08 **scheduled Plugin Catalog CI bake tick**, the first tick after Plugins#634 bumped `MW_IMAGE_DIGEST` (new framework identity ⇒ full-catalog rebake ⇒ bundle publication ⇒ portal adoption ⇒ recompile of every packaged type).
- Climb of ~2.5 GB/h through five more sync events: Plugins#635 (06:51), #637/#638/#639/#640 (08:19–08:56) — each merge auto-synced into the portal and recompiled again.
- Both pods pegged 3.7–4.2 cores (GC), peaked 17–20 GB. **Only the two silos hosting the compiling type hubs fattened** — replicas born later stayed ≤3.6 GB.
- **No OOMKill** (28Gi limit never reached), **no probe failure** (readiness AND liveness probe `/alive`, a trivial process-up check a GC-thrashing pod still answers), **no alert** — so both pods stayed in the Service endpoints for ~3.5 h serving hung/garbled requests until a human noticed.
[...]

PR #3319, opened 2026-09-04
fix(observability): alert on the GC ceiling, not only on replica divergence

rbuergi:
## What this is

#2194's fix table has kept items 4 (progress-aware `/alive`) and 5 (working-set divergence alert) open since 2026-08-25, and the GC spiral recurred on 09-03 **with fixes 1 and 2 both live in production**. The maintainer's last comment posed the discriminator: is convergence enabled-but-not-converging, or does a second cause share the curve?

Measured today, read-only, against `memex` and `memex-cloud`. It is neither of those two — it is a third answer, and it changes what items 4 and 5 have to be.

## The discriminator, settled

Counted over one replica's **entire 28-hour life** (186 137 lines), with `info:`-level lines confirmed to be shipping (2 791 of them) so a zero is a real zero and not a log-level artefact:

```
Stale-build offer:                          0
Stale-build convergence:                    0
EnrichWithNodeType: self-heal recompile #   0
BatchBake:                                  0
KEEPING its load context                    0
NOT unloading collectible context           0
Grain … deactivating: reason=               0
```

**No build was ever superseded.** `Modules:AutoRecycleOnStaleBuild` is armed on both namespaces and has never fired, because nothing has ever been stale. Fixes 1+2 could not have prevented the recurrence — they address a state that does not arise. And `deactivating: reason=` is **0 across five sampled replicas and 514 687 lines**: not one hub or grain deactivated since boot, so no ALC lease is ever released.

The compile load is a *coverage* failure, not a publication wave — the two oldest replicas booted having adopted **8 and 9 of 377** dynamic NodeTypes, with the other 368 due to "pay a Roslyn compile on FIRST ACCESS", lazily, spread over the pod's whole life. That is fix 3's lane (`Modules:RequirePrebuilt`), not fix 1's.

## Why nothing ever restarts the pod

Nothing sets `DOTNET_GCHeapHardLimit[Percent]` — not the inline `env:`, not `memex-portal-config`, not any envFrom Secret (audited by key name on both namespaces, values never read). So .NET's container default applies: the GC's heap hard limit is **75 % of the container limit = 12 GiB** at `limits.memory: 16Gi`, and it covers the **managed heap only**.

The runtime defends that ceiling with back-to-back blocking gen2 instead of letting the kubelet restart the container. Measured: **0 OOMKills across 7 replicas over 28 h** (the only two restarts were `exitCode 255`/`Unknown` at one shared timestamp — a node event), and GC pause share **0.66 sustained for 2 h 15** on 09-03 while `/alive` answered `Healthy`. The GC converts a memory shortage — which Kubernetes fixes in one second — into CPU starvation, about which Kubernetes has no opinion.

The heap growth itself is retention, not a working set: **6 856 → 10 065 MiB in 14.9 h (~215 MB/h) across 267 gen2 collections**, with the floor after each collection never returning to any earlier level. And it tracks uptime rather than load — the near-idle replica (512 log lines/30 min) holds 11 443 Mi while the busiest one (2 476 lines) holds 4 134 Mi.

## Item 5 — a detector that goes blind exactly when it matters

`PortalReplicaWorkingSetDiverged` requires `max > 3 × min`. On `memex` right now:

```
memex-portal-deployment-77898c4947-6mffn   195m   9936Mi
memex-portal-deployment-77898c4947-bfnft   167m   9409Mi     → ratio 1.06
```

Both 28 h old, both at ~80 % of the 12 GiB GC ceiling, both on the curve that produced the 09-03 outage — and the alert sits at a third of its threshold and keeps falling as they converge. Its comment argued *"a namespace whose replicas all sit at 12 GB is a sizing question, not this incident"*; that is the premise this measurement falsifies.

**This is not fixed by lowering the 3×** — a rollout legitimately runs a cold pod beside warm ones, and the rule earns its keep on the asymmetric case (it read 7.22× correctly on 09-01). The terminal case needs an absolute reading, which a ratio cannot express at all. So the divergence rule is kept as the early signal and `PortalHeapEnteringGcDefenceBand` is added for the other end: per namespace, no sibling comparison, thresholded at 75 % of the container limit — the point at which working set has reached the numeric value the GC is willing to let the managed heap alone occupy.

**Verified against the live Prometheus before committing**, because an alert that cannot fire is worse than no alert. `count(container_spec_memory_limit_bytes{container="memex-portal"})` → **9 series**, and both rules on one scrape:

| namespace | divergence `max / min` (fires > 3) | headroom `working set / limit` (fires > 0.75) | actually degraded? |
|---|---|---|---|
| `memex-cloud` | **8.45** — fires | **0.851** — fires | yes: 14 GB replicas, `Platform stalled` in the log |
| `memex` | 1.06 — blind | 0.605 — quiet | not yet: 9.4–9.9 GB, ~0.2 cores, no stalls |

**Four results, not two.** Both rules fire on the degraded namespace and both stay quiet on the healthy one — so neither fires on everything, and the new rule is not a restatement of the old. The complementarity is the first column: `memex`'s ratio is 1.06 and will **fall, not rise**, because its replicas are converging. When they reach the defence band, divergence still reads ~1.0 and only the new rule fires. That is the uncovered case.

🚨 Neither rule is armed until `values.observability.yaml` is applied. On 09-01 the live Prometheus held **zero rule groups** with the `loki` release still on its 31 May revision, so the divergence alert had been merged for a week and was watching nothing. `git grep` is evidence of authorship and never of arming; only `GET /api/v1/rules` naming the alert is.

## Item 4 — the chart contradicted itself, and the real constraint is new

`deployment.yaml` carried both of these, ten lines apart:

> readiness: *"a cheap progress-aware check is an open design item on #2194"*
> liveness: *"🚨 `/alive` MEASURES PROGRESS … It answers 503 once GC pause share has stayed above 60 % for five consecutive checks"*

The 09-04 status comment quoted the stale half as current. What `/alive` measures is **image-dependent** and this repo cannot assert it: core registers only the trivial `"self"` check on tag `live`; a progress-aware handler comes from the host (`Memex.Portal.Distributed`, in MeshWeaver.Plugins).

The consequence that is *not* recorded anywhere, and which bounds item 4:

**Readiness and liveness share the `/alive` path, and readiness trips first** — `10s × 3 = 30s` against liveness's `15s × 6 = 90s`. The moment `/alive` becomes progress-aware, a GC-bound replica is pulled from the Service **three times sooner than it is restarted**, and its traffic lands on siblings converging on the same ceiling (ratio 1.06, above). That is the 2026-07-21 readiness death-spiral, rebuilt out of the item-4 fix.

⇒ A progress-aware liveness signal needs its **own path**, or readiness must move off this one. Both comments now say so instead of contradicting each other.

## Also recorded

`AddRuntimeInstrumentation()` is on, so the portal already emits `dotnet_gc_*` — heap size, pause time, per-generation counts. The collector's metrics pipeline exports **only** `file/metrics`, a rotating JSON file on the node; logs and traces have a second path to Loki, metrics have none. That is why the only memory signal any alert can use is a kubelet-side working-set number that cannot see the GC ceiling at all, and why the new threshold is a proxy rather than a direct read of the heap.

## Verification

- `deploy/aks/scripts/check-chart-invariants.sh` → `All 3 values combinations render a self-consistent deployment.`
- `deploy/aks/scripts/check-values-are-read.sh` → `All 114 config key(s) across 3 values file(s) are read by a template`
- `dotnet build src/MeshWeaver.Documentation/MeshWeaver.Documentation.csproj -c Release -warnaserror` → `0 Warning(s) 0 Error(s)`
- `DocumentationLinkIntegrityTest` → `Passed! Failed: 0, Passed: 1`, `.trx` written.
- **Mutation-proved** that the link test actually sees the new page: adding `[Deliberately broken](../ThisPageDoesNotExistXYZ)` turned it red with
  `Doc/Architecture/WhyAGcBoundPodStaysInRotation: (../ThisPageDoesNotExistXYZ) — resolves to /Doc/Architecture/ThisPageDoesNotExistXYZ, which does not exist`
  and removing it turned it green again.
- Alert YAML re-parsed after every edit; both rules present in group `portal-degraded`.

**No cluster object was mutated.** All cluster reads were `kubectl get`/`top`/`logs`, the Prometheus query API, and one `/alive` GET through the API-server proxy. No secret value was read, printed or stored — the GC-override audit projected Secret data to **key names only**.

### Follow-up commits on this branch

- The alert's **fail-loud** edge behaviour stated: if a container's memory limit is ever removed, cAdvisor reports the denominator as 0, the ratio becomes `+Inf` and this fires. That is deliberate — with no container limit .NET applies no GC heap hard limit either. The obvious "fix" (`and container_spec_memory_limit_bytes > 0`) would silence the rule on exactly the misconfiguration it should shout about.
- `NodeTypeCompilation.md`'s in-memory retention section read as though convergence closes the gap; it now carries the 09-04 correction and links the new page, so the correction is discoverable from where the next reader actually starts.

Refs #2194 · successor RCA #3321


rbuergi:
Your shard-5 red is not yours — `ComboGateRollTest.ARedVerdict_BlocksTheRoll_AndNamesTheModuleThatWouldBreak` is a read-after-write race in that test, and this PR's diff (two YAML files, two Markdown docs) cannot reach `Memex.Portal.Shared.Test` at all.

Cured in #3324: `Record` awaited `RecordVerification(...).FirstAsync()`, but a completing `stream.Update` says the patch was ACCEPTED, not that every reader can see it — so the gate read the policy first and honestly reported "no combo verification has been recorded", and the assertion failed on text that was never about a defect. It now waits until the verdict is readable, on the same predicate a real reader uses.

Second sighting today (the first was run 33863349195, a different method, same cause). Once #3324 lands, refresh this PR's base — a re-run replays the same commit and will not pick the fix up.

Issue #3203, opened 2026-09-03
Localization: viewer-locale chrome inside authored English content yields mixed-language pages (Run = 'Ausführen', quiz = 'Question 1 of 6'); one rule for in-content controls is needed


rosalie1984:
## Observed

On an English course page (AgenticEngineering, Module 2 exercise "Retry Backoff Schedule", a learner's installed copy on the memex-cloud deployment) the code cell's Run button reads **Ausführen**. Everything the author wrote around it is English. The reader's reaction: "language should be applied as consistently as possible".

The same viewer, same course, on another instance: the node menu reads "Bearbeiten · Anheften · Verschieben · Löschen", the quiz on the same page reads "Question 6 of 6" and "This is a copy in your own space…", the lesson text is English.

## Why it happens (all by design, each piece on its own)

- The viewer's language is `User.Locale`: populated once from the browser (`LocalePreference.PopulateOnce`, write-once), changeable in settings ("choose your language", 2026-08-17). It rides on `AccessContext.Locale`.
- Every label that goes through `AccessService.Localize(key)` renders in that language; the catalog ships `en` and `de`. `common.run` → "Run" / "Ausführen"; `code.staleCell` → "Code changed — re-run" / "Code geändert — erneut ausführen".
- The Edu pack applies the same doctrine to its own chrome ("Chrome is the platform's voice, so the VIEWER's language wins", `EduTexts.ResolveChrome`); the course's declared `language` is only a fallback for a viewer who states none.
- Authored content stays in the author's language, and pack strings that never went through a table (the quiz's progress, feedback, score and notices; a dialog title in the cell views) stay English.

So a German-profile reader of an English course gets: platform chrome German, Edu chrome German where a translation exists, quiz English, content English. The rule is consistent per component and inconsistent per page, and the page is what the reader sees.

## Ask

1. **Decide one rule for controls embedded in authored content** and write it into `Localization.md`. Two coherent options:
   - (a) In-content controls (cell toolbars, quiz, exercise grid, prompt cells) follow the **content's declared language** (course root `language`, or the page's), and fall back to the viewer's locale only when the content declares none. Portal chrome (menus, settings, notifications) keeps following the viewer. A single German word inside an English lesson stops happening.
   - (b) Everything that is chrome follows the viewer, in which case *every* string on a content page must go through the seam and ship in every supported language, and the catalog needs a completeness check so a missing key cannot fall back to English mid-page.
[...]

PR #3323, opened 2026-09-04
Localization: the kernel placeholders get keys, a locale and a guard; the doc stops asserting the guard's absence (#3203)

rbuergi:
The **core half** of #3203 — the part that unblocks the rest. The reported symptom (a German
*Ausführen* in the middle of an English paragraph) was already fixed by MeshWeaver.Plugins#1318;
this is the three things #3203's 08:54Z comment says remain, all of which live in this repo.

## 1 · The kernel placeholders take a locale

`MarkdownViewLogic.DisableKernelPlaceholder` and `PendingKernelPlaceholder` rendered *"Interactive
code execution is unavailable here…"* and *"Starting interactive kernel…"* as English literals
**inside the cell frame**, and accepted no locale — so the two Blazor call sites in
MeshWeaver.Plugins had nothing to pass `ViewerLocale()` into. Under clause 1 of
[Chrome and content language](https://github.com/Systemorph/MeshWeaver/blob/main/src/MeshWeaver.Documentation/Data/Architecture/ChromeAndContentLanguage.md)
these are platform-**owned** words rendered in an author's flow, so they follow the VIEWER; a bare
literal there is the "unowned string" that clause calls a bug.

The catalog value is HTML-encoded into the notice — a translation is data going into markup, and
encoding is what stops a translator being able to alter the markup around their own sentence.

## 2 · Seven new catalog keys, in both languages

| key | en | de |
|---|---|---|
| `code.kernelDisabledNotice` | Interactive code execution is unavailable here — … | Interaktive Code-Ausführung ist hier nicht verfügbar — … |
| `code.kernelStarting` | Starting interactive kernel… | Interaktiver Kernel wird gestartet… |
| `code.saveFailed` | Save Failed | Speichern fehlgeschlagen |
| `code.codeFiles` | Code Files | Code-Dateien |
| `code.loadingCode` | Loading code… | Code wird geladen… |
| `code.enterDisplayName` | Enter display name… | Anzeigenamen eingeben… |
| `code.neverExecuted` | never executed | nie ausgeführt |

🚨 **Verified, not assumed: none had an existing key.** The issue asserts it; I checked each of the
five `CodeViews` literals against `strings.en.json` before adding. The nearest miss is
`ui.noCodeFiles` = *"No code files"* — the empty state, a different string from the *"Code Files"*
nav-group title. `Node` and `Kernel` stay English in the German values (glossary terms), matching
the existing `code.kernelUnavailable`.

The last five are consumed in MeshWeaver.Plugins#1308. They are here because a key must exist in
core before Plugins can read it, and shipping them separately would render English for German
viewers in between.

## 3 · The doc stops asserting a guard's absence

`ChromeAndContentLanguage`'s **Adoption** section still listed the label removal and the guard as
*Outstanding*, and **Enforcement** still read *"NOT YET BUILT — do not read this paragraph as a
guard that exists."* Both became false when MeshWeaver.Plugins#1318 merged. This is the inverse of
the decay that page itself documents (`AnonymousCircuitLocaleSeedTest` was cited as coverage for
months after a repo move lost it), so it is corrected rather than left to rot in the other
direction. A stale key count in the same section (*"1104 keys each"*) is replaced with a statement
that derives itself.

## 🚨 Binary compatibility — overloads, and the blind spot the pair gate has

`DisableKernelPlaceholder`, `PendingKernelPlaceholder` and `RenderKernelResultAreas` each gain a
**locale-carrying overload**; the existing signatures stay and forward. Adding a parameter to an
existing public method is a binary break — an already-compiled module bound to the old signature
stops loading.

An added overload has its own cross-repo hazard, and it is one AGENTS.md records the pair gate
**cannot see**: it can make a dependent's unqualified `<see cref>` ambiguous — **CS0419 under
`-warnaserror`**, exactly how core #3305 hit MeshWeaver.SocialMedia and MeshWeaver.Plugins#1338.
Both repos were swept:

- **Core** had two such crefs. They are qualified here, and this was *measured*: the first build
  after adding the overload failed with
  `error CS0419: Ambiguous reference in cref attribute: 'RenderKernelResultAreas'`.
- **MeshWeaver.Plugins** has one, at `src/MeshWeaver.Markdown.Test/KernelSubscribeGateTest.cs:20`.
  It is safe, and **proven rather than argued**: that project measures
  `GenerateDocumentationFile=false` (a controlled repro confirms CS0419 needs a doc file), and
  `MeshWeaver.Markdown.Test` at `origin/main` `d3b5ae01` **compiles against this core** with
  `-warnaserror` → `0 Warning(s) / 0 Error(s)`, against a byte-identical `MeshWeaver.Markdown.dll`
  (md5 `1da7e76bf294827eb469395d04462a81` on both sides).
- The two `RenderKernelResultAreas` **call sites** in Plugins (`MarkdownView.razor.cs:302`,
  `CollaborativeMarkdownView.razor.cs:578`) bind to the kept 4-arg overload by arity — no ambiguity
  is possible — and the compile above covers them.

**No `Pairs-with:` is declared because nothing public was removed**, and that is measured too:
`check-type-forwards.py --surface-json` reports `publicTypesAtBase: 1885`, `publicMembersAtBase:
11782`, **`removed: []`**; the judging mode reports *"OK — no unguarded public-type move."*

The Plugins pin is already past the blocker #3203 named: `MW_PLATFORM_REF` reads `7d644de95` in
`.github/workflows/ci.yml` (line 148, no overriding repo variable), and #3276 is closed.

## The guard, mutation-proved

`test/MeshWeaver.Documentation.Test/KernelPlaceholdersFollowTheViewerGuard.cs` — the core half of
clause 2's **enumerated** in-flow set. `InFlowChromeClause2Guard` (Plugins) covers the three
surfaces declared there; a core test cannot see a Plugins component, which is why the set is
guarded in two halves rather than one.

Broken on purpose three ways, each failing by naming the defect:

1. **Hard-coded English literal** → 2 failures, `de` fails while `en` still passes:
   > Expected `"…<div class="markdown-kernel-disabled">Interactive code execution is unavailable here — …"` to contain `"Interaktive Code-Ausf&#252;hrung ist hier nicht verf&#252;gbar — …"` **because the 'execution unavailable' notice renders inside the cell frame, so it is platform-owned text and follows the VIEWER — it must come from code.kernelDisabledNotice in the caller's language, not from a literal compiled into MarkdownViewLogic.**

2. **Locale dropped at `RenderKernelResultAreas`** while both leaf helpers stay correct — the exact
   shape of `CodeViews.BuildCellToolbar`'s unused parameter → 2 failures:
   > Expected `"…Starting interactive kernel…"` to contain `"Interaktiver Kernel wird gestartet…"` **because RenderKernelResultAreas is what the Blazor views call, so a locale dropped HERE renders English for every German viewer even with both leaf helpers correct.**

3. **A typo'd key, in BOTH source and guard** — the case every value-comparing assertion survives,
   because `LocalizationCatalog` falls back to the key and both sides then resolve to the same wrong
   string. Caught independently, by two languages resolving to identical text → 3 failures:
   > Did not expect value to be `"code.kernelStartingg"` **because 'code.kernelStartingg' resolved to its own name, which is what LocalizationCatalog does for a key it cannot find — the notice would render that raw token to the reader.**

All three restored; the suite is green.

## Verification

| check | result |
|---|---|
| `dotnet build src/MeshWeaver.Markdown -c Release -warnaserror` | `0 Warning(s)` / `0 Error(s)` |
| `dotnet build src/MeshWeaver.Messaging.Hub -c Release -warnaserror` | `0 Warning(s)` / `0 Error(s)` |
| `dotnet build src/MeshWeaver.Documentation -c Release -warnaserror` | `0 Warning(s)` / `0 Error(s)` |
| `dotnet build test/MeshWeaver.Documentation.Test -c Release -warnaserror` | `0 Warning(s)` / `0 Error(s)` |
| `dotnet build test/MeshWeaver.Messaging.Hub.Test -c Release -warnaserror` | `0 Warning(s)` / `0 Error(s)` |
| Plugins `MeshWeaver.Markdown.Test` @ `d3b5ae01` vs this core, `-warnaserror` | `0 Warning(s)` / `0 Error(s)` |
| `LocalizationTest` (incl. `EveryShippedLanguage_CoversEveryEnglishKey`) | 58 passed |
| `DocumentationLinkIntegrityTest` + `DocumentationEmbedIntegrityTest` + `WhatsNewEntryIntegrityTest` + the new guard | 18 passed |
| `check-type-forwards.py` (both modes) | `0 removed, 0 added` · `OK` |

## 🌍 Follow-up handed over, not done here

The `clients/react/src/i18n/` mirror compares against a **pinned** core commit
(`catalog-source.json`), **not** core's `main` — so these seven keys redden nothing and leave the
mirror silently stale rather than loudly broken. Core is the source of truth and merges first:
`npm run sync:i18n -- --ref <this PR's merge sha>`.

## What still needs the Plugins side

#3203 does **not** close with this. Two pieces of MeshWeaver.Plugins#1308 remain there:

1. Consume the five new `code.*` keys in `MeshWeaver.Graph.Views/CodeViews.cs` (the keys now exist —
   that was the hard block).
2. Pass `ViewerLocale()` at the two `RenderKernelResultAreas` call sites (the overload now exists —
   that was the other hard block).

Refs #3203

---

## 🚨 A finding recorded while measuring (second commit)

`code.neverExecuted` **does not finish its own line**, and that is worth having before the Plugins
half is called done. At MeshWeaver.Plugins `d3b5ae01` the provenance text is a ternary whose *other*
branch is unowned too:

```csharp
var lastRunText = codeConfig?.LastExecutedAt is { } lastRun
    ? $"last run {lastRun.Humanize()}"
      + (string.IsNullOrEmpty(codeConfig.LastExecutedBy) ? "" : $" by {codeConfig.LastExecutedBy}")
    : "never executed";
```

*"last run"* and *"by"* need keys as well — and **`Humanize()` is worse than an unowned string**.
`CodeViews.cs` carries `using Humanizer;`, and Humanizer's culture-less `DateTime.Humanize()`
resolves through **`CultureInfo.CurrentUICulture`**, which AGENTS.md bans outright ("this covers
date/number FORMATTING too"). On Blazor Server that is the container's culture, shared by every
simultaneous viewer — the ambient-resolution defect `Localization.md` exists to prevent, not merely
an untranslated word.

Recorded on the doc page's Outstanding list rather than only in the issue thread, per the
conserve-work-products rule: adding `code.neverExecuted` and stopping there would leave a German
viewer reading *"last run 3 hours ago"* in whatever language the container happens to be in.

Also checked so the handover is precise: **no Plugins test asserts on any of the five literal
strings** (at `d3b5ae01` every grep hit is a comment), so the `...` → `…` normalisation in the new
keys breaks nothing there.


rbuergi:
Copilot's one finding was right and is fixed in `cbfeb788`.

The entry was dated `2026-09-05` because that is this machine's local clock; the repo, CI and the
merge all run in **UTC, where it is still 2026-09-04** (22:38Z). The feed groups by ship day and the
doc tree sorts on `Order`, so a future-dated entry opens a group for a day that has not happened and
sorts above everything that really shipped (`-20260905` < `-20260904`). Every sibling entry from
today is `2026-09-04`. File renamed and `Order` moved with it, which `WhatsNewEntryIntegrityTest`
requires to agree — 18/18 green after the change.

Two sentences in the body were wrong for the same reason and are corrected rather than left: the
rule (#3274) was settled **earlier the same day**, not *"a day earlier"*, and the Run button's
staged release (MeshWeaver.Plugins#1318, merged 08:53Z today) was **this morning**, not *"last
week"*.

Issue #3321, opened 2026-09-04
Portal managed heap leaks ~215 MB/h through 267 gen2 collections, with ZERO sync/bake events — the cause #2194 is not tracking


rbuergi:
Filed at #2194's own instruction — *"if no wave occurred, it is (2) and needs its own RCA"*. #2194 is titled for sync/bake recompile waves; this is the cause that is actually running, and it has **zero** sync/bake events.

## The measurement

`memex-cloud`, replica `…-5nqbz`, born 2026-09-03T17:51Z, read from its own runtime watchdog line:

| time (UTC) | managed heap | gen2 collections |
|---|---|---|
| 06:55 | 6 856 MB | 342 |
| 09:47 | 7 981 MB | 410 |
| 12:48 | 8 660 MB | 454 |
| 15:41 | 8 744 MB | 537 |
| 18:36 | 9 371 MB | 575 |
| 21:51 | **10 065 MB** | **609** |

**+3.2 GB in 14.9 h (~215 MB/h) across 267 full collections, and the post-collection floor never returns to any earlier level.** A cache the GC has simply not trimmed does not survive 267 gen2 collections. This is retention.

## It is not the #2194 mechanism

Counted over that replica's **entire 28-hour life** (186 137 lines), with `info:`-level lines confirmed shipping (2 791 of them) so a zero is a real zero:

```
Stale-build offer:                          0
Stale-build convergence:                    0
EnrichWithNodeType: self-heal recompile #   0
BatchBake:                                  0
KEEPING its load context                    0
NOT unloading collectible context           0
Grain … deactivating: reason=               0
```

No publication, no bake, no recompile, no supersession. `Modules:AutoRecycleOnStaleBuild` is armed on both namespaces and has never fired because nothing has ever been stale.

## It does not track load — it tracks uptime

| replica | age | working set | log lines / 30 min |
|---|---|---|---|
| `fxbhd` | 28 h | 11 443 Mi | 512 — and zero of every hot family |
| `5nqbz` | 28 h | 14 096 Mi | 1 914 |
| `zn7nf` | 15.7 h | 4 134 Mi | **2 476 — the most of any replica** |
| `lgwrk` | 8.3 h | 1 384 Mi | 16 |

[...]

PR #3332, opened 2026-09-05
Move readiness off the progress-aware /alive onto its own /ready (#3330)

### LINKED ISSUE #3330: A progress-aware /alive is wired to BOTH probes: readiness evicts a GC-bound replica 60s before liveness restarts it, onto siblings at the same ceiling
#2194 item 4 is **landed** — `Systemorph/MeshWeaver.Plugins#1234` (merged 2026-09-03T07:49Z) added `ProcessProgressHealthCheck`, and it is the **only** check tagged `live`, which is the tag `/alive` filters on. `/alive` now answers **503** once GC pause share stays above its threshold over a window.

That fixed the blindness it was written for. It also, unnoticed, armed the failure mode the chart warns about in its own words two lines above the probe it changed.

## The defect

`deploy/helm/templates/memex-portal/deployment.yaml` points **both** post-startup probes at the same path:

```yaml
readinessProbe:
  httpGet: { path: /alive, port: 8080 }
  periodSeconds: 10        # 10 × 3 = 30 s to leave the Service
  failureThreshold: 3
livenessProbe:
  httpGet: { path: /alive, port: 8080 }
  periodSeconds: 15        # 15 × 6 = 90 s to restart
  failureThreshold: 6
```

Liveness and readiness cannot be given different **semantics** while they share a **path**. So the moment `/alive` became progress-aware, readiness became progress-aware too — and **readiness trips 60 seconds before liveness**.

A GC-saturated replica is therefore **pulled from the Service first and restarted second**. In that window it serves nothing and its traffic is redistributed to its siblings.

## Why that is a spiral rather than a blip

The siblings are at the same point on the same curve. Measured 2026-09-04 on `memex`:

```
memex-portal-deployment-77898c4947-6mffn   195m   9936Mi
memex-portal-deployment-77898c4947-bfnft   167m   9409Mi     → ratio 1.06
```

Replicas **converge** on the GC's hard limit with age rather than diverging (the derivation is in `Doc/Architecture/WhyAGcBoundPodStaysInRotation`, #3319). So the condition is nearly simultaneous across a namespace, and the pod that inherits the evicted traffic is the one closest to tipping itself.

The chart already states this hazard, on the readiness probe, as a rule:

> *"a readiness check that fails under load removes the pod, and the survivors inherit its traffic — the 2026-07-21 death-spiral. Liveness restarting ONE pegged replica has no such feedback loop."*

`ProcessProgressHealthCheck`'s own docstring reasons only about liveness — *"liveness threshold restarts it — and rejoins by itself if it recovers within the window"* — which is correct for the probe it was designed for. Nothing in either repo connects it to the readiness probe that shares its endpoint, because the two halves live in different repositories.

## What it is not

Not an argument against item 4. A GC-bound replica genuinely should be restarted, and before this it was not. **Only the readiness half is wrong**, and only because it is downstream of a path decision made when `/alive` was a trivial process-up check.

## Options

1. **Give the progress signal its own path** — liveness moves to e.g. `/alive/progress` (or the live-tagged filter moves there) and `/alive` goes back to being the trivial process-up check readiness wants. Two repos, one change each; readiness semantics are restored exactly.
2. **Move readiness off `/alive`** to a trivial endpoint, leaving liveness on the progress-aware one.
3. **Accept it deliberately**, on the argument that a GC-bound replica should leave the Service — but then the 30 s/90 s ordering wants revisiting, since it evicts before it repairs, and the convergence measurement above says it will do so nearly fleet-wide.

I have not picked one: it is a probe-semantics decision spanning this repo and MeshWeaver.Plugins, and getting it wrong in either direction has already cost an outage once (2026-07-21).

## Scope note

Not urgent in the sense of "bleeding right now": the condition needs sustained GC saturation, which is #3321's leak reaching the defence band. It is urgent in the sense that **it fires during exactly the incident it was added to contain**, and the containment and the amplifier are the same mechanism.

Refs #2194 (item 4) · #3321 (what drives pods into the band) · #3319 (the convergence measurement) · `Systemorph/MeshWeaver.Plugins#1234` (the handler)

rbuergi:
Closes #3330

## The defect, confirmed live

Both post-startup probes pointed at `/alive`. Read read-only off the cluster at **2026-09-05T01:27:57Z**, ns `memex`:

```
readinessProbe  {"path":"/alive","port":8080}  periodSeconds 10, failureThreshold 3  → 30 s
livenessProbe   {"path":"/alive","port":8080}  periodSeconds 15, failureThreshold 6  → 90 s
startupProbe    {"path":"/health","port":8080}
```

That was safe only while `/alive` stayed the trivial process-up check it shipped as: nothing carried the `live` tag its predicate filters on, and a `MapHealthChecks` matching an empty set is **vacuously healthy**. `Systemorph/MeshWeaver.Plugins#1234` then registered `ProcessProgressHealthCheck` on that tag (merged 2026-09-03) — correct, and the fix to a real blindness — and **readiness inherited it silently**, because two probes cannot be given different semantics while they share a path.

Readiness trips first. A GC-bound replica therefore left the Service a full minute before anything restarted it, and for that minute its traffic landed on siblings converging on the *same* ceiling (ns `memex`, 2026-09-04: two 28 h replicas at 9936Mi and 9409Mi — ratio **1.06**). One sick replica became a cascade: the 2026-07-21 death spiral, rebuilt out of the containment meant to prevent it. It fires during exactly the incident it was added to contain.

The chart already stated the rule, two lines above the probe, in these words: *"liveness and readiness cannot be given different SEMANTICS while they share a PATH."* It was correct and it did not stop the change, because the change was in another repository. **A rule stated in a comment is not a gate.**

## The choice, and what was rejected

**Chosen — issue option 2:** readiness moves to its own `/ready`, filtered by its own `ready` tag. `/alive` keeps the `live` tag and the liveness question.

- **Rejected — option 1** (give the *progress* signal a new path, e.g. `/alive/progress`, and hand `/alive` back to readiness). `/alive` is the Aspire/Kubernetes convention for liveness and "am I making progress" *is* the liveness question; this inverts the convention and needs a matching change in MeshWeaver.Plugins for no semantic gain.
- **Rejected — readiness excluding the `live` tag** (`Predicate = r => !r.Tags.Contains("live")`), the cheaper-looking answer named in the brief. It is a **deny-list**, so every *untagged* check joins readiness by default — and the untagged checks are the heavy DB and mesh ones `/health` is made of. That is the 2026-07-21 death spiral in its original form. It also repeats the root cause instead of removing it: the defect was a check in another repo joining readiness's predicate by carrying a tag chosen for liveness. An allow-list makes a host say "take me out of rotation" in exactly those words.
- **Rejected — a vacuous readiness endpoint** (`Predicate = _ => false`). Right behaviour today, a detector that cannot fail tomorrow — precisely how `/alive` stayed blind through 2026-08-25. The trivial `self` check carries **both** tags so `/ready` runs something.
- **Rejected — option 3** (accept it, retune 30 s/90 s). Treats the arithmetic as the defect; the defect was the shared predicate. No ordering makes readiness a safe place to express "this pod is GC-bound" on a fleet whose replicas converge.
- **Not touched:** `ProcessProgressHealthCheck`. It was right; it just needed readiness to stop listening. Both Plugins hosts already call `MapDefaultEndpoints()`, so `/ready` ships with the next platform-ref bump — **this lands entirely in this repo**.

## The guards, mutation-proven

Neither was written and assumed. Each was broken on purpose, observed red, and restored.

**Mutation A — the original defect** (readiness back to `/alive`):

> `ProbeSemanticsGuard.EachProbe_AsksItsOwnQuestion`
> `deployment.yaml points readinessProbe and livenessProbe at the same path ('/alive'). They cannot then be given different semantics: the moment that path answers a progress-aware verdict, a GC-bound replica is EVICTED at 10s×3=30s and only RESTARTED at 15s×6=90s… That is the 2026-07-21 death spiral rebuilt out of the #2194 item-4 fix — see #3330.`

> `ProbeSeparationTest.AFailingLivenessCheck_DoesNotTakeThePodOutOfRotation`
> `the chart's readinessProbe path ('/alive') answered 503 because a LIVENESS check was failing — and it answered it on the very same path the livenessProbe uses, so the two probes cannot be given different semantics at all.`

> `check-chart-invariants.py`, red on **all three** values combinations, real exit 1:
> `::error::[self-host (neutral chart defaults)] readinessProbe and livenessProbe both probe /alive`

**Mutation B — the defect in disguise** (paths differ, `/ready` wired to `LiveTag`), because path separation alone is not the invariant:

> `ProbeSemanticsGuard.EachPostStartupProbe_ResolvesToItsOwnHealthCheckTag`
> `readinessProbe ('/ready') and livenessProbe ('/alive') are different paths but both filter on the health-check tag 'live'… so they still ask the SAME question, and #3330 is back with two URLs instead of one.`

> `ProbeSeparationTest` (both cases red): `the chart's readinessProbe path ('/ready') answered 503 because a LIVENESS check was failing — so readiness is wired to the liveness predicate even though the paths differ.`

Both C# guards live inside `Consolidate test results`, the one required check, and both read the paths **out of the chart** rather than restating them.

## Verification

| step | result |
|---|---|
| `dotnet build memex/aspire/Memex.Portal.ServiceDefaults -c Release -warnaserror` | `0 Error(s)`, 22.74 s |
| `dotnet build memex/Memex.Portal.Shared -c Release -warnaserror` | `0 Error(s)` |
| `dotnet build test/Memex.Portal.Shared.Test -c Release -warnaserror` | `0 Error(s)` |
| `dotnet build test/MeshWeaver.Documentation.Test -c Release -warnaserror` | `0 Error(s)` |
| `Memex.Portal.Shared.Test` (probe + setup + drain + version + shell) | **44/44 passed** |
| `MeshWeaver.Documentation.Test` (link integrity, What's New, probe, drain) | **11/11 passed** |
| `check-chart-invariants.sh` | all 3 combinations self-consistent, exit 0 |
| `check-values-are-read.sh` | 114/114 config keys read, exit 0 |
| `test-chart-drift-compare.sh` | all classification assertions passed |

## 🚨 Rollout order: image first, chart second

`/ready` is mapped by the portal **image**. A `helm upgrade` carrying this chart onto pods running a pre-2026-09-05 image probes a path that **404s — on every replica at once**. The reverse is inert.

The fleet's normal sequence is already the safe one (CD rolls images with `set image`; a `helm upgrade` is always a deliberate human action, and self-update cannot run one). The case to watch is a namespace **pinned** to an older image. Confirm first:

```bash
kubectl -n <ns> exec deploy/memex-portal-deployment -- \
  curl -sf -o /dev/null -w '%{http_code}\n' http://127.0.0.1:8080/ready
```

**No cluster was mutated by this change** — the only cluster access was the read-only `jsonpath` above. Applying the chart to `memex` / `memex-cloud` is the operator's decision. Until it is applied, `check-chart-drift.sh` will legitimately report `DIFFERS readinessProbe` against those namespaces: the chart is the authoritative side, and the drift is work not yet done.

## Work products committed

- `Doc/Architecture/ProbeSemantics` — the three questions and their three remedies, the incident, the allow-list argument, the rollout order, and how to decide a new check's tag.
- `Doc/Architecture/WhyAGcBoundPodStaysInRotation` — the stale "still open" paragraph replaced with the resolution.
- `deploy/aks/README.md`, `values.observability.yaml`, `ServiceDefaults/README.md` — four places asserting "both post-startup probes watch `/alive`" corrected.
- A What's New entry (`Category: Fix`) — the user-visible symptom was the portal getting gradually slower and then recovering.

Nothing is claimed here about alert *arming*: `values.observability.yaml` has never been applied and Prometheus holds zero rule groups (#3321). Only its prose was corrected.

Pairs-with: none — no public type or member is removed from `src/`; the change adds `ProbeEndpoints` under `memex/` and adds one endpoint.


rbuergi:
Your shard-3 red is not yours — and it is the #3261 leak class, not a bake defect.

`BakeEquivalenceTest.MeshDrivenAndCompilerDrivenBakes_ProduceTheSameArtifacts` fails on `Assert.Null(error)` with `InvalidOperationException: bake: 'Widget/Thing' cl…`. The log shows the bake itself SUCCEEDED —

```
ok  Widget/Thing [compile:Passed render:Passed tests:Skipped]
```

— and then, during teardown:

```
[QUIESCE-TIMEOUT] Widget/Thing: 2 callback(s) still pending after 2s — forcibly cancelling.
  Pending: …=CreateOrUpdateNodeRequest@portal/nodeops-…
CompileWatcher: failed to create Release node at Widget/Thing/Release/… HubDisposedBeforeResponseException
[ReleasePostCondition] Widget/Thing: a release request was consumed and this compile succeeded, yet the no[de …]
```

So the compile's follow-on writes — the CompileStateMirror satellite and the Release node — were still in flight when the mesh went down. The post-condition then correctly reports that a consumed release request produced no Release node, and the test surfaces that as the bake's error.

`CreateOrUpdateNodeRequest@portal/nodeops-*` pending at quiesce is verbatim one of the four shapes measured on #3261 (see my comment there of 2026-09-04). This is another instance of that class, in a different suite.

Your diff is `deploy/**`, `memex/Memex.Portal.Shared/Setup/**`, `memex/aspire/**/ProbeEndpoints.cs` — readiness and health probes. It cannot reach `MeshWeaver.PluginTester.Test`, so this is not attributable to the change.

rbuergi:
## CI is red, and not for anything in this diff — holding rather than re-running

Run [33937167117](https://github.com/Systemorph/MeshWeaver/actions/runs/33937167117) failed two tests:

- `MeshWeaver.PluginTester.Test.BakeEquivalenceTest.MeshDrivenAndCompilerDrivenBakes_ProduceTheSameArtifacts` — the mesh-driven bake died on `Widget` because `mesh/…` was already `RunLevel=DisposeHostedHubs` while the bake still had work in flight.
- `MeshWeaver.Hosting.Test.ChangeFeedResetReleasesUpstreamTest.ControlArm_ReleaseIfUnwatched_OfTheSameFaultedEntry_DisposesAnUpstream` — an unwatched faulted entry whose release did not claim its upstream.

Both are teardown/disposal races. Neither is reachable from this diff: it contains **zero `src/**.cs` changes**, and neither failing project references `Memex.Portal.ServiceDefaults` or `Memex.Portal.Shared` — the only assemblies whose code changed here. This PR's own 7 new tests all passed (shard 5, `+7 ✅`).

Filed with full evidence, the log line naming the cause, and the base rate as **#3333**.

**I have not re-run the tests and have not re-pushed to get a clean run.** A flake is a real race, and the 13 Build-and-Test runs immediately before this one were all green — so clearing the red would have destroyed the only evidence. Auto-merge stays armed; this will merge itself once main is green on those two, or once someone decides they are unrelated enough to land over.

rbuergi:
Refreshed this onto `main` — not to catch up, but because one of its two reds is now fixed there.

`Run tests (shard 4)` was `ChangeFeedResetReleasesUpstreamTest.ControlArm_ReleaseIfUnwatched_…`, and
the defect was in `ReactiveWait.First`, not in that test or in anything this PR touches: the assertion
released its subscription as a **continuation on the task the test awaits**, so `await …Should()…`
returned while still subscribed, and the release refused the entry the assertion was still pinning.
Fixed and guarded in #3336, merged as `a9d05d9a9`.

`Run tests (shard 3)` was `BakeEquivalenceTest`, which is **not** fixed — tracked on #3333, where the
cause is now narrowed to `LastCompiledVersion` naming a version the assembly store never held
(shutdown-mid-bake, cross-test eviction, in-run eviction and framework-tag mismatch are all excluded
with evidence). If this run comes back red on shard 3 alone, that is the known one and not yours.

Issue #3328, opened 2026-09-04
The platform image ships Roslyn bound to SQLitePCLRaw.core 3.0.2 beside 2.1.11 — every module compiling against it fails MSB3277


rbuergi:
## What breaks

Four module bundles fail to compile against the `3.0.0-rc9.ci.7779` platform image:

```
error MSB3277: Found conflicts between different versions of "SQLitePCLRaw.core" that could not be resolved.
  There was a conflict between "SQLitePCLRaw.core, Version=2.1.11.2622" and "SQLitePCLRaw.core, Version=3.0.2.2801"
  "…2.1.11.2622" was chosen because it was primary
  References which depend on "…2.1.11.2622":
      /home/runner/work/_temp/platform-refs-effective/SQLitePCLRaw.core.dll
```

Observed on MeshWeaver.Plugins#1349 (`Module bundle` for `MeshWeaver.Publish`, `MeshWeaver.Testing`, `MeshWeaver.Hosting.Instance`, `MeshWeaver.Northwind.Application`). Every name in the conflict resolves inside `platform-refs-effective/` — the extracted image `/app` — so **the skew is inside the image's own reference set**, not between a module and the image.

## Why it can happen at all

`Directory.Packages.props` pins exactly one member of the family:

```xml
<PackageVersion Include="SQLitePCLRaw.lib.e_sqlite3" Version="3.53.3" />
```

The **managed** trio — `SQLitePCLRaw.core`, `SQLitePCLRaw.provider.e_sqlite3`, `SQLitePCLRaw.batteries_v2` — is unpinned and arrives transitively. All three are in the image (`SQLitePCLRaw.core.dll`, `.provider.e_sqlite3.dll`, `.batteries_v2.dll`), and two different `core` versions are now reachable in one closure.

That contradicts the intent the pin's own comment states:

> The provider P/Invokes the engine by ABI-stable export name, so the native lib drops in under the **2.1.x managed layer**.

So the design is *managed 2.1.x + native 3.53.3*, and a managed `3.0.2` appearing violates it. Nothing in this repo changed to cause that: `git diff 7d644de95 42f0cbc4e` touches **no `.csproj` and no `.props`** — an unpinned transitive simply resolved differently in the newer image build.

## Suggested fix

Pin the managed trio alongside the native engine, so the layer cannot drift under a floating transitive:

```xml
<PackageVersion Include="SQLitePCLRaw.core" Version="2.1.11" />
[...]

PR #3334, opened 2026-09-05
ci(#3327,#3328): assert the platform image's closure before it is promoted

### LINKED ISSUE #3327: One-producer FATAL returns from the IMAGE side: the platform host ships MeshWeaver.Markdown.Collaboration while both repos compose it as a module
## The one-producer FATAL is back, from the IMAGE side

`test-repos / Compile + render node repos (MeshWeaver from ACR)` is RED on **every** open
MeshWeaver.Plugins PR — #1337, #1335, #1334, #1322, #1338, #1348 — and on none of their diffs. The
gate names itself precisely:

```
compile: FATAL — module(s) MeshWeaver.Markdown.Collaboration are composed with --module AND
shipped by the platform host at '/app' — two builds of one assembly name in one bake. …
A module has exactly one producer: remove the assembly from the platform host's closure
(it is a module, landed from the registry) or stop composing it (#3175).
```

**Not required** (branch protection lists five contexts and this is not one), so nothing is blocked.
Filing because it is fleet-wide, it recurred after a fix, and the returning half is the one nobody
owns.

## Measured — both repos already agree it is a module

| | |
|---|---|
| `MeshWeaver.Markdown.Collaboration` in core `src/` | **absent** — it lives in MeshWeaver.Plugins |
| core `main-cd.yml` compose set | **composes it** (`modules:` + `module-artifacts:`) |
| Plugins `src/platform-shipped.txt` | **not listed** — correctly treated as a module, not platform-carried |
| Plugins#1268 (`fix/3175-markdown-collaboration-one-producer`) | merged 2026-09-04T12:48:45Z |
| pinned platform set | `3.0.0-rc9.ci.7755` — built *after* #1268, so this is **not** a stale-image lag |

So the repo side of #3175 is fixed and holding. The **platform image still carries the assembly in
`/app`**, which makes the image the second producer. Both declarations say "module"; the bytes say
otherwise.

## Why this is worth attention beyond the red

It is the same shape as #3244 ("one producer in TIME" — the registry's content-versioned bytes vs the
sealed publication's), one layer over: **not two producers in time, but two producers in one bake,
with the platform image as the second.** I have not proven that link and am not going to assert it —
but whoever picks this up should read them together, because a fix that only addresses the repo side
has now been shown to leave the image side live.

## Suggested next measurement

Resolve the image the gate actually pulled and list its `/app` — the FATAL names the assembly, so
`MeshWeaver.Markdown.Collaboration.dll` should be present there. If it is, the question is which
build put it there: the portal image compiles Plugins content, so a module compiled *into* the portal
and *also* composed as a module is the whole defect, and the fix is to remove it from the portal's
closure rather than to stop composing it (it is a module; composing it is correct).

Related: #3175 (the rule), Plugins#1268 (the repo-side fix), #3244 (the time-axis sibling).
Reported by a concurrent session working MeshWeaver.Plugins; measurements above are my own.

rbuergi:
Closes #3327. Delivers the guard both #3327 and #3328 ask for.

**Pairs-with: Systemorph/MeshWeaver.Plugins#1351** — the image-side fix for #3328. 🚦 **That one must merge first**, or the next CD run fails this gate on the current image. (No public surface is removed here, so the pair gate does not apply; the ordering is real all the same.)

## The shared gap

> *"The platform image is the compiler and the reference set"* — `Doc/Architecture/ModuleBuildArchitecture`.

So the image's `/app` is a **contract with six repositories** — and nothing asserted it. Every consumer discovered its contents by failing to compile against it, hours later, in a repository whose own diff was innocent. Two breaches arrived on 2026-09-04, in opposite directions, and **both rode one image, `ci.7758`**.

## #3327 — the diagnosis, corrected

The issue records that `ci.7755` "was built AFTER Plugins#1268 (12:48:45Z), so this is **not** stale-image lag". Re-measured against ACR, that is false, and it changes what needed fixing:

| | |
|---|---|
| `memex-portal-ai:3.0.0-rc9.ci.7755` created | **2026-09-04T11:43:19Z** — *before* 12:48:45Z |
| its Plugins commit (off the `<core>-p<plugins>` tag) | `a9aaf84` |
| `ci.7757` | `f9d2a33-ped7b270` — Plugins `ed7b270`, PR #1327, 11:30:38Z |
| `ci.7758` | `b0d4a82-p2f7a95d` — Plugins `2f7a95d`, **#1268's merge commit** |

Bisecting the extracted `/app` of every promoted image in the window:

```
ci.7756  MeshWeaver.Markdown.Collaboration.dll at /app root: yes
ci.7757  yes
ci.7758  no        ← Plugins#1268
ci.7794  no
```

So **the repo side of #3175 landed, held, and is not the problem**; the satellite red was ordinary stale-image lag on a pin that predated the fix. The rule and the refusal already existed (`BakeHost.ShippedByHostProblem`) — **the gap was WHEN it runs**: after `promote`, so the image already carries a version tag every satellite can pin, and the red lands in *their* repository. Moving that predicate before `promote` is invariant 1.

## #3328 — root cause, measured

```
ci.7755 → ci.7794, LEFT /app root:      MeshWeaver.Markdown.Collaboration.dll   ← #3327's fix
ci.7755 → ci.7794, ENTERED /app root:   MeshWeaver.Hosting.Sqlite.dll
                                        Microsoft.Data.Sqlite.dll
                                        SQLitePCLRaw.batteries_v2.dll
                                        SQLitePCLRaw.core.dll
                                        SQLitePCLRaw.provider.e_sqlite3.dll
```

`Microsoft.CodeAnalysis.Workspaces.Common` 5.9.0 declares exactly three dependencies — `Humanizer.Core`, `Microsoft.CodeAnalysis.Common`, `System.Composition` — and **no SQLitePCLRaw**, yet `Microsoft.CodeAnalysis.Workspaces.dll` binds `SQLitePCLRaw.core 3.0.2.2801`: Roslyn's optional SQLite persistence is compiled in with the package dependency excluded, so **NuGet cannot see that consumer at all**. Harmless for as long as no `SQLitePCLRaw.core.dll` was in `/app` — `ci.7755`'s reference set is provably clean. Plugins#1284 ("First-run setup: SQLite becomes selectable") supplied the other half.

The issue's proposed fix — pin the trio at 2.1.11 — would not have moved a byte: that is already what resolves. The whole story, including why moving the family *forward* would silently undo the GHSA-2m69-gcr7-jv3q native-engine pin, is in the doc page; the working fix is Plugins#1351.

## The gate

`.github/scripts/check-platform-reference-set.sh`, a step of `main-cd.yml`'s `portal-image` job, over the publish output that becomes `/app`. **Before `promote`** — the same reasoning as the identity readback beside it: a tag is not evidence about the bytes under it.

- **Invariant 1 (one producer)** reads the composed-module set **out of `main-cd.yml` itself** (`jobs.plugins-modules.with.modules`), never restated — the compose set and the set `/app` is forbidden to carry must be one list, or a drifted second list keeps the gate green while it asserts the wrong names.
- **Invariant 2 (the set resolves)** does not *model* MSBuild's binding rules, it **runs** them: a throwaway project whose `<Reference>` items are that directory, built with the module lane's own `-c Release -warnaserror`. A hand-written version comparison would be a second opinion about what `ResolveAssemblyReferences` does, and this entire defect class is two artefacts that were each individually plausible.
- **No skip trapdoor.** Every input is asserted and a missing one is RED naming what to provision: ≥50 assemblies, a surface manifest naming ≥20, a non-empty composed set, and ≥50 references RAR actually resolved. The directory is *found* by its manifest rather than assumed from a RID, so both architectures of the multi-arch publish are checked and finding none is red.

## Mutation proof

Run against the real extracted `/app` of both promoted images, and against a real local portal publish, **before the fix was written**:

| reference set | invariant 1 | invariant 2 | exit |
|---|---|---|---|
| `ci.7755` `/app` | **RED** — `MeshWeaver.Markdown.Collaboration` in the closure *and* the manifest | OK (377 refs) | `1` |
| `ci.7794` `/app` | OK | **RED** — `MSB3277`, `SQLitePCLRaw.core` 2.1.11.2622 ↔ 3.0.2.2801 | `1` |
| `ci.7794` `/app` + the fix | OK | OK (381 refs) | `0` |
| local portal publish, fix applied | OK | OK — 214 assemblies, 381 refs | `0` |
| same publish, fix reverted **in place** | OK | **RED** — `MSB3277` | `1` |

`ci.7755` red on #3327's shape and green on #3328's, `ci.7794` the exact opposite: neither half is a blanket red, and each fails on the image that actually broke.

<details><summary>The gate naming #3327 on <code>ci.7755</code></summary>

```
platform reference set: 210 assemblies at '…/app7755'; surface manifest names 44
one-producer: 4 composed module(s) read from .github/workflows/main-cd.yml — MeshWeaver.AI MeshWeaver.Maps MeshWeaver.Markdown.Collaboration MeshWeaver.Payments.Stripe
::error::ONE PRODUCER (#3175/#3327): 'MeshWeaver.Markdown.Collaboration' is composed with --module by this run AND shipped by the platform host … MeshWeaver.Markdown.Collaboration.dll in the host's closure and an entry in its surface manifest. …
reference-set probe: ResolveAssemblyReferences resolved 377 references from the host's closure
reference-set probe: OK — the closure resolves with no conflict under -warnaserror
```
</details>

<details><summary>The gate naming #3328 on <code>ci.7794</code></summary>

```
platform reference set: 214 assemblies at '…/app7794'; surface manifest names 44
one-producer: OK — no composed module's name is in the host's closure or surface manifest
reference-set probe: ResolveAssemblyReferences resolved 381 references from the host's closure
::error::THE PLATFORM IMAGE'S REFERENCE SET DOES NOT RESOLVE (#3328). …
error MSB3277: There was a conflict between "SQLitePCLRaw.core, Version=2.1.11.2622, …" and "SQLitePCLRaw.core, Version=3.0.2.2801, …"
```
</details>

<details><summary>Green, on the real publish output built from Plugins#1351</summary>

```
platform reference set: 214 assemblies at '…/bin/Release/net10.0/linux-x64/publish'; surface manifest names 44
one-producer: OK — no composed module's name is in the host's closure or surface manifest
reference-set probe: ResolveAssemblyReferences resolved 381 references from the host's closure
reference-set probe: OK — the closure resolves with no conflict under -warnaserror
platform reference set: OK — 214 assemblies, 381 resolved references, 4 composed module(s) absent from the closure
```
</details>

## Work products conserved

`Doc/Architecture/PlatformImageClosure` — both invariants, both root causes with their measurements, the corrected #3327 diagnosis, the mutation proof, and **what the gate still cannot see**. That last section records a live finding this work turned up and deliberately did not fold in: running the *broader* module-owned predicate (every `MeshWeaver.*` project the portal repo's `src/platform-shipped.txt` does not name) flags five further assemblies `/app` ships today — `MeshWeaver.Blazor`, `MeshWeaver.Blazor.Portal`, `MeshWeaver.Hosting.Blazor`, `MeshWeaver.ContentCollections.Indexing`, `MeshWeaver.ContentCollections.Indexing.Graph`. The last two are named in that file's own comment as *measured ABSENT from `/app`*, with "do NOT re-add"; they are present in both `ci.7755` and `ci.7794`. Same declaration-versus-bytes gap, one layer over, and it touches the in-flight optional-Blazor work — so it is written down rather than changed here.

Also: `Directory.Packages.props`'s SQLitePCLRaw comment now records why the managed layer is pinned and why raising `bundle_e_sqlite3` would undo the CVE pin.

## Local verification

- `check-workflow-timeouts.py` — 66 jobs checked, **0 violations**, cap 45 min
- `check-workflow-shell.py` — **0 live findings**; `shellcheck -s bash` on the new script clean
- `test-cd-steps.py` — all cases pass against the extracted `release` + `decide` steps
- `dotnet build test/MeshWeaver.Documentation.Test -c Release -warnaserror` → **0 Warning(s), 0 Error(s)**
- `dotnet test test/MeshWeaver.Documentation.Test` → **264 passed**, and `DocumentationLinkIntegrityTest` was mutation-checked with a deliberately bogus link, which it named before I removed it


rbuergi:
## 🚦 Held as a DRAFT on purpose — do not un-draft until MeshWeaver.Plugins#1351 is merged

Draft is the opt-out from `auto-arm.yml`, and that is exactly what this needs. The reason is mechanical, not cautious:

- The gate runs in `portal-image`, over the closure of the image **this run builds** — core HEAD + the plugins HEAD `gate` resolved.
- The current plugins HEAD still produces a closure that fails invariant 2 (`MSB3277`, `SQLitePCLRaw.core` 2.1.11.2622 ↔ 3.0.2.2801 — measured on the promoted `ci.7758`, `ci.7779`, `ci.7789` and `ci.7794`, and on a real local publish).
- So merging this first turns the **next CD run** red at `portal-image`, before `promote`, and **no image is published at all** until the plugins half lands. That is delivery stopped fleet-wide by the very change meant to protect it.

Merged in the right order it is self-completing, which is why the order is not merely safe but necessary: **CD is keyed on CORE's HEAD sha**, so Plugins#1351 landing on its own rebuilds nothing (the scheduled reconcile finds core's image set complete and takes `bake_only`; `main-cd.yml`'s `rebuild` dispatch input exists for exactly this case, MeshWeaver#2622). The core merge that brings this gate in **is** the core commit that rebuilds the portal — so the gate's first live run is against a closure that already carries the fix.

Written up in the PR's own doc page under *"Fixing a breach is a two-repo act, and the second half is not automatic"*.

### Status of the blocking half

Plugins#1351 is verified and armed for auto-merge, and its CI red is attributed and **not its diff**: the only failing *required* check is `Build + test the portal hosts`, which is intermittently red on Plugins `main` itself right now — `main` 33934963766 passed it, `main` 33936784055 (this branch's base) failed it on `NeverCompiledFailureRedriveTest`, and #1351 failed it on `NackReachesTheWaiterDuringTeardownTest`. Two different tests in `MeshWeaver.Hosting.Monolith.Test`, neither touched by a SQLitePCLRaw package version, in the teardown/quiesce area #3291 is live in. Not re-run, per the fleet rule and per that repository's own `retry-known-transients.yml`, which refuses anything that is not a named infrastructure signature.

**To finish:** once #1351 is green and merged, `gh pr ready 3334` and let it merge on green.

rbuergi:
Taken out of draft — the blocker cleared and the ordering it names has happened.

**`Pairs-with: Plugins#1351` has merged** (09:47Z), so the next CD run no longer fails this gate on
the current image. Confirmed downstream: ci.7818 is built from a Plugins tree carrying #1351, and
every `Module bundles` job — tests included — is green against it on Plugins#1359. The breach this
gate exists to catch is out of the image.

Also merged `main` (the branch was `dirty`). One conflict, in `Directory.Packages.props`, and both
sides were saying the same thing in different words — this branch's version, and the text
`8a4fadf0a` left after restoring the native pin. I combined them rather than picking a side, so the
comment now carries all three facts a reader needs:

* **the native pin stays**, and why — #3344 withdrew it along with the three managed pins it meant
  to withdraw, which took core CD down (NU1010, then NU1903 as resolution fell back to the
  vulnerable 2.1.11 the pin exists to avoid). Restored in `8a4fadf0a`; Plugins#1361 additionally
  states the version at the reference so the two repos no longer depend on each other;
* **the managed trio is deliberately not pinned here**, with the `CentralPackageTransitivePinningEnabled`
  reason this branch had and the "#3342 withdrawn" history `main` had;
* **never raise `bundle_e_sqlite3` to 3.0.x**, with the measured restore-graph reason.

The `3.53.3` pin itself was outside the conflict and is unchanged.

## Why this should land

Two fleet-wide breaks arrived within hours of each other yesterday and **both** were the shape this
gate asserts — nothing said what `/app` contains, so every consumer discovered it by failing to
compile, later, in a repository whose own diff was innocent. The second one (#3328) then cost a
two-hour CD outage and a fleet-wide staleness red while the pin could not move.

I checked the thing most worth checking about a gate: **it can fail.** No `continue-on-error`, no
`if:`, and when it cannot locate the published closure it exits 1 saying so —

> `no publish output under '$root' holds a meshweaver-surface.manifest, so the bytes that BECOME /app cannot be located and this gate would verify nothing. Either the publish layout moved — re-point this step, never make it conditional — …`

That is the right shape: a gate that cannot see its subject reports RED rather than green, which is
exactly what the two breaches needed and did not have.

PR #3337, opened 2026-09-05
Verify a webhook target's HMAC at the endpoint, so a drifted secret fails loudly (#3312)

### LINKED ISSUE #3312: A mismatched webhook-secret is invisible: the publication POST answers 2xx and every delivery is silently dropped
## A wrong `webhook-secret` is invisible everywhere

`node-repo-publish-bake.yml`'s **Register the publication with memex** signs the record with
`webhook-secret` and the control instance verifies `X-Hub-Signature-256` against
`Hosting:PlatformWebhookSecret`. `main-cd.yml` states the failure mode itself (L302):

> a mismatch is INVISIBLE from here — the POST still answers 2xx and the watcher silently drops
> every delivery as unverifiable

So the three states below are indistinguishable from CI:

| state | CI sees | reality |
|---|---|---|
| secret correct | 2xx, job green | record stored, dependents notified |
| secret **mismatched** | **2xx, job green** | **every delivery dropped; nobody notified** |
| secret empty | RED, named | caught today (#3311) |

Only the third has a detector. The empty case was found precisely because it fails closed; the
mismatch case fails **open** and looks identical to success.

## Why this is worth a fix rather than a note

It is the same failure family as an absent input reading as a passing gate, and the consequence is
identical to the bug #3311 just fixed — **no dependent is told, each falls back to its schedule
poll** — except that nothing anywhere goes red, so it can persist indefinitely. #3311 restores the
notification for the empty case; it cannot tell you the HMAC is right.

## Suggested shape

Make the inbox's answer carry the verification verdict rather than only transport success — e.g. it
returns a distinct status/body when the signature fails to verify, and the lane asserts on THAT, not
on `2xx`. A 2xx that means "I received bytes I then threw away" is a receipt for the wrong thing.

Whatever the mechanism, the property to hold is: **a run that publishes must be able to fail because
the record was not accepted**, not merely because it was not sent.

## Provenance

Found while fixing #3311 (core asserted `webhook-url`/`webhook-secret` and passed neither, so every
publication sealed unregistered). Deliberately not bundled into that PR — it needs its own mutation
proof, and riding it along would have made a verified fix unverifiable.

Related: `Doc/Architecture/ModuleBuildArchitecture` → "What the gate still cannot see"; #3310 (the
outcome-level bundle-identity guard, same "config-level check cannot see the outcome" shape).

rbuergi:
## The defect

`POST /api/hooks/{target}` answered **2xx for "I stored bytes"**, and CD read it as acceptance.
That is the same answer when the shared secret matches and **also** when it has drifted: the
delivery is stored, the publishing run goes green, and the consuming watcher then discards every
event as unverifiable. Nothing is red anywhere, so it can persist indefinitely while every
dependent falls back to its schedule poll. The **empty**-secret case was caught (#3311) precisely
because it fails *closed*; the mismatch failed *open*.

#3312's stated property: **a run that publishes must be able to fail because the record was not
ACCEPTED, not merely because it was not sent.**

## The fix

A target may name the configuration key holding its shared secret:

```yaml
WebhookInbox__Targets__0:                   "Hosting/PlatformBuilds"
WebhookInbox__Targets__0__SecretConfigKey:  "Hosting:PlatformWebhookSecret"
```

- **It rides ON the allowlist entry** — a configuration section carries both a value and children —
  so the record that makes a target reachable is the record that says how it is authenticated.
  No parallel list to drift, and every existing `c.Value` reader still sees a plain path, including
  the in-mesh `PaymentPathAudit` in MeshWeaver.Plugins that `dotnet build` cannot see.
- **It names a KEY, never a secret.** A value pasted in resolves to no key and refuses everything —
  loudly, fail-closed — instead of landing a live credential in a ConfigMap.

When declared, the endpoint verifies `X-Hub-Signature-256` over the raw body **before storing**:

| outcome | status | stored? |
|---|---|---|
| verifies | 200 `{"status":"accepted","signature":"verified"}` | yes |
| absent / malformed / mismatched | **401** | **no** |
| declared key empty on this instance | **500** (ours, not the caller's) | **no** |
| no declaration | 200 `…"signature":"not-required"` | yes — the dumb contract, unchanged |

Refusal order is contract: target → size → signature, always before the node is created. A
delivery that fails to verify must leave nothing behind, or the fix has only moved the silent drop
from the consumer into the store.

### Why the body carries the verdict too

`verified` and `not-required` are both 200 and mean very different things to a sender that signed.
Without the distinction, "we verify now" degrades silently to "we used to verify" the day a chart
value goes missing — the bug's own shape, one level up. Both lanes read it and `::warning::` on
`not-required`.

**That warning is deliberate, and so is its own step.** The POST step is held to *every message
ends the job* (`PlatformReleaseNotifyGuard` forbids `::warning` in it), because a release event
that never arrived must never be decorated. This is the opposite case — the event *did* arrive —
so it gets a separate step rather than softening that guard. It is a warning and not a failure
only while the declaration rolls out: it reaches an instance through its `Hosting/Deployment`
record + `helm upgrade`, never through self-update (a `set image`), so failing here would red every
publish until that lands.

## Verification

- `MeshWeaver.Graph`, `MeshWeaver.GitSync`, `Memex.Portal.Shared`, `MeshWeaver.Graph.Test` — all
  `0 Error(s)`, Release, `-warnaserror`.
- `WebhookInboxTest` 10/10; workflow + chain guards 13/13; doc-link + What's New 2/2.
- **Mutation-proved**, as the issue asked. Removing the verification block turns exactly three
  tests red — `SignedTarget_WithADriftedSecret_IsRefused_AndStoresNothing`,
  `SignedTarget_RefusesAnAbsentOrMalformedOrRebodiedSignature`,
  `DeclaredButUnprovisionedSecret_RefusesEverything_FailClosed` — and leaves the accept and
  dumb-contract tests green. The pair `…WithTheRightSecret_IsAccepted` vs
  `…WithADriftedSecret_IsRefused…` **both returned `Accepted` before this change**; that identity
  IS the defect, and if a future change makes them agree again the hole is back.

## Controls added

- `ReleaseDeliveryChainGuard` now covers the declaration for **both** slots — index 0 is not the
  same slot on every deployment (#2352). It asserts the key is *renderable*, never that it is set:
  an empty declaration is CORRECT for a slot holding a Stripe target, which signs
  `Stripe-Signature` and would 401 against this scheme on every payment. What catches a missing
  declaration on the slot that needs one is the lane reading the inbox's answer.

## Still open after this

The control instance must declare the key before it actually verifies. That is a
`Hosting/Deployment` record change plus a roll; until then both lanes say `signature:
not-required` on every publish, by design. **Escalating that warning to an error is the last step
of #3312** and is one line in each lane.

## Docs

`Doc/Architecture/WebhookInbox` rewritten (the section that documented this as *open* now
documents the fix, and keeps the two rejected alternatives so they are not re-derived), plus a
What's New entry.

Closes #3312

### DIFF
--- a/.github/workflows/main-cd.yml
+++ b/.github/workflows/main-cd.yml
@@ -2105,6 +2105,13 @@ jobs:
             2*)
               echo "### 📣 Release event for $SHA ($DIGEST) accepted by the inbox (HTTP $code)" >> "$GITHUB_STEP_SUMMARY"
               echo "The mesh STORED the delivery. Whether the wave actually ran is verified on the satellites (a repository_dispatch run + the MW_IMAGE_DIGEST bump PR), never here — see #2235." >> "$GITHUB_STEP_SUMMARY"
+              # 🚨 #3312: a 2xx says bytes landed; only the BODY says whether this instance
+              # actually checked the HMAC we just computed. Recorded here, judged in the next step.
+              if grep -Eq '"signature"[[:space:]]*:[[:space:]]*"verified"' /tmp/resp; then
+                echo "- signature: **verified** — the instance checked this POST's HMAC." >> "$GITHUB_STEP_SUMMARY"
+              else
+                echo "- signature: **NOT checked** — the instance declares no SecretConfigKey for this target (#3312)." >> "$GITHUB_STEP_SUMMARY"
+              fi
               ;;
             000)
               echo "::error::could not reach the platform-update inbox at all (DNS/TLS/connection refused). No release event was delivered, so no satellite will re-bake against $SHA until its next schedule poll."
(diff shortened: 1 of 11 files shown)


rbuergi:
## The instance half is provisioned — and measuring it caught a defect in my own first cut

Read through MCP on the control instance (`memex.systemorph.com`), the two `Hosting/Deployment`
records that matter:

| record | `webhookInboxTargets` | holds `Hosting__PlatformWebhookSecret`? |
|---|---|---|
| `Deployments/memex` (memex.systemorph.com) | `[]` — none; its targets come from chart values | no |
| `Deployments/memex-cloud` (memex.meshweaver.cloud) | `["Store/Payments", "Hosting/PlatformBuilds"]` | **yes** |

So the instance that actually receives platform-build and publication records is **memex-cloud**,
and on it `HelmValues` renders **`__0 = Store/Payments`** (Stripe) and **`__1 =
Hosting/PlatformBuilds`**.

🚨 My first commit shipped `WebhookInbox__Targets__0__SecretConfigKey:
"Hosting:PlatformWebhookSecret"` as a **chart default**. On that instance it would have demanded a
GitHub-style `X-Hub-Signature-256` from **Stripe** — which signs `Stripe-Signature` — and **401'd
every payment delivery**. Corrected in 9001d375e: the chart renders both slots' declarations and
defaults **both to empty**, because the declaration is paired with a SLOT and slot `__0` is not the
same target on every deployment.

`ReleaseDeliveryChainGuard` needed no change, which is the point of how it was written: it asserts
the key is *renderable*, never that it is non-empty. That carve-out was written for exactly this
case and is now load-bearing.

### What was provisioned

`Deployments/memex-cloud` v24 → v25, one key added to `extraPortalConfig` (deep-merged; all eight
pre-existing extras verified intact afterwards):

```diff
   "webhookInboxTargets": ["Store/Payments", "Hosting/PlatformBuilds"],
   "extraPortalConfig": {
     ...
     "PreWarm__GateReadiness": "true",
+    "WebhookInbox__Targets__1__SecretConfigKey": "Hosting:PlatformWebhookSecret"
   }
```

Index **1** — the slot that carries `Hosting/PlatformBuilds`. It names a KEY; no secret value was
read or written. `extraPortalConfig` is the sanctioned rung here: `WebhookInbox__Targets__*` is not
in `HelmValues.TypedPortalKeys`, so it passes through without colliding with the typed surface.

**This is inert until this PR merges**: the ConfigMap template carries no catch-all range, so a
values key it does not render is silently dropped. Landing the instance half first is the safe
order — the reverse (lane asserting before the instance declares) is what the old
`Doc/Architecture/WebhookInbox` correctly refused to do.

### Sequence to close #3312 completely

1. this PR merges → chart renders the key, endpoint verifies when declared;
2. memex-cloud rolls → the declaration takes effect and both lanes start reporting
   `signature: verified` instead of `::warning:: not-required`;
3. flip that warning to an error in both lanes — one line each. That is the last step, and it is
   only safe once (2) is observed, not assumed.

Issue #3310, opened 2026-09-04
The module identity guards are config-level: nothing asserts that a run's bundles actually state ONE identity


rbuergi:
## The gap

#3293 and #3306 between them make the module identity anchor correct: built from the pinned platform
source (#3293), and without the module's `-p:Version` so every job in a run produces byte-identical
bytes (#3306). Both are guarded — but **both guards are CONFIG-level**:

- `ModuleIdentityPublishGuard` asserts the lane's *text*: the anchor is built from
  `$GITHUB_WORKSPACE/meshweaver/src/MeshWeaver.Compiler`, never `$PACKDIR`, and that command line
  carries no `-p:Version`.
- `ModulePackCommand` refuses an anchor resolving inside the module directory — behavioural, but
  about the anchor's LOCATION, not about the identity the run ends up stating.

Neither can see the OUTCOME fail. A different mechanism that reintroduces a per-module identity —
another per-module property reaching that command line (`-p:InformationalVersion`,
`-p:SourceRevisionId`, a `Directory.Build.props` edit keyed on the module), a future arm that derives
the anchor some third way, or a caller passing `--framework-mvid` per module — would leave every
existing guard green.

That matters more than usual because the failure is **silent**: the bundle packs, the manifest
carries a well-formed 32-hex identity, and every downstream assertion that the field is non-blank
passes. It is exactly the shape that reached production twice already (#3211's `(unrecorded)` fleet,
then run `33874892203`'s two identities for one platform).

## The check that would catch it

One assertion, at the only place that sees the whole wave: **every bundle a run packs states the SAME
framework identity.** That is true by definition of what the field means — the identity names the
platform build, and a run has one platform.

Sketch:

1. **Record it.** The `Drop the receipt` step already has the value in hand — the inspection step
   reads `frameworkMvid` back off the packed manifest for its non-blank assertion. Add it to the
   receipt JSON alongside `compiler` and `ledger`.
2. **Assert it.** `verify` (`All selected bundles built`) already downloads every receipt for the
   lane. Fail when the distinct identity count over the lane's receipts is > 1, naming the modules
   and their identities.
3. **Scope.** The assertion holds per LANE (per `workflow_call`), which is the unit that pins one
   platform. A repo calling the lane twice (Plugins `modules-floor` + `modules-rest`) has two lanes
   and may legitimately differ if their pins differ — the lane key already scopes the receipt glob,
   so this falls out.

## Why it was not done in #3306

[...]

Issue #3333, opened 2026-09-05
Two teardown/disposal races red one Build-and-Test run: a mesh shutting down mid-bake, and a faulted change-feed entry whose release does not claim it


rbuergi:
Run [33937167117](https://github.com/Systemorph/MeshWeaver/actions/runs/33937167117) (PR #3332, head `4c98498a`) failed **two** tests in two different shards. Both are teardown/disposal races, neither is reachable from that PR's diff, and the base rate says they are not routine — so they want an owner rather than a re-run.

## The two failures

**1. `MeshWeaver.PluginTester.Test.BakeEquivalenceTest.MeshDrivenAndCompilerDrivenBakes_ProduceTheSameArtifacts`** (shard 3, [check](https://github.com/Systemorph/MeshWeaver/runs/101227994551))

```
Assert.Null() Failure: Value is not null
Actual:   "InvalidOperationException: bake: 'Widget/Thing' cl"···
  at BakeEquivalenceTest.MeshDrivenAndCompilerDrivenBakes_ProduceTheSameArtifacts() … line 225
```

The message is truncated by xUnit, but the job log says what happened. The compiler-driven bake completed both packages. The mesh-driven bake wrote `Lib.zip` and then died on `Widget`, and the next line in the log is the cause:

```
warn: MeshWeaver.Messaging.MessageService[0] Message delivery failed for CreateNodeResponse
(ID: KFydCE9clUadxTFrgWsW8Q) in portal/nodeops-O0ixdCgnbkawTtQz-bUBQw:
Hub portal/nodeops-O0ixdCgnbkawTtQz-bUBQw cannot route CreateNodeResponse to Widget/Thing —
its parent hub mesh/O0ixdCgnbkawTtQz-bUBQw is shutting down (RunLevel=DisposeHostedHubs).
```

The mesh began tearing down while the bake still had work in flight. That is the same hub and the same phase as #2543 (*"Bake wedge: portal/nodeops stops draining mid-bake"*) — likely the same defect surfacing as a failed assertion instead of a wedge.

**2. `MeshWeaver.Hosting.Test.ChangeFeedResetReleasesUpstreamTest.ControlArm_ReleaseIfUnwatched_OfTheSameFaultedEntry_DisposesAnUpstream`** (shard 4, [check](https://github.com/Systemorph/MeshWeaver/runs/101228247462))

```
AssertionException : Expected value to be true because nothing is subscribed to the faulted
[...]

PR #3339, opened 2026-09-05
State the framework identity as the platform commit instead of deriving it per job (#3308)

### LINKED ISSUE #3308: A source-built bundle's frameworkMvid is per-MODULE and unmatchable by any consumer
## What is fixed, and what is not

#3293 landed and CD's `no identity anchor for MeshWeaver.Maps` is gone — the source arm now builds `MeshWeaver.Compiler` from the platform tree instead of scavenging it out of the module's publish output. That was the outage and it is over.

What it does **not** fix is what the recorded identity *means*. The build passes `-p:Version="$VERSION"` — the **module package's** version, different for every matrix entry — so `MeshWeaver.Compiler.dll` is compiled afresh per module and its MVID differs per module. The field stays a property of the MODULE, which is the thing #3211 exists to prevent.

## The measurement

CD run **33867996265**, one run, one platform commit (`7d644de95`), from the copies publish happened to drop:

|---|---|

Two identities for one platform build. Nothing downstream can use that: a consumer comparing "what I have landed" against a per-module value gets a different answer per module.

## And neither value can match a consumer at all

A portal image is built with `-p:CIRun=true` on GitHub Actions, so `AddCommitHashMetadata` stamps `AssemblyMetadata("MeshWeaverFrameworkIdentity", "g<sha>")` on its `MeshWeaver.Compiler`, and `FrameworkIdentity.ReadIdentity` **prefers that stamp over the MVID** (#1660 WS3). So a consumer reads `g<sha>` and a source-built bundle states an MVID: never equal, by construction.

`ModuleUpdateDecision` (#3154) compares `(version, framework identity)` before answering `SkipUpToDate`. With an unmatchable identity that comparison cannot do its job, and the reconcile answers "up to date, identity could not be checked" — the exact failure #3211's own error text describes.

## The fix that was measured to work

State it instead of deriving it. In the source arm the platform IS the checkout, so its identity is the commit:

```bash
sha="$(git -C "$GITHUB_WORKSPACE/meshweaver" rev-parse HEAD)"
identity=(--framework-mvid "g$sha")
```

`--framework-mvid` is the packer's own flag for "the platform's own reading", and `g<sha>` is one of the three tokens it documents (`s<hash>`, `g<sha>`, 32-hex MVID). One identity per platform commit, identical for every module in the run, and the same token the image stamps — so the comparison starts working. It also needs no extra compiler build per matrix entry.

That change is written and was green (28/28) as #3296, which I am closing in favour of #3293 rather than reverting a landed fix. The diff is three lines of shell plus the guard and `RECIPE_VERSION` that must move with it (`module-build-key.py` — bundles packed under the old recipe attest an identity no consumer can match, so they must be repacked rather than reused).

## Suggested acceptance

A bundle packed by core's own CD and a portal image built from the same commit state the **same** framework identity — assert it once, rather than leaving it to be re-measured.

rbuergi:
## This is red on core `main` right now

Run **33950389008** (2026-09-05 07:05Z) — four entries, **one** platform commit, every leg
`plan: BUILD` (no reuse) — stated **three** identities, and #3310's agreement gate refused the lane.
Core CD has failed on it repeatedly today; the one green run in between **skipped** the module-pack
job entirely, so it attested nothing.

| module | compiler builds in the job | stated identity |
|---|---|---|
| `MeshWeaver.Maps` | **1** | `0051e7210e4144a6a18620a36e629d84` |
| `MeshWeaver.Payments.Stripe` | **1** | `0051e7210e4144a6a18620a36e629d84` |
| `MeshWeaver.Markdown.Collaboration` | **2** | `6553ea371eab49e4afdd0af20ea7ac59` |
| `MeshWeaver.AI` | **2** | `395909c22d3042d4b23fb055337db63b` |

## Why the two previous cures could not have worked

The split is **exactly the reference graph**, and the lane's own comment already says so:
`MeshWeaver.AI` and `MeshWeaver.Markdown.Collaboration` reach `MeshWeaver.Compiler` transitively.
So the **module** build compiles it first — into the same output path, under the module's own
global properties — and the anchor build that follows keeps those bytes. Maps and Payments.Stripe
never build it, so their anchor build compiles it clean and they agree with each other.

- **#3293** moved the anchor build into the platform tree — correct location, still per job.
- **#3306** removed the `-p:Version` polluting that build — correct property, still per job.

Both fixed the **anchor** build. The pollution arrives from the **module** build upstream, which is
entitled to its own version. No property removable from the anchor build can reach it.

## And determinism alone would not have been enough

A portal image is built with `-p:CIRun=true`, so `AddCommitHashMetadata` stamps
`AssemblyMetadata("MeshWeaverFrameworkIdentity", "g<sha>")`, and `FrameworkIdentity.ReadIdentity`
**prefers that stamp over the MVID**. A consumer reads `g<sha>`; a source-built bundle stated
32-hex. **Never equal, by construction** — so #3154's `(version, identity)` comparison could only
ever answer *"identity could not be checked"*, which is the exact state #3211 exists to prevent.
Even a perfectly stable MVID would have been unmatchable.

## The change

The from-source arm stops deriving and **states** it. The platform IS this checkout, its commit IS
what a consumer reads, and `--framework-mvid` is the packer's own flag for that — its own error
text names `g<sha>` as a legal token, and the only validation is "no whitespace".

```bash
sha="$(git -C "$GITHUB_WORKSPACE/meshweaver" rev-parse HEAD 2>/dev/null || true)"
# … 40-hex or RED, never a fallback …
identity=(--framework-mvid "g$sha")
```

One identity per platform commit, identical for every entry, and **no compiler build per matrix
job at all** — which also removes a full `dotnet build` from every leg.

The **refs arm is unchanged**: there the DLL genuinely *is* the platform, is already stamped, and
already yields `g<sha>`. This makes the two arms agree rather than inventing a third shape.

`RECIPE_VERSION` 2 → 3, because bundles packed under recipe 2 state an identity no consumer can
match and must be **repacked**, not reused.

## The guard moves with its subject, and gets stronger

It used to pin *where* the anchor build lived — a config-level assertion that was green through all
three incarnations of this bug. It now forbids `MeshWeaver.Compiler.csproj` in the pack step
**at all**: the per-job build is the mechanism every previous cure left in place, and it packs
GREEN while handing #3154 a value no consumer can match.

**Mutation-proved:** reintroducing that build turns the guard red (`Assert.DoesNotContain() Failure:
Sub-string found`), and it passes with it gone.

## Verification

- `MeshWeaver.Documentation.Test` — `0 Error(s)`, Release, `-warnaserror`.
- Identity guards **5/5**; `node-repo-pack-verify.py --self-test` passes.
- The acceptance #3308 asked for is now enforced by #3310's gate rather than re-measured by hand:
  a lane whose bundles disagree is refused, and after this they cannot disagree — every entry
  states the same commit.

Closes #3308

### DIFF
--- a/.github/scripts/module-build-key.py
+++ b/.github/scripts/module-build-key.py
@@ -73,7 +73,11 @@ from pathlib import Path
 # genuinely differ — and a REUSED pre-#3211 bundle would be refused at the publish step, which is the
 # right verdict for those bytes and the wrong thing to spend a run discovering. One rebuild wave, and
 # every recorded key afterwards attests bytes that state what they were built against.
-RECIPE_VERSION = "2"
+# 3: the from-source arm STATES the framework identity as the platform commit (`g<sha>`)
+# instead of deriving an MVID from a compiler DLL built per matrix job (#3308). Bundles packed
+# under recipe 2 state an identity no consumer can match, so they must be REPACKED rather than
+# reused — which is exactly what changing this constant forces.
+RECIPE_VERSION = "3"
 
 # Repo-level files that change what EVERY project compiles or tests. Presence and content both
 # count; a file the repo does not have hashes as null so adding it later changes the key.
--- a/.github/workflows/node-repo-module-pack.yml
+++ b/.github/workflows/node-repo-module-pack.yml
@@ -1753,51 +1753,60 @@ jobs:
           if [ -n "$REFS" ]; then
             anchor="$REFS/MeshWeaver.Compiler.dll"
             where="the pinned platform image's extracted /app"
+            if [ ! -f "$anchor" ]; then
+              echo "::error::no identity anchor for $MODULE — expected MeshWeaver.Compiler.dll at '$anchor', i.e. in $where. Every bundle must state the framework build its bytes were compiled against (#3211): a consumer compares it against what it has landed, and an unstated one makes every reconcile answer 'up to date, identity could not be checked' forever. Nothing here may guess it, so the pack stops."
+              exit 1
+            fi
+            # The image's own compiler, stamped by the build that produced the image, so
+            # FrameworkIdentity.ReadIdentity returns its `g<sha>` — the same token a portal image
+            # reports. Reading it off the DLL is right HERE because the DLL is the platform.
+            identity=(--graph-dll "$anchor")
           else
-            # 🚨 BUILT from the platform source, never scavenged out of $PACKDIR. Publish drops
-            # MeshWeaver.Compiler.dll beside the module ONLY when the module transitively
-            # ProjectReferences it — MeshWeaver.AI and MeshWeaver.Markdown.Collaboration both do,
-            # through MeshWeaver.Graph, and they were the ONLY two entries this lane built when
-            # the anchor landed, so "the platform ProjectReferences are real, so it IS beside the
-            # module" was measured on a population of two and read as a property of the arm.
-            # It is a property of the REFERENCE GRAPH: core's compose set grew to four on
-            # 2026-09-04 (#3290) and both new entries reach no compiler — MeshWeaver.Maps stops at
-            # MeshWeaver.Layout, MeshWeaver.Payments.Stripe is a leaf — so both were RED on their
-            # first run and took `CD delivered, or had a good reason not to` down with them for
-            # two hours (#3293).
-            anchordir="$GITHUB_WORKSPACE/meshweaver/src/MeshWeaver.Compiler"
-            # -warnaserror to match the module build above: when a module DOES reach the
-            # compiler transitively, that build is what compiles it, so omitting the flag here
-            # would let MeshWeaver.Maps accept a compiler build MeshWeaver.AI rejects — the same
-            # per-module asymmetry this change exists to remove.
+            # 🚨 STATE the identity; never DERIVE it from a DLL this job built (#3308).
             #
-            # 🚨 AND DELIBERATELY **NO** `-p:Version="$VERSION"` (#3176). #3293 passed it, to make
-            # this "the assembly publish would have copied" — but that is the one global property
-            # that must NOT reach here, because $VERSION is the MODULE's package version and the
-            # identity must be the PLATFORM's. MSBuild writes Version into the assembly's version
-            # attributes, so it lands in the metadata the MVID is computed over. Measured on THIS
-            # command, with the two package versions core CD packs, plus a control:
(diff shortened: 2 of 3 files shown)


rbuergi:
## The two reds, attributed

### 1. `CI's own shell` — mine, and the gate was right

`check-workflow-shell.py`:

> `[swallow]` a CAPTURED command substitution swallows both stderr and the exit status (step 'Pack
> the module bundle') — a failed call becomes an empty answer with no trace.

`sha="$(git … rev-parse HEAD 2>/dev/null || true)"` threw away git's message *and* its exit status.
The refusal downstream was still fail-closed, but a checkout without `.git` would have arrived at it
as a blank string with nothing in the log saying why — which is the same "cannot tell 'looked and
found nothing' from 'could not look'" shape this PR is about, one level down.

Fixed by letting it fail out loud:

```bash
if ! sha="$(git -C "$GITHUB_WORKSPACE/meshweaver" rev-parse HEAD)"; then
  echo "::error::… git rev-parse HEAD failed and its message is above …"
  exit 1
fi
if [ ${#sha} -ne 40 ] || [ -n "${sha//[0-9a-f]/}" ]; then …
```

`set -e` stays out of the way until the explanation prints, git's stderr reaches the log, and the
shape check is now a real 40-hex assertion instead of an 8-character prefix glob.

Verified locally: `check-workflow-shell.py` → **0 live findings** (the one remaining is the
pre-existing allow-listed entry). `Documentation.Test` `0 Error(s) 0 Warning(s)`,
`ModuleIdentityPublishGuard` 5/5.

### 2. `Run tests (shard 3)` — not this diff

`BakeEquivalenceTest.MeshDrivenAndCompilerDrivenBakes_ProduceTheSameArtifacts` failed inside a
**teardown/disposal race**, not on anything this PR touches:

```
HubDisposedBeforeResponseException: Hub Widget/Thing was disposed before the response arrived
ObjectDisposedException: Hub Widget/Thing is shutting down — cannot register new response subject
HostedHubsCollection: Rejecting hosted hub creation … during disposal - collection is disposing
[ReleasePostCondition] Widget/Thing: … the node names NO release at all
```

That is **#3333** ("a mesh shutting down mid-bake, and a faulted change-feed entry whose release
does not claim it"), already filed. This diff changes a workflow, one Python constant
(`RECIPE_VERSION`) and a guard test — it reaches no code in `MeshWeaver.PluginTester`, and the same
suite passed on #3337's run from the same base. Not re-running it to see whether it goes green: that
would only hide the race #3333 exists to fix.

PR #3343, opened 2026-09-05
feat(storage): DeleteMany — a subtree delete stops paying one round-trip per row

rbuergi:
## The measurement

Retiring `UWDeepfield` on memex, 2026-09-05: **9,834 nodes in ~8 minutes — about 21 nodes a second**, and the profile is flat. The subtree was not big; the per-node cost was. On Postgres every node is its own single-row `DELETE`, its own implicit transaction, and its own turn on a capacity-1 write pool.

## What this adds

`IStorageAdapter.DeleteMany` — the DELETE-side twin of `WriteMany`. Default is correct everywhere; the Postgres override groups paths by target table and sends one `DELETE … WHERE path = ANY($1) RETURNING path` per table, in **one** write-pool turn.

Two contract points, both pinned by tests:

- **The result is what was ACTUALLY removed**, not an echo of the request. The default is built on `DeleteIfExists`, not `Delete` (which emits its path either way), because `PersistenceService` unions these results to learn what really left storage — an echo would let the composite claim a delete no provider performed. `RETURNING path` gives Postgres the same exactness.
- **Order is part of the contract, INVERTED from `WriteMany`.** Callers pass children before parents, and notifications are published in that order. Batching the statement is safe; batching the feed's ordering is not, because the feed is what wakes the per-node hubs.

Forwarded through the production decorator stack and the composite, with a test that fails if any layer stops forwarding — a decorator that falls back to the default silently gives the 8 minutes back.

## Where it is wired

The two paths that already delete an authoritative, already-committed path set:
- `DeleteBatchRequest`'s handler, which has always carried a list and spent it one `Delete` at a time;
- `HandleMoveNodeRequest`'s source-subtree removal.

## What it deliberately does NOT change

`HandleDeleteNodeRequest`'s commit fan-out. That is where the remaining **2N hub round-trips** live (one `ValidateDeleteRequest` + one `DeleteNodeRequest` per descendant, each needing that node's hub to activate, ~30 ms/node) and it is the rest of the win. Replacing it is correct by the handler's own reasoning — authorization is fully decided in the permission and pre-validate stages before any mutation, and the commit already runs as System — but it needs `DeleteNodeBehaviorTest`, `RecursiveDeletePermissionTest` and `RecursiveDeleteAuthoritativeTest`, which live in the sibling repo and are not runnable from here. **It must not go in unverified**, so it is a follow-up.

## Verification

`MeshWeaver.Hosting.Test`: **356 passed, 0 failed**, including 6 new `StorageAdapterDeleteManyTest` cases and the existing `HierarchicalPathDeletionTests`.

Note: the Postgres override lives in the sibling repo (`src/MeshWeaver.Hosting.PostgreSql` is a symlink) and ships as a matching MeshWeaver.Plugins PR.

### DIFF
--- a/src/MeshWeaver.AI.AzureFoundry/MeshWeaver.AI.AzureFoundry
+++ b/src/MeshWeaver.AI.AzureFoundry/MeshWeaver.AI.AzureFoundry
@@ -1,0 +1,1 @@
+/Users/roland/code/MeshWeaver.Plugins/src/MeshWeaver.AI.AzureFoundry
--- a/src/MeshWeaver.AI.AzureFoundry/MeshWeaver.AI.AzureFoundry
+++ b/src/MeshWeaver.AI.AzureFoundry/MeshWeaver.AI.AzureFoundry
@@ -1,1 +1,0 @@
-/Users/roland/code/MeshWeaver.Plugins/src/MeshWeaver.AI.AzureFoundry
--- a/src/MeshWeaver.AI.ClaudeCode/MeshWeaver.AI.ClaudeCode
+++ b/src/MeshWeaver.AI.ClaudeCode/MeshWeaver.AI.ClaudeCode
@@ -1,0 +1,1 @@
+/Users/roland/code/MeshWeaver.Plugins/src/MeshWeaver.AI.ClaudeCode
--- a/src/MeshWeaver.AI.ClaudeCode/MeshWeaver.AI.ClaudeCode
+++ b/src/MeshWeaver.AI.ClaudeCode/MeshWeaver.AI.ClaudeCode
@@ -1,1 +1,0 @@
-/Users/roland/code/MeshWeaver.Plugins/src/MeshWeaver.AI.ClaudeCode
--- a/src/MeshWeaver.AI.Copilot/MeshWeaver.AI.Copilot
+++ b/src/MeshWeaver.AI.Copilot/MeshWeaver.AI.Copilot
@@ -1,0 +1,1 @@
+/Users/roland/code/MeshWeaver.Plugins/src/MeshWeaver.AI.Copilot
--- a/src/MeshWeaver.AI.Copilot/MeshWeaver.AI.Copilot
+++ b/src/MeshWeaver.AI.Copilot/MeshWeaver.AI.Copilot
@@ -1,1 +1,0 @@
-/Users/roland/code/MeshWeaver.Plugins/src/MeshWeaver.AI.Copilot
--- a/src/MeshWeaver.AI.OpenAI/MeshWeaver.AI.OpenAI
+++ b/src/MeshWeaver.AI.OpenAI/MeshWeaver.AI.OpenAI
@@ -1,0 +1,1 @@
+/Users/roland/code/MeshWeaver.Plugins/src/MeshWeaver.AI.OpenAI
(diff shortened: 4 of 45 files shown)


rbuergi:
**This PR's red is not a merge conflict, and it is not `DeleteMany`. It was 37 stray symlinks — removed in `a72751f7d`.**

## What was actually failing

`NoConflictMarkersGuard` does not report a conflict marker here. It **crashes**:

```
System.IO.FileNotFoundException : Could not find file
  '/home/runner/work/MeshWeaver/MeshWeaver/src/MeshWeaver.AI.AzureFoundry/MeshWeaver.AI.AzureFoundry'
   at NoConflictMarkersGuard.LooksBinary(String path)
   at NoConflictMarkersGuard.NoTrackedFileCarriesAnUnresolvedConflictMarker()
```

`File.Exists` answers true for the tracked entry, `File.OpenRead` throws on the dangling target, and
the guard dies on the first one — never reaching its assertion. So the failure *names* a
conflict-marker guard and is nothing of the kind, which is why it reads as unrelated to this PR.

## What the entries were

```
$ git ls-tree -r <head> | awk '$1=="120000"' | wc -l
37
$ git cat-file -p <blob>
/Users/roland/code/MeshWeaver.Plugins/src/MeshWeaver.AI.AzureFoundry
```

All 37 are **absolute paths into one machine's checkout** — the local dev-convenience links that sit
untracked in a working copy (`?? src/MeshWeaver.AI/`, `?? src/MeshWeaver.Blazor.Portal/`, …), swept
in by a `git add -A`. `main` carries **zero** symlinks, so nothing else in the fleet has them.

**37 of this PR's 45 changed files were these links.** The `DeleteMany` work is untouched — I removed
only mode-`120000` entries, and verified the diff against your previous head is exactly `37 files
changed, 37 deletions(-)`.

## Two things worth doing separately, not here

1. **The guard should not crash on a dangling symlink.** It exists to report offenders; a
   `FileNotFoundException` from `LooksBinary` turns a precise verdict into a stack trace that points
   at the wrong problem. A `catch (IOException) { continue; }` there — or skipping mode-`120000`
   entries outright — makes it say what it means. Worth its own PR; I have not touched the guard.
2. **A `.gitignore` entry** for these link paths would stop the next `git add -A` doing it again.

Sorry for touching your branch without asking — you had no unpushed work (worktree HEAD matched the
PR head) and the removals are mechanical and reversible. Revert `a72751f7d` if you disagree.

rbuergi:
**Correction to my comment above — I mis-attributed where those symlinks came from.**

I wrote that they were "the local dev-convenience links that sit untracked in a working copy". That
was an inference, and it is wrong:

* the primary checkout does **not** contain them. `src/MeshWeaver.AI/` there holds only stale
  `bin/` + `obj/` from before the AI exit (#2276) — the `?? src/MeshWeaver.AI/` entries I reasoned
  from are build leftovers, not symlinks;
* the links were introduced by **`aea5bdc9c` itself** — this PR's own single commit, 09-05 11:25 —
  alongside the `DeleteMany` work. They did not pre-exist anywhere.

So they were created by whatever produced this branch, most likely to build core against the Plugins
tree locally, and then swept in by a `git add -A`. Nothing about them originated in the maintainer's
working copy, and I should not have said so.

Everything else in that comment stands and is unaffected: 37 mode-`120000` entries, all absolute
paths into `/Users/roland/code/MeshWeaver.Plugins/src/…`, all removed in `a72751f7d`, with the
`DeleteMany` work untouched (`37 files changed, 37 deletions(-)` against your previous head).

The suggestion about a `.gitignore` entry also changes shape given this: it is not about protecting a
pre-existing local convention, it is about making a cross-repo build link uncommittable in the first
place. #3365 adds the other half — the guard now names a tracked symlink instead of crashing on it.

Issue #1349, opened 2026-08-12
A course root that ships its own NodeTypes stops serving its layout areas after a recycle — the Store paywall never paints, while a NodeType-free root recycled seconds earlier serves in 4.9s


rbuergi:
## Symptom

A course root that **ships its own NodeTypes** does not serve its `Store/Plugin` layout areas — the
`Subscribe` paywall and the cover's `CoverCta` — to a viewer, for **minutes**, while a NodeType-free
root recycled in the same loop seconds earlier serves in **4.9 s**. The portal is healthy throughout:
`Running=true OOMKilled=false Dead=false RestartCount=0`, and unrelated courses pass their full
journey on it during the dark window.

This is not a wedge, not OOM, not a dead container, and — since MeshWeaver#1199 — not the old
"activation parks behind an in-flight compile" behaviour either.

## The discriminator is exact

Across three CI runs of the `Systemorph/education` disposable-mesh suite, the split is clean and has
no exception:

| ships own NodeTypes | courses | `/{course}/Subscribe` |
|---|---|---|
| **yes** | AgenticOffice (4 types), AgenticPrimer (2), AgenticPrimerDe (2) | **never paints** |
| no | SwissSolvencyTest, DataModeling, AdvancedBusinessRules, RiskTransfer, ThinkInStreams, ReinsurancePractice | paints in **5–10 s** |

Not language (AgenticPrimer is English, AgenticPrimerDe German), not course kind (AgenticOffice is an
exercise course, the primers are content courses), not size (AgenticOffice 91 files vs
SwissSolvencyTest 71), not root count (AgenticOffice declares 4 install roots; DataModeling declares
22 and is fine).

## The controlled observation

Run [31637704027](https://github.com/Systemorph/education/actions/runs/31637704027), shard 3. Both
course roots are recycled by the same bootstrap loop, seconds apart:

```
+482s  20:37:13   AgenticOffice imported (content renders)     ← admin reads a GATED page: root serving
+550s  20:38:21   SwissSolvencyTest imported (content renders)
+581s  20:38:53   course roots recycled                        ← both, same loop
       20:38:54 → 20:48:04   /AgenticOffice/Subscribe paints NONE of its three states,
                             2 × 275 s (10 polls each, 27.6 s per poll)
       20:48:04 → 20:51:05   SwissSolvencyTest: cover CTA 4.9 s, install 35.9 s,
                             copies 19.8 s, exercises 16.3 s — all green
```

[...]

PR #3344, opened 2026-09-05
Withdraw the central SQLitePCLRaw pins — #1351 fixed this better, and one of mine was a latent CVE regression

rbuergi:
## Withdrawing my own merged change

#3342 added central `PackageVersion` pins for the managed SQLitePCLRaw trio.
**MeshWeaver.Plugins#1351 landed the real fix concurrently**, and it is better on both counts it
differs by. I did not check for a sibling before starting, which is how two people ended up on the
same defect.

### 1. `VersionOverride` beside the reason beats a central pin

#1351 puts `VersionOverride="3.0.2"` on `SQLitePCLRaw.core` / `provider.e_sqlite3` directly in
`MeshWeaver.Hosting.Sqlite.csproj`, with the argument spelled out there:

> nothing in the platform's own `src/` consumes SQLitePCLRaw, and a central pin would only take
> effect here anyway — while forcing this repository to move its platform pin in the same change to
> see it. The version belongs beside the reason.

That is right, and my own PR measured the first half of it: building `MeshWeaver.Hosting.Sqlite`
against the #3342 tree still resolved `2.1.11`, because a bare `PackageVersion` does not bind a pure
transitive. My change was inert *and* gated on a pin bump; theirs works immediately.

### 2. 🚨 My `bundle_e_sqlite3` pin was a latent CVE regression

#1351 warns against raising it "to finish the family". The restore graph confirms it:

```
SQLitePCLRaw.bundle_e_sqlite3 3.0.2
  → SQLitePCLRaw.config.e_sqlite3 3.0.2
  → SourceGear.sqlite3 3.50.4.2          ← NOT SQLitePCLRaw.lib.e_sqlite3
```

The 3.0 bundle swaps the **native engine package**, so the `lib.e_sqlite3` pin that exists for
GHSA-2m69-gcr7-jv3q stops being the thing supplying the engine. Inert today for the same
pure-transitive reason — which makes it a loaded gun rather than a live break, and exactly the kind
of thing I objected to leaving in place elsewhere today.

**Why my probe missed it, which is the part worth keeping:** the probe referenced
`lib.e_sqlite3 3.53.3` *explicitly* alongside the bundle, so `sqlite_version()` answered `3.53.3`
and the swap was masked. Its own `project.assets.json` restored **both** `SourceGear.sqlite3
3.50.4.2` and `lib.e_sqlite3 3.53.3` — the evidence was in the file and I read only the runtime
answer. A control held by accident rather than by design turns a green probe into a wrong
conclusion.

## What lands here

The three pins go. The comment keeps both findings so the next person reaching for "finish the
family" meets the measurement instead of repeating it:

- the managed trio is deliberately *not* pinned centrally, and where the fix actually lives;
- never pin `bundle_e_sqlite3` to 3.0.x, and why.

`MeshWeaver.Graph.Contract` — `0 Error(s) 0 Warning(s)`, Release, `-warnaserror`.

Refs #3328, MeshWeaver.Plugins#1351

### DIFF
--- a/Directory.Packages.props
+++ b/Directory.Packages.props
@@ -101,55 +101,28 @@
     <PackageVersion Include="Microsoft.AspNetCore.DataProtection" Version="10.0.10" />
     <PackageVersion Include="Microsoft.AspNetCore.SignalR.Client" Version="10.0.10" />
     <PackageVersion Include="Microsoft.Data.Sqlite" Version="10.0.10" />
-    <!-- 🚨 THE SQLitePCLRaw FAMILY IS PINNED AS A SET — native AND the managed trio (#3328).
-         Pin the patched native engine first: Microsoft.Data.Sqlite pulls
+    <!-- Pin the patched native SQLite engine: Microsoft.Data.Sqlite pulls
          SQLitePCLRaw.lib.e_sqlite3 2.1.11 transitively, which bundles a SQLite engine with the
          high-severity vuln GHSA-2m69-gcr7-jv3q (NU1903). The native engine package is versioned to
          track the bundled SQLite version, so any 3.50+ engine is past the vulnerable range.
-         Deliberately version-AGNOSTIC prose: naming a specific version here goes stale the moment a
-         pin below is bumped (it once said 3.50.3 while the pin read 3.53.3). The pins are the
-         source of truth; this comment explains WHY they exist.
+         Deliberately version-AGNOSTIC prose: naming a specific version here goes stale the moment
+         the pin below is bumped (it once said 3.50.3 while the pin read 3.53.3). The pin is the
+         source of truth; this comment explains WHY it exists.
+         Pinned explicitly in the Sqlite projects.
 
-         🚨 AND THE MANAGED TRIO IS PINNED TOO, at the 3.x family — because leaving it unpinned made
-         the PLATFORM IMAGE internally inconsistent, and that fails EVERY module compile in EVERY
-         satellite with no repo change to blame. Measured on the 3.0.0-rc9.ci.7779 image
-         (MeshWeaver.Plugins#1349, four bundles):
+         🚨 THE MANAGED TRIO IS DELIBERATELY *NOT* PINNED HERE (#3328, #3342 withdrawn). The image's
+         Roslyn binds SQLitePCLRaw.core 3.0.2 while declaring no package dependency on it, so /app
+         shipped 2.1.11 and every module bundle in every satellite died on MSB3277. The fix lives
+         beside its reason, as a VersionOverride in MeshWeaver.Plugins'
+         MeshWeaver.Hosting.Sqlite.csproj (Plugins#1351) — nothing in this repo's src/ consumes
+         SQLitePCLRaw, so a central pin here is inert for a pure transitive AND would force that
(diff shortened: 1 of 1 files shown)


rbuergi:
🚨 **This withdrawal took core CD down, and the reason is one reference #1351 did not cover.**

The reasoning here is right — the version belongs beside the reason, not in a central list this repo
does not consume. But the withdrawal removed **four** pins, and Plugins#1351 only replaced **two**:

| pin withdrawn here | replaced in Plugins#1351? |
|---|---|
| `SQLitePCLRaw.core` 3.0.2 | ✅ `VersionOverride="3.0.2"` |
| `SQLitePCLRaw.provider.e_sqlite3` 3.0.2 | ✅ `VersionOverride="3.0.2"` |
| `SQLitePCLRaw.bundle_e_sqlite3` 3.0.2 | n/a — deliberately not pinned |
| **`SQLitePCLRaw.lib.e_sqlite3` 3.53.3** | ❌ **nothing** |

`MeshWeaver.Hosting.Sqlite.csproj` carries a **versionless**
`<PackageReference Include="SQLitePCLRaw.lib.e_sqlite3" />` — the CVE pin — and its version came from
exactly the central entry withdrawn here (Plugins' `src/Directory.Packages.props` imports this list
through `$(MeshWeaverRoot)`). So since this merged, every `main-cd` run has failed:

```
MeshWeaver.Hosting.Sqlite.csproj : error NU1010: The following PackageReference items do not define
  a corresponding PackageVersion item: SQLitePCLRaw.lib.e_sqlite3
Memex.Portal.Distributed.csproj  : error NU1903: Warning As Error: Package
  'SQLitePCLRaw.lib.e_sqlite3' 2.1.11 has a known high severity vulnerability, GHSA-2m69-gcr7-jv3q
```

NU1010 for the missing version, then NU1903 because the fallback resolution reintroduced the very
2.1.11 that pin existed to avoid — the "latent CVE regression" this commit message set out to
prevent, arriving by the other door.

**ci.7813 and ci.7814 both SKIPPED `Promote: tag the full set`**, so nothing has been published since
and no image exists for the Plugins pin bump (#1349/#1359) to move to. That is the live blast radius:
core CD produces nothing at all.

**Fixed forward in Plugins#1361** — `VersionOverride="3.53.3"` on that reference, so it depends on
nothing outside its own repo. Verified on CD's own path (`-p:MeshWeaverRoot=<core checkout>`, which
is how CD builds the plugins tree): both projects restore clean, resolving `lib.e_sqlite3/3.53.3`
beside `core/3.0.2`, with `bundle_e_sqlite3` left at 2.1.11 as your note requires. **No revert needed
here** — the withdrawal itself is correct.

Worth recording for next time, because neither repo's CI could see it: a **versionless
`PackageReference` under Central Package Management is a cross-repo dependency**, and withdrawing a
central pin is a breaking change to every versionless reference that resolved through it. The
`Cross-repo pair (public surface)` gate cannot see this shape — it watches public types and members,
not package-version resolution — so it is an eighth break shape for
[CrossRepoPairGate.md](src/MeshWeaver.Documentation/Data/Architecture/CrossRepoPairGate.md). Plugins
CI compiles against the pinned IMAGE and never builds the portal host the way core CD does, so the
break could only ever surface here, one commit later.

Issue #1735, opened 2026-08-17
Per-environment content composition: what each environment pre-installs, expressed as feature flags, with required parameters routed through service discovery


rbuergi:
Maintainer-directed. Three asks, one mechanism:

1. **Configure what is pre-installed from ANY repo, per environment, in config.**
2. **Bake it into feature flags, and expose the flags by environment.**
3. **When a package needs additional parameters (a connection string, an endpoint), obtain them through something Aspire-like that can route to other services** — rather than each package inventing its own config lookup.

## What already exists (do not reinvent)

- `PluginCatalogOptions.InstallByDefault` — a list in `Source/Package` notation (`["Plugins/*"]`, `["Plugins/Store","Plugins/Essentials"]`). Already **source-scoped**, i.e. already "from any repo", and deliberately fails closed when a registry does not stamp the source. **But it SEEDS ONCE**, on an installation with no install records: it is not a policy that re-asserts, so it cannot express "this environment always has X".
- `PluginCatalogOptions.InstallPreInstalledPackages` — the platform baseline, **reconciled on EVERY boot**, cheap when up to date (content-identity gate → one catalog listing, no writes), and the only thing that heals a lost baseline (#902).
- `PluginCatalogOptions.Registries` — which registries an instance consumes, each with its own token.
- `Memex.Client.Services.FeatureFlags` — device-persisted MAUI `Preferences` flags. **Client-only by design** (no `IConfiguration` on-device). There is NO server-side feature-flag concept today.

So the gap is not "can we list packages per environment" — it is that the knob has seed-once semantics, there is no flag surface, and nothing declares or supplies a package's required parameters.

## Design constraints

**Environment dimension.** The real environments are the AKS namespaces (`memex`, `memex-cloud`, `atioz`), Aspire/Container-Apps modes, and local dev. Config reaches them through appsettings + env vars via helm today. Whatever is designed must be settable per environment WITHOUT a rebuild, and must not require hand-editing a live portal.

**Seed-once vs reconciled is a real semantic choice, not an oversight.** `InstallByDefault` seeds so an operator can later uninstall something without it coming back. A per-environment composition policy wants the opposite: it should re-assert. Do not silently change the existing knob's meaning — introduce the reconciled shape explicitly and say how the two relate (the existing doc comments spell out the reasoning; keep them true).

[...]

PR #3360, opened 2026-09-05
fix(memex-local): the mounted checkouts reconcile on every boot, not once

### LINKED ISSUE #3359: Self-registry install silently stops mirroring its mounted checkouts — the repos land on the seed-once lane, and neither auto-update mechanism applies
> **Why this repo (checked, not assumed).** Every surface that produces the defect, and every surface any of the three candidate fixes would touch, is owned by this repository:
> - `deploy/homebrew/bin/memex-local` — the line that puts the mounted repos on the seed lane (`--set pluginCatalog.installByDefault[i]=<Name>/*`)
> - `deploy/homebrew/share/values.local.self-registry.yaml` and `deploy/helm/` — the tracked layers where a reconciled-lane declaration would go
> - `src/MeshWeaver.PluginCatalog/InstanceAutoRegistrationService.cs` — the three selection lanes and the `Plugins/_DefaultInstallLedger` gate
> - `src/MeshWeaver.PluginCatalog/{CatalogLayoutAreas,PluginUpdateWatcher,RegistryUpdateReconciler}.cs` — the install orchestrator and both update mechanisms
>
> The two plausible alternatives were considered and rejected. **Memex** carries no copy of `memex-local`, the chart or `MeshWeaver.PluginCatalog`; its issues on this theme (`#153`, `#164`, `#165`) are incidents on a deployed instance, whereas this is a design defect in the mode itself, reproducible on any developer machine with no deployment involved. **MeshWeaver.Plugins** owns the packages that go stale but none of the machinery that fails to refresh them — the one Plugins-owned component in the story, `StoreManifestSource`, is a separate symptom and is left in Related rather than folded in.

### Summary

A `memex-local` install in self-registry mode (§16) mounts the developer's checkouts and serves packages from them. It installs them once and then never re-imports them: the mounted sources are configured on `pluginCatalog.installByDefault`, which is a **seed-once** lane, and neither mechanism that refreshes an installed package applies to this deployment shape. The portal quietly diverges from the working trees it exists to serve, with no error anywhere.

This is the self-registry twin of #1318 (*"A registry-only consumer never learns a module changed"*): that issue's fix, `RegistryUpdateReconciler`, states in its own doc that it does nothing when no registry is configured, and no other mechanism takes over. #1735 introduced the reconciled feature-flag lane referenced below; #2417 and Plugins#1244 are different limitations of the same deployment shape.

### Evidence (one machine, 2026-09-05)

1. Central `AgenticOffice` was last written **2026-08-31 11:58Z**. The Education checkout has moved repeatedly since; the course's lesson-1 `Quiz` node still carries no `key` fields, i.e. it predates Education#266 (merged 2026-09-04), which the mount has carried since the last sync.
2. The boot's install pass names **41 packages** — `AI, Edu, Publish, AgenticEngineering, Analysis, …` — and neither `Store` nor any course is among them. Both are on `Plugins/_DefaultInstallLedger.seeded`.
3. Nothing logs a problem: every reconciled package reports `0 written, 0 unchanged`, the correct "content hash unchanged" verdict, so the log reads healthy.

### Mechanism

1. `memex-local` configures each mounted repo as `--set pluginCatalog.installByDefault[i]=<Name>/*`.
2. `InstanceAutoRegistrationService.InstallDefaults` treats that lane as a seed: `.Where(c => c.Reconciled || !seeded.Contains(c.Package.Id))`, where `Reconciled = (baseline && PreInstalled) || IsIncluded(c)` — the platform baseline and the environment's feature flags only. The rationale ("the seed must not fight an operator") is right for a deployment seeding itself and wrong for a portal whose purpose is to mirror a working tree.
3. `PluginUpdateWatcher` needs a GitHub `workflow_run` webhook into `Admin/_Build/{owner}.{repo}`. A local install holds no GitHub credential and receives no webhooks.
4. `RegistryUpdateReconciler` covers the registry-consumer case only — its own doc: *"With no registry configured this service resolves no registry source and does nothing at all."*
5. The documented human fallback — *"a human opening the catalog page… offers the same Update"* — is unavailable out of the box: `memex-local` creates no `PluginCatalog` node, so a fresh install has **no catalog page at all** (`nodeType:PluginCatalog` → 0 nodes). Creating one by hand (`sourceRepoPath: /plugin-repos/<Name>`) restores it, and the Update then works from the mount with no credential.

### Impact

A developer testing plugin or course content on their local portal is served an arbitrary past install with no signal that it is stale — precisely the audience self-registry mode exists for. It makes "test it on your local mesh before you merge" unreliable in a way that presents as the change not working.

### Possible fixes (not prescriptive)

1. Put the mounted sources on the **reconciled** lane — as a feature flag in `deploy/homebrew/share/values.local.self-registry.yaml`, or directly in `memex-local`. Cost is nil when nothing changed: `InstallOrUpdateCore` short-circuits on the `manifest.lock` hash.
2. Better: a third lane in `InstanceAutoRegistrationService` for local `node-repo` sources. A mounted working tree is standing operator intent, not a one-time seed.
3. Have `memex-local` seed one `PluginCatalog` node per mounted repo, so the manual Update path exists on a fresh install whichever lane is chosen.

### Related

In self-registry mode the `Store` root's `packageSources` still name GitHub repos, so `StoreManifestSource` polls `github.com` unauthenticated every five minutes and logs `Octokit.NotFoundException: Not Found` per source; `Store/Package` stays empty. That component is `MeshWeaver.Plugins/Store/Catalog/Source/StoreManifestSource.cs`, and **Memex#165** already reports the credentialed end of the same path plus the `Token()` `.Catch("")` downgrade that hides which case you are in. Noted here as context only — it is not part of this issue's scope.

https://claude.ai/code/session_01HKEkt3bwipBcsvuGXQWV7F

sierragolflima:
Fixes #3359 (the first of its three options; the catalog-node option is blocked by a separate defect, see the issue comment).

## What was wrong

`helm_deploy` put every discovered checkout on `pluginCatalog.installByDefault`, which `InstanceAutoRegistrationService` treats as a **seed**: what it delivers is recorded on `Plugins/_DefaultInstallLedger` and never re-asserted, because *"the seed must not fight an operator"*. Right for a deployment seeding itself; wrong for a portal whose entire purpose is to mirror a working tree.

Nothing else covers this shape. `PluginUpdateWatcher` needs a GitHub `workflow_run` webhook a local install never receives; `RegistryUpdateReconciler` says in its own doc that it does nothing when no registry is configured. The boot is the only occasion there is, and the seed lane skips it — so a self-registry portal installs its checkouts once and then silently stops tracking them, while every boot reports `0 written, 0 unchanged` and the log reads healthy.

## The change

One loop, two lanes instead of one: the same source patterns also go on the environment's own composition lane (a feature flag, #1735), which is exempt from the seed ledger and re-asserts on every boot.

```
--set pluginCatalog.installByDefault[i]=<Name>/*     (unchanged — establishes a FRESH install)
--set features.mountedCheckouts.packages[i]=<Name>/* (new — keeps it honest afterwards)
--set features.mountedCheckouts.enabled=true         (only when at least one repo is mounted)
```

It costs nothing when nothing changed: `CatalogLayoutAreas.InstallOrUpdateCore` short-circuits on the package's `manifest.lock` hash and no node travels. Registry mode is untouched — it mounts no checkout, the name list is empty, no flag is emitted. An operator who wants out sets `features.mountedCheckouts.enabled: false` in their overlay, which the flag lane already reads as an exclusion that outranks every other signal.

## Measured, on a real install

Same portal, before and after the redeploy:

- **before** — `[DefaultInstall] 41 package(s), in dependency order — AI, Edu, Publish, AgenticEngineering, …`; no `Store`, no course.
- **after** — `[DefaultInstall] 76 package(s), in dependency order — Store, AI, Edu, AdvancedBusinessRules, AgenticBusiness, Publish, Video, AgenticEngineering, AgenticOffice, AgenticPrimer, AgenticPrimerDe, …`
- and the drift it had been hiding, on the very first boot under the fix: **`[DefaultInstall] Store → Store: 167 written, 1 unchanged`**.

The ConfigMap renders as intended (`Features__Flags__mountedCheckouts__Enabled: "true"` plus one `__Packages__N` per mounted source), the commercial-skip warning is unchanged (same 16 packages, same reasons), and `bash -n` passes.

Held as a **draft**: this changes what every developer's `memex-local update` does on the next run, and it deserves a maintainer's eye before it lands.

https://claude.ai/code/session_01HKEkt3bwipBcsvuGXQWV7F

### DIFF
--- a/deploy/homebrew/bin/memex-local
+++ b/deploy/homebrew/bin/memex-local
@@ -79,6 +79,9 @@ readonly MEMEX_LOCAL_VERSION="0.2.0"
 
 readonly LOCAL_HOME="${MEMEX_LOCAL_HOME:-$HOME/.memex-local}"
 readonly OVERLAY="$LOCAL_HOME/values.local.yaml"
+# The feature flag helm_deploy declares the mounted checkouts under, so they land on the RECONCILED
+# lane (re-asserted on every boot) and not only on the seed-once one — see the block in helm_deploy.
+readonly MOUNTED_FLAG="mountedCheckouts"
 # ── REGISTRY MODE — this install CONSUMES a remote plugin registry ──────────────────────────────
 # Written by `memex-local registry <url> [--key mwr_…] [--id ID]`, read by helm_deploy as one more
 # values file (passed AFTER the user's overlay, so it wins). Its presence IS the mode switch: with it
@@ -718,11 +721,29 @@ helm_deploy() {
   # --from-acr / --skip-image run, as the info line above says), plugin_args/install_names whenever
   # no plugin checkout was found. So the deploy died on "unbound variable" for exactly the paths the
   # script itself documents as normal. This tool is macOS-only; the guard is not optional here.
+  # 🚨 installByDefault is a SEED lane, and a seed says nothing about a portal that is already
+  # populated: InstanceAutoRegistrationService records what it delivered on
+  # Plugins/_DefaultInstallLedger and never re-asserts it ("the seed must not fight an operator").
+  # That is right for a deployment seeding itself and wrong here, where the whole point is to MIRROR
+  # a working tree — the mounted repos were installed once and then stopped tracking their
+  # checkouts, silently: measured on 2026-09-05, a portal still served a course as of 2026-08-31
+  # while the mount had moved on for a week, with every boot reporting "0 written, 0 unchanged"
+  # (MeshWeaver#3359). Neither auto-update mechanism covers this shape either — PluginUpdateWatcher
+  # needs GitHub webhooks a local install never receives, and RegistryUpdateReconciler does nothing
+  # without a registry — so the boot is the only occasion there is.
+  #
+  # The same patterns therefore ALSO go on the environment's own composition lane (a feature flag,
+  # MeshWeaver#1735), which is exempt from the seed ledger and re-asserts on every boot. It costs
+  # nothing when nothing changed: CatalogLayoutAreas.InstallOrUpdateCore short-circuits on the
+  # package's manifest.lock hash and no node travels. Both lanes are set, not one: the seed still
+  # establishes a FRESH install, and the flag keeps it honest afterwards.
   local n i=0
   for n in ${install_names[@]+"${install_names[@]}"}; do
     plugin_args+=(--set "pluginCatalog.installByDefault[$i]=$n/*")
+    plugin_args+=(--set "features.$MOUNTED_FLAG.packages[$i]=$n/*")
     i=$((i + 1))
   done
+  [ "$i" -gt 0 ] && plugin_args+=(--set "features.$MOUNTED_FLAG.enabled=true")
 
   # Layering: chart defaults < TRACKED local defaults (ship with the brew package, so existing
   # installs pick up new behaviour — e.g. the adopt-only startup — on upgrade)


sierragolflima:
Closing: this is a core change to a tool every developer runs, so it belongs with a maintainer rather than landing from here. The diff, the rationale and the local measurements now live on #3359 (https://github.com/Systemorph/MeshWeaver/issues/3359#issuecomment-5553533893) so nothing depends on this branch.

PR #3369, opened 2026-09-05
docs: a head-truncated view of ordered output is biased, not merely partial

rbuergi:
Strengthens the truncation row in `ControlsThatCannotFail` rather than adding an instance — it is the
same shape as the `dotnet-gcdump` entry it sits beside, one layer down: in the **shell**, not the
profiler.

## The new part is *why* it fools people

Not "someone forgot the output was truncated". A sweep for ambient-culture uses in a satellite ran as:

```
git grep 'CultureInfo\.Current' -- 'src/**/*.cs' | head -10
```

…and reported **zero** production offenders. The sweep had **19** lines. The real offender was number
**18**.

Comments and test files **sort early** — they cluster in doc-heavy and `*.Test` paths — so the first
ten lines were entirely commentary and tests. **The truncated view did not look partial. It looked
clean.**

Same person, same day, four hours earlier, different costume:

```
… | tee list.txt | head -30
```

`head` closed the pipe, `SIGPIPE` truncated what `tee` wrote, and a satellite-only package sweep
reported **31** where the truth was **49** — the tell being that the one package which had just taken
CD down was missing from its own sweep.

## So the rule is sharper than "beware truncation"

`grep`, `git grep`, `ls` and `find` all emit in sorted or otherwise structured order. That makes a
head-truncated view a **systematically unrepresentative** sample rather than a random one — and the
bias points at whatever sorts first, which is usually the thing you were *not* looking for.

> **Count first (`| wc -l`), filter comments and tests INSIDE the pipeline, and treat `head` as a way
> to eyeball SHAPE — never as a way to establish ABSENCE.**

## Provenance

The measurements are the core-CD session's, from its own retraction of a sweep it had handed me as
fact. The offender that sweep missed is real — `ConfiguredExcelImporter.GetCellDouble` falling back to
`CultureInfo.CurrentCulture`, filed as MeshWeaver.Plugins#1375, and guarded as the single known-debt
entry in MeshWeaver.Plugins#1376.

## Verification

`DocumentationLinkIntegrityTest` passes.

### DIFF
--- a/src/MeshWeaver.Documentation/Data/Architecture/ControlsThatCannotFail.md
+++ b/src/MeshWeaver.Documentation/Data/Architecture/ControlsThatCannotFail.md
@@ -42,7 +42,7 @@ instances. The family is larger, and it is worth recognising by shape.
 | **Blindness rendered as health** | a watcher whose read failure and whose "no matches" look identical | "I cannot see" is reported as "nothing is wrong" |
 | **One word covering three states** | `completed/success` over checked-and-passed, never-ran, and ran-and-did-nothing | the verdict is not the outcome |
 | **A tool that collapses several failures into one exit code** | `git cat-file -e "$sha:$path" \|\| echo ABSENT` | absent-path, bad-sha, missing-object and bad-quoting all read the same |
-| **A measurement that truncates and answers anyway** | a profiler capping at N objects and reporting the partial total | the wrong number is *plausible*, so it closes the investigation |
+| **A measurement that truncates and answers anyway** | a profiler capping at N objects; `\| head -10` on a sorted sweep | the wrong number is *plausible* — and on ordered output the survivors are biased, so a partial view reads as a clean one |
 | **A writer whose failure mode is a SUCCESS line** | a generator that skips its work and prints a summary anyway | nothing distinguishes "wrote it" from "declined to" |
 
 ## Eight measured instances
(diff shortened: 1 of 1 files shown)


rbuergi:
**Blocked by #3370, not by this diff.** `Consolidate test results` is red because `Run tests (shard 3)`
failed on `BakeEquivalenceTest.MeshDrivenAndCompilerDrivenBakes_ProduceTheSameArtifacts`.

This pull request changes **one markdown file**. It cannot reach a bake.

Attributed rather than assumed:

- **Not #890** — the log carries zero `PROCESS CANNOT EMIT` / `BELOW-ROSLYN` /
  `GetConsolidatedTypeParameters` markers, and that detector exists precisely to stop this
  misattribution.
- **Not the tree** — `main` (`34c3eecec`, the exact base this branch was cut from) ran the suite clean,
  as did #3368's head.
- **Not known** — absent from `.github/known-flakes.json`.

Filed as #3370 with the full evidence, including that the assertion truncates its own reason at 50
characters, which is why it is filed without a root cause. #3371 fixes that reporting.

**I have not re-run it**, and there is no legitimate re-measure available yet: `main` is still
`34c3eecec`, so merging it in would be a no-op that retriggers CI — a re-run wearing a merge's
clothes. This unblocks naturally when `main` moves (#3371 is the obvious candidate, since it touches
that very test), at which point I will merge main here and let CI re-measure against a changed tree.

🤖 Filed from Claude Code.

Issue #3370, opened 2026-09-05
Flake: BakeEquivalenceTest — the mesh-driven bake fatals on Widget after its NodeType passed, and the assertion truncates the reason at 50 chars


rbuergi:
## Observed

`MeshWeaver.PluginTester.Test.BakeEquivalenceTest.MeshDrivenAndCompilerDrivenBakes_ProduceTheSameArtifacts`
failed `Run tests (shard 3)` on
[run 33995947702](https://github.com/Systemorph/MeshWeaver/actions/runs/33995947702) (job
`101386713855`), on PR #3369 — **a documentation-only change of one markdown file**, which cannot
reach the bake.

The failing assertion is `Assert.Null(meshReport.FatalError)` (`BakeEquivalenceTest.cs:225`): the
**mesh-driven** bake reported a fatal error while the compiler-driven one completed.

```
── mesh-driven bake ──
Discovered 2 package(s), install order: Lib → Widget
── Lib: installed (3 written, 0 unchanged); checking 1 NodeType(s)…
   ok  Lib/Shared   [compile:Passed render:Passed tests:Skipped]
── Widget: installed (9 written, 0 unchanged); checking 1 NodeType(s)…
   ok  Widget/Thing [compile:Passed render:Passed tests:Skipped]
bake: Lib → 1 assembly(ies) + 3 node file(s) → /tmp/mw-bake-mesh-…/Lib.zip
          ← no Widget.zip line; the exception lands here

Assert.Null() Failure: Value is not null
Actual: "InvalidOperationException: bake: 'Widget/Thing' cl"···
```

Note the ordering: `Widget/Thing` **compiled and rendered green**, and `Lib` baked to a zip. The
failure is on the *Widget* package's own bake step, after its NodeType passed.

The compiler-driven arm finished normally — `12 node(s) from 2 package(s)`, one skipped file
(`Widget/Unparseable.json`, which is a deliberate fixture).

## Attribution

- **Not #890.** The log carries **zero** occurrences of `PROCESS CANNOT EMIT`, `BELOW-ROSLYN` or
  `GetConsolidatedTypeParameters`, so the detector that exists for exactly this misattribution says
  it is not that.
- **Not the diff.** #3369 changes one `.md` file. `BakeEquivalenceTest` builds a synthetic
  `Lib`/`Widget` fixture repo, not the documentation tree.
- **Not the tree.** `main` (`34c3eecec`, the same base this branch was cut from) and #3368's head
[...]

PR #3371, opened 2026-09-05
test(#3370): the bake assertions print WHY, instead of the first 50 characters

rbuergi:
Part of #3370. **Does not fix the flake — makes it diagnosable**, which is currently the blocker to
fixing it.

## The problem

`BakeEquivalenceTest` asserted `Assert.Null(report.FatalError)` on both arms. xUnit renders a failing
`Assert.Null` by truncating the value at **50 characters**, so when the mesh-driven bake fatalled on
shard 3 the entire diagnosis available to the reader was:

```
Assert.Null() Failure: Value is not null
Expected: null
Actual:   "InvalidOperationException: bake: 'Widget/Thing' cl"···
```

Measured: that string is exactly **50** characters. The reason is cut mid-word — `cl…` — and appears
**nowhere else** in the log: not in stdout, not in the trx. The reader gets the failure without the
diagnosis, and a second occurrence leaves them exactly as far along as the first.

#3370 *is* that occurrence, and it is why that issue is filed without a root cause.

## The change

```csharp
Assert.True(report.FatalError is null, report.FatalError);
```

Prints the whole message on failure, nothing on success.

**Applied to both arms.** The compiler-driven assertion has the identical exposure and simply has not
fired yet — which is the worse of the two states to leave alone, since it will fire first for someone
who has no reason to suspect the assertion rather than the bake.

## Why this is worth a PR of its own

It is the same shape the repo has been naming all week: a control that detects correctly and then
hands the reader something they cannot act on. The bake test *did* its job — it caught a real
divergence between the two bakes. It then reported it in a form that cannot be investigated.

## Verification

Release + `-warnaserror`, `0 Error(s)`. `BakeEquivalenceTest` **2/2 pass**, so the green path is
unchanged.

### DIFF
--- a/test/MeshWeaver.PluginTester.Test/BakeEquivalenceTest.cs
+++ b/test/MeshWeaver.PluginTester.Test/BakeEquivalenceTest.cs
@@ -207,7 +207,13 @@ public class BakeEquivalenceTest(ITestOutputHelper output)
             });
             output.WriteLine("── compiler-driven bake ──");
             output.WriteLine(treeLog.ToString());
-            Assert.Null(treeReport.FatalError);
+            // 🚨 Assert.True with the message, NOT Assert.Null — xUnit renders a failing
+            // Assert.Null by truncating the value at 50 characters, so the ONE string that says why
+            // the bake failed arrives cut mid-word ("…bake: 'Widget/Thing' cl"···) and appears
+            // nowhere else in the log (#3370). The reader gets the failure without the diagnosis,
+            // and the next occurrence costs them the same dead end. This prints the whole message on
+            // failure and nothing on success.
+            Assert.True(treeReport.FatalError is null, treeReport.FatalError);
             Assert.All(treeReport.Types, t => Assert.Null(t.Error));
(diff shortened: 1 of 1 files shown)


rbuergi:
**Superseded by #3372 — closing this one.** We built the same fix four minutes apart (this at 22:36Z,
#3372 at 22:40Z), on the same file, for the same reason.

#3372's diff is a **strict superset** of this one: besides both `FatalError` assertions it also
converts

```csharp
Assert.All(treeReport.Types, t => Assert.Null(t.Error));
```

into a form that prints `t.NodePath` alongside the error — so a per-type failure names *which type*,
which this PR leaves truncated. That is worth more than the second explanatory comment this PR
carries, and there is no reason to make a reviewer diff two branches to find that out.

The flake itself stays open as #3370; neither PR fixes it, both make the next sighting readable.

Worktree `/private/tmp/MW-3370` will be removed. No objection to #3372 landing — one note on its
issue reference over there.

🤖 Filed from Claude Code.

PR #3372, opened 2026-09-05
fix(#3333): the bake failure says what it claims — Assert.Null truncates it

rbuergi:
`BakeEquivalenceTest` has failed **three** times (#3332, #3366, #3369) and every sighting reads identically:

```
Assert.Null() Failure: Value is not null
Actual:   "InvalidOperationException: bake: 'Widget/Thing' cl"···
```

xUnit truncates a failed `Assert.Null` at ~50 characters. This bake's fatal errors are long **by design** — they name the type, the version claimed, and what the store actually holds:

> `bake: '{typePath}' claims a usable build at v{version} but the run's assembly store has NO bytes for it`

**The version is the one fact that decides whether the stamp is wrong or the store is**, and it has been cut off every time. Three investigations reasoned around it; the most recent concluded — wrongly — that `8614af280` had fixed the cause, because a green re-run was the only other evidence available. #3369 then failed on a tree that *contains* that fix.

`Assert.True(x is null, x)` prints it in full. Same for the tree-side report and the per-type errors, which also name the node path.

**This does not fix the flake.** It makes the next sighting diagnosable instead of a fourth round of inference — which is what is actually blocking progress on #3333.

Verified: `BakeEquivalenceTest` 2/2, Release `-warnaserror` clean.

Refs #3333 · `Pairs-with: none` — one test file.

### DIFF
--- a/test/MeshWeaver.PluginTester.Test/BakeEquivalenceTest.cs
+++ b/test/MeshWeaver.PluginTester.Test/BakeEquivalenceTest.cs
@@ -207,8 +207,8 @@ public class BakeEquivalenceTest(ITestOutputHelper output)
             });
             output.WriteLine("── compiler-driven bake ──");
             output.WriteLine(treeLog.ToString());
-            Assert.Null(treeReport.FatalError);
-            Assert.All(treeReport.Types, t => Assert.Null(t.Error));
+            Assert.True(treeReport.FatalError is null, treeReport.FatalError);
+            Assert.All(treeReport.Types, t => Assert.True(t.Error is null, $"{t.NodePath}: {t.Error}"));
 
             // ── the MESH-driven bake: exactly what CI runs today ──
             var meshReport = await PluginGateRunner.Run(new GateOptions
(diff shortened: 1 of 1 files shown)


rbuergi:
Concurrent duplicate on my side — I opened #3371 four minutes before this with the same two
assertions. **Yours is a strict superset** (it also gives `Assert.All` the `NodePath`), so I closed
mine; this one should land.

**One correction worth making before it does: `#3333` is closed.** It was closed 2026-09-05T08:55:56Z,
and its title is *"Two teardown/disposal races red one Build-and-Test run…"* — the disposal races,
not this test. The three sightings were discussed *in* that thread, but the thread is finished, so
both `Refs #3333` in the body and the `#3333` inside the new code comment send the next reader to a
closed issue whose title does not mention `BakeEquivalenceTest`. That is the same defect this PR is
fixing, one level up: a pointer that does not lead to the evidence.

The live home is **#3370** (open, filed 22:34Z): it carries the #3369 sighting with the truncated
string verbatim, the attribution work that rules out #890 (zero `PROCESS CANNOT EMIT` markers), and
the observation that `main` at the identical base ran clean. Suggest re-pointing the body to
`Refs #3370` and changing the in-comment `#3333` to `#3370`.

Two further facts from #3370 that are worth carrying into #3333's successor, since this PR's body is
currently the best written account of the flake:

- The failure is **not** #890. The run's log has no `PROCESS CANNOT EMIT` / `BELOW-ROSLYN` /
  `GetConsolidatedTypeParameters` markers, which is exactly what that detector exists to discriminate.
- Your body's point that a green re-run was "the only other evidence available" is the load-bearing
  one. `8614af280` is in `/Users/roland/code/MW-3333`; #3369 failed on a tree containing it, which
  falsifies that conclusion rather than merely failing to confirm it.

Happy to leave the whole thing with you — I have removed my branch.

🤖 Filed from Claude Code.

PR #3410, opened 2026-09-06
fix(#3404): the pre-warmer reads the durable GO when the subscription door is unreachable

rbuergi:
## The decisive fact

`Admin/Build` was **already `Ready`**, and a GO for the pods' own framework fingerprint
(`s2f227642…`) had been written on it at **2026-09-06T11:28:56Z**. The two refusals followed at
**11:44:32Z** and **11:49:37Z** — about **sixteen minutes later**.

The pods refused a build that had already been approved, because they could not establish a
subscription to read a verdict that was already durable. Their own hubs were healthy and idle
throughout the wait (`RunLevel=Started`, empty queue, `deliveryActionCompleted=True`).

## The two-door pattern, and where it was matched from

`Doc/Architecture/BuildCoordination` → *"The follower has two doors, and it closes them behind
it"*. Since #1440 `BuildProtocolDriver.FollowGo` ends on either of two **real, level-triggered**
events:

1. **The GO becomes visible** — `ReadBuildGo` reads it off the **durable** build root (the same
   witness `BuildNodeType.ArbitrateDurably` decides on) merged with `ObserveBuildGo` on this
   cluster's mirror, whichever answers first. The durable read is what lets a peer cluster's GO
   mean anything at all here, because a PG `NOTIFY` reaches only a live `LISTEN` session and is
   never replayed.
2. **The arbiter hands us the claim** — and on that grant the follower **re-reads the durable
   witness** to decide whether to stand down or bake.

The follower's doors only ever opened once a process was already *inside*. `BuildProtocolDriver.Run`
opens with the claim handshake, and a claim handshake is a `SubscribeRequest` to `Admin/Build` — so
the pre-warmer's **entry** had exactly one door, and exhausting `CoordinationAttempts` on it was a
terminal refusal.

**The fix:** `Run`'s whole subscription-borne composition is now treated as ONE door, and
`WhenTheSubscriptionDoorIsShut` asks the **same `ReadBuildGo`** when it cannot be opened at all.
Same read, same witness, same fault taxonomy — at the one point in the protocol that never had it.
`ProbeAfterGo`'s share probe is split out as `ProbeTheShare` so both paths share the probe.

🚨 **The GO is matched on THIS process's fingerprint by exact key.** The root's `Ready` map is keyed
by framework version precisely so an old-image pod stays ready while a new image is unproven; a GO
written for any other image reads as *no GO* and the door stays shut. Granting on a foreign GO would
certify a build this process is not running — the one wrong answer this door could give, and
strictly worse than the refusal it replaces.

## What I deliberately did NOT change

- **The transport fault.** Not diagnosed, not repaired, not papered over. `RetryUnreachableCoordination`
  still logs the unreachability at **its current severity with its current diagnostic detail** on
  both branches, and the grant adds a `Warning` naming what could not be reached and saying the
  fault is *not* cleared. It remains the only signal that the pod↔`Admin/Build`-hub path is broken.
  **The durable door changes the readiness VERDICT, never the visibility of the fault.**
- **Fail-closed.** Neither door answering re-throws the **original**
  `BuildCoordinationUnreachableException` — including the two cases `ReadBuildGo` deliberately folds
  into `null` (no durable store; the read itself failed). The sweep still faults, readiness is still
  refused, the rollout still holds the previous image, and `DescribesUnreachableCoordination` still
  separates *no verdict* from *a bad verdict* in the health payload.
- **The bounds.** `CoordinationAttempts` (3) and the hub's 60 s request budget are untouched. No
  retry loop, no watchdog, no poller, no resubscribe timer, no `catch {}`.
- **The follower's stand-down.** See below.

## This is the FAIL-SAFE; the CAUSES are elsewhere — and there are now five

🚨 **This PR does not fix the silence.** It changes the readiness verdict when the silence happens.

The issue lists three candidates for the unanswered `SubscribeRequest`: it never reached the target
(routing), the target received it and is wedged (the #2896 class), or the reply was lost. There is a
**fourth** and a **fifth**, both measured since the issue was filed.

### Fourth — deferred-queue ordering (#3408)

`MessageService.OpenGate` drained the deferred queue by **appending** it to the main queue, so a message that arrived before the gate opened but had not
yet been turned sat *ahead* of the deferred turns appended behind it — and the parked message ran
last. Appending is always the wrong end, not merely unlucky: deferral happens at TURN time, not
arrival, and the loop is strictly FIFO, so everything deferred is by construction older than
everything still waiting.

A `SubscribeRequest` to `Admin/Build` is very plausibly exactly that message — the one that triggers
the target hub's activation, and therefore the one that loses its place. Load-sensitive by
construction, green in isolation, and capable of being processed *after* the requester's 60 s budget
has expired. That fits this incident's evidence better than a wedge does: the requesting hubs were
measured healthy and idle, and the target was never observed dead. The concurrent Orleans
`ConnectionFailedException` incident (`Admin/_LogIncident/522013736000428b`, same cluster and
overlapping window) remains a candidate for the routing branch.

### Fifth — a root that STOPS EMITTING with a holder still set (`MeshWeaver.Plugins#1193`)

A controlled load experiment on `BuildCoordinationTest.Follower_StandsDown_SoTheNextBuildCanStart` —
18 cores, two arms of ten runs, **same binary and same sha**, only background load differing:

| Arm | Load | Result |
|---|---|---|
| 12 burners | ~31–39 | **10 / 10 passed** |
| 64 burners | 49–91+ | **9 passed, 1 FAILED** at 32.9 s, with the CI signature |

Instrumenting the `settled` wait recorded **exactly one** root emission in 15 seconds, then nothing.

🚨 **The control is what makes the reading honest.** The PASSING state was captured at the same point
in the same test — by temporarily making the predicate unsatisfiable, to prove the diagnostic
actually prints, since a diagnostic never seen to run reads exactly like one that works. The two
states differ in **one field**:

```
FAIL:  ClaimedBy=<machine-scoped id>   Status=Planning  RequestedClaims=[]  Ready=[fp-stand-down]
PASS:  ClaimedBy=<null>                Status=Planning  RequestedClaims=[]  Ready=[fp-stand-down]
```

So `Status=Planning` and the empty `RequestedClaims` are **not symptoms** — the healthy run carries
both, and the build completed in both. Any account resting on those two fields rests equally on the
passing run. What is measured is exactly: **`ClaimedBy` stays set, one root emission in 15 s, nothing
after.** A reader waiting on such a root waits forever.

**The mechanism is open**, and this PR does not assert one. Whether the holder cannot act or was
never told to is the question still live on `MeshWeaver.Plugins#1193`. The permanent diagnostic
landed as `MeshWeaver.Plugins#1401` (diagnostic only, no production code); the repro recipe is on
#1193 — reproduce from those rather than re-deriving.

🚨 It reproduced **with #3408 already in the core**, so #3408 does not remove it and #3131 did not
close it. Fourth and fifth are **independent holes; neither subsumes the other.**

### Why this matters for this PR

This **shape** is the one the durable door most clearly answers — and the claim is about the shape,
not about any mechanism. A root that has stopped emitting is neither dead nor slow, so nothing about
the transport is going to recover it, while the durable GO is already published and already readable.
**This change is correct whichever of the five it is, it is the only one that does not depend on
diagnosing the transport first, and in this shape it is the only fix in flight that helps at all.**

None of which makes it a cure. Each cause is a separate change, owned separately, and this PR must
not be read as having addressed any of them.

## Interaction with the live follower defect (Plugins#1193)

The subscription door's underlying protocol has a **live, load-sensitive defect** under separate
investigation — `BuildCoordinationTest.Follower_StandsDown_SoTheNextBuildCanStart`, whose controlled
measurement is the fifth cause above. **This PR does not touch it and must not be read as having
addressed it.** Two other sessions are on it from the opposite end; changing this protocol from both
ends at once is how one fix hides the other.

The durable door is deliberately **independent of the stand-down path**: it never calls
`WithdrawBuildClaim`. That is not an omission but a consequence — `RequestBuildClaim` writes its
registration through `GetMeshNodeStream(path).Update(current => …)`, and an `Update` whose stream
never delivered current state never computed a patch, so **no candidate entry can exist to hand
back**. Calling `WithdrawBuildClaim` here would also post a second write into the same unreachable
hub and re-fault the stream on the way out. The two paths share the post-GO share probe; they
deliberately do not share the stand-down.

## Falsification — real numbers

The test is `test/MeshWeaver.Hosting.Test/PreWarmerReadsTheDurableGoTest.cs`, on a **real**
`MonolithMeshTestBase` mesh: the witness is written to and read back through the mesh's own
`IStorageAdapter` and its own `JsonSerializerOptions`, and the share is probed through the mesh's
own `IAssemblyStore`. Only the fault is injected — the verbatim production
`BuildCoordinationUnreachableException` over the incident's own inner `TimeoutException` — exactly
as `BuildCoordinationRetryTest` injects one into `RetryUnreachableCoordination`. No mocks, no
`Task.Delay`, no `.ToTask()`, no `SemaphoreSlim`; the recording logger is lock-free
`ImmutableInterlocked`.

The subscription door is shut in **every** case, so a pass can only come from the durable read.

| | With the fix | Second door removed, tests unchanged |
|---|---|---|
| `ADurableGoForThisFingerprint_GrantsReadiness` | **pass** | **FAIL** — threw `BuildCoordinationUnreachableException` (the production defect, reproduced) |
| `AnUnreachableNodeIsStillReportedLoudly_EvenWhenTheDurableGoGrants` | **pass** | **FAIL** — same throw, nothing logged |
| `NoDurableGo_StillRefusesReadiness` | pass | pass *(fail-closed control — passing both ways is its job)* |
| `AGoForAnotherFingerprint_StillRefusesReadiness` | pass | pass *(fingerprint control — same)* |
| **Totals** | **Failed: 0, Passed: 4** | **Failed: 2, Passed: 2** |

The falsification reverted the behaviour under test (`WhenTheSubscriptionDoorIsShut` → pass the
subscription door straight through, which *is* the pre-fix composition) rather than the file, so the
tests compiled unchanged and produced counts. The two controls pass in both arms **by design**: they
are what proves the fix bought readiness without weakening the refusal — an implementation that
simply stopped failing closed passes the first case and fails those two.

Restored and re-run: **Failed: 0, Passed: 11** (the 4 new cases plus the 7 existing
`BuildCoordinationRetryTest` cases, which pin the retry and fail-closed contract).

`DocumentationLinkIntegrityTest` and the ratchet guards: **Failed: 0, Passed: 15**.

Builds, `-c Release -warnaserror`, one project per invocation: `MeshWeaver.Hosting`,
`MeshWeaver.Hosting.Monolith`, `MeshWeaver.Documentation`, `MeshWeaver.Hosting.Test`,
`MeshWeaver.Documentation.Test` — all `0 Warning(s), 0 Error(s)`.

## Docs

`Doc/Architecture/BuildCoordination` gains *"The PRE-WARMER has the same two doors — and it did not,
until #3404"*, recording the incident, the four load-bearing properties (per-fingerprint,
fail-closed, the fault stays visible, no stand-down) and which test pins each. Plus a What's New
entry (`Category: Fix`).

## Scope — and what remains

Fixed here: a pod no longer refuses a build that is **already approved and durably readable**. That
is the decisive complaint in the issue, and the case both 2026-09-06 refusals were.

Deliberately NOT fixed, and why this is `Refs`, not `Closes`:

1. **The silence itself is still open.** See the section above — there are now five candidate
   causes, #3408 and `MeshWeaver.Plugins#1193` are separate changes owned elsewhere, and neither
   subsumes the other. Until one of the five is closed, the fingerprint can still fire.
2. **A pod that is the FIRST to need a GO that does not exist yet** still refuses readiness under
   the same silence, and still holds the rollout. That is the correct fail-closed behaviour and this
   PR must not change it — but it means incident `4b3ea19559962a76` can recur, with a different
   consequence.
3. **The second protocol question the issue asks is untouched**: whether this subscription should
   follow the `SubscribeWithReEstablish` fault taxonomy instead of exhausting a fixed attempt budget.
   That is a separate decision about the subscription door and is not attempted here.

Refs #3404

Pairs-with: none — the diff removes no public type and no public member (`git diff --cached -U0 |
grep '^-' | grep 'public '` returns nothing); the one new member is `internal`.

### DIFF
--- a/src/MeshWeaver.Documentation/Data/Architecture/BuildCoordination.md
+++ b/src/MeshWeaver.Documentation/Data/Architecture/BuildCoordination.md
@@ -372,9 +372,25 @@ Four properties are load-bearing, and each is pinned by a case in
 - **The transport fault stays fully visible.** Nothing here widens a timeout, retries, polls or
   swallows. `RetryUnreachableCoordination` logs the unreachability at its own severity with its own
   diagnostic detail either way, and the grant logs a second `Warning` naming what could not be
-  reached. That fault is the only signal the pod↔`Admin/Build`-hub path is broken (routing loss, a
-  wedged per-node hub of the #2896 class, or a lost reply); the durable door changes the readiness
-  VERDICT, never the visibility of the fault, and does not diagnose or repair it.
+  reached. That fault is the only signal the pod↔`Admin/Build`-hub path is broken; the durable door
+  changes the readiness VERDICT, never the visibility of the fault, and does not diagnose or repair
+  it.
+
+🚨 **The durable door is a FAIL-SAFE, not a cure — and the cause is still open.** The pod's own
+diagnostic names three candidates for the silence: the request never reached the target (routing),
+the target received it and is wedged (a per-node hub that stops answering — the #2896 class), or the
+target answered and the reply was lost. There is a **fourth**, root-caused separately in #3408:
+`MessageService.OpenGate` drained the deferred queue by **appending** it to the main queue, so a
+message that arrived before the gate opened but had not yet been turned sat *ahead* of the deferred
+turns appended behind it — and the parked message ran last. Appending is always the wrong end:
+deferral happens at TURN time, not arrival, and the loop is strictly FIFO, so everything deferred is
+by construction older than everything still waiting. A `SubscribeRequest` is exactly the message
+that triggers a per-node hub's activation, so it is the one that loses its place — load-sensitive by
+construction, green in isolation, and capable of being processed after the requester's 60 s budget
+has already expired. That fits a requester measured healthy and idle against a target never observed
+dead better than a wedge does. **#3408 may reduce or remove these timeouts independently; the two
+fixes are separate and must not be read as one.** The durable door is correct whichever of the four
+it is, and is the only one of them that does not depend on diagnosing the transport first.
 - **No stand-down on this path.** The follower withdraws its claim before probing because it really
   registered one. This path did not: the registration is written by `RequestBuildClaim` through
   `GetMeshNodeStream(path).Update(current => …)`, and an `Update` whose stream never delivered
--- a/src/MeshWeaver.Documentation/Data/Architecture/BuildCoordination.md
+++ b/src/MeshWeaver.Documentation/Data/Architecture/BuildCoordination.md
@@ -336,6 +336,52 @@ on durable state, exactly like the arbiter (and see #1366 for why the poll clock
 clusters instead of at the arbiter's next pass, but it changes behaviour for every partitioned-PG
 deployment and wants its own justification and measurement. The follower is correct without it.
 
+### The PRE-WARMER has the same two doors — and it did not, until #3404
+
+The follower's two doors only ever opened once a process had already got *inside*. Everything the
+pre-warmer knew about the build arrived through ONE door: `BuildProtocolDriver.Run` opens with a
+claim handshake, and a claim handshake is a `SubscribeRequest` to `Admin/Build`. Exhaust its
+attempts and the driver threw `BuildCoordinationUnreachableException`, the sweep faulted, readiness
+was refused, and the rollout held the previous image.
+
+🚨 **A pod can therefore refuse a build that was already approved.** Measured on `memex-cloud`,
+2026-09-06: two pods of one deployment refused at 11:44:32Z and 11:49:37Z for a fingerprint whose GO
+had been written on an already-`Ready` build root at **11:28:56Z** — sixteen minutes earlier. Their
+own hubs were healthy and idle throughout (`RunLevel=Started`, empty queue); the silence was between
+them and the hub that owns `Admin/Build`. They could not open a subscription to read a verdict that
+was already durable — and the verdict was the only thing they needed.
+
+So `Run`'s subscription-borne composition is now treated as ONE DOOR, and when it cannot be opened
+at all the **durable witness is asked** (`WhenTheSubscriptionDoorIsShut`). Same read as door 1 of
+the follower — `ReadBuildGo` on the durable row, the record the claim arbiter itself decides on —
+at the one point in the protocol that never had it.
+
+Four properties are load-bearing, and each is pinned by a case in
+`PreWarmerReadsTheDurableGoTest`:
+
+- **Per-fingerprint, never "a GO exists".** The `Ready` map is keyed by framework version precisely
+  so an old-image pod stays ready while a new image is unproven. `ReadBuildGo` looks the pod's own
+  fingerprint up by exact key; a GO for any other image reads as no GO and the door stays shut.
+  Granting on a foreign GO would certify a build the process is not running — strictly worse than
+  the refusal it replaces.
+- **Fail-closed is unchanged.** Neither door answering — including the two cases `ReadBuildGo`
+  deliberately folds into `null`, no durable store and a failed read — re-throws the ORIGINAL
+  `BuildCoordinationUnreachableException`, so the sweep still faults, readiness is still refused,
+  and `DescribesUnreachableCoordination` still separates "no verdict" from "a bad verdict" in the
+  health payload.
+- **The transport fault stays fully visible.** Nothing here widens a timeout, retries, polls or
+  swallows. `RetryUnreachableCoordination` logs the unreachability at its own severity with its own
+  diagnostic detail either way, and the grant logs a second `Warning` naming what could not be
+  reached. That fault is the only signal the pod↔`Admin/Build`-hub path is broken (routing loss, a
+  wedged per-node hub of the #2896 class, or a lost reply); the durable door changes the readiness
+  VERDICT, never the visibility of the fault, and does not diagnose or repair it.
+- **No stand-down on this path.** The follower withdraws its claim before probing because it really
+  registered one. This path did not: the registration is written by `RequestBuildClaim` through
+  `GetMeshNodeStream(path).Update(current => …)`, and an `Update` whose stream never delivered
+  current state never computed a patch, so there is no candidate entry to hand back — and calling
+  `WithdrawBuildClaim` would post a second write into the same unreachable hub. The two paths share
+  the post-GO share probe; they deliberately do not share the stand-down.
+
 The probe semantics of `NodeTypeBakeGateState` are preserved unchanged — fail **closed**
 on a measured regression (the rollout stalls, the old image keeps serving), fail **open**
 when nothing is measuring (a configuration mistake must never black-hole a pod), gate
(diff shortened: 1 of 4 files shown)


rbuergi:
## CI red — attribution: not reachable from this diff

[Run 34034828870](https://github.com/Systemorph/MeshWeaver/actions/runs/34034828870) failed on **one** test.

**Failure** — `Run tests (shard 5)`, the only one in the whole shard (`MeshWeaver.Layout.Test  Total: 450, Failed: 1`):

```
MeshWeaver.Layout.Test.EditorTest.TestEditorWithoutResult [FAIL]
  ObservableAssertionException : Expected the observable to emit a value matching the predicate
  within 10s, but it did not. The observable emitted nothing at all.
    test/MeshWeaver.Layout.Test/EditorTest.cs(64,0)
  [13:06:23.021] === TEST CLASS INIT: EditorTest ===
  [13:06:23.021] === TEST START ===
  [13:06:33.070] === TEST FAILED ===
```

It is the **first** test of its class and it consumed exactly the full 10 s budget; the sibling
`TestEditorWithResult` ran immediately afterwards in ~100 ms and passed. That is a first-render /
cold-start timing shape, not a verdict about this change.

**Why this diff cannot reach it.** `MeshWeaver.Layout.Test`'s project closure is `MeshWeaver.Data`,
`MeshWeaver.Fixture`, `MeshWeaver.Layout`, `MeshWeaver.Mesh.Contract`, `MeshWeaver.Messaging.Hub`,
`MeshWeaver.Reactive.Assertions` — plus two link-compiled `PandasExplorer` sources. This PR touches
`src/MeshWeaver.Hosting/BuildProtocolDriver.cs`, `test/MeshWeaver.Hosting.Test/…`, and two markdown
files under `Data/Architecture` and `Data/WhatsNew`:

```
$ git diff --name-only origin/main...HEAD \
    | grep -E "MeshWeaver\.(Layout|Data|Fixture|Mesh\.Contract|Messaging\.Hub|Reactive\.Assertions)/|PandasExplorer"
(no matches)
```

**For this test, this branch is byte-for-byte `main`.** `MeshWeaver.Hosting` is not in its closure at
all, and neither markdown file is among the linked `.cs` sources.

**This PR's own tests passed in CI**, in the shard carrying `MeshWeaver.Hosting.Test`
(`Total: 360, Errors: 0, Failed: 0`):

```
PreWarmerReadsTheDurableGoTest.ADurableGoForThisFingerprint_GrantsReadiness            [PASS]
PreWarmerReadsTheDurableGoTest.NoDurableGo_StillRefusesReadiness                       [PASS]
PreWarmerReadsTheDurableGoTest.AGoForAnotherFingerprint_StillRefusesReadiness          [PASS]
PreWarmerReadsTheDurableGoTest.AnUnreachableNodeIsStillReportedLoudly_…GoGrants        [PASS]
```

Shards 0–4 all green.

**Not reproduced locally** on the identical closure (`DOTNET_PROCESSOR_COUNT=4`, Release, the whole
`EditorTest` class): `Failed: 0, Passed: 4`. Consistent with a load-sensitive cold-start race on the
runner rather than a deterministic break.

**What I did NOT do**, deliberately: no re-run to see whether it was a flake, no re-queue, no
`known-flakes.json` entry. That catalogue requires an evidence-bearing entry with a root-cause issue,
and its own header records that single occurrences did not meet the bar — one occurrence is not
evidence, and an entry without a root cause is a tolerance, not a fix.

**Not merging while this is red.** `EditorTest.TestEditorWithoutResult` needs an owner: it is an
uncatalogued, load-sensitive first-render timeout in `MeshWeaver.Layout`, unrelated to this change.

PR #3416, opened 2026-09-06
fix(#3407): a release re-cut no longer collides with its own first attempt

### LINKED ISSUE #3407: A release re-cut collides with its own first attempt, and the type is left advertising a build no release names
### Why MeshWeaver core — checked, not assumed

Both halves live in core's compiler pipeline and nowhere else: `src/MeshWeaver.Compiler.Pipeline/NodeTypeBuildState.cs` (`TryCreateReleaseNode` — mints the id, calls `meshService.CreateNode`, catches every exception into `null`) and `src/MeshWeaver.Compiler.Pipeline/ReleasePostCondition.cs` (detects the stale pointer and asks for the re-cut). Verified by grepping `ReleasePostCondition` and `failed to create Release node` across MeshWeaver, MeshWeaver.Plugins and MeshWeaver.Education — hits only under `MeshWeaver/src/`. No satellite can reach either: a plugin repo's only involvement is that its NodeTypes are the ones left stale (`Edu/CourseInvite` here, but any live-compiled type). `gh search issues --owner Systemorph` on "release re-cut", `ReleasePostCondition` and "Node already exists release" returns no prior report. Per conserve-work-products the durable home is a page under `src/MeshWeaver.Documentation/Data/Architecture/`; this issue is the pointer.

### How to reproduce

Timing-dependent — the collision needs the retry to land in the same second as the first attempt — so this is how to *provoke* it and how to *recognise* it, not a deterministic script. Observed once in roughly five re-cuts of `Edu/CourseInvite` on 2026-09-06.

1. On a self-registry `memex-local` install (mounted checkouts, MeshWeaver#3385 in place), change a `Source/*.cs` file of a live-compiled NodeType, regenerate its `manifest.lock`, and redeploy (`memex-local update --skip-image`) so the boot re-imports and recompiles it. Repeat a few times.
2. Watch the portal log for `CompileWatcher: failed to create Release node at <type>/Release/<id>` followed by `System.InvalidOperationException: Node already exists: <same id>`, and the `[ReleasePostCondition]` line ending "AND the release could not be re-cut".
3. `get @<type>/Release/*`: healthy re-cuts show as **pairs** a second or two apart with the same suffix; the failing one is an **unpaired** id.
4. Open any instance of the type: the `$Banner` reports the served MVID ≠ the compiled MVID and says a recycle will not clear it. `get_diagnostics` on the type still reports `Ok` — the build succeeded; the pointer did not move.

### What happens

After a compile, `ReleasePostCondition` notices `latestReleasePath` still names an earlier build and re-cuts (#781). The re-cut mints a deterministic id and **creates** it. When that id equals the one its own first attempt already created, the create throws, `TryCreateReleaseNode` catches it into `null`, and the pointer is never advanced:

```
CompileWatcher: failed to create Release node at Edu/CourseInvite/Release/20260906094912-EsO3j1xI
System.InvalidOperationException: Node already exists: Edu/CourseInvite/Release/20260906094912-EsO3j1xI
[ReleasePostCondition] Edu/CourseInvite: … the release could not be re-cut. The node advertises a
build no release names; instances will keep binding '…094902-qc4SQmpO' until a release is created for it.
```

### The evidence is in the release list

Every healthy re-cut appears as a **pair** a second or two apart sharing the random suffix — `…092506-JscoGNC5`/`…092509-JscoGNC5`, `…093540-3Z2ua7y_`/`…093541-3Z2ua7y_`, `…094901-qc4SQmpO`/`…094902-qc4SQmpO`. The failing one, `…094912-EsO3j1xI`, is the only unpaired entry: the retry landed in the same second with the same suffix.

### Consequence

The bytes were published — that release node exists — but the type does not advertise it, so every instance hub keeps executing the previous assembly. The `$Banner` (#2471) fires on each affected page, and its own text says a recycle cannot clear it. Measured on `memex.localhost` 2026-09-06: `Edu/CourseInvite` compiled to build 767 and `AgenticOffice/MyExercises` went on serving the older build, with a fix in the new one the reader could not see. A pod restart happened to clear it; nothing in the pipeline did.

### Suggested fix

Treat "already exists at the id this re-cut minted" as **success**: adopt the node and return its path so `latestReleasePath` advances. This is not a relaxation of a create-only invariant — the colliding node is one this same code created moments earlier, for these same bytes. Separately, a second-resolution id with a shared suffix is worth making collision-proof so the retry stops depending on wall-clock luck.

https://claude.ai/code/session_01HKEkt3bwipBcsvuGXQWV7F

rbuergi:
## The mechanism

`TryCreateReleaseNode` computed the release id — `{yyyyMMddHHmmss}-{8charContentHash}` — from
`DateTime.UtcNow` **inside its own body**, and wrote it with a create-only verb. The hash is a
function of the build (the durable `Collection/ContentPath`), so two attempts at one settle always
agreed on the suffix and differed only in the timestamp.

That made `ReleasePostCondition`'s re-cut — a RETRY of the same operation, for the same bytes —
address a **different node** than the attempt it was retrying. The wall clock then picked which of
two wrong outcomes you got:

| The retry lands in | The minted id | The create-only write | Result |
|---|---|---|---|
| a **later** second | a NEW id | succeeds | **two release nodes for one build** |
| the **same** second | the id the first attempt already created | refused, `Node already exists` | swallowed to `null`, `LatestReleasePath` never advanced — **the type advertises a build no release names** |

Both halves measured 2026-09-06. The duplicate, on `memex` (read back off the live mesh):
`Hosting/InstanceAction/Release/20260906113905-cgZ9cItJ` and `…113915-cgZ9cItJ` —
identical `assemblyStoreVersion: 769`, identical `assemblyContentPath`, identical source snapshot,
`createdBy: system-security` on both, `createdAt` **10.005 s** apart. That is exactly the
`.Timeout(TimeSpan.FromSeconds(10))` on the first attempt's *observation* expiring while its write
had already landed — the node's `createdDate` is 49 ms after its own mint. The collision is the one
in the issue: `Edu/CourseInvite` build 767, `…094912-EsO3j1xI` minted twice in one second,
`InvalidOperationException: Node already exists`, `AgenticOffice/MyExercises` still serving the
older build until a pod restart.

So the trigger — `newReleasePath == null` — is a statement about **observation**, not about the
write. It is a normal recurring state (a bounded observation gives up; a response is not
delivered), and the retry that answers it has to be idempotent by construction rather than lucky.

## What changed

One identity per settle, plus an idempotent write.

- **`NodeTypeBuildState.ReleaseIdentity` + `MintReleaseIdentity(result)`** — `Version`, `Hash` and
  `CreatedAt` minted **together, once**, by the settle. `CreatedAt` rides along so a re-attempt
  writes byte-identical content.
- **`TryCreateReleaseNode` takes that identity** instead of deriving one, so every attempt addresses
  the same node.
- **The write is `IMeshService.CreateOrUpdateNode`** — the framework's own verb for *"idempotent
  writes that may run concurrently or re-run … NOT a client-side
  `CreateNode().Catch(already-exists → UpdateNode())`"*. The owner decides create-vs-update by
  existence and serialises both branches, so the retry is race-free even against its own first
  attempt still in flight, and its no-op guard means an adopt of identical content never reaches the
  store. Attribution is unchanged — the upsert's create branch pins the caller onto the inner
  `CreateNodeRequest.CreatedBy`.
- **`ReleasePostCondition.Restore` re-cuts with that same identity.**
- `NodeTypeBatchBake` mints its own identity at its call site (the parameter is required, not
  defaulted — a caller cannot forget it and silently re-open this).

Both the duplicate and the collision are gone by construction: there is no second id for anything
to collide with, and no second node for a duplicate to be.

## What I deliberately did NOT do

- **No "already exists ⇒ success" catch.** That is the issue's own suggestion and it is
  exception-as-control-flow over a string match; it would also leave the duplicate half standing.
  The upsert is the framework's answer to the same problem.
- **No existence check before the write.** A point read of a node that may not exist faults the read
  stream and opens the storm breaker on that path — which fast-fails writes too.
- **No uniquifying suffix, retry count, or repair sweep.**
- **I did not remove the client-side `.Timeout(10s)` on the release write.** It *is* why the
  observation gets lost, and a client-side ceiling on a mesh write is the shape `MeshService`
  documents as wrong (#1270: *"a client-side ceiling makes the CALLER abandon a write that is still
  running"*). It survives here for a stated reason — the terminal `Status` write runs in this
  observable's `OnNext`, so a hung owner would wedge the NodeType at `Compiling` — and after this
  change its failure mode is benign: the retry it provokes adopts. Removing it is a call about that
  wedge risk, not about data loss, and it is named as such in the doc page.

## Falsification

`ReleaseRecutIsARetryTest` (real `MonolithMeshTestBase` mesh, no mocks, no `Task.Delay`, no polling,
no timing dependency — the two attempts are back to back). The settle's identity is stamped at
`UtcNow - 5min` **on purpose**: a body that re-derived the id from `DateTime.UtcNow` would silently
*agree* with an identity minted in the same second, so against a stamp of its own, disagreement is
certain rather than likely.

Three arms, each a partial revert of the `src/` change with the test file untouched, each built and
run:

| Arm | `src/` state | Result |
|---|---|---|
| **fix** | as committed | **2 passed, 0 failed** |
| **A** | identity re-derived per attempt (`MintReleaseIdentity(result)` inside the write), upsert kept | **1 passed, 1 failed** — `expected …20260906132215-qaGX-_Sn … found …20260906132715-qaGX-_Sn` |
| **B** | settle identity kept, write reverted to `CreateNode` | **1 passed, 1 failed** — `expected …20260906132317-qaGX-_Sn … but found <null>` (the literal #3407 symptom) |
| **C** | both reverted (pre-fix behaviour) | **1 passed, 1 failed** |

The arm that passes in every case is `ADifferentBuild_StillGetsItsOwnRelease` — the falsifying arm:
convergence is per *identity*, so a "fix" that stapled every attempt to one path would satisfy the
retry assertion while making a second release impossible.

Whole suite with the fix: **626 passed, 0 failed** (`MeshWeaver.Compiler.Pipeline.Test`).
`DocumentationLinkIntegrityTest`: **1 passed**.
`dotnet build -c Release -warnaserror`, one project per invocation, all `0 Error(s) 0 Warning(s)`:
`MeshWeaver.Compiler.Pipeline`, `MeshWeaver.Hosting`, `MeshWeaver.Documentation`,
`MeshWeaver.Compiler.Pipeline.Test`, `MeshWeaver.Documentation.Test`, `MeshWeaver.Graph.Test`,
`MeshWeaver.Hosting.Test`.

## Work products

- `Doc/Architecture/ReleaseRecutIdentity` — the durable page: the invariant, the two outcomes the
  wall clock used to pick between with the measured evidence, why the write is re-attempted at all,
  the shape, and what this deliberately does not do. Linked from the Architecture topic map.
- A What's New entry (`Category: Fix`) — a recompiled type serving its old code is something a user
  notices.

Pairs-with: none — no public surface is removed or changed; everything touched is `internal`
(`TryCreateReleaseNode`, `ReleasePostCondition`, `NodeTypeBatchBake`), reachable only through
`InternalsVisibleTo`, and the additions are new members.

Closes #3407


rbuergi:
Superseded by #3428, which landed on main while this was in the merge queue. #3428 is the better change: its adoption argument rests on the id embedding the durable content reference, and it additionally fixes the typed `NodeAlreadyExists` reason being discarded at both create surfaces — the reason `IsAlreadyExistsRace` was parsing English and missing one of three producers. This PR's separate finding (the later-second duplicate pair, measured on the live mesh) is recorded on #3407 so it is not lost.

Issue #2625, opened 2026-08-28
Flake on main: FaultedInit_ReachesTerminalFailedState_AndAnswersFast — hub reaches FAILED in 4 ms, the probe request is never rejected (15 s timeout)


rbuergi:
## What
`MeshWeaver.Data.Test.DataContextInitFaultedTest.FaultedInit_ReachesTerminalFailedState_AndAnswersFast` (added by #2528's fix, `5204e0fd4`, 13:55Z) failed on **main** at `22722c2a` (the #2615 merge), shard 4, run 33181004785 — and passed on main at `5610756` 35 minutes earlier. Same test source both times ⇒ a race, not a regression of the test.

## What the log says
```
[15:03:13.197] [SYNC_STREAM] OnError for BEWp1B6u9E6s7qW2AH2R3w (Reference=FaultyItem, Owner=ds/lgZYlcRS80eNi5DNLEG4aQ)
[15:03:13.197] Data source stream lgZYlcRS80eNi5DNLEG4aQ errored
[15:03:13.197] Hub sync/-BGwAgrYBkWk0QMqCgQo1Q initialization failed — a BuildupAction faulted (InvalidOperationException: boom during initial data load). Hub is now in FAILED stat…
[15:03:13.201] [SYNC_STREAM] OnError for Bp-yJrv0rkm-lHu7hh55-A (Reference=FaultyItem, …)
[15:03:28.309] === TEST FAILED: Did not expect value to be of type TimeoutException because a faulted init must be answered FAST by the rejection handler …
```
The host hub reaches FAILED **4 ms** after class init. The client's `ProbeRequest` (`Observe … FirstAsync().Timeout(15s)`) then gets **nothing** for 15 s — not a rejection, not a `DeliveryFailure`. So one of:

1. the request arrived while the hub was still initialising, was **deferred**, and the faulted arm transitioned to FAILED **without flushing the deferred buffer through the rejection handler** — the queued request is simply dropped; or
2. the request arrived **after** FAILED and the FAILED path answers only requests that passed through the init gate, so a cold arrival is not rejected.

Either way the assertion's own message names the defect exactly: "the faulted arm skipped the settle body and left the gate shut". It is the wedges-to-zero branch the watchdog is supposed to guarantee.

## How to pin it
[...]

Issue #1455, opened 2026-08-13
Workspace._localStreamCache has ReduceShared's liveness defect without its parent guard — a replaced primary would serve the old parent's child


rbuergi:
Split out of #1324, where it was found and confirmed but is off-topic. **Not manifesting today** — filed so the next person who touches either cache finds the analysis instead of re-deriving it, and so a future change cannot quietly arm it.

## The shape

#1425 added `ISynchronizationStream.ReduceShared`, a reference-keyed cache of anonymous intermediate reduces. Its first cut copied `Workspace.GetStream`'s liveness check verbatim and that was **wrong**, for a reason the PR recorded:

> 🚨 A DISPOSED PARENT IS NEVER SERVED FROM CACHE — sharing must not outlive the thing being shared FROM. The cached child's own sub-hub is a sibling under `Host`, not a child of this stream's hub, so it stays alive when this stream is disposed: a liveness check on the child alone happily hands back a mirror of a dead source. Reads then bind to a stream that will never emit and never complete, so a reader gets neither an answer nor the NACK that `DataExtensions.HandleGetDataRequest`'s disposal arm owes it — it hangs for its whole budget.
> — `src/MeshWeaver.Data/Serialization/SynchronizationStream.cs:118-127`

`ReduceShared` therefore bypasses its cache once `isDisposed` (`:127`). **`Workspace._localStreamCache` has the identical construction and no such guard.**

## The two caches, side by side

| | `ReduceShared` | `Workspace._localStreamCache` |
|---|---|---|
| declaration | `SynchronizationStream.cs:113` | **`Workspace.cs:634`** |
| key | `WorkspaceReference` alone | `WorkspaceReference` alone |
| child-liveness check | `SynchronizationStream.cs:154-158` | **`Workspace.cs:667-671`** — *character-for-character the same predicate* |
| parent guard | ✅ `:127` `if (isDisposed) return Reduce(...)` | ❌ **none** |

`Workspace.cs:667-671`:

```csharp
// Same liveness contract as GetExternalClientSynchronizationStream: a cached stream
// whose sub-hub is gone (its parent was torn down, or a consumer disposed it) is
// replaced rather than handed out dead.
if (stream.Hub?.RunLevel <= MessageHubRunLevel.Started
    && stream.Hub is not MessageHub { IsDisposing: true })
    return (ISynchronizationStream<TReduced>)stream;
```

[...]

PR #3427, opened 2026-09-06
fix(#3321): step 3 — a disposed stream releases its hub

rbuergi:
Step 3 of 3 for #3321 — the step that actually reclaims the leak. Steps 1 (#3380, make liveness
askable) and 2 (#3386, migrate the dereference sites) landed first, in that order, because doing
this one before them is the production NRE this issue is about.

`Closes #3321`.

## 🚨 `Dispose()` clearing the field would have reclaimed 11 of 1 496

The step is usually described as "`SynchronizationStream.Dispose()` clears `Hub`". Read against the
heap dump on the issue, that description is a **sixth of the fix**:

```
SynchronizationStream total=8461 disposed=11
   1485  <MeshNode>     hub RunLevel=Dead, stream NOT disposed
     11  <JsonElement>  hub RunLevel=Dead, stream DISPOSED
```

A `sync/` hub is a HOSTED hub: its parent disposes it during the parent's own teardown — a Blazor
circuit ending, a `DisposeRequest`, a recycle — while the stream that created it is owned by a
workspace somewhere else and **is never told**. `Dispose()` never runs for those 1 485.

So the release happens at **both ends**, through one private `ReleaseHub()` (idempotent, CAS'd,
callable from any thread; it disposes nothing — it only drops a reference):

| End | Where | Covers |
|---|---|---|
| the stream is disposed | `Dispose()`, after `Hub.Dispose()` | the 11 |
| the hub dies underneath it | `syncHub.RegisterForDisposal(_ => ReleaseHub())`, from the constructor | the 1 485 |

The constructor's hook is registered **after** `RegisterForDisposal(resyncSubscription)`, and that
ordering is load-bearing: that call is what hooks `streamDisposables` onto the hub, and a hub's
composite disposes in registration order — so the stream's own registrants run first and still see a
bound `Hub`, and the release runs after them. A hub already past accepting registrants disposes the
hook inline, so the constructor re-checks and refuses with `HubDisposingException` exactly as its
`syncHub is null` branch does.

No new state enters the liveness vocabulary: `StreamLiveness.IsUsable` has always treated
`current.Hub is not { } hub` as dead, so a released stream reports itself unusable through the
predicate every cache already consults.

## The site classification — 29 found, 4 untouched, 25 resolved

A fresh sweep of `src/` for every `.Hub` whose receiver resolves to an `ISynchronizationStream`
found **29 occurrences on 27 lines**. Four stay untouched and must: two ARE the predicate
(`StreamLiveness.IsUsable`'s chain walk, `TryGetHub`'s own read), one is already `stream?.Hub?.…`,
one is `Workspace`'s cache probe, which is itself a liveness check.

The other **25** are this step's work — the 20 step 2 named, plus five it did not:

- the **three** `LayoutAreaHost` reads step 2 called "hub-turn confined". That rested on *a hub
  executing a turn is by construction alive* — a statement about the HUB. Step 3 makes the FIELD
  independently clearable, so a queued `UpdateStreamRequest` can now run after the release. The
  premise expired with the change that needed it.
- `Workspace.EvictClientSubscriptions` — step 2 placed it in "already answerable" for the right
  reason (an `IsUsable` guard there would SKIP a disposal), but a null check is still needed.
- `WorkspaceExtensions.ApplyChanges` — **step 2's list missed it** (`stream!.Hub.Version`, behind a
  null-forgiving `!`).

| Group | Sites | Resolution |
|---|---|---|
| **Patch reducers** — `StandardReducers` (10), `MeshDataSource.PatchMeshNode` (2) | 12 | `RequireHub()`, absorbed by their single gateway |
| **creation-path `Subscribe` callback** — `WorkspaceStreams` (2), `PartitionedHubDataSource` (2) | 4 | `HubIfHeld()` once at the top; **skip the emission** |
| **inside a `stream.Update(…)` lambda** — `LayoutAreaHost` (3), `VirtualDataSource` (1) | 4 | `HubIfHeld()`, then `return null` — `Update`'s documented no-op |
| `JsonSynchronizationStream`'s `reduced.Hub.Register` | 2 | `HubIfHeld()` once; **don't register** |
| `GenericUnpartitionedDataSource.Initialized` | 1 | a released stream contributes **no task to wait for** |
| `WorkspaceExtensions.ApplyChanges` | 1 | `RequireHub()` — same shape as a reducer |
| `Workspace.EvictClientSubscriptions` | 1 | `Hub is { }` — deliberately not a liveness check |
| | **25** | |

**The reducers were the design problem, and their single gateway is the answer.** A reducer's
signature is `… → ChangeItem<T>`: it MUST return a change, so it has no absent value to hand back.
But every patch reducer in the codebase is invoked from exactly ONE place —
`JsonSynchronizationStream.ToChangeItem`, the only call site of `ReduceManager.PatchFunction` — and
`null` is what `ToChangeItem` **already returns** for a stream with no `PatchFunction` registered. So
the reducers read the hub once through `RequireHub()`, which refuses with the TRANSIENT
`HubDisposingException` (an `ObjectDisposedException` → `ErrorType.ShuttingDown`, the "ask again"
answer, never the terminal `DeliveryFailure` the predecessor's raw NRE produced), and `ToChangeItem`
catches **only that type** and answers `null` — which both callers already handle
(`BuildChangeItem` falls through to its type-registry path; `UpdateStream` falls back to
deserializing through `Host.JsonSerializerOptions`). Not a swallow, and not a
`catch (NullReferenceException)`: every other exception a reducer throws still propagates.

## 🚨 PRESENCE is not LIVENESS — and the wrong one is a behaviour change, measured

The first draft wrote every owner-side guard as `TryGetHub()`, on the reasoning that step 2's rule
was "migrate onto `TryGetHub()`". **That is wrong, and one existing test proved it before the change
left the branch.**

`DataSource.Initialized` is `Task.WhenAll(streams.Select(s => s.Hub.Started))`. Guarded with
`TryGetHub()`, a stream whose **initial load faulted** is excluded — because it is not `IsUsable`,
which is exactly what `IsFaulted` is for. So the WhenAll no longer held the task carrying the fault,
`Initialized` completed **successfully**, and a hub whose data source had thrown went on answering
requests as though nothing had happened. `DataContextFaultedInitBeforeStreamHubBoundTest` went red
on precisely that — the #2625 wedge, re-introduced by a guard that looked like a tightening of
safety.

So there are two accessors for two questions:

| Accessor | Question | Where |
|---|---|---|
| `TryGetHub()` (`IsUsable`) | *should I use this stream?* — chain walk, faulted store, winding-down hub | a consumer, a cache, a read — step 2's 28 sites |
| `HubIfHeld()` / `RequireHub()` | *is the reference still there?* — one field read | an owner writing into its own stream |

**The rule for an owner-side guard is: match the guard the site's own gate already applies.** Each of
them ends in `stream.OnNext` or `stream.Update`, whose refusal is `isDisposed || Hub is null`, so a
presence check refuses in exactly the cases the write would have been dropped anyway and never one
more. `TryGetHub()` there is a silent policy change smuggled in under a null-safety fix — the shape
step 2 already declined at `Workspace`'s cache probe. `RequireHub()` is written in terms of
`HubIfHeld()`, so there is one definition of presence and one of liveness, never a fourth
hand-copied predicate (#1455).

## And every read inside the stream now resolves the field once

Step 2 excluded `SynchronizationStream`'s own bare `Hub` reads as "already guarded". They were —
against a field that could not change. Once the field can be nulled, a guard followed by a *re-read*
is check-then-act across threads, so every read below a guard now uses the local the guard resolved:
`OwnerVersion(hub)`, `BuildChangeItem(hub, …)`, `BuildFullChangeItem(hub, …)`, and the locals in
`RegisterForDisposal` / `DeliverMessage` / `OnNext` / `OnError`.

One log level moved for a reason, not for debugging: `OnError`'s deliberately LOUD "the stream has no
Hub bound" `LogError` (#2625) gained a second, benign way to be reached — the hub was released
because it died — and that arm logs Debug instead. The original arm, for a stream whose hub was never
bound, is unchanged.

## Tests, and the measurement that says they can fail

`test/MeshWeaver.Data.Test/StreamReleasesItsHubTest.cs`, 6 cases. The flagship assertion is a
**`WeakReference`**, not `Hub is null`: "the field reads null" is a statement about one field;
**"the hub is unreachable"** is the statement this issue is about, and it is the only one that goes
red on the defect for the right reason. The hub is created and dropped inside a non-inlined helper
so no live local on the test's frame can root it. It passes.

Falsified by disabling **only** the two `ReleaseHub()` call sites — everything else, `RequireHub` and
every guard included, left in place, so the experiment isolates exactly what step 3 does — and
re-running both suites unchanged:

| Suite | release disabled | release enabled |
|---|---|---|
| `StreamReleasesItsHubTest` (`MeshWeaver.Data.Test`) | **4 failed, 1 passed** | **5 passed** |
| `TornDownStreamCallSitesTest` (`MeshWeaver.Layout.Test`) | **4 failed, 1 passed** | **5 passed** |

(Measured before the sixth case was added.) The live control passed in both runs of both suites — a
release that fired unconditionally, or a guard that answered "dead" always, would satisfy every
negative at once, and the controls are what rule that out. The four pre-change failures name four
different facts: `Hub` not null after `Dispose()`, `Hub` not null after the hub died underneath,
`RequireHub` not throwing, and `ToChangeItem` still deriving a change off the corpse.

A *full* `src/` revert does not compile the tests, because `RequireHub` is itself part of step 3 —
which is why the falsification reverts the release rather than the whole change.

Two existing assertions INVERT here, deliberately kept rather than deleted, because they are the
lines that pin which contract is in force:
`TornDownStreamCallSitesTest` and `StreamLivenessIsConsumerReachableTest` both asserted
`stream.Hub.Should().NotBeNull()` after disposal — "the non-nullable contract still holds, which is
precisely why a consumer needed a separate way to ask". Step 3 is what makes that stop holding.

## Verification

Local, `-c Release -warnaserror`, one project per invocation:

| | Result |
|---|---|
| **whole solution** (64 projects) | **0 Warning(s), 0 Error(s)** |

| Suite | Result |
|---|---|
| `MeshWeaver.Data.Test` | 484 passed, 0 failed |
| `MeshWeaver.Layout.Test` | 450 passed, 0 failed |
| `MeshWeaver.Graph.Test` | 1366 passed, 0 failed |
| `MeshWeaver.Documentation.Test` | 279 passed, 0 failed |

## Work products conserved

`Doc/Architecture/StreamLivenessAndTheHubReference` gains "What step 3 did" — the 11-vs-1 485 split,
the two-ended release with the registration-order constraint, the 29/4/25 site table with the reason
per group, the measured PRESENCE-vs-LIVENESS finding, and the falsification numbers. Plus a What's
New entry (`Category: Fix`).

## i18n

No new user-visible strings. `HubDisposingException`'s message is existing framework text.

## Cross-repo

`Pairs-with: Systemorph/MeshWeaver.Plugins#1405`

🚨 **Not because this removes public surface — it removes none** (zero public types, zero public
members; the gate does not fire). This is **shape 7**: `ISynchronizationStream.Hub` is a public
member of a public interface, and step 3 changes what it ANSWERS behind an unchanged signature,
which no surface detector can see by construction.

A by-hand sweep of every satellite checkout, on the RECEIVER type and including in-mesh `.json`
nodes, found the blast radius is one repo and one cluster: **11 production dereferences in
`MeshWeaver.Plugins`'s Blazor view layer**, all on the inherited
`ISynchronizationStream<JsonElement>? Stream`. The `!` in `Stream!.Hub.Something` is on `Stream`,
not on `.Hub`, so the compiler stays silent; one of them sits inside a bare `catch { }`, so it would
have been silent at runtime too. The four content satellites are clean structurally
(`ISynchronizationStream` appears in zero files in four of them).

**The dependent landed FIRST**, and this PR was held as a draft until it did — its guards are no-ops
on a live stream, so they are safe against the old core, while core's release is not safe against an
unguarded dependent. The full table is in the doc page under
"The dependent half — and why no gate could have told you".

### DIFF
--- a/src/MeshWeaver.Data/GenericUnpartitionedDataSource.cs
+++ b/src/MeshWeaver.Data/GenericUnpartitionedDataSource.cs
@@ -309,13 +309,27 @@ public abstract record DataSource<TDataSource, TTypeSource>(object Id, IWorkspac
 
     /// <summary>
     /// A task that completes once every created stream's hub has started.
+    ///
+    /// <para>🚨 A stream that has RELEASED its hub contributes nothing to wait for (#3321 step 3):
+    /// the hub is gone, so its <c>Started</c> task cannot be reached, let alone settle.</para>
+    ///
+    /// <para>🚨 <b>And the guard is PRESENCE (<c>HubIfHeld</c>), never liveness
+    /// (<c>TryGetHub</c>) — measured.</b> Writing this with <c>TryGetHub()</c> broke
+    /// <c>DataContextFaultedInitBeforeStreamHubBoundTest</c>: a stream whose initial load FAULTS is
+    /// not <c>IsUsable</c>, so it was excluded from the WhenAll, this task completed
+    /// SUCCESSFULLY, and a hub whose data source had thrown went on answering requests as though
+    /// nothing had happened. A faulted stream's <c>Started</c> task is the very thing that must
+    /// still be awaited — it is how the fault reaches <c>DataContext</c>'s gate.</para>
     /// </summary>
     public Task Initialized
     {
         get
         {
             lock (Streams)
-                return Task.WhenAll(Streams.Values.Select(s => s.Hub.Started));
+                return Task.WhenAll(Streams.Values
+                    .Select(s => s.HubIfHeld())
+                    .Where(h => h is not null)
+                    .Select(h => h!.Started));
         }
     }
     /// <summary>
--- a/src/MeshWeaver.Data/ISynchronizationStream.cs
+++ b/src/MeshWeaver.Data/ISynchronizationStream.cs
@@ -86,14 +86,22 @@ public interface ISynchronizationStream : IDisposable
 
     /// <summary>The message hub associated with this stream.
     ///
-    /// <para>🚨 <b>Non-nullable in the contract, but not in life.</b> A stream whose owner has been
-    /// torn down keeps answering this property, and the hub it hands back may already be winding
-    /// down — dereferencing it is how a recycle window produced an NRE inside
-    /// <c>LayoutAreaHost</c>'s constructor that escaped to the subscriber as a TERMINAL
-    /// <c>DeliveryFailure</c> (Systemorph/MeshWeaver#3321). On any path that can run concurrently
-    /// with teardown, prefer <see cref="SynchronizationStreamLiveness.TryGetHub"/>, which answers
-    /// <c>null</c> instead, or gate on
-    /// <see cref="SynchronizationStreamLiveness.IsUsable(ISynchronizationStream?)"/>.</para>
+    /// <para>🚨 <b>Non-nullable in the contract, but not in life — and since #3321 step 3 it is
+    /// ABSENT on a stream that has let its hub go.</b> A disposed stream, and a stream whose hosted
+    /// sub-hub was killed by its parent's teardown, both RELEASE this reference: that release is
+    /// what reclaims the <c>stream → dead hub → resolved state</c> graph that held ~580 MB of a
+    /// production replica's heap. The declaration stays non-nullable because these assemblies ship
+    /// as packages and changing it would break every downstream implementer.</para>
+    ///
+    /// <para><b>So never dereference this on a path that can run concurrently with teardown.</b> Use
+    /// <see cref="SynchronizationStreamLiveness.TryGetHub"/>, which answers <c>null</c>, or gate on
+    /// <see cref="SynchronizationStreamLiveness.IsUsable(ISynchronizationStream?)"/>. Where the
+    /// surface has no absent value to return — a patch reducer, whose signature is
+    /// <c>… → ChangeItem&lt;T&gt;</c> — use
+    /// <see cref="SynchronizationStreamLiveness.RequireHub"/>, which refuses with the TRANSIENT
+    /// <c>HubDisposingException</c> rather than an NRE. A raw dereference here is how a recycle
+    /// window produced an NRE inside <c>LayoutAreaHost</c>'s constructor that escaped to the
+    /// subscriber as a TERMINAL <c>DeliveryFailure</c>.</para>
     /// </summary>
     IMessageHub Hub { get; }
     /// <summary>The hub that hosts the underlying data source backing this stream.</summary>
--- a/src/MeshWeaver.Data/Persistence/PartitionedHubDataSource.cs
+++ b/src/MeshWeaver.Data/Persistence/PartitionedHubDataSource.cs
@@ -98,13 +98,27 @@ namespace MeshWeaver.Data.Persistence
                 .Synchronize()
                 .Subscribe(changes =>
                 {
+                    // 🚨 #3321 step 3 — same shape as WorkspaceStreams' combined subscription: a
+                    // callback in flight when the stream is torn down finds `ret.Hub` released.
+                    // Dropping the frame is what `ret.OnNext` would do anyway on a released stream,
+                    // so it is the answer this void callback already models. PRESENCE, not liveness
+                    // (HubIfHeld, not TryGetHub): this guard must match OnNext's own
+                    // `isDisposed || Hub is null` exactly, so it refuses precisely when the
+                    // emission would be dropped and never a frame that would still have landed.
+                    if (ret.HubIfHeld() is not { } retHub)
+                    {
+                        Logger.LogDebug(
+                            "PartitionedHubDataSource: combined stream {StreamId} is torn down — "
+                            + "dropping this frame", ret.StreamId);
+                        return;
+                    }
                     if (!isInitialized)
                     {
                         var initialStore = changes
                             .Where(c => c.Value != null)
                             .Aggregate(new EntityStore(), (acc, change) => acc.Merge(change.Value!));
 
-                        ret.OnNext(new ChangeItem<EntityStore>(initialStore, ret.StreamId, ret.Hub.Version));
+                        ret.OnNext(new ChangeItem<EntityStore>(initialStore, ret.StreamId, retHub.Version));
 
                         foreach (var change in changes)
                             processedChangeItems.Add(change);
@@ -120,7 +134,7 @@ namespace MeshWeaver.Data.Persistence
                                 .Where(c => c.Value != null)
                                 .Aggregate(new EntityStore(), (acc, change) => acc.Merge(change.Value!));
 
-                            ret.OnNext(new ChangeItem<EntityStore>(updatedStore, ret.StreamId, ret.Hub.Version)
+                            ret.OnNext(new ChangeItem<EntityStore>(updatedStore, ret.StreamId, retHub.Version)
                             {
                                 ChangedBy = string.Join(", ", newChanges.Select(c => c.ChangedBy).Where(cb => !string.IsNullOrEmpty(cb))),
                                 Updates = newChanges.SelectMany(c => c.Updates).ToArray(),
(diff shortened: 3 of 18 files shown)


rbuergi:
## 🚨 HELD AS DRAFT — the dependent half has to land first

A read-only sweep of every satellite checkout on this machine for `.Hub` **on a stream receiver**
found the break this change would cause outside core. It is shape 7 of the seven — a method's
BEHAVIOUR changing behind an unchanged signature — which no surface detector can see by
construction, and which the `Cross-repo pair (public surface)` gate therefore does not fire on
(measured: this diff removes **zero** public types and **zero** public members).

### What the sweep found

| Repo | stream-receiver `.Hub` | Raw `.Hub` lines classified |
|---|---:|---:|
| **MeshWeaver.Plugins** | **22** (11 production, 11 test) | 2 051 |
| education | 0 | 483 |
| MeshWeaver.Education | 0 | 11 |
| MeshWeaver.Reinsurance | 0 | 152 |
| MeshWeaver.SocialMedia | 0 | 73 |
| MeshWeaver.Manufacturing | 0 | 15 |

The four content satellites are clean *structurally*, not just by token match:
`ISynchronizationStream` appears in ZERO files in four of them and in exactly one file in
Reinsurance, where the stream is only `.Update(…)`d. Every in-mesh carrier was swept too — the 27
`.json` nodes carrying `.Hub` inside a C# `configuration` string all resolve to
`IWorkspace.Hub` / `LayoutAreaHost.Hub` / an `IMessageHub` local.

### The 11 that would NRE

All in `MeshWeaver.Plugins`, all on one inherited property —
`BlazorView.razor.cs:62`, `ISynchronizationStream<JsonElement>? Stream`:

- `BlazorView.OnClick` → `Stream.Hub.Post(new ClickedEvent(…))`
- `FormComponentBase.OnBlur` → `Stream.Hub.Post(new BlurEvent(…))`
- `DialogView.HandleClose` / `HandleDialogResult` → `Stream!.Hub.Post(new CloseDialogEvent(…))`
- `ViewModelExtensions.GetControl`, `FormComponentBase.ConversionToValue`,
  `DataGridView.GetPropertyExpression`, `RadzenChartView.BindData` ×2,
  `RadzenPivotGridView.BindData`, `RadzenChartView.razor:171` →
  `Stream!.Hub.JsonSerializerOptions`

🚨 **The `!` is on `Stream`, not on `.Hub`** — so after this change the NRE moves one level to the
right, past the operator that currently suppresses the warning, and **the compiler stays silent**.
And `RadzenChartView.razor:171` sits inside a bare `catch { }`, so there it would be invisible at
runtime too: a wrong chart instead of an exception.

That is precisely the incident this issue exists to avoid, relocated one repo over — an NRE in a
Blazor view during a recycle window. Shipping core first would ship it.

Also flagged, not an NRE but a silent test regression:
`MeshWeaver.Hosting.Monolith.Test/StreamCacheLivenessTest.cs:146-151` is already null-tolerant
(`as` + `?.`), but its `.Where` predicate becomes permanently `false` once `Hub` can be null, so it
would stop observing the cascade and **hang to its 60 s timeout** rather than fail an assertion.

### Order

Plugins first, core second. Un-drafting this once the Plugins counterpart is merged.

`Pairs-with:` will be filled in with that PR before this leaves draft.

rbuergi:
## Hold released — the dependent half merged

Systemorph/MeshWeaver.Plugins#1405 is merged into that repo's `main` (`cdfaef28`), so the 11 Blazor-view dereferences that would have NRE'd on a released hub are guarded before this lands. Order held: dependent first, platform second.

Its CI also surfaced one thing worth recording here, because it is the shape that repeats: **every red on that PR cascaded from a single stale `Radzen/manifest.lock`** — editing the chart views changes the module's content hash, and the module-bundle gates fail RED rather than skip when an input job fails (#890), so `Compile every NodeType (vs core)`, both `test-repos` jobs and the two Tests-area gates were all reporting `validate=failure` rather than a finding of their own. Regenerating the lock cleared all six at once. 🚨 `gen-manifests.py` also rewrote **15 other locks** with an identical `moduleVersion` and only a bumped `version`/`sourceCommit`; committing those would have published new versions of 15 untouched modules and collided with every other in-flight PR, so only Radzen was kept — and `validate-repos.py` agrees (259 nodes valid across 60 repos).

Marking ready. Auto-arm will enqueue this on its own.

Issue #3454, opened 2026-09-06
Three satellites verify nothing about the digest they pin — SocialMedia, Manufacturing and Crm have no pin-consistency gate at all, so a half-moved set merges green


rbuergi:
## Three satellites have NO pin-consistency gate — a wrong digest merges green

Measured 2026-09-06 while staging the coordinated platform-pin move for the 3.0.0 wave.

| repo | `check-platform-pins.py` | "Both platform pins name ONE promoted build" | pin sites a move must pass through |
|---|---|---|---|
| MeshWeaver.Plugins | ✅ | ✅ | **14** |
| MeshWeaver.Reinsurance | ✅ | ✅ | **12** |
| **MeshWeaver.SocialMedia** | ❌ **absent** | ❌ **absent** | 4 |
| **MeshWeaver.Manufacturing** | ❌ **absent** | ❌ **absent** | 1 |
| **MeshWeaver.Crm** | ❌ **absent** | ❌ **absent** | 1 (a step output, not `env:`) |

**0 matches** in the bottom three — this is not a gate that skips, it is a gate that does not exist.
🚨 The distinction matters because the fleet's usual failure is a *skipped* gate wearing a passing
tick; here there is nothing to skip. A pin move that names one build in one place and a different
build in another **merges green** in those three repos, and the first symptom is whatever the wrong
reference set does at bake or compile time.

## Why this is a blocker-list item rather than a nice-to-have

The pins are moving **right now** as one coordinated set for the release wave, and they move again at
the tag. In Plugins and Reinsurance a half-moved set is caught mechanically. In SocialMedia,
Manufacturing and Crm it is caught by whoever notices.

The precedent is on record and it was expensive: **a withdrawal that took a line it did not add
dropped a 4th pin, took core CD down for an hour, and every check was green** (#3344). That is
exactly the shape these three repos cannot detect.

## What the gate would have to check

Not merely "is the digest well-formed". The two properties that matter:

1. **Every pin in the repo names ONE promoted build** — the set name, the source ref, the tester
   digest and the portal digest must be mutually consistent, including sites outside the `env:` block.
2. **The digests actually resolve** in ACR at the tag the set names (`--check-tags`). 🚨 This one is
   load-bearing given #3438: ACR retention has already deleted a *pinned* tester digest and wedged an
   entire repo's CI including its own nightly.

🚨 **A placeholder must fail RED, not silently pass.** `check-platform-pins.py` matches
[...]

Issue #3399, opened 2026-09-06
Dependabot PRs are red fleet-wide: the Dependabot secret store is a partial mirror of Actions secrets


rbuergi:
## Every Dependabot PR in the fleet is red at `preflight`, and the gate is right to be

Dependabot workflow runs read secrets from a **separate store** (`Secret source: Dependabot` in the
run log) — the repo's Actions secrets are simply not visible to them. Each satellite's Dependabot
store is a *partial* mirror of its Actions store, so `preflight` finds an empty required secret and
fails red:

```
##[error]Required CI inputs are not provisioned, so the heavy gate cannot run.
         A gate that cannot run fails RED — it must never go grey and read as a pass.
##[error]missing: secrets.REGISTRY_PUBLISH_TOKEN — …
```

**The preflight is behaving exactly as designed** (AGENTS.md: *"a gate NEVER tests its own inputs —
no skip-trapdoors"*). This is a **provisioning** gap, not a workflow bug, and the fix must not be an
`if:` that asks whether a secret is set — that is the precise anti-pattern the rule forbids.

## The gap, per repo

| repo | missing from the **Dependabot** store | how established |
|---|---|---|
| **Manufacturing** | `ACR_USERNAME`, `ACR_PASSWORD`, `MW_REGISTRY_INSTANCE_KEY` | measured — run [33357319017](https://github.com/Systemorph/MeshWeaver.Manufacturing/actions/runs/33357319017) names all three |
| **SocialMedia** | `REGISTRY_PUBLISH_TOKEN` | measured — run [33354469022](https://github.com/Systemorph/MeshWeaver.SocialMedia/actions/runs/33354469022) names it; `ACR_*` are already mirrored there |
| **Reinsurance** | `AZURE_CLIENT_ID`, `AZURE_SUBSCRIPTION_ID`, `AZURE_TENANT_ID`, `MW_REGISTRY_KEY` | store diff (Actions ∖ Dependabot) against its preflight's own `missing+=` list |
| **Crm** | *none* | store diff — **and Crm is the one satellite that drained to zero today**, which is the control that makes the whole table credible |
| **Education** | *none by this guard shape* | its preflight asserts differently; not yet measured |

Note the two naming variants across the fleet — `MW_REGISTRY_INSTANCE_KEY` (Manufacturing) vs
`MW_REGISTRY_KEY` (the rest) — worth converging separately.

## Remediation

One command per missing secret, same value as the Actions secret:

```bash
[...]

PR #3462, opened 2026-09-06
A dead image pin is named, not discovered by a repo's CI dying (#3438 guard half)

rbuergi:
## What this lands

The **guard half** of #3438: a daily sweep that reds when a repository's pinned image digest no
longer resolves in ACR, naming the repository, the declaration and the digest.

It does **not** stop the deletion. It converts *"an entire repository's CI is inexplicably dead,
including its nightly"* into one named red. The retention half is deliberately **not** in this PR —
see below.

- `.github/scripts/check-pinned-digests.py` — the sweep
- `.github/workflows/pinned-digests.yml` — schedule 05:20 UTC (after `purge-old-images` at 03:00 and
  `purge-old-ci-releases` at 04:00), `workflow_dispatch`, and push-to-`main` on its own two files
- `.github/workflows/dotnet-test.yml` — `--self-test` in the workflow-shell lane, so the extractor is
  proven on every PR
- `src/MeshWeaver.Documentation/Data/Architecture/PinnedImageRetention.md` + topic-map entry + a
  What's New (`Category: Fix`)

## Why Memex#122's rule does not cover this

That rule is *"if this tag were deleted tonight, would anything recreate it?"* — and `mw-plugin-test`
**passes** it. It is republished many times a day. The hazard is not "nothing rebuilds this
repository", it is:

> Something **PINS** a specific digest of it — and `--keep N` is counted in **newer builds**, so
> frequent republishing is what **destroys** the pin rather than what protects it.

A pin is *designed* to sit still for weeks (`Doc/Architecture/ModuleVersioning` — pins move as one
set, deliberately and rarely). `--ago 7d --keep 10` guarantees any pin older than about a week names
bytes that no longer exist. The retention window and the intended pin lifetime were in direct
contradiction and nothing reconciled them.

## Falsification — both arms, against the live registry (2026-09-06)

**Dead arm** — `--repos Systemorph/MeshWeaver.Education` → **exit 1**, and it names **three** gone
pins, not just the tester #3438 reports:

```
    distinct pins found                   3
    …that RESOLVE                         0
    …that are GONE                        3

      GONE  memex-migration@sha256:d81df6cc79ad…  ci.yml:MW_MIGRATION_DIGEST
      GONE  memex-portal-ai@sha256:dab2a7b3a7e1…  ci.yml:MW_PORTAL_DIGEST
      GONE  mw-plugin-test@sha256:df19f10afc1f…   ci.yml:MW_TEST_DIGEST
```

🚨 **New finding: Education's portal and migration pins are gone too.** Both repositories are in
`purge-old-images`'s filter list, so the same task took all three. #3438 names only the tester.

**Live arm** — `--repos Systemorph/MeshWeaver.Plugins,Systemorph/MeshWeaver.Reinsurance` →
**exit 0**, 6/6 resolve.

**Fleet, with the denominator printed:**

```
    repositories scanned                  8
    …declaring at least one digest pin    6
    …declaring NO digest pin              2  (Systemorph/MeshWeaver, Systemorph/Memex)
    …whose workflows could not be read    0
    distinct pins found                   12
    …that RESOLVE                          9
    …that are GONE                         3
    …INDETERMINATE (registry not reached)  0
    pin-shaped declarations MALFORMED      0
```

🚨 **#3454's "SocialMedia, Manufacturing, Crm have no pin gate — 0 matches" is a zero with two
causes.** All three **do** pin: Manufacturing as `MW_IMAGE_DIGEST`, SocialMedia and Crm inside a
`$GITHUB_OUTPUT` step output. And Education pins under `MW_PORTAL_DIGEST` / `MW_MIGRATION_DIGEST` /
`MW_TEST_DIGEST`. A check keyed on the two common names reports four of six pinning repositories
clean while verifying none of their pins — which is why extraction here is by **shape**, and why the
report states both halves of the denominator.

## Built not to pass on no evidence

- **The denominator is printed**, and a sweep that extracted **no** pins at all fails — that proves
  the extractor broke, not that the fleet is clean.
- **Placeholder vacuity is a failure.** `check-platform-pins.py` matches `sha256:[0-9a-f]+`, so
  `sha256:PLACEHOLDER` reads to it as "no literal pin present" and passes. Here an *attempted* digest
  that is not one reds — and so does a truncated one that a loose matcher would accept. The predicate
  is deliberately narrow (value starts with `sha256`) so a nightly sweep cannot cry wolf on
  `digest: $SHELL_VAR` or a YAML key with a block beneath it; the self-test asserts both non-firings.
- **The fleet list is discovered**, from `/installation/repositories`, never committed.
  `MeshWeaver.Crm` pins and is **absent** from `.github/shared-rules.json`'s `repos` array.
- **"Cannot reach" is never "gone".** A control probe runs first; only the registry's own
  `manifest unknown` counts as absence; every other `az` answer is INDETERMINATE and reds naming the
  output.
- **No skip-trapdoor.** `preflight` asserts every input and fails RED naming what to provision; the
  sweep `needs:` it with no input-shaped `if:` and no `continue-on-error:`. Every job carries a
  literal `timeout-minutes` ≤ 45.
- **No `pull_request` trigger** — a registry's live state is not a property of a PR, and a
  credentialed PR-reachable job would enter the Dependabot store's blast radius (#3399) for no gain.
  The extractor's falsification runs on every PR instead.

Local gates, all clean: `check-workflow-timeouts.py --root .` (68 jobs, 0 violations),
`check-pr-secret-preflight.py --root .`, `check-workflow-shell.py --root .` (0 live findings),
`actionlint`, `DocumentationLinkIntegrityTest` (Passed, and both new pages verified embedded in the
built assembly so the pass is not vacuous).

## 🚨 The retention half is NOT in this PR, and cannot be

**The purge task definitions are cloud-only.** Measured three ways: `az acr task show` returns an
`EncodedTask` with `contextPath: null` (the YAML is base64 in Azure, no source repository); an
org-wide code search for `acr purge` finds only doc pages; and the deployment repo carries no bicep,
terraform or task definition for them (`deploy/aks/infra/main.bicep` treats `meshweaver.azurecr.io`
as an existing `sharedAcrLoginServer`). **There is no IaC PR to open** — changing retention is a live
change to shared registry infrastructure and is the maintainer's to apply.

The design is in the doc page with the exact commands. In brief:

- **`acr purge` skips locked manifests by default**; `--include-locked` is the opt-in to delete them,
  and neither task passes it. So `az acr repository update --delete-enabled false` on a pinned
  manifest is a hard protection against this exact configuration, not a hint.
- **Raising `--ago`/`--keep` is explicitly not a fix** — it moves the cliff. A pin left stable for a
  quarter walks off a 30-day window exactly as it walked off a 7-day one.
- **The two tasks having diverged is itself the defect, and measurably so:** four of the five
  repositories `purge-old-ci-releases` (30d) names are already swept an hour earlier by the stricter
  `purge-old-images` (7d, keep 10), so for those four the 30-day task is a **no-op**. The only
  repository unique to it is `memex-portal`; the only one unique to the aggressive task is
  `mw-plugin-test` — the one that broke. One task, one filter list removes the "which of two lists?"
  question by construction.
- 🚨 **But merging them must NOT tighten `memex-portal`** (corrected after review — see the comment
  below). The two lists differ by exactly two repositories and they are not symmetric: the one
  unique to the 30-day task is the **production portal image**. Unioning the filters under the
  stricter window would move it to `7d --keep 10` with nobody deciding to — and this PR's guard
  cannot cover it, because `memex-portal` is pinned on the **other axis** (committed tags in the
  deployment overlays, Memex#141), which a scan of `.github/workflows` never enumerates. The task
  in the doc is therefore **behaviour-preserving**: two steps, each keeping the window that
  repository already has. It removes the second *task*, not the second *window*. The lock set is
  correspondingly the **union of both pin axes**.

## Not closing #3438

`Refs #3438`, deliberately. Half of it — the retention change — is outstanding and is the
maintainer's to apply.

Also worth recording: `Systemorph/Memex`'s `scripts/check-image-pins.py` covers the sibling axis
(tags committed in helm overlays, Memex#141) and its own header says it is **not wired into CI**.
That gap is unchanged by this PR.

Pairs-with: none — this change set adds only `.github/` scripts and documentation; it removes no
public type or member from `src/`.


rbuergi:
🚦 **Held as DRAFT deliberately — not because anything here is unfinished.**

Measured just now (2026-09-06 ~17:45Z): `main-cd` run **7919 is `pending`** on `cb675ecfd`, behind
7917 `in_progress`, and five `ci/pin-set-next-seal` draft PRs are open across the fleet
(Plugins#1415, Reinsurance#170, SocialMedia#138, Manufacturing#56, Crm#56) all titled *"awaiting the
seal"*.

Under `main-cd`'s one-pending-slot rule (#3379) **a merge to `main` cancels the pending run and
re-queues on the newer tip** — which is the mechanism `ContinuousDeliveryContract` records as having
made three sessions write a wrong run number inside one hour today. That makes merging this a shared
action against a seal somebody is gating on, so it is AGENTS.md's *"hold and say so"* case rather
than a merge-on-green.

The change set itself is complete and CI is running on it. Nothing here is time-critical: the guard
fires nightly, and the pins it would red today (Education's three) are already dead.

**Mark ready and let auto-merge take it once the pin-set seal has landed** — or say the word and I
will.

rbuergi:
## Independently re-measured before landing — both arms, against the live registry

Re-ran every claim in this PR from a fresh worktree rather than inheriting it (2026-09-06, ~19:00Z).

### Ground truth on the two digests #3438 names

```
mw-plugin-test@sha256:df19f10a…            ERROR: manifest unknown        (az exit 3)
mw-plugin-test:3.0.0-ci.7917               sha256:642f686fed743289…       (az exit 0)
```

### Arm 1 — DEAD. `--repos Systemorph/MeshWeaver.Education` → **exit 1**

```
    repositories scanned                  8
    …declaring at least one digest pin    6
    …declaring NO digest pin              2  (Systemorph/MeshWeaver, Systemorph/Memex)
    …whose workflows could not be read    0
    distinct pins found                   12
    …that RESOLVE                          9
    …that are GONE                         3
    …INDETERMINATE (registry not reached)  0
    pin-shaped declarations MALFORMED      0

  GONE  memex-migration@sha256:d81df6cc79ad…  ci.yml:MW_MIGRATION_DIGEST
  GONE  memex-portal-ai@sha256:dab2a7b3a7e1…  ci.yml:MW_PORTAL_DIGEST
  GONE  mw-plugin-test@sha256:df19f10afc1f…   ci.yml:MW_TEST_DIGEST
```

Each is a separate `::error::` naming repository, declaration and digest. Reproduces the PR's
figures exactly.

### Arm 2 — LIVE. Five pinning repos → **exit 0**, `10/10` resolve

```
    repositories scanned                  5
    …declaring at least one digest pin    5
    distinct pins found                   10
    …that RESOLVE                         10
    …that are GONE                         0
```

🔎 **The guard tracked the release wave mid-run.** Between the two sweeps `MeshWeaver.SocialMedia`
went from one pin to two and adopted `mw-plugin-test@sha256:642f686fed74…` — the `3.0.0-ci.7917`
digest above. The shape-driven extractor picked up the new `platform-image-digest` declaration with
no change, which is the property a name-keyed check would not have had.

### Arm 3 — the resolver itself, on exactly the two briefed digests

```
sha256:df19f10a…  (known-dead)              -> GONE
sha256:642f686f…  (mw-plugin-test:3.0.0-ci.7917) -> RESOLVES
```

## Deliverable (B): cloud-only — re-established independently

- `az acr task list` → both tasks are `EncodedTask` with **`contextPath: null`**.
- Decoded both task bodies; the filter lists, windows and the Memex#122 reasoning-in-one-task-only
  match this PR's table **exactly**.
- Org code search: `purge-old-images` → **0 hits**; `"acr purge"` → 2 hits, both doc pages here.
- `Systemorph/iac` contains only `.gitignore` and `README.md`.
- `deploy/aks/infra/main.bicep` defaults `sharedAcrLoginServer` to `meshweaver.azurecr.io` and
  instantiates `acr.bicep` only `if (!useSharedAcr)`; that module declares no `tasks` child.

**There is no IaC PR to open.** Confirmed against upstream `Azure/acr-cli` that `acr purge` skips
locked manifests by default and `--include-locked` is the opt-in to delete them — neither task
passes it, so `--delete-enabled false` is a hard protection under this exact configuration.

## Landing checks

`meshweaver-cloud` is installed org-wide (`repository_selection=all`, 34 repos), so `--discover`
sweeps the whole org rather than a committed six. `actions/checkout@v7`,
`actions/create-github-app-token@v3`, `azure/login@v3` all match this repo's uniform usage, and all
five preflight secrets exist on the repo. Local gates re-run clean: workflow timeouts (68 jobs, 0
violations), PR-secret preflight, workflow-shell (0 live), `--self-test`,
`DocumentationLinkIntegrityTest` (Passed), Release `-warnaserror` build `0 Warning(s) 0 Error(s)`.

⚠️ **Expect this workflow to be RED on `main` the moment it lands**, until Education's pin move
completes — Education's three pins are gone right now. That red is the guard working, and it gates
nothing: `main-cd` selects on `Consolidate test results` **by name**, not on all check runs.

rbuergi:
## Three corrections since the falsification above

### 1 + 2 — both automatic-review findings, fixed (they were this gate's own defect class, inward)

- **`discover_repos()` could lose a page of the fleet.** It recovered `gh api --paginate`'s
  per-page JSON documents by splitting the text on braces and **swallowed** the `JSONDecodeError`
  when a chunk split badly. A lost page is a lost *repository*, and that repository then appears in
  the report neither as a failure nor as a measured zero — simply absent, while the run goes green.
  Now `--jq '.repositories[].full_name'`, which is also character-for-character what the workflow's
  preflight asserts with, so the two cannot disagree about who is in the fleet.
- **"distinct pins found" counted declaration SITES.** Measured now: **15 declarations over 9
  distinct digests**. The old label invited *"15 images are pinned"* when 9 manifests are. Both are
  reported and labelled — declarations are what a pin move must edit, distinct digests are what
  retention must keep alive — and the failure summary leads with *N declarations name M missing
  manifests*, because 3-from-1 is one retention event and one fix while 3-from-3 is a different
  size of problem.

### 3 — 🚨 the proposed retention merge would have tightened `memex-portal`. Fixed.

This is the one worth a maintainer's attention, because the doc hands over commands to run against
shared registry infrastructure.

The two filter lists differ by exactly two repositories and they are **not symmetric**:

| unique to | repository | what it is |
|---|---|---|
| `purge-old-images` (7d, keep 10) | `mw-plugin-test` | the CI tester — the one that broke |
| `purge-old-ci-releases` (30d) | **`memex-portal`** | the **production portal image** |

The obvious merge — union the filters, keep the stricter window — moves `memex-portal` from
`--ago 30d` to `--ago 7d --keep 10` **with nobody deciding to**. Worse, this PR's guard cannot cover
it: `memex-portal` is pinned on the **other axis**, committed image TAGS in the deployment overlays
(Memex#141), which `check-pinned-digests.py` never enumerates because it reads `.github/workflows`.
A lock set derived from the guard alone leaves every overlay-pinned manifest unlocked *while the
window that deletes them gets shorter* — which is exactly how Memex#122's victim was pinned.

The proposed task is now **behaviour-preserving by construction**: two steps, each keeping the
window that repository already has. It removes the second **task**, not the second **window**, and
the task body says why. Retuning `memex-portal` is a separate decision with its own evidence. The
page now also states that the lock set is the **union of both pin axes**.

Re-verified after all three: `--self-test` green · dead arm still exit 1 naming Education's three ·
`--discover` on a user token fails RED naming the installation token rather than returning an empty
fleet · workflow timeouts (68 jobs, 0 violations) · PR-secret preflight · workflow-shell (0 live) ·
Release `-warnaserror` `0 Warning(s) 0 Error(s)` · `DocumentationLinkIntegrityTest` Passed.

rbuergi:
## 4th correction — found by falsifying the gate, not by reading it

**An absent repository read as one with no pins, and the sweep went green.**

Listing `.github/workflows` answers **404 for two opposite situations**:

| 404 because | correct verdict |
|---|---|
| the repository is fine and has no workflows | a **measured zero** |
| the repository is absent, renamed, or unreadable by the token | **nobody looked** |

The script believed the first. Measured before the fix:

```
--repos Systemorph/MeshWeaver.Plugins,Systemorph/ThisRepoDoesNotExist12345
  Systemorph/ThisRepoDoesNotExist12345: 0 workflow file(s), no digest pin declared
  All 4 pin declaration(s) — 2 distinct digest(s) — across 1 repository(ies) resolve.     exit 0
```

A repository nobody looked at, printed **in the denominator** as swept and clean — this PR's own
thesis failing inside this PR's own gate. The repository is now probed before any absence is
believed. Both arms after the fix:

```
Systemorph/ThisRepoDoesNotExist12345  ->  UNREADABLE, ::error::, exit 1
Systemorph/iac  (real repo, no workflows)  ->  "0 workflow file(s), no digest pin declared", exit 0
```

`--discover` could not hit this — names come from the API — but `--repos` is the documented local
and manual mode, and a renamed or access-revoked repository reaches the same 404 there.

## Landing readiness, re-checked on the final head

- **The nightly will actually be able to log in.** `main-cd.yml` uses the same three Azure secrets
  with `azure/login@v3` on `main` and reads ACR manifests with them; its recent runs are green, so
  the OIDC subject (`ref:refs/heads/main`, which is also what a `schedule` run presents) and the
  registry read permission are both proven for the exact identity this workflow will use.
- **The fleet list will not be short.** `meshweaver-cloud` is installed org-wide
  (`repository_selection=all`, 34 repositories), so `--discover` sweeps the org, not a committed six.
- **Fleet re-measured, unchanged:** 8 scanned · 6 pin · 2 do not · **15 declarations over 9 distinct
  digests** · 12 resolve · **3 GONE** (Education's, still) · 0 indeterminate · 0 malformed.

⚠️ Expect `Pinned image digests` to be **RED on `main` immediately after this lands**, until
Education's pin bump completes. That red is the guard working and it gates nothing — `main-cd`
selects on `Consolidate test results` **by name**, not on all check runs.

Issue #3472, opened 2026-09-06
Adopt-time identity guard: a process must refuse a bake keyed to a framework identity that is not its own, and compilationStatus must not read Ok for a type whose hubs cannot activate


rbuergi:
## Measured on memex.systemorph.com, 2026-09-06

Two CRM types stopped serving their layout areas while their NodeType nodes report `compilationStatus: Ok`:

| type | compiled at (UTC) | `compiledFrameworkVersion` | `latestAssemblyPath` | behaviour |
|---|---|---|---|---|
| `Crm/Client` | 19:19:38 | `s2f227642d43f78eab13720c4af06a1be` | `Crm_Client/v31703-s2f22764-…` | renders |
| `Crm/Mail` | 19:19:41 | `s2f227642d…` | `Crm_Mail/v11-s2f22764-…` | renders (Approve/Send page fine) |
| `Crm/Offer` | 19:28:04 | **`sc273ee39fdccbfc088f9aaf1fc548a9a`** | `Crm_Offer/v22488-sc273ee3-…` | `get @PartnerRe/Esl/EslProposal/area/Offer` → activation **timeout** |
| `Crm/Opportunity` | 19:31:30 | **`sc273ee39f…`** | `Crm_Opportunity/v13760-sc273ee3-…` | `get @PartnerRe/Esl/area/Deal` → **"Area not found"**, only the generic areas listed; recycling the instance hub does not help |

Same sources for all four (`shared=@Crm/Source`, unchanged since 2026-09-05 12:47Z; `Crm/_GitSync` last synced 2026-08-30). The `!toolchain` mvid differs too (`22f6327b…` on the s2f22764 pair, `847e4ad9…` on the sc273ee3 pair). `Ops/Status/memex` reads *Degraded, 2 of 3 replicas ready, runningImage 3.0.0-rc9.ci.7647* (sampled 2026-09-03), and the only place `sc273ee3` shows up for this repo is the identity of the tester set MeshWeaver.Plugins pins as `3.0.0-ci.7859` (MeshWeaver.Crm #55 moved its pin there at ~09:50Z today and its publish-bake ran green).

[...]

Issue #3441, opened 2026-09-06
The release gate's denominator is read from the artifact it judges: a package whose bake breaks silently leaves the check, so the fleet keeps rolling onto releases it has not baked


rbuergi:
> Maintainer, 2026-09-06: *"let's see that the deploy rolls only when **all** packages of all plugins are baked"* — *"we kept rolling without edu being properly baked."*

The release-availability gate (#1754) already asks the right question and already holds the roll when the answer is no. It nonetheless let environments roll onto releases whose Education content was not baked, **while reporting a clean pass**. The defect is its DENOMINATOR.

## The defect

`ReleaseAvailabilityService.RequiredPackages` builds the gate's package list from the environment's install records — a mesh query, genuinely independent, and correct. It then decides per package whether a baked bundle is required:

```csharp
var adoptedToday = PublishedBundleCatalogue.SealedBundlesForIdentity(
    publishedRoot, PrebuiltAssemblySeeder.LiveFrameworkMvid, logger);

new RequiredPackage(manifest.Id, manifest.Id, LiveFloorOf(manifest.MinMeshVersion),
                    HasContent: adoptedToday.Contains(manifest.Id))
```

`HasContent` **is** the denominator, and it is read from the artifact store the gate is about to judge. So:

- Education's bake stops producing a sealed publication;
- its packages leave `adoptedToday`;
- leaving `adoptedToday` sets `HasContent: false` — *"ships no content"*;
- a package that ships no content cannot have a missing bake, so **it is exempt**;
- every later roll is green about precisely the packages that regressed.

The denominator **erodes**, and it erodes exactly where the gate is needed. A package broken today is exempt tomorrow and cannot re-enter the set on its own.

### The limiting case is worse

If the live identity has **no** sealed publication at all (a fresh storage target, a bake lane pointed at the wrong share, a `BAKE_PUBLISH_TARGETS` that stopped being written), `adoptedToday` is empty, *every* installed package becomes non-content-bearing, and the content half of the gate checks **nothing** while answering "updatable" — a green tick over zero evidence, with nothing anywhere saying the expected set was empty. That is the `0 expected, 0 found, green` shape AGENTS.md forbids in CI gates, living in the runtime.

## Fix shape

[...]

Issue #3433, opened 2026-09-06
EaGraphAuth.LoadAsync renders every failure as 'this user never connected' — a swallowed 10s timeout tells a connected user to re-consent, and there is no token cache so the timeout is the expected outcome


rbuergi:
## The defect

`memex/Memex.Portal.Shared/Authentication/EaGraphAuth.cs`, `LoadAsync`:

```csharp
// line ~198, INSIDE the try
    .Take(1).Timeout(TimeSpan.FromSeconds(10)).FirstAsync()
...
// line ~212
catch (Exception ex)
{
    ...
    return (null, null);
}
```

`(null, null)` is the **same value** this method returns when a user genuinely has no stored
credential. So every failure mode — the 10-second timeout on its own node read, a transport fault, a
deserialization problem — is rendered to the caller as *"this user has never connected their
mailbox."* `GetAccessTokenAsync` then answers `null`, and the Executive Assistant shows the
re-consent link.

There is a second one at line ~223: a bare `catch { return null; }`.

A user who IS connected is told to connect again. Pressing that link re-consents something that was
never revoked, so it "works", which is why this has survived: the failure looks like a successful
recovery from a state that never existed.

## Why the timeout fires — the second half

**There is no token cache at all.** Every EA tool call:

1. reads the credential node,
2. redeems the refresh token,
3. on rotation — i.e. essentially always — **writes that same node**.

A single agent round issuing ~20 parallel `SearchMail` calls therefore puts ~20 writes and ~20 reads
on ONE single-threaded node hub. The 10-second read timeout is not a remote possibility under that
load; it is the expected outcome. In the 2026-09-06 incident the credential write at 12:29:49.375
was one of them, 372 ms after the round went silent.

So the two halves compound: contention makes the read time out, and the swallow turns the timeout
into a lie about the user's connection state.

## What was excluded (so nobody re-derives it)

- **The #1369 `ReplaySubject` latch** — excluded by construction: no cache exists, so nothing can
  latch a transient fault.
- **Rotation invalidating the stored token** — a `client_secret` is sent, and confidential clients do
  not get single-use refresh tokens. Rotation matters here for node CONTENTION, not validity.
- **The credential being absent** — read directly at 12:35 and present.

**Not excluded:** Entra answering non-2xx. One named query distinguishes it —
`{app="memex-portal"} |= "EaGraphAuth"` over 12:29:40–12:33:00Z. The `Hosting/LogEntry` window on
[...]

PR #3483, opened 2026-09-06
fix(#3429): install-time prebuilt adoption says WHY it adopted nothing

### LINKED ISSUE #3429: Install-time prebuilt adoption silently adopts nothing for a package installed after boot — Edu's 10 baked assemblies sat in /bundles while its types compiled in-mesh, and the seed logged neither adoption nor a reason
Found while reading why Education's disposable-mesh shards take 30 minutes to fail (PR #250, run [34037514537](https://github.com/Systemorph/MeshWeaver.Education/actions/runs/34037514537), shard 1, artifact `e2e-mesh-report-shard1` → `mesh-diagnostics/portal.log`). The harness half is fixed in Education#250; this is the platform half it uncovered.

## What happened, from the shard's portal log

Mounted bundles: 40 (`/bundles`, the bake's output for this very run — the bake mesh adopted `Edu.zip` **10/10**, framework `sa55f8190d2f21a651c19ad14d56107a7`).

| time | event |
|---|---|
| 14:17:18 | `Package 'Store': … 188 nodes installed` |
| 14:20:31 | `ShippedPrebuiltBundles: bundle Store.zip: adopted 5/5 prebuilt assembly(ies); 12 were already current` — the install-time seed did its job for Store, 3 min after the writes |
| 14:32:35 | `Package 'Edu': … 99 nodes installed`, `Node created at Edu` |
| 14:32:38–39 | `Edu/Exercise/_Activity/compile-state`, `Edu/Lesson/_Activity/compile-state` … created — in-mesh compiles begin |
| 14:32:35 → 14:46:51 (end of log) | **no** `bundle Edu.zip: adopted …`, **no** `Install: adopted N prebuilt assembly(ies)`, **no** `Install: prebuilt adoption attempt failed after …` |

The shard's consumption verdict: `bundles mounted: 40 · assemblies backed by bundles: 17 · declined: 0 · in-mesh compiles: 14`. The 17 are Store's. Edu's ten were neither adopted nor declined — they were never candidates — and Edu's types paid Roslyn on a mesh that had their bytes on disk. That is why the exercise pages had no workbench inside the harness's window after Edu finally installed.

## Why it is silent

`PackageInstaller.SeedPrebuiltAssemblies` (called after the writes, `result.Written > 0`) logs only two outcomes: `adopted > 0` → `Install: adopted …`; a timeout/fault → `Install: prebuilt adoption attempt failed after {bound}`. A seed that completes with **0** says nothing. `ShippedPrebuiltBundles.SeedForTypes` → `SeedBundles` likewise reports per bundle only what it backed. So "ten matching assemblies on disk, none adopted, no reason given" is indistinguishable in the log from "no bundles mounted".

I have not established which filter dropped Edu's payloads (the "NodeType nodes the mesh already holds" selection at the moment the seed runs right after the same install's writes, or a fingerprint/identity comparison against the just-written type nodes) — that needs the seeder's own trace, which this log does not carry. What is established: same run, same `/bundles`, same framework identity, Store adopted at install, Edu not.

## Why it matters

- A package installed after boot is the normal case for every consumer of the registry (the boot lane cannot see types that arrive later — `e2e/mesh/docker-compose.yml` documents exactly that, and names the install-time seed as "the piece that matters"). If the install-time seed silently adopts nothing, every such install compiles in-mesh, and on a saturated mesh (#2543 shape) that is the 15 minutes this run lost.
- Education's e2e budget is sized for adopted assemblies; in-mesh compiles of Edu's ten types alone exhaust it.

## Asks

1. `SeedPrebuiltAssemblies` / `SeedForTypes` report a zero with its reason per type path — *no bundle covers this path*, *payload fingerprint differs from the installed node*, *framework identity differs*, *node not yet visible to the seeder* — the way the boot lane already prints `DECLINED: … built against mvid:X, live is mvid:Y`. A silent zero is the class of gate that cannot fail.
2. Then the actual filter, once the reason is visible. Adjacent to #3359/#3385 (seed-once vs reconcile) rather than the same defect: this is the install-time lane, and the bundles were present.

<sub>Read from the run's own artifacts; nothing on the mesh was touched. Filed from the Education side, where the cost lands.</sub>

rbuergi:
Closes #3429

Pairs-with: none — this PR removes no `public` type and no `public` member from `src/`. It only ADDS (`AddPrebuiltAssemblyConsumption`, the shortfall reporting) and widens two `private` signatures; `SeedTally` moved `private` → `internal`, which was never public surface.

## The defect, and why it is the day's recurring shape

A package installed **after boot** depends entirely on install-time adoption: the boot sweep filters bundles against the NodeTypes the mesh already **holds**, and a package that arrives later was not in that snapshot. On the measured run (MeshWeaver.Education e2e, run 34037514537 shard 1) 40 bundles were mounted at `/bundles` — including `Edu.zip`, which the bake had adopted **10/10** under the very framework identity the harness was running. `Store` installed and adopted 17. Three minutes later `Edu` installed 99 nodes, adopted **nothing**, and its ten NodeTypes went to Roslyn.

The log carried no adoption line, no decline line and no reason. **A completed seed that covered zero said exactly as much as a deployment with no bundles at all** — the same family as #3441 (a gate reading its denominator from the store it judged), #3433 (a swallowed timeout returning the "never connected" value) and a check-suite with zero check-runs read as all-green: *a failed or absent action rendering indistinguishably from a successful one*.

Four silent zeros existed:

| where | the zero |
|---|---|
| `PackageInstaller.SeedPrebuiltAssemblies` | logged only `adopted > 0` and a timeout |
| same | returned 0 in silence when no `IPrebuiltAssemblyConsumer` was registered |
| same | returned 0 in silence when the install recognised no NodeType definitions |
| `ShippedPrebuiltBundles.SeedBundles` | an absent or empty bundle directory logged at `Debug`, invisible where it matters (CI and prod both run at `Information`) |

## What a non-adoption now says, and where

**`ShippedPrebuiltBundles.SeedForTypes`** — the one place every consumer route converges (install, install-delta, cross-package dependents, git push, on-demand compile watcher) — witnesses its own coverage and, when it covered less than it was asked for, emits **one** line naming the uncovered type paths plus exactly one reason:

| the pass observed | the line says | level |
|---|---|---|
| no source directory exists | this deployment has no bundle lane; names both config keys | `Information` |
| sources exist, no bundles | nothing published **for this framework identity** | `Information` |
| bundles read, **no entry named these types** | ← **Edu's shape**: never candidates; compare the bundles' node paths with the install's | `Information` |
| entries **did** name them and coverage fell short | the bytes were here and did not land | **`Warning`** |

The discriminator is `EntriesMatched` — bundle entries that named a **requested** type, counted before anything can decline them. Answer 4 is also where #3472 lands: a portal adopting bytes stamped for a *different* framework identity is worse than adopting none, and both must be distinguishable from success. `DescribeShortfall` / `IsShortfallAFailure` are **pure**, so the reasoning is pinned with no mesh, no bundle and no disk.

**`PackageInstaller.SeedPrebuiltAssemblies`** reports its outcome **unconditionally** and names the **package**, which the seeder cannot:

```
Install: Edu: adopted 3 prebuilt assembly(ies) for 10 installed type(s) — …          Information
Install: Edu: adopted NO prebuilt assembly for any of 10 installed type(s) — …        Warning
Install: Edu: the install recognised no NodeType definition among the nodes it wrote  Information
Install: Edu: NO prebuilt assembly was adopted … this host registers no
             IPrebuiltAssemblyConsumer — it consumes no bundle source at all          Warning
```

`NodeTypeRecompileExtensions` says the same for the push lane's null consumer.

## The established mechanism for "after boot"

Building the repro found a concrete one: **the only way to obtain an `IPrebuiltAssemblyConsumer` was `AddDynamicTypePreWarming()`**, which also registers a hosted service requiring `IHostApplicationLifetime`. A composition without a generic host therefore had **no consumer**, the installer's `GetService` answered null, and adoption was skipped silently no matter how many bundles were mounted beside it. (The repro hit exactly this: the test mesh could not even resolve the pre-warmer.) Two unrelated decisions are now two calls — `AddPrebuiltAssemblyConsumption()` and `AddDynamicTypePreWarming()`, the latter a superset, so every existing host is unchanged.

That is *a* mechanism, not necessarily *the* one that hit Edu — the shard's log cannot say which filter dropped its payloads, which is the whole point of the issue. What this PR guarantees is that the next occurrence names itself.

## Verification — both arms, with numbers

**With the fix**

| suite | result |
|---|---|
| `InstallTimePrebuiltAdoptionTest` (new) | **2 / 2** |
| `PrebuiltShortfallSpeaksTest` (new) | **9 / 9** |
| `MeshWeaver.Hosting.Test` | 385 / 385 |
| `Memex.Portal.Shared.Test` | 828 / 828 |
| `MeshWeaver.Documentation.Test` (incl. `DocumentationLinkIntegrityTest`) | 304 / 304 |

Actual lines produced by the run:

```
INSTALL Information: Install: PrebuiltPkg: adopted 1 prebuilt assembly(ies) for 1 installed type(s) …
INSTALL Warning:     Install: PrebuiltPkg: adopted NO prebuilt assembly for any of 1 installed type(s) …
MESH   Information:  ShippedPrebuiltBundles: adopted no prebuilt assembly for 1 of 1 requested
                     NodeType(s) — PrebuiltPkg/Widget. All 1 bundle(s) under … name only NodeTypes
                     OUTSIDE this set, so none of these types was ever a candidate. Either the bake
                     did not cover this package, or the paths it baked differ from the paths this
                     install wrote — compare the bundle manifests' node paths with the names above.
```

**Reverting `src/` with the tests unchanged**

- The 9 pure tests **do not compile** — `DescribeShortfall` / `IsShortfallAFailure` do not exist and `SeedTally` is inaccessible. 0 pass.
- Keeping ONLY the registration split (so the integration arms still compile) and reverting all the loudness: **0 passed / 2 failed**. The adoption arm fails on its **log** assertion having already passed the MVID assertion — i.e. the adoption happened and nothing said so, which is the defect exactly.

The positive-control arm is not decoration: without it the silence arm would pass just as well against a harness in which adoption is unreachable, which is the shape of a gate that cannot fail.

## Docs

- New: `Doc/Architecture/InstallTimePrebuiltAdoption` — why the boot lane cannot serve a post-boot install, the measured timeline, the four answers a zero must keep apart, the registration split, the grep lines, and what this deliberately does **not** cover (store modules under `/data/modules` are pinned at process start; that lane's gap is a pod restart).
- Linked from the Architecture index; `DocumentationLinkIntegrityTest` green.
- What's New (`Category: Fix`).


rbuergi:
All three findings are valid and are **accepted**, fixed and verified locally — but this PR had already entered the merge queue at position 1 with two other PRs behind it (#3482, #3481) when the push landed, and GitHub refused the update rather than ejecting it. A dequeue restarts every queued entry's build, so I am not imposing that on two other sessions to improve a log line; the fixes land immediately as a follow-up on top of this one.

**1. `DescribeShortfall`'s denominator (correct, and the most important of the three).** `tally.Covered` and `EntriesMatched` are per-bundle-**entry** sums, so a type carried by both the image bundle and the CI-published root contributes two entries and one path — `uncovered + Covered` could read "1 of 3 requested" for a pass that asked for two, and `EntriesMatched > Covered` could misclassify the Warning/Information level. That is precisely the confidently-wrong signal this PR exists to remove, reproduced inside the cure.

Fixed at the root rather than by patching the arithmetic: the pass now witnesses **matched paths** as a second per-subscription set (written where `present` is computed, before anything can decline an entry) alongside the covered set; the requested count is passed in explicitly; `SeedTally.EntriesMatched` is deleted. The line got strictly better for it — answer 4 now *names* the offered-but-uncovered paths rather than quoting a count. Pinned by a new test, `ADenominatorIsTheREQUESTEDSet_NotASumOfBundleEntries`, which fails against the derived denominator.

**2. The `consumer is null` + empty `nodeTypePaths` corner.** Correct — the guard order made the method's own "reported unconditionally" doc false, which is the *prose asserts a guarantee the code does not keep* trap. The empty-type-set check now runs first (it is also the more specific, more actionable fact), so every path emits exactly one line.

**3. Temp directory never deleted.** Correct. Cleaned up in a `finally`, matching this suite's own convention in `PrebuiltPublicationTest`.

Verified after the fixes: `MeshWeaver.Hosting.Test` 386/386, `Memex.Portal.Shared.Test` 828/828, `MeshWeaver.Documentation.Test` 304/304, every touched project clean at `-c Release -warnaserror`.

Issue #3473, opened 2026-09-06
IEmailSender has no Cc/Bcc — a message with copied recipients cannot be delivered as one mail


rbuergi:
`IEmailSender.SendEmail(toAddress, subject, htmlBody, attachments, delivery)` takes ONE recipient and no Cc/Bcc, and `SendDocumentDispatch.ExportAndSend` therefore sends one mail per recipient. A client mail with a To and seven Cc — the ordinary shape of a business mail — cannot be delivered as the one message the sender approved: the recipients would each get a separate mail with no Cc line, and replies would not reach the copied people. The Graph implementation builds a `Message` that supports Cc, so the gap is the contract, not the transport.

Ask: an overload (or an `EmailMessage` record) carrying `To`, `Cc`, `Bcc`, `Subject`, `HtmlBody`, `Attachments`, `EmailDelivery`, with the no-op and SMTP senders refusing (not silently dropping) Cc they cannot honour. This is what would let `Essentials/Email`'s Send actually deliver — through the user's connected mailbox, with the PDF exports of its `attachments` — instead of only recording the send, and show the plan (from, to, cc, subject, attachments) before the click.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

rbuergi:
**Core half in flight: Systemorph/MeshWeaver#3493** (paired with #3450 — both change `IEmailSender`,
so they land as one contract change rather than two cross-repo cycles).

**What landed in core**

`EmailMessage` carries the whole envelope — `To`, `Cc`, `Bcc`, `Subject`, `HtmlBody`, `Attachments`,
`Delivery` — and `IEmailSender.SendEmail(EmailMessage)` (plus `hub.SendEmail(EmailMessage)`) sends it
as ONE mail. `VisibleRecipients` is To ⧺ Cc, de-duplicated case-insensitively, and **never** Bcc; a
blind address appearing there would be the disclosure the field exists to prevent.

**The refusal is the part this issue asked for explicitly, and it is honoured.** The default
implementation forwards to the existing single-address overload when the message fits one (`To` of
exactly one, no copies) — so every sender written before this member keeps serving every call it
could already serve — and otherwise **faults** with a `NotSupportedException` naming the counts. It
never delivers a quietly smaller mail: doing so is the same class of defect as stamping an
undelivered mail `Sent` (#2023).

`NoOpEmailSender` is the one deliberate exception, and for the reason the issue implies rather than
[...]

Issue #3450, opened 2026-09-06
CanSendAsUser has two answers for three states: an undetermined read shows a connected user the "connect your mailbox" panel


rbuergi:
Found while reviewing the MeshWeaver#3433 fix. It is the **same defect shape, one size smaller**, and it is deliberately NOT being fixed inside #3433 — see "Why not in #3433" below.

## The shape

`IEmailSender.CanSendAsUser` returns `IObservable<bool>`. Two answers, three states of the world:

| world | answer today |
|---|---|
| the user has connected their mailbox | `true` |
| the read completed and found no credential | `false` |
| **the read did not complete** (timeout, transport fault, unreadable content) | **`false`** |

MeshWeaver#3433 is exactly this collapse one layer up, in the Executive Assistant's tools, where it
rendered as *"please connect your mailbox"* to a user whose grant was stored and valid. Here the
third row is folded into the second by the signature itself.

## Where it is user-visible, precisely

`CanSendAsUser` has exactly two consumers, both in
`src/MeshWeaver.Markdown.Export/Layout/SendDocumentLayoutArea.cs`:

1. **At send (~line 321)** — `false` calls `AskWhichMailbox(...)`, which shows the user which mailbox
   the mail would leave from and lets them **choose**. That is a prompt, not a capability claim, and
   the code says so: *"The send does NOT proceed on an assumption either way."* **This half is fine
   and needs no change.**

2. **At dialog open (~line 173)** — `false` renders `ui.sendDocument.connectFirst` plus a **Connect**
   button. That *is* a user-visible "you have not connected" statement, and on the third row of the
   table it is false.

**Two things make it milder than #3433, and they are already in the code.** The panel is an OFFER,
not a gate — the comment at that site is explicit: *"the form below is fully usable, and sending
without connecting simply goes from the shared mailbox with a reply-to"* — and the send path
re-probes and asks rather than asserting. So the worst case is a connected user shown a Connect
panel they do not need, on a form that still works. In #3433 the tool **refused to act** and the link
was the only path offered.

## 🚨 This is broader than the Undetermined state, and predates it

Both consumers already wrap the probe in `.Catch((Exception _) => Observable.Return(false))`. So a
**faulted** probe has always become `false` at both sites, with nothing logged. The reactive-seam
change (MeshWeaver#3446 + its Plugins half) does not introduce this; it makes a newly *modelled*
Undetermined join a pre-existing swallow, and it adds the Warning line — naming the user and the
[...]

PR #3493, opened 2026-09-06
Model the third send-as state, and let a mail carry Cc/Bcc (#3450, #3473)

rbuergi:
Two defects on the same interface — `IEmailSender` — fixed as **one coherent contract change**, so
the second does not rebase onto a signature the first moved.

`Pairs-with: none` — measured, not asserted. `scripts/check-type-forwards.py --base HEAD` reports
`removed: []`: this diff removes **0 public types and 0 public members**. Both new members are
default interface methods, so the `Cross-repo pair (public surface)` gate does not fire. See
"Purely additive" below.

---

## (1) #3450 — `CanSendAsUser` had two answers for three states

| world | before | now |
|---|---|---|
| the user connected their mailbox | `true` | `Available` |
| the check completed, no credential | `false` | `Unavailable` |
| **the check did not complete** | **`false`** | **`Undetermined`** |

The send dialog renders `false` at OPEN time as `ui.sendDocument.connectFirst` — *"Your Microsoft 365
mailbox is not connected yet"* — beside a Connect button. On a transient fault that is a
**user-visible statement that is false**, shown to a user who **had** connected. Both consumers
already wrapped the probe in `.Catch(_ => Observable.Return(false))`, so the collapse **predates**
the #3433 reactive seam and nothing was logged — the third state was not even greppable.

**The rendering rule lives on the answer, not in each UI.** `EmailSendAsCapability.OffersConnect` is
true for `Unavailable` **alone** — deliberately *not* `!IsAvailable`, because that expression is the
defect written out. A consumer cannot re-derive it by accident.

`hub.ObserveSendAsCapability(...)` routes a **faulted or silent** probe into `Undetermined` and logs
it at Warning naming the user and the diagnostic — once, where the sender is resolved, instead of
every consumer having to remember. A probe that completes without emitting gets the same state: that
is exactly how a dropped mesh read looks.

🚨 **The send path is deliberately unchanged.** `hub.CanSendAsUser(...)` still folds `Undetermined`
to `false`, so pressing Send still shows `AskWhichMailbox` rather than assuming an identity. That is
the *correct* answer for a genuinely binary decision, and it is the positive control below.

## (2) #3473 — `IEmailSender` had no Cc/Bcc

Every overload took ONE address, so `ExportAndSend` sent one mail per recipient. A To with seven Cc —
the ordinary shape of a business mail — went out as **eight** messages: no recipient could see who
else received it, Reply-All reached only the sender, the thread split eight ways, and **Bcc had no
expressible form at all**.

`EmailMessage` carries `To`/`Cc`/`Bcc`/`Subject`/`HtmlBody`/`Attachments`/`Delivery`;
`IEmailSender.SendEmail(EmailMessage)` sends it as ONE mail. `VisibleRecipients` is To ⧺ Cc and
never Bcc.

🚨 **A sender that cannot carry copies REFUSES, naming the counts — it never delivers a quietly
smaller mail.** Delivering to fewer people than the sender addressed while reporting success is the
same class of defect as stamping an undelivered mail `Sent` (#2023). The one exception is
`NoOpEmailSender`: it delivers nothing to anybody and says so, so it accepts the full envelope and
logs every count — and still refuses outright on a mail-enabled-but-unconfigured install.

---

## Purely additive — and why that is the whole design

Both members are **default interface methods whose defaults point in opposite directions and never
recurse**:

- the interface's `ObserveSendAsCapability` folds `CanSendAsUser`;
- the hub extension `CanSendAsUser` folds `ObserveSendAsCapability`.

So an **old sender** (overriding only `CanSendAsUser` — the shape of every implementation in a repo
pinning an older core) compiles untouched and answers exactly what it answered before, in two states.
A **new sender** overriding only the three-state member is folded correctly at every consumer.

This is what makes **shape 8** (an added interface member breaking external *implementers*)
inapplicable: a DIM has a body, so nothing is stranded. Every implementer found is listed in the
report. `CanSendAsUser` is kept as the two-state shim and is deliberately **not** `[Obsolete]`, for
the reason spelled out on `IEaGraphAuth`'s retiring surface — MeshWeaver.Plugins builds
`-warnaserror` against a core checkout at a pinned ref, and the attribute would red that repo before
its own half could land.

`CS0419` (shape 3) checked in **both** repos: no unqualified `<see cref>` to any member that gained
an overload. One *was* caught locally, inside `Mesh.Contract`, and fixed.

## Falsification — both arms, with numbers

Real monolith mesh (`MonolithMeshTestBase`), no mocking of core interfaces, no `Task.Delay`, no
`.ToTask()`.

**Arm 1 (#3450)** — restore the old `.Catch(… => Unavailable)` swallow:

```
Failed: 2, Passed: 11, Total: 13
  ✗ AProbeThatFaults_IsUndetermined_AndTheDialogDoesNotClaimNotConnected
      Expected value to be true because a probe that faulted produced NO answer
      about this user's mailbox, but found False.
  ✗ AProbeThatNeverAnswers_IsUndetermined
  ✓ AnUndeterminedProbe_StillFoldsToFalse_SoTheSendPathAsksWhichMailbox  (×3)  ← passes BOTH ways
```

**Arm 2 (#3473)** — make the default silently forward the first address and drop the rest:

```
Failed: 2, Passed: 7, Total: 9
  ✗ ASingleAddressSender_RefusesCopiedRecipients_RatherThanDroppingThem
      Expected value to be OnError because a Cc that cannot be honoured is a
      refusal, not a smaller success, but found OnNext.
  ✗ AMessageWithNoRecipients_IsRefused  (IndexOutOfRange — the drop shape cannot even express it)
  ✓ AMessageWithCcAndBcc_IsDeliveredAsOneMessage                              ← positive control
```

**Restored:** 22 new tests green; **54/54** across every email suite in the project
(`EmailSendAsCapabilityTest`, `EmailCopiedRecipientsTest`, `NoOpEmailSenderRefusalTest`,
`EmailMisconfigurationRefusalTest`, `InvitationEmailTest`, `OutboundSendQueueTest`) — the last four
pre-existing, proving no regression to the senders already shipping.

The Cc/Bcc numbers, stated as counts: **1** send carrying **4** recipients, **3** mutually visible,
**1** blind — against **4** sends carrying **1** each, **0** able to see anyone, which
`TheOldPerRecipientShape_SplitsOneMailIntoFour` measures as the control.

## Builds (`-c Release -warnaserror`, one project per invocation)

`MeshWeaver.Mesh.Contract` · `MeshWeaver.Messaging.Hub` · `MeshWeaver.Graph` ·
`Memex.Portal.Shared` · `Memex.Portal.Shared.Test` · `MeshWeaver.Documentation.Test` ·
`MeshWeaver.Messaging.Hub.Test` — all `0 Warning(s), 0 Error(s)`.

Guards run: `DocumentationLinkIntegrityTest` ✓ · `HubReachableAsyncGuard` ✓ (nothing added returns
`Task`/`ValueTask`) · `LocalizationTest` 58/58 ✓.

## 🚧 What is NOT in this PR, and why

**The user-visible half of both issues lives in `MeshWeaver.Plugins`**, which builds `-warnaserror`
against a core checkout at `MW_PLATFORM_REF` (`bbcb22f2`). It cannot compile until that pin carries
these members, and this session is fenced from moving any platform pin. The order is therefore
**core → pin → Plugins** — exactly the scoping in
https://github.com/Systemorph/MeshWeaver/issues/3450#issuecomment (maintainer, on the issue).

Because the change is additive, landing core alone strands nothing: no signature moved, no
implementer broke, and every consumer keeps its current behaviour until it opts in.

`SendingEmail.md` names the remaining diff **file by file** (`GraphEmailSender`,
`SendDocumentLayoutArea`, `SendDocumentDispatch`) so the next session can execute it without
re-deriving the design. The two localization keys the third panel needs
(`ui.sendDocument.connectUnknown`, `ui.sendDocument.checkAgain`) ship **here**, in `en` and `de`, so
the consumer half adds no catalog entries.

🚨 **i18n mirror handover** — this PR adds 2 catalog keys. After merge:
`npm run sync:i18n -- --ref <merged core sha>` in MeshWeaver.Plugins. The mirror guard compares
against a pinned core commit, so these keys red nothing and leave it silently stale until that runs.

## Docs

`Doc/Architecture/SendingEmail` gains two sections — the three-answer probe (with the table above and
the send-path exception) and Cc/Bcc as one message — plus the remaining-work table. Two What's New
entries, `Category: Fix`.

Refs #3450 · Refs #3473 — deliberately **not** `Closes`: each issue's user-visible half is the
Plugins consumer. Both stay open until that lands.

### DIFF
--- a/memex/Memex.Portal.Shared/Email/NoOpEmailSender.cs
+++ b/memex/Memex.Portal.Shared/Email/NoOpEmailSender.cs
@@ -41,7 +41,35 @@ public sealed class NoOpEmailSender(EmailOptions options, ILogger<NoOpEmailSende
 
     public IObservable<bool> SendEmail(
         string toAddress, string subject, string htmlBody, IReadOnlyCollection<EmailAttachment> attachments)
+        => SendEmail(new EmailMessage
+        {
+            To = [toAddress],
+            Subject = subject,
+            HtmlBody = htmlBody,
+            Attachments = attachments,
+        });
+
+    /// <summary>
+    /// 🚨 Overridden so a message with <b>Cc/Bcc</b> is handled by this sender rather than refused
+    /// by the interface default (#3473).
+    ///
+    /// <para>The default refusal exists to stop a single-address transport from quietly delivering
+    /// a DIFFERENT message than the one it was handed. That reasoning does not apply here: this
+    /// sender delivers nothing at all, to anybody, and says so. Refusing copied recipients would
+    /// break every local-dev and test send that carries one, while protecting a delivery that is
+    /// not happening either way.</para>
+    ///
+    /// <para>What it does still do is name the FULL envelope in its log line — To, Cc and Bcc
+    /// counts — so a developer reading the log sees the message that would have gone out, copies
+    /// included, instead of a line that silently describes fewer recipients than were asked for.
+    /// The <c>Email:Enabled=true</c> refusal (#2023) is unchanged and applies to the whole
+    /// envelope.</para>
+    /// </summary>
+    /// <param name="message">The message, its recipients, its attachments and its sender identity.</param>
+    public IObservable<bool> SendEmail(EmailMessage message)
     {
+        var recipients = string.Join(", ", message.AllRecipients);
+
         if (options.Enabled)
         {
             // ExplainRefusal, not Explain: this install may have the module and be missing only a
@@ -50,14 +78,18 @@ public sealed class NoOpEmailSender(EmailOptions options, ILogger<NoOpEmailSende
             var explanation = EmailDeliveryGuard.ExplainRefusal(
                 options, "This send is REFUSED rather than reported as delivered.");
             logger?.LogError(
-                "Refusing to send to {To} (subject: {Subject}, attachments: {Attachments}). {Explanation}",
-                toAddress, subject, attachments.Count, explanation);
+                "Refusing to send to {To} (subject: {Subject}, to: {ToCount}, cc: {CcCount}, "
+                + "bcc: {BccCount}, attachments: {Attachments}). {Explanation}",
+                recipients, message.Subject, message.To.Length, message.Cc.Length,
+                message.Bcc.Length, message.Attachments.Count, explanation);
             return Observable.Throw<bool>(new InvalidOperationException(explanation));
         }
 
         logger?.LogInformation(
-            "Email disabled (Email:Enabled=false) — skipping send to {To} (subject: {Subject}, attachments: {Attachments})",
-            toAddress, subject, attachments.Count);
+            "Email disabled (Email:Enabled=false) — skipping send to {To} (subject: {Subject}, "
+            + "to: {ToCount}, cc: {CcCount}, bcc: {BccCount}, attachments: {Attachments})",
+            recipients, message.Subject, message.To.Length, message.Cc.Length,
+            message.Bcc.Length, message.Attachments.Count);
         return Observable.Return(true);
     }
 }
--- a/src/MeshWeaver.Documentation/Data/Architecture/SendingEmail.md
+++ b/src/MeshWeaver.Documentation/Data/Architecture/SendingEmail.md
@@ -223,11 +223,125 @@ Two identities exist, and the recipient can tell them apart — so the choice is
 The delegated scope needed is `Mail.Send`, and it is **already part of `EaGraphAuth.Scopes`** —
 connecting the personal assistant grants it, so there is no separate consent step for sending.
 
-- **Probe before composing**: `hub.CanSendAsUser(objectId)`, so the UI can STATE the identity rather
-  than the user discovering it in the recipient's inbox.
+- **Probe before composing**: `hub.ObserveSendAsCapability(objectId)`, so the UI can STATE the
+  identity rather than the user discovering it in the recipient's inbox.
 - **Never fall back silently.** If the user is not connected, offer `/auth/ea/connect`; the shared
   mailbox is only ever an explicitly chosen second option — and then set
   `EmailDelivery.AsSharedMailboxReplyingTo(userEmail)` so a reply still reaches the human.
+
+### 🚨 The probe has THREE answers, and only one of them may say "you have not connected"
+
+`IEmailSender.CanSendAsUser` returns `IObservable<bool>` — two answers for three states of the
+world. Both consumers wrapped it in `.Catch(_ => Observable.Return(false))`, so a probe that timed
+out or faulted became the same `false` as a probe that completed and found no credential. The send
+dialog renders `false` at open time as *"Your Microsoft 365 mailbox is not connected yet"* beside a
+Connect button — a sentence that is simply untrue on a transient fault, shown to a user who **had**
+connected. Nothing was logged, so the third state was not even greppable. That is
+[#3450](https://github.com/Systemorph/MeshWeaver/issues/3450), and it is
+[#3433](https://github.com/Systemorph/MeshWeaver/issues/3433) one layer down.
+
+`IEmailSender.ObserveSendAsCapability(userObjectId)` answers with `EmailSendAsCapability`:
+
+| `EmailSendAs` | Means | What the UI may say |
+|---|---|---|
+| `Available` | The check completed; this sender can act as the person | Name the address the mail leaves from |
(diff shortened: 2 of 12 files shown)


rbuergi:
**Third falsification arm** (added in 73fb0e5c3, a self-review catch on the first commit).

`ObserveSendAsCapability` carried a `.Take(1)` that nothing needed and that broke two things:

- the send dialog `CombineLatest`s this stream into the form it renders, so it **is** a live
  data-bound view — and `.Take(1)` on one of those freezes the binding (AGENTS.md, the GUI rule);
- it is the mechanism the **Check again** affordance needs: a sender that re-probes must be able to
  replace an `Undetermined` answer with a real one *in place*, without the dialog being torn down.

Neither downstream operator required a single emission — `Catch` replaces only the tail after a
fault, `DefaultIfEmpty` fires only on an empty completion — and the send-time caller already takes
its own `.Take(1)` in `SubmitSend`.

Pinned by `TheProbeStaysLive_SoARetryCanReplaceAnUndeterminedAnswerInPlace`. Falsified by re-adding
`.Take(1)`:

```
Failed: 1, Passed: 0, Total: 1
  Expected collection to have 2 item(s) because truncating the probe to one answer
  freezes the dialog's binding and makes 'Check again' unable to replace the panel
  it is offered on, but found 1.
```

Restored: **55/55** green across every email suite in the project.

rbuergi:
**Re the Copilot review — the finding is correct, and here is the measurement plus why I am not
ejecting the queue for it.**

Confirmed: `date -u` reads `2026-09-06 22:46Z` while local reads `2026-09-07 00:46 +0200`. I named
both What's New entries from the harness's *local* date, so they are one day ahead of UTC — and this
PR will merge on **2026-09-06 UTC**. The theme page says an entry is *"produced automatically when a
pull request merges"*, so the merge date is the right anchor and `2026-09-06` is what these should
have been.

Precedent measured, both directions:

| entry | committed (UTC) |
|---|---|
| `2026-09-06-*.md` | `2026-09-06T21:15Z` — same day |
| `2026-09-05-*.md` | `2026-09-05T21:15Z` — same day |
| `2026-09-02-*.md` | `2026-09-03T10:41Z` — dated *before* its commit |

So entries have lagged their commit before; none has *led* it. Mine would be the first future-dated
pair, which is the direction worth avoiding.

**Not fixing it here, deliberately.** The PR is already in the merge queue (`added_to_merge_queue`
22:43:48Z), and a push to a queued branch ejects it. Ejecting a green, queued PR — and burning a
second full matrix run — to rename two release-note files is poor value against the standing
directive that every CI run is paid for. It is cosmetic: no link resolves to these filenames
(`DocumentationLinkIntegrityTest` is green), and `Order:` only affects sort position within What's
New.

**The rename rides the next change that touches this area** — the Plugins-half follow-up for #3450 /
#3473, which needs a `MW_PLATFORM_REF` bump and will touch these docs anyway. Recording it here so it
is not lost: rename both `2026-09-07-*.md` → `2026-09-06-*.md` and set `Order: -20260906`.

No other Copilot finding to address — the review raised this one point and no inline comments.

rbuergi:
## Copilot's one finding, weighed and declined — with the reasoning, since it is not obviously wrong

> The two new What's New entries are future-dated (2026-09-07) relative to the PR timestamp (2026-09-06)

**Half right, and the half that is right does not justify the fix.**

- Right: every other entry in `Data/WhatsNew/` is dated `2026-09-06`, and UTC at the time of writing is `2026-09-06 22:47Z`. So `2026-09-07-…` is ahead of the tree and ahead of UTC by 73 minutes. The date came from local time (CEST, `2026-09-07 00:47`).
- Not right enough to act on: the only way to rename the files is a **push to this branch, and a push to a queued branch EJECTS it** — `EVERY queue mutation restarts every build`. Paying two full queue builds (this PR's and every entry behind it) to move a release-note filename by one day, on a date that becomes correct in 73 minutes, is not a trade this repo's CI budget should make. The maintainer's directive is explicit that CI is paid for.

No gate objects — the run is green on this shape, and #3487 carries a `2026-09-07-…` entry through the same lane.

**What it does leave behind, and is worth fixing once rather than twice:** *nothing decides which clock names a What's New file.* Both spellings are legal today, which is why two PRs on the same night disagree. That is a one-line convention in `AuthoringDocumentation.md` plus a check, not a rebase of two queued PRs — filed separately rather than smuggled in here.

PR #3494, opened 2026-09-06
docs(#3413): a layout area that never renders is silent in every sink — collect the message trace in the shard lane

### LINKED ISSUE #3413: Flake or defect: EditorTest.TestEditorWithoutResult times out at exactly its 10s budget having emitted NOTHING — on a branch whose diff cannot reach its closure
## `EditorTest.TestEditorWithoutResult` timed out on a branch that cannot reach it

Failed on [PR #3410](https://github.com/Systemorph/MeshWeaver/pull/3410)'s run, shard 5 — **one failure in 450**:

```
MeshWeaver.Layout.Test.EditorTest.TestEditorWithoutResult
  the observable emitted nothing at all
```

after **exactly its 10 s budget**. It is the **first** test of the class; the sibling case passed ~100 ms later.

## Why this is attributed to `main`, not to that PR

`MeshWeaver.Layout.Test`'s closure is Data · Fixture · Layout · Mesh.Contract · Messaging.Hub · Reactive.Assertions, plus two linked PandasExplorer sources. `git diff --name-only origin/main...HEAD` on that branch matches **none** of them — `MeshWeaver.Hosting` (all of the PR's code) is not in the closure, and its two markdown files are not linked sources.

**For this test, that branch is byte-for-byte `main`.** So the failure belongs to `main`, and the PR is the messenger.

## What was deliberately NOT done

- **Not re-run.** House rule: never re-run a test to see whether it was a flake — flakes here are real races, and re-running hides them.
- **Not catalogued in `.github/known-flakes.json`.** That catalogue has zero entries and its header records that single occurrences did not meet the bar; an entry also requires a root-cause issue, which is what this is.
- **Not re-queued**, and no steward action taken.

## Not reproduced locally — which narrows it rather than clearing it

Run on the identical closure under `DOTNET_PROCESSOR_COUNT=4` (the setting that reproduces CI's scheduling pressure): **Failed: 0, Passed: 4**.

Combined with "first test of the class, timed out at exactly the budget, sibling passed 100 ms later", the shape is a **load-sensitive first-render/first-subscription race** — something that is not yet subscribed when the assertion begins to wait, and which a warm process does not exhibit.

🚨 That shape has precedent in this repo and it is NOT usually a test bug: an assertion that fires before the thing it tests, a wait that re-derives its stream and opens a second subscription, or a cold observable nobody subscribed. Worth checking those three before concluding the test is merely slow. **Raising the 10 s budget would be a band-aid** — the observable emitted *nothing at all*, so the question is why the emission never happens, not whether it happens late.

## Evidence

- PR #3410 run, shard 5 (`MeshWeaver.Layout.Test`), 1 failed / 450
- Shards 0–4 green; `MeshWeaver.Hosting.Test` shard: `Total: 360, Errors: 0, Failed: 0`
- Local control: identical closure, `DOTNET_PROCESSOR_COUNT=4`, 0 failed / 4 passed

Surfaced while fixing #3404; filed rather than worked around, and it currently blocks that PR from merging.

🤖 Filed from Claude Code.

rbuergi:
## Why

Investigating #3413 (`EditorTest.TestEditorWithoutResult` timed out at exactly its 10 s budget
having emitted nothing) I could not establish the cause — but I could establish **why the failing
run's artifacts were unable to say anything**, and that turned out to be the #890 defect shape a
second time.

`LayoutAreaHost` carries two diagnostics written for exactly this wedge:

- `COMPLETED WITHOUT RENDERING` (`LogError`)
- `was torn down having never rendered` (`LogWarning`)

Both pass **no exception object**, so the trace sink's documented gate —
`exception is not null && logLevel >= LogLevel.Warning`, in both `XUnitFileOutputHelper.Log` and
`LoggingBuilderExtensions` — refuses them; and both fire on a pool thread or during teardown, where
`outputHelper` is null / `IsInTestMethod()` is false and the trx sink is closed too. In a
`HubTestBase` test the disposal one is `LogDebug` regardless (the `Hub.IsDisposing` branch, #2679),
and mesh teardown is precisely how such a test disposes its layout host.

**Measured** on the #3413 occurrence (run
[`34034828870`](https://github.com/Systemorph/MeshWeaver/actions/runs/34034828870), shard 5): across
all **450** trx results and the whole **4,050-line** trace, `never rendered` → **0** and
`COMPLETED WITHOUT` → **0**. (`grep FAULT-BUDGET` → 0, so nothing was suppressed — the silence is
real.) A layout area that renders nothing is silent in *every* sink. That is what made #1081 cost
four sessions, and it is what leaves #3413 open.

## What this changes

**1. `dotnet-test.yml` collects the per-message trace.** `MessageTrace`
(`MeshWeaver.Messaging.Hub/MessageService.cs`) writes `ROUTED` / `DEFERRED gates=[…]` /
`GATE_FAILED` / `DROPPED_GATE_STUCK` per delivery to `$TMPDIR/meshweaver-msg-trace.log` when
`MESHWEAVER_MSG_TRACE=1`. That answers the exact three-way question this failure class poses — was
`InitializeHubRequest` processed, was an `UpdateStreamRequest` ever posted, was anything deferred.

`node-repo-module-pack.yml` already copies that file into its artifact (two places);
`dotnet-test.yml` collected only `_meshweaver-test-trace.log` and `_meshweaver-memory-delta.log`.
This closes that asymmetry, so turning the trace on for a run is a **one-variable** change instead
of a variable *plus* a workflow edit. The variable stays OFF by default (a lock + file append per
message), so the copy is a **no-op on every ordinary run**.

**2. `ReadingCiSignals.md` records the second instance.** That page already carries the #890 lesson
— *"for a fault the trace log is the sink, and reaching it is a property of the CALL, not of the
level or the wording"* — with a "know which questions the sinks cannot answer" section. This adds
the `LayoutAreaHost` instance beside it with the measurement, plus where the message trace is now
collected, so the next session checks rather than rediscovers.

## What this deliberately does NOT do

- **It does not fix #3413** — hence `Refs`, not `Closes`. The cause is still not established; the
  exclusions and the refutation of the standing hypothesis are recorded in
  [this comment](https://github.com/Systemorph/MeshWeaver/issues/3413#issuecomment-5562611015).
- **It does not touch the render path.** Making `LayoutAreaHost`'s diagnostics reach the sink (and
  say *which stage* the pipeline reached) is the right follow-up, but it changes the render path of
  every area in the platform and deserves its own issue and review — not a drive-by on a flake
  investigation.
- No test was re-run, no budget widened, no retry/delay/spin added, no assertion relaxed, nothing
  added to `.github/known-flakes.json` (still `"entries": []`).

## Verification

| check | result |
|---|---|
| `dotnet build test/MeshWeaver.Documentation.Test -c Release -warnaserror` | **`0 Warning(s)`, `0 Error(s)`**, 14.5 s (real compile, not a no-op) |
| `DocumentationLinkIntegrityTest` | **`Passed! - Failed: 0, Passed: 1`**, `.trx` written |
| `check-workflow-timeouts.py` | 68 jobs checked, **0 violations** |
| `check-workflow-shell.py` | 0 live findings |
| `check-shared-rules.py` | exit 0 |
| YAML parse of `dotnet-test.yml` | OK, 12 jobs |

No `public` type or member is removed from `src/` — `Pairs-with: none — the diff touches only a
workflow and a documentation page; no compiled surface changes.`

### DIFF
--- a/.github/workflows/dotnet-test.yml
+++ b/.github/workflows/dotnet-test.yml
@@ -1388,8 +1388,21 @@ jobs:
         # outside the test-method window (so the trx never sees them either). This trace answers
         # the question those cannot: was InitializeHubRequest processed, was an UpdateStreamRequest
         # ever posted, was anything deferred. See Doc/Architecture/ReadingCiSignals.
-        if [ -f /tmp/meshweaver-msg-trace.log ]; then
-          cp /tmp/meshweaver-msg-trace.log collected-logs/_meshweaver-msg-trace.log
+        # 🚨 Derive the directory the SAME way the writer does, never a literal. MessageTrace
+        # writes to `Path.Combine(Path.GetTempPath(), "meshweaver-msg-trace.log")`
+        # (MessageService.cs:27, and identically in OrleansRoutingService and RoutingGrain), and on
+        # Unix `GetTempPath()` returns $TMPDIR when it is set and /tmp only when it is not. A
+        # hard-coded /tmp therefore works by the coincidence that GitHub's ubuntu images leave
+        # TMPDIR unset — and the day anything sets it (a container action, a setup step, a
+        # self-hosted runner) this `-f` test answers false, the copy is skipped, and the step stays
+        # GREEN having collected nothing. That is precisely the failure class this whole step exists
+        # to make visible, so it must not be the shape of the collector itself.
+        msg_trace="${TMPDIR:-/tmp}/meshweaver-msg-trace.log"
+        if [ -f "$msg_trace" ]; then
+          cp "$msg_trace" collected-logs/_meshweaver-msg-trace.log
+          echo "collected the per-message trace from $msg_trace"
+        else
+          echo "no per-message trace at $msg_trace (MESHWEAVER_MSG_TRACE is off unless set for the run)"
         fi
         # Native-crash mini dumps (DOTNET_DbgEnableMiniDump) — present only when a
         # test host died on a signal; the exit-marker gate reds the shard, this is
--- a/.github/workflows/dotnet-test.yml
+++ b/.github/workflows/dotnet-test.yml
@@ -1374,6 +1374,23 @@ jobs:
         if [ -f /tmp/meshweaver-memory-delta.log ]; then
           cp /tmp/meshweaver-memory-delta.log collected-logs/_meshweaver-memory-delta.log
         fi
+        # Per-MESSAGE trace (MessageTrace, MeshWeaver.Messaging.Hub/MessageService.cs): the
+        # ROUTED / DEFERRED gates=[…] / GATE_FAILED / DROPPED_GATE_STUCK line for every delivery.
+        # OFF unless MESHWEAVER_MSG_TRACE=1 is set for the run, so this copy is a no-op on an
+        # ordinary shard and costs nothing — but when a wedge needs it, flipping ONE env var is
+        # then the whole job, instead of also having to change this workflow first.
+        #
+        # 🚨 It is the instrument the "observable emitted nothing at all" class actually needs, and
+        # it was reachable in node-repo-module-pack.yml and NOT here (#3413): a layout area that
+        # never renders is silent in every other sink, because LayoutAreaHost's two
+        # "delivered nothing" diagnostics carry no exception (so the fault sink's
+        # `exception is not null && level >= Warning` gate refuses them — the #890 shape) and fire
+        # outside the test-method window (so the trx never sees them either). This trace answers
+        # the question those cannot: was InitializeHubRequest processed, was an UpdateStreamRequest
+        # ever posted, was anything deferred. See Doc/Architecture/ReadingCiSignals.
+        if [ -f /tmp/meshweaver-msg-trace.log ]; then
+          cp /tmp/meshweaver-msg-trace.log collected-logs/_meshweaver-msg-trace.log
+        fi
         # Native-crash mini dumps (DOTNET_DbgEnableMiniDump) — present only when a
         # test host died on a signal; the exit-marker gate reds the shard, this is
         # the forensic payload.
--- a/src/MeshWeaver.Documentation/Data/Architecture/ReadingCiSignals.md
+++ b/src/MeshWeaver.Documentation/Data/Architecture/ReadingCiSignals.md
@@ -522,11 +522,21 @@ cost four sessions and what leaves #3413 open.
 **The instrument that does answer it** is `MessageTrace`
 (`MeshWeaver.Messaging.Hub/MessageService.cs`): `MESHWEAVER_MSG_TRACE=1` makes every delivery write
 a `ROUTED` / `DEFERRED gates=[…]` / `GATE_FAILED` / `DROPPED_GATE_STUCK` line to
-`$TMPDIR/meshweaver-msg-trace.log` — enough to say whether the sync sub-hub processed its
+`Path.Combine(Path.GetTempPath(), "meshweaver-msg-trace.log")` — enough to say whether the sync
+sub-hub processed its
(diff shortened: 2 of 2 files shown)


rbuergi:
🚦 **Held as DRAFT deliberately — not because it is unfinished.**

Everything is verified locally (see the body's verification table: `0 Error(s)` `-warnaserror`, `DocumentationLinkIntegrityTest` `Passed! - Failed: 0, Passed: 1`, all three workflow guards clean).

The hold is the shared-action rule: at the time this was opened, main had a **`Continuous Delivery (main)` run in progress** on `ddb95e6` (with a cancelled one on the same sha minutes earlier), and a **release tag is pending** on the sealed set the five satellites just landed on (`3.0.0-ci.7917`). Merging would add a commit to main and start a competing CD run while someone may be gating on that one. Draft is the documented opt-out from `auto-arm.yml`, so this cannot enqueue itself in the meantime.

**To release the hold:** `gh pr ready 3494` once the CD run in flight has landed and the release tag is cut. Nothing else is needed — the branch is behind main only in the harmless sense (`strict: false`; the one required context is `Consolidate test results`).

Note this PR does **not** fix #3413 and must not be read as closing it — it only removes the reason the next sighting would be as uninformative as this one.

rbuergi:
## Both findings accepted and fixed — the first one is the shape this PR is about

Neither was argued away; both are in the branch (`da46bd3dc`, `d00bb080b`).

### 1 · The hard-coded `/tmp` — you are right, and it matters more than it looks

The writer is `Path.Combine(Path.GetTempPath(), "meshweaver-msg-trace.log")` — `MessageService.cs:27`, and identically in `OrleansRoutingService.cs:37` and `RoutingGrain.cs:28`. On Unix `GetTempPath()` returns `$TMPDIR` **when that is set** and `/tmp` only when it is not.

So the literal worked by the coincidence that GitHub's ubuntu images leave `TMPDIR` unset. The day anything sets it — a container action, a setup step, a self-hosted runner — the `-f` test answers false, the copy is skipped, and **the step stays GREEN having collected nothing.**

🚨 That is precisely the failure class this step exists to expose: an instrument that is silent when it should speak (#3413 / the #890 shape). It must not be the shape of the collector itself. Now:

```bash
msg_trace="${TMPDIR:-/tmp}/meshweaver-msg-trace.log"
if [ -f "$msg_trace" ]; then
  cp "$msg_trace" collected-logs/_meshweaver-msg-trace.log
  echo "collected the per-message trace from $msg_trace"
else
  echo "no per-message trace at $msg_trace (MESHWEAVER_MSG_TRACE is off unless set for the run)"
fi
```

Both branches now name the path they examined, so *"nothing was collected"* and *"nothing was there"* stop reading the same in the log — which was the second half of the same defect.

### 2 · `$TMPDIR` in the doc — also right, and it would have propagated the confusion

`ReadingCiSignals` now says `Path.Combine(Path.GetTempPath(), "meshweaver-msg-trace.log")` rather than `$TMPDIR/…`, plus a short rule stating why they are not interchangeable (`%TEMP%` on Windows) and why the collector derives rather than hard-codes. A page that names an env var teaches a contract that does not exist.

### Verification

`actionlint` clean · `check-workflow-shell.py` 0 live findings · `check-workflow-timeouts.py` 68 jobs, 0 violations · YAML parses · `MeshWeaver.Documentation.Test` Release `-warnaserror` **0 Warning(s), 0 Error(s)** · `DocumentationLinkIntegrityTest` **1/1 passed** with a real `.trx`.

*Note on process: this PR was already in the merge queue when the review landed, so the push was refused (`protected branch hook declined`) until it was dequeued. I checked first that it was the only live queue entry — the other `gh-readonly-queue` refs are stale leftovers — so the dequeue restarted no sibling build.*

Issue #3498, opened 2026-09-06
A re-ask's NotFound is dropped as 'resubscribe failed' instead of terminating the stream — a reader of a deleted node waits out its budget (CD 7946, #1029's other door)


rbuergi:
## Measured: CD 7946 (`ddb95e65f`, the first candidate set carrying #3480) died before its seal

`Plugins: pack the module bundles the bake composes / Module tests (MeshWeaver.AI)` — 1 of 1868 failed, `MeshPluginTest.FullCrudWorkflow_CreateGetUpdateDelete` (Plugins at `ea9aa9934`), the first test of its class, 29 s into the run:

```
[xUnit.net 00:00:29.54]     MeshWeaver.AI.Test.MeshPluginTest.FullCrudWorkflow_CreateGetUpdateDelete [FAIL]
  Failed MeshWeaver.AI.Test.MeshPluginTest.FullCrudWorkflow_CreateGetUpdateDelete [21 s]
  Error Message:
   System.TimeoutException : Get(@ACME/CrudTest_5b2574eeaee942e0954d0afe7c01afc8) never reported the node gone within 20s
  Stack Trace:
     at MeshWeaver.AI.Test.MeshPluginTest.FullCrudWorkflow_CreateGetUpdateDelete() in /home/runner/work/MeshWeaver/MeshWeaver/src/MeshWeaver.AI.Test/MeshPluginTest.cs:line 630
--- End of stack trace from previous location ---
  Standard Output Messages:
 [22:50:51.795] === TEST CLASS INIT: MeshPluginTest ===
 [22:50:52.513] === TEST START: MeshPluginTest.FullCrudWorkflow_CreateGetUpdateDelete ===
 [22:50:52.586] [Warning] [MeshWeaver.Messaging.Serialization.ObjectPolymorphicConverter] Received '$type':'Space' which is NOT registered in this (receiving) hub — it can only be read as an untyped JsonElement (the value renders empty / reactive waits time out). Register it in this hub too (WithType
 [22:50:52.811] [Warning] [MeshWeaver.Hosting.MeshNodeStreamCache] MeshNodeStreamCache.GetStream: Content for ACME/CrudTest_5b2574eeaee942e0954d0afe7c01afc8 stayed an untyped JsonElement after deserialization (TypeRegistry lacks the $type discriminator) — downstream 'Content is X'/'as X' consumers wi
[...]

Issue #3477, opened 2026-09-06
LateNackReenqueueTest: after LATE_NACK_REENQUEUE nothing lands or answers within 45 s (1/330, Plugins shard 3, first run on 3.0.0-ci.7917; 28/28 locally on both pins)


rbuergi:
## Observed

MeshWeaver.Plugins main run 34059060596 (`ef51fd505`, the first run on platform set `3.0.0-ci.7917` / core `1b5350d54`), `Portal hosts (shard 3)`, migrated suite `src/MeshWeaver.Hosting.Monolith.Test`:

```
Failed MeshWeaver.Hosting.Monolith.Test.LateNackReenqueueTest.LateOwnerDisposingNack_AfterOptimisticEmit_ReenqueuesAndLands [45 s]
System.TimeoutException : The operation has timed out.
   at LateNackReenqueueTest.LateOwnerDisposingNack_AfterOptimisticEmit_ReenqueuesAndLands() in src/MeshWeaver.Hosting.Monolith.Test/LateNackReenqueueTest.cs:line 148

[write] patch posted with marker post-nack-4b7556aa598346; owner merge is parked
[dispose] owner per-node hub disposal invoked for TestData/late-nack-node
[20:56:25.860] [Warning] [MeshNodeStreamHandle] [UpdateRemote] LATE_NACK_REENQUEUE hub=cache/Z86EFbZ420G_0zLpfgUKSQ target=TestData/late-nack-node attempt=1 code=OwnerDisposing — the patch was never applied; re-enqueueing the original update, rebased on state newer than 0
[20:57:10.860] === TEST FAILED: The operation has timed out.
```

329 of 330 passed in the same assembly. Line 148 is the wait for the re-enqueued write to be **persisted** (`storage.Read(path)` reaching `Name == marker` within 45 s). So: the owner's late `OwnerDisposing` NACK fired, `UpdateRemote` re-enqueued the original update (attempt 1), and in the following 45 s **nothing landed and nothing logged** — no `LATE_ACK`, no second NACK, no `LATE_NACK_REENQUEUE failed`, no error to the caller. A write that was NACKed, re-enqueued and then neither applied nor answered is the write-loss class this test exists to refuse (#3197 family; #3363 is the sibling on the queued-write handoff).

## Controls — this is a race, not (yet) the pin

| where | core | result |
|---|---|---|
| Plugins main, last four runs before the pin move | `bbcb22f25` (ci.7859) | shard 3 `success` ×4 |
[...]

PR #3505, opened 2026-09-06
fix(#3499): a teardown fault is a RETRYABLE verdict, not a terminal one — an install survives its own root recycling

### LINKED ISSUE #3499: A package root is disposed mid-install under a reconcile — Chess, then Hosting, then Chess again; main-cd cannot seal
## A package's ROOT hub is disposed mid-install, and every write under it then dies on the owner bound

Three independent occurrences in six hours, on three different heads, hitting **two different packages**. It is not a package's content and it is not any one pull request.

| run | head | when | package that wedged |
|---|---|---|---|
| core `main-cd` **#7937** ([34058018897](https://github.com/Systemorph/MeshWeaver/actions/runs/34058018897)) | `85ab1dfbe` | 20:26Z | **Chess** — `GATE FAILED — install: Chess` |
| core `main-cd` **#7941** ([34061838257](https://github.com/Systemorph/MeshWeaver/actions/runs/34061838257)) | `b8accc2b2` | after #3475 | **Hosting** (144 files) — same shape; **Chess adopted 3/3 and passed** |
| Plugins #1422 gate ([34059652860](https://github.com/Systemorph/MeshWeaver.Plugins/actions/runs/34059652860), job `101564568189`) | — | 22:08Z | **Chess** again |

`#7938` in between **sealed fine**. So it is a race that lands on whichever package's root recycles under a reconcile at the wrong moment.

🚨 **#7937 predates the merge of #3475 and carries unstripped content — that is the control.** It rules out the operational-strip work as the cause; #1422 is a victim of this, not its origin.

## The sequence, from the Plugins gate shard

```
22:08:20-25  bake-seed: Prebuilt assembly ADOPTED for Chess/GambitHunt|Game|History at version 0
             — no compile needed; bundle Chess.zip: adopted 3/3          ← no Roslyn ran
22:08:29     [QUIESCE-TIMEOUT] Chess: 1 callback(s) still pending after 2s — forcibly cancelling.
             Pending: CreateOrUpdateNodeRequest…
             [PluginGating] reconcile failed for Chess: Hub Chess was disposed before the
             response arrived
~22:09:00    VERDICT_TIMEOUT / OwnerUnreachable at ~31s on Chess/History, GambitHunt, Game,
             _Policy, _Access/Anonymous_Access, GambitHunt/_Activity/compile-state
~22:09:21    MeshNode Unknown … TimeoutException (~20s)
22:09:38     idempotence: re-install failed: Install of 'Chess/_Policy' failed:
             Inner UpdateNode faulted: TimeoutException
             GATE FAILED — idempotence: Chess — 1 new failure(s)
```

Everything downstream is consequence: the root is gone, so every write under `Chess/**` waits out the owner bound. The types themselves were fine — `compile=Ok render=ok tests=ok` on all three, and `1/1 content asset(s) served`.

## The likely mechanism, and what would confirm it

The Chess root is a `Store/Plugin` instance (`PluginContent`), so this is **not** the installer's retyped-root recycle. Something disposes the root while `PluginGating`'s reconcile is in flight — the shape reads like a rebind recycle (a hub activated before its node was visible, then retyped by the install's root write). The recycler logs at Debug in that tester, so the log does not name it.

🚨 **The pending item is a `CreateOrUpdateNodeRequest` — correlated work — and there is a MEASURED mechanism that produces exactly one of those, unanswerable, at teardown.** From #3261 (`DanglingNodeTypeUpdateTest`, `MESHWEAVER_MSG_TRACE=1` + `MeshWeaver.Messaging: Debug`):

> A hub that has already logged `[QUIESCE-OK]` **still accepts new correlated requests**, and even creates hosted hubs for them — the intake gate only refuses from `DisposeHostedHubs` onward. A request taken on in that window can never be answered.

```
11.928 [QUIESCE-START] TestData/dnt…: 0 pending callbacks at dispose entry: <none>
11.928 [QUIESCE-OK]    TestData/dnt…: drained 0 callback(s) in 0ms
11.968 Start processing SubscribeRequest (ID: p2qsFux3…, cache/iIN4… -> TestData/dnt…)
11.969 Creating new hosted hub for address sync/4wzOafy… in host TestData/dnt…
```

**Stated as a fit, not a diagnosis** — that trace shows `[QUIESCE-OK]` then late intake, while this shows `[QUIESCE-TIMEOUT]` with something already pending. They may be two bugs. What the intake window *does* explain is how a request that can never be answered comes to be pending at all, which is the part the gate log does not show.

**The discriminator:** whether the `CreateOrUpdateNodeRequest` was accepted **before or after** `[QUIESCE-START]`. That needs `MeshWeaver.Messaging` at Debug in the tester image, alongside the recycler, so the intake timestamp can be read against the quiesce line.

## 🚨 Not the fix: widening the 2s quiesce budget

That is the explicit conclusion from #3261's measurement and it holds independently of whether the two shapes are one bug: a bigger budget converts a leaked callback into a slower leaked callback. The same goes for the 31s owner bound and the install timeout — those are reporting a root that is gone, not a bound that is too tight.

## Why it matters now

`main-cd` cannot produce a sealed set. The last one is **#7938 (`aa4075832`)**, which **predates #3475**; every post-merge run since has either failed the seal on this wedge or skipped it behind a module-test flake (#7946, `MeshWeaver.AI FullCrudWorkflow` timeout). So this race is currently the thing standing between a merged core fix and any deployable wave — Education's pin bump and every downstream instance wait on it.

## Acceptance

- A package root is not disposed while a reconcile addressed to it is in flight; or, if it must be, the requester is told (a NACK) rather than left to time out.
- The tester image logs the recycler and `MeshWeaver.Messaging` at a level that names *which* recycle fired, so the next occurrence is attributed in one read.
- No quiesce, owner or install bound is widened.

## Provenance

Gate-shard analysis and the three-run attribution by the session working MeshWeaver.Education; the #3261 intake-window measurement and the acceptance framing by this one. Filed jointly so neither half is lost.

rbuergi:
Closes #3499.

## The discriminator, answered: **BOTH — and the run that wedged the seal was BEFORE**

The issue asks whether the pending `CreateOrUpdateNodeRequest` was accepted before or after
`[QUIESCE-START]`. The request-fate trails in the three runs carry `RECEIVED runLevel=<level>@<hub>`
at the issuing hub, which answers it without any new instrumentation:

| run | hub | runLevel at intake |
|---|---|---|
| core #7937 | Cornerstone, Publish, RemoteControl, Voice | `Quiescing` |
| core #7937 | **Chess** — the one that failed the gate | **`Started`** |
| core #7941 | RolePlay | `Started` |
| core #7941 | Video, Essentials | `Quiescing` |
| Plugins #1422 | Chess | `Quiescing` |

So both shapes are live. **Six of nine are #3261's intake window** — and every one of them is
benign: the requester gets `HubDisposedBeforeResponseException`, which is a real answer, and eight
packages carrying that shape passed the gate in run #7937 alone. **The wedge is the other one.**

## What actually wedged the seal

Not the recycle, and not the pending callback. In **ten of ten** `VERDICT_TIMEOUT` trails on the
Plugins gate run (and, per #2543's own measurement, thirteen on core #7937), the patches the dying
subtree hubs owed acks for had **already committed**:

```
HANDLER_ENTER → PATCH_MERGE_DISPATCHED → HANDLER_EXIT Processed
→ PATCH_MERGE_TURN entered → PATCH_MERGE_STAMPED v=2 refused=0 → PATCH_ECHO_SEEN  ⇒ (nothing, 31s)
```

Only the durability leg was outstanding, and it faulted on a lifetime scope that was closing. That
fault reaches `ClassifyPatchException` — the ONE funnel every owner-side patch fault passes through —
and fell out of its `_ =>` arm as `MeshNodeErrorCode.Unknown`, which is **TERMINAL** by the enum's own
contract:

> `OwnerDisposing` and `OwnerNotReady` "are the only auto-retried NACK codes; every other code stays
> terminal."

So the writer reported a **lost write for a patch that had committed**, and the installer failed the
package. That is `GATE FAILED — install: Chess` (#7937), `install: Hosting` (#7941) and
`idempotence: Chess` (#1422).

**#2543 does not fix this** — and this matters, because it merged at 21:49Z, *after* all three runs,
so it looks like the answer. It catches the escaping throw so the writer is no longer left in 31 s of
silence; it then hands over a prompt `Unknown`. Same failed install, 31 seconds sooner. Its own test's
`because` reads *"a NACK the writer can retry on"* while the verdict it carried could not be retried;
this PR makes that assertion enforceable and true.

## The fix

The rule was already written down one layer over, in `HubDisposingException.IsDisposedContainer`:

> *"a hub whose scope has been closed cannot serve anything, so a delivery that faults this way must
> be answered as RETRYABLE (the address reactivates) rather than as a result."*

`ClassifyPatchException` now consults it **first**. Any `ObjectDisposedException` in the exception
**graph** (which covers both typed teardown exceptions — `HubDisposingException` and
`HubDisposedBeforeResponseException` both derive from it), or a `DeliveryFailure` carrying
`ErrorType.ShuttingDown`, answers `OwnerDisposing`. `MeshNodeStreamExtensions` then re-enqueues
against the fresh activation, capped at `MaxOwnerDisposingReenqueues` and re-diffing each attempt, so
a merge that did commit is a no-op and one that died with its hub lands.

The **graph**, not `InnerException` in a loop: a teardown at any aggregate index must still decide the
classification, or it is settled by arrival order.

## Against the acceptance criteria

- ✅ **"…or, if it must be, the requester is told (a NACK) rather than left to time out."** The root
  is still recycled — that is correct behaviour, and an install that cannot survive one is the defect.
  The requester is now told, *truthfully*, in the verdict its own machinery already acts on.
- ✅ **"The tester image logs the recycler and `MeshWeaver.Messaging` at a level that names *which*
  recycle fired."** `mw-plugin-test` ran at Warning while every candidate recycler announces itself at
  Information, which is why attributing this cost a day and still could not name the recycle. Three
  categories are raised: `MeshWeaver.Graph.Configuration` (rebind watcher, stale-build convergence,
  overlay self-heal), `MeshWeaver.PluginCatalog.PackageInstaller` (the retyped-root recycle and its
  declines), and `MeshWeaver.Messaging.MessageHub` for `[QUIESCE-START]` — whose pending-callback
  snapshot AT dispose entry is exactly the discriminator above, and is unreadable from the
  Warning-level `[QUIESCE-TIMEOUT]` alone. Cost stated at the call site: two lines per hub disposal
  plus a handful per package.
- ✅ **"No quiesce, owner or install bound is widened."** None moved. No watchdog, no poller, no
  `catch {}`, no sleep. The retry already existed and is already capped; only the classification
  changed.

## Not covered, and not described as covered

**#3261's intake window is a second bug and is not fixed here.** A hub past `[QUIESCE-OK]` still
accepts new correlated requests because `MessageService.ScheduleNotify`'s gate only refuses from
`DisposeHostedHubs` onward. Six of the nine trails above are that shape. It is benign in every
occurrence measured (the requester gets a typed answer), and the callback that wedged #7937 was
accepted while the hub was `Started`, so it is not this. Filed separately.

## Verification

Release, `-warnaserror`, one project per invocation — `MeshWeaver.Data`, `MeshWeaver.Data.Test`,
`MeshWeaver.PluginTester`, `MeshWeaver.Documentation.Test`: **0 Error(s), 0 Warning(s)**.
`MeshWeaver.Data.Test` **490/490**; `MeshWeaver.Documentation.Test` **304/304** including
`DocumentationLinkIntegrityTest`.

**Falsified.** With the classification reverted and the pins kept, all three fail:

```
Expected value to be OwnerDisposing because a disposal fault is the OWNER going away,
not a verdict about the patch …, but found Unknown.
```

## Why this merges promptly

`main-cd` has produced no sealed set since #7938 (`aa4075832`), so a merged core fix cannot become a
deployable wave. Per the session that owns the 3.0.0 tag: `main-cd`'s one-pending-slot eviction
cancels only the *pending* run, never the one in flight, and the tag's set is the first sealed set at
or after both this and #3498's merges — so holding this back is the only thing that can hurt the
release.

Doc: `WriteVerdictTotality.md` gains *"Totality is not enough — the verdict's CODE decides whether the
write survives"*. What's New (Fix).

Pairs-with: none — no public type or member is removed; the change is behavioural, plus one internal
predicate and one private static field.

### DIFF
--- a/src/MeshWeaver.Data/DataExtensions.cs
+++ b/src/MeshWeaver.Data/DataExtensions.cs
@@ -1181,8 +1181,24 @@ public static class DataExtensions
     /// disposed subject, a disposed CTS and the two typed teardown exceptions
     /// (<see cref="HubDisposingException"/>, <c>HubDisposedBeforeResponseException</c> — both derive
     /// from it) are all the same fact. And the direction of a mistake is asymmetric, exactly as
-    /// <c>MeshOperations.IsWriteDenial</c> argues for its own default: a false "retryable" costs at
-    /// most two idempotent re-diffs, while a false "terminal" loses a write and fails an install.</para>
+    /// <c>MeshOperations.IsWriteDenial</c> argues for its own default: a false "terminal" loses a
+    /// write and fails an install, every time; a false "retryable" costs an idempotent re-diff,
+    /// capped at two.</para>
+    ///
+    /// <para>🚨 <b>That asymmetry is real but the cheap side is NOT fully bounded, and this is the
+    /// honest statement of the cost.</b> The re-enqueue this classification feeds is
+    /// <c>MeshNodeStreamExtensions</c>'s <c>OwnerDisposing / OwnerNotReady / Conflict</c> arm, and
+    /// <b>#3477</b> is a measured, open, unreproduced case where that arm went DARK: after
+    /// <c>LATE_NACK_REENQUEUE … code=OwnerDisposing attempt=1</c>, nothing landed and nothing logged
+    /// for 45 s — no <c>LATE_ACK</c>, no second NACK, no error to the caller (1/330,
+    /// <c>LateNackReenqueueTest</c>, MeshWeaver.Plugins shard 3). Routing more faults here makes
+    /// that class MORE likely, not less, and it is silent when it happens. Traced end to end for
+    /// #3499: these faults take the SAME arm — and the closed-lifetime-scope shape takes its LATE
+    /// half preferentially, because a hosted hub's scope is closed on <c>DisposalCompleted</c>, by
+    /// which point <c>PostImplGeneric</c> refuses the owner's own post and <c>RoutePatchVerdict</c>
+    /// falls through to <c>ILatePatchVerdictSink</c> — which is exactly where #3477 was measured.
+    /// So the trade is <i>rarely and recoverably</i> against <i>always and fatally</i>, which is
+    /// still the right way round; it is not "at most two re-diffs".</para>
     ///
     /// <para>Everything else keeps its previous mapping; unknown exception types still fall through
     /// as <see cref="MeshNodeErrorCode.Unknown"/> with the exception type prefixed — visible at the
--- a/src/MeshWeaver.Data/DataExtensions.cs
+++ b/src/MeshWeaver.Data/DataExtensions.cs
@@ -1150,13 +1150,54 @@ public static class DataExtensions
     }
 
     /// <summary>
-    /// Maps owner-side patch exceptions to structured <see cref="MeshNodeError"/>
-    /// codes. Unknown exception types fall through as
-    /// <see cref="MeshNodeErrorCode.Unknown"/> with the exception type prefixed
-    /// — visible at the consumer GUI so the gap is diagnosable, not silent.
+    /// 🚨 TEARDOWN IS NOT A VERDICT ABOUT THE WRITE — issue #3499, and the reason a package
+    /// install could not survive its own root recycling.
+    ///
+    /// <para>An owner-side patch fault raised because something the answer needed had been
+    /// DISPOSED says exactly one thing: this activation went away. It says nothing about whether
+    /// the caller's patch is acceptable, and <see cref="HubDisposingException.IsDisposedContainer"/>
+    /// already writes the rule down — <i>"a hub whose scope has been closed cannot serve anything,
+    /// so a delivery that faults this way must be answered as RETRYABLE (the address reactivates)
+    /// rather than as a result."</i> This funnel — the ONE place every owner-side patch fault is
+    /// classified — did not consult it, so every such fault fell through to
+    /// <see cref="MeshNodeErrorCode.Unknown"/>, which is TERMINAL by the enum's own contract
+    /// (<see cref="MeshNodeErrorCode.OwnerDisposing"/> and <see cref="MeshNodeErrorCode.OwnerNotReady"/>
+    /// "are the only auto-retried NACK codes; every other code stays terminal").</para>
+    ///
+    /// <para><b>Measured.</b> CD 7937 / 7941 and the MeshWeaver.Plugins gate on 2026-09-06: a
+    /// package root recycles mid-install, its subtree hubs die owing acks for patches whose merge
+    /// had ALREADY committed (<c>PATCH_MERGE_STAMPED v=2 refused=0 → PATCH_ECHO_SEEN</c>, ten of
+    /// ten trails), and the writer was left either with 31 s of silence (fixed by #2543) or — once
+    /// that silence became a NACK — with a TERMINAL <c>Unknown</c> for a write it could simply
+    /// have re-applied. Both read at the gate as <c>GATE FAILED — install: Chess</c>. Classifying
+    /// the teardown correctly hands the writer the code its OWN machinery already acts on:
+    /// <c>MeshNodeStreamExtensions</c> re-enqueues an <c>OwnerDisposing</c> NACK against the fresh
+    /// activation, capped at <c>MaxOwnerDisposingReenqueues</c>, re-diffing each time — so a merge
+    /// that DID commit is a no-op and one that died with its hub lands. No bound moves.</para>
+    ///
+    /// <para>🚨 <b>Why every <see cref="ObjectDisposedException"/> in the graph counts</b>, not just
+    /// the Autofac-scope shape <see cref="HubDisposingException.IsDisposedContainer"/> matches.
+    /// Nothing on this path disposes anything as a way of REFUSING a patch: a disposed scope, a
+    /// disposed subject, a disposed CTS and the two typed teardown exceptions
+    /// (<see cref="HubDisposingException"/>, <c>HubDisposedBeforeResponseException</c> — both derive
+    /// from it) are all the same fact. And the direction of a mistake is asymmetric, exactly as
+    /// <c>MeshOperations.IsWriteDenial</c> argues for its own default: a false "retryable" costs at
+    /// most two idempotent re-diffs, while a false "terminal" loses a write and fails an install.</para>
+    ///
+    /// <para>Everything else keeps its previous mapping; unknown exception types still fall through
+    /// as <see cref="MeshNodeErrorCode.Unknown"/> with the exception type prefixed — visible at the
+    /// consumer GUI so the gap is diagnosable, not silent.</para>
     /// </summary>
     private static MeshNodeError ClassifyPatchException(Exception ex, string path)
     {
+        if (IsOwnerTeardown(ex))
+            return new MeshNodeError(
+                MeshNodeErrorCode.OwnerDisposing, path,
+                $"the owner went away while this patch was being answered ({ex.GetType().Name}: {ex.Message}) — "
+                + "the write is NOT confirmed; safe to retry against the fresh activation: a re-enqueue "
+                + "re-diffs against the freshest state, so a merge that did commit is a no-op",
+                ex.StackTrace);
+
         var (code, prefix) = ex switch
         {
             UnauthorizedAccessException => (MeshNodeErrorCode.AccessDenied, "Access denied"),
@@ -1168,6 +1209,26 @@ public static class DataExtensions
         };
         return new MeshNodeError(code, path, $"{prefix}: {ex.Message}", ex.StackTrace);
     }
+
+    /// <summary>
+    /// True when <paramref name="ex"/> is — or carries, at any depth and through any
+    /// <see cref="AggregateException"/> branch — the fact that the OWNER went away while this
+    /// patch was being answered. See the remarks on <see cref="ClassifyPatchException"/> for why
+    /// this is a retryable verdict and not a result.
+    ///
+    /// <para>The graph, never <c>InnerException</c> in a loop: a reactive ack path composes, so the
+    /// teardown can arrive at any index of an aggregate and a chain walker would classify by
+    /// arrival order (<see cref="ExceptionChain"/>'s own remarks).</para>
+    ///
+    /// <para><see cref="ErrorType.ShuttingDown"/> is the messaging layer's word for the same fact —
+    /// <c>MessageService.NackThroughParent</c> stamps it, <c>HubDisposingException</c>'s remarks
+    /// name it as the same contract, and <c>PackageInstaller</c> already retries on it — so a
+    /// delivery failure carrying it belongs here rather than in the <c>Unknown</c> bucket.</para>
+    /// </summary>
+    internal static bool IsOwnerTeardown(Exception? ex)
+        => ExceptionChain.Contains<ObjectDisposedException>(ex)
+           || ExceptionChain.Contains(ex, static e =>
+               e is DeliveryFailureException { Failure.ErrorType: ErrorType.ShuttingDown });
 
     /// <summary>
     /// 🚨 Totality for a one-shot watcher: runs <paramref name="onEmptyCompletion"/> when
--- a/src/MeshWeaver.Documentation/Data/Architecture/WriteVerdictTotality.md
+++ b/src/MeshWeaver.Documentation/Data/Architecture/WriteVerdictTotality.md
@@ -192,6 +192,94 @@ Missing both routes is then a checked fact (nobody in this mesh is armed, and th
 and is logged rather than swallowed, because the remaining case — a caller in another process — is
 one the registry cannot see.
 
+## Totality is not enough — the verdict's CODE decides whether the write survives
+
+A terminal that reaches the writer can still be the wrong terminal, and the difference is not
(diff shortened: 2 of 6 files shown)


rbuergi:
### Base rate added in `665a1d1` — the issue's stated cause is falsified by its own control

Counting all four bake jobs, each with per-request tracing on (`handler-side fate` present in every
one, so no count is a missing denominator):

| bake job | verdict | `[QUIESCE-TIMEOUT]` | `reconcile failed … disposed before the response arrived` | `VERDICT_TIMEOUT` |
|---|---|---:|---:|---:|
| core `main-cd` #7937 | FAILED `install: Chess` | 6 | 10 | **5** |
| core `main-cd` **#7938** | **SEALED** | **10** | **12** | **0** |
| core `main-cd` #7941 | FAILED `install: Hosting` | 5 | 8 | **0** |
| MeshWeaver.Plugins #1422 gate | FAILED `idempotence: Chess` | 3 | 7 | **10** |

**A package root disposed under an in-flight reconcile is at its MAXIMUM in the run that sealed.** It
is normal, it is answered, and it is not the cause. What separates a seal from a failure is the third
column — a write that reached no verdict at all — which is what this PR fixes.

It also means the three occurrences are **not one bug**. #7941 has zero unanswered write verdicts and
instead carries two `JsonSynchronizationStream … resubscribe failed` lines (zero in the sealed
control), whose parked stream leaves the outer `CreateOrUpdateNodeRequest` with no terminal one layer
*above* the verdict watcher. That is #3498's shape, owned elsewhere — both fixes are needed before
`main-cd` can seal.

rbuergi:
## Review finding: the asymmetry argument's cost side is understated, and #3477 is the evidence

The classification fix is right and the reasoning for it is the strongest part of this PR — a
teardown genuinely is not a verdict about the write, and `IsDisposedContainer` had already written
that down one layer over. This is about **one load-bearing sentence** in the justification.

> *"the direction of a mistake is asymmetric … a false 'retryable' costs at most two idempotent
> re-diffs, while a false 'terminal' loses a write and fails an install."*

That holds **only if the re-enqueue answers**. There is a measured, open, unreproduced case where it
does not: **#3477**.

```
[Warning] [MeshNodeStreamHandle] [UpdateRemote] LATE_NACK_REENQUEUE hub=cache/Z86EFbZ4…
    target=TestData/late-nack-node attempt=1 code=OwnerDisposing — the patch was never applied;
    re-enqueueing the original update, rebased on state newer than 0
[20:57:10.860] === TEST FAILED: The operation has timed out.
```

`LateNackReenqueueTest`, Plugins shard 3, 1/330. After the re-enqueue on **`code=OwnerDisposing`,
attempt=1**, in the following 45 s **nothing landed and nothing logged** — no `LATE_ACK`, no second
NACK, no `LATE_NACK_REENQUEUE failed`, no error to the caller. #3477's own words: *"a write that was
NACKed, re-enqueued and then neither applied nor answered is the write-loss class this test exists to
refuse."*

**Why this bears on this PR specifically:** the re-enqueue arm is
`MeshNodeStreamExtensions.cs:~1981`, unchanged by this diff, and its trigger is
`lateErr.Code is OwnerDisposing or OwnerNotReady or Conflict`. This PR **increases the traffic
through that arm** — that is precisely its mechanism. So the cost of a false "retryable" is not
bounded at *"two idempotent re-diffs"*; it is bounded at that **plus whatever #3477 is**, which is
currently *silent write loss* with no caller notification.

That does not make the change wrong. Terminal-`Unknown` was losing writes *loudly and always*;
this loses them *rarely and silently*, and rarely-and-recoverable beats always-and-fatal. But the
sentence as written would let a reader conclude the downside is fully bounded, and it is not.

### Two smaller things, both cheap

**1 · A comment at the re-enqueue site is now false.** `MeshNodeStreamExtensions.cs:~1983` says:

> *"`OwnerDisposing` / `OwnerNotReady` mean the patch never reached a merge, so there is no newer
> version to wait for"* — and sets `lateRebaseFrom = 0` on that basis.

After this PR, `OwnerDisposing` is **also** produced for patches whose merge *did* reach
`PATCH_MERGE_STAMPED v=2 refused=0`. The behaviour stays correct — `update` is a function, so
re-running it against freshest state re-diffs to a no-op — but the *stated precondition* of that arm
is no longer true, and it is the comment a future reader will rely on when deciding whether
`lateRebaseFrom = 0` is still right. Worth one sentence there, since this PR is what makes it stale.

**2 · The PR body's direction-of-risk line** (*"its failure mode is a slower install, not a lost
write"*) inherits the same gap and would be worth softening the same way.

### What would falsify this finding

- If #3477 is shown to be a **test-harness** artefact rather than a re-enqueue defect — its own
  controls do not establish that (8/8 and 12/12 locally on both pins, 1/330 in CI, suspect
  `67e71f53d` unconfirmed), and the trail after `LATE_NACK_REENQUEUE` goes dark by construction,
  which is why #3477 asks for `LATE_NACK_REENQUEUED requestId=<new>` instrumentation.
- Or if the re-enqueue for the newly-reclassified faults takes a **different** path from the one
  #3477 exercises. Reading `MeshNodeStreamExtensions.cs` I do not believe it does — same file, same
  arm, same code — but I have not traced it end to end and would rather be corrected than believed.

**Suggested disposition, not a merge block:** land this, and add #3477 to what the first candidate
seal is watched for — it becomes materially more likely to be hit now that more faults route through
that arm, and it is currently invisible when it happens.

rbuergi:
## Finding CONFIRMED, not falsified — and it is worse than you stated, in the direction you suspected

You asked to be corrected rather than believed. Traced end to end; you were right, and the trace
sharpens it.

**Both re-enqueue arms are the same mechanism reached two ways**, and both trigger on
`OwnerDisposing or OwnerNotReady or Conflict`:

- **direct** — `MeshNodeStreamExtensions.cs:~2160`, `OWNER_NACK_REENQUEUE`: the caller's `Observe`
  callback is still pending, i.e. the verdict arrived inside `UpdateResponseWaitBound` (2 s);
- **late** — `~1981`, `LATE_NACK_REENQUEUE`: the verdict arrived after that, **or** the owner's own
  `Post` was refused and `RoutePatchVerdict` fell through to `ILatePatchVerdictSink.Dispatch`.

**The dominant shape of the faults I reclassify takes the LATE half — #3477's exact arm.** The
mechanism #2543 attributes the sync flush fault to is a closed lifetime scope, and
`HostedHubsCollection` closes a hosted hub's scope on `DisposalCompleted`. A hub at that point is
past `DisposeHostedHubs`, so `MessageService.PostImplGeneric` **refuses** its own post
(`POST_REFUSED_SHUTTING_DOWN`) and `PostPatchVerdict` routes the verdict through the late-verdict
sink by construction. So it is not merely "the same file, same arm, same code" — the teardown
population is *selected for* the late path.

So the sentence was wrong and I have replaced it, in both homes, rather than softened it:

> a false "terminal" loses a write and fails an install, **every time**; a false "retryable" costs an
> idempotent re-diff, capped at two — **but the cheap side is NOT fully bounded**: #3477 is a
> measured, open case where that arm went dark (no `LATE_ACK`, no second NACK, no error to the
> caller, 45 s). Routing more faults here makes that class more likely, and it is silent when it
> happens. The trade is *rarely and recoverably* against *always and fatally*, which is still the
> right way round — it is not "at most two re-diffs".

### The two smaller items, both taken

1. **`MeshNodeStreamExtensions.cs:~1983`** — the `lateRebaseFrom = 0` precondition. You are right that
   this PR is what makes it stale, and I have written the corrected reasoning at the site rather than
   just flagging it: the *premise* changes (`OwnerDisposing` can now carry a patch that DID reach
   `PATCH_MERGE_STAMPED`) but the *conclusion* survives, for the opposite reason to the one stated —
   there is no version to wait for because the version we would wait for died unpersisted with its
   hub, so filtering on it would park until the bound. `update` is a function: the re-run recomputes
   the patch against the fresh activation's pre-merge state and lands it, or, if the commit did
   survive, re-diffs to an empty no-op. Only `Conflict` needs the version. The twin comment on the
   direct arm (`~2178`) now carries a one-line pointer to it.
2. **The direction-of-risk line** — corrected as quoted above, in the doc page and in the
   `ClassifyPatchException` remarks. (The exact phrase *"a slower install, not a lost write"* isn't in
   the PR body; the sentence carrying that gap was the asymmetry claim, and it is the one fixed.)

### Disposition

Landing as you suggest, and **#3477 is added to the doc page as what a first candidate seal must be
read for** — a `LATE_NACK_REENQUEUE` with no terminal after it — with the note that #3477's own ask
(`LATE_NACK_REENQUEUED requestId=<new>` linking the re-enqueued request's trail to the original) is
what would make it loud. That is the honest disposition: this change makes an existing silent class
more reachable, and the mitigation is instrumentation on #3477, not withholding a fix that turns an
always-fatal verdict into a recoverable one.

Re-verified after the edits: `MeshWeaver.Mesh.Contract`, `MeshWeaver.Data`, `MeshWeaver.Data.Test`
Release `-warnaserror` clean; `MeshWeaver.Data.Test` green; `DocumentationLinkIntegrityTest` green.

PR #3532, opened 2026-09-07
fix(#3477): correlate a re-enqueued write with the NACK that caused it

### LINKED ISSUE #3477: LateNackReenqueueTest: after LATE_NACK_REENQUEUE nothing lands or answers within 45 s (1/330, Plugins shard 3, first run on 3.0.0-ci.7917; 28/28 locally on both pins)

rbuergi:
Closes #3477.

`Pairs-with: none` — adds no public type or member and removes none; format arguments, one new log line, one test.

**Parser-delta: none** — does not touch `scripts/check-type-forwards.py`.

---

## The defect: a fresh `requestId` per attempt, so the chain broke exactly where it mattered

```csharp
var requestId = Guid.NewGuid().AsString();      // ~1941, per ATTEMPT
lateRegistry?.Register(requestId, _path!, resp => { … });
```

The re-enqueue calls `UpdateRemote(update, attempt + 1, …)`, which runs that line again and registers under a **different** id. So after

```
[UpdateRemote] LATE_NACK_REENQUEUE … attempt=1 code=OwnerDisposing
```

there was **no id to follow**, and the three outcomes were indistinguishable in the log: the re-enqueue never left the cache hub; it left and never activated the owner; it activated and never answered. #3477's own words — *"a write that was NACKed, re-enqueued and then neither applied nor answered is the write-loss class this test exists to refuse"* — and it could not be attributed.

## The fix: one correlation id, minted at entry, inherited by every re-attempt

`UpdateRemote` gains `correlationId`; `corr = correlationId ?? Guid.NewGuid().AsString()` mints once at the caller's entry, and **both** re-enqueue arms pass `correlationId: corr`.

`corr=` now appears on `BEGIN`, `LATE_ACK`, `LATE_NACK_REENQUEUE`, `LATE_NACK_REENQUEUE failed`, `OWNER_NACK_REENQUEUE` and `VERDICT_TIMEOUT` — the last being the line an operator actually reaches for, which now carries the instruction outright: *`grep corr=<id>` yields every attempt of this write, including re-enqueues that registered under a different requestId.* A new `LATE_VERDICT_REGISTERED` names the per-attempt `requestId` beside the stable `corr`.

**Purely additive.** Nothing new is awaited, bounded or retried; no behaviour changes.

## 🚨 Why it matters more since tonight

#3505 reclassifies owner-side teardown faults as `OwnerDisposing`, which feeds **this arm** — and not incidentally. Once `HostedHubsCollection` closes a hosted hub's scope, the hub is past `DisposeHostedHubs`, `PostImplGeneric` refuses its own post, and the verdict routes through `ILatePatchVerdictSink` **by construction**: the *late* half, which is the half #3477 measured going dark. #3505's cost argument — *"a false retryable costs at most two idempotent re-diffs"* — holds only if the re-enqueue answers.

## Log levels, stated rather than assumed

`src/` levels are committed contract, not a verbosity knob. `LATE_VERDICT_REGISTERED` and `BEGIN` are **Debug** — one per write and one per late-registered write, on the hot path. The **Warning** that pays for them is `VERDICT_TIMEOUT`, which fires only once a write has already failed to settle, i.e. exactly when the chain is worth its cost.

## 🚨 What the test covers — and what it does NOT

`LateNackReenqueueCorrelationTest` drives the real sequence on the sibling's proven rig: park the owner's merge executor, write cross-hub, fence on the armed late watch, fence on the handler having run, dispose the owner so its NACK is necessarily late. It then asserts the `LATE_NACK_REENQUEUE` line **carries** a correlation id — which before this change it did not. **Passes in ~190 ms**, deterministic: a disposed owner's post is refused, so the verdict routes through the late sink immediately rather than after the 2 s response bound.

It deliberately does **not** assert on the write landing — its sibling `LateNackReenqueueTest` does, and is 1-in-330 flaky on exactly that wait, which is #3477's own open subject.

**And it does not pin INHERITANCE.** That needs the re-attempt's own `BEGIN`, which is `LogDebug` and is dropped by the harness's filter before any provider sees it. Two attempts to raise that filter for one category did not reach the factory the hub resolves. The honest options were to promote a hot-path line to Information so a test could see it — **forbidden** — or to say so. So the inheritance is guaranteed by the two call sites passing `correlationId: corr`, which the compiler checks and no test does. That gap is stated in the test's own doc comment and on the issue, rather than left for someone to find behind a green suite.

## Verification

- `MeshWeaver.Mesh.Contract` Release `-warnaserror` → **0 Warning(s), 0 Error(s)**
- `MeshWeaver.Graph.Test` Release `-warnaserror` → **0 Warning(s), 0 Error(s)**
- `LateNackReenqueueCorrelationTest` → **1 passed**, real `.trx`, and the captured line verified in the run output: `LATE_NACK_REENQUEUE … code=OwnerDisposing corr=s00FLGaix0yWYK2z8alWWQ`

### DIFF
--- a/src/MeshWeaver.Documentation/Data/Architecture/WritingTests.md
+++ b/src/MeshWeaver.Documentation/Data/Architecture/WritingTests.md
@@ -384,6 +384,73 @@ Alternatively, start the assertion first without awaiting it (`var assertion = o
 
 ---
 
+## 🚨 Never `await` an Observable the Code Under Test Signals
+
+**`await someObservable` resumes its continuation INLINE on whatever thread called `OnNext`.** If
+that thread belongs to the code under test — a hub's action-block thread, a grain turn — then
+*everything after the await* runs there: the remaining assertions, the `finally`, the method return,
+and the framework's teardown.
+
+If the test is also waiting on that same component (disposing it, draining it, parking it), it
+**wedges in total silence**. No assertion fails and no `.Timeout()` fires, because the thread that
+would carry the test forward is the one being waited on.
+
+### The fingerprint: absent teardown lines
+
+```
+=== TEST START: … ===
+[+0.2s]  [Warning] … LATE_NACK_REENQUEUE … corr=…      ← the awaited line WAS written
+[+5.2s]  [Warning] … Dropping UnsubscribeRequest …
+                                                        ← 84 seconds of nothing
+[+89.8s] [FAIL] Test execution timed out after 90000 milliseconds
+```
+
+**No `=== TEST END ===` and no `[DISPOSE]` lines.** Those are present on every ordinary failure, so
+their absence means the method never returned — which separates "an inner wait lost" from "the
+thread is gone". Read that before theorising about which bound expired.
+
+### Why it presents as a flake
+
+A capture that replays a buffer and then concats a live subject has **two paths that resume on
+different threads**:
+
+| path taken | continuation resumes on | observed |
+|---|---|---|
+| the **replay** of already-buffered values | the test's own thread | 663 ms, green |
+| the **live subject** | the component's own thread | 90 s, killed |
+
+So it passes locally and on the PR build, and fails in the merge queue. **Re-running proves
+nothing** — local runs may take the replay path every time. Measured on #3477/PR #3532: five green
+local runs did not discriminate, because the *un-fixed* code also passed locally.
+
+### The control that names it in one step
+
+Its sibling test — identical park/dispose choreography, differing only in that it asserted on
+durable storage rather than on a log line — **passed in the same shard of the same run**. When two
+tests of one mechanism disagree on a single host, diff what they *observe*, not what they *do*.
+
+### What to write instead
+
+**Poll a buffer from `Observable.Interval`**, so the continuation resumes on the scheduler's thread:
+
+```csharp
+// ✅ the continuation lands on the scheduler's thread, never on the hub's
+return Observable.Interval(TimeSpan.FromMilliseconds(50)).StartWith(0L)
+    .Select(_ => lines.ToImmutableArray().Select(l => re.Match(l)).FirstOrDefault(m => m.Success))
+    .Where(m => m is not null).Select(m => m!)
+    .FirstAsync().Timeout(within);
+```
+
+Delete the subject rather than leaving it unused — keeping one invites the await back. And **prefer
+asserting ground truth (what landed in storage) over a log line**: a log line buys determinism only
+if *observing* it cannot perturb what produced it.
+
+🚨 **Avoiding `.ToTask()` does not avoid this.** The ban on `.ToTask()` exists because *"a Task
+completed inside an Rx pipeline resumes its awaiter inline on the signalling thread"* — and that
+reasoning applies verbatim to awaiting the observable directly. In the incident above, the offending
+method's own comment cited that rule to justify returning `IObservable`, and then the caller awaited
+it. The hazard is the **inline resume**, not the `Task`.
+
 ## "One Emission Carrying Everything" — Batched or Late? The First Snapshot Decides
 
 A test that asserts progress **streams** (several distinct snapshots, not one lump) fails with a
--- a/src/MeshWeaver.Mesh.Contract/MeshNodeStreamExtensions.cs
+++ b/src/MeshWeaver.Mesh.Contract/MeshNodeStreamExtensions.cs
@@ -1590,17 +1590,29 @@ public sealed class MeshNodeStreamHandle : IObservable<MeshNode>
     /// refused this write against, so the re-attempt rebases on something newer
     /// (<see cref="RebaseSource"/>). 0 for a caller write and for the re-enqueue codes that never
     /// reached a merge.</param>
+    /// <param name="correlationId">
+    /// 🚨 #3477: the id that makes a re-enqueued write followable. Each attempt registers under a
+    /// FRESH <c>requestId</c> (the late registry mints one per attempt), so after
+    /// <c>LATE_NACK_REENQUEUE</c> the trail went dark — nothing related the child's registration to
+    /// the parent's, and "never left the cache hub", "never activated the owner" and "activated and
+    /// never answered" were indistinguishable in the log. That is the ambiguity #3477 exists to
+    /// remove. Minted once at the caller's entry and passed UNCHANGED into every re-attempt, so one
+    /// <c>grep corr=&lt;id&gt;</c> yields the whole chain across attempts.
+    /// </param>
     private IObservable<MeshNode> UpdateRemote(
         Func<MeshNode, MeshNode> update, int attempt = 0, long refusedBaseVersion = 0,
-        MeshNode? pendingSelfWrite = null, Action<MeshNode?>? onLocalState = null)
+        MeshNode? pendingSelfWrite = null, Action<MeshNode?>? onLocalState = null,
+        string? correlationId = null)
         => Observable.Create<MeshNode>(observer =>
         {
+            // Minted here on the FIRST attempt only; a re-enqueue passes the parent's through.
+            var corr = correlationId ?? Guid.NewGuid().AsString();
             var diagLogger = _workspace.Hub.ServiceProvider
                 .GetService<Microsoft.Extensions.Logging.ILoggerFactory>()
                 ?.CreateLogger("MeshWeaver.Mesh.MeshNodeStreamHandle");
             diagLogger?.LogDebug(
-                "[UpdateRemote] BEGIN hub={Hub} target={Path} attempt={Attempt}",
-                _workspace.Hub.Address, _path, attempt);
+                "[UpdateRemote] BEGIN hub={Hub} target={Path} attempt={Attempt} corr={Corr}",
+                _workspace.Hub.Address, _path, attempt, corr);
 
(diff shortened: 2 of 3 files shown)


rbuergi:
Held as draft — **my own test in this PR is flaky, and it is flaky by construction rather than by chance.** The merge-queue steward ejected this twice (runs [34088390261](https://github.com/Systemorph/MeshWeaver/actions/runs/34088390261), [34089526911](https://github.com/Systemorph/MeshWeaver/actions/runs/34089526911)) and correctly declined to re-queue it. Not re-queueing; fixing it.

## The failure is not what the label says

`LateNackReenqueueCorrelationTest` did **not** fail an assertion. The line it asserts on *was written*, 2.4s in:

```
[06:12:24.227] LATE_NACK_REENQUEUE hub=cache/w4Pa… target=TestData/corr-node
               attempt=1 code=OwnerDisposing corr=yW4yTuav70WJZrf5mU5lDg
```

and the test was then killed at **90 000 ms** with no inner wait reporting anything.

## Root cause: the park and the dispose are the same hub, so they wait on each other

The test parks the owner's merge executor, then calls `nodeHub.Dispose()` on **that same hub**. `Dispose()` drains the executor; the executor is sitting in

```csharp
SpinWait.SpinUntil(() => Volatile.Read(ref releaseGate) == 1 || owner.IsShuttingDown,
                   TimeSpan.FromSeconds(60));
```

and `releaseGate` is only written in the `finally`, which cannot run until after the assertion. So the only early exit is `owner.IsShuttingDown` flipping **before** dispose blocks — a race.

The arithmetic is what makes this conclusive rather than a guess:

| bound | value on CI | where from |
|---|---|---|
| outer test kill | **90 s** | `TestTimeouts.TestMilliseconds` = `Convergence * OuterMargin`, `OuterMargin = 2.0` |
| my assertion wait | **45 s** | `TestTimeouts.Convergence` |
| the park's spin cap | **60 s** | literal in the test |

For the **outer** bound to fire while the **inner** 45 s wait never reports, the test must burn more than 45 s before it ever reaches its assertion. The only thing there that can is the dispose. And because the spin cap (60 s) exceeds the whole remaining budget, that cap cannot rescue the test — **it guarantees the kill.** The 663 ms local pass and the 90 s CI kill are the two sides of one race, not two different behaviours.

## Why the sibling looked like a safe pattern to copy

`LateNackReenqueueTest` has the identical park/dispose/`finally` structure, and it is green — so the shape read as established. It differs in what it *asserts*: ground truth in durable storage, with its own 45 s poll **after** the dispose. Mine asserts on a **log line**, which buys no extra tolerance and adds nothing to the budget. The sibling therefore carries the same latent race and merely wins it more often; that is worth a separate look and I am not touching it in this PR.

## What changes

The production fix in `MeshNodeStreamExtensions.cs` is unaffected — no assertion about it failed, and the `corr=` it emits is visible in the failure log above. Only the test's choreography is wrong. I am reworking it so the assertion does not depend on winning a disposal race, then re-arming.

rbuergi:
**Correction to my comment above — I named the wrong mechanism, twice.** Root cause found; both earlier claims are falsified by reading the code they were about.

### What was wrong

1. **"`Dispose()` blocks draining the parked executor for up to 60 s."** It does not. `MessageHub.Dispose()` sets `disposalStarted = true` as the *first* statement inside a short lock — so `IsShuttingDown` flips immediately — and its own summary says *"Returns immediately; observe `DisposalCompleted` for completion."* It never blocks on the drain, so the park's `|| owner.IsShuttingDown` exit cannot be starved by it. The circular wait I described does not exist.
2. **"The 90 000 ms is `TestTimeouts.TestMilliseconds`."** It is this test's own `[Fact(Timeout = 90_000)]` literal. The arithmetic I built on `Convergence * OuterMargin` was numerically the same and causally irrelevant.

### The actual cause

The control settles it: on the same queue-build (34089526911), in the **same shard**, on the same host, the sibling `LateNackReenqueueTest` **passed** while this test wedged. The structural difference is not the park — the two are identical there — it is that this is the only one of the two that **awaits a live `Subject<string>` fed from a logger**.

`CapturingLoggerProvider.Add` called `emitted.OnNext(line)`, and `Add` runs **on whatever thread logged** — for `LATE_NACK_REENQUEUE`, the owner hub's action-block thread. Awaiting an observable resumes its continuation **inline on the signalling thread**. So everything after the await — the assertions, the `finally` that releases the parked merge turn, the method return, and the framework's teardown — ran **on the hub's own action thread while that hub was mid-disposal**.

That is why the failure is *silent*: there is no assertion to fail and no timeout to report, because the thread that would carry the test forward is the thread being waited on. It also explains the pass/hang split exactly, because `FirstMatching` had two paths:

| path taken | continuation resumes on | observed |
|---|---|---|
| the **replay** of the already-buffered lines | the test's own thread | 663 ms, green |
| the **live Subject** | the hub's action-block thread | 90 s, killed, **no `TEST END`, no `[DISPOSE]`** |

The absent teardown lines are the fingerprint — the method never returned.

There is a small irony worth recording: the XML comment on that very method congratulated itself for returning an `IObservable` rather than a Task, citing *"a Task completed inside an Rx pipeline resumes its awaiter inline on the signalling thread"*. That reasoning is correct and applies verbatim to awaiting the observable. Avoiding `.ToTask()` does not avoid the hazard `.ToTask()` is banned for.

### The fix

`FirstMatching` now **polls the buffer** from `Observable.Interval` instead of awaiting a subject, so the continuation resumes on the scheduler's thread and never on the hub's. The `Subject<string>` is deleted outright rather than left unused — keeping it would invite the await back.

I also removed a claim from the class doc that measurement had falsified: it asserted this test was *"deterministic where the sibling is not."* The opposite is what happened. Asserting on a log line instead of on storage buys determinism only if observing the line cannot perturb what produced it — here, observing it captured the producer's thread.

Re-verified locally across repeated runs before re-arming; the production change in `MeshNodeStreamExtensions.cs` is untouched throughout.

PR #3533, opened 2026-09-07
fix(cd): "incomplete because a publish is in flight" is a third verdict, not a stuck delivery (#3513)

### LINKED ISSUE #3513: The hourly main-cd reconcile reds on its own workflow_run twin's in-flight publish ('INCOMPLETE image set … published nothing') — 7963, 7965, 7940 tonight
## Measured 2026-09-07: the hourly `schedule` main-cd run reds on its own workflow_run twin's in-flight publish

| run | event | head | started | outcome |
|---|---|---|---|---|
| 7962 | workflow_run | `6577052de` | 01:02Z | **sealed** 01:38:49Z |
| **7963** | schedule | `6577052de` | ~01:07Z | **failure** at `CD delivered, or had a good reason not to` (01:07:43–01:07:47Z) |
| 7964 | workflow_run | `93dd88941` | ~01:39Z | **sealed** 02:32:25Z |
| **7965** | schedule | `93dd88941` | ~01:42Z | **failure**, same step (01:42:56–01:43:01Z) |
| 7966 | schedule | `93dd88941` | ~02:25Z | success — `Plugins: bake + seal` skipped (already complete) |

Both reds say:

```
##[error]main `6577052` is GREEN and has an INCOMPLETE image set, and this reconcile published nothing.
Delivery is stuck: every self-updating install stays on the previous image. This job ex…
```

Delivery was not stuck. The image set was incomplete because the `workflow_run` twin on the same commit was **between `Build + push` and `Promote`** at that moment — 7962's promote ran 01:19–01:20Z, the reconcile looked at 01:07Z. The scheduled job read a snapshot mid-publish as a terminal fact and failed itself with a message that names the wrong condition, forty minutes before the same commit sealed.

It recurs every hour a publish straddles the top of the hour (7940 at 21:24Z is the same shape on `62039c996`; #3355's slot watch shows it thrice tonight), and it costs a red on `main`'s CD history that a reader must reason away — the *"failed CD run still promotes"* trap from the other direction.

## Expected

Before declaring `INCOMPLETE image set … reconcile published nothing` a failure, the reconcile checks for an in-flight `main-cd` run on the same head (REST `actions/workflows/main-cd.yml/runs?head_sha=…&status=in_progress`) and, when one exists, reports "publish in flight on run N since T" and exits green as a no-op — the same verdict 7966 reached by arriving after the twin finished. A genuinely stuck delivery is one where NO run on that head is in flight and the set is still incomplete; that case keeps its red.

Not this: a longer wait inside the reconcile, or suppressing the red outright — the stuck case is real (#2543's family produced it) and must stay loud.

rbuergi:
Closes #3513.

## The defect

The hourly `main-cd` reconcile red on its own `workflow_run` twin's in-flight publish — twice inside forty minutes on 2026-09-07, on a night when people were reading CD conclusions to decide a release.

| run | event | head | outcome |
|---|---|---|---|
| 7962 | `workflow_run` | `6577052` | publishing; **sealed 01:38:49Z** |
| **7963** | `schedule` | `6577052` | **FAILED 01:07:47Z** — "delivery is stuck" |
| 7964 | `workflow_run` | `93dd889` | publishing; **sealed 02:32:25Z** |
| **7965** | `schedule` | `93dd889` | **FAILED 01:43:01Z** — same message |
| 7966 | `schedule` | `93dd889` | success — arrived *after* 7964 finished |

`delivery-verdict` derived a claim about the **future** ("delivery is stuck") from four gate outputs that describe main at **one instant** (`reason=reconcile publish=false complete=false green=true`). Those four facts are also exactly what a reconcile looks like while its twin is halfway through creating the image set.

**The gate had already worked it out and the verdict could not hear it.** From run 7963's own gate log:

```
⏳ An older run of this workflow is still publishing `6577052`:
   https://github.com/Systemorph/MeshWeaver/actions/runs/34069402974 — deferring,
   so one commit yields ONE image set (MeshWeaver#3376).
```

`decision false "…"` writes only a **boolean**. The reason reached the log and the step summary and nothing the verdict could read, so the verdict re-derived a diagnosis from `publish=false` and got a different one.

## The shape chosen — a third verdict, not a skip

Skipping the reconcile while a run is in flight would trade a false red for an invisible hole (GitHub paints a skipped job the same colour as a passed one), and the window in which publishes happen **is** the window this alarm exists to watch. So the step **concludes** instead of declining to conclude. It asks the question its claim depends on, itself, at verdict time:

| finding | verdict |
|---|---|
| a run of this workflow is `queued`/`in_progress` on this exact commit | **incomplete because a publication is in flight** — green, naming the run, its status and its start |
| no such run | **incomplete and nobody is fixing it** — the original red, now carrying the negative finding that makes it a measurement |
| the probe could not be answered | **red, fail-closed** — an unanswered probe is never reassurance |

Four properties, each closing a way this could have become a band-aid:

- **It re-reads; it does not relay.** Piping the gate's `⏳` decision through as an output would have been shorter and would have repeated the defect one level along — the gate's reading is *also* a snapshot.
- **Self-limiting without a timer.** No sleep, no retry, no widened window. Suppression lasts exactly as long as a run is *observably live*, and every job here carries a literal `timeout-minutes` ≤ 45 — so **GitHub** ends a hung publication, after which the next tick finds nothing in flight and reds.
- **`.id != $RUN_ID` excludes this run and nothing else.** Dropping it is the catastrophic edit: every reconcile would find *itself* in flight and the alarm would be silenced forever, with a green tick.
- **The probe sits inside one branch.** Red main, empty gate output, cancelled tip and the legitimate PR no-op are exactly as loud as before, live run or not.

## Verification — both arms, demonstrated

`.github/scripts/test-cd-steps.py` **extracts** this step from `main-cd.yml` by its new `id: verdict` and executes it against fixtures, driving the step's **real `--jq` filter** over a real `workflow_runs` payload. A hand-supplied answer would prove nothing about the filter, and the filter is the whole discriminator. 22 cases:

```
  PASS  a green, incomplete main with NOTHING in flight is still RED      ← coverage kept
  PASS  ...and the red now carries the negative finding that makes it a measurement
  PASS  a reconcile does NOT count ITSELF as the publication in flight
  PASS  a live run of ANOTHER workflow on the same sha does not count
  PASS  a COMPLETED run on the same sha does not count as in flight
  PASS  run 7963's exact conditions no longer red — its twin 7962 was mid-publish   ← the false red
  PASS  run 7965's exact conditions no longer red either — twin 7964 was mid-publish
  PASS  a QUEUED run counts as a live publication, and is reported as queued
  PASS  a NEWER live run on this sha falsifies 'stuck' just as an older one does
  PASS  a FAILING probe fails the step — it never silences the alarm
  PASS  a push path on a RED main is still red, live run or not
  PASS  an EMPTY gate verdict is still red, live run or not
  PASS  a PR-triggered workflow_run is still the legitimate no-op it always was
  PASS  a cancelled, superseded push is still a green no-op / a cancelled TIP is still red
```

**A gate nobody has watched fail is not a gate**, so the fix was sabotaged three ways and each reds:

| sabotage | cases red |
|---|---|
| drop `.id != $RUN_ID` from the probe's filter | 7 |
| make the branch `exit 0` without probing — the "just skip it" fix | 16 |
| treat an unanswerable probe as "in flight" | 3 |

**Live against the real API**, not only fixtures — the exact probe, on a sha with a CD run in flight and on `6577052` today:

```
── 19b077868…  (CD run 7976 in_progress)
   probe answer='34085298189 in_progress 2026-09-07T05:02:30Z …/runs/34085298189'
   ⇒ VERDICT: deferred (publication in flight)
── 6577052de…  (7962 completed/success, 7963 completed/failure)
   probe answer=''
   ⇒ VERDICT: stuck (RED)
```

## Coverage, provably unchanged

The red is suppressed **only** by a `queued`/`in_progress` run of this workflow on main's exact HEAD — i.e. only when something demonstrably *is* completing the set. The genuinely-stuck case (`#2543`'s family) has no such run and still reds, which is the first case in the list above. Worst case on a hung publication is one hour of extra latency, bounded by GitHub's own job cap rather than by anything written here.

## Gates

- `check-workflow-timeouts.py --root .` — 71 jobs, 0 violations, cap 45
- `check-workflow-shell.py` — 0 live findings
- `check-pr-secret-preflight.py` — 0 violations
- `actionlint` on `main-cd.yml` — clean
- `MeshWeaver.Documentation.Test` (link integrity, embed integrity, all `main-cd.yml` guards) — 75 passed, 0 failed
- `dotnet build -c Release -warnaserror` — 0 Warning(s), 0 Error(s)

## Work product committed

- `Doc/Architecture/ContinuousDeliveryContract` → **"INCOMPLETE is a snapshot; STUCK is a claim about the future — the third verdict"**: the measurement, why the gate's correct answer could not reach the verdict, why "skip while busy" is the wrong fix, and the four load-bearing properties.
- `Doc/Architecture/ReadingCiSignals` → the inverse of its "delivery stops behind green ticks" section: a CD red on main that means nothing, and to check whether the run named an in-flight publication before treating it as an incident.

Pairs-with: none — CI and documentation only; the diff removes no public type or member from `src/` (no `.cs` file is touched).

### DIFF
--- a/.github/scripts/test-cd-steps.py
+++ b/.github/scripts/test-cd-steps.py
@@ -41,6 +41,7 @@ from pathlib import Path
 WORKFLOW = ".github/workflows/main-cd.yml"
 STEP_ID = "release"
 DECIDE_STEP_ID = "decide"
+VERDICT_STEP_ID = "verdict"
 
 SHORT_SHA = "aaf95af"
 VERSION = "3.0.0-rc8.ci.6360"
@@ -113,12 +114,42 @@ def fixture(tagged: bool) -> list[dict]:
 # will still be there after someone edits the other.
 GH_STUB = """#!/usr/bin/env bash
 echo "gh $*" >> "$GH_CALLS"
-# The ONE read the decide step makes through gh: "is an older run of this workflow on this sha
-# still running?" (MeshWeaver#3376). A case supplies the answer the real `--jq` would have
-# produced (the older run's URL, or nothing) in GH_RUNS_RESULT; every other invocation stays
-# silent, exactly as before.
+# Two reads are modelled, because two steps make them:
+#  * `actions/runs?head_sha=…` — "is a run of this workflow live on this sha?", asked by `decide`
+#    (older runs only, MeshWeaver#3376) and by `verdict` (any live run, MeshWeaver#3513). A case
+#    supplies the answer the real `--jq` would have produced in GH_RUNS_RESULT.
+#  * `commits/main` — main's tip, which `verdict` uses to tell "superseded" from "stuck".
+# GH_RUNS_FAIL makes ONLY the first one fail, so a case can drive the fail-closed arm without
+# also breaking the tip resolution. Every other invocation stays silent and succeeds, exactly as
+# before — the safety property (this stub can mutate nothing) is unchanged.
+#
+# 🚨 GH_RUNS_FIXTURE runs the step's OWN `--jq` program, with real jq, over a real
+# `workflow_runs` payload. That is deliberately stronger than answering GH_RUNS_RESULT: the whole
+# discriminator lives in that filter, and its most dangerous defect — dropping `.id != $RUN_ID`,
+# so every reconcile finds ITSELF "in flight" and the alarm is silenced forever — is invisible to
+# any case that hands the step a pre-computed answer.
 case "$*" in
-  *actions/runs*) printf '%s' "${GH_RUNS_RESULT:-}" ;;
+  *actions/runs*)
+    if [ -n "${GH_RUNS_FAIL:-}" ]; then
+      echo "gh: HTTP 502 (https://api.github.com/repos/x/y/actions/runs)" >&2
+      exit 1
+    fi
+    if [ -n "${GH_RUNS_FIXTURE:-}" ]; then
+      prog=""
+      while [ $# -gt 0 ]; do
+        [ "$1" = "--jq" ] && prog="$2"
+        shift
+      done
+      if [ -z "$prog" ]; then
+        echo "stub gh: an actions/runs call with no --jq — the harness models the filter, not the payload" >&2
+        exit 97
+      fi
+      jq -r "$prog" < "$GH_RUNS_FIXTURE"
+      exit 0
+    fi
+    printf '%s' "${GH_RUNS_RESULT:-}"
+    ;;
+  *commits/main*) printf '%s' "${GH_TIP_RESULT:-}" ;;
 esac
 exit 0
 """
@@ -154,7 +185,8 @@ def die(msg: str):
 
 
 # ── running one case ────────────────────────────────────────────────────────────────────────
-def run_step(body: str, env: dict[str, str], rows: list[dict] | None, az_fail: bool = False):
+def run_step(body: str, env: dict[str, str], rows: list[dict] | None, az_fail: bool = False,
+             runs: list[dict] | None = None):
     with tempfile.TemporaryDirectory() as td:
         tmp = Path(td)
         binp = tmp / "bin"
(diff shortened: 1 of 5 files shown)


rbuergi:
Both review nits addressed in 47810d29f:

- `main-cd.yml`: the fail-closed comment's "indistinguish-/able" line split is gone — reworded so the point (GitHub renders a skipped job identically to a passed one) carries without the hyphenation.
- What's New: "sealing it about eighty" now reads "about eighty minutes".

No behaviour change — all 22 verdict cases, `check-workflow-timeouts`, `check-workflow-shell` and `actionlint` re-run clean, and the documentation guard suite passes 34/34.

Issue #2221, opened 2026-08-25
Registry publish drops a module's static web assets: every landed view pack has no wwwroot (EntityViews CSS 404 in prod; blocks the Views/Graph flip)


rbuergi:
## The defect

`PluginBundleEndpoints.MapPublish` — the registry's module-publish endpoint — never passes a bundle's **static web assets** to the shelf. It calls:

```csharp
ModulePublish.Validate(plugin, manifest, files, version, packagePath)   // no staticAssets
ShelveModule(accepted.Module, accepted.Files, …, minMeshVersion: …)     // no staticAssets
```

Every other part of the lane was already built for them: `Validate` takes a `staticAssets` parameter, `Accepted` carries `StaticAssets`, `ShelveModule` takes `staticAssets`, `LandCore` materializes them into the generation directory, and the CONSUMER path (`PluginBundleClient.LandFromBundle`) passes them correctly. Only these two arguments were missing — so **every module ever published to the registry landed with assemblies and no `wwwroot`**, and no consumer could receive assets no matter how complete the packed bundle was (a consumer never reads the producer's archive; it fetches what the shelf holds).

## Evidence

Production, memex-cloud, landed generation directories (`/data/modules/<name>@<gen>`):

| module | files | wwwroot |
|---|---|---|
| MeshWeaver.Blazor.Views@3badbda5 | 11 | **0** |
| MeshWeaver.Blazor.Graph@9877c683 | 3 | **0** |
| MeshWeaver.Blazor.EntityViews@12497147 | 8 | **0** |
| MeshWeaver.Blazor.Radzen@dc9459b4 | 3 | **0** |
| MeshWeaver.Blazor.GoogleMaps@26755191 | 2 | **0** |
| MeshWeaver.Blazor.Analysis@59b6417d | 2 | **0** |

The image copy of the same pack carries 58 files including `wwwroot/{Components/, _content/, MeshWeaver.Blazor.Views.styles.css, MeshWeaver.Blazor.Views.modules.json}`.

The producer side is healthy — I ran the workflow's exact sequence locally on `MeshWeaver.Blazor.Views` (`dotnet build` → `dotnet publish --no-build` → `module-pack --deps-closure`): the publish output carries `wwwroot/` (8 entries, incl. the scoped-CSS bundle and JS initializers) and the packed bundle's manifest declares **254 `staticAssets`** with the tree under `meshweaver/moduleassets/`. So neither the RCL publish nor the packer is at fault.

[...]

Issue #1728, opened 2026-08-17
A runtime-loaded module cannot ship NATIVE assets: the runtimes/ prune blocks the OgCard/SkiaSharp and Speech flips


rbuergi:
Sibling of Systemorph/MeshWeaver#1724. Same root cause, different mechanism, so they need separate fixes — filing this so nobody scopes the UI half as if it were the whole problem.

**Root capability gap: a runtime-loaded module cannot ship non-IL assets.** Two faces:

| face | asset | blocked work | issue |
|---|---|---|---|
| **static web assets** | `_content/<pack>/*.js/.css`, scoped CSS | Radzen, GoogleMaps, Analysis view packs | Systemorph/MeshWeaver#1724 |
| **native libraries** | `runtimes/<rid>/native/*.so/.dylib/.dll` | OgCard (SkiaSharp), Speech | **this** |

## Mechanism (verified in `memex/MeshModulesPublish.targets`)

The closure lane's **prune (0)** deletes the RID tree outright, and correctly so:

```xml
<!-- Prune (0): RID trees. A module loads via Assembly.LoadFrom, which never consults the
     module's deps.json — the fallback probe is the module's FLAT folder only — so a
     runtimes/<rid>/ tree under a module folder is unreachable dead weight by construction …
     A module that truly needs native/RID assets (Speech is the known exception, per the Systemorph/MeshWeaver#1644
     design) needs explicit flat placement or NativeLibrary resolution in ITS flip, not a RID
     tree. -->
<RemoveDir Directories="$(_AppDir)modules/%(MeshModuleClosure.Filename)/runtimes" … />
```

So the constraint is not the prune — the prune is right. `Assembly.LoadFrom` genuinely never reads the module's `deps.json`; a RID tree under a module folder is unreachable whether we delete it or not. Deleting it just stops shipping dead weight.

Compounding it: per Systemorph/MeshWeaver#1675/#1676 every module MSBuild invocation strips RID globals by design (a flipped module is outside the host's restore graph, so a RID flowing in from CD's per-arch publish fails `NETSDK1047` — the 2026-08-16 CD outage). A module publish is therefore **always portable**, which means natives **always** land under `runtimes/` and are **always** pruned. There is no configuration escape.

## What it blocks concretely

[...]

PR #3535, opened 2026-09-07
fix(ci): the reusable lanes land a module's static web assets, not only its DLLs

### LINKED ISSUE #3514: Reusable lanes land a module bundle's DLLs but not its meshweaver/moduleassets/wwwroot — a browser-driven mesh 404s every _content/<Module>/ asset
## The reusable lanes land a module bundle's DLLs, not its browser assets — a browser-driven mesh 404s every `_content/<Module>/…` the module ships

`node-repo-gate.yml` (~L909) and `node-repo-publish-bake.yml` (~L949) compose an external module with

```
cp -R "$u"/meshweaver/modules/. "$EXT/$name/"
MODULE_ARGS="$MODULE_ARGS --module /ext/$name/$name.dll"
```

A module bundle carries its static web assets separately, under `meshweaver/moduleassets/wwwroot/…` (`NuGetPackageWriter.ModuleAssetFolder`; `ModulePackCommand` lays a Razor project's `wwwroot/` there, dependencies namespaced under `wwwroot/_content/<Dep>/`). The portal serves a landed module's assets from `modules/<Name>/wwwroot` (`MeshModuleStaticAssetExtensions`) and logs only at Debug when there is nothing there — so a module landed this way loads fine and then 404s every asset in the browser. The runtime consumer (`ServedModuleBytes`) handles the prefix correctly; production is unaffected (both portals serve `…/_content/MeshWeaver.Markdown.Collaboration/Components/collaborativeMarkdownView.js` with 200).

**Where it costs today.** For the lanes' own use the gap is latent: the bake is mesh-free and the gate mesh runs no browser. It bites the moment a browser renders against a mesh whose modules were landed by this pattern. MeshWeaver.Education's `e2e/mesh/fetch-upstream.sh` (a local copy of this landing block) is exactly that: since MeshWeaver.Plugins#1268 moved the collaboration views INTO `MeshWeaver.Markdown.Collaboration` (2026-09-04 12:48Z), every Education mesh run has thrown

```
Unhandled exception rendering component: Failed to fetch dynamically imported module:
http://localhost:8323/_content/MeshWeaver.Markdown.Collaboration/Components/collaborativeMarkdownView.js
  at MeshWeaver.Blazor.Components.CollaborativeMarkdownView.OnAfterRenderAsync
```

and every exercise page rendered its brief and then a bare Code node card where the workbench belongs (Playwright screenshot in run [34070110529](https://github.com/Systemorph/MeshWeaver.Education/actions/runs/34070110529)'s `e2e-install-report`) — 4 × 180 s per course. `essentials.module.nupkg` ships the JS; the landing never copied it. The last green Education run (09-04 12:16Z) predates #1268 by thirty minutes.

**Fix pattern** (Education #274, `fetch-upstream.sh`), for the two lane sites:

```
if [ -d "$u/meshweaver/moduleassets/wwwroot" ]; then
  mkdir -p "$EXT/$name/wwwroot"
  cp -R "$u"/meshweaver/moduleassets/wwwroot/. "$EXT/$name/wwwroot/"
  # fail CLOSED: shipped vs landed asset counts must match
fi
```

Simulated against the run's own set: Markdown.Collaboration lands 270 asset(s) (its JS/CSS plus the namespaced dependency assets), the EntityViews/Graph view packs their scoped CSS. Low urgency for the lanes as they stand; filed so the next browser-driven consumer of these blocks (or the next satellite that copies them) does not re-derive it. Related: #2384 (the same class for the OpenStreetMap module's `.razor.js`), Plugins#808 (scoped CSS shipped in neither image nor module).

rbuergi:
Closes #3514.

## Which half was wrong

**The lane, not the packer.** A module legitimately carries static web assets and the platform has
a complete lane for them: `ModulePackCommand` packs `wwwroot/**` → `meshweaver/moduleassets/`,
`ModuleLandingService` writes each asset at its **verbatim module-relative path** beside the entry
assembly, and `MeshModuleStaticAssetExtensions.ModuleWwwrootPath` looks for `wwwroot` **beside the
loaded assembly** and serves it at `_content/<Name>/…`. Production is unaffected because
`ServedModuleBytes` reads the `moduleassets/` prefix correctly.

The recorded constraint *"modules cannot carry native or static assets"* is a different claim
wearing the same words: **native** payloads genuinely cannot ride (`Assembly.LoadFrom` never reads
a module's `deps.json`, #1728); **static web** assets can, and that half was corrected twice
already — most recently when #2221 fixed a registry shelf that discarded them.

So the two lanes' `cp -R "$u"/meshweaver/modules/. "$EXT/$name/"` was simply landing half the
bundle. Nothing was producing assets a module should not have.

## What changed

Both `node-repo-gate.yml` and `node-repo-publish-bake.yml` now also copy
`meshweaver/moduleassets/.` into the same module directory, and assert **per file** that what the
bundle ships is what the directory holds — failing closed with both numbers and the missing path.

Two deliberate refinements over the pattern in the issue:

* **module-relative, not `wwwroot`-specific.** The packer's contract (and `ValidateAssetPath`'s) is
  a relative path; `wwwroot/` is merely the only root it emits today. `cp -R moduleassets/.`
  survives a second root, `cp -R moduleassets/wwwroot/.` does not.
* **a denominator in the log**, not just a count: `external modules: 30 composed, 6 of them
  carrying static web assets, 294 asset file(s) landed`. An uncounted zero is what let this ride —
  24 of 30 real modules legitimately ship none, so "0 landed" looked exactly like the healthy case.

`node-repo-compile-check.yml` (`cp modules/*.dll refs/`) and `node-repo-module-pack.yml` are
deliberately unchanged: a Roslyn reference set and a bundle inspection have no request path.

## Verification

`.github/scripts/test-module-asset-landing.py` **executes** the lanes rather than asserting about
them, and runs in `dotnet-test.yml` beside the other workflow-shell gates.

* Both lanes' step is **extracted by `id:`** — a copy would pass while the real thing rots. Losing
  the id is RED, not a skip (watched: it dies with *"no step with id 'ext-modules'"*).
* The verdict is read **off the landed bytes' sha256**, never off the step's log.
* The bundle layout is read out of `NuGetPackageWriter.ModuleAssetFolder` in `src/`, so renaming
  the constant reds the harness instead of orphaning the lane.
* Three controls (no-asset bundle still composes · no-entry-DLL bundle is refused · the DLL closure
  still lands) so the harness can tell "it lands assets" from "it always passes".

**Both falsification arms, watched:**

| arm | result |
|---|---|
| whole asset block removed from the REAL `node-repo-gate.yml` | harness exit **1**, 12 findings, `0/12 landed` |
| only the `cp` removed, the lane's own check kept | the **step** exits non-zero naming the first missing asset — the check is not a no-op |
| `id: ext-modules` removed | harness exit **1**, refuses to run |
| a broken link added to the new doc page | `DocumentationLinkIntegrityTest` **fails** naming it |

**The loop closed on the URL, not the directory.** Real bundle → the lane step extracted from
`node-repo-gate.yml` → a real `WebApplication` with `AddMeshModuleStaticAssets()` /
`UseMeshModuleStaticAssets()` over the landed tree → an actual fetch. Against
`MeshWeaver.Blazor.Chat` 1.0.23 (23 assemblies, 283 assets, 5 mounts):

| request | with the fix | pre-fix |
|---|---|---|
| `_content/MeshWeaver.Markdown.Collaboration/Components/collaborativeMarkdownView.js` — the literal URL in #3514's stack trace | **200**, `text/javascript`, 12,391 B, byte-identical | **404** |
| `_content/MeshWeaver.Blazor.Chat/chatResizer.js` (own asset, re-based) | **200**, 4,273 B, byte-identical | **404** |
| `_content/MeshWeaver.Blazor.Chat/ChatMessageList.razor.js` (collocated JS) | **200**, 22,739 B, byte-identical | **404** |
| `_content/MeshWeaver.Blazor.Chat/background.png` (binary) | **200**, byte-identical | **404** |
| the same URL with `Accept-Encoding: br` | **200** + `Content-Encoding: br`, byte-identical to the landed `.br` sibling | **404** |
| `_content/MeshWeaver.Blazor.Chat/does-not-exist.js` — negative control | **404** | **404** |

The negative control is what makes the first five mean something. The pre-fix host's entire
complaint, at Debug: `Module MeshWeaver.Blazor.Chat contributes no static assets — no wwwroot at
…/MeshWeaver.Blazor.Chat/wwwroot`.

**Real-bundle measurement** — the extracted lane step run against all **30** module bundles of
MeshWeaver.Plugins run [34086016940](https://github.com/Systemorph/MeshWeaver.Plugins/actions/runs/34086016940) (2026-09-07):

| | modules | with assets | asset files shipped | landed byte-identical |
|---|---|---|---|---|
| with the fix | 30 | 6 | 294 | **294** |
| pre-fix | 30 | 6 | 294 | **0**, exit code **0** |

The 6: `Blazor.Chat` 283, `Blazor.OpenStreetMap` 7, `Blazor.Analysis`/`.EntityViews`/`.Graph`/
`.Radzen` 1 each (their scoped-CSS aggregates).

Local: `actionlint` (all workflows) · `check-workflow-shell.py` · `check-workflow-timeouts.py`
(71 jobs, 0 violations) · `check-pr-secret-preflight.py` · `MeshWeaver.Documentation.Test`
**304/304** Release `-warnaserror`.

## Work product

`Doc/Architecture/ModuleStaticAssets` — the four hops, the landing invariant, every site that lands
a module and what each owes, why the failure is invisible to review/CI/production, and the measured
denominators. Indexed from `Doc/Architecture`; What's New entry added.

## Two things the issue's premise gets slightly wrong

1. **Education #274 is OPEN, not landed.** The issue cites it as the fix pattern; its branch does
   carry the `moduleassets` copy, but the PR is a different change (`ci: fetch the upstream once,
   gate the bake in parallel…`), is `mergeable_state: blocked`, and Education's `main` still has the
   broken block — so Education's mesh e2e is still 404ing. That half is theirs to land; this change
   does not touch it.
2. **There is a third latent copy.** `MeshWeaver.Plugins/scripts/mesh-test.py` unpacks bundles whole
   and hands the mesh `--module …/meshweaver/modules/<Name>.dll`, so `ModuleWwwrootPath` resolves to
   `…/meshweaver/modules/wwwroot` while the assets sit at `…/meshweaver/moduleassets/wwwroot`. Same
   shape, latent only because nothing points a browser at it — filed as
   Systemorph/MeshWeaver.Plugins#1440.


rbuergi:
Copilot's three findings are all correct and all message/comment drift, no behaviour: the `LANES` comment said `(workflow, job id, step id)` for a `(workflow, step id)` tuple, and both lanes' summary line said `landed under /ext/<module>/wwwroot` when the copy is deliberately module-relative — which contradicts the very generality the change argues for.

They landed a minute too late: auto-merge had already enqueued this PR (`gh-readonly-queue/main/pr-3535-459e35f8`), and a push now would eject it. They are fixed and verified locally (harness + actionlint + `check-workflow-shell` + `check-workflow-timeouts` + `MeshWeaver.Documentation.Test` 304/304) and land as a follow-up immediately after this merges, rather than churning the queue for three strings.

PR #3562, opened 2026-09-07
fix(#3516): onboarding's refusal messages are catalog entries, in en and de

rbuergi:
Adds the five catalog keys the onboarding form's refusal messages resolve from. Half one of
Systemorph/MeshWeaver#3516; the page that reads them lives in `MeshWeaver.Plugins` (core has no
`.razor` files), and core is the source of truth for the catalog, so this merges first.

## Why

`/onboarding` is the one screen whose **first field is a language picker** — deliberately, so that
the rest of the form and the screens after it arrive in the language the visitor picked. Every label
on it already follows that choice. The messages that come back when the form **cannot** be accepted
did not: all five were English literals compiled into the page.

```
Username '…' is already taken. Please choose a different one.
This email is already assigned to user '…'. Please sign in with that account.
This portal is invitation-only and your email has not been invited. …
Registration is closed. …
Failed to save profile: …
```

A hard-coded UI string is a bug, and these are the ones that matter most: they arrive at the moment
the visitor most needs to know what to do next.

## What is here

| key | en | de |
|---|---|---|
| `onboarding.usernameTaken` | Username '{0}' is already taken. … | Der Benutzername „{0}" ist bereits vergeben. … |
| `onboarding.emailAlreadyAssigned` | This email is already assigned to user '{0}'. … | Diese E-Mail-Adresse ist bereits dem Benutzer „{0}" zugeordnet. … |
| `onboarding.invitationRequired` | This portal is invitation-only … | Dieses Portal ist nur mit Einladung zugänglich … |
| `onboarding.registrationClosed` | Registration is closed. … | Die Registrierung ist geschlossen. … |
| `onboarding.saveFailed` | Failed to save profile: {0} | Das Profil konnte nicht gespeichert werden: {0} |

The German follows the catalog's existing **impersonal** register (`auth.usernamePlaceholder`:
"Benutzernamen wählen", not "Wählen Sie einen Benutzernamen") and the same `„…"` quoting as the
`chat.model*` entries. Placeholders stay positional so a translator can reorder them.

## Verification

`LocalizationTest` (the completeness gate — every English key present in every shipped language):
**58 passed, 0 failed**, Release build `0 Error(s) / 0 Warning(s)`.

## Sequencing

- **No public surface is added, removed or changed** — two JSON files. No `Pairs-with:` /
  `Implementers:` obligation.
- **No What's New entry here, deliberately.** One was written and then removed on review: it
  described the onboarding form answering in the visitor's language, which is true of the PAIR and
  not of this half. The docs ride core at `MW_PLATFORM_REF` and the page rides Plugins `main`, so
  the two can reach a portal separately and a note that ships ahead of its behaviour is worse than
  none. The user-facing note is in Systemorph/MeshWeaver.Plugins#1455, with the behaviour.
- The Plugins half (the page: typed refusals, severity, adoption of these keys) follows once
  `MW_PLATFORM_REF` carries this merge. Its own suite asserts these keys resolve, so it is red
  until the pin moves rather than silently shipping raw key tokens.
- The React mirror (`MeshWeaver.Plugins/clients/react/src/i18n/`) does not consume these — they are
  server-rendered on a Blazor page with no React counterpart. No `npm run sync:i18n` is required by
  this change; the mirror's pinned `catalog-source.json` stays valid.


rbuergi:
Both findings taken, both real — `57beea61b`.

**The quoting convention.** Correct: the file uses literal `„ … “` everywhere else
(`chat.noUsableModel` and its siblings around line 232) and I had emitted `„` / `“`
escapes. JSON-equivalent, but a second convention inside one catalog is a convention nobody can
follow. They are literal characters now.

**The release note.** Also correct, and the fix is to **delete** it rather than reword it — the
reasoning is worth writing down because the timing is not obvious.

The two halves can reach a portal **separately**. The doc tree ships from core at
`MW_PLATFORM_REF`; the page ships from `MeshWeaver.Plugins` `main` via the `memex-portal-ai` image.
So a pin bump that lands this commit *before* Systemorph/MeshWeaver.Plugins#1455 merges would
publish a note describing behaviour the page does not yet have — which is exactly what was flagged,
and it is a live ordering, not a theoretical one (that PR is currently blocked on the pin).

Rewording it to a "coming soon" note would be worse: a changelog entry that describes nothing a
reader can go and see. Core's half here is a catalog addition with **no user-visible effect on its
own**, so the honest place for the one user-facing note is with the behaviour — it is in #1455, and
it already covers the localization half explicitly.

Re-verified after both changes: `Release -warnaserror` → `0 Error(s) / 0 Warning(s)` on both touched
test projects, `LocalizationTest` 58/58, and the whole `MeshWeaver.Documentation.Test` suite 304/304
— that last one reads the `WhatsNew` tree, so its green is the evidence the removal left nothing
dangling.

PR #3570, opened 2026-09-07
fix(#3567): a build candidate's stand-down states a fact the arbiter acts on

### LINKED ISSUE #3567: A build candidate's stand-down is decided on its own mirror, so a grant that lands inside the window survives it
## A candidate's stand-down is DECIDED on its own mirror and APPLIED as a merge patch, so a grant that lands in between survives it

`MeshWeaver.Plugins`' `BuildCoordinationTest.Follower_StandsDown_SoTheNextBuildCanStillBeClaimed`
has now failed on three separate cores, including ones that contain #3131 — the fix that was
supposed to close it. This is the remaining hole, and it is the mirror image of the one #3131 shut.

### The measurement that was already on the record

From MeshWeaver.Plugins#1193, with the permanent state dump #1401 landed:

```
FAIL: ClaimedBy=<the follower>  Status=Planning  RequestedClaims=[]  Ready=[fp-stand-down]
PASS: ClaimedBy=<null>          Status=Planning  RequestedClaims=[]  Ready=[fp-stand-down]
```

**One emission in fifteen seconds** — terminal, not slow. Three of the four fields are identical.
The whole difference is a holder that was never cleared, and the build had already completed
(`Ready` carries the fingerprint).

### The mechanism, and why it is not #3131

`WithdrawBuildClaim` runs on the CANDIDATE's hub, which does not own `Admin/Build`. So its
`stream.Update` lambda is evaluated against **that hub's mirror** and travels to the owner as an
RFC 7396 merge patch of the fields it changed —
`MeshNodeStreamHandle.UpdateQueued` → `ComputeMergePatchDiff(currentNode, updatedNode)`. A field
the lambda leaves unchanged is **absent from the patch**, and by RFC 7396 an absent member leaves
the owner's value untouched.

Its second half is conditional on exactly such a field:

```csharp
var grantedNotStarted = state.ClaimedBy == holder && state.Status is BuildStatus.Planning;
```

So on a mirror the grant has not reached yet, `grantedNotStarted` is **false**, the patch carries
the registration removal and **no `claimedBy` at all**, and a grant the arbiter committed inside
that window survives the stand-down.

The arbiter, by contrast, writes the node it OWNS (`workspace.GetMeshNodeStream()` with no path →
`IsOwn` → `UpdateOwn`), so ITS lambda is serialised against fresh state. **That asymmetry is the
bug**: #3131's own words are *"conditional writes make each half safe on its own; they cannot order
two writers"* — and it fixed the ordering at `ApplyGrant`, the one point that could see a
stand-down that had ALREADY happened. It could not fix this one, because here the publication was
**legitimate at the moment it happened**: the candidate was still registered. What arrives late is
the stand-down.

| order | outcome |
|---|---|
| stand-down decided **and applied** before the grant | `ApplyGrant` refuses (winner neither holder nor candidate), `HandBackAStoodDownGrant` drops the lock. Closed by #3131. |
| grant applied, **then** stand-down decided | withdraw's `grantedNotStarted` is true → clears the holder. Always worked. |
| **stand-down decided before the grant, applied after it** | patch carries no `claimedBy` → **the holder survives**. ← this issue |

`HolderStillHoldsIt` then defends that holder **by design** — the follower's process really is
alive (#1355: a stopped heartbeat on a running process means busy, not dead) — so with cluster
membership the claim is never taken over at all, and without it every later candidate waits out
`ClaimStaleAfter` and is handed the same dead claim again. No builder, no bake, no ready pod.

### Hypotheses excluded (so nobody re-derives them)

- ❌ *the follower stands down without withdrawing* — both doors of `FollowGo` call
  `WithdrawBuildClaim` first. Falsified on the Plugins issue in 2026-09.
- ❌ *withdraw clears the registration but not a granted claim* — it covers both
  (`grantedNotStarted`) and chains `ReleaseBuildClaim`.
- ❌ *a stale pending registration is the wedge* — the test's own comment says so and the state
  dump says the opposite: `RequestedClaims` is **EMPTY**. The registration was consumed, not left
  behind.
- ❌ *re-registration through `RetryUnreachableCoordination`* — `holder` is minted once, before the
  retry, so a retried registration overwrites its own key.
- ❌ *load is the cause* — load only widens the window. The failure has also been measured 2 in 15
  on a completely idle machine.

### The fix

The candidate cannot evaluate a condition the owner will apply, so it must state a **fact** the
owner reacts to — the `RequestedX`-plus-owner-watcher shape AGENTS.md prescribes for exactly this,
and `RequestedStatus` is the same idea one field over on this very record.

- `BuildState.StoodDown` — holders that have withdrawn, written **unconditionally** by
  `WithdrawBuildClaim` under its own key (merge-safe against every other candidate, and true
  whatever the owner's state turns out to be).
- `BuildNodeType.ReleaseStoodDownClaim` — the arbiter releases a `Planning` claim whose holder has
  stood down, and consumes the mark. A `Building` holder is never released: a mark from an earlier
  life of the same id must not pull the lock out from under a running compile.
- `ApplyGrant` also refuses a winner that carries a mark — the same refusal stated off the fact
  rather than off the absence of a registration.
- The durable path drops the claim LOCK when the mirror released, so the two records agree — the
  debris can land on either, per `HandBackAStoodDownGrant`'s own note.
- Marks are CONSUMED: on release, on re-registration (`RequestBuildClaim` clears its own key), and
  otherwise aged out on `ClaimStaleAfter`, so the durable row cannot accumulate one per process
  that ever followed a build.

No timer, no retry, no widened bound, no relaxed assertion: the release publishes on the mirror's
change feed, which is the arbiter's own trigger, so the next candidate is elected immediately.

### Coverage

`test/MeshWeaver.Graph.Test/BuildGrantPublicationTest.cs` — the interleaving is BUILT, never raced
for, in the same family as the rest of that file. It drives the **real** merge diff
(`MeshNodeStreamHandle.ComputeMergePatchDiff`) to show the patch cannot carry a `claimedBy` the
mirror has not seen, and carries the control that says why this needed a fix at all: the election
ALONE never frees the wedged state, however long anyone waits.

Closing MeshWeaver.Plugins#1193 needs that repo's `MW_PLATFORM_REF` to carry this commit — the test
lives there and runs against the pinned core.

rbuergi:
## A candidate's stand-down was DECIDED on its own mirror and APPLIED as a merge patch

`MeshWeaver.Plugins`' `BuildCoordinationTest.Follower_StandsDown_SoTheNextBuildCanStillBeClaimed`
has failed on three separate cores, **including ones that contain #3131** — the fix that was
supposed to close it. This is the remaining hole, and it is #3131's mirror image.

## The measurement

With the permanent state dump MeshWeaver.Plugins#1401 landed:

```
FAIL: ClaimedBy=<the follower>  Status=Planning  RequestedClaims=[]  Ready=[fp-stand-down]
PASS: ClaimedBy=<null>          Status=Planning  RequestedClaims=[]  Ready=[fp-stand-down]
```

**One emission in fifteen seconds** — terminal, not slow. Three of the four fields are identical,
the build had completed (`Ready` carries the fingerprint), and the whole difference is a holder
that was never cleared.

## The mechanism

`WithdrawBuildClaim` runs on the **candidate's** hub, which does not own `Admin/Build`. So its
`stream.Update` lambda is evaluated against *that hub's mirror* and travels to the owner as an RFC
7396 merge patch of the fields it changed — `MeshNodeStreamHandle.UpdateQueued` →
`ComputeMergePatchDiff`. **A field the lambda leaves unchanged is absent from the patch**, and by
RFC 7396 an absent member leaves the owner's value untouched.

Its second half is conditional on exactly such a field:

```csharp
var grantedNotStarted = state.ClaimedBy == holder && state.Status is BuildStatus.Planning;
```

On a mirror the grant has not reached yet that is **false**, so the patch carries the registration
removal and **no `claimedBy` at all** — and a grant the arbiter committed inside that window
survives the stand-down.

The arbiter, by contrast, writes the node it **owns** (`GetMeshNodeStream()` with no path →
`IsOwn` → `UpdateOwn`), so *its* lambda is serialised against fresh state. **That asymmetry is the
bug.** #3131's own words are *"conditional writes make each half safe on its own; they cannot order
two writers"*, and it fixed the ordering at `ApplyGrant` — the one point that could see a
stand-down that had **already** happened. It could not fix this one, because here the publication
was legitimate at the moment it happened: the candidate was still registered. What arrives late is
the stand-down.

| order | outcome |
|---|---|
| stand-down decided **and applied** before the grant | `ApplyGrant` refuses, `HandBackAStoodDownGrant` drops the lock. **Closed by #3131.** |
| grant applied, **then** stand-down decided | `grantedNotStarted` is true → clears the holder. Always worked. |
| **stand-down decided before the grant, applied after it** | patch carries no `claimedBy` → **the holder survives.** ← this PR |

`HolderStillHoldsIt` then defends that holder **by design** — the process really is alive (#1355) —
so with cluster membership the claim is never taken over at all, and without it every later
candidate waits out `ClaimStaleAfter` and is handed the same dead claim again. No builder, no bake,
no ready pod.

## The fix

A candidate cannot evaluate a condition the owner will apply, so it states a **fact** the owner
acts on — the `RequestedX`-plus-owner-watcher shape AGENTS.md prescribes for exactly this, and
`RequestedStatus` is the same idea one field over on this very record.

- **`BuildState.StoodDown`** — holders that withdrew, written **unconditionally** under their own
  key: merge-safe against every other candidate, and true whatever the owner's state turns out to
  be. It is the half that is always there to be read.
- **`BuildNodeType.ReleaseStoodDownClaim`** — the arbiter releases a `Planning` claim whose holder
  has stood down and consumes the mark. A **`Building`** holder is never released: a mark from an
  earlier life of the same id must not pull the lock out from under a running compile — the same
  line `WithdrawBuildClaim` already drew, and the same family as this file's existing
  `ARunningBakeIsNotMistakenForAStandDown`.
- **`ApplyGrant`** also refuses a marked winner — the same refusal stated off the fact rather than
  off the absence of a registration.
- **The durable path drops the claim LOCK** when the mirror released. The debris lands on *either*
  record (`HandBackAStoodDownGrant`'s own note), so fixing one would just move the wedge from the
  host that decides on the mirror to the one that decides on the lock.
- **Marks are consumed**, never accumulated: on release, on re-registration
  (`RequestBuildClaim` clears its own key), else aged out on `ClaimStaleAfter` — the claim's own
  budget.

**No timer, no retry, no widened bound, no relaxed assertion, no allow-file entry.** The release
publishes on the mirror's change feed, which is the arbiter's own trigger, so the next candidate is
elected immediately; `GrantOnMirror` releases and re-grants inside **one** serialised write.

The withdraw's decision moves to `BuildNodeType.StandDown`, beside `ApplyGrant` and `Arbitrate`, so
the three writers of a claim read as one set — and so a test can drive the **real** patch this
ships rather than a stand-in that would drift from it.

## Coverage

`test/MeshWeaver.Graph.Test/BuildGrantPublicationTest.cs`, in that file's existing family: the
interleaving is **built, never raced for**, with no wall-clock and no concurrency.

- `TheStandDownPatch_CannotCarryAClaimTheMirrorHasNotSeenYet` drives
  `MeshNodeStreamHandle.ComputeMergePatchDiff` — the real diff a cross-hub write ships — and shows
  the patch has no `claimedBy` member while `stoodDown` is unconditionally present.
- `TheElectionAlone_NeverFreesAStoodDownHolder` is **the control**: the pre-fix pass leaves the
  wedged state untouched, which is what "one emission in fifteen seconds" measured. Without it the
  rest would not show that a fix was needed at all.
- release + elect in one pass, release with nobody queued, a running bake never released, stale
  marks pruned, a quiet node returned unchanged (so running this every tick is free), and a grant
  refused to a marked winner.

**15 / 15 passed.** `MeshWeaver.Graph` and `MeshWeaver.Graph.Test` build `-c Release -warnaserror`
→ `0 Error(s)`; `MeshWeaver.Documentation` likewise, `DocumentationLinkIntegrityTest` passed.

## Hypotheses excluded, so nobody re-derives them

- ❌ *the follower stands down without withdrawing* — both doors of `FollowGo` call
  `WithdrawBuildClaim` first.
- ❌ *withdraw clears the registration but not a granted claim* — it covers both and chains
  `ReleaseBuildClaim`.
- ❌ *a stale pending registration is the wedge* — the test's own comment says so; the state dump
  says the opposite. `RequestedClaims` is **empty**.
- ❌ *re-registration through `RetryUnreachableCoordination`* — `holder` is minted once, before the
  retry.
- ❌ *load is the cause* — load only widens the window; the failure has also been measured 2 in 15
  on an idle machine.

## Surface

Additive only: a property on the `sealed record BuildState`, and three static methods on the static
`BuildNodeType`. **No public type or member leaves `src/`**, so no `Pairs-with:` is owed; **no
interface or abstract member is added**, so no `Implementers:` is owed.

## What this does not close by itself

The failing test lives in MeshWeaver.Plugins and runs against that repo's pinned core, so
MeshWeaver.Plugins#1193 closes when its `MW_PLATFORM_REF` carries this commit — measured today at
`aa4075832`, which predates it.

Closes #3567


rbuergi:
**Full-suite regression check, since this touches the claim protocol every host runs:** `MeshWeaver.Graph.Test` in full — **1399 / 1399 passed** (1 m 51 s), and `MeshWeaver.Hosting.Test` filtered to the build / pre-warm / witness families — **18 / 18 passed**. Both `-c Release`, both built `-warnaserror` with `0 Error(s) / 0 Warning(s)` first.

Recording it because the change adds a writer to `ClaimedBy`, and "the new tests pass" would not have said whether the existing arbitration rules still hold.

rbuergi:
Added `Doc/Architecture/ConditionalWritesAcrossHubs` (`9e51db003`), because the finding outlives the fix.

The defect is not specific to build coordination: **on a node this hub does not own, `stream.Update`'s lambda runs on the CALLER's mirror and ships a merge patch of the fields it changed**, so a field the lambda *decided not to write* is absent from the patch — and by RFC 7396 an absent member leaves the owner's value untouched. Every conditional write across a hub boundary has this shape, and it is invisible at the call site: no exception, no log, no failed write, and a symptom several layers away.

The page states the mechanic, the rule (*a lambda may write **values**; it may not decide from its own mirror that a field needs **no** write*), the replacement — state a fact under your own key and let the owning hub's watcher draw the conclusion, which is AGENTS.md's `RequestedX` prescription, with the three properties that make it safe: unconditional, own-key, **consumed** — and the four wrong answers, including the retry loop and the version compare-and-set that this protocol had already tried.

Indexed under "Reading & writing nodes" beside `Write Verdict Totality`, which is the same family of write-path surprise. `MeshWeaver.Documentation.Test` **304/304** — that is what checks every link resolves against the page's real node path, and it caught two of mine (a leaf page is a container, so a sibling needs `../`).

PR #3571, opened 2026-09-07
fix(#3558): the property overview asks the type registry instead of building the form over JsonElement

### LINKED ISSUE #3558: BuildPropertyOverview consults no type registry — an unregistered discriminator builds the form over JsonElement's own members
## `BuildPropertyOverview` is the one degrade seam that consults no type registry — it builds the form over `JsonElement`'s own members

Found while investigating MeshWeaver.Plugins#1279. Not that issue's cause (established there), but a
real defect with the exact signature AGENTS.md's payload rule warns about: **renders empty / renders
nonsense, no exception, nothing to grep.**

### The code

`src/MeshWeaver.Graph/OverviewLayoutArea.cs:33-37`:

```csharp
var instance = node.Content;
if (instance == null)
    return Controls.Stack;

if (instance is JsonElement je)
    instance = JsonSerializer.Deserialize<object>(je.GetRawText(), host.Hub.JsonSerializerOptions)!;

var contentType = instance.GetType();
```

`JsonSerializer.Deserialize<object>` runs through `ObjectPolymorphicConverter`, whose unregistered-type
branch (`src/MeshWeaver.Messaging.Hub/Serialization/ObjectPolymorphicConverter.cs`, `ReadObject`) is:

```csharp
// If no type discriminator or unknown type, return as JsonElement
return cleanedElement.Clone();
```

So when the discriminator is not registered on this hub, `instance` comes back a **`JsonElement` again**,
`contentType` becomes `typeof(JsonElement)`, and the property form is built over `JsonElement`'s own
members (`ValueKind`, …) instead of the content's. No exception, no warning from this call site.

### Why this one is different from its four siblings

There are five places a mesh node's content can be read while its discriminator is unregistered. **Four
of them consult the mesh-wide `IMeshContentTypeRegistry`; this one consults nothing:**

| seam | recovers? |
|---|---|
| `MeshNodeStreamExtensions.cs:599-746` `TypedContentObserver` | ✅ arms a wait on `Registrations` (`:667`) and re-emits typed |
| `ObjectPolymorphicConverter.ReadObject` | ✅ tries `contentTypeRegistry.TryRecover` before degrading |
| `MeshNodeTypeSource.ResolveJsonElementContent` (`:1089-1195`) | ⚠️ one-shot per emission, but warns explicitly (`:1149-1159`) |
| `MeshNodeStreamCache.DeserializeContent` (`:2665-2725`) | ⚠️ one-shot, but consults the registry |
| **`OverviewLayoutArea.BuildPropertyOverview` (`:33-37`)** | ❌ **nothing — no registry, no warning** |

### Why it matters

The registry's `TryRecover` exists precisely so a late registration is not fatal, and
`TypedContentObserver` proves the framework's intent is to recover rather than latch. This seam opts
out of both: it neither recovers nor reports, so a node whose type registers late (or never) renders a
form over the wrong type and the only symptom is a page that looks wrong.

It is also a direct instance of the AGENTS.md rule *"Never cast an `object` payload — `.As<T>()` /
`.ContentAs<T>()`, always"*: `instance.GetType()` on a value that may still be JSON is the same
trap-door in a different spelling.

### Suggested shape (not implemented)

Consult the registry the way the other seams do, and when the discriminator is genuinely unresolvable,
render an explicit "content type `X` is not registered on this hub" state rather than a form over
`JsonElement`. Silently building the wrong form is the outcome worth removing; which of the two
recoveries is right (a `TryRecover` attempt, or the `Registrations` wait `TypedContentObserver` uses) is
a judgement call I would rather leave to whoever owns this area.

Filing rather than fixing: it is tangential to the investigation that found it, and the "what should an
unresolvable type render?" half is a design decision, not a defect fix.

Related: MeshWeaver.Plugins#1279 (where this was found; **not** its cause).

<sub>Read-only, 2026-09-07, against `main`. Nothing edited or executed.</sub>

rbuergi:
## The defect

`OverviewLayoutArea.BuildPropertyOverview` (`src/MeshWeaver.Graph/OverviewLayoutArea.cs:29-37`) read a node's content like this:

```csharp
if (instance is JsonElement je)
    instance = JsonSerializer.Deserialize<object>(je.GetRawText(), host.Hub.JsonSerializerOptions)!;
var contentType = instance.GetType();
```

`ObjectPolymorphicConverter`'s unregistered-type branch returns `cleanedElement.Clone()` — **a `JsonElement` again**. So `contentType` became `typeof(JsonElement)` and the property form was built over `JsonElement`'s own members (`ValueKind`, …). No exception, no warning from this call site, nothing to grep: the page looked finished and described nothing.

Of the five seams that can read a node's content while its discriminator is unregistered, four consult the mesh-wide `IMeshContentTypeRegistry`. This one consulted nothing.

## The fix — two halves

**1. Ask the EXACT route.** This seam holds the NODE, so it can ask `TryRecoverForNodeType(node.NodeType, …)` — the question that always has one right answer. The wire converter can only ask the NAME route (`TryRecover`), which must REFUSE a discriminator two declarations both claim (#1299), so a node whose type is contested resolved nowhere before. `TryRecoverForNodeType` falls back to the name route itself: strictly more recovery, never less.

**2. When nothing resolves, say so.** The notice names the discriminator and separates the two indistinguishable causes — an in-mesh NodeType whose Roslyn compile has not registered it YET (transient; `MeshNodeStreamHandle`'s typed read boundary waits on `IMeshContentTypeRegistry.Registrations` and re-emits, which re-renders this area) from a name no declaration will ever claim. Free-form JSON content, legal by design, gets its own note because there is nothing to wait for there. Both strings are catalog keys in `strings.en.json` + `strings.de.json`, read through `host.Localize`.

No new recovery machinery: the `Registrations` wait already exists one layer down, so adding a second one here would have been a duplicate watcher, not a fix.

## Measurement

`OverviewUnresolvedContentTypeTest` — 4 cases, run at `-c Release`:

| | with the fix | on `origin/main` |
|---|---|---|
| unresolvable `$type` ⇒ notice naming it, never `ValueKind` | pass | **fail** (`StackControl`) |
| no `$type` at all ⇒ its own note | pass | **fail** (`StackControl`) |
| contested `$type` + NodeType ⇒ the real form (non-regression control) | pass | pass |
| contested `$type`, no NodeType ⇒ notice | pass | **fail** (`StackControl`) |

`Passed! - Failed: 0, Passed: 4` with the fix; `Failed: 2, Passed: 1` on main before the fourth case was added, then 3 of 4 failing with it. `LocalizationTest`: `Failed: 0, Passed: 58`. `dotnet build -c Release -warnaserror` on `MeshWeaver.Graph` and `MeshWeaver.Graph.Test`: `0 Error(s)`.

## Follow-up (not in this PR)

Two catalog keys are added here, so the React mirror in `MeshWeaver.Plugins/clients/react/src/i18n/` needs `npm run sync:i18n -- --ref <this merge sha>` after it lands. The mirror's guard compares against a pinned core commit, so nothing reds until that sync.

Pairs-with: none — no public type or member is removed; the only surface change is one new private helper.

Closes #3558


rbuergi:
Both review findings are addressed in **#3584** — they could not land here because the merge queue refuses a push to a queued branch, and dequeuing restarts the builds of the whole group for other people's PRs.

- The `LogWarning` fired for content with **no `$type` at all** — free-form JSON, which `ContentDiscriminatorValidator` declares legal. There is no name that "resolves nowhere" in that case, so the line accused a legal, permanent shape once per render. The two states were already separated in the rendered notice; they are separated in the log now too.
- `JsonDocument.Parse(...).RootElement.Clone()` never disposed the document. `Clone()` is what makes the element independent of it, so it can be released immediately: `using var`.

Verified on merged main: both projects `0 Error(s)` at `-warnaserror`, `Passed! - Failed: 0, Passed: 14`.

PR #3572, opened 2026-09-07
fix(#3548): the migration Job mounts the SecretProviderClass it reads, so its vault values are current

### LINKED ISSUE #3548: Migration Job reads the CSI-synced secret before the driver rotates it — a deploy that changes an SPC mapping runs the embedding backfill with the previous key (1,260 × 401, green)
## What happened

`memex` release, 2026-09-07, migration job `memex-migration-33` (ci.7981), right after a deploy that changed the SecretProviderClass mapping of `Embedding__ApiKey` from the Azure object to the OpenRouter object:

```
[DocBackfill] 1274 documentation pages found (embeddings ON)
EmbeddingRequestException: The embeddings backend at 'https://openrouter.ai/api/v1/embeddings'
  refused the request for model 'openai/text-embedding-3-small' with HTTP 401 (Unauthorized)
  — the bearer was rejected …                                        × 1260
[DocBackfill] done: 1274 upserted (0 embedded), 0 unchanged, 1274 total
```

The key in the vault is correct: a portal pod that started a few minutes later holds the same object under `Embedding__ApiKey` and gets **200** (sha-compared against the chat key — identical). The Job simply read the synced Secret **before** the CSI driver had replaced the previous value, so it presented the old Azure key to OpenRouter for its whole run. `memex-cloud`, whose synced Secret had no prior `Embedding__ApiKey`, embedded everything in the same deploy (0 refusals).

## Why it is structural, not a one-off

#3468 gave the migration Job `envFrom` on the CSI-synced Secret (`memex-portal-keyvault`, `optional: true`). That Secret is materialised and rotated by the driver **for the pods that mount the SPC volume** — the portal pods — on their rotation poll. The Job mounts nothing, starts the instant `helm upgrade` applies, and `envFrom` is resolved once at pod start. So on any deploy that changes an SPC mapping, the Job runs with the value from *before* the change, and the backfill logs-and-skips per row: 1,260 warnings, `Database migration completed`, green.

## Fix

Mount the SPC CSI volume in the migration Job itself (same `volumeName`/`mountPath` as the portal, from `memex.keyVaultClasses`). A CSI mount forces a fresh fetch from the vault at pod start, so the Job gets the current objects by construction, independent of the portal pods' rotation timing. Keep the `envFrom` (the projected Secret is still how the values become env); the mount is what makes it current.

Until then: after any deploy that changes an SPC mapping, the backfill needs a second `helm upgrade` (a new Job) to embed. Measured today.

rbuergi:
## Which half, and where it lives — measured first

The issue asked whether the chart/Job half is here or in the private `Systemorph/Memex` repo. **It is here**: `deploy/helm/templates/memex-migration/job.yaml` is the only definition of that workload, and `Memex` carries per-environment *overlays* (`deployments/aks/<ns>/values.*.public.yaml`, `portal-patch.json`), not the template. `chart-drift.yml` compares the live ConfigMap and portal Deployment against this chart, so nothing in Memex needs to move for this fix. No `az`, `kubectl` or `helm upgrade` was run — this is a template change, verified with `helm template` / `helm lint` locally.

## The defect

The Secrets Store CSI driver materialises and rotates the synced Kubernetes Secret **for the pods that MOUNT the SecretProviderClass volume**. #3468 gave the migration Job `envFrom` on that Secret (`optional: true`) but no mount, which made it a **free rider**: it reads whatever the portal pods' mount last wrote, and `envFrom` is resolved once, at container start. The Job starts the instant `helm upgrade` applies — before the portal pods have rolled — so on any deploy that *changes* an SPC mapping it runs with the value from before the change.

Verified against the issue's own measurement (`memex` release 2026-09-07, job `memex-migration-33`, the deploy that repointed `Embedding__ApiKey` from the Azure object to the OpenRouter one):

```
[DocBackfill] 1274 documentation pages found (embeddings ON)
EmbeddingRequestException: … HTTP 401 (Unauthorized) — the bearer was rejected …   × 1260
[DocBackfill] done: 1274 upserted (0 embedded), 0 unchanged, 1274 total
Database migration completed.
```

**The green is the defect's signature.** `MeshNodeEmbeddingBackfill` logs-and-skips per row rather than throwing, so the Job's exit code says nothing about whether it had a working key — a check that cannot fail. A portal pod that started minutes later held the same key name and got 200s, so the vault was right and only the Job's copy was stale. `memex-cloud`, whose synced Secret had no prior `Embedding__ApiKey`, embedded everything in the same deploy.

## The fix — structural, not a retry

The Job now renders the CSI volume + `volumeMount` from the **same** `memex.keyVaultClasses` block that renders its `envFrom`, so the class name, the synced-Secret name, the volume name and the mount path cannot disagree. A CSI mount is set up before *any* container in the pod starts and `NodePublishVolume` fetches from the vault at that moment, so the kubelet resolves this container's `envFrom` against a freshly written Secret **by construction** — no second `helm upgrade`, no rotation-poll timing, nothing to retry.

Deliberately not done: a retry loop, a wait-for-key init container, or a "re-run the backfill later" job — each would leave the stale read in place and paper over it.

One consequence, documented in `values.yaml` and `DeploymentAKS.md`: a declared vault object that does not exist now stalls the migration too, exactly as it already stalls the portal. That is the correct direction — loudly stuck beats green-and-wrong.

## The gate

`KeyVaultCsiFreshnessGuard` generalises the invariant: **every pod-bearing chart template that reads `$class.syncedSecret` must also carry `secrets-store.csi.k8s.io` and `$class.mountPath`.** `secretproviderclass.yaml` names `syncedSecret` legitimately (it is the declaration) and is excluded by the `kind:` filter, not by an allow-list. The guard asserts its own denominator — the templates directory is non-empty, `_keyvault.tpl` still exists, and at least two workloads read a class — so it cannot pass having examined nothing.

## Measurement

| | result |
|---|---|
| `helm template` with two declared classes | Job renders `volumeMounts` (2) + `volumes` (2, CSI) — see below |
| `helm template` with chart defaults (no Key Vault) | byte-identical to before: no `volumeMounts`, no `volumes` |
| `helm lint` | `1 chart(s) linted, 0 chart(s) failed` |
| guard **with** the fix | `Passed! - Failed: 0, Passed: 1` |
| guard **with the mount block removed** (negative control) | `Failed: 1` — *"these workloads read a Key Vault CSI-synced Secret through envFrom but do not mount the SecretProviderClass volume that feeds it: deploy/helm/templates/memex-migration/job.yaml"* |
| `MigrationWorkloadModelGuard` + `DocumentationLinkIntegrityTest` + What's New | `Passed! - Failed: 0, Passed: 6` |

```yaml
          imagePullPolicy: "IfNotPresent"
          volumeMounts:
            - name: "kv-secrets"
              mountPath: "/mnt/secrets-store"
              readOnly: true
      volumes:
        - name: "kv-secrets"
          csi:
            driver: "secrets-store.csi.k8s.io"
            readOnly: true
            volumeAttributes:
              secretProviderClass: "memex-portal-keyvault"
```

Pairs-with: none — a Helm template, a values comment, a doc page and a new guard; no public surface changes.

Closes #3548

### DIFF
--- a/deploy/helm/templates/memex-migration/job.yaml
+++ b/deploy/helm/templates/memex-migration/job.yaml
@@ -67,10 +67,64 @@ spec:
                    pod has started yet, and a missing non-optional secretRef would keep the Job
                    from starting at all. Skipping is safe there and only there: a first install has
                    no pre-existing rows to backfill, and the run still says so out loud via
-                   `[EmbeddingBackfill] … skipped` plus EmbeddingCapabilityReporter. */}}
+                   `[EmbeddingBackfill] … skipped` plus EmbeddingCapabilityReporter.
+
+                   🚨 The envFrom alone gives the Job a value, not a CURRENT value — which is why
+                   the volumeMounts below are not optional decoration. See their comment. */}}
             {{- range $class := (include "memex.keyVaultClasses" . | fromYamlArray) }}
             - secretRef:
                 name: "{{ $class.syncedSecret }}"
                 optional: true
             {{- end }}
           imagePullPolicy: "IfNotPresent"
+          {{- /* 🚨 THE JOB MOUNTS THE SPC ITSELF — this is what makes the values above CURRENT
+                 (MeshWeaver#3548).
+
+                 The synced Secret the envFrom reads is materialised and rotated by the CSI driver
+                 FOR THE PODS THAT MOUNT THE SecretProviderClass VOLUME. Until this block existed
+                 that was the portal pods and nobody else: the Job mounted nothing, started the
+                 instant `helm upgrade` applied, and `envFrom` is resolved once at container start.
+                 So on any deploy that CHANGES an SPC mapping, the Job read the value from BEFORE
+                 the change — and the backfill logs-and-skips per row rather than failing.
+
+                 Measured on the `memex` release 2026-09-07, job memex-migration-33, the deploy that
+                 repointed `Embedding__ApiKey` from the Azure object to the OpenRouter one:
+                     [DocBackfill] 1274 documentation pages found (embeddings ON)
+                     EmbeddingRequestException: … HTTP 401 (Unauthorized) — the bearer was
+                       rejected …                                              × 1260
+                     [DocBackfill] done: 1274 upserted (0 embedded), 0 unchanged, 1274 total
+                     Database migration completed.
+                 1,260 refusals and a GREEN Job. A portal pod that started minutes later held the
+                 SAME key name and got 200s, so the vault was right and the Job's copy was stale.
+
+                 A CSI mount is set up before ANY container in the pod starts, and NodePublishVolume
+                 fetches the objects from the vault at that moment — so the driver refreshes the
+                 synced Secret before the kubelet resolves this container's envFrom. The Job then
+                 gets the current objects BY CONSTRUCTION, independent of the portal pods' rotation
+                 poll. That is the whole point: not a retry, not a second `helm upgrade`, an
+                 ordering that cannot be lost.
+
+                 Same volumeName/mountPath as the portal, from the one `memex.keyVaultClasses`
+                 block, so the class, the volume, the mount and the envFrom can never disagree on a
+                 name. Rendered only when a class is declared — a namespace with no Key Vault gets
+                 the pod spec it had before. */}}
+          {{- $keyVaultClasses := (include "memex.keyVaultClasses" . | fromYamlArray) }}
+          {{- if $keyVaultClasses }}
+          volumeMounts:
+            {{- range $class := $keyVaultClasses }}
+            - name: "{{ $class.volumeName }}"
+              mountPath: "{{ $class.mountPath }}"
+              readOnly: true
+            {{- end }}
+          {{- end }}
+      {{- if $keyVaultClasses }}
+      volumes:
+        {{- range $class := $keyVaultClasses }}
+        - name: "{{ $class.volumeName }}"
+          csi:
+            driver: "secrets-store.csi.k8s.io"
+            readOnly: true
+            volumeAttributes:
+              secretProviderClass: "{{ $class.name }}"
+        {{- end }}
+      {{- end }}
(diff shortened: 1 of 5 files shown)


rbuergi:
**A boundary this fix deliberately does not cross, measured while writing it.**

The Job's `envFrom` renders the chart's own ConfigMap + Secret and then one entry per `memex.keyVaultClasses` class. It renders **no** `.Values.extraEnvFrom` — grep the template: zero occurrences. So on an environment still using the LEGACY hand-made SecretProviderClass shape (`extraEnvFrom` + `extraVolumes` + `extraVolumeMounts`, which is what `memex-cloud`'s AI provider keys ride on), the migration Job receives none of those values at all, before or after this change.

That is a **separate** gap and it is not fixed here, on purpose:

* `extraEnvFrom` and `extraVolumes` are verbatim Kubernetes objects supplied by an operator, with no declared correspondence between an env source and the volume that feeds it — the chart cannot pair them, which is exactly the disagreement `memex.keyVaultClasses` exists to make impossible. Mounting every `extraVolume` in the migration pod would mount volumes that have nothing to do with it.
* `KeyVaultCsiFreshnessGuard` therefore cannot see that shape either: it keys on `$class.syncedSecret`, the chart-owned wiring. It is a guard over the wiring the chart owns, and it says so.

The durable cure for the legacy shape is the one the values file already recommends — declare the secrets under `keyVaultSecrets` / `keyVaultSecretClasses` so one block renders the SecretProviderClass, the volume, the mount and the `envFrom` — at which point this fix and its guard cover it automatically. Filing this here rather than widening the change: it needs a per-environment migration, not a template edit.

rbuergi:
The review finding is addressed in **#3585** — it could not land here because the merge queue refuses a push to a queued branch, and dequeuing restarts the builds of the whole group for other people's PRs.

`ExecutableLinesOf` stripped Helm comment blocks and whole-line `#` comments but not **trailing** ones, so `name: "kv-secrets"   # {{ $class.mountPath }}` stayed in the scan — a template could have satisfied the mount half of the invariant *from a comment* while its real mount was gone. Each line is now cut at its first whitespace-preceded `#`.

Re-falsified rather than assumed, since changing the masker invalidates the earlier negative control: green with the Job's mount block, **red without it** naming `deploy/helm/templates/memex-migration/job.yaml`; `0 Error(s)` at `-warnaserror` on merged main; `Passed: 4` alongside `MigrationWorkloadModelGuard`.

PR #3573, opened 2026-09-07
ci(#3568): the platform-pin lane asks for permission to write the file it edits

### LINKED ISSUE #3568: The platform-ref bump lane can never move a pin: its App token omits the workflows permission, and the pin lives in ci.yml
`node-repo-platform-ref-bump.yml` mints an App token **without** the `workflows` permission, and the
only file it ever edits is `.github/workflows/ci.yml`. GitHub refuses that push outright, so the lane
has **never once moved a pin** — in either repo that calls it.

## Measured, 2026-09-07

Both scheduled runs this morning failed at the same step (`Open (or update) the bump PR`) with the
same line:

```
! [remote rejected] chore/platform-ref-c7fef7a2d -> chore/platform-ref-c7fef7a2d
  (refusing to allow a GitHub App to create or update workflow `.github/workflows/ci.yml`
   without `workflows` permission)
error: failed to push some refs to 'https://github.com/Systemorph/MeshWeaver.Plugins'
##[error]Process completed with exit code 1.
```

| caller | run | when | conclusion |
|---|---|---|---|
| `MeshWeaver.Plugins` | [34083504064](https://github.com/Systemorph/MeshWeaver.Plugins/actions/runs/34083504064) | 04:32:51Z | failure — `remote rejected … without workflows permission` |
| `MeshWeaver.SocialMedia` | (same workflow) | 04:51:24Z | failure — identical line, branch `chore/platform-ref-37f431a8c` |

Those are the **only** runs the API lists for either caller, so this is not a regression — the lane
has been in this state since the callers were added.

## The cause is one omission, and it is in the lane, not in the callers

```yaml
# .github/workflows/node-repo-platform-ref-bump.yml
- name: Mint an installation token whose pull requests actually start CI
  uses: actions/create-github-app-token@v3
  with:
    …
    # Requested explicitly so an installation missing either fails HERE, on a step whose name
    # says what it wanted, rather than degrading to a token that pushes fine and then opens a
    # pull request nothing runs on.
    permission-contents: write
    permission-pull-requests: write
```

The comment states the design intent exactly — *request explicitly so a missing grant fails on a step
whose name says what it wanted* — and then omits the one permission this lane structurally requires.
`contents: write` is enough to push a branch; it is **not** enough to push a branch that touches
`.github/workflows/**`, which is the lane's entire diff. So it fails at the last step, three minutes
in, with a message about a permission nobody requested.

## Why it stayed invisible

The lane's own header argues at length that it must never `continue-on-error`, and it does not — it
goes red honestly. But it runs at 04:17Z on a schedule, against nobody's branch, on satellite repos.
Nothing surfaces a scheduled red there, so *"the pin is not being maintained"* — the exact statement
the lane was written to make — has been made daily into an empty room.

The cost is the failure mode the lane's header describes as its reason for existing. As of writing,
`MeshWeaver.Plugins`' `MW_PLATFORM_REF` is `aa4075832` (2026-09-06 20:27Z), **117 commits behind
core `main`** — inside the staleness gate's bounds, and therefore silent, but only just.

## Fix

1. Add `permission-workflows: write` to the token mint in
   `.github/workflows/node-repo-platform-ref-bump.yml`.
2. 🚨 That is necessary but may not be sufficient: the **installation** must actually hold
   *Workflows: Read and write*. If it does not, `create-github-app-token` fails at the mint — which
   is the good failure this step's comment is asking for, naming what to provision, instead of a
   push rejection four steps later. Granting it is an org-admin act and cannot be done from CI.
3. Worth considering as the durable alternative: keep the pin somewhere that is **not** a workflow
   file (a `platform-ref.json` the workflow reads), so the bump lane never needs a permission that
   also lets it rewrite CI. That removes the blast radius rather than widening the grant.

Found while landing Systemorph/MeshWeaver#3516, whose Plugins half
(Systemorph/MeshWeaver.Plugins#1455) is waiting on exactly this pin move.

rbuergi:
## The lane's entire diff is a workflow file, and its token could not push one

`node-repo-platform-ref-bump.yml` mints an App token with `contents: write` and
`pull-requests: write`. The **only** file it ever edits is the caller's own
`.github/workflows/ci.yml` — `workflow-file`'s default, because the pin lives in the workflow's own
`env:`. GitHub refuses an App push touching `.github/workflows/**` without the `workflows` grant,
so **the lane has never once moved a pin**, in either repo that calls it.

## Measured, 2026-09-07

```
! [remote rejected] chore/platform-ref-c7fef7a2d -> chore/platform-ref-c7fef7a2d
  (refusing to allow a GitHub App to create or update workflow
   `.github/workflows/ci.yml` without `workflows` permission)
error: failed to push some refs to 'https://github.com/Systemorph/MeshWeaver.Plugins'
```

| caller | run | when | conclusion |
|---|---|---|---|
| `MeshWeaver.Plugins` | 34083504064 | 04:32:51Z | failure — the line above |
| `MeshWeaver.SocialMedia` | — | 04:51:24Z | failure — identical, branch `chore/platform-ref-37f431a8c` |

Those are the **only** runs the API lists for either caller, so this is the lane's state since the
callers were added — not a regression.

## 🚨 Why it stayed invisible, which is the interesting half

The lane is right about everything else. Its header argues at length that it must never
`continue-on-error` — *"a red here is the actionable, once-per-day statement that the pin is not
being maintained"* — and it does not. **It went red honestly, every night.** At 04:17Z, on a
schedule, against nobody's branch, on satellite repos, where nothing surfaces a scheduled red.

So the statement the lane exists to make has been made daily into an empty room. The cost is the
exact failure mode its header names as its reason for existing: `MeshWeaver.Plugins`'
`MW_PLATFORM_REF` is `aa4075832`, **117 commits behind core `main`** — inside the staleness gate's
bounds, and therefore silent, but only just. Three separate in-flight changes are currently waiting
on a pin move that was supposed to be automatic.

This is worth stating plainly because it is not the usual shape: **no gate was skipped, nothing was
tolerated, no `if:` swallowed anything.** The gate fired correctly and nobody was in the room. A
red that nothing is attached to is a different failure from a red that is suppressed, and it needs
a different remedy.

## The fix

One line, and it is the discipline the step's own comment already states:

```yaml
# Requested explicitly so an installation missing any of them fails HERE, on a step whose
# name says what it wanted, rather than degrading to a token that pushes fine and then
# opens a pull request nothing runs on.
permission-contents: write
permission-pull-requests: write
permission-workflows: write     # ← this lane's entire diff is under .github/workflows/**
```

The comment had the intent exactly right and then omitted the one permission the lane structurally
requires.

## What this does NOT do, stated up front

**Asking for a permission is not having it.** `permission-workflows` on
`actions/create-github-app-token` narrows the token to what the *installation* already grants; if
the installation does not carry `workflows`, this step now fails at the **mint** rather than at the
push.

That is strictly better and is the outcome the step's own comment asks for — the failure is
reported by the step whose name says what it wanted, on the first thing that runs, rather than
three minutes later with a message about a permission nobody requested. But it is not the same as
the pin moving, so it should not be read as one. **A red at the mint step is a provisioning task,
not a workflow bug**, and the header now says so with the exact settings path, because nothing in
CI can grant this and nothing in CI can read whether it is granted.

## No caller change

The push and the PR are both the App's — the checkout carries the minted token, and every later
step passes `GH_TOKEN: ${{ steps.bump-token.outputs.token }}`. The callers' job-level
`permissions:` block scopes `GITHUB_TOKEN`, which does none of this, so the documented caller
contract is unchanged.

## Verification

| | |
|---|---|
| `.github/scripts/check-workflow-timeouts.py` | 71 jobs checked, 2 reusable-call exempt, **0 violations**, cap 45 min |
| workflow parses | `yaml.safe_load` OK; the `bump` job keeps its literal `timeout-minutes: 10` |
| public surface | none — a workflow and a What's New entry, so no `Pairs-with:` and no `Implementers:` is owed |

Closes #3568

### DIFF
--- a/.github/workflows/node-repo-platform-ref-bump.yml
+++ b/.github/workflows/node-repo-platform-ref-bump.yml
@@ -33,6 +33,15 @@ name: Node repo platform-ref bump (reusable)
 # above, silently, on the day the App loses a grant. `ArmedMergeMustTriggerMainsPushLanesGuard`
 # fails core's build if anyone puts it back.
 #
+# 🚨 THE APP INSTALLATION MUST GRANT `workflows`, not just `contents` and `pull_requests` (#3568).
+# The pin lives in the calling repo's OWN workflow file (`workflow-file`, default
+# `.github/workflows/ci.yml`), so this lane's entire diff is under `.github/workflows/**`, and
+# GitHub refuses an App push that touches that path without the grant — by name, at the push, on
+# the last step. The mint below asks for it explicitly so a missing grant is reported by the step
+# that wanted it. Provisioning it is a repository-settings act (Settings → GitHub Apps → the
+# installation → Repository permissions → Workflows: Read and write); nothing in CI can grant it,
+# and nothing in CI can read whether it is granted.
+#
 # Caller contract:
 #
 #   name: Bump the platform pin
@@ -140,11 +149,36 @@ jobs:
           private-key: ${{ secrets.MESHWEAVER_APP_PRIVATE_KEY }}
           owner: ${{ github.repository_owner }}
           repositories: ${{ github.event.repository.name }}
-          # Requested explicitly so an installation missing either fails HERE, on a step whose name
-          # says what it wanted, rather than degrading to a token that pushes fine and then opens a
-          # pull request nothing runs on.
+          # Requested explicitly so an installation missing any of them fails HERE, on a step whose
+          # name says what it wanted, rather than degrading to a token that pushes fine and then
+          # opens a pull request nothing runs on.
           permission-contents: write
           permission-pull-requests: write
+          # 🚨 THIS LANE'S ENTIRE DIFF IS A WORKFLOW FILE, so `contents: write` alone cannot push it
+          # (#3568). `workflow-file` defaults to `.github/workflows/ci.yml` — the pin lives in the
+          # workflow's own `env:` — and GitHub refuses an App push that touches
+          # `.github/workflows/**` without this grant, by name:
+          #
+          #   ! [remote rejected] chore/platform-ref-<sha> -> chore/platform-ref-<sha>
+          #     (refusing to allow a GitHub App to create or update workflow
+          #      `.github/workflows/ci.yml` without `workflows` permission)
+          #
+          # Measured 2026-09-07: that is EVERY run this lane has ever had — MeshWeaver.Plugins
+          # 34083504064 at 04:32:51Z and MeshWeaver.SocialMedia at 04:51:24Z, the only two runs the
+          # API lists for either caller. So the lane has never once moved a pin, and it failed at
+          # the LAST step, three minutes in, naming a permission nobody had asked for.
+          #
+          # It stayed invisible because the lane is right about everything else: it refuses to
+          # `continue-on-error`, so it went red honestly — at 04:17Z, on a schedule, against
+          # nobody's branch, where nothing surfaces a scheduled red. "The pin is not being
+          # maintained" is exactly the statement this lane exists to make, and it made it daily
+          # into an empty room while MeshWeaver.Plugins' MW_PLATFORM_REF drifted 117 commits behind
+          # core main — inside the staleness gate's bounds, and therefore silent.
+          #
+          # Requesting it here is the same discipline the two lines above already state: if the
+          # installation lacks the grant, this step fails with the permission in its own name,
+          # instead of a push failing later with a message about something nobody requested.
+          permission-workflows: write
 
       # The checkout carries the minted token, so the later `git push` is the App's push. With the
       # default credential the push would be GITHUB_TOKEN's, and the branch would exist with no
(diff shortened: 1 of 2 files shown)


rbuergi:
**Checked the thing that would have falsified this:** `permission-workflows` is a real input on `actions/create-github-app-token@v3`, not a name I inferred from the GitHub permission key.

```
$ gh api repos/actions/create-github-app-token/contents/action.yml --jq .content | base64 -d | grep -nE '^  permission-(workflows|contents|pull-requests):'
49:  permission-contents:
119:  permission-pull-requests:
143:  permission-workflows:
```

Worth doing rather than assuming: the action ignores unknown `with:` keys, so a mistyped input would have shipped as a **silent no-op** — a one-line fix that changes nothing, on a lane whose failure nobody watches. That is the same class as the omission it is fixing, and it would have looked identical from here (green PR, workflow parses, gate passes) until the next 04:17Z run failed exactly as before.

PR #3577, opened 2026-09-07
fix(#3554): a declared module floor is checked against the platform the bundle is built against

### LINKED ISSUE #3554: A declared minMeshVersion floor is never checked against a platform build that actually exists — an rc floor is unsatisfiable for ever
## The gap

A module declares `content.minMeshVersion` — a platform FLOOR — and `ModulePlatformFloor.DeclineReason`
refuses to land the module until the running platform satisfies it. Nothing checks that the declared
floor is satisfiable by a platform build **that actually exists**, so a floor can be declared that no
published image can ever meet. The module then holds for ever, and the hold message names two versions
that look like they are in the right order.

## Measured, 2026-09-07

`ModulePlatformFloor` compares with `NuGetVersionComparer` (SemVer). SemVer §11.4 compares pre-release
identifiers as **text**, so `"ci"` < `"rc"`. Against the real comparer (`Compare(running, floor) < 0`
⇒ DECLINED):

| running \ floor | `3.0.0-rc4` | `3.0.0-rc8` | `3.0.0-rc9` | `3.0.0` |
|---|---|---|---|---|
| `3.0.0-ci.1` | DECLINED | DECLINED | DECLINED | DECLINED |
| `3.0.0-ci.7989` | DECLINED | DECLINED | DECLINED | DECLINED |
| `3.0.0-ci.999999999` | DECLINED | DECLINED | DECLINED | DECLINED |
| `3.0.0` | ok | ok | ok | ok |
| `3.1.0-ci.1` | ok | ok | ok | ok |

Observed on both AKS portals at ~08:20Z while they ran `3.0.0-ci.7981`:

```
HOLDING 3.0.0-ci.7989 — AI: the module requires platform 3.0.0-rc8 or newer
                         but this deployment runs 3.0.0-ci.7989
```

And the registry has nothing that could satisfy those floors:
`az acr repository show-tags -n meshweaver --repository memex-portal-ai` → **1268 tags — 798
`staging-*`, 48 `3.0.0-ci.*`, ZERO `3.1.0-*`, ZERO `rc*`.** Every `rc` image is untagged, so an `rc`
floor is unsatisfiable in the strongest sense: there is no such build to run.

The values themselves are being corrected in Systemorph/MeshWeaver.Plugins#1447. This issue is about
the *class*: nothing stops the next one.

## Why it belongs on a CORE lane, not the Plugins gate

The obvious home — the Plugins module gate — deliberately cannot host it. That gate runs on forks and
on Dependabot, where reading ACR needs a credential, and gating the check on whether that credential
is present is exactly the skip-trapdoor AGENTS.md forbids: GitHub paints a skipped job the same colour
as a passed one, so "the gate never ran" and "the gate passed" become indistinguishable.

A core lane that **already** holds ACR credentials — `main-cd`'s bake — needs no such exemption. The
check is cheap there: list the portal repository once (already done for other steps), and assert that
every `minMeshVersion` declared by a package in the publication is satisfied by at least one listed
platform tag. Fail RED naming the package, the floor and the newest tag that does not reach it.

## Related

- MeshWeaver#3542 / #3551 — the sibling defect at the *other* call site: the self-updater ordered
  candidates by SemVer, so a mislabelled `3.1.0-ci.7841` outranked every later sealed `3.0.0-ci.79xx`
  set for ever. Fixed by ranking on the sealed-publication lineage. The shared root is that neither
  call site may compare `ci.<n>` against `rc<m>` by SemVer.
- `Doc/Architecture/SelfUpdateTargetSelection` §4 — why "newer than" and "satisfies the floor of" are
  **different predicates** over the same key, so the floor cannot simply reuse the updater's
  comparison: for an updater `3.0.0` must outrank `3.0.0-ci.7977`; for a floor the same two strings
  must give the opposite answer.
- #3552 does not cover this: its link probe reads
  `ModulePlatformFloor.DeclineReason(minMeshVersion) ?? LinkHoldReason()`, so it short-circuits
  *behind* the declared floor and never runs unless that already passes.

rbuergi:
## First, a measurement that moves the fix

The issue proposed `main-cd`'s bake lane. **Measured: core declares no module floors at all** — zero `index.json` under this repo carries `content.minMeshVersion` (`grep -rln minMeshVersion --include=index.json .` → nothing). A gate there would iterate an empty set and pass vacuously, for ever. The floors live in the satellites, and the credentialed core-owned lane that already *reads* one is `node-repo-module-pack.yml`'s `pack` job (`.../node-repo-module-pack.yml:1331`, `floor=$(jq -er '.content.minMeshVersion' "$PACKAGE/index.json")`). That is where the gate goes.

## The defect

`ModulePlatformFloor.DeclineReason` decides at RUNTIME whether the platform a deployment happens to run satisfies a module's declared floor. It answers *"not yet"* — and from inside one process it cannot distinguish that from *"never"*. So a floor no published image can meet produces a hold indistinguishable from one waiting on the next release, for ever, with nothing red anywhere.

Verified against the issue's measurement (both AKS portals, 2026-09-07, running `3.0.0-ci.7989`):

```
HOLDING 3.0.0-ci.7989 — AI: the module requires platform 3.0.0-rc8 or newer
                         but this deployment runs 3.0.0-ci.7989
```

Two versions that read as though they are in the right order. SemVer §11.4 compares pre-release identifiers as **text**, so `"ci" < "rc"` and `3.0.0-ci.<n>` is below `3.0.0-rc8` for **every** `n`; a clean `3.0.0` floor outranks every pre-release of `3.0.0` the same way; and one package named `3.0.0-rc14`, a platform that never existed. 42 packages, and `memex-portal-ai` held **zero** `rc*` tags — no image could have rescued any of them.

## The check

`pack` now asserts the declared floor against the platform the bundle is being **compiled against**, read as that image's own `MESHWEAVER_PLATFORM_VERSION` — the string a running portal identifies itself by, i.e. the exact value `ModulePlatformFloor` compares at landing time. `prepare` publishes it as a job output from the image it has **already staged**, so the reading is one `docker image inspect`, not a registry trip.

That predicate is deliberately stronger than "some tag satisfies this", and needs no listing: a bundle records the framework identity it was built against and a consumer only adopts a bundle whose identity matches its own platform, so **a floor above the build platform is unsatisfiable by construction** — no deployment can both satisfy it and accept the bundle.

Two properties make it a gate rather than a comment:

- **It cannot skip.** The step is unconditional, and an unknown platform version is RED, naming the caller input to pin. "The floor could not be checked" must never read as "checked and fine" — that equivalence *is* the defect. An image carrying no `MESHWEAVER_PLATFORM_VERSION` fails in `prepare` for the same reason; that gap has shipped once (every `memex-portal-ai` image between 2026-08-25 and 09-01 reported `1.0.0` while tagged `3.0.0-rc9.ci.<n>`), and a floor gate reading `1.0.0` would pass everything while proving nothing.
- **It cannot drift from the runtime.** The ordering is written once in `NuGetVersionComparer`, mirrored in `.github/scripts/check-module-platform-floor.py`, and pinned by `ModulePlatformFloorScriptParityTest`: for every case the script's exit code must equal "the runtime found no reason to decline". Two call sites folding one rule differently either never converge or never fire, and both are silent.

## Negative control — the gate was seen RED, then GREEN

```
floor=3.0.0-rc8          platform=3.0.0-ci.7989      exit=1   ← the reported defect
floor=3.0.0-rc9          platform=3.0.0-ci.7989      exit=1
floor=3.0.0-rc14         platform=3.0.0-ci.7989      exit=1   ← names a platform that never existed
floor=3.0.0              platform=3.0.0-ci.7989      exit=1   ← a release outranks every pre-release
floor=3.0.0-ci.7845      platform=3.0.0-ci.7989      exit=0   ← the floor Plugins#1447 moves all 42 to
floor=3.0.0-ci.7845      platform=3.0.0-rc9.ci.7534  exit=0
floor=3.0.0-ci.7845      platform=3.0.0              exit=0
floor=3.0.0-ci.7845      platform=3.1.0-ci.1         exit=0   ← forward across the next two line moves
floor=3.0.0-ci.900       platform=3.0.0-ci.3758      exit=0   ← the numeric-vs-text trap
floor=3.0.0-ci.3758      platform=3.0.0-ci.900       exit=1
floor=(none)             platform=3.0.0-ci.7989      exit=0
floor=3.0.0-ci.7845      platform=(unknown)          exit=1   ← fails closed
```

13/13 as intended. `ModulePlatformFloorScriptParityTest`: `Passed! - Failed: 0, Passed: 18`. `check-workflow-timeouts.py`: `71 job(s) checked … 0 violation(s)`. `check-workflow-shell.py` and `check-pr-secret-preflight.py`: 0 live findings. `DocumentationLinkIntegrityTest` + What's New: `Passed: 2`.

## Relationship to MeshWeaver.Plugins#1447 — complementary, not duplicated

#1447 corrects the 42 values and adds a **static** repo-local check refusing a floor that names the retired `rc` line or the unreleased clean `3.0.0`. That one has to stay static on purpose: the Plugins validate lane runs on forks and Dependabot, where reading a registry would need a credential, and gating on whether that credential is present is the skip-trapdoor AGENTS.md forbids. The `pack` job already **requires** ACR credentials (it errors by name when they are absent), so this half needs no exemption. Between them: the static check catches a bad floor on the pull request that writes it; this one catches any floor the build platform cannot satisfy, whatever its shape.

## Blast radius

Satellites pin `node-repo-module-pack.yml@<40-char sha>`, so **nothing changes for any repo until it moves its pin**. Two notes for whoever does:

- MeshWeaver.Plugins should land #1447 first — its current floors are exactly what this gate refuses.
- `MeshWeaver.SocialMedia`'s `ci.yml` documents (its own comment, "MEASURED 2026-09-04, NOT FIXED HERE") a run where `platform-image-digest` resolved **empty** and the platform-image steps silently skipped. On that lane this gate will now go red naming the input instead of packing a module against nothing — which is the intended direction, but it is a red they should expect rather than discover.

Pairs-with: none — a reusable workflow, a CI script, a test and doc pages; no public type or member is removed.

Closes #3554

### DIFF
--- a/.github/scripts/check-module-platform-floor.py
+++ b/.github/scripts/check-module-platform-floor.py
@@ -1,0 +1,166 @@
+#!/usr/bin/env python3
+"""Is a module's declared `content.minMeshVersion` FLOOR satisfiable by the platform it is
+being built against?
+
+    check-module-platform-floor.py --package AI --floor 3.0.0-rc8 --platform-version 3.0.0-ci.7989
+    exit 0 = the floor is satisfied (or none is declared)
+    exit 1 = it is not, or the platform version is unknown
+
+🚨 WHY THIS EXISTS (Systemorph/MeshWeaver#3554). A module declares a platform FLOOR, and
+`ModulePlatformFloor.DeclineReason` refuses to land the module until the running platform
+satisfies it. Nothing checked that the declared floor could be satisfied by a platform build
+that ACTUALLY EXISTS, so a floor could be declared that no image can ever meet. The module then
+holds for ever, and the hold message names two versions that look like they are in the right
+order:
+
+    HOLDING 3.0.0-ci.7989 — AI: the module requires platform 3.0.0-rc8 or newer
+                             but this deployment runs 3.0.0-ci.7989
+
+SemVer §11.4 compares pre-release identifiers as TEXT, so `"ci" < "rc"` and `3.0.0-ci.<n>` is
+below `3.0.0-rc8` for EVERY n. Measured 2026-09-07: 42 packages declared rc-line or clean-`3.0.0`
+floors, and every self-update candidate was held on both AKS portals while the registry held
+1268 tags — 48 `3.0.0-ci.*`, ZERO `rc*`, ZERO `3.1.0-*`. The failure is silent at runtime: the
+portal logs a hold and carries on.
+
+🚨 WHAT IT COMPARES, AND WHY THAT IS THE RIGHT QUESTION. Not "does some tag somewhere satisfy
+this", but "does THE PLATFORM THIS BUNDLE IS BEING BUILT AGAINST satisfy it". That is strictly
+stronger and needs no registry listing: a module bundle records the framework identity it was
+built against and a consumer only adopts a bundle whose identity matches its own platform, so a
+floor ABOVE the build platform is unsatisfiable by construction — there is no deployment that
+both runs a satisfying platform and would accept this bundle. It is also the reading the lane
+already has in hand, so the check costs nothing.
+
+🚨 IT MUST AGREE WITH THE RUNTIME, NOT MERELY RESEMBLE IT. The ordering below mirrors
+`src/MeshWeaver.Plugin.Packaging/NuGetVersionComparer.cs` exactly — numeric core parts compared
+as numbers, a release outranking any pre-release, dot-separated pre-release identifiers compared
+numerically when both are numeric, numeric ranking below alphanumeric, otherwise ordinal — and
+`ModulePlatformFloorScriptParityTest` in `test/Memex.Portal.Shared.Test` pins the two together on
+a table of cases. Two call sites computing the same fold differently either never converge or
+never fire, and both are silent.
+
+🚨 FAIL CLOSED ON AN UNKNOWN PLATFORM. An empty `--platform-version` is exit 1, not a skip: "the
+floor could not be checked" must never be reported as "checked and fine" — that equivalence IS
+the defect. It mirrors `ModulePlatformFloor.DeclineReason`, which refuses to land a module on
+faith when the running version is unknown.
+"""
+
+from __future__ import annotations
+
+import argparse
+import sys
+
+
+def split(version: str) -> tuple[list[int], list[str]]:
+    """Numeric core + pre-release identifiers, build metadata discarded (SemVer ignores it)."""
+    plus = version.find("+")
+    if plus >= 0:
+        version = version[:plus]
+    dash = version.find("-")
+    core, pre = (version, "") if dash < 0 else (version[:dash], version[dash + 1:])
+    parts = []
+    for p in core.split("."):
+        parts.append(int(p) if p.isdigit() else 0)
+    return parts, (pre.split(".") if pre else [])
+
+
+def compare_core(left: list[int], right: list[int]) -> int:
+    for i in range(max(len(left), len(right))):
+        l = left[i] if i < len(left) else 0
+        r = right[i] if i < len(right) else 0
+        if l != r:
+            return -1 if l < r else 1
+    return 0
+
+
+def compare_pre(left: list[str], right: list[str]) -> int:
+    for i in range(max(len(left), len(right))):
+        # Fewer identifiers ranks LOWER when all preceding are equal: rc3 < rc3.ci.1.
+        if i >= len(left):
+            return -1
+        if i >= len(right):
+            return 1
+        lnum, rnum = left[i].isdigit(), right[i].isdigit()
+        if lnum and rnum:
+            l, r = int(left[i]), int(right[i])
+            if l != r:
+                return -1 if l < r else 1
+            continue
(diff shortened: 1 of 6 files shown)


rbuergi:
Follow-up in `053ce112`: the first commit read the platform version only from a pinned platform IMAGE, so a caller that pins none — a supported shape, the one whose bundle identity is the platform CHECKOUT's own commit (`g<sha>`, #3308) — would have gone RED on any declared floor. That is a lane stranded by a new guard, not a defect found.

`pack` now falls back to the LINE that checkout declares (`$(PlatformVersion)` out of the platform's own `Directory.Build.props`). Weaker on purpose and said out loud: a line satisfies floors that no published build of it does, so a source lane cannot catch an unpublished-`rc` floor. What it still refuses is a floor above the platform's own line — and, the part that matters, **it is never a skip**. Only a platform identifiable neither way is RED, naming what to pin.

The "Blast radius" note above stands unchanged: satellites pin this workflow by SHA, so nothing moves for a repo until it bumps.

PR #3597, opened 2026-09-07
ci: a duplicate mapping key in a workflow is red, at the first job (#3579)

### LINKED ISSUE #3579: Workflow YAML: duplicate mapping keys are accepted silently (last wins) — a pin can move in the diff and not in the job
## The failure

`yaml.safe_load` — and every other loader the fleet's guards use — accepts **duplicate mapping keys silently, last one wins**. In a workflow file that means an edit can add a second `with:` (or a second `platform-ref:`) to a job, and:

- the file parses,
- `actionlint`-style shape checks pass,
- the job runs,
- and it uses the **last** value, not the one the diff appears to set.

The diff looks like it moved a pin. The job kept the old one.

## Measured, 2026-09-07 — a near-miss in Systemorph/MeshWeaver#3560's satellite wave

Adopting #3560's `centralized-gen-manifests` input needs a `with:` block on each caller's `validate:` job. My patcher found the job's end with `s.index("\n  ", i)` — which also matches `"\n    with:"`, because that starts with a newline and two spaces. So it cut the block after `uses:` and appended a **second** `with:`. Result, in MeshWeaver.Education and MeshWeaver.Crm:

```yaml
  validate:
    uses: Systemorph/MeshWeaver/.github/workflows/node-repo-validate.yml@<NEW>
    with:
      platform-ref: <NEW>
      centralized-gen-manifests: true
    with:                                   # ← silently wins
      platform-ref: <OLD>
```

`yaml.safe_load` accepted it. The `uses:` sha had moved and the effective `platform-ref` had not — so the lane would have fetched its central guard scripts *and the canonical gen-manifests.py* at the **old** ref while running the **new** lane. Caught before pushing, by reading the rendered block rather than trusting the parse.

## What would have caught it, and what would not

`Platform pins name one build` (`.github/scripts/check-pin-set-consistency.py`) models this consequence and reds on it:

> `MeshWeaver.Crm — ci.yml:362 calls .github/workflows/node-repo-validate.yml@c7fef7a2… but passes platform-ref 0a2b9017… (line 367).`

— MeshWeaver.Crm#62, run 34083378598, head `437fbffe` (head_sha verified against the PR head).

🚨 **That PR is a different mechanism with the same signature, and the distinction matters.** Crm#62 is a Dependabot bump whose whole diff is `+1/-1` on `.github/workflows/ci.yml`: it moved the `uses:` line and left the `platform-ref` line alone. No duplicate key involved. It is evidence that *this class of stale-pin outcome reaches main's queue routinely*, **not** evidence that anything today detects duplicate keys.

Nothing does. And the pin gate only covers the cases it models:

| | |
|---|---|
| catches a duplicate key whose effect is a pin mismatch | only in repos where `Platform pins name one build` is a **required** context — measured: MeshWeaver.SocialMedia and MeshWeaver.Crm **yes**; MeshWeaver.Reinsurance and MeshWeaver.Education **no** |
| catches a duplicate key with any other effect (a duplicated `if:`, `env:`, `needs:`, `timeout-minutes:`) | **nothing, anywhere** |
| when it catches it | after a full CI run, not at the first job |

## Proposal

Assert it centrally, in `node-repo-validate.yml`, beside the two guards it already fetches at `platform-ref`. It needs no credential, so it runs on forks and on Dependabot PRs; it is milliseconds; and it fails at the first job instead of after a wasted run. A `--self-test` proving it fails on a duplicate and passes on a clean file is the usual bar.

Deliberately **not** folded into #3560: that PR is the gen-manifests centralization, it was at 10 green with four satellite PRs queued behind it, and restarting it for an unrelated gate is how a wave half-lands. This is the same reason the check should not stay a per-repo script — a repo hand-rolling its own copy is behind, not different ([Module Build Architecture](/Doc/Architecture/ModuleBuildArchitecture) → *"scripts are centralized — the lane fetches the platform's copy at the pin; repos keep only allow-files"*).

## The loader, so it does not evaporate

```python
"""yaml.safe_load hides duplicate mapping keys (last wins) — that is how a second `with:` kept an
old platform-ref while the `uses:` moved. Fail on any duplicate key in a mapping."""
import sys, yaml

class Strict(yaml.SafeLoader):
    pass

def no_dupes(loader, node, deep=False):
    seen, out = set(), {}
    for k, v in node.value:
        key = loader.construct_object(k, deep=deep)
        if key in seen:
            raise yaml.constructor.ConstructorError(
                None, None, f"duplicate key {key!r} at line {k.start_mark.line + 1}", k.start_mark)
        seen.add(key)
        out[key] = loader.construct_object(v, deep=deep)
    return out

Strict.add_constructor(yaml.resolver.BaseResolver.DEFAULT_MAPPING_TAG, no_dupes)

ok = True
for path in sys.argv[1:]:
    try:
        yaml.load(open(path), Strict)
        print(f"  ok   {path}")
    except Exception as ex:
        ok = False
        print(f"  FAIL {path}: {ex}")
sys.exit(0 if ok else 1)
```

Run over the four satellites' patched `ci.yml` files it prints `ok` for each; run over the duplicated-`with:` version it names the key and the line.

Found while closing Systemorph/MeshWeaver.Plugins#1426's fleet half.

rbuergi:
## The defect

`yaml.safe_load` accepts a **duplicate mapping key silently and keeps the LAST one** — and every guard in this fleet reads workflow files through it. So a workflow carrying a second `with:` parses, passes every shape check, runs the job, and uses a value the diff never shows.

Measured **2026-09-07**, adopting #3560's `centralized-gen-manifests` input across the satellites. The patcher found the job's end with `s.index("\n  ", i)` — which also matches `"\n    with:"` — so it cut the block after `uses:` and appended a **second** `with:` in MeshWeaver.Education and MeshWeaver.Crm:

```yaml
  validate:
    uses: Systemorph/MeshWeaver/.github/workflows/node-repo-validate.yml@<NEW>
    with:
      platform-ref: <NEW>
      centralized-gen-manifests: true
    with:                                   # ← silently wins
      platform-ref: <OLD>
```

The `uses:` sha had moved and the effective `platform-ref` had not, so the lane would have fetched its central guard scripts — and the canonical `gen-manifests.py` — at the **old** ref while running the **new** lane. It was caught by reading the rendered block, by nothing else.

`check-pin-set-consistency.py` (`Platform pins name one build`) models only the *consequence*: it sees a duplicate whose **effect is a pin mismatch**, only in the repos where that gate is a **required** context (measured: SocialMedia and Crm yes; Reinsurance and Education no), and only **after a full CI run**. A duplicated `if:`, `env:`, `needs:` or `timeout-minutes:` is invisible to it and to everything else.

## What this adds

`.github/scripts/check-workflow-yaml-keys.py` — same shape as the two guards beside it (`--self-test`, `--root .`, `::error` annotations, the name on line 1 load-bearing for the fetch's sanity check).

- **Refuses any duplicate key in any mapping** in every workflow YAML under `.github/workflows/`, **and** in every composite-action `action.yml`/`action.yaml` in the tree. Composite actions are scanned **when present** and never required to exist — a repo with no composite action is not a repo with a missing gate; a repo with no *workflows* directory fails, because nothing to gate is a failure, not a pass. (No satellite carries a composite action today.)
- **The verdict names the file, the mapping, the key and BOTH line numbers**, so the near-miss is reproducible from the message alone:
  > `::error file=.github/workflows/ci.yml,line=10::duplicate key `with` in mapping `jobs.validate` — first written at line 7, shadowed by the one at line 10. YAML takes the LAST, so everything under line 7 is dead; ...`
- **A second arm**: two keys written *differently* that **resolve to the same key**. `on:`, `yes:` and `true:` are all the boolean `True` under PyYAML's YAML-1.1 resolver, so `safe_load` merges them while GitHub's YAML-1.2 parser keeps them apart — a disagreement between what the gates see and what the runner sees is the same hazard, so it is reported.
- **Anchors, aliases and merge keys — a decision, proven in the self-test, not claimed.** The scan walks the composed **node graph** (`yaml.compose_all`), never a constructed dict:

  | Shape | Verdict |
  |---|---|
  | `&base` … `*base` | **silent** — legitimate; an aliased node is the same object at every use site, so it is checked once, never double-reported, and a recursive anchor terminates |
  | `<<: *base` where the merged mapping supplies a key the local mapping also writes | **silent** — YAML defines the explicit key as the winner; that is a documented override, not a shadowed value, and redding it would be redding correct YAML |
  | `<<: [*a, *b]` | **silent** — the spec's own multi-merge form |
  | two separate `<<:` keys in one mapping | **fires** — a literal duplicate with no defined meaning |
- **Needs no credential**, reads nothing external, takes milliseconds — so it runs on forks and on Dependabot pull requests. No `if:` on a secret anywhere, no `continue-on-error:` on any step.
- **`--self-test` is non-vacuous**: 15 planted cases plus three assertions that carry the point — that the message names the file, the key `with`, the mapping `jobs.validate` and both line numbers; that a composite action is scanned; and that a tree with **no workflows at all fails** rather than passing vacuously.

## Wired in both halves, no per-repo copy

1. **Core, on itself** — `dotnet-test.yml`, in the workflow-gates job immediately after the 45-minute-cap guard: `--self-test`, then `--root .`.
2. **Every satellite, centrally** — `node-repo-validate.yml` fetches it at the caller's `platform-ref` with the same no-trapdoor fetch block as its two neighbours (`[ -n "${PLATFORM_REF:-}" ]`, a `gh api … || { echo "::error::…"; exit 1; }`, and the `head -c 400 … | grep -q` sanity check), self-tests it, and runs it against the **caller's** tree.

This is the issue's own proposal — *a repo hand-rolling its own copy is behind, not different*. **No allow file**, deliberately: a duplicate key has no legitimate form to exempt.

## 🎯 NEGATIVE CONTROL — the gate was watched failing in CI

The `--self-test` proves the checker's logic; it does not prove the checker is *wired to a job that can fail the run*. So the branch was pushed with the guard **and a deliberately planted duplicate key** in one of core's own workflows, chosen to be the shape nothing else can see: `pinned-digests.yml`'s `preflight` job writing `timeout-minutes` **twice** — `10`, then `45`. Both are under the cap, so `check-workflow-timeouts.py` reads the LAST through `safe_load` and stays **green** while the first line is silently dead.

| | run | `Workflow shell + gates` | `check-workflow-timeouts.py` |
|---|---|---|---|
| **RED**, plant present (`__RED_SHA__`) | __RED_URL__ | **failure** | passed — 0 violations |
| **GREEN**, plant removed (`__GREEN_SHA__`) | __GREEN_URL__ | **success** | passed |

The exact red log line:

```
__RED_LINE__
```

`pinned-digests.yml` triggers on `push: branches: [main]`, `schedule` and `workflow_dispatch`, so the plant was inert on this branch except as input to the new gate, and `main` never sees it.

## The fleet was measured before pushing

Every satellite's `main`, all workflows and composite actions, run through this guard:

| repo | head | workflow files | verdict |
|---|---|---|---|
| MeshWeaver.Plugins | `a5d08aa25c1a` | 7 | clean |
| MeshWeaver.Education | `58c4062a910d` | 2 | clean |
| MeshWeaver.Reinsurance | `4dd9fadd791d` | 4 | clean |
| MeshWeaver.SocialMedia | `9408df755cd9` | 5 | clean |
| MeshWeaver.Crm | `d9e2768cba6d` | 4 | clean |
| MeshWeaver.Manufacturing | `cfa72caecc37` | 2 | clean |

**No satellite is currently dirty**, so merging this reds nothing. The Education and Crm duplicates from the near-miss never reached `main`. No follow-up issue is needed and no allow entry was added.

## Docs

`Doc/Architecture/WorkflowDuplicateKeys` — the defect class, the near-miss, why every existing gate was blind (with the Crm#62 distinction: same signature, different mechanism), the anchor/alias/merge-key decisions, where the guard runs, and the fleet sweep. Linked from the Architecture index, [Pin Set Consistency](https://github.com/Systemorph/MeshWeaver/blob/main/src/MeshWeaver.Documentation/Data/Architecture/PinSetConsistency.md) and [Reading CI Signals](https://github.com/Systemorph/MeshWeaver/blob/main/src/MeshWeaver.Documentation/Data/Architecture/ReadingCiSignals.md). `DocumentationLinkIntegrityTest` green locally (1 passed, 48 ms, trx written).

## No What's New entry, and why

This is a CI gate. Nothing a portal user can notice changes — no surface, no string, no behaviour. Judged against practice rather than asserted: the two most recent CI-only merges, #3560 (`gen-manifests.py` centralization) and #3573 (pin-bump lane), added no What's New entry either. Manufacturing one here would be noise in the place users go to find what changed for them.

Closes #3579


rbuergi:
🚨 **The negative control's plant was still in the tree on this branch** — pushed a removal commit.

Commit `cdf00939c` deliberately planted a second `timeout-minutes: 45` in
`.github/workflows/pinned-digests.yml` so the new gate could be seen to fire on it. Its own comment
read `NEGATIVE CONTROL for #3579 — REMOVED BEFORE MERGE`; the removal never happened (the session
writing it stopped first), so the branch carried a real, deliberate defect into review.

Removed, and the verification re-run afterwards with the exit codes read **directly** rather than
through a pipe (a pipeline's exit status is the pager's, which is how this class hides):

| tree | exit | result |
|---|---|---|
| planted copy | **1** | `duplicate key \`timeout-minutes\` in mapping \`jobs.preflight\` — first written at line 69, shadowed by the one at line 70` |
| this branch, clean | **0** | `31 workflow file(s) and 0 composite action(s) checked, 0 violation(s)` |
| `--self-test` | 0 | 7/7, each case firing on its defect and silent on its fix |

The denominator is printed on the clean run, so an empty scan cannot be mistaken for a pass.

Worth saying plainly: a negative control that ships is worse than no control, because it looks like
evidence. The control did its job — it demonstrated the gate fires — and then nearly became the very
defect the gate exists to catch.

PR #3609, opened 2026-09-07
fix(#3595): the migration Job renders the legacy escape hatch too — both workloads carry the same secret set

### LINKED ISSUE #3595: The migration Job renders no .Values.extraEnvFrom — a hand-made-SPC environment gets NONE of those values, and the CSI freshness guard is blind to it
## The gap

#3548 made the migration Job mount the SecretProviderClass CSI volume, so the values it reads
through `envFrom` are **current**. That fix is keyed entirely on `memex.keyVaultClasses` — the helper
that enumerates the chart-owned `keyVaultSecrets` / `keyVaultSecretClasses` declarations
(`deploy/helm/templates/memex-portal/_keyvault.tpl:32-65`).

An environment whose SecretProviderClass is **hand-made** attaches it through the documented legacy
escape hatch instead:

```yaml
# deploy/helm/values.yaml:347-365
extraEnvFrom:
  - secretRef:
      name: memexcloud-portal-ai-secrets
extraVolumes: [ … the CSI volume behind that SPC … ]
```

`values.yaml` names that shape explicitly — *"the memex-cloud AI-keys shape"*.

**The migration Job renders none of it.** Measured 2026-09-07T12:25Z on `origin/main` `fbf9e9be1`,
`grep -rn extraEnvFrom deploy/`: exactly **one** rendering site in the whole chart —
`deploy/helm/templates/memex-portal/deployment.yaml:177`, the portal Deployment. The Job
(`deploy/helm/templates/memex-migration/job.yaml`) renders neither `extraEnvFrom` nor
`extraVolumes`/`extraVolumeMounts`.

## Why this is a different defect from #3548, not the same one

#3548 was **staleness**: the Job had the value, from before the rotation. This is **absence**: for a
key delivered through `extraEnvFrom`, the Job has no value at all. `MeshNodeEmbeddingBackfill` —
which is the only thing that embeds rows written before a provider existed — then takes the
`optional: true` / not-configured path and logs-and-skips, and the Job still reports
`Database migration completed`. Same green run, same silent outcome, different mechanism and
different remedy.

## Why no guard catches it

`test/MeshWeaver.Documentation.Test/KeyVaultCsiFreshnessGuard.cs` (added with #3548, hardened by
#3585) keys on the two markers `$class.syncedSecret` and `$class.mountPath` — i.e. on chart-owned
classes only. A hand-made SPC referenced by name through `extraEnvFrom` carries neither marker, so
the guard is blind to this by construction. It cannot go red on it, and did not.

## Two candidate fixes

1. **Render the escape hatch on the Job too** — `.Values.extraEnvFrom` in the container's `envFrom`
   and `.Values.extraVolumes` / `.Values.extraVolumeMounts` on the pod, mirroring the portal
   Deployment. Cheapest, and keeps the two workloads' secret sets identical by construction. It also
   inherits the mount-makes-it-current property #3548 established, provided the environment's
   `extraVolumes` actually carries the CSI volume (which is the point of the twin key).
2. **Migrate those environments off the escape hatch** onto `keyVaultSecretClasses`, which renders
   the SPC, the volume, the mount and the `envFrom` from one declaration — and then extend
   `KeyVaultCsiFreshnessGuard` to refuse a chart that still declares `extraEnvFrom` for a CSI-synced
   Secret. Better end state; needs the affected environments' values files (private
   `Systemorph/Memex`) to move first.

Whichever is chosen, the guard should be able to see it — otherwise this recurs the next time a
workload is added.

Split out of #3548 rather than reopening it: that issue's stated scope ("the Job reads the
CSI-synced secret before the driver rotates it") is met and closed.

<sub>Measured 2026-09-07T12:25Z on `origin/main` `fbf9e9be1`, read-only: `grep -rn extraEnvFrom
deploy/`, `cat deploy/helm/templates/memex-migration/job.yaml`, `sed -n '335,395p'
deploy/helm/values.yaml`, and `grep -n` over `KeyVaultCsiFreshnessGuard.cs`. Nothing edited, nothing
deployed.</sub>

rbuergi:
Closes #3595.

## The defect

The migration Job carried the chart-**owned** Key Vault classes and nothing else. An environment
whose `SecretProviderClass` is **hand-made** attaches it by name through `.Values.extraEnvFrom` —
the shape `values.yaml` itself calls *"the memex-cloud AI-keys shape"* — and the whole chart had
exactly **one** rendering site for that escape hatch: the portal Deployment.

So for a key delivered that way the Job had **no value at all**. `MeshNodeEmbeddingBackfill` — the
only thing that embeds rows written before a provider existed — took the not-configured path,
logged-and-skipped every row, and the Job still reported `Database migration completed`.

**This is ABSENCE, not #3548's STALENESS**, and it needs a different remedy. #3548 (merged as #3572)
made the Job *mount* the chart-owned SPC so its values are current; it is keyed entirely on
`memex.keyVaultClasses`, so it cannot reach a class the chart does not own.

## The fix

The Job now renders the escape hatch exactly as the portal Deployment does — `.Values.extraEnvFrom`,
and its `.Values.extraVolumes` / `.Values.extraVolumeMounts` twins. The mount is what makes the CSI
driver fetch the vault objects and write the synced Secret *before any container in the pod starts*,
so the Job gets the current value by construction, the same property #3548 established.

**Verbatim, and deliberately NOT `optional: true`.** The chart-owned entries above are optional
because the chart *renders* their SecretProviderClass and knows a first install has nothing to
backfill. The chart owns nothing here — the operator declared this source, and the volume the Job
now mounts is what materialises it. Marking it optional would recreate the exact silent-skip this
change closes. The portal Deployment has identical semantics for identical entries, so nothing can
fail here that would not already fail there.

## Measurements

**Byte-identical for every shipped configuration** (`helm template` before ⇢ after, `cmp`):

| combination | result |
|---|---|
| self-host (`deploy/helm/values.yaml`) | `cmp` **rc=0** — sha256 `3593ec21…` both sides |
| AKS overlay (`+ deploy/aks/values.aks.yaml`) | `cmp` **rc=0** |
| memex-local (`+ deploy/homebrew/share/values.local.defaults.yaml`) | `cmp` **rc=0** |

**With the escape hatch set, the diff IS the fix** — nothing else in the whole render changes:

```diff
             - secretRef:
                 name: "memex-migration-secrets"
+            - secretRef:
+                name: memexcloud-portal-ai-secrets
           imagePullPolicy: "IfNotPresent"
+          volumeMounts:
+            - mountPath: /mnt/secrets-store
+              name: kv-ai-secrets
+              readOnly: true
+      volumes:
+        - csi:
+            driver: secrets-store.csi.k8s.io
+            readOnly: true
+            volumeAttributes:
+              secretProviderClass: memexcloud-portal-ai-secrets
+          name: kv-ai-secrets
```

**Parity, parsed out of the rendered manifests** with a chart-owned class *and* the hatch declared
together — both workloads, same sources, same order (order is precedence: k8s keeps the last
`envFrom` on a clash):

```
JOB    envFrom: [configMapRef:memex-migration-config, secretRef:memex-migration-secrets,
                 secretRef:memex-portal-keyvault, secretRef:memexcloud-portal-ai-secrets,
                 configMapRef:memexcloud-extra-config]
PORTAL envFrom: [configMapRef:memex-portal-config,    secretRef:memex-portal-secrets,
                 secretRef:memex-portal-keyvault, secretRef:memexcloud-portal-ai-secrets,
                 configMapRef:memexcloud-extra-config]
JOB    volumes: [kv-secrets, kv-ai-secrets]      ← the hatch's CSI volume is mounted by the Job
```

**Gates, run locally:** `helm lint` → `1 chart(s) linted, 0 chart(s) failed` · `check-chart-invariants.sh`
→ *All 3 values combinations render a self-consistent deployment* · `check-values-are-read.sh` →
*All 119 config key(s) across 3 values file(s) are read by a template* ·
`test-chart-drift-compare.sh` → *all classification assertions passed* ·
`dotnet build -c Release -warnaserror` on `MeshWeaver.Documentation` and
`MeshWeaver.Documentation.Test` → `0 Warning(s) 0 Error(s)`.

## The guard, and its negative controls

`KeyVaultCsiFreshnessGuard` was blind to this **by construction** — it keys on
`$class.syncedSecret` / `$class.mountPath`, markers a hand-made class does not carry. Two facts
added to it:

1. `EveryWorkloadThatCarriesTheChartsKeyVaultSecrets_AlsoRendersTheLegacyEscapeHatch` — the parity half;
2. `EveryWorkloadThatRendersTheEscapeHatchEnvFrom_AlsoRendersItsVolumeAndMount` — #3548's freshness
   argument applied to the class the chart does not own.

Each asserts its own denominator (templates found, ≥2 readers, ≥2 hatch renderers) and both assert
that `values.yaml` still **declares** the three keys, so a rename there fails loudly instead of
leaving the guard matching nothing.

**Falsified, three ways** — each mutation applied to the tree, tests run, then reverted and re-run:

| control | mutation | result |
|---|---|---|
| **A** | the Job stops rendering `.Values.extraEnvFrom` (**the #3595 state itself**) | ❌ both new facts fail · ✅ **the #3548 fact still passes** — the blindness, demonstrated rather than asserted |
| **B** | the Job keeps `extraEnvFrom`, drops the volume + mount | ❌ only the freshness fact fails · ✅ the parity fact correctly stays green |
| **C** | `values.yaml` renames `extraEnvFrom` → `extraEnvSources` | ❌ both new facts fail on the denominator — a rename cannot silence the guard |
| — | restored | ✅ `Passed! - Failed: 0, Passed: 6` (with `MigrationWorkloadModelGuard`), 14/14 with the doc-integrity tests |

Control A also proves the guard does not pass on its own documentation: the mutated Job still
contains the literal `.Values.extraEnvFrom` **inside a Helm comment block**, and the scan strips
comments before matching, so it went red anyway.

## Work product conserved

`Doc/Architecture/DeploymentEnvLayers` gains *"Two workloads, one secret set — and the escape hatch
has to reach both"*: the #3548-vs-#3595 table (staleness vs absence, and why both produced a green
run), why the three values keys are a set, and why the original guard could not see this.

Pairs-with: none — this diff removes no public type and no public member from `src/`; it touches the
Helm chart, one test file and two documentation assets.

### DIFF
--- a/deploy/helm/templates/memex-migration/job.yaml
+++ b/deploy/helm/templates/memex-migration/job.yaml
@@ -76,6 +76,30 @@ spec:
                 name: "{{ $class.syncedSecret }}"
                 optional: true
             {{- end }}
+            {{- /* ⚠️ LEGACY ESCAPE HATCH — the SAME env sources the portal Deployment renders
+                   (memex-portal/deployment.yaml), and for the SAME reason the block above exists
+                   (MeshWeaver#3595).
+
+                   The block above covers only the classes the CHART owns. An environment whose
+                   SecretProviderClass is HAND-MADE attaches it by name through `.Values.extraEnvFrom`
+                   instead — the shape values.yaml calls "the memex-cloud AI-keys shape" — and until
+                   this block existed the Job rendered NONE of it. That is a different defect from
+                   #3548, which was STALENESS: for a key delivered this way the Job had no value at
+                   all, so `MeshNodeEmbeddingBackfill` took the not-configured path, logged-and-
+                   skipped every row, and the Job still reported `Database migration completed`.
+                   Same green run, same silent outcome, different mechanism.
+
+                   VERBATIM, and deliberately NOT `optional: true`. The chart-owned entries above
+                   are optional because the chart RENDERS their SecretProviderClass and knows a
+                   first install has no pre-existing rows to backfill. The chart owns nothing here:
+                   the operator declared this source, and the volumes below are what materialise it
+                   before this container starts. Marking it optional would recreate the very defect
+                   this block closes — an absent value read as "not configured", green and silent.
+                   The portal Deployment has identical semantics for the identical entries, so an
+                   env whose Secret is genuinely missing already fails there. */}}
+            {{- range .Values.extraEnvFrom }}
+            - {{ toYaml . | nindent 14 | trim }}
+            {{- end }}
           imagePullPolicy: "IfNotPresent"
           {{- /* 🚨 THE JOB MOUNTS THE SPC ITSELF — this is what makes the values above CURRENT
                  (MeshWeaver#3548).
@@ -109,15 +133,24 @@ spec:
                  name. Rendered only when a class is declared — a namespace with no Key Vault gets
                  the pod spec it had before. */}}
           {{- $keyVaultClasses := (include "memex.keyVaultClasses" . | fromYamlArray) }}
-          {{- if $keyVaultClasses }}
+          {{- if or $keyVaultClasses .Values.extraVolumeMounts }}
           volumeMounts:
             {{- range $class := $keyVaultClasses }}
             - name: "{{ $class.volumeName }}"
               mountPath: "{{ $class.mountPath }}"
               readOnly: true
             {{- end }}
+            {{- /* ⚠️ LEGACY: the mount half of the escape hatch, and the half that carries the
+                   whole freshness argument above for a HAND-MADE SecretProviderClass. The envFrom
+                   entry names the synced Secret; THIS is what makes the CSI driver fetch the vault
+                   objects and write that Secret before any container in this pod starts. Without
+                   it the Job would free-ride on the portal pods' mount — #3548's defect, in
+                   escape-hatch clothing. KeyVaultEscapeHatchParityGuard asserts the pair. */}}
+            {{- range .Values.extraVolumeMounts }}
+            - {{ toYaml . | nindent 14 | trim }}
+            {{- end }}
           {{- end }}
-      {{- if $keyVaultClasses }}
+      {{- if or $keyVaultClasses .Values.extraVolumes }}
       volumes:
         {{- range $class := $keyVaultClasses }}
         - name: "{{ $class.volumeName }}"
@@ -126,5 +159,11 @@ spec:
             readOnly: true
             volumeAttributes:
               secretProviderClass: "{{ $class.name }}"
+        {{- end }}
+        {{- /* ⚠️ LEGACY: the volume half — typically the secrets-store CSI volume behind a
+               hand-made SecretProviderClass. Verbatim Volume objects, mounted by
+               .Values.extraVolumeMounts above. */}}
+        {{- range .Values.extraVolumes }}
+        - {{ toYaml . | nindent 10 | trim }}
         {{- end }}
       {{- end }}
--- a/deploy/helm/values.yaml
+++ b/deploy/helm/values.yaml
@@ -345,7 +345,13 @@ keyVaultSecrets:
 #         - vaultSecret: memex-connectionstring
 #           key: ConnectionStrings__memex
 keyVaultSecretClasses: []
-# ---- Extra env sources for the portal (verbatim EnvFromSource objects) ----
+# ---- Extra env sources for the portal AND the migration Job ----
+# (verbatim EnvFromSource objects)
+# 🚨 BOTH workloads render these, and that parity is the point (#3595):
+# the migration Job rendered none of it until then, so a key delivered this way
+# reached the portal and NOT the backfill — which logged-and-skipped every row
+# and still reported `Database migration completed`. KeyVaultEscapeHatchParityGuard
+# holds the two renders together.
 # ⚠️ LEGACY ESCAPE HATCH — kept for environments whose SecretProviderClass is
 # still hand-made. Declare the secrets under `keyVaultSecrets` above instead;
 # that renders the SecretProviderClass AND the envFrom, so the pod's secret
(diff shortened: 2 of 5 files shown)


rbuergi:
### Two more shapes verified: the ASYMMETRIC escape hatch, and an accidental control on unfixed `main`

The PR body measures the symmetric case (`extraEnvFrom` + its volume twin). An environment may
legitimately declare only one half — `extraVolumes` alone carries the jobrunner token mount that
`values.yaml:159` documents, and `extraEnvFrom` alone is right for a plain, non-CSI Secret. Both had
to render a valid pod spec on the Job, so both were rendered and parsed:

```
--- extraEnvFrom only ---
  envFrom     : [memex-migration-config, memex-migration-secrets, plain-k8s-secret]
  volumeMounts: []
  volumes     : []
  every mount has a volume: True (mounted=∅)

--- extraVolumes + extraVolumeMounts only ---
  envFrom     : [memex-migration-config, memex-migration-secrets]
  volumeMounts: [(token, /var/run/token)]
  volumes     : [token]
  every mount has a volume: True (mounted={token})
```

Neither emits an empty `volumes:` / `volumeMounts:` key, and in both the mount set is a subset of
the declared volume set — so no shape of the escape hatch can produce an invalid pod spec on the
Job.

**An accidental control worth recording.** The first attempt at the above rendered the chart from a
worktree parked on a *different* branch, so it rendered `origin/main`'s chart rather than this one.
Both shapes came back with the Job carrying **nothing**:

```
envFrom: [memex-migration-config, memex-migration-secrets]   volumeMounts: []   volumes: []
```

That is #3595 reproduced from the other direction, independently of the guard: on unfixed `main` the
migration Job renders none of the escape hatch for **either** shape. I nearly filed it as a failure
of this change before checking which tree helm had actually read — so it is reported here as what it
is, a control I did not plan.

Issue #3504, opened 2026-09-06
MeshWeaver.Plugins runs NO secret preflight and neither it nor Manufacturing calls node-repo-validate — a hand-rolled lane opts out of every guard the lane grows later


rbuergi:
## Measured: two repos do not call `node-repo-validate`, and one of them runs NO secret preflight at all

Found while applying the rule that #3454's work produced, rather than by looking for it:

> **The guards live INSIDE the shared lane. A hand-rolled copy therefore opts out of every guard that
> lane grows LATER — silently, retroactively, and invisibly from the repo that made the copy.**

`check-workflow-timeouts.py` and `check-pr-secret-preflight.py` are both invoked *inside*
`node-repo-validate.yml`. Neither existed when the older copies were made, so nobody chose to skip
them — the copy chose, months of commits later. A repo that forked a lane before a guard was written
looks identical to one that passes it.

## The adoption matrix, measured 2026-09-07 against each repo's `main`

Read from each repo's actual workflow files, not from AGENTS.md — which named Education a caller
while Education called none (that is the precedent, #258).

| repo | **validate** | compile-check | gate | module-pack | publish-bake | tag-modules | platform-ref-bump |
|---|:---:|:---:|:---:|:---:|:---:|:---:|:---:|
| MeshWeaver.SocialMedia | ✅ | ✅ | ✅ | ✅ | ✅ | ✅ | ✅ |
| MeshWeaver.Reinsurance | ✅ | ✅ | ✅ | — | ✅ | ✅ | — |
| MeshWeaver.Crm | ✅ | ✅ | ✅ | — | ✅ | ✅ | — |
| MeshWeaver.Education | ✅ | — | — | — | ✅ | ✅ | — |
| **MeshWeaver.Manufacturing** | **❌** | ✅ | ✅ | — | ✅ | — | — |
| **MeshWeaver.Plugins** | **❌** | — | ✅ | ✅ | ✅ | — | ✅ |

A missing lane is not automatically wrong — a repo that packs no modules has no use for
`module-pack`. **`validate` is different**, because it is where the fleet-wide guards run.

## What each of the two actually has

| repo | `check-workflow-timeouts.py` | `check-pr-secret-preflight.py` |
|---|---|---|
| MeshWeaver.Manufacturing | ✅ inline in `ci.yml` (+ `auto-arm.yml`) | ✅ inline in `ci.yml` |
[...]

PR #3624, opened 2026-09-07
docs(#3504): an unadopted repo has NO verdict — the Plugins adoption row was filled in from a pull request

rbuergi:
## A column headed "verdict on its `main`" was filled in from a pull request

`Doc/Architecture/CiContentBake` → *Calling the lane is NECESSARY and NOT SUFFICIENT* carries the
adoption table whose last column is **`check-pr-secret-preflight` verdict on its `main`**. The
MeshWeaver.Plugins cell read:

```
| MeshWeaver.Plugins | c7fef7a2d | ✅ | 0 (adopted 2026-09-07, #3504) |
```

Measured 2026-09-07 against `Systemorph/MeshWeaver.Plugins` `origin/main`:

```
$ git grep -n "node-repo-validate"        origin/main -- .github/   → (nothing)
$ git grep -n "check-pr-secret-preflight" origin/main -- .github/   → (nothing)
```

No workflow calls the lane; nothing invokes the guard. **The cell recorded a verdict no run has
produced**, and the pin column named a sha that exists only on a branch. It was anticipating
Plugins#1453, which is open and blocked on that repo's own platform-pin move (Plugins#1463, red on
`Portal hosts (shard 3)` since 11:12Z).

🚨 **Zero and UNMEASURED are different readings**, and this is the one table in the tree whose whole
job is to tell them apart — its own control arm is Reinsurance, *"the only pre-existing caller whose
pin carried the guard, and the only one at zero"*. A cell that says `0` where nothing ran makes the
largest satellite in the fleet look exactly like a repo the guard passes, which is the failure the
surrounding section is about.

## What changes

- The row names what `main` actually has: **no lane**, therefore no pin and no guard, and the
  verdict cell says **UNMEASURED** rather than a number — while still carrying the **3** violations
  the guard did name the first time it ran on the branch (`MESHWEAVER_APP_ID`,
  `MESHWEAVER_APP_PRIVATE_KEY`, `REGISTRY_PUBLISH_TOKEN`, each now a reasoned allow entry).
- A short paragraph beside the table records why it happened and the rule that prevents it: a column
  about `main` may never be filled in from a pull request.

When Plugins#1453 lands, the row becomes an ordinary measured `0` like Manufacturing's — one line,
after the fact.

## Verification

- **Manufacturing's row re-checked, and it is correct**: its `main` calls
  `node-repo-validate.yml@1b5350d547…` with the paired `platform-ref: 1b5350d547…` (read from the
  repo's `ci.yml` over the contents API). Only the Plugins row was wrong.
- **Swept for the same claim elsewhere**: `CrossRepoPairGate.md` cites #3504 only as a sibling
  shape, and `WorkflowDuplicateKeys.md` already states the true position (*"Plugins runs none of the
  three central workflow guards"*). Neither needed a change.
- Doc-only, no new internal links, so `DocumentationLinkIntegrityTest`'s subject is unchanged.

**No What's New entry**: internal engineering record, no user-visible effect — the documented skip case.

**Pairs-with: none** — doc-only; the diff removes no public type or member from `src/`.

Part of #3504.

### DIFF
--- a/src/MeshWeaver.Documentation/Data/Architecture/CiContentBake.md
+++ b/src/MeshWeaver.Documentation/Data/Architecture/CiContentBake.md
@@ -741,10 +741,20 @@ does not run it. Measured against every satellite's `main`, 2026-09-07:
 |---|---|:---:|---|
 | MeshWeaver.Reinsurance | `1b5350d54` | ✅ | **0 violations** |
 | MeshWeaver.Manufacturing | `1b5350d54` | ✅ | 0 (adopted 2026-09-07, #3504) |
-| MeshWeaver.Plugins | `c7fef7a2d` | ✅ | 0 (adopted 2026-09-07, #3504) |
+| MeshWeaver.Plugins | *(none — `main` calls no validate lane)* | — | 🚨 **UNMEASURED: the guard does not run here.** Adoption is Plugins#1453, open. Its first execution named **3** (`MESHWEAVER_APP_ID`, `MESHWEAVER_APP_PRIVATE_KEY`, `REGISTRY_PUBLISH_TOKEN`) — all now reasoned allow entries on that branch, where the lane reads 0 |
 | MeshWeaver.Crm | `0a2b9017d` → `c7fef7a2d` | ❌ → ✅ | **4** — incl. `MW_REGISTRY_KEY` → 0 (adopted 2026-09-07 12:30Z, `9228b8be`) |
 | MeshWeaver.SocialMedia | `0a2b9017d` → `c7fef7a2d` | ❌ → ✅ | **4** — incl. `MW_REGISTRY_KEY` → 0 (adopted 2026-09-07 12:30Z, `9dea3668`) |
 | MeshWeaver.Education | `8ffbe4762` → `c7fef7a2d` | ❌ → ✅ | **4** → 0 (adopted 2026-09-07, Education#283) |
+
+🚨 **The Plugins row is the reason this table says `main` and means it.** An earlier revision read
+*"0 (adopted 2026-09-07, #3504)"* — anticipating a merge that has not happened. Measured
+2026-09-07 against `Systemorph/MeshWeaver.Plugins` `origin/main`: **no workflow calls
+`node-repo-validate.yml` and nothing invokes `check-pr-secret-preflight.py`**, so the cell recorded
+a verdict no run has produced. In the largest satellite of the fleet, that is the exact failure the
+section above describes — a repo that looks identical to one the guard passes. A column headed
+*"verdict on its `main`"* must never be filled in from a pull request; an unadopted repo has **no
+verdict**, which is a different thing from zero. The adoption is blocked on that repo's own
+platform-pin move (Plugins#1463), not on the change.
 
 **Reinsurance is the control**, and it is what makes the table evidence rather than an assertion:
 it is the only pre-existing caller whose pin carried the guard from the start, and it was the only


rbuergi:
## Held as a draft — the premise may be about to expire, and one sentence in it is already wrong

**Why draft.** This PR's whole content is "MeshWeaver.Plugins' `main` has no lane, so the verdict
cell is UNMEASURED". Plugins#1453 — the adoption — is now green-path and running CI on a head that
carries the platform pin (`88dd5c16`, `MW_PLATFORM_SET 3.0.0-ci.8032`, staleness re-measured at
1.1 h / 25 commits against bounds of 24 h / 120). If it merges, the original row (`0 (adopted
2026-09-07, #3504)`) becomes correct and landing this would introduce the same class of error in
the opposite direction. Drafting is the hold that also disarms auto-merge.

**And a correction I owe to this PR's own body.** It says the adoption is *"blocked on that repo's
own platform-pin move (Plugins#1463)"*. That reasoning is wrong, and the mistake is worth keeping
because several people made it at once:

> 🚨 **The staleness gate reads the PULL REQUEST'S OWN `ci.yml`.** So Plugins#1463 landing on `main`
> would not have unblocked #1453 at all — each branch must carry the pin itself. And auto-merge
> cannot bridge that: it acts only once checks are green, which they never would be. Waiting on the
> pin PR was waiting on something that could not arrive.

Resolution, once #1453 settles:

- **merged** → close this unmerged; the row is correct as written.
- **not merged** → re-open, drop the `Plugins#1463` clause for the paragraph above, and land it.

rbuergi:
**Hold STANDS — still waiting on Systemorph/MeshWeaver.Plugins#1453**, and that PR is not slow, it is *structurally* stuck.

Measured 2026-09-08 07:3xZ. #1453's head `3eab127d` is **completely green** — its `Plugin Catalog CI` run [`34191771207`](https://github.com/Systemorph/MeshWeaver.Plugins/actions/runs/34191771207) is `success` across all 106 jobs, `validate / Validate node repos` included. It nonetheless reads `mergeable_state: blocked`, for one reason:

```
MeshWeaver.Plugins branches/main/protection → required contexts:
  Validate node repos                          ← never published again by #1453
  Compile every NodeType (vs core)             ✅
  Build + test the portal hosts                ✅
  Module bundles / All selected bundles built  ✅
  Module bundles (floor) / …                   ✅
```

#1453 adopts the reusable lane, which renames that job's context to `validate / Validate node repos` and splits the repo-specific half out as `Repo policy gates`. The **bare** name is never reported again, and MeshWeaver.Plugins uses **classic** branch protection (not a ruleset), where an unreported required context blocks forever rather than counting as satisfied. #1453's own comment says so and expects branch protection to move in the same change.

So the remaining blocker is a **branch-protection edit on MeshWeaver.Plugins**, not any code or CI work, and it is deliberately not being made from this PR because it has fleet-wide blast radius: dropping the old context and adding the two new ones blocks every open Plugins PR that has not yet merged #1453 (7 non-draft ones today, several owned by other sessions). Doing it in the wrong order is the exact failure #1453 warns about — *"an absent required context counts as SATISFIED, so getting this wrong leaves a silently unguarded repo"*.

This doc PR is correct as written and should land the moment #1453 does; until then the row it fixes is still `UNMEASURED`, which is what this PR says it is.

rbuergi:
Closing as superseded, following the original draft hold: Systemorph/MeshWeaver.Plugins#1453 merged on 2026-09-08 (10768c9f015ffc314f7f73fc7c4feaf0d1d1b0ad), and current Plugins main calls node-repo-validate.yml. The required-check follow-through is now complete under #3504. Landing this diff would incorrectly claim the adoption is still absent.

PR #3629, opened 2026-09-07
fix(Plugins#1483): the view-pack check asserts the packs are LOADED, and covers the one that draws every input

rbuergi:
The second defect named in Systemorph/MeshWeaver.Plugins#1483: `memex-local verify` reported a
portal USABLE that could not draw a single input control, and printed `ok update complete` after it.

`memex-local` lives here (`deploy/homebrew/bin/memex-local`), so this is the core half of that
issue. The render fix is Systemorph/MeshWeaver.Plugins#1486.

## Two gaps, one shape: the answer printed was to a different question

**1 — the probe named two packs and there are three.** `VIEW_PACK_MODULES` covered
`MeshWeaver.Blazor.Views` and `MeshWeaver.Blazor.Graph`. The one it left out,
`MeshWeaver.Blazor.EntityViews`, is every input in the product — text, number, date and choice
fields plus the Editor / EditForm / Property skins. A portal without it can be read and not
operated: no quiz answered, no coupon redeemed, no dialog field filled in. It passed check 3.

**2 — "resolvable on disk" was reported as "present".** Those are different states. A module reaches
the running process only if boot put it in the effective set — `Modules:Assemblies` ∪ the landed
activation sidecar (`ModuleActivationBoot.ComputeEffectiveModuleEntries`, called from
`memex/Memex.Portal.Shared/MemexConfiguration.cs:268`) — and bytes sitting in `modules/` that
neither names are inert. #1483's reporter observed exactly that: the probe said
`MeshWeaver.Blazor.Graph` **present** while `MeshNodePickerControl` still rendered as its
`ToString()`. The file was really there; the question being answered was not the one being printed.

## What it asserts now, and why that is reachable from a shell script

The portal already says which copy of which module it is about to load — one
`[ModuleLoad] <name> ← <path> (source=…)` line per module, at boot, written from the SAME array it
then loads, which is the whole point of `ModuleLoadReport` (*"the report and the load cannot
disagree"*, MeshWeaver#2223). That report **is** the activation signal, and `kubectl logs` can read
it. So `assert_view_packs_loaded` reads it and names any pack that is on disk and not running,
with the remedy that actually applies (rebuild an image that declares the pack, or switch to
registry mode) rather than the restart that cannot help.

🚨 **It never passes on no evidence.** With no `[ModuleLoad]` line visible at all the boot report has
scrolled out of the container's log — the check learned nothing, and "learned nothing" is reported
as a failure naming its remedy. That is the same rule the existing `FAKE_EXEC_FAILS` arm already
follows, applied to the second input this check now has.

The `--repair` arm no longer prints `view packs present and activated` straight off
`rollout status`: the restart is the ACT, the new pod's boot report is the EVIDENCE, and asserting
the thing you are trying to establish without looking at it is how this check got here.

## `values.local.self-registry.yaml`

Its cost note listed EntityViews among the missing "Radzen charts, Analysis, EntityViews, GoogleMaps
or Speech", which reads as a loss of visualisations. That framing is what let the defect sit: it is
the loss of every way to enter anything. Corrected, with the reason index 2 nonetheless stays blank
recorded beside it — un-blanking would require the module of an image that may predate the seed, and
a required module that cannot land answers `/health` 503 and times the rollout out (Memex#131).

## Verification

`bash deploy/homebrew/test/run-tests.sh` → **passed 64, failed 0**. `bash -n` and
`shellcheck -S error` clean on the CLI.

**Falsification — the four new/changed assertions against the pre-change script**
(`git show origin/main:deploy/homebrew/bin/memex-local`) → **passed 60, failed 4**:

```
FAIL  a missing EntityViews is caught — a portal nobody can type into is not usable
FAIL  a pack on disk but NOT in the boot report is not green
FAIL  no boot report means the check FAILS, never passes
FAIL  a healthy portal verifies green          # the green line now names the claim it makes
```

🚨 The first of those did **not** discriminate on my first draft, and the reason is worth recording:
the suite's `kubectl` stub echoed the whole `FAKE_MODULES` fixture regardless of which modules the
probe asked about, so a fixture line for a pack outside `VIEW_PACK_MODULES` was still read as a
missing pack — the test passed against a script that never looked at EntityViews. The stub now
models the real in-pod probe and reports only the modules named in its last argument, which is what
turns that test into a control.

Core's own `WhatsNewEntryIntegrityTest` passes with the new entry (1 passed / 0 failed, after
`dotnet build -c Release -warnaserror` of `MeshWeaver.Documentation.Test` → `0 Warning(s) 0 Error(s)`).

## Merge order

**After Systemorph/MeshWeaver.Plugins#1486.** This PR arms a detector; that PR ships the fix it
detects the absence of. Landing the detector first would red every self-registry install for a state
whose remedy is not yet buildable.

No public surface leaves `src/` — `Pairs-with: none — this diff touches `deploy/homebrew/` and one
doc page; it removes no public type or member.`

### DIFF
--- a/deploy/homebrew/bin/memex-local
+++ b/deploy/homebrew/bin/memex-local
@@ -915,9 +915,16 @@ uninstall_launchd() {
 # manual recovery step.
 # ---------------------------------------------------------------------------
 
-# The two packs a Blazor portal cannot render without. MODULE (assembly) names — what the landing
-# lane writes on disk; the catalog package names are DefaultViews and GraphViews.
-readonly VIEW_PACK_MODULES="MeshWeaver.Blazor.Views MeshWeaver.Blazor.Graph"
+# The packs a Blazor portal cannot render without. MODULE (assembly) names — what the landing
+# lane writes on disk; the catalog package names are DefaultViews, GraphViews and EntityViews.
+#
+# 🚨 EntityViews is here because leaving it out made this check report a portal USABLE that could
+# not draw a single input (MeshWeaver.Plugins#1483). It is not a visualisation pack: EditorView,
+# TextFieldView, NumberFieldView, DateTimeView, RadioGroupView, the comboboxes, checkboxes and
+# switches plus the Editor / EditForm / Property skins all live in it, so without it every quiz
+# option, the Subscribe coupon box and every dialog field render as the control's own ToString().
+# A portal that cannot be typed into is not a portal, and it passed check 3.
+readonly VIEW_PACK_MODULES="MeshWeaver.Blazor.Views MeshWeaver.Blazor.Graph MeshWeaver.Blazor.EntityViews"
 
 # One in-pod probe answering the whole of check 3: for each module, is it resolvable at all — landed
 # under <Modules:Root>/modules/<name>[@<generation>]/<name>.dll, or shipped in the image's own app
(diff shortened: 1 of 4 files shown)


rbuergi:
Held as a DRAFT on purpose — this is a merge-ORDER hold, not a work-in-progress one. The diff is complete and its suite is green (64/0, with the four new assertions falsified against the pre-change script at 60/4).

This PR arms a detector; Systemorph/MeshWeaver.Plugins#1486 ships the fix it detects the absence of. Landing this first would red `memex-local verify` on every self-registry install for a state whose remedy is not yet buildable. I will mark it ready the moment #1486 is merged.

PR #3654, opened 2026-09-07
Doc: self-hosted CI runners on AKS, and the reserve that sizes them

rbuergi:
## What

Records the Actions Runner Controller install on the shared AKS cluster as a doc page, plus a What's New entry, and reconciles it with the shape `ChartOwnershipAndRunnerPool` had already recorded.

Docs only — no code, no public surface touched.

## Why

CI queues behind an org-wide ceiling of 60 concurrent GitHub-hosted jobs. Measured 2026-09-07: **61 hosted jobs running org-wide, 18 MeshWeaver.Plugins runs queued behind them**, and **$1,987 of a $3,000 monthly Actions budget spent on day 7** with `prevent_further_usage: true` armed. One Plugins PR is 105 jobs / ~350 job-minutes. Self-hosted minutes do not count against that budget.

## The part worth reviewing: the cap is a reserve, not headroom

Sizing this from `kubectl top` would have suggested a limit **five times too large**. The silos pool reads 2–14% CPU in use — but the portals *request* 4 CPU + 8Gi per replica, and `memex-cloud` is a KEDA `ScaledObject` at **min 2 / max 8** sitting at **74% of an 80% memory target**, i.e. one nudge from adding replicas.

So the page reserves production's growth first and takes the remainder:

| | CPU | memory |
|---|---|---|
| pool allocatable (3 × D16s_v5) | 47,220m | 177,432 Mi |
| already requested | 26,517m | 54,466 Mi |
| reserved: KEDA +2 replicas | 8,000m | 16,384 Mi |
| reserved: rolling surge +2 pods | 8,000m | 16,384 Mi |
| **left for runners** | **4,703m** | **90,198 Mi** |

At a 500m CPU request that is 9.4 runners ⇒ **`maxRunners: 8`**. Both reserves are then satisfied **without relying on preemption or on a fourth node** — those are the second and third lines of defence, not the plan.

## Three brakes, and why the priority class is -10 exactly

- `maxRunners` (a setting), a **`ResourceQuota`** (API-server enforced, one runner above the cap), and a **`LimitRange`** that makes an unbounded pod — the one shape that could crowd a portal regardless of priority — unschedulable.
- Priority **-10** with `preemptionPolicy: Never`: every portal pod runs at the default priority 0, so runners rank below all of them and can never preempt. **-10 and not lower** is deliberate — below the cluster autoscaler's `--expendable-pods-priority-cutoff` (default -10) a runner cannot grow the pool *and can be deleted mid-job*.

## Reconciling with the earlier record

`ChartOwnershipAndRunnerPool` recorded the plan as *"on a dedicated node pool so CI cannot starve the portal"*. What shipped shares the existing pool, because the headroom was measured after that page was written. Rather than leave the two contradicting, both now cross-link, and the dedicated pool is written up as the **next** step with its trigger (steady rather than bursty load, or a reserve that stops closing).

## Verification

- `DocumentationLinkIntegrityTest` passes, and the new pages are confirmed **embedded resources** in the built `MeshWeaver.Documentation.dll` — so the test saw them rather than passing on files it never loaded.
- `dotnet build src/MeshWeaver.Documentation -c Release -warnaserror` → `0 Warning(s), 0 Error(s)`.
- Portals verified `2/2` and `4/4` Ready and **HTTP 200 before and after** the cluster work.

## What is deliberately NOT in this PR

**No runner is registered and no CI job has run on one yet**, so no workflow is pointed at a self-hosted label. ARC needs a GitHub App with organization *Self-hosted runners: Read and write*; the org's existing App does not have it — measured, not assumed: an installation token minted from it answers **HTTP 403 `Resource not accessible by integration`** on both `GET /orgs/{org}/actions/runners` and the registration-token endpoint. An App's permissions are not API-writable and App creation is a browser flow, so that step is the maintainer's. The environment-specific record and the exact finishing steps are in the private deploy repo (Systemorph/Memex#238); the credential itself is tracked as Systemorph/Memex#239.

https://claude.ai/code/session_01CCM41UknXQwgXqH39tLo9u


rbuergi:
All three review findings taken — the priority one was a genuine correctness defect, not a wording nit.

**1 · `-1` → `-10`.** The entry said runners "run at a negative one"; the design and the architecture
page specify **-10**, and the difference is operational rather than cosmetic. The autoscaler's
`--expendable-pods-priority-cutoff` defaults to **-10**: below that cutoff a pod cannot grow the node
pool *and can be deleted mid-job*. An operator copying `-1` from this entry would get preemption
behaviour that does not match the page it links to. Now reads `priority **-10**`.

**2 · "its own namespace" → the namespace pair.** The design uses two — the controller in one, the
runners in another — so "stops at its own namespace" was ambiguous about which. Reworded to name the
pair, and the permissions sentence now says the controller *watches only its own runner namespace*,
which is the property that actually matters (it cannot see the portal namespaces at all).

**3 · `plugins` → `MeshWeaver.Plugins`.** Repository name, now spelled as it is elsewhere on the page.

**Also merged `main`** (this PR had gone `DIRTY`). One conflict, in `Architecture.md`: my side adds
`Self-hosted CI runners on AKS`, main had gained `NuGet Package Retirement` and
`What a Synthetic Probe May Assert` in the same table row. Resolved by taking **main's row and
appending my entry** — all three survive; I asserted that programmatically rather than eyeballing it,
and confirmed each link target exists (`SelfHostedRunners.md`, `NuGetPackageRetirement.md`,
`SyntheticProbeTargets.md`).

**Verified locally:** `MeshWeaver.Documentation.Test` builds Release `-warnaserror` with
`0 Error(s)`, and `DocumentationLinkIntegrityTest` + `WhatsNewEntryIntegrityTest` pass **2/2**.

Picking this up because the session that opened it had finished; nothing else about the change has
moved.

Issue #3670, opened 2026-09-08
memex-portal-ai has no 'latest' tag while mw-plugin-test does — every release-following bake fails with no reference set


rbuergi:
`memex-portal-ai` carries **no `latest` tag** in ACR while `mw-plugin-test` does, so every
release-following bake in every satellite fails at preflight with no reference set. Measured
2026-09-08 ~01:17Z.

## The failure, as the gate reports it

MeshWeaver.Reinsurance `main`, runs `34175767679` and `34175763793`, both
`event=repository_dispatch`, job `Required CI inputs`, step *"Resolve the bake target"*:

```
release image meshweaver.azurecr.io/mw-plugin-test:latest resolves to sha256:1807b71d38752f5c…
##[error]'meshweaver.azurecr.io/memex-portal-ai:latest' returned no digest — the bake has no
         reference set. The tester (mw-plugin-test:latest) resolved, so the promoted set is
         incomplete or the portal repository name changed.
```

The gate names its own two hypotheses. **It is the first one.**

## Measured directly against the registry

```
$ az acr repository show -n meshweaver --image mw-plugin-test:latest --query digest
sha256:476244f72f956c854404640e8a45022376e65f4d5d11d8cd07a5e2d2fcf322fb

$ az acr repository show -n meshweaver --image memex-portal-ai:latest --query digest
ERROR: the specified tag does not exist.
```

`memex-portal-ai` is not empty — its newest tags are `3.0.0-ci.8066`, `main`, `52c6103-p857c921`,
`52c6103`, `staging-52c6103-34173814738`. There is simply **no `latest` among them**, while the
sibling tester repository has one. So one half of a promoted pair carries the moving tag and the
other does not.

## Blast radius — precise, not fleet-wide panic

Every satellite lane references this image (`memex-portal-ai` occurrences in `main`'s `ci.yml`:
Plugins 9, Education 8, Reinsurance 4, SocialMedia 2, Manufacturing 2, Crm 2), **but the `latest`
form is only reached when `FOLLOW_RELEASE=true`** — the release/publication-wave dispatch path. A
`push` or `pull_request` run bakes against the pinned digests (`MW_IMAGE_DIGEST` /
`MW_PORTAL_IMAGE_DIGEST`) and is unaffected.

So: **release-following bakes are broken fleet-wide; ordinary PR CI is not.** That matches the
observed pattern — Reinsurance's PRs are green while its `main` is red on two dispatch runs.

[...]

PR #3671, opened 2026-09-08
fix(#3578): the prod probe asserts the platform floor and the target's own declaration

### LINKED ISSUE #3578: memex.systemorph.com: the Prod synthetic probe names AgenticPrimer, which is not installed there — decide whether to install it or repoint the probe
## The whole `/api/content/**` route 404s on `memex.systemorph.com` — not one missing asset

`Prod synthetic probe` has failed **59 of its last 60 runs**. Last green:
**2026-09-06T10:46:07Z** — so it has been red for roughly **22 hours**, continuously.

It is not a required check and it runs on a schedule against `main`, so nothing surfaces it. That
is the same "a red in an empty room" shape as #3568, which is how I found this: the probe is
correct, loud, and unwatched.

## What the probe says, and why it under-describes the fault

```
sample 1: 404  https://memex.systemorph.com/api/content/AgenticPrimer/content/og.png
sample 2: 404  …
sample 3: 404  …
::error::content route unhealthy on memex.systemorph.com — 3/3 samples failed
```

Three clean 404s, never a 5xx and never a timeout, so it reads as *"that one asset is gone"*.
It is not. Measured just now:

| request | `memex.systemorph.com` | `memex.meshweaver.cloud` |
|---|---|---|
| `/healthz` | **200** | — |
| `/AgenticPrimer` (the page) | **200** | — |
| `/AgenticEngineering` (the page) | **200** | — |
| `/api/content/AgenticPrimer/content/og.png` | **404** | **200** |
| `/api/content/AgenticEngineering/content/og.png` | **404** | **200** |
| `/api/content/AgenticPrimer/content` | **404** | — |
| `/api/content/AgenticPrimer` | **404** | — |
| `/api/content/DoesNotExistXYZ/content/og.png` | **404** | — |

Three things follow, and the second is the one the probe cannot say:

1. **The portal is up.** `/healthz` is 200 and both space pages render.
2. 🚨 **The content route answers 404 for EVERY path, including ones that exist** — a real asset and
   a deliberately bogus one are indistinguishable. This is the route, not the content.
3. **The very same asset serves 200 on the other portal**, so the content itself is fine and the
   probe's own target is not stale.

## Probable cause — medium confidence, stated as a hypothesis

A route that answers 404 uniformly while the rest of the app serves normally is the signature of the
handler not being mapped at all — i.e. the content-serving module did not load on this deployment —
rather than of a lookup returning empty.

**Neighbour worth reading first:** #3485 — *"A shipped package is missing a source node on
memex.systemorph.com since 2026-08-26 and a reinstall did not restore it"*. Same host, same family
(something shipped is absent on this one portal and only this one). These may be one fault; I have
not established that, and saying they are without checking would be exactly the confident-wrong
answer this fleet keeps writing memories about.

## The discriminator for whoever picks it up

Ask the running pod which endpoints are mapped and whether the content-collections module loaded —
that separates "handler never registered" (hypothesis above) from "handler registered, index empty".
The outside-in evidence cannot tell those apart, because both answer 404.

## Scope

Filed rather than fixed: this is outside the work I was doing, it blocks no gate (the ruleset's only
required context on `main` is `Consolidate test results` — measured, not remembered), and it is a
deployment-state question rather than a code one. But it is production, it has been red for a day,
and the probe's message names one asset when the whole route is down — so the next person to look
would start in the wrong place.

rbuergi:
Fixes #3578.

## The red, and why it outranks its size

`Prod synthetic probe` has failed **59 of its last 60 runs** since 2026-08-31 because its matrix pointed `memex.systemorph.com` at `/api/content/AgenticPrimer/content/og.png` — a package that is not installed on the staff portal and never was. The portal was healthy throughout.

**A check red in an empty room is worse than no check.** After a day nobody reads it, and every real failure it would have caught arrives wearing the colour everyone has learned to ignore. Same class as a gate that skips.

## Why I did not take either option the issue offered

The issue framed it as *install the content, or repoint the probe*. Both put a fresh guess in the same slot.

A per-deployment install decision is **not a property of the platform**, and this repository cannot see which portal has what — the deployment inventory is deliberately private (`chart-drift.yml`: *"do NOT move the overlays into this public repo to dodge the credential — that would publish the deployment inventory"*). So a public workflow that hard-codes what a named portal serves is encoding a private fact it has no way to keep current. It could only ever drift, and when it drifted it blamed the portal. Installing course content on the staff portal to satisfy a monitor is the tail wagging the dog.

The probe had to stop guessing.

## What each deployment declares — measured 2026-09-08

`/sitemap.xml` is anonymous and is the only unauthenticated surface that says what a deployment serves (`SeoEndpoints` enumerates every top-level `Store/Plugin` / `Store/Catalog` / `Space` root passing `AnonymousGate`, per node, fail-closed):

| portal | roots declared | `AgenticPrimer` |
|---|---|---|
| `memex.systemorph.com` | **79** | absent |
| `memex.meshweaver.cloud` | **101** | present |

It already encodes the difference the matrix used to guess at.

## What the probe now asserts

Everything is either **shipped in the image** or **read from the deployment at probe time**. Nothing is hard-coded about what a named portal contains.

1. **Platform floor** — `/api/content/Doc/Architecture/content/platform-overview.svg`. An `EmbeddedResource` in `MeshWeaver.Documentation`, published on the content route by `AddDocumentation` (`isStatic: true` — the default is *not publishable, and the route answers 404*), anonymous-readable via `Doc/_Policy`. It resolves through `ContentFileResolver`'s same longest-node-prefix match and same collection-config request to the owning hub, so it crosses the hop the probe exists to watch while depending on no install decision. This is the only assertion that can answer *"is the route healthy"*.

2. **Negative control** — an impossible path under the same route that must **404**. Its absence is what made this issue unreadable: the old probe sampled **one URL three times** and reported `content route unhealthy`, but three samples of one URL cannot tell *"asset missing"* from *"route down"* — they are the same evidence. The issue was filed on exactly that wrong inference, and two subsequent readers each had to re-measure from scratch to establish the route was fine.

   | platform asset | control | verdict |
   |---|---|---|
   | 200 | 404 | green |
   | 404 | 404 | route discriminates ⇒ **content** red |
   | 5xx / timeout | any | **route** red |
   | any | not 404 | route answers the same for everything ⇒ **route** red |

3. **The deployment's own declaration** — read `/sitemap.xml`, sample roots from it by index (first / middle / last: deterministic so a red is reproducible, spread so it is not the same three alphabetically-first names on every portal), require 200. A portal is now only failed for contradicting **itself**, and content installed later is covered with no edit here.

## The empty denominator fails loudly

The root count is asserted **before** the sample loop — a loop over an empty list exits 0 having asserted nothing, and GitHub paints that green. Reading the declaration produces **three distinct reds**, deliberately not one message: the sitemap could not be fetched; the response was 200 but is not a sitemap; a well-formed sitemap declaring zero roots.

## Two defects found by falsifying the steps, not by reviewing them

- 🚨 **A 200 is not a sitemap.** The portal's SPA fallback answers **200 `text/html`** for any unrouted path, so *"the sitemap endpoint is gone"* arrived looking exactly like *"the portal publishes nothing"* — a statement about the portal rather than about the probe's own input. The step now asserts the shape (`<urlset`) before counting.
- 🚨 **`code=$(curl … || echo 000)` concatenates.** On a timeout curl prints its own `000` *and* exits non-zero, so the capture read **`000000`** — a six-digit status nobody can look up, produced in exactly the situation where the message is all the reader has. Fixed at all six call sites, including the pre-existing Store check.

## The guard

`SyntheticProbeNamesNoInstalledContentGuard` — the next wrong target will not look wrong; it will look exactly like a target that used to work. Review cannot see that difference.

- every `/api/content/…` path in the workflow is under `Doc/` or is the negative control (a citation on a comment line is allowed, so the file can name its own history);
- the matrix names hosts only — no paths;
- the control, the sitemap read and the zero-count assertion are all still present;
- and it **asserts the workflow file exists first** — a guard whose subject moved and whose own path did not passes having checked nothing.

## Verification

**The probe's own run blocks, executed against both live portals** (extracted from the YAML, run under bash 5):

```
memex.meshweaver.cloud  step0 EXIT=0  content route OK (platform asset 200, impossible path 404)
memex.meshweaver.cloud  step1 EXIT=0  all 3 sampled root(s) of 101 declared serve 200
memex.meshweaver.cloud  step2 EXIT=0  Store: 200
memex.systemorph.com    step0 EXIT=0  content route OK (platform asset 200, impossible path 404)
memex.systemorph.com    step1 EXIT=0  all 3 sampled root(s) of 79 declared serve 200
memex.systemorph.com    step2 EXIT=0  Store: 200
```

**Four failure branches, each exit 1 with the right message:** platform asset absent while the control still 404s ⇒ *content* red, not route; control pointed at a real asset ⇒ *route* red; 200 with zero `<loc>` ⇒ empty-denominator red; 200 that is not a sitemap ⇒ not-a-sitemap red.

**The guard falsified, five ways** — reintroducing the package target (1 fail), a path in the matrix (2), deleting the control (2), deleting the zero-count check (1), renaming the workflow (**3 — all of them**). Restored: 3 passed.

**Local gates:** `check-workflow-shell.py` → 0 live findings · `check-workflow-timeouts.py` → 77 jobs checked, 0 violations · `check-workflow-yaml-keys.py` → 31 workflows, 0 violations · `dotnet build -c Release -warnaserror` → **0 Warning(s), 0 Error(s)** · `DocumentationLinkIntegrityTest` + `WhatsNewEntryIntegrityTest` + the new guard → **5 passed, 0 failed**.

*(The heredoc-at-column-0 and `set -e` `[ … ] && arr+=(…)` bugs in my first draft were caught by running those linters locally — worth noting that they earned their keep here.)*

## Conserved

`Doc/Architecture/SyntheticProbeTargets` — *What a synthetic probe may assert*, linked from the Architecture index and from `ProbeSemantics`.

No What's New entry: pure-internal (CI monitoring), no user-visible effect.


rbuergi:
## Verified in real CI, not just locally — both portals pass

Dispatched this branch's version of the workflow through `workflow_dispatch` so it ran on GitHub's own runners rather than only against my extracted run-blocks:

**Run [34176822153](https://github.com/Systemorph/MeshWeaver/actions/runs/34176822153)** — `completed/success`

```
Probe memex.systemorph.com    completed/success
Probe memex.meshweaver.cloud  completed/success
```

For contrast, the two scheduled runs on `main` either side of it — same 15-minute cadence, the old workflow — are `completed/failure` (01:25Z and 00:58Z). So the eight-day red is green, in the environment that actually runs it, and green for a true statement rather than a lowered bar: the staff portal is still not asserted to have `AgenticPrimer`, because it does not.

## Added since opening: which half is wrong, the assertion or the environment?

Prompted by #3670, filed the same night. It is the mirror image of this issue and the two are easy to confuse, because from outside they look identical — *a check is red because the thing it names is not there* — while the correct repairs are opposites.

The discriminator now in the doc page: **name the mechanism that was supposed to make the assertion true.**

- `AgenticPrimer` on the staff portal was guaranteed by **nothing**. No mechanism produces it there. The *assertion* was the wrong half — repointing is the fix.
- `memex-portal-ai:latest` (#3670) **is** guaranteed — the promotion produces it, and the sibling `mw-plugin-test:latest` had one. The *environment* is the wrong half, and the bake preflight refusing rather than falling back to the pin is it working correctly.

So if you can name the mechanism, the red is a finding and repointing destroys it. If you cannot, it was a guess — and the replacement has to be something a mechanism does guarantee, or you have only moved the guess. That is the rule that makes this change a fix rather than a lowered bar, and it seemed worth writing down where the next person will meet it.

Issue #3675, opened 2026-09-08
Modules GC fails closed forever on a duplicate-proposal NOTICE from ModuleSetStore.Read, so set records and generations are never pruned (memex-cloud: 687 set files, 843 generations, /health probe 10 s)


rbuergi:
## Measured on memex-cloud (pod memex-portal-deployment-d6bfd5dcb-qk5gf, 8055, 2026-09-08 ~03:30Z)

- `/data/modules`: **843 generation directories** (22 per module — `MeshWeaver.AI.Anthropic@…` ×22, `MeshWeaver.Teams@…` ×22, …), 42,720 files; `/data/modules/sets`: **687 set records** (sequences 137 … 716, current pair `000000716-725a8a9bc07c61a4.{adopted,proposed}.json`); `activation.d`: 43 entries. On memex: 38 entries, 3,172 files.
- Every activation entry and every set record **parses** (`python3 json.load` over all of them: 0 bad).
- The pod's GC log, once per boot: `Modules GC: 16 activation entry file(s) could not be read, so the reference set is incomplete — SKIPPING every generation delete this pass (unreadable is never unreferenced, #2509).` The 16 `fail:` lines before it are all `ModuleSetStore.Read`'s **"Module set sequence N was proposed by more than one replica — the ordinally smallest id wins ('…'), and the next landing wave folds every landing back in. No module is lost; the mesh runs one set."** (sequences 319, 322, 323, 332, 339, …).
- `Modules GC: removed … superseded module set record(s)` never appears: `ModuleSetStore.Prune` runs only inside `CollectGarbage` under `readFaults == 0` (`ModuleLandingService.cs:240`).

## Cause

`ModuleSetStore.Read(baseDirectory, onCorrupt)` delivers the duplicate-proposal **notice** — a decided, harmless outcome — through the same `onCorrupt` callback `CollectGarbage` increments as a read fault. One duplicate pair on the volume therefore makes every GC pass fail closed **forever**: no generation is ever reclaimed (#2509's rule, applied to something that is not unreadable) and, because Prune is gated on the same counter, no set record is ever pruned either — so the pairs that trigger the notice are never removed, and the condition is permanent. The volume grows by one generation per module per landing wave and one set record per proposal.

[...]

Issue #2509, opened 2026-08-27
Module generations GC deletes lazily-loaded dependency DLLs out from under a running process; activation can point at a never-landed generation


rbuergi:
Prod, 2026-08-27 (~15:00Z), memex-cloud — chat via the OpenAI factory failed live: `Could not load file or assembly 'OpenAI, Version=2.12.0.0…'. The system cannot find the file specified.` for every OpenRouter-driven model (the whole tier roster).

Two defects, measured on the pod:

1. **GC deletes generations a RUNNING process still needs.** The process had loaded `MeshWeaver.AI.OpenAI@32f8e990` at boot (13:10Z ModuleLoad line); by 15:00 that directory was GONE from /data/modules while the process still served. The module's own DLL was safe in memory, but its dependency DLLs (`OpenAI.dll`) load LAZILY on first use — first chat → FileNotFound. `CollectGarbage`'s "nothing holds open (skip-on-locked)" is FALSE for not-yet-loaded dependency files: a generation any live process loaded its main DLL from must be GC-immune until process exit (e.g. pin via a per-process marker/lease, not open-file detection).

2. **activation.json referenced a generation that did not exist on disk** — `MeshWeaver.AI.OpenAI@8f4e6f6b` active while only `@0f9679fc / @27718c7b / @5356063c` existed (and `@27718c7b` was itself INCOMPLETE: no OpenAI.dll — a partial landing that was still eligible for activation reference). Same dangling shape on systemorph (`@661a3510` active, only `@7124fd6d` on disk — its chat survived only because the loaded generation's dir was intact). Activation must advance only after the landing is verified complete (all closure files present), and a landing must be atomic (temp dir + rename).

Mitigation applied: rollout restart of memex-cloud — boot resolved to the complete `@5356063c` and chat recovered. systemorph left running (loaded generation intact) but will hit the dangling reference at next boot.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01QVbVvrHeU1gm8oq4PuMVjF

PR #3677, opened 2026-09-08
fix(#3675): a duplicate module-set proposal is a notice, and the modules GC counts only faults

rbuergi:
Stacked on #3676 (retarget to `main` once it lands). Closes #3675.

**Measured on memex-cloud (2026-09-08 ~03:30Z, pod qk5gf, 8055):** 843 generation directories (22 per module), 687 set records (100 duplicate sequences), 43 activation entries — every file parses. The GC's once-per-boot pass logged `16 activation entry file(s) could not be read … SKIPPING every generation delete`; the 16 were `ModuleSetStore.Read`'s *"sequence N was proposed by more than one replica — the ordinally smallest id wins … No module is lost"* notices, delivered through the same `onCorrupt` the GC counts as a read fault. So the GC failed closed on every boot, `Prune` (gated on the same counter) never ran, the pairs that trigger the notice were never removed, and reading the pile put `/health`'s set read at 9–10 s over the probe's 5 s — no new memex-cloud pod has passed its startup probe since 00:23Z (#3676 takes the probe path).

- `ModuleSetStore.Read(baseDirectory, onCorrupt, onNotice)`: the duplicate-proposal notice goes to `onNotice`; the two-argument overload forwards `onNotice: onCorrupt`, so `PendingModuleActivations` and `MemexConfiguration` keep every note they show today. Binary-compatible (the old signature stays a real overload).
- `CollectGarbage` counts only `onCorrupt` toward the fail-closed reference set, logs notices at Information, and its warning names "activation entry or set record file(s)".
- Tests (`ModuleSetConvergenceTest`): a duplicate proposal is reported on the notice channel and never on the fault channel, the two-argument overload still says it, an unreadable deciding record is still a fault; with a duplicate pair on the volume the GC still prunes the records below the current set and reclaims an orphan generation, and still reclaims nothing when a deciding record cannot be read (#2509 untouched).

Local, `-c Release -warnaserror`: convergence + store + pending-activation + floor classes green; `ConfiguredModuleActivationTest` + `PlatformUpdateKeepsThePluginThenTakesTheRebuildTest` 25/25.

The production volume itself still needs one pass to catch up: once this and #3676 run there, the next boot prunes 685 records and ~800 generations at SMB speed, off the readiness path. Doing the record prune by hand first is the maintainer's call (proposed on #3675, not run).

### DIFF
--- a/src/MeshWeaver.PluginCatalog/ModuleLandingService.cs
+++ b/src/MeshWeaver.PluginCatalog/ModuleLandingService.cs
@@ -226,12 +226,18 @@ public sealed class ModuleLandingService : IDisposable
         // the 2026-08-27 outage from the other side. Both retained sets count (the one the mesh is
         // on and the one it has proposed), and a set directory that cannot be read counts as a read
         // fault for the same fail-closed reason the entries do.
+        // 🚨 #3675: a record that could not be READ is a fault; a sequence that two replicas
+        // proposed is a NOTICE — decided deterministically, nothing lost. The store used to deliver
+        // both through one callback, so a single duplicate pair on the volume made this pass fail
+        // closed forever (memex-cloud, 2026-09-08: 100 such sequences, 687 records, 843 generation
+        // directories, never one reclaimed). Only the fault channel counts here.
         var setIndex = ModuleSetStore.Read(baseDirectory,
-            msg =>
+            onCorrupt: msg =>
             {
                 readFaults++;
                 logger?.LogError("{Message}", msg);
-            });
+            },
+            onNotice: msg => logger?.LogInformation("Modules GC: {Message}", msg));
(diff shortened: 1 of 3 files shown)


rbuergi:
Reviewed (comment; one identity fleet-wide): the split is the right shape — the two-argument `Read` forwarding `onNotice: onCorrupt` keeps `PendingModuleActivations`/`MemexConfiguration` byte-for-byte on today's notes, the conflict loop is the only line touched inside my reader, and the GC counts only faults while logging notices at Information. Auto-arm does not arm a PR whose base is a feature branch, so I merge this into `fix/module-set-store-reads-only-what-it-decides-on` myself when its Build and Test is green; #3676 then carries both to main in one queue run.

Issue #3510, opened 2026-09-07
A package root is disposed while its own install is in flight (CD 7950): the creates' write-through streams lose their owner, no reply is ever sent, and the install runs out its bound — neither #3501 nor #3505 covers it


rbuergi:
## Measured: CD 7950 (`f55b9f4ab`, carries #3480 and #3475, neither #3501 nor #3505) failed its seal on a THIRD mechanism

`[FAIL] Hosting — install: TimeoutException` after the install's own 10-minute bound (`── Hosting: installing 145 file(s)…` at 23:36:36Z; the next package starts at 23:46:36Z). Every discriminator that names the two mechanisms fixed tonight reads **zero** over the whole bake log: `VERDICT_TIMEOUT=0` (#3499/#3505), `resubscribe failed=0` (#3498/#3501), `LATE_NACK_REENQUEUE=0`, `OWNER_NACK_REENQUEUE=0`, `PATCH_FLUSH_FAULTED_SYNC=0`. The control 7938 (sealed) reads zero on all of them too — so this run's failure is something else, and the trail says what.

## The trail (request-fate ledger, printed by `[STALE-CALLBACK]`)

Four creates for Hosting's own nodes, all `CreateOrUpdateNodeRequest → portal/nodeops-1SDY…`, pending 30–60 s:

```
Hosting/Backup: … ROUTED onTarget=True state=Submitted@portal/nodeops(+5518ms) → HANDLER_ENTER → UPSERT_READ existing → update
  → UPSERT_WRITE_THROUGH_STREAM path=Hosting/Backup/_Activity/compile-state → HANDLER_EXIT state=Processed(+5519ms)
  ⇒ a handler was entered and no reply, completion or fault has been recorded since
(same for Hosting/BackupStore, Hosting/ConfigAudit, Hosting/Deployment)
```

and, in between, the owner of every one of those paths stops answering:

```
23:37:44  [UpdateQueue] ADVANCE_WITHOUT_HANDOFF path=Hosting/Admin/_Activity/compile-state       seq=338 bound=5000ms — the owner never acknowledged this write
23:37:44  [UpdateQueue] ADVANCE_WITHOUT_HANDOFF path=Hosting/ConfigAudit/_Activity/compile-state seq=339
23:37:44  [UpdateQueue] ADVANCE_WITHOUT_HANDOFF path=Hosting/BackupStore/_Activity/compile-state seq=340
[...]

PR #3681, opened 2026-09-08
fix(#3680): the installer stops healing a partition the gating reconcile owns

rbuergi:
Fixes the seal failure in #3680 (CD run [34190841613](https://github.com/Systemorph/MeshWeaver/actions/runs/34190841613), commit `4fedf454`).

## Which delivery broke

**Neither of the two the issue names — the third one.** `preflight`, every image job, `promote`, `verify-images` and `notify-platform-update` were all **green**: a complete image set was published and promoted, and the release event did reach the control instance. The red was `Plugins: bake + seal … / Bake + publish NodeType assemblies to portal storage`, and `Register the publication with memex` was **skipped**.

So the set is **promoted but not SEALED** — the images exist, the event went out, and no install can adopt them (`heldReason: no sealed content bake …`). Not the budget cap: the failing job executed 37 steps, and September spend is ≈ $2,482 of $3,000.

## Root cause

The bake itself was fine — `compile: 89/89 NodeType(s) compiled`. The `--seed` gate failed on package-install **idempotence**:

```
[FAIL] Chess (34 node(s), 3 type(s))
    idempotence: re-install failed: InvalidOperationException: Install of 'Chess/_Policy'
    failed: Inner UpdateNode faulted: MeshNode Unknown at 'Chess/_Policy':
    TimeoutException: The operation has timed out.
GATE FAILED — idempotence: Chess — 1 new failure(s), 0 stale allow entr(ies).
```

**Two components own a partition's `_Policy` and its children's `_Access` denies, and each undoes the other.**

| Writer | Lives in | Reads | Runs |
|---|---|---|---|
| `PackageInstaller.EnsureDeclaredAccess` | this repo | the package **manifest** | per install + boot repair |
| `PluginGate.SeedGating` | MeshWeaver.Plugins — **in-mesh source** | the **root node's** `PluginContent` | every root activation + every subtree change |

For a free, non-pre-installed package with no `publicSegments` (Chess: `preInstalled` absent, `publicSegments: []`, `price: 0`) they state opposite rules in their own comments — core: *"a free package that a catalog hands out must be readable by everyone"*; the gating reconcile: *"there is no open-content tier … (The old rule left non-purchasable plugins' content readable — guides, dashboards, even imported personal LinkedIn data were world-visible)"*.

So on the idempotence re-install core retired **26** denies and reopened the policy, `PluginGating` immediately logged `Chess: reconcile is NOT CONVERGING — rewrote …, which this hub already wrote`, and the heal's own `Chess/_Policy` write starved behind the contention for 20 s and faulted. That is the write in the error message.

**The control confirms it.** Run [34186868747](https://github.com/Systemorph/MeshWeaver/actions/runs/34186868747), 65 minutes earlier on the **same commit**, *sealed* — while doing the same thing at a smaller amplitude: `retiring 9` instead of `retiring 26`, and no `NOT CONVERGING`. Identical content, different count. A steady state that differs run to run on unchanged inputs is a race, and the count is just how far one writer got before the other looked.

(`Chess/_Policy` has been measured at version **210,801** before, on 2026-08-25 — but that is **not** this defect: `PluginGate`'s own comments attribute it to the reconcile writing off blind subtree snapshots, since fixed. It is cited here only as evidence that these nodes are a contended surface with more than one way to loop.)

## The survey mistake

The heal's guard was the fingerprint **plus a claim**:

> "Current code cannot produce that (the scoped branch requires `declared.Count > 0`); only a pre-#902 installer could."

True of this repository's code. The producer is **in-mesh source** — a `.cs` node compiled at runtime in the portal: no `dotnet build` here compiles it, no core test executes it, no `grep --include='*.cs'` over this repo can see it. AGENTS.md states this blind spot for *deletions* of public surface; it applies identically to any premise of the form *nothing writes this any more*.

## The change

The legacy heal (deny sweep + policy rewrite) now runs **only on a pre-installed partition** — the case the #902 incident was actually about ("its 8 pre-installed partitions carried 136 legacy denies, while the partitions installed after #902 were correct"), and the one case where the two components **agree**: `SeedGating`'s pre-installed arm retracts the very denies this sweep retires, so nothing fights and the heal sticks.

- The **create-only** path is untouched — a create cannot loop, so a partition with no policy still gets one.
- A shipped gated policy still survives untouched.
- This does **not** decide whether a free plugin should be world-readable. It decides **where that question is answered**: in one component, so the answer can be changed by changing one rule instead of by winning a race.

## Behaviour change, stated plainly

A free, non-pre-installed plugin's partition previously oscillated: reopened by every install and every boot repair, re-gated by the reconcile seconds later. It is now consistently gated by the Store model, and the brief post-boot window in which its content was readable by anyone closes. Nothing that was deliberately set is touched, and nobody who could already read a plugin's content loses access.

## 🚨 Cross-repo: break shape 7, which no gate sees

This changes BEHAVIOUR behind an unchanged signature (#3276's shape). It removes no public type or member and adds no interface member, so no `Pairs-with:` / `Implementers:` line applies — and no gate will catch it either.

`MeshWeaver.Plugins`' `src/MeshWeaver.PluginCatalog.Test/DeclaredAccessInstallTest.cs` has `LegacyScopedGate_OnAFullyPublicPackage_IsHealed`, which asserts the heal fires on a **free** package (`FreePlug`). That test encodes the same false premise and must move to a pre-installed package. It builds against `MW_PLATFORM_REF` (a pinned core commit), so merging this reds nothing immediately — it would red at the next pin bump. **A follow-up PR in MeshWeaver.Plugins updates that test and adds a case pinning that a free plugin's gating is NOT healed; it must land before the pin crosses this commit.**

## Verification

- `dotnet build src/MeshWeaver.PluginCatalog/MeshWeaver.PluginCatalog.csproj -c Release -warnaserror` → `Build succeeded. 0 Warning(s) 0 Error(s)`, 9 projects, 15.2 s.
- `dotnet build test/MeshWeaver.Documentation.Test -c Release -warnaserror` → clean; new doc confirmed embedded in the built `MeshWeaver.Documentation.dll`.
- `DocumentationLinkIntegrityTest` → `Passed! Failed: 0, Passed: 1`.

## Also in this PR

- `Doc/Architecture/PartitionAccessOwnership` — who owns a partition's access shape, the measured ping-pong, and the survey rule.
- A What's New entry (`Category: Fix`) for the user-visible half.

## Deliberately not in this PR

The **recycle race** is a separate defect and is reported on #3510, whose central open question this run answers: `[PackageInstaller] recycling root Chess …` names the disposer as the installer's own retyped-root recycle (`SettleRetypedRoot`), contradicting #3499's stated hypothesis that it was *"not the installer's retyped-root recycle"*. Fixing that alone would **not** have fixed this red — the heal-vs-gating ping-pong is independent of it.

### DIFF
--- a/src/MeshWeaver.Documentation/Data/Architecture/PartitionAccessOwnership.md
+++ b/src/MeshWeaver.Documentation/Data/Architecture/PartitionAccessOwnership.md
@@ -1,0 +1,100 @@
+---
+Name: Who Owns a Partition's Access Shape
+Category: Architecture
+Description: Two components write a plugin partition's _Policy and its children's _Access denies — the installer in this repository and the Store's gating reconcile, which is in-mesh source in another one. When both claimed ownership they undid each other on every pass, and the ping-pong failed a CD seal. The rule that settles it, and the survey mistake that produced it.
+Icon: <svg xmlns="http://www.w3.org/2000/svg" width="24" height="24" viewBox="0 0 24 24" fill="none" stroke="currentColor" stroke-width="2" stroke-linecap="round" stroke-linejoin="round"><path d="M12 22s8-4 8-10V5l-8-3-8 3v7c0 6 8 10 8 10z"/><path d="M9 12l2 2 4-4"/></svg>
+---
+
+# Who Owns a Partition's Access Shape
+
+**A partition's `_Policy` node and its children's `_Access` deny assignments have exactly ONE
+owner. Where two components both wrote them, neither converged, and the loser was the platform.**
+
+Two components write that shape today:
+
+| Writer | Lives in | Reads | Runs |
+|---|---|---|---|
+| `PackageInstaller.EnsureDeclaredAccess` | this repository, `src/MeshWeaver.PluginCatalog/` | the package **manifest** (`preInstalled`, `price`, `publicSegments`) | once per install, and on the boot repair pass |
+| `PluginGate.SeedGating` | MeshWeaver.Plugins, `Store/Licensing/Source/` — **in-mesh source** | the **root node's** `PluginContent` | on every plugin-root activation, and on every subtree change |
+
+For a **pre-installed** partition the two agree: the installer publishes it fully public, and the
+gating reconcile's pre-installed arm retracts child denies. Nothing fights.
+
+For a **free, non-pre-installed** package that declares no `publicSegments`, they used to disagree
+outright — and both were emphatic about it in their own comments:
+
+> *the installer:* "a free package that a catalog hands out must be readable by everyone, signed in
+> or not."
+
+> *the gating reconcile:* "the cover + declared public segments are the ONLY public surface … there
+> is no open-content tier: a product without a price simply has no self-service way in yet. (The old
+> rule left non-purchasable plugins' content readable — guides, dashboards, even imported personal
+> LinkedIn data were world-visible via the cover grants.)"
+
+## What the disagreement cost
+
+The installer's legacy heal treated the pairing *policy-withholding-public-read + Public/Anonymous
+child denies* as damage left by a pre-#902 installer, retired the denies and reopened the policy.
+That pairing is precisely what the gating reconcile writes, on purpose, every pass. So each pass of
+each component undid the other:
+
+- the installer retires N denies and opens `_Policy`;
+- the gating reconcile re-denies and re-gates, logging
+  `[PluginGating] <plugin>: reconcile is NOT CONVERGING — rewrote …, which this hub already wrote`;
+- the next install repeats it.
+
+Measured on CD run `34190841613` (2026-09-08), package `Chess`: the idempotence re-install retired
+**26** deny assignments, the reconcile immediately rewrote them, and the heal's own `Chess/_Policy`
+write starved behind the contention for 20 s and faulted —
+`MeshNode Unknown at 'Chess/_Policy': TimeoutException`. That failed the package-install idempotence
+gate, so `Plugins: bake + seal` went red and `Register the publication with memex` never ran: a
+complete image set was promoted and verified, and **no installation could adopt it**, because a
+promoted set that is not sealed is held.
+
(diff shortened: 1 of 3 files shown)


rbuergi:
Cross-repo half filed as Systemorph/MeshWeaver.Plugins#1506 — it carries the exact diff and must ride the `MW_PLATFORM_REF` bump past this PR's merge commit (the test builds against the pinned core, so nothing is red until the pin moves).

It also records why that suite could never have caught this: every fixture root is `"nodeType":"Space"`, and `AddPluginGating` is registered on the `Store/Plugin` type — so `PluginGate.SeedGating`, the other writer of `_Policy`, never runs in it. A contradiction between two writers is invisible when only one is present.

rbuergi:
Coverage check — the heal fired on **four** packages in the failing run, not just Chess:

```
Cornerstone  retiring 1     RemoteControl  retiring 3
RolePlay     retiring 10    Chess          retiring 26
```

All four are free, non-pre-installed, no declared `publicSegments` (measured from each package's `index.json` on MeshWeaver.Plugins `origin/main`):

| package | preInstalled | publicSegments | price |
|---|---|---|---|
| Chess | absent | `[]` | 0.0 |
| RolePlay | absent | `[]` | 0 |
| RemoteControl | absent | `[]` | 0.0 |
| Cornerstone | absent | absent | absent |

So this change stops the fight for **every** heal observed in that run, not just the one that timed out — and it is self-consistent in the other direction too: for a genuinely pre-installed partition `SeedGating`'s pre-installed arm retracts the same denies, so `ContradictingDenies` finds none and the heal is a no-op there anyway. The heal only ever had work to do where the two components disagreed.

Also checked: `VerifyDeclaredAccess`'s post-condition tests `persistence.Exists(marker)` only, and for these packages the marker is `{partition}/_Policy` — which still exists (the create-only path is untouched). No new "declared access never converged" error.

PR #3698, opened 2026-09-08
fix(bake): an empty source set needs a positive witness — #3663's three-hour readiness stall

### LINKED ISSUE #3663: 🚨 Every rollout since 3.0.0-ci.8059 stalls at the bake gate: #3628's MeshNodeContentDegradedException fires during the NodeType bake for instance nodes typed by their own NodeType's source (4 Doc types 'regressed')
## What happened

Rolling the fleet to **3.0.0-ci.8059** (core a57a6b52b × plugins a7abdb2c) on 2026-09-08 00:22–00:50Z: on **memex.systemorph.com** the new pod (`memex-portal-deployment-7d5d458cc4-cbztk`) never passes its startup probe — `/health` answers **503** from `nodetype_bake`:

```
Health check nodetype_bake with status Unhealthy … NodeType bake regressed on this image — refusing readiness so the rollout stalls with the previous image still serving. 4 NodeType(s) regressed on this image: Doc/Architecture/BusinessRules/Cession, Doc/DataMesh/PythonPandasNode/PandasExplorer, Doc/DataMesh/SocialMedia/Post, Doc/DataMesh/SocialMedia/Profile
```

The same pod's log names the cause for each of the four — one instance node per type, read during the bake:

```
MeshWeaver.Mesh.MeshNodeContentDegradedException: MeshNodeStreamCache.GetStream: content for 'Doc/Architecture/BusinessRules/Cession/MotorXL' (nodeType 'Doc/Architecture/BusinessRules/Cession') stayed an untyped JsonElement — the $type discriminator resolved to no registered CLR type on the reading hub and the mesh-wide content-type registry did not recover it …
  … 'Doc/DataMesh/PythonPandasNode/PandasExplorer/LiveFrame' ($type PandasExplorer)
  … 'Doc/DataMesh/SocialMedia/Post/Post-001' ($type SocialMediaPost)
  … 'Doc/DataMesh/SocialMedia/Profile/Roland-LinkedIn' ($type SocialMediaProfile)
```

The gate did its job: the old 8055 pods keep serving, the deploy run (Memex 34173391343) reported the stalled rollout red at its 25-minute budget. **Nothing broken is serving.** But the release is un-rollable.

## Root cause

**8055 (core 7bfe2d907) rolled clean on the same portal 2 h earlier** (Job 36 complete, 3/3 ready). The only core change between the sets is **#3628** (`7bfe2d907..a57a6b52b`, "the untyped-content gate can fire"), which introduced `MeshNodeContentDegradedException` and throws it from `MeshNodeStreamCache.ConvertContentJsonElementToTyped` when content stays a `JsonElement`. The four regressed types are exactly the doc types whose instance nodes carry a `$type` declared in the type's OWN `Source/` (`CessionData`, `PandasExplorer`, `SocialMediaPost`, `SocialMediaProfile`): during the bake, the instance is read on a hub where that type is not yet registered (the type is what the bake is compiling), the content is legitimately untyped for that moment, and what used to be a silent degradation is now a throw inside the bake — which the sweep scores as a regression. `LateContentTypeRegistrationTest` (added by #3628) was meant to cover the late-registration case; the bake path is a case it does not model.

Both portals show it: memex-cloud's 8059 pod logs the same exception for `Posts/ClaudeAbstractsErrors` (`SocialMedia/Post`), plus `CS0246: 'CessionData' could not be found` on `Doc/Architecture/BusinessRules/Cession` — the same type, failing to compile its own source in the same window.

## What needs to happen

The throw must not fire for content whose type is registered by the NodeType being baked — the bake must read those instances after their type compiled, or the conversion must degrade (as before) rather than throw while that type's compilation is in flight. Until then every set ≥ a57a6b52b stalls every rollout at the bake gate; the fleet stays on 8055 (whose migration image lacks Plugins#1488's budget — the databases are already migrated to v55 by the 8059 Job, so 8055 pods serve fine).

Evidence: Memex runs 34173111536 (memex-cloud) and 34173391343 (memex); pods `memex-portal-deployment-7d5d458cc4-cbztk` (memex) and `memex-portal-deployment-8b96fdbd4-h7cxj` (memex-cloud).

rbuergi:
Fixes #3663.

## The stall, and the one number that explains it

`BatchBake` logs the size of every source-discovery pass. Every boot on **`memex` (memex.systemorph.com — the namespace #3663 opens with)**, 09-07 22:40Z → 09-08 07:37Z:

| boot (UTC) | Code nodes resolved | `compileErrors=` | readiness |
|---|---|---|---|
| 09-07 22:40:02 | 1236 | 2 | granted |
| 09-07 22:43:42 | 1236 | 2 | granted |
| **09-08 00:31:22** | **1145** | **6** | **REFUSED** |
| 09-08 03:32:37 | 1237 | 2 | granted |
| 09-08 03:35:30 | 1237 | 2 | granted |
| 09-08 07:34:44 | 1241 | 2 | granted |
| 09-08 07:36:57 | 1241 | 2 | granted |

**Denominator: 29 boots across both portals in 15 hours; exactly one refused**, and three of the six that granted readiness ran the *same image*. `2` is that portal's standing baseline of already-broken types, which `MarkOutcome` correctly refuses to let gate. The four extra are the entire `Doc` partition's NodeTypes — **4 of 4**, not 4 of 40. `memex-cloud` never shows it (6837 → 6841 nodes; `compileErrors=` steady at 4–5 either side of 8059); its simultaneous stall was the 10-second `pending_module_activation` check over 687 set records, i.e. #3676, a different defect conflated into the same issue.

## The defect

On the short pass those four types matched ZERO sources. Roslyn was handed each type's generated provider file alone and emitted a completely honest `CS0246 'CessionData' could not be found` about a type declared in that very NodeType's own `Source/`. The gate did exactly its job on a genuine `CompileError`.

The invariant written to stop precisely this (`Assemble` → `SourceDiscoveryFailedException`, #1216, whose doc comment even quotes the identical `CS0246 'ScopeLibrary'` shape) did not fire:

```csharp
if (matched.Count == 0 && pending.DeclaresSources && !pending.SourcesKnownDeleted)
```

`DeclaresSources` is false for a type on the DEFAULT queries — which is very nearly the whole population, exactly as `DynamicTypePreWarmer.ClassifyCompileFailure` had already had to be corrected to say in **#1391** (*"an empty `Sources` does not mean configuration-only — it means uses the DEFAULT queries"*). The same conjunct sat here in the opposite direction. Verified on the live node: `Doc/DataMesh/SocialMedia/Profile` has no `sources` and a populated `currentSourceVersions` naming its 2 files — the mesh's own record said "two source files" while the pass matched none, and nothing read it.

## The change

`NodeTypeBatchBake.DiscoveryUnestablished` is now the one predicate, and it asks a **positive** witness rather than the absence of one. `PendingType` carries `KnownSourceCount` (`CurrentSourceVersions?.Count`) in place of the boolean, so all three shapes stay distinguished:

- `0` (explicitly empty) — the mesh agrees; established, classifies `NoSources`, still must not gate (**#1204**);
- `n > 0` — the record contradicts the zero match; **unestablished**, batch abandoned for the activation-driven sweep;
- `null` (absent) — no witness; falls back to the pre-existing declared-queries rule, unchanged.

It cannot launder a real regression: one matched source makes the set established, the compile runs, and a genuine failure gates as before.

## Why the stall lasted three hours — worth recording

`RetractRegression` is level-triggered on the type *recompiling*, and nothing recompiled (the content never changed, so the park registry's source-change retry never fired). What ended it was the kubelet: `startupProbe` is `failureThreshold: 1080` × `periodSeconds: 10` = exactly 3 h, and the container died at 03:29:52Z having started at 00:30:23Z (2 h 59 m 29 s). **#3676 did not clear it** — falsified, not assumed: the recovery is at 03:33Z, 47 minutes *before* #3676 merged, on `3.0.0-ci.8059`, which cannot contain a later commit. Had #3676 been the cause the recovery would have had to postdate that merge and arrive on a later image.

## Verification

- `EmptySourceSetNeedsAWitnessTest` (4): the regression case; #1204's corroborated-empty case in both declared and default form; the absent-snapshot fallback in both directions; and three shapes of "any matched source is always established".
- **Control arm:** the same test class against the pre-#3663 predicate — **1 of 4 fails** (the regression case, `Expected: True / Actual: False`), the other 3 pin unchanged semantics on both.
- `MeshWeaver.Hosting.Test` **461/461** with the fix; **460/461** with the control arm, the single failure being that one case.
- `DocumentationLinkIntegrityTest` + `WhatsNewEntryIntegrityTest` 2/2.
- `dotnet build -c Release -warnaserror`, one project per invocation, `0 Warning(s) 0 Error(s)`: `MeshWeaver.Hosting`, `MeshWeaver.Documentation`, `MeshWeaver.Hosting.Monolith`, `MeshWeaver.Hosting.Orleans`, `MeshWeaver.Hosting.Test`, `MeshWeaver.Documentation.Test`.

## Left open deliberately (not this PR)

The same boot logged `ShippedPrebuiltBundles: bundle Doc.zip: adopted 4/4 prebuilt assembly(ies)` at 00:31:08, and ten seconds later the sweep enumerated `204 of 209 … need building — 5 already on the share` and recompiled them anyway. Adoption seeding and the sweep's store probe disagree on a cold boot; that is what put these four on the compile path at all. It costs time, not correctness, and is recorded in the doc page rather than fixed here.

Doc: `Doc/Architecture/SourceSetEstablishment` — the three shapes of zero, the measurement above, and the two LogQL queries to re-measure with. What's New (Fix): `2026-09-08-a-short-read-no-longer-condemns-a-healthy-node-type`.

Pairs-with: none — no public type or member removed; the changed record is `private`, the new predicate is `internal`.
Implementers: none — no interface member added.
Mirror-sync: none — no localization catalog key added or reworded.


rbuergi:
The two open halves this issue named are now filed, so the chain is followable from either end:

- **#3704 — why the discovery pass was short.** The actual unexplained fact. Carries the 1145-vs-1236/1237/1241 table, the candidate (`RunQuery`'s 1-second quiet window; `QueryResultChange<T>` has no terminal marker), the discriminating measurement (per-query change-event count + largest inter-chunk gap, short pass vs complete pass **on the same portal**), and an explicit ruling-out of widening the window with the three reasons.
- **#3703 — the upstream half.** `Doc.zip: adopted 4/4` at 00:31:08, `5 already on the share` at 00:31:18, and the control that makes it a disagreement rather than a cold-store fact (a boot with `77 already current` sees 189 baked; a boot with `78 adopted now` sees 5). This is what put those four types on the compile path at all, so it is what made the short pass reachable.

#3698 fixed what a short pass may CONCLUDE. Neither of the above is fixed.

PR #3717, opened 2026-09-08
Publish: a supersession by this bake's own content is not a failure

rbuergi:
Two commits, both `Refs #3461`: the **false red that is failing core CD today**, and the **writer
half of the generation layout** that removes the race itself.

---

## 1 — A supersession by this bake's own content is not a failure

**Measured, 2026-09-08.** Core CD runs `34205409381` and `34206854855` both published `plugins` at
source `cfac152ef023bc8e16203511aa60b50f581d3161` for identity `s057b1e7785fba6c6ba079e9d84a0c00d`.
`BAKE_PUBLISH_TARGETS` holds two shares; **each run won one and each went red on the other**, with
the same verdict — *40 of 45 file(s) were overwritten … the remaining 5 are byte-identical*. Both
shares ended sealed with exactly the right bytes; both CD runs failed; neither produced a sealed set;
the platform pin behind them did not move.

The 40 are bundle zips and module packages — the compile is not reproducible byte for byte. The 5 are
`source-commit.txt`, `repository.txt`, `architecture.txt`, `modules/_index` and
`platform-surface.json`, byte-identical *because it is the same content*.

**The defect.** The two ends of `publish-bake-bundles.sh` answered the same question and disagreed.
The sealed-skip keys on **content × framework** and says a *different run's bytes for the same key*
are already-published. `verify_publication` can only ask *"are these MY bytes"*, so it called the
same state a failure.

**The change.** `ours = 0, foreign > 0` now **converges** instead of failing, on positive byte-level
proof of all three — nothing inferred, nothing waited for:

1. every **non-payload** file byte-identical to this bake's (foreign markers `0` **and** the
   neutral-marker count asserted against `MARKER_COUNT`, so it cannot pass having checked nothing);
2. every foreign file naming **one** publication, and that publication **stamped** (`<unstamped>` is
   not a legal token, so an unstamped file can never satisfy it);
3. `_complete` on the shelf, its digest the SHA-256 of **this bake's own listing**, sealed by **that
   same** publication.

Anything less is the superseded **red**, unchanged, with the reason printed. It never touches the
**mix** verdict, never seals, never deletes, and reports `targets-converged` separately from
`targets-published`.

## 2 — The writer learns the generation layout, behind a per-caller selector

Phase 2 of `Doc/Architecture/SealedPublicationGenerations`. Each publication is written under its own
run-unique token, sealed there, and a one-line `_current` pointer says which one applies — so two
publishers never write the same bytes and a mix becomes **unrepresentable** rather than merely
detectable.

🚨 **The obvious design was measured and rejected.** Staging plus an atomic directory rename is **not
available on this store through this client**. Read off azure-cli 2.90.0: `az storage file` has copy
· hard-link · metadata · symbolic-link · delete · delete-batch · download · download-batch · exists ·
generate-sas · list · resize · show · update · upload · upload-batch · url and **no `rename`**;
`az storage directory` has create · delete · exists · list · show — **no `rename`**, and its delete is
documented *"Delete the specified **empty** directory"*. Neither group has a `lease` command, so a
lease per identity is not reachable either. Even in REST, `Rename Directory` cannot **replace** an
existing directory, so the swap is rename-away plus rename-in — two operations with a gap in which
the live path does not exist. **The smallest write this store offers is one small file, and that is
what the pointer is.**

- Selector `publication-layout` on `node-repo-publish-bake.yml`, default `flat`, carried in as
  `BAKE_PUBLICATION_LAYOUT`. An unrecognised value is **refused**, never read as flat. **Nothing
  anywhere writes a generation until a caller opts in.**
- `bake-scope.sh` and `carry-forward-bundles.sh` resolve the pointer in the SAME commit, by the
  reader's rules, so the writer and the two Azure-direct readers cannot disagree about what is live.
- **One deliberate deviation**, documented in place: the pointer moves **before** the flat
  compatibility copy. "Last" is about the *generation*; the flat copy is a different audience and the
  one part another publisher can still be writing. Ordering it after the pointer means an overlap on
  the flat copy costs the *compatibility copy* instead of a publication that is already whole.
- **Retention is not implemented**, and that is a precondition on flipping, not a follow-up: the
  sweep deletes from the production share and must land as its own reviewed change. Nothing here
  deletes anything it did not create.

## The denominator

Of 30 core-CD runs examined, 3 had not reached the bake job, 16 were cancelled before it and 2
decided not to publish — **9 executed it**. Counting every publication of *either* lane that had a
same-identity run overlapping it in time (4 core-CD + 5 satellite `publish-bake`): **2 failures / 9
overlapping publications = 22%**, both being the two halves of the one incident above. 🚨 **All 9
overlaps were same-lane; zero cross-lane** — the same finding as 2026-09-06 on a different day, and
the second independent measurement saying "one owner per prefix" addresses none of this.

## What is still open, plainly

- The convergence verdict does **not** cover a sibling that has not sealed yet when this run's sweep
  ends — 21 seconds on one target of the incident, −0.24 on the other, so of those two reds one
  converges and one does not. More checking cannot close it; disjoint directories can.
- Flipping a prefix to `generation` needs **every producer of it** past this commit (`plugins` has
  two: core CD and the satellite), and even then the **flat compatibility copy still races** — the
  reds go when that copy goes, or when the publication becomes digest-addressed.
- `Doc/Architecture/SealedPublicationGenerations` now records what an **OCI registry** does and does
  not close: content-addressed blobs make a mix unrepresentable, but a **tag** is a mutable
  last-writer-wins reference and moves the defect unless the seal names **digests** — every bundle
  recorded by digest, each publication also given an immutable identity-qualified tag, and the sealed
  set a list of `(package, digest)` pairs. Plus the two invariants that migration must carry across:
  the **framework identity must stay in the reference** (a path like `plugins/<source>/<pkg>:<ver>`
  carries none, and two surfaces' bakes would then collide on one reference), and the `_releases`
  marker must survive.

**The refusal is not weakened anywhere.** A mix is still never sealed, and a sentinel covering one is
still removed.

## Verification (nothing here triggered a CD run)

- `test-publish-bake-overlap.py` — **67 assertions**, executing the REAL script, reading every
  verdict off the bytes on a stub share. Three convergence arms and five generation arms, including
  **two interleaved publishers each sealing their OWN generation with neither directory holding a
  byte of the other, both complete, and `_current` naming exactly one**.
  🚨 **Negative control: 16 of the 67 fail against `main`'s script.**
- `bake-scope.sh --self-test` — 22 cases. The pointer case is discriminating by construction: the
  flat copy and the generation record **different** baselines, so the verdict says which was read.
- `carry-forward-bundles.sh --self-test` — 12 cases, incl. a **decoy** bundle planted at the prefix
  under the same name and different bytes.
- `bash -n` × 3 clean · `shellcheck -S warning` × 3 clean · `check-workflow-shell.py`,
  `check-workflow-yaml-keys.py`, `check-workflow-timeouts.py`, `check-pr-secret-preflight.py` clean ·
  `DocumentationLinkIntegrityTest` passed ·
  `dotnet build test/MeshWeaver.Documentation.Test -c Release -warnaserror` → 0 Warning(s), 0 Error(s).

Refs #3461

### DIFF
--- a/.github/scripts/publish-bake-bundles.sh
+++ b/.github/scripts/publish-bake-bundles.sh
@@ -342,6 +342,55 @@ sha256_of() { # <file> — hex digest, portable across the CI runner and a devel
   fi
 }
 
+# 🚨 THE RENDERING IS MEASURED, NOT ASSUMED, and getting it wrong is silent. knack's `format_tsv`
+# treats a TOP-LEVEL list as ROWS: `--query "[a, b]" -o tsv` prints a's value and b's value on TWO
+# LINES, not as two columns — so `awk '{print $2}'` would read EMPTY for every file, every owner
+# would compare unequal to $PUBLICATION, `ours` would be 0 on every publish and the postcondition
+# would refuse EVERY publication in the fleet as "superseded". Hence a list-of-ONE-ROW: `[[a, b]]`
+# renders one tab-separated line. The `|| '-'` guards exist because a null field renders as the
+# LITERAL STRING 'None', not as empty. Both measured against azure-cli 2.90.0's own jmespath +
+# knack formatter, and re-asserted with the real jmespath engine by test-publish-bake-overlap.py.
+STAMP_QUERY="[[metadata.digest || '-', metadata.publication || '-']]"
+
+# The ONE read-back, in one place. Both the per-file verification sweep and the sentinel check the
+# convergence verdict makes parse the SAME rendering, so a query change cannot leave one of them
+# reading a shape the other no longer produces. Answers through globals rather than stdout: an
+# `::error::` written inside a command substitution would be CAPTURED instead of reaching the log.
+#
+#   rc 0  STAMP_DIGEST / STAMP_OWNER hold the values ('' where az rendered '-')
+#   rc 1  unreadable — a transient fault, an expired credential, or the file is simply not there
+#   rc 2  the CLI's rendering changed; STAMP_RAW/STAMP_FIELDS say what came back
+#
+# 🚨 `< /dev/null` because the sweep below runs this INSIDE a `while read` loop fed by the
+# manifest: a command that consumed stdin would swallow the remaining lines and the loop would end
+# early, having verified a PREFIX of the publication while reporting no foreign bytes at all.
+STAMP_DIGEST=""
+STAMP_OWNER=""
+STAMP_RAW=""
+STAMP_FIELDS=0
+read_stamp() { # <account> <share> <path>
+  local account="$1" share="$2" path="$3" props fields
+  STAMP_DIGEST=""; STAMP_OWNER=""; STAMP_RAW=""; STAMP_FIELDS=0
+  if ! props=$(az storage file show --account-name "$account" --share-name "$share" \
+      --path "$path" --auth-mode login --backup-intent \
+      --query "$STAMP_QUERY" -o tsv --only-show-errors < /dev/null); then
+    return 1
+  fi
+  STAMP_RAW="$props"
+  fields=$(printf '%s' "$props" | awk -F'\t' 'NR == 1 { print NF }')
+  STAMP_FIELDS="${fields:-0}"
+  # The field count is ASSERTED rather than trusted: if that rendering ever changes, this is a loud
+  # refusal on the next publish instead of a fleet-wide false verdict.
+  if [ "$STAMP_FIELDS" -ne 2 ]; then
+    return 2
+  fi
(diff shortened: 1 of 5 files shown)


rbuergi:
🚨 **Only part 1 of this PR merged.** The queue took it on `f9fbff6bc` (the convergence verdict) four
minutes before the second commit was pushed, so the body above describes two halves and `main` got
one.

Part 2 — the generation writer behind the `publication-layout` selector — is #3724, unchanged and
with the same verification. `f9fbff6bc` is on `main` (`git merge-base --is-ancestor`); `cb7b697db` is
not.

PR #3724, opened 2026-09-08
Publish: the writer learns the generation layout, behind a per-caller selector

rbuergi:
Phase 2 of `Doc/Architecture/SealedPublicationGenerations`. The second half of #3717 — it was pushed
four minutes after the merge queue took that PR, so it landed on the branch and not on `main`.

**It changes nothing anywhere until a caller opts in.** The selector defaults to `flat`, and the three
flat control cases in the overlap harness are the regression suite proving the default path is
byte-identical to what it was.

## Why

The #3496 postcondition turns an overlap from a silent mix into a loud refusal. It is a
**postcondition, not mutual exclusion**, and it leaves the interval between its last verification read
and the seal. This removes the shared directory instead: each publication is written under its own
run-unique token, sealed there, and a one-line `_current` pointer says which one applies. Two
publishers never write the same bytes, so **a mix becomes unrepresentable rather than merely
detectable**.

## 🚨 The obvious design was measured and rejected

Staging the publication and renaming it over the live one is **not available on this store through
this client**. Read off azure-cli 2.90.0 — the version the read-back query was measured against:

| group | commands |
|---|---|
| `az storage file` | copy · hard-link · metadata · symbolic-link · delete · delete-batch · download · download-batch · exists · generate-sas · list · resize · show · update · upload · upload-batch · url |
| `az storage directory` | create · delete · exists · list · show |

**No `rename` in either**, `directory delete` is documented *"Delete the specified **empty**
directory"*, and neither group has a `lease` command — so a lease per identity is not reachable from
this lane either. Even in REST, `Rename Directory` (2021-04-10) cannot **replace** an existing
directory: the swap is rename-away plus rename-in, two operations with a gap in which the live path
does not exist at all.

**So the smallest write this store offers is one small FILE, and that is what the pointer is.** A
rename design would have reintroduced a partial-visibility window wearing a better story.

## What is in it

- `publication-layout` on `node-repo-publish-bake.yml`, default `flat`, carried in as
  `BAKE_PUBLICATION_LAYOUT`. An unrecognised value is **refused**, never silently read as flat.
- `bake-scope.sh` and `carry-forward-bundles.sh` resolve the pointer **in the same commit**, by the
  reader's rules (`ShippedPrebuiltBundles.PublicationDirectoryOf`), so the writer and the two
  Azure-direct readers cannot disagree about which publication is live. `carry-forward` resolves it
  itself rather than being handed the directory — its caller may be pinned to a workflow copy that
  knows nothing about generations, and its one-publication postcondition is what pins it to a single
  generation if the pointer moves mid-read.
- **One deliberate deviation from the design page, documented in place:** the pointer moves **before**
  the flat compatibility copy, not after it. "Last" is about the *generation* — a reader must never be
  pointed at a directory still being filled in — and the flat copy is a different audience and the one
  part another publisher can still be writing. Ordering it after the pointer means an overlap on the
  flat copy costs the *compatibility copy* (refused as today; the run goes red) instead of costing a
  publication that is already whole, sealed and disjoint.
- 🚨 **Retention is NOT implemented, and that is a precondition on flipping rather than a follow-up.**
  Generations accumulate at ~45 small files each until a sweep removes the ones nothing names, and
  that sweep DELETES from the production share, so it must land as its own reviewed change with its
  own harness. **Nothing here deletes anything it did not create.**
- The doc also records what an **OCI registry** does and does not close, for the migration that is
  about to touch these files: content-addressed blobs make a mix unrepresentable, but a **tag** is a
  mutable last-writer-wins reference and moves the defect unless the seal names **digests** — every
  bundle recorded by digest, each publication also given an immutable identity-qualified tag, and the
  sealed set a list of `(package, digest)` pairs (the rule already in force for images via
  `MW_IMAGE_DIGEST`). Plus the two invariants that migration must carry across: the **framework
  identity must stay in the reference**, and the `_releases` marker must survive.

## What this does NOT do

- **Nothing writes a generation yet.** Phase 3 is each producer's `platform-ref` reaching this commit;
  phase 4 is flipping per prefix — `plugins` in ONE change set, because it is the only prefix with two
  producers (core CD's `plugins-bake` and the satellite's own `publish-bake`), and a new writer beside
  an old one on one prefix is the half-migration that serves stale silently.
- 🚨 **Flipping does not by itself stop the publish reds.** The flat compatibility copy is still
  replaced in place and still races. What the flip buys immediately is that the *publication* survives
  an overlap intact and pointed-to instead of being lost. The reds go when the flat copy goes
  (phase 5), or when the publication becomes digest-addressed.
- **The refusal is not weakened anywhere.** A mix is still never sealed, and a sentinel covering one
  is still removed.

## Verification (nothing here triggered a CD run)

- `test-publish-bake-overlap.py` — **67 assertions**, executing the REAL script against a stub share
  and reading every verdict off the BYTES. Five generation cases: one publisher seals under its own
  token and the pointer names it; **two interleaved publishers each seal their OWN generation, neither
  directory holds a byte of the other, both are complete, and `_current` names exactly one of them**;
  "already published" resolves *through* the pointer, not off the prefix; every unusable pointer shape
  (escaping, rooted, `..`, blank, dangling) degrades to the prefix; an unrecognised selector is
  refused. 🚨 **Negative control: 16 of the 67 fail against the pre-change script.**
- `bake-scope.sh --self-test` — the pointer case is **discriminating by construction**: the flat copy
  and the generation record *different* baselines (a diverged commit vs. an ancestor), so the verdict
  itself says which was read — a resolver that ignored the pointer answers `full`, one that follows it
  answers `narrowed` with the generation's commit. The dangling-pointer control asserts the fallback
  SAYS so, so silence cannot pass for resolution.
- `carry-forward-bundles.sh --self-test` — a **decoy** bundle is planted at the prefix under the same
  name and different bytes, so carrying the publication's own bytes forward is only possible by
  following the pointer.
- `bash -n` × 3 clean · `shellcheck -S warning` × 3 clean · `check-workflow-shell.py`,
  `check-workflow-yaml-keys.py`, `check-workflow-timeouts.py`, `check-pr-secret-preflight.py` clean ·
  `DocumentationLinkIntegrityTest` passed ·
  `dotnet build test/MeshWeaver.Documentation.Test -c Release -warnaserror` → 0 Warning(s), 0 Error(s).

Refs #3461

### DIFF
--- a/.github/scripts/bake-scope.sh
+++ b/.github/scripts/bake-scope.sh
@@ -72,23 +72,27 @@ if [ "${1:-}" = "--self-test" ]; then
   mkdir -p "$ST_TMP/bin"
   cat > "$ST_TMP/bin/az" <<'AZ'
 #!/usr/bin/env bash
+group="$2"          # file | directory
 verb=""
 for a in "$@"; do case "$a" in exists|download) verb="$a"; break;; esac; done
-account=""; share=""; path=""; dest=""
+account=""; share=""; path=""; dest=""; name=""
 while [ $# -gt 0 ]; do
   case "$1" in
     --account-name) account="$2"; shift 2;;
     --share-name)   share="$2";   shift 2;;
     --path)         path="$2";    shift 2;;
     --dest)         dest="$2";    shift 2;;
+    --name)         name="$2";    shift 2;;
     *) shift;;
   esac
 done
 file="$MOCK_AZ_ROOT/$account/$share/$path"
-case "$verb" in
-  exists)   { [ -f "$file" ] && echo true; } || echo false;;
-  download) [ -f "$file" ] || exit 1; mkdir -p "$(dirname "$dest")"; cp "$file" "$dest";;
-  *) exit 1;;
+dir="$MOCK_AZ_ROOT/$account/$share/$name"
+case "$group/$verb" in
+  directory/exists) { [ -d "$dir" ] && echo true; } || echo false;;
+  file/exists)      { [ -f "$file" ] && echo true; } || echo false;;
+  file/download)    [ -f "$file" ] || exit 1; mkdir -p "$(dirname "$dest")"; cp "$file" "$dest";;
+  *) echo "stub-az: unmodelled '$group $verb' — teach the stub rather than letting the script fall back" >&2; exit 64;;
 esac
 AZ
   chmod +x "$ST_TMP/bin/az"
@@ -127,6 +131,19 @@ AZ
     mkdir -p "$dir"
     printf '%s\n' "$sha" > "$dir/source-commit.txt"
     printf '%s\n' "$@" > "$dir/_complete"
+  }
+
+  # A publication under a GENERATION directory, with the pointer naming it (MeshWeaver#3461).
+  seal_generation() {  # <account/share[/base]> <generation> <source-sha> <bundle…>
+    local target="$1" gen="$2" sha="$3"; shift 3
+    local account="${target%%/*}" rest="${target#*/}"
+    local share="${rest%%/*}" base=""
+    case "$rest" in */*) base="${rest#*/}";; esac
+    local prefix="$ST_TMP/remote/$account/$share/${base:+$base/}prebuilt-bundles/ID1/plugins"
+    mkdir -p "$prefix/$gen"
+    printf '%s\n' "$sha" > "$prefix/$gen/source-commit.txt"
+    printf '%s\n' "$@" > "$prefix/$gen/_complete"
+    printf '%s\n' "$gen" > "$prefix/_current"
   }
 
   run_case() {  # <event> <targets> [<head-sha>]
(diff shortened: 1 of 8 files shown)


rbuergi:
## 🚦 Held as draft — a CD run on main's tip is gating a portal roll

Not a problem with this PR. CI can run; the hold is on the merge.

Core CD run **34216985031** started at 10:43:53Z on `da892b004` — the merge commit of #3717, main's
tip. That run matters twice over:

1. it is the run that produces the sealed set a third session is holding a portal roll on, which has
   not been produced since the 2026-09-08 double failure;
2. it is the **first real execution** of the convergence verdict that merged in #3717 — this script
   talks to Azure Files and runs only on `main`, so a production publish is the only place any edit
   to it executes.

Merging this PR moves `main`, starts a new CD run and **supersedes** that one. Of the 30 core-CD runs
swept for the denominator, 16 were cancelled exactly this way.

**Un-draft and merge once 34216985031 has produced its sealed set** (or has otherwise concluded).
Nothing here needs to change; the branch is complete and verified locally, and it is inert by
construction — `publication-layout` defaults to `flat` and no caller opts in.

Watch what 34216985031's `Plugins: bake + seal the publication for this identity` job prints: a
`Superseded by an equivalent publication` warning with `targets-converged=` in the summary line is the
convergence verdict firing for the first time, and a green job where the previous two runs went red.

rbuergi:
Both review findings were right and are fixed in `34579f0ee`.

The order was deliberately changed to move `_current` as soon as the generation is sealed and
**before** the flat compatibility copy — so an overlap on the flat copy (which still races) costs that
copy rather than a publication that is already whole, sealed and disjoint. Two statements were left
describing the old order:

- the layout selector's summary comment still said the pointer is *"moved LAST"*;
- 🚨 the `generation layout:` **log line** announced *"then the flat compatibility copy, then
  `_current`"* — the wrong order, in the one diagnostic a reader has in front of them during an
  incident. That is the more serious of the two and the review is right to call it out.

The third occurrence — *"the pointer is moved LAST"* in the long note beside the postcondition — is
about the **generation** and is still exactly right, so it is disambiguated rather than changed.

**And the order is now observed rather than read off the code.** Nothing in the finished state records
it, and a comment is precisely what drifted, so the harness hooks the flat copy's first upload and
asserts `_current` is already there, plus that the announcement names them in that order. 70
assertions now; 18 fail against the pre-change script.

PR #3753, opened 2026-09-08
Bound the teardown intake exemption to the phase that can use it (#3647)

### LINKED ISSUE #3647: Transient sync hub teardown discards a turn accepted into the shutdown window
## What is failing

The shutdown-drain contract — accepted work is processed before a hub goes down ([Teardown Layers](/Doc/Architecture/TeardownLayers)) — did not hold for a `sync/<id>` worker hub: it reached the terminal `ShutDown` phase with one accepted turn still queued, and `messageService.Dispose()` (which stops the pump) discarded that turn unprocessed. The turn was not a deferred delivery (none had answered `ShuttingDown`), and the last turn the pump executed was the `ShutdownRequest` itself — i.e. the turn was accepted behind the teardown's terminal phase transition. `sync/<id>` are the transient per-operation GitSync worker hubs (the family whose `[QUIESCE-OK]` lines show a clean "drained 0 callbacks" dispose), so the whole Quiescing → DisposeHostedHubs → ShutDown chain can complete in milliseconds and the race window is correspondingly narrow.

## Probable cause

A producer posted into the hub's queue inside the shutdown window that the [Hub Disposal Model](/Doc/Architecture/HubDisposalModel) leaves open by design — message intake stays open until `DisposeHostedHubs`, and the post evidently passed that gate. **Which producer is not identifiable from the evidence**: the log carries the discarded count but not the discarded message's type or id, so beyond "a late post beat the pump by milliseconds" the cause is low confidence. Two concrete candidates to check first:

- **a late completion/progress/cancel notification racing the owner's dispose** — these hubs are disposed the moment their operation finishes, so the operation's own final notification and its owner's cleanup dispose can arrive in the same instant;
- **a duplicate shutdown request** — two owners of the hub's lifetime (activity-finished cleanup and the sync completion path) both disposing it, the second request landing behind the first teardown's phases and being indistinguishable from real work in the current log shape.

Whichever it is, the immediate fix candidate is to make `[DISPOSE-DISCARD]` log the discarded turn's message type/id — the 2026-09-04 shutdown change ([Shutdown lets work finish, and names what it had to stop](/Doc/WhatsNew/2026-09-04-shutdown-lets-work-finish-and-names-what-it-had-to-stop)) intended these errors to carry "the hub, the message or activity", and one recurrence with that field answers the question outright.

## Impact

A one-off on a single transient worker hub — not an outage. The discarded turn never ran, so there is no half-applied state; the realistic worst case is one unprocessed turn's worth of work (at most a late progress/completion update) for one sync operation. The value here is diagnostic: this log site went live with the 2026-09-04 teardown change and this is its first production occurrence, so early recurrences are the cheapest way to pin the producer down.

## Where to look

- The `[DISPOSE-DISCARD]` site and `MessageService.Dispose` in `src/MeshWeaver.Messaging.Hub` — canonical shutdown chain: `MessageHub.Dispose` / `HandleShutdownCore` / `OnQuiesceComplete` / `DisposeImpl`.
- [Hub Disposal Model](/Doc/Architecture/HubDisposalModel) — the phase machine, where the pump stops, and "The creation window" (intake open until `DisposeHostedHubs`) that a producer can post into.
- Producer side: the lifetime of a `sync/<id>` hub — who disposes it on completion and what can post into it concurrently (activity completion on the control plane, sync finish/cancel paths). The build-triggered GitSync import in the current deployment wave (`UpdateToProvenCommitFromGitHub`, [The Sync-Ref Contract](/Doc/Architecture/SyncRefContract)) changed when imports fire and therefore hub churn; the timing is suggestive, but the evidence does not tie this discarded turn to it.

## Already known?

No existing incident or issue covers this log site. The nearest neighbours — the August `MessageHub`/`MessageHubGrain` shutdown incidents (e.g. `Admin/_LogIncident/1b713d0145006a84`, `Admin/_LogIncident/d79c8f1a68873ff6`) — belong to the retired out-of-band watchdog teardown, not this drain contract: related family, not duplicates.

---

### Evidence

| | |
|---|---|
| Fingerprint | `abf46b345ef20a96` |
| Category | `MeshWeaver.Messaging.MessageService` |
| Severity | Error |
| Namespace | `memex` |
| Pods | `memex-portal-deployment-84d966d9bd-nlv6z` |
| Occurrences | 1 |
| First seen | 2026-09-07 21:39:38Z |
| Last seen | 2026-09-07 21:39:38Z |

```
2026-09-07 21:39:38Z memex-portal-deployment-84d966d9bd-nlv6z fail: MeshWeaver.Messaging.MessageService[7302]
      [DISPOSE-DISCARD] Hub sync/KptJzzVNdk2qCCf4nMvcSA is disposing with 1 turn(s) still queued and unprocessed (the pump stops with this call). RunLevel=ShutDown; 0 deferred delivery(ies) already answered ShuttingDown; last turn executing: ShutdownRequest. Accepted work must be drained before a hub goes down — find what posted into the shutdown window.
```

</details>

<sub>Opened automatically from `Admin/_LogIncident/abf46b345ef20a96`. Recurrences are folded into this issue rather than opening new ones.</sub>

rbuergi:
Fixes #3647.

## The two sites

**Admits the turn** — `MessageService.RefusesIntake` (`src/MeshWeaver.Messaging.Hub/MessageService.cs`):

```csharp
// Teardown's OWN traffic gets in at every level — it is what advances the phases at all.
if (delivery.Message is ShutdownRequest or DisposeRequest)
    return false;
```

The comment carries the reason; the rule does not. Past `Quiescing` a `DisposeRequest` can advance
nothing — `MessageHub.Dispose()` is the only thing that moves `RunLevel` off `Started` and it is
idempotent, `IsShuttingDown` is already set so there is no recycle announcement, and `HandleDispose`
returns on its first line. Admitting it buys one turn slot and nothing else.

**Drops it** — `MessageService.Dispose()`, the queued-turn half of `[DISPOSE-DISCARD]` (event 7302):
it read `mainQueue.Count` and filed every entry at **Error** as *"still queued and unprocessed (the
pump stops with this call)"*.

## The denominator, and what the caller sees

**Provable, not estimated.** At the moment `messageService.Dispose()` reads the queue, the ShutDown
`ShutdownRequest` is the executing turn, so everything accepted before it has already run (FIFO) and
the queue can only hold what the gate admitted at `RunLevel >= DisposeHostedHubs` — where tier 2
admits *only* `ShutdownRequest` and `DisposeRequest`. `ShutdownRequest` is `internal` and absent from
the default `TypeRegistry`, and its three in-process posters each fire once per hub, so it is
**always a `DisposeRequest`** — modulo a delivery preempted between the gate read and `EnqueueTurn`,
a few instructions wide.

**The caller is not left waiting, and that is the finding.** `Dispose()` runs *inside* the
ShutdownRequest turn, so `DrainLoop` is one frame below on the same stack and takes the next turn the
instant that turn returns — the pump does not stop with that call and cannot. Measured on the unfixed
tree with a fixture that reproduces the production line byte for byte:

```
[Error]   [DISPOSE-DISCARD] Hub victim/late-post is disposing with 1 turn(s) still queued and
          unprocessed (the pump stops with this call). RunLevel=ShutDown; 0 deferred delivery(ies)
          already answered ShuttingDown; last turn executing: ShutdownRequest.
[Warning] Hub victim/late-post is disposing. Not processing DisposeRequest (id=r4FlWq…)   ← +3 ms
```

The second line is the pump dequeuing the very delivery the first called unprocessed, and the
disposing seam in `RunHandler` answering it — a transient `ShuttingDown` NACK for anything a sender
awaits, a silent drop for `[CanBeIgnored]` traffic nobody awaits. Nothing was left waiting; the drain
contract held. **The Error was the defect** — an incident filed about a hub that did what
[Teardown layers](src/MeshWeaver.Documentation/Data/Architecture/TeardownLayers.md) says it must.

Observable in production (once, 2026-09-07) and now deterministically in a test. It is *not* a
Blazor/sync-only shape: any hub that receives a recycle while it is tearing down reaches it.

## The fix — at the acceptance and at the ordering

1. **The exemption is bounded by the phase that can use it.** `ShutdownRequest` while a phase is left
   to advance (refused from `ShutDown` on); `DisposeRequest` only before the hub's own teardown has
   begun (refused from `Quiescing` on). The production occurrence disappears at its cause.
2. **The verdict measures instead of narrating.** 7302 stays an Error only where "queued and
   unprocessed" can be true — `drainsInFlight == 0`, nothing is running the loop, nobody will take
   those turns. Otherwise it is a Debug `[DISPOSE-DRAIN]` line. This is a deliberate level change
   with a stated trade-off: the line was shipping to Loki and filing incidents for a state that is
   normal, benign and self-resolving; the state where it is *not* benign keeps the Error, the event
   id and the incident.
3. **It names the message.** The queues carried a bare `Func<IObservable<IMessageDelivery>>` — a
   closure, opaque from outside — so the report could print a count and nothing else, which is
   exactly why #3647 could get no further than "a late post beat the pump by milliseconds". They now
   carry `QueuedTurn` (a struct over the already-allocated delivery plus the same closure: nothing
   extra is allocated per message), and both forms print type, id and sender. Rendered only when a
   line will actually be written — a method argument is evaluated whatever the level, and this file
   already carries the scar of that on this path (#3044/#3056).

No band-aid: no bound widened, no retry, no sleep, no watchdog, no `catch {}`.

## A second thing the bound secures

`TeardownVerdictsAreCausal` tells every caller and every test to wait for `RunLevel == Dead` and then
read state. Two of the three phase idempotency guards in `HandleShutdownCore` test
`RunLevel == <their own phase>` rather than `>=`, so a `ShutdownRequest` handled after
`RunLevel = Dead` would *not* hit its guard: it would re-enter the phase and assign the run level
back down. Nothing can produce one today (three posters, one shot each, `internal`, not on the wire)
— the gate makes that a property rather than an accident of who happens to post.

## Evidence

| | |
|---|---|
| Unfixed | `ShutdownWindowAdmissionTest` — subject **fails** on the `[DISPOSE-DISCARD]` assertion, control arm passes |
| Fixed | both arms pass |
| `MeshWeaver.Messaging.Hub.Test` | 374/374 |
| `MeshWeaver.Data.Test` | 496/496 |
| `MeshWeaver.Graph.Test` | 1452/1452 |
| `MeshWeaver.Documentation.Test` | 358/358 |
| `-c Release -warnaserror` | `MeshWeaver.Messaging.Hub`, `MeshWeaver.Documentation`, `MeshWeaver.Messaging.Hub.Test` — 0 Warning(s), 0 Error(s) each |

The test asserts two absences, so it is paired with two positive controls: the registrant records the
`RunLevel` it actually posted from (`ShutDown` — the post really was made in the window), and a
control arm sends the same message one phase earlier and still recycles the hub.

## Conserved

- [TeardownLayers](src/MeshWeaver.Documentation/Data/Architecture/TeardownLayers.md) → "What a hub
  still discards" — corrected; it asserted the claim this PR falsifies.
- [TeardownVerdictsAreCausal](src/MeshWeaver.Documentation/Data/Architecture/TeardownVerdictsAreCausal.md)
  → a new section on why `Dead` is safe to read.

Pairs-with: none — no public type or member is removed; every changed symbol (`QueuedTurn`,
`EnqueueTurn`, `mainQueue`/`deferredQueue`, `RefusesIntake`) is private to `MessageService`.

---

### Merged with main (#3739 touches the same struct)

`fix/plugins-1394-gate-open-straddle-fifo` landed a `QueuedTurn(long Seq, Func<…> Run)` on the same
two queues while this was in flight. Resolved by keeping **both** sides — `QueuedTurn(long Seq,
IMessageDelivery Delivery, Func<…> Run)` — with every one of main's `Seq` producers
(`EnqueueTurn`, the deferral, both `TryRequeueBehindOlderTurns` inserts, the `OpenGate` restore)
preserved verbatim. `git diff origin/main` on the file is exactly the three changes above and
nothing else. Re-verified after the merge: `MeshWeaver.Messaging.Hub.Test` **383/383** (main's 9
new tests included), `-c Release -warnaserror` clean.


rbuergi:
## The shard-4 red, attributed — and I did not treat "adjacent to my diff" as "not mine"

`MeshWeaver.Graph.Test.LateNackReenqueueCorrelationTest.AReenqueuedAttemptInheritsTheCorrelationIdOfTheNackThatCausedIt` — `Test execution timed out after 90000 milliseconds` (job 102189254946).

🚨 **This diff touches `MessageService.cs` — the message pump — so "unrelated" was not available as an assumption.** The late-NACK re-enqueue path runs through the same file this PR changes, and the change bounds which messages are admitted during teardown. That is exactly the shape that *could* break a re-enqueue.

### What I measured rather than assumed

| | |
|---|---|
| the failing test, on this branch, locally | **passes in 1 s** (CI: 90 s timeout) |
| the same run after merging `main` (11 commits) | **2 passed** — it and `ShutdownWindowAdmissionTest` together |
| my three `MessageService.cs` changes after the merge | present — `RefusesIntake` bound, `drainsInFlight` verdict, the named delivery |

So the change does **not** deterministically break it: a 1-second local pass against a 90-second CI timeout is not a marginal difference.

### Why I think it is the known intermittent, stated as a belief and not a proof

MeshWeaver#3477 documents this family — *"after `LATE_NACK_REENQUEUE` nothing lands or answers within 45 s"*, measured at **1 in 330**, first seen on a platform-set change, `28/28` locally on both pins. Same shape: a timeout with nothing landing, in the late-NACK re-enqueue path. Different class (`LateNackReenqueueTest` in the Plugins-hosted suite versus `…CorrelationTest` here), same mechanism.

🚨 **What I have NOT established:** that this specific test passes on `main` under CI load. One local pass discriminates little, and I did not re-run CI to fish for a green — the fresh run this push produces is a consequence of a forward-merge that was **genuinely needed** (the branch was 11 commits behind, and #3739 had changed the same file), not a re-run dressed up as one.

**If it fails again on this head, that changes the attribution** and it should be treated as caused by this diff until shown otherwise — not re-run a third time.

Recording a sighting on #3477 so the rate there is measured against every occurrence rather than only the ones someone happened to file.

PR #3759, opened 2026-09-08
fix(#2543): a response continuation no longer runs on the responding hub's action block

rbuergi:
## What this changes

Node CRUD in a mesh was serialised through **one actor's queue**. Every create/delete/move/copy
without an opted-in ancestor executes on `portal/nodeops-<meshId>`, and a create parked in its
validator held that hub so completely that a second, unrelated create posted to the same hub sat
unprocessed at `Queue(buffer=1,…)` for its entire budget.

**The root cause is not in the node-CRUD path.** A response subject is signalled from *inside* the
turn that handled the response, and Rx runs a continuation on the thread that signalled it — so
every `hub.Observe(x).SelectMany(…)` chain in the framework ran its whole remainder **on that hub's
action block, inside that turn**, and the turn could not end until the chain did.

The create handler is detached and correct — `HANDLER_EXIT state=Processed` at **+1 ms** on the fate
trail. The pipeline then climbed *back onto* the block when its partition bootstrap's nested
response came home, and ran validators, save and change feed there. Delete, move and copy inherit
the same shape, which is why one line fixes all of them.

`MessageHub.ContinueOffBlockRestoringUserContext` now hops the continuation off the block before
restoring the caller's identity. **The order matters:** `.Do(SetContext)` must run on the thread
that will run the continuation, or the chain resumes unauthenticated.

## This PR started as the opposite claim — that half is corrected, not deleted

It originally argued that the create pipeline never holds the block, reading `Executing(T, N ms)` as
time-on-the-block. That field is cleared in the handler **observable's** `.Finally`, so it measures
the handler's in-flight *lifetime*, including hops the chain already left the block for. With a
create parked it read `Executing(CreateNodeResponse, 36001ms)` and tracked the observer's own
sampling window exactly.

So **#2543's `Executing(CreateNodeRequest, 24888ms)` never established that the block was held for
24.9 s** — the `buffer=45` beside it is the part that was real. `BakeSealNodeOpsSaturation.md` keeps
the original analysis and adds a correction section saying which two inferences failed and how they
were falsified; the `drainsInFlight` table there inherits the same error and is corrected with it.

## Evidence

Measured behaviourally, because the snapshot fields cannot decide it:

| test | hop disabled | hop enabled |
|---|---|---|
| probe: an independent create completes while one is parked | **FAIL** | PASS |
| `TwoCreates_AreInFlightSimultaneously` (two validators entered before either is released) | **FAIL** | PASS |
| **positive control**: the same probe with nothing parked | PASS | PASS |

The control passing in both worlds is what makes the other two mean "the pump is held" rather than
"this probe never completes anyway". Both now fail deterministically in isolation (11 s / 21 s)
rather than only under assembly load, which is how the original red hid: isolated it passed 10/10.

## Verification

- **4,426 tests, 8 projects, 0 failures** — Graph 1454, Compiler.Pipeline 735, Hosting 514,
  Data 496, Layout 455, Messaging.Hub 383, Documentation 361, ContentCollections 28
- Full solution `dotnet build -c Release -p:CIRun=true -warnaserror` — **0 Error(s), 0 Warning(s)**

## Blast radius, stated plainly

This changes continuation threading for **every** request/response pair in the framework, not just
node CRUD. Anything that implicitly relied on continuations being serialised per-hub can now run
concurrently. That is the point of the change, and it is why the verification above is broad rather
than scoped to the touched path.

Related: #2543.

### DIFF
--- a/src/MeshWeaver.Documentation/Data/Architecture/BakeSealNodeOpsSaturation.md
+++ b/src/MeshWeaver.Documentation/Data/Architecture/BakeSealNodeOpsSaturation.md
@@ -391,6 +391,55 @@ hop. The first draft of that test asserted the opposite and failed. **So a creat
 block for 24.9 s is doing something this path does not normally do**, and "the create pipeline is
 expensive" is not the answer.
 
+## 🚨 CORRECTION (2026-09-08): half of the paragraph above is wrong, and the fix is one line
+
+Everything above about the *request turn* holds. Two inferences drawn from it do not, and both were
+falsified by measurement rather than by re-reading the code.
+
+**Correction 1 — `Executing(T, N ms)` does NOT measure time on the block.** The mechanism is stated
+correctly earlier in this page (`currentlyExecutingMessageType` is cleared in the handler
+observable's `.Finally`) and then read the wrong way round. `.Finally` fires when the handler's
+**observable completes** — i.e. when the whole detached chain finishes — not when the pump is
+released. So the field measures the handler's in-flight LIFETIME, including every pooled hop the
+chain has already left the block for. Demonstrated: with a create parked in its validator, that
+counter read `Executing(CreateNodeResponse, 36001ms)` and tracked the observer's sampling window
+exactly. **Therefore #2543's `Executing(CreateNodeRequest, 24888ms)` never established that the
+block was held for 24.9 s** — the `buffer=45` beside it is the part that was real. The
+`drainsInFlight` table above inherits this error and reads `> 0` as "a thread is inside the
+handler"; it is "a chain is still in flight", which is not the same claim.
+
+**Correction 2 — the create pipeline DID hold the block, via its continuation.** The request turn
+is genuinely free (`HANDLER_EXIT state=Processed` at +1 ms, confirmed on the fate trail). But the
+detached chain does a partition bootstrap, whose nested response comes back to the SAME hub as a NEW
+turn — and the rest of the pipeline, validators included, ran inline inside that response's turn.
+The earlier reading of an idle hub was taken at a single instant before that turn began, and in
+isolation the path differs; under load the test caught it at 0–1 ms and the assertion read it as
+noise.
+
+**How it was settled — behaviourally, because the snapshot fields cannot decide it.** With a create
+parked, post an INDEPENDENT node create to the same hub:
+
+| | result |
+|---|---|
+| probe alone, nothing parked (**positive control**) | completes promptly |
+| same probe, while a create is parked | never runs — `Queue(buffer=1,…)` for the whole budget |
+
+The control is what makes the second row mean "the pump is held" rather than "this probe never
+completes anyway".
+
+**The fix, and it is not in the node-CRUD path at all.** A response subject is signalled from inside
+the turn that handled the response, and Rx runs a continuation on the thread that signalled it — so
+*every* `hub.Observe(x).SelectMany(…)` chain in the framework ran its remainder on that hub's action
+block, inside that turn. `MessageHub.ContinueOffBlockRestoringUserContext` now hops the continuation
+off the block (`ObserveOn`) before restoring the caller's identity — the hop must precede the
+identity restore, or the chain resumes unauthenticated. Create, delete, move and copy are fixed by
+the same line because they all await nested node ops the same way.
+
+`NodeCrudDoesNotOccupyTheExecutionBlockTest` carries the three tests: the probe, its positive
+control, and `TwoCreates_AreInFlightSimultaneously`, which parks two creates at once and requires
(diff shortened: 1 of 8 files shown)


rbuergi:
## CI after the terminating guard: one failure left, and it names a SIXTH invariant

Run [34316509035](https://github.com/Systemorph/MeshWeaver/actions/runs/34316509035) on `8738db0f7`:

- ✅ the host crash is gone (shard 5 clean, no `MASKED`, no `[FATAL ERROR]`)
- ✅ `ReadDuringDisposalWindowTest` — the NACK-path failure did not recur
- ❌ `LogonActionFrameworkTest.Two_concurrent_logons_run_a_once_action_exactly_once` — still a clean
  30 s `TimeoutException` (05:56:21.740 → 05:56:51.750). Nothing emitted, nothing errored.

**So the terminating guard was necessary but is not the cause of this one.**

### What this test is actually standing on

`LogonActionRunner.RunFor` carries a comment that describes, in advance, precisely the thing this PR
changes:

> 🚨 The identity is scoped with `RunAs`, NEVER `Observable.Using(access.ImpersonateAsSystem, …)`.
> Impersonation is an **AsyncLocal store/restore pair**; `Observable.Using` opens it on the
> SUBSCRIBING thread and disposes it when the inner observable TERMINATES — for a cross-hub write,
> **the owning hub's response thread** — so the two halves land on different threads and the
> subscriber is left latched as the impersonated identity (#1790). **`RunAs` opens and closes inside
> one synchronous `Subscribe`.** Each individual write re-establishes it too, because `Concat`
> subscribes a later action on whichever thread the previous one completed on.

That is a design that reasons explicitly about *which thread a response continuation completes on*,
because the identity is `AsyncLocal`. This PR moves every response continuation onto a pool thread.
`ContinueOffBlockRestoringUserContext` re-seeds `accessService.SetContext(capturedCtx)` after the hop
— which is why the hop is placed before the restore — but `AccessService.RunAs` is a **different**
mechanism from the hub's captured context, and re-seeding the captured identity is not obviously the
same as preserving a `RunAs` scope that was opened synchronously around a `Subscribe`.

### Running tally of what the action block was underwriting

1. serialisation of continuations (removed deliberately — this is the parallelism)
2. a fault boundary → replaced by `GuardContinuationFaults`
3. ordering with DI-scope disposal → logged as transient during teardown
4. fail-fast rather than hang → the guard must TERMINATE, not just report (previous commit)
5. which NACK path answers a teardown race (did not recur; unexplained)
6. **thread affinity for `AsyncLocal` identity scopes** ← this one

### Position

This is the fourth CI run to surface an invariant nobody had written down, and #6 is the first that
is about *correctness of identity* rather than liveness. I am not going to keep discovering these
one run at a time on a framework-wide change.

Staying DRAFT. The scope question I put to the maintainer stands and this sharpens it: either
enumerate the invariants up front and stage the general decoupling deliberately, or narrow the hop to
the node-CRUD path where the parallelism win is concrete (proven here by
`TwoCreates_AreInFlightSimultaneously`) and the surface is small.

The two findings that stand regardless of that decision, and are already merged or committed:
`Executing(T, N ms)` is a handler lifetime rather than block occupancy (in `MessageService`'s field
comment and `Doc/Architecture/DebuggingMessageFlow`), and `buffer=N` is the occupancy signal.

rbuergi:
## Superseded by #3802 — and this branch must not be merged as-is

Two reasons, one of them mechanical and load-bearing.

**1. This branch is 14,689 deletions behind `main`.** `git diff origin/main --stat` on it: *175 files changed, 2,340 insertions, 14,689 deletions.* Five PRs merged today (#3795, #3797, #3798, #3799, #3800) and this branch predates all of them, so a PR from it would **revert** them — including files it shows as deletions that are its own author's work from this morning. Merging `main` in would mean resolving conflicts in `MessageHub.cs` and `MessageService.cs`, which are exactly the files #3798 changed.

**2. The scope call this PR surfaced is resolved, the conservative way.** The choice was *"fix the general decoupling forward (error boundary + disposal-aware + bounded pool) or narrow the hop to the node-CRUD path"*. #3802 takes the second: `IDetachedResponseContinuation` is declared by the six node-CRUD requests and by nothing else, so every other request keeps the semantics it has always had.

Everything this branch established is carried over — the seven host crashes and the **six invariants** they named, `PooledContinuationScheduler` withholding `ISchedulerLongRunning`, the guard that REPORTS AND TERMINATES (a first version only reported, which turned a crash into a hang), and the behavioural test with its positive control. What is not carried over is the global hop, which is what broke `LogonActionFrameworkTest` and reddened `ReadDuringDisposalWindowTest`; both are green in #3802 and re-run by name.

The findings are now durable rather than living in this thread: `Doc/Architecture/DetachedResponseContinuations`.

Closing this draft in favour of #3802.

Issue #3353, opened 2026-09-05
Serve container images from the mesh: a read-through OCI mirror in front of ACR


rbuergi:
## Proposal

Serve OCI images from the mesh, as a **read-through mirror** in front of ACR — not a replacement.

Design committed as [`Doc/Architecture/ContainerRegistryInMemex`](https://github.com/Systemorph/MeshWeaver/blob/main/src/MeshWeaver.Documentation/Data/Architecture/ContainerRegistryInMemex.md) (PR #3352). This issue tracks whether we do it.

## Why

An image's contents are **opaque until something fails to compile against them.** That is not a
hypothetical: it is #3328, which cost a fleet-wide module-bundle break, a ~2 h core CD outage, and a
staleness red on every open Plugins PR — because `/app` is the reference set every satellite compiles
against and nothing could say what was in it.

Four concrete gains, each traceable to something that actually went wrong:

| gain | the incident it answers |
|---|---|
| the image's closure becomes queryable mesh data | #3328 — diagnosed with `docker run … ls /app`; #3334's gate re-implements this as a shell script because there is nowhere else to put it |
| provenance as typed references, not a tag convention | `<core-short>-p<plugins-short>` string-splitting |
| a pin becomes ONE reference | `ci.yml` carries **six** digest literals; half-moving them is a documented failure mode, and I nearly shipped exactly that today |
| retention that refuses to evict a digest a live deployment names | an ACR purge once deleted a pinned production tag |

## The constraint

**A registry that runs inside the thing it deploys cannot serve the image that boots it.** Kubernetes
pulls the portal image before any MeshWeaver process exists, and putting the registry in the portal
merges two failure domains that are independent today.

**⇒ The boot image stays on ACR, permanently.** Any proposal that does not accept this is wrong.

## Scope of the first increment

- pull side only: `GET /v2/`, manifest `GET`, blob `GET` + range, `tags/list`, token handshake
- cache-through to ACR on miss; record closure + provenance as mesh nodes
- pushes unchanged (ACR), boot image unchanged (ACR)
- reversible: switching the mirror off is a config change, not a migration

**Explicitly out of scope for v1:** push, GC, being anything's only copy.

## Acceptance — measurable, not "images pull"

1. the closure of a promoted image can be answered from mesh data with no `docker run`;
2. #3334's `check-platform-reference-set.sh` can be re-expressed as an assertion over that data;
3. a pin bump moves one reference instead of six literals;
[...]

PR #3760, opened 2026-09-08
CD publishes every image and every sealed bundle publication to cr.meshweaver.cloud (#3353)

rbuergi:
Closes the last step of the fleet-registry program (#3353): **every CD publication now reaches `cr.meshweaver.cloud`** — the three images and every sealed bundle publication — beside ACR, which stays (the host of the registry pods, `memex.meshweaver.cloud`, keeps pulling from ACR by design).

Pairs-with: none — CI only

## Images (`main-cd.yml`)

- Each `Build + push …` leg (`memex-portal-ai`, `memex-migration`, `mw-plugin-test`) logs in to cr as the publisher and, **in the same job right after the ACR push**, copies its staging tag there with the new `.github/scripts/mirror-image-to-registry.sh`: `docker buildx imagetools create` with a single source (the copy-as-is operation `promote` has always used for GHCR), then the digest is **read back off cr and compared to ACR's — a mismatch or an unreadable digest fails the job RED**.
- `promote` writes every consumer-visible tag on cr in the same phase it writes it on ACR (identity tags, `main`/`latest`, the arming `memex-portal-ai:<version>` inside phase C right after the ACR PUT, the line pointers), every one through the same script and the same readback.
- `preflight` asserts `secrets.MW_REGISTRY_PUBLISHER_PASSWORD` RED, naming both secret stores. `registry-login.sh` retries transport faults bounded and refuses a rejected password immediately.

## Bundles

- `push-bundle-publication.sh` (from #3729, called by nothing until now) is wired into **all three** seal lanes, after the share seal succeeds and never instead of it: `main-cd.yml`'s `publish-bake` (`meshweaver-content`, per architecture, with `--release <version>` so `plugins/releases:<version>` is recorded like the share's `_releases` marker), `main-cd.yml`'s `plugins-bake` (through the reusable lane), and `node-repo-publish-bake.yml` (every satellite). Index `plugins/<source>:<identity>` + the immutable `<identity>-<run id>` tag, tag move last, exactly as the design page states.
- `publish-bake-bundles.sh` gains `BAKE_PUBLICATION_DIR`: after every share target sealed (or already held the content) it materialises the exact publication from its own verified upload plan — byte-checked against the digests the shares were verified against, `_complete` last — so the registry and the shares hold ONE set.
- `push-bundle-publication.sh` gains `--skip-if-published`: the registry-side sealed-skip, keyed on content × identity like the share's (tag resolves + sidecar config's `sourceCommit` equals ours ⇒ skip; anything less pushes, because a push is idempotent). Without it every core merge under an unchanged identity would push a full set of new blobs to a registry with no GC.
- The index digest is **printed** (job log + step summary, `bundle_index_digest` lane output, RED if missing after a publish); `register-publication`'s record has no field for it — the TODO is on the doc page's Publishing section, not in code.
- `install-oras.sh` is the ONE pinned ORAS install (1.3.4, sha256 per platform); the chart-gate harness now installs through it instead of its own pin table.

## Preflight and gates — all run locally the way `chart-gate.yml` / `shared-rules.yml` run them

- `check-pr-secret-preflight.py --root .` → 0 violations (`--self-test` green). No allow entry needed: `main-cd.yml` never runs on a PR and the lane is `workflow_call`; satellite callers gate `publish-bake` to push/dispatch/schedule.
- `check-workflow-timeouts.py --root .` → 81 jobs, 0 violations.
- `check-workflow-shell.py` → 0 live findings (new scripts shellcheck-clean at `-S warning`).
- `test-bundle-registry.py` against the real distribution + docker_auth images → **92 assertions, 0 failures**, including new cases for `--skip-if-published` (skip on equal content, republish on a different commit with the previous generation still resident, refusal without a source commit) and the install script.
- `MeshWeaver.Documentation.Test` guards (`PlatformBakeLaneGuard`, `UpstreamBuildGateGuard`, `InboxSignatureVerdictGuard`, `NodeRepoLaneHostGuard`, `CdImage*`, …) + `WhatsNewEntryIntegrityTest` + `DocumentationLinkIntegrityTest`: 292 passed, 0 failed (Release, `-warnaserror`, `.trx` written).

## Maintainer: provisioning (the value is in Key Vault `Systemorph` → `memexcloud-Registry-PublisherPassword`)

Core, both stores:

```bash
az keyvault secret show --vault-name Systemorph --name memexcloud-Registry-PublisherPassword --query value -o tsv | gh secret set MW_REGISTRY_PUBLISHER_PASSWORD --repo Systemorph/MeshWeaver
az keyvault secret show --vault-name Systemorph --name memexcloud-Registry-PublisherPassword --query value -o tsv | gh secret set MW_REGISTRY_PUBLISHER_PASSWORD --app dependabot --repo Systemorph/MeshWeaver
```

Every satellite that calls `node-repo-publish-bake.yml` needs the same pair AND, when it moves its `uses:` pin past this commit, must pass `bundle-registry-publisher-password: ${{ secrets.MW_REGISTRY_PUBLISHER_PASSWORD }}` on the `publish-bake` job (the lane's preflight fails RED naming it otherwise — never skips):

```bash
for repo in MeshWeaver.Plugins MeshWeaver.Education MeshWeaver.Reinsurance MeshWeaver.SocialMedia MeshWeaver.Manufacturing MeshWeaver.Crm; do
  az keyvault secret show --vault-name Systemorph --name memexcloud-Registry-PublisherPassword --query value -o tsv | gh secret set MW_REGISTRY_PUBLISHER_PASSWORD --repo Systemorph/$repo
  az keyvault secret show --vault-name Systemorph --name memexcloud-Registry-PublisherPassword --query value -o tsv | gh secret set MW_REGISTRY_PUBLISHER_PASSWORD --app dependabot --repo Systemorph/$repo
done
```

The chart's `registry.publisher.passwordBcrypt` must be the bcrypt hash of that same value (`registry-login.sh` names both on a refusal).

## Not verifiable without the credential

- A real push to `cr.meshweaver.cloud` (login, `imagetools create` cross-registry copy, ORAS push) — the first main-cd run after the secret is provisioned is the measurement; until it is, `preflight` reds naming the secret and nothing downstream runs.
- Whether `docker buildx imagetools create` preserves the index digest on the ACR→cr copy in production: it is the fast path buildx takes for a single unmodified source (and what the ACR→GHCR mirror relies on), and the digest readback is precisely the gate that turns a wrong assumption into a RED job rather than a silent divergence.

## Docs

`ContainerRegistryInMemex.md` (status: CD publishes to cr, with the verification and what still reads ACR), `PluginBundlesInTheRegistry.md` (Publishing: wired, the three callers, the skip rule, the per-architecture release tag, the record TODO, who reads which copy), What's New `Category: Feature`.


rbuergi:
**Conflict resolution 2026-09-12 (merge of `origin/main`, 595 commits):** one hunk in `.github/workflows/node-repo-publish-bake.yml`, where core #4084's `RESP="${RESP:-/tmp/resp}"` default and this PR's fleet-registry `INDEX_DIGEST` assertion landed on the same lines. Kept both — the `RESP` default LEADS its `run:` block (each `run:` is a fresh shell, so the second block keeps its own default too), the assertion follows, and the first `curl -o "$RESP"` comes after. Verified by the session that owns #4084. Rule for the next resolver of an older branch touching this file: the `RESP` default leads its block; everything else follows.

rbuergi:
**Red diagnosed and fixed — `ad109155f`.**

**Failing assertion.** `register-publication` → *Sign and POST the publication record* (`node-repo-publish-bake.yml`, added by this PR) asserted `[ -n "${INDEX_DIGEST:-}" ]` unconditionally, *before* the `curl` that writes `$RESP`/`body`. `PluginPublicationProvenanceTest` executes that step's real shell supplying neither `BUNDLE_REGISTRY` nor `INDEX_DIGEST`, so the step exited 1 before `body` existed and every case in the class failed with `FileNotFoundException: …/body`.

**Fix.** The assertion stays where it is (its point is to refuse before recording a publication the registry cannot be asked about) and is now keyed on `BUNDLE_REGISTRY`: *if we pushed to the fleet registry we must have its index digest; a repo not publishing there has none to demand.* Verified by content that this is the variable that decides the push — the `registry` step runs on `steps.scope.outputs.scope != 'none'` (the same condition as the share publish whose `published` output is this job's `if:`), targets the same `inputs.bundle-registry`, and publish-bake's unconditional preflight refuses an empty one. So on the lane a push happened iff this job runs with the value set: no path pushes with the assertion disabled. The finding is written as the comment next to the guard.

**Tests** (`dotnet test test/MeshWeaver.Documentation.Test -c Release -warnaserror`):
- new negative control `AFleetRegistryPushWithoutItsIndexDigestIsRefusedBeforeTheRecordIsWritten` — registry set, digest empty: exit ≠ 0, no `body`, no HTTP;
- new positive companion `AFleetRegistryPushWithItsIndexDigestIsRecordedAndNamed` — both set: exit 0, index named in log and step summary;
- `~PluginPublicationProvenance`: 34/34 cases; whole project: 436/436. actionlint on both workflows: no new findings (8 pre-existing style/info hits, identical on `main`).

Also merged `origin/main` (conflict in `main-cd.yml`'s preflight: kept both `MW_REGISTRY_PUBLISHER_PASSWORD` and the `FLEET_READER_APP_*` assertions). Draft state untouched; auto-merge remains armed by `meshweaver-cloud[bot]` (merge).

🤖 Filed from Claude Code.

rbuergi:
Rebuilt on `main@69e587c5` (was based on `043d2cb6`; the queue ejected this PR as UNMERGEABLE).

**One conflict hunk**, in the `publish-bake` job's `outputs:` map — both sides appended an output at
the same line. Resolved by **keeping both**, not by taking a side:

- `bundle_index_digest` (this PR) — `steps.registry.outputs.index_digest`, read by `register-publication`;
- `platform-set` (main, `39caeed99`) — `steps.digests.outputs.platform-set`, the promoted set when the
  lane resolves the platform itself.

Different steps, independent readers; dropping either breaks the job that reads it.

Both content rules established earlier on this file survive, verified by reading the merged text: in
`register-publication`'s shell `RESP="${RESP:-/tmp/resp}"` still **leads** the block, the fleet-registry
`INDEX_DIGEST` assertion follows it, still wrapped `if [ -n "${BUNDLE_REGISTRY:-}" ]`, and the comment
recording why `BUNDLE_REGISTRY` is the deciding variable is intact. main's self-resolve work (the
`digests` step, the header, and `image`/`platform` now reading `steps.digests.outputs.*`) is taken
unmodified on its own terms.

**#4108** is still open and touches `node-repo-module-pack.yml`, **not** this file — it cannot
re-conflict this lane when it lands.

Verified locally: `actionlint` clean over all workflows; `check-workflow-{shell,timeouts,yaml-keys,permission-pairing}`
clean; `dotnet test test/MeshWeaver.Documentation.Test -c Release -warnaserror` → **441 passed, 0 failed**,
including `AFleetRegistryPushWithoutItsIndexDigestIsRefusedBeforeTheRecordIsWritten`.

PR #3776, opened 2026-09-09
fix(operator): run.sh signs in to Azure through Workload Identity before the first step

rbuergi:
## run.sh signs in to Azure through Workload Identity before the first step

**Measured 2026-09-09 01:33Z**, `Deployments/pearl-provision-20260909` — the first `Provision` ever run through the lane, on operator 0b79490 after #3757 / #3769 / #3770 and Memex #256 had cleared every earlier wall. Fourteen steps rendered; step 1/14 `az postgres flexible-server db create`:

```
ERROR: Please run 'az login' to setup account.
```

Nothing in the operator runs `az login`. The Job is federated (pod label `azure.workload.identity/use`, the ServiceAccount annotation, the identity from backups.bicep) and the webhook projects a token and sets `AZURE_CLIENT_ID` / `AZURE_TENANT_ID` / `AZURE_FEDERATED_TOKEN_FILE` — the Azure **SDKs** pick those up, the **CLI** never does. No run had reached an az step before: memex's Reconcile / Restart are kubectl + helm only, which authenticate in-cluster. Nothing was created on Azure (step 1 was the first az call).

### What changes (`run.sh`, before the first step)
- `AZURE_FEDERATED_TOKEN_FILE` set → require `AZURE_CLIENT_ID` and `AZURE_TENANT_ID` by name, `az login --service-principal --username … --tenant … --federated-token <file contents> --allow-no-subscriptions`, log `signed in as <id> (subscription <s>)`, report `::hosting:: az_login=true`.
- A failed login is a hard refusal before any step, naming the federated credential's subject and issuer (the one place a mismatch shows). An unreadable token file is named.
- No token variable (a kubectl+helm-only run) → said, `az_login=false`, the plan runs; an az step then fails by name with the hint above it in the log.
- The token is passed as an argument by this process and never printed — az is deliberately not wrapped in `hosting::do`, which narrates argv.

### Tests (`stubs/run/az` records argv)
signed in with the projected token · token absent from the run's output · `az_login=true` · sign-in before the first step marker · failed login = hard refusal, first step never ran · unreadable file named · missing tenant named · no-token run reports false. **146 passed, 0 failed.**

### Chain
#3775 (adopt before helm) + this → operator image → `Deployments/{memex,pearl}.operator.image` (API) → memex Reconcile (expect `adopted=1`, success) and pearl Provision re-filed.

### DIFF
--- a/deploy/aks/operator/bin/run.sh
+++ b/deploy/aks/operator/bin/run.sh
@@ -69,6 +69,38 @@ if [ -n "${HOSTING_VALUES:-}" ]; then
   fi
 fi
 
+# ── Azure sign-in through Workload Identity ──────────────────────────────────────────────────
+# The Job runs as the hosting-operator ServiceAccount, federated to the operator's managed identity
+# (backups.bicep); the workload-identity webhook projects a token into the pod and sets
+# AZURE_CLIENT_ID / AZURE_TENANT_ID / AZURE_FEDERATED_TOKEN_FILE. 🚨 az DOES NOT READ THOSE ON ITS
+# OWN — the Azure SDKs do (WorkloadIdentityCredential), the CLI does not. Measured 2026-09-09 01:33Z
+# on the first Provision ever run through the lane (Deployments/pearl-provision-20260909): step 1/14
+# `az postgres flexible-server db create` answered "ERROR: Please run 'az login' to setup account."
+# Every earlier run (memex Reconcile/Restart) was kubectl+helm only, which authenticate in-cluster,
+# so no run had reached an az step before. Sign in ONCE here, as the identity, and say so.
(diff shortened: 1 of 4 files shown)


rbuergi:
Token lifetime, for the record: the sign-in happens once at t=0 with the projected file's contents; the Job's `activeDeadlineSeconds` is 3600 (`HostingOperator.DeadlineKey` default), the same as the workload-identity webhook's default token expiration, and az keeps the exchanged access token for the session — so a run cannot outlive its assertion. If `Hosting:Operator:DeadlineSeconds` is ever raised past 3600, the sign-in must re-read the file before long steps; not built, because nothing can reach it today.

PR #3793, opened 2026-09-09
fix(#3764): an unconvertible bound value must not end the binding

### LINKED ISSUE #3764: LayoutClientExtensions.ConvertNumericValue throws for non-IConvertible bound values (raw JsonElement), breaking Label bindings
**What is failing**

The shared client-side binding conversion in `MeshWeaver.Layout.Client.LayoutClientExtensions` crashes whenever the value a control is bound to is not an `IConvertible`. The per-emission conversion chain (`DataBind` → `ConvertSingle` → `ConvertNullableOrNumericValue` → `ConvertNumericValue`) ends in `Convert.ChangeType`, which throws `InvalidCastException` ("Object must implement IConvertible") for such a value; the exception propagates through the Rx `Select` and kills the binding subscription. In this instance the Label view was binding a value named `gates` in the nested area `Overview/1/2/1/1/1/15/2` of a node's Overview page, and that label failed to load.

**Probable cause** — medium confidence on the mechanism; the data producer is unidentified

`System.Text.Json.JsonElement` does not implement `IConvertible`. A Label bound to a pointer whose value is structured JSON (an object or an array — `gates` reads like a list of gates) receives a raw `JsonElement`, and the conversion helper has no `JsonElement` branch before its `Convert.ChangeType` last resort. The fix belongs in `ConvertNumericValue`/`ConvertSingle`: convert `JsonElement` properly (deserialize to `T`, `GetString()`, or `GetRawText()`), and degrade to the default value instead of letting the exception tear down the subscription. Which portal surface publishes the `gates` value could not be identified from the mesh (no matching source found) — the assignee will need the pod's surrounding logs or the node behind that Overview path to pin it down. The "value arrives as raw `JsonElement`" failure class is already called out in the framework docs ([Data Binding](/Doc/GUI/DataBinding)) for the node-stream cache.

**Impact**

Isolated so far — a single label on one node's Overview page failed to render. Judged from the numbers this is a nuisance, not an outage, but the conversion path is shared by every bound control on every client, so any label or field bound to non-scalar data will crash the same way.

**Where to look**

- `src/MeshWeaver.Layout/Client/LayoutClientExtensions.cs` (Systemorph/MeshWeaver): `DataBind` (~line 129), `ConvertSingle` (~line 299), `ConvertNullableOrNumericValue` (~line 339), `ConvertNumericValue` (~line 354)
- Log site: the Label Blazor view (category `MeshWeaver.Blazor.Components.Label`), `src/MeshWeaver.Blazor/Components/Label.razor.cs`
- Binding architecture: [Data Binding in MeshWeaver Layout](/Doc/GUI/DataBinding)

Duplicate check: no other `LogIncident` nodes exist in the mesh and no issue covering this component + symptom was found — this looks like a fresh report.

---

### Evidence

| | |
|---|---|
| Fingerprint | `f9d90a1aa0623909` |
| Category | `MeshWeaver.Blazor.Components.Label` |
| Severity | Error |
| Exception | `System.InvalidCastException` |
| Top frame | `MeshWeaver.Layout.Client.LayoutClientExtensions.ConvertNumericValue[T](Object value)` |
| Namespace | `memex-cloud` |
| Pods | `memex-portal-deployment-bf77c847-jl7xh` |
| Occurrences | 1 |
| First seen | 2026-09-08 20:18:34Z |
| Last seen | 2026-09-08 20:18:34Z |

```
2026-09-08 20:18:34Z memex-portal-deployment-bf77c847-jl7xh fail: MeshWeaver.Blazor.Components.Label[0]
      Loading 'gates' in area 'Overview/1/2/1/1/1/15/2' in Area Overview/1/2/1/1/1/15/2
      System.InvalidCastException: Object must implement IConvertible.
         at System.Convert.ChangeType(Object value, Type conversionType)
         at MeshWeaver.Layout.Client.LayoutClientExtensions.ConvertNumericValue[T](Object value) in /home/runner/work/MeshWeaver/MeshWeaver/src/MeshWeaver.Layout/Client/LayoutClientExtensions.cs:line 354
         at MeshWeaver.Layout.Client.LayoutClientExtensions.ConvertNullableOrNumericValue[T](Object value, T defaultValue) in /home/runner/work/MeshWeaver/MeshWeaver/src/MeshWeaver.Layout/Client/LayoutClientExtensions.cs:line 339
         at MeshWeaver.Layout.Client.LayoutClientExtensions.ConvertSingle[T](IMessageHub hub, Object value, Func`3 conversion, T defaultValue) in /home/runner/work/MeshWeaver/MeshWeaver/src/MeshWeaver.Layout/Client/LayoutClientExtensions.cs:line 299
         at MeshWeaver.Layout.Client.LayoutClientExtensions.<>c__DisplayClass2_0`1.<DataBind>b__0(Object x) in /home/runner/work/MeshWeaver/MeshWeaver/src/MeshWeaver.Layout/Client/LayoutClientExtensions.cs:line 129
         at System.Reactive.Linq.ObservableImpl.Select`2.Selector._.OnNext(TSource value)
```

</details>

<sub>Opened automatically from `Admin/_LogIncident/f9d90a1aa0623909`. Recurrences are folded into this issue rather than opening new ones.</sub>

rbuergi:
Closes #3764.

## What was wrong

`ConvertNumericValue` and `ConvertNumericSafely` both finish at `Convert.ChangeType`, which throws
`InvalidCastException` — *"Object must implement IConvertible"* — for a value that is not
`IConvertible`. A `System.Text.Json.JsonElement` is exactly such a value.

🚨 **The damage is not one bad emission.** The conversion chain runs inside the binding's Rx
`Select`, so the exception propagates out of the *live sequence* and **terminates the
subscription**. The control keeps whatever it was showing and never updates again for the life of
the page, with nothing on screen saying why. The reported instance was a Label bound to `gates` that
simply never loaded.

## The fix, in two arms

The issue named `ConvertNumericValue`. That is only half — **a numeric target reaches a different
arm**, and my first attempt missed it: the test for `int` still threw, and said so.

| arm | reached when | now |
|---|---|---|
| `ConvertNumericValue` | non-numeric target | degrades to default |
| `ConvertNumericSafely` | **numeric** target | degrades to default |

`JsonElement` also gets a *real* conversion rather than only a guard, so the fix converts what it can
and falls back only where there is genuinely no reading:

- a JSON string → its text (not its quoting)
- a structured value in a scalar slot → its raw text, so the control shows what it was given
- `null` / `undefined` → default

## Why degrading is the correct disposition, not a swallow

A control rendering its default is a **bounded, visible** loss for one update. A dead binding is an
**unbounded, invisible** one — the value is never right again and nothing reports it. The exception
was not carrying information to anyone; it was ending a subscription.

## Verification

- The new tests assert **values**, not merely "did not throw", so a fix that degraded *everything* to
  default would still fail them.
- **Falsified:** removing the numeric guard turns the `int` case red again; restoring it, green.
- One expectation was **corrected against measurement rather than the code changed to match it** — a
  JSON array already had a reading through `ConvertJson` and renders as `1, 2`, so the test records
  that instead of the `[1,2]` I had assumed.
- `MeshWeaver.Layout.Test` **463 passed**, `MeshWeaver.Documentation.Test` **361 passed**, built
  `-c Release -warnaserror` clean.

Includes a `Category: Fix` What's New entry.

### DIFF
--- a/src/MeshWeaver.Documentation/Data/WhatsNew/2026-09-09-a-control-no-longer-stops-updating-when-one-value-cannot-be-converted.md
+++ b/src/MeshWeaver.Documentation/Data/WhatsNew/2026-09-09-a-control-no-longer-stops-updating-when-one-value-cannot-be-converted.md
@@ -1,0 +1,21 @@
+---
+Name: A control no longer stops updating when one value cannot be converted
+Category: Fix
+Description: When a value bound to a control could not be converted to the type the control expected, the error did not just skip that one update — it ended the binding, and the control silently stopped updating for as long as the page stayed open. Bindings now survive an unconvertible value.
+Icon: Link
+Order: -20260909
+---
+
+# A control no longer stops updating when one value cannot be converted
+
+A label, or any bound control, could stop updating entirely and show nothing — with nothing on the
+page to say why.
+
+The cause was the conversion that turns a bound value into the type a control expects. When it met a
+value it had no reading for, it raised an error. That error travelled out of the live binding rather
+than being confined to the single update that caused it, and ending a binding ends it for good: the
+control kept whatever it was showing and never changed again while the page was open.
+
+A value the conversion cannot read now leaves the control at its default for that update, and the
+binding keeps running. Structured values — a list or an object bound where text was expected — now
+render as their text instead of stopping the control.
--- a/src/MeshWeaver.Layout/Client/LayoutClientExtensions.cs
+++ b/src/MeshWeaver.Layout/Client/LayoutClientExtensions.cs
@@ -296,13 +296,6 @@ public static class LayoutClientExtensions
         // `fail` line on EVERY NavLink render (prod error storm) while the icon rendered as nothing.
         if (typeof(T) == typeof(Icon) && TryGetStringValue(value, out var iconString))
             return (T?)(object?)Icon.Parse(iconString);
-        // A label can receive the CLR collection directly from a local binding, or its JSON
-        // representation after transport. Both must use the same display conversion. Collections,
-        // JsonNodes and values such as Guid do not implement IConvertible, so ChangeType cannot
-        // render them. Keep scalar coercion and caller-supplied converters unchanged.
-        if (typeof(T) == typeof(string) && value is not null && value is not IConvertible)
-            return hub.ConvertJson(JsonSerializer.SerializeToElement(value, hub.JsonSerializerOptions),
-                null, defaultValue);
         return value switch
         {
             null => defaultValue,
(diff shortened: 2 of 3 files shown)


rbuergi:
Integration regression found after merge: this diff removes the CLR-collection-to-string branch and both tests from merged #3788. The new numeric fallback is a separate useful behavior, but native gate-name arrays now reach the non-IConvertible fallback and render default instead of readable text. I am restoring the prior value/parity regression tests and collection conversion alongside the new numeric handling in an isolated follow-up; not reverting the numeric fix.

Issue #3625, opened 2026-09-07
The untyped-content gate cannot fire: the phrase it greps is logged without an exception, and the trace sink it scans only takes records that carry one


rbuergi:
Found while checking whether a deliberately-degraded fixture in #3619 would red the shard. It would not — and the reason is that **the gate cannot fire at all** in core's test lane.

## The gate

`dotnet-test.yml:1504` runs `check-untyped-content.sh collected-logs`, which greps the collected shard logs for

```
stayed an untyped JsonElement
```

and reds the shard on a hit. It was made load-bearing on 2026-09-02 (#3056 removed the accidental type-registration that `MessageService.Post`'s Debug serialize was performing, and the degradations it had been masking started surfacing as this warning). The script is careful in every way that matters *given its input*: it separates "found nothing" from "the scan failed" on grep's exit code, refuses a missing directory, and explains the fix.

## The phrase cannot reach `collected-logs`

**Both producers log it as a Warning with NO exception:**

- `src/MeshWeaver.Hosting/MeshNodeStreamCache.cs:2283` — `logger.LogWarning("MeshNodeStreamCache.GetStream: Content for {Path} stayed an untyped JsonElement after …", node.Path, TruncateRaw(je))`
- `src/MeshWeaver.Hosting/MeshNodeStreamCache.cs:2717` — the same, from `GetQuery`

**The sink that feeds the collected file takes a record only when there IS an exception:**

- `test/MeshWeaver.Fixture/LoggingBuilderExtensions.cs:193` — `var isFaultCandidate = exception is not null && logLevel >= LogLevel.Warning;`
- `test/MeshWeaver.Fixture/XUnitFileOutputHelper.cs:507` — the same condition guards `TestTraceLog.AppendFault`

so neither warning ever reaches `collected-logs/_meshweaver-test-trace.log`.

**The other route into the directory is off in this lane.** The collect step globs `*/bin/*/test-logs/*.log`; those per-method files are opt-in via `MESHWEAVER_TEST_FILE_LOGS`, which is set **only** in `node-repo-module-pack.yml` (lines 2163, 2553) — never in `dotnet-test.yml`.

[...]

Issue #2952, opened 2026-09-01
A node whose NodeType has not compiled yet stays untyped FOREVER — the missed message behind the Pandas flake


rbuergi:
The `PandasExplorerLayoutAreaTest` flake (a named member of the unowned flake family, alongside Acme ColdCycle / GitSync Bidirectional / ScopeTeardown) is not a timing wobble to be re-run. It is a **missed message**, and the mechanism is exact.

## What the failing run shows

```
[08:19:23] '$type':'PandasExplorer' is NOT registered in this (receiving) hub
[08:19:29] '$type':'PandasExplorer' is NOT registered in this (receiving) hub
[08:19:29] MeshNodeTypeSource[Doc/DataMesh/PythonPandasNode/PandasExplorer/LiveFrame]:
           content discriminator '$type': 'PandasExplorer' is not a registered type on the
           owning hub — content stays an untyped JsonElement
[08:20:13] FAILED after 50s. Last of 1 emission(s) was:
           MarkdownControl { Id = area-not-found } "No renderer is registered for area `Explorer`"
```

**One emission.** The stream produced the terminal verdict and never spoke again.

## The chain

1. `PandasExplorer` is an in-mesh NodeType — `Source/*.cs` compiled by Roslyn at runtime. The type exists only once that compile has registered it.
2. The `LiveFrame` node's content carries `$type: PandasExplorer`. `MeshNodeTypeSource.ResolveJsonElementContent` (`src/MeshWeaver.Graph/MeshNodeTypeSource.cs:1084`) tries to type it, fails, tries `IMeshContentTypeRegistry.TryRecoverForNodeType`, fails — because the compile has not finished — then **logs a warning and returns the node untyped**.
3. Untyped content ⇒ the hub cannot see it as a `PandasExplorer` instance ⇒ the NodeType's layout areas are never registered.
4. `LayoutDefinition` injects `area-not-found` **only when the render stream COMPLETES** without producing the area — completion being the deliberate structural signal that "no more content is coming". The menu-only render completes, so the placeholder is emitted as a **terminal** answer.
5. The compile finishes moments later and registers the type — **and nothing re-types the already-loaded node.** `ResolveJsonElementContent` runs once, on load. There is no path back.

## The defect, stated plainly

[...]

PR #3797, opened 2026-09-09
fix(#3645): the untyped-content gate reports the VERDICT, not the event

### LINKED ISSUE #3645: The untyped-content gate reports a normal TRANSIENT degradation as a defect — the two seams disagree about the same event
## The two seams describe the same event and disagree about whether it is a failure

`MeshNodeStreamCache` warns at the **instant of a read**, and at that instant it cannot know the
type will register moments later. That is the normal state during portal boot — before the NodeType
compiles land — and it is precisely the race #2952 fixed, by re-typing every live reader when the
registration arrives.

The sibling seam one layer down already says so, in its own message:

> `MeshNodeTypeSource[…/Live]: content discriminator '$type': 'X' is not a registered type on the
> owning hub — content stays an untyped JsonElement **for now** … Two causes look identical here:
> (a) the NodeType's runtime compile has not registered 'X' **YET, which is TRANSIENT** … or (b) no
> declaration will ever claim this discriminator …`

`MeshNodeTypeSource` passes **no exception**, so it never reaches `_meshweaver-test-trace.log` and
reds nothing. `MeshNodeStreamCache` (since #3625) passes `MeshNodeContentDegradedException`, so it
does. **One event, two descriptions, opposite CI consequences.**

## Why this matters

`check-untyped-content.sh` reds a shard on any occurrence, so it currently reports a *normal
transient* as a defect. Measured on #3628: two of `LateContentTypeRegistrationTest`'s three cases
degrade **and recover inside a single test** — the recovery is the assertion — yet both produce a
gate-reddening record. A portal boot that reads an in-mesh NodeType's content before its compile
lands produces the same record for the same non-reason.

The gate is still worth having: a *final* degradation is a real, silent defect, and the gate's first
working run found one (`Ops/Modules/{deployment}`, fixed in #3628). The problem is the denominator —
a hit currently means "content was unreadable at a read", not "content is unreadable".

## Why #3628 did not fix it

#3628's red was fixed minimally and correctly: the three tests capture their own degradation records
and **assert** them (see `Doc/Architecture/ReadingCiSignals` → *"How a test declares that a
degradation is its own SUBJECT"*). Distinguishing transient from final is a different change — it
needs a window that nothing in a process naturally closes — so it is a rework rather than a repair,
and expanding #3628 into it would have been the wrong call.

## Direction (a hypothesis, not a decision)

The platform already HAS the recovery event: #2952 wired `IMeshContentTypeRegistry.Registrations` to
re-type every live reader. So the emitter is not missing information, it is missing a place to
decide *"this one never recovered"*. Candidates, none measured yet:

- report at the **recovery** seam instead — emit the marker only when a degraded path is evicted /
  the reader completes without ever having been re-typed;
- carry the verdict in the record (`NodeTypeCompilationStatus` at the time of the read), so the gate
  can key on `Ok`-and-still-untyped, which is the (b) the message already describes;
- leave the emitter alone and give the GATE the second half — cross-reference each reported path
  against the NodeType's `compilationStatus` in the same run.

## Acceptance

A gate hit means *"content was unreadable and stayed unreadable"*, and a portal boot that reads a
runtime-compiled NodeType's content before its compile lands produces **no** gate hit — with a
falsification in both directions, as #3628 did.

Refs #3625, #2952, #3628.

rbuergi:
Closes #3645.

## The gate was right; its denominator was wrong

A node whose content carries a `$type` the reading hub cannot resolve does not fail. The value
degrades to a raw `JsonElement` and reads as absent everywhere downstream — an `is MyType` check
misses, a view renders empty, a reactive wait never completes. That silence is why there is a gate.

It keyed on `MeshNodeContentDegradedException`, which is emitted **at the instant of a read** — and
at that instant nothing can know whether the type registers a moment later. That is the ordinary
state during a portal boot: a NodeType's runtime compile lands after its first readers are served,
and #2952 exists precisely to re-type those readers when the registration arrives. The seam one
layer down already said so in its own message — *"content stays an untyped JsonElement **for now** …
(a) … has not registered it **YET, which is TRANSIENT** … or (b) no declaration will ever claim this
discriminator"* — and, passing no exception, reddened nothing.

One event, two descriptions, opposite CI consequences. Measured on #3628: **two of
`LateContentTypeRegistrationTest`'s three cases degrade and RECOVER inside a single test** — the
recovery *is* the assertion — and both produced a gate-reddening record. So a hit meant *"content
was unreadable at a read"*, never *"content is unreadable"*.

## The fix invents nothing — it asks again

Both halves already existed. `ContentDegradationRegistry` keeps one entry per node type (added for
`/health`), and `IMeshContentTypeRegistry` answers *"is this resolvable"* as a **pure map lookup** by
either route `TryRecoverForNodeType` takes. So the verdict is computed by re-asking, and needs no
window, no timer and no retained content:

- **`Record`** now keeps the stored `$type` beside the NodeType, so both routes can be re-asked
  later. (First-seen wins — a later read whose element carries no `$type` must not erase the name
  the first one gave us.)
- **`Unresolved(registry)`** drops every entry whose type has since registered. 🚨 A null registry
  means UNRESOLVED, never resolved: a missing instrument may not read as a clean result.
- **`MeshNodeStreamCache.Dispose`** logs one `MeshNodeContentUnresolvedException` per surviving
  entry — **first**, before any teardown, because a verdict nobody can emit is the exact failure
  #3625 was.
- **`check-untyped-content.sh`** keys on that type. The per-read records stay in the logs and stay
  useful: once the verdict names a type, they are how you find which reads degraded on it.

| record | when | meaning | reds a shard? |
|---|---|---|---|
| `MeshNodeContentDegradedException` | at each degraded read | *this read was untyped* — an EVENT, possibly the boot race | **no** (this PR) |
| `MeshNodeContentUnresolvedException` | once per node type, at cache teardown | *the registry was re-asked and still says no* — a VERDICT | **yes** |

## `/health` was wrong the same way

`ContentTypeHealthCheck` fed `Snapshot()`, so a replica stayed `Degraded` after a boot race until
some later read of that type happened to clear the entry — a probe answering about the past. It
reads `Unresolved(...)` now, so a race that healed disappears immediately and only a type that
genuinely never arrived keeps the replica degraded.

## The acceptance criterion, falsified in both directions

Two new cases in `LateContentTypeRegistrationTest` — the file whose subject this already is, which
is also why they belong there rather than in a new file (the diverted-logger guard requires the
capture and the assertion to live together).

- **`ARecoveredDegradation_IsNotTheVerdict`** — degrade, then `registry.Register(...)`:
  `Unresolved` is empty while `Snapshot` is not. *Reduce `Unresolved` to `Snapshot()` and it fails.*
- **`AnUnrecoverableDegradation_IsTheVerdict_AndReachesTheSink`** — nothing ever registers the type;
  drives the **production** teardown (`Dispose`, idempotent) and asserts the record the trace sink's
  own predicate would take (`exception is not null && level >= Warning`), that it names the
  discriminator, and that it carries the phrase the script greps. *Remove the report and it fails on
  "the teardown must REPORT what never resolved".*

Both were run against the reverted code: **2 failed / 3 passed**, each failing exactly its own half.

`UntypedContentDegradationGate` grew from 4 assertions to 6 and now pins: the script keys on
`nameof(MeshNodeContentUnresolvedException)`; the emitter constructs it; **the report is computed
from `Unresolved(`** (a report built from `Snapshot()` would re-red every boot race *invisibly*,
because the marker would still be constructed and the gate would still match); and both read seams
still emit the per-read record the verdict is derived from.

🚨 One thing the guard caught on me: both messages keep the gate's phrase **contiguous in one string
literal**. The guard greps the source line by line, so a phrase the formatter splits across two
literals is emitted correctly and yet reads as absent — a rewording that "passes" while retiring the
coupling. Noted in a comment at both sites.

## Verification

`dotnet build -c Release -warnaserror`, one project per invocation — `MeshWeaver.Data.Contract`,
`MeshWeaver.Hosting`, `Memex.Portal.ServiceDefaults`, `MeshWeaver.Documentation`, and the three test
projects: **0 Warning(s), 0 Error(s)** each. Tests: `MeshWeaver.Graph.Test` **1483/1483**,
`MeshWeaver.Hosting.Test` **534/534** (includes `UntypedContentDegradationReachesTheTraceSinkTest`,
unaffected), `MeshWeaver.Documentation.Test` **368/368** (includes the link-integrity test).

## Docs

- `Doc/Architecture/ReadingCiSignals` → *TRANSIENT from FINAL* rewritten: it recorded this as
  *"needs a window nothing in a process naturally closes — a rework, deliberately left to its own
  issue"*, and that is the issue.
- What's New: *A degradation that recovered is not a defect* (`Category: Fix`).

## Surface

`MeshNodeContentUnresolvedException` is **added**, nothing public is removed, and no interface gains
a member — so neither the cross-repo pair gate nor the interface-additions gate applies.
`ContentDegradation` gains an optional trailing parameter and `Record` an optional trailing
argument, so every existing caller still compiles.

### DIFF
--- a/.github/scripts/check-untyped-content.sh
+++ b/.github/scripts/check-untyped-content.sh
@@ -58,15 +58,45 @@
 # asserted by UntypedContentDegradationReachesTheTraceSinkTest, which drives the PRODUCTION emitter
 # and evaluates the sink's own condition against the captured record. Modelled on
 # EmitDeadAttributionReachesTheTraceSinkTest, which exists for the same trap one diagnostic over.
+#
+# 🚨 WHY THE KEY MOVED FROM THE EVENT TO THE VERDICT (MeshWeaver#3645).
+#
+# For its next stretch this gate could fire, and fired on the WRONG POPULATION. It keyed on
+# MeshNodeContentDegradedException, which is emitted at the INSTANT of a read — and at that instant
+# nothing can know whether the type registers a moment later. That is the ordinary state during a
+# portal boot: a NodeType's runtime compile lands after the first readers have been served, and
+# MeshWeaver#2952 exists precisely to re-type those readers when the registration arrives. The
+# sibling seam one layer down (MeshNodeTypeSource) already said so in its own message — "content
+# stays an untyped JsonElement FOR NOW … (a) … has not registered it YET, which is TRANSIENT … or
+# (b) no declaration will ever claim this discriminator" — and, passing no exception, reddened
+# nothing. One event, two descriptions, opposite CI consequences.
+#
+# Measured on #3628: two of LateContentTypeRegistrationTest's three cases degrade and RECOVER inside
+# a single test — the recovery IS the assertion — and both produced a gate-reddening record. So a
+# hit meant "content was unreadable at a read", never "content is unreadable".
+#
+# The denominator is now the FINAL state, and the platform already had the machinery for it.
+# ContentDegradationRegistry keeps what a replica could not type; MeshNodeStreamCache.Dispose calls
+# Unresolved(), which RE-ASKS the mesh-wide content-type registry per entry — both routes
+# TryRecoverForNodeType takes, two pure map lookups, no content needed — and logs one
+# MeshNodeContentUnresolvedException per node type that is STILL unresolvable. A boot that read
+# before its compile landed leaves nothing behind; a discriminator no declaration will ever claim
+# leaves exactly one record, naming the node type.
+#
+# So the marker below is the VERDICT, not the event. The per-read
+# MeshNodeContentDegradedException records stay in the logs and stay useful — they are how you find
+# WHICH read degraded once the verdict tells you a type never recovered — but they no longer red a
+# shard on their own.
 set -euo pipefail
 
-# The event's IDENTITY. Primary key: compiler-bound, and pinned to this string by
+# The verdict's IDENTITY. Primary key: compiler-bound, and pinned to this string by
 # UntypedContentDegradationGate via nameof(...).
-MARKER='MeshNodeContentDegradedException'
-# The event's DESCRIPTION. Secondary net, kept deliberately: a per-test file log
+MARKER='MeshNodeContentUnresolvedException'
+# The verdict's DESCRIPTION. Secondary net, kept deliberately: a per-test file log
 # (MESHWEAVER_TEST_FILE_LOGS, on in node-repo-module-pack.yml) carries the formatted message with
-# no exception attached at all.
-PHRASE='stayed an untyped JsonElement'
+# no exception attached at all. 🚨 It must match the FINAL report only — the per-read warnings say
+# "stayed an untyped JsonElement", which is the transient population this gate no longer reds on.
+PHRASE='was NEVER resolvable on this replica'
 DIR="${1:?usage: check-untyped-content.sh <collected-logs-dir>}"
 scan_err="$(mktemp)"
 trap 'rm -f "$scan_err"' EXIT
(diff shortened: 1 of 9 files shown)


rbuergi:
## Red, and the surface gate is right — my PR body was wrong

`Public surface (binary compatibility)`:

```
✗ src/MeshWeaver.Hosting/ContentDegradationRegistry.cs
    record ContentDegradation — arity
        primary constructor went from 5 to 6 parameter(s). Adding one — even WITH a default —
        replaces the signature: every assembly compiled against the old arity calls a constructor
        this one no longer has.
```

I wrote *"`ContentDegradation` gains an optional trailing parameter … so every existing caller still
compiles"*. True for **source**, false for **binary**, which is the one that matters for a module
bundle compiled against a previous platform. **No allow entry** — the break is avoidable: the
discriminator becomes a plain `init` property on the record instead of a primary-constructor
parameter, which leaves the constructor arity untouched and reads better anyway.

Second failure, `Run tests (shard 1)`:
`MeshWeaver.Data.Test.ReadDuringDisposalWindowTest.ReadThatFaultsOnTeardown_IsNackedShuttingDown_NotAnsweredAsAbsent`
— the NACK named `Hub … is shutting down (RunLevel=Quiescing …)` where the test wants
`HubDisposingException`. Nothing in this diff touches disposal NACKs, hub creation or
`MeshWeaver.Data`; it is not in `.github/known-flakes.json`. Classifying it before the next push
rather than assuming — it is either pre-existing on `main` or a real race, and both need saying out
loud.

rbuergi:
## Arity break fixed at the source; the shard-1 failure classified as far as it can be from here

**`Public surface (binary compatibility)` — fixed, no allow entry.** `Discriminator` is now an
`init` property on `ContentDegradation` rather than a sixth primary-constructor parameter, so the
record's constructor arity is unchanged. Locally:

```
✓ 6 changed C# file(s): no binary-breaking primary-constructor changes
```

**`Run tests (shard 1)` — `ReadDuringDisposalWindowTest.ReadThatFaultsOnTeardown_IsNackedShuttingDown_NotAnsweredAsAbsent`.**
What is measured:

| | |
|---|---|
| locally, this branch, Release | **passes, 305 ms** (1/1) |
| in `.github/known-flakes.json` | no |
| the six `MeshWeaver Build and Test` runs on `main` and its merge queue since 06:53Z | **all success** |
| files this diff touches in `MeshWeaver.Data` or on any disposal-NACK path | none |

So it is CI-only, which this repo reads as a real timing/state race rather than as noise. I am not
claiming it is unrelated: `ReportUnresolvedContentTypes` runs FIRST in
`MeshNodeStreamCache.Dispose`, and anything added to a disposal path can in principle move a race.
Against that — it returns on `degradations is null || degradations.IsEmpty` before doing anything
else, so on a mesh that degraded nothing (which is every mesh in that shard) the added work is a
single `ConcurrentDictionary.IsEmpty` read.

**Not a re-run to see.** The code under test changed with the commit above, so this run is a fresh
measurement, and it is also the discriminator: if shard 1 fails on the same test again, the finger
points here and I will chase it; if it passes, the race is real and belongs in its own issue with
this run as its first sighting rather than being quietly forgotten.

rbuergi:
## Green — and the shard-1 failure was not this diff

The run after the arity fix is clean: **zero non-success jobs**, including `Public surface (binary compatibility)` and `Run tests (shard 1)`.

So `ReadDuringDisposalWindowTest.ReadThatFaultsOnTeardown_IsNackedShuttingDown_NotAnsweredAsAbsent` passed on the same shard, on a diff that still adds `ReportUnresolvedContentTypes` to `MeshNodeStreamCache.Dispose`. That is the discriminator I said it would be: had the teardown addition moved that race, it would have moved it again here.

It is therefore a **CI-only race in `MeshWeaver.Data`**, seen once, on a test not listed in `.github/known-flakes.json`. Recording it rather than letting it dissolve into 'it went away' — the NACK read `Hub data-owner/1 is shutting down (RunLevel=Quiescing, activation #02FEE553)` where the test requires `HubDisposingException`, i.e. the hub reached `Quiescing` before hosted-hub creation was frozen, so a *different* refusal answered first. If it recurs, that ordering is where to look.

PR #3805, opened 2026-09-09
The provenance line becomes a module control, and a container reports what happened UNDER it

sierragolflima:
## What

`Type · Created · Updated` is drawn by `BuildHeaderMetaRow`, reachable only through `BuildHeader`. A node type that renders its **own** default area therefore ships without it — a CRM client, a board. That is structural rather than a decision: the data is on the node, and every Markdown page already shows the line.

- **`BuildMetaEntries(node, zoneId, lastActivity?)` is pure** — no host, no `AccessService`, stamps through the deterministic `DisplayTimeExtensions.ToDisplayTime`. That is what lets the line be asserted with no mesh and no mock.
- **`BuildMetaRow(host, node, lastActivity?)`** renders it for a module page that has already drawn its own title block and wants only the line.
- **`NodeActivity` replaces the `Updated` segment on a container.** A partition root is written when it is opened and barely again, so its own row reports whatever last touched the ROOT — on a retyped CRM client that is the migration, stamped `system-security` — while the deals underneath moved days later. `Updated 2026-08-28 by system-security` on an account worked yesterday is a wrong statement, not a missing one.

## Two decisions worth reviewing

1. **It replaces rather than joins.** Two last-changed answers on one line, one of them a migration's, is worse than the weaker of the two. A `null` override falls back to the node's own row, so an empty container does not read as one that never existed.
2. **`BuildHeader` gains a fifth OVERLOAD, not a fifth parameter.** Its own doc comment states it is a module-facing contract and that adding a parameter — default or not — replaces the signature every already-compiled module was built against. In-mesh modules compile against the platform image, so this obeys that rule literally.

## Scope note

The four labels are now localized (`node.meta.*`, en + de). They had been hard-coded English since the line was written. I localized them because they are the lines this PR rewrites, not as an adjacent cleanup — a row with three English literals and one translated label would be worse than either.

## Verification

- `dotnet build test/MeshWeaver.Graph.Test -c Release -p:CIRun=true -warnaserror` → `0 Warning(s), 0 Error(s)`
- `NodeMetaEntriesTest` → **10 passed**, 0 failed (red first: the API did not exist)
- `LocalizationTest` → **58 passed** (en/de parity, duplicates, orphans)

## Follow-on

`Systemorph/MeshWeaver.Crm` consumes this to put the line on the client, deal, contact, offer and interaction pages. That PR is **blocked until this ships in a platform image and the Crm pin moves** — the module compiles against the published image (currently `3.0.0-ci.8055`), not against this branch.

Draft on purpose: opened for review, not to land on green.


sierragolflima:
Closing unmerged in favour of the module-local route.

The core change was never strictly necessary: `Crm/Source/Display.cs` already carries both pieces the provenance line needs — the `DisplayTimeExtensions.ToDisplayTime(instant, zoneId).ToString("yyyy-MM-dd HH:mm", …)` formatter and the `AccessService.Context.TimeZoneId` → `CircuitContext.TimeZoneId` resolution. The only thing core withheld was the shared implementation (`BuildHeaderMetaRow` being private), not the capability. Landing this first would have gated a single-repo CRM change behind merge → platform image → pin move, for a de-duplication that pays off only when a second module wants the same line.

Split into two issues:

- Systemorph/MeshWeaver#3823 — the framework gap, as a suggestion. Not urgent: merging it alone changes nothing a user sees, since every page must still call the row.
- Systemorph/MeshWeaver#3824 — the live i18n bug this PR incidentally fixed. `Type:` / `Created:` / `Updated:` are hard-coded English literals and render that way for German-profile viewers on every Markdown and NodeType page. That one is a Fix and stands on its own.

**Branch `fix/node-provenance-byline` is deliberately NOT deleted** — it holds the working implementation plus `NodeMetaEntriesTest` (10 cases, host-free) and the verified `node.meta.*` keys, so both issues are a cherry-pick rather than a rewrite.

PR #3806, opened 2026-09-09
ci(scripts): a platform canary — tell a node repo on the DAY core breaks it, not at the next pin bump

rbuergi:
Implements the platform-canary proposal in Systemorph/MeshWeaver.Plugins#1519. The caller workflow is a separate change in that repository.

The scheduled canary compares selected dependent suites at their current core pin and at one resolved core-main commit. It reports compatibility regressions without becoming a required PR check or moving the pin.

## Evidence and outcomes

- Each arm uses its own caller checkout and builds Release with CI warnings-as-errors flags, then runs tests against that build.
- Test regressions require a recorded **Passed** result at the pin and **Failed** at main, read from fresh TRX files. A missing failure line is never interpreted as a pass.
- A compiler failure at main is reported as a build regression only when the same suite built successfully at the pin. Other process/build failures are incomplete evidence.
- Missing, partial, empty, skipped-only, ambiguous, or inconsistent results prevent a clean verdict. An incomplete comparison cannot close an existing tracking issue. A baseline build failure does not cancel against another failed build and become green.
- A missing or skipped candidate test also cannot establish that an earlier failure healed.
- Per-suite build/test logs, TRX files, and arm evidence are retained as artifacts. The 45-minute job cap remains; each build and test invocation is bounded at eight minutes.
- Tracking-issue discovery uses REST and refuses multiple matching issues.

## Verification

`python3 .github/scripts/test-platform-canary.py`: **15 tests pass**, covering the demonstrated baseline-build-failure false regression, real pass-to-fail comparisons, compiler versus infrastructure failure, missing/skipped tests, partial results, host exits, duplicate names, and theory names. The parser also successfully read an actual five-test xUnit TRX from the local radio-binding reproduction.

Workflow shell, YAML-key, timeout, and secret-preflight checks pass. The actual scheduled comparison still needs its first run after the dependent caller is wired; fixture tests do not establish that the dependent suites pass against core main.

No What's New entry: internal CI reporting only.

Pairs-with: none — CI only; no public framework surface changes.

### DIFF
--- a/.github/scripts/test-platform-canary.py
+++ b/.github/scripts/test-platform-canary.py
@@ -1,0 +1,109 @@
+#!/usr/bin/env python3
+"""Exercise the exact embedded canary evidence code without a GitHub runner."""
+from pathlib import Path
+import tempfile
+import unittest
+import xml.etree.ElementTree as ET
+
+workflow = Path(__file__).resolve().parents[1] / 'workflows/node-repo-platform-canary.yml'
+source = workflow.read_text().split("<<'PYTHON'\n", 1)[1].split('\n          PYTHON', 1)[0]
+namespace = {'__name__': 'canary_under_test'}
+exec(compile('\n'.join(line[10:] for line in source.splitlines()), str(workflow), 'exec'), namespace)
+compare = namespace['compare']
+read_results = namespace['read_results']
+
+
+def arm(**results):
+    return {'suite': {'results': results}}
+
+
+class CanaryEvidenceTest(unittest.TestCase):
+    def test_only_a_recorded_pass_can_be_a_regression(self):
+        delta, incomplete = compare(arm(good='Passed', old='Failed'), arm(good='Failed', old='Failed'))
+        self.assertEqual(['suite: good'], delta)
+        self.assertEqual([], incomplete)
+
+    def test_candidate_compiler_failure_has_a_successful_build_control(self):
+        pin = {'suite': {'build': 'passed', 'results': {'test': 'Passed'}}}
+        main = {'suite': {'build': 'failed', 'error': 'CS0246'}}
+        self.assertEqual((['suite: BUILD FAILED at main (build passed at pin)'], []), compare(pin, main))
+
+    def test_candidate_infrastructure_failure_is_not_a_build_regression(self):
+        pin = {'suite': {'build': 'passed', 'results': {'test': 'Passed'}}}
+        main = {'suite': {'build': 'unknown', 'error': 'host died'}}
+        delta, incomplete = compare(pin, main)
+        self.assertEqual([], delta)
+        self.assertTrue(incomplete)
+
+    def test_baseline_build_failure_never_establishes_a_pass(self):
+        delta, incomplete = compare({'suite': {'error': 'BUILD FAILED'}}, arm(test='Failed'))
+        self.assertEqual([], delta)
+        self.assertTrue(incomplete)
+
+    def test_both_builds_failing_is_not_a_clean_comparison(self):
+        failed = {'suite': {'error': 'BUILD FAILED'}}
+        self.assertTrue(compare(failed, failed)[1])
+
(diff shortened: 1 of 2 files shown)


rbuergi:
Additional evidence for the canary comparison at head `5221b8f553531b6b3eeedd75c9435792e5fb9171`: I extracted the exact `canary-run.sh` from this workflow and ran it against a synthetic dotnet command. At the PIN the command returned a compiler error (no tests ran); at MAIN it returned `Example.Tests.TestOne [FAIL]`. Both arm runners exited 0. Their files were:

```text
pin.txt:  src/Example.Test: BUILD FAILED
main.txt: Example.Tests.TestOne
```

The workflow's set difference then reports `Example.Tests.TestOne` as “passes at the pin, fails at core main”, although it never ran at the pin. This is separate from a missing output file: both expected files exist and contain a classified result. Removing `continue-on-error` or the `touch` alone will not fix it.

Please make the baseline require positive per-test pass evidence (fresh TRX), and classify build/host failures as an incomplete comparison rather than a test pass. Incomplete arms must not close a prior tracking issue. A fixture covering baseline-build-failure + candidate-test-failure should pin this distinction. This is a local workflow-logic reproduction, not a real failure in a dependent suite.

Also, the earlier review's nested-quote warning on `echo "behind=$(gh api "repos/..." ...)"` is not by itself a Bash syntax error: command-substitution parsing has its own quoting context. The evidence/comparison problem above is the actionable one.

rbuergi:
Verified update: d36f769427a5f31e913abe6dfa219910fa6cc00f is now the remote head of this PR. The first push was rejected by the merge queue; I dequeued the PR through dequeuePullRequest, pushed successfully without force, and verified the head with REST. Fifteen focused tests and workflow checks passed locally. Fresh CI/review must complete before this correction merges.

rbuergi:
## 🚨 There are now TWO canaries from Plugins#1519 — this lane, and an inline one in MeshWeaver.Plugins

Flagging a duplication I am partly responsible for. Both landed on 2026-09-09 from the same issue:

| | |
|---|---|
| this PR | `node-repo-platform-canary.yml`, **reusable**, 332 lines — merged 10:31Z |
| MeshWeaver.Plugins | `.github/workflows/platform-canary.yml`, **inline per-repo**, 272 lines |

I filed #1519 and built the shared lane; someone independently built the Plugins one. Neither of us
saw the other.

### They are not redundant copies — each has something the other lacks

The Plugins one is better structured in two ways I did not do:

- a **matrix over the two arms** ("check out the platform at *this arm's* commit"), so the pin arm
  and the main arm run in parallel rather than sequentially;
- a **comparator self-test** — *"Prove the comparator can fail"* — which is exactly the anti-vacuity
  discipline this repo applies everywhere else, and which my delta step does not have.

This lane has what that one does not: it **files and closes a tracking issue**, which is the half
that makes a scheduled red visible at all (the lesson from the platform-ref bump lane going red daily
into an empty room for weeks — Plugins#1547).

### The convergence, and it is the pattern we just used twice today

Same shape as `compile-check.py` (core #3761) and `gen-manifests.py` (Plugins#1426): **port the
better parts UP into the shared lane, then make the caller call it.** Doing it the other way — having
Plugins keep an inline copy — recreates exactly the drift the fleet just finished removing, and it
does so in the one lane whose whole purpose is catching cross-repo drift.

Concretely: this lane takes the matrix and the comparator self-test; Plugins' `platform-canary.yml`
becomes a ~20-line caller with its `suites:` list; the five satellites then get the same canary for
free, which none of them has today.

I am not doing that unilaterally — the Plugins implementation is somebody's fresh work and they may
have reasons for the inline shape I cannot see (its "build the canary set as one graph" step suggests
a build strategy this lane does not model). Whoever owns it: say which way, and I will do the port.

Leaving both as they are until then is the one option that should not stand — two canaries mean two
answers to "would core main break this repo", and the first time they disagree nobody will know which
to believe.

PR #3807, opened 2026-09-09
feat(#3488): the platform gets its first meter — hubs alive, by kind and run level

### LINKED ISSUE #3488: The platform emits NO metrics of its own — zero custom meters in src/, so answering any question about live hub/stream state requires a heap dump that restarts the replica
## The platform emits no metrics of its own — so every question about live state costs a replica restart

Measured 2026-09-06, twice independently:

```
grep -rn "new Meter(" src/                     → 0 results
grep -rln "System.Diagnostics.Metrics" src/    → 0 results
```

**Not one custom meter in the entire codebase, and the namespace is not referenced.** Every counter
any investigation has relied on is a .NET runtime built-in (`dotnet.timer.count` and friends), used as
a proxy for what we actually want to know.

## What this costs, measured tonight

**#3432** asks why 6,925 `Started` `sync/` hubs hold ~2.7 GB of a 3.5 GB heap. Nothing reports:

- live hub count, or the `sync/`-vs-node split
- the `RunLevel` histogram (`Started` / `Dead` / `Starting`)
- stream totals versus disposed
- anything that would show the curve moving after a fix

So the only way to answer is a **heap dump against a live production replica** — and 🚨 **that
restarts the replica**:

```
21:37:38Z  dotnet-dump collect -p 1 --type Heap    (process suspended)
21:39:17Z  kubelet: liveness probe failed, will be restarted
21:39:26Z  Complete                                (106 s of freeze)
21:42:18Z  pod 1/1 Running, RESTARTS 1
```

`/alive` is 15 s × failureThreshold 6 ≈ 90 s. And the freeze is **not write-bound** — `dd oflag=direct`
measured 169 MB/s on the same node, so a 5.5 GB dump is ≈33 s of the 106; the rest is `createdump`
walking ≈50 M objects with threads suspended.

🚨 **The freeze therefore scales with heap size, which produces a catch-22**: the population under
investigation only exists on a replica fat enough that dumping it restarts the replica — and the
restart destroys the state being measured. A freshly restarted portal was back to 2 977 MiB within
five minutes, so ~3 GiB is the FLOOR, not the accumulation.

## Why this is worth fixing rather than tolerating

1. **It makes a whole class of question unaskable in production.** Not merely expensive —
   *self-defeating*, because the measurement destroys its own subject.
2. **It is why #3432 says "cause NOT established" after two sessions.** That title is honest, and the
   honesty is forced by the absence of instrumentation rather than by the difficulty of the defect.
3. **The data mostly already exists in-process.** `HostedHubsCollection` holds the hubs; `RunLevel` is
   a field; the streams are known. This is exposing what the runtime already has, not computing
   anything new.
4. **It closes the loop on fixes.** #3321 step 3 released 1,495 `Dead` hubs and its own author said to
   *expect the curve to flatten, not vanish* — and **nobody can see the curve**. A merged fix whose
   effect cannot be observed is indistinguishable from one that did nothing.

## Suggested v1 — deliberately small

A single meter with the counters the last two investigations actually needed:

- live `MessageHub` count, tagged by address kind (`sync/` vs node vs other) and `RunLevel`
- streams created / disposed / live
- (worth considering) live `IoPool` slots, since a parked pool has been a root cause before

🚨 **Constraints this repo will hold you to:** a meter is process-wide state, so it must be an
**instance on a mesh-scoped singleton**, never `static` (`NoStaticState.md`); nothing `async` in
hub-reachable code; and 🚨 **an observable gauge that enumerates a live collection on the metrics
thread must not park the action block** — the cure would otherwise be a smaller version of the
disease.

🚨 **And a counter that cannot distinguish "zero hubs" from "the meter never ran" is the night's
recurring defect.** Six instruments were measured today returning confident wrong answers rather than
no answer. Emit a denominator, or make absence explicit.

## Scope

Filed rather than actioned: **instrumenting the platform is a maintainer-level scope call**, not
something to slip into an investigation. Recorded here so the cost is visible and the decision is
deliberate.

Related: #3432 (the investigation this blocks), PR #3481 (`Doc/Architecture/PortalHeapIsHubs` — the
dump hazard and the ClrMD walk recipe), #3325.

rbuergi:
Closes #3488.

## The platform had no meter at all

Measured twice, and the issue's own numbers:

```
grep -rn "new Meter(" src/                  →  0 results
grep -rln "System.Diagnostics.Metrics" src/ →  0 results
```

Every counter any investigation had used was a .NET runtime built-in standing in for what was
actually wanted.

**That is self-defeating, not merely expensive.** #3432 asks why ~6,925 `Started` `sync/` hubs hold
~2.7 GB of a 3.5 GB heap; nothing reported live hub counts, the `sync/`-vs-node split or the RunLevel
histogram, so the only way to answer was a heap dump against a live replica — which suspends it
~106 s against a 90 s liveness budget and therefore **restarts** it. The freeze scales with heap
size, so the population only exists on a replica fat enough that measuring it guarantees losing it.
And #3321's author asked readers to *"expect the curve to flatten, not vanish"* while **nobody could
see the curve**.

## What this adds

One meter, `MeshWeaver.Platform`, one instrument: `meshweaver.hubs.live`, tagged `kind` and
`runlevel`.

### The two choices that make the number trustworthy

🚨 **Observable gauges, never counters.** A counter must be incremented at every add site and
decremented at every remove site. Miss one — #3321's **1,485** dead hubs died down a path nobody had
thought about — and the number drifts from reality *while still looking like a measurement*. A gauge
walks the live tree when the collector asks and cannot disagree with it. This repository's own rule:
a field that cannot fail is not a measurement.

🚨 **The tag is the address KIND, never the address.** Per-address means one time series per
`sync/{guid}` — the exact population under investigation — and the meter becomes the second thing
nobody can afford to collect.

**An instance owned by the mesh, not the conventional `static readonly Meter`.** A static would
survive mesh disposal, bleed across every test in a process, and report two meshes' hubs as one
population — the shape `Doc/Architecture/NoStaticState` forbids.

### Wired end to end

A meter nobody collects is the same silence one step later, so this does not stop at the type:

- `ServiceDefaults`' OpenTelemetry block subscribes the name beside `Microsoft.Orleans`;
- `PlatformMetricsHostedService` arms the instance once the root hub exists and registers it for
  disposal **on the hub** — the same shape and the same reason as `RootMeshHubReplyStreamService`
  beside it;
- a fault there is logged and the host starts anyway. **Instrumentation must never keep a host from
  booting**: a portal that will not come up because its meter threw is a far bigger outage than the
  question the meter was added to answer.

`MessageHub.LiveHubTree` is the walk — `internal`, lazy, and bounded by the same
`MaxHostedHubRecursionDepth` as the disposal walks beside it, so a hosted-hub cycle costs a truncated
answer rather than a hung scrape.

## Tests

`PlatformMetricsTest` — 3 cases, 297 ms, each asserting the **measurement** rather than the plumbing
(a meter that exists and reports nothing is worth what the silence it replaced was worth):

- **`TheMeterReportsLiveHubs_ByKindAndRunLevel`** — the acceptance criterion: a real scrape through
  a `MeterListener`, finding root + hosted hub, tagged by kind and by `Started`.
- **`TheKindTagIsTheAddressTYPE_NotTheAddress`** — the cardinality guard: two hubs of one kind must
  fold into **one** series carrying a count of two, and the instance id must never reach a tag.
- **`ADisposedRoot_ReportsNothing_RatherThanThrowing`** — an exception out of a gauge callback is
  swallowed by the metrics infrastructure and the instrument goes **silent**, which is
  indistinguishable from a mesh with no hubs — the exact failure mode this issue is about.

**Falsified:** make `KindOf` return the full address and **2 of 3 fail** (the cardinality guard and
the by-kind assertion); restore it and all three pass.

## Verification

`dotnet build -c Release -warnaserror`, one project per invocation — `MeshWeaver.Messaging.Hub`,
`MeshWeaver.Hosting`, `Memex.Portal.ServiceDefaults`, `MeshWeaver.Documentation` and three test
projects: **0 Warning(s), 0 Error(s)** each.

| suite | |
|---|---|
| `MeshWeaver.Messaging.Hub.Test` | **387** (384 + 3 new) |
| `MeshWeaver.Hosting.Test` | **539** |
| `MeshWeaver.Documentation.Test` | **368** |

## Scope

Deliberately v1, as the issue asks: **hub counts only**. Streams, queues and pending callbacks each
want their own owner and their own scrape cost, and a meter that tries to answer everything on day
one is how instrumentation becomes the thing nobody trusts. What this unblocks immediately is
#3432 — the RunLevel histogram it needs is now a query rather than a heap dump — and the ability to
see whether a fix moved the curve.

## Surface

`PlatformMetrics` is **added**; `MessageHub` gains an `internal` method. Nothing public is removed
and no interface gains a member, so neither cross-repo gate applies.


rbuergi:
## The red, and why this is not a re-push

The first run failed on
`LogonActionFrameworkTest.Two_concurrent_logons_run_a_once_action_exactly_once` — a 30 s
`TimeoutException`, the **same signature** as invariant #6 of the six broken by the global
continuation hop in the abandoned draft #3759.

What the evidence said:

| check | result |
|---|---|
| that test, 3 local runs on this branch | **green**, all three |
| `main`, last 5 runs | green |
| in `.github/known-flakes.json`? | **no** |

So the evidence settled it in neither direction — which is exactly the situation in which
"re-run and see" is the wrong move.

## What was actually wrong with the diff

The arming lived in `MeshHostApplicationBuilder`. That puts a hosted service **resolving the root
hub** into *every mesh in the fleet* — including the several thousand a test run builds, **none of
which has a collector**. So the one plausible way this diff could touch a concurrency test was a
change in mesh startup timing, and the diff had no business being on that path in the first place.

Instrumentation that alters the startup of hosts that will never be scraped is cost with no reader,
and it makes every future timing-sensitive failure a question about the meter.

## The change

The arming moves to the portal's `ServiceDefaults`, **beside the `AddMeter` subscription it pairs
with** — both halves of the wiring in one place, in the host that actually collects:

```csharp
builder.AddOpenTelemetryExporters();

// 🚨 The OTHER half. Subscribing the meter NAME above only says "collect this if it
// exists"; something must construct the instance, and the gauge closes over the root hub.
builder.Services.AddPlatformMetrics();
```

Exactly one site arms it now — verified by grep across `src/`, `memex/`, `test/` and `tools/`:

```
memex/aspire/Memex.Portal.ServiceDefaults/ServiceDefaults.cs:168:  builder.Services.AddPlatformMetrics();
```

**A test mesh is now untouched by this PR at all** — that removes the doubt rather than assuming it
away.

## Verified locally on the new head

```
solution build, Release -warnaserror ...... 0 Warning(s)  0 Error(s)
MeshWeaver.Graph.Test ..................... 1486 passed, 0 failed   (incl. the failing test)
MeshWeaver.Hosting.Test ...................  539 passed, 0 failed
MeshWeaver.Messaging.Hub.Test .............  387 passed, 0 failed
```

The reasoning is recorded on `PlatformMetricsHostedService` itself, so the next person who reaches
for the mesh builder meets it at the rule.

rbuergi:
All four review findings were real and are fixed. One of them is a genuine defect in the thing this
PR exists to guarantee.

### 1. `KindOf` re-admitted host ids to the tag — the cardinality the tag exists to bound

`ToString()` is `Path + '~' + Host`. Splitting it at the first `/` is right only when the path has
**two or more** segments; with one, that `/` comes out of the **host chain** and the kind became
`solo~portal`. One series per host id — the exact failure the design paragraph forbids, shipped
looking green because the existing test uses a two-segment address where the parse lands inside the
path.

```csharp
- var text = address?.ToString();
- var slash = text.IndexOf('/');
- return slash > 0 ? text[..slash] : text;
+ var type = address?.Type;                       // Segments[0], host-free, no parsing
+ return string.IsNullOrEmpty(type) ? "unknown" : type;
```

🚨 **My first test for this passed against the old code.** A hub whose own address carries a host
chain is not constructible through the public hub API this rig uses, so the scrape never saw a `~`
and the negative control came back green. An assertion that cannot fail is not one, so it is
replaced by `ToStringCarriesTheHostChain_TypeDoesNot`, which pins both halves on `Address` itself,
where the property is decidable:

```
hosted.ToString() == "solo~portal/xyz"          // and the first '/' is at index 11 — in the HOST
hosted.Type       == "solo"
```

### 2. `Observe` promised something it did not implement

The summary said it reports nothing for a torn-down root; nothing *throws* on a disposed hub, so it
happily tallied one. Now guarded on the terminal run level — **the root only**: a `Dead` hub *inside*
a live tree is precisely what the runlevel tag exists to show.

### 3. `StopAsync` claimed dispose-once and did not do it

The field is now cleared so it does. `Meter.Dispose` happens to be idempotent today; that is not a
property to lean on during shutdown, and a comment asserting semantics the code lacks is worse than
neither.

### 4. The vacuous assertion

`NotBeNull` on a list that can never be null tested the half named in the method name and not the
half named in the sentence. It now asserts emptiness — **and waits for the terminal state first**,
because `Dispose()` returns *before* the hub is `Dead` (teardown drains in-flight work). Scraping
straight after the call measures a hub winding **down**, not one that is gone, and asserting
"nothing" there would have been asserting the wrong property.

### Unrelated: the red on the previous run was mine, and not the meter

`EveryArchitecturePage_IsListedInTheTopicMap` failed because a `git add -A` here swept in
`DepartedPlatformAssemblies.md` from an untracked file belonging to #3810 — without the Architecture
index line, which lives on that branch. The gate was right. #3810 has since merged, so the page is on
`main` with its index line; this branch no longer carries it.

### Verified on the new head

```
solution build, Release -warnaserror ...... 0 Warning(s)  0 Error(s)
MeshWeaver.Messaging.Hub.Test .............  388 passed, 0 failed
MeshWeaver.Documentation.Test .............  368 passed, 0 failed
MeshWeaver.Hosting.Test ...................  539 passed, 0 failed
MeshWeaver.Graph.Test ..................... 1486 passed, 0 failed
```

Issue #3706, opened 2026-09-08
The catalog lists three packages no repository publishes (Agent, Skill, PlatformUI) — their retired floors hold every self-update on memex forever


rbuergi:
## The catalog lists packages no repository publishes — and their retired floors hold every self-update

Measured 2026-09-08 09:0xZ on `memex` (the registry). `search nodeType:Package namespace:Plugins`
returns 77 packages. Three of them exist in **no** repository's `main` today, yet still carry the
retired `rc` / clean-`3.0.0` floors that MeshWeaver#3554 now refuses at build time and that
MeshWeaver.Plugins#1447 moved for every package that *does* have a home:

| catalog node | version | last modified | declared floor (from the hold) | found in a repo? |
|---|---|---|---|---|
| `Plugins/Agent` | 10 | 2026-08-25 | `3.0.0` | no — absent from Plugins and core `main` (present only in worktrees at older commits) |
| `Plugins/Skill` | 20 | 2026-08-27 | `3.0.0` | no — same |
| `Plugins/PlatformUI` | 1 | 2026-09-02 | `3.0.0-rc8` | no — same |

(`Plugins/Crm` — `3.0.0-rc4` — *does* have a home; MeshWeaver.Crm's fix is its own PR.)

## Why it matters

memex's `Admin/UpdatePolicy` at 22:27Z yesterday held `3.0.0-ci.8057` with, among others:

```
Agent: the module requires platform 3.0.0 or newer but this deployment runs 3.0.0-ci.8057 …
Skill: the module requires platform 3.0.0 or newer …
PlatformUI: the module requires platform 3.0.0-rc8 or newer …
```

RollSelect requires "63 plugin(s) … from the install records (Plugins/*, nodeType:Package)". A
package no repository publishes can never be re-baked for a new identity and can never have its
floor moved by a PR — so as long as the install records name these three, **no candidate can ever
clear** for this install, regardless of what every other repo fixes. The same shape shows in the
portal log's gating lines: *"a package listed by the registry but never PROVISIONED on this mesh …
retrying cannot help"*.

## The decision (not taken here — it changes catalog content)

1. **Delist / retire** the three catalog nodes (and the install records that name them), if the
   packages are genuinely gone (Agent/Skill moved with the AI engine, MeshWeaver#2276).
2. **Republish** them from whichever repo now owns them, with ci-line floors.
3. Make the updater treat "required, but no publisher exists" as a named, non-blocking
   condition rather than an eternal hold.

[...]

Issue #3651, opened 2026-09-07
A platform roll is gated on measured loadability on the target, never on floors; a missing bake is a boot compile, not a hold


rbuergi:
## Rule (maintainer, 2026-09-07): a platform roll is gated on what can LOAD on the target, never on declared floors — and a missing bake is a boot compile, not a hold

Today `ReleaseAvailability.Evaluate` declines a whole release when any installed package (a) declares a floor above the target tag (`ModuleFloorExceedsTarget`), or (b) has content but no sealed bake for the target identity (`ContentBakeMissing`), or (c) the sealed set is inconsistent. On 2026-09-07 (a) held memex-cloud on 8009 for the whole day while every candidate would have loaded; (b) would have held it again because the satellites' bakes for the new identity did not exist, although a boot compile of those courses succeeds on every PR.

### Change

- (a) is removed as a blocker (see the floors-advisory issue).
- (b) becomes **advisory**: "would recompile at boot: education, crm" is reported in the verdict and on the Updates tab, but does not hold the roll. `Modules:RequirePrebuilt` (the strict mode) keeps its meaning for instances that opt in.
- The **only** hold on the module lane is measured: for each installed module that has no build published for the target identity, `ModulePlatformLink.Check(landed generation bytes, target surface)` must say `Linkable`. This needs the target's surface at gate time: `publish-bake-bundles.sh` writes `platform-surface.json` (assembly → exported type names, the same set `ModulePlatformSurface.OfRunningProcess` builds) beside `_complete` for every identity; `PublishedBundleCatalogue` reads it; `ReleaseAvailability` consumes it. A release with no surface file is `Indeterminate` for the link check only — and Indeterminate is reported, not treated as clearance and not treated as a hold (the boot-time probe + keep-the-old fallback + readiness stall are the safety net; see the policy page).
- `RollSelection.Select` therefore walks newest-first and stops at the first release with no UNLOADABLE module; "no complete release" (every candidate has an unloadable module) stays put and names the module.

### Tests
`SealedSetConsistencyTest`/`ReleaseGateDenominatorTest`: bake-missing rows → advisory; new `ReleaseLinkGateTest`: a module unlinkable against the target surface holds and names the missing type; surface absent → Indeterminate reported, roll proceeds; `RollSelectionTest` updated.

Part of the module-adoption policy (Doc/Architecture/ModuleAdoptionPolicy). Depends on the floors-advisory change.

rbuergi:
[...]

Issue #3648, opened 2026-09-07
Module floors are advisory: a declared minMeshVersion never refuses, holds or skips — loadability is measured


rbuergi:
## Rule (maintainer, 2026-09-07): a declared version floor is advisory — loadability is measured, never declared

Today `content.minMeshVersion` is compared with `NuGetVersionComparer` at **eight** decision points and a miss REFUSES or HOLDS: `ModuleUpdateDecision` (`SkipPlatformBelowFloor`), `PluginBundleClient.LandFromBundle` ("DECLINED … nothing landed"), `ModuleLandingService.LandCore` (`held = ModulePlatformFloor.DeclineReason(...) ?? LinkHoldReason()`), `ModuleActivationBoot.ComputeEffectiveModuleEntries` (boot SKIP), `ReleaseAvailability.Evaluate` (`ModuleFloorExceedsTarget` blocks the whole release), `RequiredModuleStatus` (`ExpectedLater`), `PendingModuleActivations`, `ModuleActivationStatus`. The comparator ranks `ci < rc < clean`, so on 2026-09-07 every installed `Plugins/*` Package record carrying an `rc` or `3.0.0` floor made memex-cloud's self-updater decline all 11 candidate releases ("77 plugins required … every one declined"), while #3552's measured link probe would have loaded every one of them.

### Change

`minMeshVersion` becomes **advisory**: it is logged (Information) and shown on the module's status row as *"declares platform ≥ X; running Y"*, and it decides nothing. The only platform gate on the module lane is the measured one — `ModulePlatformLink.Check` (type-level link against the running surface) at landing and at boot (`MeshBuilder.InstallAssemblies`), plus the actual load. Concretely:

- `ModuleUpdateDecision.Decide`: drop step 2; a served bundle whose floor exceeds the platform proceeds to `Land`, where the link probe decides.
- `PluginBundleClient.LandFromBundle`: the floor check becomes a log line; the name-drift refusal stays.
- `ModuleLandingService.LandCore`: `held = LinkHoldReason()` only; the floor is written to the entry and logged.
- `ModuleActivationBoot.ComputeEffectiveModuleEntries`: no floor skip; the DLL-exists check stays; the link probe in `MeshBuilder` remains the refusal.
[...]

Issue #3506, opened 2026-09-06
A hub past [QUIESCE-OK] still accepts new correlated requests — the intake gate opens one phase too late


rbuergi:
## A hub past `[QUIESCE-OK]` still ACCEPTS new correlated requests, and they can never be answered

Split out of #3499, which measured this shape but was wedged by a different one. #3499's fix does
NOT cover this; nothing does.

## The defect

`MessageService.ScheduleNotify`'s shutdown gate refuses new deliveries only from
`MessageHubRunLevel.DisposeHostedHubs` onward:

```csharp
if (hub.RunLevel >= MessageHubRunLevel.DisposeHostedHubs
    && delivery.Message is not ShutdownRequest and not DisposeRequest)
```

The **Quiescing** phase is therefore wide open. A hub that has entered `[QUIESCE-START]` — and even
one that has already logged `[QUIESCE-OK]`, having drained everything it owed — keeps accepting
inbound deliveries, keeps handling them, and keeps registering response callbacks for requests it
POSTS. #3261 measured it creating hosted hubs in that window too:

```
11.928 [QUIESCE-START] TestData/dnt…: 0 pending callbacks at dispose entry: <none>
11.928 [QUIESCE-OK]    TestData/dnt…: drained 0 callback(s) in 0ms
11.968 Start processing SubscribeRequest (ID: p2qsFux3…, cache/iIN4… -> TestData/dnt…)
11.969 Creating new hosted hub for address sync/4wzOafy… in host TestData/dnt…
```

Work taken on after the drain has, by construction, no drain left to finish it: the quiesce budget
has already been spent, so the next phase cancels it.

## Measurement (2026-09-06, three independent bake runs)

Counted from the `RECEIVED runLevel=<level>@<issuing hub>` stage in the request-fate trails printed
by `[QUIESCE-TIMEOUT]`. **Six of nine** pending callbacks were registered by a hub that was ALREADY
`Quiescing`:

| run | hub | runLevel at intake |
|---|---|---|
| core `main-cd` #7937 ([34058018897](https://github.com/Systemorph/MeshWeaver/actions/runs/34058018897)) | Cornerstone, Publish, RemoteControl, Voice | `Quiescing` |
| core `main-cd` #7937 | Chess | `Started` |
| core `main-cd` #7941 ([34061838257](https://github.com/Systemorph/MeshWeaver/actions/runs/34061838257)) | RolePlay | `Started` |
[...]

PR #3815, opened 2026-09-09
fix(#3812): a hold nothing can recompute is history, not the current verdict

### LINKED ISSUE #3812: A hold nothing can recompute still renders as the live reason — a frozen heldReason beside a fresh lastCheckedAt
## A hold nothing can recompute still renders as the live reason

`UpdatePolicyContent.IsHeld(tag)` answers **"is there a hold record for this tag"**. Both production
readers use it to answer a different question — **"is this why the install is not moving right
now"** — and those stopped being the same question the moment a check could decline to evaluate.

Measured on `memex`, `get @Admin/UpdatePolicy`, 2026-09-09 10:42Z:

```
heldAt          : 2026-09-07T22:27:17Z
heldTag         : 3.0.0-ci.8057
lastCheckedAt   : 2026-09-09T10:41:24Z          <- 36 h later
lastCheckVerdict: "updates are disabled on this install (Admin/UpdatePolicy = None);
                   the registry was not listed."
```

`HeldAt`/`HeldReason` are written **only** by `RecordHold`, which runs only when a candidate is
actually evaluated. `LastCheckedAt` is written by `RecordCheck` on **every** check. Under
`UpdatePolicyKind.None` the check records "updates are disabled" and never evaluates — so the hold
is never revisited, never cleared, and the fresh `LastCheckedAt` beside it makes a two-day-old
record read as a current verdict.

**#3795 (merged today) sharpens this**: under policy `None` the `BuildCompletion` /
`ModuleSetProposed` triggers are now filtered out before they check, so the hold is frozen harder
than it was.

## It has already caused one misdiagnosis

#3706 was filed on this exact frozen record. Its headline — *"their retired floors hold every
self-update on memex forever"* — read three floor lines as a live verdict. They were computed at
2026-09-07 **22:27Z**; #3648 (floors become advisory) merged **00:22Z** and #3651 (only measured
unloadability holds) merged **01:10Z** the next morning. The quoted evidence predates its own fix by
two hours, and nothing in the record says so. See
[#3706 (comment)](https://github.com/Systemorph/MeshWeaver/issues/3706#issuecomment-5600560051).

## Where it surfaces

| reader | line | what it says |
|---|---|---|
| `UpdatePolicySettingsTab.cs:211` | `if (content.IsHeld(tag))` | renders `ui.updateHeld` — *"this install is not moving because…"* — with the stale reason quoted underneath |
| `PlatformUpdateStatus.cs:99` | `policy.IsHeld(latest) \|\| blocked` | returns `UpdateHeld` on the About surface |

The tab does print `ui.updateHeldAt`, so the date is on screen — but it reads as *when the hold
started*, not as *the last moment anything checked*. A reader has no way to tell a live hold from a
frozen one.

The actual reason this install is not moving is already recorded, one field away:
`updates are disabled on this install (Admin/UpdatePolicy = None)`.

## What correct looks like

The record is fine — a durable note of the last evaluation is worth keeping. The **presentation** is
what conflates a record with a verdict.

1. A predicate that distinguishes them, on the content itself (it already carries `Policy`): a hold
   is **operative** only while the checker that wrote it would still clear it.
2. The tab renders a non-operative hold as **history** — *"when last checked on «date», «tag» was
   held because …"* — and states the current reason (updates are disabled) as the current reason.
   New keys go in both `strings.en.json` and `strings.de.json`.
3. `PlatformUpdateStatus.Derive` stops returning `UpdateHeld` off a non-operative note. A **Red combo
   verification** is a recorded fact and keeps holding — the existing comment already draws exactly
   that line (*"the recorded verdict is the fact; IsHeld is the poller's note about it"*); this
   extends it to the case where the note can no longer be refreshed.

🚨 Not "clear the hold when updates are disabled" — that destroys the last real evaluation, which is
the only diagnostic an operator has when they turn updates back on.

Found while correcting #3706.

rbuergi:
Closes #3812.

## Two fields, two clocks, one apparent story

`HeldTag` / `HeldReason` / `HeldAt` are written **only** by `RecordHold`, which runs only when a
candidate is actually evaluated. `LastCheckedAt` is written by `RecordCheck` on **every** check.
While updates are switched on those move together and nothing distinguishes them. Switch updates
off and they come apart: the check keeps recording *"updates are disabled on this install"*, the
last real verdict sits underneath unchanged — and the fresh timestamp beside it makes it read as
current.

Measured on `memex`, `get @Admin/UpdatePolicy`, 2026-09-09 10:42Z:

```
heldAt          : 2026-09-07T22:27:17Z
heldTag         : 3.0.0-ci.8057
lastCheckedAt   : 2026-09-09T10:41:24Z          <- 36 h later
lastCheckVerdict: "updates are disabled on this install (Admin/UpdatePolicy = None);
                   the registry was not listed."
```

**#3795, merged this morning, sharpened it**: under policy `None` the `BuildCompletion` /
`ModuleSetProposed` triggers are now filtered out *before* they check, so nothing revisits the hold
at all.

## It already cost a day

#3706 was filed on that exact frozen record, arguing three retired packages *"hold every self-update
on memex forever"*. Those lines were computed at **22:27Z on 09-07**; #3648 made floors advisory at
**00:22Z** and #3651 reduced the hold to measured unloadability at **01:10Z** the next morning. The
evidence predated its own fix by two hours and the record could not say so. Of the eleven lines in
that `heldReason`, ten are advisories under today's rules; the one real blocker is #3732's two-build
inconsistency. Full correction: [#3706 (comment)](https://github.com/Systemorph/MeshWeaver/issues/3706#issuecomment-5600560051).

## The change

`IsHeld(tag)` answers *"is there a record"*. Both readers used it to answer *"is this why the install
is not moving right now"*. Those are now different questions, so they are different methods:

| | asks | true when |
|---|---|---|
| `IsHeld(tag)` | is there a record | `HeldTag == tag` |
| `IsHoldOperative(tag)` | is it a live verdict | …and a poller is running that would clear it |

- **Updates tab** — an operative hold renders as before; a frozen one renders as history (*"This is
  a record, not a current verdict … nothing has re-checked `{tag}` since «date»"*). The
  `(held «date»)` suffix goes with the live framing, because that phrasing describes something still
  happening.
- **`PlatformUpdateStatus.Derive`** — stops answering `UpdateHeld` off a note nothing can refresh. A
  **Red combo verification keeps holding** regardless: that is a recorded fact about the build, not a
  note about the poller. The surrounding comment already drew exactly that line; this extends it to
  the case it did not anticipate, and there is a positive control pinning it.

🚨 **The record is deliberately NOT cleared when updates are switched off.** The last real evaluation
is the only diagnostic an operator has *before* turning updates back on. Destroying it to avoid
showing something stale trades a misleading answer for no answer. Only the framing changes — the
reason is still quoted in full.

## The negative control

The new suite was run against the pre-fix predicate (`IsHoldOperative` ≡ `IsHeld`) to prove it is not
vacuous:

```
FrozenHoldIsHistoryTest.TheAboutSurface_DoesNotReportAFrozenHoldAsHeld        [FAIL]
FrozenHoldIsHistoryTest.WithCheckingOFF_TheHoldIsRecordedButNotOperative      [FAIL]
FrozenHoldIsHistoryTest.AFrozenHoldWithNoTimestamp_StillReadsAsHistory        [FAIL]
FrozenHoldIsHistoryTest.AFrozenHold_ReadsAsHistory_NeverAsTheCurrentVerdict   [FAIL]
Failed! - Failed: 4, Passed: 5
```

The five that still passed are the operative-path assertions and the Red-verdict control — correctly
unaffected.

## One existing test changed, and why

`PlatformUpdateStatusTest.ANewerTagTheAvailabilityGateHolds_ReadsAsHeld_NotAsAvailable` did not set
`Policy`, and `Policy`'s default is `None`. Its premise — *a poller is running* — always mattered and
was never written down; it is now stated. That makes it a different case from the frozen record
rather than an accidental duplicate of it.

## i18n

Two keys, `ui.updateHoldHistorical` and `ui.updateHoldHistoricalUndated`, added to **both**
`strings.en.json` and `strings.de.json` (1433 keys each, parity verified). Inserted beside
`ui.updateHeldAt` rather than re-serialising the catalogue.

🚨 **The mirror does not see this.** `MeshWeaver.Plugins/clients/react/src/i18n/` compares against a
*pinned* core commit, not core's `main`, so adding keys here reddens nothing anywhere and leaves the
mirror silently stale — which is why the handover is declared rather than assumed:

Mirror-sync: run `npm run sync:i18n -- --ref <this PR's merge sha>` in MeshWeaver.Plugins after this merges, to carry `ui.updateHoldHistorical` and `ui.updateHoldHistoricalUndated` (en + de) into `clients/react/src/i18n/`. The sync runs in that repository and cannot be done from here.

## Observed, deliberately not changed

Under policy `None` the poller does not list the registry, so `LatestAvailableTag` is **also** frozen
— the About surface can therefore say "update available" about a tag that is no longer the newest.
That predates this PR and is not what it is about; the Updates tab dates the tag (`ui.updateCheckedAt`)
and the new wording states outright that nothing has re-checked. Flagging it rather than folding an
unrelated behaviour change into this diff.

## Verified locally

```
solution build, Release -warnaserror ......... 0 Warning(s)  0 Error(s)
MeshWeaver.Documentation.Test ................  368 passed, 0 failed
MeshWeaver.Messaging.Hub.Test ................  384 passed, 0 failed
Memex.Portal.Shared.Test ..................... 1188 passed, 1 failed  (see below)
```

🚨 **The one failure is PRE-EXISTING on `main`, not from this diff.**
`AnInstanceReportsWhatItRunsTest.TheReport_CarriesBothModuleShapesAndThePlatform_AndLandsWhereAReportedOneWould`
fails here, and it fails **2 of 2 runs on a clean `origin/main` worktree with no part of this
change** — so it is deterministic on the baseline rather than a flake this PR provoked. It passes in
isolation (`--filter`), so it is parallelism-dependent. Filed separately; nothing in this diff is on
`DeploymentReportService`'s path (three pure functions and two catalogue keys).

Pairs-with: none — additions only; no public type or member is removed.


rbuergi:
CI red is **#3827**, not this diff: `ReadDuringDisposalWindowTest.ReadThatFaultsOnTeardown_IsNackedShuttingDown_NotAnsweredAsAbsent`, which fails on a **clean `origin/main` worktree at `3ba04254c`** with no diff (run 1 of 2 red locally), and reddened #3797 earlier today.

It is not a plain flake: the test relies on *"message intake stays open until `DisposeHostedHubs`… by construction"*, and **#3506 closed exactly that window** — `QuiescingHubRefusesNewWorkTest` describes the same sentence as the defect it fixed. What is left is an unordered race between the test's read and the `ShutdownRequest` that `Dispose()` posts from inside the `DisposeRequest` handler. Full analysis on #3827.

This branch changes a settings-tab predicate, two catalogue keys and docs — nothing on the hub teardown path. Re-running the failed jobs on that evidence.

(The `i18n catalog` gate is green now: it reads the body from the `pull_request` **event payload**, so the earlier `--failed` re-run replayed the pre-edit body. A pushed commit was needed to carry the `Mirror-sync:` line — now documented in `Doc/Architecture/Localization`.)

rbuergi:
Merging. Required check **`Consolidate test results`: success**.

The one red is `Bundle registry scripts (executed)`, which is **not a required check** and failed on infrastructure, not on this diff:

```
Traceback (most recent call last):
urllib.error.HTTPError: HTTP Error 500: Internal Server Error
```

GitHub was returning 500s in that window — my own `git push` to this repo took one at 11:51Z. Nothing in this diff (a settings-tab predicate, two catalogue keys, docs, tests) reaches bundle registry scripts.

The earlier test red was #3827, with a control on clean `origin/main` (1 in 4, no diff).

PR #3821, opened 2026-09-09
diag(#3510): [QUIESCE-START] names who asked for the teardown

rbuergi:
Addresses one of #3510's three Expectations — *"and `[QUIESCE-START]` on a root should name who
asked"* — which is the one that gates the others, because the issue could not identify the disposer
at all.

## The unknown this removes

#3510's wedge: the Hosting root was disposed at 23:37:51Z **while its own 145-file install was in
flight**. Its per-node children went with it, the writes they owed acks for were stranded
(`ADVANCE_WITHOUT_HANDOFF`), four creates never got a reply, and the install ran out a ten-minute
bound. The issue's own words:

> **Who disposed the root is not in the log at this level** — no `[QUIESCE-START] Hosting`, no
> dispose source line. The leading candidate is the NodeType rebind recycle […] i.e. #3499's
> hypothesis (2), which was never measured because (1) was found first.

So the leading hypothesis stayed a hypothesis through the whole investigation.

## 🚨 Why the obvious source cannot supply it

`Dispose()` posts its `ShutdownRequest` **to itself**:

```csharp
Post(new ShutdownRequest(MessageHubRunLevel.Quiescing, Version));
```

so that message's `Sender` is always the dying hub. **The field that looks like it answers "who
asked" is the one field that cannot.** That is why nothing was there to read, and why reaching for
the sender on the shutdown turn would have produced a confidently wrong answer.

The discriminating fact is one frame earlier: whether a `DisposeRequest` arrived **over the bus** at
all, and from where. It is recorded in `HandleDispose`, at the point such a request is *honoured* —
not merely received, since the root-mesh refusal above it returns without disposing.

## What the line says

```
[QUIESCE-START] Hosting: requested by portal/installer (routed DisposeRequest); 4 pending callbacks …
[QUIESCE-START] Hosting: requested by itself — a self-posted DisposeRequest (Hosting), i.e. a rebind or self-heal recycle; …
[QUIESCE-START] Hosting: requested by a direct Dispose() (no routed DisposeRequest); …
```

| reading | means | rules out |
|---|---|---|
| a named sender | another hub asked — `PackageInstaller` posting to a package root is #3510's shape | a recycle; a host teardown |
| `itself — a self-posted DisposeRequest` | an automatic recycle: `NodeTypeRebindWatcher`, `WithOverlaySelfHeal` | an external actor |
| `a direct Dispose() (no routed DisposeRequest)` | host teardown, an owner disposing children, a `using` | **the whole message path** |

🚨 **The third is not an absence of information** — it rules out an entire class of cause in one
reading, which is the deduction #3510 could not make.

🚨 **The self-posted case is called out rather than printed as an address.** The automatic recycles
post to their OWN hub (`NodeTypeRebindWatcher`, `WithOverlaySelfHeal`), so a bare sender would render
`Hosting: requested by Hosting` — true, useless, and easy to misread as a routing oddity. That case
is #3510's leading hypothesis, so it is the one reading that must not be ambiguous.

## The negative control

`QuiesceStartNamesTheAskerTest` pins all three readings. Run against the previous log line:

```
ADirectDispose_SaysNoRoutedRequestBroughtItDown              [FAIL]
ASelfPostedDisposeRequest_SaysSoRatherThanNamingTheHubTwice  [FAIL]
ARoutedDisposeRequestFromAnotherHub_NamesThatSender          [FAIL]
Failed! - Failed: 3, Passed: 0
```

🚨 **And the first version of that test observed nothing at all** — a precondition guard caught it
rather than a vacuous pass. `TestBase` binds log-level filters from `test/appsettings.json`
(`logging.AddConfiguration("Logging")`), which leaves `MeshWeaver` above `Information`, so the line
is dropped **before any provider sees it**. The sibling `DisposalDeadlockDiagnosticsTest` works
without noticing because the verdicts it captures are `Error`s. The test now raises its own filter in
its own DI — the src-tree `appsettings.json` is committed contract and is not edited to make a test
see a line.

## What this does NOT do

It does not stop a shutdown landing mid-install. #3510's other two Expectations — defer a recycle
while an install holds the root, **or** NACK every in-flight write under it so the install fails in
milliseconds with a name instead of running out a ten-minute bound — are untouched, and the issue
stays open for them. This makes the next occurrence explicable instead of a dead end.

Docs: a new section in `Doc/Architecture/DisposalStallVerdicts` (the three readings and what each
rules out) plus a What's New entry.

Pairs-with: none — additions only; no public type or member is removed.

### DIFF
--- a/src/MeshWeaver.Documentation/Data/Architecture/DisposalStallVerdicts.md
+++ b/src/MeshWeaver.Documentation/Data/Architecture/DisposalStallVerdicts.md
@@ -256,6 +256,50 @@ mesh, which the attached recursive snapshot lists.
 
 ---
 
+## Who asked for the teardown
+
+`[QUIESCE-START]` names the requester (#3510):
+
+```
+[QUIESCE-START] Hosting: requested by portal/installer (routed DisposeRequest); 4 pending callbacks …
+[QUIESCE-START] Hosting: requested by itself — a self-posted DisposeRequest (Hosting), i.e. a rebind or self-heal recycle; …
+[QUIESCE-START] Hosting: requested by a direct Dispose() (no routed DisposeRequest); …
+```
+
+**Why it was missing, and why the obvious source could not supply it.** `Dispose()` posts its
+`ShutdownRequest` to ITSELF, so that message's `Sender` is always the dying hub — the field that
+looks like it answers "who asked" is the one field that cannot. The discriminating fact is one frame
+earlier: whether a `DisposeRequest` arrived over the bus at all, and from where. That is recorded in
+`HandleDispose`, where such a request is *honoured* (not merely received — the root-mesh refusal
+returns without disposing).
+
+**The three readings, and what each rules out:**
+
+| line says | means | rules out |
+|---|---|---|
+| a named sender | another hub asked — e.g. `PackageInstaller` posting to a package root | a recycle; a host teardown |
+| `itself — a self-posted DisposeRequest` | an automatic recycle: `NodeTypeRebindWatcher`, `WithOverlaySelfHeal` | an external actor |
+| `a direct Dispose() (no routed DisposeRequest)` | host teardown, an owner disposing its children, a `using` | **the whole message path** |
+
+🚨 The third is not an absence of information — it is the deduction #3510 could not make. That issue
+turned on one unknown: the Hosting root was disposed at 23:37:51Z *while its own 145-file install was
+in flight*, its per-node children went with it, the writes they owed acks for were stranded, four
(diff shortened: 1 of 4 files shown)


rbuergi:
CI red is **#3826**, not this diff: `LogonActionFrameworkTest.Two_concurrent_logons_run_a_once_action_exactly_once`, which fails **1 in 3 on a clean `origin/main` worktree at `3ba04254c`** with no diff at all, and has reddened three unrelated pull requests today (#3807, this one, and #3825's local run). This branch adds a log line and its test; there is no path from it to `LogonActionRunner`.

Re-running the failed jobs on that evidence — the control is in #3826, so nothing is being waved through on a hunch.

PR #3837, opened 2026-09-09
Protect official releases independently of deployment pins

rbuergi:
An official release with no current deployment pin was absent from the protection set and could even be selected by the optional unlock arm. Release support must not depend on which version today's deployments happen to pin.

Include clean release-tagged manifests in the protected union only for the repositories promoted by release.yml (memex-migration, mw-plugin-test, memex-portal-ai), matching release.yml's numeric major.minor.patch tags and retaining legacy v-prefixed tags. Unknown support-end mappings preserve protection. This does not authorize retiring releases or unlock anything; it adds no credential or permission requirements. Ordinary unpinned CI builds remain outside this classification. The policy now states the requested 30-day age-based cleanup for continuous artifacts explicitly.

Validation: the existing offline self-test and retention-record check pass. Removing the new protection deterministically fails five official-release assertions, including unlock prevention; restored implementation passes. Controls cover CI tags, duplicate release aliases, unknown support mappings, missing digests, cached third-party images and independently versioned helpers. The retention-record guard compares the protected repository set with release.yml; adding an unprotected publisher repository deterministically fails. A live report-only inventory across nine fleet repositories identified and drove the scope correction; no registry writes occurred. Documentation.Test builds with Release CIRun warnings-as-errors (0 warnings/errors); What's New integrity passes.

Related to #3438. This is the official-release protection prerequisite, not closure of retention: the independent cloud purge still needs interlocking with protection, its 7d/keep10 policy needs the 30-day age policy, and the actual purge-tool version must enforce age for untagged manifests. No live cleanup settings or registry deletions were changed in this work.

### DIFF
--- a/.github/scripts/lock-pinned-digests.py
+++ b/.github/scripts/lock-pinned-digests.py
@@ -543,6 +543,12 @@ def build_plan(axis1: list, axis2: list[OverlayScan]) -> Plan:
     return plan
 
 
+# These are the ACR/GHCR image repositories promoted by release.yml. Numeric tags
+# in cached third-party images or independently versioned helpers are not MW releases.
+# --check-retention-record checks this set against the publisher's explicit repo loops.
+OFFICIAL_RELEASE_REPOSITORIES = frozenset({"memex-migration", "mw-plugin-test", "memex-portal-ai"})
+
+
 def resolve_and_classify(plan: Plan, registry: Registry,
                          inventory: dict[str, list[Manifest]],
                          inventory_errors: dict[str, str]) -> None:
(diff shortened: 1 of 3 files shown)


rbuergi:
Addressed the review's classification-scope concern with a live report-only inventory across nine fleet repositories. The original broad numeric-tag rule would have included cached Grafana Loki and independently versioned helper images. The corrected rule is limited to the three repositories promoted by MeshWeaver's official release workflow; ordinary fleet-pin protection remains unchanged.

New controls reject the overbroad classification and reject a release-workflow repository addition absent from the protection set. Both negative controls failed as intended, the full offline self-test and retention-record checks pass, and the updated documentation builds with Release CIRun warnings-as-errors. No registry mutation, unlock, deletion, or permission change was performed.

PR #3869, opened 2026-09-09
Give every generated form field an accessible name (#3863)

### LINKED ISSUE #3863: Generated Edit form question titles are not associated accessible input labels
The native Edit-generated assessment form displays DisplayName questions as definition terms, followed by a definition containing a textbox with no accessible name. Reproduced on a disposable portal 3.0.0-ci.8195 with LearningJourneyContent properties annotated [UiControl<TextAreaControl>] and [DisplayName(...)].

The accessibility snapshot reports `term: 1. What topics ...`, `definition: textbox` instead of an associated textbox label. Playwright getByLabel(question) finds nothing; locating the adjacent definition textbox works. Explicit TextAreaControl.Label produces an aria-label on the Fluent host, so the generated-editor label relationship is the gap.

Ensure generated form labels associate with the editable input (including Fluent shadow-DOM controls) and add a browser accessibility regression using getByRole(textbox, {name: question}). Durable evidence: https://github.com/Systemorph/MeshWeaver.Plugins/blob/codex/edu-learning-journey/Edu/LearningJourneyGuide.md

rbuergi:
Fixes #3863 — **core half**. The observable fix needs the Plugins half as well; see below.

## The defect

The native `Edit`-generated form renders a question as the `<dt>` of a definition list and the input
in the `<dd>` beside it. The input had **no accessible name at all**: the accessibility snapshot read
`term: 1. What topics …` then a bare `definition: textbox`, and Playwright's
`getByRole('textbox', { name: question })` matched nothing.

Two independent causes, and both have to change:

1. **Here.** `EditorExtensions.MapToControl` builds the control with `label = null` **on purpose** —
   the caption is already painted by the surrounding `PropertySkin`, and setting it on the control
   too would render it twice. The consequence nobody accounted for is that the control was then left
   with nothing to name it: the control's own `Label` is what produces the `aria-label` on the Fluent
   host, which is why an explicit `TextAreaControl.Label` worked and a generated field did not.
2. **MeshWeaver.Plugins.** `PropertyView.razor` emitted `<label for="@(Skin?.Name)">` — a target that
   *cannot exist*, because a Fluent input keeps its real `<input>` inside a **shadow root** and
   `for`/`id` do not associate across that boundary.

`aria-label` is the only one of the two that survives the shadow boundary, and it needs no id
plumbing through `DispatchView` — so that is the association this uses.

## What this half changes

* `IFormControl` gains `AriaLabel` and `WithAriaLabel(object)`, **both with default implementations**
  (`null` / return-unchanged) so no out-of-repo implementer is obliged to write code before it can
  compile. `FormControlBase<TControl>` carries the real record copy, so every form control has an
  accessible name, exactly as `DataGridControl.WithAriaLabel` already does for grids.
* `DateTimeControl.AriaLabel` **moves up** to `FormControlBase` — it was the only control that had
  one. It is still readable and settable on `DateTimeControl` by inheritance; re-declaring it there
  would shadow the base property and give the record two backing fields.
* `EditorExtensions` sets `AriaLabel` on every generated control from `GetEditorLabel()` — the SAME
  single source the `PropertySkin` caption is read from, so the accessible name and the visible term
  can never drift apart (WCAG *label in name*). **`label` stays null**, so the caption is still not
  painted twice.
* `Doc/GUI/Editor` gains an "Accessible names" section recording why it is `aria-label` and not
  `<label for>`.

**Scope, stated narrowly on purpose** (review finding, second commit). The guarantee is over
`IFormControl` — text, multi-line text, number, date, checkbox, switch, select, combobox, listbox,
radio group. TWO generated surfaces are still unnamed, and both are recorded in `Doc/GUI/Editor` and
at the code site rather than left to be rediscovered from a symptom:

| Surface | Why not covered here |
|---|---|
| `[Markdown]` / `[UiControl<MarkdownEditorControl>]` (the `SpecialControls` branch) | `MarkdownEditorControl` is not an `IFormControl`; it is a composite editor with its own toolbar and view, so its accessible name is an `aria-labelledby` question about that composite, not an `aria-label` on one input. Guessing at it here would ship an unmeasured answer. |
| The click-to-edit path (`MapToToggleableControl`) | It renders its caption as a sibling `LabelControl`, which emits a `FluentLabel` with no `for` — the same association hole, on a different feature surface. |

## Control result — the test fails without the fix

`EditorTest.GeneratedFieldsCarryTheirQuestionAsAccessibleName` renders the exact shape from the issue
(`[DisplayName]` + `[UiControl<TextAreaControl>]`) and asserts both halves of the tension: the
control carries the question as `AriaLabel`, **and** still carries no visible `Label`.

Reverting `src/MeshWeaver.Layout/EditorExtensions.cs` and rebuilding fails it on the first assertion:

```
Expected value to be "1. What topics do you want to cover?" because the generated input must
expose the question as its accessible name (#3863), but found <null>.
```

With the fix: `Failed: 0, Passed: 1`.

## What this half does NOT do

Nothing observable. The rendered `aria-label` comes from the Blazor views in **MeshWeaver.Plugins**,
so the browser regression the issue asks for
(`getByRole('textbox', { name: question })`) can only go green once that half ships. It is open as
**Systemorph/MeshWeaver.Plugins#1578**, which also carries the DOM-level regression
(`GeneratedFieldAccessibleNameTest`) and the What's New entry. Per AGENTS.md the core half lands
first.

Pairs-with: none — nothing left the public surface. DateTimeControl.AriaLabel was not deleted, it moved UP to its base class FormControlBase, so every existing read, object-initializer write and `with` expression on DateTimeControl still compiles unchanged by inheritance — including in-mesh C#, which the compiler cannot see. The detector reports the move as a member removal because it compares declaration sites, not effective surfaces.

Local verification (Release, warnings-as-errors, one project per invocation): MeshWeaver.Layout,
MeshWeaver.Documentation, MeshWeaver.Mesh.Operations, MeshWeaver.Hosting.AspNetCore,
MeshWeaver.Layout.Test, MeshWeaver.Documentation.Test — all `0 Warning(s) 0 Error(s)`;
`DocumentationLinkIntegrityTest` green.

### DIFF
--- a/src/MeshWeaver.Documentation/Data/GUI/Editor.md
+++ b/src/MeshWeaver.Documentation/Data/GUI/Editor.md
@@ -250,6 +250,41 @@ Each property becomes a named area inside the `EditorControl`. The reactive over
 
 ---
 
+# Accessible names
+
+A generated field paints its caption **outside** the input — the `PropertySkin` renders it as the
+`<dt>` of a definition list and the control itself is created with **no `Label`**, so the caption is
+not drawn twice. That is deliberate, and it costs the input its accessible name unless something
+puts one back.
+
+So `MapToControl` sets **`AriaLabel`** on every generated control, read from the same
+`GetEditorLabel()` source the visible caption comes from — `[Display(Name = …)]`, then
+`[DisplayName(…)]`, then the property name word-split. The two can therefore never disagree, which
+is what WCAG's *label in name* asks for.
+
+```csharp
+public record Assessment
+{
+    [DisplayName("1. What topics do you want to cover?")]
+    [UiControl<TextAreaControl>]
+    public string Topics { get; init; } = null!;
+}
+// renders: <dt>1. What topics …</dt>
+//          <dd><fluent-text-area aria-label="1. What topics …"> …
+```
+
+🚨 **`aria-label`, not `<label for>` — and this is not a style preference.** A Fluent input is a web
+component that keeps its real `<input>` inside a **shadow root**, and `for`/`id` do not associate
+across that boundary. The generated form used to emit `<label for="topics">` pointing at an element
+that could not exist: the caption was visible, the textbox was unnamed, and
+`getByRole('textbox', { name: question })` matched nothing (MeshWeaver#3863). `aria-label` sits on
+the host element, so it survives the boundary and needs no id plumbing.
+
+Setting `Label` yourself still works and still renders a second, visible caption — use it for a
+control you compose by hand, not for a field the editor generates.
+
+---
+
 # See Also
 
 - [Property Attributes](../Attributes) — every supported attribute in detail
--- a/src/MeshWeaver.Documentation/Data/GUI/Editor.md
+++ b/src/MeshWeaver.Documentation/Data/GUI/Editor.md
@@ -257,10 +257,11 @@ A generated field paints its caption **outside** the input — the `PropertySkin
 not drawn twice. That is deliberate, and it costs the input its accessible name unless something
 puts one back.
 
-So `MapToControl` sets **`AriaLabel`** on every generated control, read from the same
-`GetEditorLabel()` source the visible caption comes from — `[Display(Name = …)]`, then
-`[DisplayName(…)]`, then the property name word-split. The two can therefore never disagree, which
-is what WCAG's *label in name* asks for.
+So `MapToControl` sets **`AriaLabel`** on every generated **input** control — every `IFormControl`:
+text, multi-line text, number, date, checkbox, switch, select, combobox, listbox, radio group. It is
+read from the same `GetEditorLabel()` source the visible caption comes from — `[Display(Name = …)]`,
+then `[DisplayName(…)]`, then the property name word-split. The two can therefore never disagree,
+which is what WCAG's *label in name* asks for.
 
 ```csharp
 public record Assessment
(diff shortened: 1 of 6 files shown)


rbuergi:
Addressed the review finding in `64a9fef` — by narrowing the claim rather than by widening the fix, and here is why.

**`SpecialControls` / the markdown editor.** `MarkdownEditorControl` is not an `IFormControl`: it is a composite editor with its own toolbar and its own Blazor view, so its accessible name is an `aria-labelledby` question about that composite, not an `aria-label` on a single input. I have not measured what its rendered markup needs, and shipping an unmeasured answer to an accessibility question is worse than shipping a stated gap.

**The click-to-edit path (`MapToToggleableControl`)** has the same hole for a third reason again: it renders its caption as a sibling `LabelControl`, which emits a `FluentLabel` carrying no `for`.

So the guarantee is now stated over `IFormControl`, and both gaps are recorded in `Doc/GUI/Editor` and in a comment at the `SpecialControls` branch itself — the two places a reader would actually be standing when the question comes up. Neither is left to be rediscovered from a symptom.

rbuergi:
The renderer revision in Systemorph/MeshWeaver.Plugins#1578 (0d26ca08e966439efeeb4c1ef08e265b731c4584) now carries the displayed PropertySkin label and its target through named-area wrappers using existing Blazor cascades. It needs no new core API. The exact final renderer passed the real assessment browser regression (question-based textbox lookup, label-click focus, reload), 62 renderer tests, and all 89 node-source compilation checks. This draft is now a candidate for closure as superseded once the replacement passes review and CI and lands; please keep it unmerged in the meantime.

rbuergi:
Closing as superseded by Systemorph/MeshWeaver.Plugins#1578, merged as 3f7686da284e39d460a5d89c0009aa9c0faeea51 after [Plugin Catalog CI 34427522111](https://github.com/Systemorph/MeshWeaver.Plugins/actions/runs/34427522111) completed successfully.

The landed implementation carries the displayed question label and its input target through named-area wrappers using existing Blazor cascades. It needs no new core API. The review's coverage finding was addressed with 68 renderer tests, alongside the real browser focus/name regression and all 89 node-source compilation checks.

Issue #3863 remains open for publication and production behavior verification. This closes the alternative draft, not that delivery requirement.

Issue #3880, opened 2026-09-10
Mounted node-repo source fingerprint ignores equal-length content edits


rbuergi:
Found while tracing the mounted-source feed requested in Systemorph/MeshWeaver.Plugins#1563.

`PackageSources.LocalDirectoryFetch` computes its snapshot ref from each relative filename and `FileInfo.Length`. It never hashes the file contents. Changing `ABC` to `DEF` in either a text file or a binary asset keeps the ref unchanged. For a package without a versioned manifest, `NodeRepoPackageSource.ListPackages` uses that snapshot ref as its advertised version.

Deterministic reproduction on core 2287decb702e5470b72ec0725a8acfae9905d3e0: a real temporary node repo, read through public `PackageSources.FromRepo` on a real mesh. Preserve byte count and mtime, change the bytes, list again. Both the Markdown and PNG cases fail the requirement that the version change; an unchanged second listing retains the version as expected.

The correction hashes the `RepoFile.Bytes` already captured for the returned snapshot. It retains the existing file enumeration, text/binary reading, source prefix, and declared-release-version precedence. No extra filesystem read, Store visibility change, or credential change is needed.

Acceptance: equal-length text and binary edits move the source ref; unchanged content and ignored dot-directory changes do not; existing mounted reconciliation and fetched-source seed behavior pass. Follow the normal PR/publication path and verify the affected mounted-source behavior before closing. This is independent of the pending choice about which mounted packages Store should advertise.

rbuergi:
Closing on a measurement, not on the existence of a merged PR.

## The fix

**#3881** — *Fingerprint mounted package sources from their content* — merged **2026-09-10T03:08:41Z** (merge commit `e29d40aa`). Core `main` merges through the merge queue, which does not fire `Closes #…` links, which is why this issue was still open.

It does hash **content**, not a substitute for it. `PackageSources.LocalDirectoryFetch` now folds `SHA256(file.Bytes)` — the payload already read to build the snapshot, so no second filesystem read — into the ref, replacing `FileInfo.Length`:

```csharp
var contentHash = Convert.ToHexString(SHA256.HashData(file.Bytes));
hash.Append(relative.Length).Append(':').Append(relative)
    .Append(':').Append(contentHash).Append(';');
```

That is exactly the reported case: an **equal-length** edit. A change that merely added mtime or a counter would not have covered it.

## Re-measured against `origin/main` today (2026-09-10)

[...]

PR #3881, opened 2026-09-10
Fingerprint mounted package sources from their content (#3880)

rbuergi:
Mounted node-repo snapshots hash filenames and byte counts, so replacing `ABC` with `DEF` leaves their advertised source version unchanged. Hash the already-read text or binary payload as well, keeping one disk read per file and the existing `local-` ref format. Declared package release versions retain their existing precedence.

Addresses #3880, found while investigating Systemorph/MeshWeaver.Plugins#1563. This does not enable the mounted Store feed or choose its visibility policy. Keep #3880 open until publication and affected mounted-source behavior are verified.

Validation:
- Deterministic negative control against unchanged core: both real-mesh Markdown and PNG cases fail when content changes with the same byte count and timestamp.
- With the correction: all four focused tests pass, including unchanged/dot-directory controls and the existing mounted reconciliation and fetched-source seed tests; zero skips.
- Core solution builds in Release with CIRun and warnings-as-errors: zero warnings, zero errors.
- All 368 documentation/repository guards pass, including the timeout-literal ratchet. The four focused regressions pass after adopting TestTimeouts.Quick. A user-facing Fix entry documents the corrected source fingerprint.

No public API removal, authentication change, filesystem visibility expansion, or second file read.

### DIFF
--- a/src/MeshWeaver.Documentation/Data/WhatsNew/2026-09-10-mounted-source-content-version.md
+++ b/src/MeshWeaver.Documentation/Data/WhatsNew/2026-09-10-mounted-source-content-version.md
@@ -1,0 +1,9 @@
+---
+Name: Mounted source fingerprints detect content edits
+Category: Fix
+Description: Mounted package source fingerprints now detect edits even when the file size stays the same.
+Icon: FolderSync
+Order: -20260910
+---
+
+Mounted package sources now include file contents in their fingerprints. Editing text or replacing an image with the same byte length changes the source fingerprint used as the version of packages without a declared release version. Unchanged files keep their fingerprint.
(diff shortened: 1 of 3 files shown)


rbuergi:
The first CI run failed the timeout-literal ratchet: the new test added four hard-coded 30-second waits. The checksum regressions themselves passed. Replaced those waits with the existing TestTimeouts.Quick budget; no production code or ratchet baseline changed.

Release/CIRun/warnings-as-errors build is clean. All 368 documentation/repository guards and the four focused source/reconciliation tests now pass, zero skips. The corrected assembly timestamp precedes the recorded test-run start. Pushed 646bbaded for a fresh normal CI run; no unchanged failing run was retried.

PR #3884, opened 2026-09-10
Retain registry errors when image-set verification fails (#3882)

rbuergi:
CD 8239 reported `memex-migration:main` missing when its registry read failed. The tag subsequently resolved to the exact promoted digest, with a last-update timestamp before the failed check. The verifier discarded the diagnostic needed to explain that read failure.

Keep Azure CLI error output and include its exit status for image-index and core/plugins pair-tag reads. Suppress only the CLI preview warning, and report an unverified set without asserting that an image was deleted or built from the wrong source. All failures still fail the gate; no retries, permissions, publication ordering or image checks change.

Validation: the executable shell regression failed on all six lost-error scenarios before the correction and now passes, including successful eleven-tag enumeration and single-architecture rejection. It verifies each failed read happens once and later checks still execute. Wired into CI's existing shell job. Shellcheck, actionlint, workflow timeout/YAML/secret guards and diff checks pass. Embedded documentation and dependencies build Release with CIRun and warnings-as-errors: zero warnings/errors. A read-only run of the corrected checker against actual set8239 verified all eleven identities/pointers successfully.

Related to #3882; deliberately does not close it or claim the original registry failure's cause is solved. The incident evidence and diagnostic contract are recorded in Doc/Architecture/ImageTagContract.

No What's New entry: internal CI diagnostics with no product behavior change.

### DIFF
--- a/.github/scripts/check-image-set.sh
+++ b/.github/scripts/check-image-set.sh
@@ -90,8 +90,8 @@ while [ $# -gt 0 ]; do
     --pointers)
       CHECK_POINTERS=1
       # Deliberately tolerant of an EMPTY value here: an unresolved version is reported below,
-      # AFTER the per-sha diagnostics, so a run whose image leg died still gets "<repo>:<sha> is
-      # MISSING" as its headline instead of a usage message about a flag.
+      # AFTER the per-sha diagnostics, so a run whose image leg died still gets "<repo>:<sha>
+      # could not be verified" as its headline instead of a usage message about a flag.
       if [ $# -ge 2 ]; then POINTER_VERSION="$2"; shift 2; else shift; fi
       ;;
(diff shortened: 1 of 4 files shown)


rbuergi:
Addressed both findings in automatic review5162268928 in commit3eb13a74d. The regression now asserts the exact eleven expected tag identities on both successful and failed-read paths. A negative control replacing the three main pointers with unrelated names still makes eleven distinct reads and is correctly refused. The older documentation paragraph and shell comment now describe the retained registry error and unverified-image diagnosis.

All three executable regressions pass; shellcheck and diff checks pass; the embedded documentation and dependencies rebuilt Release/CIRun/warnings-as-errors with zero warnings/errors (31.77s). No verifier behavior changed in this follow-up.

PR #3885, opened 2026-09-10
Seed past a publication whose seal vanishes mid-read (#3876)

rbuergi:
#3877 fixed the three `PublishedBundleCatalogue` seal readers and the four prebuilt HTTP routes: absence is decided AT the file open, an existing-but-unsealed source answers 503 + `Retry-After`, an absent source directory answers 404. That closes the 500 the incident reported.

The boot seeder is the **fourth reader of the same file**, and it kept the racing sequence — `File.Exists(sentinel)` followed by an unguarded `File.ReadAllLines(sentinel)` in `ShippedPrebuiltBundles.CompletePublishedBundlesOf`.

## Why it matters more there than on the routes

The publisher unseals before it republishes and re-seals last (several times an hour during a release), and retention unseals before it removes an identity. Where a route answered a 500, here the `FileNotFoundException` — or `DirectoryNotFoundException`, which is what a removed parent throws — left the per-source loop and reached `SeedBundles`' outer `Catch`, which abandons **the whole identity's adoption pass**. One source being replaced during a boot therefore made every *other* sealed source on that identity recompile too, at prebuilt-content scale.

Skipping an unsealed source is what this loop is *for*. It now reads through the same operation and skips only that source. The warning says which absence it saw — a publication directory present but unsealed (it died before the seal, or is being replaced right now) versus one that has been removed — so "not sealed yet" and "gone" are not the same line. A seal that is present but **unreadable** (locked, denied) still surfaces: only absence is classified.

Two implementations of one classification is how the seeder kept the race after the catalogue's was fixed, so there is now one: `ShippedPrebuiltBundles.ReadSealLines`, beside the sentinel's own name. `PublishedBundleCatalogue.ReadSealLines` delegates to it — `MeshWeaver.PluginCatalog` depends on `MeshWeaver.Hosting`, not the reverse, so the shared operation can only live on that side.

## Validation

Release builds with `-p:CIRun=true -warnaserror`, one project per invocation — `0 Warning(s), 0 Error(s)` on each: `MeshWeaver.Hosting`, `MeshWeaver.PluginCatalog`, `MeshWeaver.Documentation`, `Memex.Portal.Shared`, `MeshWeaver.Hosting.Test`, `Memex.Portal.Shared.Test`, `MeshWeaver.Documentation.Test`.

- New `SeedingSurvivesASealRemovedMidReadTest` — 5 passed. It removes the seal (and, in the second theory case, its whole directory) **at the read operation**, after an existence check could have observed it, on the first of two sealed sources; the surviving bundle proves the pass continued.
- **Negative control**: restoring the pre-fix sequence (`File.Exists` leading an unguarded read), **rebuilding**, and re-running gives **Failed: 2, Passed: 3** — the two race cases fail with `System.IO.FileNotFoundException` and `System.IO.DirectoryNotFoundException` thrown out of `CompletePublishedBundlesOf`. The three non-race cases (intact identity, present-but-unreadable seal, one-operation pin) pass in both states, so they are controls rather than duplicates. Fix restored and re-verified: 5 passed.
- `Memex.Portal.Shared.Test` seal/publication suites (incl. #3877's own `PublicationSealReadTest`) — 40 passed, so the delegation preserves the catalogue's behaviour.
- `DocumentationLinkIntegrityTest` + `WhatsNewEntryIntegrityTest` — 2 passed.

Refs #3876. The issue stays open through production verification, as its existing comment records; this does not change the publication-writer or retention work.


rbuergi:
**CI status on `db92e47f6`: one failure, filed as #3892 — not reachable from this diff.**

`MeshWeaver Build and Test` [34441078365](https://github.com/Systemorph/MeshWeaver/actions/runs/34441078365) — 1 failing test, `BakeEquivalenceTest.MeshDrivenAndCompilerDrivenBakes_ProduceTheSameArtifacts`. The two bakes' dependency records differ only by `!input`, an entry `CompiledDependencies.ContentKey` documents as absent by construction on an adopted-prebuilt / disk-cache-hit / hydration path.

Evidence that this diff cannot reach it:

- `main` is green at `eb18374ca`, this branch's exact merge base — so the branch is not behind anything relevant.
- The test passes locally in isolation on this branch (Release, `-p:CIRun=true`): 2 passed.
- The diff changes the seal read's behaviour **only** when it throws `FileNotFoundException`/`DirectoryNotFoundException` after an existence check had just succeeded — a concurrent unseal. No test performs one; with a readable seal the new path is byte-for-byte the old one.

Per the house rule I have not re-run the shard to see whether it comes back green — the details are in #3892 for whoever picks it up.

This PR is deliberately a **draft**: it is complete and verified, and draft is the repo's opt-out from `auto-arm`, which would otherwise land it on green. Mark it ready when you want it to merge.

PR #3056, opened 2026-09-02
fix: the report, the ack and the log about a delivery stop carrying it (#3044, #3045)

rbuergi:
## Consolidation first: six issues, three defects

#3044, #3045, #3046, #3047, #3048 and #3049 were auto-filed within 14 minutes from the same pod
(`memex-portal-deployment-85859db964-7sqbk`, `memex-cloud`, 02:54–03:08Z on 2026-09-02), all
`OutOfMemoryException` while serializing a message body. They are **three** defects:

| Defect | Survivor | Duplicates | State |
|---|---|---|---|
| The **failure report** carries, renders and re-renders the payload | **#3044** | #3049 (same site, second frame — it says so itself) | **fixed here** |
| The **acknowledgement** echoes the payload back across every grain boundary | **#3045** | — | **fixed here** |
| The **forward** directed pod-hub delivery to `import/GpdhJsyWQUuhnvNbsiY6HQ` | **#3046** | #3047, #3048 (the same failed send recorded by two more loggers — #3048 says so itself) | **stays open — producer work** |

#3047, #3048 and #3049 are closed as duplicates with the reasoning posted on each.

## Why "add another bound" is not the fix

Four bounds have landed in this family (#1890, #2897, #2885 twice) and the pattern each time was
*the guard exists, but not on this leg*. So the question here was never "add a guard" but **which of
the five legs covers this site, and if none does, why not**. The answer is that two of these are not
producer legs at all — no bound placed on the way out could ever have reached them — and the third
is a producer leg that **was** guarded and the delivery went through anyway.

And an `OutOfMemoryException` *during* serialization means the **allocation was the failure**, so a
bound that refuses the delivery afterwards is too late, and a `try/catch` around it (which the
incident's own log line already is) turns a lost message into a lost message *plus* a lost report.

## Defect 1 — the failure report (#3044, #3049)

```
fail: MeshWeaver.Messaging.MessageService[0]
      Failed to post DeliveryFailure message for RawJson (ID: …) — breaking error cascade
      System.OutOfMemoryException
```

Three things on that path copied the body:

1. **The NACK carried the payload it was reporting on.** That rule shipped twice as a *hand-applied
   call at one site each* — `RoutingGrain.PostFailure` (#1890) and
   `OrleansRoutingService.SendDeliveryFailure` (#2885). This was a **third** site nobody had told:
   `MessageService.ReportFailure`, which every one of the burst's occurrences ran through. There are
[...]

PR #3897, opened 2026-09-10
Type the existing snapshot by the node's own identity, never the proposal's (#3803)

### LINKED ISSUE #3803: NodeUpdatePipeline types the existing snapshot as the proposal's content type, so in-place retypes (NodeType changes) fail content recovery and break the validators' typed-content guarantee
## What is failing

An in-place NodeType change reaches `NodeUpdatePipeline` with existing content of one CLR type and proposed content of another. The pipeline tries to recover the existing snapshot using the proposal's content type, producing a `JsonException`. The reported update committed despite this recovery failure, leaving a concern about the typed-content contract seen by update validators.

## Investigation

- Check the existing-side typing step before `NodeValidationContext` is constructed.
- When the NodeType changes, establish how the existing snapshot should retain its original content type.
- Extend `UpdateValidatorSeesTypedExistingContentTest` with two distinct content types and a NodeType-changing proposal.
- Verify that validators receive the intended existing and proposed content and that recovery failures cannot silently bypass their comparisons.
- Review `UpdateValidatorsSeeTypedContent.md`, whose current explanation assumes an unchanged NodeType.

## Evidence handling

The original automated report included private application content in its exception excerpt. That content has been removed from this issue body; it is not needed to describe or reproduce the defect. The disclosure through automatic incident reporting requires a separate security investigation.

The underlying retyping defect remains unresolved. #3804 is a duplicate of #3803.

rbuergi:
Fixes #3803.

## The defect, established before the fix

`NodeUpdatePipeline` owes Update validators the existing node and the proposed one **typed alike**,
and it delivered that by deserialising the existing snapshot as the *proposed* content's CLR type.
The method states its own precondition — *"same node, same NodeType"* — and **an in-place NodeType
change is exactly where that is false**. Retyping a node is a supported operation (`patch` refuses
`nodeType` outright, so a full `update` is the sanctioned repair for a mistyped node —
`DanglingNodeTypeUpdateTest` exists to keep that route open). On such an update the proposal's type
describes what the node is moving **to**; it says nothing about the snapshot it is moving **from**.

Typing the snapshot by it is not recovery, it is **manufacture** — and `System.Text.Json` makes the
manufacture **silent**. The hub options set `UnmappedMemberHandling = Skip`, so the old node's bytes
deserialise *cleanly* into the new type whenever the new type's members are present or defaultable.
Two records sharing one member name are enough:

```text
stored     {"title":"Original","sequence":5}    (NodeType A, record A(Title, Sequence))
proposed   B("Retyped", "…")                    (NodeType B, record B(Title, Reason))
handed to validators as ExistingNode.Content:
           B("Original", null)                  ← a state the node was never in
```

Nothing throws, nothing logs, and a validator written the way the fleet writes them —
`ExistingNode.Content is TNew e && Node.Content is TNew p && …` — now *succeeds*, comparing the
proposal against a ghost with `Sequence` dropped and `Reason` defaulted. Where the shapes do not
line up the conversion throws instead, `As` returns null, the snapshot stays untyped, and the typed
comparison skips its own check and answers Valid: the #3056 silent pass again, this time with a
`JsonException` in the log. The original report saw the loud half; the silent half is worse.

## Cast trap-door, or routing mistake? Neither — a wrong type AUTHORITY

The read already uses the safe accessor (`ObjectAsExtensions.As`), so this is not the
`payload as MyType` trap-door. It is also not a routing/registration mistake to paper over: the
right hub is already running the validators, and moving the work elsewhere would not help. The
defect is that the pipeline chose **the wrong evidence for what the snapshot IS** — the proposal's
type instead of the node's own identity. That reading was decided by the machinery, not by argument:
`MeshNodeStreamExtensions.EnsureTypedContent` runs on **every** emission of the stream this snapshot
came from, and already offers it the exact recovery — `IMeshContentTypeRegistry.TryRecoverForNodeType`,
keyed on the node's **own** NodeType — plus a late re-type wait for a type that is not registered
yet. So a snapshot still untyped when the pipeline sees it is untyped because *nothing in the process
can type it*, and there is no third answer for the pipeline to give beyond saying so.

## The fix

- Proposal-typed recovery now applies **only while the NodeType is unchanged**.
- `RetypesTheNode` requires **both** sides to name a type: a proposal that omits `NodeType` is not
  changing it, and reading "absent" as "changed" would route ordinary partial updates down the
  no-recovery branch and reopen #3056 wholesale.
- On a retype the snapshot is left exactly as it arrived; when it is untyped the pipeline logs a
  **Warning** naming both NodeTypes and the node path (never content), saying validators will see
  that side untyped. Silence was the thing worth fixing, not the outcome.
- The retype still lands. The gate changes what validators are *shown*, never whether the update is
  accepted — a validator that wants to refuse a retype refuses it on the NodeType, which both sides
  always carry.

## Why two earlier attempts could not reproduce it

The monolith rig does not produce a degraded snapshot by itself: a create hands the store a LIVE CLR
instance and the same process reads that instance straight back, so the pipeline's typing step is
never reached and the defect cannot appear. `UpdateRetypeExistingContentTest` therefore **creates**
its probe node with content already in the degraded shape — a bare `JsonElement` with no `$type`,
which is literally what `ObjectPolymorphicConverter.Write` persists for a content type no hub
registered ("type not registered, serialize without type information"), and the state every later
read then reproduces forever. That is the prod state the whole typing step was written for.

## Control, both directions, on the same test code

| | result |
|---|---|
| fix reverted (`git checkout --` on the one src file, rebuilt) | **FAILED** — `AnInPlaceRetype_DoesNotHandValidatorsAnExistingContentOfTheProposedType`: *"Did not expect value to be of type RetypeAfterContent because an in-place NodeType change is precisely where the pipeline's own premise — 'same node, same NodeType' — is false … but it was."* (1 failed, 1 passed) |
| fix re-applied, rebuilt | **2 passed** |

Wider: `UpdateRetypeExistingContentTest` + `UpdateValidatorSeesTypedExistingContentTest` +
`DanglingNodeTypeUpdateTest` → 11 passed. `MeshWeaver.Documentation.Test` (link integrity + every
ratchet guard) → 368 passed. `dotnet build -c Release -warnaserror` clean (`0 Error(s)`) on
MeshWeaver.Hosting, MeshWeaver.Documentation and MeshWeaver.Graph.Test.

## Work products conserved

[Update Validators See Typed Content](src/MeshWeaver.Documentation/Data/Architecture/UpdateValidatorsSeeTypedContent.md)
gains the "Same NodeType is a precondition, not an aside" section (the worked manufacture, why
nothing else is available to the pipeline, and what the test's degraded-shape create is for), plus a
What's New entry (Category: Fix).

### DIFF
--- a/src/MeshWeaver.Documentation/Data/Architecture/UpdateValidatorsSeeTypedContent.md
+++ b/src/MeshWeaver.Documentation/Data/Architecture/UpdateValidatorsSeeTypedContent.md
@@ -64,21 +64,26 @@ proposal's CLR type is the type of the content the node is moving **to**; it say
 snapshot the node is moving **from**.
 
 **Typing the snapshot by it is not recovery, it is manufacture — and `System.Text.Json` makes the
-manufacture silent.** The hub options set `UnmappedMemberHandling = Skip`, so the old node's bytes
-deserialise *cleanly* into the new type whenever the new type's members are present or defaultable.
-Two records that merely share a member name are enough:
+manufacture silent.** The hub options set `UnmappedMemberHandling = Skip`, so the old content
+deserialises *cleanly* into the new type whenever the new type's members are present or defaultable.
+
+The sharpest case needs no degraded JSON at all, because a `$type` discriminator is a **package-local
+name**: `As` recovers a foreign runtime type exactly when `value.GetType().Name == type.Name`, and one
+customer repo ships `Currency` in four packages (see `IMeshContentTypeRegistry`). So two packages that
+each declare a `PackageContent` are enough, with both types registered and nothing degraded anywhere:
 
 ```text
-stored   {"title":"Original","sequence":5}     (NodeType A, content record A(Title, Sequence))
-proposed A(Title, Reason) → B(Title, Reason)   (NodeType B, content record B(Title, Reason))
-manufactured existing     B("Original", null)  ← a state the node was never in
+stored    RetypeFrom.PackageContent("Original", 5)     (NodeType A, registered, reads back typed)
+proposed  RetypeTo.PackageContent("Retyped", "…")      (NodeType B, registered)
+                     ↓ short names match, so As round-trips it
+manufactured existing RetypeTo.PackageContent("Original", null)   ← a state the node was never in
 ```
 
 Nothing throws, nothing logs, and the validator's `ExistingNode.Content is B e && Node.Content is B
 p && …` now *succeeds* — against a ghost, with `Sequence` dropped and `Reason` defaulted. Where the
-conversion instead throws, `As` returns null, the snapshot stays untyped and the typed comparison
-skips: the #3056 silent pass all over again, this time with a `JsonException` in the log. That is
-issue #3803; the first report saw the loud half, and the silent half is the worse one.
+conversion instead throws, or where the two names differ, `As` returns null and the snapshot is left
+as it was. That is issue #3803; the first report saw the loud half, and the silent half is the worse
+one.
 
 **So the pipeline gates the recovery on the NodeType being unchanged** (`RetypesTheNode` requires
 BOTH sides to name a type — a proposal that omits `NodeType` is not changing it, and reading
@@ -93,12 +98,21 @@ plus a late re-type wait for a content type that is not registered yet. A snapsh
 the time the update pipeline sees it is untyped because nothing in the process can type it. The
 honest outcome is to say so, not to invent an answer.
 
-`UpdateRetypeExistingContentTest` (MeshWeaver.Graph.Test) pins both halves: a retype must never hand
-a validator an existing content of the *proposed* type, and an update that keeps the NodeType must
-still get the #3056 recovery. Its probe node is created with content already in the degraded shape
-(a bare `JsonElement`, no `$type`) — the monolith rig does not produce that state by itself, because
-a create hands the store a live CLR instance and the same process reads it straight back, so the
-typing step is never reached at all.
+`UpdateRetypeExistingContentTest` (MeshWeaver.Graph.Test) pins all three halves: a retype must never
+hand a validator an existing content of the *proposed* type; an update that keeps the NodeType must
+still get the #3056 recovery; and the exact NodeType-keyed recovery must still run upstream, which
+is **measured** rather than asserted from a code read — the probe stores bytes carrying no `$type`
+under a NodeType that declares its content type, so a typed result can only have come from
+`TryRecoverForNodeType`.
+
+🚨 **Its fixture is the cross-package collision, not unreadable JSON, and that distinction is
+enforced.** An earlier revision seeded discriminator-less bytes under a NodeType that declared no
+content type; that node was never resolvable, so `check-untyped-content.sh` redded the shard at
+teardown — correctly, because *content nothing can read* is a different defect, and that gate has no
+allow-list by design. Seeding a live, REGISTERED value of a foreign same-short-named type reproduces
+this defect through the mechanism a running mesh actually produces, degrades nothing, and leaves the
+gate with nothing to report. If you are modelling "present but unreadable as `T`" anywhere, that is
+the shape to reach for.
 
 ## What this does not change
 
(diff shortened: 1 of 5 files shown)


rbuergi:
## Control, both directions — full rebuild, no `--no-build`

`NodeUpdatePipeline.cs` reverted to `origin/main` (`git checkout origin/main -- …`; `origin/main`
does not touch that file, so this is exactly the pre-fix source), test file kept, and `dotnet test`
left to do its own build both times.

**Reverted — 1 failed, 2 passed** (`Failed! - Failed: 1, Passed: 2, Skipped: 0, Total: 3`):

```
Failed MeshWeaver.Graph.Test.UpdateRetypeExistingContentTest.AnInPlaceRetype_DoesNotHandValidatorsAnExistingContentOfTheProposedType [2 s]
  Error Message:
   MeshWeaver.Reactive.Assertions.AssertionException : Did not expect value to be of type
   RetypeAfterContent because an in-place NodeType change is precisely where the pipeline's own
   premise — 'same node, same NodeType' — is false, so the PROPOSAL is not the type authority for
   the EXISTING snapshot. System.Text.Json ignores unmapped members, so the old node's bytes
   deserialise cleanly into the proposed type and the validator receives a well-formed instance of
   a state the node was never in, every member the old content did not carry silently defaulted,
   but it was.
  Stack Trace:
     at …ObjectAssertions`2.NotBeOfType[T](String because, Object[] becauseArgs)
     at …UpdateRetypeExistingContentTest.AnInPlaceRetype_…() in UpdateRetypeExistingContentTest.cs:line 149
```

**Re-applied — 3 passed** (`Passed! - Failed: 0, Passed: 3, Skipped: 0, Total: 3`).

## Does a retype now leave validators permissive? No — measured

New test in this PR: `ARetypeStillGetsTheExactRecovery_WhenTheExistingNodeTypeDeclaresItsContentType`.
Same discriminator-less bytes (`{"title":"Original","sequence":5}`), but stored under a NodeType that
DECLARES its content type (`WithContentType<RetypeBeforeContent>`), then retyped. The validator sees
**`RetypeBeforeContent`**. The bytes name no type at all, so the only route that can produce that is
`IMeshContentTypeRegistry.TryRecoverForNodeType` keyed on the node's OWN NodeType — which **measures**
that the exact recovery runs upstream, on the very stream `NodeUpdatePipeline` reads. That claim was
previously a source-read inference; it is now a passing assertion.

So a retype hands validators untyped content **only where nothing in the process can type it** — which
is #3056's own precondition, and the state in which the old code was equally broken: it manufactured a
ghost of the proposal's type (silent, wrong answers) or, when the conversion threw, left the snapshot
untyped anyway. The gate removes the wrong-answer branch from a window that already existed; it does
not open a new one.

## What FAILS, versus what merely warns

Nothing fails the update, deliberately — refusing a retype would close the sanctioned repair route for
a mistyped node (`patch` cannot write `nodeType`), which `DanglingNodeTypeUpdateTest` exists to keep
open. A validator that wants to refuse a retype refuses it on the **NodeType**, which both sides always
carry.

But the untyped state is **not** signal-free, and that is the difference from #3056. The framework
already reports it loudly at the READ, before the pipeline runs — visible in the control run's own log:

```
[Warning] [MeshWeaver.Hosting.MeshNodeStreamCache] MeshNodeStreamCache.GetStream: Content for
TestData/rt4e0cd418 stayed an untyped JsonElement after deserialization (TypeRegistry lacks the
$type discriminator) — downstream 'Content is X'/'as X' consumers will fail …
Exception: MeshWeaver.Mesh.MeshNodeContentDegradedException: … the $type discriminator resolved to
no registered CLR type on the reading hub and the mesh-wide content-type registry did not recover it.
```

`stayed an untyped JsonElement` is the exact string AGENTS.md names as a pre-prod sweep FAILURE, and
`MeshNodeContentDegradedException` is a typed diagnosis. #3056 had none of this — its silent pass came
from an accidental registration disappearing with nothing saying so. On top of that, the pipeline now
adds a retype-specific Warning naming both NodeTypes and the path.

**One thing I did not do, and am flagging rather than burying:** turning "validators could not see typed
content on a retype" into a hard REFUSAL would be a new acceptance policy on a path that has always been
permitted. #3803's earlier comment flagged exactly that as awaiting a maintainer decision. I left
acceptance behaviour unchanged; if the refusal is wanted, it is a separate, deliberate call.

Issue #1601, opened 2026-08-14
A lesson's example cell gets NO C# completions or diagnostics — Monaco falls back to word-based (alphabetical) suggestions


rbuergi:
## Symptom the user reports

*"Still dull ordering of autocompletes alphabetically."* Ranking is not the cause — **our language
service contributes nothing at all** in a lesson's example cell, so Monaco falls back to its built-in
`wordBasedSuggestions`, which scrape identifiers out of the document text and are ordered
alphabetically. #1499's rank ordering is correct; it ranks a list that never arrives.

## Measurement (disposable mesh, portal `179d66bb` — the build memex runs)

Learner's own copy of a lesson, cell rendered as an inline Monaco editor (so `BuildCellEditor` DID
wire `WithLanguageServer`):

- typing `e2eProbe.` → **no suggest widget at all**, four attempts, 20s apart;
- `Ctrl+Space` → `["a","AG","All","Alpina","already","ATT","Attritional","book","by","c","C1","C2"]`
  — words scraped from the cell's own comments and string literals, alphabetical;
- `.squiggly-error` / `.squiggly-warning` count: **0**, so `CheckSpeculative` is silent too;
- no browser console error (the JS provider logs one when the interop call throws — it did not).

Both callbacks silent together is the signature: either they were never wired, or the service answers
empty/late for this node.

## Two candidates, not yet discriminated

1. **Circuit-side DI.** `CodeEditorView.razor` resolves `IMeshLanguageService` from the component's
   `Services`; the registration is `.WithServices(...)` on the mesh hub
   (`GraphConfigurationExtensions`). If that instance is not reachable from the circuit's provider,
   the `else` branch nulls `diagnosticsCallback` **and** `codeCompletionCallback` — exactly what was
   measured.
2. **The 5s cap vs. node resolution.** `MonacoEditorView.GetCodeCompletions` bounds the call at
   **5s**, while `MeshNodeLanguageService.ResolveNode` alone allows **15s**
   (`hub.GetMeshNode(path, 15s)`). A cold or cross-partition owner lookup that takes >5s yields an
   empty list every time, silently. An unresolvable owner also returns empty by design
   (`CompletionEnvironment.Unknown` — "no suggestions beats wrong ones").

Server-side is known good in general: `lsp_diagnostics_for_node` over MCP answers for the same
subtree on memex.

## Repro

1. Install a course as a learner, open a lesson with an example cell (the code renders as an editor).
[...]

PR #3913, opened 2026-09-10
Give the click-to-edit form's inputs an accessible name (#3863)

### LINKED ISSUE #3863: Generated Edit form question titles are not associated accessible input labels

rbuergi:
Fixes #3863 — **core half**. Pairs with a **MeshWeaver.Plugins** pull request opened alongside this one, which carries the renderer and the markup control (linked in a comment below).

## Re-measured first: the reported surface is fixed, the OTHER generated form is not

The instance in the issue — the `Edit` macro / `EditorControl` / `PropertySkin` path, questions as
`<dt>` of a definition list — was fixed by MeshWeaver.Plugins#1578 (merged 2026-09-10 02:35Z). Its
`PropertyView` now mints a per-instance id, points `<label for>` at it, and cascades the question
down to the input as an `aria-label`. Measured on `origin/main` of that repo: nine input kinds
covered by `PropertyViewLabelTest`.

**The node editor's OTHER generated form was left exactly as #3863 described it.**
`EditorExtensions.MapToToggleableControl` — what `EditLayoutArea.BuildPropertyForm` builds for
`Edit`/`Overview` on a mesh node — paints the caption as a **sibling** `LabelControl`:

```csharp
return Controls.Stack
    .WithView(Controls.Label(displayName).WithStyle("font-weight: 600; …"))
    .WithView((h, _) => editStateStream.Select(isEditing => isEditing && isEditable
        ? BuildEditControl(h, property, dataId, editStateId, isToggleable, boundDataContext)
        : BuildReadonlyControl(…)));
```

There is no `PropertySkin`, so the merged cascade never reaches it; there is no `<label for>`,
because a `FluentLabel` emits none and there is no id to point one at; and `BuildEditControl` created
the input with no `Label` (the caption is already painted) and nothing else. So every field of the
node Edit form renders as a bare `textbox` with **no accessible name at all** — the same defect, on
the more central form. The Plugins suite already records the shape:
`StandaloneInput_DoesNotAcquireAnEmptyAccessibleName` asserts an input outside `PropertyView` carries
no name.

`aria-label` is the only channel that reaches it: it needs no id plumbing and, unlike `<label for>`,
it survives the Fluent web component's shadow root.

## What this half changes

* **`IFormControl` gains `AriaLabel` and `WithAriaLabel(object)`, both with default implementations**
  (`null` / return-unchanged) so no out-of-repo implementer is obliged to write code before it can
  compile. `FormControlBase<TControl>` carries the real record copy, so every form control can hold
  an accessible name — exactly as `DataGridControl.WithAriaLabel` already does for grids.
* **`DateTimeControl.AriaLabel` moves UP to `FormControlBase`** — it was the only control that had
  one. Still readable and settable on `DateTimeControl` by inheritance; re-declaring it there would
  shadow the base property and give the record two backing fields.
* **`MapToToggleableControl` → `BuildEditControl` names its input** from `GetToggleableDisplayName`,
  the SAME single source the sibling caption is read from, so the accessible name and the visible
  caption cannot drift (WCAG *label in name*). The `[ContentItem]` field+Browse row is a composite,
  so its text field is named where the row is built rather than skipped silently.
* **`RenderControl` names the skinned path too**, from `GetEditorLabel()` — belt and braces with the
  renderer cascade, and the channel that reaches views the cascade does not (the mesh-node picker).
  A control that carries its own visible `Label` is left alone: the Fluent host paints an associated
  label of its own, and an `aria-label` would only shadow it.
* **`ui.numericInput`** replaces the hard-coded English `aria-label="Numeric input"` fallback the
  number field emits (Plugins side). An `aria-label` is a user-visible string.
* **`Doc/GUI/Editor`** gains an "Accessible names" section: the table of which form names its input
  how, why it is `aria-label` and not `<label for>`, and the one surface still uncovered.

**Scope, stated narrowly on purpose.** The guarantee is over `IFormControl`. A `[Markdown]` /
`[UiControl<MarkdownEditorControl>]` property is NOT covered and is recorded as such in
`Doc/GUI/Editor` and at the code site: `MarkdownEditorControl` is not an `IFormControl`, it is a
composite editor with its own toolbar and view, so its accessible name is an `aria-labelledby`
question about that composite, not an `aria-label` on one input. Guessing at it here would ship an
unmeasured answer.

## Control — both directions, and it fails without the fix

Two view-model tests drive the REAL generators through a real layout area:

| test | what it drives |
|---|---|
| `EditorTest.GeneratedFieldsCarryTheirQuestionAsAccessibleName` | `hub.Edit(new AssessmentForm())` — the exact `[DisplayName]` + `[UiControl<TextAreaControl>]` shape from the issue |
| `MapToToggleableControlTest.ClickToEditInput_IsNamedByItsCaption` | the click-to-edit form: click into edit mode, then compare the input's name with the sibling caption's own text |

Reverting `src/MeshWeaver.Layout/EditorExtensions.cs` to `main` and rebuilding fails both:

```
Failed MeshWeaver.Layout.Test.EditorTest.GeneratedFieldsCarryTheirQuestionAsAccessibleName
  Expected value to be "1. What topics do you want to cover?" because the generated input must
  expose the question as its accessible name (#3863), but found <null>.
Failed MeshWeaver.Layout.Test.MapToToggleableControlTest.ClickToEditInput_IsNamedByItsCaption
  Expected value to be "Title" because the click-to-edit input is named only by the caption beside
  it, which carries no <label for> to associate with (#3863), but found <null>.
Failed!  - Failed: 2, Passed: 0
```

With the fix: `Passed! - Failed: 0, Passed: 476` over the whole `MeshWeaver.Layout.Test` project.

🚨 **These assert the VIEW MODEL, which is all core can see — core ships no renderer.** They are the
INPUT to the accessibility guarantee, not the guarantee. The markup control is in the Plugins half:
a real `HtmlRenderer` over the real dispatch chain, asserting the `aria-label` the Fluent host
actually emits, in both directions. Both tests say so in their own doc comments so neither is ever
read as more than it is.

## Ordering

Nothing here is observable on its own — the rendered `aria-label` comes from the Blazor views in
MeshWeaver.Plugins. Per AGENTS.md the core half lands FIRST, and the Plugins half is open alongside
it with the browser-level regression the issue asks for.

Local verification (Release, warnings-as-errors, one project per invocation): MeshWeaver.Layout,
MeshWeaver.Messaging.Hub, MeshWeaver.Documentation, MeshWeaver.Layout.Test,
MeshWeaver.Documentation.Test, MeshWeaver.Messaging.Hub.Test — all `0 Warning(s) 0 Error(s)`;
`DocumentationLinkIntegrityTest` and `LocalizationTest` (58) green.

Pairs-with: none — nothing left the public surface. DateTimeControl.AriaLabel was not deleted, it moved UP to its base class FormControlBase, so every existing read, object-initializer write and `with` expression on DateTimeControl still compiles unchanged by inheritance — including in-mesh C#, which the compiler cannot see. The detector reports the move as a member removal because it compares declaration sites, not effective surfaces.

Implementers: IFormControl.AriaLabel, IFormControl.WithAriaLabel — both ship with default implementations (`=> null` and `=> this`), so no implementer is obliged to write code. Checked in this repo: FormControlBase<TControl> is the only implementer of IFormControl and it now carries the real record copy; searched MeshWeaver.Plugins for `: IFormControl` and `IFormControl,` and found no other implementer, only consumers (FormComponentBase<,,> binds the interface, never implements it).

Mirror-sync: npm run sync:i18n -- --ref <this PR's merge sha> — this adds one key, `ui.numericInput`, so the Plugins react/RN catalog mirror goes stale on merge and nothing reds to say so. Handed over on the paired MeshWeaver.Plugins pull request opened alongside this one, which consumes the key.

### DIFF
--- a/src/MeshWeaver.Documentation/Data/GUI/Editor.md
+++ b/src/MeshWeaver.Documentation/Data/GUI/Editor.md
@@ -250,6 +250,56 @@ Each property becomes a named area inside the `EditorControl`. The reactive over
 
 ---
 
+# Accessible names
+
+A generated field paints its caption **outside** the input, and both generated forms do it
+differently:
+
+| Form | Where the caption is painted | What names the input |
+|---|---|---|
+| `Edit<T>()` / `EditorControl` | the `PropertySkin`'s `<dt>` | a `<label for>` the renderer targets at the input's id, **plus** `aria-label` |
+| The node editor's click-to-edit form (`MapToToggleableControl`) | a **sibling** `LabelControl` above the field | `aria-label` only — a `FluentLabel` carries no `for`, and there is no id to point it at |
+
+In both the control itself is created with **no `Label`**, so the caption is not drawn twice. That is
+deliberate, and it costs the input its accessible name unless something puts one back.
+
+So the generators set **`AriaLabel`** on every generated **input** — every `IFormControl`: text,
+multi-line text, number, date, checkbox, switch, select, combobox, listbox, radio group, mesh-node
+picker. It is read from the same source the visible caption comes from — `[Display(Name = …)]`, then
+`[DisplayName(…)]`, then the property name word-split — so the two can never disagree, which is what
+WCAG's *label in name* asks for.
+
+```csharp
+public record Assessment
+{
+    [DisplayName("1. What topics do you want to cover?")]
+    [UiControl<TextAreaControl>]
+    public string Topics { get; init; } = null!;
+}
+// renders: <dt><label for="property-…">1. What topics …</label></dt>
+//          <dd><fluent-text-area id="property-…" aria-label="1. What topics …"> …
+```
+
+🚨 **`aria-label` is not decoration on top of `<label for>`.** A Fluent input is a web component that
+keeps its real `<input>` inside a **shadow root**, and the generated form used to emit
+`<label for="topics">` pointing at an element that could not exist: the caption was visible, the
+textbox was unnamed, and `getByRole('textbox', { name: question })` matched nothing
+(MeshWeaver#3863). `aria-label` sits on the host element, needs no id plumbing, and is the only one
+of the two the click-to-edit form can use at all.
+
+Setting `Label` yourself still works and still renders a second, visible caption — use it for a
+control you compose by hand, not for a field the editor generates. A control that carries its own
+`Label` is left alone: the Fluent host paints its own associated label, and an `aria-label` would
+only shadow it.
+
+**One generated surface is NOT covered**, and it is named here rather than left to be rediscovered
+from a symptom: a `[Markdown]` / `[UiControl<MarkdownEditorControl>]` property. `MarkdownEditorControl`
+is not an `IFormControl` — it is a composite editor with its own toolbar and its own view, so its
+accessible name is an `aria-labelledby` question about that composite, not an `aria-label` on one
+input.
+
+---
+
 # See Also
 
 - [Property Attributes](../Attributes) — every supported attribute in detail
--- a/src/MeshWeaver.Documentation/Data/GUI/Editor.md
+++ b/src/MeshWeaver.Documentation/Data/GUI/Editor.md
@@ -265,9 +265,21 @@ deliberate, and it costs the input its accessible name unless something puts one
 
 So the generators set **`AriaLabel`** on every generated **input** — every `IFormControl`: text,
 multi-line text, number, date, checkbox, switch, select, combobox, listbox, radio group, mesh-node
-picker. It is read from the same source the visible caption comes from — `[Display(Name = …)]`, then
-`[DisplayName(…)]`, then the property name word-split — so the two can never disagree, which is what
-WCAG's *label in name* asks for.
+picker.
+
+**Each form reads it from the same helper its own caption comes from**, so within a form the
+accessible name and the visible term can never disagree — which is what WCAG's *label in name* asks
+for. The two helpers do not resolve the caption the same way, and that predates this:
+
+| Form | caption helper | resolution order |
+|---|---|---|
+| `Edit<T>()` | `GetEditorLabel` | `[Display(Name = …)]` → `[DisplayName(…)]` → the property name word-split |
+| click-to-edit | `GetToggleableDisplayName` | `[Display(Name = …)]` → the viewer-localized `[Description]` / `[Translation]` → the property name word-split |
+
+🚨 So `[DisplayName("…")]` names an `Edit<T>()` field and does **not** name a click-to-edit one, which
+falls through to the wordified property name. That divergence is stated here rather than smoothed
+over: it is a question about the visible CAPTION, not about the accessible name, and changing it
+would re-word every click-to-edit caption in the portal.
 
 ```csharp
 public record Assessment
@@ -276,8 +288,8 @@ public record Assessment
     [UiControl<TextAreaControl>]
     public string Topics { get; init; } = null!;
 }
-// renders: <dt><label for="property-…">1. What topics …</label></dt>
-//          <dd><fluent-text-area id="property-…" aria-label="1. What topics …"> …
+// Edit<T>() renders: <dt><label for="property-…">1. What topics …</label></dt>
+//                    <dd><fluent-text-area id="property-…" aria-label="1. What topics …"> …
 ```
 
 🚨 **`aria-label` is not decoration on top of `<label for>`.** A Fluent input is a web component that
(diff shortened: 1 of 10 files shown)


rbuergi:
Plugins half: Systemorph/MeshWeaver.Plugins#1601 (draft — that repo resolves the platform to the newest SEALED set of core's main-cd, so it cannot compile until this merges and a set seals). It carries the renderer, the markup control asserted through a real HtmlRenderer in both directions, the What's New entry, and the `ui.numericInput` consumer.

Mirror sync handover for the one added key: `npm run sync:i18n -- --ref <this PR's merge sha>`, to be run on #1601 once this lands.

rbuergi:
## Copilot review — three findings taken, one falsified by measurement

### 1. `DateTimeControl.AriaLabel` moving to the base — **measured, no break** 🔬

> assemblies compiled against the previous API call `DateTimeControl.get_AriaLabel`/`set_AriaLabel`, and those accessors no longer exist on `DateTimeControl`, causing `MissingMethodException`

This is the right question to ask in this repo — MeshWeaver#2370 was exactly a break of that shape —
but it does not hold for a member moving UP its own inheritance chain inside the same assembly. A
`MemberRef` naming `DateTimeControl::get_AriaLabel` is resolved by walking the type **and its base
classes** (ECMA-335 §II.22.25), so an inherited accessor satisfies it.

I did not want to settle that by argument, so I measured it. A probe assembly compiled against the
**pre-change** `MeshWeaver.Layout` — where `DateTimeControl` declares the property — exercising a
read, an object-initializer write and a record `with`:

```csharp
var withInitializer = new DateTimeControl(DateTime.UnixEpoch) { AriaLabel = "set through an object initializer" };
var copied = withInitializer with { AriaLabel = "set through a record `with`" };
Console.WriteLine(copied.AriaLabel);
Console.WriteLine(typeof(DateTimeControl).GetProperty("AriaLabel")!.DeclaringType);
```

Run against the OLD assembly (the control):

```
initializer  -> set through an object initializer
with-copy    -> set through a record `with`
read         -> set through a record `with`
declared on  -> MeshWeaver.Layout.DateTimeControl
OK: no MissingMethodException
```

Then **the identical, unrecompiled binary** with the NEW `MeshWeaver.Layout.dll` swapped in:

```
initializer  -> set through an object initializer
with-copy    -> set through a record `with`
read         -> set through a record `with`
declared on  -> MeshWeaver.Layout.FormControlBase`1[MeshWeaver.Layout.DateTimeControl]
OK: no MissingMethodException
```

Same behaviour; only reflection notices, and it reports the new declaring type. `Public surface
(binary compatibility)` agrees (green on this PR).

The suggested remedy would also cost more than it bought: `DateTimeControl` is a `record`, so a
forwarding accessor has to be `new` (the base property is not virtual), and a shadowing property on a
record gives it a second backing field — breaking `with`, `Equals` and `ToString` for that member.
That is why the code comment at the old declaration site says so explicitly rather than just deleting
the lines.

### 2 + 3. `GetToggleableDisplayName` never reads `[DisplayName]` — **correct, docs were wrong** ✅

Taken, and it is the better catch of the four. Both pages said the accessible name comes from
`[Display(Name)]` then `[DisplayName]`; that is `GetEditorLabel`, which only `Edit<T>()` uses. The
click-to-edit form reads `[Display(Name)]` → the viewer-localized `[Description]`/`[Translation]` →
the wordified property name, and never looks at `[DisplayName]`.

`Doc/GUI/Editor` now tables the two resolutions side by side and states plainly that
`[DisplayName("…")]` names an `Edit<T>()` field and does **not** name a click-to-edit one. I did NOT
change `GetToggleableDisplayName` to match: that is a question about the visible CAPTION, not about
the accessible name, and re-wording every click-to-edit caption in the portal is not something this
PR should do silently. The invariant this change rests on is unaffected — *within* a form the name
and the caption come from the same helper, so they cannot drift.

### 4. The What's New entry overstated the coverage — **taken** ✅

A `[Markdown]` property still renders an unnamed composite editor. The entry is narrowed to the
controls actually covered and names that exception; `Doc/GUI/Editor` already carried it.

Fixed in 048e09e1a.

PR #3914, opened 2026-09-10
fix: the @-reference drill-down runs behind the caller's identity (#3890)

### LINKED ISSUE #3890: Autocomplete runs with useSecurityFilter:false and no caller identity — the @-reference drill-down names nodes (and their titles) in partitions the caller cannot read
## What is failing

`@`-reference **autocomplete runs with no access filter and no caller identity**, so it names nodes
— label, path, node type — in partitions the caller cannot read. `get` and `search` on the very
same paths refuse, as they should.

Measured on `memex.systemorph.com`, 2026-09-10:

```
autocomplete '@/Helvetia/'
→ Helvetia (Crm/Client), Helvetia/Engagement (Crm/Opportunity),
  Helvetia/Prospect-1, Helvetia/Workshop-20260630, Helvetia/CommercialIndication-3

get    @Helvetia            → Not found
get    @Helvetia/Engagement → Not found
search 'namespace:Helvetia scope:subtree' → 0
```

Same for `BinaryClickerV2` (names its `Space` and its `NodeType`). The negative control is
`PearlTechnology`, a partition this identity *can* read: `autocomplete`, `get` and `search` all
answer, so the instrument is not simply broken.

Labels are business content, not just ids — the `@/PearlTechnology/` drill-down returns strings like
*"Pricing Comparison (Internal)"* and *"E-Mail-Entwurf an Thomas"*. For a partition the caller is
denied, that is the document titles of someone else's workspace.

## Where it comes from

The whole chain, with nothing filtering at any hop:

```
ChatCompletionOrchestrator.ProducePartitionDrillDown          (src/MeshWeaver.Hosting/Completion)
  → IMeshService.Autocomplete
  → MeshService.Autocomplete            → _query.Autocomplete(...)   ← does NOT call StampIdentity,
                                                                        unlike MeshService.Query
  → MeshQuery.Autocomplete              → per-provider Autocomplete
  → StorageAdapterMeshQueryProvider.Autocomplete
        UserId = null,                                         ← identity dropped
        RunQueryNodes(queryRequest, options, useSecurityFilter: false)
```

`CollectMatched` is where the RLS validator lives:

```csharp
return useSecurityFilter
    ? ValidateRead(meshNode, userId).Where(valid => valid).Select(_ => (object)meshNode)
    : Observable.Return<object>(meshNode);
```

`IMeshQueryProvider.Query` passes `true`; `IMeshQueryCore.Query` passes `false` **because
`SecurityService` consumes it and re-entering would be a DI cycle** — that is the documented reason
the raw surface exists. `Autocomplete` passes `false` too, with no such reason recorded and no
compensating filter downstream. It reads like identity was never plumbed through this path rather
than a decision.

## Neighbourhood

Same family as #1250 (*the change-feed supplement bypasses RLS, so a subscriber can transiently
receive a node it cannot read*), #603 (*a node stays readable but vanishes from all listings*) and
#2899. Those were each closed on one surface at a time; this is the surface that was not swept.

## Two things to settle before fixing

1. **Is it deliberate for the partition LIST?** `@/` enumerating partition NAMES off
   `Admin/Partition` may be intended (a name is not content). The **drill-down** is the part that
   returns other people's node titles, and that is hard to defend either way.
2. **Cost.** `Autocomplete` deliberately over-fetches (`limit*50`, min 1000) and scores in memory;
   filtering per row adds a `ValidateRead` per candidate on a latency-sensitive path. The right
   shape is probably to narrow at the source (restrict the base paths to readable partitions before
   the walk) rather than to validate the over-fetched set.

## The other half — it is also a diagnostic people now rely on

🚨 Note before changing it: **this bypass is currently the only way to tell a DENIED read from a
DELETED node**, and #3886 documents it as exactly that positive control (`get` answers `Not found`
for both). If autocomplete starts filtering, that control disappears and #3886's guidance needs a
replacement in the same change — otherwise we re-open the door that let #1391 be closed on a false
negative and re-filed four weeks later as #3883.

## How it was found

Chasing #3883, establishing whether `BinaryClickerV2/BinaryToggle` still existed. Related: #3883,
#3886, #3888, #1250, #603, #2899.

rbuergi:
Fixes #3890.

## What leaked, and to whom

`StorageAdapterMeshQueryProvider.Autocomplete` built its request with a hard-coded `UserId = null`
and ran it through `RunQueryNodes(…, useSecurityFilter: false)`. Same storage, same query shape and
the same `ValidateRead` chain as the secured read — only those two inputs differed, which is what
made the `@`-reference drill-down a witness to content the caller is refused.

**The projection that crossed the wire** is `QueryResult { Path, Name, NodeType, Icon, Score,
ProviderName }` — path, **display name**, node type, icon. Never `Content`, and never anything the
caller could then open. The `Name` is the part that makes this a disclosure rather than an
enumeration: it is a document title.

**Reachable by any authenticated caller, with no admin gate anywhere on the path.** The MCP
`autocomplete` tool is gated on `RequireAuthenticatedUser()` and nothing else; the chat composer's
`@` box reaches the same `ChatCompletionOrchestrator.ProducePartitionDrillDown`, which dispatches on
the *text shape* of the reference (`@/<Partition>/…`) with no role or permission check of its own.

**Correcting the issue's framing on one point.** The issue reads the whole chain as unfiltered. The
production Postgres provider's own autocomplete *does* carry a user id into its SQL. It reached
production anyway because `MeshQuery.SelectMatchingProviders` hands **every** registered
`IMeshQueryProvider` every autocomplete and unions the snapshots by path, and
`AddPartitionedCoreAndWrapperServices` registers this pedestrian provider **alongside** a native
backend rather than instead of it ("Native query backends (Postgres, Cosmos) register their OWN
IMeshQueryProvider alongside this one"). The native provider's filtered rows were merged with this
provider's unfiltered ones — and `Autocomplete` is deliberately exempt from the `#20` defer-to-native
gate, so the pedestrian walk runs even on a partitioned Postgres mesh.

## Was the bypass load-bearing for autocomplete's function?

**No.** The filter is applied to rows that were already materialised, inside the existing
`CollectMatched` fold — the over-fetch is unchanged and **nothing extra is fetched**. Two consequences
worth stating:

- Because filtering happens before scoring, rows the caller may not read no longer consume slots in
  the post-scoring `.Take(limit)`. A caller now sees **more** of their own suggestions, not fewer.
- The per-row cost is one `RlsNodeValidator.Validate`, which short-circuits **synchronously** for
  System, for `MainNode == userId`, and for the caller's own partition — the common drill-down. The
  drill-down shape the issue measured over-fetches `max(limit*5, 100)`; the `max(limit*50, 1000)`
  applies only to the empty-basePath shape.
- Measured locally: the same three tests run in 774 ms with the filter and 869 ms without, i.e. no
  measurable cost at suite scale. **Production latency was not measured** and this PR does not claim
  it was.

The issue suggests narrowing at the source (restrict the base paths to readable partitions before
the walk) instead. That was rejected on correctness, not effort: grants exist **below** a partition
root, so a caller denied `Helvetia` may still hold Read on one node inside it, and a partition-level
narrowing would hide it. There is also no "partitions this user may read" oracle to narrow with —
the native SQL providers express it as an inline predicate, never as a list.

## The coupling: a documented diagnostic died with the bypass, and is replaced here

`NodeTypeCompilation.md` (landed in #3886) told readers to separate a DENIED partition from a
DELETED node with three reads, and step 2 was exactly this bypass. It flagged that whichever change
landed had to replace it in the same diff. **AGENTS.md carried the same guidance** and is updated too.

The replacement does not reinstate a caller-run oracle, and says so:

1. `get @Admin/Partition/<Ns>` still says whether the **partition** was torn down — it never answers
   about a node.
2. **The pod's `nodetype_bake` payload** is the one instrument with the full denominator: the boot
   sweep enumerates `nodeType:NodeType` mesh-wide under `ImpersonateAsSystem`, and it already
   carries the distinction — `Regressions`/`Unevaluated`/`ContentBroken` name a type that EXISTS and
   did not build, while `Retired`/`Removed` is the sweep reporting the node is gone
   (`PreWarmStatus.Removed`, established by a listing that came back). Its arming caveat is already
   on the page and is cross-referenced.
3. Otherwise the answer is the partition **owner's** — ask, or elevate (break-glass).

And the rule that actually prevents the #1391 → #3883 recurrence: **a `Not found` from outside your
readable denominator has told you nothing and cannot close an issue.** The sweep's question is "is
anything broken", not "does this node exist", and (2) answers that over the whole mesh regardless of
who may read what — so naming the denominator makes denied-versus-deleted stop deciding anything.

## Control, in both directions

`test/MeshWeaver.Hosting.Test/AutocompleteHonoursReadAccessTest.cs`, on a mesh built from
`ConfigureMeshBase` (no `PublicAdminAccess` seed, so RLS is genuinely enforced).

**Fails before the fix** — `ADeniedNodeIsNotNamedByTheDrillDown`, run against `main`'s
`StorageAdapterMeshQueryProvider`:

```
Did not expect collection {"TestData/PricingComparison", "TestData"} to contain
"TestData/PricingComparison" because autocomplete reads the same storage, through the same query
shape, as the `get` that refuses this caller — a suggestion the caller cannot then open is a
disclosure of the node's existence and its place in someone else's tree (#3890).
```

**Passes after.** The other two pass on **both** trees, on purpose:

- `AReadableNodeIsStillNamedByTheDrillDown` — the filter did not simply empty the feature, and it is
  what proves the snapshot under assertion is the CONVERGED one rather than `CombineLatest`'s empty
  seed (otherwise the denial assertion could pass having measured nothing).
- `TheOutsiderReallyIsDenied` — the premise, measured against `hub.CheckPermission`, the same fold
  `RlsNodeValidator` consults, asserted `false` for the outsider **and** `true` for the admin so a
  fold that denied everyone would fail here rather than silently validating the leak test.

Runs: pre-fix `Failed: 1, Passed: 2`; post-fix `Failed: 0, Passed: 3`. Full owning project in
Release: `Failed: 0, Passed: 555`.

## Residuals, deliberately not in this diff

- **`StaticNodeQueryProvider` has no access filter at all** — visible in the failing-before output as
  the extra `"TestData"` row. It serves the built-in catalogs (Role/Model/Agent) and `AddMeshNodes`
  bootstrap seeds, which are published to everyone on purpose; it takes no `AccessService` and no
  validators, so gating it is a feature, not a two-line fix. On a real portal, partition roots come
  from the native provider's `partition_access`-filtered top-level index.
- **`CosmosMeshQuery.Autocomplete` and `SqliteVectorMeshQuery.Autocomplete` (MeshWeaver.Plugins) have
  the same gap** — Cosmos does not resolve a user id at all and never calls its own
  `IsReadableAsync`, though its `Query<T>` does. Neither backs `memex`. Filed separately rather than
  smuggled into a core PR.
- **`McpMeshPlugin.Autocomplete` (MeshWeaver.Plugins) does not wrap in `AsCaller(...)`** the way
  `Get`/`Search`/`SearchChunks` do. It works today because the viewer is resolved at CALL time on the
  request thread (this PR resolves at call time rather than subscribe time partly for that reason),
  but it is a structural outlier from that file's own defence-in-depth pattern.

## Gates

No public type or member removed, no interface member added, no `PackageVersion` removed, no
workflow touched — `Pairs-with`/`Implementers`/`Satellite-pins` do not apply. This is a **behaviour
change behind an unchanged signature** (AGENTS.md shape 7), which no surface gate can see by
construction: `MeshWeaver.Plugins`' `BlazorAutocompleteService` already re-establishes the circuit
user across the JS-interop hop expressly so the query's RLS sees them, so the dependent is already
written for a filtered autocomplete.


rbuergi:
Addressed in `328a39b7b`. Four of the five were real; here is the disposition of each.

**`Timeout = 30000` ×3 (critical) — already fixed in `881cb3a0b`**, which landed before the review
was posted (the review read `0fcb5a916`). `TestTimeoutLiteralRatchetGuard` caught it locally first:
*"Hand-written test timeouts rose to 389 (baseline 386)"*. The attribute is simply gone —
`test/xunit.runner.json` already sets `methodTimeout: 30000` for every method in the tree, and each
wait in the suite is bounded by `TestTimeouts.Convergence`, so writing the guess down added nothing.

**Missing denied-root / readable-descendant regression (moderate) — agreed, and it was the gap that
mattered most.** The provider comment argues a partition-level narrowing would be *wrong* because a
grant can sit below a partition root, and nothing tested that claim.
`ADeniedPartitionStillSuggestsTheOneNodeSharedWithYou` now does: the outsider holds Viewer on
`TestData/SharedNotes` alone, and ONE snapshot must contain it **and** omit the ungranted sibling.
A partition-level narrowing fails the first assertion; no filter at all fails the second — so the
test discriminates in both directions rather than only forbidding the leak.

**The resolution comment promises behaviour that is not present (nit) — correct, and thank you.**
It said leaving `UserId` null lets "the provider's own late resolution" run. There is no later
moment on this path today, exactly as you describe: `MeshQuery.Autocomplete` invokes the provider
eagerly and `CollectMatched` resolves `GetEffectiveUserId` while composing. Restated to say what is
actually true — leaving it null preserves that resolution byte-for-byte, reaching the same answer at
the same instant, and is the one form that stays correct if this path ever grows a `Defer` or a pool
hop (which is why the stamp is written this way rather than pinning Anonymous).

**The instrument-table row (nit) — agreed on both halves.** `autocomplete` does not belong in a row
about compilation instruments at all: it returns suggestion projections (path, name, node type,
icon) and never a `compilationStatus`. Removed from the row, and noted in the control section
instead. The denominator wording was imprecise *before* this PR and my edit made it worse: it is per
NODE, not per partition, for the same reason the new test pins. Now says "only the NODES the caller
may read", with a paragraph saying why the count has to be counted rather than inferred from the
list of partitions you can open.

**Suppressed comment on the bake buckets — this was the best catch in the review.** Lumping
`Regressions` / `Unevaluated` / `ContentBroken` as "exists and did not build" is its own false
negative in a section about telling denied from deleted: `Unevaluated` is *no verdict*, and
`ContentBroken` (`NoSources` / `UpstreamContentBroken`) means nearly the opposite — the type node
survives, its **sources** were deleted out from under it. Now a four-row table saying what each
bucket licenses you to conclude.

**What's New scope (nit) — scoped, without advertising the gap.** The guarantee now names the store
every deployment runs, and the "what this does not change" section says an installation that added
an optional third-party index gets its half with a later update. The remaining providers are
`CosmosMeshQuery.Autocomplete` and `SqliteVectorMeshQuery.Autocomplete`, tracked as
Systemorph/MeshWeaver.Plugins#1600 — neither backs `memex`, and on those deployments the pedestrian
provider this PR fixes is registered alongside the native one and is now filtered either way.

Local verification after the change: `MeshWeaver.Hosting`, `MeshWeaver.Documentation` and both test
projects build `-c Release -warnaserror` at `0 Error(s)`; the four #3890 cases pass; and
`MeshWeaver.Documentation.Test` is 368/368 including `DocumentationLinkIntegrityTest`,
`WhatsNewEntryIntegrityTest` and the timeout ratchet.

Issue #2819, opened 2026-08-30
Flake: FileSystemObservableQueryTests.ObserveQuery_ScopeExact_OnlyNotifiesExactPath emits nothing within 30 s on a loaded shard — ejected #2814 from the merge queue


rbuergi:
## What failed

`MeshWeaver.Persistence.Test.FileSystemObservableQueryTests.ObserveQuery_ScopeExact_OnlyNotifiesExactPath` on the **merge-group build of #2814** (run [33326470909](https://github.com/Systemorph/MeshWeaver/actions/runs/33326470909), shard 5, `MeshWeaver.Persistence.Test exit=1 TESTFAIL`, elapsed 85 s), which ejected #2814 from the merge queue at 18:03Z:

```
ObservableAssertionException : Expected the observable to emit a value within 30s, but it did not.
The observable emitted nothing at all.
   at FileSystemObservableQueryTests.cs(301,0)
```

## Attribution

- #2814 touches `MeshWeaver.Mesh.Contract` (a new `InstanceConsent` record), `MeshWeaver.Graph` (its node-type registration), `MeshWeaver.PluginCatalog`, `Memex.Portal.Shared`, i18n and tests — nothing under `MeshWeaver.Persistence` or the file-system storage adapter the test drives.
- `test/MeshWeaver.Persistence.Test/FileSystemObservableQueryTests.cs` last changed 2026-08-27 (`e7bcefcbf`, #2423).
- main's last five Build-and-Test runs (afb5a5ec9 … 2160d6525) are green; #2814's own PR run was green (six shards executed) minutes before the queue build.
- No existing issue names this test.

## Shape

A 30 s wait on an `ObserveQuery` over the **file-system** storage expecting an exact-path notification — the class of test that depends on a file-watcher event arriving on a loaded shard. Whether it is the watcher (coalesced/dropped event under load) or the exact-scope filter swallowing the only event is what needs pinning; the assertion note ("if you filtered upstream … this counts only values that PASSED the filter") points at the second.

Re-queued #2814 (main moved; the next merge group is a different tree). Recording on the flake umbrella #1384 as well.

rbuergi:
## The failing await is a WRITE, not the query — the test's name is a red herring for triage

`FileSystemObservableQueryTests.cs(301,0)` is not a subscription or a notification wait. It is:

```csharp
[...]

PR #3930, opened 2026-09-10
fix(#3477): the phantom guard asks the write path's OWN no-op question, and the test observes the write's contract

rbuergi:
Closes the residual of #3477 in the write path, and fixes the instrument that has kept every sighting
of it unattributable.

## Which of the three candidates it was

Both of the ones that matter, and the evidence separates them:

| candidate | verdict |
|---|---|
| the production re-enqueue path is wrong | **yes** — one no-write exit was still unguarded (below) |
| the test observes the wrong signal | **yes** — it polled a proxy on a bound below the framework's own, and its `[Fact(Timeout)]` sat below its inner waits |
| the harness/fixture races | **no evidence either way** — nothing in the two CI transcripts distinguishes it, and the local statistical repro produced 0 failures in 4 full-assembly runs on the pre-fix binary, so it neither supports nor excludes this. Not pursued; if a sighting survives this change, it is the first candidate to revisit |

## 1. Production: the write path has TWO no-write exits; #3633 guarded ONE

`UpdateRemote` decides not to post a patch in two places:

| # | gate | logs |
|---|---|---|
| 1 | `ReferenceEquals(updated, current) \|\| Equals(updated, current)` | `NO-OP … contentType=…` (Debug) |
| 2 | `ComputeMergePatchDiff(...).Count == 0` | `NO-OP … diff empty after serialisation` (Debug) |

Both post nothing, both complete the caller as a **success**, both are Debug — one decision wearing
two coats. #3633's phantom guard (`ReattemptBaseSource`) was written inline at its call site as **gate
1 alone**, so a phantom base that fails record equality but serialises identically walked past it and
out through gate 2, into exactly the silent loss the guard exists to refuse.

**That pair is this path's ordinary shape, not a contrived one.** `MeshNode.ContentEquals` compares
two `JsonElement`s structurally but refuses a MIXED pair — one `JsonElement`, one typed — outright,
and says why: *"no `JsonSerializerOptions` is available here to bridge representations"*. The mirror
`UpdateRemote` reads lives on the **cache hub**, which in `MeshNode.Equals`'s own words *"does not
know domain types"*, so its content **is** a `JsonElement` while the caller's lambda yields a **typed**
content. Every such write fails gate 1 by construction and is decided at gate 2.

The fix asks the write path's own question once, and both gates are now spelled through the same
helpers so they cannot drift apart again:

```csharp
internal static bool PostsNothing(MeshNode current, MeshNode updated, JsonSerializerOptions o)
    => IsRecordNoOp(current, updated)
       || ComputeMergePatchDiff(ToJsonObject(current, o), ToJsonObject(updated, o)).Count == 0;
```

The extra serialisation is paid **only where the guard runs at all** — a re-attempt whose owner said
the write never applied. A first attempt, a `Conflict` re-attempt and any write yielding a real diff
are byte-for-byte the path they were.

## 2. The instrument: the tests could not report their own failure

Three defects, all in the tests, all arithmetic rather than opinion:

1. **`LateNackReenqueueTest` waited on the proxy before the contract.** Ground truth was
   `storage.Read(path)` reaching the marker; the caller's terminal — the only wait with a bound and a
   diagnosis of its own — was checked *afterwards*. So the framework's `OwnerUnreachable` + `corr=`
   request trail was always discarded in favour of `System.TimeoutException : The operation has timed
   out.`, which is verbatim what #3477 recorded. #2819's rule restated.
2. **The decisive bound was a hand-written `45.Seconds()`** — not CI-scaled while every other wait in
   the same test was `TestTimeouts.Convergence`, and below what one re-enqueue may legitimately cost.
   The issue's elimination table read every terminal's deadline from a common origin; they compose
   **additively** (`BaseStateWaitBound` 30 s from subscribe, then `WriteVerdictBound` 31 s from the
   *post*), so T+30…T+61 is silent **by design** and the 45 s window sat inside it.
3. **`[Fact(Timeout = 90_000)]` was BELOW the inner waits on a runner** (`TestTimeouts.Convergence` is
   108 s at the CI factor), so xunit killed both tests anonymously before any inner wait could report.
   Core merge-queue run `34332482683` (2026-09-09, shard 4) is an instance: *"Test execution timed out
   after 90000 milliseconds"*, no assertion, no named wait.

The corrected order asserts the caller's terminal first and durable storage second, which is a
**strictly stronger** claim: the owner acks only after its flush, so once the caller holds a success
the store already holds the value — and a phantom success now fails on the storage assertion and says
so, instead of hiding inside an anonymous timeout.

## Control, both directions

**Deterministic, at the seam.** `ReattemptBaseIsAuthoritativeTest` gains a case whose phantom holds the
value as untyped JSON against a lambda producing it typed. It asserts first that the pre-#3477
predicate does *not* see that base (a positive control on the gap), then the behaviour.

- **RED with the pre-fix predicate planted** (`PostsNothing => IsRecordNoOp(current, updated)`):
  `Expected value to be "initial" because the owner has just stated it does not hold this write … but
  found "post-nack-4b7556aa5983".`
- **GREEN restored**: 17/17 over `ReattemptBaseIsAuthoritativeTest`, `LateNackReenqueueTest`,
  `LateNackReenqueueCorrelationTest`, `WriteBaseStateTotalityTest`, `BaseReadBoundIsReachableTest`,
  `UpsertAnswersWhenTheOwnerGoesAwayTest`; full `MeshWeaver.Graph.Test` and
  `MeshWeaver.Documentation.Test` green.
- The cases now call `MeshNodeStreamHandle.PostsNothing` — the predicate `UpdateRemote` itself passes
  — instead of a local copy, so the production guard can no longer be narrowed without a case going
  red.

**Statistical, honestly reported.** The 1/330 race did **not** reproduce here: 4 full-assembly
`MeshWeaver.Graph.Test` runs (1506 tests each, `DOTNET_PROCESSOR_COUNT=2`, load 6–29) on the *pre-fix*
binary, 0 failures. At this rate a clean streak is close to no evidence, which is exactly why the fix
is pinned at the seam and not by a green run.

## 🚨 Still open — a maintainer decision, not part of this change

`LatePatchResponseRegistry.WriteVerdictBound` documents itself as *"the outer bound on a caller-visible
mesh WRITE"*, and `TestTimeouts.Convergence` derives from it so every waiter in the fleet dominates it
by construction. **For a write that re-enqueues it is false**: each attempt arms its own deadline from
its own post, `MaxOwnerDisposingReenqueues` is 2, and each attempt also pays a base read *outside* that
deadline — worst case ≈ 3 × (30 + 10 + 31) ≈ 213 s against a published 31 s. The two candidate fixes
(mint the deadline once per write and thread the remaining budget through re-attempts, vs. publish the
composite bound) are written up in the doc page; the first changes caller-visible behaviour on the slow
path, so it is stated rather than smuggled in.

## 🚨 Plugins twin: port obligation

`test/MeshWeaver.Graph.Test/LateNackReenqueueTest.cs` is one test kept in two files; Plugins'
`TeardownTwinParityTest` compares its copy against core's body **at `MW_PLATFORM_REF`**. It reddens in
the **pin bump**, not here — the copy in
`MeshWeaver.Plugins/src/MeshWeaver.Hosting.Monolith.Test/LateNackReenqueueTest.cs` must carry this body
in the same change that moves the pin past this commit. That Plugins copy is the one that actually
failed in #3477's original sighting, so it keeps the `45.Seconds()` / `Timeout = 90_000` defects until
that port lands.

Doc: `Doc/Architecture/PhantomBaseAfterOwnerDisposal` — extended with the second exit, the additive-bounds
correction to the original elimination, the observation contract, and the open bound question.

Refs #3477.

### DIFF
--- a/src/MeshWeaver.Documentation/Data/Architecture/PhantomBaseAfterOwnerDisposal.md
+++ b/src/MeshWeaver.Documentation/Data/Architecture/PhantomBaseAfterOwnerDisposal.md
@@ -177,11 +177,15 @@ cases those are in two places of its own:
   a re-parsed `JsonElement` …, or a content record holding collections (compared by reference) all
   read as 'changed' while the persisted JSON is byte-identical"*.
 
-And the mixed pair is what this path always has. `UpdateRemote` runs on the **cache hub**, whose
-mirror content is a `JsonElement` because — in `MeshNode.Equals`'s own words — that hub *"does not
-know domain types"*; the caller's lambda produces a **typed** content. So every such write fails gate
-1 by construction and is decided at gate 2, which is precisely where the phantom guard was not
-looking.
+And a mixed pair is what this comparison gets **whenever the content type resolves**. `UpdateQueued`
+wraps the caller's lambda as `update(EnsureTypedContent(node, …))`, so the lambda's **output** carries
+typed content while `current` — the raw mirror emission it is compared against — still carries the
+`JsonElement`. Gate 1 therefore cannot fire, and gate 2 decides, which is precisely where the phantom
+guard was not looking.
+
+When the type does **not** resolve, `EnsureTypedContent` degrades back to a `JsonElement`, both sides
+stay untyped, and `JsonElement.DeepEquals` settles it at gate 1. So gate 2 is the resolved-type case
+— the ordinary one — not literally every write.
 
 The fix is to ask the write path's own question once:
 
--- a/src/MeshWeaver.Documentation/Data/Architecture/PhantomBaseAfterOwnerDisposal.md
+++ b/src/MeshWeaver.Documentation/Data/Architecture/PhantomBaseAfterOwnerDisposal.md
@@ -126,6 +126,142 @@ bounded below the 45 s the test waited:
 Exactly one terminal is both silent and leaves storage unchanged, and it is the one that requires the
 mirror to already carry the marker. That is the reading; the code that produces it is above.
 
+### 🚨 A correction to that elimination: the bounds compose ADDITIVELY
+
+The table above reads every terminal's deadline from a **common origin** — the NACK — and concludes
+that all of them fall inside the 45 s the test waited. They do not. The re-attempt's bounds are
+**sequential**, and the same page of `LatePatchResponseRegistry` says so about its own window: *"The
+owner-side paths were enumerated as ALTERNATIVES, taking their maximum; in
+`ApplyMeshNodePatchInTurn` they compose ADDITIVELY."* The re-attempt is the same shape:
+
+```
+BaseStateWaitBound (30 s, from SUBSCRIBE)  →  post  →  WriteVerdictBound (31 s, from the POST)
+                                              └─ plus GetMeshNode's 10 s when the base is a phantom
+```
+
+So `VERDICT_TIMEOUT` is due at *(base-read latency) + 31 s*, not at 31 s. A re-attempt whose base
+arrives after 14 s produces its first Warning at 45 s or later — **outside the window the test was
+watching**. The region from T+30 to T+61 is therefore silent BY DESIGN, and the failing observation
+sat inside it.
+
+That does not make the phantom reading wrong; it makes the sighting **under-determined**. It is
+consistent with two things, and this page's change closes the first while the observation contract
+described below closes the second:
+
+1. the re-attempt settled silently against a phantom base (this page), or
+2. the re-attempt had simply not reached any terminal yet, and the test stopped watching first.
+
+## The residual: the write path has TWO no-write exits, and #3633 guarded ONE
+
+`UpdateRemote` decides not to post a patch in two separate places:
+
+| # | gate | where | logs |
+|---|---|---|---|
+| 1 | `IsRecordNoOp(current, updated)` | before anything is serialised | `NO-OP … contentType=…` (Debug; Warning only for a raw `JsonElement`) |
+| 2 | `ComputeMergePatchDiff(...).Count == 0` | after serialisation | `NO-OP … diff empty after serialisation` (Debug) |
+
+Both post nothing, both complete the caller as a **success**, both are Debug. They are one decision
+wearing two coats — and `ReattemptBaseSource`'s phantom test was originally written at its call site
+as **gate 1 alone**. A phantom base that fails record equality but serialises identically therefore
+walked straight past the guard and out through gate 2, into precisely the silent loss the guard
+exists to refuse.
+
+**That pair is not contrived — it is the cache hub's ordinary shape.** Gate 2 exists because "a
+rebuilt-but-identical content slips past the record-Equals check above", and `MeshNode` says which
+cases those are in two places of its own:
+
+- `MeshNode.ContentEquals` compares two `JsonElement`s structurally, but a **MIXED** pair — one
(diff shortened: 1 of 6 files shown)


rbuergi:
## Review addressed in `9f0bcb755` — all four findings were real

| finding | what changed |
|---|---|
| the seam takes an **injected** predicate, so a production regression to `IsRecordNoOp` keeps the test green | `ReattemptBaseSource` now takes the caller's `update` lambda + the hub's `JsonSerializerOptions` and computes `PostsNothing` **itself**. There is one predicate and no way to pass a different one — which is exactly the hole that let #3633's guard cover one of the two exits. |
| `callerTerminal` / `callerError` handed over through plain captured fields | `Volatile.Write` / `Volatile.Read` both ways. Polling rather than awaiting is kept **deliberately**: awaiting resumes the continuation inline on the signalling thread — the 90 s wedge documented on `CapturingLoggerProvider.FirstMatching`. |
| both waits still ended in a bare `.Timeout(...)`, so a phantom success is still reported anonymously | each bound is turned into a **value** (`false` / `null`) and the assertion carries the diagnosis. A phantom now reads: *"the caller has just been handed a SUCCESS … so a store that does not carry the marker means the write was reported saved while it is in no store at all"*. |
| the regression case's mirror JSON was a literal, pinning only default options | the mirror is produced by the **same serializer the diff uses**, so both sides are byte-identical under whatever options are supplied — hub naming policy and `$type` included. |

And one **overstatement corrected**, which the review caught: gate 2 is reached for every write whose
content type **resolves**, not literally every write. `UpdateQueued` wraps the lambda as
`update(EnsureTypedContent(node, …))`, so `updated` is typed while `current` is still the raw
mirror's `JsonElement` and `MeshNode.ContentEquals` refuses that mixed pair; when the type does not
resolve, both sides stay untyped and `DeepEquals` settles it at gate 1. (Note the mechanism runs the
other way round from the review's reading — `EnsureTypedContent` is applied to the lambda's INPUT,
never to `current`, which is what makes the pair mixed rather than what rescues it.)

### On the composite-bound comment

> Use a bound for the composite reattempt or fix the production write budget before relying on this assertion.

Agreed as a diagnosis, and it is this PR's **stated open item** rather than something to absorb into a
test bound: `WriteVerdictBound` is published as *"the outer bound on a caller-visible mesh WRITE"* and
is per-attempt, so a re-enqueue chain's worst case is ≈213 s against a published 31 s. Widening the
test to cover 213 s would encode the defect instead of reporting it, so the wait keeps dominating the
**published** bound and now says so when it is the one that fires:

> *🚨 If THIS is what failed, the write is unbounded rather than merely slow … the open defect
> recorded in `Doc/Architecture/PhantomBaseAfterOwnerDisposal`, not a number to widen here.*

Fixing the budget properly (mint the deadline once per write, thread the remaining budget into every
re-attempt) changes caller-visible behaviour on the slow path, which is a maintainer call — written up
with both candidate shapes in the doc page.

Control re-run after the restructure: planting the pre-fix predicate is still RED
(`Expected value to be "initial" … but found "post-nack-4b7556aa5983"`), and it can no longer be
masked by the test supplying its own predicate. Green restored — 17/17 targeted, 1507/1507 full
`MeshWeaver.Graph.Test`, 368/368 `MeshWeaver.Documentation.Test`, Release `-warnaserror` `0 Error(s)`.

Issue #2593, opened 2026-08-28
Move the Swiss-German voice model off the GitHub release — it is the last pin keeping pre-rewrite history alive


rbuergi:
## Context

Core's history was rewritten on 2026-08-28 (maintainer-ordered): the departed package trees (94
roots — the AI family, Blazor family, clients/, content, old samples/tests/tools) were pruned from
history, tree-identical at every branch tip. Main and all branches/tags moved.

**One pin remains**: the `voice-model-swiss-german` tag + release. Its 574MB `ggml-swiss-german-turbo-q5_0.bin`
release asset is load-bearing — `clients/voice-gateway/run-local.sh` (and its README) download it
from `releases/download/voice-model-swiss-german/…` — and the release is **immutable**, so the tag
cannot be moved, and deleting the release would kill the URL irrecoverably (immutable-release tag
names are permanently reserved; we learned this the hard way on `v3.0.0-rc1` / `v3.0.0-preview1`,
whose release pages are now gone and whose names are burned).

Because that tag's commit reaches the full pre-rewrite history, **every clone still fetches the
~10,000 pre-rewrite commits (≈118MB instead of ≈83MB), and the pruned trees remain reachable**
through it.

## The fix

1. Host the model where large binaries belong — ACR (an OCI artifact) or the blob account the
   deployments already use.
2. Repoint `clients/voice-gateway/run-local.sh` + README to the new URL.
3. After a deprecation window, delete the `voice-model-swiss-german` release + tag (release first;
   accept the name is burned — that is fine, nothing will reuse it).
4. Fresh clones then drop to ~83MB and the departed code truly leaves history.

## Guardrails

- 🚨 Do NOT delete the release before step 2 ships — the URL dies irrecoverably.
- 🚨 Any future release of a large binary goes to artifact storage, never a release asset on this
  repo: release assets + immutability + history pins interact exactly as badly as this issue
  documents.

🤖 Filed by Claude Code as the follow-up completing the 2026-08-28 history prune.

rbuergi:
Progress 2026-08-28: the model is re-hosted — release **`voice-model-swiss-german`** now exists on **MeshWeaver.Plugins** (published 2026-08-28T11:39Z) with the `ggml-swiss-german-turbo-q5_0.bin` asset. The core repoint (`clients/voice-gateway/run-local.sh` + README) is in flight as **#2602**.

[...]

PR #3956, opened 2026-09-10
fix: the whisper container bakes its model — the release moved, the pin cannot follow (#3906)

### LINKED ISSUE #3906: deploy/whisper cannot fetch its model — #2593 moved the release and left the chart behind
**Why this repo:** the broken pin is `deploy/whisper/helm/values.yaml` and `deploy/whisper/docker-compose.yml`, and the fetch runs in this repo's chart. No module is involved.

## The chart cannot start

`values.yaml:22` points the model initContainer at

    https://github.com/Systemorph/MeshWeaver/releases/download/voice-model-swiss-german/ggml-swiss-german-turbo-q5_0.bin

which returns **HTTP 404** — `gh release view voice-model-swiss-german --repo Systemorph/MeshWeaver` reports no such release. The asset moved to the **private** `Systemorph/MeshWeaver.Plugins` repo under the same tag in `6d726d361` (#2593, 2026-08-28). That commit updated `deploy/whisper/README.md`, `clients/voice-gateway/README.md` and `run-local.sh` — but not `values.yaml` or `docker-compose.yml`, and nothing has touched `deploy/whisper` since. #2593 is CLOSED.

## Why repointing the URL is not enough

The initContainer runs a bare, unauthenticated fetch under `set -e`:

    curl -fSL --retry 3 --retry-delay 5 "$MODEL_URL" -o /models/model.bin.tmp

`-f` fails on the 404 and `set -e` aborts, so the pod never starts. A **private** release asset also needs an `Authorization` header and the API asset URL rather than the browser download URL — so this needs a token-carrying fetch, the model staged into a volume or the image, or the asset republished somewhere the cluster can reach unauthenticated.

## Impact

Speech-to-text cannot be enabled on any deployment. On memex.systemorph.com the module is loaded — `ISpeechTranscriber` resolves to `WhisperContainerTranscriber` — but `Speech:Enabled=false` and `Speech:Endpoint` is empty, so `/api/speech/transcribe` answers 503. Standing the container up is the blocker for the Swiss-German transcription use case, and it fails at the first step.

Verified 2026-09-10: URL still returns 404, `values.yaml:22` unchanged.

rbuergi:
Fixes #3906.

## The premise, re-verified today

`deploy/whisper/helm/values.yaml:22` pointed a model initContainer at a GitHub release on **this** repo. Measured 2026-09-10:

| URL | Status |
|---|---|
| `github.com/Systemorph/MeshWeaver/releases/download/voice-model-swiss-german/ggml-swiss-german-turbo-q5_0.bin` (the removed pin) | **404** |

#2593 moved that asset to the **private** `Systemorph/MeshWeaver.Plugins` on 2026-08-28, updated three READMEs and `run-local.sh`, and left `values.yaml` and `docker-compose.yml` behind. The initContainer runs `curl -fSL … "$MODEL_URL"` under `set -e`, so the 404 aborts the pod in `Init:Error` and speech-to-text could not be enabled on any deployment.

## The decision

Repointing the URL is not a fix — a private asset needs an `Authorization` header and the API asset URL. Three real options; I picked the third.

### ❌ 1. Token-carrying fetch — rejected on blast radius

The only credential that can read a private release asset is one that can read the **repository**, and `Systemorph/MeshWeaver.Plugins` is the private home of the entire AI engine, the plugin sources and the fleet's module trees. That buys one 547 MB blob with read access to the whole private platform, held as a standing namespace secret that **does not exist today** and would need rotating forever. It also inverts the plugin-catalog credential model, whose whole point is that the registry holds the git credential so consumers never need one.

### ❌ 2. Republish it anonymously — rejected on LICENCE, and again on prior intent

**I checked the licence rather than assuming it.** The model derives from [`Flurin17/whisper-large-v3-turbo-swiss-german`](https://huggingface.co/Flurin17/whisper-large-v3-turbo-swiss-german), published under **CC BY-NC-4.0** — Attribution, **NonCommercial**. Verified two ways on 2026-09-10: the model card states *"This model is distributed under the Creative Commons Attribution-NonCommercial 4.0 license (`cc-by-nc-4.0`)"*, and the HF API returns `cardData.license: cc-by-nc-4.0` plus the tag `license:cc-by-nc-4.0`. The base model `openai/whisper-large-v3-turbo` is **MIT**, but that is not the binding half — a derivative cannot be more permissive than the work it derives from, so our GGML file inherits the NonCommercial restriction. **Publishing it for anonymous world download from a commercial product's infrastructure is exactly what NC restricts.**

Independently, #2593 already decided this: it moved the asset *off* such a URL deliberately, recording that *"this is a Systemorph-tuned model, and anonymous world-download was an accident of hosting"*, that private hosting is *"a feature, not a cost"*, and the guardrail *"any future release of a large binary goes to artifact storage, never a release asset on this repo."*

For completeness, both anonymous URLs that already exist were measured and both **404**: the old core release asset, and the portal content route `/api/content/MeshWeaver/static/Speech/…` documented in `OnDeviceVoice.md` — which is *expected* to 404 (that page's own caveat says the `static-assets` share is not yet mounted) and is an **access-controlled** route in any case, so it was never anonymous.

### ✅ 3. Stage the model into the image — chosen

Bake it at `/models/model.bin` and let the image registry distribute it. The pod **already authenticates to that registry to pull its own image**, so this adds **zero new credentials, zero secrets, zero standing access**, and the model never leaves the authenticated boundary the licence requires. It is also the option #2593 named first — *"ACR (an OCI artifact) …"* — in the form kubelet can already fetch (a bare OCI artifact would need `oras` in an image that ships only `curl`).

Two bugs it fixes beyond the 404: the `emptyDir` was **per-pod**, so every restart re-downloaded 547 MB; and the failure now lands at **build** time, on the machine of the person who can fix it, naming the file and the exact `gh release download` command.

**Cost, stated plainly:** the image grows by **574,041,195 bytes (~547 MiB)**, and the build requires the model in the build context first (`gh auth`).

## Also fixed, because a half-fix is what caused this

`deploy/whisper/docker-compose.yml` carried the same stale pin — repaired. Its `./models:/models:ro` bind mount is **removed**, because an empty host dir would now *shadow* the baked model. `NOTES.txt` no longer tells operators the pod downloads the model, and the `.gitignore` note that "the Dockerfile bakes no model" is corrected.

## What I did NOT do

None of the forbidden softenings: `-f` stays, no `|| true`, the init is not made non-blocking, no retry count raised. Each turns a loud 404 into a pod that starts without a model and serves a silent 503.

`image.tag` is deliberately **unchanged** at `1.7.4`: it names the whisper.cpp release the binary is built from, and moving it to a tag no image carries would red the nightly `lock-pinned-digests` **axis-2** lane (*"does not resolve in the registry, so there is nothing to protect"*). I verified the extractor still finds the pin in the edited file.

## 👤 What a person must still do — two named, actionable items

1. **Build and push a model-bearing `whisper-swiss-german` image** (needs `gh auth` to the private repo + registry push; no new credential is created):
   ```bash
   gh release download voice-model-swiss-german --repo Systemorph/MeshWeaver.Plugins \
     --pattern 'ggml-swiss-german-turbo-q5_0.bin' --output deploy/whisper/models/model.bin
   az acr build -r meshweaver -t whisper-swiss-german:1.7.4 deploy/whisper
   ```
   Until this happens, `meshweaver.azurecr.io/whisper-swiss-german:1.7.4` has **no model in it** and the pod will start and fail to load one. I cannot do this: it needs registry push, and operations here go through the memex API — I did not touch a cluster.
2. **Mirror this change into `Systemorph/Memex`** at `deployments/aks/memex-cloud/whisper/helm/` — filed as **Systemorph/Memex#271** so it is tracked rather than living in this PR body. That is a **verbatim vendored copy** of this chart, not a dependency — I read it and it carries the identical broken URL. Nothing makes the two agree.

Separately, and **not** resolved by me because it is not mine to resolve: CC BY-NC-4.0 is a **NonCommercial** licence, and this model is used in a commercial product. That question predates this issue (it applies equally to the shipped on-device MAUI path) and my change does not alter the posture — but it is now written down rather than implicit, and it deserves a decision.

## Conserved work product

`src/MeshWeaver.Documentation/Data/Architecture/VoiceModelDistribution.md` (`/Doc/Architecture/VoiceModelDistribution`) — the licence and how it was verified, the standing rule (*this asset travels only through channels that already authenticate their reader*), all three options with why two lost, **every** consumer including the Memex vendored copy, and a checklist for the next move. This is the second time the asset moved and left a consumer behind, and the reasoning existed nowhere in the repo. Linked from `Architecture.md`, `OnDeviceVoice.md`, `CentralizedSpeech.md` and the whisper README.

## Verification — each of these can fail

| Check | Result |
|---|---|
| removed pin (old core release) | **404** |
| private asset, **unauthenticated** | **404** — this is why a repoint cannot work |
| private asset via `gh api` (the documented path) | **200**, `574,041,195` bytes, `sha256 2d56e77…5ea612` |
| portal content route from `OnDeviceVoice.md` | **404** (and access-controlled) |
| `meshweaver.azurecr.io/v2/` anonymously | **401** — the authenticated channel, which is the point |
| `helm lint` + `helm template` | clean |
| rendered Deployment | asserted: **no** initContainer, **no** `/models` volume/mount, **no** `MODEL_URL` — and the assertion was **falsified against the pre-change chart** (exit 1, naming all six) |
| `DocumentationLinkIntegrityTest` | **Passed** — after first injecting a broken link into the new page and confirming it goes **RED naming that page** |
| `lock-pinned-digests.py --self-test` | passed; axis-2 still extracts `whisper-swiss-german:1.7.4` |
| `check-workflow-timeouts.py` / `check-pr-secret-preflight.py` | 0 violations |
| `dotnet build -c Release -warnaserror` (Documentation.Test) | 0 Warning(s), 0 Error(s) |

**No What's New entry**, and deliberately so: nothing here is user-visible yet. Speech was already 503 on every deployment and stays 503 until item 1 above is done and `Speech:Enabled` is set — this change makes that possible, it does not perform it. An entry now would announce a fix nobody can observe.


rbuergi:
## The red on `842c1b5a` was not this diff, and this is not a flake-reroll

`Consolidate test results` failed on head `842c1b5a` — **1 failure out of 6,920 tests**, on shard 4:

```
MeshWeaver.Graph.Test.NackReachesTheWaiterDuringTeardownTest
  .OwnerDisposingUnderMeshTeardown_StillAnswersTheWaitingCaller

AssertionException : Did not expect collection {"3YOdHJMOHEKUFJlwHFjJmg"} to contain
"3YOdHJMOHEKUFJlwHFjJmg" because the owner minted an OwnerDisposing NACK for this patch,
and a caller was armed and waiting for it … a watch still armed here is the #2778 defect
NackReachesTheWaiterDuringTeardownTest.cs(242)
```

That is the regression guard for #2778 (closed `completed`, 2026-08-31) — a mesh hub-teardown NACK race.

**This diff cannot reach it.** `git diff --name-only origin/main...HEAD` matched **zero** `.cs` and zero `.csproj` files: the change is `deploy/whisper/**`, `.gitignore`, and markdown under `src/MeshWeaver.Documentation/Data/`. There is no mechanism by which a Helm chart, a Dockerfile and some embedded markdown alter hub disposal ordering in `MeshWeaver.Graph.Test`.

**What I checked before touching anything** (rather than assuming "flake" and re-running, which AGENTS.md forbids):

| Question | Answer |
|---|---|
| Catalogued in `.github/known-flakes.json`? | **No** — `entries` is empty |
| An existing issue for this test? | **No** — one closed PR (#2890, 2026-08-31), nothing open |
| Is my first push a usable control? | **No.** `d6e43eec`'s run was **cancelled** by my second push (`Build solution (once) | cancelled`), so its `Consolidate` failure means *superseded*, not *test failed*. A completed run is not a conclusive one |
| Is `main` green? | **Yes** — `03f4d7b67`, completed `success` at 18:37Z, i.e. the same guard passing minutes later on a tree carrying all 15 commits my branch was behind |

So: the guard fails on my merge-base tree and passes on main's tip. AGENTS.md's rule for exactly this case is *"merge main only when the PR is `DIRTY` or **CI fails on something your diff cannot reach**"* — which is why this head is a merge of `origin/main` and **not** a re-run of the same tree. The tree under test genuinely changed.

Re-verified locally on the merged tree before pushing: `dotnet build -c Release -warnaserror` → 0 Warning(s), 0 Error(s); `DocumentationLinkIntegrityTest` + `ArchitectureTopicMapTest` → 5 passed (the topic-map guard matters here — main restructured `Architecture.md` while this PR was open, and my entry had to survive the merge on its own line; it did, line 202); `helm lint`/`template` clean with the rendered Deployment still asserted to carry no initContainer, no `/models` volume and no `MODEL_URL`; and the axis-2 overlay pin still extracts as `whisper-swiss-german:1.7.4`.

**If shard 4 fails the same way again**, that is deterministic evidence the #2778 guard is genuinely racy or its fix has regressed — a real defect that belongs to #2778's owner, not to this PR. I will file it and stop rather than roll again.

PR #3959, opened 2026-09-10
fix: collect superseded generations inside a retained identity (#3461)

rbuergi:
## What this is

The **precondition on flipping a prefix to the generation layout**, and the reason phase 4 of #3461
has been blocked: generation retention.

`PrebuiltBundleStore` swept at the **identity** level only. Under an identity its rules KEEP — the
running one, a released one, a pinned one, a recent one — it removed nothing, so once generations
start being written, one ~45-file publication per CI run accumulates inside a live identity for
ever. A share that fills up **truncates writes silently**: memex.systemorph.com reached 3 MiB free
on 2026-09-08 with every runtime recompile landing as `Bad IL format` and CD's read-back answering
`ResourceNotFound` for 39 of 45 files.

## The rule — the same one, one level down

A generation inside a **retained** identity is collected only when **all** of these hold, and kept
when any one fails:

- **no `_current` names it** — the live publication is kept however old;
- **the pointer that would say so resolved cleanly** — unreadable, blank, refused or dangling keeps
  **every** generation of that source, because which one applies is then unknown;
- **its own seal could be read** — unreadable is never unreferenced;
- **its newest write is older than `MinimumAge`** (at least 30 days).

Generations under a *collectable* identity are deliberately not listed: that directory goes whole
and its byte count already includes them. An aborted plan collects nothing, at either level.

## Why it is sound without a per-generation consumer inventory

That is the one thing that could make it unsound, so it is argued rather than assumed. **A
non-current generation is unreachable, not merely unfashionable.** Every read of a published source
directory composes under `ShippedPrebuiltBundles.PublicationDirectoryOf` — the boot seeder, the
same-major adopted fallback, `SealedPublicationIndex`, `PublishedBundleCatalogue` and
`ServedModuleBytes` alike — and nothing anywhere enumerates a source's subdirectories looking for a
publication. So the set an inventory would have to protect is exactly *{what `_current` names}* ∪
*{everything young}*, and both are kept by construction.

The sweep therefore uses **the readers' own resolution** instead of a second copy of the rules:
`ResolvePublicationPointer` is `PublicationDirectoryOf` with the fallback *reason* kept rather than
discarded, and `PublicationDirectoryOf` now calls it. A sweep that decided reachability differently
from the readers would delete bytes a portal was still resolving, silently.

And *"publication must establish protection before moving `_current`"* is satisfied by **age, not a
lease**: a generation is protected from its first byte, so a publication in flight can never be
collected and the publisher needs no claim, no lock and no ordering with the sweep — which is what
makes this implementable across two repositories that share no lock. `UnsealedGrace` is deliberately
not consulted: hours against a 30-day floor, so the branch could never fire, and a check that cannot
fail is not a check.

**A directory is a generation only on POSITIVE evidence that it is a publication** —
`source-commit.txt`, `repository.txt`, the sentinel, or the pointer names it. Never "every
subdirectory but the ones I know about": the publisher's `modules/` sits beside the generations in
the flat compatibility copy, and under an exclusion list every bookkeeping directory added later
becomes collectable the day it is added, silently.

## Controls, in both directions

13 new cases, **47/47 in the file**, 584/584 in `MeshWeaver.Hosting.Test`, 373/373 in
`MeshWeaver.Documentation.Test`, 15/15 in the reader half (`PublicationGenerationTest` — the
`PublicationDirectoryOf` refactor is behaviour-preserving).

- **positive** — `ASupersededGeneration_OlderThanTheWindow_IsCollected_AndTheLiveOneSurvives`: a fix
  that refused everything would pass a one-sided test.
- **the load-bearing claim, asserted** — `WhatIsCollected_IsExactlyWhatNoReaderResolves`: what the
  plan collects is asserted not to be what `PublicationDirectoryOf` returns.
- **refusals, each with its reason** — the live generation however old
  (`it is the live publication`); a dangling pointer and an empty pointer (`licenses no deletion`,
  `the pointer is empty`) protecting *every* generation of the source; an unreadable seal; a
  non-publication directory (`modules/`) that is never even enumerated; an aborted plan; the flat
  layout, which has no generations at all.
- **end to end** — `SweepCore_RemovesASupersededGeneration_AndLeavesTheLiveOne`, report-only removes
  nothing, and a failed removal is counted and re-planned.

## Measured before deciding — and it reproduces the earlier finding a third time

`Doc/Architecture/SealedPublicationReads` → "Reproduced a third time, on a different day". **61 bake
jobs that ran to a terminal conclusion over 22 h (2026-09-09 20:00Z → 2026-09-10 18:09Z)**, 33 core
CD and 28 satellite, every job log fetched, **zero unreadable identities**. 36 sealed publications
over 19 prefixes.

- 🚨 **7 of 19 prefixes were written by BOTH lanes — and in all 7 the two lanes carried DIFFERENT
  source commits.** #3461's premise is confirmed, not narrowed: the identity is a property of the
  platform image's reference surface and carries no content commit, so the content × framework
  sealed-skip cannot separate the lanes when they meet.
- 🚨 **Concurrent publications on one prefix: 4 — all same-lane, zero cross-lane.** Third
  independent day, same answer. **"One owner per prefix" would have addressed 0 of the 4**, exactly
  as it addressed 0 of the ones measured on 09-06 and 09-08. GitHub `concurrency:` cannot serialise
  them either — it is per-repository and these lanes live in two repos.

The one cross-lane near-miss shows the *mechanism*, not luck: on `s546f29f9…`, core CD `34430130924`
sealed at **03:27:37.327Z**; satellite `34430082656` reached its publish at **03:30:57Z**, found
*"holds a COMPLETE publication of THIS content"* on both shares and skipped — they were baking the
same plugins commit `3f7686da2…`, because core CD resolves the plugins tip at its gate.

So the design already chosen stands: the *layout* is what removes contention, because a publication
written into its own directory is disjoint from every other publication whoever wrote it.

## Not this PR

Phase 4 (flipping `plugins`, both producers in one change set) and phase 5 (dropping the flat copy).
Retention was the stated blocker on phase 4; it is no longer one. **Nothing behaves differently on
any installation today** — `publication-layout` still defaults to `flat`, so no producer writes a
generation and there is nothing to collect yet.

Pairs-with: none — nothing public was removed; the change adds two types and six members under
`MeshWeaver.Hosting` and removes nothing (verified with `scripts/check-type-forwards.py`: 11 removed
entries in its output are `origin/main`'s own, my base being 12 commits behind; 0 removals in this
diff).

Mirror-sync: none — no localization catalog key is added, changed or re-worded.

Refs #3461

### DIFF
--- a/src/MeshWeaver.Documentation/Data/Architecture/SealedPublicationGenerations.md
+++ b/src/MeshWeaver.Documentation/Data/Architecture/SealedPublicationGenerations.md
@@ -144,7 +144,7 @@ directory:
 | `.`, `..`, anything with a separator, anything rooted | 🚨 a pointer is a NAME. It must never be able to address bytes outside its own source directory — refused and logged as a warning, because a publisher wrote something this reader will not follow |
 | names a directory that is not on disk | a retention sweep outran the pointer; logged as a warning |
 
-### Retention
+### Generation retention
 
 Follow [Released Artifact Retention](/Doc/Architecture/ReleasedArtifactRetention)
 (#3842). The 30-day age window and release/consumer references supersede the earlier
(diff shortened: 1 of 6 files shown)


rbuergi:
🚨 **Found while verifying this change is reachable at runtime, and it is not — filed as #3963.**

`AddPrebuiltBundleRetention` has **zero callers**: not in core `src/`, not in core `test/`, not in
MeshWeaver.Plugins `src/`. The only other mention of `PrebuiltBundleRetentionHostedService` anywhere
is a *logger category* in `DeploymentPinnedReferences.cs:47`. The control that makes this conclusive
rather than a naming guess: the sibling `AddAssemblyCacheRetention` **is** registered in both portal
hosts (`Memex.Portal.Distributed/Program.cs:138`, `Memex.Portal.Monolith/Program.cs:49`), and this
one is registered in neither.

So the sweep this PR extends has never executed on any portal — written in response to the
2026-09-08 share-full outage and dark ever since, in the same shape as a CI gate that skips: code
present, tests green, documentation describing it as the store's cleanup, and no observable
difference between "found nothing to collect" and "never ran".

**Deliberately not fixed here.** The registration is one line per host, but it lives in
MeshWeaver.Plugins and `Delete` defaults to `true`, so arming it is a first pass that deletes against
defaults nobody has checked against a live store — exactly the "its own reviewed change" the
retention rule asks for, and #3963 carries the report-only-first rollout. This PR's collection is
correct and tested either way, and it changes nothing today regardless, since no producer writes a
generation yet.

rbuergi:
Addressed the review point in 955ed3a5d, structurally rather than as a tweak — thank you, it was a real cost.

`ReadIdentity` already walked the identity recursively **and then walked each source recursively again**; adding a walk per generation made it 1 + S + G enumerations, touching every file three times per pass. It is now **one recursive walk per identity**, with the identity's bytes and newest write, each source's newest write, and each generation's bytes, newest write and marker files all derived from it. That also removes the pre-existing per-source walk, so this is strictly fewer enumerations than before the feature — which matters because the root is an Azure Files share where every enumeration is a network round trip, and the cost would otherwise have grown with the exact quantity this sweep exists to bound.

Rather than deferring the deep read to delete-time, two things dropped out honestly:

- `IsSealed` on a generation became **`HasSentinel`** and now says what it measures — the sentinel is *there*, not that the publication is whole. Verifying the listing (as the source-level `IsSealed` does) would cost one file read per generation per pass to answer a question no rule asks: a superseded generation is collected on age whether it was sealed, torn or abandoned. It is a note for the operator — it distinguishes a wave of *abandoned* publications from ordinary supersession — and decides nothing. The torn-generation test now pins that meaning instead of the old stronger one.
- the per-generation `ReadFault` is gone and its protection has **not** weakened: a generation the walk cannot enter contributes no files, so it is never *identified* as a publication and never becomes a candidate. Unreadable still means kept — by not being enumerated rather than by a branch, which is the stronger of the two.

587/587 in `MeshWeaver.Hosting.Test`, 373/373 in `MeshWeaver.Documentation.Test`, 15/15 on the reader half.

PR #3964, opened 2026-09-10
fix(ci): explain divergent module hazard accurately

### LINKED ISSUE #3962: The divergent-build guard's stated reason stops being true under #3934's floor — keep the refusal, change the sentence
Follow-up to #3934 / #3946, raised because a control's stated reason stops being true the moment #3946 ships — and a control whose cited consequence has been fixed elsewhere reads as stale to the next person who checks it.

## The guard, verbatim

`.github/workflows/node-repo-gate.yml:1102` and `node-repo-publish-bake.yml:1147`, the `ext-modules` composition step:

```
::error::$dupe reaches this mesh as more than one build — the loader keeps whichever it
sees first, and the other's NodeTypes are declined at adoption (MeshWeaver#3732).
```

Two clauses. **The first survives #3946 untouched. The second does not.**

* *"the loader keeps whichever it sees first"* — unchanged. #3946 does not alter which copy `Assembly.LoadFrom` returns, nor which one the bake host stamps records against.
* *"and the other's NodeTypes are declined at adoption"* — this is precisely the decline #3946 removes, **conditionally**: it is removed exactly when both copies report the same `AssemblyInformationalVersion`, which is the same-publication / same-platform-pin case (the only case the guard has ever actually fired on). It is **not** removed when the versions genuinely differ — a floor still declines a live copy below it.

So the guard's own sentence now names a consequence that, for the case it fires on, no longer happens. That is the shape recorded on `Doc/Architecture/ControlsThatCannotFail` from the other direction: a control asserting a platform ABSENCE dies when the platform improves, and it then reports the improvement as your defect.

## Position: the guard should STAY, and its message should change

Not a weakening. Three reasons the refusal is still right, none of which is the record:

1. **The loader coin toss is a behaviour hazard, not only a record hazard.** `Doc/Architecture/ModuleOwnedSiblingsRide` → "The invariant that replaces it": two copies under one simple name are ONE assembly identity, the loser's bytes are never in memory, and *whichever copy loads first wins the process*. If the two builds differ in anything but their MVID — a partially-updated lane, a half-landed source change, two different commits reaching two `build-workspace` calls — the process silently runs one and every caller compiled against the other gets behaviour it was not built against. The dependency record never *detected* that; it was a **proxy** that happened to fire on the same input. #3946 stops the proxy firing, which makes the composed-set assertion MORE load-bearing, not less.

2. **This is the only reading taken where every copy is in one hand.** A consumer holds one copy and cannot see a second by construction. `PublishedBundleCatalogue.ArtifactsForIdentity` / `ReleaseAvailability` can see both, and #3946 deliberately kept that check firing under floors (`CompiledDependencies.IsModuleLaneId`, with a pinned control in `SealedSetConsistencyTest`) — but both are downstream of the seal, and a torn publication that reaches the shelf has already cost something.

3. **It is stated policy, independent of the mechanism.** `Doc/Architecture/ModuleAdoptionPolicy` → "What the platform roll gates on": *"two builds of one platform assembly in one identity is a torn publication, and torn publications are refused whole."*

And the practical argument: the producer remedy is already in flight (#3933 / Plugins#1614 — one compilation per publication; Plugins#1609 already took this assembly from three builds to two). Relaxing the guard now would remove the pressure to land it and would hide the *next* divergence, which may not be the benign same-version kind.

## The change this issue asks for

**Message only, plus the doc line it quotes.** Replace the decline clause with the harm that actually survives:

```
::error::$dupe reaches this mesh as more than one build — the loader keeps whichever it
sees first and the other copy's bytes are never in memory, so callers compiled against
the loser silently run the winner (MeshWeaver#3732). Since MeshWeaver#3934 a dependency
record states a version FLOOR, so this no longer shows up as a NodeType decline — which
is why it is refused here, at the one place every copy is in one hand.
```

Same for the sibling in `node-repo-gate.yml`, and the corresponding sentence in `ModuleOwnedSiblingsRide` → "The invariant that replaces it", which currently ends on the decline quote.

Deliberately **not** proposed here: narrowing the refusal to "only when the copies differ in a way the floor cannot absorb". That trades a definite refusal for a prediction about the consumer, on the fleet's upstream, and the copies-differ case is exactly where the loader hazard is real.

## What this does NOT fix, said plainly

**#3946 does not unred core CD.** Its diff touches zero files under `.github/` (verified: `git diff --name-only origin/main...` returns none). The guard refuses to **SEAL**, and publication is all-or-nothing, so no bundle is published at all — a consumer's adoption policy is irrelevant to a publication that does not exist. Anyone reading "floor, not pin" and expecting the red to clear is reading the consumer half as if it were the producer half. What unreds CD is the second compilation going away (#3933 / Plugins#1614).

rbuergi:
## What

- Keep the divergent-build refusal in both reusable publication lanes.
- Replace the stale “the other NodeTypes are declined” claim with the surviving loader hazard: only the first build enters memory, so callers compiled against the loser silently run the winner.
- Explain why version-floor adoption does not make two builds safe.
- Strengthen the real workflow regression so both lanes must emit the corrected explanation.

## Why

MeshWeaver#3934 changed module dependency records from exact MVID pins to version floors. Same-version duplicate builds can now remain adoptable, so the old diagnostic described a consequence that no longer necessarily occurs. The load-order hazard remains, and this is the only stage where every competing copy is visible.

## Verification

- python3 .github/scripts/test-module-set-consistency.py: 2 lanes × 7 cases passed, including both falsification arms
- workflow YAML-key guard: 34 workflows checked, 0 violations
- workflow timeout guard: 84 jobs checked, 0 violations
- actionlint YAML/expression validation: success
- MeshWeaver.Documentation Release build with warnings as errors: 0 warnings, 0 errors
- MeshWeaver.Documentation.Test Release build with warnings as errors: 0 warnings, 0 errors
- MeshWeaver.Documentation.Test: 374 passed, 0 failed, fresh TRX

Fixes #3962.

### DIFF
--- a/.github/scripts/test-module-set-consistency.py
+++ b/.github/scripts/test-module-set-consistency.py
@@ -210,6 +210,8 @@ def case_divergent_is_refused(lane: str, body: str, tmp: Path) -> None:
     combined = result.stdout + result.stderr
     missing = [needle for needle in
                ("MeshWeaver.Shared reaches this mesh as more than one build",
+                "other copy's bytes never enter memory",
+                "A version floor can keep the affected NodeTypes adoptable",
                 "as a sibling riding",
                 "1 carried at more than one BUILD")
                if needle not in combined]
--- a/.github/workflows/node-repo-gate.yml
+++ b/.github/workflows/node-repo-gate.yml
@@ -1102,7 +1102,7 @@ jobs:
               echo "::error::$dupe reaches this mesh as more than one build — the loader keeps whichever it sees first and the other copy's bytes never enter memory, so callers compiled against the loser silently run the winner (MeshWeaver#3732)."
               awk -v n="$dupe" '$1 == n {printf "          sealed %s %s %s\n", substr($2, 1, 16), ($4 == "declared" ? "as the declared module of" : "as a sibling riding"), $3}' "$DIGESTS" | sort -u
             done < <(awk '{if (!seen[$1 SUBSEP $2]++) d[$1]++} END {for (k in d) if (d[k] > 1) print k}' "$DIGESTS" | sort)
-            echo "::error::the composed module set carries $DIVERGED assembly name(s) at more than one build. A version floor can keep the affected NodeTypes adoptable (MeshWeaver#3934), but cannot make the loader use both builds or prove their behavior agrees. Refusing here is the only point where every copy is visible. Build the whole set in ONE compilation, or stop the divergent copy riding — Doc/Architecture/ModuleOwnedSiblingsRide."
+            echo "::error::the composed module set carries $DIVERGED assembly name(s) at more than one build. A version floor can keep the affected NodeTypes adoptable (MeshWeaver#3934), but cannot make the loader use both builds or prove their behaviour agrees. Refusing here is the only point where every copy is visible. Build the whole set in ONE compilation, or stop the divergent copy riding — Doc/Architecture/ModuleOwnedSiblingsRide."
             exit 1
           fi
           echo "external modules: $MODULE_COUNT composed, $ASSET_MODULES of them carrying static web assets, $ASSET_FILES asset file(s) landed module-relative under /ext/<module>/"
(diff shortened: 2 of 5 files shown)


rbuergi:
Closing this duplicate draft: #3973 merged the same #3962 root fix, including the surviving loader-hazard wording, enforcement coverage, green CI, and the completed issue. No unique change remains here to ship.

Issue #3934, opened 2026-09-10
Dependency records must express a FLOOR, not an MVID pin — a module rebuild must never trigger a runtime recompile


rbuergi:
## The position

Measured on memex today, and stated as a design decision by the platform owner:

1. **There is no incompatibility.** Two builds of one module assembly link fine — `AssemblyVersion` is synchronised fleet-wide, so `Assembly.LoadFrom` returns the already-loaded copy and ignores the second path. Nothing throws.
2. **It should be a floor, not a pin.** A record should say "I need at least X" and accept anything at or above.
3. **It must never recompile.** A moved MVID is not a reason to rebuild a NodeType at runtime.
4. **It must not crash.** If a floor genuinely is not met, that must be visible — never a silent fallback that drops the type's views.

## What the code does today

`CompiledDependencies.CreateIdResolver`:

```csharp
if (moduleMvidByName.TryGetValue(name, out var moduleMvid))
    return MvidScheme + moduleMvid;   // a MODULE  → exact build, checked FIRST
if (!name.StartsWith("MeshWeaver.", StringComparison.Ordinal))
    return null;
if (surfaceByName.TryGetValue(name, out var surface))
    return RefAsmScheme + surface;    // a PLATFORM assembly → API surface hash
```

A platform assembly is compared by **API surface**, so a rebuild with an unchanged API compares equal and adopts. A module is compared by **raw MVID**, so a rebuild of *identical source* compares unequal and every NodeType binding it is declined. The MVID changes on every compilation by construction.

`ModuleOwnedSiblingsRide` already concedes the point: *"Module-vs-module is `mvid:` on both sides. Identical bytes compare equal. The conflict is fixable by agreement"* — and then makes byte-identity the remedy. That is treating the symptom: it requires every producer of every copy to agree byte-for-byte, forever.

## What it costs, measured 2026-09-10

`SocialMedia/Post`, `PostsHub` and `Profile` are the only types calling `.AddComments()`, so they are the only ones binding a module (`MeshWeaver.Markdown.Collaboration`). All three were declined at adoption; `/Posts`, `/Posts/*` and `/Profiles/*` returned HTTP 200 and rendered nothing for the whole day. Every other NodeType on the mesh binds platform assemblies only, resolves `ref:`, and was unaffected throughout.

The recompile loop is directly observable on the node's own fields:

```
[...]

Issue #3244, opened 2026-09-03
One producer in TIME: the registry's content-versioned module bytes can disagree with the sealed publication's, and the set check cannot see it


rbuergi:
## The residual from #3175: ONE producer in TIME, not just in the set

#3175 is closed by #3242 — the availability gate now confirms the sealed set is mutually
CONSISTENT: every dependency record's `mvid:` matches the module sealed for the same identity, and
no assembly name appears in the set at two MVIDs, riding sibling copies included (#3221).

**That asserts consistency WITHIN one sealed publication. It cannot see across two producers in
time**, and `ModuleBuildArchitecture` → "What the gate still cannot see" records why:

- core CD's `plugins-modules` rebuilds `MeshWeaver.AI` / `MeshWeaver.Markdown.Collaboration` per
  platform release to feed the Plugins seal (mvid A);
- the registry's package endpoint (`/api/plugins/bundles/<pkg>/<version>`) is content-versioned
  (Plugins#931) and serves the Plugins lane's last publication (mvid B) — **the bytes a portal
  actually installs**.

Both sets can be internally consistent and still disagree with each other. The sealed-set check
passes; the instance still logs `dependency record mismatch … live is mvid:` at adoption. That is
the gap, not a new defect.

## Two ways to close it (either, not both)

1. **The seal composes the registry's bytes** for the root source (`registry-modules` in core CD).
   Blocked today: core holds no registry credential — `MW_REGISTRY_KEY` is not among its secrets.
2. **The instance adopts module bytes for its identity from the sealed publication** it already
   adopts bundles from, instead of from the registry's content-versioned endpoint. This needs no
   new credential and is probably the cheaper half.

This is fix-shape part 2 of #3175 ("fail the SEAL, not the roll"), which the maintainer's own
comment there called *"the right end state … but an optimisation and a stronger guarantee, not a
prerequisite"* for the set check that has now landed.

## Also worth doing while here

Record the composed upstream MVIDs in the bundle manifest at bake time, and have the registry refuse
a publication whose dependency records point at upstream bytes it does not serve for that identity.
Then the set is consistent BY CONSTRUCTION and #3242's check becomes a string compare instead of
`O(bundles)` of PE reads on every availability poll.

Refs: #3175, #3221, #3242, Plugins#931, `Doc/Architecture/ModuleBuildArchitecture`,
`Doc/Architecture/ModuleOwnedSiblingsRide`.

rbuergi:
[...]

Issue #1751, opened 2026-08-17
A module must always carry a resolvable link to its shipped DLL (per framework identity, per arch) — routable via the registry, so nothing compiles


rbuergi:
Maintainer-directed. Several statements, one design:

> *"modules should always have link to dll shipped. can route via memex.meshweaver.cloud"*
> *"it must receive all the pre-built releases (we can even ship release nodes)"*
> *"then no compile should be needed. just give path to correct arch"*
> *"we should probably disable any pre-bake on instance level and just do it dynamically when user reaches place"*

## What already exists (build on it, don't reinvent)

`PluginBundleEndpoints` (`/api/plugins/bundles`) already serves:
- `index.json` — a top-level `frameworkMvid` plus, per installed package, `{plugin, version, url}` where **`url` is absolute and built from the REQUEST host**, so an instance behind an ingress or a custom domain advertises a reachable address.
- `/{plugin}/{version}` — the bundle itself.

`PluginBundleClient.Adopt` consumes that, comparing the framework identity ONCE before downloading anything. Since #1719 each assembly also carries a per-type dependency record, validated at adopt time.

So the transport exists and works — prod boot measured `compiled=0 alreadyBaked=84` once the four node repos published under the live identity.

## The gaps

1. **The link is an index you poll, not a property of the module.** A module/package does not itself record where its shipped DLL lives. The maintainer's ask is that it *always* does — so resolution is "read the module, follow its link", not "ask a registry what it happens to have installed". Note the mesh already models releases as nodes (`<Package>/Release/<stamp>`, nodeType `Release`) — shipping a module's release AS a node, carrying the identity and the assembly link(s), is the maintainer's own suggestion and fits what is already there.
2. **The index only advertises what THIS instance has installed** (`InstalledPackages`). A module that exists but is not installed here is unreachable through it.
3. **No arch dimension.** Native payloads are the open blocker (#1728): the closure lane prunes `modules/<Name>/runtimes/`, and `Assembly.LoadFrom` never reads a module's deps.json, so a RID tree is unreachable *by the current loader*. The maintainer's direction is explicitly **not** "modules cannot carry natives" but **"just give path to correct arch"** — i.e. the link resolves per-RID and the loader is pointed at it. That reframes #1728 from a structural impossibility into a resolution problem.
[...]

PR #4002, opened 2026-09-11
fix(#3768): the bundle index answers the CALLER's lane, served from the publication sealed for it

### LINKED ISSUE #3768: Registry hands an s414bfb2 consumer bundles baked for s72c27af — 25 FrameworkDeclined adoption misses on memex@8131 while the consumer's identity is sealed on both shares
**Measured 2026-09-09 00:2xZ on memex.systemorph.com (3.0.0-ci.8131, live framework `s414bfb274aa125ad3e02d8d33c62bec3`), `/health`:**

```
bundle_adoption: Degraded — 25 adoption attempt(s), 0 assembly/assemblies adopted, 25 MISS(es) — content the registry was meant to serve is compiled here instead: AI: FrameworkDeclined (built against framework s72c27afab89c100727e79e0559c59e32, live framework is s414bfb274aa125ad3e02d8d33c62bec3); Anthropic: FrameworkDeclined (…); AppleIntelligence, Maps, AzureBlob, AzureFoundry, Chat, Export, Mcp, Notifications, …(+15): same
```

At the same time the registry share (memex.meshweaver.cloud, `/data/prebuilt-bundles`, read through the memex API) holds the consumer's identity sealed for every source:

```
s414bfb274aa125ad3e02d8d33c62bec3 09-08 18:24Z: crm(sealed 18:57Z), education(sealed 19:23Z), meshweaver-content(sealed 18:24Z), plugins(sealed 23:34Z), reinsurance(sealed 19:18Z)
```

and memex's own share holds the same identity's `plugins` publication (55 bundles; `ShippedPrebuiltBundles … A bundle entry under /data/prebuilt-bundles/s414bfb274aa125ad3e02d8d33c62bec3 DID name 1 of them`).

So the registry (itself still on 3.0.0-ci.8079 = `s72c27af…`) hands an `s414bfb2…` consumer bundles baked for `s72c27af…`, every one is declined by the identity rule, and 25 packages compile locally instead of adopting. Not user-visible (the content compiles), but it is exactly the situation `Modules:VersionStrictness` (#3727) and the per-identity seal exist to avoid, and it costs a full local compile of every registry package on every boot.

**Hypotheses to measure, in order:** (1) the registry's catalog (`GET /api/plugins`, `RegistryPackageSource`) selects the publication by the REGISTRY's own identity rather than the caller's (the caller's identity travels on the request? check `InstanceRegistryAuthenticator` / the catalog handler); (2) the consumer's `RegistryUpdateReconciler` asks with `ref=HEAD` and takes whatever the index names without stating its identity; (3) the `plugins` seal for `s414bfb2` on the registry share landed at 23:34Z, after the consumer's last catalog read — a stale catalog cache on the consumer (the health check counts attempts since boot at ~23:30Z).

Related: #3727 (Family strictness), #3732 (downstream republish), #3651 (roll gate on measured loadability). Evidence scripts (read-only, on the meshes): `rbuergi/Script/list-prebuilt-identities` (registry), `rbuergi/Script/loki-memex-crm-compile` (consumer Loki).

rbuergi:
Fixes #3768.

## It still reproduces — measured today, on code that already carries #3946

`https://memex.systemorph.com/health`, 2026-09-11 ~05:35Z, image `3.0.0+45306a33e` (which **does**
contain #3934/#3946 — verified by `git merge-base --is-ancestor 26bf0a60c 45306a33e`):

```
bundle_adoption: Degraded — 25 adoption attempt(s), 0 assembly/assemblies adopted, 25 MISS(es)
  AI: FrameworkDeclined (built against framework s01d65c9542e19652c09f851ffb6bf92a,
                         live framework is s3e3c80238a740a8cb895a72b0bfb9cd6)
  …25 of them, the same pair
```

Same count, same kind, a **different identity pair** from the one this issue was filed on. #3946
changed a different comparison (a dependency record's module entry, `CompiledDependencies`); this
one is `PrebuiltAssemblySeeder.DeclineReason` against the INDEX's single top-level stamp, and it is
untouched by it — correctly, and it stays untouched here too.

**Denominator:** 25 is the whole population of that process's adoption attempts, not a truncation —
`BundleAdoptionLedger.Capacity` is 500 and the payload names ten then counts the rest. The
installation holds 63 `Package` records and its seed ledger lists 61; 25 is what the default-install
lane processed on that boot, and 0 of 25 adopted.

## What decides which bundles a consumer is handed

`Index()` stamped `frameworkMvid = FrameworkMvid` — *this portal's own bake* — and
`PluginBundleClient.Adopt` compares that ONCE and declines the whole index before requesting any
package. `DownloadOverHttp` has stated the consumer's lane since #1751 and `ServedModuleBytes` has
read the sealed publication for it since #3244, but neither is reachable: `FetchIndex` asked with no
lane at all, so the registry answered about itself and the consumer walked away.

## Why the issue's own proposed fix is not sufficient on its own

A registry cannot answer an off-lane caller out of its **mesh**. A `ReleaseArtifact` is minted in
exactly one place — a compile, in that mesh, stamping the compiling process's own identity
(`NodeTypeBuildState.TryCreateReleaseNode`); adoption mints none (`PrebuiltAssemblySeeder.Seed`
writes `LastCompiledVersion` only); and an in-portal compile writes `collection: "local"`, which
`FileSystemAssemblyStore` defines as pod-local. Measured on memex-cloud 2026-09-11:
`Store/Plugin/Release/20260911011107-Br61zt34` **does** name the consumer's identity
`s3e3c80238a740a8cb895a72b0bfb9cd6` — and its artifact is
`collection: "local", contentPath: "Store_Plugin/v13725-s3e3c802-…dll"`, written by a pod that has
been replaced. So stamping the index alone would turn one cheap named decline into N downloads that
each resolve nothing.

The durable cross-lane source is the publication the bake sealed, on the share the registry already
mounts. Measured the same morning: 561 identity directories, with
`s3e3c80238a740a8cb895a72b0bfb9cd6 09-11 00:13Z: crm(sealed 00:38Z), meshweaver-content(sealed
00:13Z), plugins(sealed 00:33Z)`. **The bytes were on the registry's own disk while every adoption
was declined.**

## What changed

1. `PluginBundleClient.FetchIndex` states this instance's lane (`?identity=&arch=`), exactly as
   `DownloadOverHttp` already does.
2. `Index` answers for the requested lane **when this registry holds a sealed publication for it**,
   and its own identity when it does not — so a bake gap keeps reading as a bake gap. A caller that
   states no lane (every already-deployed consumer) gets today's answer byte for byte.
3. `Assemble`, off-lane, serves that publication's bundle — its assemblies, dependency records and
   source fingerprints (#2813) — with the module section composed as on every other serve. New
   `SealedLaneBundles` is the lookup, reading through `PublishedBundleCatalogue.SealedPublicationOf`
   so a torn publication is never served and the generation pointer (#3461) is honoured.

**No gate is widened, no retry, no fallback, no catch.** The served archive states the identity it
was sealed for, and `PrebuiltAssemblySeeder.DeclineReason` (ordinal equality) still decides — first
for the archive as a whole, then per assembly as it seeds.

## Control, both directions

`test/Memex.Portal.Shared.Test/PluginBundleSealedLaneTest.cs`, 3 tests, asking the SAME endpoint
with three different asks:

- **positive** — `ACallerWhoseLaneIsSealedHere_IsToldThatLane_AndServedThePublicationsBytes`:
  `Assert.Equal(ConsumerIdentity, index.GetProperty("frameworkMvid").GetString());` and, on the
  bundle, `Assert.Equal(ConsumerIdentity, manifest?.FrameworkMvid);` /
  `Assert.Equal(NodeTypePath, Assert.Single(assemblies).NodePath);` /
  `Assert.Equal("fp-sealed", Assert.Single(assemblies).SourceFingerprint);`
- **negative** — `ACallerWhoseLaneIsNotSealedHere_IsToldThisRegistrysOwnLane_AndGetsNoOtherLanesBytes`:
  `Assert.Equal(PrebuiltAssemblySeeder.LiveFrameworkMvid, index.GetProperty("frameworkMvid").GetString());`
  `Assert.NotEqual(UnsealedIdentity, …);` then `Assert.NotEqual(ConsumerIdentity, manifest?.FrameworkMvid);`
  and `Assert.Empty(assemblies);` — a publication for a different identity is present on the same
  root and must not be substituted.
- **unchanged shape** — `ACallerThatStatesNoLane_IsAnsweredWithThisRegistrysOwnIdentity`.

**Proven falsifiable** rather than assumed: with the positive expectation replaced by a probe the
run went red with `Expected: "FALSIFICATION-PROBE" / Actual: "s11111111111111111111111111111111"`,
i.e. the index really returned the caller's lane. `DocumentationLinkIntegrityTest` was falsified the
same way (a deliberate broken link reddened it) before being run green.

## Verification

- `dotnet build -c Release -warnaserror`: `MeshWeaver.PluginCatalog`, `Memex.Portal.Shared`,
  `Memex.Portal.Shared.Test`, `MeshWeaver.Documentation.Test`, `MeshWeaver.Compiler.Pipeline.Test`,
  `tools/MeshWeaver.PluginTester` — each `0 Error(s)`, `0 Warning(s)`.
- `Memex.Portal.Shared.Test` in full: **1275 passed, 0 failed** (2 m 52 s) — no regression in
  `PrebuiltPublicationTest`, `SealedModuleCompositionTest`, `PluginBundleIndexArtifactTest`,
  `PluginBundleArtifactFetchTest` or `PluginBundleLaneResolutionTest`.
- `MeshWeaver.Documentation.Test` in full: **374 passed, 0 failed**.

## Work product

`Doc/Architecture/SealedPublicationReads` is EXTENDED (not a fifth page): "The four consumers" now
names the registry serving an off-lane caller, with the `ReleaseArtifact`/`collection: "local"`
finding, both live measurements with their denominators, and the three-rule statement of what the
index stamp means. Plus a What's New entry (`Category: Fix`).

Pairs-with: none — no public top-level type or member is removed from `src/`; the change adds
`SealedLaneBundles`/`SealedLaneBundle` and edits private handlers.

### DIFF
--- a/memex/Memex.Portal.Shared/Api/PluginBundleEndpoints.cs
+++ b/memex/Memex.Portal.Shared/Api/PluginBundleEndpoints.cs
@@ -138,8 +138,18 @@ public static class PluginBundleEndpoints
         // IMessageHub makes an UNAUTHENTICATED request depend on the mesh being resolvable — it
         // throws (500) instead of the 401 the filter would have returned. The rejection path must
         // not need anything but the header.
+        // 🚨 `identity`/`arch` are the CONSUMER's lane here for exactly the reason they are on the
+        // bundle route below (#1751) — and until #3768 the index did NOT read them, which made that
+        // route's lane-awareness unreachable: the index announced the registry's OWN bake, and
+        // `PluginBundleClient.Adopt` declines the whole index on a mismatch before it ever asks for
+        // a package. Absent ⇒ this instance's own lane, which is what every pre-#3768 client asks
+        // for, so the answer for them is unchanged.
         group.MapGet("/index.json", (HttpContext http, CancellationToken ct) =>
-            Index(http, RootHub(http), Caller(http), ct));
+            Index(
+                http, RootHub(http),
+                Requested(http, "identity", FrameworkMvid),
+                Requested(http, "arch", ReleaseArchitecture.Live),
+                Caller(http), ct));
 
         // 🚨 `identity`/`arch` are the CONSUMER's lane (#1751), not a filter the caller invents: they
         // say which framework build identity and which architecture the caller can actually run, and
@@ -895,10 +905,23 @@ public static class PluginBundleEndpoints
     /// <summary>
     /// What this instance can serve, and the framework identity it serves it for.
     ///
-    /// <para>The framework MVID is at the TOP, not per-bundle: every assembly here was produced by
-    /// this portal's own bake, so they all share it. A consumer compares it once and skips the whole
-    /// fetch when it does not match — downloading bundles it is then obliged to decline is pure
-    /// waste, and the decline itself is silent (see <see cref="PrebuiltAssemblySeeder"/>).</para>
+    /// <para>The framework MVID is at the TOP, not per-bundle: every bundle listed here is resolved
+    /// for ONE lane, so a consumer compares it once and skips the whole fetch when it does not
+    /// match — downloading bundles it is then obliged to decline is pure waste, and the decline
+    /// itself is silent (see <see cref="PrebuiltAssemblySeeder"/>).</para>
+    ///
+    /// <para>🚨 <b>That lane is the one the CALLER asked for, not this portal's own bake</b>
+    /// (#3768). It used to be <see cref="FrameworkMvid"/> unconditionally, on the premise that
+    /// "every assembly here was produced by this portal's own bake" — a premise #1751 retired when
+    /// the bundle route learned to resolve another lane, and #3244 retired again for the module
+    /// half. The consequence was measured on 2026-09-09 and again on 2026-09-11: a consumer one
+    /// image ahead of its registry declined the whole index — 25 attempts, 0 adopted, 25
+    /// <c>FrameworkDeclined</c> — while the publication sealed for its identity sat on the very
+    /// share this registry mounts. The stamp now names the lane the answer is FOR, which is the
+    /// same value each served bundle's own manifest carries, and it is claimed only when this
+    /// registry HOLDS a sealed publication for it: with none, the caller is told this portal's own
+    /// identity and declines exactly as before, because a bake that does not exist here is a bake
+    /// gap and must not read as a serving offer.</para>
     ///
     /// <para>Absolute URLs are built from the REQUEST rather than configuration: an instance reached
     /// through an ingress, a port-forward or a custom domain must advertise the host the caller
@@ -917,10 +940,21 @@ public static class PluginBundleEndpoints
     /// download route records every decision it makes — one per actual fetch.</para>
     /// </summary>
     private static Task<IResult> Index(
-        HttpContext http, IMessageHub rootHub, AuthenticatedInstance? caller, CancellationToken ct)
+        HttpContext http, IMessageHub rootHub, string identity, string architecture,
+        AuthenticatedInstance? caller, CancellationToken ct)
     {
         var baseUrl = $"{http.Request.Scheme}://{http.Request.Host}{RoutePrefix}";
         var ledger = Ledger(http);
+        // The lane this answer is FOR. On this instance's own lane it is FrameworkMvid and every
+        // bundle resolves from LastCompiledVersion, exactly as before. Off it, the claim is only
+        // made when a sealed publication for that identity is on the share this registry mounts —
+        // the one durable cross-lane source it has (SealedLaneBundles).
+        var servesOwnLane =
+            string.Equals(identity, FrameworkMvid, StringComparison.Ordinal)
+            && ReleaseArchitecture.Matches(ReleaseArchitecture.Live, architecture);
+        var servedLane = servesOwnLane
+            ? new ServedLane(FrameworkMvid, ReleaseArchitecture.Live)
+            : ServableLane(http, identity, architecture, caller);
         // Resolved NOW, while the request scope is alive: a late fault arrives after
         // HttpContext.RequestServices has been disposed, so resolving inside the lambda would throw
         // exactly when the report is needed.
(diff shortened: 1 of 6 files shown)


rbuergi:
## All six review findings addressed in `cacd18d` — four were real, and one was the safety property itself

The PR had already gone green on `f48f7a7`, and that sha carried the traversal. I dequeued it rather
than let auto-merge land it.

**1 + 4 — path traversal (`identity` → `Path.Combine`).** Correct, and NEW: the identity used to be
compared as an opaque string and nothing else; since this change it selects a directory. Both routes
now answer `400` for a stated identity that is not a bare name — the rule and the answer the
`/prebuilt/{identity}/{source}` segments already applied — and `SealedLaneBundles.IsBareName`
restates it *at the layer that composes the path*, exactly as the review asked, so a second entry
point added later inherits the check rather than the hole.

**2 + 5 — the (source, package) grant boundary.** Correct. `Locate` now takes the source the
entitlement decision resolved (`decision.Source ?? asked.Source` — the anchor before the cached
stamp) and is restricted to it, case-insensitively because the publication prefix is lowercase while
a registry source name need not be. Only a genuinely unknown binding widens the scan; a named source
that matches no directory answers "nothing sealed for you here" and the caller compiles, which is
what it did before this lookup existed.

**3 — the archive's own identity was read and never validated.** This one is the whole safety
argument and the review is exactly right: `BuildResult` restamps the outgoing manifest with the
requested identity, so a mislabelled archive would have reached the consumer wearing a manifest that
agrees with its live framework. `ReadSealed` now refuses unless the archive's own claim passes
`PrebuiltAssemblySeeder.DeclineReason` against the requested identity — the same function the
consumer applies to the same bytes, so producer and gate cannot disagree. The directory a bundle is
filed under is a filing convention; only the manifest's claim may be believed.

**6 — synchronous ZIP read on the request thread.** Taken. The open/parse/inflate now runs on the
filesystem `IIoPool` (`IoPoolNames.FileSystem`), composed into the observable, as
`PublishedBundleCatalogue.Observe` does for the same read.

### Controls

9 now (3 → 9), and each new one **proven falsifiable**, not assumed:

- disabling the archive-identity check reddens
  `ABundleFiledUnderALaneButBuiltForAnotherIsRefused_NeverRestamped`;
- ignoring the source scoping reddens
  `TheBundleComesFromTheSourceTheEntitlementResolved_NotWhicheverSortsFirst`
  (the decoy source is named `education` so it sorts *before* `plugins` — the old lookup would have
  served it);
- those two only; nothing else moved.

`AnIdentityThatIsNotABareNameIsRefused_OnBothRoutes` is a `[Theory]` over `../../etc`, `..`, `a/b`,
`c:evil`, asserting `400` on the index and the bundle route.

`Memex.Portal.Shared.Test`: **1281 passed, 0 failed**. `MeshWeaver.Documentation.Test`: **374 passed,
0 failed**. Every touched project builds `-c Release -warnaserror` at `0 Error(s) / 0 Warning(s)`.

`Doc/Architecture/SealedPublicationReads` gains "Three refusals the lookup carries, and why each is
not optional", so the next reader finds the reasoning rather than the checks alone.

PR #4012, opened 2026-09-11
YamlDotNet 16 -> 18: two majors, verified against 1,817 authored front matters

rbuergi:
Moves `YamlDotNet` from **16.3.0 to 18.1.0** — two major boundaries — under the standing directive to run the latest version of every library. Majors were held back because the directive says "at least within minor"; this one is explicitly authorised.

`Directory.Packages.props` is the only version change. No source file needed adapting.

## First, the premise — measured, not inherited

The target version came from an earlier session's analysis, so it was re-measured against the live registry, with every read gated on **HTTP status AND byte count** (a failed request's empty body greps to zero and reads exactly like "no such version").

| question | measurement |
|---|---|
| Does an 18.x exist, and what is newest? | `https://api.nuget.org/v3-flatcontainer/yamldotnet/index.json` → HTTP **200**, **5,860 B**. 270 versions; newest stable **18.1.0** (18.0.0 and 18.1.0 are the only 18.x). |
| Licence at the new major? | **MIT, unchanged.** `<license type="expression">MIT</license>` in both the 16.3.0 nuspec (HTTP 200, 1,337 B) and the 18.1.0 nuspec (HTTP 200, 1,338 B). No licence-terms change of the kind holding the Aspire family — nothing to escalate. |

The local `Dependency licences` gate agrees: `Licence gate passed: 114 shipping packages checked`.

## What actually changed across the two boundaries

Rather than trusting release notes, both `.nupkg`s were downloaded and the **public API of the real assemblies** enumerated with `MetadataLoadContext` and diffed (version stamps normalised out, or they dominate the diff).

**Six public members are removed across both majors**, and that is the whole removal set:

- `Accepts(Type)` on the five built-in date/time converters — refactored onto the new `ScalarConverterBase<T>`, so still present by inheritance.
- `YamlDotNet.Core.Tokens.Scalar.IsKey`.

`IYamlTypeConverter` is **byte-identical** between 16.3.0 and 18.1.0.

Our entire usage across both repositories is six symbols — `DeserializerBuilder`, `SerializerBuilder`, `IDeserializer`, `ISerializer`, `DefaultValuesHandling`, `YamlException` — all present in both, and there are **zero** custom `ITypeInspector` / `IYamlTypeConverter` / `INodeDeserializer` / `TypeInspectorSkeleton` implementations anywhere.

**The two declared breaking changes both miss us, and both would have been invisible in a build:**

- **18.0.0 added `HasParseMethod` and `Parse` to `ITypeInspector`** with no default implementation. That breaks *implementers*, not callers — and neither repository implements it.
- **18.1.0 introduced a default maximum YAML nesting depth of 130.** Front matter is a flat map plus the occasional string list — depth 2 to 3. Real behaviour change, distant failure mode, `WithMaximumRecursion(n)` is the lever.
- Checked and not reaching us: **17.1.0** capped `MergingParser` at 100k events. Neither repository constructs one.

## The risk that is not in the compiler: 1,817 authored front matters

A green build proves very little here. Nothing in `src/` calls YamlDotNet in an interesting way — the blast radius is **authored content**: every `.md` node's front matter is read by this parser, and nothing compiles that data, so nothing fails if it starts parsing differently.

So the same probe program was built **twice** — identical source, two project files differing only in the pin — configured exactly as `MarkdownFileParser` configures the deserializer and serializer, and run over every `.md` in core and MeshWeaver.Plugins:

**1,940 files, 1,817 with front matter, 0 differing lines.** Identical parsed-value hashes, identical re-serialised hashes, and the same 18 files raising the same `SemanticErrorException` on both sides (files whose leading `---` is a horizontal rule — the parser behaves identically, and `MarkdownFileParser` already has its documented defensive fallback for them).

**The instrument was controlled, because a probe that cannot detect a difference reports zero for the wrong reason:**

- *Negative control* — 1,798 distinct hashes among the 1,799 OK rows, so the hash discriminates.
- *Positive control* — mutating a single cell of one output made the diff fire.

In-mesh source was swept separately: **no `.cs` stored in a mesh node, and no NodeType `configuration` lambda, references YamlDotNet at all** — the only matches under `samples/` and `src/` were `obj/` build artefacts.

## Cross-repo coupling — declared, though no gate asks for it

`MeshWeaver.Plugins` has **no `Directory.Packages.props` of its own**; its `src/Directory.Packages.props` imports this repo's at `MW_PLATFORM_REF`, and `src/MeshWeaver.AI` (skill/agent front matter) plus `src/MeshWeaver.Publish` (slide front matter) carry **versionless** `YamlDotNet` references. So this one line decides what the AI engine's front-matter parser is built from.

The `Satellite package pins` gate does **not** fire here — it covers a pin being *removed*, not bumped — so this is declared voluntarily rather than to satisfy a check. Both satellite consumers were included in the 1,817-file corpus run, and neither implements an extension point that 18.0.0 broke. The satellite picks the change up at its next pin move; nothing is required of it in the meantime.

No public type or member is removed, no i18n catalog key is added or re-worded, and no `PackageVersion` is removed — so `Pairs-with`, `Implementers`, `Mirror-sync` and `Satellite-pins` do not apply. All three body-reading gates were run locally against this body with `--pr-body-file` rather than by burning a CI round.

## Verification

- `dotnet build src/MeshWeaver.Hosting/MeshWeaver.Hosting.csproj -c Release -warnaserror` → **0 Error(s), 0 Warning(s)** (20.07 s)
- `dotnet build test/MeshWeaver.Hosting.Test/... -c Release -warnaserror` → **0 Error(s), 0 Warning(s)**
- `dotnet build test/MeshWeaver.Documentation.Test/... -c Release -warnaserror` → **0 Error(s), 0 Warning(s)** — rebuilt after the doc edits, since docs are embedded resources and `--no-build` would have been a false pass
- The bump was confirmed to reach the binaries, not just the props file: the `YamlDotNet.dll` copied beside both test assemblies reports **18.1.0**
- Full `MeshWeaver.Hosting.Test` → **608 passed, 0 failed**
- Full `MeshWeaver.Documentation.Test` → **390 passed, 0 failed**, including `DocumentationLinkIntegrityTest`, `ArchitectureTopicMapTest`, `WhatsNewEntryIntegrityTest` and `SkillFrontMatterGuard`

## Work product

The method is committed, not just the result: `Doc/Architecture/DependencyMajorUpgrades` records the four things a green `-warnaserror` build cannot see at a major boundary, the differential-corpus technique with its two controls, and the ledger entry for this crossing. The next person bumping a major — `System.Reactive` 6 → 7 is the immediate one — needs that more than they need this diff.

### DIFF
--- a/Directory.Packages.props
+++ b/Directory.Packages.props
@@ -4,10 +4,21 @@
     <NoWarn>$(NoWarn);NU1608</NoWarn>
   </PropertyGroup>
   <ItemGroup>
-    <!-- Used DIRECTLY by MeshWeaver.AI (skill/agent front matter) and MeshWeaver.Hosting
-         (MarkdownFileParser). It used to arrive only TRANSITIVELY via EPS.Extensions.YamlMarkdown,
-         so removing that unused package broke the build - an accidental dependency, now explicit. -->
-    <PackageVersion Include="YamlDotNet" Version="16.3.0" />
+    <!-- Used DIRECTLY by MeshWeaver.Hosting (MarkdownFileParser). It used to arrive only
+         TRANSITIVELY via EPS.Extensions.YamlMarkdown, so removing that unused package broke the
+         build - an accidental dependency, now explicit.
+
+         🚨 THIS ENTRY IS ALSO MeshWeaver.Plugins' VERSION SOURCE, and nothing here can see that.
+         Plugins has no `Directory.Packages.props` of its own; its `src/Directory.Packages.props`
+         IMPORTS this file at `MW_PLATFORM_REF`, and `src/MeshWeaver.AI` (skill/agent front matter)
+         plus `src/MeshWeaver.Publish` (slide front matter) carry VERSIONLESS references. So this
+         line decides what the AI engine's front-matter parser is built from - a satellite the
+         `Satellite package pins` gate does not watch, because that gate fires on a pin being
+         REMOVED, not bumped. The satellite reads the change only when its pin moves.
+
+         16.3.0 -> 18.1.0 crosses two majors; the adaptation and the two behavioural changes that
+         came with them are recorded in Doc/Architecture/DependencyMajorUpgrades. -->
+    <PackageVersion Include="YamlDotNet" Version="18.1.0" />
     <PackageVersion Include="AspNet.Security.OAuth.Apple" Version="10.0.0" />
     <!-- 🚨 THE ASPIRE FAMILY IS HELD AT 13.4.6 ON PURPOSE — do not let a dependency sweep move it.
          Every Aspire.Hosting version depends on JsonPatch.Net; only the VERSION differs, and the
--- a/src/MeshWeaver.Documentation/Data/Architecture.md
+++ b/src/MeshWeaver.Documentation/Data/Architecture.md
@@ -436,6 +436,7 @@ Each theme starts with its introductory page, followed by related architecture t
 ### Licensing
 
 - **Start here:** [Dependency Licensing](DependencyLicensing) — Apache-2.0/MIT compatible only; the CI gate that enforces it
+- [Dependency Major Upgrades](DependencyMajorUpgrades) — the four things a green `-warnaserror` build cannot see at a major boundary (in-mesh source, authored content, the satellites that import this repo's package list, and behaviour behind an unchanged signature), the differential method that replaces them, and the ledger of boundaries actually crossed
 
 ## Getting started
 
--- a/src/MeshWeaver.Documentation/Data/Architecture/DependencyMajorUpgrades.md
+++ b/src/MeshWeaver.Documentation/Data/Architecture/DependencyMajorUpgrades.md
@@ -41,10 +41,22 @@ Never adopt a target version asserted by an earlier session, an issue, or a memo
 and **gate the read on HTTP status AND byte count** — a failed request's empty body greps to zero
 and reads exactly like "no such version".
 
+🚨 **PRINTING the status and the byte count is not GATING on them.** Plain `curl -sS -o file -w …`
+exits **0** on an HTTP 404, writes the error body into the output file, and prints a tidy
+`HTTP=404 BYTES=14` that a hurried reader skims as evidence — this repository's own "a verification
+step that cannot fail is not a verification step" rule, broken in the one place it is being taught.
+Make the command *refuse*:
+
 ```bash
-curl -sS -o idx.json -w "HTTP=%{http_code} BYTES=%{size_download}\n" \
-  https://api.nuget.org/v3-flatcontainer/<package-id-lowercased>/index.json
+curl -sS --fail-with-body -o idx.json \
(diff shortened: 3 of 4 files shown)


rbuergi:
Pushed `908b53dd2` addressing both Copilot findings — both were right, and both were about this page teaching a rule it then broke.

**1. The commands printed the gate instead of enforcing it.** `curl -sS -o file -w "HTTP=%{http_code} BYTES=%{size_download}"` exits **0** on an HTTP 404, writes the error body into the output file, and prints a tidy `HTTP=404 BYTES=14` that skims as evidence. Step 1 and step 2 now carry `--fail-with-body`, an explicit `[ -s file ]` test, and a non-zero exit naming what is unmeasured. This is not hypothetical: while verifying this PR's claims, a guessed path for Rx's suppression file answered exactly 404/14 bytes, and `grep -c` over that body scores 0 — which unguarded reads as "zero suppressions, all clear".

**2. "Additions cannot break a consumer" was false, and the page contradicted itself twelve lines later.** Additions cannot break a *caller*; they routinely break an *implementer* — and this PR's own ledger is the instance (`ITypeInspector.HasParseMethod`/`Parse`, no default implementation, the single declared breaking change of 18.0.0). Step 3 now partitions the addition set instead of dismissing it, points at core's mirror-image `Interface additions (implementers declared)` gate, and names the `CS0419` ambiguous-cref shape.

Verified after the edit — docs are embedded resources, so `--no-build` would have been a false pass: `MeshWeaver.Documentation` and `MeshWeaver.Documentation.Test` rebuilt `-c Release -warnaserror` (0 Error(s), 0 Warning(s)), then **390 passed / 0 failed** including `DocumentationLinkIntegrityTest` and `ArchitectureTopicMapTest`.

---

Two notes from independently re-verifying this PR's claims rather than inheriting them, neither of them a defect in the change:

- **The API claim reproduces exactly.** A `MetadataLoadContext` diff of `16.3.0 lib/net8.0` → `18.1.0 lib/net10.0` gives precisely the 6 removals stated — the five `Accepts(Type)` plus `Scalar.IsKey` — with `ScalarConverterBase`1.Accepts` among the additions (so still present by inheritance), `IYamlTypeConverter` member-identical, and all six symbols this repo uses present in both. Registry and licence re-measured too: index HTTP 200/5,860 B, 270 versions, 18.1.0 newest stable; MIT in both nuspecs (HTTP 200, 1,337 B and 1,338 B).
- **18.1.0 adds a `lib/net10.0` asset**, so we now resolve `net10.0` where 16.3.0 gave us `net8.0`. Its public surface is byte-identical to 18.1.0's own `net8.0`, so nothing follows from it — worth recording only because the resolved TFM changed under the bump.
- `Licence gate passed: 114 shipping packages checked` is a **denominator**, not just a pass: the gate resolves from the *restored* graph, so it reflects what that worktree had restored. After a full solution restore it reports **257** here. The gate passes either way; the doc page now says to quote the number.

PR #4019, opened 2026-09-11
System.Reactive 6 -> 7: a packaging release, and the split Rx family it closes

rbuergi:
Moves `System.Reactive` from **6.1.0 to 7.0.0** under the standing directive to run the latest version of every library. Majors were held back because the directive says "at least within minor"; this one is explicitly authorised.

**The version being old was not the defect.** `Microsoft.Reactive.Testing` was *already* pinned at **7.0.0** while `System.Reactive` read **6.1.0**, five lines apart in the same file.

## The split, measured

`Microsoft.Reactive.Testing` 7.0.0's nuspec declares `<dependency id="System.Reactive" version="7.0.0" exclude="Build,Analyzers" />`, and `CentralPackageTransitivePinningEnabled` is not set — it occurs in `Directory.Packages.props` three times, every one of them in prose, never as a property — so the higher version wins, **but only in the projects that reference the testing package**. Exactly **8** test projects do: `Messaging.Hub`, `Data`, `Layout`, `Graph`, `Documentation`, `Compiler.Pipeline`, `Hosting`, `Hosting.Orleans`.

So the suite that decides whether the platform is fit to ship was exercising a different Rx **major** from the one the platform shipped. Nothing was red; nothing failed to restore. After this change **all 52** Rx-resolving projects in the tree report `System.Reactive/7.0.0` and **none** reports 6.1.0.

This also means the Rx 7 *runtime* has been under the test suite for as long as that pin has been at 7.0.0 — the bump moves what `src/` **compiles** against to match what the tests were already **running**.

## Premise re-measured, not inherited

Every read gated on **HTTP status AND byte count** — a failed request's empty body greps to zero and reads exactly like "no such version". (This is not hypothetical: reaching for the suppressions file below at a guessed path returned HTTP **404, 14 bytes**, which a `grep -c` would have scored as "zero suppressions, all clear".)

| question | measurement |
|---|---|
| Does Rx 7 exist; what is newest? | `https://api.nuget.org/v3-flatcontainer/system.reactive/index.json` → HTTP **200**, **974 B**. 47 versions; newest stable **7.0.0**, the only stable 7.x. |
| Licence at the new major? | **MIT at both ends**, unchanged: `<license type="expression">MIT</license>` in the 6.1.0 nuspec (HTTP **200**, **2,364 B**) and the 7.0.0 nuspec (HTTP **200**, **2,814 B**). Nothing of the OSMFEULA shape holding the Aspire family, so nothing to escalate. Assembly identity holds too: public key token `94bc3704cddfc263` unchanged; only `AssemblyVersion` moves 6.1.0.0 → 7.0.0.0. |

## What changed — measured on the assets this repo actually resolves

`project.assets.json` names them exactly: `compile = ref/net8.0/System.Reactive.dll`, `runtime = lib/net8.0/System.Reactive.dll`. Rx 6.1.0 ships no `ref/` at all and no `net8.0`, so the pairs are:

| surface | from → to | removals | additions |
|---|---|---|---|
| compile | `6.1.0 lib/net6.0` → `7.0.0 ref/net8.0` | **0** | — |
| runtime | `6.1.0 lib/net6.0` → `7.0.0 lib/net8.0` | **0** | 16 |

The 16 are `ObservableEx.CombineLatest`/`Zip` overloads for 9–16 sources, on a type nothing here references. Enumerated with `MetadataLoadContext`, version stamps normalised out (unnormalised they produce 69 phantom "removals" — they dominate the diff).

🚨 **The probe excludes members by ACCESSOR visibility, and that is load-bearing.** `Type.GetMembers` with `BindingFlags.NonPublic` returns non-public properties; a naive dump lists them and invents removals. There are exactly **138** such properties on each side — the previous pass reported `CurrentThreadScheduler.Time` as a removal purely because of this. It has no public or protected accessor, so no consumer could ever have called it. A removal you cannot call is not a removal.

The surfaces this codebase leans on are member-identical: `System.Reactive.Linq.Observable` at **656 members** with an empty diff, and every `Subject`/`ReplaySubject`/`BehaviorSubject`/`AsyncSubject`, every `System.Reactive.Disposables` type, and every scheduler.

## Two claims from the earlier pass that did not survive checking

Both were in the direction that reads as reassurance, which is why they are called out rather than quietly fixed.

**1. The new analyzer is INERT here — and that is not the same as passing.** Rx 7 adds `analyzers/dotnet/System.Reactive.Analyzers.dll` (6.1.0 ships none) emitting **RXNET0001–RXNET0004** at Warning, which `-warnaserror` would promote to errors. The earlier pass swept the source for the Windows UI types the rules fire on, found zero, and concluded *"that is why the build stays clean"*. **It isn't.** csc never loads the analyzer: it appears only in `project.assets.json`'s `libraries` file manifest, never in a `targets` entry, and `dotnet build <proj> -getItem:Analyzer` on a project referencing Rx returns **zero** Rx analyzer entries — while still returning the .NET analyzers, so the instrument discriminates rather than being broken. The package places the assembly directly under `analyzers/dotnet/` with no language subfolder, which is not where NuGet selects a C# analyzer from.

So the clean build is *not* evidence about those four rules. The real evidence is separate and is source-level: this repo declares **no `*-windows*` target framework at all**, references none of the Windows UI types (0 hits), and references no pre-6.x Rx facade package — swept against a positive control of the **581** `.cs` files under `src/` and `test/` that *do* contain `IObservable`, so the sweep is not looking at an empty corpus. Both facts are needed and neither substitutes for the other. If upstream ever fixes the analyzer's path, these rules go live for every consumer at once.

**2. The upstream suppression list must be read per-TFM, not in aggregate.** `ApiCompatSuppressions.xml` at tag `rxnet-v7.0.0` (HTTP **200**, **22,564 B**) does carry **57** entries, as stated. But "none is a behavioural or core-API break — all are `ref`-vs-`lib` UI omissions" is wrong: 51 are `CP0001` (40 naming Windows-UI/WinRT types, 11 `AsyncInfoObservable`/`IEventPatternSource`), and the other six — 5 `CP0008` plus 1 `CP0002`, an outright **member removal** — name *core* types, `ThreadPoolScheduler` and `NotificationKind`.

What actually rescues them is narrower and checkable: every one of those six compares `netstandard2.0` against `uap10.0.18362`, TFMs this repo never resolves. Both types and the `CP0002` member are present in `ref/net8.0` and `lib/net8.0` alike — consistent with the 0-removal diff above. The question is never "is the list clean", it is "does any entry name the asset pair I actually resolve".

## The risk that is not in the compiler

**Behaviour behind unchanged signatures** — the shape no surface detector can see. Scheduler defaults, `ObserveOn`/`SubscribeOn` semantics, error-propagation timing and completion/disposal ordering are unchanged, which matters here specifically because the action block is an **error boundary and a disposal fence**, so a shift in when Rx signals disposal would be a correctness change rather than a performance note. ADR 0004 (`Rx.NET/Documentation/adr/0004-onerror-to-throw.md` at tag `rxnet-v7.0.0`, HTTP **200**, **14,436 B**), which *would* change when `OnError` becomes a thrown exception, reads `Status: Proposed`.

🚨 **But all of that is upstream's word about itself, so it is paired with our own measurement of the property that matters.** `MeshWeaver.Messaging.Hub.Test` carries **40** `[Fact]`/`[Theory]` tests across **31** files whose whole subject is disposal, teardown, quiescing, shutdown admission and turn/backlog ordering — `RoutedDisposeRequest_Announces_Once_AndBeforeTheTeardownStarts`, `GateOpenMustNotLetARunningTurnOvertakeTheBacklog`, `InitializationStopsAtTeardownStart`, `ShutdownWindowAdmission`, the `DisposalStall*`/`DisposalRace*` family. Every one passes on Rx 7 (409/409 in that suite). The suite total alone would not have told you the ordering tests were even in it, which is why the denominator is stated. Two changes often attributed to Rx 7 — the `RefCount` fixes and `ResetExceptionDispatchState` — actually shipped in 6.1.0 and were already in force here. (The GitHub release notes are no help: 7.0.0, 7.0.0-rc.1 and 7.0.0-preview.20 have byte-identical bodies, an auto-generated changelog reaching back to 2021.)

The single compile-surface change is a nullability annotation: `OfType<TResult>(this IObservable<object>)` became `IObservable<object?>`, which widens what is accepted.

**Where a source break would be, and why a `lib`-only diff misses it.** Rx 7 keeps the Windows UI surface (`DispatcherScheduler`, `ControlScheduler`, …) in `lib/net472`, `lib/net8.0-windows10.0.19041` and `lib/uap10.0.18362` but omits it from the matching `ref/` assemblies — source-breaking, not binary-breaking, and invisible to a diff taken against `lib`. Confirmed directly: those symbols are present in the 7.0.0 `lib/` Windows assets and absent from `ref/`. It does not reach us, because we resolve the non-Windows `net8.0` pair and the repo has no Windows TFM.

**In-mesh source**, which no `dotnet build` compiles: **31 of 638** in-mesh artefacts reference Rx — 30 of 93 `.cs` node sources under the `Data/` trees, and 1 of 545 node `.json` (`samples/Graph/Data/Northwind/AnalyticsCatalog.json`, Rx inside escaped source). Since the removal set on the resolved assets is empty, there is nothing for any of them to have used.

## The control that could not have caught this

`PackageFamiliesMoveTogetherGuard` keys families on a shared **prefix**, and `System.Reactive` / `Microsoft.Reactive.Testing` share none — so the split was invisible to it by construction. The guard now carries a second shape, `NamedFamilies`, listing a family's members explicitly, plus a scanner check that every named member is actually pinned (so a rename cannot shrink a family silently).

The split-detection is extracted into one routine that the real check *and* the negative control both call — a guard tested through a replica of its own logic is not tested at all. The control asserts both directions: the exact 6.1.0/7.0.0 pairing that was live here must be reported, and an agreeing family (and a one-member family) must stay silent.

## Verification

- `dotnet build MeshWeaver.slnx -c Release -warnaserror` → **0 Error(s), 0 Warning(s)**, 57.93 s. Not an up-to-date no-op: **873** assemblies under `bin/Release/net10.0` carry a post-build mtime and **zero** predate the build; all 66 projects in the solution emitted.
- `Dependency licences` gate, after a **full** solution restore: `Licence gate passed: 257 shipping packages checked (59 declared-unused ignored)`, with `System.Reactive 7.0.0` in the graph. 🚨 That count is a denominator, not decoration — the gate resolves licences from the RESTORED graph, so a worktree that restored only a few projects prints the same "passed" over a fraction of the packages. Noted on the doc page, because it is easy to quote as a clean sweep when it is a partial one.
- The bump reaches the binaries, proven by hash rather than by a version string: the `System.Reactive.dll` copied beside the test assemblies is SHA-256 `e8070137fc07efeb8fd00e1607a096c7718ee1814c6b2032e4cb865f5cf042a9`, byte-identical to `lib/net8.0/System.Reactive.dll` from the downloaded 7.0.0 `.nupkg` — which also confirms *which* TFM asset resolved.
- Full suites, run as the test **executables** rather than through `dotnet test` (see below) — **0 failures**:

| suite | tests | failed |
|---|---:|---:|
| `MeshWeaver.Graph.Test` | 1545 | 0 |
| `Memex.Portal.Shared.Test` | 1293 | 0 |
| `MeshWeaver.Compiler.Pipeline.Test` | 859 | 0 |
| `MeshWeaver.Hosting.Test` | 619 | 0 |
| `MeshWeaver.Data.Test` | 508 | 0 |
| `MeshWeaver.Layout.Test` | 496 | 0 |
| `MeshWeaver.Messaging.Hub.Test` | 409 | 0 |
| `MeshWeaver.Documentation.Test` | 397 | 0 |
| `MeshWeaver.PluginTester.Test` | 362 | 0 |
| `MeshWeaver.Hosting.Orleans.Test` | 259 | 0 |
| `MeshWeaver.ContainerImages.Test` | 112 | 0 |
| `MeshWeaver.ContentCollections.Test` | 28 | 0 |
| `MeshWeaver.Testing.Xunit.Test` | 23 | 0 |
| `MeshWeaver.Cli.Test` | 19 | 0 |
| `MeshWeaver.Deployment.Contract.Test` | 15 | 0 |
| **15 suites** | **6,944** | **0** (2 skipped) |

  `MeshWeaver.Security.MeshTest` is the only test-tree project not run, and correctly so: it is `OutputType=Library` with zero `[Fact]`/`[Theory]` — a support library, not a suite. `MeshWeaver.Hosting.Orleans.Test` ran all **259** of its tests — including `PodHubClaimReassertionTest`, which passed here.

  `MeshWeaver.Documentation.Test` was then REBUILT and re-run after the doc edits, because docs are embedded resources and the earlier run had the pre-edit copy compiled in: **397 passed, 0 failed**, covering `DocumentationLinkIntegrityTest`, `ArchitectureTopicMapTest`, `WhatsNewEntryIntegrityTest` and the 4 `PackageFamiliesMoveTogetherGuard` tests.

🚨 A note on the instrument, since it produced a false pass on the way: `dotnet test` drives these xUnit v3 executables through the VSTest bridge, and on `Messaging.Hub.Test` that bridge died with `SIGABRT` after ~70 s, printed **no results at all**, and still exited **0**. The direct runner prints an explicit `=== TEST EXECUTION SUMMARY ===`. Every figure above comes from the direct runner, and this is written into the doc page.

## The one CI red, and why it is not this diff

The first full CI run on this branch failed: shard 2 aborted `MeshWeaver.Compiler.Pipeline.Test` with `exit=134 SIGNAL SIGABRT` and a synthetic `MeshWeaver.Compiler.Pipeline.Test.HOST_CRASHED` trx entry. That signal shape reads as a native crash, which is exactly what an Rx major is suspected of causing. It is neither a crash nor this diff, and it was cleared by measurement rather than by re-running.

**What it actually is.** The abort is a deliberate `Environment.FailFast` from the **CI-only** memory watchdog in `MonolithMeshTestBase`: process RSS crossed a fixed 6,144 MiB threshold (**6,256 MiB observed — 1.8% over**), and the `FailFast` message names its own cause, *"the cumulative Autofac `Reflection.Emit` factory leak from non-shared `MonolithMeshTestBase` classes"*. The guard arms only when `CI`/`GITHUB_ACTIONS` is set; on a dev machine it logs the breach and keeps running, by design. **So "it passes locally" is worth nothing here** — the guard is deliberately silent locally, and my clean local run of that suite was not a control.

**Exonerated by two positive measurements:**

1. **That suite already ran Rx 7 before this PR.** `MeshWeaver.Compiler.Pipeline.Test` references `Microsoft.Reactive.Testing` — it is one of the eight projects on the wrong side of the very split this PR closes. Read off the binary rather than the props file: I checked out the merge base, rebuilt that project, and hashed the `System.Reactive.dll` beside its test assembly — `e8070137fc07efeb8fd00e1607a096c7718ee1814c6b2032e4cb865f5cf042a9`, byte-identical to the 7.0.0 `lib/net8.0` asset. **The bump cannot change what that suite runs**; it changes only what `src/` compiles against, and that surface has zero removals.
2. **A/B of the property that actually failed, on one machine.** Peak RSS of that suite, measured with `/usr/bin/time -l`: **merge base 1,843 MiB, this branch 1,616 MiB** — lower, not higher. Both green (851 and 841 tests; the merge base has since gained 10).

Supporting context: #4012's run of the same shard set passed, recent `main` runs are green, and the shape is in neither `known-flakes.json` nor any other recent run — so it is a borderline shard crossing a fixed threshold on run-to-run noise, not a catalogued flake. I have not re-run it, catalogued it, or re-queued anything.

Both discriminators are now written into the doc page's step 6, because the first is peculiar to closing a split family and nobody will think of it unprompted: *the suite you are about to blame may already have been running the new version.*

## Scope

No public type or member is removed (the only `.cs` touched is a test guard), no i18n catalog key is added or re-worded, and no `PackageVersion` is removed — so `Pairs-with`, `Implementers`, `Mirror-sync` and `Satellite-pins` do not apply.

`MeshWeaver.Plugins` consumes this pin **versionless**, so this one line decides what that repo compiles against. Verified by reading that checkout directly: it has no `Directory.Packages.props` of its own, its `src/Directory.Packages.props` imports this repo's list, and **7** projects reference the family with no version — `MeshWeaver.Blazor`, `Blazor.Views`, `InstanceSync`, `LogWatcher` on `System.Reactive`, and `Observability.Test`, `AI.Orleans.Test`, `Query.Test` on `Microsoft.Reactive.Testing`. It picks the change up at its next pin move; nothing is required of it in the meantime.

🚨 **One correction to the earlier pass here too.** It waved this away with "Blazor Server is not a Windows-UI framework in the analyzer's sense". That is true but it is not the relevant fact, because **Plugins does have a Windows-UI project**: `clients/desktop/windows/MeshWeaver.Windows.csproj`, `net10.0-windows` with `UseWindowsForms=true`. The reasons it is unaffected are specific and checkable — it references **no** Rx package at all (only `Microsoft.Web.WebView2`), it sets `ManagePackageVersionsCentrally=false` so it does not consume this pin, and it is deliberately outside `MeshWeaver.slnx` and built separately on Windows. Across the whole Plugins checkout there are **0** references to the Windows UI Rx types, against a positive control of **494** `.cs` files under its `src/` that contain `IObservable`.

## Relationship to #4012 — resolved

#4012 has **merged**, and `main` is merged in here with the add/add conflict resolved exactly as this body said it would be:

- **Method section:** this branch's version, a strict content superset — six steps rather than five, adding step 4 ("Ask the library whether IT thinks it broke"), the phantom-removal warning in step 3, the per-TFM warning in step 4, the analyzer and VSTest-bridge warnings in step 6, and the red-triage discriminators. Verified by set-differencing the two method sections: every line present only on main's side is a renumber or a sentence this side expands.
- **Ledger:** *both* entries. #4012's YamlDotNet 16.3.0 → 18.1.0 crossing is appended verbatim — it is the whole work product of that PR, and taking "ours" wholesale would have silently deleted it.

🚨 **Taking `--ours` on `Architecture.md` also dropped two lines main had added** — `Catalog Action Identity` and `Compile Cache Input Freshness`, from PRs that merged while this branch was open. Both restored at their original positions; `ArchitectureTopicMapTest` would otherwise have reddened. The general trap is that resolving a conflict by taking a whole **file** reverts every unrelated change the other side made to it, so the resolution was re-checked by set-differencing both files line by line: the only line of main's now absent is the topic-map entry this branch deliberately supersedes.

`Directory.Packages.props` auto-merged: `YamlDotNet 18.1.0` and `System.Reactive 7.0.0` / `Microsoft.Reactive.Testing 7.0.0` all present, both comment blocks intact, XML re-validated. After the merge the pin still reaches the binaries — the `System.Reactive.dll` beside the test assemblies hashes `e8070137…` (the 7.0.0 `lib/net8.0` asset), 52 projects resolve 7.0.0 and none resolves 6.1.0 — and the full suites were re-run on the merged tree.


rbuergi:
Held as **draft on purpose** — this is the merge-order opt-out, not unfinished work.

Both this PR and #4012 add `Doc/Architecture/DependencyMajorUpgrades` and a line to `Architecture.md`, so whichever lands second meets an add/add conflict. This PR's body states that **#4012 lands first**; both were auto-armed and this one's CI started first, which would have inverted that and made the body untrue. Draft disarms auto-merge while leaving CI running, so the green signal still arrives.

**Ready-up sequence once #4012 merges:** merge `main` in, resolve by keeping **this** version of the method section and appending #4012's ledger entry (the method here is a strict content superset — verified by diffing the two method sections and confirming every remaining difference is a renumber or an expansion), rebuild `MeshWeaver.Documentation` + `.Test` (embedded resources, so `--no-build` would be a false pass), re-run that suite, then `gh pr ready`.

PR #4048, opened 2026-09-11
fix(#1605): release the shared mesh provider before its unload drain

rbuergi:
## What

Follow-up to #4046, answering its review comment (https://github.com/Systemorph/MeshWeaver/pull/4046#discussion_r3992357382). #4046 set `ServiceProvider = null!` before the collectible-unload drain on the per-test teardown path, but the shared-mesh path was left out: `SharedMeshProvider.DisposeAsync` kept its get-only `ServiceProvider` populated while it awaited `CollectibleUnloadDrain.WaitUntilCollectedAsync`. That holder still roots the disposed mesh and its `MeshContentTypeRegistry`, and with them the collectible contexts the drain is waiting on.

`SharedMeshProvider` now holds the provider in a nullable field. `DisposeAsync` resolves the tracker, disposes the provider, then clears both the field and its own local before awaiting the drain. After that, reading `ServiceProvider` throws `ObjectDisposedException`, so a late reader surfaces loudly instead of getting a disposed mesh. A second `DisposeAsync` returns without doing anything.

## Why this matters (Plugins#1605)

A teardown ends only when the mesh's collectible contexts have really unloaded (#4042). Any owner still holding the disposed mesh during that wait turns "collected" into "retained". This is the shared-path twin of #4046's frame and SP release.

## Verification

- `dotnet build test/MeshWeaver.Hosting.Monolith.TestBase -c Release -warnaserror`: 0 warnings, 0 errors.
- `MeshWeaver.Graph.Test` `AppIconAdoptionTest` (the one `ShareMeshAcrossTests => true` class in core) passes on the changed path. No test reads the shared provider after its collection disposed it.
- Limit: no core shared-mesh class compiles dynamic NodeTypes, so the shared drain has no retirements to wait on here. The retention effect is the same mechanism #4046 measured on the per-test path.

Pairs-with: none — test-base only; no public `src/` surface changes.

### DIFF
--- a/test/MeshWeaver.Hosting.Monolith.TestBase/MonolithMeshTestBase.cs
+++ b/test/MeshWeaver.Hosting.Monolith.TestBase/MonolithMeshTestBase.cs
@@ -273,8 +273,15 @@ public abstract class MonolithMeshTestBase : Fixture.TestBase
     private sealed class SharedMeshProvider(IServiceProvider serviceProvider, string testClassName)
         : IAsyncDisposable
     {
+        // Nullable, and cleared by DisposeAsync BEFORE the unload drain: while this holder keeps the
+        // disposed provider, it roots the mesh and its MeshContentTypeRegistry, and with them the very
+        // collectible contexts the drain waits for (Plugins#1605) - the shared twin of the per-test
+        // path's `ServiceProvider = null!`.
+        private IServiceProvider? serviceProvider = serviceProvider;
+
         /// <summary>The provider every test of the class shares.</summary>
-        public IServiceProvider ServiceProvider { get; } = serviceProvider;
+        public IServiceProvider ServiceProvider =>
+            serviceProvider ?? throw new ObjectDisposedException(nameof(SharedMeshProvider), testClassName);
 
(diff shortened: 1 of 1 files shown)


rbuergi:
Regression test for the review's 'focused shared-path teardown test' request: #4053. It adds TeardownWaitsForCollectibleUnloadsTest, whose shared case drives a shared-mesh fixture through the collection teardown that disposes SharedMeshProvider. Deleting the shared drain turns it red: Expected collection {} to contain "DISPOSE_SHARED_UNLOADS_COLLECTED".

Correction to this PR's own claim: the provider release it adds could NOT be pinned. With the provider disposed cleanly, keeping it roots nothing a harness can build. Autofac clears the singletons it built on dispose: a container-built holder resolved from the root provider was collected with this change reverted. Anything registered is also held by the fixture's own Services collection. The release stays (an in-drain gcroot put the provider on the chain, and it costs nothing), but the comment's 'it roots the mesh and its MeshContentTypeRegistry' is not a measured fact. #4053 rewrites it and the doc accordingly.

Issue #3859, opened 2026-09-09
Retention: interlock cleanup with protection updates and active publication


rbuergi:
Part of #3438; policy contract #3842.

The existing lock job and ACR purge run on separate clocks. The prior evidence in #3438 shows that purge can run even when protection collection did not run or failed, and that references can change between collection and deletion. Runtime platform resolution also creates in-flight build consumers that a static repository-pin scan cannot represent.

Scope: make cleanup eligibility depend on a successfully completed protection decision and preserve artifacts used by active builds/publications throughout the operation. Choose the coordination mechanism with the maintainer before implementing this architectural boundary. A wider retention window, larger keep count, or another independent timer does not close the race.

Acceptance:
- Failed/unavailable/incomplete protection collection cannot be followed by deletion.
- A build or publication becoming active during cleanup cannot lose its resolved platform images or bundles.
- A deterministic race test or controlled non-destructive rehearsal exercises collection failure and a new consumer appearing between scan and removal.
- The old independent destructive schedule is replaced only after the coordinated path is verified; the deployed task definition and run evidence are read back.
- Continuous artifacts use at least 30 days by age, with no build-count quota.

Depends on the consumer inventory and artifact-closure work linked from #3438. Related #3461 concerns two publishers writing one bundle namespace; this issue concerns cleanup racing its producers/consumers and must not duplicate that fix. No live purge, lock release, permission change or bypass is authorized by this tracking split.

rbuergi:
## Reopened — the 2026-09-10 sweep was not a triage decision

This issue was closed `not_planned` at **2026-09-10T14:30:19Z** by a bulk sweep that PATCHed **every open issue in all 13 `Systemorph` repositories** — 311 issues in twelve minutes, no comment on any of them — after a session was asked to *"close all issues in all repos"*. Nothing on this thread was read, and no judgement was made about this issue.

All 82 issues the sweep closed in the MeshWeaver fleet were reviewed afterwards. This one is **unresolved**: Open part of #3438; no PR exists.

Sixteen of the 82 were left closed — their fix had actually landed, or the author had withdrawn them; those should have read `completed`, not `not planned`.

rbuergi:
## STAYING OPEN, narrowed — the software half landed in #4081; the one thing left is the schedule itself


> 🚨 **Status, 2026-09-12 — the purge is PAUSED, so no date applies.** Roland ordered it stopped
[...]

Issue #3842, opened 2026-09-09
No platform pins in module repos — build against the released platform, and be TOLERANT of drift (same major serves)


rbuergi:
## Requirement (platform owner, 2026-09-09)

> We must not pin anything. It must be tolerant.

Stated after the 2026-09-09 incident on memex.systemorph.com (#3583, comments of today): a one-line CSS merge in Plugins (#1555) left every `Essentials/Email` page dead for hours. Part of the chain was a **pin**:

```
12:41:44Z  Plugins main, push run for 4c4019b46: FAILURE — "Repo policy gates: 2 staleness problem(s)
           in .github/workflows/ci.yml" — the hand-maintained MW_PLATFORM_SET / MW_PLATFORM_REF pin
           went stale the moment core released a new framework; every later run on main failed the
           same way until Plugins#1560 re-pinned to 3.0.0-ci.8195 at 15:30Z.
```

So a core release turns every module repo's `main` red until a human bumps a number, and while it is red the registry still serves HEAD to live portals. Meanwhile the portal itself runs a THIRD version (3.0.0-ci.8057, `s414bfb2…`), which no pin anywhere names — see also #3768.

## What "no pins, tolerant" means concretely

1. **Module repos do not carry a platform pin.** The build lane resolves the platform to build against at run time — the latest *released* (sealed) set, and/or the identities the fleet actually runs (`Hosting__Deployment` / `Hosting__ReportTo` know them). No `MW_PLATFORM_SET`, no `MW_PLATFORM_REF`, no `uses: …@<sha>` lane refs that a policy gate has to police for staleness. The staleness gate itself then has nothing to check and goes away.
2. **Compatibility is a VERSION rule, not an exact match.** A module built against platform X runs on platform Y unless one of them declares the break (a MAJOR bump). The same rule for module bundles vs. module source (#3583): same major ⇒ the adopted build keeps serving, with an honest "stale, vNEXT pending" status; major bump ⇒ "incompatible, awaiting bundle" on the page. Never a dead page, never an error caused by delivery alone.
3. **A red or cancelled main is invisible to portals.** The registry serves the last green set, never HEAD.
4. **Readiness is announced.** When a module version is actually adopted on a portal, the portal says so (notification + Updates page). Today the only signal between "merged" and "page dead" is the dead page.

## Where the pins live today

[...]

PR #4081, opened 2026-09-12
#3438: retention protects what is CONSUMED, and refuses when it cannot see it all

### LINKED ISSUE #3858: Retention: require a complete and fresh fleet inventory before releasing artifact protection
Part of #3438; policy contract #3842.

A remote portal can run platform Y while still executing a same-major fallback compiled for X. Core #3849 and Plugins Systemorph/MeshWeaver.Plugins#1564 carry the actual adoption stamps plus an explicit completeness signal. That producer/receiver pair is necessary but does not by itself prove that every expected consumer has reported recently.

Scope: establish the expected consumer set from the existing deployment inventory; account for running and adopted artifact identities; detect missing, legacy, incomplete and stale reports before deciding an artifact is unused. Do not infer zero consumers from absent evidence. Define the handling of deliberately retired installations explicitly.

Acceptance:
- A fresh complete report preserves both running Y and adopted X.
- A missing expected installation or stale/incomplete/legacy report prevents releasing potentially required protection, with the affected installation named.
- Retirement is explicit and auditable; a temporarily unavailable portal is not treated as retired.
- Verify report readback on the production receivers and demonstrate the decision with report-only output; no deletion is required to test the guard.

Dependencies: core #3849 and Plugins#1564. This issue does not authorize new credentials, cross-partition access, or live deletion. Inventory authority and retirement semantics need the maintainer involved before a major architectural change. Parent #3438 remains open.

### LINKED ISSUE #3860: Retention: preserve the full artifact set required by supported releases and advertised fallbacks
Part of #3438; policy contract #3842 and Doc/Architecture/ReleasedArtifactRetention.

Core #3843 protects age floors and last sealed bundle sets per source/major; #3846 applies the 30-day floor to assembly-cache cleanup. End-to-end retention still needs one verifiable mapping from retained releases and consumers to every artifact required to pull, start and serve them. Preserving only a version label or one top-level image is insufficient.

Scope: resolve the required artifact set for supported official releases, advertised last-green sets, running/adopted versions and active builds. Include image indexes and platform-specific manifests, tester/portal/migration images, bundle payloads and seals wherever required by the supported release pipeline. Use the actual published manifest structure rather than guessing names.

Acceptance:
- A retained release has a report listing every required image and bundle identity; each resolves.
- Both supported image architectures and the corresponding migration/tester artifacts remain usable when required by that release.
- An older same-major adopted fallback and a last-green release older than 30 days remain protected despite red/cancelled successors.
- Supported official releases remain protected for the underlying .NET release lifetime; plugin support also requires supported counterparty collaboration. Missing support metadata is unresolved protection, not permission to delete.
- Unreferenced continuous artifacts younger than 30 days remain; older ones become eligible only after inventory and coordination guards succeed, without a retained-build quota.

Requires the maintainer involved if implementing this changes major publication or retention architecture. Test with manifest fixtures and report-only production readback before any live cleanup change. Parent #3438 owns the final production acceptance.

rbuergi:
Closes #3858, closes #3860. **#3859 stays OPEN** — narrowed to the one thing this cannot build (see
"Not done" below) — and #3438 remains the parent for the rest of the residue.

## 🚨 Two things a reviewer cannot see from the diff, and they are the reasons to trust it

**1. Recording the truth would have reddened every pull request in the repository.**
`purge-old-images` was disabled on the morning of 2026-09-12. Writing `"status": "Disabled"` into
`.github/acr-retention/tasks.json` makes `read_purge_schedules` return *"records no ENABLED purge
task … either the record is stale or retention stopped; both are worth a red rather than a silent
pass"* — correct reasoning, and correct until that day, when a third cause became the true one. So
the record gains a `pause` block and the guard accepts an empty purge set **only** when that
declaration is present and complete (`inForce`, `since`, `reason`, `reEnableWhen`). That is the whole
argument for the declared-absence rule and it is invisible from the diff: **an absence is acceptable
only when it is declared; an undeclared one stays RED.** Same rule as `instances.json`.

**2. The pause made four of my own falsification arms vacuous.** The lock-before-purge arms were
driven on the **live** record. With the purge paused that list is legitimately empty, and
`ordering_problems(anything, [])` compares nothing and returns no problem — so all four would have
passed **while proving nothing**, on the day their subject was switched off. A reviewer who knows
those arms exist would reasonably assume they bite; they bite only because they were moved to a
fixture, with the shipped record asserted separately in *both* of its states. The refusal branch was
unfalsifiable for the same reason — the arms asserted what the record *says*, not whether the
function would still refuse an undeclared absence — and is now driven against four fixtures.

Neither was found by reading the code. The self-test failed, and that is why they are here.

## The four issues are one design

#3438 (parent), #3858 (a complete and fresh fleet inventory), #3859 (interlock cleanup with
protection), #3860 (the full artifact set) describe one mechanism:

> **A cleanup whose protection set is derived from a stale or incomplete source deletes something
> that is still in use.**

They were split on 2026-09-09 as "independently verifiable pieces", but the inventory, the closure
and the software half of the interlock are three properties of one decision — you cannot build a fresh inventory that
does not also decide what a refusal means, and you cannot interlock cleanup against a decision that
has no completeness verdict. So this is one change set in the two places a cleanup derives a
protected set: the container registry, and the portal's prebuilt-bundle store.

## Measured before building, read-only, 2026-09-12

| | |
|---|---|
| the nightly lock lane | **RED** ([run 34664099031](https://github.com/Systemorph/MeshWeaver/actions/runs/34664099031), 01:12Z — two hours before the 03:00 purge) on *"AXIS 1 found ZERO digest pins"* |
| …why | that assertion is about the **fleet**, and #3842 expired it. This repository's own extractor now finds **0** digest pins and **6** repositories naming a platform image: the satellites RESOLVE the platform set at run time. Pinning did stop. |
| …and its twin | `check-pinned-digests.py` carries the identical assertion, so the 05:20 guard was red on the same false premise |
| tag `memex-portal-ai:3.0.0-ci.8372` (what **both** production overlays pin) | `deleteEnabled: **true**` |
| the manifest it resolves to | `deleteEnabled: false` (locked) |
| tags of `memex-portal-ai` | 1,396, **0 locked** |
| memex-cloud | running an image set its committed pin did not name (Memex#219's third shape) |

`Azure/acr-cli` deletes a tag on the **tag's own** attribute:

```go
if includeLocked || (*(*tag.ChangeableAttributes).DeleteEnabled && *(*tag.ChangeableAttributes).WriteEnabled) { … }
```

So the protection that existed saved the **bytes** and lost the **reference**: the purge deletes the
tag, the locked manifest survives untagged, and the pin answers `manifest unknown` to the next pull.

## What lands

**AXIS 3 — ask the installations.** Each one the deployment overlays declare
(`Hosting__Deployment` + `ingress.host`) is asked through its own unauthenticated `/api/version`
what it is RUNNING, and the closure of image sets built from that commit is protected in every
repository its overlay pins — portal and migration twin together. 🚨 The bare short-sha tag is a
**moving** tag (core `4c99ec26` produced two sets ninety minutes apart and the tag followed the
newer one), so the axis protects the closure and says so rather than guessing a member of it.
Locking a superset destroys nothing; guessing does.

**The tag is locked beside the manifest**, `deleteEnabled` only, with the same read-back assertion
the manifest lock already has. `writeEnabled: false` would refuse the retag `release.yml` promotes
with and buys no purge protection.

**Completeness is a verdict with a denominator.** Installations expected / answered / not answered /
declared non-live, printed every run. One that did not answer is **named**, makes the run INCOMPLETE
and **refuses the unlock arm** — never a shorter protection set. Retirement is a declaration in
`.github/acr-retention/instances.json` with a reason; the overlays stay the denominator, so a
missing entry makes a run stricter and never looser.

**The instrument is measured, not inferred from the fleet.** `extractor_control()` drives
`extract()` over a fixture with known digest pins on every run in both scripts — which fires even
when the fleet happens to declare pins, where the fleet-shaped assertion could not.
`check-pinned-digests.py` additionally refuses to report a plain green over an empty subject.

**The same contract in the portal.** `DeploymentPinnedReferences.Resolve` takes the
`Hosting/Deployment` records as the expected-consumer denominator: no report, an unreadable report,
or one whose `sampledAt` is older than 24 h (twenty-four missed ticks of the one-hour reporter)
refuses the pass naming the instance. A retired record leaves the set explicitly and that is logged.
🚨 Zero expected is a **true** answer on a host that is nobody's fleet — the source is registered on
every portal, so refusing there would wedge retention fleet-wide — but a host that RECEIVES reports
and holds **no** records is refused rather than read as a fleet of zero.

## Falsification

12 sabotages against `lock-pinned-digests.py` and 5 against `check-pinned-digests.py`, each watched
to go RED **and to name the arm that caught it** — a traceback was rejected as a caught sabotage and
the harness was fixed until every sabotage failed an assertion instead.

Two arms exist only because the first pass found the wiring unfalsified: deleting
`extractor_control()` from `run()` left **both** self-tests green, because the harness drives its own
copy of the sequence. `ARM 29` now asserts, from the AST, that `run()` both performs every step the
harness drives and is driven by it.

24 xUnit tests in `DeploymentPinnedReferencesTest` (14 new), 432 documentation tests, 53 hosting
tests, all green; `dotnet build -c Release -warnaserror` clean on every touched project.

## The live run

Report-only against the real fleet, **exit 0**:

```
    AXIS 1 — digest pins in .github/workflows
      repositories scanned                     8
      …declaring at least one digest pin       0        ← would have been RED before this change
    AXIS 3 — what the installations are RUNNING (/api/version, live)
      installations the overlays declare       3
      …expected to answer                      2
      …that ANSWERED                           2
      …declared not-live in instances.json     1  (pearl: not-installed)
        memex              core 74d4c85 → 3 manifest(s) in its closure
        memex-cloud        core fce6780 → 2 manifest(s) in its closure
    TAG REFERENCES — the OTHER lock, and the one the purge reads to delete a tag
      tag references depended on               20
      …already protected (deleteEnabled false)  0      ← the live exposure
      …TO LOCK                                 20
```

## Would this have prevented today's failure?

**Yes for both halves of it, and it is worth separating them.**

1. The run **died** on the axis-1 zero. That assertion is gone, replaced by a control on the
   extractor plus a denominator that is still impossible (a fleet naming no platform image at all).
   The live run above is the same fleet, green.
2. The run **locked the wrong set** — the committed overlay's, not the instance's. AXIS 3 asks the
   instance. In the live run `hosting-operator` is a standing example: the overlay pins `4c99ec2`
   while installation `memex` is RUNNING `74d4c85`, and **both** are protected only because both
   axes are read.

And it would have prevented the shape nobody had named yet: every one of those 20 tag references is
deletable today with its manifest locked.

## No band-aids

No `--ago` raised, no `--keep` raised, no retry, no `continue-on-error`. Those are the knobs this
family exists to make unnecessary, and the purge task's own header says raising them moves the cliff.

## Not done, and deliberately the maintainer's

Listed in the doc under "What is still the maintainer's":

1. **The two clocks are still two clocks — and this is why #3859 stays open.** The structural interlock #3859 ultimately asks for is to
   make cleanup a step *downstream* of the protection decision — one lane, the destructive step
   `needs:` the complete verdict, the ACR timer task disabled. That is a registry change and a
   destructive-schedule change; **nothing destructive was run here** (no purge, no dry run, no
   registry mutation, no deploy). Until it lands, what this buys is that the exposure is **visible
   instead of silent**.
2. `purge-old-images` carries `memex-portal-ai` on its 7-day step — the image both production
   portals run — while the 30-day step covers only the retired `memex-portal` repository.
3. The `:latest` question; the `Container Registry Repository Writer` grant.
4. **A one-off state cleanup is owed and not performed:** 20 tag references are unlocked right now.
   Tonight's scheduled run locks them if it is allowed to; no hand command was issued.

Also still not inventoried, and named rather than implied: released **module bundles** and **active
builds** as consumers (#3860's remaining surface), tracked on #3438.

---

## Since the first push

- **`38dd2a8d`** — two `Hosting/ModuleInventory` nodes can describe one deployment (a control
  instance files its own report locally while remote ones arrive through the inbox). Keying them
  into a dictionary made the freshness verdict depend on query order: a stale duplicate beside a
  current report would read as a stale instance, and a current one beside a stale report as fresh.
  Every report contributes its references; the verdict comes from the newest. Falsified by making
  the oldest decide.
- **`9b45b5ad`** — *"how many, over what window"* was half-answered. The report named everything it
  protects and never named what it protects them from, and the two numbers that decide whether a
  lock was needed at all (`--ago`, `--keep`) live in a cloud-only ACR task whose only committed copy
  is `.github/acr-retention/`. The enabled purge steps now print beside the protected set. The
  verdict on that record stays in its own step; this only reads it, so an unstatable window prints
  as unstatable rather than being omitted — falsified by making a missing record report no problem.
## Copilot review — all nine addressed, and two of them changed the outcome

| # | Finding | What happened |
|---|---|---|
| 1 | duplicate roster ids are last-write-wins | **Fixed.** A repeated `id` is a blocker: which declaration is in force would be decided by the order of a JSON array, so neither is. Arm 24d, falsified (S17) |
| 2 | the reported version tag is not protected when a portal is ahead of its pin | **Fixed, and it was a real hole.** The pod spec names `memex-portal-ai:3.0.0-ci.N`, `/api/version` answers the core commit, and the axis matched short-sha tags. While an instance is ahead of its pin that version tag is in **no** committed file. Every tag of a running manifest is now protected — from outside the cluster there is no way to tell which name the spec used, so protecting one of them is a guess. Arm 21, falsified (S18) |
| 3 | multi-arch children are never traversed | **Fixed, and it turned out worse than a missing closure member** — acr-cli's locked-manifest `continue` comes *before* the walk that ignore-lists an index's children, so protecting the index alone made its children **less** safe than leaving it unprotected. Live: 12 platform manifests entered the closure that a locked index would not have covered. Arms 30/30b/30c, falsified (S14/S15/S16) |
| 4 | the unlock gate is not the purge interlock, so #3859 must not be claimed closed | **Agreed, and the plan changed: #3859 stays OPEN**, narrowed to the one architectural change (move cleanup behind the protection decision; disable the independent ACR timer). #3858 and #3860 close; #3438 remains the parent. This PR never claimed the schedule half — the doc, the PR body and the issue comments all say so — but leaving the ticket open is the honest bookkeeping |
| 5 | duplicate report identities are last-write-wins | **Already fixed** in `38dd2a8d`, before this review ran: reports are grouped per deployment, every one contributes references, and the freshness verdict comes from the newest. Falsified by making the oldest decide |
| 6 | the zero-consumer case is fail-open on a stale empty snapshot | **Fixed, and it is the sharpest of the nine.** `Records` reads an eventually-consistent index, so a control instance whose Deployment *and* inventory queries have both not landed answers exactly as an ordinary portal. Configuration is authoritative where the query is not: `Hosting:Deployment` set with `Hosting:ReportTo` unset **is** a control instance, and an empty record set on one is now a refusal |
| 7 | a future `SampledAt` is treated as fresh | **Fixed.** A negative age passed the budget trivially, so one skewed clock would have made every report that producer files permanently fresh. A future stamp is an unknown age |
| 8 | mutable `Dictionary` in `src/` against the collections policy | **Fixed.** The index is an `ImmutableDictionary<string, ImmutableList<MeshNode>>` built by `GroupBy`, claimed entries removed functionally, and the refusal list is an `ImmutableList` builder |
| 9 | `1 of 0 expected instance(s)` is a misleading denominator | **Fixed.** The refusal now reports the refusal COUNT beside separate expected / retired / orphan counts, because an orphan report can be the refused subject and is not a fraction of the expected set. Covered by a test asserting exactly that shape |

Three of these (1, 5, 9) are the same defect in three places — *an ordering decides a verdict that should be decided by evidence* — which is the family this change set is about, turned on its own code.

The self-test is now 18 sabotages, each watched to go RED **and to name the arm that caught it**; a traceback is rejected as a caught sabotage.

---

## 🚨 Status correction — the purge is PAUSED, so this is not a race

Roland ordered the purge stopped while protection was incomplete, and `purge-old-images` was disabled
at ~08:5xZ on 2026-09-12 (Memex#219). Confirmed here with `az acr task list`: **both tasks read
`status: Disabled`**. A disabled task still reports an `Enabled` **timer trigger** at `0 3 * * *` —
read the task's status, not the trigger's.

**No date is quoted anywhere in this PR or its docs.** The pause is the **mitigation** and this change
set is the **fix**, which makes it the precondition for safely turning the task back on rather than a
countdown. The condition is recorded in `.github/acr-retention/tasks.json` under `pause.reEnableWhen`,
and the protection report prints it where it would otherwise print a window.

Later commits:

- **`a2b0ba6e`** — the exposure sized and conserved, including the correction that **`--keep` is
  applied AFTER `--ago`**, over the eligible set only. Measured: 1,402 tags on `memex-portal-ai`,
  **none locked**; **193.1 tags/day**, so a 10-tag keep window is worth **~1.2 h**; the oldest tag in
  the repository was **exactly 7 days** — the horizon is already hard and `--ago` is the whole policy.
- **`ec9e276c`** — the record tells the truth, plus the two fixes described at the top of this body.
- **`690dd818`** — the operator page. `ImageCleanup.md` was the *source* of the wrong `--keep` model
  ("keep the 10 newest … encode the keeper protection with a `--keep` count"). It now answers "is the
  purge running", names the disabled-task/enabled-trigger trap, says the retention directory is a
  record rather than the registry, and corrects `--keep`.

Touches none of the five files reserved for `docs/daily-wave-build-process`.


rbuergi:
## Copilot review — all nine addressed, and two of them changed the outcome

| # | Finding | What happened |
|---|---|---|
| 1 | duplicate roster ids are last-write-wins | **Fixed.** A repeated `id` is a blocker: which declaration is in force would be decided by the order of a JSON array, so neither is. Arm 24d, falsified (S17) |
| 2 | the reported version tag is not protected when a portal is ahead of its pin | **Fixed, and it was a real hole.** The pod spec names `memex-portal-ai:3.0.0-ci.N`, `/api/version` answers the core commit, and the axis matched short-sha tags. While an instance is ahead of its pin that version tag is in **no** committed file. Every tag of a running manifest is now protected — from outside the cluster there is no way to tell which name the spec used, so protecting one of them is a guess. Arm 21, falsified (S18) |
| 3 | multi-arch children are never traversed | **Fixed, and it turned out worse than a missing closure member** — acr-cli's locked-manifest `continue` comes *before* the walk that ignore-lists an index's children, so protecting the index alone made its children **less** safe than leaving it unprotected. Live: 12 platform manifests entered the closure that a locked index would not have covered. Arms 30/30b/30c, falsified (S14/S15/S16) |
| 4 | the unlock gate is not the purge interlock, so #3859 must not be claimed closed | **Agreed, and the plan changed: #3859 stays OPEN**, narrowed to the one architectural change (move cleanup behind the protection decision; disable the independent ACR timer). #3858 and #3860 close; #3438 remains the parent. This PR never claimed the schedule half — the doc, the PR body and the issue comments all say so — but leaving the ticket open is the honest bookkeeping |
| 5 | duplicate report identities are last-write-wins | **Already fixed** in `38dd2a8d`, before this review ran: reports are grouped per deployment, every one contributes references, and the freshness verdict comes from the newest. Falsified by making the oldest decide |
| 6 | the zero-consumer case is fail-open on a stale empty snapshot | **Fixed, and it is the sharpest of the nine.** `Records` reads an eventually-consistent index, so a control instance whose Deployment *and* inventory queries have both not landed answers exactly as an ordinary portal. Configuration is authoritative where the query is not: `Hosting:Deployment` set with `Hosting:ReportTo` unset **is** a control instance, and an empty record set on one is now a refusal |
| 7 | a future `SampledAt` is treated as fresh | **Fixed.** A negative age passed the budget trivially, so one skewed clock would have made every report that producer files permanently fresh. A future stamp is an unknown age |
| 8 | mutable `Dictionary` in `src/` against the collections policy | **Fixed.** The index is an `ImmutableDictionary<string, ImmutableList<MeshNode>>` built by `GroupBy`, claimed entries removed functionally, and the refusal list is an `ImmutableList` builder |
| 9 | `1 of 0 expected instance(s)` is a misleading denominator | **Fixed.** The refusal now reports the refusal COUNT beside separate expected / retired / orphan counts, because an orphan report can be the refused subject and is not a fraction of the expected set. Covered by a test asserting exactly that shape |

Three of these (1, 5, 9) are the same defect in three places — *an ordering decides a verdict that should be decided by evidence* — which is the family this change set is about, turned on its own code.

The self-test is now 18 sabotages, each watched to go RED **and to name the arm that caught it**; a traceback is rejected as a caught sabotage.

Issue #4093, opened 2026-09-12
SelfUpdate refuses cr.meshweaver.cloud because the plugin registry is on memex.meshweaver.cloud — every fleet-registry instance cannot self-update (measured on build, 2026-09-12)


rbuergi:
Measured 2026-09-12 12:44Z in the first boot of the fleet build instance (`Deployments/build`, Systemorph/Memex#305), image `cr.meshweaver.cloud/memex-portal-ai:3.0.0-ci.8411` (core `c84c6c0`):

```
[SelfUpdate] starting (event-driven, with a 01:00:00 safety net); version=3.0.0-ci.8411, registry=cr.meshweaver.cloud/memex-portal-ai listed as an OCI Distribution registry (the mirror, instance-key auth), canPatch=False, retryInterval=06:00:00.
[SelfUpdate] check (Startup): check FAILED: InvalidOperationException: SelfUpdate:Registry is 'cr.meshweaver.cloud', an OCI registry the self-updater authenticates with this installation's plugin-registry instance key — but no plugin registry (PluginCatalog:Registries / PluginCatalog:RegistryUrl) is configured on that host, so there is no key to present. The mirror an installation pulls images from is the installation it installs modules from; configure both on the same host, or set SelfUpdate:Registry back to the upstream Azure Container Registry.
```

**The configuration is the fleet's documented shape, not a mistake on this record.** Per Systemorph/Memex `docs/new-deployment.md` (step 2, step 4) and `Doc/Architecture/ContainerRegistryInMemex` (#3353): the plugin registry is `https://memex.meshweaver.cloud` (`PluginCatalog:Registries:0:Url`, the `Plugins` mount), the image registry is `cr.meshweaver.cloud` (`imageRepository` → `selfUpdate.registry` derived by `HelmValues`), and ONE `mwi_` key serves both — `cr.meshweaver.cloud`'s `docker_auth` validates it against `https://memex.meshweaver.cloud/api/instances/token` (`RegistrySpec.validationUrl`). The kubelet pulls with exactly that key (pull Secret `registry-pull`, verified by `hosting-pull-secret`; the image pulled in 6 s). The self-updater has the same key in `PluginCatalog__RegistryToken` / `PluginCatalog__Registries__0__Token` but refuses to use it because the HOST strings differ.

[...]

Issue #3201, opened 2026-09-03
Rotate the plugin-registry token that sat in plaintext on both portals 2026-09-03 → 2026-09-11 (the shadow itself is removed, and the two tokens were EQUAL)


rbuergi:
Found while auditing `PluginCatalog__RegistryToken` for #3199. Reporting separately because it is a live production state, and because it changes the ORDER the Memex#148 cleanup has to be done in.

## memex carries two different registry tokens, and the managed one is the dead one

`PluginCatalog__RegistryToken` is set **twice** on `memex`:

| where | reaches the pod via |
|---|---|
| `secret/memex-portal-secrets` | `envFrom` — the chart's own path (`secrets.yaml`), Key Vault CSI provisioned |
| inline `env:` on `memex-portal-deployment` | **overrides `envFrom`** |

Both values are **47 characters**, and they are **NOT equal** (compared inside the cluster; neither value was read out or logged — only the verdict and the lengths).

So the token the platform provisioned through Key Vault is **dead**, and the portal authenticates to the plugin registry with a different token that exists **only as plaintext in the Deployment spec** — readable by anything that can `get deploy` in that namespace, and present in no committed source.

`memex-cloud` differs: its `memex-portal-secrets` carries only `PluginCatalog__OpenRegistration__Key`, so there the inline entry is the **only** copy.

## 🚨 Why this changes the cleanup order

The obvious cleanup — *"delete the inline entry, the secret already supplies it"* — is **wrong here and would break catalog auth**. The two tokens differ, so deleting the inline entry does not restore the status quo; it switches the portal onto a token nobody has verified is still issued and granted. On `memex-cloud` it would leave no token at all.

This is the `SHADOWS` class from #3168 — with the twist that the shadowed side is a **Secret**, not the ConfigMap. **#3168's detector does not catch this one**: it compares inline `env:` names against the rendered/live *ConfigMap* only, so a shadowed `envFrom` **secret** is invisible to it. Worth extending, and worth knowing before anyone treats that report as complete.

## Suggested order

1. Decide which of the two tokens is the one actually registered and granted at the registry (`memex.meshweaver.cloud`) — that is a lookup on the registry's instance records, not a guess.
2. Put the winner in Key Vault for **both** namespaces (`memex-cloud` has no secret entry at all today).
3. Only then delete the inline `env:` entries — and verify the catalog still loads before and after, because this is the credential the Store reads with.
[...]

PR #4094, opened 2026-09-12
fix(#4093): the self-updater accepts a registry whose DECLARED validator holds the key

rbuergi:
> 🚨 **Refs #4093 — deliberately NOT `Fixes`, in the body and in the commit message.** This change
> makes the fix **possible**; it does **not** fix `build` or `pearl`. Both stay refused until
> `SelfUpdate__RegistryValidationUrl` is declared on their records in Systemorph/Memex (a config-repo
> PR, tracked below). **#4093 stays OPEN past this merge** — auto-closing it would mark an issue
> resolved whose subject is still broken in production.

An installation that pulls its platform images from the fleet registry `cr.meshweaver.cloud`
boots, serves, pulls its image in seconds with its `mwi_` instance key — and then never
self-updates. `OciTagLister.ResolveCredential` decided *which* plugin-registry key to present by
matching the **host**, and the fleet's registry deliberately validates at **another** portal
(`docker_auth` forwards the key to `https://memex.meshweaver.cloud/api/instances/token`), so the two
hosts differ by design and host matching can never succeed on the shape the fleet deploys. Core's
own contract says as much on `RegistrySpec.ValidationUrl`: *"so a consumer holds ONE credential for
both."*

## The fix is a DECLARED pairing, not a loosened check

🚨 **The guard is a credential-disclosure control** — it stops the installation presenting its
instance key to whatever host `SelfUpdate:Registry` happens to name — so dropping it, or replacing
it with "the sole mount carries a key" or a suffix / registrable-domain match, is not acceptable.
The rule is now:

| the key is presented to… | …because |
|---|---|
| the container registry's **own host**, when a plugin registry is configured there | unchanged — today's behaviour |
| the host named by **`SelfUpdate:RegistryValidationUrl`**, when a plugin registry is configured there | the operator declared that portal as the registry's validator **and** this installation holds a key for it |
| **nothing else** | an absent declaration is refused, never read as permission |

Two explicit statements are required for either row. Hosts are compared **whole** and
case-insensitively, through one shared reader (`SelfUpdateOptions.HostOf`) used on both sides so they
cannot drift; a value carrying userinfo (`https://memex.meshweaver.cloud@evil.example`) is **refused, not
stripped** — its host is `evil.example` and its userinfo is `memex.meshweaver.cloud`, so an
implementation that stripped the userinfo would have declared the attacker's host trusted.

## Where the running portal learns the pairing

**It cannot today, and that is why this is a declaration and not an inference.** `RegistrySpec` lives
on the record of the instance that *hosts* the registry (`Deployments/memex-cloud`), on the control
instance; a consumer's record (`build`, `pearl`) carries no `registry` block, `HelmValues` renders
`registry.validationUrl` only for the hosting record, and reaching across instances at credential
time would put a network dependency inside the update path. So the operator restates the registry's
own `validationUrl` on the consuming side:

- `SelfUpdateOptions.RegistryValidationUrl` (config key `SelfUpdate:RegistryValidationUrl`)
- chart: `selfUpdate.registryValidationUrl` → renders `SelfUpdate__RegistryValidationUrl`
- an existing instance is fixed with no chart change through a record's `extraPortalConfig` /
  an overlay's `config.memex_portal` — the same rendered key (verified with `helm template`, both
  paths, and the default render is `""`)

## The falsified tests

Both directions were measured, not assumed (`test/Memex.Portal.Shared.Test/OciTagListerTest.cs`):

- **Guard deleted** (`match = registries.FirstOrDefault()`, the "one mount, present it" fix) →
  **3 failed / 6 passed**: `TheSameRegistry_WithNoDeclaredValidator_IsStillRefused`,
  `ADeclaredValidatorThisInstallationHoldsNoKeyFor_IsRefused` and the pre-existing
  `NoPluginRegistryOnThatHost_FaultsNamingTheHost` all go red. The negatives assert `Requests == 0`
  **and** `PresentedSecret == null`, so they fail on disclosure, not only on the message.
- **Validator branch removed** (today's `main`) → **1 failed / 8 passed**:
  `ADeclaredValidator_PresentsTheKeyHeldForThatPortal_AndListsTheRegistry` goes red — the positive
  measures the fix rather than restating it.
- **Boundary validation removed** (the review round, below) → **2 failed / 9 passed**:
  `ARegistryTargetCarryingUserinfo_IsRefused_AndNeverEchoed` and
  `AUrlFormRegistryTarget_IsRefused_NamingTheBareHostToUse`.
- 🚨 **Two false passes happened while proving these, and both are recorded rather than tidied away**
  — the reusable lesson is that a negative control must be proven to have **RUN**, not merely to be
  green, because a control that silently became a no-op converts "I did not check" into "I checked
  and it was fine":
  1. A falsification patch that **did not compile**, run with `--no-build`, reported a confident
     `11/11 passed` off the **stale assembly** — the exact trap AGENTS.md lists.
  2. The disclosure assertion was **vacuous**. `PresentedSecret` is assigned only inside the
     `/v2/token` branch, reached solely for hosts the fake serves, so a request to
     `evil.example.test` was 404'd before any secret was recorded and the assertion held whatever
     the code did. Isolating it with the guard deleted **passed**, which is how it was caught. Root
     cause is the protocol: the OCI client presents Basic only *after* a 401 challenge, and a fake
     that answers 404 can never observe a leak. An unknown host now behaves like a **hostile
     registry** — challenges, serves `/v2/token`, records whatever it is handed, before the host
     check. The same experiment now fails with **"found 2 item(s)"**, and every negative asserts
     `CredentialsSeen`.
- Restored: **1382/1382** in `Memex.Portal.Shared.Test`, `MeshWeaver.Documentation.Test` **432/432**,
  chart invariants **7/7**, `check-values-are-read` **123/123**, and `0 Error(s)` on
  `-c Release -warnaserror` for `MeshWeaver.Hosting`, `Memex.Portal.Shared` and the test project.

## Blast radius (measured against every `Hosting/Deployment` record in Systemorph/Memex)

| instance | image registry | lister | after this change, as configured today | once the operator declares the validator |
|---|---|---|---|---|
| `build` | `cr.meshweaver.cloud` | OCI | **unchanged — still refuses** | self-update starts working |
| `pearl` | `cr.meshweaver.cloud` | OCI | **unchanged — still refuses** | self-update starts working |
| `memex` | `meshweaver.azurecr.io` | ACR | **nothing changes** (`OciTagLister` is not on the ACR path) | n/a |
| `memex-cloud` | `meshweaver.azurecr.io` | ACR | **nothing changes** (hosts the registry; pulls from ACR) | n/a |

🚨 **"Would newly present a key somewhere it does not today": EMPTY, by measurement.** No record, no
overlay, no values file and no `appsettings` in this repo or in Systemorph/Memex sets
`SelfUpdate__RegistryValidationUrl` / `selfUpdate.registryValidationUrl`, and the chart's default
renders `""`. So on every instance in the fleet as it stands, the resolved credential after this
change is byte-identical to before; the key can only reach a new host after a human writes the
declaration.

### What the credential path emits — nothing derived from the secret

The one new `Information` line is quoted in full. Its **only** two arguments are hosts; everything
else is literal text naming a **configuration key**, not a value:

```csharp
logger?.LogInformation(
    "[SelfUpdate] presenting the instance key configured under PluginCatalog:Registries "
    + "for {PluginRegistryHost} to the container registry {Registry}: "
    + "SelfUpdate:RegistryValidationUrl declares that host as the portal which validates "
    + "this installation's key there.",
    declaredValidator, registry);
```

No token, no prefix, no suffix, **no length**, no hash. `Information` leaves the pod for Loki, so the
comment above the call says so explicitly and tells the next author not to add a length "for
diagnostics" — a length is a fact about a credential.

🚨 **One hardening beyond the brief, prompted by re-reading this line.** It first logged
`paired.Url`, a *configured URL* — and a URL may carry userinfo
(`https://instance:mwi_…@memex.meshweaver.cloud`), which would have put the key in Loki inside a
field nobody thinks of as a credential: the #3201 shape, where the registry key sat in a pod spec
`kubectl describe` printed and is still owed a rotation. It now logs the **parsed host**
(`declaredValidator`, which *is* that registry's host — it is what the match was made on). The same
substitution was applied to the pre-existing "no instance key could be resolved for the plugin
registry at {match.Url}" message on the identical path. Audited: the resolved token is only ever
*returned* as the credential, and `token.Length > 0` is a predicate, never an output.

The refusal messages name hosts and config-key names only. The startup line now also says when **no**
validator is declared, so the gap is visible before the first check fails.

## Conserved

- `src/MeshWeaver.Documentation/Data/Architecture/SelfUpdateRegistryCredential.md` (new), linked
  from the Architecture topic map and from `ContainerRegistryInMemex`, whose now-stale "the plugin
  registry whose host equals the container registry host" sentence is corrected in the same change.
- What's New entry, `Category: Fix`.

Pairs-with: none — nothing public is removed or renamed; the change adds `SelfUpdateOptions.RegistryValidationUrl`, `RegistryValidatorHost`, `RegistryValidatorDeclared` and `HostOf`, and `RegistryTokenResolver.ResolveDurableKey` (round 2); `SelfUpdateOptions` is a plain record and `RegistryTokenResolver` a sealed class, neither with implementers.
Mirror-sync: none — no localization catalog key is added or re-worded (the i18n gate script confirms: "No catalog keys added and no values changed").

### 🚨 Measured: with the target check removed, the instance key reaches an arbitrary host — twice

Not an argument, a measurement. Delete the bare-host requirement, point `SelfUpdate:Registry` at
`instance:mwi_…@evil.example.test`, declare any validator this installation holds a key for, and the
fixture's attacker host receives **two credentials** — the Basic credential at the token realm it
names, then the Bearer on the listing.

**That is what makes the bare-host requirement non-negotiable.** The design discussion below supports
it; it does not lead.

### The most important sentence in this PR

> **The fix introduced the vulnerability it exists to prevent.** Host equality could never match
> `instance:mwi_…@evil.example`, so the old code always refused it. The declared-validator branch
> matches on the *declared* host and never looks at the *target* — so the new trust path made a
> previously unreachable value reachable, and `OciRegistryClient` would have built
> `https://{registry}/` around it and presented the Basic credential to `evil.example`.

That is why `SelfUpdate:Registry` must be host-EQUAL to its parsed host, not merely parse to one.
**Do not simplify it back to `HostOf(registry)`** — normalizing silently would accept a URL form here
while `PortalImage`/`MigrationImage`, which interpolate the same value, stayed malformed, and the
constraint would have lost the reason it exists. The rule and this attack are recorded together in
`Doc/Architecture/SelfUpdateRegistryCredential`, so the constraint travels with its justification.

Measured, not argued: with the target check deleted, the fixture's attacker host received **two**
credentials — Basic at the token realm, then the Bearer.

**And on the earlier "the credential path is clean":** that audit was not wrong, its **denominator**
was. The two places I inspected (`paired.Url`, and the pre-existing `match.Url`) were clean; the
parameter *between* them — `registry`, on the same call — was the hole.

### The vacuity audit — "could this assertion FAIL?"

A vacuous negative is a **class**, and the userinfo case gave up its mechanism: *a fake that refuses
early can never observe what it was built to catch.* Every negative in this change was audited
against that question — by deleting the guard it protects, stripping the test to the assertion under
audit, and measuring what reported. **Two of five shared the shape**, and the hostile-host fixture
fixed both:

| negative | guard deleted ⇒ the assertion reports | was it vacuous? |
|---|---|---|
| no plugin registry on that host | **2 credentials** at `other.example.test` | **YES** — 404'd before a credential could be observed |
| the same registry, nothing declared | **3 credentials** | no — the target is a host the fake serves |
| a declared validator holding no key | **3 credentials** | no — same reason |
| a target carrying userinfo | **2 credentials** at `evil.example.test` | **YES** — the case that started the audit |
| a URL-form target | **4 requests** under the `HostOf`-without-equality simplification | no — and it is what blocks that simplification |

🚨 **"no plugin registry on that host" predates this PR** and had the identical defect: its
disclosure assertion had never been able to fail. It became a real control only when the fake learned
to challenge.

The two non-fake negatives cannot take this shape: one asserts values from a pure function with no
fake in the path; the other drives a host the fake *does* serve and asserts a throw, which an
empty-list regression could not satisfy.

**The same audit is owed across the wider suite — that is its own piece of work and is deliberately
not folded in here.**

### This is a TRUST EXPANSION, and the doc page now says so

Before: the key could reach only a host with a configured plugin registry. After: **any** host,
provided some configured plugin registry is declared its validator. That is the feature working as
designed and it is fail-closed, but it widens who can receive a credential. Three things bound it,
each load-bearing rather than defence in depth: an **explicit declaration**; a **bare-host target**;
and a **plugin registry actually configured at the declared validator**.

### Review round 2 (hand-run) — eleven findings, addressed in order

`review_on_push: false` on the ruleset means Copilot does not re-review a push, so this round was run by hand; the bar is the reviewer's sign-off. Every falsification below was **built under `-c Release -warnaserror` before it ran** — the runner voids a patch that does not compile instead of replaying the stale assembly, and the very first attempt (F1) was voided that way (`CS0219`) and re-done.

| # | finding | done | falsified by ⇒ what fired |
|---|---|---|---|
| 1 | 🚨 boot line logged `SelfUpdate:Registry` **verbatim** at `Information` — a `user:mwi_…@evil` value shipped the key to Loki at boot, before the lister's non-echoing refusal ever ran | **fixed** — `SelfUpdateOptions.HostOf(_options.Registry) ?? "(unreadable — see SelfUpdate:Registry)"`. Swept every other read of `Registry`/`RegistryValidationUrl` in `Memex.Portal.Shared` + `MeshWeaver.Hosting`: no other log, message, node or report carries the raw value (the lister's messages name the *normalized* host; the ACR lister in MeshWeaver.Plugins logs its `registry` at Debug and is out of this PR's path — noted, not touched) | new `TheBootLine_NamesTheRegistryHost_NeverTheConfiguredValue` against a captured logger; reverting to `_options.Registry` ⇒ `Did not expect "… registry=instance:mwi_smuggled…@evil.example.test/…"` |
| 2 | 🚨 the validator branch presented whatever `ResolveToken` yields — for an auto-registered instance that is `StoredCredential → Exchange → mwa_…`, and the declared validator **is** `/api/instances/token`, which refuses a token by shape ⇒ 401 on every check; only `build`/`pearl` (raw Token) were fixed | **fixed** — `RegistryTokenResolver.ResolveDurableKey` (configured token, else the stored key decrypted, **no exchange**); `ResolveToken` and it share one `ConfiguredToken` rule and differ only in whether the STORED key is exchanged. The validator branch resolves through it; the same-host branch is unchanged (the mirror's `/v2/token` accepts both shapes). First cut of the refactor routed the configured token through the exchange too — the fake's new exchange route caught it (3 same-host tests 401'd) before it left the worktree | new `AnAutoRegisteredInstallation_PresentsItsStoredDurableKey_NeverAnExchangedToken`: `Token` blank, key only in `Admin/PluginRegistryCredential/…`, fake serves `POST /api/instances/token` and refuses `mwa_` at the realm; switching the branch to `ResolveToken` ⇒ `Expected ExchangeRequests to be 0 … but found 1` |
| 3 | `HostOf` refuses userinfo, so `PluginCatalog:Registries:0:Url=https://instance:mwi_x@host` silently stopped matching on BOTH branches and was diagnosed as "no plugin registry configured" | **decided: refuse and name the cause, at the lister** — a registry on that host (or the declared validator) whose URL carries userinfo is refused with "its PluginCatalog:Registries URL carries credentials … move the key to `PluginCatalog:Registries:N:Token`", URL not echoed. Not at `PluginCatalogOptions` binding: that would widen the change to every catalog consumer, the catalog never honours userinfo as a credential anyway (the HTTP client drops it) and already logs `registry.Url` verbatim on several pre-existing paths, and no fleet record uses the shape | new `APluginRegistryWhoseUrlCarriesCredentials_IsRefusedNamingTheCause_NeverAsAbsent` (both rows); removing the diagnosis ⇒ `Expected "… but no plugin registry … is configured …" to contain "carries credentials"` |
| 4 | `RegistryValidatorHost` collapsed "declared but unreadable" into "not declared" — the boot line said NO validator declared and the refusal told the operator to set the key they had set | **fixed** — `SelfUpdateOptions.RegistryValidatorDeclared` (`!IsNullOrWhiteSpace`) beside the reading; the lister refuses first with "`SelfUpdate:RegistryValidationUrl` is set but does not name an http(s) host … value not repeated"; the boot line has THREE states | new `ADeclaredButUnreadableValidator_IsRefusedAsMalformed_NeverAsAbsent` (userinfo, `mailto:`, `htps://`) + `ADeclaration_IsToldApartFromAReadableOne`; removing the refusal ⇒ all three report the old `"… and no validator is declared for it …"` diagnosis; the two-state boot line ⇒ `Expected "… NO validator declared …" to contain "is SET but does not name an http(s) host"` |
| 5 | a legal bare `host:443` was refused, and the message claimed it "carries a scheme or a path" | **fixed** — "bare" is a textual test (`IsBareHost`: no `://`, no `/ \ ? #`; userinfo already refused by `HostOf`), not equality with the normalized host; client and messages then use the normalized host. The message is now true of every value that reaches it | new `ABareHostNamingTheDefaultPort_IsAccepted_AndListed`; restoring the equality form ⇒ `SelfUpdate:Registry must be a bare registry host, but it carries a scheme or a path; it names the host 'mirror.example.test'` — for `mirror.example.test:443`. `AUrlFormRegistryTarget…` is now a theory (`https://host`, `host/v2`, `https://host:443/v2/`) and asserts the message is true |
| 6 | `FakeMirror.ServedHosts` was appended only for KNOWN hosts, so the positive's `OnlyContain(RegistryHost)` could not see an un-credentialed request to a third host | **fixed** — `ReachedHosts` records every host unconditionally, before the fake decides what it serves; the `known` gate stays for what the fake serves; `ServedHosts` deleted | injecting an un-credentialed `GET https://third.example.test/` in the paired branch ⇒ all three paired positives: `Expected all items in the collection to match the predicate` — with the old gate `third.example.test` was not `known` and would never have been appended |
| 7 | with one registry configured, nothing distinguished "the key held for the DECLARED validator" from "some registry's key" | **fixed** — new `ADeclaredValidator_SelectsTheKeyHeldForThatPortal_NotAnotherRegistrys`: A (`https://a.example.test`, `mwi_issued-by-a…`) and B (the mirror, `InstanceKey`), B declared; the fake accepts both keys so a wrong selection reports as the wrong SECRET | `registries.FirstOrDefault()` in place of `On(registries, declaredValidator)` ⇒ `Expected value to be "mwi_this-installations-own…" … but found "mwi_issued-by-a-example-test-and-not-the-validator"` (and the "declared validator holding no key" negative leaks **3 credentials**) |
| 8 | `SelfUpdateOptions.HostOf` duplicated `RegistryUpdateReconciler.SameRegistry` with different port semantics | **fixed — one rule.** `SameRegistry` now reads both sides through `HostOf`. **The two subsystems now agree**: a scheme-default port is not part of the host, any other port is; userinfo, non-http schemes and non-URLs name no registry. Behaviour change, deliberate and pinned: `("not a url", "not a url/")` no longer "match" (two copies of an unreadable string are not a registry), and `https://memex…@evil.example.test` vs `https://evil.example.test` is `false`; `http://x:443` vs `https://x` — same under the old rule — now differs (the only case the two rules disagreed on that could have been a real registry, and a config nobody writes) | reverting `SameRegistry` to its own reader ⇒ the `not a url` and the userinfo theory rows fail (`Expected: False Actual: True`) |
| 9 | What's New `Icon: RefreshCcw` is a Lucide name | **fixed** — `ArrowSync` (Fluent, the feed's most-used self-update icon) | `WhatsNewEntryIntegrityTest` passes (it does not validate icon names — noted) |
| 10 | `HostOf(match.Url) ?? "(unreadable URL)"` was a dead fallback | **fixed** — the matched HOST (`registry` or `declaredValidator`) is carried as `matchedHost`; `match.Url` is never read for a message | — (dead code removed; the empty-credential message still names the host) |
| 11 | altitude: derive the pairing on the control instance at render time; also the second trust rule (`PluginBundleClient.DownloadArtifact`) | **filed, not implemented** — [#4123](https://github.com/Systemorph/MeshWeaver/issues/4123), linked from the doc page's "Where the running portal learns the pairing" (new subsection "The alternative: derive it on the control instance — recorded, not done") | — |

**Negatives restructured so a falsification reports WHAT LEAKED:** every negative now asserts `CredentialsSeen` / `Requests` first through `Record.ExceptionAsync`, and the message second — `Assert.ThrowsAsync` used to fail with "no exception was thrown" before the disclosure assertion could run.

**Guarded before push, in this worktree:** `MeshWeaver.Hosting`, `Memex.Portal.Shared` and `Memex.Portal.Shared.Test` each `0 Error(s)` on `-c Release -warnaserror`, one project per invocation, dll timestamps read after each; lister + `SameRegistry` **32/32**; the doc guards (`DocumentationLinkIntegrityTest`, `WhatsNewEntryIntegrityTest`, `ArchitectureTopicMapTest`) executed by name — present in the `.trx` — and the whole `MeshWeaver.Documentation.Test` **432/432**; full `Memex.Portal.Shared.Test` **1397/1397** (was 1382 — 15 new cases).

#### The vacuity audit, re-run — every negative in `OciTagListerTest`

| negative | guard deleted ⇒ the assertion reports | vacuous? |
|---|---|---|
| no plugin registry on that host | **2 credentials** at `other.example.test` | was (round 1); no |
| the same registry, nothing declared | **3 credentials** | no |
| a declared validator holding no key | **3 credentials** | no |
| a target carrying userinfo | **2 credentials** at `evil.example.test` | was (round 1); no |
| a URL-form target ×3 | the `no plugin registry` message for the raw value; under the equality form `host:443` refused with an untrue message | no |
| a declaration set but unreadable ×3 | the old **"no validator is declared"** diagnosis | no |
| a plugin registry whose URL carries credentials ×2 | the old **"no plugin registry is configured"** diagnosis; **3 credentials** when the host guard goes | no |
| the boot line | the line with **`instance:mwi_…@evil.example.test`** in it; the two-state line saying **"NO validator declared"** | no |
| a refused key faults, never an empty list | `No exception was thrown` under `.Catch(Observable.Return([]))` | no |
| a third host reached (`ReachedHosts`, all three paired positives) | `Expected all items in the collection to match the predicate` | **was** (this round, finding 6); no |
| the auto-registered shape presents the durable key | **`ExchangeRequests` 1, expected 0** | no |
| the key held for the DECLARED validator, two registries | **A's key** presented instead of B's | no — and could not exist with one registry |

Pure-function negatives (the validator-host reader, declared-vs-readable) have no fake in the path.

**Public surface added (no pairing obligation — additions only):** `SelfUpdateOptions.RegistryValidatorDeclared`, `RegistryTokenResolver.ResolveDurableKey`. `RegistryUpdateReconciler.SameRegistry` stays `internal`.

### Follow-ups this PR does not do

1. **The declaration itself**, in Systemorph/Memex — `extraPortalConfig:
   { "SelfUpdate__RegistryValidationUrl": "https://memex.meshweaver.cloud/api/instances/token" }` on
   `Deployments/build` and `Deployments/pearl`. That is the change that actually restores
   self-update, and **#4093 closes on it, not on this merge.** It is a config-repo PR and the
   deployment lane is another session's.
2. **`canPatch=False` on `build`** — the second question in #4093, untouched here and not folded in.
3. **Derive the pairing on the control instance at render time**, and give `PluginBundleClient.DownloadArtifact` the same trust rule — [#4123](https://github.com/Systemorph/MeshWeaver/issues/4123), recorded during review round 2; deliberately not part of this PR.

### DIFF
--- a/deploy/helm/templates/memex-portal/config.yaml
+++ b/deploy/helm/templates/memex-portal/config.yaml
@@ -456,6 +456,16 @@ data:
   {{- if (.Values.selfUpdate).registry }}
   SelfUpdate__Registry: "{{ .Values.selfUpdate.registry }}"
   {{- end }}
+  # 🚨 The portal that VALIDATES this installation's instance key at that registry — the registry
+  # record's own `validationUrl` (MeshWeaver#4093). The fleet registry cr.meshweaver.cloud decides a
+  # pull by forwarding the caller's key to memex.meshweaver.cloud, so the image registry and the
+  # plugin registry are DIFFERENT hosts by design, and without this declaration the self-updater
+  # refuses to present the key it holds — an instance boots fine and then never self-updates.
+  # A string, so an absent value IS inert: no declaration ⇒ the key is presented only to a host that
+  # is itself a configured plugin registry, exactly as before. Never read as permission, and never
+  # inferred from name resemblance. It may be set through an overlay's config.memex_portal or a
+  # record's extraPortalConfig; the values path below is what the instance record renders.
+  SelfUpdate__RegistryValidationUrl: "{{ (.Values.selfUpdate).registryValidationUrl | default .Values.config.memex_portal.SelfUpdate__RegistryValidationUrl | default "" }}"
   # ---- Container image mirror (MeshWeaver#3353, src/MeshWeaver.ContainerImages) --------------
   # The /v2 pull surface THIS installation serves. Rendered only when containerImages.upstream is
   # set — an absent block reaches no container and the mirror stays OFF (every /v2 route 404s).
--- a/deploy/helm/values.yaml
+++ b/deploy/helm/values.yaml
@@ -64,6 +64,22 @@ selfUpdate:
   # 🚨 PAIRED with portal.imagePullSecret below: the images the updater rolls to are named on this
   # host, so the kubelet needs a credential for it. AcrPull on the node identity covers ACR only.
   registry: ""
+  # 🚨 THE PORTAL THAT VALIDATES THIS INSTALLATION'S KEY AT THAT REGISTRY (MeshWeaver#4093) — the
+  # registry record's own `validationUrl`, copied verbatim (a bare host works too; only the host is
+  # read). Set it whenever `registry` above is a registry that decides a pull by FORWARDING the
+  # caller's key to a portal, which is the fleet's shape: cr.meshweaver.cloud asks
+  # https://memex.meshweaver.cloud/api/instances/token whether the key is good, so the image
+  # registry and the plugin registry are two DIFFERENT hosts by design
+  # (Doc/Architecture/ContainerRegistryInMemex). Leave it empty when the registry IS the portal that
+  # serves the catalog — a portal serving its own /v2 mirror needs no declaration.
+  #
+  # 🚨 It is a GRANT, and an absent one is a refusal, never permission: the self-updater presents
+  # this installation's instance key ONLY to a host that is a configured pluginCatalog registry, or
+  # to `registry` when this names that registry's validator AND a pluginCatalog registry is
+  # configured there. Two explicit statements, both required; name resemblance grants nothing.
+  # Without it an instance on the fleet registry boots fine and then never self-updates — every
+  # check fails with "no plugin registry is configured on that host".
+  registryValidationUrl: ""
 # ---- Container image mirror (the /v2 pull surface THIS installation serves) -----------------
 # The read-through OCI mirror (src/MeshWeaver.ContainerImages, MeshWeaver#3353): this installation
 # proxies container images from an upstream registry with ONE credential it holds, while every
--- a/deploy/helm/values.yaml
+++ b/deploy/helm/values.yaml
@@ -56,10 +56,21 @@ selfUpdate:
   # Container Registry, listed through ACR's API with the workload identity above. A registry
   # login-server host here — `memex.meshweaver.cloud`, the read-through mirror another
   # installation serves at /v2 — makes the self-updater list tags through the OCI Distribution API
-  # instead, authenticated with THIS installation's plugin-registry instance key (the mirror host
-  # must equal a pluginCatalog registry's host; no second secret), and roll to images named on
-  # that host. Every installation EXCEPT the one serving the mirror sets this — the mirror instance
-  # cannot serve the image that boots it (Doc/Architecture/ContainerRegistryInMemex).
+  # instead, authenticated with THIS installation's plugin-registry instance key (no second
+  # secret), and roll to images named on that host. Every installation EXCEPT the one serving the
+  # mirror sets this — the mirror instance cannot serve the image that boots it
+  # (Doc/Architecture/ContainerRegistryInMemex).
+  #
+  # 🚨 WHICH key is presented is decided by a PAIRING, and there are exactly two ways to qualify:
+  # this host IS a pluginCatalog registry (a portal serving its own /v2 mirror), or
+  # `registryValidationUrl` below declares the portal that validates this installation's key here
+  # AND that portal is a pluginCatalog registry. The fleet's own registry needs the SECOND — do not
+  # read the first as a requirement to put both on one host, which is what #4093 was.
+  #
+  # 🚨 A BARE `host` or `host:port` ONLY — never a URL, a path, or a value carrying credentials.
+  # The same value is interpolated into the portal and migration image references, and a value of
+  # the form `user:secret@host` would be a valid URI naming a DIFFERENT host as the one to hand the
+  # instance key to. Anything else is refused at startup of the listing.
   #
   # 🚨 PAIRED with portal.imagePullSecret below: the images the updater rolls to are named on this
   # host, so the kubelet needs a credential for it. AcrPull on the node identity covers ACR only.
--- a/memex/Memex.Portal.Shared/SelfUpdate/OciTagLister.cs
+++ b/memex/Memex.Portal.Shared/SelfUpdate/OciTagLister.cs
@@ -79,7 +79,7 @@ public sealed class OciTagLister(
         //    never match such a value, so the declared-validator branch below is what would make it
         //    reachable — this refusal is the reason that branch cannot open a disclosure path.
         //
-        // Requiring host-equality (not merely "parses to a host") also keeps this in step with
+        // Requiring a bare host (not merely "parses to a host") also keeps this in step with
         // SelfUpdateOptions.PortalImage/MigrationImage, which interpolate the SAME value into an
         // image reference and need a bare host for it to be well-formed.
         var registry = SelfUpdateOptions.HostOf(configured);
@@ -92,7 +92,13 @@ public sealed class OciTagLister(
                 + "credentials, and never one this platform cannot also use as the registry half of "
                 + "an image reference. The configured value is not repeated here because a value of "
                 + "this shape can carry a credential."));
-        if (!string.Equals(registry, configured, StringComparison.OrdinalIgnoreCase))
+        // 🚨 "Bare" is a TEXTUAL test on the configured value, NOT equality with the host HostOf
+        // read from it: HostOf drops a scheme-default port, so `cr.example.test:443` — a legal
+        // `host:port` the documentation promises — would fail an equality test and be refused with
+        // a message claiming it carries a scheme or a path when it carries neither (#4094 review).
+        // A value that parsed (so: no userinfo) and carries no scheme and no path IS a bare host,
+        // whatever port it names; the client and the messages then use the normalized host.
+        if (!IsBareHost(configured))
             // Safe to name: HostOf refuses userinfo, so a value that parsed cannot have carried one.
             return Observable.Throw<IReadOnlyList<string>>(new InvalidOperationException(
                 $"SelfUpdate:Registry must be a bare registry host, but it carries a scheme or a path; "
@@ -104,6 +110,15 @@ public sealed class OciTagLister(
                 hub, registry, ResolveCredential(registry), options.RegistryUsername, HttpClientName, logger)
             .ListTags(repository);
     }
+
+    /// <summary>
+    /// Whether a value that <see cref="SelfUpdateOptions.HostOf"/> could read is ALSO nothing but
+    /// <c>host[:port]</c>: no scheme, no path, no query, no fragment. Userinfo is not tested here
+    /// because HostOf already refused it. Pure.
+    /// </summary>
+    private static bool IsBareHost(string configured) =>
+        !configured.Contains("://", StringComparison.Ordinal)
+        && configured.IndexOfAny(['/', '\\', '?', '#']) < 0;
 
     /// <summary>
     /// 🚨 The plugin-registry credential this installation may present to the container registry —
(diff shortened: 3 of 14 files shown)


rbuergi:
Addressed in a86d12a5a — thank you, the critical finding was real and it was a regression **this PR introduced**.

**The target is now normalized and validated before any credential is resolved.** You were right on both halves, and the disclosure half is worse than the comment states: `OciRegistryClient` builds `https://{registry}/` when the value has no scheme (`OciRegistryClient.cs:128-130`), so `instance:mwi_…@evil.example` is a *valid* URI whose host is `evil.example` — it would have received the Basic credential — and the raw value is interpolated into a dozen of that client's exception messages as well as the audit line added here.

The part worth recording: host equality could never match such a value, so step 1 always refused it. **The declared-validator branch is what made it reachable**, because it matches on the declared host and never looks at the target. So the one thing this design exists to prevent was introduced by the design itself.

`ListTags` now requires the configured value to already BE a bare `host[:port]` — `HostOf` must parse it **and** agree with it — which also closes the asymmetry you identified. Host-equality rather than merely "parses to a host" is deliberate: the same value is interpolated into `PortalImage`/`MigrationImage`, so accepting a URL form in the lister while the image reference stayed malformed would trade one silent divergence for another. The unparseable branch does **not** echo the value, since the userinfo is exactly where a key would be.

Two tests pin it (`ARegistryTargetCarryingUserinfo_IsRefused_AndNeverEchoed`, `AUrlFormRegistryTarget_IsRefused_NamingTheBareHostToUse`); deleting the check reddens both.

**Your three suppressed comments were all valid and are all fixed:**

- **`SelfUpdateRegistryCredential.md`** — you were right that it misstated where the key goes. The page now separates the two questions explicitly: the key is always *sent to* the container registry; the validator host only *selects which* credential and is never a destination. The table gained a "the key is sent to" column saying so on every row.
- **`values.yaml`** — the pre-existing guidance telling operators the mirror host "must equal a pluginCatalog registry's host" is removed; it described the exact shape #4093 is about. It now states the two ways to qualify, and that the value must be a bare host.
- **What's New** — reworded and renamed: it adds the setting, it does not switch anything on. It now opens by saying an affected instance keeps refusing until an operator declares the pairing.

Local: `Memex.Portal.Shared.Test` 1382/1382, `MeshWeaver.Documentation.Test` 432/432, chart invariants 7/7, `0 Error(s)` on `-c Release -warnaserror`.

PR #4105, opened 2026-09-12
hosting-registry-register: issue the instance's plugin-registry key and store it (Plugins#1720)

rbuergi:
Operator half of MeshWeaver.Plugins#1720 (the plan half — a "Register at the plugin registry" step before "Create Key Vault secrets" — is the paired Plugins PR; `Pairs-with: none — this PR removes no public surface`).

## The hand step it replaces

Memex `docs/new-deployment.md` step 2, *"The instance key → `<prefix>PluginCatalog-RegistryToken`"*:

> `curl -sS -X POST https://memex.meshweaver.cloud/api/instances/register … | jq -r .instanceKey > /tmp/<id>-key` … `az keyvault secret set --vault-name Systemorph --name <prefix>PluginCatalog-RegistryToken --file /tmp/<id>-key`

## What changes

New command `hosting-registry-register --registry-url … --instance-id … --home-url … --vault … --object … [--display-name …] [--bootstrap-secret …]`:

- **vault object present** → its key is presented to `GET /api/instances/self` and must authenticate as THIS instance → `::hosting:: registry_registration=present`, nothing issued, nothing written. A key the registry rejects (401/403) or that belongs to another instance is a **refusal** naming the re-issue path (Instance grants ▸ re-issue → `az keyvault secret set … --file`) — never a silent second registration, and the object is kept.
- **vault object absent** → `POST /api/instances/register` with an EMPTY bootstrap key (open registration, free plan — what the runbook does) or the `mwr_` key read from `--bootstrap-secret`; the returned `mwi_` key is stored through a mode-600 `--file`, read back by id, then **proven** to authenticate before the step reports `registry_registration=registered`, `registry_plan=free`, `registry_key_hash=<sha256>`.
- `409` (id taken, key not in this vault) → refusal naming re-issue; `400` names the id rule; `401/403` names `--bootstrap-secret`; `404`/no answer name the registry.
- **Never prints a key**: the body is composed by `jq` from the environment and reaches curl via `--data @-`; the bearer rides `hosting::registry_call`. Deliberately reports `registry_key_hash`, not `key_hash` (which `OperatorOutput.KeyHash` reads as a rotation).

## Rights

[...]

PR #4103, opened 2026-09-12
hosting-kv-copy: materialise a fleet-shared vault object under an instance's prefix (Plugins#1723)

rbuergi:
Operator half of MeshWeaver.Plugins#1723 (the plan half — `keyVaultSecrets.secrets[].copyFrom` → a "Copy fleet-shared secrets" step — is the paired Plugins PR; `Pairs-with: none — this PR removes no public surface`).

## The hand step it replaces

Memex `docs/new-deployment.md` has no step for it — it was done ad hoc for `build` on 2026-09-12 (`build.json` notes): 

> `build-GitHub-App-PrivateKey` — a byte-identical COPY (sha256-verified) of `memexsystemorph-GitHub-App-PrivateKey`, made 2026-09-12 by piping `az keyvault secret show` into `az keyvault secret set --file /dev/stdin`

The Memex PR at the end of this series adds it to the runbook as "done by Provision since Plugins#1723".

## What changes

New command `hosting-kv-copy --vault <v> --copy <target>=<source> […]`:

- **absent target** → source read into a variable, written through a mode-600 `--file` (never argv), read back by id; `::hosting:: kv_copy_created=n`;
- **existing target** → **KEPT, never rewritten** (the master-key rule); compared with the source **by sha256** and a difference is reported as `kv_copy_drift=n` — a fact for the operator, not a failure;
- no value is ever printed; every pair is gated before the vault is touched; every refusal names the exact hand command; dry run reads no value.

Why a copy at all: `hosting-kv-purge` deletes by PREFIX on teardown, so a cross-prefix mapping would let the first teardown take the fleet's shared PEM with it. The copy makes the object the instance's own, purged with it.

## Rights the operator identity needs

Key Vault **secret get** on the source object and **set** on the target — both under the existing *Key Vault Secrets Officer* grant on `Systemorph`. Nothing new.

## Gates run

```
$ bash deploy/aks/operator/test/run-tests.sh
319 passed, 0 failed          (291 on main; +28 new hosting-kv-copy cases)
$ shellcheck -S warning deploy/aks/operator/bin/hosting-* deploy/aks/operator/test/stubs/kv-copy/az
(clean)
```

`hosting-operator.yml`: `hosting-kv-copy` added to the plan-command list, tracked-file floor raised to 26 (measured on the tree merged with main 8aaf35f: 23 commands + run.sh + _common.sh + _audit.jq; main is at 25), stub added to the tracked-stub list.

## Ships how

[...]

PR #4109, opened 2026-09-12
hosting-signin-app: keep, verify and add the sign-in app's client secret — never rotate, never show (Plugins#1719)

rbuergi:
Operator half of MeshWeaver.Plugins#1719 (the plan half — an "Ensure sign-in app secret" step after the vault step — is the paired Plugins PR; `Pairs-with: none — this PR removes no public surface`).

## The hand step it replaces (partly — read on)

Memex `docs/new-deployment.md` step 1, *"The sign-in app → `<prefix>Authentication-Microsoft-ClientSecret`"*:

> `APP_ID=$(az ad app create --display-name "<Name> Portal (<host>)" --sign-in-audience AzureADMultipleOrgs --web-redirect-uris "https://<host>/signin-microsoft" …)` · `az ad sp create` · `az ad app credential reset … --query password -o tsv > /tmp/<id>-secret 2>/dev/null` · `az keyvault secret set … --file`

## What is automated, what is verification-only, and why

New command `hosting-signin-app --name <id> --host <host> --vault … --object … [--client-id <app id>] [--years 1]`:

| State | What the step does | Right it needs |
|---|---|---|
| vault object **present** | **kept** (never minted over, never rotated) and, with `--client-id`, the app is **verified** to list `https://<host>/signin-microsoft` — a readable app missing it is RED naming `az ad app update`; an identity that cannot read the app reports `signin_app_verify=unknown` with the hand command and does **not** fail the provision | Graph `Application.Read.All` for the verify — **not granted** to `hosting-operator` today, so on the fleet this reports `unknown` until it is |
| **absent**, record names the app | a credential is **added** (`az ad app credential reset --append` — existing credentials stay valid; this adds, it never rotates), az's stderr echo of the secret discarded, the value stored through a mode-600 `--file`, read back, `signin_secret=created` + expiry | Graph `Application.ReadWrite.OwnedBy` on an app the identity **owns** (or `Application.ReadWrite.All`) — **not granted**; without it the step is RED with the exact hand commands |
| **absent**, no `signIn.microsoftClientId` | **RED, deliberately**: registering the app needs the Graph right AND writing the new client id back to the git-owned record — the message carries the runbook's exact commands and the record field to set | — |

So today, on the fleet, the step is the **verification half** for the two absent cases (RED with the manual command) and a no-op keep for the present case. Granting `hosting-operator` `Application.ReadWrite.OwnedBy` + making it an owner of each instance's app turns the middle row on with no code change.

## Gates run

```
$ bash deploy/aks/operator/test/run-tests.sh
325 passed, 0 failed          (291 on main; +34 new hosting-signin-app cases — the stub prints the minted secret on stderr like the real az, proving it is discarded)
$ shellcheck -S warning deploy/aks/operator/bin/hosting-* deploy/aks/operator/test/stubs/signin-app/az
(clean)
```

`hosting-operator.yml`: plan list, tracked-file floor 23 → 24 (will need `main` merged and the floor re-counted once the sibling PRs land), stub list.

## Ships how

Hop (C): `hosting-operator:<sha>` on merge to `main`; reaches a Provision when `operator.image` on `Deployments/memex` moves. Merging is not shipping.

### DIFF
--- a/.github/workflows/hosting-operator.yml
+++ b/.github/workflows/hosting-operator.yml
@@ -65,8 +65,8 @@ jobs:
         run: |
           set -euo pipefail
           tracked=$(git ls-files deploy/aks/operator/bin/ | wc -l | tr -d ' ')
-          if [ "$tracked" -lt 23 ]; then
-            echo "::error::only ${tracked} file(s) tracked under deploy/aks/operator/bin/ — expected 23 (20 commands + run.sh + _common.sh + _audit.jq). The floor is the REAL count — raise it with every command added, or a lost file passes."
+          if [ "$tracked" -lt 25 ]; then
+            echo "::error::only ${tracked} file(s) tracked under deploy/aks/operator/bin/ — expected 25 (22 commands + run.sh + _common.sh + _audit.jq). The floor is the REAL count — raise it with every command added, or a lost file passes."
             echo "::error::Almost certainly .gitignore matching this bin/ again. `git add` on an ignored path exits 0 and does NOTHING, so the loss is silent until a fresh checkout builds an image with no scripts in it."
             git check-ignore -v deploy/aks/operator/bin/run.sh || true
             exit 1
@@ -116,6 +116,7 @@ jobs:
           hosting-redirect
           hosting-registry-key
           hosting-restore
+          hosting-signin-app
           hosting-tls
           hosting-verify
           hosting-verify-backup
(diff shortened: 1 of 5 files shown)


rbuergi:
Corrected the tracked-file floor in `hosting-operator.yml`: `main` already tracks **24** files under `deploy/aks/operator/bin/` (21 commands + `run.sh` + `_common.sh` + `_audit.jq`) while the guard declared 23 and its comment said "20 commands" — stale, so one lost file would have passed. This PR adds one command, so the floor is now **25** with the comment re-counted. 🚨 The four sibling PRs (#4103, #4105, #4109, #4110) each do the same on the same line: whichever lands after the first must merge `main` and set the floor to the real count at that point (26, 27, 28), not keep 25.

rbuergi:
**Merge resolved by content; floor settled at 25 — measured, not assumed.**

`origin/main` merged in (no rebase), one push. All three conflicts were *both sides added*, and both sides are kept:

| File | Resolution |
|---|---|
| `deploy/aks/INSTANCE-LIFECYCLE.md` | main's extended `hosting-kv-ensure` bullet (#4100) **and** this PR's guarantee bullet |
| `deploy/aks/operator/test/run-tests.sh` | main's `hosting-kv-ensure` block **and** this PR's block |
| `.github/workflows/hosting-operator.yml` | tracked-stub list = the **union** of both lists |

**The floor, and how it was counted.** `git ls-tree -r --name-only origin/main deploy/aks/operator/bin/ | wc -l` = **24** on `main` (tip `69e587c5c`): 21 `hosting-*` commands + `run.sh` + `_common.sh` + `_audit.jq`. `main` *declares* 23 with a comment saying "20 commands" — stale, so one lost script would have passed. This PR adds exactly one command, so the true count for this tree is **25 (22 commands + 3)**, and the guard and its comment now say 25. Verified on the pushed head: tracked=25, commands=22, floor=25.

**Guards run on the pushed head:** `deploy/aks/operator/test/run-tests.sh` green (both test blocks coexist), `shellcheck -S warning` clean over `_common.sh`, `run.sh` and every `hosting-*`. No run was re-run wholesale.

**The follow-up is still needed and is now precisely scoped.** Each of #4103/#4105/#4109/#4110 sets the floor true for *its own* merge, so as the siblings land the real count rises 25 → 26 → 27 → 28 while the floor stays 25 — up to 3 files of slack. After all four have merged, a one-line PR sets `-lt 28` and `expected 28 (25 commands + run.sh + _common.sh + _audit.jq)`. It cannot be pre-set here: 28 would fail the guard for whichever lands first.

PR #4110, opened 2026-09-12
hosting-image-mirror: copy the pinned platform images from ACR into the fleet registry when it lacks them (Plugins#1722)

rbuergi:
Operator half of MeshWeaver.Plugins#1722 (the plan half — `Mirror platform images` as the first Provision/Roll step — is the paired Plugins PR; `Pairs-with: none — this PR removes no public surface`).

## The hand step it replaces

Memex `docs/new-deployment.md` step 4:

> `crane copy meshweaver.azurecr.io/memex-portal-ai:<tag> cr.meshweaver.cloud/memex-portal-ai:<tag>` · `crane copy meshweaver.azurecr.io/memex-migration:<tag> cr.meshweaver.cloud/memex-migration:<tag>`

## What changes

New command `hosting-image-mirror --registry <fleet host> --repository <path> --tag <tag> [--also <path>]… --vault … --publisher-secret … [--source <acr host>] [--publisher <user>]`:

- **present** in the fleet registry → **KEPT, never re-pushed**; its digest is compared with the source's and a difference is the fact `image_drift` (a re-tagged upstream build is the operator's call);
- **absent** → the source is asked (absent there too → refusal naming the sealed set to pin), `crane copy` runs, and **both digests are read back and must match** before it reports;
- portal AND migration ride the same tag (Memex#141) via `--also`;
- the publisher password (vault) and the ACR token (`az acr login --expose-token`) reach crane on **stdin** (`--password-stdin`) into a per-process `DOCKER_CONFIG` — never argv, never printed, never a fact;
- a non-ACR `--source` is refused (the fleet's upstream is ACR and the token comes from `az`); `--registry` equal to `--source` is refused.

**Dockerfile**: adds `crane` (go-containerregistry v0.20.3, selected per `TARGETARCH`) — nothing else in the image can do a manifest+blob copy, and the image's own tool check now asserts it.

## Rights the operator identity still lacks

**`AcrPull` on `meshweaver`** (plus Reader on the registry resource so `az acr login --name meshweaver --expose-token` resolves it) — `hosting-operator` does not hold it today. Until it is granted the step is RED naming the grant and the exact `crane copy` fallback; nothing is copied and nothing else is touched. Key Vault get on `memexcloud-Registry-PublisherPassword` is under the existing *Key Vault Secrets Officer* grant.

## Gates run

```
$ bash deploy/aks/operator/test/run-tests.sh
326 passed, 0 failed          (291 on main; +35 new hosting-image-mirror cases)
$ shellcheck -S warning deploy/aks/operator/bin/hosting-* deploy/aks/operator/test/stubs/image-mirror/*
(clean)
```

The crane/az stubs record logins by host and user only — never a credential — and the suite asserts no sentinel appears in any output or argv.

`hosting-operator.yml`: plan-command list, tracked-file floor 23 → 24, stub list, and `crane` added to the in-image tool assertion. (The floor line conflicts textually with the sibling PRs #4100/#4103/#4105/#4109, which bump the same number; whichever lands last merges `main` and re-counts.)

## Ships how

Hop (C): `hosting-operator:<sha>` on merge to `main`; reaches a Provision when `operator.image` on `Deployments/memex` moves to it. Merging is not shipping.

### DIFF
--- a/.github/workflows/hosting-operator.yml
+++ b/.github/workflows/hosting-operator.yml
@@ -65,8 +65,8 @@ jobs:
         run: |
           set -euo pipefail
           tracked=$(git ls-files deploy/aks/operator/bin/ | wc -l | tr -d ' ')
-          if [ "$tracked" -lt 23 ]; then
-            echo "::error::only ${tracked} file(s) tracked under deploy/aks/operator/bin/ — expected 23 (20 commands + run.sh + _common.sh + _audit.jq). The floor is the REAL count — raise it with every command added, or a lost file passes."
+          if [ "$tracked" -lt 25 ]; then
+            echo "::error::only ${tracked} file(s) tracked under deploy/aks/operator/bin/ — expected 25 (22 commands + run.sh + _common.sh + _audit.jq). The floor is the REAL count — raise it with every command added, or a lost file passes."
             echo "::error::Almost certainly .gitignore matching this bin/ again. `git add` on an ignored path exits 0 and does NOTHING, so the loss is silent until a fresh checkout builds an image with no scripts in it."
             git check-ignore -v deploy/aks/operator/bin/run.sh || true
             exit 1
@@ -106,6 +106,7 @@ jobs:
           hosting-dns
           hosting-export
           hosting-federate
+          hosting-image-mirror
           hosting-inline-env-retire
           hosting-kv-ensure
           hosting-kv-purge
@@ -192,7 +193,7 @@ jobs:
         # through to a REAL az/kubectl on a runner that has one.
         run: |
           set -euo pipefail
-          for f in deploy/aks/operator/test/stubs/pull-secret/az deploy/aks/operator/test/stubs/pull-secret/kubectl deploy/aks/operator/test/stubs/pv-resize/kubectl deploy/aks/operator/test/stubs/inline-env/kubectl deploy/aks/operator/test/stubs/kv-rotate/az; do
+          for f in deploy/aks/operator/test/stubs/pull-secret/az deploy/aks/operator/test/stubs/pull-secret/kubectl deploy/aks/operator/test/stubs/pv-resize/kubectl deploy/aks/operator/test/stubs/inline-env/kubectl deploy/aks/operator/test/stubs/kv-rotate/az deploy/aks/operator/test/stubs/image-mirror/az deploy/aks/operator/test/stubs/image-mirror/crane; do
             git ls-files --error-unmatch "$f" >/dev/null 2>&1 || { echo "::error::${f} is not tracked"; git check-ignore -v "$f" || true; exit 1; }
             [ "$(git ls-files -s "$f" | awk '{print $1}')" = "100755" ] || { echo "::error::${f} is not committed executable (mode 100755)"; exit 1; }
           done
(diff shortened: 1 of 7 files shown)


rbuergi:
Corrected the tracked-file floor in `hosting-operator.yml`: `main` already tracks **24** files under `deploy/aks/operator/bin/` (21 commands + `run.sh` + `_common.sh` + `_audit.jq`) while the guard declared 23 and its comment said "20 commands" — stale, so one lost file would have passed. This PR adds one command, so the floor is now **25** with the comment re-counted. 🚨 The four sibling PRs (#4103, #4105, #4109, #4110) each do the same on the same line: whichever lands after the first must merge `main` and set the floor to the real count at that point (26, 27, 28), not keep 25.

rbuergi:
Why `crane` rather than the mesh's own registry package, recorded because a reviewer will ask: the in-mesh **Container Registry** package (the `/v2` surface a portal served from blob storage) was **retired on 2026-09-08** — no image was ever published into it and portal hosts now answer 404 there (`MeshWeaver.Plugins/WhatsNew/2026-09-08-…`, MeshWeaver#3353/#3697). `cr.meshweaver.cloud` is a separate deployed service (`distribution` + `docker_auth`), so there is no in-mesh API to push through, and the operator image carried no registry client at all (`az`, `kubectl`, `helm`, `pg_dump`, `jq`).

The **push** credential is deliberately the registry's `publisher` account, not an instance key: an `mwi_` key is what the auth service grants **pull** for and cannot push. That is the same credential the hand step in MeshWeaver.Plugins#1722 used.

rbuergi:
**Merge resolved by content; floor settled at 25 — measured, not assumed.**

`origin/main` merged in (no rebase), one push. All three conflicts were *both sides added*, and both sides are kept:

| File | Resolution |
|---|---|
| `deploy/aks/INSTANCE-LIFECYCLE.md` | main's extended `hosting-kv-ensure` bullet (#4100) **and** this PR's guarantee bullet |
| `deploy/aks/operator/test/run-tests.sh` | main's `hosting-kv-ensure` block **and** this PR's block |
| `.github/workflows/hosting-operator.yml` | tracked-stub list = the **union** of both lists |

**The floor, and how it was counted.** `git ls-tree -r --name-only origin/main deploy/aks/operator/bin/ | wc -l` = **24** on `main` (tip `69e587c5c`): 21 `hosting-*` commands + `run.sh` + `_common.sh` + `_audit.jq`. `main` *declares* 23 with a comment saying "20 commands" — stale, so one lost script would have passed. This PR adds exactly one command, so the true count for this tree is **25 (22 commands + 3)**, and the guard and its comment now say 25. Verified on the pushed head: tracked=25, commands=22, floor=25.

**Guards run on the pushed head:** `deploy/aks/operator/test/run-tests.sh` green (both test blocks coexist), `shellcheck -S warning` clean over `_common.sh`, `run.sh` and every `hosting-*`. No run was re-run wholesale.

**The follow-up is still needed and is now precisely scoped.** Each of #4103/#4105/#4109/#4110 sets the floor true for *its own* merge, so as the siblings land the real count rises 25 → 26 → 27 → 28 while the floor stays 25 — up to 3 files of slack. After all four have merged, a one-line PR sets `-lt 28` and `expected 28 (25 commands + run.sh + _common.sh + _audit.jq)`. It cannot be pre-set here: 28 would fail the guard for whichever lands first.

rbuergi:
The merge-queue red on this PR's group build ([`34744471576`](https://github.com/Systemorph/MeshWeaver/actions/runs/34744471576), shard 4, `DanglingNodeTypeUpdateTest.Upsert_WithTheNamedImportEscapeHatch_StillLands`, `OwnerUnreachable … no verdict within 31s`) is **unrelated to this PR**: the diff here is `deploy/aks/operator/*` + `hosting-operator.yml`, zero `src/`, and `main` is green on the same base (`cdb2878bb`, run 34720891908). The failure is a race in the hub reply path — the owner posted its `PatchDataResponse` at +2 ms and the requester never received it — filed as #4159 with the trail and the instrument that will name the losing hub next time (#4154). Re-queueing this PR is safe with respect to that failure.

PR #4122, opened 2026-09-12
ci(module-pack): fetch module-pack-batch.py at the lane's scripts-ref — Plugins main published nothing since #4096

rbuergi:
## MeshWeaver.Plugins `main` has published NOTHING since 16:43Z today

Every batch leg of Plugins main (run [34707906260](https://github.com/Systemorph/MeshWeaver.Plugins/actions/runs/34707906260), head `7726c7e0`, and every push since) fails in `Module bundles / Module bundle (batch n/7: …)` at step **This leg's modules**:

```
python3: can't open file '/home/runner/work/MeshWeaver.Plugins/MeshWeaver.Plugins/meshweaver/.github/scripts/module-pack-batch.py': [Errno 2] No such file or directory
##[error]Process completed with exit code 2
```

All 7 batch legs red → `All selected bundles built` red → every gate hanging off it red → no seal, no publication. Stalled since #4096 merged at **2026-09-12 16:43:55Z**.

## Root cause (verified by content)

`.github/workflows/node-repo-module-pack.yml` (core `origin/main`, `65bdead0d`):

- **`pack` L1809** and **`tests` L3660**: `Check out Systemorph/MeshWeaver at the pinned ref` — `ref: ${{ inputs.platform-ref }}`, `path: meshweaver`.
- **`pack` L1827** / **`tests` L3674**: `python3 meshweaver/.github/scripts/module-pack-batch.py --state … init` — the first of ~40 `bk()` invocations per leg.
- **`select` L927**: checkout at `inputs.build-logic-ref || inputs.platform-ref`; **L951**: `python3 meshweaver/.github/scripts/module-pack-batch.py --self-test`, unconditional.

`platform-ref` for an unpinned caller is the newest **sealed** platform set — the failing run resolved `a7450f3aa` (merged 14:46Z), two hours before `module-pack-batch.py` existed on core (`dcabce799`, via #4096). The script only lands in a caller's checkout once a seal carries it; batching is default-on (`batch-size` 6) and `select` self-tests the script unconditionally, so no caller could opt out. Plugins' `select` survived only because it passes `build-logic-ref` (core main's tip); `pack`/`tests` read the platform checkout and died. MeshWeaver.SocialMedia (calls the lane `@main`, passes neither) is stranded in `select`.

This is the **scripts-ref vs platform-ref split** the gate lane already lives by (#3842; #4095 hit the same trap for `compose-gate-host.sh` and fetched it at `scripts-ref` with curl). `github.job_workflow_sha` is empty inside a reusable called cross-repo, so the lane's own ref must be an input.

## Mechanism (same as #4095)

- New input **`scripts-ref`** on `node-repo-module-pack.yml`, the gate's contract: empty (default) → `build-logic-ref` → `platform-ref`. Plugins passes `build-logic-ref` = core main's tip at run time, so **it is unstranded by this merge with no caller change**. A caller of the lane `@main` that pins nothing should pass `scripts-ref: main` (SocialMedia's own `no-pins` rule already says lanes with a `scripts-ref` input get `main` — that is a one-line follow-up there, not touched here).
- **`select`, `pack`, `tests`** each get the step `Fetch module-pack-batch.py at the lane's scripts ref` before the first use: `curl -fsSL --retry 3` with `Accept: application/vnd.github.raw+json` at `?ref=${SCRIPTS_REF}` (curl, not `gh` — `pack`/`tests` honour `inputs.runner` and the self-hosted image ships no `gh`), written **into `meshweaver/.github/scripts/module-pack-batch.py`** so every `bk()` reads the fetched bytes; asserts non-empty, asserts `#!`, runs `--self-test` on the fetched bytes. RED by name on a miss (`could not fetch module-pack-batch.py at <ref> … a caller of this lane at @main passes scripts-ref: main`), on an empty ref, on a non-script. The existing self-test line in `select` stays; nothing is weakened or skipped.
- Simulated locally against the real API: at `main` → fetched (31,888 bytes), self-test green, exit 0; at `a7450f3aa` (the sealed set) → 404 → `::error::could not fetch module-pack-batch.py at a7450f3aa…`, exit 1; empty ref → `::error::… all empty`, exit 1.
- `batch-size` default unchanged. MeshWeaver.Plugins not touched.

## Guard so it cannot recur — `ScriptsRefFetchGuard` (test/MeshWeaver.Documentation.Test)

Scans every job of every workflow for `meshweaver/.github/scripts/<x>.(py|sh)` references (comments stripped). Each must be (a) fetched **in that job** at `?ref=${SCRIPTS_REF}` with `SCRIPTS_REF: ${{ inputs.scripts-ref || … }}`, and the fetching step must precede the first use — or the job's whole `meshweaver` checkout is at `inputs.scripts-ref` (the gate's `merge` job shape) — or (b) on an explicit allow-list of scripts old enough to be in every sealed set, each entry carrying its landing date (`88201f021` 09-07 ×8, `3a5f44b02` 09-09 ×2, `0c7e29e50` 09-10). The trap is stated in the class doc and in the failure message. A stale allow-list entry is RED too (ratchet). Plus a theory pinning the fetch step's shape in `select`/`pack`/`tests` (curl, raw media type, target path, `#!`, `--self-test`, no `continue-on-error`, no `::warning::`) and the `scripts-ref` input.

**Deletion proof (by content):**
- Lane **as merged on `origin/main`** (the defect): `EveryLaneScriptReadFromTheMeshweaverCheckout_IsFetchedAtScriptsRef_OrOldEnough` **fails**, naming `select`, `pack`, `tests` — *"runs meshweaver/.github/scripts/module-pack-batch.py out of the platform-ref checkout without fetching it at ${SCRIPTS_REF}, and it is not on the old-enough allow-list"*.
- Fixed lane with only `pack`'s fetch step deleted: **2 failures** — the general rule naming `pack`, and `ModulePack_FetchesTheBatchScriptAtItsScriptsRef…(job: "pack")`: *"the step `- name: Fetch module-pack-batch.py at the lane's scripts ref` is gone"*.
- Fixed lane: 5/5 green; full `MeshWeaver.Documentation.Test`: **446/446**.

## Guards run

- `actionlint -shellcheck= -pyflakes=` (CI's flags, all workflows): clean.
- `check-workflow-shell` / `-timeouts` / `-yaml-keys` / `check-pr-secret-preflight` / `check-workflow-permission-pairing` — `--self-test` all green; `--root .` 0 violations each (shell: 1 pre-existing allow-listed finding, 0 live); permission-pairing `--fleet .github/lane-caller-grants.yml`: 56 pairs, 0 violations.
- `check-combo-verify`, `check-pin-set-consistency`, `test-cd-steps`: green.
- `dotnet test test/MeshWeaver.Documentation.Test -c Release -warnaserror`: 446 passed.

Also: `ModuleBuildArchitecture.md` gets the note under the batching section; the script's docstring no longer claims it is fetched at `platform-ref`.

**After merge:** Plugins' next `main` run (lane `@main`) fetches the script at `build-logic-ref` and publishes again — no seal needed. Hop: this is the lane, not an image; nothing to roll.

Refs #4096

https://claude.ai/code/session_01DqTDZe4Skkn81GuGkwMrXS

### DIFF
--- a/.github/scripts/module-pack-batch.py
+++ b/.github/scripts/module-pack-batch.py
@@ -3,7 +3,9 @@
 per-module bookkeeping that keeps a batch from becoming a coupling.
 
 (The name on this first line is load-bearing: the module-pack lane fetches this file at the
-caller's `platform-ref` and its `select` job self-tests it on every run.)
+lane's `scripts-ref` — NOT out of its `platform-ref` checkout, which for an unpinned caller is the
+newest sealed set and did not have it (#4096, Plugins main 2026-09-12) — and every job that runs
+it self-tests the fetched bytes first.)
 
 WHY THIS EXISTS (measured 2026-09-05..12)
 ------------------------------------------
--- a/.github/workflows/node-repo-module-pack.yml
+++ b/.github/workflows/node-repo-module-pack.yml
@@ -386,6 +386,18 @@ name: node-repo module pack
 # (`Module bundles`), which is untouched. A caller that wants the old one-leg-per-module shape
 # passes `batch-size: 1`. Doc: Doc/Architecture/ModuleBuildArchitecture ("The pipeline, end to
 # end", the 2026-09-12 batching note).
+#
+# 🚨 module-pack-batch.py is the LANE's script and is FETCHED at the lane's SCRIPTS ref
+# (`scripts-ref`, then `build-logic-ref`, then `platform-ref`) in `select`, `pack` and `tests` —
+# never read out of the `meshweaver/` checkout those jobs make at `platform-ref`. The checkout is
+# the PLATFORM's commit, and for an unpinned caller that is the newest SEALED set, hours behind a
+# lane called at `@main`: #4096 merged at 16:43Z on 2026-09-12 and every `pack` leg of Plugins main
+# read the script out of a checkout sealed at a7450f3aa (14:46Z) — `python3: can't open file` —
+# so Plugins published nothing until a seal carried it. This is the scripts-ref / platform-ref
+# split node-repo-gate.yml already lives by (#4095, compose-gate-host.sh). A caller of this lane
+# at `@main` passes `scripts-ref: main`. ScriptsRefFetchGuard (MeshWeaver.Documentation.Test) fails
+# any lane job that runs a `meshweaver/.github/scripts/*.py` it neither fetched at scripts-ref nor
+# lists as old enough to be in every sealed set.
 
 on:
   workflow_call:
@@ -410,6 +422,26 @@ on:
         description: The Systemorph/MeshWeaver commit the modules build against (a pin, not a branch).
         required: true
         type: string
+      scripts-ref:
+        description: >-
+          The Systemorph/MeshWeaver ref this lane's OWN helper scripts are fetched at — today
+          `module-pack-batch.py`, the batching bookkeeping every `pack` and `tests` leg drives and
+          `select` self-tests. Same contract as node-repo-gate's `scripts-ref`: distinct from
+          `platform-ref`, which names the PLATFORM the modules build against and is, for an
+          unpinned caller, the newest SEALED set — hours behind the lane called at `@main`.
+          Empty (the default) falls back to `build-logic-ref` (this lane's existing pin for its
+          selection scripts — an immutable core SHA a caller resolves at run time) and then to
+          `platform-ref`, right for a caller that pins the lane and the platform to one commit.
+          🚨 A caller that pins NOTHING passes `main` here: on 2026-09-12 (#4096) this lane's
+          batching script landed on core main at 16:43Z, every `pack` leg of MeshWeaver.Plugins
+          main read it out of the checkout at `platform-ref` — a set sealed at a7450f3aa, two
+          hours earlier — and died with `python3: can't open file '…/module-pack-batch.py'`, so
+          Plugins published nothing until a seal carried the file. `github.job_workflow_sha` cannot
+          stand in: it is EMPTY inside a reusable called from another repository (measured
+          2026-08-30), so the lane's own ref is an input, never inferred.
+        required: false
+        type: string
+        default: ''
       publish:
         description: >
           Hand the packed bundles to the registry. TRUNK-ONLY, not push-only — a PR packs and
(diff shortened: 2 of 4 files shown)


rbuergi:
**Superseded — do not merge as-is.** While this was being written, #4119 (merged 2026-09-12 18:40:47Z, 8aaf35f14) fixed the same root cause with a different mechanism: every job now checks out core at `inputs.build-logic-ref || job.workflow_sha` into `build-logic/` and runs `$MODULE_PACK_BATCH` from there; `meshweaver/` keeps only the platform-semantic probes (`check-module-platform-floor.py`, `module-owned-platform.sh`). `job.workflow_sha` is documented as "the commit SHA of the workflow file that defines the current job" (unlike `github.job_workflow_sha`, which the gate measured empty cross-repo on 2026-08-30) — so the read path this PR fetches into no longer exists on `main`, and the PR is `mergeable_state: dirty` against it. Auto-merge is armed but cannot fire on a conflict; I have not force-pushed (one push, per the brief) and cannot convert to draft over REST.

What is still unclaimed by #4119 and could be lifted into a small follow-up if wanted: **`ScriptsRefFetchGuard`** (test/MeshWeaver.Documentation.Test) — a fleet-wide rule that every `meshweaver/.github/scripts/<x>` a lane job runs is fetched in that job at the lane's own ref, or is on an explicit allow-list of scripts old enough to be in every sealed set (with landing dates; stale entries RED). It fires by content on the lane exactly as #4096 merged it, naming `select`, `pack` and `tests`. On today's `main` it would pass (the two remaining `meshweaver/` reads date from 2026-09-07) and hold the line for the next script that lands with the step that runs it.

⚠️ Not yet proven in production: no MeshWeaver.Plugins `main` push has run since #4119 landed (last: 34707906260 at 17:19Z, red), and no core CD run on 8aaf35f14 has reached the lane's `select` — the first cross-repo run is what shows `job.workflow_sha` resolving. Plugins main publication has been stalled since 16:43Z.

rbuergi:
Closing: superseded by #4119 (merged 18:40:47Z), which fixes the same root cause by checking the lane's own tooling out at `job.workflow_sha` into `build-logic/` — the fetch path this PR added no longer exists on main. The one part #4119 does not carry, the `ScriptsRefFetchGuard` that refuses a lane script read from the platform-ref checkout, is worth landing on its own; filed as a follow-up rather than rebased here.

🤖 Filed from Claude Code.

PR #4154, opened 2026-09-13
fix: a handler that faults past DisposeHostedHubs answers its requester (#4072)

### LINKED ISSUE #4072: A handler that FAULTS past DisposeHostedHubs never answers its requester: ReportFailure declines, NackThroughParent is not reached (the live half of #1151, and Reinsurance#178's compile-state wave)
## What

A `SubscribeRequest` whose handler ENTERS and FAULTS on a hub that is past `DisposeHostedHubs` is **never answered**. The requester hears nothing, holds its callback open for its whole budget, and re-asks — which is the climbing-sequence retry wave [`MeshWeaver.Reinsurance#178`](https://github.com/Systemorph/MeshWeaver.Reinsurance/issues/178) reports as its second signature on a loaded gate host.

It has a **14-second in-process reproduction on core `main`**, in a test that **PASSES**.

## Measured — core `main` `4c99ec2691b893bcac34bb7c30e03121924c7ad3`, 2026-09-12

```
dotnet test test/MeshWeaver.Compiler.Pipeline.Test -c Release \
  --filter "FullyQualifiedName~LeavingHubAdoptionSweepTest"
→ Passed!  -  Failed: 0, Passed: 2, Skipped: 0, Total: 2, Duration: 14 s
```

The verdict is green; the finding is in the dispose diagnostics in the `.trx` (they do not reach stdout — read `TestResults/*.trx`). From `ABundleWhoseFingerprintDisagreesWithTheLiveSource_IsDeclinedBeforeItWrites`:

```
[DISPOSE] LeavingHubAdoptionSweepTest: still waiting after 3002ms — snapshot:
Hub mesh/… RunLevel=DisposeHostedHubs Disposal=Pending Queue(buffer=0,deferred=0,drainsInFlight=0,openGates=0,draining=False)
HostedHubs (1):
  Hub cache/… RunLevel=Quiescing Disposal=Pending Queue(buffer=0,…)
    PendingCallbacks=1[vP64TxXvQ0295XzJIZoIOw=SubscribeRequest@type/StaleBundleType/_Activity/compile-state(3008ms)]
  handler-side fate (what happened to the delivery):
    … → ROUTED onTarget=True state=Submitted@type/StaleBundleType/_Activity/compile-state(+15ms)
      → HANDLER_ENTER@type/StaleBundleType/_Activity/compile-state(+16ms)
      → HANDLER_FAULT TargetInvocationException@type/StaleBundleType/_Activity/compile-state(+16ms)
  ⇒ the chain FAULTED and no reply was posted — the fault is the cause;
    find why its error arm does not answer the requester.
```

The fault lands at **+16 ms**. Three seconds later the requester has still heard nothing, and disposal is held open by that one callback. (Teardown then completes at 4003 ms — the budget absorbs it, which is why nothing is red.)

## Where the answer is lost — read, not guessed

`MessageService.cs` `.Catch((Exception e) => …)` (around line 2131) forks three ways:

1. `delivery.Message is ExecutionRequest` → the caller's `ExceptionCallback`.
2. `HubDisposingException.IsHubDisposal(e)` → `NackThroughParent(...)`, with the tombstone fork (authoritative `NotFound` for a deleted address, transient `ShuttingDown` for one that may reactivate).
3. **`else` → `logger.LogError(...)` then `ReportFailure(delivery.Failed(e.ToString()))`.**

And `ReportFailure` (line ~858) contains:

```csharp
// Don't post DeliveryFailure during shutdown - recipients are likely also disposing
// and the messages just clog the pipeline
if (hub.RunLevel >= MessageHubRunLevel.DisposeHostedHubs)
```

So on arm **3**, past `DisposeHostedHubs`, the NACK is computed and then **declined**. The requester is never told.

🚨 That is the *same* loss the long comment block above arm 2 describes and fixes — *"ReportFailure posts through OUR OWN hub, and it declines to post at all once RunLevel >= DisposeHostedHubs … so the NACK it so carefully classifies was computed, logged, and then silently dropped — the sender heard nothing and burned its whole RequestTimeout."* `NackThroughParent` is the primitive built for exactly this. **Arm 2 was given it; arm 3 was not.**

## What I checked and ruled OUT

- **It is not a missing unwrap.** `HubDisposingException.IsHubDisposal` walks the whole chain (`ExceptionChain.Contains<HubDisposingException>`), and its XML doc names `TargetInvocationException` from reflective dispatch explicitly. So the wrapper is not why arm 2 is skipped.
- **It is not #1151 regressing.** #1151 separated two questions and closed on the first: *"why `SubscribeToClient` throws for this node"*, plus the accepted-then-dropped delivery on a disposing hub (#1153 + #1154). Its own body separated the second — *"why a faulted handler does not NACK the requester — the second turns any such fault into a 60s stall instead of an immediate, legible error"* — and that half is what arm 3 still is.

## The candidate this points at, stated as a hypothesis

`HubDisposingException.IsDisposedContainer` exists, is documented for precisely this situation, and **is not consulted anywhere in `MessageService`'s fault chain**. Its own XML doc:

> 🚨 This is a TEARDOWN fact, not a fault of the work that hit it … a hub whose scope has been closed cannot serve anything, so a delivery that faults this way must be answered as RETRYABLE (the address reactivates) rather than as a result. … `HubDisposingException` covers only the case where the hub announced its own disposal, and **a scope closed underneath a live delivery never does**.

A handler faulting because its Autofac scope closed underneath it is therefore invisible to arm 2 by construction and falls to arm 3. That fits the snapshot exactly (mesh at `DisposeHostedHubs`, the cache hub `Quiescing`, the fault 16 ms after the handler entered).

🚨 **I have NOT established that this is the exception.** The `.trx` carries only the wrapper type (`TargetInvocationException`), not the inner one. **What would settle it:** log the full chain at the arm-3 `LogError` (it currently logs `{Exception}`, so the inner type should already be in the test host's log — it just is not in the dispose snapshot), or assert on it directly in a test that drives a handler fault while the hub is past `DisposeHostedHubs`.

## Why it matters beyond a slow teardown

On a test host the 3 s is absorbed by the disposal budget and nothing reds — which is why this has survived. On a **loaded gate host** the same shape is a wave: each retry enters, faults, answers nothing, and the requester re-asks, so the sequence number climbs and the backlog never clears. That is `MeshWeaver.Reinsurance#178`'s second signature verbatim, and it is the half of that issue which is not explained by the node-CRUD serialisation work (#2543 / #3802) or the update-queue cold-factory fix (#3818).

## Suggested fix

Give arm 3 the same route arm 2 has: `NackThroughParent` first, `ReportFailure` only as the fallback for a hub with no parent or a parent already past `DisposeHostedHubs` — preserving the "ONE request, ONE failure" rule. Classifying a closed-scope fault as retryable (via `IsDisposedContainer`) is the separate, smaller half, and it is what decides whether the NACK says `ShuttingDown` or `Unknown`.

## Acceptance

A test that drives a handler fault on a hub past `DisposeHostedHubs` and asserts the requester receives a `DeliveryFailure` rather than nothing — and `LeavingHubAdoptionSweepTest`'s dispose snapshot showing no `PendingCallbacks` for a faulted delivery.

https://claude.ai/code/session_01KFBiMVHLTZM2axNgVvaZN2

rbuergi:
Fixes #4072.

## The defect, reproduced

`LeavingHubAdoptionSweepTest` (core `main`, 14 s, **passing**) tears its mesh down with a `SubscribeRequest` in flight from `cache/…` to a per-node hub. With the fate ledger stamped at every carrier (this PR), the responder's trail reads:

```
HANDLER_FAULT TargetInvocationException→TargetInvocationException→HubDisposingException (+56ms)
  → NACK_DECLINED reason=parent-DisposeHostedHubs (+62ms)
  → FAILURE_REPORTED errorType=ShuttingDown runLevel=Dead (+63ms)
  → RESPONSE_POSTED type=DeliveryFailure target=cache/… (+63ms)
  → REPLY_REFUSED_SHUTTING_DOWN runLevel=Dead parent=mesh/…@DisposeHostedHubs (+63ms)
```

Before this PR the trail ended at `HANDLER_FAULT`: `ReportFailure` declined on its own run level (`>= DisposeHostedHubs`) before reaching the post seam — the loss the issue names — and `NackThroughParent` had already declined because the parent was past `DisposeHostedHubs`. The requester (`cache/…`, Quiescing, `PendingCallbacks=1`) re-armed its quiesce budget for the answer ([QUIESCE-WAIT] — "a shutting-down sibling answers before it goes") and got `HubDisposedBeforeResponseException` at 4 s. Mesh teardown: **4.0 s** with a "still waiting" snapshot at 3 s → **2.0 s** and no snapshot after this PR.

Why a handler runs on a hub past `DisposeHostedHubs` at all: `AccessControlPipeline` detaches `next` from the turn (decision outside the gate), so the handler runs concurrently with the hub's own shutdown phases — the hub read `Dead` 7 ms after the fault.

## The fix (all in `MeshWeaver.Messaging.Hub`)

1. **`ReportFailure` has no run-level gate any more.** The gate predates `PostImplGeneric`'s forward-through-a-live-parent seam and could only lose answers. The seam decides, and stamps `REPLY_REFUSED_SHUTTING_DOWN runLevel=… parent=…` on the request's trail where it cannot carry.
2. **`MessageHub.TryDeliverNackInProcess`** — the NACK's carrier of last resort: a `DeliveryFailure` carrying `RequestId` that no transport can take is handed straight to the in-process hub its requester lives under (host chain climbed to the root, descended, never created), when that hub still admits replies (`< DisposeHostedHubs`). NACKs only — nothing to reorder — and offered only where the drop happens: the post seam's refusal arm, the two routing drop arms after `IUndeliverableReplySink` declined, and (through `ReportFailure`) the intake gate's tier 2, which used to return without answering because "there is genuinely no route".
3. **`IsDisposedContainer` is a teardown fact** in the handler fault chain: answered `ShuttingDown` in the owner's refusal vocabulary (`ShutdownNack.IsAnsweredByOwner`), not `Unknown`.
4. **The instrument**: `HANDLER_FAULT` names the inner types; every carrier tried or declined is a stage (`NACK_DECLINED reason=…`, `NACKED_THROUGH_PARENT`, `FAILURE_REPORTED`, `REPLY_FORWARDED_THROUGH_PARENT`, `NACK_DELIVERED_IN_PROCESS`, `REPLY_REFUSED_SHUTTING_DOWN`); the verdict reads the last post-seam stage instead of `RESPONSE_POSTED` alone.

## Test and falsification

`HandlerFaultPastDisposeHostedHubsTest` (4 cases, 0.5 s): a delivery-pipeline stage detaches the handler exactly as production does and releases it once the fence is read back — targeted teardown (parent live, genuine fault → `Unknown` through the parent) and whole-tree teardown (parent past `DisposeHostedHubs`, requester Quiescing on a 5-minute budget; `HubDisposingException` / disposed-scope → `ShuttingDown`, genuine → `Unknown`; NACK arrives in ~2 ms, the requester leaves Quiescing on the answer).

Reverting the production change (tests kept): **4/4 fail** — every requester times out unanswered (`TimeoutException` after 36 s), and the whole-tree cases hold the mesh teardown open on the pending callback for 46 s. Restored: 4/4 green.

Also green with the fix: `MeshWeaver.Messaging.Hub.Test` 413/413; the teardown/NACK slice of `MeshWeaver.Graph.Test` (62/62 — `NackReachesTheWaiterDuringTeardownTest`, `RefusedReplyReachesTheWaiterTest`, `UndeliverableReplyShapesTest`, `LateNackReenqueue*`, `UpsertAnswersWhenTheOwnerGoesAwayTest`, …); `LeavingHubAdoptionSweepTest` 2/2; `DocumentationLinkIntegrityTest`.

## Residue, named

A **typed** reply refused by a closed parent during a whole-tree teardown still drops (the documented `RefusedRepliesDuringTeardown` asymmetry): in the same repro `nodeops` answers `type/StaleBundleType`'s compile-state upsert with `CreateOrUpdateNodeResponse.Fail`, which no carrier takes, so that sibling still pays one 2 s quiesce budget. That is the remaining 2.0 s.

Doc: `Doc/Architecture/RefusedRepliesDuringTeardown` (new section "The NACK's own carrier of last resort — #4072"). What's New: Fix.

Pairs-with: none — no public type or member leaves `src/`; the new members are `internal`/`private`.


rbuergi:
Addressed in `7a2cccba1`, all three:

- **`NackThroughParent` on a refused parent `Post`** — now reads the returned delivery; a `Failed` (the parent crossed the shutdown boundary after the check, or its post pipeline rejected) stamps `NACK_DECLINED reason=parent-post-refused` and returns `false`, so `ReportFailure` and the in-process carrier are still tried.
- **`TryDeliverNackInProcess` discarding `DeliverMessage`'s verdict** — it now returns the intake's verdict (`Failed`/`Ignored` ⇒ not delivered), so a requester that crossed `DisposeHostedHubs` between the check and the call is not reported as answered.
- **`IsDisposedContainer` over-classifying** — the arm now uses the probe-gated `hub.IsTerminatedByScopeTeardown(e)` (an `ObjectDisposedException` in the chain AND this hub's own scope disposed), the same classifier `HandleInitialize` and the permission fold use; a handler that reached into some other disposed scope keeps its genuine-fault verdict. The test's disposed-scope case now expects `Unknown` on a live scope for exactly that reason.

Plus, since the review: the ledger now writes a reply's own journey onto the request's trail (`↩ reply#1: …`), because this morning's merge-queue red (#4159) ended at `RESPONSE_POSTED ⇒ chase the response delivery` with nothing to chase it by.

PR #4164, opened 2026-09-13
docs(teams): the conversation link is content at Admin/_TeamsConversation/{key}, never under the thread (Plugins#1665)

rbuergi:
> 🚨 **HELD — do not flip on green.** This page describes the link placement that Systemorph/MeshWeaver.Plugins#1773 introduces, so it lands WITH or AFTER #1773. #1773 is draft under the fleet-wide Plugins drafting until the build is fixed. The coordinator flips both; a green wall here is not the signal.

Docs follow-up to Systemorph/MeshWeaver.Plugins#1773 (fixes Plugins#1665).

`Doc/AI/TeamsBot` and the remark on `TeamsConversationNodeType.Segment` said the `TeamsConversation` link lives at `{threadPath}/_TeamsConversation/…`. That is under the configured `_Thread` satellite segment, where no content query on either backend could find it (the in-memory provider mirrors Postgres' table split) — which is why `FindThread` never found a link and `TeamsReplySender` never posted a reply. #1773 moved the link to `Admin/_TeamsConversation/{key}` (the WhatsApp channel's shape); this PR records that where the next reader looks, and names the separate sender→user mapping gap (Plugins#1774) beside the sentence that describes the mapping.

Doc-only plus one XML remark: `DocumentationLinkIntegrityTest` + `NoConflictMarkersGuard` green; `MeshWeaver.Graph` Release `-warnaserror` clean.

Pairs-with: none — nothing removed.
Mirror-sync: none.


rbuergi:
## Review round (d0161e5f) — thread replies answer HTTP 502/500 at the moment, so here:

- **"Every content query" too broad:** Taken: the sentence now says a node beneath a thread path is excluded from every UNTARGETED content query (a query that targets the satellite path is routed to that table), and names the old mesh-wide lookup as the untargeted one.
- **`TeamsConversation.cs` still advertises the thread placement:** Taken: the record's own summary now names Admin/_TeamsConversation/{key} and says why the thread placement was found by nobody; the Id remark no longer calls it a satellite id.

PR #4165, opened 2026-09-13
feat: leg 5 of the #890 canary — the pristine-COMPILER control (compiler=PRIVATE-COPY-EMITS / THREW)

### LINKED ISSUE #890: CI: one Roslyn Emit poisons the whole process — every later compile NREs in Cci.MetadataWriter.GetConsolidatedTypeParameters, shard ends exit=124
Split from #612 — a SECOND, distinct defect observed in PR #884 run 31153582625 shard 2 (2026-08-07), NOT explained by the empty-source-snapshot race fixed there.

## Evidence

Every compile in that test process failed with `CompilationError = Object reference not set to an instance of an object.` — including `CodeEditRecompileTest`'s deliberately-invalid `BrokenType`, whose failure should be a Roslyn diagnostic, i.e. the NRE strikes PRE-Roslyn. Source discovery was healthy (`IsDirty=True`) and release triggers were handled (fresh `LastCompilationActivityPath` minted per retry). The wedge cascaded across classes on the shard (OverlaySelfHealInstanceRecycle → LinkedInTelemetryImport → PandasExplorer → CodeEditRecompile → ReleaseRequestWatcherHighWater), consistent with a process-wide condition.

## Why it is currently undiagnosable, and what changed

Every sink reduced the fault to `.Message` — no stack anywhere. The #612 fix adds: exception TYPE + full stack on the activity log for non-Roslyn aborts, `{TypeName}: {Message}` on the node's `CompilationError`, and the exception object on the terminal `LogWarning`. **The next CI occurrence will name its throw site** — triage should start from that.

## Probed and excluded (locally green)

- `DOTNET_ENVIRONMENT=Development` + `Logging__LogLevel__Default=Warning`
- `#if CIRun`-gated code (none exists)

Remaining suspects: Linux-only or load/saturation-dependent null-race in the pre-Roslyn compile pipeline.

rbuergi:
Builds the control #890 has been waiting for since its 2026-09-10 specification and its 2026-09-12 supersession check ("the decisive control is still not implemented"). Does not close #890 — the closing condition is the READING of this leg on an occurrence.

## What it is

`PrivateRoslynCopy.Emit` — **leg 5** of the emit canary, the pristine-COMPILER control. Legs 1–4 all execute the one shared `Microsoft.CodeAnalysis*.dll` (same mapped image, same native code, same statics), so none of them can separate "the process cannot emit" from "the shared copy of Roslyn cannot emit" — which is why `BELOW-ROSLYN` has been reading "not the references and not the source" as "not Roslyn" on every occurrence.

On the same two verdicts legs 3 and 4 ride (`BELOW-ROSLYN`, `DIVERGENT`), it loads `Microsoft.CodeAnalysis` + `Microsoft.CodeAnalysis.CSharp` a second time — from freshly read bytes into a collectible `AssemblyLoadContext` that serves the two to each other and lets every other dependency fall through to the default context — drives `EmitCanarySource` through that private copy by reflection against an image-backed CoreLib the private copy builds itself, and unloads.

| `compiler=` | meaning |
|---|---|
| `PRIVATE-COPY-EMITS` | a never-executed Roslyn emits the source the shared one cannot, same process, same CLR ⇒ the broken state lives IN the shared copy (a static, a cache, a miscompiled method); scope the compiler out per generation — not a runtime bug |
| `PRIVATE-COPY-THREW … at …` | even fresh compiler code cannot emit ⇒ `BELOW-ROSLYN` earned; the private copy's frame is the repro to file |
| `PRIVATE-COPY-DIAGNOSTICS(…)` | compiled and refused — a shape no other leg has produced; read the ids |
| `PRIVATE-COPY-UNAVAILABLE(…)` / `NOT-RUN` | the leg could not run / no probe — stated, never folded into a verdict |

Cost: two loads (~15 MB read), one tiny emit, one `Unload()`, on the already-failing terminal path only. Measured ~1 s on a healthy process including both loads.

## Test

`EmitCanaryPrivateCompilerLegTest`: on a healthy process the reading is `PRIVATE-COPY-EMITS`, twice in a row (the context is re-created per run) — a control that cannot run would retire the discriminator silently, and a Roslyn upgrade that reshapes `Create`/`Emit` now reds this test instead of turning the next occurrence into `UNAVAILABLE`. Both verdict shapes render the `compiler=` reading; its absence is named; a faulting probe reports `UNAVAILABLE`. The existing canary suites (`EmitCanaryControlTest`, `EmitCanaryFlatLegTest`, `EmitCanaryDissectionTest`) stay green.

The reflection binds to the private copy's own types by full name and fills every optional parameter with its declared default, so it names nothing of Roslyn's optional-parameter surface.

Doc: `Doc/Architecture/NodeTypeCompilation` → "Leg 5 — the pristine-COMPILER control". What's New: Feature.

Pairs-with: none — a new `internal` class and an optional parameter on an `internal` method.


rbuergi:
Addressed in the follow-up commit, all four inline and the two suppressed:

- **`THREW` only from the emit stage** — the reflective stage is tracked; a throw from parse / reference / options / create reports `compiler=UNAVAILABLE(stage=… threw …)`: the private copy was never asked to emit, so it has not answered.
- **Immutable collections** — the diagnostic ids accumulate in an `ImmutableList<string>`; the load context holds an `ImmutableDictionary<string, Assembly>` replaced whole via `SetItem`.
- **One `compiler=` format** — the doc's 2026-09-10 specification section is the canonical contract; the new section is now the implementation note pointing at that table, with the tokens aligned (`UNAVAILABLE(…)` / `NOT-RUN`) and `PRIVATE-COPY-DIAGNOSTICS(<ids>)` added to the table.
- **The `EMITS` inference** — reworded to the specification's: the fault travels with *this process's copy of the compiler* (its image, its mapping, or the native code produced for it), not with the CLR heap; the tiering residual (one emit does not separate fresh mapping from fresh native code) is restated in the verdict, the class remarks and the doc.
- **The memory emit path** — `EmitToMemory` now stamps the canary verdict on an emit-phase throw exactly as the disk path does; a deployment emitting to memory used to lose every `#890` reading, legs 1–5 alike.

PR #4162, opened 2026-09-13
fix: a transient infrastructure fault at init retires the activation instead of latching it FAILED (#4067, #4068)

rbuergi:
Fixes #4067. Fixes #4068.

## The defect

Both incidents are one mechanism: a hub whose initialization met a **transient infrastructure fault** entered the FAILED state — `InitializationError` set, a front-of-chain rule answering every later request with a terminal `DeliveryFailure{Failed}`, streams errored, nothing ever re-run — until a recycle or a process restart. That latch is the right outcome for a fault *of the activation* (a NodeType that does not compile, a handler that throws). It is wrong in kind for a fault *of the dependency*: nothing about the activation is broken, the database was away for a moment, and the latch turned a two-minute DNS blip into "this address is broken until the process restarts".

- #4068: `DataContext.SettleInitializationGate` — three DataContexts FAILED on `SocketException: Name or service not known` in two minutes (memex-cloud, 2026-09-12). The three `mesh/…` ids are three boots of one pod: the ROOT's recovery is the process, and that is what happened.
- #4067: `MessageHub.HandleInitialize` — 19 `sync/*` BuildupActions FAILED on `NpgsqlException: Failed to connect … ---> TimeoutException` across three weeks (memex).

Neither is a `ReplaySubject` latch nor a failed `Lazy`: it is the FAILED state machine applied without classifying the fault, and no path back.

## The fix

1. **`InfrastructureFault.IsTransient`** (Messaging.Contract) — `DbException { IsTransient: true }` (every ADO.NET provider classifies its connection failures there; Npgsql for exactly the two shapes measured, no provider reference in core) or a `SocketException`, anywhere in the chain (`ExceptionChain`, so an `AggregateException` or reflective wrapper does not hide it). A bare `TimeoutException` is deliberately NOT transient — the init time-box mints one for a hang, and a hang is a defect.
2. **`WithReactivationOnDemand()`** on `MessageHubConfiguration` — "demand routing re-creates this hub after it goes away" — declared in the one funnel every per-node activation passes (`NodeTypeRebindWatcher.WithNodeTypeRebind`, Monolith routing and `MessageHubGrain` alike).
[...]

Issue #3112, opened 2026-09-02
Gate seed race: one bundle's adoption write times out at the owner during the mass install (ADVANCE_WITHOUT_HANDOFF → LATE_NACK_TERMINAL on Radzen/Gallery) and the consumption postcondition reds a satellite PR


rbuergi:
## Symptom

`node-repo-gate.yml` (core `4f6a727c`) went RED on MeshWeaver.Manufacturing PR #48, run [33623113056](https://github.com/Systemorph/MeshWeaver.Manufacturing/actions/runs/33623113056), job `test-repos / Compile + render node repos (MeshWeaver from ACR)`:

```
GATE FAILED — fatal: bake consumption: the gate adopted 101 of 102 baked assembly(ies) for the types it installed — 1 were DECLINED and compiled locally … DECLINED: Radzen/Gallery.
```

The decline is not a framework-identity or manifest mismatch. The seed's own log (category `bake-seed`) says the adoption WRITE never landed:

```
11:14:52.305 warn: MeshNodeStreamCache [UpdateQueue] ADVANCE_WITHOUT_HANDOFF path=Radzen/Gallery seq=681 bound=5000ms — the owner never acknowledged this write inside the bound …
11:14:57.329 warn: MeshNodeStreamHandle [UpdateRemote] LATE_NACK_TERMINAL hub=cache/HbeB6n8CD0GpZEF02br4DA target=Radzen/Gallery code=Unknown attempt=0 msg=TimeoutException — the write did NOT apply and is not auto-retryable
11:14:57.332 warn: bake-seed ShippedPrebuiltBundles: seeding Radzen/Gallery from Radzen.zip did not complete — the sweep compiles it instead  MeshNodeStreamException: MeshNode Unknown at 'Radzen/Gallery': TimeoutException
11:14:57.335 info: bake-seed ShippedPrebuiltBundles: bundle Radzen.zip: adopted 0/1 prebuilt assembly(ies)
11:14:57.341 fail: MessageHub ROUTER_TRAFFIC: RawJson has the mesh hub as sender (sender: mesh/HbeB6n8CD0GpZEF02br4DA, target: Radzen). The mesh hub is the ROUTER and must not execute work
```

This happened while `PackageInstaller` was installing the same package (`── Radzen: installing 7 file(s)…` at 11:14:4x) — the seed's adoption write and the installer's node writes race on the `Radzen/Gallery` owner, and the owner answered nothing for 10 s.

[...]

PR #4183, opened 2026-09-13
fix: a reply is an ANSWER at every tier, and a refused one is reported to the REQUESTER (#4170)

### LINKED ISSUE #4170: A TYPED reply a tearing-down responder cannot send is dropped and the requester waits out its budget — two seams (post guard: 2 s; intake gate tier 2: 31 s)
## The shape

A **typed** reply (not a `DeliveryFailure`) whose responder can no longer post it — its own pump is closed and its parent is past `DisposeHostedHubs`, the state every sibling is in during a whole-tree teardown — is DROPPED, and its requester waits out its own budget. #4154 closed this for NACKs (`MessageHub.TryDeliverNackInProcess`) and deliberately did **not** extend the carrier to typed replies: an ack must not overtake the state change it acknowledges (`Doc/Architecture/RefusedRepliesDuringTeardown` — tried twice at the patch-verdict seam, reverted twice). So the class is narrowed, named, and still live. This issue is the residue, filed so it does not live only in a PR body.

## Measured — `LeavingHubAdoptionSweepTest`, core `main` with #4154 (2026-09-13)

The mesh teardown is **2.0 s**, and that 2.0 s is one hub's quiesce budget spent on exactly this:

```
09:27:46.415  Mesh.Dispose() invoking on mesh/jhOR2F5OlUGD9PKS23wb_A
09:27:46.466  [CreateOrUpdate] inner UpdateNode faulted for type/LeavingSweepType/_Activity/compile-state
09:27:48.417  [CompileStateMirror] type/LeavingSweepType: satellite write failed (state stays on the node; next change retries)
              HubDisposedBeforeResponseException: Hub type/LeavingSweepType was disposed before the
              response arrived (request type CreateOrUpdateNodeRequest, target portal/nodeops-jhOR…)
09:27:48.420  Mesh.Disposal completed in 2006ms
```

`portal/nodeops` produced a `CreateOrUpdateNodeResponse` for a request `type/LeavingSweepType` was awaiting; the requester heard nothing and was answered by its own teardown two seconds later. Before #4154 the same teardown took 4.0 s and printed a pending-callback snapshot at 3 s — that half is fixed; this half is the remainder.

**Cost in production shape:** one quiesce budget (default 2 s) per hub that is owed a typed reply at teardown, paid in sequence down the tree, plus the work the reply carried being reported as unknown (`satellite write failed … next change retries`).

## What is established, and what is not

- **Established:** the reply existed, the requester never received it, and the requester's own teardown is what ended the wait (`HubDisposedBeforeResponseException`, which only the REQUESTER's `CancelCallbacks` raises).
- **NOT established:** *where* it was lost. Two candidates, and they have different fixes:
  1. the responder's post seam refused it (`PostImplGeneric`'s teardown guard, the correlated-reply forward declining because the parent is past `DisposeHostedHubs`) — the documented asymmetry;
  2. the reply was posted and routed fine, and the requester had already cancelled its callbacks (it reached `ShutDown` first), in which case nothing was dropped and the cost is ordering, not loss.

Asserting (1) from the `HubDisposedBeforeResponseException` alone would be the inference this fleet keeps catching itself in.

## What would close it — one reading, and the instruments are already in `main`

On the next occurrence, read the **request's** trail (printed by `[STALE-CALLBACK]`, `[QUIESCE-TIMEOUT]` and the disposal snapshot, all in #4154):

- `REPLY_REFUSED_SHUTTING_DOWN runLevel=… parent=…@DisposeHostedHubs` on the trail ⇒ candidate **(1)**: the responder's seam refused it. The fix is then a carrier for typed replies that cannot reorder — the only safe shape is one that hands the reply to the requester's own INTAKE (its gate and pump decide, as for a routed delivery), and the open question is whether that is safe for an ack whose state change is still in the responder's own queue. `Doc/Architecture/RefusedRepliesDuringTeardown` states why the post-site bypass is not.
- `↩ reply#1: … → RECEIVED@<requester>` with the callback already gone ⇒ candidate **(2)**: the transport did its job and the requester's teardown outran it. The fix is then in the requester's quiesce ordering (`AReplyIsOwedByAShuttingDownLocalHub` already re-arms for a shutting-down sibling — the question is why it did not here), not in the carrier.

A repro exists and is cheap: `LeavingHubAdoptionSweepTest` prints this every run (it is PASSING — read the 2.0 s and the `CompileStateMirror` warning, not the verdict).

## Related

- #4072 / #4154 — the NACK half, fixed; this is the typed half it deliberately left.
- #3303 — the original "teardown drops the ANSWER" class; `IUndeliverableReplySink` covers a reply with an ARMED late watch (patch verdicts). A `CreateOrUpdateNodeResponse` has no such watch, which is why it falls through.
- #2543 — a different mechanism (a NodeType hub not taking turns); listed so the two are not merged by resemblance.

rbuergi:
Fixes #4170.

## The defect, from the trail CI printed

A reply's own post enters **its own hub's intake first**. `RefusesIntake` tier 1 (`Quiescing`) exempts a delivery carrying `PostOptions.RequestId` — *"an ANSWER, never new work … precisely what the quiesce drain is waiting for"* — and tier 2 (`DisposeHostedHubs`) refused everything. The two checks can see **different run levels for one delivery**:

```text
  → PATCH_ACK ok@TestData/dnt48714d51(+1ms)
  → RESPONSE_POSTED type=PatchDataResponse target=cache/kRcuyK2j…@TestData/dnt48714d51(+1ms)
↩ reply#1: RECEIVED runLevel=Quiescing@TestData/dnt48714d51(+1ms)
  → DROPPED_SHUTTING_DOWN runLevel=DisposeHostedHubs@TestData/dnt48714d51(+1ms)
  → NACK_DECLINED reason=not-awaited@TestData/dnt48714d51(+1ms)
  → FAILURE_REPORTED errorType=ShuttingDown@TestData/dnt48714d51(+1ms)
  → RESPONSE_POSTED type=DeliveryFailure target=TestData/dnt48714d51@TestData/dnt48714d51(+1ms)
  → RESPONSE_ARRIVED_NO_SUBJECT type=DeliveryFailure@TestData/dnt48714d51(+17ms)
```

The owner had merged (`v=3`) and acked; the reply died in its own intake; the abandonment path answered `delivery.Sender`, **which for a reply is the responder** — a `DeliveryFailure` posted to itself and forwarded back to nobody — while `cache/…` burned its whole 31 s `WriteVerdictBound` for a write that had committed. It reddened core's merge queue twice on 2026-09-13 (#4110's group at 07:21Z, #4162's at 10:43Z) via `DanglingNodeTypeUpdateTest`, whose retype recycles the very hub that owes the verdict — production shape for any importer that retypes and writes again.

## The fix — both halves, because neither alone is enough

1. **A reply is an answer at every tier.** From `DisposeHostedHubs` until `ShutDown`, a correlated delivery is ADMITTED and routed: it registers no callback, owes no work and is addressed elsewhere, so tier 2's "no new work" never applied to it. One turn, then a route to a live parent.
2. **The exemption stops at `ShutDown`** — the bound `ShutdownRequest` already carries, for the same reason: past it `messageService.Dispose()` has drained the queues, so an admitted delivery would never be dequeued by anyone. That would trade a bounded wait for a permanent leak (#3647's shape in a new costume).
3. **Past that bound the refusal is reported to the REQUESTER** — the party the reply was FOR, named by the delivery itself (`Target` + `RequestId`) — through a live parent, else through #4154's in-process carrier. Never to `delivery.Sender`. The trail says `REPLY_REFUSED_REQUESTER_NACKED` / `REPLY_REFUSED_UNREPORTABLE`.
4. 🚨 **`ShuttingDown`, never `Failed`.** The work the lost reply reported may have committed — on the measured trail it had — so "failed" is a lie that makes a caller re-apply a write that landed (#3112's mistake). `ShuttingDown` is the truth and is what `UpdateRemote` re-enqueues on, so a 31 s silence becomes an immediate retry.

**The typed reply itself is never handed to a bypass.** What travels past the transport is a *classified NACK about it*, which acknowledges no state and therefore cannot overtake the change it reports — so the reorder question #4170 carried did not have to be answered to close this seam.

## Control, both directions

`ARefusedReplyReachesItsRequesterTest` — a responder held at `DisposeHostedHubs` by a parked child (no race: the phase is read back before the act):

| | with the fix | **production change reverted** |
|---|---|---|
| `AtDisposeHostedHubs_ACorrelatedReply_IsAdmittedAndReachesItsRequester` | **50 ms**, the requester receives the typed `SlowAck` | **FAILS** — `TimeoutException` after 38 s: the requester's silence |
| `PastShutDown_ARefusedReply_NacksTheRequesterTransiently_NotTheResponder` | **171 ms**, `DeliveryFailure{ShuttingDown}` answered by the owner, `REPLY_REFUSED_REQUESTER_NACKED` on the trail | **FAILS** — `Assert.Throws … Actual: TimeoutException` after 38 s |

I watched both fail on the reverted tree before restoring it.

Also green: `MeshWeaver.Messaging.Hub.Test` **416/416**, `MeshWeaver.Data.Test` **508/508**, the Graph teardown slice **25/25** (including `DanglingNodeTypeUpdateTest`, the test that reddened the queue), `DocumentationLinkIntegrityTest`.

Doc: `Doc/Architecture/RefusedRepliesDuringTeardown` — it listed two seams and this is a third; the page now carries it, the two bounds and why the NACK-not-the-reply choice sidesteps the reorder hazard. What's New: Fix.

Pairs-with: none — no public type or member leaves `src/`; the change is inside `MessageService`.


rbuergi:
Addressed in `3bad2b744` — the comment is right, and by the shape of my own branch: it keys on `PostOptions.RequestId` + a `Target` that is not this hub, and **every** NACK posted with `ResponseFor` satisfies both, so a refused `DeliveryFailure` would have been answered with a `DeliveryFailure` addressed at the previous requester — and past `ShutDown` several hubs are in that state at once, which is the cascade.

**`MayAnswer()` is the established answer-once test on this path**, so no other predicate was needed: it reads the contract off the ENVELOPE (#1485) rather than the CLR type — which matters here, because a reply that crossed a hub boundary arrives as `RawJson` — and it is false for a `DeliveryFailure` and for `[CanBeIgnored]` lifecycle traffic. It is what `ReportFailure` and the routing paths already use. A delivery it excludes falls through to the sender-addressed paths below, which suppress it for the same reason, so nothing changes for those.

Note the guard is on the **reporting** branch only, deliberately: a correlated `DeliveryFailure` is still ADMITTED by the tier-2 exemption while the hub can route it (below `ShutDown`). Admitting a NACK is not a cascade — it is the answer somebody is waiting for, and routing it onward is the whole point; minting a *second* one is what had to stop.

**The case is a control, not a decoration.** `ARefusedDeliveryFailure_IsNotAnsweredWithAnotherOne`: a correlated `DeliveryFailure` refused past `ShutDown` must leave the requester's callback as it was and stamp no `REPLY_REFUSED_REQUESTER_NACKED`. Falsified by removing the guard — it then **fails**, with the trail showing `RESPONSE_POSTED type=DeliveryFailure target=refused-reply-requester/1@mesh/1 → REPLY_REFUSED_REQUESTER_NACKED runLevel=Dead` and the second NACK delivered all the way to the requester. Restored ⇒ green.

`MeshWeaver.Messaging.Hub.Test` **417/417** locally with the guard in place.

On the localization nit: the message is a `ShutdownNack` banner — a wire/log string the transient classifiers (`MeshNodeStreamCache.IsTransientOwnerFailure`, `AreaErrorClassifier.IsTransientHubFailure`, `ShutdownNack.IsAnsweredByOwner`) match on, and no viewer surface. Like every other `ShutdownNack` sentence it stays English by the same rule that keeps wire identifiers untranslated.

PR #4210, opened 2026-09-13
fix(#3461): the two Azure-direct publication readers resolve `_current` (phase 3)

rbuergi:
`Refs #3461` — phase 3's **reader half**, which is the precondition the design doc puts on flipping any prefix. The flip itself (phase 4) is a scope call, measured and stated below; phase 5 is a deployment fact.

## Mechanism, at file:line on the merge base

Phase 1 routed every reader the **portal image** carries through `ShippedPrebuiltBundles.PublicationDirectoryOf`. Two readers are not in that image, so they kept composing paths under the bare `prebuilt-bundles/<identity>/<source>/` prefix:

| reader | sites |
|---|---|
| `.github/scripts/compose-sealed-modules.sh` (the `--storage-target` OIDC fallback) | `:145` (`_complete` + `modules/_index`), `:183` (each module bundle) |
| `.github/workflows/node-repo-gate.yml`, step `id: seed` | `:1445` (`$dir`), `:1448` (`_complete`), `:1460` (`download-batch`) |

`SealedPublicationGenerations.md` names both by name: *"Route them through the same resolution before any prefix flips… at phase 5, when the flat copy is dropped, they break outright."*

The gate's Azure branch is **reachable, credentialed and preflighted**, not vestigial: `node-repo-gate.yml:876` does a real `azure/login@v3` `if: inputs.upstream-seed != '' && inputs.registry-url == ''`, and the preflight at `:859` asserts the Azure secrets and the `az` binary for exactly that path.

🚨 **And #4172 (merged this morning) made this load-bearing rather than tidy.** A downstream publication now seals only its OWN modules — `node-repo-publish-bake.yml:1267`: *"a consumer composes each package from the FIRST upstream whose seal lists it (compose-sealed-modules.sh)"*. There is no downstream copy of an upstream's modules left to fall back on, so this reader is now the single path to those bytes.

## What changes

**Both readers resolve `_current`**, by the rules the reader contract states (absent, blank, unreadable, not a single path segment, or naming a directory that is not there ⇒ the source directory). Copies 4 and 5 of `resolve_publication_dir`; the house precedent for these two artefacts is explicit duplication with a *change one, change all* note (`node-repo-gate.yml:1313` says so for `registry_get`, for the same pinning reason — the workflow is pinned by the caller's `uses:` while the scripts are fetched at `platform-ref`, so a shared file would be a third independently-pinned artefact).

**`compose-sealed-modules.sh` now records the generation its module *index* came from** and pins each module download to it. That is the storage-path equivalent of the registry path's `If-Match`/412, threaded through the `$genfile` argument that already existed — and a property that lane never had: a pointer moving between the listing and a download can no longer mix two publications' module bytes, because a generation directory is not rewritten in place.

🚨 **The gate's `seed` is the one reader that may NOT take the reader's fall-back, and it is because of the BATCH.** `download-batch` **recurses** and the CLI **flattens**. For a point read, "fall back to the source directory" yields the previous publication *whole*; for a recursive batch over a prefix holding generations it yields a **union** — and same-named files silently overwrite each other, so the mix is invisible to the step's own `n=$(ls "$SEED_DIR"/*.zip | wc -l)` count as well as to the compiler. So a pointer that **exists** and could not be followed is a **refusal** there, not a fall-back. It is transient by construction (the pointer is one small file write) and a re-run reads the publication that now applies. The divergence from the reader contract is stated in place, with its reason.

## The control, watched fail on main

🚨 **Neither Azure path had ever been executed by anything.** `test-sealed-module-compose.py` runs the script with `--registry-url` only, and its gate case runs `seed` with `TARGETS=""` — so the entire `download-batch` block (`node-repo-gate.yml:1437-1470`) was dead to every harness in the repository.

`test-publication-pointer-readers.py` executes both — the real script, and the gate's step **extracted by `id:`** (refusing red if the step grew an expression or stopped issuing the `download-batch`, never skipping) — against a stub `az` over a directory tree. The fixture is **discriminating**, the same trick `bake-scope.sh --self-test` uses: the flat copy and the generation hold the same file names with different bytes, plus one bundle the flat copy carries and the generation does not, so the union is countable as well as readable.

Against the pre-change readers (`git checkout origin/main -- <the two files>`):

```
a prefix whose _current names a generation:
  FAIL compose-sealed-modules.sh --storage-target composes the GENERATION's module bytes:
       composed 'FLAT-COPY-BYTES', wanted 'GENERATION-BYTES'
  FAIL node-repo-gate's seed download-batch seeds the GENERATION's bundle bytes:
       seeded 'FLAT-COPY-BYTES', wanted 'GENERATION-BYTES'
  FAIL …and it did NOT also flatten the flat copy's bundles into the same seed directory:
       the seed holds ['Legacy.zip', 'Store.zip'], wanted exactly ['Store.zip']
the same prefix with NO pointer (today's layout, and the one that must not change):
  OK   compose-sealed-modules.sh reads the prefix itself when there is no pointer
  OK   the gate's seed reads the prefix itself when there is no pointer
a pointer naming a generation that is not on the share:
  OK   compose-sealed-modules.sh falls back to the prefix, exactly as the reader contract says
  FAIL …and it names the dangling pointer instead of falling back in silence
  FAIL the gate's seed REFUSES rather than seeding a possible mix
::error::5 case(s) failed
```

**5 of 8 fail; the 3 that pass are the no-pointer controls** — which is what makes this a measurement rather than a harness that fails everything, and which pins today's flat behaviour as byte-identical. With the change: `8 case(s) over 2 readers × 3 pointer states — all green.`

## Also: a proof the doc cites that no lane ran

`bake-scope.sh --self-test` is one of the two phase-2 reader proofs `SealedPublicationGenerations.md` leans on (*"four pointer-resolution assertions"*). Measured 2026-09-13, a repo-wide grep finds it only in its own usage comment and in that doc — **no workflow executed it**, while its sibling `carry-forward-bundles.sh --self-test` is wired. A proof a doc cites and no lane executes is a claim, not a control. Both it and the new harness are wired into `dotnet-test.yml` beside the others.

## Why phase 4 (the flip) is NOT in this PR — the measurement

Over 2026-09-12T08:00Z → 09-13T08:00Z, every `plugins` publish job of both lanes: core `plugins-bake` **25** executed (23 sealed), Plugins `publish-bake` **10** (8 sealed); **62 of 62** target verifications read *"N/N file(s) hold this run's bytes … 0 byte-identical from another"*; **0** MIX refusals; 20 distinct identities, **6 written by both lanes** — all 6 at different Plugins commits, hours apart, with no overlap in the window. The defect is **latent, not active**, and the #3496 postcondition is carrying it.

The flip is still worth doing, but it is not a core-only change and it is not risk-free:

- `plugins` must flip in **one change set across BOTH producers**, and one of them (`MeshWeaver.Plugins`' own `publish-bake`) lives in another repository. A new writer and an old writer on one prefix is the half-migration the selector exists to avoid — the new one moves the pointer, the old one replaces the flat copy in place and never touches it, so a pointer-following reader keeps serving its generation and never sees the old writer's *newer* publication. A stale serve, silent, with nothing red.
- Measured 2026-09-13, **no caller anywhere passes `publication-layout`**: core's `main-cd.yml` `plugins-bake` (`with:` block at `:2662-2689`) does not, so the prefix's core-side producer is `flat` too. The input's default is `flat` at `node-repo-publish-bake.yml:335`.
- Flipping **before** this PR would have been actively wrong: the gate's recursive batch would have seeded a union the moment generations appeared under a prefix.

So what phase 3 removes is the reason a flip could not be attempted at all. **The flip is a scope call for whoever owns the cross-repo change set.**

## Verification

```
bash -n .github/scripts/compose-sealed-modules.sh                      → clean
bash -n <the `seed` step body, extracted>                              → clean
python3 .github/scripts/test-publication-pointer-readers.py            → 8/8 (5/8 on main)
python3 .github/scripts/test-sealed-module-compose.py                  → 6/6 (unchanged)
bash .github/scripts/bake-scope.sh --self-test                         → 12 + 2 + 4 + 4, all green
check-workflow-shell / -yaml-keys / -timeouts / check-pr-secret-preflight → 0 violation(s)
dotnet test test/MeshWeaver.Documentation.Test -c Release --no-build   → 447/447
```

CI-lane change only; nothing user-facing, so no What's New entry. No public type or member leaves `src/`; no interface gains a member; no i18n catalog key changes.


rbuergi:
All three review findings are addressed in #4211 — this PR auto-merged on green before they could land here.

1. **Orphan generation** (`node-repo-gate.yml`): you were right that checking `_current` is not sufficient. A generation publish seals its directory and *then* moves the pointer, so an interrupted pointer upload leaves a generation with no pointer at all. The refusal is now a property of the **bytes** — whenever the resolution lands on the prefix, refuse if the prefix holds any subdirectory other than `modules` — and an unreadable listing refuses too.

2. **Seed ↔ composition consistency** (`compose-sealed-modules.sh`): the seed now exports the generation it read and the composition takes it as `--generation`, which replaces the resolution. It is passed only when the script fetched at this lane's ref understands it (the same vintage probe `node-repo-publish-bake.yml` uses for `SEAL_MODULES_OWN`), because the workflow is pinned by the caller's `uses:` while the script is fetched at `scripts-ref` — a new workflow meeting an old script would exit 2 on every satellite.

3. **The no-pointer control**: correct, and it passed by coincidence — `find` order decided which `Store.zip` won. `stage()` now takes `generation` as its own knob, so the flat control is the real flat layout and the orphan is its own case. 13 cases over 5 share states, 7 of 13 failing on the pre-#4210 readers.

#4211 also closes a fourth hole the review prompted me to look for: `SealedPublicationIndex.ReadFor` is a reader in the portal image that phase 1 missed. It read the markers and the sentinel from the **source** directory while the same sweep read the **bundles** through `PublicationDirectoryOf` — so once the flat copy is dropped it would report every source unsealed, `SealedSyncGate` would see an empty `mine`, and every repository would advance ungated.

Issue #3845, opened 2026-09-09
Adopt-then-sync, hole 4: the seal marker cannot name a NodeType, so a fully satisfied gate still leaves types unadopted


rbuergi:
Follow-up to #3583, split out of https://github.com/Systemorph/MeshWeaver/pull/3844.

That PR makes requirement (1) airtight — a NodeType never errors because of delivery; the last build keeps serving, keyed on module-version compatibility — and delivers (3), the readiness notification. Requirement (2) — **sources must not run ahead of bytes on a live portal** — is not addressed there and remains open:

- `SealedSyncGate` (#3600, fixed for Plugins attribution in #3612) is the intended mechanism, but per the #3583 comments it can only arm on a seal a post-#3612 lane wrote for an identity a portal actually runs; already-drifted sources are not repaired, and `ModuleDiscoveryService.FirstImport` still imports at branch tip.
- The instance-side half is MeshWeaver.Plugins#1430 (import at the publication's `sourceCommit` sealed under the instance's framework identity, never at latest green).

Acceptance: on a live portal, after a Plugins merge that touches a NodeType's source, the affected type's `CurrentSourceFingerprint` does not move until a bundle for the portal's framework identity whose fingerprint matches exists — i.e. the type goes straight from `AdoptedVerified` to `AdoptedVerified`, never through `StaleAdopted`. Until then `StaleAdopted` is the honest, serving, announced intermediate state.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01PcriF2ByxHM2Dnn2eA7k3Y

rbuergi:
## Reopened — the 2026-09-10 sweep was not a triage decision

This issue was closed `not_planned` at **2026-09-10T14:30:27Z** by a bulk sweep that PATCHed **every open issue in all 13 `Systemorph` repositories** — 311 issues in twelve minutes, no comment on any of them — after a session was asked to *"close all issues in all repos"*. Nothing on this thread was read, and no judgement was made about this issue.

All 82 issues the sweep closed in the MeshWeaver fleet were reviewed afterwards. This one is **unresolved**: Requirement 2 of #3583; no PR exists.

Sixteen of the 82 were left closed — their fix had actually landed, or the author had withdrawn them; those should have read `completed`, not `not planned`.

rbuergi:
## Scoped against today's `main` (`d1d0705e7`) — **stays OPEN**, and it now owns requirement 2 outright

[...]

Issue #4063, opened 2026-09-12
A publication seal that stops advancing freezes every GitSynced Space of that repository, silently and unboundedly


rbuergi:
## The defect

**A green main build publishes, and GitSynced Spaces on the live meshes do not converge to it.** On
both production portals nothing has converged automatically since `24c2d024` (2026-09-11 ~18:55Z),
although MeshWeaver.Plugins `main` has gone green eight times since and has tagged `Hosting/v1.17.1`,
`Hosting/v1.18.0` and `DeepSign/v1.1.0` in that window.

## Verified state (read-only, 2026-09-12 04:0x–06:0xZ)

| | memex.meshweaver.cloud | memex.systemorph.com |
|---|---|---|
| `Hosting/_GitSync` | `24c2d024`, `Imported`, attempted == synced, `lastAttemptWasFinal: true`, synced 18:55:16Z | same, synced 18:53:55Z |
| `/api/version` | core `74d4c852` | core `74d4c852` |
| `/health` bake report | `framework=s6649734…` | `framework=s6649734…` |
| `Admin/_Build/Systemorph.MeshWeaver.Plugins` | rewritten **03:44:49.895Z** (run 34668111508, `4b97be19`, success) | rewritten **03:44:49.896Z**, same run |

`Hosting/PlatformBuilds/plugins`, written by the publishing lane at 03:44:43Z:
`identity: s8e1f76cc22bcd7051a7ad998debbfeb3`, `version: 3.0.0-ci.8392`, `sha: 4b97be19`.

`DeepSign/_GitSync` had never moved off its provisioning commit (`7660ca73` / `78e6d92e`) until a
peer ran `git_hub_sync update` by hand at 03:47Z; before that, attempted == synced, i.e. **no attempt
at all, not a failed one**. Every space sourced from a satellite (Reinsurance, SocialMedia,
Manufacturing, Crm) is legitimately AT its repo's current head, so the stall is specific to the one
repository whose `main` is moving.

## Mechanism

`GitHubWebhookProcessor.MatchingBuildTargets` (`src/MeshWeaver.GitSync/GitHubWebhookProcessor.cs`)
calls `SealedSyncGate.Decide` per candidate. The gate proceeds only when some publication **sealed
under this instance's own framework identity** has `source-commit.txt == headSha`
(`SealedPublicationIndex.ReadFor(root, PrebuiltAssemblySeeder.LiveFrameworkMvid)`).

[...]

Issue #3583, opened 2026-09-07
Bundles are baked for ONE framework identity per run, so a portal's GitSynced sources stay days behind main (or its types go unadopted) — options 1-4 undecided


rbuergi:
> ## 🚦 Status, 2026-09-14 — re-read against `main` `85be5d7937`. Everything below this block is the ORIGINAL 2026-09-07 filing; three parts of it are now stale and are corrected here rather than edited out, because the measurement is still the evidence.
>
> **What remains is ONE maintainer decision and nothing else: the identity-churn call, options 1–4, costed on [`Doc/Architecture/FrameworkIdentityChurn`](https://github.com/Systemorph/MeshWeaver/blob/main/src/MeshWeaver.Documentation/Data/Architecture/FrameworkIdentityChurn.md) → *The options, costed*.** It is the same call #4063's item 2 waits on, seen from the consumer end; answering it once answers both.
>
> | the filing says | today |
> |---|---|
> | requirement 1 — a NodeType must not error because of delivery | **closed** (#3844) |
> | requirement 2 — sources must not run ahead of bytes | tracked on **#3845**; hole 2 closed (#4212), holes 1/3/4 are three *different* decisions, priced in [this comment](https://github.com/Systemorph/MeshWeaver/issues/3845#issuecomment-5656511871) |
> | requirement 3 — announce readiness | built in the narrow per-NodeType shape; the package-level "ready for identity I on this replica" state is unbuilt, and is the scope question on `FrameworkIdentityChurn` → *A related scope question* |
> | "4b" (resolve the newest platform-SEALED set on `schedule`) | **was never open** — implemented since ≥ 2026-09-10 in every satellite's `ci.yml`, and it does not remove the red it was expected to. Falsified with a control run in #4248 |
> | the stronger reading of 4b (*resolve the newest set every upstream has ALSO published*) | **refused**, in those words, by `Doc/Architecture/CiContentBake`: *"an upstream that stops publishing would leave every dependent's poll green forever"* — and contrary to the maintainer's *"for compile always find latest package of platform and plugins"* |
>
[...]

PR #4212, opened 2026-09-13
fix(#3845): the AutoSync first import pins the sealed commit, not the branch tip

rbuergi:
## The last unattended branch-tip import, closed by a pin the residue's own rationale overlooked

`ModuleDiscoveryService.FirstImport` called `UpdateToLatestFromGitHub` — which resolves the branch
**at fetch time** — unattended, as System, on boot and on every catalog scan. That is exactly what
`UpdateToProvenCommitFromGitHub`'s own XML doc forbids a machine trigger:

> Trigger `UpdateToLatestFromGitHub` only from a human action.

It is **hole 2 of #3845's five** (`ModuleDiscoveryService.cs:694`, called from `:476` and `:644`) and
the last unattended branch-tip import in the tree.

### Why this is implementable and not a decision

`SyncRefContract` named it a deliberate residue, on the grounds that *a consumer instance that never
receives GitHub webhooks has no `BuildCompletion` record to pin to*. That is true, and beside the
point: such a consumer **does** hold a commit a build proved, on its own disk and with no webhook
involved — `source-commit.txt` beside the bundles **sealed for its own framework identity**. That
commit names a tree CI proved *and* the tree the bytes this instance actually runs were baked from.
Strictly better evidence than a branch tip.

So this is not a new policy. It is `SealedSyncGate`'s already-sanctioned envelope applied to the one
path that skipped it, using the established unattended-import API.

### What it does

`SealedSyncGate.DecideFirstImport` — pure, three answers:

| the seal says | the first import does |
|---|---|
| nothing of this repository is sealed for this identity | resolve the branch — **the residue's case, unchanged** |
| one sealed commit, attributed by the `repository.txt` marker | `UpdateToProvenCommitFromGitHub` at it |
| attributable but torn / unknown commit / two seals disagreeing | import nothing, log the reason at Warning, retry next scan |

The original rationale stays intact where it applied: a repository this instance runs no publication
of is not attributable, so it still provisions at the tip, and AutoSync on a webhook-less consumer is
as alive as it ever was. The blast radius of a hold stays "a new empty partition", never live content
going dark.

🚨 **A hold is NOT self-releasing without qualification — review caught that claim and it is now
stated exactly.** Both release paths are boot-driven: `ModuleDiscoveryService` enqueues scans at boot
and on `BuildCompletion` emissions, and `SealedPublicationSyncReconciler` runs from
`ShippedPrebuiltBundles.SeedPublishedRoot`. So on an instance receiving no build webhooks *both
reduce to the next process start*, and a seal completing mid-process is not noticed until then. That
is #4063's shape and this hold inherits it. The Warning the scan logs names the event to wait for
rather than saying "later".

🚨 **Every attributable publication must be usable, not merely one of them** (review finding, fixed in
`dbd60ffe22`). Selecting the sealed entries first let a good seal override a torn sibling of the same
repository — `plugins` sealed at C beside a `plugins-extra` with no completion sentinel pinned the
Space at C while part of that repository's bytes were missing here, contradicting this method's own
contract. The refusal is now evaluated over every attributable publication before any selection
narrows it, and the hold names a witness plus how many others share its state. This is deliberately
stricter than `Decide`: that admits a proved tree onto a source already carrying content, while this
decides what a Space holding *nothing* is first populated with — no partial state to preserve, and no
second chance to be careful later.

🚨 A first import can be attributed **only by the repository marker** — it has neither a built commit
nor a last-sync commit, so both commit-attribution legs of `BelongsTo` have nothing to compare
against. A seal predating that marker reads as unattributable and keeps today's behaviour.
Unattributable is never a guess, in either direction.

### The test could fail

Two independent mutation experiments, each reding exactly the right tests:

- Mutating `DecideFirstImport` back to *"always resolve the branch"* reds **6 of the 9** original new
  tests; the 3 that stay green assert the unattributable → branch-tip path, which the mutation
  preserves — the right 3.
- Mutating it back to the **pre-review** logic (decide on the sealed subset) reds **exactly the 3**
  tests added for that finding, and nothing else.

Restored: **28/28 green** in `MeshWeaver.Documentation.Test`, `MeshWeaver.slnx -c Release
-warnaserror` = **0 Warning(s), 0 Error(s)**, `DocumentationLinkIntegrityTest` green.

### What this does NOT close

#3845 stays open on the other four holes, stated as what they are:

| hole | why it is not in this PR |
|---|---|
| 1 — the gate fails open when nothing is sealed (`SealedSyncGate.cs:56-57`) | pinned as intended behaviour by `SealedSyncGateTest.NothingSealedForThisIdentity_IsNotThisGatesBusiness`; changing it holds every repo with no seal, course repos included — a **decision** |
| 3 — the two GUI paths (`GitHubActionArea.cs:86`, `GitHubSyncSettingsTab.cs:407+`) | `SyncRefContract` states the opposite as contract: *"A human clicking Update is asking for exactly that and gets it."* Holding a human's explicit action contradicts a documented, deliberate rule — a **decision** |
| 4 — the gate keys on a seal MARKER, not on a bundle for this NodeType | true adopt-then-sync; a design change, not a defect fix |
| 5 — starvation when the gate does engage | #4063 |

`Refs #3845`, `Refs #3583` — deliberately not `Fixes`: one hole of five.

### Scope checks

No public type or member removed; nothing added to a public interface; no i18n catalog change (the
pinned import reuses the already-localized `activity.gitsync.updateToProvenCommit.*` keys, so a
first import that lands on a sealed commit is titled in the viewer's language today).


rbuergi:
## Review addressed — one real bug fixed, two claims tightened, one declined with the measurement

**1. Torn sibling could be overridden by a good seal — REAL, fixed (`dbd60ffe22`).** `DecideFirstImport` selected the sealed publications first and decided on those, so `plugins` sealed at C beside a `plugins-extra` with no completion sentinel pinned the new Space at C while part of that repository's bytes were missing here. That contradicted this method's own documented contract (*"attributable but torn … import nothing"*), and the test only covered the case where the torn seal was the **only** one — so nothing caught it. The refusal is now evaluated over every attributable publication **before** any selection narrows it, and the hold names a witness plus how many others share its state.

Three tests added, and the pre-review logic reds **exactly those three**: a torn sibling beside a good seal, an unknown-commit sibling beside a good seal, and the `and N more` count. (Stated as stricter than `Decide` on purpose — `Decide` admits a proved tree onto a source that already carries content; this decides what a Space holding *nothing* is first populated with, where there is no partial state to preserve and no second chance to be careful later.)

**2. "The hold is self-releasing" — too strong, corrected.** Both release paths are boot-driven: `ModuleDiscoveryService` enqueues scans at boot and on `BuildCompletion` emissions, and `SealedPublicationSyncReconciler` runs from `ShippedPrebuiltBundles.SeedPublishedRoot`. So on an instance receiving no build webhooks **both reduce to the next process start**, and a seal completing mid-process is not noticed until then. That is #4063's shape and this hold inherits it. Said plainly in both XML docs, in `SyncRefContract`, and in the Warning the scan logs — which now names the event to wait for instead of "later".

**3. `SealedPublicationIndex.ReadFor` inline — declined, with the measurement recorded at the call.** The suggestion was to use the `IIoPool` boundary *"as the other bundle readers do"`. Measured: **there is no such reader.** All four call sites of that pure function read it inline — `GitHubWebhookProcessor` inside its own `Select`, `ShippedPrebuiltBundles` twice during seeding, and this one — so wrapping this one alone would make four call sites disagree about a shared function. It also does not reach the hub: the scan pipeline is `ObserveOn(TaskPoolScheduler.Default)` + `Concat` **precisely so a scan never runs on the thread that queued it**, so the exposure is one pool thread and the next queued scan, on a path that already does a git fetch and a full import. Moving all four behind the pool is worthwhile and is a change to `SealedPublicationIndex`'s contract, not to this call — the reasoning is now in the code so the next reader does not re-derive it.

`MeshWeaver.slnx -c Release -warnaserror` = 0 Warning(s), 0 Error(s); 28/28 in `MeshWeaver.Documentation.Test`.

PR #4215, opened 2026-09-13
ci(#3732): read one-build-per-assembly-name over the PUBLICATION's module set

rbuergi:
## The guard's denominator is 4; the defect's population was ~40

`node-repo-{gate,publish-bake}.yml` already refuse *"one assembly name, one BUILD"*. They assert it
over the set a bake **composes**, and that set is small — its own verdict line says so:

| lane | printed |
|---|---|
| `MeshWeaver.Reinsurance` gate, 2026-09-11 05:54Z | `4 MeshWeaver.* assembly file(s) across 4 bundle(s), 4 distinct name(s), 0 shared, 0 diverged` |
| `MeshWeaver.Plugins` publish-bake, 2026-09-11 05:30Z | `5 … across 5 bundle(s), 5 distinct name(s), 0 shared, 0 diverged` |

**The defect #3732 was filed on did not live there.** It was measured on a memex pod —
`MeshWeaver.Markdown.Collaboration` in **15 copies and THREE builds** — spread across the ~40
*module bundles* a publication carries, the population `node-repo-module-pack.yml` produces and most
of which no single bake composes. So every green reading of the existing assertion has been green
**about a different population**. A denominator of 4 against a population of 40 is not a weak
measurement; it is a measurement of something else, and #3732 has said so twice in the instrument's
own words.

### Where the whole population is in one hand

Not in a pack leg — each holds ONE bundle. The one job that sees the whole wave is `verify`, which
already collects a receipt per module and says so in its own comment (*"IT IS ALSO THE ONLY PLACE
THAT SEES THE WHOLE WAVE (#3310)"*). So the reading is split to match:

- **`module-set-census.py census`**, per pack leg — every `MeshWeaver.*` assembly the bundle carries,
  digested **out of the archive**, with its role (`declared` per the manifest's
  `module.assemblyName`, else `riding`). Never from an unpacked folder: two bundles declaring one
  entry assembly unpack over each other, which is the accident of glob order that hid this. It rides
  the receipt.
- **`module-set-census.py verdict`**, in `verify` — the same arithmetic over the receipts,
  denominator printed on every run.

🚨 **It reads every receipt in the RUN and passes no `--lane`** — the opposite narrowing from the
build accounting beside it, on purpose (review finding, fixed in `da79740f30`). MeshWeaver.Plugins
invokes this lane **twice in one run** (`modules-floor`, `modules-rest`) with different lane stamps,
and the 15-copies/3-builds measurement spanned exactly those calls; a lane filter would make each
verifier drop the other's receipts and print a confident zero — this PR's own denominator error, one
level up. The verdict therefore **names the lanes it folded** on every run, so "publication-wide" is
checkable rather than assumed, and a test pins that a divergence *across* two lanes is found with both
lanes named.

🚨 **The reader step carries no `continue-on-error`** (same review). The script exits 0 when it
*finds* divergence — that is the report posture — so the only way the step can fail is that the
reading did not happen: a malformed receipt, a helper the build-logic ref does not carry, a changed
CLI. Tolerating that would leave a required job green while promising a reading nobody took, which is
the skip-trapdoor rule verbatim. An explicit helper-present assertion names why it is missing.

### 🚨 It REPORTS; it does not yet refuse — and the reason is evidence, not timidity

`verdict` exits 0 on divergence unless `--enforce`, and no lane passes it.

Nobody has ever read this population, so **what a refusal would refuse is unknown**. A `reuse` leg
republishes bytes packed in an earlier run *by construction*, so a shared sibling riding both a
reused and a freshly built bundle would diverge — possibly on every wave, in every satellite, the
moment this landed. `verify` is a required context in five satellites, and turning an unread
measurement into a refusal there is how a new guard takes CD down fleet-wide.

The population is not unguarded meanwhile — it is guarded **late**, at the consumer, where
`PublishedBundleCatalogue` answers `SealedSetInconsistent` and holds the roll for the whole fleet.
Moving that refusal earlier is the point; doing it before one wave has been read is guessing. The
promotion condition is written into the code, the doc and the test:

> Read the `publication module set: …` line on the first waves. Zero divergence ⇒ add `--enforce`.
> Non-zero ⇒ the reading names which assemblies and which bundles, and THAT is the defect to remove.

### The denominator distinguishes its two zero-shapes

A reading over nothing must never be spelled like a clean reading over something. So: an absent
receipts directory prints an explicit zero and says nothing was measured; and a receipt carrying **no
census** (a producer at an older scripts ref) is counted apart and **named** — `not measured: N
bundle(s)` — never folded into the clean count. "Absent" reading as "agrees" is this kind of gate's
own failure mode.

### The test could fail, and it was watched failing

`test-module-set-census.py`, wired into `dotnet-test.yml` beside the composed-set harness, runs both
halves over **real zip fixtures** (real archives, real sha256):

- **both directions** — divergent copies named with both builds and both carriers; identical riding
  copies clean, which is the whole `ModuleOwnedSiblingsRide` decision and what a
  flag-every-duplicate checker would fail;
- **the denominator**, including both zero-shapes;
- **`--enforce` in both positions**, so the report→refusal promotion cannot happen by halves;
- **a falsification arm that mutates the real subject** — with the digest comparison removed, the
  divergent fixture goes quiet (asserted);
- **the wiring read off the real workflow file** — `MODULE_SET_CENSUS`, the `census` call, the
  preflight `--self-test`, and `verdict`'s `--lane`/`--declared`, so a renamed step or a dropped
  argument is red here rather than silently unmeasured in every satellite.

Local: `module-set-census.py --self-test` ok; harness **all checks passed**; the three sibling
harnesses (`test-module-set-consistency`, `test-module-publication`, `test-module-asset-landing`)
still pass; `check-workflow-timeouts` 87 jobs / 0 violations; `check-workflow-yaml-keys` 36 files / 0;
`check-workflow-permission-pairing` 0; `DocumentationLinkIntegrityTest` green.

A build-logic ref that predates the helper fails **one** preflight by name (`[ -f "$MODULE_SET_CENSUS" ]`
plus `--self-test`), rather than failing every pack leg inside a subshell whose recorded reason names
neither the helper nor the ref.

`Refs #3732` — deliberately not `Fixes`. Defect 1 (a downstream composing against its upstream's
sealed bytes rather than carrying copies) is untouched; this is the guard's population only.

### Scope checks

Workflow + scripts + one doc page. No public surface removed or added, no interface member, no i18n
catalog change.


rbuergi:
## Review addressed (`da79740f30`) — both findings were right, and the first would have reproduced the exact blindness this PR removes

**1. The lane filter made the reading publication-CALL-wide.** Correct, and it is the same denominator error as the composed-set guard, one level up. MeshWeaver.Plugins invokes this reusable lane **twice in one run** (`modules-floor`, `modules-rest`) with different lane stamps, and the 15-copies/3-builds measurement this whole reading exists for spanned exactly those calls — so each verifier would have dropped the other's receipts and printed a confident zero.

`verify` now downloads **every** receipt in the run (`module-pack-receipt-*`) and passes **no `--lane` and no `--declared`**. That is deliberately the opposite narrowing from the build accounting beside it, which needs both because attributing a sibling call's evidence to this one was a real defect (Plugins#1077) — the two answer different questions, and the script now says so where the filters are defined rather than leaving a reader to wonder why one reading narrows and the other does not.

And because "publication-wide" is only true if it saw every call, the verdict now **prints the lanes it folded** on every run:

```
publication module set: 2 MeshWeaver.* assembly file(s) across 2 module bundle(s) of 2 read, …
  folded 2 lane(s): modules-floor, modules-rest
```

A test pins that a divergence **across** two lanes is found and both lanes are named — i.e. the case a lane-filtered reading would miss.

**2. `continue-on-error` on the reader.** Also correct, and it is the skip-trapdoor rule verbatim. The script exits 0 when it *finds* divergence (that is the report posture), so the only way the step can fail is that **the reading did not happen** — a malformed receipt, a helper the build-logic ref does not carry, a changed CLI. Tolerating that leaves a required job green while promising a reading nobody took. Removed, with an explicit helper-present assertion that names *why* it is missing (a skipped build-logic checkout downstream of a failed `select`, or a ref predating #3732).

The harness pins the absence by matching the **key** `continue-on-error:`, not the phrase — the phrase appears in the step's own explanatory comment, and the first version of that check reported a pass as a failure.

Local: `module-set-census.py --self-test` ok; harness all checks passed; `test-module-set-consistency`, `check-workflow-timeouts` (87/0), `check-workflow-yaml-keys` (36/0), `check-workflow-shell` (0 live) all green; `DocumentationLinkIntegrityTest` green.

PR #4216, opened 2026-09-13
fix: a NodeType is not judged against a source set that has not landed (#4208)

### LINKED ISSUE #4208: Gate: a NodeType activated before its shared sources landed adopts NOTHING and is PARKED red — 'MISSING SOURCES: 3 of 3 queries matched NO nodes', live fingerprint e3b0c442 (empty)
## Measured (MeshWeaver.Reinsurance#204, run 34766381762 attempt 2, gate shard 1/1, 2026-09-13 16:16:50Z)

`Reinsurance/AggregateSection` declares three source queries — `namespace:Source scope:subtree`, `shared=@Reinsurance/Source`, `shared=@Reinsurance/SampleData/Source` — the same three every Reinsurance NodeType declares. In this run it was the ONLY one of them that went red, and the log says why in order:

```
16:16:50.758 bake-seed  Prebuilt assembly ADOPTED for Reinsurance/AggregateSection at version 0 (framework s5c186d0c…, module version 1.0.21) — no compile needed
16:16:50.760 CompileWatcher [AdoptedSourceStamp] Reinsurance/AggregateSection: the source moved past the adopted build (bundle fingerprint 2c12e765f7b3f3e9, live e3b0c44298fc…)
16:16:50.820 bake-seed  Prebuilt assembly for Reinsurance/AggregateSection DECLINED before writing (#2813): the bundle records source fingerprint 2c12e765f7b3f3e9 but the live sources are e3b0…
16:16:55.310 MeshNodeCompilationService  Failed to compile assembly for node 'Reinsurance/AggregateSection'
        MISSING SOURCES: 3 of 3 declared source queries for 'Reinsurance/AggregateSection' matched NO nodes on this mesh
        CS0246 (line 68): 'Premium' could not be found … CS0246 (line 135): 'ReinsuranceSection' …
16:16:55.383 CompileWatcher  NodeType 'Reinsurance/AggregateSection' PARKED after compile failure — further activations serve the cache
    RED Reinsurance/AggregateSection [compile:Failed render:Skipped tests:Skipped]
```

`e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855` is the SHA-256 of the EMPTY string: at the moment the adoption stamp was checked, the mesh had imported NONE of the 26 source nodes this NodeType reads (`AggregateSection` sorts first in the namespace, ahead of `Reinsurance/Source/*`). The stamp check therefore refused a correct prebuilt, the fallback compile ran against an empty source set — which the #3929 diagnostic names precisely — and the node was PARKED, so when the sources landed milliseconds later nothing re-activated it. `AmountType` and every later sibling, with the identical queries, passed (`ok … [26 source(s), 7 dependency record entr(ies)]`).

The same run's earlier attempt and this repo's 07:24Z run adopted the same bundle green (`ADOPTED … framework s8cea1c45…`, 26 sources live). Crm#97, Manufacturing#86 and SocialMedia#185 — the same fleet-process change, same afternoon — passed. It is an ordering race in the gate's import → activate sequence, not content.

## What a fix looks like

Either of these closes it, both are better:

1. **Activation waits for the declared source queries to be non-empty** before the adoption stamp is computed and before a compile is attempted — a NodeType whose `shared=@…` query matches nothing yet is not ready, it is early. The `MISSING SOURCES: N of N` diagnostic already knows the condition; today it only explains the red.
2. **PARKED is not terminal while the source set is still growing** — a compile failure recorded with `MISSING SOURCES: all` should re-arm on the next source-node arrival rather than serve the cache.

Until then the gate re-runs the shard (`rerun-failed-jobs`); this issue is the place the second occurrence goes. Related: #3929 (the diagnostic), #2813 (the stamp check that correctly declined).

rbuergi:
Fixes #4208

## The mechanism, at file:line on `origin/main`

`Reinsurance/AggregateSection` was adopted and then judged 2 ms later against a live source
fingerprint of `e3b0c44298fc1c14` — which is `NodeTypeSourceFingerprint`'s fold of an **empty**
source set. The type sorts first in its namespace, ahead of the `Reinsurance/Source/*` nodes its
`shared=@…` queries read; those had not been written yet.

`src/MeshWeaver.Compiler.Pipeline/NodeTypeCompilationHelpers.cs:1220` guards the absence that #2813
already knew about:

```csharp
if (def.CurrentSourceFingerprint is not { Length: > 0 } live)
    // …Not a mismatch — nothing has been compared.
    return stamped with { BuildProvenance = BuildProvenance.AdoptedUnverified };
```

An empty fold is a perfectly well-formed 16-character value, so it walks straight past that guard —
and past the identical `is { Length: > 0 }` in `InstallAdoptedSourceStampWatcher`'s `Where`
(`:1503`), in the release-request watcher (`:2155`), and in `PrebuiltAssemblySeeder.SeedObserved`
(`:720`). The comparison then ran, the values differed, and the conclusion drawn was *"the source
moved past the adopted build"*. It had not moved; there was nothing there yet to have moved.

Everything after that followed honestly from a false premise: the refusal dispatched a compile
(`DispatchPending`), that compile ran with **3 of 3 declared source queries matching NO nodes** —
which `SourceCoverage.Describe` named precisely — and Roslyn's completely genuine-looking `CS0246`s
parked the type. `AmountType` and every later sibling, declaring the identical queries, adopted
green once the 26 source nodes had landed milliseconds later.

The two unanswerable states are the same state: **nothing has been compared.** Only one of them was
guarded, because only one of them looks empty.

## The fix

- `NodeTypeCompilationHelpers.CanJudgeAdoption` — the ONE predicate every writer asks before
  spending the one-shot stamp request, so no two of them can disagree about when the three-way is
  answerable. Three ways it is answerable and nothing else is: no adopted fingerprint (a legacy
  bundle), the two fingerprints **equal** (including the honest empty==empty of a type that
  genuinely compiles from no sources), or **a declared source query matched a node**. It subsumes
  the existing `#2813` guard exactly.
- 🚨 **The witness is the source SNAPSHOT, not the hash** (review finding, `1d978d8`). The tick
  snapshot counts every MATCHED node; the fingerprint is folded over `CollectCompileSources`, which
  drops executable cells and blank files — so a type whose only matched node is one of those has an
  established source set AND the empty fold, and a hash-based witness would have deferred its
  adoption *indefinitely*, trading this park for a permanent hold.
  `NodeTypeSourceFingerprint.EmptySourceSet` remains as a **diagnostic** (it names what was measured
  in the deferral log) and its doc says it must never be the witness.
- When it is false the judgement is **DEFERRED, not refused**: `ApplyAdoptedSourceStamp` returns the
  record untouched, so `RequestedSourceStampAt` is left **STANDING**. The wait is reactive with
  nothing added — the sources watcher re-publishes the full set on every source arrival (a live
  synced query; no timer, no poll, no `Task.Delay`) and that publication fulfils the standing
  request in the same write. At that point the fingerprints agree and the adoption **verifies**.
- Nothing is stranded meanwhile: `CompiledSources` and `CurrentSourceVersions` are both empty, so
  `IsDirty` is false and the install's release request is satisfied by the adopted build instead of
  compiling it.
- The seeder's pre-write decline gets the same predicate, for the same reason — its premise
  (*"sources only ever move the live fingerprint further from a bundle baked earlier"*) does not
  hold for a set nobody established — and now says so in the log when it defers.
- `ApplyAdoptedSourceStampAndReport` logs the deferral at **Warning**, naming which of the two
  absences it is. On the gate mesh, which runs at Warning, that is the one line saying a NodeType
  was activated ahead of the sources it declares — where previously the same moment produced a page
  of compiler errors about somebody else's code.

## 🚨 Two negative controls

**The production path, on a real mesh** — `AdoptionDeferredUntilSourcesLandTest`: a NodeType whose
declared queries match nothing, a real `PrebuiltAssemblySeeder.SeedDetailed` with a disagreeing
fingerprint, then the source node arrives and the standing request converges to `AdoptedVerified`
with nothing driven by hand (and a `NotEmit` proving nothing re-drives the type meanwhile — a
standing request must not become a poll). **It failed on its first run and the bug was mine**: at
the seeder's pre-write point the adoption has not happened, so `AdoptedSourceFingerprint` is null on
the observed node, my call took the legacy-bundle row and the decline fired on the very absence it
was meant to defer on. There is now a three-argument overload taking the producer's fingerprint
explicitly. Every pure test was green throughout.

**The decision, pinned pure**

`test/MeshWeaver.Compiler.Pipeline.Test/AdoptionIsNotJudgedFromAnEmptySourceSetTest.cs` — pure, no
mesh, no timing. With the guard removed and **everything else unchanged**, exactly the two #4208
assertions fail:

```
MeshWeaver.Reactive.Assertions.AssertionException : Expected value to be 9/13/2026 4:16:50 PM +00:00
because the stamp request is ONE-SHOT — answering it from a set nobody established spends the only
chance to refuse and closes the question for good, but found <null>.

MeshWeaver.Reactive.Assertions.AssertionException : Expected value not to be <null>
because precondition: the request is standing.

Failed! - Failed: 2, Passed: 12, Skipped: 0, Total: 14
```

…while every control still passes — a real disagreement with a **major bump** is still
`AdoptionRefused` with the bytes cleared; within the same major it is still `StaleAdopted` with the
compile dispatched; a type that genuinely compiles from **no** sources still verifies (equality
outranks the guard); and the predicate table. `EmptySourceSet` is pinned to BOTH the fold's own
output and the literal `e3b0c44298fc1c14` the gate log recorded, so a drifted constant cannot make
the rest vacuous.

## Two reworked fixtures — and why that is not "fixing the test"

`StaleDeclineNeverDanglesTest` and `LeavingHubAdoptionSweepTest` build a decline on a **sourceless**
fixture type, whose live fingerprint therefore *is* the empty fold. Their subject — *"a bundle the
owner would refuse is declined before it writes"* — post-#4208 requires a live source set that
establishes something, so those inputs no longer reach it: they were exercising the defect. Each
fixture now creates **one real source node**; the fingerprint is still stated with the product's own
function over that node, so the watcher's recompute writes the same value and cannot move it
mid-test (the race that reddened the merge queue twice on #3143 stays closed). Nothing compiles —
both records carry `CompilationStatus.Ok` and a usable build, so no kickoff arms — and each now
asserts `Establishes(expected)` as an explicit precondition, so a future fixture that drifts back to
the empty fold fails loudly instead of testing nothing.

## Verified locally with CI's flags

```
dotnet build src/MeshWeaver.Compiler.Pipeline/... -c Release -warnaserror   0 Warning(s)  0 Error(s)
dotnet test  test/MeshWeaver.Compiler.Pipeline.Test  Passed!  931/931
dotnet test  test/MeshWeaver.Graph.Test              Passed! 1592/1592
dotnet test  test/MeshWeaver.Documentation.Test      Passed!  447/447
```

(Those are the only two test projects in the repo that reference `CurrentSourceFingerprint` /
`AdoptedSourceFingerprint`.)

What's New: `Category: Fix` — a user sees a type that installed as failed now install cleanly.

The finding is conserved in `Doc/Architecture/NodeTypeCompilation` — the `ApplyAdoptedSourceStamp`
table gains this row (and its pre-#3583 "matches: no" row is corrected in the same pass, since it
still described an unconditional refusal).

Pairs-with: none — no `public` top-level type or member is removed from `src/`; the change adds one
public member (`NodeTypeSourceFingerprint.EmptySourceSet`), two `internal` predicate overloads, and
behaviour behind existing signatures.

Mirror-sync: none — no `strings.{en,de}.json` key is added or re-worded.


rbuergi:
## All three review findings acted on — and one of them found a real bug in my own change

### 2. `Establishes` infers the source set from the HASH, but the fold drops nodes the snapshot keeps — **correct, and fixed**

This is the substantive one. `CurrentSourceVersions` counts every MATCHED node; the fingerprint is
folded over `NodeCompileShaping.CollectCompileSources`, which deliberately drops executable cells and
blank files. A NodeType whose only matched source node is one of those therefore has an **established
source set** and the **empty fold** — and my predicate would have deferred its adoption *indefinitely*,
trading #4208's park for a permanent hold. That is a new failure mode, introduced by the fix.

The witness is now the **snapshot**, not the hash:

```csharp
internal static bool CanJudgeAdoption(
    string? adoptedFingerprint, string? liveFingerprint, IReadOnlyDictionary<string, long>? liveSources)
{
    if (adoptedFingerprint is not { Length: > 0 } adopted) return true;   // legacy bundle
    if (liveFingerprint   is not { Length: > 0 } live)    return false;   // #2813 — not published yet
    return string.Equals(adopted, live, StringComparison.Ordinal)         // a match is a match
           || liveSources is { Count: > 0 };                              // #4208 — did anything match?
}
```

"Did any declared source query match a node?" is the question the incident is actually about
(*MISSING SOURCES: 3 of 3 declared source queries matched NO nodes on this mesh*). `Establishes` is
**removed**; `EmptySourceSet` stays as a diagnostic — it names what was measured in the log line — and
its doc now says explicitly that it must never be used as the witness, with the reason.
`AMatchedSourceTheFoldDROPS_IsStillAnEstablishedSourceSet` pins exactly your case: a blank Code node
folds to `EmptySourceSet`, and the adoption is still judgeable.

### 1. Equality is consulted before the witness — **considered, kept, and named as residue**

Both halves of this are real, and both are narrower than they look:

- **empty==empty consuming the request while an import is in flight.** Reachable only when the
  *producer* also hashed nothing, i.e. the bundle was baked from no sources and cannot contain the
  type's code anyway. This is **unchanged from `main`** — the equality branch has always taken it —
  and ordering the witness *before* equality would demote the honest verification of a type that
  genuinely compiles from no sources (pinned by `ATypeThatCompilesFromNoSourcesAtAll_StillVerifies`).
  A witness that distinguishes the two does not exist on the record.
- **a real deletion leaves the request standing forever.** True, and deliberate. It is bounded and
  benign: the adopted build keeps serving, `CompiledSources` and `CurrentSourceVersions` are both
  empty so `IsDirty` is false, nothing recompiles and nothing parks, and the owner logs a **Warning
  naming the type** on every judgement attempt. The alternative is what #4208 measured — a compile
  against nothing, `CS0246`s about correct code, and a parked type. It is now written into the
  predicate's doc and into `Doc/Architecture/NodeTypeCompilation` as named residue rather than left
  implicit.

### 3. The tests never run the seeder or the watcher — **correct, and it caught a bug I had just written**

Added `AdoptionDeferredUntilSourcesLandTest` on a real mesh: a NodeType whose declared queries match
nothing, a real `PrebuiltAssemblySeeder.SeedDetailed` with a disagreeing fingerprint, then the source
node arrives and the standing request converges — nothing driven by hand.

**It failed on the first run**, and the failure was mine, not the fixture's:

```
Expected value to be Adopted …, but found DeclinedStaleSources.
```

At the seeder's pre-write point the adoption has not happened, so `AdoptedSourceFingerprint` is
**null** on the observed node — my `CanJudgeAdoption(observed, …)` call therefore took the
legacy-bundle row, answered "judgeable", and the decline fired on the very absence it was meant to
defer on. The producer's fingerprint is the **bundle's**, and it arrives as an argument; there is now
a three-argument overload whose doc says exactly that, and the seeder passes `producerFingerprint`.
The pure tests were green through all of it — which is your point, made in full.

The mesh test also asserts, with `NotEmit`, that nothing re-drives the type while the sources are
still missing: a standing request must not become a poll.

### Green

```
dotnet test test/MeshWeaver.Compiler.Pipeline.Test   Passed! 931/931   (was 927 — 4 new)
dotnet test test/MeshWeaver.Documentation.Test       Passed! 447/447
dotnet build … -c Release -warnaserror               0 Warning(s)  0 Error(s)
```

(The new mesh test's waits are `TestTimeouts.*`, not literals — `TestTimeoutLiteralRatchetGuard`
caught two literals I had copied from the fixtures I was modelling on.)

PR #4219, opened 2026-09-13
fix: wait-for-postgres probes every database host the pod will open (#4173)

### LINKED ISSUE #4173: wait-for-postgres gates only MEMEX_HOST — an explicit ConnectionStrings__orleans naming another host is gated by nothing
## What is wrong

The portal's `wait-for-postgres` init container exists to make one thing deterministic — *"Postgres and the portal start concurrently on a fresh install; the portal's hub init FAILS (and the host aborts) if Postgres isn't yet accepting connections. Gate startup on Postgres TCP readiness so the ordering is deterministic."* It probes **one** host, and that host comes from a **different input** than the connection strings the process actually opens at boot.

`deploy/helm/templates/memex-portal/deployment.yaml:104-111`:

```yaml
- name: "wait-for-postgres"
  image: "busybox:1.36"
  command:
    - "sh"
    - "-c"
    {{- $pgHost := .Values.postgres.enabled | ternary "memex-postgres-service" (.Values.config.memex_portal.MEMEX_HOST | default "memex-postgres-service") }}
    {{- $pgPort := .Values.postgres.enabled | ternary "5432" (.Values.config.memex_portal.MEMEX_PORT | default "5432") }}
    - "until nc -z {{ $pgHost }} {{ $pgPort }}; do echo 'waiting for postgres'; sleep 2; done"
```

The probe's host is `config.memex_portal.MEMEX_HOST` — a **config** value, on AKS derived from the deployment record (`DeploymentPortalConfig.DatabaseHost`, `src/MeshWeaver.Deployment.Contract/DeploymentPortalConfig.cs:69-73`: explicit FQDN → Azure server FQDN → in-cluster service).

The connections the boot actually opens are **secrets**, rendered from the Key Vault values half that no repository holds (`deploy/helm/templates/memex-portal/secrets.yaml`):

| key | line | host it names |
|---|---|---|
| `ConnectionStrings__memex` | `:29` | whatever `secrets.memex_portal.ConnectionStrings__memex` says, else the in-cluster default |
| `ConnectionStrings__orleans` | `:74`, `:94` | **explicit** (`secrets.memex_portal.ConnectionStrings__orleans`, *"it always wins"*), else **derived** from `ConnectionStrings__memex` + `Search Path=orleans`, else the in-cluster default |

Nothing renders, asserts or logs that those hosts equal `MEMEX_HOST`. `grep -rn MEMEX_HOST deploy/helm/templates/ scripts/` returns four hits: the two probes and the two ConfigMaps that publish the value — no comparison anywhere.

So:

- **Explicit `ConnectionStrings__orleans` naming another host** (a dedicated `orleans` server, or the same server under another name) — that host is gated by **nothing**. The init container can pass on a host that is up while the silo then opens a connection to one that is not.
- **Derived `ConnectionStrings__orleans`** (the AKS default for an external mesh database) inherits `ConnectionStrings__memex`'s host — covered by the probe **only if** that host string equals `MEMEX_HOST`, which nothing checks. They are set from two different halves of the values.
- **The migration Job has the same shape**: `deploy/helm/templates/memex-migration/job.yaml:70` gates on `config.memex_migration.MEMEX_HOST` and then migrates against `ConnectionStrings__memex`.

## What an ungated host produces at boot

Exactly the fingerprint of [#3780](https://github.com/Systemorph/MeshWeaver/issues/3780): the pod passes `Init:1/1`, the silo reaches lifecycle stage `RuntimeGrainServices`, and `MembershipTableManager` fails on the orleans connection —

```
'MembershipTableManager' failed to start due to errors at stage 'RuntimeGrainServices (8000)'.
Npgsql.NpgsqlException: Name or service not known
 ---> System.Net.Sockets.SocketException: Name or service not known
   at System.Net.Dns.GetHostEntryOrAddressesCore(...)
```

— and the container exits, because the portal fails closed. The chart rolls surge-first (`maxUnavailable: 0`), so the cost of one occurrence is a delayed roll rather than an outage; a persistent one stalls every roll of the instance. **The init container's passing tells you nothing about the connection that failed**, which is what makes this read as a transient resolver blip rather than as a gate that was never covering that host.

## 🚨 What this is NOT

It is **not** established as the cause of #3780. That incident's values live in Key Vault (`helm-values-memex`), which I cannot read, so I do not know whether `memex`'s `ConnectionStrings__orleans` is explicit or derived, nor whether its host string equals `MEMEX_HOST`. #3780's own measurement — two later boots of the same deployment through the same lifecycle stage on the same secret — excludes a *permanently* wrong host there. This issue is the **hole in the guard**, filed on its own because it is true whatever #3780's cause turns out to have been: a deployment whose orleans host differs from `MEMEX_HOST` has an init container that is not gating the connection the silo makes.

## What would close this

The gate covers every host the boot connects to, or says loudly that it cannot:

1. **Derive the probe's host(s) from the rendered connection strings** rather than from a parallel config value — probe each distinct `Host=` the pod's secret carries (mesh + orleans), so "the gate passed" means "every database this pod will open is accepting connections"; or
2. **Assert agreement at render time** — `fail` in the template when an explicit `ConnectionStrings__orleans` (or `ConnectionStrings__memex`) names a host other than `MEMEX_HOST`, so a divergence is a red render naming both values instead of a silent hole. The chart already uses `fail` for the values-shape traps it refuses to guess at.

Either way the verification is the same and does not need a cluster: render the chart with `postgres.enabled=false`, `MEMEX_HOST=a`, and an explicit `ConnectionStrings__orleans` naming host `b`, and assert that the output either probes `b` or fails the render. Today it probes `a` and says nothing about `b`.

Found while re-measuring #3780 (2026-09-13); read-only, nothing changed.

rbuergi:
Fixes #4173
Refs #3780 — see "What this means for #3780" below; that issue is not closed here.

## The mechanism, at file:line on `origin/main`

`deploy/helm/templates/memex-portal/deployment.yaml:104-111` (and the identical shape at
`memex-migration/job.yaml:70`):

```yaml
{{- $pgHost := .Values.postgres.enabled | ternary "memex-postgres-service" (.Values.config.memex_portal.MEMEX_HOST | default "memex-postgres-service") }}
- "until nc -z {{ $pgHost }} {{ $pgPort }}; do echo 'waiting for postgres'; sleep 2; done"
```

The probe's host is a **config** value. The connections the boot actually opens are **secrets**
(`memex-portal/secrets.yaml:29` and `:74`/`:94`), rendered from the Key Vault values half that no
repository holds. `grep -rn MEMEX_HOST deploy/helm/templates/ scripts/` returns four hits — the two
probes and the two ConfigMaps that publish the value. **No comparison anywhere.**

So an explicit `ConnectionStrings__orleans` naming another host was gated by *nothing*, and a
derived one was covered only if its inherited host happened to equal `MEMEX_HOST`, which nothing
checks. What an ungated host produces at boot is #3780's exact fingerprint: `Init:1/1`, the silo
reaches `RuntimeGrainServices`, and `MembershipTableManager` dies on `Name or service not known`
with no TCP attempt ever made.

🚨 **The expensive part is the init container's PASS.** It says nothing about the connection that
failed, so a gate that was never covering that host reads as a transient resolver blip. That is the
same shape as a skipped CI job wearing a passing tick.

## The fix

- **`deploy/helm/templates/_database.tpl`** — every database connection string the chart renders,
  and the endpoints derived *from* those strings, as ONE definition consumed by the Secret that
  carries them and the init container that waits for them. `memex-migration/secrets.yaml` already
  *claimed* this property — *"both are derived by the same rule from the same input rather than
  typed twice"* — while the rule was in fact written out twice; now it is written once.
- Both `wait-for-postgres` init containers probe **every distinct `host:port` the half will open**
  (mesh always, orleans under AdoNet), deduplicated, **naming each endpoint as they wait** — so the
  log says *which* host is not answering.
- **No skip-trapdoor**: `memex.dbProbeTargets` calls `fail` when no host can be derived. An init
  container running `until nc -z ; do …` would pass having probed nothing, and *"the gate could not
  determine what to wait for"* must never render as *"the gate passed"*.
- **Invariant 16** in `check-chart-invariants.sh`/`.py`, for BOTH halves: every host the rendered
  Secret's connection strings name must appear in that half's `wait-for-postgres` command. The
  migration Job is matched by name **prefix** (its name carries `.Release.Revision`), so the lookup
  cannot silently find nothing and make half the check vacuous — and an absent object or Secret is
  a **finding**, not an early return, so the check cannot evaluate nothing while the gate reports
  success.
- **`deploy/aks/scripts/deploy.sh` hands the connection string to helm** instead of patching both
  Secrets afterwards. A value that arrives after the render cannot reach the init container the
  render produced — so on that path the probe would have kept naming `memex-postgres-service` while
  the process opened the Flexible Server. It also removes the hazard `memex-portal/secrets.yaml`
  documents and this script was still creating: helm replaces a Secret wholesale, so any later
  `helm upgrade` overwrote the working connection string.
- **A multi-host `Host=a,b` failover list is covered in full**, and the waiter treats the group as
  *any ONE of these answers* — the process's own requirement. Requiring all of them would block a
  deployment whose standby is legitimately down; covering only the first would let the gate pass
  having waited for `a` while the process connects to `b`.
- **A new fixture**, `values.dedicated-orleans-host.yaml`: mesh on `mesh-db.example.test:5432`,
  Orleans membership on `orleans-db.example.test:6432`, `MEMEX_HOST` deliberately left at the chart
  default naming neither. That is the shape nothing in this repo rendered, and it is #4173's own
  stated verification.

## 🚨 Negative controls — both measured

**1. The invariant can fail.** Old probe restored, everything else unchanged:

```
::error::[two plain replicas, KEDA off (the memex shape)] portal: ConnectionStrings__memex names host 'pg.example.test:5432' that wait-for-postgres does not probe
::error::[two plain replicas, KEDA off (the memex shape)] portal: ConnectionStrings__orleans names host 'pg.example.test:5432' that wait-for-postgres does not probe
::error::[two plain replicas, KEDA off (the memex shape)] migration: … ×2
[two plain replicas, KEDA off (the memex shape)] 4 contradiction(s) across 22 invariants.
::error::[a dedicated orleans server (fixture)] portal: ConnectionStrings__memex names host 'mesh-db.example.test:5432' that wait-for-postgres does not probe
::error::[a dedicated orleans server (fixture)] portal: ConnectionStrings__orleans names host 'orleans-db.example.test:6432' that wait-for-postgres does not probe
::error::[a dedicated orleans server (fixture)] migration: … ×2
[a dedicated orleans server (fixture)] 4 contradiction(s) across 22 invariants.
```

(With the failover pair in the fixture it now reports **6** contradictions there, naming `orleans-a`
*and* `orleans-b`.) Meanwhile the **six** combinations whose hosts already agreed stay green, so the
invariant is not trivially true.

**1b. The gate cannot skip a half.** Suppressing `memex-migration-secrets` behind `{{- if false }}`
now fails **all 8** combinations (`migration: memex-migration-secrets was not rendered, so there is
nothing to compare the probe to`). Before `5d3aaf0` that suppression failed **none** of them — the
check returned early without counting anything, which is the skip-trapdoor shape this script exists
to prevent.

**1c. The waiter was EXECUTED**, against a stub `nc`, because a shell loop nobody ran is an
assumption: it skips an unreachable failover member and accepts the reachable one, and it retries a
host that is down for two attempts and then proceeds. Neither a no-op nor a hang. Note that `two plain replicas, KEDA off` is the **memex.systemorph.com shape** — a
combination this repo already shipped was un-gated, not just the new fixture. With the fix in place:
`All 8 values combinations render a self-consistent deployment.`

**2. The refactor moved no connection string.** Rendering all seven previously-shipped combinations
before and after, the only substantive line-level difference across the entire manifest set is the
init-container command:

```
-  - "until nc -z memex-postgres-service 5432; do echo 'waiting for postgres'; sleep 2; done"
+  - "for t in memex-postgres-service:5432; do h=${t%:*}; p=${t##*:}; until nc -z $h $p; do echo 'waiting for postgres at '$h:$p; sleep 2; done; echo 'postgres ready at '$h:$p; done"
+  - "for t in pg.example.test:5432; do …"
```

**Every rendered Secret is byte-identical.**

## What this means for #3780

#3780's cause was never established — the issue named two candidates (a transient resolver blip in
the boot window, or a `ConnectionStrings__orleans` naming a host that does not exist) and its values
live in Key Vault, which neither I nor the issue's author can read. **This change makes the posture
question in it moot rather than deciding it**, which is why it is `Refs`, not `Fixes`:

- **If it was a blip**, the retry now exists where waiting is free and visible — the init container
  holds the pod at `Init:0/1` until the orleans host resolves, *before* the silo starts, so a blip
  costs a delayed pod rather than a crash loop.
- **If it was a wrong host**, the pod now holds at `Init:0/1` with the offending host named in its
  own log, instead of passing and dying in `MembershipTableManager` on an error that names nothing.

Either way the silo's **fail-closed stays correct**: by the time it runs, every host it will open
has already resolved and accepted TCP, so a failure there is a real answer rather than a race. No
bounded-resolution-retry is added to the silo, no timer, no watchdog.

I am not closing #3780 on this: it has not recurred, so there is no measurement that would confirm
the cure, and it should be closed after this ships if the boot shape is what the change predicts.

## Verified locally

```
deploy/aks/scripts/check-chart-invariants.sh      All 8 values combinations render a self-consistent deployment.
shellcheck -S warning deploy/aks/scripts/deploy.sh          (clean)
deploy/aks/scripts/check-values-are-read.sh       All 125 config key(s) across 3 values file(s) are read by a template
deploy/aks/scripts/test-chart-drift-compare.sh    all classification assertions passed
python3 deploy/aks/scripts/test-observability-storage.py   OK (2 tests)
python3 deploy/aks/scripts/test-observability-values.py    OK (9 tests)
dotnet test test/MeshWeaver.Documentation.Test -c Release  Passed! 447/447
```

What's New: `Category: Fix` — an operator sees a roll that stalled visibly, naming the host, instead
of a pod that passed its gate and then died.

Pairs-with: none — no `public` top-level type or member is removed from `src/`; the C# side of this
diff is documentation only.

Mirror-sync: none — no `strings.{en,de}.json` key is added or re-worded.


rbuergi:
## All five review findings acted on — and two of them were real holes in the gate I had just written

Pushed as `5d3aaf0` (`fix: close three holes the review found in the new probe and its gate`).

### 1–3. `deploy.sh` patches the Secrets AFTER `helm upgrade` — **correct, and fixed**

`deploy/aks/scripts/deploy.sh` patched both Secrets with `MEMEX_PG_CONN` once the upgrade had
finished, under the note *"Chart-gen gap: the secret template hardcodes the in-cluster pg connection
string"*. That gap closed when the template started taking `secrets.<half>.ConnectionStrings__memex`
from values; the patch outlived its reason and was harmful twice over:

- **helm replaces a Secret wholesale**, so any later `helm upgrade` overwrote the working connection
  string — the exact hazard `memex-portal/secrets.yaml` documents, and this script was the thing
  still creating it;
- **a value arriving after the render cannot reach the init container the render produced**, so on
  that path the probe would have kept naming `memex-postgres-service` while the process opened the
  Flexible Server. My PR body claimed "every host this pod will open"; on that path it would not
  have been true.

The string now goes in **through helm**, as a values file rather than `--set` (a connection string
carries `;`, `=` *and* `,` — three ways for `--set` to mangle it silently), written to a `mktemp`
file that is deleted on `trap … EXIT` because it holds the password. The post-helm patch loop is
gone. `bash -n` and `shellcheck -S warning` both clean.

### 4. The invariant could **skip a half** — correct, and it is the rule this script exists for

This is the one that should not have shipped. `_probe_coverage` returned early *without counting a
check* when the object or its Secret was absent, so a template regression that removed
`memex-migration-secrets` would have left invariant 16 evaluating **nothing** while the gate reported
success. Both the portal Deployment and the migration Job are unconditional in this chart, so either
going missing is itself a defect — and "the object is gone" must never read as "the object's gate is
fine" (AGENTS.md → *A gate NEVER tests its own inputs*).

It now counts the check **first** and **reports** the absence. Measured, by suppressing
`memex-migration-secrets` behind `{{- if false }}`:

```
::error::[self-host (neutral chart defaults)] migration: memex-migration-secrets was not rendered, so there is nothing to compare the probe to
…8 of 8 combinations
```

Before the change that suppression failed **none** of them.

### 5. A multi-host `Host=a,b` was covered by its first entry only — correct, and fixed

Npgsql accepts a comma-separated failover list and may open either member, so a probe of `a` could
pass while the process connects to `b` — the same "the gate passed says nothing about the connection
that failed" defect one level down.

`memex.dbHostGroup` now keeps **every** member, and the waiter treats the group as **any ONE of
these answers** (`break 2` on the first that responds). Requiring *all* of them would be the opposite
error: the point of a failover list is that one member suffices, so demanding a standby be up would
block a valid deployment. The invariant requires every member to be **probed**.

The `values.dedicated-orleans-host.yaml` fixture now carries a failover pair, so the new coverage is
itself exercised — with the old probe restored it reports **6** contradictions there, naming
`orleans-a` *and* `orleans-b`.

And the rendered waiter was **executed** against a stub `nc`, twice, because a shell loop nobody ran
is an assumption:

```
# a member that never answers, one that does
postgres ready at mesh-db.example.test:5432
postgres ready at orleans-b.example.test:6432

# a host that is down for two attempts and then comes up
waiting for postgres at mesh-db.example.test:5432
waiting for postgres at mesh-db.example.test:5432
postgres ready at mesh-db.example.test:5432
```

So the loop is neither a no-op nor a hang.

### Still green

```
deploy/aks/scripts/check-chart-invariants.sh    All 8 values combinations render a self-consistent deployment.
deploy/aks/scripts/check-values-are-read.sh     All 125 config key(s) across 3 values file(s) are read by a template
shellcheck -S warning deploy/aks/scripts/deploy.sh   (clean)
```

The What's New entry gained a paragraph for the `deploy.sh` change and one for the failover-list
rule, since both are things an operator can notice.

rbuergi:
### The `deploy.sh` path, rendered end to end

The script now writes a values file rather than patching afterwards, so that file is an input a
render can be run against. Driving the AKS overlay with exactly what it writes, for a deliberately
hostile connection string (`;`, `:`, `#`, `,` and spaces in the password — every character `--set`
would have mangled):

```
$ MEMEX_PG_CONN='Host=pg-x.postgres.database.azure.com;Port=5432;Username=memexadmin;Password=p@ss:w#rd,1;Database=memex;SslMode=Require;Trust Server Certificate=true'
$ # …the exact heredoc deploy.sh emits → values file → helm template -f values.yaml -f values.aks.yaml -f <that file>

memex-portal-secrets    ConnectionStrings__memex   = 'Host=pg-x.postgres.database.azure.com;Port=5432;Username=memexadmin;Password=p@ss:w#rd,1;Database=memex;SslMode=Require;Trust Server Certificate=true'
memex-portal-secrets    ConnectionStrings__orleans = '…;Search Path=orleans'
memex-migration-secrets ConnectionStrings__memex   = '…'   (identical)
memex-migration-secrets ConnectionStrings__orleans = '…;Search Path=orleans'

wait-for-postgres: for g in pg-x.postgres.database.azure.com:5432; do …
```

The value round-trips byte-for-byte through the YAML literal block, the derived orleans string picks
up the same server, and **the probe waits for the external host** — which the post-helm patch made
structurally impossible, because it landed after the render that produced the init container.

PR #4221, opened 2026-09-13
fix(#3438): an installation served by ANOTHER REGISTRY is named, not read as pinning nothing

rbuergi:
> **Stacks on #4213** (same branch line, same issue). Until #4213 merges this PR's diff shows both
> commits; it cleans up automatically once #4213 lands. Merge #4213 first.

## 🚨 The protection lane is RED right now, and its stated reason is FALSE

`Lock pinned image digests`, scheduled run 2026-09-13T01:18Z
([34730206406](https://github.com/Systemorph/MeshWeaver/actions/runs/34730206406)):

```
##[error]AXIS 3 — installation `build` runs core c84c6c0 and its overlay
(Systemorph/Memex deployments/aks/build/values.build.public.yaml) pins no image at all,
so there is no repository in which to protect what it runs.
```

**That overlay pins two images.** Run the shipped extractors over the actual committed bytes:

| overlay | ACR pins | foreign pins |
|---|---|---|
| `deployments/aks/build/values.build.public.yaml` | **0** | **2 → `cr.meshweaver.cloud`** (`memex-portal-ai:3.0.0-ci.8411`, `memex-migration:3.0.0-ci.8411`) |
| `deployments/aks/memex/values.memex.public.yaml` | 6 | 0 |

`REGISTRY_HOST_RE` is `(?P<host>[a-z0-9]+)\.azurecr\.io` — correct for a lane that locks one ACR,
but it made *"pins its images elsewhere"* spell identically to *"pins nothing"*. The sentence the
run prints is an instrument answering confidently wrong, and the fix it suggests is to go and repair
an extractor that has no defect.

## And the red was not free

`pause.reEnableWhen` in `.github/acr-retention/tasks.json` is literally *"lock-pinned-digests is
green on a fleet whose tag references and index closures are locked"*. So **one live installation
moving registry was standing between the fleet and re-enabling cleanup** — and because the run
refuses as a whole, `memex` (six pins in this ACR, and the one actually exposed to the ACR purge)
got **no locks either**. One instance's registry question held the other's protection hostage.

`build` is the fleet's build server (`build.meshweaver.cloud`, MeshWeaver.Plugins#1784) and the
first **live** installation on `cr.meshweaver.cloud`. `pearl` pins there too and is declared
`not-installed`, so it was never asked — which is why this surfaced only now.

## What changed

Foreign references are **extracted and named**. They are never locked — nothing here can write to
another registry — but four facts stay distinct:

| the overlay | the run |
|---|---|
| pins in this ACR | protected, unchanged |
| pins **only** elsewhere, **undeclared** | **RED**, naming the registry and what to write. Silently skipping would make an unprotected installation look like a protected one |
| pins **only** elsewhere, **declared** | accepted, and printed on its **own denominator line** — `…declared OUT OF THIS REGISTRY'S SCOPE` — so it can never be counted among the answered and read as covered |
| pins **nothing**, in any registry | RED, and the message now says "in ANY registry" so it means what it says |

`registry` in `instances.json` is a **scope statement, never an exemption from answering**: the
installation stays live, stays expected to answer `/api/version`, stays counted. It is checked
**both ways** against what the overlay actually pins, so it cannot go stale in silence — a
declaration naming a registry the overlay does not pin in is RED, and so is one on an installation
that **also** pins in this ACR (half in is not out; excusing it wholesale would leave the half that
*is* here unprotected). Refused on a non-live entry, and refused without a `reason`.

## The negative controls — eleven arms, all falsified and restored

**Sabotage A** — `extract_foreign_pins` stubbed to `return []`, i.e. exactly the pre-fix behaviour.
It reproduces the live run's own false sentence and reds six arms:

```
::error::ARM 32: the foreign extractor did not see the two `cr.meshweaver.cloud` pins in the real overlay shape: []
::error::ARM 32: an UNDECLARED installation served by another registry did not red naming that registry:
         ['AXIS 3 — installation `build` runs core 1234567 and its overlay (…/values.build.public.yaml)
           pins no image at all, in ANY registry, …']
::error::ARM 32: the run still says an overlay with two pins in it 'pins no image at all'. That sentence
         is FALSE and it sends the reader to fix an extractor that is working
::error::ARM 32: a declared out-of-scope installation was not marked as such, so it would be counted
         among the answered and read as PROTECTED
… (2 more)
```

**Sabotage B** — the two declaration checks short-circuited, so a scope statement is honoured
without being verified:

```
::error::ARM 32: a STALE registry declaration was honoured: []
::error::ARM 32: an installation pinning in BOTH registries was excused wholesale by a scope
         declaration, which would leave its ACR half unprotected: []
```

Restored: `--self-test` exits 0. The fixture is the **real** `build` overlay shape, and the
extractors were additionally run over the **actual committed bytes** of both overlays (the table
above) — so the fix is measured against the file that broke the lane, not only against a fixture.

## What this does NOT claim

🚨 **What retains `cr.meshweaver.cloud` is not established by this.** The declaration says those
images are out of *this lane's* reach; it claims nothing about what keeps them. That is now an open
item on #3438 — the same question this mechanism answers for the ACR, asked again about a second
store, and it grows with every installation provisioned on the fleet's own registry, which is the
default for new ones.

Read-only against Azure and the registry throughout: no purge, no lock, no dry run, no `apply`, no
registry mutation of any kind.

## Verification

- `lock-pinned-digests.py --self-test` → 0; ARM 32's arms falsified in two directions and restored
- `lock-pinned-digests.py --check-retention-record .` → 0
- `check-pinned-digests.py --self-test`, `check-pin-set-consistency.py --self-test`,
  `check-workflow-shell.py --root .` → 0
- `dotnet build test/MeshWeaver.Documentation.Test -c Release -warnaserror` → `0 Warning(s) 0 Error(s)`
- `dotnet test test/MeshWeaver.Documentation.Test -c Release --no-build` → **447 passed, 0 skipped**

Docs: `Doc/Architecture/ArtifactRetentionInterlock` gains *A second registry*, and
`cr.meshweaver.cloud`'s own retention is added to the maintainer list. What's New entry included.

Refs #3438


rbuergi:
## Review round: six findings addressed, and the measurement changed the design

All six Copilot findings were real. Measuring the fleet before fixing finding 3 turned up a
**seventh** — my own extractor was matching inside **comments**.

### What the measurement said

Across the fleet's **twelve** deployment overlays, 2026-09-13, before deciding anything:

| host | references | what they are |
|---|---|---|
| `cr.meshweaver.cloud` | 4 | `build` ×2, `pearl` ×2 — real pins, our images |
| `ghcr.io` | 2 | one real (`ghcr.io/distribution/distribution`, the registry service's own image, in memex-cloud's overlay) and **one line of PROSE** in `ci-runners` describing what the runner image is built from |

So "every undeclared foreign host reds", applied literally, would have reddened the lane over a
third-party image **and over a sentence**. This issue already paid for that lesson: of the 31
`sha256:` tokens in its own 2026-09-08 hand count, **11 were comments**, several narrating this
incident inside the files it broke.

### The six findings

1. **Declared-drift exemption keyed only by task name** → fixed in **#4223** (it is a #4213 defect,
   and that PR was queued when the review landed).
2. **`*.azurecr.io` treated as "any ACR"** → `--registry meshweaver` locks one ACR; a pin in a
   different one is now foreign. Extracting its repository and looking it up here is this bug
   reintroduced one registry along.
3. **A mixed-registry overlay with no declaration was silently accepted** (the `if not repositories`
   branch is never reached when the ACR half is non-empty) → an undeclared host now reds **wherever
   it appears**. And **half covered is not covered**: pinning both here and in a `fleet-unlockable`
   registry is a blocker, because locking the ACR half and reporting success is a
   partially-protected installation reading as a protected one.
4. **A declaration must cover the COMPLETE foreign set** → 🚨 **the unit of declaration is now the
   REGISTRY, not the installation.** `instances.json` gains a top-level `registries` table and the
   per-instance `registry` key is **refused**: a per-instance field answers *"is this one out of
   scope"* and can never answer *"is every registry the fleet pins in accounted for"*. Two
   dispositions with a mandatory reason — `fleet-unlockable` (ours, unlockable here) and
   `third-party` (never ours to protect); unknown disposition, missing reason and an undeclared host
   are each RED.
5. **Helm's split `repository:`/`tag:` shape** is handled on the foreign path too — without it a
   foreign-only overlay written that way extracted as *nothing* and fell back into `pins no image at
   all`: this bug in its second shape.
6. 🚨 **The report claimed protection nobody had checked.** The out-of-scope line read *"protected by
   that registry's own retention"*; it now reads **"NOT locked here, and whether anything retains
   that registry is UNVERIFIED"**. A summary claiming otherwise would make a green run read as
   covered and support re-enabling cleanup on a false premise — this mechanism's own failure mode,
   committed by its own report. This one was the most important of the six.

Plus (7) whole-line **comments are stripped** before extraction, on both paths.

### Controls — 23 arms, falsified in two directions

- an **inert classifier** reds six arms;
- **removing comment-stripping and the host comparison** reds four more, including
  `ARM 32: a reference inside a COMMENT was extracted as a pin`.

Re-run over the **actual committed bytes of all twelve overlays** with the fixed code: the comment
match is gone (`ghcr.io` 2 → 1), `memex-cloud`'s 3 ACR + 1 third-party is **not** a blocker, `build`
and `pearl` are out of scope, `memex` is 6 ACR + 0 foreign, and **both hosts are accounted for by
the committed table** — so nothing in the fleet blocks today, and the mixed-registry arm fires
exactly when the migration reaches an installation that still pins here.

Rebased onto `main` now that #4213 has merged; this PR is a single commit pair again.

Verification re-run after the rebase: `lock-pinned-digests.py --self-test` 0 ·
`--check-retention-record .` 0 · `check-pinned-digests.py --self-test` 0 ·
`check-pin-set-consistency.py --self-test` 0 · `check-workflow-shell.py --root .` 0 ·
`check-workflow-yaml-keys.py --root .` 0 · Release `-warnaserror` 0 warnings 0 errors ·
Documentation.Test **447 passed, 0 skipped**.

PR #4235, opened 2026-09-13
fleet process: the five satellites converged — roster rows become real; two rules of the day

rbuergi:
Education#326, Reinsurance#204, SocialMedia#185, Crm#97 and Manufacturing#86 merged on 2026-09-13 (16:10Z–17:20Z). The roster rows core #4201 recorded as `pending:` for their arriving callers are now measured callers, so this PR removes the marker from the ten rows (gate with `actions: read`, resolve-locks) and drops the five superseded gate rows. The ci-failure rows stay `pending:` — those callers have not landed.

`Doc/Architecture/FleetCiProcess.md`: the convergence table is filled in, and two rules measured today are written down — the module-pack receipt must agree with the tests matrix (#4203), and lifting Plugins' `MW_PLATFORM_REF` while main's runs are backed up resolves pull requests to an OLDER set than the freeze (measured 18:01Z: 8484 chosen over sealed 8506), with the twin-port ordering that follows.


rbuergi:
Scope note before this merges: the five satellite PRs landed the **wiring** — the caller, the grant, the expression — and this PR records that. They have not yet been observed to **function**: `Resolve generated-lock conflicts` has failed on every run since 14:58Z on all five (the lane fetches the resolver into `$RUNNER_TEMP`, so `Path(__file__).parent.parent` resolves the repo root to `/home/runner/work` → `fatal: not a git repository`), fixed by #4231. Plugins is unaffected because its copy runs in-tree.

So read this page's convergence table as "every satellite calls the shared lane with the right grant", which is what the roster asserts, and not as "the resolver has resolved a PR anywhere but Plugins". The first green `Resolve generated-lock conflicts` run on a satellite after #4231 is that proof.

Issue #2013, opened 2026-08-21
A blocking bridge in a test wedges the whole shard: methodTimeout cannot abort a thread parked in a native wait


rbuergi:
Found while diagnosing #1991's shard-2 timeout, and generalised because the signature is greppable and has almost certainly bitten before.

## The mechanism

A test that bridges an observable to a blocking call — `.ToEnumerable()`, Rx's `IObservable<T>.Wait()`, `.Result`, `.GetAwaiter().GetResult()` — parks the calling thread on a semaphore until the source produces. When the source schedules onto **that same thread**, it self-deadlocks.

Which thread it lands on depends on the scheduler the operator picks, so **it is intermittent**: in #1991 one full local run passed with the offending test present, and two later runs wedged for **31 and 15 minutes at ~0% CPU**.

## Why it costs a whole shard rather than one test

🚨 **xUnit's `methodTimeout` cannot abort a thread parked in a native wait.** So instead of a 30 s test failure you get an unbounded host wedge:

```
[CI] MeshWeaver.Graph.Test exit=124 TIMEOUT (8m wall-clock cap hit — likely fixture/init hang)
```

— `exit=124`, **no failing test named**, and the marker's own guess ("fixture/init hang") points away from the real cause. That is why the 30 s bound appeared to have been bypassed.

The stack that identifies it:

```
SemaphoreSlim.Wait()
System.Reactive…ObservableImpl.GetEnumerator.MoveNext()
System.Linq.Enumerable.ToArray
<the test>
```

## Why this is not just "the codebase bans .Wait()"

It is the same rule AGENTS.md already states for product code — *`async` IS a deadlock; compose with `.Select`/`.SelectMany`/`.Subscribe`* — but tests have been treated as exempt, and the consequence there is **worse**, not milder: product code that deadlocks fails one request, while a test that deadlocks takes the shard's remaining tests with it and reports a cause that is not the cause.

## Current exposure

- `.ToEnumerable()`: **0** remaining (the #1991 instance is fixed on that branch).
- `.Wait()`: **37 call sites across 10+ test files** — `SystemScopeDoesNotEscapeTest`, `PostgreSqlTransientRetryTest`, `SyncedQueryPgTest`, `ContentChunkSearchTest`, `ChunkNavigationStoreTest`, `OverlayReEvaluationReadTest`, `NodeTypeBakeStatusTest`, `ThreadStreamingIdentityTest`, …

[...]

PR #4237, opened 2026-09-13
fix(install-completeness): a declared count must name the record it was taken over

### LINKED ISSUE #4200: Install-completeness sweep reports ABSENT for a node that exists — two Store nodes created 2h42m before the sweep denied them (distinct from #3659's phantom class)
## What is failing

The boot-time install-completeness sweep reports a node **ABSENT** that `get` returns, and that was created **hours before** the sweep ran. This is the opposite direction from #3659's defect: that one counted files that were never nodes and so invented names to miss; this one **denies a node that exists**.

## The measurements

**The sweep's own line** — `memex-cloud`, pods `memex-portal-deployment-8ff67f8c7-gcf24` (2026-09-12 **09:01:15Z**) and `-vm8d7` (2026-09-12 **09:04:10Z**), folded on incident `Admin/_LogIncident/b03482717d5ba39a`:

```
fail: MeshWeaver.PluginCatalog.InstalledPackageRepairService[0]
      [InstallCompleteness] Store → 'Store': 2 of 201 declared node(s) are ABSENT.
      Missing: [Store/CoverActionDispatch, Store/Plugin/Test/CoverActionIdentityTests].
      Counted over: 201 file(s) declared, 0 of them not node files (README/manifest/content
      assets, or an extension no parser claims) → 201 distinct node path(s) compared.
      The install record says this package is up to date; the mesh disagrees.
```

**Both nodes, read on memex.meshweaver.cloud (2026-09-13, read-only):**

| path | nodeType | version | state | created / last modified |
|---|---|---|---|---|
| `Store/CoverActionDispatch` | `Markdown` | 1 | `Active` | `2026-09-12T06:18:50.865293+00:00` (`createdBy system-security`) |
| `Store/Plugin/Test/CoverActionIdentityTests` | `Code` | 1 | `Active` | `2026-09-12T06:18:45.684602+00:00` |

Both existed **2 h 42 min** before the 09:01:15Z sweep called them absent, both are still there a day later, and both are version 1 — never rewritten, never recycled. The same two names were reported by two different pods three minutes apart, so this is not one replica's bad moment.

## Why this is not #3659's defect

[MeshWeaver#3659](https://github.com/Systemorph/MeshWeaver/issues/3659) is the **phantom** class: `NodePathForFile` counted carry-along files (`Chess/gui/rn/chess.tsx`) as declared nodes, so the sweep missed names no install ever writes. #3685 fixed it by making the declared side and the write side one predicate, and that fix is **running** — these very lines carry its population clause (`201 file(s) declared, 0 of them not node files`), and the phantom names (`Chess/gui/rn/chess`, `Store/README`) are gone from the samples.

So the declared side is now correct and the *observed* side is wrong. Different half of the comparison, different fix.

## What I do NOT know

The read is not a query, so "stale index" is not available as an explanation off the shelf — `InstallCompleteness.Observe` is explicit (`src/MeshWeaver.PluginCatalog/InstallCompleteness.cs:233`):

```csharp
return persistence.ReadMany(declared, options)
    .Select(n => n.Path)
```

with the remark *"never a query (eventually consistent, and a stale negative here would manufacture a shortfall)"*. Three candidates remain, and I cannot separate them from outside:

1. **An access-filtered read.** `ReadMany` runs as whoever the boot sweep runs as. If that identity cannot read these paths, they come back missing and are spelled identically to damage. 🚨 Note the direction: my read is RLS-**filtered** and the sweep's is the system's, so this is the *opposite* of the usual trap — it would mean the sweep sees **less** than I do. Both nodes are `createdBy system-security` in a partition I can plainly read, which makes this the less likely of the three but not excluded.
2. **A read-path cache holding a negative** for a path first written by GitSync rather than by the installer — both nodes landed at 06:18:4x, in the same second-range, which is a GitSync-shaped write, not an install.
3. **A genuine read race** between the batched `ReadMany` and a write that had already committed — which the timestamps argue against (2 h 42 min is not a race window), unless the *storage adapter's* view of the partition was rebuilt later.

## The reading that separates them

At the sweep's own call site, log — for each declared path the comparison marks absent — the identity `ReadMany` ran under, and re-read that one path immediately afterwards **as system** with the cache bypassed:

- re-read returns the node ⇒ **(1)** or **(2)**: compare the two identities; equal identities point at **(2)**, different at **(1)**;
- re-read also returns nothing, while `get` from a client does ⇒ the adapter's view of that partition is the subject, i.e. **(3)** and the trail is the partition's storage, not the sweep;
- re-read returns nothing and `get` also stops returning it ⇒ the node really did go, and the sweep was right at that instant.

One boot sweep with that reading settles it; no further occurrence is needed.

## Impact

Cry-wolf, one level down from the one #3485 was built to end. An unactionable ABSENT is spelled exactly like an actionable one, and here it is worse than a phantom: it names a **real, healthy node**, so the operator's natural response — reinstall the package — is work that cannot change the outcome. It also inflates the `Incomplete` verdict that `CatalogLayoutAreas.InstallOrUpdateCore` consumes.

## Related

- **MeshWeaver#3659** — where this was first observed, as a comment. That issue's own subject (the phantom class) is fixed and running; this is excluded from it.
- **MeshWeaver#3485** — the sweep this is a defect in.
- **Systemorph/MeshWeaver.Plugins#1795** — unrelated to the cause, but it stops this folding onto #3659's incident, which is why it needs its own issue to be visible at all.

rbuergi:
## What this fixes — a count with no provenance

`#4200`'s line reported

```
[InstallCompleteness] Store → 'Store': 2 of 201 declared node(s) are ABSENT.
Missing: [Store/CoverActionDispatch, Store/Plugin/Test/CoverActionIdentityTests].
Counted over: 201 file(s) declared, 0 of them not node files … → 201 distinct node path(s) compared.
```

Reconciling that `201` against the install record took a version-by-version read of `Plugins/Store` **plus** a commit-by-commit count of the package's source folder. The result:

| | |
|---|---|
| record v31, written 2026-09-11T21:30:55Z | 193 files, declares **neither** name |
| record v32, written 2026-09-12T02:00:24Z | **byte-identical map**, declares neither name |
| source snapshot `85ccf1305`, 2026-09-11T21:18:59Z | **201** node-candidate files, containing **exactly** those two |

The two record versions straddle the 09:01:15Z sweep. So the count was right about *something* — a real 201-file population that existed before the sweep — and nothing in the line could say what.

🚨 **That is the same defect as a missing denominator.** A count taken over the current record and one taken over a stale record render identically. And the record is precisely the half that *can* be stale: `InstallCompleteness.Observe` goes to considerable lengths to keep the OBSERVED side off a query —

> *"never a query (eventually consistent, and a stale negative here would manufacture a shortfall)"*

— while the DECLARED side arrives through `InstalledPackageRepairService.InstalledRecords`, an eventually-consistent `GetQuery` whose answer nothing identified.

## The change

- **`InstallCompletenessVerdict.RecordIdentity`** — init-only, default `null`. Additive: the positional constructor is untouched, exactly like `DeclaredFiles`/`NonNodeFiles`/`ModuleSourceFiles` before it.
- **`InstallCompleteness.DescribeRecord(path, version, lastModified, manifest)`** — pure and public, because only the caller that reads the record NODE knows its version.
- **`Observe(…, string? recordIdentity = null)`** stamps it on **every** arm, `NotObserved` included: a failed verification nobody can attribute to a record version is as unactionable as a shortfall nobody can attribute.
- **`InstalledRecords`** selects `version,lastModified` and builds the identity.
- Both verdict log lines print `taken over {Record}`:

```
Counted over: 201 file(s) declared, … , taken over Plugins/Store v32
(written 2026-09-12T02:00:24Z; version 1.9.16, moduleVersion b53d4f05232ce57f,
 installedAtUtc 2026-09-10T12:55:28Z, 201 file(s) in the map)
```

The map's own size sits beside the count it produced, so a record that has gone **backwards** is visible in the line rather than only in an archaeology of its versions.

🚨 And where no caller supplies one, `Provenance` says **"the install record was NOT identified"** rather than rendering a blank that reads like an identified one — the same rule that makes `NotObserved` not a pass.

## What this is NOT

It does not make the declared read fresh, and it **does not fix #4200** — hence `Refs`, not `Fixes`. It makes the next occurrence attributable in one line instead of a day of archaeology, and it removes the ambiguity that made this issue's central number unreadable.

## Verification — both directions

🚨 **The test was watched failing before it was trusted to pass.**

| | |
|---|---|
| `EveryVerdictNamesTheRecordVersionItWasTakenOver` | **1 passed**, `.trx` written, 2 s |
| with `Provenance` sabotaged to return `""` | **1 failed** |
| restored | **1 passed** again |

`Release -warnaserror`, one project per invocation: `MeshWeaver.PluginCatalog`, `Memex.Portal.Shared.Test`, `MeshWeaver.Documentation.Test` — **0 warnings, 0 errors** each. `DocumentationLinkIntegrityTest`: 1 passed. `check-record-signatures.py` (self-test + live): `✓ 3 changed C# file(s): no binary-breaking primary-constructor changes`.

The test awaits the observable (`await … .FirstAsync().Timeout(30.Seconds())`) rather than bridging it to a blocking wait — `BlockingBridgeInTestRatchetGuard` (#2013).

## Related

- **Systemorph/MeshWeaver#4200** — where the unattributable count was measured. The same pass also restored that issue's central claim on a sound argument (the files' only source commit is 2026-09-10T03:41:11Z and neither stored timestamp carries it, so neither is source-derived) and ruled out two of its candidates.
- **#3659** — the population clause this extends, and the same log site.


rbuergi:
All four findings were real and are fixed in `c406b8dc`. Three were watched failing before the fix was trusted.

| finding | verdict | control |
|---|---|---|
| the What's New entry uses the Plugins front-matter convention, not core's | **correct — would have reddened CI** | with `Name:` put back to `title:`, `WhatsNewEntryIntegrityTest` FAILS (1 failed); restored, 2 passed |
| adding an optional parameter replaces the six-parameter metadata signature | **correct** | see below |
| nothing exercises the `InstalledRecords` projection | **correct** | the `select:` is now asserted |
| the new summary orphaned `NotCheckedIsNeverClean`'s | **correct** | the test moved below it |

🚨 **The second is the one that could not have been caught here.** Nothing in this repository compiles against an older core, so a lost metadata signature produces no diagnostic anywhere in CI — it surfaces as `MissingMethodException` in a consumer at run time. The seven-parameter `Observe` now takes `recordIdentity` with **no default**, and a real six-parameter overload forwards `null`, so both signatures stay in the metadata. (`check-record-signatures.py` does not cover this — it reads primary constructors, and reports `✓ 3 changed C# file(s): no binary-breaking primary-constructor changes` either way.)

🚨 **And the third was the difference between an instrument and a comment.** With only `DescribeRecord` exercised on hand-supplied values, dropping `version` or `lastModified` from the listing's `select:` would leave every verdict naming a record it cannot identify, with nothing red — the same "declared count with no provenance" defect this PR exists to close, one level up. The query is now `InstalledPackageRepairService.RecordsQuery` and the test asserts the projection, with `content` as the control: a listing that stops selecting the manifest declares nothing at all.

Release `-warnaserror` on all three touched projects: 0 warnings, 0 errors. Tests: 2 passed (`EveryVerdictNamesTheRecordVersionItWasTakenOver`, `NotCheckedIsNeverClean`) and 2 passed (`WhatsNewEntryIntegrity`, `DocumentationLinkIntegrity`).

rbuergi:
### One more caught locally, and it was a CI red: the timeout ratchet

Running the whole guard/ratchet sweep (not just the tests I touched) turned up a failure the PR would have carried into CI:

```
🚨 Hand-written test timeouts rose to 388 (baseline 386). … Do NOT raise the Baseline.
   Convert the sites you added.
```

The two `.Timeout(30.Seconds())` calls in the new test were the only literals in the file — the rest already used `TestTimeouts.Convergence`. Both converted (`99078811`). A literal 30 s is a guess about one machine, and on the CI runner it also sits one second under the framework's own write bound, so a test that hits it can never report why it gave up (#2819).

**Verification now reads:**

| | |
|---|---|
| `TestTimeoutLiteralRatchetGuard` | 3 passed |
| whole guard/ratchet sweep | **312 passed** (was 311 passed / 1 failed) |
| `InstallCompletenessTest` (whole class, not just the new test) | **15 passed** — the 11 pre-existing ones unaffected |
| `WhatsNewEntryIntegrity` + `DocumentationLinkIntegrity` | 2 passed |
| Release `-warnaserror`, one project per invocation | `MeshWeaver.PluginCatalog`, `Memex.Portal.Shared.Test`, `MeshWeaver.Documentation.Test` — 0 warnings, 0 errors |

Issue #4126, opened 2026-09-12
Registry module bundles cannot carry native runtime assets: both closure derivations drop runtimeTargets, the bundle format has no native section, and the container evaluator refuses a native-only PackageReference (measured: MeshWeaver.AppleMessages needs libe_sqlite3)


rbuergi:
Measured 2026-09-12 while moving `MeshWeaver.AppleMessages` (the iMessage module, MeshWeaver.Plugins) off a system `sqlite3` shell-out onto `Microsoft.Data.Sqlite` — the fleet rule being that **a module brings its own dependencies** (Doc/WhatsNew/2026-09-01-a-module-brings-its-own-dependencies-again), and the maintainer's addition (2026-09-12, verbatim: "must also be module … and then mounted"): the engine the module brings must travel INSIDE its sealed bundle, native runtime assets included, so a runner that mounts the sealed publication from the platform volume gets a working module with no install.

**It cannot, today. The runtime half exists; the bake half does not.** Every reference is `origin/main` at c7262a5e8.

## What works: the runtime resolver (#1728)

`src/MeshWeaver.Mesh.Contract/ModuleNativeAssets.cs:131` probes `modules/<Name>/runtimes/<rid>/native/<lib>` (then the flat folder) from `AssemblyLoadContext.Default.ResolvingUnmanagedDll`, keyed on the REQUESTING assembly's folder; `MeshBuilder.InstallModules` subscribes it. The landing service writes a bundle into `modules/<generation>/` (`src/MeshWeaver.PluginCatalog/ModuleLandingService.cs:907`), whose parent is `modules`, so a `runtimes/` tree laid out there WOULD resolve. The in-image closure lane already keeps that tree: `memex/MeshModulesPublish.targets:465` (`_ClosureRidNative`, kept since #1728).

## What strips it: the registry bundle, on both build paths

1. **SDK path** — `src/MeshWeaver.Plugin.Build/DepsClosure.cs:149` takes only `runtime` entries ending in `.dll`; `:152` DETECTS `runtimeTargets` and `:184` turns them into a WARNING (`'{name}' declares native runtimeTargets the bundle does not carry — if the module needs them at runtime, they must ship another way`). Nothing ships them another way.
[...]

PR #4239, opened 2026-09-13
feat: registry module bundles derive and CARRY native runtime assets (#4126, stage 1)

rbuergi:
## The honest staging answer first

The prior pass verified all eight claims and called this *"a four-component feature, not small"*. It
is. I re-derived the split and it is **four stages, not one**, and the reason is not size — it is that
stages 2–4 each depend on something the stage before it establishes, and stage 4 (`warning → refusal`)
would break a working module if landed early.

**This PR is stage 1: DERIVE and CARRY.** Stages 2–4 are named in the doc and on the issue, with what
each depends on, so the residue is a plan rather than a rediscovery.

## 🚨 What I found that the issue did not state: the warning could not fire for the measured case

The body says `DepsClosure` *"DETECTS `runtimeTargets` and turns them into a WARNING"*. It does — for
a package that ALSO contributes a managed assembly. The loop short-circuited first:

```csharp
if (!nodes.TryGetValue(name, out var node) || node.RuntimeFiles.Count == 0)
    continue;                                    // ← a pure-native package leaves here
files.AddRange(node.RuntimeFiles);
if (node.HasNativeAssets)
    warnings.Add($"'{name}' declares native runtimeTargets the bundle does not carry — …");
```

`SQLitePCLRaw.lib.e_sqlite3` — the package this whole issue was measured on — contributes no managed
assembly at all. It was `continue`d before the native branch ran. **It warned about nothing and
dropped everything, silently.**

**Watched failing on unmodified `main`** (`d1556d7a37`), with a probe written against main's own API:

```
a package whose ONLY contribution is native must not vanish without a word — warnings=0, files=0
```

Zero warnings, zero files. The package vanished. That probe is now the permanent case
`APackageWhoseONLYContributionIsNative_IsCarried_NotSkippedBeforeItIsLookedAt`.

## What stage 1 does

**Derived.** `DepsClosure.Result.Natives` — `NativeAsset(RelativePath, Rid, Package)` from
`runtimeTargets` entries whose `assetType` is `native`. An `init` property, because a fifth
constructor parameter is a binary break.

**Carried.** A third manifest-declared bundle section, `meshweaver/modulenatives/`, with the
module-relative path PRESERVED — the path IS the contract: `ModuleNativeAssets` probes
`<moduleDir>/runtimes/<rid>/native/<lib>` and nothing else.

It could not reuse either existing section, and both reasons are load-bearing:

| section | why not |
|---|---|
| `meshweaver/modules/` | every consumer filters to entries with **no `/` in the remainder** (`ServedModuleBytes.cs:204-206`, `PublishedBundleCatalogue.cs:519-521`) — a native written there is *silently skipped*, not laid out. And `ModulePublish.Validate` refuses a flat name containing `/`, so the other spelling is a declined publish. |
| `meshweaver/moduleassets/` | means static WEB assets; the landing service's logs and docs say wwwroot. Conflating a loadable binary with a served file makes every future rule about one apply to the other. |

**Three things are still dropped — each now NAMED rather than silent:**

| declaration | outcome | why |
|---|---|---|
| `native` at `runtimes/<rid>/native/<file>` | **carried** | the exact layout the loader probes |
| `native` at any other shape | warned | bytes at a path nothing looks at read as shipped and behave as absent |
| `runtime` (RID-specific MANAGED assembly) | warned | the flat closure has one slot per assembly name, no way to choose a RID at pack time |
| `.a` / `.lib` | excluded | link-time inputs, never loaded — the exclusion the in-image lane has applied since #1728 |

## The negative controls — quoted

**Two on the round trip**, so the positive assertion is not reading a constant of every bundle:

```csharp
// ── negative control 1: it is NOT in the flat closure, which is filtered to entries with
// no '/' at every consumer — a native declared there would be silently skipped.
Assert.DoesNotContain(manifest.Module.Assemblies!,
    a => a.Contains('/') || a.EndsWith(".so", StringComparison.Ordinal));
Assert.DoesNotContain(files, f => f.FileName.Contains('/'));
```

```csharp
// ── negative control 2. If the section were written empty rather than omitted, a consumer
// could not tell "this producer ships none" from "this producer predates the section" —
// and the assertion above would be reading a constant of every bundle.
Assert.Null(manifest!.Module!.NativeAssets);
Assert.Empty(BundleReader.ReadModuleNativeAssets(bundle));
```

**One on the derivation**, separating "declares runtimeTargets" from "declares a NATIVE":

```csharp
// The negative control for the case above: `assetType: "runtime"` is NOT a native, so it
// must not sneak into the native section — the flat closure has one slot per assembly name
// and no way to choose a RID. If the derivation keyed on `runtimeTargets` alone rather
// than on the asset TYPE, this would be carried and the assertion fails.
Assert.DoesNotContain(result.Natives, n => n.Package == "RidPicky");
```

## The binary-compat gate caught this, and the fix was the right one

`ModuleRef` first took `NativeAssets` as a 5th optional positional parameter. `check-record-signatures.py`
refused it:

```
✗ record ModuleRef — arity
    primary constructor went from 4 to 5 parameter(s). Adding one — even WITH a default —
    replaces the signature: every assembly compiled against the old arity calls a constructor
    this one no longer has.
```

Converted to an `init` property, the same shape `Manifest.SourceCommit` and
`AssemblyRef.SourceFingerprint` already use. Gate now: `✓ 6 changed C# file(s): no binary-breaking
primary-constructor changes`.

## Stages 2–4 and what each depends on

1. **Landing** — `ModuleLandingService` does not lay the section out under `modules/<generation>/`,
   and `PluginBundleClient.LandFromBundle` does not pass it. `ValidateAssetPath` already accepts the
   shape (`runtimes/linux-x64/native/libX.so` passes) and the write loop already handles trees, so
   this is signatures plus **`GenerationIdOf`, which must fold the natives into the hash** or two
   bundles differing only in natives collide on one generation directory. Depends on nothing but
   itself. 🚨 **Until it lands, a native rides the bundle and is not written to disk** — carried,
   declared, ignored. That is why there is no What's New entry yet: it belongs with stage 2, when a
   module author can actually ship one.
2. **Registry re-serve** — `ModuleBundleSource.CollectVersion` and `ServedModuleBytes.FromSealed`
   collect flat files plus `wwwroot/**`, so a registry re-serving a shelved module drops the section.
   Depends on (1) being the layout it re-serves.
3. **The container lane** — `ContainerReferenceSet.cs:307` reads only the `runtime` section of the
   image's deps.json, and `:360-370` `Resolve` marks a native-only package `Supplied=false`, so a
   `build: container` module cannot even DECLARE the CVE-patched engine (`ProjectBuild.cs:951-958`
   reds it). Independent of (1) and (2) — it is the other closure derivation. Also
   `node-repo-module-pack.yml:2046-2051` (`--with` refuses a path component) and `:2349-2358` (the jq
   provenance gate reads `module.assemblies`), neither of which stage 1 disturbs because the natives
   are a separate manifest section.
4. **Warning → refusal** — only once (1)–(3) are in. Today a module needing a native the bundle
   cannot carry still has the host-supplies-it fallback; refusing now would break it with no route
   forward.

## Verification

```
dotnet build src/MeshWeaver.Plugin.Build -c Release -warnaserror         → 0 warnings, 0 errors
dotnet build src/MeshWeaver.Plugin.Packaging -c Release -warnaserror     → 0 warnings, 0 errors
DepsClosureTest + ModulePackCommandTest                                  → 38/38
BundleReaderTest + ModuleNativeAssetTest                                 → 17/17
MeshWeaver.Documentation.Test                                            → 459/459
python3 scripts/check-record-signatures.py                               → ✓ 6 changed C# files, clean
the silent-drop probe on unmodified main (d1556d7a37)                    → RED: warnings=0, files=0
```

## Conserved

`Doc/Architecture/Modules` → **"The registry bundle: derived and CARRIED (#4126, stage 1)"**, replacing
the gap statement the prior pass wrote there, with the drop table and the four stages.

Refs #4126 · Pairs-with: none — `DepsClosure.Result` and `ModuleRef` GAIN init properties and
`BundleReader` gains a method; nothing leaves `src/`.


rbuergi:
## The red on `e562e142fd` is a test this diff provably cannot reach

Run [34782436418](https://github.com/Systemorph/MeshWeaver/actions/runs/34782436418), shard 1:

```
[CI] MeshWeaver.Data.Test exit=1 TESTFAIL (1 failing test(s) recorded in trx; the host completed normally)
    MeshWeaver.Data.Test.StreamResyncGivesUpTest.AResyncThatNeverConverges_FaultsTheMirror_AndAFreshSubscriberRecovers [FAIL]
```

**Not "it fails elsewhere too" — a REFERENCE-CLOSURE argument, which is the strong form.**
`test/MeshWeaver.Data.Test` references `MeshWeaver.Data`, `MeshWeaver.Mesh.Contract`,
`MeshWeaver.Fixture` and `MeshWeaver.Data.TestDomain`; their own references close over
`MeshWeaver.Data.Contract`, `MeshWeaver.Messaging.Hub`, `MeshWeaver.Domain`,
`MeshWeaver.Messaging.Contract` and `MeshWeaver.Layout`. **`MeshWeaver.Plugin.Build` and
`MeshWeaver.Plugin.Packaging` are in none of them**, and this PR's entire `src/` diff is those two
projects plus one Markdown page:

```
src/MeshWeaver.Documentation/Data/Architecture/Modules.md
src/MeshWeaver.Plugin.Build/DepsClosure.cs
src/MeshWeaver.Plugin.Build/ModulePackCommand.cs
src/MeshWeaver.Plugin.Packaging/BundleReader.cs
src/MeshWeaver.Plugin.Packaging/NuGetPackageWriter.cs
test/MeshWeaver.Graph.Test/…
```

There is no compilation or runtime path from this diff to `StreamResyncGivesUpTest`. It is also not
in `.github/known-flakes.json`, so the steward will not re-queue it on evidence — that catalogue
entry is a decision for whoever owns that test, not something I will add from here to unblock my own
PR.

**I have not re-run it to see.** The push above is a new run caused by a real code change (the four
review findings below), not a retry — and if the same test reds again on the new head I will say so
rather than push again.

PR #4265, opened 2026-09-14
ci: the platform resolver gets a FLOOR, and a ceiling that cannot read its own fixtures

rbuergi:
Routed from **Systemorph/MeshWeaver.Plugins#1826**. Both halves are canonical-resolver work: patching a satellite's vendored copy would have re-forked what #4171/#1565 just paid to unify (Plugins' `scripts/resolve-platform.py` is byte-identical to `.github/scripts/resolve-platform.py` today — 1868 lines, `cmp` clean — and `check-resolver-copy.py` reds on drift from **2026-09-15**).

### Mechanism

Two controls, answering the same kind of question:

* **the ceiling** — `main_passed_ceiling()`, *which set is vouched*;
* **the freeze** — `MW_PLATFORM_REF` → `ceiling_for()`, *which set to take*, and it overrides the ceiling entirely.

Neither says which sets a repository's **source** can still compile against. `grep -n "floor\|minimum\|not go below"` returned nothing.

### What that cost, measured

MeshWeaver.Plugins#1821. `src/Memex.LocalMesh` adopted `MeshWeaver.Mesh.Operations.OperationSentinel`. The adopting PR compiled clean and merged — `MW_PLATFORM_REF` was frozen to `3.0.0-ci.8506` (core `7ee11bc76`), which carries the type. The freeze was lifted, a repo VARIABLE edit with no gate and no diff:

```
set=3.0.0-ci.8492
source=the newest sealed platform set this repo's `main` has passed; 50 newer run(s) passed over
…
src/Memex.LocalMesh/LocalMeshApiEndpoints.cs(240,23):
error CS0103: The name 'OperationSentinel' does not exist in the current context
```

Every open pull request red on a fault none of their diffs could reach.

### The fix

A repository may declare `platform-floor.json` **beside the resolver** naming the oldest core-CD run number its tree compiles against, and *why*. It moves in the diff that needs it, so an adoption and its requirement land together. The refusal names the set, the floor, the reason, and the one way down.

**It binds every path, a freeze included.** A freeze chooses among the sets that can build a tree; it cannot make an older set carry a symbol that set does not have, and resolving below the floor has exactly one outcome.

🚨 **Nothing changes for this repository or any vendoring one until a declaration is written.** Verified with none present:

```
current_floor() with no declaration -> (None, '')
floor_refusal on that floor         -> None
```

🚨 **The path is the resolver's SIBLING, explicitly** — not derived from a repository root. The canonical lives at `.github/scripts/` here and at `scripts/` in every vendoring repo; a root-relative path would have to be right for both (it happened to be, by a coincidence of the constant's shape — a coincidence a future move breaks silently). A sibling is right for both by construction.

**Boundary, deliberate and in the code comment:** the floor guards `main()`, which is every CI resolution. It does not guard `fetch-refs.py`, which calls `choose` directly for local tooling and takes the newest sealed set with no ceiling and no freeze — no mechanism to resolve backwards past an adoption.

### Watched failing first

```
  File ".github/scripts/resolve-platform.py", line 1610, in _floor_case
    got = floor_refusal(floor, set_name, number)
NameError: name 'floor_refusal' is not defined
```

### Mechanism

The ceiling is recovered from the `::notice title=Platform for this run::` each run publishes. In MeshWeaver.Plugins the resolving job **also runs the resolver's self-tests in the same step**, and those reach `emit`, so FIXTURE set ids were published under the production title on the very job `main_passed_ceiling` parses. And it took the **first** match and `break`ed.

Measured on three consecutive green `main` runs (34800174441, 34798629153, 34798196752), each publishing:

```
Platform for this run :: 3.0.0-ci.8539 — core 77451a10b (frozen by the repo VARIABLE …)  ← the verdict
Platform for this run :: 3.0.0-ci.8207 — core aaaaaaaaa (the newest sealed platform set) ← a fixture
Platform for this run :: 3.0.0-ci.8207 — core aaaaaaaaa (the newest sealed platform set) ← a fixture
```

`core aaaaaaaaa` is `A = 'a' * 40` in the test module. It answered correctly only because the verdict happened to sort first — which the API does not promise and which is **not emission order**: the fixtures are logged about **seven seconds earlier**.

### The fix — a refusal, not a better guess

Collect **every** annotation under the production title; if they name two different sets, the run's own verdict cannot be told from a fixture, so the **run is skipped** with both numbers in the note. `best` is a max over the other runs examined, so one poisoned run costs a data point, never a wrong ceiling.

### Watched failing first

```
✗ resolve-platform self-test: 3 failure(s)
  - a FIXTURE set id before the verdict must NOT be taken (#1826): ceiling 8207, expected None
  - …nor after it — the reader must not depend on annotation order either: ceiling 8539, expected None
  - …and the note names BOTH sets, so the poisoning is readable from the log: ceiling 8207, expected None
```

`ceiling 8207` is the fabricated id taken as production data; the second line is the order-dependence in the other direction. (The **emitter** half — the satellite's own test module no longer publishing under the production title — landed separately in Plugins, and needs no canonical change.)

## Counts, and what was not widened

`resolve-platform.py --self-test`: **70 → 88 cases**, run against this repository's canonical. Nothing the resolver accepts is widened — both changes are **refusals**: an ambiguous run it previously guessed at, and sets it previously took. `MeshWeaver.Documentation.Test`: **459 passed**, 0 failed, `Release -warnaserror` clean.

## 🚨 The general lesson, and why it is documented here

Hours after the above, `MW_PLATFORM_REF` was moved to `3.0.0-ci.8547` — the *newest sealed set* — and reddened Plugins `main` plus every open pull request on the first job of every run:

```
the freeze names main-cd #8547 (core 1c1d62adf), but its source/release is unverified:
platform-bake job 103845949792 has an incomplete or inconsistent final publication receipt.
```

That job reported `completed success` and emitted one well-formed receipt carrying **`targets-published=0`** — sealed, and published to no target at all. `--verify-source` refusing was correct.

So **"pin the newest sealed set" is necessary and not sufficient**; the set must also VERIFY its source, and only one of those is visible from a set's name or its run's colour. `CiContentBake.md` now says to resolve a candidate with `--verify-source` *before* setting a freeze. The floor does not help there — a floor is a lower bound and `8547 > 8506` passes it — which is precisely why both are written down.

Work product in `src/MeshWeaver.Documentation/Data/Architecture/CiContentBake.md`.


rbuergi:
**Blast radius on the vendored copies — measured, and it is gentler than it looks.**

`check-resolver-copy.py` runs in each **satellite's** CI (`node-repo-validate.yml`), comparing that repository's `scripts/resolve-platform.py` against this repository's canonical. So changing the canonical is what puts a copy into drift, and the guard's `RED_FROM` is **2026-09-15T00:00:00Z** — about a day away. Worth being precise about when that bites.

It does **not** bite on merge. The lane fetches the canonical at the caller's own pin, not at `main`:

```yaml
SCRIPTS_REF: ${{ inputs.scripts-ref || inputs.platform-ref }}
…
gh api "repos/Systemorph/MeshWeaver/contents/.github/scripts/resolve-platform.py?ref=${SCRIPTS_REF}"
```

So every satellite keeps comparing against the canonical **as of the ref it pins**, and its copy stays code-identical until it moves that pin past this commit. The drift arrives per repository, at a moment that repository chooses — which means **the re-copy can travel in the same change as the pin move**, one repo at a time, and this merge creates no fleet-wide deadline of its own.

Two consequences worth stating plainly:

1. **A satellite that moves its pin past this commit without re-copying will be advisory-then-red** — advisory only until 2026-09-15T00:00:00Z, error after. Pin move and re-copy belong together from here on, which is the same coupling #1565 documented.
2. **Nothing needs to be co-ordinated before merging this.** A satellite on an older pin is unaffected and stays green.

The six copies were byte-identical to the canonical before this change (measured today: MeshWeaver.Plugins `cmp`-clean at 1868 lines), so a re-copy is a verbatim overwrite with no merge to reason about — which is exactly the state #1565 worked to reach and the reason this change belongs here rather than in a fork.

rbuergi:
**Live evidence from production, 2026-09-14 ~04:37Z: the ambiguity this PR refuses is not hypothetical — it is happening now, and it produces a `(set, sha)` pair that never existed.**

Measured by the session that owns `MW_PLATFORM_REF`, on an **unfrozen** resolve: one invocation wrote **both** pairs into a single `GITHUB_OUTPUT` file —

```
sha=1c1d62adf…  set=3.0.0-ci.8548
sha=f41ac5e7d…  set=3.0.0-ci.8548
```

Same set name, two different core shas, from one invocation. `1c1d62adf` is **8547's** core commit, so the unfrozen path paired **8548's name with 8547's commit**. A consumer taking the first match gets a pair that corresponds to no real sealed set — it would check out one commit while believing it had the other's bytes.

That is precisely the disagreement this PR's ceiling change refuses rather than guesses at, and it means #4265 is not only a hardening: the behaviour it replaces is **presently wrong in production**. Until it merges, an incident freeze is what stands in front of it — which is a poor instrument to be relying on, since a freeze also has to satisfy two independent conditions (the set must be the newest seal so `/opt/platform` carries it, AND it must verify its source — `3.0.0-ci.8547` satisfied the first and failed the second).

Recorded here so the reasoning survives the merge rather than living in a chat thread.

Issue #4259, opened 2026-09-14
8 of 19 Hosting NodeTypes are at compilationStatus:Error on memex-cloud — three declared source nodes never landed and the install logged nothing


rbuergi:
## 8 of 19 `Hosting/*` NodeTypes are dead on memex.meshweaver.cloud, and the install that killed them logged nothing

Measured live **2026-09-14**, re-run immediately before filing (a recompile wave can be live, so this is a snapshot of a moving target).

```
search 'nodeType:NodeType content.compilationStatus:Error'   → 8, truncated: false
search 'namespace:Hosting scope:children nodeType:NodeType'  → 19, truncated: false
```

**8 of 19**, all `Hosting/*`: `Backup`, `Deployment`, `FleetConsole`, `InstanceAction`, `InstanceRequest`, `Issue`, `PlatformBuildInbox`, `TriageStatus`. Every one `lastModified` 2026-09-13 **21:56:06–22:03:01Z**.

That is the Hosting control plane — the Fleet Console, the Deployment record, `InstanceAction` (the node every `Roll` / `Restart` / `Sample` / `Logs` operation is written as). On this portal it is a consumer copy, not the control instance; **memex.systemorph.com's Hosting types are `Ok`** (`Hosting/Backup` v668 etc., and its own `nodeType:NodeType content.compilationStatus:Error` sweep returns 0 of 147 readable). So the fleet's operations are not down — but this instance's Hosting UI renders empty and its watchers cannot activate.

## Root cause: declared source nodes that never arrived

`get @Hosting/TriageStatus`:

```
compilationStatus: Error
compilationError:
  MISSING SOURCES: 2 of 2 declared source queries for 'Hosting/TriageStatus' matched NO nodes
    on this mesh ('namespace:Source scope:subtree',
                  'shared=@Hosting/Deployment/Source/TriageIntake')
  CS0246 (line 58): 'TriageStatusContent' could not be found
  CS1061 (line 58): 'LayoutDefinition' … no 'AddTriageStatusLayoutAreas'
  CS0103 (line 58): 'TriageStatusTestsArea' does not exist
  Matched Code nodes (0): (none)
currentSourceVersions:     {}
currentSourceFingerprint:  e3b0c44298fc1c14      ← SHA-256 of the EMPTY STRING
failedBuildInputs:         fw=sd608997…;mod=66780795…;src=0
[...]

Issue #4174, opened 2026-09-13
Two sample NodeTypes fail on every boot of memex-cloud because of the samples MIRROR, not their source — a pre-rename Type/article leftover and Product's unrebased shared=@Northwind refs


rbuergi:
## Two NodeTypes fail to compile on every boot of memex.meshweaver.cloud — neither is an author error

Both sit under `MeshWeaver/samples/Graph/Data/…` — this repository's `samples/Graph/Data` tree mirrored into a `MeshWeaver` partition on that portal — and both fail for a reason created by the mirroring, not by their source. They share the log fingerprint `c0b1424c7beb28e0` with [#3883](https://github.com/Systemorph/MeshWeaver/issues/3883) (whose normaliser masks the node path), so they have been arriving on that thread as if they were recurrences of a defect they have nothing to do with. **They are excluded from #3883's denominator**; this issue is where they live.

All readings below are `get` on memex.meshweaver.cloud, 2026-09-13 ~08:1xZ, read-only.

---

## 1. `MeshWeaver/samples/Graph/Data/Type/article` — a pre-rename leftover the sync never retired

Both spellings exist on the portal, and the comparison is the whole finding:

| node | created | `compilationStatus` | evidence |
|---|---|---|---|
| `…/Type/article` (lowercase) | **2026-08-12T04:49:24Z** | **`Error`** | `failedSourceQueries: ["namespace:Source scope:subtree"]`, `failedBuildInputs: …;src=0`, `CS0246 'Article'`, *Matched Code nodes (0)* |
| `…/Type/Article` (capital) | **2026-08-13T20:11:28Z** | **`Ok`** | `compiledSources: { …/Type/Article/Source/Article: 1 }`, `latestReleasePath: …/Type/Article/Release/20260908182043-2zKzh56V`, `latestAssemblyPath: MeshWeaver_samples_Graph_Data_Type_Article/v272-….dll` |

2026-08-13 is the day core's `e1ecd8e61` renamed the file, and its message names this exact failure:

> `Type/article.json -> Type/Article.json`. Here the id ("Article") was already relative and the FILENAME was the wrong half: the NodeType imported as `Type/article` while its sources import as `Type/Article/Source/Article.cs`. With no explicit Sources, the default query is `namespace:{self}/Source`, i.e. `namespace:Type/article/Source` — which matched 0 Code nodes, so the type compiled against nothing and gated portal readiness.

[...]

Issue #3485, opened 2026-09-06
A shipped package is missing a source node on memex.systemorph.com since 2026-08-26 and a reinstall did not restore it — masked 11 days by an AdoptedUnverified assembly, and the proximate cause of #3472


rbuergi:
## A shipped package is missing a source node in production, and a reinstall did not restore it

**`memex.systemorph.com`'s `Feedback` install is missing `Feedback/Feedback/Source/FeedbackContent`.**

- Absent since **2026-08-26**.
- A **reinstall on 2026-09-03 did not restore it**.
- The compile record shows **4 matched Code nodes against the bake's 5** (`src=4`).
- Masked for **eleven days** by an `AdoptedUnverified` assembly — the type kept working on bytes
  adopted from a publication, so nothing surfaced.

Established while implementing #3479 (roll selection), by comparing the instance's install against
what CD run 7926 actually published. **7926 shipped `Feedback` correctly** — its bake+seal logged
`ok Feedback/Feedback [5 source(s)…]`, `published: …/sc273ee39…/plugins/Feedback.zip`,
`sealed: …/_complete (36 bundle(s))` on both storage targets. So this is not a publication defect.

## Why this matters more than one missing node

🚨 **This is the proximate cause of tonight's outage trigger.** The roll to `3.0.0-ci.7926` was
refused by the bake gate with *"1 NodeType(s) regressed on this image: Feedback/Feedback"* — and the
regression was **not in the image**. It was this instance's incomplete copy failing to build against
it, while the previous image's adopted assembly had been hiding the gap.

The refused-but-still-running pod then stamped module generations alongside the healthy replicas
(#3478), two images served one mesh for two hours, and `Crm/Offer` and `Crm/Opportunity` went
unloadable — every deal and offer page on the client portal dead for ~2.5 h (#3472).

**So a single missing source node, invisible for eleven days, produced a client-facing outage through
three intermediate mechanisms.** None of them would have fired had the install been complete.

## The part that is a defect in its own right

**A reinstall did not restore a missing node.** That is the load-bearing finding here, and it is worth
more than the specific file: whatever `Feedback` is missing today, some other package can be missing
tomorrow, and the remedy an operator would reach for first has already been shown not to work.

Questions worth answering with evidence, not assumption:

- **Why did the 2026-09-03 reinstall not restore it?** An unchanged-check comparing something other
  than node presence (a version, a fingerprint, a manifest hash) would skip a node that is absent —
[...]

PR #4277, opened 2026-09-14
fix(#4259): an UPDATE restores a declared node that is absent from the mesh

rbuergi:
## An UPDATE never looked at the mesh — so a lost node survived every update, each reporting success

Closes the last open half of #4259.

### The defect

`CatalogLayoutAreas.InstallOrUpdate` has three exits. Two of them observe the mesh; the third did not.

| exit | taken when | looked at the mesh? |
|---|---|---|
| skip / heal | module hash EQUAL | **yes** — #3485's `InstallCompleteness.Observe` + `SkipOrHeal` |
| full install | no record, or no file map | **yes** — `DecideAndWrite` writes whenever `current is null` |
| **incremental update** | module hash MOVED **and** the record has a file map | **no** — fetched exactly `newManifest.DiffFrom(record.InstalledFiles)` |

The incremental fetch set was a comparison of two *declarations*: the candidate's `manifest.lock`
against the record the installer itself stamped. A node lost **after** a previous install has an
unchanged file hash, so it never entered `delta.AddedOrChangedFiles`, was never fetched, and never
reached `DecideAndWrite` — the one place that would have restored it. The presence-awareness sat one
layer below a set the absent file could not enter.

🚨 **And it is the exit an update always takes.** #3485's heal lives on the module-hash-EQUAL exit,
which an *update* never takes by definition — and every new publication moves the hash again. So a
package that loses a node stays short across arbitrarily many updates, each one green.

### The measured case — `Plugins/Hosting` on memex.meshweaver.cloud

Read 2026-09-14T06:0xZ, read-only (`get` / `search`); portal at
`3.0.0+c84c6c05503228860df03c4a8b596e497e6d218c`.

**The files were there and they compiled.** `Hosting/TriageStatus/Release/20260913081235-bylQeIj_`
(created on *this* portal 2026-09-13T08:12:35.446Z, `status: Succeeded`) names its inputs in
`sourceVersions`: `Hosting/Deployment/Source/TriageIntake`,
`Hosting/TriageStatus/Source/TriageStatusContent`, `…/TriageStatusLayoutAreas` (ticks =
2026-09-13T08:12:26Z), plus two `Test/` nodes. A release records the source nodes a successful
compile actually read.

**The update wrote only what had changed.** Record stamped 2026-09-13T22:03:03.221Z,
`installedNodeCount: 224`. Exactly two nodes carry that install's timestamps:

| node | version | written |
|---|---|---|
| `Hosting/Deployment/Source/AksOpsResult` | 1 (new file) | 22:02:54.601Z |
| `Hosting/Deployment/Source/PlatformBuildInboxWatcher` | 16 (changed file) | 22:02:57.255Z |

**Eleven declared files have no node.** Anchored probes, each with a same-anchor positive control:

| probe | result |
|---|---|
| `namespace:Hosting/Deployment/Source scope:subtree nodeType:Code` | **14**, `truncated: false` — no `TriageIntake`; the 14 are the control |
| `namespace:Hosting/Deployment/Test scope:subtree nodeType:Code` | **3** — no `TriageIntakeTests` |
| `namespace:Hosting/TriageStatus scope:subtree` | **1** — the Release node alone |
| `namespace:Hosting/TriageItem scope:subtree` | **1** — the Release node alone |
| `namespace:Hosting scope:children nodeType:NodeType` | **19**, `truncated: false` — the denominator |
| `namespace:Hosting scope:children nodeType:NodeType content.compilationStatus:Error` | **8 of 19** |

All eleven are in the record's own `installedFiles` map. Control instance, same probes:
**0 of 19** at Error, and its `Hosting/Deployment/Source` holds **15** including `TriageIntake` —
same package, same declaration, one instance short.

### The fix

`IncrementalUpdate` asks the mesh before it fetches:

1. Derive the candidate manifest's declared node paths through `PackageInstaller.NodePathForFile` —
   the installer's own file→node rule, not a second implementation of part of it.
2. `InstallCompleteness.ObservePresent` — **one** batched `IStorageAdapter.ReadMany` over that
   bounded, known set. Never a query (a stale negative would re-fetch a present file) and never N
   point reads of possibly-absent paths (which opens a storm breaker on the owning hub).
3. `InstallCompleteness.FilesToRestore` — **pure**, so every arm is pinnable offline — returns the
   declared files the delta is not already fetching whose node is absent.
4. Fetch `delta.AddedOrChangedFiles ∪ restore`.

One batched read per incremental update — the same read the post-install verification already pays
on this exit, moved to where it can still change the outcome instead of only reporting it.

**Three answers, never two.** `ObservePresent` returns `null` when the mesh was NOT read (no
adapter, or a faulted batched read) and never an empty set — an empty set would say "the mesh holds
none of them", which on a Postgres read that faulted (`42P01`, not "absent") would re-fetch every
file the package ships, every time. That `null` yields an empty restore set and the caller says so
**at Warning**, spelled so it cannot be read as a pass.

**A restored SHARED `Source/`/`Test/` file falls back to a full install**, by the same rule the
existing changed-file guard applies and for the same reason: those are compile inputs for every type
in the package, so the delta's owner-derivation would recompile only the nominal owner.

**And the other silent hole in the same method.** `IPackageSource.FetchPackageFiles(…, paths)`
filters, and its contract is explicit that *"paths absent from the package simply don't appear in the
result"* — so a requested path the source does not serve vanished with no exception and no line,
after which the record was stamped with the full declared map anyway. The returned set is now
compared against the requested one and the shortfall named at Error. It reports only; it never fails
the update.

### Verification

- **Deterministic repro, and it is not vacuous.**
  `AnIncrementalUpdateOverAMissingNode_RestoresIt` installs a package, deletes one node, then runs an
  update in which the module hash moves and the deleted node's own file hash does **not**.
  - reverting only the `CatalogLayoutAreas.cs` change and rebuilding: **FAILED in 2 m** (the
    `WaitForNode(present: true)` timeout — the node stays gone)
  - with the change: **PASSED in 3 s**
- `Memex.Portal.Shared.Test.InstallCompletenessTest` — **21 of 21 passed**, 27 s. The class held 18
  `[Fact]`s at the branch base and holds 21 here: the behavioural repro plus two arms pinning
  `FilesToRestore` (every branch, offline) and `ObservePresent` (live read, absence, and the
  no-adapter `null` that must never be an empty set).
- `MeshWeaver.Documentation.Test` link integrity / topic map / conflict markers — **6 of 6 passed**.
- `dotnet build -c Release -warnaserror`, one project per invocation: `MeshWeaver.PluginCatalog`
  **0 Error(s) 0 Warning(s)**, `MeshWeaver.Documentation` **0/0**, `Memex.Portal.Shared.Test` **0/0**.

### Work products conserved

- **[Declared Is Not Landed](src/MeshWeaver.Documentation/Data/Architecture/DeclaredIsNotLanded.md)** —
  the three exits, the measured case, the batched presence read, and the **sweep hazard**:
  `Hosting/TriageItem` reads `compilationStatus: Ok` while all six of its `compiledSources` are
  absent, because it has not been re-driven since its last successful compile. So
  `content.compilationStatus:Error` is a sweep for types that *attempted* a compile and failed —
  "8 of 19 report Error" understates the damage, and the cross-check that sees it is
  `compiledSources` (or a release node's `sourceVersions`) against the mesh.
- **NodeTypeCompilation.md** gains the `shared=@…` decision that #4174 deliberately left open:
  `@` is **absolute and never rebases** (`CodeQueryResolver.Expand`), because the dominant use is a
  genuinely cross-partition include; the consequence is that a tree carrying an absolute
  cross-reference is not importable under a prefix, and the remedy is at the install boundary
  (install the package at the root its sources were authored for), not in the resolver.
- Topic-map entry + a `Category: Fix` What's New entry.

### What this does NOT do

- It does not explain **why** the nodes went missing between 08:12:35Z and 21:56:07Z. The release
  node proves they were written and compiled; no instrument a session can run names what removed
  them. This change makes the loss self-correcting on the next update instead of permanent.
- It does not re-arm a **PARKED** NodeType — that is #4208. Restoring content and recycling the type
  stay two steps on an image without it.

Refs #4259, #4257, #3485, #4208, #4174


rbuergi:
## Held as draft — author terminated, CI red, three review findings standing

I have converted this to **draft**, which disabled auto-merge (`auto_merge: false` verified after the conversion). Draft alone is the correct hold here because the PR was **not** admitted to the queue — `mergeable_state: blocked` on failing checks, so there was no queue entry to retract. (Had it been admitted, draft would not have held it: an admitted entry merges regardless of draft state, and the hold would have been GraphQL `dequeuePullRequest` **first**, then draft.)

**Why it is held.** This PR was armed with `--auto` and carries three unaddressed Copilot findings, one of which is a correctness defect. That combination — a change set set to merge on checks alone while a valid review finding stands — is precisely the gap #4269 exercised. Its author was terminated mid-work by an API rate limit that resets **2026-09-17 13:00 Europe/Zurich**, so nobody was going to answer them.

### 1. CI red — one failing test, and it is a rule this repo enforces deliberately

`MeshWeaver.Documentation.Test.TestTimeoutLiteralRatchetGuard.TheHandWrittenTimeoutCountOnlyEverGoesDown`
(shard 5, run [34813326232](https://github.com/Systemorph/MeshWeaver/actions/runs/34813326232); 1 ❌ of 4 022, classified `TESTFAIL` — ordinary failing tests, not a host death).

This diff added hand-written `TimeSpan` literal(s) in tests. The ratchet **only goes down**: the fix is `TestTimeouts.*`, never raising the guard's count. Small and mechanical, but it must be done rather than worked around.

### 2. Three Copilot findings, none answered

- 🚨 **Correctness — the install baseline advances through a fetch shortfall.** `InstallNodeRepoDelta` unconditionally stamps `newManifest` as the next `InstalledFiles` baseline. If a requested changed file is omitted while its old node is still present, `VerifyLanded` sees the node present and reports completeness — so the record claims the new hash while the old content remains, and *the next same-module update skips it permanently*. That is durable stale content, not a transient miss. Do not advance the baseline when `missing` is non-empty.
- **Cost** — `Verified(...)` still calls `VerifyLanded` after `InstallNodeRepoDelta`, so every incremental update now performs **two** full-manifest `ReadMany` probes, not the one the PR description claims. Either account for that explicitly or redesign so the second read is not needed.
- **A synchronous throw bypasses the `Catch`** — `ReadMany` is invoked before the Rx chain exists, so an adapter that throws synchronously skips the documented `null`/Warning path and falls back to a full package install. `Observable.Defer` gives synchronous and asynchronous failures the same conservative outcome.

The first one is worth restating: it is the same shape as the defect this PR exists to fix (#4259 — *"three declared source nodes never landed and the install logged nothing"*). Shipping it as-is would replace a silent miss with a silent miss that is also durable.

### Resuming

The branch is `fix/4259-incremental-repair`, pushed and intact. Work is: fix the timeout literal, answer the three findings, undraft. **Do not undraft it before the findings are answered** — undrafting re-arms `auto-arm.yml`, and the checks will pass long before a reviewer does.

---

rbuergi:
## All three findings answered, ratchet fixed, undrafting

Picked this up after its author was terminated by the rate limit. `2c3e46a4cb`.

### 1. The install baseline advancing through a fetch shortfall — **fixed, and you were right**

This was the real one. `ReportFetchShortfall` only logged, and `InstallNodeRepoDelta` then stamped `newManifest` as the next `InstalledFiles` baseline regardless. With the old node still present — the ordinary case — `VerifyLanded` sees a present node and reports completeness, so the record claims the new hash over old content and **every later update of that module diffs clean and skips it permanently**.

It is now `EnsureFetchComplete` and it **throws** `"full install required"` — the same mechanism the absent-shared-source arm a few lines above already uses, so the caller falls back to a full install and the user still gets the package.

I also corrected the XML doc, which argued the opposite: *"Reports only; it never fails the update. The install that did run is a fact, and collapsing 'landed short' into 'failed' would make a working update a new way to break."* That reasoning does not survive what happens next — the failure was never *"landed short"*, it was *"landed short and then lied about it"*, and the fallback means nobody loses an update.

Second defect found while there: the shortfall was computed only `if (logger is not null)`, so the guard disappeared exactly where diagnostics are off. Detection no longer depends on anyone listening.

### 2. `ReadMany` bypassing the `Catch` on a synchronous throw — **fixed**

Wrapped in `Observable.Defer`, with the reason in a comment so it is not "tidied" away later: it is a method call evaluated before the chain exists, so a synchronous throw escaped the `null`/Warning path and forced a full re-fetch on every such fault.

### 3. The second full-manifest probe — **accounted for explicitly, not removed**

Taking the first of the two options you offered. The reads are not redundant and neither can go: the preflight asks *which declared nodes are ABSENT so the delta must re-fetch them*, **before** the write — that is the whole of #4259 — and `VerifyLanded` asks *did what we just wrote land*, **after** it. Collapsing them gives either a re-fetch decision made on a pre-install verdict, or a lost node discovered only on the next update. The added cost is one `ReadMany` over `declaredNodePaths`, bounded by the manifest rather than by mesh size, and paid only on the incremental path. That is now stated at the call site.

### The CI red

`TestTimeoutLiteralRatchetGuard.TheHandWrittenTimeoutCountOnlyEverGoesDown` — this branch added one `30.Seconds()`, now `TestTimeouts.Convergence`. The guard is deliberately narrow (only the copied 30 s convention), so the reasoned 60/120/180 s bounds in this file are untouched. Guard passes 3/3.

### Verification

`EnsureFetchComplete` is `internal` (`InternalsVisibleTo` already covers `Memex.Portal.Shared.Test`) so the guard is reachable at all — three tests pin throw-on-shortfall, no-throw-when-complete, and detection-with-no-logger.

**Negative control run, not assumed:** with the throw disabled, **2 of 3 fail** and the no-throw control still passes; restored, **3/3 green**. Builds: `MeshWeaver.PluginCatalog` and `Memex.Portal.Shared.Test`, Release `-warnaserror`, `0 Error(s)` each.

---

PR #4282, opened 2026-09-14
fix(#4281): the gate waits for a platform set that is arriving or not yet reached, instead of refusing with the wrong remedy

### LINKED ISSUE #4281: The gate's platform-volume refusal names two causes and the common one is a third: the set is ARRIVING — and its prescribed remedy (bump the caller's digest) is wrong for it
`resolve-gate-platform.sh` refuses when the CI platform volume carries no set for the tester digest a shard pinned. Its message names **two** possible reasons and prescribes one remedy:

> …so this pin is either older than those three or has never been sealed. Fix it by bumping the caller's image-digest/platform-image-digest to a set the refresh has installed…

There is a **third** reason, it is the common one, and the prescribed remedy is wrong for it.

## The third case: the set is ARRIVING

`ci-platform-refresh` extracts into `<digest>.tmp.<pod>` and only then makes the set complete. A shard that looks while that extraction is in flight sees no completed set and refuses — while the digest it wants is **sitting right there in the listing it prints**.

Measured on Education, 2026-09-14 05:42Z ([run 34809613064](https://github.com/Systemorph/MeshWeaver.Education/actions/runs/34809613064), job `103870839607`). The shard wanted `sha256:cda259b4…`, and the volume listing inside that same refusal contained:

```
sha256-cda259b436636a4464525bbc80ae8f8f1c6f47dace311947fedb592f420a70e1.tmp.ci-platform-refresh-29822740-7x7xb
```

The exact digest, mid-write by refresh pod `29822740-7x7xb`. Not older than the three retained. Not unsealed. **Arriving.**

## And a fourth: the caller is AHEAD of the volume

Measured on Manufacturing [PR #89](https://github.com/Systemorph/MeshWeaver.Manufacturing/pull/89), ~06:5xZ: the shard wanted `sha256:0860c392…` while the volume held `4c327117`, `cda259b4`, `ce95ac38`. By then `cda259b4` had finished installing — so the volume had advanced and the wanted digest had advanced past it.

This is structural, not bad luck: the refresh is a **CronJob on a 10-minute schedule**, while a satellite's gate run is triggered by the **seal**. A run is therefore ahead of the volume *by design*, for up to one period plus one install.

## Why the wrong message is expensive

In both cases the caller's digest is **right and merely early**, and the message tells the reader to change it. Following that advice pins a gate to an older platform to work around a wait — the opposite of what anyone wants, and invisible afterwards.

This is the failure shape the fleet keeps paying for: a message that describes an ABSENCE in the vocabulary of a different condition. The same defect in the catalog path is `MeshWeaver.Plugins#1802`, where the wrong sentence produced the wrong remedy **twice** on one issue.

Live cost on 2026-09-14: Education `main` red all morning on this; Reinsurance `main` not baked since 2026-09-13 17:13Z; Manufacturing PR #89 red. Fleet-side analysis and the refresh's own half are on **Systemorph/Memex#343** (whose title premise — "carries ONE set" — is also stale: the refresh keeps the **3 newest**).

## Direction

Distinguish the cases and treat them differently, because two of them resolve themselves and one never will:

- **ARRIVING** — `<SET_DIR>.tmp.<pod>` exists, or `<SET_DIR>` exists without its completion marker. Wait on the **actual condition** (the marker appearing) under a bounded deadline. Not a sleep: the share is SMB, so there is no inotify and `actimeo` caches a negative dentry, which forces a poll — but the poll's exit is the condition, never the clock.
- **AHEAD** — the caller's set has a higher core-CD run number than the newest installed. Same bounded wait.
- **GONE** — older than the oldest kept, or overtaken (the refresh only ever installs the NEWEST sealed set, so a set that was skipped will never arrive). **Red immediately** — waiting could never end.

Distinguishing AHEAD from GONE needs the set NAME, not just the digest; without it only ARRIVING is detectable and the other two must stay red at once.

🚨 **Do not add a `docker pull` fallback.** The current script is right to refuse one, and its reason should survive any fix: a gate that quietly reached the registry would hide a dead refresh job for as long as the registry answers.

---

rbuergi:
Closes #4281.

## The defect

`resolve-gate-platform.sh` refuses when `/opt/platform` carries no completed set for the tester digest a shard pinned, and its message named **two** causes — *"older than the three kept, or never sealed"* — prescribing one remedy: bump the caller's digest.

Two refusals measured 2026-09-14 were **neither**, and for both that remedy was wrong because the digest was right and merely **early**:

| repo | wanted | what the volume actually showed | case |
|---|---|---|---|
| Education, [run 34809613064](https://github.com/Systemorph/MeshWeaver.Education/actions/runs/34809613064) 05:42Z | `sha256:cda259b4…` | the listing printed **inside that refusal** contained `sha256-cda259b4….tmp.ci-platform-refresh-29822740-7x7xb` | **ARRIVING** |
| Manufacturing [PR#89](https://github.com/Systemorph/MeshWeaver.Manufacturing/pull/89) ~06:5xZ | `sha256:0860c392…` | held `4c327117`, `cda259b4` (by then complete), `ce95ac38` | **AHEAD** |

**AHEAD is structural, not luck.** The refresh is a CronJob on a 10-minute schedule; a satellite's gate run is triggered by the **seal**. A run is therefore ahead of the volume *by design*, for up to one period plus one install.

Live cost: Education `main` red all morning, Reinsurance not baked since 2026-09-13 17:13Z, Manufacturing PR#89 red.

## The change

An absent set becomes three cases with different remedies:

- **ARRIVING** — `<SET_DIR>.tmp.<pod>` exists, or the set dir exists without its completion marker.
- **AHEAD** — the caller's set has a higher core-CD run number than the newest installed.
- **GONE** — older than the oldest kept, or overtaken (the refresh only ever installs the *newest* sealed set, so a skipped one will never arrive).

The first two **wait on the actual condition** — the completion marker appearing — under a bounded deadline. GONE is **red at once**, because that wait could never end. The wait polls because the share is SMB (no inotify, and `actimeo` caches a negative dentry), but its exit is the condition, never the clock.

**No `docker pull` fallback**, and the existing refusal to have one keeps its reason: a gate that quietly reached the registry would hide a dead refresh job for as long as the registry answers.

## 🚨 Two defects in the draft, both caught before this ran

This change was drafted by a session that was cut off mid-work; I finished it and found two things in it that would have shipped:

1. **The caller read `steps.platform.outputs.set`.** That step prints `platform-set=` — so the value read **empty**, and an empty `--set` degrades *silently*: an in-flight install is still visible, but "not reached yet" and "moved past" become indistinguishable, so **AHEAD would have stayed red at once** and the wait would never have fired for the one case that is structural. The fix would have looked applied and done half of nothing. Now `platform-set`, with the trap named at the call site.
2. **The header cited a census of 8 refusals across six satellites** that I could not reproduce. Replaced with the two refusals actually measured, each with its run id, so the comment claims exactly what was checked.

## Verification

- `resolve-gate-platform.sh --self-test`: **21 cases** — 1 container, 2 volume, 8 refusals, 3 usage refusals, **7 wait**. Every refusal case passes `--wait-seconds 0`, so a refusal assertion cannot pass by timing out instead of refusing.
- `check-workflow-timeouts.py`: 84 jobs checked, 0 violations, cap 45 min.
- YAML parses.

## Scope note

`node-repo-gate.yml` is the shared reusable lane every satellite calls `@main`, so this reaches all six on merge. That is the intent — it is what unblocks their gates — but it is also why the self-test carries the wait cases rather than leaving the new branch to be exercised first in production.

The other half of this — retention and the non-atomic install on the writer's side — is **Systemorph/Memex#343**, whose own title premise ("carries ONE set") is stale: the refresh keeps the **3 newest**.

---


rbuergi:
## Held, fixed, undrafting — the review was right and it was a correctness defect

**How it was held.** This PR was already **admitted to the merge queue** (position 2) when I read the finding, so draft alone would not have stopped it — an admitted entry merges regardless of the head PR's draft state, as #4269 proved this morning. The hold was GraphQL `dequeuePullRequest` **first**, then draft, verified by reading `repository.mergeQueue.entries` back (only #4283 remained). That is the rule this session learned the hard way; recording that it was applied.

**The finding.** `WAIT_SECONDS=420` budgeted **120 s** for the CronJob's schedule — the comment even said *"`*/10` until that change is applied"* — i.e. it did the arithmetic for a cadence change that had not been applied. Deployed is `*/10`, so a run that is AHEAD at the instant a tick fires waits one full **600 s** period before the next tick starts installing. The wait ran out at 7 minutes and still reddened the ordinary AHEAD case — the one this change exists to wait out.

**Fix, `e9ec2299ff`:**
- `600 + 210 + 90 = 900 s` — arithmetic over the values **as deployed**, with the review finding recorded beside it so the next reader knows why 120 was wrong.
- **The `gate` job cap moves 30 → 45**, and this is because of the wait, not the gate. The same file documents that an unsharded full-set gate measured **17.9–30.4 min on its own** and that **four runs were cut at the 30-min cap** — reporting `cancelled`, unreadable, the exact failure that note exists to prevent. A healthy 15-min wait plus a healthy gate does not fit in 30. 45 is the fleet ceiling `check-workflow-timeouts.py` refuses to exceed (84 jobs, 0 violations), so it is the one raise available, and the comment says not to raise it again — a job needing more is stuck, not slow.

Self-test unchanged at 21 cases (its wait cases pass explicit `--wait-seconds`, so the default is not what they measure).

---

Issue #1868, opened 2026-08-18
Hub Build constructs other hubs (1,350× per green FutuRe run) — a disposal races a TREE of constructions, not one frame


rbuergi:
Split out of #613, where it was investigated as a *crash* hypothesis and **falsified as one** (the faulting thread there is the background GC thread, with zero managed frames — [analysis](https://github.com/Systemorph/MeshWeaver/issues/613#issuecomment-5333868102)). What survives that investigation is a separate, real design defect that deserves to be judged on its own rationale rather than on whether a SIGSEGV stops.

## The fact

`MessageHubConfiguration.Build` runs `SyncBuildupActions` inline, before `StartMessageProcessing()`. Two of those actions reach code that creates hubs:

- `DataExtensions.GetDefaultConfiguration` registers `WithInitialization(h => h.GetWorkspace())` on the **synchronous** overload → `Workspace..ctor` → `DataContext.Initialize()` → `DataSource.GetStream` → `SynchronizationStream..ctor`;
- `KernelContainer.<Configure>` registers `StartActivityControlPlane` the same way → `WatchControlPlane` → `SubscribeWithReEstablish` → `AcquireStream` → `Workspace.GetStream` → `SynchronizationStream..ctor`;

and `SynchronizationStream`'s constructor **always** calls `Host.GetHostedHub(SynchronizationAddress…, HostedHubCreation.Always)`.

So every data-enabled hub builds at least one `sync/{clientId}` sub-hub — and a second Autofac container — **inside its own `Build`**. Measured with a `[ThreadStatic]` depth counter on a green run of `MeshWeaver.FutuRe.Test` (`-c Release`, native host):

```
Total: 59, Errors: 0, Failed: 0, Time: 23.195s      exit=0
1,350 nested Builds — all depth=2, all inner=sync/{clientId}
top outers: FutuRe/Analysis 156 · FutuRe/EuropeRe/Analysis 150 · FutuRe 132 · … · mesh 92
```

## Why it matters (and it is not the crash)

[...]

PR #4352, opened 2026-09-14
fix(#4300): starting a data source no longer mints a spare sync/ hub — the discarded full-reference reduce was the sixth hub

### LINKED ISSUE #4300: UNATTRIBUTED: ReadPathStreamMintingTest.AnUncachedConfiguredReduce_MintsOneSyncHubPerCall counted +6 sync hubs for 5 calls on a diff that cannot reach MeshWeaver.Data (#4292 run 34823051999)
## `ReadPathStreamMintingTest.AnUncachedConfiguredReduce_MintsOneSyncHubPerCall` — UNATTRIBUTED failure: `+6` sync hubs for 5 calls, on a diff that cannot reach `MeshWeaver.Data`

**This is filed as an unattributed failure, not a flake** (working agreement, 2026-09-14: a test failing on a diff that cannot reach it proves only that the diff did not cause it — not that the defect is nondeterministic).

### Occurrence

| | |
|---|---|
| Run | `MeshWeaver Build and Test` [34823051999](https://github.com/Systemorph/MeshWeaver/actions/runs/34823051999), shard 1, head `75ae08358bcbc377107b042c5408a1ef1df8ec11` (#4292) |
| Result | 1 454 tests, 1 453 ✅, **1 ❌** — this test; the earlier head of the same PR (`115e0a9d`) ran the identical suite green (1 454 ✅) 30 min earlier |
| Message | `DIAG uncached: calls=5 ownerSyncHubs=+6` → `Expected value to be 5 … but found 6` |
| The diff | #4292 touches `MeshWeaver.PluginCatalog`, string literals in `MeshWeaver.GitSync`, two doc pages and a test in `Memex.Portal.Shared.Test`. `MeshWeaver.Data.Test`'s closure is `MeshWeaver.Data`, `MeshWeaver.Fixture` (`Mesh.Contract`, `Messaging.Hub`), `Data.TestDomain` — none of it references PluginCatalog or GitSync. The diff cannot reach the code under test. |

No earlier occurrence is findable by search (`in:comments,title,body` for the test name returns only #3952, the PR that introduced the test). This is the first recorded one.

### What the number says

The test takes `LiveSyncHubs(host)` (the count of `sync/` hubs in the host's `HostedHubsCollection`), makes 5 uncached `GetStream(reference, x => x.WithClientId(…))` calls, and reads the count again. `HostedHubsCollection.GetHub(…, Always)` constructs synchronously, so each of the 5 calls contributes exactly one hub at return. **A `+6` therefore means one `sync/` hub was created on the host by something other than those five calls, inside the window between the two counts** — a window of a few milliseconds in the failing run (baseline at 08:36:46.919, count at 08:36:46.932).

The test's own guard against a foreign hub is `LongHeartbeat = 5 min` ("so every hub counted is one this test caused"), which covers heartbeat resubscribes and nothing else.

### The shape to look for — a deterministic one, not a re-run loop

Candidates for the foreign creation, each a real ordering rather than noise:

1. **The precondition's stream registering after its first frame.** `await workspace.GetObservable<BusinessUnit>() … Match(x => x.Count > 0)` completes when a frame arrives; if any part of that read's sync-hub creation lands on the host after the frame (a second hub the reduce mints on subscription, a partition/data-source stream created lazily on first reduce), the baseline is taken one low.
2. **The client hub connecting.** `HubTestBase` builds a client as well; a client-side subscription that mints a `sync/` hub on the HOST (its initial data mirror) is not caused by this test and lands whenever the client's startup gets there — on a loaded CI runner, plausibly inside this window.
3. **The class's previous test's stream still settling** (tests in a class run sequentially, but a stream from `ReadingTheSameReferenceRepeatedly_…` or `UnifiedUpdate_…` may complete a late subscription on the shared fixture).

Each of these reproduces at 100% by inserting the ordering deliberately (e.g. delaying the client's first subscription until after the baseline, or taking the baseline before the precondition's frame). That is the falsification to run before this is touched; a re-run that goes green says nothing about which of the three it was.

### Handling on #4292

Re-run of the failed jobs requested once, to unblock, with this issue as the record. Not added to `known-flakes.json`. A recurrence on any PR is a signal — append it here, do not re-run past it.

rbuergi:
## Defect

`ReadPathStreamMintingTest.AnUncachedConfiguredReduce_MintsOneSyncHubPerCall` counted **+6 `sync/` hubs for 5 calls** on run 34823051999 (#4292, shard 1) — a diff that cannot reach `MeshWeaver.Data`. Filed as unattributed (#4300) with three candidate mechanisms.

## Root cause — pinned by naming the hub, not by re-running

A probe subscribed to the host's `HostedHubsCollection.HubAdded` and recorded every `sync/` hub's address, thread and creation stack across the test. The baseline population was **three** hubs, not one:

| Hub | Thread | Minted by |
|---|---|---|
| `sync/JHJu…` | host init turn | `DataContext.InitializeDataSources` → `TypeSourceBasedUnpartitionedDataSource.Initialize()` → `GetStream(GetReference())` → **`GetStreamForPartition(null)`** — the primary `EntityStore` stream |
| `sync/E6Dr…` | host init turn, **8.5 ms later** | the same call's second half, **`.Reduce(reference)`** — an uncached configured reduce (`x => x`) of the store to its own full reference, **result discarded** |
| `sync/BTEH…` | test thread | the precondition's own reduce |

The test's precondition waited for a **frame**. `GetHost()` returns when `Build` returns; the data sources start on the hub's init turn (#1868), on another thread; the frame is produced by the primary stream's own initialization on a third. The init turn's second mint raced that frame and, on a loaded runner, lost: it landed after the baseline, inside the measurement, and read as the sixth hub. None of the issue's three hypotheses was it (no client is created by this test; the precondition's hub is minted synchronously in the `SynchronizationStream` constructor; the mesh is per test).

**The product defect underneath:** every `Initialize()` override (`TypeSourceBasedUnpartitionedDataSource`, `TypeSourceBasedPartitionedDataSource`, `HubDataSource`, `PartitionedHubDataSource`) called `GetStream(reference)` for the side effect of opening the primary stream and discarded the reduced stream it returns — a second `SynchronizationStream`, a second hosted `sync/` hub registered for disposal on the hub-lifetime primary, reachable by nobody: **one permanent, unused hub per data source per partition on every hub that carries data.** The same shape #3432 pinned on the read path, paid once at start instead of once per read. `DataContext` even documents the eager call as "a sync/ sub-hub per source"; it was two.

## Fix

- The four `Initialize()` overrides open the primary stream (`GetStreamForPartition`) and reduce nothing. `IDataSource.Initialized` is `WhenAll` over the partition map, which never held the reduced stream, so the init gate is unchanged; a `HubDataSource` mirror posts its `SubscribeRequest` inside `CreateExternalClient` at construction, so nothing depended on the discarded reduce as a trigger.
- `ADataSourceStart_MintsOnlyItsPrimaryStream` (new): waits for `host.Started` — which settles only after the init turn ran every `Initialize()` — and asserts exactly one `sync/` hub per data source. **Negative control run before the fix: red at 2; after: 1.**
- `AnUncachedConfiguredReduce_MintsOneSyncHubPerCall` keeps its frame precondition and takes its baseline only after `host.Started`: the baseline of a population count is the steady state, and a frame is evidence of a stream, not of a finished start.
- `DataSourceStartMintingTest` (new, after review): the partitioned type-source override (a concrete record over the abstract `GenericPartitionedDataSource`, beside an unpartitioned source on one host — 2 sources → 2 hubs) and `PartitionedHubDataSource` with two initialized partitions on a client (2 partitions → 2 mirrors). **Negative control with the four pre-fix lines restored: both read 4.**
- Finding conserved in `Doc/Architecture/ReadPathStreamMinting` §8; What's New entry (`Category: Fix`).

## Verification

- `dotnet build test/MeshWeaver.Data.Test -c Release -warnaserror` → `0 Error(s)`.
- `ReadPathStreamMintingTest` class: 11/11 green (`DIAG start: dataSources=1 ownerSyncHubs=1`, uncached `+5`, shared `+0`); before the fix the new test failed with `found 2`.
- Full `MeshWeaver.Data.Test` suite: **517/517 passed** (Release). `MeshWeaver.Documentation.Test` built `-warnaserror` at `0 Error(s)`; `DocumentationLinkIntegrityTest` 1/1 passed.

Closes #4300


rbuergi:
Review addressed in e0dd1019cb and ee53102b56:

- **Coverage of the partitioned paths (the substantive finding):** `DataSourceStartMintingTest` now pins the two overrides the unpartitioned fixture could not reach, with the same exact count after `Started` — a concrete record over the abstract `GenericPartitionedDataSource` beside an unpartitioned source on one host (2 sources → 2 hubs), and a `PartitionedHubDataSource` with two `InitializingPartitions` on a client (2 partitions → 2 mirrors). No in-repo source derives from the abstract generic, which is why that override had no test at all. **Negative control with the four pre-fix lines restored: both tests read 4** (one primary plus one discarded reduce per stream opened); with the fix, 2 and 2. Full `MeshWeaver.Data.Test`: 519/519.
- The two XML summaries are reworded ("the full-reference reduce this method used to perform and discard minted …").
- The `DataContext` note now says one `sync/` sub-hub **per stream a source's `Initialize()` opens**, which holds for the per-partition loop as well.

Issue #4158, opened 2026-09-13
[ModuleLoad] cannot name the commit a bundle was built from, nor the assembly a NodeType's content type resolved to — a stale adopted build and a stale registry shelf are indistinguishable from the consumer


rbuergi:
## What is missing

`[ModuleLoad]` prints, per module: the path, `source=`, `mvid=` and `written=`. Two questions that decided a night of production diagnosis are NOT on that line, and nothing else answers them from the consumer:

1. **Which commit the loaded bundle was built from.** On 2026-09-10 (MeshWeaver.Plugins#1585) the loaded `MeshWeaver.AI@a7971ab5…` had the newest generation, the newest `written=`, and types that predated two merged PRs. `mvid` and `written` said "newest"; only the served JSON schema said "old". The reading "the registry serves stale bytes" was written down, acted on (three RefreshModules, two restarts), and later falsified by `/health` (`bundle_adoption: … AI: FrameworkDeclined` — the bundle was never adopted; the previously adopted build kept serving under the same-MAJOR rule of #3844).
2. **Which assembly a NodeType's content type actually resolved to.** The `/schema/<Type>` endpoint and the serializer answer from whatever assembly won the TypeRegistry, and `[ModuleLoad]` names the assembly that was *loaded*, which is not the same statement when two builds of one assembly name are in the process (#3732's shape).

## Ask

- Record the built-from commit (and the framework identity it was compiled against) in the bundle manifest at pack time, and print both on `[ModuleLoad]` beside `mvid`/`written` — so "the generation is newest" and "its types are current" stop being read as one fact.
- Make the per-NodeType diagnostics name the assembly (path + mvid) the content type resolved to — the thing `/schema/<Type>` is actually serialising from — so a stale adopted build and a stale registry shelf can be told apart from the consumer.

Both are additive diagnostics; neither changes what an installation runs. Measured context and the 05:52Z falsification: MeshWeaver.Plugins#1585 (closed on the portal outcome). Adjacent: #3583 (adoption vs sweep count), #3732 (two builds of one assembly name), #3844 (stale-but-serving rule).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01BRLdhAiX5DjdZC2WUfPg9E

rbuergi:
## 2026-09-13 check — precisely stated; half the data exists one level up

[...]

Issue #3240, opened 2026-09-03
Arm the registry's own 400: ModulePublish.Validate still accepts a manifest that states no framework identity


rbuergi:
`ModulePublish.Validate` still accepts a manifest that states no framework identity, so the registry can still shelve a null and advertise it on `/api/plugins/bundles/index.json`. #3237 (part 3 of Plugins#931) closed the blind spot at the producer — the packer refuses to write such a bundle, the pack step refuses to pack without naming the anchor, and the **publish step refuses to POST one** — but deliberately did not arm the registry's own 400.

## Why it was not armed with #3237

The precondition #3211 states: *"every producer CI in the fleet must be emitting the field before the refusal is armed, or the refusal takes the fleet's publishes down."*

Measured on 2026-09-03, before #3237: **none of them was.** MeshWeaver.Plugins run 33773265959 packed all 34 bundles `built-against MVID (unrecorded)`, on both the sdk and the container path.

The registry rolls from core `main` automatically; every producer pins `node-repo-module-pack.yml@<sha>` and bumps it deliberately, per repo, in its own PR. Arming the 400 on #3237's merge would therefore have redded every satellite main-push publish for as long as its pin lagged — the nulls would have survived and the publishes would not.

## The arming criterion

Arm when **both** hold:

1. Every repo that publishes modules has bumped its `node-repo-module-pack.yml` pin past #3237's merge commit. That is exactly two repos — **MeshWeaver.Plugins** (`modules-floor` + `modules-rest`) and **MeshWeaver.SocialMedia**; MeshWeaver.Education and MeshWeaver.Reinsurance call the lane not at all.

   ```
   gh api repos/Systemorph/<repo>/contents/.github/workflows/ci.yml --jq '.content' \
     | base64 -d | grep -n 'node-repo-module-pack.yml@'
   ```

2. The registry's own log shows no recent unstated-identity publish. #3237 added the measurement for exactly this — the publish endpoint logs, per publish:

   > `Module publish for {Plugin}: '{Module}' version {Version} states NO framework identity … Its producer packs on a lane older than MeshWeaver#3211; bump that repo's node-repo-module-pack.yml pin. This becomes a 400.`

   A quiet window across at least one full publish wave of each producer is the signal.

## Then

- `ModulePublish.Validate` returns a named refusal (`the bundle states no framework identity …`) → the route's existing `400` path, with a test beside `ModulePublishShelfTest`.
[...]

Issue #4303, opened 2026-09-14
ci: Consolidate test results reads the PREVIOUS attempt's shard artefact after rerun-failed-jobs (latest-by-id is the older upload) — a shard that re-ran green cannot turn the required check green


rbuergi:
### Measured, run [34823051999](https://github.com/Systemorph/MeshWeaver/actions/runs/34823051999) (#4292, head `75ae0835`)

| attempt | `Run tests (shard 1)` | `Consolidate test results` | why |
|---|---|---|---|
| 1 | ❌ one failing test (`ReadPathStreamMintingTest.AnUncachedConfiguredReduce_MintsOneSyncHubPerCall`, #4300) | ❌ | correct |
| 2 (`rerun-failed-jobs`, 08:46Z) | ✅ 1 454/1 454, artefact `testResults-shard1` id **10338734845** uploaded 08:49:55Z | ❌ `##[error]MeshWeaver.Data.Test.ReadPathStreamMintingTest.AnUncachedConfiguredReduce_MintsOneSyncHubPerCall` · `[CI] MeshWeaver.Data.Test exit=1 TESTFAIL` | it downloaded `testResults-shard1 (ID: 10338854604, Size: 297521)` — **attempt 1's** artefact (uploaded 08:38:28Z) |

The run now holds two artefacts named `testResults-shard1`:

```
id 10338854604  size 297521  created 08:38:28Z   ← attempt 1, carries the failure
id 10338734845  size 298484  created 08:49:55Z   ← attempt 2, green
```

`actions/download-artifact@v8` with `pattern: testResults-shard*` reports `Found 12 artifact(s)` — 6 `build-output-*` + 6 `testResults-shard*` after its own per-name de-duplication — and the one it keeps per name is the **highest id**, which here is the OLDER upload: artefact ids are not monotonic across attempts (attempt 2's id is lower). So the consolidate reads the stale trx and re-declares the failure the re-run just cleared. The exit-marker gate reads the same directory and agrees, for the same reason.

[...]

PR #4366, opened 2026-09-15
fix(#4126): a module's native payloads are derived by BOTH closures and land where the loader probes

### LINKED ISSUE #4126: Registry module bundles cannot carry native runtime assets: both closure derivations drop runtimeTargets, the bundle format has no native section, and the container evaluator refuses a native-only PackageReference (measured: MeshWeaver.AppleMessages needs libe_sqlite3)

rbuergi:
## A module's native payloads are derived by BOTH closures, and they land where the loader probes

Closes #4126.

Since #4239 a registry bundle could **carry** a RID-specific native payload, and nothing ever
wrote one to disk: *carried, declared, ignored*. The consumer's landing took the assemblies and
the `wwwroot` tree and nothing else — so a module that declares an engine published 200, appeared
in the catalog, installed, loaded, rendered, and threw `DllNotFoundException` at its first
P/Invoke, with nothing in any log connecting the two. The **other** closure derivation, the
container lane, could not even DECLARE such a package.

### Root cause, measured before it was fixed

Not taken from the issue's description — re-derived from a real `SQLitePCLRaw.lib.e_sqlite3`
publish (2026-09-15), because the two deps.json shapes are the whole container half:

| publish | what the deps.json says | what is on disk |
|---|---|---|
| portable | 29 `runtimeTargets` entries, `assetType: "native"` | the `runtimes/<rid>/native/` tree, preserved |
| `-r linux-x64` (**what every MeshWeaver image is**) | a ONE-entry `native` section, key still `runtimes/linux-x64/native/libe_sqlite3.so` | `libe_sqlite3.so`, **flattened** beside the app |

`ContainerReferenceSet` read **neither** — only the `runtime` section — and `Resolve` matched a
package by the ASSEMBLY it contributes (falling back to `<id>.dll`). A package contributing only
natives therefore answered `Supplied = false` against an image that demonstrably ships its engine,
and `ProjectBuild` reds an unresolved `PackageReference`. That is why `MeshWeaver.AppleMessages`
carries a conditional pin and depends on its host.

And the collection ran **after** the `runtime` short-circuit in both derivations — the same bug in
the same place the SDK derivation had (#4239's note), skipping exactly the packages it exists for.

### What changed

**Landing.** `LandModule` / `ShelveModule` take `nativeAssets`; the publish endpoint and
`PluginBundleClient.LandFromBundle` read them off the bundle; `GenerationIdOf` folds them into the
content address — two bundles differing *only* in an engine must address two generations, or the
second landing adopts the first's directory and serves the wrong engine under the right version.

🚨 **The layout is validated by ONE predicate at every door**: `NuGetPackageWriter.IsModuleNativeLayout`
— the derivation, the packer, the bundle reader, `ModulePublish.Validate` (a named 400, so a
publisher is told rather than the server blamed) and `ModuleLandingService` itself, before a byte
touches disk. A payload at any other shape is REFUSED, never landed: it would be bytes that read
as shipped and behave as absent, which is the one outcome the section exists to prevent.

**Re-serve.** `ModuleBundleSource.NativeAssetsOf` reads the shelf's tree — an **additive** call,
not a fourth element on `CollectVersion`'s tuple, because `MeshWeaver.Plugins` destructures that
into three and widening it would red a consumer's trunk for a capability it does not use.
`ServedModule.Natives` (an `init` property — `check-record-signatures.py`) carries it from the
shelf AND out of a sealed publication, and the download route writes the section back into the
bundle it composes.

**The container derivation.** `ContainerReferenceSet` reads both deps.json shapes through one
reader (`NativeContributions`, shared with the shelf so the two cannot drift), `Resolve` treats a
native-only package as SUPPLIED, `PrivateClosure` carries `NativeRide`s and NAMES a declared
payload neither source has, `ProjectBuild` lays them out under the pack input and writes
`module-natives.txt`, and the pack lane declares them with `--with-native`.

🚨 `--with-native` takes the **opposite** argument to `--with`, deliberately: `--with` refuses a
path component because the flat closure has no place for one; this requires the path, because the
path is what the loader probes. Two flags and two builder manifests, because a lane feeding one to
the other would be wrong whichever way it was written. The pack lane refuses a native with no
builder provenance, exactly as it does a non-MeshWeaver assembly.

🚨 **The gate and publish-bake lanes compose into `/ext/modules/<Name>/`, one level deeper.**
`ModuleNativeAssets` accepts a directory as a module folder only when its PARENT is named
`modules`, so at the shallower path a module's natives would land on disk and never be probed.

**Also fixed at a call site this change already touched:** the publish endpoint never passed
`accepted.SourceCommit` to `ShelveModule`, so #4158's shelf half recorded nothing — the same
unwired-argument shape as #2221 and as this issue itself.

### Evidence

Every positive assertion was driven RED against the pre-fix code, by mutating the real thing:

| control | result |
|---|---|
| `ContainerReferenceSet` reverted to reading only `runtime` | **4 of 8** `NativeClosureTest` cases red, incl. `ANativeOnlyPackageIsSUPPLIED…` |
| the publish endpoint stops passing the natives | **2 of 6** `PluginBundlePublishNativesTest` cases red |
| both lanes' native block removed (`test-module-asset-landing.py` revert arm) | 0/2 natives land — red |
| both lanes' native `cp` removed, the in-lane check kept | the step's own assertion fails closed, naming the files |

Plus anti-vacuity controls that pass and must: a package whose payload is DECLARED but absent is
`Supplied = false` (so the verdict is about the image, not the declaration); a module with no
engine grows no `runtimes/` tree; a bundle with no natives declares no section at all; `.a`/`.lib`,
`runtimes/<rid>/nativeassets/<tfm>/…` and `assetType: "runtime"` are excluded and named.

`test-module-asset-landing.py` now runs **7 cases per lane** and EXECUTES both lanes' real steps,
asserting the bytes, the exact four segments **and** the depth.

- `dotnet build MeshWeaver.slnx -c Release -warnaserror` → 68 assemblies, **0 Warning(s), 0 Error(s)**
- `NativeClosureTest` 8/8 · `PluginBundlePublishNativesTest` 6/6 · `ModulePackCommandTest` 32/32 ·
  the module suites of `Memex.Portal.Shared.Test` 76/76 · `PrivateClosureTest`/`ContainerReferenceSetTest`/
  `ModuleLibrariesShelfTest` green · doc link integrity green
- `test-module-asset-landing.py` PASS (2 lanes × 7 cases, all four falsification arms red)
- `check-record-signatures.py` ✓ (16 changed C# files, no binary-breaking primary constructors) ·
  `check-type-forwards.py` **0 removed**, 1 added (`ServedModule.Natives`, an init property) ·
  `check-interface-addition.py` 0 obliging members · `check-workflow-timeouts.py` ✓

`test-module-set-consistency.py` reports one failure **on `origin/main` too, byte-identical** (a
pre-#3732-publisher case) — this diff cannot reach it.

### What is deliberately NOT done

1. **Warning → refusal** (the issue's stage 5). The pack log's two drop warnings stay warnings:
   arming the refusal before a full publish wave has been measured clean takes the fleet's
   publishes down instead of the drops — the #3240 shape, which was armed on two measurements
   rather than on faith. Filed as #4367, with the measurement it needs stated.
2. **The container lane carries ONE rid's engine**, because it resolves against ONE image. A host
   on another rid falls back to the runtime's own probing — strictly better than the nothing it
   carried before, and stated in the builder's log rather than left to be deduced.
3. **The module-libraries shelf's native half** is built (`DeclaredNativesOf` / `NativeFileFor`)
   and unexercised: no curated package declares one yet.

Conserved on `Doc/Architecture/Modules` → *"LANDED, RE-SERVED, and derived by BOTH lanes"*,
replacing the stage list the previous pass wrote there, plus a What's New entry.

`Pairs-with: none — nothing is removed or renamed; every public change is additive (two optional
trailing parameters, one additive static, two init properties), and the sibling repo's existing
call sites compile unchanged.`


rbuergi:
## CI is red on a dated tolerance that expired at 00:00Z today — not reachable from this diff

`MeshWeaver Build and Test` fails on ONE step, `CI's own shell` → *Execute the lanes' one-build-per-assembly-name assertion*:

> node-repo-publish-bake.yml: with a PRE-#3732 publisher on disk the step sealed [] (rc=1), expected all 5

That is `test-module-set-consistency.py`'s legacy-publisher case reading the CLOCK: `node-repo-publish-bake.yml` carries `LEGACY_PUBLISHER_RED_FROM="2026-09-15T00:00:00Z"`, and past it the step REFUSES instead of sealing all five — which is the branch the harness's positive expectation was written against.

**Measured, in a full checkout, with `origin/main`'s copies of both lanes AND both harness scripts checked out over mine:** identical single failure. And fleet-wide — every `MeshWeaver Build and Test` run created after 00:00Z today is failing the same step, with one exception: `ci/harness-tolerance-reads-the-branch` (#4369), which is the fix.

🚨 My first reading of this was WRONG and worth recording: I ran the harness against a *minimal* tree holding only the five files it names, saw the same failure, and concluded "pre-existing". That tree was missing `publish-bake-bundles.sh` — the very file the failing case stages — so the case could not have worked there. The conclusion happened to be right; the measurement was not. What settled it was **main's own green run on this PR's merge base** (bca2013a, 2026-09-14T22:48Z — before the cutoff), where both harness steps are `success`.

Holding for #4369. Once it lands this PR takes main (a re-run cannot green it — `rerun-failed-jobs` reads the previous attempt's artefact, #4303).

Everything else on this run is green, and the change set's own evidence is in the description: 68 assemblies at `-c Release -warnaserror` with 0 warnings, every touched suite green, and all four falsification arms driven red.

Issue #638, opened 2026-07-24
Top-level Space create is not atomic — churn leaves content-less, grant-less ghost roots


rbuergi:
Creates during pod churn landed the node row without content or the creator's Admin grant. The partitions were then un-writable, un-deletable (30s delete timeouts) and writes into them HANG on hub activation instead of failing fast. Create must be all-or-nothing; writes into a broken partition must fail fast with a speaking error.

From the 2026-07-24 plugin-install retrospective.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

rbuergi:
Status audit against current main (2026-08-08) — PARTIAL; re-scoping to what remains.

**Fixed — the hang half:** writes into a confirmed-absent partition fail fast with a speaking error (PartitionWriteGuardValidator.cs:250-260 — 'no partition, no write'); PathResolutionService.cs:357-380 refuses to synthesize a placeholder root for an absent partition (404 instead of activating a grain onto a missing schema, pinned by PgOnlyProdShapeTests); grain activation is bounded and always terminal (MessageHubGrain FirstNodeResolutionTimeout 30s, 'no path leaves _hubReadyRaw pending forever', real cause surfaced via GrainActivationFailureRegistry, pinned by PartitionRootActivationTest). #909 additionally made ghost-root repair writes actually land (version floor vs MonotonicWriteGuard, MeshExtensions.cs:2781-2870, pinned by GhostPartitionRootRecoveryTest).

**Remaining (this issue now tracks exactly these):**
1. **Create is still not atomic**: order is provision → write row → post-creation grant, and a failed grant returns Fail while the row STAYS ('Node persisted but post-creation handler failed', MeshExtensions.cs:693-718) — no compensating delete. The #638 residue (row present, grant absent) is still producible; only its reporting improved.
2. **Self-heal gaps for exactly that residue**: EnsurePartitionBootstrap treats a content-less ghost row as 'root exists' (rootExists = root is not null, :1192) so it never repairs it, and the heal is gated on CheckPermission(Create) (:1208) — a grant-less partition denies precisely the user whose grant went missing, so it self-heals only under System/platform-admin.
3. A provisioned-but-grant-less partition still surfaces as generic RLS 'access denied' (fast, but not the speaking partition-broken error asked for).

PR #4376, opened 2026-09-15
fix(#4061): the Space creator grant no longer latches the create flow as System

### LINKED ISSUE #4061: Space create: the Admin/Partition announcement declares no access context — intermittent 'Create permission required' on delete-then-recreate
## What happens

`MeshWeaver.Hosting.PostgreSql.Test.SpaceDeletionPartitionDropTests.DeletingSpace_DropsWholePartition_AndSameIdCanBeRecreated` (MeshWeaver.Plugins) fails intermittently at its LAST step — creating a child in a Space that was just deleted and recreated under the same id:

```
ObservableAssertionException: Expected the observable to emit a value within 30s, but it errored:
Access denied: Create permission required for node 'pgdropd50d977293/page'
```

Seen on Plugins#1678, job 103426172251, head `fc488678`. Its test output shows the sequence (all within 2.4 s):

```
21:52:01.207  TEST START
21:52:01.524  [Warning] AccessControlPipeline: Access denied: user 'Roland' lacks Update permission
              on 'Admin/Partition/pgdropd50d977293' (triggered by message=DataChangeRequest sender=portal/node…)
21:52:01.945  [Warning] RoutingServiceBase: RouteMessage: NotFound for RawJson → Admin/Partition/pgdropd50d977293
21:52:03.604  [Warning] MeshWeaver.Mesh.CreateNode: Validator rejected node creation at pgdropd50d977293/page:
              Access denied: Create permission required for node 'pgdropd50d977293/page'
21:52:03.606  [Warning] AccessControlPipeline: Access denied: user 'Roland' lacks Update permission on 'Admin/Partition/…'
21:52:03.615  TEST FAILED
```

🚨 The SAME denial appears at the FIRST create (21:52:01.524) and is survived there. So the denial is the normal state of this write, and what varies is whether the child create is refused.

## Two findings

**1. The additional-node announcement posts with NO impersonation (contained).**
`MeshExtensions.RunPostCreationHandlers` persists the nodes a post-creation handler returns from `GetAdditionalNodes` — for a Space that is the `Admin/Partition/{id}` `PartitionDefinition` — and then announces them:

```csharp
.Select(extra => persistence.WriteAndPublishCreated(extra with { State = MeshNodeState.Active }, …)
    .Do(saved =>
    {
        hub.Post(DataChangeRequest.Update([saved]), o => o.WithTarget(new Address(saved.Path)));
```

That `Post` carries the CALLER's identity, so for a Space created by an ordinary user it is access-checked as that user and denied on `Admin/Partition/{id}` — which is precisely the logged warning, sender `portal/node…`. `SpaceNodeType`'s own doc says the partition definition is written so that it "primes `PgPartitionCache` (positive) across the mesh and drives the `partition_changes` pg_notify provisioning"; a denied announcement is exactly the "announce-loss" shape the surrounding comment warns about, and `NotFound for … Admin/Partition/{id}` 400 ms later is consistent with it.

The sanctioned shape for an infrastructure post is explicit impersonation (AGENTS.md: "a write that must run as system/hub says so explicitly"), and the overload exists: `PostOptions.ImpersonateAsHub(Address hubAddress)` (`src/MeshWeaver.Messaging.Hub/PostOptions.cs:137`). Proposed fix:

```csharp
hub.Post(DataChangeRequest.Update([saved]),
    o => o.WithTarget(new Address(saved.Path)).ImpersonateAsHub(hub.Address));
```

The creator-grant write in `SpaceNodeType` already does this correctly — `Observable.Using(() => accessService.ImpersonateAsSystem(), _ => meshService.CreateNode(assignmentNode))` — so the scope covers the subscribe. This announcement simply has no scope at all.

**2. The fatal denial is a grant-visibility race on recreate (infrastructure, not contained).**
The child create needs the creator's `AccessAssignment` at `{id}/_Access`, and the security layer reads grants through the per-node stream (`SecurityServiceExtensions` over `IMeshNodeStreamCache`). The recreate happens ~1.5 s after the delete, well inside `RecentlyDeletedRegistry`'s 30 s tombstone TTL (`Ttl = TimeSpan.FromSeconds(30)`), and that registry is consulted on read paths (`MeshDataSource`, `MeshNodeTypeSource`) while `Supersede(path, version)` is what clears a tombstone when the path exists again (`SubtreeDeletionGuardStorageAdapter`). A recreated `{id}/_Access` whose tombstone has not yet been superseded, or whose stream snapshot still reads deleted, denies `Create` on the child — intermittently, by timing. Fixing this belongs with whoever owns the tombstone / stream-cache interplay; I did not attempt it.

## It is NOT a regression from Plugins#1605 (measured)

It was suspected of being caused by core moving 8372 → 8380 (a range of #1605 commits) plus main gaining #1677.

- **Core `src/` differs ONLY in Markdown between the two sets.** `git diff --name-only 74d4c8527b cf4306b47 -- src/ | grep -v '\.md$'` is empty. Access control, partitions, persistence and `RecentlyDeletedRegistry` are byte-identical. The range's code changes are `test/` only: `CollectibleUnloadDrain.cs` (+9) and core's `MonolithMeshTestBase.cs` (+24/−3), both teardown-time.
- **#1677 cannot reach this suite.** `MeshWeaver.Hosting.PostgreSql.Test` references CORE's `test/MeshWeaver.Hosting.Monolith.TestBase`, not the Plugins copy that #1677 changed.
- The teardown drain itself already existed at 8372 (core #4042), so the delta for this suite is one `Task.Yield` plus a provider release.

Measured locally (macOS arm64, Testcontainers Postgres), same Plugins tree, core checkout the only variable:

| probe | core 8372 | core 8380 |
|---|---|---|
| the single test, isolated, 15 runs | 15/15 pass, 0 failures | 15/15 pass, 0 failures |
| the whole 1106-test suite under 6-way CPU load, 3 runs | 3/3 green, target failed 0× | 3/3 green, target failed 0× |

Local detailed run of the same test (set 8380, Postgres + RLS, PASSING) shows the contrast that matters:
**no `Access denied … lacks Update permission` warning at all**, but the SAME
`NotFound for RawJson → Admin/Partition/{id}` warning as the CI run. So the denial is not a constant of this
write: the announcement normally inherits a privileged ambient context and is allowed, and in CI it
sometimes runs as the caller and is denied. That is the lost-ambient-context shape — the write declares no
context of its own.

So the local rate is 0 in both sets and the race did not reproduce off a loaded CI runner; the ownership verdict rests on the byte-identical-source argument, which does not depend on the rate.

**Verdict: pre-existing race, exposed by CI timing — not caused by the #1605 teardown work.** Plugins#1678 is red on it and is unblocked as a known pre-existing flake, not by a change to that PR.

## A control for finding 1

It needs a deterministic control that fails when the announcement is unimpersonated: create a Space as an ordinary (non-admin) user and assert that `Admin/Partition/{id}` is ROUTABLE right after the create returns (no `NotFound`), and that no `AccessControlPipeline` denial was recorded for it. Before the fix that assertion fails on the denied announcement; after it, it passes. It must not be written as "no warning logged" alone — routability is the property that matters.

rbuergi:
## The source of #4061's intermittency, not another symptom

#4061's finding 1 was "the `Admin/Partition/{id}` announcement declares no access context", and #4197
fixed it by declaring one. That was correct, and it left the question its own measurement raised
unanswered: **why was the ambient identity `system-security` on some runs and the caller on others?**

It is a defect of its own, one stage earlier. `SpacePostCreationHandler.Handle` — the creator-Admin
grant, which genuinely must run as the platform, because a brand-new partition root is a path its own
creator holds nothing on — opened its impersonation as:

```csharp
Observable.Using(() => accessService.ImpersonateAsSystem(), _ => meshService.CreateNode(grant))
```

Impersonation is an `AsyncLocal` store/restore pair. Rx runs the resource factory on the
**subscribing** thread and disposes the resource when the inner observable **terminates** — for this
CROSS-HUB create, the owning hub's response thread. `AccessContextScope.Dispose` is thread-affine, so
it writes nothing over there, and **nothing ever closes the scope on the subscriber**. That
subscriber is `MeshExtensions.RunPostCreationHandlersObs`, so the rest of the create carried
`system-security`.

That is exactly the identity #4197 measured on the announcement 41 ms apart in one run: the
announcement *rode a sibling handler's scope* when the latch was still in effect and ran as the
caller — denied on `Admin/Partition/{id}` — when it was not. Declaring the post's identity fixes that
one post. Every other stage composed after the grant still inherited `Permission.All` by accident,
and an accidental `Permission.All` fails in the direction that does not show up: a write silently
succeeding where the user would have been refused (#1444).

`SpaceNodeType.cs` was on `test/ImpersonationScopeSites.allow` for precisely this shape. The line is
retired and `TotalBudget` lowered 66 → 65.

## Measured, both directions

`test/MeshWeaver.Graph.Test/SpaceGrantScopeDoesNotLatchTheCreateFlowTest.cs` asserts
SYNCHRONOUSLY, on the subscribing thread, that subscribing to the grant hands that thread back the
identity it subscribed with — no timing, no polling, no bridge.

| | on `origin/main` (product code reverted) | with the fix |
|---|---|---|
| `SubscribingToTheGrantLeavesTheSubscribersIdentityIntact` | **FAIL** — `Expected "grantlatchprobe" … found "system-security"` | pass |
| `TheCreateStillWorksAndTheCreatorOwnsTheSpace` | pass | pass |

The revert was a real revert (`git checkout` the one file, **rebuild**, re-run), not a reasoned one.

**Its negative control is what makes it non-vacuous.** Before the property, the same fixture runs the
SAME cross-hub write through the old `Observable.Using` idiom on the same thread and requires it to
latch. Without that, a write that happened to complete synchronously would make the property pass for
a reason unrelated to the seal — `Using` *does* restore correctly when the inner observable is
synchronous, which is `SystemScopeDoesNotEscapeTest`'s first case in as many words.

`TheCreateStillWorksAndTheCreatorOwnsTheSpace` is the other guard-rail: the grant is declared
`FailsCreateOnError`, so a seal that stopped the write authorising would not degrade quietly — it
would ship ownerless Spaces (#638). It creates a Space and then a child page in it, which is the
literal symptom #4061 reports (`Access denied: Create permission required for node '{id}/page'`).

Also green locally: `ImpersonationScopeSiteRatchetGuard` (3), and
`PostCreationAnnouncementContextTest` / `SpaceRecreateGrantVisibilityTest` /
`AFreshGrantIsVisibleToTheNextCheckTest` / `DenialTheStoreDisagreesWithSaysSoTest` /
`PartitionTeardownCoverageTest` (10). `MeshWeaver.Graph` built `-c Release -warnaserror`, `0 Error(s)`.

## Scope

**This does not close #4061.** Finding 2 — the delete-then-recreate grant-visibility race in
`SpaceDeletionPartitionDropTests` — is untouched and is being pursued where its PG harness is
(Plugins#1870). What this removes from that investigation is the accidental-System identity: the
create flow now runs as the caller throughout, so the next measurement of the race is not confounded
by a stage that was privileged by timing.

No public type or member is removed and no interface gains a member, so neither cross-repo gate
applies (`SpacePostCreationHandler` is private; `TotalBudget` is a private const). No i18n key
changes.

## Also conserved

`Doc/Architecture/AccessContextPropagation` → "Choosing the identity for an infrastructure post"
gains the other half: the symptom/source table, why fixing the post is not fixing the flow, and one
sibling that is **located but not measured** — the transient→Active confirmation fan-out in
`MeshExtensions.CreateNode` carries no identity of its own either. It is written down as unmeasured
rather than changed on a guess.

Closes nothing on its own; refs #4061.


rbuergi:
## Review addressed — and one of the findings was load-bearing

All four Copilot findings were real. Two of them changed what the test **proves**, not how it reads:

1. **The error arms masked a failed write.** They called `Settle(subject)` on `OnError`, so a
   failed cross-hub grant completed the test as a success. Forwarding the fault instead reddened the
   very next run with `Node already exists: grantlatch/_Access/grantlatchprobe_Access` — **the
   sealed grant write had been failing all along**, and both identity assertions were standing on a
   write that never ran. Cause: the Space's real create already grants its creator, so re-invoking
   the handler for the SAME creator duplicates the node. The property probe now grants a second
   subject (`GrantSubject`); the creator is untouched. This is the exact class the fix is about —
   an answer that reads like a pass — so thank you, it was the right catch.
2. **The end-to-end control was vacuous.** It ran as the fixture's own circuit identity, which is
   entitled to the child create independently. Both creates now run as `Probe`, an identity holding
   nothing anywhere, through `access.RunAs` rather than an ambient scope — the cold create decides
   its identity at Subscribe, which the assertion helper performs, so an ambient `using` would not
   have covered it. Probe may create a top-level Space because `SpaceAccessRule` allows any
   authenticated identity, and may create a child ONLY because the grant made it the Space's Admin.
3. **`TotalBudget` lowered to 64, not 65.** You are right that 65 against a file summing to 64 left
   a free slot — a new allowance could have been added without tripping the total check. Note the
   slack was pre-existing (65 vs 66 on `main`); it is now spent, and the seed is exact.
4. **try/finally around the waits and disposal.** The subscriptions were disposed only on the success
   path — i.e. never on the path this test exists to take.

Also took the suppressed What's New note: the entry implied #4061's "Create permission required" was
gone. Finding 2 (delete-then-recreate grant visibility) is untouched and still open, and the entry
now says which of the two causes this removes and which remains.

Re-measured after the hardening — still **fails** on reverted product code (`found
"system-security"`) and passes with the fix.

The PR was dequeued via `dequeuePullRequest` before pushing, never by pushing onto a queued branch.

Issue #385, opened 2026-07-09
Inconsistent visibility for private/invited users: inaccessible nodes listed in 'last edited', invited module missing from navigation, search and access menu expose users


rosalie1984:
## Context
Follow-up to #384. After completing sign-in as a **private user** (invited via private email, Google sign-in), the portal itself loads and generally looks OK — but node **visibility** is inconsistent with the user's actual access rights, in both directions (shows too much, and too little).

## Problem 1 — 'Last edited' feed lists nodes the user has no access to
The portal's last-edited nodes view shows nodes that the private user cannot access. Access being denied on open is correct — but these nodes should not be **listed** at all. Showing their existence/titles/paths to an unauthorized user is a metadata leak; row-level security should filter the last-edited feed the same way it filters direct reads.

## Problem 2 — Invited module 'Agentic Engineering' does not appear as a node
The user was explicitly invited to the educational module **Agentic Engineering**, but it does not show up as a node anywhere in the portal (navigation / node listings). Navigating directly via the browser URL works and access is granted as expected — so the permission itself is in place, only the discovery/navigation surface misses it. Invited-to content should be visible in navigation, otherwise invitees can only reach it via out-of-band links.

## Problem 3 — Search exposes users of the educational program
When reaching the module via search instead of the direct URL, the search results included **user nodes of other participants in the educational program**. A private invitee should not be able to see (or enumerate) other users; these should be excluded from their search results entirely.

## Problem 4 — Access menu lists all users of the space
The **access menu** shows the private user **all users of the space**. A private invitee should not be able to enumerate the space's user base from the access/sharing UI — the user picker / access list should be restricted (e.g. to users they legitimately need to see, or hidden entirely for users without manage-access rights).

## Expected behavior
One consistent access model across all surfaces — direct navigation, 'last edited' feed, navigation tree, search, and the access menu should all reflect exactly what the user is entitled to see:
- No listing of inaccessible nodes in feeds or search (Problems 1 and 3).
- Content the user IS entitled to (invited module) discoverable in navigation, not only via direct URL (Problem 2).
- No enumeration of the space's users via the access menu (Problem 4).

## Notes
- Problems 1, 3 and 4 are the security-relevant half (information disclosure / user enumeration); Problem 2 is a UX/consistency gap.
- Suggests the row-level-security filter is applied on direct node reads but not (or differently) on the last-edited query, the navigation tree source, the search index results, and the access-menu user list.
[...]

PR #4377, opened 2026-09-15
feat(home): a Spaces scope for the Apps launcher, plus the strings its filter row needs

rbuergi:
Companion to MeshWeaver.Plugins#1876 (the Apps band becomes a filter over one query). Two things, both data rather than API:

1. **A second scope on `BuildAppsBand`: Spaces.** The Plugins GUI renders every scope past the first as one more entry in the launcher's filter combobox, so this is what puts the viewer's workspaces on the launcher — icons, most recently used first (`SortByAccess`, the viewer's own log). The query is the partition roots the reader can see (`RootTypeFilter`, the same allow-list the content list's root leg uses) plus a union leg for the spaces they were invited into (#385), which no scope query can reach. `Sortable = false`: the arrangement lives on `App` records and a space has none.
2. **Twenty localization keys (en + de)** for the launcher's chrome: Recent / Favorites / Filter / Categories / Browse / the view switch / the tile menu's category and favorite items / four empty states. The Plugins pack reads them through its own catalog-miss fallback (the `CatalogText.Or` rule, restated locally since that pack does not reference the pack holding it), so the two PRs land in either order — these keys are what makes the German portal read German rather than English.

`BuildAppsBand` now takes the shared targets, screened by the presentation mode exactly like the content section's own legs.

Evidence: `MeshWeaver.Graph.Test` 38/38 green, including a new case pinning the Spaces scope's shape and both query legs, and `SpacesQuery(null)`/`([])` carrying only the roots leg.

Mirror-sync: `npm run sync:i18n -- --ref <this PR's merge commit>` in MeshWeaver.Plugins, tracked on MeshWeaver.Plugins#1898 — the PR that consumes these twenty keys. Its launcher reads every one through a catalog-miss fallback, so the React mirror being a step behind degrades to the authored English rather than to a raw key; the sync runs there because the mirror is pinned to a core commit and cannot be moved from here.


rbuergi:
Back in the queue on a new head (e73cf1c1f9, a trunk merge — 117 commits behind main, merged cleanly; catalogs verified 1540 + 20 keys in both languages, no duplicates).

The 07:58 ejection was `ARefusedReplyReachesItsRequesterTest.PastShutDown_…` getting a router `NotFound` where it expects the owner's `ShuttingDown` — a teardown ordering race this diff cannot reach (a layout query, two localization files, a Graph test). Measured locally in Release: the whole class ×25 on clean `main` 25/25 green, ×25 on this branch merged with `main` 25/25 green; this PR's own run was 7,557/7,557. Tracked as #4432 with the evidence, not catalogued (one occurrence is below the catalogue's bar).

Issue #2993, opened 2026-09-01
A node can reference a NodeType that no longer resolves: two holes, each needing a decision


rbuergi:
## Summary

An instance can carry a `nodeType` that resolves to nothing. Two separate code paths allow it, and
each has a **real counterparty**, so both need a decision rather than a patch. Filing rather than
implementing, per maintainer direction — a guard that breaks the importer would be worse than the
hole.

Prerequisite already handled in PR #2992: until that lands, neither of these is *observable*, because
the diagnostic that names a dangling NodeType could never be delivered (two equal timeouts).

Found while auditing `memex.systemorph.com` and `memex.meshweaver.cloud`. One live instance:
`rbuergi/_Draft/PartnerRe_EslProposalQA` (`nodeType: EmailDraft`) reads as `Unavailable` and never
reaches a verdict.

---

## Hole A — `update` accepts a `nodeType` that does not exist

**Create validates it.** `MeshExtensions.cs:756-786` rejects with
`NodeCreationRejectionReason.InvalidNodeType`, against the rule
`FindStaticNode(nodeType) is not null || persistence.Exists(nodeType)`.

**Update does not.** `UpdateAccordingToSourceNode` (`MeshExtensions.cs:4612`) copies `NodeType`
through unvalidated, and no registered `INodeValidator` ever produces `InvalidNodeType` for
`NodeOperation.Update` — every producer is on the create path. `ContentDiscriminatorValidator`
explicitly returns `Valid()` when the type resolves to nothing.

So `update` is a supported route to **create** the orphan condition.

> **Counterparty.** `StaticRepoImporter` (~`:1026`) *deliberately relies* on update skipping this
> check, as the ordering escape hatch for cases `ImportWriteOrder` cannot sequence (cycles, a type
> carried by no source). A naive validator on the update path would break imports.

**Decision needed:** refuse on update (and give the importer an explicit, narrow bypass), or warn
loudly without refusing?

---

## Hole B — pruning a NodeType strands its instances

`ComputePrunableNodes` (`StaticRepoImporter.cs:412-441`) has five guards and **no NodeType-aware
guard and no instance check**. A NodeType node is pruned exactly like a Markdown node, recursively
(source, activity, release history). Same in `PackageInstaller.PruneRemovedNodes:2915`. A rename
(`HandleMoveNodeRequest:4697`) never rewrites instances.

The inverse direction — writing an instance before its type — *was* fixed (`ImportWriteOrder`,
[...]

PR #4383, opened 2026-09-15
fix: 33 doc pages whose front matter YAML refused or silently truncated

rbuergi:
Prompted by MeshWeaver.Crm#105, I swept every `*.md` front matter in the fleet with a real YAML parser. The same defect is here, at scale, in this repository's own documentation tree — **33 pages ship with no description**, and every one of them is green.

## Half one — 20 pages the parser REFUSES

7 Architecture pages (`ReadingCiSignals`, `ContinuousDeliveryContract`, `DenialIsAnAnswer`, `BakePublicationReceipt`, `PortalHeapIsHubs`, `BaseStateTimeoutCensus`, `ReadingASiloEviction`), 12 What's New entries, one GUI page. Two causes: a `": "` inside an unquoted plain scalar, and a single-quoted scalar with an undoubled apostrophe.

**Why nothing went red.** `MarkdownFileParser` catches the deserializer failure and falls back to a regex rescue that recovers `NodeType`, `Name`/`Title`, `Category`, `Icon`/`Thumbnail`, `State` and `Order` — and **not** `Description`/`Abstract`:

```csharp
Description = frontMatter?.Abstract,   // null once the YAML failed
```

The node lands, the page renders, the title is right, and the summary is absent from every catalog card, TOC entry and Postgres search row.

## Half two — 13 pages that PARSE and are cut short

The new guard's **second arm** found these, and the parse check cannot see them by construction: an unquoted ` #` opens a YAML comment, so the document is valid and the value is simply shorter than it was written.

| page | kept | lost |
|---|---|---|
| `DanglingNodeTypes` | "… the repair path both decisions had to leave open — issue" | `#2993.` |
| `ModuleOwnedSiblingsRide` | "The settled answer to" (8 words) | 58 words |
| `NodeTypeWithNuGet`, `NugetPackages` | "… using the" | the `#r "nuget:…"` directive that is the page's whole subject |

## The guard — and why it is not new

`SkillFrontMatterGuard` already enforces exactly this, correctly, with both arms. Its **subject** was just `.claude/skills/*/SKILL.md`. `DocumentationFrontMatterGuard` points the same two checks at `src/MeshWeaver.Documentation/Data`, with a >100-file floor so a moved tree cannot turn it into a green tick over an empty scan.

## 🚨 The test that was watching could not see either half

`WhatsNewEntryIntegrityTest` asserts every entry carries a `Description:` — but reads the field with a **regex**, the same way the rescue does. A line the regex finds and the parser refuses satisfies it. Twelve entries were green in that test and description-less in the mesh at the same time: two readers, one artefact, opposite verdicts.

## Verification

- `dotnet build -c Release -warnaserror` → `0 Error(s)`
- **Both guard arms watched FAILING before they passed.** Reverting `ReadingCiSignals.md` → `front matter does NOT parse — SemanticErrorException`; reverting `DanglingNodeTypes.md` → `'Description' is unquoted and contains ' #'`. Restored → green.
- `Documentation.Test` (front-matter + What's New + doc-link + topic-map): **10 passed, 0 failed**
- Every fixed value verified to round-trip to byte-identical text.

## Fleet scope

The same sweep found this defect in three sibling repos, each fixed in its own repository: MeshWeaver.Crm (2 — the agent and skill, #105), MeshWeaver.Plugins (3 — two deck slides and a live agent node), MeshWeaver.Education (7 — still to do). Reinsurance, SocialMedia, Manufacturing and Memex are clean.


rbuergi:
All three findings addressed in `ccfebc7a50` — and the first two are this PR's own subject turned back on it, which is the best kind of review.

**1. The block scalar would have degraded the very test I was criticising.** Exactly right: `WhatsNewEntryIntegrityTest` reads `StartsWith("Description:")`, so against `Description: >-` it sees the literal marker `">-"` as a non-empty description however empty the folded body. I had complained that the test uses a different reader from the runtime and then handed it a form that widens the gap.

`ReadFrontMatter` now **parses** the block and returns a dictionary; `Field` is a lookup. One artefact, one reader — the same one the runtime uses.

Measured both directions: a probe entry with `Description: >-` and no body is now rejected —

```
2026-09-15-zz-probe.md: missing 'Description:' — the one-line summary shown in the feed
```

— and the old `StartsWith` reader accepted it.

**2. The unterminated block escaped my own guard.** Also right, and worse than it looks: `FrontMatterOf` returned `null` for a block that opens `---` and never closes, and `FrontMatterFiles` filtered `null` out as "carries no front matter" — so the one shape **no parser can read** was the one shape the guard could not see. That is precisely the skip-trapdoor the file exists to close.

It now returns a sentinel and the guard reports it by name:

```
…/ZzUnterminatedProbe.md: the front-matter block OPENS with '---' and never closes,
so no parser can read it and every field in it is lost.
```

Watched failing on a probe, then the probe removed. The `#`-truncation arm skips the sentinel, since the sibling test already names it.

**3.** Fixed — "an issue reference".

Verification: `-c Release -warnaserror` → `0 Error(s)`; front-matter + What's New + doc-link filter → **6 passed, 0 failed**.

PR #4385, opened 2026-09-15
fix: the bundle asset gate reads a BOM-prefixed manifest, and refuses instead of crashing

rbuergi:
🚨 **This is blocking every MeshWeaver.Plugins bundle right now.** All eight `Module bundles / Module bundle (…)` batches are red on PRs whose diffs cannot reach them, and the Plugins bundle lane gates the fleet's publication seal.

```
File ".../check-bundle-static-assets.py", line 138, in declared_assets
  document = json.loads(manifest.read_text(encoding="utf-8"))
json.decoder.JSONDecodeError: Unexpected UTF-8 BOM (decode using utf-8-sig)
```

**Timeline:** the gate merged at **07:36:03Z** (`84007d1097`, #2384). Plugins `main`'s last green was **07:04Z**; the first bundle red was **08:04Z**. Nothing else changed.

## Two halves, both measured

**The manifest carries a BOM, and always has.** `NuGetPackageWriter.cs:187` writes it with `new StreamWriter(stream, Encoding.UTF8)` — and .NET's `Encoding.UTF8` emits the byte-order mark by design.

**Python's `"utf-8"` codec treats that mark as content**, so `json.loads` raises on the first character.

The fix is on the **reader**: `utf-8-sig` decodes a plain UTF-8 file identically and additionally strips the mark, so it is correct for both producers and can never be the narrower choice. Re-encoding the manifest instead would change every bundle's bytes, and those are content-addressed.

## Why the gate shipped broken

Its 14 self-test cases all build their manifest with the file's own `_manifest()` helper, which writes through **Python**. No case could observe how a manifest written by anything **else** decodes — so the gate was green on every case it had and red on the first real bundle it met.

Two cases now exercise the reader directly, outside that helper:

- a **BOM-prefixed** manifest (written as raw bytes, as .NET writes it) must be READ
- a **malformed** one must REFUSE BY NAME

## And a crash is not a verdict

The uncaught `JSONDecodeError` printed a traceback naming neither the module nor the file, so eight red legs read as the harness dying rather than one unreadable manifest. `declared_assets` now raises `CannotReadManifest`, and `check` turns it into an `::error::` naming the module, the manifest and the decode error — then fails closed with exit 1.

## Verification

- `--self-test` → **16 cases, all pass** (was 14)
- **Both new cases watched FAILING first.** Reverting only the encoding fails the BOM case (`exit 1, wanted 0`). Reverting the refusal wrapper as well makes **both** raise `CannotReadManifest` out of `check` — which is exactly the crash this change removes.

## Not in scope

`grep` finds ~20 more `read_text(encoding="utf-8")` sites under `.github/scripts/`, several feeding `json.loads`. Most read git-authored or Python-written files and are fine today; any that comes to read a .NET-written artifact has this same trap waiting. Worth a follow-up sweep, not a same-PR change while the lane is red.


rbuergi:
Both findings valid, both fixed — and both are this change's own subject one step further in.

**1. Parsing is not validating.** Right: the wrapper only caught decode/IO, so `[]`, `{"module": [1]}` and a non-list `staticAssets` still raised `AttributeError`/`TypeError` out of `check` — the exact crash this PR exists to replace. `declared_assets` now checks the shape of `document`, `module` and `module.staticAssets` inside the same try and refuses by name.

One thing the new case caught that I got wrong first: **order matters.** `x or {}` silently rescues an *empty* list, so a check written after the coercion passes `"module": []` and one written before refuses it. My first attempt validated after coercing and the case failed:

```
FAIL [bom] a STRUCTURALLY malformed manifest REFUSES too: exit 0 (wanted 1)
```

Each value is now read, `None`-defaulted, then type-checked — never coerced first.

**2. The summary didn't substantiate the claim.** Also right, and it stings in this file specifically: `--self-test` printed `14 case(s)` while running 16, and my commit message said 16. That is an answer that reads like a pass, in the gate whose job is to stop those. The reader cases are a named collection now and the summary counts both groups:

```
18 case(s) — 14 asset-shape + 4 manifest-reader: 9 must FAIL the gate, 9 must PASS it.
```

Four reader cases now: BOM, not-JSON, `{"module": []}`, and a top-level `[]`.

**Negative controls.** Both new structural cases watched failing with the validation removed — and the top-level case reproduces your predicted crash verbatim:

```
FAIL [bom] a manifest whose top level is not an object REFUSES:
     raised AttributeError: 'list' object has no attribute 'get'
```

Pushed as `HEAD`. This PR is holding every MeshWeaver.Plugins bundle leg red, so I'd like it in as soon as checks allow.

Issue #4299, opened 2026-09-14
Review is advisory on core main: three PRs merged today with valid findings standing because checks outran the reviewer — the third (#4291) left three security-relevant findings on main until #4297


rbuergi:
Three times on 2026-09-14, a PR on core `main` **merged with valid review findings standing**, purely because the required checks finished before Copilot did — and in the third case the findings were security-relevant.

| PR | what merged past review | how the hold failed |
|---|---|---|
| **#4269** (05:33Z) | a change to the fleet's most-read publication prefix, with a finding that the pointer could move backwards | draft was applied at 05:30:17Z — 61 s **after** the queue admitted it at 05:29:16Z; an admitted entry merges regardless of draft. Fix landed as #4273, 29 min later. |
| **#4282** (07:5xZ) | *caught in time* — the wait bound was arithmetic for a cadence change that had not been applied (420 s against a 600 s period) | the PR was already at queue position 2 when the finding was read; `dequeuePullRequest` first, then draft, worked. The only one of the three that was held. |
| **#4291** (08:4xZ) | the self-update control lane, with **seven** findings — three security-relevant | the findings arrived while the PR was in the queue; the push to address them was **refused by the queue's branch lock**. Merged as-is; follow-up #4297 is on first CI. |

The three #4291 findings that are on `main` right now (`29f166a7e1`), until #4297 lands:

1. **A successful HTTP status is accepted as proof the control inbox verified the HMAC.** An empty or malformed response, and even an explicit `signature: not-required` answer, is converted into a successful hand-over. An *unverified* answer counts as a verified one. (`SelfUpdateHandover.cs:338`)
2. **A bare `Hosting/PlatformBuilds` target with no `SecretConfigKey` is treated as intentionally unsigned** — so it *looks* like it has `Hosting:PlatformWebhookSecret` when it does not. (`SelfUpdateHandover.cs:276`)
3. **The manual apply path never consults `ComboVerificationGate`.** A candidate with a recorded **RED** combo verdict can still be handed over or self-applied from the Updates tab. (`UpdatePolicySettingsTab.cs:140`)

None of these is exploitable by an unauthenticated party as far as I can tell — they weaken verification on a lane that already requires an operator — but they are exactly the class a review exists to catch before merge, and they were caught, and it merged anyway.

## What this is not

[...]

Issue #1984, opened 2026-08-21
Every .md-authored Skill imports as nodeType: Markdown — the two .md parsers disagree about front-matter casing, silently


rbuergi:
Found while measuring #1580's exposure. **The documented, preferred authoring form for skills does not produce a Skill node.** It produces a plain Markdown page — no error, no log, nothing to notice.

## Measured on memex

```
Skill/thread                    nodeType: Markdown   name: thread        content.$type: MarkdownContent
Hosting/Skill/deployment        nodeType: Markdown   name: deployment
Hosting/Skill/deployment-activity  nodeType: Markdown
Hosting/Skill/instance          nodeType: Markdown
Hosting/Skill/platform-update   nodeType: Markdown
Hosting/Skill/ui-extensibility  nodeType: Markdown
Feedback/Skill/feedback         nodeType: Markdown
```

Every one of their repo files declares `nodeType: Skill` in front matter. `Feedback/Skill/feedback.md` is the example MeshWeaver.Plugins' AGENTS.md tells authors to copy:

> A **skill** is a `.md` with `nodeType: Skill` front matter … whose body is the skill's procedure (see `Feedback/Skill/feedback.md`).

Meanwhile the `.json`-authored skills are fine — `Essentials/Skill/access`, `Essentials/Skill/code`, `DataModelling/Skill/ModelCube` all carry `nodeType: Skill` and the `/slash` name.

**Consequence**: `SkillNodeType.Build` filters on `node.NodeType == "Skill"`, so none of these 11 skills reach the slash-command list. Their `name` also loses its `/slash` form and their `icon` is dropped, because those front-matter keys are lost by the same mechanism.

## Root cause — proven executably

Two parsers claim `.md` and configure YAML differently:

```csharp
// AgentFileParser.cs:25
new DeserializerBuilder()
    .WithNamingConvention(CamelCaseNamingConvention.Instance)   // ← binds nodeType:
    .IgnoreUnmatchedProperties().Build();

// MarkdownFileParser.cs:24
new DeserializerBuilder()
    .IgnoreUnmatchedProperties().Build();                       // ← no convention: PascalCase only
```

`AgentFileParser` claims only `nodeType: Agent` and returns null otherwise, so a skill falls through to `MarkdownFileParser`, whose deserializer never matches a lowercase key — and `IgnoreUnmatchedProperties()` swallows the miss. `NodeType` then falls back to its `?? "Markdown"` default.

[...]

PR #4386, opened 2026-09-15
review(#4383): the What's New test parses, and an unterminated front-matter block is a failure

rbuergi:
#4383's auto-merge landed at **08:10:19Z** on head `3dd737105` — while I was fixing the three review findings. `ccfebc7a50` never got in. This is that commit, rebased onto the merge.

That is the #4299 shape once more, and slightly worse than usual: the review was read and acted on, and the checks still outran it. Worth noting on #4299 as another measured instance.

Two of the three findings are #4383's own subject turned back on it.

## 1. The block scalar would have degraded the very test the PR was criticising

`WhatsNewEntryIntegrityTest` reads fields with `StartsWith("Description:")`. Against the new `Description: >-` form it sees the literal marker `">-"` as a non-empty description — however empty the folded body. #4383 complained that the test uses a different reader from the runtime, then handed it a form that *widens* the gap.

`ReadFrontMatter` now **parses** the block and returns a dictionary; `Field` is a lookup. One artefact, one reader — the same one the runtime uses.

Measured both directions. A probe entry with `Description: >-` and no body is now rejected:

```
2026-09-15-zz-probe.md: missing 'Description:' — the one-line summary shown in the feed
```

The old `StartsWith` reader accepted it.

## 2. An unterminated block escaped the new guard

`FrontMatterOf` returned `null` for a block that opens `---` and never closes, and `FrontMatterFiles` filtered `null` out as "carries no front matter". So the one shape **no parser can read** was the one shape the guard could not see — precisely the skip-trapdoor the file exists to close.

It now returns a sentinel and the guard reports it by name:

```
…/ZzUnterminatedProbe.md: the front-matter block OPENS with '---' and never closes,
so no parser can read it and every field in it is lost.
```

Watched failing on a probe, then the probe removed. The `#`-truncation arm skips the sentinel, since the sibling test already names it.

## 3. Grammar

"a issue reference" → "an issue reference".

## Verification

- `dotnet build -c Release -warnaserror` → `0 Error(s)`
- Documentation.Test front-matter + What's New + doc-link filter → **6 passed, 0 failed**
- Both new behaviours watched failing on a probe first


rbuergi:
All three valid; fixed in `7af3e4953d`. The third one is the most useful — it says my reader was wrong in the *opposite* direction from the bug this PR is about.

**1. The unterminated arm had no negative control in CI.** Exactly right, and I should have caught it: I proved that arm with a temporary probe file and then deleted the probe. On the committed (clean) tree, reverting `end < 0 ? Unterminated` to `null` would have made the guard silently skip the one shape no parser can read, with every test still green. A guard arm proved only by a probe the reviewer has to take on trust is not proved.

`FrontMatterOf_TellsTheThreeShapesApart` now pins all three classifications on synthetic strings — no front matter, a `---` that is not at the start, an opener with no closer, and a properly closed block. Watched failing with the sentinel reverted:

```
Failed!  - Failed: 1, Passed: 0  (FrontMatterOf_TellsTheThreeShapesApart)
```

**2. `ToString()` made structured values look like scalars.** Right: `Description: {text: …}` and `Description: [a, b]` are valid YAML, and `ToString()` turns each into a non-empty CLR type name that satisfies every presence check while the feed renders nothing useful. Non-scalar values are filtered out now, so such an entry reads as **absent** and fails — rather than passing on a type name.

**3. The reader was stricter than the runtime, and its comment claimed the reverse.** This is the one worth spelling out. `MarkdownFileParser` sets `WithCaseInsensitivePropertyMatching()` **on purpose** (#1984 — eleven skills imported as plain Markdown because a PascalCase-only deserializer never bound `nodeType:`, and 1146 plugin-repo files author the same keys PascalCase). So `description:` and `Description:` both bind at runtime.

The test read `StringComparer.Ordinal` and justified it with *"a lowercase `category:` would be silently dropped rather than accepted here"* — which stopped being true when #1984 landed. It would have redded an entry the feed renders perfectly. That is the same two-readers-one-artefact defect this PR is about, pointing the other way, with a stale comment keeping it in place. Now `OrdinalIgnoreCase`, with the reason recorded.

Verification: `-c Release -warnaserror` → `0 Error(s)`; front-matter + What's New filter → **6 passed, 0 failed**.

PR #4394, opened 2026-09-15
fix(#4382): a declared Space stops re-writing itself on every import

### LINKED ISSUE #4382: Space.CreatedAt defaults to UtcNow — every declared Space is non-idempotent by construction (3 live in samples/)
`Space.CreatedAt` defaults to `DateTimeOffset.UtcNow`, so **any Space declared in a content repo is non-idempotent by construction**: the declared JSON omits `createdAt`, every deserialization stamps a fresh clock reading, the stored node differs from the incoming one, and the installer's unchanged-skip can never hold.

```csharp
// src/MeshWeaver.Graph/Configuration/SpaceNodeType.cs:57
public DateTimeOffset CreatedAt { get; init; } = DateTimeOffset.UtcNow;
```

## How it surfaced

MeshWeaver.Plugins#1901 added a `Publish/Decks` folder declared as `nodeType: Space`. Its content CI failed:

```
[FAIL] Publish (32 node(s), 2 type(s)) [1/1 content asset(s) served]
    idempotence: re-install of the unchanged snapshot wrote 1 node(s)
    (expected 0 — the unchanged-skip regressed): Publish/Decks
GATE FAILED — idempotence: Publish — 1 new failure(s)
```

The PR's own fix is to use the `Markdown` folder idiom instead (a Space is a tenant container, not a folder — its post-creation handler also emits an `Admin/Partition/{id}` `PartitionDefinition`, so a documentation folder would have provisioned a Postgres schema). That is merged separately; **this issue is the framework half**, which no content repo can work around.

## Denominator — this is already live in `src/`

Every Space declared in this repo's own sample trees omits `createdAt`, measured 2026-09-15:

| file | Space | content keys |
|---|---|---|
| `samples/Graph/Data/Systemorph.json` | `Systemorph` | `$type, description, email, isVerified, location, logo, name, website` |
| `samples/Graph/Data/ACME.json` | `ACME` | `$type, category, description, name` |
| `samples/Graph/Data/MeshWeaver.json` | `MeshWeaver` | `$type, description, email, isVerified, location, logo, name, website` |

None declares `createdAt`. So each of these three is re-written on every static-repo import — a write per Space per import, forever, and a version bump each time. The Publish gate is the first instrument that *reported* it; it has been true for as long as the default has existed.

## Why it reads like a pass

Nothing goes red. The re-write succeeds, the node is correct, the portal serves it. The only visible trace is a version number that climbs with no author — which is exactly the shape catalogued as [an answer that reads like a pass](https://github.com/Systemorph/MeshWeaver/issues/4299). The idempotence gate is the one place that asks "would this write have happened if nothing changed?", and it only runs where a pack is installed.

## The fix, and the thing to decide

`CreatedAt` is *derived provenance*, not authored content: a declaring file should not have to write it, and a re-import should not move it. Two candidate shapes, and the choice is a maintainer call because it changes a persisted field's meaning:

1. **Make it nullable** (`DateTimeOffset? CreatedAt`) and have the create path stamp it once, on first materialization, leaving a re-import to preserve whatever is stored. Idempotent for both declared and UI-created Spaces; existing rows keep their value.
2. **Drop it from `Space`** and read creation time from the node's own version history, which already records it. Removes a duplicated fact, but it is a public property removal — `Pairs-with:` applies, and any consumer reading `content.createdAt` has to move.

**A default of `default(DateTimeOffset)` is NOT a fix**: it would make the field silently wrong (0001-01-01) for every Space rather than merely unstable, trading a visible instability for an invisible falsehood.

Worth a sweep for the same shape before fixing just this one: any `record` in `src/` used as node content with a `DateTimeOffset.UtcNow`, `Guid.NewGuid()` or `DateTime.Now` property initializer has this defect, and each one is invisible until a pack that declares it meets the idempotence gate.

rbuergi:
`Space.CreatedAt` defaulted to `DateTimeOffset.UtcNow`, so **any Space declared in a content repo was non-idempotent by construction**: the file omits `createdAt`, every deserialization stamps a fresh clock reading, the incoming content differs from the stored content, and the installer's unchanged-skip can never hold.

Caught by MeshWeaver.Plugins#1901's idempotence gate:

```
idempotence: re-install of the unchanged snapshot wrote 1 node(s)
(expected 0 — the unchanged-skip regressed): Publish/Decks
```

And live in this repo the whole time: `samples/Graph/Data/{Systemorph,ACME,MeshWeaver}.json` each declare a Space and each omits `createdAt`, so **all three have been re-written on every static-repo import** for as long as the default has existed. Silently — a re-write that produces the *correct* node goes green; the only visible trace is a version number climbing with no author behind it.

## Why stamping it on create is not the fix

This is the part worth reading, because it is the obvious alternative and it does not work: a declaring file that omits the field would then present `null` against a **stored** stamp, differ again, and re-write again. Provenance a declaration does not carry cannot live on the declared content at all. The node's own version history already records when it first materialised, and that is the reading to use.

A default of `default(DateTimeOffset)` is not the fix either, and a test says so: it would write `0001-01-01` onto every declared Space — trading an unstable value for a false one.

## Nullable rather than removed

Measured 2026-09-15: the declaration is the **only** occurrence of `Space.CreatedAt` in core (`src/`, `test/`, `samples/`) or in MeshWeaver.Plugins — nothing reads it. Keeping it nullable preserves rows that already carry a value and removes nothing from a public surface, so no `Pairs-with:` counterpart is owed.

## Verification

- `dotnet build -c Release -warnaserror` on `MeshWeaver.Graph`, `MeshWeaver.GitSync`, `MeshWeaver.PluginCatalog` → `0 Error(s)` each
- `DeclaredSpaceIsIdempotentTest` → **3 passed**, and **watched failing (2 of 3)** with the clock default restored, then restored
- Documentation.Test What's New + front-matter filter → 5 passed

## Worth a sweep, not in this PR

15 property initializers across 14 types in `src/` stamp a clock or a guid. Restricted to node trees a portal actually imports, only two content types are declared at all: `Space` (3 files, **3 of 3 omit the field** — this defect) and `Comment` (10 files, **0 of 10 omit it** — clean). I expected `Comment` to be the larger instance and it is not. The trap stays armed for anything anyone declares next, which is exactly how Plugins#1901 hit it by declaring the first `Space` in a content repo.

Closes #4382


rbuergi:
Both findings are fair. One I have answered with a wider sweep; the other I have made explicit in the code rather than argued away — and the first one deserves a caveat I cannot close myself.

## 1. The property-type change is a binary break — in general, and here is what it costs in this fleet

You are right that `DateTimeOffset` → `DateTimeOffset?` changes the CLR getter signature, and that a repository-wide source sweep is not the same statement as binary compatibility. I had conflated the two in the commit message. So I widened the sweep past the two repos I had checked:

| where | readers of `Space.CreatedAt` |
|---|---|
| core `src/`, `test/`, `samples/` | the declaration itself, nothing else |
| MeshWeaver.Plugins | **one** file casts to `Space` at all — `src/MeshWeaver.Graph.Views/SpaceLayoutAreas.cs` — and it contains **zero** occurrences of `CreatedAt` |
| .Education, .Reinsurance, .SocialMedia, .Manufacturing, .Crm, Memex | no `Space` cast in any of them |
| the live mesh | two `search_chunks` sweeps, both `searched: true`, `count: 0` — a real zero, not the `searched: false` shape that is a failed sweep |

And the propagation model matters as much as the census: satellites compile **from source** against a sealed platform set, in-mesh NodeTypes compile at runtime against the deployed assembly, and baked bundles are re-baked per framework identity. Consumers here arrive by recompilation, not by binding to a pinned `MeshWeaver.Graph`. With an empty reader set, a recompile has nothing to break either.

🚨 **The residual is not mine to rule out.** Any assembly compiled against `MeshWeaver.Graph` from **outside** these eight repositories would take the break, and I have no instrument that sees those. If such a consumer exists, the alternative that is binary-compatible is to drop the initializer and keep the type — `default(DateTimeOffset)` is stable, so it fixes the idempotence too, at the cost of writing `0001-01-01` onto every declared Space. I did not choose that because a false value with no reader is worse than a nullable one; if the external-consumer risk is real, it is the better trade and I will switch.

## 2. The first import drops a previously stamped value — intended, now said out loud

Correct, and it was missing from the PR. `681e5c2f31` documents it at the declaration:

> The first import after this change rewrites each declared Space **once**, and that is intended rather than overlooked: the owner-side upsert replaces the whole `Content`, so a stamp a previous import invented is dropped. **Nothing real is lost** — that value was a *different clock reading on every previous import*, which is the defect — and after that one write the node is stable forever. A Space whose declaration **does** carry `createdAt` keeps it, on that import and every one after.

`ADeclaredCreatedAt_IsPreserved` already pins the second half.

Verified after the change: `-c Release -warnaserror` on `MeshWeaver.Graph` → `0 Error(s)`.

Issue #4373, opened 2026-09-15
Notifications are written as rendered English strings, so a German viewer reads every bell row in English


rbuergi:
## What is wrong

`NotificationService.CreateNotification(…, string title, string message, …)` takes **rendered
strings**, and every call site in `src/` passes an English literal:

| Call site | Example |
|---|---|
| `PackageUpdateReconciler.Notify` (3 kinds: update-available, authorization-required, sync-owned-partition) | `"Update available: {name}"` |
| `ModuleDiscoveryService` | a module-discovery reminder |
| `StaticRepoImporter` | an import reminder |
| `NotificationService` itself (the fan-out at line 309) | — |

The portal ships English + German, so a German viewer reads every bell row in English. Per
`Doc/Architecture/Localization` and the ownership rule in AGENTS.md, platform-owned text follows the
**viewer** — and a notification's viewer is not known when it is written.

## Why the obvious fix is not one

"Call `accessService.Localize(...)` at the call site" does not work here, and doing it would look
like a fix while changing nothing:

- A notification is raised on a **background reaction** — the package-update reconciler, the module
  discovery scan, an import. There is no viewer, and the writer runs as SYSTEM. Resolving a key
  there picks the system default (English), so the German viewer still reads English.
- It also **bakes** one language into a durable node. The row outlives the write by days and can be
  read by several people with different locales.

## The shape that would work

The notification's content has to carry the **key and its arguments**, not the rendered sentence, and
the bell has to render it against `AccessContext.Locale` at read time — the same split the platform
already uses for activity logs (`LogMessage.WithKey(...)` + `LogMessageLocalizationExtensions.Localize(message, accessService)`).

Sketch:

- add `TitleKey` / `MessageKey` + arguments to the `Notification` content (additive; an older row
  with only the rendered strings keeps rendering them, so nothing has to be migrated);
- give `CreateNotification` an overload taking `LogMessage`-shaped values;
- render in the bell / notification list through the existing localization extension;
- move the five call sites over, adding the keys to
  `src/MeshWeaver.Messaging.Hub/Localization/strings.{en,de}.json` — and the React mirror
  (`MeshWeaver.Plugins/clients/react/src/i18n/`), core first.

## Where this came from

Copilot's review of #4364 (MeshWeaver#4355) raised it against that PR's new
`sync-owned-partition` hold reminder. That reminder is one of four literals in the same method and
the fourth instance of a surface-wide gap — localizing one of them would leave the surface
[...]

Issue #4348, opened 2026-09-14
MeshWeaver.Plugins PRs compile against the set its own main last PASSED — when main is red because the set moved, the fix cannot be proven (the five satellites do NOT follow main)


rbuergi:
A satellite's PR lane compiles against **the newest sealed set its own `main` has PASSED**, not the newest sealed set. That is deliberate and right in the normal case — `pull requests follow main` — but it produces a deadlock in exactly the case where a fix is most urgent: **when `main` is red *because* the set moved.**

## Measured today, on the PR fixing the fleet's publication outage

`MeshWeaver.Plugins#1873` exists to fix two suites that core changes broke on arrival in set **3.0.0-ci.8596**. Its own platform job says:

```
skip main-cd #8596 …: sealed, but this repo's `main` has not passed on it yet
                      (main's newest passed set is core CD #8587)
set=3.0.0-ci.8587
```

The resolver is behaving exactly as documented (`resolve-platform.py:1016-1019`). The consequences, all measured on that PR:

1. **The PR lane cannot exercise the regression the PR exists to fix.** It builds and runs against the set *before* the break, so it can go green having proven nothing about the set `main` will actually use.
2. **It is worse than uninformative — it is hostile to the clean fix.** The first attempt used the API the breaking core change introduced (`PackageInstaller.SetUpdatePolicy`, which is what the catalog card itself calls). The PR went **red**: `CS0103 PackageUpdatePolicy`, `CS0117 SetUpdatePolicy`, `CS1061 EffectiveUpdatePolicy` — the set it compiles against has none of those symbols.
3. **So a fix for a set-move regression must compile on the OLD set AND pass on the NEW one, and may name no symbol the new set introduced.**

For #1872 that was resolvable — the fix builds the record's legacy shape as JSON (`content["autoUpdate"] = true; content.Remove("updatePolicy")`), which names nothing new **and is the more honest test**, since that shape is exactly the population production holds: every record installed before the change. CI offers one oracle, so both were taken locally against two core worktrees — set 8587: 9/9; core main: 9/9, `PluginCatalog.Test` 734/734, `FutuReAnalysisTest` 46/46.

**But that was luck of this particular break.** A set-move regression whose only honest fix *requires* the new API has no green path at all: the PR cannot compile, and `main` cannot go green until the PR merges.

## Why the current design is still right

[...]

PR #4397, opened 2026-09-15
feat(#4373): a notification carries its catalog keys, so the bell can speak the viewer's language

rbuergi:
Steps 1 and 2 of #4373's own plan. `Notification` can now carry `TitleKey`/`TitleArgs` and `MessageKey`/`MessageArgs`, `CreateNotification` accepts them, and `NotificationLocalizationExtensions` resolves them at **render** time — deliberately the same shape as `LogMessage.WithKey` + `LogMessageLocalizationExtensions` (#3236), so a reader who knows one knows this.

## Why the key and not the sentence

Every notification the platform raises comes from a background reaction — the package-update reconciler, a module-discovery scan, a startup import — running as **SYSTEM with no viewer in scope**. Resolving a catalog key there picks the system default (English) and then bakes it into a durable row that outlives the write by days and is read by people whose languages differ.

`ModuleDiscoveryService` already calls `accessService.Localize(...)` at its write site, and that is precisely why its rows are English for everybody. This is the shape that fixes it.

## Additive, and safe on every existing row

The four fields are nullable. `Title` and `Message` stay the stored English and remain the **fallback**, so:

- a row with no key renders exactly as before (every row written before this);
- a key that has since left the catalog still renders the sentence the writer meant, never a raw `plugins.…` token.

A renderer may therefore switch to `LocalizeTitle`/`LocalizeMessage` unconditionally, and nothing has to be migrated.

**The node's `Name` stays English on purpose** — it is the row's identity in listings, search and logs, and a per-viewer value there would make two readers disagree about what the same node is called. The bell renders off the *content*.

## What this does not do

- **The renderer** is `MeshWeaver.Blazor.Portal/Components/NotificationCenter*` in MeshWeaver.Plugins. It cannot be written until a sealed platform set carries these fields (the satellite ceiling, #4348), so the core half lands first — the ordering this fleet already uses for `Implementers:` and `Mirror-sync:`.
- **The call sites are not migrated.** `plugins.discovery.*` uses **positional** `{0}`/`{1}` placeholders while `GetNamed` binds **by name**, so adopting them means rewriting those templates in both catalogs — a change with its own mirror obligation. Migrating some sites and not others is exactly the inconsistency #4373 warns about, so the surface moves in one later step rather than partly here.

No catalog key is added or re-worded (no `Mirror-sync:` owed), no public member is removed (no `Pairs-with:` owed), no interface gains a member.

## Verification

- `dotnet build -c Release -warnaserror` on `MeshWeaver.Mesh.Contract`, `MeshWeaver.Graph`, `MeshWeaver.PluginCatalog` → `0 Error(s)` each
- `NotificationCarriesItsKeysTest` → **6 passed**, covering: an un-keyed row unchanged in every locale; keys and named args stored; a **real two-viewer case** on the existing named-placeholder key `activity.compile.started` — one stored row renders `Compile started for Acme/Thing` in `en` and different text in `de`, with `Message` itself untouched; a missing key falling back to the stored English; scalarized arguments surviving the JSON round-trip; a blank key changing nothing.


rbuergi:
Closing as a duplicate — **#4373 was already fixed and merged before I started this**, and I did not check.

`4ea1f4b799` — *"fix(#4373): notifications carry their key, so the bell renders in the reader's language"* — landed on `main` at **08:17:44Z**. I created this branch at ~08:46Z. Same four fields (`TitleKey`/`TitleArgs`/`MessageKey`/`MessageArgs`), same extension shape, same reasoning about a `Localize` call with no viewer being a rendering rather than a localization — and **theirs also migrates the call sites**, which mine deliberately did not. It is strictly the better change.

The process error is mine and worth naming precisely: I checked that **issue #4373 was open** and treated that as "unclaimed". An issue's state does not track its fix — the fix had merged and the issue was closed out of band a few minutes later. The check I should have run is `git log origin/main --oneline -- <the files the fix would touch>` before writing a line, not `gh issue view`.

Two review findings on this PR were valid regardless and apply to what landed, so I am not dropping them:

1. **Appending optional parameters to `CreateNotification` is a binary break** — optional arguments are substituted at the call site, so an already-compiled consumer still looks for the old signature and gets `MissingMethodException`. Worth checking whether `4ea1f4b799` extended the existing method or added an overload; if the former, a compatibility wrapper is cheap now and expensive later.

2. **A test documented as covering the JSON round-trip that never serializes** — mine asserted "arguments are scalarized so they survive the JSON round-trip" while every assertion read the pre-serialization object. A serializer regression would have kept it green. That is this repository's dominant defect class in my own test, and if the landed change has a similar one it deserves the same fix.

I have raised both on #4373's thread rather than leaving them in a closed PR.

rbuergi:
Correction to my own closing note: **I checked the landed change against both findings and neither applies.** I said I would raise them on #4373; I am not going to, because there is nothing to raise.

1. **Binary compatibility** — `4ea1f4b799` keeps the ten-parameter `CreateNotification` **verbatim** and forwards it to a new `CreateLocalizableNotification` with `LocalizableText.Verbatim(...)`. That is exactly the compatibility wrapper the finding asked for, done before the finding existed.

2. **The JSON round-trip** — `NotificationLocalizationTest` writes through a real mesh and reads back with `node.ContentAs<Notification>(Mesh.JsonSerializerOptions)`, so `TitleArgs` genuinely survives storage. That is a stronger control than the one I wrote, which never serialized at all.

They also added `LocalizationTest.PersistedNamespaces_UseNamedPlaceholders` — a guard refusing a positional placeholder in a persisted namespace, which is the catalog half I had deferred as "a separable step with its own mirror obligation".

So the duplicate was mine in full: later, narrower, and weaker on both points the review raised against it. Leaving this note so the closing comment above is not read as an outstanding concern on a change that does not have one.

PR #4418, opened 2026-09-15
shared-rules: never-hand-roll-ui required in all seven repos — step 3 of 3

rbuergi:
Step 3 of the `never-hand-roll-ui` rollout (maintainer, 2026-09-15: *"see also in all the Agents.md that we make clear to always use this and not create own ui frameworks"*).

- Step 1: Systemorph/MeshWeaver#4412 (merged) — the block wrapped in core and registered hub-only, so no spoke marker was ever undeclared.
- Step 2: Systemorph/MeshWeaver.Plugins#1916, Systemorph/MeshWeaver.Education#337, Systemorph/MeshWeaver.Reinsurance#211, Systemorph/MeshWeaver.SocialMedia#190, Systemorph/MeshWeaver.Manufacturing#91, Systemorph/Memex#363 — the block verbatim in each spoke.
- **This PR:** widens `required-in` to all seven, so a repo that drops or drifts the block is red in core's shared-rules check from here on.

Also answers Copilot's finding on #4412: the register's own `$schema-note` said *"wrap every listed repo … land it as ONE change set"*, which contradicts the gate (red both for a listed repo with no markers and for a marker whose id the register does not declare). The note now states the three steps, matching `SharedRuleBlocks.md`.

**Draft until all six spokes are on their default branches** — the PR half of the gate reads them from there.

Verified locally: `check-shared-rules.py --self-test` 18/18; `--local` for all seven repos green (`✓ 6 spoke(s) match the hub` for `never-hand-roll-ui`).

Pairs-with: none — the register only; no public surface changes.


rbuergi:
Refreshed and taken out of draft. What was actually wrong, measured rather than assumed:

**The red was two days stale, not a real drift.** The last run was 2026-09-15T10:53Z and named Reinsurance, SocialMedia and Manufacturing as missing the block. All three step-2 pull requests (Reinsurance#211, SocialMedia#190, Manufacturing#91) merged this morning at 05:58–06:00Z, so the run was reporting a fleet state that no longer exists. The stated hold in the body — *"draft until all six spokes are on their default branches"* — is therefore met: I read all six default branches and each carries `<!-- shared-rule:begin never-hand-roll-ui -->`.

**Replayed the gate against the live fleet before touching anything**, with this PR's widened register and the six spokes' current `AGENTS.md`:

```
── never-hand-roll-ui  (hub: Systemorph/MeshWeaver; 7 repos; slots: ['reference'])
     ✓ 6 spoke(s) match the hub
✅ All 5 shared rule block(s) are identical everywhere they are carried, across 7 repos.
```

**`MeshWeaver.Crm` is NOT a blocker here, contrary to a diagnosis going round.** It carries no `never-hand-roll-ui` markers, but it is absent from the register's `repos` list, so this gate never reads it and nothing in this PR requires it. Registering Crm is a separate change that needs the read-only `fleet-reader` App installed there *and* Crm carrying the blocks already registered — an unreadable repo is a RED in this gate by design.

**Merged `main` in — necessary, not tidy.** The branch was 439 behind, and core#4641 has since registered a second block, `dispose-to-serve-new-state`. Its step-2 spoke pull requests are open now; the moment one lands, a spoke would carry a marker whose id this branch's stale register does not declare, which the gate reds on. The merge resolved cleanly and both blocks are intact on the branch (`never-hand-roll-ui` → 7 repos, `dispose-to-serve-new-state` → 1).

One thing that changed since this was written: `Automatic review answered` is now a **required** check, so the automatic review's findings have to be answered before this can merge.

PR #4424, opened 2026-09-15
docs(#4414): xUnit1069 is a warning by default; Plugins' tests made it an error

rbuergi:
Corrects one paragraph that #4419 added to `DependencyMajorUpgrades.md`.

That paragraph said `xUnit1069` is an **error by default** under xunit.v3 4.x. It is a **warning**. The error I measured came from MeshWeaver.Plugins' `src/Directory.Build.props`, which applied `TreatWarningsAsErrors` to its `*.Test` projects. Core's own test projects never had that policy.

MeshWeaver.Plugins#1920 aligned the Plugins test projects with core. After that, main-cd run 34964034139 built `MeshWeaver.AI.Test` with the advisories as warnings and ran the suites green. The paragraph now says so, and keeps the point that a timed test ignoring the token still needs the token threaded into its waits.

Pairs-with: none — documentation only.


rbuergi:
Correction: the table-name fix for #4415 cannot ride here, because the row it edits exists only once #4415 merges. It will follow as its own PR after #4415 lands.

Issue #3945, opened 2026-09-10
GitSync: a source that never converges re-clones its WHOLE repository on every green build — `lastSyncCommitSha` is both the cheapness gate and the convergence guard


rbuergi:
## Summary

`GitHubWebhookProcessor.SkipReason` makes a green-build delivery free for exactly one reason —
`lastSyncCommitSha == headSha` — and `GitHubSyncService.MayAdvanceBaseline` deliberately **holds**
that field whenever an import did not fully converge (`Preserved > 0`, `Failed > 0`, or outcome
`Failed`). Both behaviours are individually correct and each has an incident behind it (#675/#677
for the horizon, #2229 item C for the sha). Together they mean:

> **A sync source that cannot converge pays a full `git fetch --depth 1` of its entire repository,
> plus a full parse, on EVERY publish-signal delivery — and the rate is the source repository's CI
> cadence, not anything about the source.**

The allow-list that admits `schedule` and `repository_dispatch` states the opposite as its
justification, in `PublishSignalTriggers`:

> **Widening this cannot cause churn.** A sync source already sitting on the built sha is skipped by
> `SkipReason` ("already at this commit"), so a scheduled or dispatched re-verification of an
> unchanged default branch triggers no import at all.

That invariant holds for every source that converges and for no source that does not — and the
sources that do not converge are precisely the ones whose imports are most expensive.

## Measured

**Trigger rate** — `Systemorph/MeshWeaver` `main`, 24 h ending 2026-09-10T17:35Z: **254 green
publish-signal runs** (`gh api repos/Systemorph/MeshWeaver/actions/runs?branch=main&status=success`),
of which 87 `Prod synthetic probe` (`schedule`, `cron: "*/15 * * * *"` — an HTTP probe that builds
nothing), 47 `Chart Gate`, 47 `Hosting Operator`, 46 `MeshWeaver Build and Test` (all `push`), 19
`Continuous Delivery (main)`, 8 other. Independently confirmed by the node the webhook writes:
`Admin/_Build/Systemorph.MeshWeaver` on memex.meshweaver.cloud read **version 4778** at
2026-09-10T17:21Z, first written 2026-08-06T07:38:37Z — **4,775 deliveries in 35 days**, 256 in the
last 24 h (10.7/h). So one merge is ~3 deliveries at the same sha, plus 96/day from a cron whose
`head_sha` does not move between merges.

[...]

Issue #3146, opened 2026-09-02
StaticRepoImport: a FAILED import at fingerprint F is re-run in full on every trigger — memex-cloud re-imports the core samples every ~10 min (425 failing upserts + NodeType compiles per pass)


rbuergi:
## A static-repo import that FAILED at fingerprint F is re-run in full on every trigger at the same F — memex-cloud re-imports the core repo's samples every ~10 min, 425 failing upserts + sample NodeType compiles per pass

**Measured on memex.meshweaver.cloud (image ci.7231), 2026-09-02 15:40–18:40Z, Loki `namespace=memex-cloud`:**

- `[StaticRepoImport] MeshWeaver: upsert of MeshWeaver/samples/Graph/Data/… failed (continuing)` — **19 full passes in 3 h**, each ≈425 `InvalidOperationException`s: `A 'Space' owns its partition, so it must be top-level`, `A 'User' owns its partition…`, `NodeType 'Northwind/Article' is not registered`, `NodeType 'Systemorph/Marketing/Post' is not registered`, `Node must have a NodeType or Content set`. Same 425 nodes, same errors, every pass.
- Each pass also compiles the sample NodeTypes (`CS1591 Warning … MeshWeaver_samples_Graph_Data_Type_articleMeshNodeProviderAttribute`, `…Northwind_ProductMeshNodeProviderAttribute`, 110 compiler lines per pass on one pod) and writes `…/_Activity/compile-state`, which then hits `[UpdateQueue] ADVANCE_WITHOUT_HANDOFF … bound=5000ms` and `LATE_NACK_TERMINAL … TimeoutException` on the loaded pods.
- The source is an `InMemoryStaticRepoSource` minted by `GitHubSyncService` (GitSync of `Systemorph/MeshWeaver` into partition `MeshWeaver`); the trigger is the core repo's `workflow_run` webhook to `https://memex.meshweaver.cloud/webhooks/github` — every green core CI completion re-imports. Core main merged ~30 times today.
- `StaticRepoImporter` stamps `ActivityStatus.Failed` for the run (correct) but the next trigger at the **unchanged** fingerprint re-imports everything again: the skip arm ("already at {Fingerprint} — content skipped") only fires on a *Succeeded* stamp. A source that fails deterministically is therefore re-imported forever, at full cost, on every trigger.

[...]

Issue #3101, opened 2026-09-02
A Space whose content sync is refused stops syncing SILENTLY — AgenticEngineering's assets are over the frame limit outright


rbuergi:
## A Space whose assets stopped syncing, silently

Found while fixing #2885 (the import-hub OOM), and **not the same defect** — #2885's producer fix does not close this one, so it needs its own tracking.

`MeshWeaver.Education`'s `AgenticEngineering` Space holds **106,070,300 bytes** under `content/**`. A GitSync content sync carries those binaries **inline**, and base64 inflates them:

| Space | `content/**` | base64 form | verdict |
|---|---|---|---|
| `AgenticBusiness` | 28,484,421 B | ~38 MB | under `MaxMessageBodySize` — this was #2885's OOM |
| **`AgenticEngineering`** | **106,070,300 B** | **~141 MB** | **over the frame limit outright** |

So once the oversized-delivery refusals landed (#1890 / #2897 / #3018 / #3032), that Space's content sync began being **refused rather than delivered**. The assets stopped syncing — and nothing said so.

## Why this is worse than an OOM

An OOM is loud: it kills a pod, files an incident, and someone looks. A refusal is quiet by design — the refusal is the *correct* behaviour of the guards, working exactly as intended. There is no error surfaced to whoever authored the content, no ledger entry saying "this Space's assets are stale", and nothing that distinguishes "synced and unchanged" from "refused every time since the guards landed".

The person who notices is a learner opening a course page whose video or image is missing.

## Relationship to #2885 / #3097

#3097 partitions the sync so a delivery is `≤ budget + largest single file`, which fixes `AgenticBusiness` and every Space whose *aggregate* was the problem. It does **not** fix this one on its own if any **single file** in `AgenticEngineering` is itself over budget — #3097's own residual, recorded in `Doc/Architecture/OversizedDeliveryRefusal`:

> a single file over budget still travels whole — a different axis, belonging behind a content-store handle.

So the two questions this issue should answer, in order:

1. **Does `AgenticEngineering` now sync after #3097?** If its files are individually small and only the aggregate was over, #3097 already fixed it and this issue closes with a measurement. That is the cheap check and it should be done first.
2. **If not, which files are individually over budget?** Those need the content-store handle rather than an inline payload — a real design change, not a bound.

## The part that should be fixed regardless

[...]

PR #4467, opened 2026-09-16
fix: a content verdict is final for the NODE that earned it, never for the partition (#4459, #4456)

### LINKED ISSUE #4456: A proven-commit import applies modifications but not additions — the partition is left referenced-but-incomplete, and a force update imports 0 nodes
## What is failing

A GitSync import at a proven commit applies **modifications** to a partition but does not **create** files the same commit adds. The partition is then referenced-but-incomplete, its NodeTypes cannot compile, and the bake gate holds readiness — so the instance cannot roll. A forced re-import does **not** heal it: it imports **0 nodes**.

Measured on memex.systemorph.com, 2026-09-15, rolling onto sealed set `3.0.0-ci.8692` (core `3cd1935709`, plugins sealed at `85515f62`).

### The import

`Hosting/_Activity/10f16763` — *"Update Hosting to the built commit 85515f62"*, 19:00:52 → 19:01:48:

```
Recompiling 8 NodeType(s): Hosting/Admin, Hosting/Deployment, Hosting/Backup, Hosting/FleetConsole,
                           Hosting/InstanceAction, Hosting/InstanceRequest, Hosting/Issue, Hosting/PlatformBuildInbox
Adopted 8 prebuilt assembly(ies) — the rest compile.
Pruned 118 node(s) absent from the repo: Hosting/Triage/ci-failure/…, Hosting/Triage/ci-green/…, Hosting/Triage/Status
Re-imported ImportedWithErrors (41 node(s)) at 85515f62.
```

`Hosting/_GitSync` afterwards: `lastAttemptedCommitSha: 85515f62`, `lastSyncOutcome: ImportedWithErrors`, `lastSyncCommitSha` **still `c7f03952`**.

### What landed and what did not

In `Hosting/Deployment/Source/`, that import **updated** two existing nodes — `PlatformBuildInboxWatcher` (v14) and `HelmValues` (v8), both `lastModified 2026-09-15T19:01:12Z`. In the same commit range it **did not create** `Hosting/Deployment/Source/SelfUpdateRouting`. The partition holds 13 Code nodes in that folder and it is not among them.

`SelfUpdateRouting.cs` is present at `85515f62` (`git ls-tree`), and at that same commit it is referenced by `Hosting/Deployment/Source/PlatformBuildInboxWatcher.cs` — the very file the import **did** apply — and by `Hosting/PlatformBuildInbox/Test/PlatformBuildInboxTests.cs`. So the import wrote the reference and skipped the referent.

🚨 `Hosting/` is **byte-identical between `85515f62` and today's `main`** (`git diff --stat 85515f62 origin/main -- Hosting` is empty). This is not a "newer content" or seal-ordering problem: the importer had one consistent tree and applied it partially.

### The consequence

On the 3.0.0-ci.8692 replica (`memex-portal-deployment-6546c779d4-4ngqv`):

```
CS0103 Error (line 6362): The name 'SelfUpdateRouting' does not exist in the current context
CS0246 Error (line 7308): The type or namespace name 'SelfUpdateRouting' could not be found
DynamicTypePreWarmer: REFUSING READINESS — 5 NodeType(s) regressed on this image:
  Essentials/Mailbox, Hosting/InstanceAction, Hosting/Issue, Hosting/ModuleReportInbox, Hosting/PlatformBuildInbox
BuildProtocol: NOT publishing GO for s4b2836629f0c58ad4ed0aa683190981a — 5 gating regression(s)
```

The gate behaved correctly: the previous replicas kept serving and there was no outage. But the rollout cannot complete, and the helm-release deploy failed at *"Observe the rollout until it completes"* after 28 minutes.

### The repair path does not repair

`git_hub_sync update force: true` on `Hosting` (branch HEAD = `main`, whose `Hosting` tree is identical to the sealed commit) — `Hosting/_Activity/89ae5941`:

```
Pruned 6 node(s) absent from the repo: Hosting/Triage/…
Imported ImportedWithContentErrors (0 node(s)).
```

**Zero nodes imported**, and the missing file is still missing. A force update is documented as "overwriting local changes"; it cannot create the absent node either. That is what makes this a hard stop rather than an inconvenience: there is no supported operation that restores the partition.

### Same shape, second partition

`Essentials/Mailbox` fails identically on `MailboxLiveTests` (`CS0103`/`CS0246`, 7 sites): the referencing test-area file is in the mesh, `Essentials/Mailbox/Test/MailboxLiveTests.cs` is not. Unlike Hosting, `Essentials/` on `main` is 10 files ahead of `85515f62`, so that partition additionally carries content newer than the seal — the two-writer mix of #4355 — but the immediate failure is the same missing-addition shape.

## Why this is worth fixing as its own defect

#4355 is about two writers disagreeing over one partition. This is narrower and, I think, more serious: **one importer, one consistent tree, and additions silently dropped** — with `ImportedWithErrors` / `ImportedWithContentErrors` as the only signal, and no per-node error in the activity log naming what failed. The outcome string is on the sync config where nothing reads it; the activity's `maxSeverity` is `Information` and its `status` is `Succeeded`.

Suggested, in order of value:

1. **An import that could not write a node must say which node and why**, at Warning or Error, in the activity log — `ImportedWithErrors (41 node(s))` names neither the failures nor the count of them.
2. **A force update must be able to create a missing node.** Today it imported 0 while a referenced file was absent from the partition.
3. **A partition left referenced-but-incomplete should not be handed to the compiler**, or the compile failure should name the import that produced it — the operator's visible symptom is `CS0103` on a symbol whose file is in git.

## Evidence

- `Hosting/_Activity/10f16763`, `Hosting/_Activity/89ae5941`, `Hosting/_GitSync`, `Essentials/_GitSync` on memex.systemorph.com
- pod `memex-portal-deployment-6546c779d4-4ngqv` (3.0.0-ci.8692), 18:56Z onwards
- Memex deploy run 35010439754 (failed at *Observe the rollout*), `Ops/Actions/helm-memex-20260915-deploy-8692`
- `git ls-tree -r 85515f62 -- Hosting/Deployment/Source/SelfUpdateRouting.cs`; `git diff --stat 85515f62 origin/main -- Hosting` (empty)

### LINKED ISSUE #4459: GitSync: a partial import records the content fingerprint anyway, so five Hosting NodeTypes are parked on a source node the import dropped and re-import is skipped
**Measured on memex.systemorph.com, 2026-09-15 20:30–20:40Z.**

A `Hosting` GitSync import dropped exactly one source node, reported `ImportedWithContentErrors`, and then **recorded the content fingerprint anyway**. Five NodeTypes are parked as a result, and no further import can repair it: the next `update` is skipped as already-imported.

**The missing node.** `Hosting/Deployment/Source/SelfUpdateRouting` exists in no partition on memex (`search partitions:all name:*SelfUpdateRouting*` → 0). The repository has 14 files under `Hosting/Deployment/Source/`; memex holds 13 nodes. The only absentee is `SelfUpdateRouting.cs`, added to Plugins main at 11:22Z (36ee62f3d) and carrying the same `// <meshweaver>` header as its siblings.

**What that breaks.** `PlatformBuildInboxWatcher`, which calls it, DID import, at 19:01Z. So five NodeTypes fail to compile, each on the symbol that is missing:
```
CS0246: The type or namespace name 'SelfUpdateRouting' could not be found
CS0103: The name 'SelfUpdateRouting' does not exist in the current context
```
Parked: `Hosting/Backup`, `Hosting/Deployment`, `Hosting/InstanceAction`, `Hosting/InstanceRequest`, `Hosting/PlatformBuildInbox`. **No instance action can run on memex while `Hosting/InstanceAction` is parked** — that is Sample, Logs, HelmRelease and every provisioning action on the control instance.

**Why it cannot heal.** `Hosting/_GitSync` still reports `lastSyncedAt` 2026-09-14T22:18Z and `lastSyncCommitSha` c7f03952: the outcome is not a success, so the recorded commit never advances and each attempt re-imports the same delta and fails the same way. A second `update` at 20:37Z answered:
```
Imported Skipped — an earlier FULL import already recorded this exact content at
fingerprint bb9801101e859a21 (Hosting/_Activity/import-bb9801101e859a21), so the
partition was not re-read.
```
So the fingerprint marker claims completeness for an import that lost a node, and the operator's one repair (re-import) is refused.

**The defects, separable.**
1. An import that could not write every node records a content fingerprint that says it did. A partial import must not mark the content complete.
2. `ImportedWithContentErrors` names a count (9), never the paths. The nodes that failed should be named, or nobody can tell what is missing.
3. A NodeType parked on a symbol whose source the import dropped cannot be repaired by Compile: the source is not there to compile.

**How to reproduce the state.** `get @Hosting/_GitSync`, `get @Hosting/_Activity/6ac3721c` (the 20:31Z import at Plugins 04a14676), and `get_diagnostics @Hosting/InstanceAction`.

Found while carrying MeshWeaver.Plugins#1935 to memex; the park predates that merge.

rbuergi:
Fixes #4459
Fixes #4456

Pairs-with: none — this diff removes no public type and no public member from src/. It ADDS one public record (FailedImport) and one init property (StaticRepoImportResult.FailedPaths); the only signature change is to the private StaticRepoImporter.WriteManifest.

Mirror-sync: npm run sync:i18n -- --ref <this PR's merged core sha> — two keys added (activity.import.itemRefusedBefore, activity.gitsync.failedNodes), no existing value re-worded.

Implementers: none — no member is added to any public interface or abstract class.

## The root cause

On 2026-09-15 one file — `Hosting/Deployment/Source/SelfUpdateRouting.cs` — carried a **literal NUL byte (0x00)** inside a char literal. PostgreSQL cannot store a NUL in a text column, so the Hosting GitSync import refused that ONE row while forty others landed.

A pass whose failures were *all* deterministic earns `ImportedWithContentErrors`. That word means **"every failure was a verdict about the bytes"**, not "everything failed" — but it was recorded on the **content-addressed marker**, whose whole job is to say something about the **partition**. Every later import read it and answered:

```
Imported Skipped — an earlier FULL import already recorded this exact content at
fingerprint bb9801101e859a21 (Hosting/_Activity/import-bb9801101e859a21), so the
partition was not re-read.
```

A claim about forty nodes, made on the evidence of one, and false: the import lost a node. Five Hosting NodeTypes stayed parked on `CS0246 SelfUpdateRouting` — among them `Hosting/InstanceAction`, so **no instance action would run on the control instance at all** (no Sample, Logs, HelmRelease or Roll) — and the operator's one repair, a re-import, was refused with that same sentence.

**So #3146's rule is right and its GRANULARITY was the defect.** Re-issuing a write the owner refused cannot help — that is real, and it cost memex-cloud 19 full passes in 3 h. But a content verdict is earned by a **node**, and recording a per-node fact as a per-partition verdict is what turned one unstorable byte into a frozen partition.

#4456 reads the same evidence as *"applies modifications but not additions"*. It is the same single dropped file: nothing named it, so the shape of the loss had to be inferred. Its *"a force update imports 0 nodes"* is the same file being refused again — correctly, on bytes that genuinely cannot be stored.

## The fix

**1. The memory moves to the per-node import manifest.**

`WriteManifest` may only claim what the run **wrote**. A refused node used to be recorded under its plain source token — the ledger asserting the partition holds content that is provably not in it, which is the marker's false claim one layer down, and it had to become honest before the marker's skip could be retired.

| what happened to the node | what the manifest now records |
|---|---|
| written | its source token |
| refused, deterministically | `!` + its source token |
| failed, retryably | nothing at all — which is what makes the next pass look again |

The `!` prefix is deliberately a prefix and not a schema change: a reader that does not know it compares `!abc` against the token it computes (`abc`), sees a mismatch, and re-evaluates the node — the safe direction, and exactly the behaviour before the field existed.

**2. The skip moves with it.** The import loop skips a node whose manifest entry is a refusal *of this exact token* — reported, no write attempted — and `ImportedWithContentErrors` stops being a fingerprint-recording outcome. The marker keeps it as a **record**; it is no longer a licence to answer `Skipped` without reading the partition. `Force` and `Reconcile` bypass the refusal memory exactly as they bypass the marker.

Both properties then hold at once: the failing write is not re-issued (no storm), and every other node is still evaluated (no freeze) — so a partition that has drifted from its marker can be repaired, which is the repair the whole-partition skip refused.

*Cost:* a partition holding a refused node re-evaluates on every trigger instead of short-circuiting — **no node writes and no recompiles**, just one subtree read, the prune evaluation and two bookkeeping writes. That is the same cost the marker already accepts for an absent or non-converged verdict.

**3. An import that loses a node NAMES it.** Every other thing a pass leaves behind already named its paths (written, pruned, blocked creates, held NodeTypes, refused content); failures were the one category reported as a bare number. `StaticRepoImportResult.FailedPaths` now carries each failed node with the write path's own reason and whether it was deterministic — #3101's argument, for nodes this time. The terminal summary names them, and the GitSync activity logs them as their own line.

**4. A partial import no longer renders green.** Every outcome line in `GitHubActivityExtensions` was written at `Information` whatever the import did, and `RunActivity` derives the activity's `maxSeverity` and terminal `status` from those levels — which is how the import that lost `SelfUpdateRouting` reported `maxSeverity: Information`, `status: Succeeded`. The level now follows the outcome: `Failed` → error, any `ImportedWith…` → warning.

## The control

Both new behavioural tests use **only pre-existing API**, so the fix could be reverted and the suite still compiles.

Fix reverted (`src/MeshWeaver.Graph/StaticRepoImporter.cs` + the GitSync half back at `main`), test project **rebuilt**, run — **3 failed / 2 succeeded**, with the defect printed verbatim:

```
outcome=ImportedWithContentErrors count=1 failed=1 writeRequests=3 written=[Rf9a6279eb/Good]
[StaticRepoImport] Rf9a6279eb already FAILED at b7b778c951b8aabb on content rules — skipping.
outcome=Skipped count=0 failed=0 writeRequests=0 written=[]
=== TEST FAILED: Did not expect value to be "Skipped" …
```

Fix restored, rebuilt, re-run — **6 succeeded / 0 failed**.

| suite | result |
|---|---|
| `ARefusedNodeDoesNotFreezeThePartitionTest` + `FailedImportIsNotRetriedAtTheSameFingerprintTest` | 6/6 |
| `MeshWeaver.Graph.Test` (full) | **1717/1717** |
| `MeshWeaver.Hosting.Test` (full) | **728/728** |
| `MeshWeaver.Documentation.Test` (full) | **476/476** |
| `LocalizationTest` | **60/60** |

`dotnet build -c Release -warnaserror` clean (`0 Warning(s), 0 Error(s)`) on `MeshWeaver.Graph`, `MeshWeaver.GitSync`, `MeshWeaver.Documentation` and each test project. `check-record-signatures.py` and `check-type-forwards.py` pass locally against `origin/main`.

🚨 **#3146's own pins changed deliberately.** They asserted the word `Skipped`; they now assert `WriteRequests == 0` — the quantity that storm was actually made of (≈425 failing upserts plus a compile, per pass). A word can stay right while the writes come back; a request count cannot.

## 🚨 What this deliberately does NOT fix

Both issues list a **third** item, and it is the same one: a NodeType parked on a symbol whose source the import could not write still fails on `CS0246` for a file that is plainly in git, with nothing in that error connecting it to the import. Split out as **#4469** rather than left implied — closing two issues while one of their listed defects stays live, without saying so, is the failure mode where an answer reads like a pass.

What changed is the **other direction**: an operator reading the sync activity now learns which file is missing and why, and the activity is no longer green. The facts #4469 needs are already recorded here (the manifest's `!<token>` entries say which declared nodes a partition is missing), so it is a wiring problem, not a measurement one — with the granularity trap noted in it: scoping a compile refusal to a whole partition would re-create exactly the over-broad reading this PR removes.

Also unchanged on purpose: a remembered refusal is still reported as a failure, so `MayAdvanceBaseline` still **holds** the git baseline (the mesh genuinely is not at that commit), and `VerdictIsFinal` is untouched, so #3945's cheap-delivery path is unaffected.

## Work products

- `Doc/Architecture/AContentVerdictIsPerNode` — the incident, the granularity argument, the manifest encoding, the cost, the residue (#4469) and what is deliberately unchanged. Linked from `Architecture.md` and from `ImportMarkerRecordsConvergence.md`.
- #4469 — the compiler-side half, filed rather than implied.


rbuergi:
## Review response

Thanks — two of these were real and are fixed, two were already fixed in a commit you did not see, and the critical one I am declining with reasoning. Detail per finding.

### ✅ Fixed — `StaticRepoImport.md` still promised the old contract (suppressed comment, AContentVerdictIsPerNode.md:101)

Correct, and the most valuable finding here. The canonical page's outcome table still read *"`ImportedWithContentErrors` → `Failed` → **skips**, logging the recorded verdict"* — the two pages were promising opposite behaviour for the same outcome, which is the two-homes-drift failure. The row now reads *"re-imports, but issues no write for the refused node"*, with the granularity argument and a pointer. (`3a40180c6f`)

### ✅ Fixed — the test's existence gate was a point query (ARefusedNodeDoesNotFreezeThePartitionTest.cs:239)

Correct, and it is a house-rule violation independent of this test: existence of a specific path is not a valid query use, because a point read of an absent node is a routing NotFound that terminates the stream and opens the storm-breaker on that path. It is now a `scope:children` listing of the parent, `.Complete()`.

🚨 And your underlying worry — *"this wait can complete while the import still sees `Good`"* — deserved more than a shape change, because the new shape has the mirror-image vacuity: waiting for a node to **leave** a listing is satisfied instantly by a listing that never contained it. So the test now asserts the listing **sees** `Good` before the delete, and that guard is measured, not assumed: with the behavioural fix disabled the suite still fails on the real assertion (`Did not expect value to be "Skipped"`), never on the new guard.

### ✅ Fixed — duplicate source paths (suppressed comment, StaticRepoImporter.cs:3205)

Correct. A path that **also landed** is no longer recorded as refused: the written set is now the authority, so a refusal can never be stamped over a node that is in the mesh. Good catch on `g.First()` not necessarily being the entry that won.

### ✅ Already fixed in `1cb50463db` — "NOT in the mesh" vs a failed UPDATE (GitHubActivityExtensions.cs:506, StaticRepoImporter.cs:1663, and the suppressed comment at 2127)

You reviewed `09f3fe0917`; this landed in `1cb50463db`, before your review was submitted. The reasoning is the same as yours — a content verdict can refuse an UPDATE as easily as a CREATE, leaving the node **present and stale**, and telling an operator it is absent sends them looking for the wrong thing. All three sentences and both catalog values now say *"the mesh does NOT hold this content"*, which is what is true in both cases.

### ✅ Named, not fixed — legacy manifest entries (suppressed comment, StaticRepoImporter.cs:3215)

You are right that a manifest written before this change stores a refused **update** under its plain token, so the incremental skip still passes over it. But this is **pre-existing, not introduced**: it is precisely the lie this change removes going forward. Before this PR that same entry existed and that same skip applied.

I have named it in the doc rather than migrating, for two reasons. A migration would have to decide, for every pre-existing entry, a fact no pre-existing entry records — there is nothing stored that separates "written" from "refused" in the old encoding. And it is self-limiting in the same way the marker's absent-verdict residue is: any edit to that file moves its token and the node is evaluated again. The **create** case needs nothing at all — a refused create leaves no node, and the incremental skip already requires the node to be present.

### ❌ Declining — `VerdictIsFinal` / the webhook path (StaticRepoImporter.cs:240)

This one I think is wrong, and the suggested change would re-create the defect `VerdictIsFinal` was built to fix.

`SkipReason`'s last arm skips only when **`LastAttemptedCommitSha == headSha`** — the *same commit*, re-built (a flake re-run, a manual re-dispatch). A green build of a **new** commit carries a different sha, does not match, and re-imports the whole source unscoped; the arm's own doc says so (*"the next new commit re-attempts"*). So it cannot produce the freeze this PR is about, which was permanent at **every** future commit.

And at the same commit, the delivery it drops is one that provably changes nothing under this PR: the refused node is skipped by the per-node memory, every other node's token is unchanged and present, so the pass would issue zero writes. That is exactly what #3945 measured the cost of — `Essentials/_GitSync` paying a full `git fetch --depth 1` of its entire repository on every green build for 31 days and 436 commits, while 33 siblings on the same webhook were skipped for free. Making content-error attempts non-final would restore that.

The repair the issue actually names — an operator pressing **Update** — does not consult this predicate at all (`ReimportAtCommit`, noted in the same doc block), and unblocking it is what this PR does.

One thing your finding did make me check, and it is worth recording: the change makes the second pass report `Failed = 1` where it previously returned `Skipped` with `Failed = 0`. `MayAdvanceBaseline` requires `Failed == 0`, so the new behaviour **holds** the git baseline where the old one did not obstruct advancing it — the safer direction, and correct, since the mesh genuinely is not at that commit.

### ↩︎ Not changed — `failedNote` is unkeyed (StaticRepoImporter.cs:2127)

`failedNote` is appended to the importer's composed terminal summary, which is unkeyed **in its entirety** and deliberately so: `UnkeyedActivityLogMessageRatchetGuard` carries a budget of 3 for this file and its own doc names *"a composed multi-clause import summary"* as a case no catalog can carry. `prunedNote`, `blockedNote`, `refusedNote` and `heldNote` are all in that same sentence.

The surface an operator actually reads **is** keyed: the GitSync activity line this PR adds is `activity.gitsync.failedNodes`, in both catalogs, with the paths as arguments — and the per-file ⚠ lines were already `activity.import.itemFailed`. So the German viewer does get a German sentence around the paths; what stays English is the internal bookkeeping summary on the attempt node, which was already so.

rbuergi:
**Disposition for the one review finding the agent's table did not cover** — recording it rather than letting it merge silently.

Copilot on `StaticRepoImporter.cs:2133`: *"`failedNote` is persisted in the import activity as an unkeyed `LogMessage`, so it renders …"*

**Accepted as-is, deliberately.** `failedNote` is one of **seven** sibling fragments interpolated into a single activity summary — `preservedNote`, `claimedNote`, `prunedNote`, `failedNote`, `blockedNote`, `refusedNote`, `heldNote` — and every one of the other six is unkeyed today. The new fragment follows the shape of the line it joins. Keying one of seven would leave a summary that is half-localised and read worse in both languages than either consistent choice.

Localising that whole summary is a real and separate change (one key with seven optional segments, or seven keys composed in order), and it is not this PR's subject. Where this PR adds genuinely new user-facing text it *did* localise — `activity.import.itemRefusedBefore` and `activity.gitsync.failedNodes`, both in `strings.en.json` and `strings.de.json`, with `Mirror-sync:` declared in the body.

Merging on that basis: 36/36 checks green, `mergeable_state: clean`, and the other six findings are fixed or declined with reasoning in the thread above.

PR #4475, opened 2026-09-16
fix(ops): keep Prometheus retention within its disk budget

rbuergi:
## Summary

Fix the shared monitoring stack's unbounded Prometheus TSDB. Declare a 6 GiB retained-data budget on the existing 8 GiB PVC, alongside the existing 15-day time retention. Prometheus removes oldest complete blocks and reserves capacity for its WAL/head and compaction.

The live failure was reproduced read-only: the 8 GiB metrics volume was full, oldest blocks began on 26 August, newest completed blocks ended on 10 September, and compaction could not advance the timestamp used by time retention. The user explicitly approved pruning the oldest metrics. No portal data, logs, credentials, or volume capacity are changed.

Extend the existing observability checker to inspect the rendered Prometheus command line and follow the TSDB mount to its PVC. Fail on missing/zero/oversized budgets, an omitted workload, unexpected rendering, or ephemeral storage. The check runs unconditionally in the existing chart lane.

## Validation

- Added eight regression controls: all eight failed against the previous checker while the nine existing controls passed.
- All 17 checker tests now pass.
- `check-observability-values.sh` passes against the real loki-stack 2.10.3 chart and actual Loki configuration validator; the Prometheus flag and PVC render correctly.
- `git diff --check` passes.
- No .NET project or embedded document assets changed; Release builds are not applicable to these Python/YAML-only changes.

This is internal monitoring configuration and gate coverage, so no end-user What's New entry is added. The third-party chart version remains unchanged; no MeshWeaver build pin is introduced.


rbuergi:
Production verification after the user's explicit approval to prune oldest metrics:

- Helm dry-run compared every rendered resource with revision 2 and proved the only change was adding `--storage.tsdb.retention.size=6GB` to `Deployment/loki-prometheus-server`.
- Applied through the existing Helm release, same loki-stack 2.10.3, reusing existing values. Revision 3 is deployed; no other resource/settings changes, no PVC resize, no manual file deletion.
- Before: 8 GiB PVC, filesystem 100% full. After: 60% used, 3.1 GiB available; Prometheus 1/1 ready.
- Native retention removed eight oldest blocks covering 26 August through 31 August (2,675,517,504 bytes). Oldest retained sample now 1 September 00:00 UTC. Other old block IDs changed through successful compaction, not additional history deletion.
- Since restart: 14 successful compactions, zero failed, two size-retention passes, one time-retention pass, one WAL truncation. Head time is current and completed blocks now extend to 16 September.
- Fresh Prometheus query reports the Memex portal's two replicas available. A normal `Hosting/InstanceAction` Sample completed Done with no observation warnings, and the previously stalled restart completed its verification and outcome phases without intervention.

The committed guard/config prevents the original unbounded setting from returning on later installs. The older metric history was deliberately pruned; no backup was created by this operation. Portal conversations/documents and Loki log history are untouched.

rbuergi:
Addressed the review's trailing-slash finding: `/data/` now resolves to the `/data` mount. The new regression failed before the fix and passes afterwards.

The suppressed fractional-size finding does not apply to the pinned Prometheus 2.41.0 dependency: [its exact units parser](https://github.com/alecthomas/units/blob/b94a6e3cc137/util.go) parses fractional digits, and [ParseBase2Bytes](https://github.com/alecthomas/units/blob/b94a6e3cc137/bytes.go) accepts both GB and GiB as binary quantities. Added positive controls for `5.5GB` and `6GiB`, plus a negative control for the unsupported `6G` spelling. The production value remains `6GB`.

Validation after the review fixes: **20 tests passed**, `git diff --check` clean, and the real chart 2.10.3 render plus real Loki binary validation passed. Live production retention remains 15 days / 6 GiB on its existing 8 GiB PVC; no additional data deletion was performed for these guard-only review fixes.

rbuergi:
Final independent review identified a real boundary case in the CLI conversion: `0.5B` truncates to zero bytes, which disables size retention. Added a regression that failed against the previous guard, then matched the CLI's integer-byte conversion before checking positivity.

Final local result: **21 tests passed (9.358 s)**, real chart 2.10.3 / Loki binary validation passed, and `git diff --check` clean. The actual deployed `6GB` / `15d` policy is unchanged by this guard-only fix.

rbuergi:
Merged as `66cc3df5d70594fc26dceccc0bd57010c66ba82b` at 2026-09-16 07:46:07 UTC after PR CI and merge-queue CI both completed successfully. Main's Chart Gate run 35070223194 also completed successfully.

Final live verification after merge:

- Helm `loki`, namespace `monitoring`, revision 3: deployed.
- Prometheus: 1/1 ready.
- Existing metrics filesystem: 61% used, 3,202,468 KiB available (about 3.1 GiB).
- Runtime retention flags: `15d` and `6GiB`.
- Since restart: 14 compactions completed, zero failed; two size-retention passes and one time-retention pass.

The user-authorized pruning removed the eight oldest blocks covering 26–31 August. History from 1 September onward was retained. No backup was created by this operation; recovery of pruned samples would require an independently existing backup. Portal documents/conversations and Loki log history were not touched. No platform-image deployment is needed for this Helm-only configuration change.

Issue #3903, opened 2026-09-10
A NodeType whose Source subtree is absent retries a doomed compile forever — rbuergi/OperationRequest failing since 2026-09-06 on memex.meshweaver.cloud


rbuergi:
## What was measured

Found by running the mandated pre-prod sweep on **memex.meshweaver.cloud**, 2026-09-10 ~06:45Z:

```
search 'nodeType:NodeType content.compilationStatus:Error'
→ count: 1, truncated: false
  rbuergi/OperationRequest  (v97, lastModified 2026-09-10T06:02:56Z by system-security)
```

Denominator, stated per the rule #3886 just corrected: **1 of >200 readable NodeTypes** (the plain `nodeType:NodeType` listing hits the 200 max and reports `truncated: true`, so the total is a lower bound; the Error sweep itself is NOT truncated, so the numerator is complete over partitions this caller can read).

## The failure

`rbuergi/OperationRequest` has **not compiled since 2026-09-06** and retries on every boot:

| field | value |
|---|---|
| `compilationStatus` | `Error` |
| `lastCompileSucceededAt` | **2026-09-06T07:08:49Z** |
| `lastCompileStartedAt` | 2026-09-10T06:43:45Z (still trying) |
| `lastCompiledVersion` | **33** — while the node is at version **97** |
| `latestAssemblyCollection` | `local` |
| `compiledFrameworkVersion` | `s2f227642d43f78eab13720c4af06a1be` |

```
CS0246 (line 11909): The type or namespace name 'OperationRequestContent' could not be found
CS1061 (line 11909): 'MessageHubConfiguration' does not contain a definition for 'AddOperationRequestControlPlane'
CS1061 (line 11909): 'LayoutDefinition' does not contain a definition for 'AddOperationRequestLayoutAreas'
```

Its `configuration` is the package's:

```csharp
config => config.WithContentType<OperationRequestContent>()
    .AddDefaultLayoutAreas()
    .AddOperationRequestControlPlane()
    .AddLayout(layout => layout.AddOperationRequestLayoutAreas().WithDefaultArea("Overview"))
```

## The cause: the NodeType is there, its SOURCE SUBTREE is not

```
search 'namespace:rbuergi/OperationRequest scope:subtree'   → count: 0
search 'namespace:Essentials/OperationRequest scope:subtree' → count: 21
```

`Essentials/OperationRequest` (v125, **compiles clean — it is not in the Error set**) carries 14 `Release` nodes plus:

[...]

PR #4478, opened 2026-09-16
fix(compile): a NodeType parked on a refused source names the import, not just CS0246 (#4469)

### LINKED ISSUE #4469: A NodeType parked on a source node the import could not write should name the import, not just CS0246
Split out of #4459 (defect 3) and #4456 (suggestion 3), which are fixed on the IMPORT side by #4467. This is the half #4467 deliberately does not reach, recorded rather than quietly dropped.

## What is still dark

When a static-repo / GitSync import cannot write a source node, the partition is left **referenced-but-incomplete**: the files that reference the missing symbol DID land, so every NodeType built from them fails to compile — and the only thing an operator sees is

```
CS0246: The type or namespace name 'SelfUpdateRouting' could not be found
CS0103: The name 'SelfUpdateRouting' does not exist in the current context
```

on a symbol whose file is **plainly in git**. Nothing in that error, and nothing on the NodeType, points at the import that produced the state. On 2026-09-15 that gap is what turned a one-line repo fix into an evening of investigation: five Hosting NodeTypes parked, `Hosting/InstanceAction` among them, so no instance action would run on the control instance at all.

#4467 fixes the import half — the refusal is now remembered per node instead of freezing the partition, and the failed path and reason are carried to the caller, the import summary and the GitSync activity (at `Warning`, so the activity no longer renders green). **An operator who reads the sync activity now learns which file is missing.** What is still missing is the link in the other direction: an operator who starts from the *compile error* still has nothing connecting it to the import.

## The two shapes a fix could take

1. **The compile failure names the import.** When a NodeType's compile fails on an unresolved symbol, and the partition's import bookkeeping records a node that did not land, say so on the compile diagnostic / the NodeType's own status: *"`Hosting/Deployment/Source/SelfUpdateRouting` was refused by the import at <activity> — fix the source file"*. This is the smaller change and probably the more useful one: it meets the operator where the symptom is.

2. **A partition known to be incomplete is not handed to the compiler at all** (or is, but with the incompleteness recorded as the compile's cause). Larger, and needs care — refusing to compile a partition because ONE unrelated node was refused would be the same over-broad granularity #4467 just removed from the import, so any version of this has to be scoped to the types that actually reference the missing node.

## Where the facts already are

Everything needed is already recorded as of #4467 — this is a wiring problem, not a measurement one:

- `{Partition}/_Activity/import-manifest` records a refused node as `!<token>` (see `StaticRepoImporter.RefusedTokenPrefix`), so "which declared nodes is this partition missing" is a read, not an inference.
- `StaticRepoImportResult.FailedPaths` carries each failed path with the write path's own reason.
- The import attempt activity carries the per-file ⚠ line.

## Related

- #4459, #4456 — the import-side defects, fixed in #4467
- #4259 — the same *shape* on the install path (declared source nodes never landed, and the install logged nothing); worth checking whether the reporting fix in #4467 should be mirrored there
- `Doc/Architecture/AContentVerdictIsPerNode` — the incident and what #4467 does and does not cover

rbuergi:
Closes #4469.

An import that cannot write one source node leaves the partition **referenced-but-incomplete**: the files that *reference* the missing symbol land perfectly well, so every NodeType built from them fails on `CS0246`/`CS0103` for a symbol whose file is **plainly in git** — and nothing in that error, and nothing on the NodeType, points at the import that produced the state.

Measured on `memex.systemorph.com` 2026-09-15: one literal NUL byte in `Hosting/Deployment/Source/SelfUpdateRouting.cs`, one row refused out of forty-one, five Hosting NodeTypes parked — `Hosting/InstanceAction` among them, so **no instance action ran on the control instance at all** — and an evening spent re-deriving a fact the system already held.

#4467 fixed the import half. This is the direction it deliberately did not reach: the operator who starts from the **compile error**.

## The surface, and why

| where | what it carries |
|---|---|
| `NodeTypeDefinition.CompilationError` | the sentence **leads** the recorded error, ahead of the diagnostics — so `get_diagnostics`, the Settings → Progress page and `search 'content.compilationStatus:Error'` all carry it. Same placement, and the same argument, as #3903's missing-source-query line. |
| the compile activity | the same fact as a catalog-keyed `Error` line, so an operator reads it in their own language. The path and the reason ride as arguments — neither is translatable. |
| `NodeTypeDefinition.CompilationImportRefusals` | the structured half, queryable. |

🚨 **A park must not erase it.** That is the state an operator actually finds the system in: the compile that formed the verdict ran once, and every later access is short-circuited through the park gate with the registry's *bare* Roslyn text. `ApplyGateSettle` therefore re-composes the sentence from `CompilationImportRefusals` (which survives the settle by construction) rather than remembering it as prose. The **batch bake** asks the same question for the same reason — `CompilationStatus`/`CompilationError` is one shared record with two writers, and the boot-time bake is exactly the writer that runs after a roll.

## The join

Everything needed was already recorded — this is wiring, not measurement.

- The import manifest's refusal entry now carries the write path's own **reason** (`!{token}|{reason}`). A token is `Convert.ToHexString` output, so the first `|` always ends it; both directions of compatibility hold exactly as the `!` sigil's already do, and an entry written before this reports *"the refusal predates reason recording"* — a different sentence from "no reason". A **remembered** refusal now reports the original reason instead of *"re-import with force to find out why"*.
- `IPartitionImportRefusals` (Graph.Contract) is the one-question seam, because `MeshWeaver.Graph` references the compile pipeline and not the other way round — the same shape `IPartitionSourceTracking` already is. `StaticRepoImportRefusals` reads the manifest through `GetMeshNodeOutcome`, so `Absent` (a real answer) and `Unavailable` (nothing established) cannot collapse at the instrument.
- `ImportRefusalDiagnosis` accuses an import only when **both** halves hold: the bookkeeping **records** a refusal for a node under a `Source`/`Test` segment, **and** this compile failed on a name-resolution diagnostic (`CS0246`/`CS0103`/`CS0234`/`CS0426`/`CS0400`) naming that node's identifier **as a whole word**. It reads the identifier out of the path and looks for it *inside* the diagnostic, never the reverse: Roslyn's message text is localized, a C# identifier and a `CS` id are not, so the join holds on a portal running in any language. Evidence is read **per diagnostic**, so an unrelated `CS0246` elsewhere in a transcript cannot lend its id to a mention of the symbol somewhere else.

## Three answers, never two

A symbol is equally missing when it was **deliberately deleted**, or when the module carrying it **is not loaded on this replica** (#3583 — a different defect with a different fix). Asserting an import refusal that was not established is **worse than today's silence**: it sends the operator to the repository for a file that was never there.

- `null` — NOT DETERMINED: not asked, no bookkeeping on this mesh, or the read did not come back.
- empty — read, and it explains none of these names.
- non-empty — established.

And the scoping is per type, never per partition: accusing every failed compile in a partition because one unrelated node was refused would re-create the over-broad reading #4467 had just removed from the import, one layer up.

## Proof

**Verified red first.** With the join wiring removed (`importRefusals: null` at the `ApplyCompileFailure` call site), the end-to-end pair is **2 failed / 0 succeeded**; restored, **2 passed / 0 failed**. Both arms go red, including the control — which also depends on the *determined-empty* answer.

A third case pins a hole the first commit opened: the reason now travels in the ledger, and it is a message from the **write path** — so it can carry the literal NUL that started all of this, into the same kind of Postgres text column that could not take it. Measured: the validator's message reached the ledger verbatim, control characters and all. `SanitizeLedgerReason` maps control characters to spaces, collapses runs and caps without splitting a surrogate pair; the case asks the seam directly (so it pins import → manifest → seam with no compile in the way) and asserts the message **survives** as well as being clean — a sanitizer that truncated at the first bad byte would silently drop the locator after it. Verified red on its own: **1 failed / 0 succeeded** with the sanitizer disabled.

The end-to-end test runs a **real import** whose validator refuses one shared source node, then a **real Roslyn compile** of a type that only *references* it — the shape four of the five parked Hosting types had, since only one of them owned the file. Its control arm compiles a **second** type in the **same** partition under the **same** standing refusal, failing on a **different** unresolved name, and requires it to be told nothing at all; without that, a blanket per-partition accusation would pass. Diagnostic ids are asserted against the set the join actually reads, not one literal — the same missing file earns `CS0103` in expression position and `CS0246` in type position.

Local runs, all `-c Release`:

| suite | filter | result |
|---|---|---|
| MeshWeaver.Graph.Test | `AParkedNodeTypeNames…`, `ARefusedNodeDoesNotFreeze…`, `NodeTypeOperationalContentTest`, `ShippedNodeTypeStateTest`, `FailedImportIsNotRetried…` | 24 passed |
| MeshWeaver.Compiler.Pipeline.Test | `ImportRefusalDiagnosisTest`, `NodeTypeCompileStampTest`, `NodeTypeCompileStateTest`, `GateSettleFailedBuildInputsTest`, `BuildDeliveryHoldTest`, `HydrateIsNotACompileTest`, `ADoomedCompile…` | 80 passed |
| MeshWeaver.Hosting.Test | `Bake`, `DeclaredSourcesMissing…`, `SweepSelector…` | 70 passed |
| MeshWeaver.Messaging.Hub.Test | `LocalizationTest` | 60 passed |
| MeshWeaver.Documentation.Test | `DocumentationLinkIntegrityTest`, `UnkeyedActivityLogMessageRatchetGuard` | 10 passed |

`dotnet build -c Release -warnaserror` on every touched project — Graph.Contract, Compiler.Pipeline, Graph, Hosting, Messaging.Hub and the four test projects — **0 Warning(s), 0 Error(s)**.

## Review round

Copilot raised 7 findings; **6 were valid** and are fixed in `d716557` — two of them were this change set violating contracts it states itself. Full disposition in [this comment](https://github.com/Systemorph/MeshWeaver/pull/4478#issuecomment-5693941436). The sharpest three:

- **`ApplyGateSettle` prepended the finding to *any* replacing reason.** A source change un-parks a type without clearing the field, so a later delivery-hold or availability verdict would have carried an import diagnosis — the unfounded accusation this mechanism exists to refuse. `formedUnderLiveInputs` already separates the two call sites exactly, and it is the same discriminator `FailedBuildInputs` beside it already uses.
- **The manifest read reported a corrupt-but-present manifest as determined-empty** — the null-versus-empty contract this seam is built on, violated by the caller that depends on it most. `TryParseManifest` keeps the failure; one parse, two readings.
- **The `{reason}` catalog change would have rewritten the past** (see `Mirror-sync:` below).

Finding 2 (localize the `CompilationError` lead) is declined with reasoning — that string is baked into the node at write time and read back by tools and `search`, so there is no viewer to follow when it is composed; the viewer-facing surface is the keyed activity line.

The guard that *should* have caught finding 5 structurally cannot: `NodeTypeOperationalContentTest` asserts only `MemberNames ⊆ the record`, never the reverse. Checking the other way found three **pre-existing** runtime-state properties already missing (`LatestAssemblyMvid`, `CompiledModulesHash`, `CompiledDependencies`). Not fixed here — the reverse guard would red this PR on unrelated debt — filed as #4480.

## Docs

Design conserved as [`Doc/Architecture/AParkedTypeNamesTheImport`](src/MeshWeaver.Documentation/Data/Architecture/AParkedTypeNamesTheImport.md), with the incident, the join rule, the granularity trap and the three-answers table. `AContentVerdictIsPerNode` no longer lists #4469 as open and links across. What's New entry included (`Category: Fix`).

---

**Pairs-with:** none — this diff removes no public top-level type and no public member from `src/`. It is additive only: one new interface (`IPartitionImportRefusals`), one new record (`ImportRefusal`), two new classes (`StaticRepoImportRefusals`, `ImportRefusalDiagnosis`), one new `NodeTypeDefinition` property, and one new **optional** parameter on the `internal` `ApplyCompileFailure`.

**Implementers:** `IPartitionImportRefusals.Refusals` — a brand-new interface, not a member added to an existing one, so no implementer outside this diff can exist yet. `StaticRepoImportRefusals` is its only implementation and ships in the same commit; it is resolved with `GetService<T>()` and a mesh with none registered answers NOT DETERMINED, so a host that never registers it compiles and behaves exactly as before. Checked `MeshWeaver.Plugins` for an existing type of that name: none. No member was added to any existing public interface or abstract class.

**Mirror-sync:** `npm run sync:i18n -- --ref <merged core sha>` in `MeshWeaver.Plugins/clients/react/src/i18n/`. This PR **adds two keys and modifies none** in both `strings.en.json` and `strings.de.json` — `activity.compile.importRefusedSources` and `activity.import.itemRefusedBeforeWithReason` — so the mirror is two keys further behind until that runs, with **zero value drift**. (An earlier revision of this PR added a `{reason}` argument to the EXISTING `activity.import.itemRefusedBefore` instead; that is reverted byte-for-byte, because activity messages are persisted with their arguments and `GetNamed` leaves an unknown placeholder visible, so every historical import activity would have rendered a literal `{reason}` for ever.)


rbuergi:
Thanks — six of the seven are valid, and **two of them were this change set violating contracts it states itself**. All fixed in d716557; auto-merge was disarmed while I worked through them so a green wall could not carry them past.

| # | finding | disposition |
|---|---|---|
| 1 | `EvidenceOf` falls back on the raw count, not the filtered errors | **Fixed.** Real: `BuildFailureDiagnostics` re-derives its rows from a *second*, LSP-shaped compilation, so it can return warnings-only while the transcript still holds the authoritative `CS0246` — the diagnosis went silently missing on a shape it is written for. Decision now turns on the filtered set; new unit case `StructuredDiagnosticsWithNoERRORS_FallBackToTheFlatTranscript`. |
| 3 | `ApplyGateSettle` prepends the finding to an unrelated gate verdict | **Fixed, and it is the sharpest of the seven.** A source change un-parks *without* clearing the field, so a later delivery-hold or availability reason would have carried an import diagnosis — the exact unfounded accusation this mechanism exists to refuse. `formedUnderLiveInputs` already separates the two call sites precisely: `false` re-serves the remembered compile failure (the verdict the finding belongs to), `true` forms a **new** verdict. Quoted only in the first case; the field is cleared in the second. |
| 4 | not cleared by the other terminal success/adoption paths | **Fixed.** Now cleared wherever `CompilationError`/`CompilationDiagnostics` are: `BuildDeliveryHold.Settle`'s serving state, `PrebuiltAssemblySeeder`'s adoption, `NodeTypeContractHandler`'s hydrate, and the adoption short-circuit in the helpers. |
| 5 | not in `MemberNames`, not mirrored on the compile-state satellite | **Fixed** — added to both, plus `IsEmpty`. Worth flagging: the pinning test only asserts `MemberNames ⊆ the record`, so it structurally *could not* have caught this, which is why it stayed green. Checking the other direction shows three **pre-existing** runtime-prefixed properties also absent — `LatestAssemblyMvid`, `CompiledModulesHash`, `CompiledDependencies`. That is someone else's scope, not this change's, so I have not added the reverse guard here (it would red on unrelated debt); filing it separately. |
| 6 | parse failure reported as determined-empty | **Fixed, and this one violated my own stated contract.** A corrupt-but-present manifest was reported as "the import lost nothing" **on the evidence of a parse failure**, suppressing attribution at exactly the moment the bookkeeping is broken. The importer's parse degrades to empty on purpose (for *it*, unreadable and empty both correctly mean "do a full pass"); `TryParseManifest` keeps the failure — one parse, two readings, so they can never drift about what a manifest *says*. |
| 7 | `{reason}` on an existing key breaks persisted rows | **Fixed, and this one would have rewritten the past.** Activity messages are persisted *with their arguments* and `GetNamed` deliberately leaves an unknown placeholder visible, so every historical import activity would have rendered a literal `{reason}` for ever. The old key and template are restored **byte-for-byte**; the with-reason wording has its own key — which doubles as the correct wording for a ledger entry that genuinely has no reason to report. The catalog diff is now two pure additions per language, zero modifications. |

**Not changed — #2, `Describe` is not localized.** Deliberate, and consistent with its immediate neighbour `SourceCoverage.Describe`. `CompilationError` is **baked into the node at write time** and read back later by `get_diagnostics`, by `search 'content.compilationStatus:Error'` and by the Progress page — there is no viewer to follow at the moment it is composed, and the Roslyn diagnostics it leads are English for exactly the same reason. Localizing it would either fork the field's register or require a per-viewer render the field does not have. The viewer-facing surface **is** the compile activity line, which carries `activity.compile.importRefusedSources` plus its arguments and is resolved per viewer at render time; `CompilationImportRefusals` is the structured data a localized renderer would bind to. Happy to revisit if the Progress page grows a localized renderer for the whole field.

Re-verified after the fixes: **244 tests green** — Graph.Test 24 (incl. `NodeTypeOperationalContentTest` 13, run separately to confirm it executes), Compiler.Pipeline.Test 80, Messaging.Hub.Test 60, Hosting.Test 70, Documentation.Test 10 — and every touched project builds `-c Release -warnaserror` with `0 Warning(s), 0 Error(s)`.

rbuergi:
Filed the pre-existing half of finding 5 as #4480 — `MemberNames` is pinned only as `MemberNames ⊆ the record`, which is why the omission on this PR stayed green, and three runtime-prefixed properties (`LatestAssemblyMvid`, `CompiledModulesHash`, `CompiledDependencies`) are already absent from it. Deliberately not fixed here: the reverse guard would red on unrelated debt, and each of the three needs the same per-property judgement the existing entries got.

PR #4483, opened 2026-09-16
fix(deployment): a record can state the COMPLETE required-module set, not an index overlay (#4476)

### LINKED ISSUE #4476: A record's requiredModules overrides the image's Modules:Required BY INDEX, so an empty list requires MORE and a short list requires modules the record never names
## What is wrong

`DeploymentPortalConfig.ModuleEntries` renders a record's `requiredModules` as `Modules:Required:N`
entries, and its own doc comment states the mechanism and then a conclusion that does not follow
from it:

> 🚨 These entries override the image's own list **BY INDEX** (an array key never appends), so a
> non-empty list is the COMPLETE required set.

An indexed override replaces the entries it names and **leaves the rest of the image's list in
place**. So the record cannot express either of the two things an operator will try:

1. **"Require nothing."** An empty `requiredModules` renders NO entries at all, so the image's own
   `Modules:Required` stands in full. Emptying the list does not relax the requirement — it
   restores it.
2. **"These are the required modules."** A list SHORTER than the image's overrides the leading
   indices and the tail survives, so the instance requires modules its record never names. The doc
   comment's "a non-empty list is the COMPLETE required set" holds only when the record's list is
   at least as long as the image's — which nothing checks or states.

## Measured on pearl.meshweaver.cloud (2026-09-16)

The image (`src/Memex.Portal.Distributed/appsettings.json`) requires six:

```
Radzen, Analysis, EntityViews, GoogleMaps, Speech, Social       (indices 0..5)
```

`Deployments/pearl` named five (`Blazor.Radzen, Blazor.Analysis, Blazor.EntityViews,
Blazor.GoogleMaps, Speech`) — indices 0..4 — and **never named `Social`**. The portal demanded
`Social` anyway, because image index 5 was never overridden. That is defect 2, and it is what put a
module nobody had asked for into the instance's missing-module report.

Defect 1 was then measured directly. `Deployments/pearl` was edited to `requiredModules: []` at
`06:08:24Z`. The pod created by the re-provision that followed
(`memex-portal-deployment-5977c7bfbf-tjrwl`) reported at `06:18:28Z`:

```
Health check required_modules with status Degraded … '5 required module(s) are store-delivered and
not here yet … MeshWeaver.Blazor.Radzen … MeshWeaver.Blazor.Analysis … MeshWeaver.Blazor.GoogleMaps
… MeshWeaver.Speech … MeshWeaver.Social'
```

An empty record list, and a pod requiring five modules — read off a pod that started AFTER the edit.

## Why it matters beyond one instance

The record is the single authoritative description of an instance, and a reader cannot tell what an
instance requires by reading it: the answer is the record's list index-merged over whatever list the
image happens to ship. The `Deployments/pearl` notes now say the list "is now EMPTY rather than
unsatisfiable", which is what the record implies and not what the instance does.

## Two ways out (not implemented — this needs a decision)

- **An explicit authority flag.** Render `Modules:RequiredIsAuthoritative=true` beside the entries
  and have the reader take the record's list as the complete set when it is present — including
  when it is empty. Says what the operator meant; costs a reader change in core.
- **Pad the override.** Emit empty entries for the indices beyond the record's list so the image's
  tail is blanked, and have the reader skip blanks. No reader contract change, but it silently
  depends on knowing how long the image's list can get, which is the coupling this issue is about.

Either way `ModuleEntries`' doc comment should state what an indexed override actually does with a
shorter or empty list, and `HelmValuesTests` should pin it — today no test covers a record list
shorter than the image's.

<sub>Measured 2026-09-16 by Claude Opus 5 through the control instance's Logs action (read-only).</sub>

rbuergi:
Fixes #4476.

## The root cause, confirmed

The issue's hypothesis is right, and the mechanism is a property of ASP.NET configuration rather
than of anything in this repository: **`Modules:Required` is an ARRAY, and configuration merges
arrays BY INDEX.** A later provider replaces the entries it NAMES and leaves every other index of
the earlier one standing — so an array can express *"replace entry N"* and can never express
*"these and only these"*. A record's `requiredModules` was therefore an OVERLAY on the image's list,
not a description of what the instance requires:

- a list **shorter** than the image's leaves the image's tail required (pearl demanded `Social`,
  which its record never names);
- an **empty** list renders nothing at all, so the image's list stands in full — emptying it does
  not relax the requirement, it restores it.

`ModuleEntries`' doc comment stated the mechanism and then a conclusion that does not follow from it
("a non-empty list is the COMPLETE required set"), which holds only when the record's list is at
least as long as the image's — and nothing checked or stated that.

**Two corrections to the issue's measurements, both worse than reported.** The image's list is
**nine** entries today, not six (`Social`, `Blazor.Chat`, `Markdown.Collaboration`, `AI` were added
after it was written), so every fleet record — `memex`, `memex-cloud`, `pearl` — is short by four,
not one. And `Deployments/memex-cloud` names `requiredModuleSlots: {"7": "MeshWeaver.Mcp.dll"}`
against an image whose index 7 is now `MeshWeaver.Markdown.Collaboration.dll`: the Memex#131
shadowing has silently recurred because the image's list grew underneath the "first free slot"
advice. That one is an operator fix on the private record, not code — the boot report names it.

## The decision the issue asked for

The issue offers an authority flag or padding. **Padding is rejected**: it needs the image's length,
which is exactly the coupling that expires (see above). **An authority flag alone cannot work
either** — a flag gives no bound, so a claimed-but-shorter list still inherits the tail.

What lands is the flag *read off provider order*, which needs no bound at all:

- **`Modules:RequiredIsAuthoritative`** — a SCALAR key, so no index merge can touch it, and the one
  thing an EMPTY list can still state (with no indexed keys rendered, "require nothing" and "this
  record has no opinion" are otherwise the same bytes). Rendered from the new record field
  `requiredModulesAuthoritative` on both routes (portal ConfigMap + the operator's catalog config
  file) and by the chart. Never rendered as `false`: the reader takes an explicit false as a
  *withdrawal* of a claim layered under it.
- **`MeshBuilderModuleActivation.RequiredEntries(configuration)`** — the ONE reading the boot path
  and the health check share. No claim ⇒ the merged array, byte for byte what every caller read
  before. Claim ⇒ the entries the claiming provider *and anything layered after it* supply, and
  those only. No count, no padding, no knowledge of the image's length.
- **`MeshBuilderModuleActivation.UnstatedRequired(configuration)`** — the quiet half no other check
  can see, reported at boot. `ShadowedRequired` sees an entry a deployment **replaced**; this sees
  the effective entries its overlay never **reached**. That is the shape that let the record and the
  instance disagree in silence.

**Opt-in, deliberately.** The image's list is the platform's floor — the AI engine, the chat
renderer and the collaboration pack are each named there because losing one silently is a measured
outage — and every fleet record names five against nine. Taking a partial list as authoritative *by
default* would have un-required four modules on every instance the day this shipped.

## The chart ceiling moves 9 → 19

The portal ConfigMap names every key literally (no `range`, by contract with the key-literal
guards), so the `Modules__Required__N` block has a hand-written ceiling. It was 9 while an overlay
was a three-slot *append*; the claim renders a **complete set**, and memex-cloud's (the nine plus
MCP) already fills 0..9 exactly. One more module in the image and a claimed set would silently lose
its last entry — which under the claim means NOT REQUIRED. The stale "the baseline is 7 entries"
comment is corrected to nine in the same block.

## Reproduction and negative control

Reproduced first, against the public API as it stood, over a REAL two-provider configuration (the
image's appsettings, then the ConfigMap overlay) — the layering the defect lives in:

```
Test run summary: Failed!   total: 42   failed: 5   succeeded: 37
  failed AnEmptyAuthoritativeList_RequiresNOTHING              (image's nine survived an empty list)
  failed ADeploymentThatStatesItsListIsComplete_…              (Social, Chat, Collaboration, AI inherited)
  failed AShorterDeploymentListLeavesTheImagesTailRequired     (nothing named the four unstated)
  failed AnAuthoritativeListMayStillPlaceAnEntryAtAnExplicitSlot
  failed ABlankedEntryInAnAuthoritativeList_IsStillNotRequired
```

With the fix: `42 / 0 failed`. **Negative control:** the two readings were reverted to today's
semantics in place (signatures kept, so the suite still compiles) and the *same five* went red
again — `total: 42, failed: 5` — then restored to `42 / 0`. The tests pin the MERGE semantics the
issue asked for, including a record list shorter than the image's, which nothing covered before.

Also green locally, Release + `-warnaserror`, `0 Error(s)` each: `MeshWeaver.Mesh.Contract`,
`MeshWeaver.Deployment.Contract`, `MeshWeaver.PluginCatalog`, `MeshWeaver.Documentation`,
`MeshWeaver.Compiler.Pipeline.Test` (42/42), `MeshWeaver.Deployment.Contract.Test` (24/24),
`MeshWeaver.Documentation.Test` (476/476, incl. `PlatformBakeLaneGuard`, `ArchitectureTopicMapTest`
and `DocumentationLinkIntegrityTest`), `MeshWeaver.PluginTester.Test/ExternalModuleGateTest` (3/3).

## Still open (named, not silently left)

`/health`'s `required_modules` check lives in `Memex.Portal.Distributed` (MeshWeaver.Plugins) and
still reads the raw array. For every record today the two readings are identical, so nothing
disagrees; a record that makes the claim needs that half adopted. The core half lands first because
the satellite compiles against a pinned core.

Pairs-with: none — this diff removes no public type and no public member (`check-type-forwards.py`
measured against the merge base: 2168 / 2168 public types, 13816 → 13822 members, **0 removed, 6
added**), and it adds no interface member (implementer obligations 456 → 456).

Mirror-sync: none — no key is added to or re-worded in
`src/MeshWeaver.Messaging.Hub/Localization/strings.{en,de}.json`. The one new user-visible label is
a `[Description]` on a `DeploymentContent` property, whose German text lives in the Labels table of
`Doc/Architecture/ConfiguringAnInstanceFromAspire` (that assembly carries zero MeshWeaver references
and cannot read the catalog).


rbuergi:
The queue merged this while the Copilot review on it was still being worked. Three of the five findings were valid and each was a way the authority claim could be *believed* while the module it covered was quietly not required — they are fixed in #4488, together with the two hygiene ones:

1. `BootConfigurationEntries` rendered the contiguous list alone and dropped every explicit `requiredModuleSlots` entry, so the two delivery routes stated two different COMPLETE sets.
2. The chart's `Modules__Required__N` ceiling was a number in a comment; a slot above it reaches no container, which under the claim means NOT REQUIRED rather than "the image's entry stands". Now a constant read back out of the template, plus a reported problem instead of a silent drop.
3. The cross-assembly key assertion sat in a project that references only the renderer's assembly, so it compared a constant against its own spelling and could never catch a rename of the reader's. Moved to a project that sees both.

Plus the stale "six entries" wording in the `UnstatedRequired` diagnostic and two `static readonly string[]` test fixtures.

Issue #4484, opened 2026-09-16
The sanctioned retype repair races the owner recycle the PREVIOUS retype triggered — DanglingNodeTypeUpdateTest refused at DisposeHostedHubs, 65 ms in


rbuergi:
Found while running the full `MeshWeaver.Graph.Test` project for #4480 (the diff there cannot reach this test — see "Attribution" below).

## What failed

`MeshWeaver.Graph.Test.DanglingNodeTypeUpdateTest.Upsert_WithTheNamedImportEscapeHatch_StillLands`, on the THIRD step — the repair upsert that retypes the stranded node back to `Markdown`:

```
Expected value to be true because retyping a stranded node to a type that DOES resolve is the
sanctioned repair. Error was: Inner UpdateNode faulted: MeshNode Unknown at 'TestData/dntad82f245':
Hub TestData/dntad82f245 is shutting down (RunLevel=DisposeHostedHubs, activation #007C42DF)
— cannot process RawJson; the address may reactivate (recycle / restart). Rejecting now.
```

The whole test took **65 ms** — the first upsert succeeded at `+19 ms`, the repair was refused at `+62 ms`. This is not a timeout; the owner was mid-recycle and fast-refused.

## The race

It is the same window #4159 documented, one leg further on. Step 2 retypes the node `Markdown → No/Such/NodeType` under `AllowUnresolvableNodeType`; that stamped merge makes **the owner's rebind watcher recycle the owner** (#4159: *"after the stamped merge the owner's rebind watcher recycles the owner, and a fresh activation lands with `EnrichWithNodeType … applying error overlay`"*). Step 3 then posts the repair at the same address. If it arrives while that recycle is in `DisposeHostedHubs`, the intake gate refuses it — correctly, per its own contract (*"the address may reactivate"*) — and `CreateOrUpdateNodeRequest` surfaces the refusal as `success=false`, which the test reads as the repair being unavailable.

The full trail from the run:

```
10:56:09.110 [CreateOrUpdate] TestData/dntad82f245: writing NodeType 'No/Such/NodeType' WITHOUT the existence check …
             upsert allowUnresolvable=True success=True
10:56:09.153 [UpdateRemote] response wait errored hub=cache/… target=TestData/dntad82f245
             DeliveryFailureException: Hub TestData/dntad82f245 is shutting down (RunLevel=DisposeHostedHubs, activation #007C42DF)
[...]

Issue #4159, opened 2026-09-13
Merge-queue race: PatchDataResponse posted by the owner at +2 ms never reaches the cache hub's callback — DanglingNodeTypeUpdateTest OwnerUnreachable after 31 s (run 34744471576, shard 4)


rbuergi:
## What failed

Core **merge queue**, group build [`34744471576`](https://github.com/Systemorph/MeshWeaver/actions/runs/34744471576) (2026-09-13 07:14Z, #4110 + `main`; #4110 touches only `deploy/aks/operator/*` and `hosting-operator.yml`, zero `src/`), shard 4:

`MeshWeaver.Graph.Test.DanglingNodeTypeUpdateTest.Upsert_WithTheNamedImportEscapeHatch_StillLands` — `Inner UpdateNode faulted: MeshNode OwnerUnreachable at 'TestData/dnt34d0ffdc': no verdict within 31s`.

`main` is green on the same commit (`cdb2878bb`, run 34720891908); 25/25 green locally in isolation (0.6–0.9 s each). A race, i.e. a defect — the steward left #4110 dequeued (`queue-rejected`), not catalogued.

## The trail — the owner did everything, the reply never arrived

```
AWAITING PatchDataRequest→TestData/dnt34d0ffdc@cache/s_oMZAhQcUGfUcamZTnAiA(+0ms)
  → POSTED → RECEIVED@cache → ENQUEUED@cache → QUEUED@cache
  → RECEIVED@mesh → ENQUEUED@mesh → QUEUED@mesh → ROUTED onTarget=False state=Forwarded@cache
  → RECEIVED runLevel=Started@TestData/dnt34d0ffdc(+0ms) → …(+6 suppressed)…
  → HANDLER_EXIT state=Processed@TestData/dnt34d0ffdc(+1ms)
  → PATCH_MERGE_TURN entered → PATCH_MERGE_STAMPED v=2 refused=0 → PATCH_FLUSH_BOUND_ARMED
  → PATCH_ECHO_SEEN(+1ms) → PATCH_ACK ok(+2ms)
  → RESPONSE_POSTED type=PatchDataResponse target=cache/s_oMZAhQcUGfUcamZTnAiA@TestData/dnt34d0ffdc(+2ms)
  → PATCH_FLUSH_SUBSCRIBED(+2ms)
  ⇒ a reply WAS posted for this correlation and the callback is STILL pending — the reply was lost
    between the responder and the requester, so chase the response delivery, not the handler.
[...]

Issue #2544, opened 2026-08-27
NodeType recompile storm: one event mints N release requests, and the watcher queues them instead of coalescing


rbuergi:
## Summary

One logical event — a single merge, a single GitSync push, a single plugin install — produces **N NodeType release requests**, and the release watcher **queues** them instead of coalescing them. Each surplus trigger is dispatched by the previous compile's own terminal `Ok` write-back, so N triggers become N *sequential* Roslyn compiles of the same type against the same inputs.

Every compile invalidates the type's instance hubs and raises the *"A newer build of this type is available … Recycle"* adornment. Observed in production on `memex-cloud`: compiles arriving in near-simultaneous **pairs** (two 65 ms apart on `Training/Tour`; three inside 2 s on `Publish/Deck`), and **seven compiles for one merge** on `Store/Plugin`. Reported from the user side as a slide deck that "keeps recompiling while being viewed" (Systemorph/MeshWeaver.SocialMedia#75).

### A. A release request is identified by *when* it was asked, never by *what* it asks for

`src/MeshWeaver.Graph/NodeTypeReleaseExtensions.cs:153` mints a fresh timestamp on every call, unconditionally — no "already requested" check, no coalescing, no debounce:

```csharp
var triggerAt = DateTimeOffset.UtcNow;
…
RequestedReleaseAt = triggerAt,
RequestedReleaseForce = force,
```

Two identical requests for the same content are therefore two *distinct* values that can never be recognised as one.

### B. The watcher parks surplus triggers rather than dropping them

`src/MeshWeaver.Graph/Configuration/NodeTypeCompilationHelpers.cs:1408-1419`:

```csharp
.Where(node => node?.Content is NodeTypeDefinition def
    && def.RequestedReleaseAt is { } req
    && dispatchHighWater.IsPast(req)
    && (def.LastReleaseRequestHandledAt is null || req > def.LastReleaseRequestHandledAt.Value)
    && def.CompilationStatus is not CompilationStatus.Pending
                              and not CompilationStatus.Compiling),
```

A trigger that arrives while the type is `Pending`/`Compiling` fails the predicate, and the high-water mark is deliberately **not** advanced for it (correctly — see the `#185` comment). So it re-fires on the next settled emission, which is the first compile's own `Ok` write-back. The compile does not re-trigger itself; it is the **clock that releases the next queued trigger**. That is the pair.

[...]

PR #4485, opened 2026-09-16
fix(mesh): every NodeTypeDefinition member is owned, and the mask is pinned in BOTH directions (#4480)

### LINKED ISSUE #4480: NodeTypeOperationalContent.MemberNames is pinned in one direction only, and three runtime-state members are already missing from it
Found while reviewing #4478 (#4469). Not fixed there — it is pre-existing debt outside that change's scope, and adding the guard below would have reddened an unrelated PR.

## The gap

`NodeTypeOperationalContent.MemberNames` is the set of `NodeTypeDefinition` members that are RUNTIME COMPILE STATE: stripped on export, preserved from the live node on import, so a repo file can neither carry them nor forge them. Its pinning test, `NodeTypeOperationalContentTest.OperationalMembers_AreRealNodeTypeDefinitionProperties`, asserts **one direction only**:

```csharp
foreach (var member in NodeTypeOperationalContent.MemberNames)
    Assert.True(properties.Contains(member), "...the list drifted from the record.");
```

i.e. `MemberNames ⊆ the record`. It cannot see a runtime-state property that is **missing** from `MemberNames`, which is the direction that actually loses data — and it stayed green through exactly that omission on #4478 (caught by review, not by the guard).

## The three that are already missing

Cross-checking the record against `ShippedNodeTypeStateTest.RuntimeStatePrefixes` (`Compilation`, `Compiled`, `LastCompil`, `LastRelease`, `LatestAssembly`, `LatestRelease`, `RequestedRelease`, `CurrentSource`, `Failed`) — the list that already defines "written by the runtime, never authored" — gives 26 runtime-prefixed properties, of which three are absent from `MemberNames`:

- `LatestAssemblyMvid`
- `CompiledModulesHash`
- `CompiledDependencies`

## Why it matters

Both halves of the seam are affected, in opposite directions:

- **Export / change-detection.** `StripOperational` leaves them in the repo file and in the change token, so a node's runtime state leaks into authored content and moves the token when only the runtime state changed.
- **Import.** `WithTypedMembersReset` does not preserve the live values for them, so an import can overwrite a measurement taken on THIS mesh with whatever a repo file happens to carry — and for `CompiledModulesHash` in particular an authored value is the same class of forgery the `adoptedSourceFingerprint` entry's comment warns about: it participates in deciding whether a build is stale.

I have deliberately not assumed these are oversights — each needs the same judgement the existing entries got (three of them carry a reasoned comment saying *why* they are operational). One of them may be intentionally excluded for a reason the list does not record, which is itself worth writing down.

## Proposed

1. Decide, per property, operational or not — and record the reason inline next to the entry, as the existing ones do.
2. Add the **reverse** assertion to `NodeTypeOperationalContentTest`: every `NodeTypeDefinition` property matching a `RuntimeStatePrefixes` entry is in `MemberNames`, with an explicit, reasoned allow-list for any deliberate exclusion. Single-source the prefix list with `ShippedNodeTypeStateTest` rather than copying it — two copies of that list drifting is the same failure one layer up.

Without step 2 the next runtime-state member added to the record will be missed the same way, silently.

rbuergi:
Fixes #4480.

## Root cause

`NodeTypeOperationalContent.MemberNames` is the set of `NodeTypeDefinition` members the sync seams treat as MESH-OWNED — stripped on export, taken from the live node on import, ignored by the change token. It was pinned by one assertion, `MemberNames ⊆ the record`, which catches an entry naming no property and is **structurally blind** to a runtime-state property MISSING from the mask. That is the direction that loses data, and four members were missing that way.

| Member | What it decides | What an authored value does |
|---|---|---|
| `latestAssemblyMvid` | the IDENTITY of the bytes the build produced (#2471); bind time compares it against the bytes served | forges a MATCH and turns off the stale-build detector — or a MISMATCH and refuses a correct bind |
| `compiledModulesHash` | `HasUsableBuild` invalidates a build stamped with a different non-null hash (#1664) | a hash matching the importing deployment declares a FOREIGN build usable, suppressing the recompile a module update requires |
| `compiledDependencies` | `HasUsableBuild` / the bake probe's `Classify` / `IsAlreadyAdopted` (#1707) | a record matching the bundle makes a FRESH install read as already-adopted — the bytes are never seeded |
| `dispatchedBuildInputs` | the token the IN-FLIGHT compile was dispatched for (#2544) | absorbs a live release request against a compile nobody dispatched, so that release is simply lost |

## What the issue's own instrument could not see

The issue proposed deriving the reverse assertion from `ShippedNodeTypeStateTest.RuntimeStatePrefixes`. That finds three of the four: **`DispatchedBuildInputs` is spelled outside the convention**, so no prefix-based guard would ever have named it — the same defect, one name away. The convention is a *sufficient* condition and never a necessary one (`BuildProvenance`, `ReleaseNotes`, `RequestedSourceStampAt` and the whole `Adopted*` family are mesh-owned and none is spelled like it either).

So the guard is three, and the convention now lives in **one** place (`NodeTypeMemberOwnership`), read by both test classes instead of copied:

1. `MemberNames ⊆ the record` — unchanged.
2. **convention ⊆ `MemberNames`** — the issue's reverse assertion.
3. **the partition** — every serialised member of the record is classified repo-authored, mesh-owned-and-masked, or mesh-written-but-deliberately-unmasked, **exactly once**. This one does not depend on what a member is called.

`ShippedNodeTypeStateTest`'s banned set now unions all three sources rather than deriving from the convention alone.

## The one reasoned exclusion

`PendingRetirement` is mesh-written and stays UNMASKED, deliberately: nothing in `src/` ever writes null back to it, so the only thing that clears it is the repo shipping the type again — an upsert replaces the content wholesale, and an unmasked member absent from the file goes away. Mask it and a re-shipped type stays marked retired forever, and the bake gate reads a stamped type's compile failure as "do not hold the rollout". The committed-file ban covers it instead.

`NodeTypeCompileState` carries the four as well — its member set is pinned to the mask, so the compile-state satellite would otherwise silently drop them.

## Negative control

The three new guards are RED on the pre-fix tree, naming exactly `CompiledDependencies, CompiledModulesHash, LatestAssemblyMvid` (reverse assertion) plus `DispatchedBuildInputs` (partition), and `dispatchedBuildInputs` missing from the banned set — then green with the fix. Verified both ways in the full project population, not only under a filter.

Measured locally, Release + `-warnaserror`, `0 Error(s)` on every touched project:

- `MeshWeaver.Graph.Test` **1720/1721**
- `MeshWeaver.Compiler.Pipeline.Test` **970/970**
- `Memex.Portal.Shared.Test` **1606/1606**
- `MeshWeaver.Documentation.Test` link integrity + topic map green

The one Graph failure is a known pre-existing race on `DanglingNodeTypeUpdateTest` (#4159's window, one leg further on) that this diff cannot reach — every node there is typed `Markdown`/`No/Such/NodeType`, so `IsNodeTypeNode` is false and all three seam functions return the input untouched. With the `src/` half reverted and rebuilt, the same full run passed it. Filed separately as #4484; deliberately not band-aided here.

## Notes

- Work product conserved: [Who Owns a NodeType Member](src/MeshWeaver.Documentation/Data/Architecture/NodeTypeMemberOwnership.md), listed in `Architecture.md`, plus a `Category: Fix` What's New entry.
- One-time effect: the first sync after this lands drops the four members from repo files that still carry them, and the change token moves once for those nodes. That convergence is the correction.
- No public surface removed and no interface member added; no i18n catalog change.


rbuergi:
Addressed all four review findings in f35a553490.

**1 — `PendingRetirement` leaked through the EXPORT** (the substantive one, and correct). Leaving it out of `MemberNames` also left it out of `StripOperational`, which `GitHubSyncService.SerializeOne` calls — so the live stamp rode every export into the repository, where this PR's own `ShippedNodeTypeStateTest` bans it, and a file could then forge a retirement the mesh never measured.

The bucket was under-specified, not the exclusion. A member can belong to the mesh at the EXPORT and CHANGE-TOKEN seams and still have to lose at the IMPORT seam, because dropping it on re-import is the only thing that clears it. That is now what the bucket **means** — `NodeTypeOperationalContent.StrippedButNotPreserved`, carrying its reason — and the strip paths run over the **union** while the preserve path keeps `MemberNames` alone. The test project reads the production set rather than declaring a fourth copy of the list this PR exists to keep single.

Negative control: with the union reverted, the new test `StrippedButNotPreserved_LeavesTheFile_AndIsNeverRestoredFromLive` fails on *"the strip removed NOTHING from a node carrying a live pendingRetirement"*; green with it.

**2 — the partition's denominator was one member short.** Correct, and it is the same class of defect as the one this PR fixes: it excluded every property carrying a `JsonIgnoreAttribute`, but `IncludeGlobalTypes` carries `JsonIgnore(Condition = Never)`, i.e. ALWAYS written. It now filters on the CONDITION.

**3 — the dangling `<see cref="Classified"/>`.** Fixed. `CS1574` is an error under `-warnaserror`; the identical mistake in `Mesh.Contract` failed that build during this round.

**4 — mutable `List<T>` accumulators.** Replaced with immutable projections.

Re-measured after the change, Release + `-warnaserror`, `0 Error(s)` per project: Graph **1725/1725**, Compiler.Pipeline **990/990**, Memex.Portal.Shared **1606/1606**, doc link integrity + topic map green.

PR #4488, opened 2026-09-16
fix(deployment): the two routes agree slot for slot, the chart's ceiling is stated, and the key test can fail (#4476 follow-up)

### LINKED ISSUE #4476: A record's requiredModules overrides the image's Modules:Required BY INDEX, so an empty list requires MORE and a short list requires modules the record never names

rbuergi:
Follow-up to #4483 (fixed #4476), which merged through the queue while its Copilot review was still
being worked. Three of the five findings were valid, and each was a way the authority claim could be
*believed* while the module it covered was quietly not required. Two more are hygiene.

## 1. The two delivery routes described different sets

`BootConfigurationEntries` — the operator's catalog config file (`HOSTING_CATALOG_CONFIG`) — rendered
the **contiguous** `requiredModules` list alone and dropped every explicit `requiredModuleSlots`
entry, while `PortalConfig` renders `ModuleSlots`: both. That was harmless while the entries were
read as a by-index **overlay** (a dropped slot just left the image's entry standing). Beside the
claim the two routes state two different **complete** sets, and the route that dropped a slot says
that module is not required *at all*.

It now renders `ModuleSlots` too, and `BOTHRoutesRenderTheSameSlots_ExplicitOnesIncluded` compares
the two routes slot for slot — which is the drift `DeploymentPortalConfig` exists to make impossible
("a difference between the two is a bug here rather than a drift between two renderers").

## 2. The chart's ceiling was a number in a comment

The portal ConfigMap names every `Modules__Required__N` key literally (a Helm `range` renders
correctly and is invisible to the key-literal guards), so the block stops somewhere and a slot above
it reaches **no container** — silently, with helm reporting success. Before the claim that meant "the
image's entry at that index stands", which is merely wrong; under the claim it means **NOT
REQUIRED**.

- `DeploymentPortalConfig.MaxRenderedRequiredModuleSlot` states the ceiling.
- `TheDeclaredCeilingIsTheOneTheChartRenders` reads the number back **out of the template** (and
  asserts the block is contiguous from 0), so the constant and the chart cannot drift — a constant
  that drifts is the Memex#128/#131 shape wearing a different hat.
- `RequiredModuleProblems`, folded into a new `SpecProblems(DeploymentContent?)` overload, **names**
  a slot above the ceiling instead of letting the render drop it — the same "report, never drop"
  rule `Validate` already applies to a half-configured plugin mount.

## 3. The key test could not fail for the reason it existed

The claim's key is spelled twice, in two assemblies that may not reference each other
(`MeshWeaver.Deployment.Contract` carries zero MeshWeaver references by design), and
rendered-but-never-read is indistinguishable from not rendered. The assertion sat in
`MeshWeaver.Deployment.Contract.Test`, which references only the **renderer's** assembly — so it
compared a constant against its own spelling and could never see a rename of the **reader's**.

It moves to `ConfiguredModuleActivationTest`, which sees both assemblies through
`MeshWeaver.PluginCatalog`. What stays behind asserts the key's *shape* and says in its own doc
comment why it is not the real check, so nobody "strengthens" it back into a tautology.

## 4 & 5 — hygiene

- The `UnstatedRequired` diagnostic still described the image's list as six entries while the rest of
  #4483 corrects it to nine; it now states the pearl measurement without contradicting the count.
- The two test fixtures were `static readonly string[]` — a static **mutable** collection, which
  AGENTS.md forbids (`static readonly` is for immutable lookups). Now `ImmutableArray<string>`.

## Verified

Release + `-warnaserror`, `0 Error(s)` each: `MeshWeaver.Deployment.Contract`,
`MeshWeaver.Mesh.Contract`, `MeshWeaver.PluginCatalog`, `MeshWeaver.Documentation`, and the three
test projects. Suites: `ConfiguredModuleActivationTest` **43/43** (was 42 — the cross-assembly key
assertion), `MeshWeaver.Deployment.Contract.Test` **27/27** (was 24), `MeshWeaver.Documentation.Test`
**476/476** including `PlatformBakeLaneGuard`, `ArchitectureTopicMapTest` and
`DocumentationLinkIntegrityTest`.

Both new assertions were checked to be able to fail: the ceiling test compares the constant against
the template's own `Modules__Required__N` block, and the cross-assembly test resolves the reader's
constant from `MeshWeaver.Mesh.Contract` (it did not compile until the reference was qualified,
which is itself evidence it reaches the other assembly).

Pairs-with: none — this diff removes no public type and no public member (`check-type-forwards.py`
against the merge base: 2168 / 2168 public types, 13816 → 13824 members, **0 removed, 8 added**), and
it adds no interface member (implementer obligations 456 → 456).

Mirror-sync: none — no key is added to or re-worded in
`src/MeshWeaver.Messaging.Hub/Localization/strings.{en,de}.json`.


rbuergi:
Copilot's finding on `DeploymentPortalConfig.cs:359/362` is correct and is fixed in `cad6bfe9cc`.

The ceiling is a property of the **Helm chart** (its literal-key `Modules__Required__N` block), not of every renderer: the Aspire route injects whatever `PortalConfig` emits, ceiling and all, so it delivers slot 20 correctly. That made two things false at once — the message said "the highest slot **any renderer** delivers", and folding the check into a route-neutral "why the spec cannot bring an instance up" surface would have rejected a record an Aspire run brings up perfectly.

- `MaxRenderedRequiredModuleSlot` → `MaxChartRenderedRequiredModuleSlot`, documented as the chart's and naming the Aspire route.
- `RequiredModuleProblems` → `ChartModuleSlotProblems`, and the `SpecProblems(record)` overload is **gone** rather than carrying a chart-specific verdict — a Helm renderer asks the chart check *alongside* the spec problems.
- The message now says the slot reaches no container **in Kubernetes** and that the Aspire route delivers it, which is the nastiest property of the shape rather than a footnote: it works on a laptop and the module disappears in the cluster.
- `TheCeilingIsTheCHARTS_SoTheRouteNeutralSpecProblemsDoesNotCarryIt` pins exactly that — the route-neutral `SpecProblems` stays empty for such a record, and **both** renderers are asserted to emit the key for it, so "the chart would not carry it" is measured rather than assumed.

Local, Release + `-warnaserror`, `0 Error(s)`: `MeshWeaver.Deployment.Contract` and the three test projects. Suites: deployment contract **28/28**, `ConfiguredModuleActivationTest` **43/43**, `MeshWeaver.Documentation.Test` **476/476**.

PR #4492, opened 2026-09-16
fix(review): the third ownership bucket is STRIPPED but not preserved, and the partition's denominator was one short (#4480)

rbuergi:
Follow-up to #4485 (issue #4480). Copilot's review landed four findings on that PR; auto-merge fired on the previous head at 09:37:05Z while the fixes were being pushed, so they are not on `main`. This is that commit, unchanged, on top of `main`.

## 1 — `PendingRetirement` leaked through the EXPORT (the substantive one)

#4485 introduced a third ownership bucket for members the mesh writes but the import must **not** preserve, and gave it the wrong meaning: "masked at no seam at all". Leaving `PendingRetirement` out of `MemberNames` also left it out of `StripOperational` — which `GitHubSyncService.SerializeOne` calls — so a held type's live retirement stamp rode every export into the repository, where #4485's own `ShippedNodeTypeStateTest` bans it, and a file could then forge a retirement the mesh never measured (the bake gate reads a stamped type's compile failure as `Retired`, i.e. as a verdict that must **not** hold a rollout).

The exclusion was right; the bucket was under-specified. A member can belong to the mesh at the EXPORT and CHANGE-TOKEN seams and still have to lose at the IMPORT seam, because dropping it on re-import is the only thing that clears it — nothing in `src/` ever writes null back to `PendingRetirement`. That is now what the bucket **means**:

- `NodeTypeOperationalContent.StrippedButNotPreserved` — production-side, carrying the reason per entry;
- the strip paths (export, the typed import reset, the change token) run over the **union** with `MemberNames`;
- the preserve path keeps `MemberNames` alone. **That asymmetry is the bucket.**

The test project now reads the production set instead of declaring its own copy — a fourth copy of exactly the list #4485 exists to keep single.

**Negative control:** with the union reverted to `MemberNames`, the new test `StrippedButNotPreserved_LeavesTheFile_AndIsNeverRestoredFromLive` fails on *"the strip removed NOTHING from a node carrying a live pendingRetirement — the stamp rides every export into the repository, where ShippedNodeTypeStateTest bans it"*; green with it.

## 2 — the partition guard's denominator was one member short

It excluded every property carrying a `JsonIgnoreAttribute`, but `NodeTypeDefinition.IncludeGlobalTypes` carries `JsonIgnore(Condition = Never)` — which means it is **always** written. So the guard silently counted one serialised member fewer than it claimed, which is the same class of defect #4480 is about. It now filters on the `JsonIgnoreCondition`, not on the attribute's presence.

## 3 — a `<see cref>` naming a symbol that no longer exists

`CS1574` is an error under `-warnaserror`; the identical mistake in `Mesh.Contract` (a `NodeTypeDefinition` cref that assembly cannot resolve) failed that build during this round, which is the evidence the finding was real.

## 4 — mutable `List<T>` accumulators

Replaced with immutable projections, per the collections policy.

## Measured

Release + `-warnaserror`, `0 Error(s)` on every touched project. Full suites after the change: Graph **1725/1725**, Compiler.Pipeline **990/990**, Memex.Portal.Shared **1606/1606**; doc link integrity + topic map green. Re-verified on this branch after merging current `main`.

No public surface removed, no interface member added, no i18n catalog change.


rbuergi:
Round 2 — all six findings from the last review addressed in 544aff39b3, including the two suppressed ones that are the same defect class this PR is about (a guard that can silently check one thing fewer).

**Denominator regression assertion (the important one).** `EverySerializedMember_IsClassified_ExactlyOnce` only ever iterates `SerializedProperties`, so a filter that drops a member drops its classification too and the test still passes. `IncludeGlobalTypes` is now named explicitly, with `BuildCreate` and `UnknownMembers` as the negative controls for what must stay OUT. **Negative control:** with the attribute-PRESENCE filter restored the test fails *Assert.Contains() Failure: Item not found in collection*; green with the condition-based one.

**`StrippedButNotPreserved` was not pinned against the record.** A typo there names no property, so the member it was meant to strip silently never is — and the partition guard cannot see it, because that one iterates the RECORD's properties. The real-property assertion now covers both lists, with its own diagnostic.

**Two XML contracts went stale with the behaviour change**, both now correct: `PreserveLiveOperational` promised every operational member comes from the live node or is absent (no longer true for the third bucket — it is stripped from the incoming node and never restored, which is what clears it), and `WithTypedMembersReset` still said its member list was `MemberNames` when it is the union. An asymmetry nobody can read from the code is one a maintainer undoes.

**Immutability**: `StrippedButNotPreserved` and `FileExcludedMembers` are `ImmutableHashSet` now — a read-only interface over a mutable process-wide collection is not immutable — and the bucket projection builds its labels from an `ImmutableArray` of (claimed, label) pairs in a named helper rather than a `new[]`.

Measured: Graph **1725/1725**, Compiler.Pipeline **998/998**, Memex.Portal.Shared **1606/1606**, doc link integrity + topic map green; Release + `-warnaserror`, `0 Error(s)` per project.

Note: the PR was in the merge queue while this review was open — converting it to draft does not dequeue — so I dequeued it via `dequeuePullRequest` before pushing. It is held as a draft until CI is green on this head.

PR #4497, opened 2026-09-16
Separate "was this type working" from "is there a baseline to protect" (#4496)

### LINKED ISSUE #4496: DynamicTypePreWarmer: a never-built type reports WasHealthyBeforeBake=false in the satellite lane — core green at the same sha, and it pins every satellite's platform ceiling
`DynamicTypePreWarmerTest.BatchBake_CompilesPendingTypes_InDependencyOrder_AndStampsTheRecords`
fails in MeshWeaver.Plugins' `portal-hosts` lane on core source that core's own CI calls green, and
it is currently what holds that repo's `main` red.

### The failure

MeshWeaver.Plugins run **35082273730** (main, 2026-09-16 10:20Z), leg `portal-hosts (network-133 ·
leg 1/16)`, suite `src/MeshWeaver.Hosting.Monolith.Test`:

```
failed MeshWeaver.Hosting.Monolith.Test.DynamicTypePreWarmerTest
         .BatchBake_CompilesPendingTypes_InDependencyOrder_AndStampsTheRecords (3s 396ms)
  AssertionException : Expected all items in the collection to match the predicate because
  a never-built type is not damaged goods — nothing here may read as a regression that gates a roll
```

i.e. `outcomes.Should().OnlyContain(o => o.WasHealthyBeforeBake, …)`. The preceding assertion
passed — all three types reported `Compiled`, and the run's own output shows the FIRST-BAKE path:

```
[Warning] [BatchBake] DynamicTypePreWarmer: FIRST BAKE — all 3 NodeType(s) are NeverBuilt on this
instance, so there is no previous image to protect.
PreWarmTestBatch/Unrelated  → Compiled
PreWarmTestBatch/Upstream   → Compiled
PreWarmTestBatch/Dependent  → Compiled
```

So at least one never-built type came back with `WasHealthyBeforeBake == false` — the exact
condition the assertion exists to refuse, because that value is what `nodetype_bake` reads when it
decides whether to refuse a new replica's readiness.

### Why it is not "the satellite's diff"

| | |
|---|---|
| core's own `MeshWeaver Build and Test` at the SAME sha `3385e18fb` | **success** (both the `push` and the `merge_group` run) |
| the satellite's diff that main run carried (Plugins#1965) | deletes a dead AI prior-messages loader — it cannot reach the pre-warmer |
| the same leg one platform set earlier — set `3.0.0-ci.8721`, core `ea35bbe1a` (Plugins run 35073843357) | **passed** |
| set `3.0.0-ci.8725` = core `3385e18fb` | **fails** |

The 8721 → 8725 delta is `038b9f708` (#4472, *first-rollout-bake-gate*), `66cc3df5d` and
`3385e18fb` — and #4472 is precisely about the first-rollout / never-built branch this assertion
guards. So either that change made `WasHealthyBeforeBake` reachable as `false` for a never-built
type in an environment core's hosted runners do not reproduce, or the value races the record stamp
and the satellite's self-hosted `aks-silos` runners lose that race.

Both readings matter for the same reason: `WasHealthyBeforeBake` is what decides whether a bake
counts as a REGRESSION that refuses readiness. A never-built type reporting `false` there is the
shape that stalls a roll.

### Why it is urgent beyond one repo

A satellite's pull requests follow *the newest sealed platform set its own `main` has PASSED on*. As
long as `main` cannot pass on 8725, MeshWeaver.Plugins' ceiling stays at **8716** and no PR in that
repo can use anything core shipped after it — today that blocks adopting `--warning-baseline`
(the in-mesh warning ratchets, MeshWeaver#4473), which exists only from 8721. It compounds
MeshWeaver#4491: the one run that DID pass on 8721 lost its annotation to a partial re-run, so that
data point is gone too.

Please read the failure before re-running it — the test passed on the previous set, and "re-run to
see if it was a flake" is exactly what would hide a real race here.

rbuergi:
Fixes #4496. Unblocks MeshWeaver.Plugins' `main`, and with it every satellite PR whose platform
ceiling is pinned behind it.

## What went wrong

#4472 (mine, merged this morning) suppressed the first-rollout readiness gate by emptying
`DynamicTypePreWarmer.RegressionBaseline`, and then stamped the **per-type fact**
`PreWarmOutcome.WasHealthyBeforeBake` from that emptied set:

```csharp
var healthyBefore = RegressionBaseline(report);            // empty on a first bake
.Select(o => o with { WasHealthyBeforeBake = healthyBefore.Contains(o.TypePath) });
```

The gating verdict was right; the statement was false. Every outcome of a first bake claimed
*"this type was broken on the way in"* about types nothing had ever built — and
`NodeTypeBakeEntry.WasHealthy` says the opposite by construction (`State is not PreviouslyBroken`,
so `NeverBuilt` is healthy).

**It also forked a rule that is deliberately written twice.** `BuildProtocolDriver.OutcomesOf`
stamps the same field straight off the entry, and `IsGatingFailure` reads it under a comment
saying the two copies are *"kept in one shape here so the GO and the gate cannot disagree about
what a regression is"*. After #4472 they disagreed — same type, same first bake, GO held, gate did
not. Nothing was red for it: the field's meaning forked, not its signature.

## The fix

Two questions, two fields, and only their conjunction gates.

| field | question | source |
|---|---|---|
| `WasHealthyBeforeBake` | was THIS TYPE working? | `NodeTypeBakeEntry.WasHealthy`, faithful again |
| `HasRegressionBaseline` *(new)* | is there anything here to protect? | `!IsFirstBake(report)` |

`RegressionBaseline` stops special-casing the first bake. Both stampers (`WarmPending` and
`OutcomesOf`) set both fields off the same report, and both readers
(`NodeTypeBakeGateState.MarkOutcome`, `BuildProtocolDriver.IsGatingFailure`) test the pair — so
first-rollout leniency and established-instance strictness are both unchanged, and the two copies
of the rule cannot drift apart again. The new field defaults to `true`, so every hand-built
outcome keeps the strict reading.

## Verification

**Across the repo boundary**, which is where this broke — the failing test lives in
MeshWeaver.Plugins (`src/MeshWeaver.Hosting.Monolith.Test`), so core's suites structurally cannot
run it, which is the whole reason core stayed green:

```
$ dotnet build …/MeshWeaver.Hosting.Monolith.Test.csproj -c Release \
    -p:MeshWeaverRoot=<this worktree> -p:NoWarn=xUnit1051%3BxUnit1069%3BCS1591%3BCS1573 > log 2>&1
  REAL EXIT CODE: 0        Build succeeded.   0 Warning(s)   0 Error(s)

$ ./MeshWeaver.Hosting.Monolith.Test -method "*BatchBake_Compiles*AndStampsTheRecords*"
  REAL EXIT CODE: 0
  Total: 1, Errors: 0, Failed: 0, Skipped: 0, Not Run: 0, Time: 6.529s
```

with the loaded assembly PROVEN to be this branch's, not a leftover:
`md5 MeshWeaver.Hosting.dll` in the satellite's `bin/` is byte-identical to this worktree's build,
and `strings … | grep -c HasRegressionBaseline` is 3 (`BaselineStamp` 2). The suppressed
diagnostics are unrelated to this diff and none are in a file it touches: `xUnit1069` is the known
platform-xunit condition when a satellite suite is built against a core *source* checkout, and the
three `CS1591`/`CS1573` are doc-comment gaps in `InMemoryStorageAdapter.cs:568`,
`ShippedPrebuiltBundles.cs:503` and `AddNodeTypeBakeGate`.

<sub>🚨 My first attempt at this was a false pass and I want it on the record: I piped the build
into `tail`, so the shell reported the pager's exit 0 while the build had printed `386 Error(s)`;
nothing was written; and the `Total: 1, Failed: 0` I read came from a **Sep-4 binary** still sitting
in `bin/Release/`. It is why the md5/symbol check above is quoted rather than assumed.</sub>

The negative control for that green: `git log 3385e18fb..origin/main` over the three touched files
is **empty**, so the code Plugins run 35082273730 measured is byte-identical to today's `main`.

**In core**: `MeshWeaver.Hosting.Test` 36/36 (Release, `-warnaserror`, `0 Warning(s) 0 Error(s)`),
including a new `OnAFirstBake_TheGoAndTheGateReachTheSameVerdict` that asserts the GO and the gate
agree on one first-bake outcome — then flips only `HasRegressionBaseline` and requires both to
gate — plus `AnAlreadyBrokenTypeOnAnEstablishedInstance_StillDoesNotGate` for the other half of the
conjunction. There was no test asserting the two copies of the rule agree; that is the gap that let
this through.

**The controls fired** (a proof step that cannot fail is not a proof step): deleting
`HasRegressionBaseline = hasBaseline` from `BaselineStamp` and rebuilding reds **2 of 2** of the new
stamp tests (`TEST EXIT: 2`, `failed: 2`); restoring it returns the file to **11 of 11 green**.

Pairs-with: none — no public type or member is removed; `HasRegressionBaseline` is an addition to a
record (not an interface), and `RegressionBaseline`'s signature is unchanged. It was introduced
hours ago in #4472 and has zero callers outside `src/` — checked across MeshWeaver.Plugins,
.Education, .Reinsurance, .SocialMedia, .Manufacturing, .Crm and Memex (0 files each).

Mirror-sync: none — no localization key added or re-worded.


rbuergi:
All three taken — and the first two were the important ones, so I fixed the cause rather than adding
assertions around it.

**On "neither production stamper is covered" (the `OutcomesOf` and `WarmPending` findings).** Both
were right, and both defaulted to the *strict* value, which is the dangerous direction: deleting
either assignment would have reverted first-rollout leniency with every test in the file still
green. That is one field away from how #4496 itself arrived.

But the reason there were two uncovered stampers is the same reason the bug existed: **two copies of
one rule**, each deriving the gating facts its own way — the sweep from a set, `OutcomesOf` straight
off the entry. Adding coverage to each copy would have left the fork in place. So:

```csharp
public static Func<PreWarmOutcome, PreWarmOutcome> BaselineStamp(NodeTypeBakeReport report)
```

is now **the only place either field is written**, computed once per report, and both call sites are
a single `stampBaseline(...)`. `OutcomesOf` is `internal` (the test project already has
`InternalsVisibleTo`) so the call site itself is asserted, not just the rule.

Two new tests:

- `TheStampItself_MarksAFirstBakeHealthyAndUnprotected` — the production stamp on a real first-bake
  report yields `WasHealthyBeforeBake == true` **and** `HasRegressionBaseline == false`, with two
  controls: one baked entry restores full strictness, and a `PreviouslyBroken` entry still reports
  `WasHealthyBeforeBake == false`.
- `TheGosOwnProjection_UsesTheSameStamp` — `BuildProtocolDriver.OutcomesOf` over the same first-bake
  report, asserting both fields on every emitted outcome, plus the established-report control.

Deleting either assignment inside `BaselineStamp`, or reverting either call site to deriving its
own value, now reds these.

**On the stale `WithoutBaseline` doc.** Correct, and thank you — it still described the emptied
baseline. Rewritten to name the two independent facts it actually holds
(`WasHealthyBeforeBake == false`, or `HasRegressionBaseline == false`), pointing at `IsFirstBake`
rather than `RegressionBaseline`.

The end-to-end evidence is unchanged and re-run after the refactor: MeshWeaver.Plugins'
`BatchBake_CompilesPendingTypes_InDependencyOrder_AndStampsTheRecords` — the assertion this broke —
passes against this branch.

PR #4498, opened 2026-09-16
An unsubscribe's sender is correlation state, not router traffic (#4489)

### LINKED ISSUE #4489: ROUTER_TRAFFIC residue: JsonSynchronizationStream releases its subscription as the router, and hopping it is a correlation decision (split from #1140)
Split out of #1140 so it is tracked rather than only described. #1140 covered the router-as-an-end
defect class; PR #4487 fixed every instance that reproduces and has a seam to hop onto, and named
this one as the residue that deliberately does **not**.

## The site

`src/MeshWeaver.Data/Serialization/JsonSynchronizationStream.cs`, in `CreateExternalClient`'s release
disposable:

```csharp
var release = new AnonymousDisposable(
    () => hub.Post(new UnsubscribeRequest(reduced.StreamId), o => o.WithTarget(owner)));
```

`hub` is `workspace.Hub`. Where that workspace belongs to the root mesh hub, the unsubscribe leaves
stamped `Sender = mesh/{id}` and `ROUTER_TRAFFIC` reports the router as an end of a work delivery —
the same shape #1140 reports, and `UnsubscribeRequest` is named in #1140's own evidence.

## Why it was not fixed with the rest

**Hopping it is not a no-op**, which is what every other site in that PR relied on. Both issuing
seams (`NodeOperationIssuingHub`, `ReadIssuingHub`) are the identity function off the router, so
adopting one changes nothing for a non-router caller — but this delivery's SENDER is not incidental:
it is what the owner's per-subscriber bookkeeping sees when the subscription is released, and the
subscribe that opened it came from the same hub. Moving only the unsubscribe's origin would leave
the owner holding a subscription opened by one hub and released by another. That is a correlation
question, not a routing one, and it needs a design decision rather than a mechanical hop.

## What "done" looks like

Either

- the unsubscribe and the SUBSCRIBE that pairs with it both move to the same off-router hub (so the
  owner still sees one correlated pair), with whatever `SubscribeRequest.Subscriber` /
  per-subscriber-stream bookkeeping that implies checked explicitly; or
- a reasoned decision that this delivery's sender must stay the stream's own hub, written into
  [Router Traffic Detection](https://github.com/Systemorph/MeshWeaver/blob/main/src/MeshWeaver.Documentation/Data/Architecture/RouterTrafficDetection.md)
  and — since `RouterTrafficRule` is the shared predicate both detector sites evaluate — an
  exclusion there rather than a line in an allow file, so the detector stops reporting a shape
  nobody intends to change.

Note that neither `src/` ratchet currently sees this site: `JsonSynchronizationStream.cs` declares no
router-capable receiver (it calls no seam), so it is outside the receiver-keyed denominator, and
`UnsubscribeRequest` is not a lifecycle message, so it is outside the message-keyed one. The
instrument that names it is the runtime `ROUTER_TRAFFIC ORIGIN` line.

## Verifying it is still live

The measurement in PR #4487 was static (a receiver walk over `src/`). Before working this, re-read
the site — and confirm on a live deployment whether the origin detector is actually reporting
`UnsubscribeRequest` with the mesh hub as sender, since a workspace whose `Hub` is never the root
mesh hub would make this theoretical.

rbuergi:
Fixes #4489. Takes the **second** of the two paths that issue lists under "What done looks like" —
a reasoned decision that this delivery's sender must stay the stream's own hub, recorded in
[Router Traffic Detection](src/MeshWeaver.Documentation/Data/Architecture/RouterTrafficDetection.md)
and carried as an exclusion in `RouterTrafficRule`, the shared predicate both detector sites
evaluate.

## Why the second path and not the first

The first path — move the unsubscribe **and** its subscribe to the same off-router hub — is not
available at an acceptable price, and the code says so at the site:

- `JsonSynchronizationStream.CreateExternalClient` binds `var hub = workspace.Hub` once and posts
  **both** halves on it: the subscribe at `postSubscribeRequest`, the release in its disposable. The
  owner's per-subscriber stream is keyed on the subscriber that opened it, so hopping **only** the
  unsubscribe leaves the owner holding a subscription opened by one hub and released by another —
  exactly what #4489 warned about.
- Hopping **both** changes `SubscribeRequest.Identity`. That file calls `MintSubscribeRequest` *"THE
  ONE PLACE a C# subscriber mints a SubscribeRequest"*, and documents the failure mode of getting its
  identity wrong: a wiped `AccessContext` fails closed → the owner denies the subscribe → the
  consumer re-opens → a flood of denied `SubscribeRequest`s.

So the sender here is **correlation state, not an incidental origin**, and a `mesh/{id}` sender on
this release is correct. Reporting it asks for a change nobody may make, which is the shape that
trains people to mute the channel — the same argument the rule already makes for `HeartBeatEvent`
and for the routing NACK.

## The shape of the exclusion

A marker interface in `MeshWeaver.Messaging.Contract`, implemented by `UnsubscribeRequest`:

```csharp
public interface ICorrelatedBySender;
```

- **A type check, not a name match** — matching the `HeartBeatEvent` exclusion's own reasoning that
  *"a rename must not silently turn this exclusion off"*. A direct `message is UnsubscribeRequest`
  is impossible: that type lives in `MeshWeaver.Data.Contract`, which references
  `MeshWeaver.Messaging.Contract` and not the reverse.
- **On the message rather than in an allow file**, as #4489 asks — the claim travels with the
  contract and is visible where the decision applies.
- **Exactly as narrow as the site**: `new UnsubscribeRequest(` appears **once** in all of `src/`,
  at the site in question.

The interface's doc says what implementing it claims, because the risk of a marker is that someone
uses it to quiet an inconvenient report: it is a statement about the RECEIVER — that an earlier
delivery already told it to remember this sender — and a message whose sender the receiver does not
correlate has no business carrying it.

## Verification

Core, Release with `-warnaserror`, `0 Warning(s) 0 Error(s)` on all four touched projects;
`RouterTrafficRuleTest` **15/15** and the new `AnUnsubscribeIsCorrelatedByItsSenderTest` **2/2**.

**The controls fired.** Removing the `ICorrelatedBySender` branch from the rule and rebuilding:

| | |
|---|---|
| `AMessageCorrelatedByItsSender_IsNeverReported` | **1/1 failed** |
| `AnUnsubscribeIsCorrelatedByItsSenderTest` | **1 of 2 failed** |

The 1-of-2 is the discriminating part: `TheReleaseCarriesTheClaim` reds, while
`TheSubscribeDoesNotCarryIt_AndIsStillReported` keeps passing — so the pair distinguishes the
exclusion from the ordinary behaviour rather than moving together. Restoring the branch returns both
to green.

Two tests keep the exclusion from widening: `AnOrdinaryMessageInTheSamePositions_IsStillReported`
pins that a plain message is still reported as `target` / `sender` / `sender AND target`, and the
link test pins that `SubscribeRequest` deliberately does **not** carry the marker — silencing both
halves would remove the detector's view of subscription traffic entirely.

## The live measurement behind it

Posted on #4489 earlier today: over a 700-minute window covering both `memex` pods' startup, the
origin detector reported `DisposeRequest` and `RawJson` and **not** `UnsubscribeRequest`. That
supports the site being theoretical as it stands, but it is deliberately **not** the argument for
this change — the detector dedups once per role+type per process lifetime, so a clean window proves
"not since these pods started", never "never occurs". The argument above stands on the code.

Pairs-with: none — no public type or member is removed. `ICorrelatedBySender` is a new marker
interface with no members (so nothing to declare under the interface-additions gate), and
`UnsubscribeRequest` gains an interface, which is additive.

Mirror-sync: none — no localization key added or re-worded.


rbuergi:
Both findings are correct. I **dequeued this PR from the merge queue** to fix them rather than let it
land and follow up — the first one is a false claim in the doc and the tests, which is the kind of
thing that is much more expensive to unwind after it has merged.

## 1. The exclusion does not reach the receiver site — confirmed, and I had already measured it

Verified in the code rather than taken on trust:

```csharp
// MessageHub.DeliverMessage
ReportRouterTraffic(delivery);                              // :2007
var ret = delivery.ChangeState(MessageDeliveryState.Submitted);
var result = messageService.RouteMessageAsync(ret, default); // :2009  ← unpacking happens here
```

And my own live reading on `memex`, posted on #4489 earlier today, is exactly the pair your comment
predicts — I had the evidence in hand and still wrote "both sites":

```
ROUTER_TRAFFIC ORIGIN: DisposeRequest was POSTED with the mesh hub as sender   ← typed
ROUTER_TRAFFIC:        RawJson has the mesh hub as sender                      ← packed
```

**What I changed, and what I deliberately did not.** I narrowed the claim instead of carrying the
marker through the delivery envelope. Putting a detector concern into the wire format for one
message is a worse trade than stating the scope: the ORIGIN line is the one #4489's evidence names
and the only one that carries a call site an engineer can act on, and the receiver-side `RawJson`
line names no message type and no call site and prints for **every** cross-hub delivery with the
router as an end — a pre-existing property, not a hole this opened. Narrowing that is a different
change, about where the receiver detector runs relative to unpacking.

So: the doc's "excluded at both sites" is gone, replaced by a scope section that quotes the call
order and the measured pair; the rule's comment says which caller still holds the type; and there is
a new test, `TheExclusionCannotReachAPackedDelivery`, asserting that a packed `RawJson` carrying an
unsubscribe **is still reported** (`RoleOf("portal", "mesh", packed) == "sender"`). That pins the
boundary, so the next reader cannot assume cross-hub coverage the way I did.

I did not add a cross-hub integration test: the unit assertion above pins the same fact at the
predicate, and a hub-level test would be asserting `MessageHub`'s packing order rather than this
change.

## 2. The stale #4489 residue block — fixed

You are right that one page gave opposite instructions. The older block still read *"One named
residue stays open and is deliberately not fixed here"* and described the hop as an open design
question. It now records that the correlation question was **answered** — the sender is correct, the
detector stops asking — and points at the new section for the argument, the marker, and the scope
note, so the two halves of the page agree.

PR #4516, opened 2026-09-16
A node landing page carries provenance by default (#4500)

### LINKED ISSUE #4500: Provenance is opt-in per layout area, so at least three node types lose it silently
## Why this belongs in MeshWeaver core

Everything involved is core's and unreachable from a consumer repo. `MeshNodeLayoutAreas.BuildMetaRow` (`MeshNodeLayoutAreas.cs:783`) composes the row; `BuildHeader` places it as row 2 of the standard header (line 563); the `node.meta.*` keys live in `MeshWeaver.Messaging.Hub/Localization/strings.{en,de}.json`; and every layout area that does or does not call it is a built-in serving a built-in NodeType. A module can write its own header — the CRM does — but it cannot change what the framework's own pages render.

## What

`BuildHeader` is **opt-in**. A node type whose overview is custom-built shows no provenance unless its author remembered to call it, and nothing flags the omission. In a current checkout, three files call it and six do not:

| calls `BuildHeader` | does not |
|---|---|
| `MarkdownOverviewLayoutArea` · `NodeTypeLayoutAreas` · `MeshNodeLayoutAreas` | `CodeLayoutAreas` · `MarkdownLayoutAreas` · `MarkdownNotebookLayoutArea` · `ActivityLayoutAreas` · `UserActivityLayoutAreas` · `PresentationLayoutArea` |

## Measured, 2026-09-16

Renders the row:

| node | type | portal |
|---|---|---|
| `CustomerJourneys/B2C/Pro` | `Markdown` | memex.meshweaver.cloud |
| `PearlTechnology` | `Crm/Client` | memex.systemorph.com |
| `PearlTechnology/Commercials` | `Crm/Offer` | memex.systemorph.com |

Does not:

| node | type | what renders instead |
|---|---|---|
| `CustomerJourneys` | `Space` | icon, title, description, then an **empty** `StackControl` — `{"areas":[],"style":"gap: 24px; margin-top: 12px; flex-wrap: wrap;"}` |
| `Doc` | `Space` | the same empty stack, identical style, one index earlier (no description) |
| `Templates/Export/Pdf` | `Code` | no row and no type link — zero matches for `Created:` or `Updated:` in the rendered area |
| `Essentials/Skill/communicate` | `Skill` | no standard header at all: `Overview` is `<h1>/communicate</h1>` plus three markdown blocks |

The state is present on every one of these nodes — e.g. `CustomerJourneys` carries `createdBy: rsalzmann`, `createdDate: 2026-09-15T06:53:47Z`, `lastModifiedBy: rsalzmann`, `lastModified: 2026-09-15T07:17:01Z`, `version: 2`. Only the header omits it.

## Three observations that shape the fix

1. **`Code` is not a partition root**, which rules out "roots are special" as the explanation. The behaviour tracks the layout area, not the node's position.
2. **`Skill` is the worst case.** It has no standard header at all, so a reader cannot see the type, the author, or the date on a node whose entire purpose is to be shared and reused.
3. **`Space` reserves the slot and leaves it blank.** Both instances build a container styled exactly like the meta row and put nothing in it. That reads as a row that was intended and lost, rather than one never considered — worth a look at the history before assuming it was always so.

## Coverage — what I did not check

Verified: `Markdown`, `Crm/Client`, `Crm/Offer` (have it); `Space` ×2, `Code`, `Skill` (do not).

**Not verified**, and suspected from the table above: the types served by `MarkdownNotebookLayoutArea`, `ActivityLayoutAreas`, `UserActivityLayoutAreas` and `PresentationLayoutArea` — notebooks, `Activity`, `Thread` and presentation pages. `NodeType` should have it, since `NodeTypeLayoutAreas` calls `BuildHeader`, but I did not confirm it on a live node. The list above is a floor, not a census.

I also could not locate the Space overview renderer: the style string `margin-top: 12px; flex-wrap: wrap` appears in neither a current `Systemorph/MeshWeaver` nor `Systemorph/Memex` checkout, and there is no `SpaceOverview`/`SpaceHome` file in either. Cloud appears to run ahead of both.

## The question worth deciding

Patching the Space overview fixes one of three. The design question is whether the standard header should be **opt-out** rather than opt-in — rendered by default, with a custom overview explicitly declining it — so that the next custom overview cannot lose provenance by omission.

If opt-in stays, the cheaper mitigation is a test that walks every registered NodeType's default area and asserts the meta row is present or the type is on an explicit allow-list, which turns a silent omission into a decision someone has to write down.

## Not proposed here

Which types *should* show the row. A `Thread` or an `Activity` may have good reason to omit it; that is a judgement for whoever owns those pages, and this issue asks the question rather than presuming the answer.

rbuergi:
Closes #4500.

## The hypothesis held, and the number is an order of magnitude larger than the floor

The issue reported "at least three" node types losing the provenance line. Census, 2026-09-16,
over landing pages that REPLACE the framework's `Overview` renderer:

| Repository | Replace the renderer | Carry the line | Do not |
|---|---:|---:|---:|
| MeshWeaver (`src/`) | 7 | 3 | **4** |
| MeshWeaver.Plugins (compiled + in-mesh) | 92 | 10 | **82** |
| **Total** | **99** | **13** | **86** |

Core's other **61** node types never replace it and were never at risk. The four in core that had
lost it are `PluginCatalog`, `GlobalSettings`, `User` and `Partition` — the last being the quietest,
because `PartitionNodeType` mentions no page at all: it repoints the default area with
`WithDefaultArea(SearchArea)` and the landing page moves off the framework renderer as a side
effect.

Two corrections to the issue's table, both from reading today's `main`: `MarkdownLayoutAreas` and
`CodeLayoutAreas` are registration/constant files — Markdown's page *does* call `BuildHeader`
(`MarkdownOverviewLayoutArea.cs:235`), and `Code`'s renderer is not in this repository at all
(`MeshWeaver.Graph.Views`, in Plugins). `Skill` is likewise the AI engine's, not core's. The
file-level instrument counted files, not landing pages.

## Why not three call sites

At 87% of pages that opt out of the framework renderer — and with 82 of the 86 in a repository no
core reviewer reads — patching sites leaves the next page to be forgotten. The default had to
change.

`MeshNodeLayoutAreas.WithNodePage(area, page)` replaces the bare `WithView` for a landing area, and
its **no-argument form composes the provenance line above the page**. A landing page written by
somebody who never thought about provenance now ships with it; the failure mode flips from "the
line silently disappears" to "the line silently appears", and only the second is visible.

The two other answers are declared, never inferred:

- `NodePageProvenance.RenderedByThePage` — the page calls `BuildHeader`/`BuildMetaRow` itself.
- `NodePageProvenance.Declined("<why>")` — **the reason is mandatory**; `Declined("")` throws. A
  blank decline records that somebody decided without recording what, which is the state the issue
  is about.

`LayoutDefinition.GetNodePageProvenance(area)` reads the verdict back, and `null` ("nobody decided")
stays distinguishable from `Declined` ("decided no") — the distinction the issue's second commenter
identified as the whole defect.

### The verdict cannot go stale

`LayoutDefinition.WithNamedRenderer` — which every `WithView` overload funnels through — now
**removes** the area's verdict. A node type taking over a landing page is exactly the move that
loses the line; if the framework's own verdict survived that takeover, every check downstream would
pass having measured a renderer that is no longer registered.

🚨 That is a behaviour change behind an **unchanged public signature** (break shape 7). It is
additive in effect — nothing outside this repository reads `NodePages`, which did not exist before —
but it is stated here rather than left to be discovered.

## What core decided for its own four

| Node type | Verdict | Why |
|---|---|---|
| `PluginCatalog` | framework-supplied | a catalog node is declared, published and re-pointed by people |
| `GlobalSettings` | framework-supplied | "who last changed the platform's settings" is the first thing an admin on that page wants |
| `Partition` | framework-supplied | an administrative record whose first question is who declared the space, and when |
| `User` | **declined, with the reason in code** | the subject is a PERSON and the node's stamps describe the ROW — `LastModified` moves when a preference is flipped, `CreatedBy` is the sign-in flow — so "Updated … by system-security" across somebody's home states a database fact to a reader asking about a colleague |

The three compile/refusal overlays also decline: they are not the node's page but what stands in its
place while the page cannot be built, and the node's stamps describe the INSTANCE while the broken
thing is its TYPE.

## The guard, and what it deliberately cannot see

`NodePageProvenanceGuard` scans `src/` for a landing-area renderer registered through the bare
`WithView`/`WithNamedRenderer` and reds naming the file; exceptions are reasoned lines in
`test/NodePagesWithoutProvenance.allow`, which shrinks only (a stale entry reds, so the reason is
re-read when it stops being true). Two entries today, both `Search`-as-a-tab.

It **over-flags on purpose**: an area name is a landing page on one hub and a tab on another
(`Search` is `Partition`'s landing page and every other node's search tab) and a scanner cannot tell
which hub a registration lands on. Over-flagging costs a reasoned line; under-flagging costs a page
nobody notices is missing something.

🚨 **It cannot see the 82.** Most of the fleet's landing pages are in-mesh `Source/*.cs` and NodeType
`configuration` lambdas that compile at RUNTIME in the portal — invisible to every compiler and
every scan in this repository. Those are reached by the changed DEFAULT, not by the guard: a page
that moves to `WithNodePage` gains the line, one that stays on `WithView` stays exactly as it is
(nothing in this PR changes their rendering). Widening the guard's roots would produce a guard that
scans a tree the defect does not live in. That is written down in the doc page and in the allow
file's header rather than left implicit.

## Reproduction and negative controls

**The defect, reproduced against a real mesh** (`NodePageLosesProvenanceTest`) — one variable, three
halves, same node, same viewer, same page content, same area key `Overview/NodeMeta`:

| Registration | Page renders | Provenance strip |
|---|---|---|
| `WithView(OverviewArea, page)` | yes | **absent** |
| `WithNodePage(OverviewArea, page)` | yes | **present** |
| `WithNodePage(…, Declined(reason))` | yes | absent, by declaration |

The negative half waits for the page's own control *first* and only then asserts the strip's
absence — otherwise it would also pass on an empty screen.

**Negative control on the anti-stale invariant, measured.** With `NodePages.Remove(area)` deleted
from `WithNamedRenderer`, `A_verdict_does_not_survive_the_renderer_it_describes` fails with
*"Expected value to be null … but found RenderedByThePage"* — i.e. the takeover inherits the page it
replaced. 1 of 8 failed; restored, 8/8 pass.

**Negative control on the guard, measured.** Reverting one migrated page to `WithView` reds the
guard naming it: *"src/MeshWeaver.PluginCatalog/CatalogLayoutAreas.cs: 1 landing-area registration(s)
without a provenance verdict, 0 allowed"*. Restored, green. The guard also carries a planted-snippet
self-test so a regex that stopped matching cannot leave it green forever.

## Measuring a rendered page

The strip renders into a stable area id, `MeshNodeLayoutAreas.NodeMetaArea` (`NodeMeta`) — so its
store key always ends in `/NodeMeta`. Previously the only way to answer "does this page carry
provenance?" was to read the rendered HTML for the English word `Created:`, which asserts that a
German reader sees an English page.

## Verification

- `dotnet build -c Release -warnaserror` — `MeshWeaver.Layout`, `MeshWeaver.Graph`,
  `MeshWeaver.PluginCatalog`, `MeshWeaver.Graph.Test`, `MeshWeaver.Layout.Test`,
  `MeshWeaver.Documentation.Test`, and the whole `MeshWeaver.slnx`: **0 Warning(s), 0 Error(s)**.
- `MeshWeaver.Documentation.Test` whole project: **496/496** (covers the topic-map entry, the
  What's New `Order:`, doc-link integrity and the new guard).
- `MeshWeaver.Layout.Test` whole project: **504/504**.
- `MeshWeaver.Graph.Test`: the whole project, plus the new `NodePageProvenanceVerdictTest` (8),
  `ABareWithViewLosesProvenanceTest`, `WithNodePageSuppliesProvenanceTest`,
  `ADeclinedNodePageCarriesNoStripTest`.

Mirror-sync: none — this change adds no localization key and re-words no catalog value. The composed
strip reuses the existing `node.meta.type` / `.created` / `.updated` / `.lastActivity` / `.by` keys
through `BuildMetaRow` → `host.Localize`, so `clients/react/src/i18n/` has nothing to take.

Work product: [Node Page Provenance](src/MeshWeaver.Documentation/Data/Architecture/NodePageProvenance.md)
(`get Doc/Architecture/NodePageProvenance`) carries the census, the reasoning for inverting the
default rather than patching sites, the per-type decisions, and both blind spots.


rbuergi:
## Review findings — all four acted on, plus the CI red (which was mine)

**CI red, root cause.** The two NEGATIVE halves of the experiment used `NotEmit(TestTimeouts.Quick)`. A negative window always spends its whole budget, and `Quick` is `Convergence / 3` = **36 s on a CI runner** (Factor 3.0) — above the 30 s `methodTimeout` in `test/xunit.runner.json`. So they passed locally (`Quick` ≈ 12 s) and timed out *anonymously* on every runner: `Test execution timed out after 30000 milliseconds`, no assertion, nothing named.

Neither raising a bound nor writing a shorter literal is the fix — the timer was never needed. `ComposeProvenance` emits the strip and the page as **one** control, `Stack(NodeMeta, page)`, so there is no window in which the page is on screen and the strip is still in flight. The assertion is now on the landing area's **root control**: does its area list carry a child whose rendered key ends in `/NodeMeta`? Decidable from one rendered value — no budget, no CI scaling, and a stronger statement than "nothing emitted within N seconds". Each half still proves the page RENDERED in the same tree it claims the strip is missing from, and the positive half is the detector's own control. The three classes now run in **1.7 s** instead of 25 s.

**1. `ScannedRoots` is static mutable state** — correct, and it is the house rule rather than a style point (`Doc/Architecture/NoStaticState`: only an *immutable* static readonly lookup is allowed). Now `ImmutableArray<string>`.

**2 & 3. Both front-matter descriptions ended `…had.`** — correct. The elliptical "86 of the 99 … had [lost it]" is defensible prose but reads as a fragment in a catalog subtitle, which is the only place that string appears. Both now read "had lost it."

**4. "Every node page shows a provenance line" overstates what shipped** — the most useful of the four, and correct. A landing page still registered with a bare `WithView` renders exactly as it did; only the default changed. The opening is re-scoped, and the release note now names the eighty-two plugin pages as **unchanged**, with the handover tracked as Systemorph/MeshWeaver.Plugins#1973 — which is a better outcome than softening the sentence, because that debt was previously visible only inside this PR's body.

Re-verified after the change: `MeshWeaver.Documentation.Test` 496/496, the four new classes 11/11, both touched test projects built `-c Release -warnaserror` at 0 Warning(s) / 0 Error(s).

Issue #2394, opened 2026-08-26
MeshWeaver.Hosting.Monolith.Test wedges in TEARDOWN with exit=124 — a post-#2335 recurrence with a different mechanism from #2013/#2341


rbuergi:
## The signature

```
[CI] MeshWeaver.Hosting.Monolith.Test (part 2/2) exit=124 TIMEOUT (8m wall-clock cap hit — likely fixture/init hang)
```

Three occurrences in the last 250 `MeshWeaver Build and Test` runs (2026-08-25T21:59Z → 2026-08-26T14:16Z), on three different branches — so it is branch-independent, roughly a 1.2 % per-run flake, and it reds whichever PR it lands on:

| run | branch | when |
|---|---|---|
| [32939560960](https://github.com/Systemorph/MeshWeaver/actions/runs/32939560960) | **`main`** | 2026-08-26T06:45Z |
| [32965944797](https://github.com/Systemorph/MeshWeaver/actions/runs/32965944797) | `refactor/test-suites-drop-ai-batch` | 2026-08-26T11:56Z |
| [32977048720](https://github.com/Systemorph/MeshWeaver/actions/runs/32977048720) | `fix/2341-unthrottled-resubscribe-busy-wait` (#2390) | 2026-08-26T13:54Z |

🚨 **The 11:56Z and 13:54Z occurrences are AFTER #2335 merged (08:58Z)**, so this is not residue #2335 already closed — and it is a different mechanism from both #2013 (a blocking bridge parking a thread) and #2341 (a busy-wait spinning one).

## Where it hangs — teardown, not the test body

From the failing shard's own artifact (`collected-logs/_meshweaver-test-trace.log`, run 32977048720 shard 2): the wedged host, pid 3520, logs **92 `INIT_DONE` against 91 `DISPOSE_DONE`**. The unbalanced class is **`MeshNodeLanguageServiceTest`**, and it is the *second* instance of that class — the first disposed cleanly in 2026 ms.

```
14:09:13.189 [MeshNodeLanguageServiceTest] CTOR
14:09:13.227 [MeshNodeLanguageServiceTest] INIT_DONE elapsed=37ms
14:09:14.269 [MeshNodeLanguageServiceTest] DISPOSE_START
14:09:14.269 [MeshNodeLanguageServiceTest] DISPOSE_HOSTED_SERVICES_STOPPED elapsed=0ms
[...]

PR #4523, opened 2026-09-16
Teardown no longer mints the thread that delivers a terminal (#4448)

### LINKED ISSUE #4448: IoPoolTest.Dispose_doesNotBlockOnASlowPooledSubscriptionTeardown: disposal did not terminate the pooled subscription within 5 s (queue, once; mechanism open)
**`IoPoolTest.Dispose_doesNotBlockOnASlowPooledSubscriptionTeardown` failed once in the merge queue: disposal did not terminate the pooled subscription within 5 s. The mechanism is not established.**

## The occurrence
- **Run:** queue group build [35003438933](https://github.com/Systemorph/MeshWeaver/actions/runs/35003438933), job 104498266018 (`Run tests (shard 3)`), a group of one: #4423 alone, on `main` `d97160f07`.
- **#4423's diff:** `KernelContainer`'s idle timer. It cannot reach `IoPool`.
- **#4423's own PR run** was green, including this test.
- **The failure:**
  ```
  MeshWeaver.Reactive.Assertions.ObservableAssertionException : Expected the observable to emit a value
  within 5s because disposal must still terminate the subscription, but it did not. The observable emitted nothing at all.
    at IoPoolTest.Dispose_doesNotBlockOnASlowPooledSubscriptionTeardown() IoPoolTest.cs:line 1120
  ```
- **Load:** `MeshWeaver.Hosting.Test` ran 718 tests in 333 s on that shard. The test's own log window is about 8.4 s, from the previous test's last line at 17:54:16.97 to the failure at 17:54:25.36. The preceding `SpinWait.SpinUntil(() => source.HasObservers, 5 s)` **passed**, so the timings fit the pooled subscribe landing about 3.4 s late on a busy ThreadPool (`SubscribeOn(TaskPoolScheduler.Default)`), followed by `Within(5 s)` expiring.

## What reading the code rules out
- **ThreadPool starvation of the cancel:** `Dispose` issues `_poolCts.Cancel()` from a dedicated `IoPool-cancel` thread (`StartCancelOffCallerThread`), never the ThreadPool or the pool's own scheduler.
- **Resources disposed under the cancel:** the cancel holds its own gate region, taken on the caller's thread before `Start()`, so `TryFinishDisposal` cannot dispose `_poolCts` first.
- **A lost terminal:** the drain registration calls `inner.Dispose(); observer.OnCompleted()` once through a latch. The setup's error arm also turns an `OperationCanceledException` into `OnCompleted`.

## What is still open
Nothing found by reading holds the cancel thread for 5 s, so this is **unattributed**, not "a flake". The next measurement is a deterministic or stress repro:
- the test class under a saturated ThreadPool (for example, blocking N ThreadPool workers while the test runs);
- with the `IoPool-cancel` thread's start and the drain callback timestamped.

That shows whether the terminal is late (scheduling) or never comes (a missed registration).

**Not catalogued.** `.github/known-flakes.json`'s own seed note records that single occurrences did not meet its bar, and this is one. #4423, whose diff cannot reach this test, stays `queue-rejected` until this is fixed or the evidence meets that bar.

rbuergi:
Fixes #4448 (the half #4466 did not explain).

## The defect

`Drain()` and `Dispose()` called `new Thread(...).Start()` **per call**, so every pooled
subscription's terminal waited on an OS thread being *created* and then *first scheduled*. Two
reasons that is a defect rather than a cost:

* **It is a blocking call inside the method whose contract forbids blocking.** `Thread.Start()` takes
  the runtime's thread store lock and can queue behind a GC suspension. `Dispose()`'s own comment,
  three lines above the call, is *"DISPOSE MUST NOT BLOCK"* — the same shape as the inline `Cancel()`
  that #2394 removed from it, one layer down.
* **The latency it adds is invisible.** `Dispose` returns in 0 ms, the awaited subject is a latched
  `AsyncSubject`, no gate permit moves, `CurrentInFlight`/`CurrentlyWaiting` are both 0 — which is
  exactly what the issue's earlier probes measured and why they found nothing.

## The change

The canceller is **started with the pool and parked** on a one-shot latch (a
`CancellationTokenSource`'s wait handle: raised once, never resets, and held for the pool's whole
lifetime, which is the one wait a *slim* spin-then-block primitive is documented not to be for).
`Drain()`/`Dispose()` raise the latch instead of minting a thread, and `Drain()` joins on a
`ManualResetEventSlim` the canceller sets once `Cancel()` has *returned* — under the same budget,
exactly like `_blockingIdle` — in place of `Thread.Join`.

The requester still takes the gate region **before** raising the latch, as it did before `Start()`,
so `_poolCts` is provably alive when the canceller wakes. The canceller publishes completion
**before** handing that region back, because the hand-back can complete disposal and dispose the very
event a `Drain()` would be waiting on. This is not the #4466 inversion: what a joiner waits on is
still the last thing the fact it asserts publishes.

`IoPoolRegistry` creates pools lazily by name, so the cost is one parked stack per resource class
actually in use, and the thread exits as soon as the one cancel it exists for has run.

## The measurement

The race is two thread-state transitions wide and 26 class runs across three load regimes reproduced
nothing, so it is established the way #4466's was — by **widening the window in place**. Two
wideners, because there are two latencies and only one is the pool's to remove:

| widener (6 s, in place) | before | after |
|---|---|---|
| **A** — thread *creation* costs 6 s (sleep before `Start()`) | **FAILS**: `Expected 00:00:06.0026424 to be less than 00:00:02 … Dispose must return immediately` | **PASSES** (6.1 s test, all of it in the constructor) |
| **B** — the canceller gets no timeslice for 6 s (sleep before `Cancel()`) | **FAILS**: *"…emit a value within 5s … The observable emitted nothing at all"* | **FAILS**, identically |
| none | passes | passes |

**A is the discriminator.** **B is the honest half**: it reproduces the merge-queue failure verbatim,
message included, and the fix does not change it — "the OS did not run the thread" cannot be designed
away while the cancel is forbidden to run on the caller. What the change removes from that path is
thread creation, runtime registration and first-scheduling; what remains is waking a thread that
already exists.

So the next occurrence is no longer ambiguous: `Dispose_doesNotBlockOnASlowPooledSubscriptionTeardown`
now asserts `CancellerIsAlive` **before** the disposal it measures, and
`TheCancellerThreadIsStartedWithThePool_NotAtTeardown` pins the property on its own. A green line
there over a red terminal is a host verdict, not a pool defect — and still never a reason to widen
the 5 s bound.

## One forensic correction, recorded in the doc

The original triage read the failure as the pooled subscribe landing ~3.4 s late on a busy
ThreadPool, on the strength of *718 tests in 333 s* on that shard. `test/xunit.runner.json` sets
`maxParallelThreads: 1` and `parallelizeTestCollections: false`, so **that assembly runs one test at a
time** — the 718 tests were never concurrent demand, and only one `IoPoolTest` case ran in that job.
The 8.39 s hole in the runner log is this test plus the preceding class's mesh teardown. The test's
own precondition also rules the ThreadPool leg out directly: `source.HasObservers` is true only once
`source.Subscribe(observer)` has run, so the subscribe had already landed before `Dispose()`.

## Also in this PR

`IoPoolQueueReportTest.ABlockingAdmissionCancelledBeforeItStarts_DoesNotLeakTheWaitingGauge` waited
on `CurrentlyWaiting > 0` — already true with **both** leaves queued, because the gauge is raised on
subscribe — and then asserted exactly `1`. This change's timing exposed it (`but found 2`). It now
waits for the state it is about: one leaf running, one queued.

## Verification

`MeshWeaver.Mesh.Contract`, `MeshWeaver.Hosting`, `MeshWeaver.Hosting.Test`,
`MeshWeaver.Documentation.Test` and `MeshWeaver.Hosting.Orleans.Test` all build `-c Release
-warnaserror` with **0 Warning(s) / 0 Error(s)**. All 55 `IoPool*` tests pass, all 3
`IoPoolSiloTeardownTest`, and `MeshWeaver.Documentation.Test` 496/496 whole (which is what caught the
blocking-bridge ratchet on an earlier spelling of the park — the park is written as a latch on its
merits, not spelled around the marker).

Analysis committed to `Doc/Architecture/ControlledIoPooling`.


rbuergi:
Review addressed in 2b7a023cfe. Three findings were valid and are fixed; two are answered below.

**Fixed — the losing `GetOrAdd` candidate (critical, and a regression this PR introduced).** Right:
`ConcurrentDictionary.GetOrAdd` does not promise its value factory runs once, and the discarded pool
used to be plain garbage. It no longer is — a pool owns a started thread from its constructor, and a
thread is a **GC root**, so a dropped candidate parked an `IoPool-cancel` thread for the process's
life. `Get` now records what it built and disposes anything that lost the race, which wakes that
canceller and lets it exit (the same thing the refusal path already did with `raced`).

**Fixed — `ExecutionContext` capture.** Also right, and I had not considered it. `Thread.Start()`
flows the starting thread's context for the thread's whole life; the canceller is started from
wherever a pool is first resolved, which can be a hub turn serving a viewer, and the cancel it later
runs executes every pooled subscription's downstream teardown. That would have run all of it under
that user's `AsyncLocal` identity — `AccessService.Context` included — and pinned it until the pool
died. Now `UnsafeStart()`, which restores what the old shape gave for free: the thread used to be
created on the mesh-teardown thread, which carries no context.

**Fixed — two stale identifiers.** The constructor comment still named `_cancelSignal`, and
`Doc/Architecture/WhichKindOfSilence` still cited `IoPool.StartCancelOffCallerThread`. My sweep for
the old name covered `*.cs` only; the doc tree is exactly where that misses.

**Answered — `Thread.Start()` moving onto a lazy first `Get`, which can be a hub turn.** A real
observation and a deliberate trade, not an oversight. It is now **once per pool name per mesh**,
against **once or twice per pool per teardown** before — and the teardown one sat on a path with a
5 s wall-clock contract that it measurably missed, which is what this PR fixes. `Get` already builds
a `SemaphoreSlim`, a `LimitedConcurrencyLevelTaskScheduler`, a `TaskFactory` and a
`CancellationTokenSource` on that same thread, so the caller already pays construction there; a
thread start is more expensive but the same category, and it is paid once. Moving it to a startup
boundary would mean eagerly minting every pool name, which is the thing the registry's laziness
exists to avoid ("Wave-2/3 pool names cost nothing until a leaf actually touches that resource
class"). If pool construction on a hub turn is itself worth removing, that is a separate change to
`Get`, not to this thread.

**Answered — `CancellerIsAlive` does not prove every terminal runs on that thread.** Correct, and it
does not claim to. Its `because` is *"the canceller is started with the pool, not at teardown"*, and
that is the whole property: no terminal waits on an OS thread being **created**. The late-
registration path you describe — `Register` on an already-cancelled token running inline on the
subscribing thread during a `Drain()` — is real and **pre-existing** (`Drain` sets `_draining` but
not `_disposing`, so `TryEnterGateRegion` still admits), and it is a subscriber that arrived at a
dead pool getting its own teardown on its own thread. It is not reached by this change and is not
what #4448 measured; filing it separately is the right home for it, not widening this PR.

rbuergi:
The late-registration path from the last review point is now tracked as #4524 — pre-existing, narrow, not observed in the wild, and recorded as a reading of the code with the two candidate shapes and why neither belongs in this PR.

PR #4527, opened 2026-09-16
The catalog listing transferred the whole repository to read its manifests (#4222)

### LINKED ISSUE #4222: The plugin registry still serves the pre-#4246/#4527 catalog listing — memex.meshweaver.cloud is 934 commits behind and has never run either fix
## What is failing

`GET /api/plugins` — the plugin registry's **catalog listing** — does not answer inside a 30-second
attempt budget, roughly **60 times a day**, from a single consumer.

This is the residue of #1134 after that issue's own defect (the timeout events carried an empty
`Operation Key` and could not be attributed to any call path) was fixed by #1133/#1137. The
attribution is live and now names the client; what it names is a real latency defect that no issue
tracked.

## The measurement

Read from the incident node the log watcher folds these events onto —
`Admin/_LogIncident/9ca334c1e8dad9ca` on memex.systemorph.com, 2026-09-13:

| | |
|---|---|
| occurrences | **2,028** |
| first seen | 2026-08-10 01:10:16Z |
| last seen | 2026-09-12 16:10:52Z |
| window | 33.6 days ⇒ **~60 / day** |
| namespace | `memex` — the control instance, calling the registry on `memex.meshweaver.cloud` |
| all 10 retained samples | `EventName: 'OnTimeout', Source: 'plugin-registry-standard//Standard-AttemptTimeout'` |

Every retained sample is an **attempt** timeout on the `plugin-registry` pipeline. The one
`Standard-TotalRequestTimeout` line in the incident's history is from 2026-08-10, under the unnamed
`-standard` defaults (10 s attempt / 30 s total) that #1133/#1137 replaced.

## Which call

`plugin-registry` (30 s attempt / 90 s total, registered at
`memex/aspire/Memex.Portal.ServiceDefaults/ServiceDefaults.cs:94`) is resolved by exactly three
consumers:

| consumer | call | frequency |
|---|---|---|
| `InstanceRegistrationClient.Register` (`src/MeshWeaver.PluginCatalog/InstanceRegistrationClient.cs:154`) | `POST` the registration payload | once, at an installation's first boot |
| `GitHubOidcKeyService` (`…/GitHubOidcKeyService.cs:97`) | GitHub's JWKS | a different host entirely |
| **`RegistryPackageSource`** (`…/RegistryPackageSource.cs:53`) | **`GET /api/plugins?ref=…`** (`ListPackages`, line 84) | see below |

So the repeated caller is the catalog listing. It is driven by every registry contact the
installation makes — a catalog open, an install, the `/Store` count — plus the reconciler's safety
net every `PluginCatalogOptions.ReconcileSafetyNetInterval` (**30 minutes** by default,
`PluginCatalogOptions.cs:147`). Two replicas × 48 polls/day ≈ 96 listings/day, of which ~60 exceed
30 s: consistent, and it explains why the rate is steady rather than bursty.

Bundle transfers are NOT this: they resolve `plugin-registry-bundles` (120 s / 5 min), whose events
would read `plugin-registry-bundles-standard//…`. None do.

## The mechanism, server side

`GET /api/plugins` (`memex/Memex.Portal.Shared/Api/PluginRegistryEndpoints.cs:99` → `List` →
`ListAll`, line 158) does three things **per request**, with no cache between them and the caller:

1. **`s.Source.ListPackages(s.GitRef)` per configured source.** For the fleet registry that source is
   `GitHubPackageSource`, whose `ListPackages`
   (`src/MeshWeaver.PluginCatalog/GitHubPackageSource.cs:66`) mints an installation token and then
   **fetches the plugins repo tree from GitHub**, parsing every `package.json` in it. There are ~70
   modules. That is the dominant cost, and it is paid again on the next request.
2. **`Artifacts(hub)` (line 182)** — a mesh read of the registry's pushed publications, once per
   listing.
3. **`IsGranted(caller, source, package)` per package**, before the merge, so the grant is evaluated
   once per package per request.

The comment that already sits in `ServiceDefaults.cs:55-74` measured the sibling endpoint from inside
two production portals on 2026-08-26 — `GET /api/plugins/bundles/index.json`, an 8.7 KB document,
connects in ~0.03 s and takes **12–19 s to first byte** — and says outright that raising the client
budget is a mitigation: *"a registry this slow to serve 8.7 KB is its own defect and wants caching"*.
This issue is that defect, on the listing endpoint, with the consumer-side rate now measured.

## Impact

**Not an outage, and that is why it has been invisible.** The 90 s total budget covers a retry, so
the listing usually succeeds on a later attempt — the installation's catalog and its reconcile still
work. What it costs:

- Every catalog open and every `/Store` render pays up to 30 s before the first attempt is even
  abandoned. That is a page, not a background job.
- The failure mode when the registry is merely a little slower is the one #1133/#1137 already
  documented three consequences deep: modules stop landing, `MeshWeaver.SelfUpdate.Aks` never lands,
  and self-update is silently off while the instance keeps reporting an available version.
- ~60 red log lines a day that a reader has to recognise and discount.

## What would fix it

Not a wider budget. The listing's inputs change at the cadence of a merge to the plugins repo, and
are identical for every caller of the same `(source, ref)` — the per-caller part is only the grant
FILTER, which runs over an already-materialised list. So the repo snapshot and the parsed manifests
are cacheable per `(repoUrl, gitRef)` with the existing change signal (the registry already receives
the plugins repo's webhook) as the invalidation, leaving `IsGranted` and `Artifacts` per request.

🚨 **The cache must be keyed so it cannot serve one instance's catalog to another.** That is #3768's
shape — the registry answering with its OWN identity rather than the caller's — and it is a
disclosure bug, not a performance one. Caching the SOURCE snapshot (pre-filter) rather than the
RESPONSE is what keeps that impossible by construction: the grant filter still runs per caller.

## Closing condition

`Admin/_LogIncident/9ca334c1e8dad9ca`'s `occurrences` stops advancing over a full week with the
control instance still polling — **or** a measured TTFB for `GET /api/plugins?ref=…` from a consumer,
under the 30 s attempt budget with margin, taken with a real instance key.

Refs #1134 (which is closed on its own defect — the attribution — being fixed and live).

rbuergi:
## The measured cause — and it is not the one the issue names

`GET /api/plugins` builds its catalog from **two small files per plugin** (`<Plugin>/index.json` and
`<Plugin>/manifest.lock`) and obtained them by **transferring the entire source repository**, then
discarding almost all of it.

Measured against `MeshWeaver.Plugins` at `25dc26d7`, 2026-09-16:

| | files | bytes | wall clock |
|---|---|---|---|
| what the transfer moved | 4,995 | 47.8 MB compressed | **13.0 s** |
| what `ListPackages` actually reads | 143 | 0.8 MB | 0.08 s |
| the narrow read — same commit, same answer | 143 | 1.3 MB | **3.3 s** |

**13.0 s lands exactly in the 12–19 s TTFB two production portals measured on 2026-08-26**, which is
what identifies the transfer as the dominant term rather than anything the endpoint computes per
request.

Two attributions that were on record are wrong for this endpoint, and both were checked rather than
assumed:

- **Not the parse.** Reading all 4,995 files off warm disk is **83 ms**. So filtering *while reading
  the worktree* — which is exactly what `GitProtocolRepoClient` did — saves nothing a caller feels.
- **Not entitlement or a mesh query.** `ServiceDefaults`' comment blames *"entitlement per package
  and a mesh query on every request"*. That is true of the sibling **bundle index** (`InstalledPackages`
  and `HeldPartitions` really are per-request mesh queries) and false here: `IsGranted` is
  `caller.Allows(...)`, pure in-memory over a grant the authenticator already resolved, and
  `Artifacts(hub)` is `NoPublicationArtifacts` — `Observable.Return([])` — on every deployment in the
  fleet (no other implementation exists in this repo or in MeshWeaver.Plugins). A measurement taken on
  one endpoint had been carried to another.

### Relationship to #4246

#4246 (merged, `bd7876bf6e`) was right and is not being revisited: it changed **how often** the
repository is read. It could not change **what one read costs**, and the first request after every
eviction or window expiry still pays it in full, on a page-facing request. This is the other half.

🚨 **Neither half is live yet.** The registry these timeouts are measured against is
**memex.meshweaver.cloud**, whose `/api/version` reads `3.0.0+c84c6c05` (2026-09-12) — and `c84c6c05`
is **not** a descendant of #4246's merge. So the incident keeps advancing (2,078 at 18:53Z today)
because the registry is still serving the uncached, whole-repository listing. Rolling it remains the
step that closes the issue; this PR makes the read that then happens fit the budget with margin.

## The fix

`IGitHubRepoClient`'s filtered `Fetch` overload has always *stated* this contract — *"downloads ONLY
the blobs whose path satisfies `pathFilter` … without pulling the rest of the repo"* — and
`OctokitGitHubRepoClient` honoured it. `GitProtocolRepoClient` did not, on the reasoning that the git
protocol costs the same number of REST calls (zero) either way: **true of calls, false of bytes**.
And nothing in the tree called the overload, so the listing never even asked for it.

The git-protocol client now selects **before** it transfers:

1. `git fetch --depth 1 --filter=blob:none` — a blobless partial clone: commits and trees only, so
   the whole path list is known before one file's bytes move (1.5 s / 372 KB for the repo above).
2. `git ls-tree -r --name-only -z FETCH_HEAD` — that list, for free. `-z` because `ls-tree` quotes
   unusual paths otherwise, and a quoted path would be read under a name the worktree does not have.
3. The caller's predicate selects from it; the survivors become `--no-cone` sparse-checkout patterns,
   so the checkout materialises exactly those in **one** batched lazy fetch.

A filter matching nothing skips the checkout entirely and moves no blobs at all. The worktree read
still applies the same predicate, so the **answer is identical however the transfer went**.

Then the two package sources actually ask for it: `NodeRepoPackageSource` and `GitHubPackageSource`
take an optional narrow-fetch delegate that `PackageSources.Build` supplies, and `ListPackages` uses
it with a predicate written against the very loops that parse the result.

**Only the LISTING is narrowed.** `FetchPackageFiles` — the install path — still reads the whole
package folder (narrowing it would install an empty package), and the unfiltered `Fetch` is untouched,
which is what a content sync wants.

### Two hazards handled explicitly

- 🚨 **A remote that cannot serve partial clones does not fail.** `uploadpack.allowFilter` is a server
  capability; with it off, git prints *"filtering not recognized by server, ignoring"* and **exits 0**
  having transferred everything. The answer stays correct, but the saving is gone — and an unannounced
  loss of it is precisely how this latency would return unnoticed, so the client detects that line and
  says so. Only a hard fetch failure takes the fallback branch, and a failure of the fallback propagates.
- 🚨 **An unsubscribe is not a refusal.** `OperationCanceledException` is excluded from the fallback, so
  a cancelled read cannot kick off a second — and this time whole-repository — fetch for an answer
  nobody will read.

A selection that cannot be expressed as sparse patterns (a path carrying gitignore metacharacters, or
more matches than an argv should carry) falls back to a whole checkout — still correct, just not narrow.

## Evidence it holds

- **`GitProtocolNarrowFetchTest` (5 cases, all green)** pins **which path ran**, not only what it
  answered — a narrow fetch that silently regressed to a whole checkout still returns the right files,
  so asserting the files alone would pass over the very defect the test exists to prevent. Cases: the
  narrow path; a remote *without* partial clone answering identically and announcing the lost saving;
  a filter matching nothing; a **subdirectory**, where the predicate sees the sub-relative path while
  the sparse pattern must be repo-root-relative (confusing them would answer an empty catalog); and the
  unfiltered control.
- **Negative control run, not assumed**: reverting the implementation to the old whole-fetch behaviour
  fails **3 of the 5** cases, and the unfiltered control correctly keeps passing. Restored, rebuilt from
  the restored source, green again — so the assembly under test is the new one.
- **`Memex.Portal.Shared.Test` — 22/22**, including #4246's own `PackageListingCacheTest` (with its
  two-caller disclosure control) and `LocalSourceContentVersionTest`, the test that caught the first
  version of #4246 caching a local source.
- **Doc guards 6/6**: `WhatsNewEntryIntegrityTest`, `ArchitectureTopicMapTest`, `DocumentationLinkIntegrityTest`.
- **`dotnet build -c Release -warnaserror`, one project per invocation, `0 Warning(s) 0 Error(s)`**:
  `MeshWeaver.GitSync`, `MeshWeaver.PluginCatalog`, `Memex.Portal.Shared`, `MeshWeaver.Hosting.Test`,
  `MeshWeaver.Documentation.Test`, `Memex.Portal.Shared.Test`.

## Found and NOT fixed

- 🚨 **The incident this issue closes on is shared with a second, unaddressed defect.**
  `Admin/_LogIncident/9ca334c1e8dad9ca`'s fingerprint masks the `Source:` value, so every Polly
  `OnTimeout` folds onto it. Of the **10 retained samples read today, 6 are
  `plugin-registry-bundles-standard`** (2026-09-15T20:50–20:55Z, four pods) — the **bundle transfer**
  pipeline, 120 s attempt budget, which #4222 explicitly scoped out and which neither #4246 nor this PR
  touches. The 09-15 triage noted it at n=1 and said it wants its own issue; it is now n=6 in one
  five-minute window. **So this incident going flat is not a clean signal for this issue**, and a future
  reopen should read `samples[]` before being believed.
- **The bundle index (`/api/plugins/bundles/index.json`) genuinely does the per-request work the
  2026-08-26 comment describes** — `InstalledPackages` and `HeldPartitions` mesh queries, two activation
  reads, and per-package module-byte filesystem probes, none of them cached. That is the endpoint the
  8.7 KB / 12–19 s measurement was actually taken on, and it is untouched here.
- **Registration runs once per POD BOOT, not once per installation.** The issue discounts
  `InstanceRegistrationClient.Register` as "once, at an installation's first boot", but it shares the
  30 s `plugin-registry` pipeline and the control instance has cycled through ~180 pods. The empty
  `Operation Key` means a sample cannot distinguish it from the listing, so some share of the
  `plugin-registry` timeouts may not be the listing at all.

Fixes #4222

Pairs-with: none — the diff removes no public type or member; it only adds (two `init` properties and
private methods). No interface member was added, so nothing is owed to implementers.


rbuergi:
## Review addressed — all seven findings, five of them real defects

Thanks — three of these were defects I could not have argued with, and one of them defeated a guard I
had written myself two commits earlier.

### Fixed as defects

**1. The abbreviated-SHA fallback swallowed cancellation, defeating the guard below it.** Correct, and
the sharpest of the findings: I excluded `OperationCanceledException` from the *inner* partial-clone
fallback and left the *outer* short-SHA `.Catch` catching everything — so for a short-SHA filtered
read, a cancelled subscription still launched `octokit.Fetch`. The outer catch now propagates
cancellation too, which also fixes the pre-existing behaviour.

**2. The full-fetch fallback fired on every non-cancellation error.** Correct. An invalid ref, an auth
failure or a transient transport error would each have triggered a second, whole-repository read —
doubling page-facing work before failing anyway, which is the shape this PR exists to remove. The
fallback is now gated on `IsPartialCloneRefusal(ex)` (git's own wording for an unsupported or
malformed filter); everything else propagates.

**3. Trimming `\n`/`\r` from `-z` records corrupts valid git paths.** Correct, and it would have been
a silent contract violation rather than an error: git permits both characters inside a path, so the
predicate would have been evaluated against a name the worktree does not have, no sparse pattern
would be generated, and the filtered fetch would quietly omit the file. `-z` records are now taken
verbatim and only a genuinely empty (trailing) record is skipped.

**4. The 2,000-match cap does not bound argv.** Correct — a few hundred long paths reach the
platform's limit long before the count does, and `sparse-checkout set` would then fail the whole
fetch instead of taking the documented whole-checkout fallback. Now bounded on **bytes** as well
(`MaxNarrowArgumentBytes`, 128 KiB), and exceeding either bound takes the same fallback.

**5. The wiring was not actually covered — the existing suite could not have failed.** This is the
best catch of the set and exactly the trap I try hardest to avoid. `CountingRepoClient` implements
only the four-argument `Fetch`, so the interface's DEFAULT five-argument member forwarded to it and
`PackageListingCacheIsArmedTest` passed whether or not `PackageSources` wired `NarrowFetch` at all —
the listing could have regressed to whole-repository reads with the suite green. The client now
overrides **both** overloads and counts them separately, and a new test pins both halves:
`TheListingAsksForTheNarrowFetch_AndTheInstallStillReadsTheWholeFolder` asserts the listing takes the
filtered overload **and** that `FetchPackageFiles` still takes the unfiltered one (narrowing the
install would ship an empty package).

### Fixed as raised

**6. Mutable accumulation in the test logger.** Replaced with `ImmutableInterlocked.Update` over an
`ImmutableList` — the client logs from pool threads, so this is the one place the mutable collection
was genuinely shared-while-written. **7. `List<string>` returned through `IReadOnlyList<string>`** in
`WantedPatterns` — now an `ImmutableArray` built and frozen.

### Partly addressed, and where it stops — the sparse-transfer proof

> selection logging does not prove sparse checkout occurred … it would still pass if the code logged
> `3 of 5` and then performed a full checkout

Fair, and I strengthened it as far as the current seam allows: there is now a second line emitted
**only after `sparse-checkout set` has actually succeeded**, so it reports a command that ran rather
than an intention, and the capable-path test asserts it. A path that selected three and then checked
out the whole tree cannot produce that line (the whole-tree branch logs a different one).

What that still is not is a byte-level assertion that `A/big.txt` never moved. Observing the lazy
blob requests needs either an instrumented `GitCli` or retaining the temp clone, and the client
deletes it on termination by design. So I am naming this as a residual gap rather than claiming it is
closed. What covers the regression today is the **measured negative control**: reverting the
implementation to the old whole-fetch behaviour fails **3 of the 5** cases, with the unfiltered
control correctly still passing — I ran that, restored, rebuilt from the restored source and
re-confirmed green, rather than asserting it.

All suites re-run after these changes: `GitProtocolNarrowFetchTest` 5/5, the package/cache suite
green including the new wiring test, and `dotnet build -c Release -warnaserror` clean on every
touched project.

Issue #4471, opened 2026-09-16
The fleet watch stopped writing at 00:19:40Z and nothing noticed for ~5.7h — a frozen Ops/Status reads as a current one


rbuergi:
**The fleet watch on memex.systemorph.com stopped writing at 2026-09-16T00:19:40Z and nothing noticed for ~5.7 hours — because nothing watches the watcher.**

## Measured

`search 'namespace:Ops/Status scope:descendants sort:lastModified-desc'` on memex.systemorph.com (`coverage.partitions: ["ops"]`), 2026-09-16 ~06:0xZ:

| node | lastModified | verdict it is frozen on |
|---|---|---|
| `Ops/Status/pearl` | 2026-09-16 05:54:53Z | Unknown — **hand-run `Sample`** |
| `Ops/Status/memex` | 2026-09-16 05:46:13Z | Unknown — **hand-run `Sample`** |
| `Ops/Status/memex-cloud` | **2026-09-16 00:19:40Z** | **Degraded** |
| `Ops/Status/build` | **2026-09-16 00:19:31Z** | **Degraded** |
| `Ops/Status/partnerre` | **2026-09-15 22:07:04Z** | Unknown |

Only the two nodes someone refreshed by hand carry recent stamps. The automatic watch's last write is **00:19:40Z**; `partnerre` had already been stale for two hours before that.

## 🚨 The property that makes this serious

**A frozen reading is indistinguishable from a fresh one.** `memex-cloud — Degraded` and `build — Degraded` render exactly as they would if the watch had just evaluated them. The detectors judge *what the watcher last read*; none of them asks *when it last read anything*, so a watcher that dies produces a fleet console that looks like a working fleet console. Six hours of "no alerts" meant "no observations", and the two are the same colour.

This is the same failure class as a skipped CI gate reading as a passed one: the absence of a report presenting as a clean report.

## It is NOT #4422 — eliminated by measurement, not by argument

The obvious hypothesis is the kernel idle-disconnect (#4422, fixed by #4423): a recurring in-mesh activity whose hub goes idle between ticks. It does not hold.

```
memex.systemorph.com /api/version → 3.0.0+afde4eabe0740ff658f3dae8a3073c33a2b3ea02
afde4eabe = "Merge pull request #4460 …", built 2026-09-15 21:50:58Z
#4423 merged as 7fc9f9cd8a at 2026-09-15 19:48:47Z

[...]

Issue #4320, opened 2026-09-14
The pre-prod NodeType sweep would have passed a broken mesh today: compilationStatus is one shared field over a PER-REPLICA compile, so the gate cannot see the failure a roll produces


rbuergi:
The one-call sweep AGENTS.md prescribes as **the** pre-prod gate was run against `memex.meshweaver.cloud` at ~13:0xZ today, while **six NodeTypes were failing to compile on that very portal** (#4314, failing since 06:59:28Z; the replica's own `/health` reads `previouslybroken=11`).

```
search 'nodeType:NodeType content.compilationStatus:Error partitions:all' limit:200
→ 8 of N ≥ 200 over M unknown   (companion count truncated:true; envelope carried NO coverage field)
```

**It named none of the six.** A deploy gated on that reading would have shipped.

## Two independent reasons, and the second is not in AGENTS.md

1. **RLS filtering** — documented. The sweep runs as the caller; a type parked at `Error` in a partition the caller holds no grant on is silently not counted.

2. 🚨 **`compilationStatus` is ONE SHARED FIELD over a PER-REPLICA compile.** A NodeType compiles on each replica as it boots. The field on the node records *a* compile, not *every* replica's. So a type that fails on replica A and succeeds on replica B can read `Ok` — and #4314's own evidence shows exactly this shape: one replica's `/health` reports `previouslybroken=11` while the shared field names eight, and not the same eight.

Reason 2 means the sweep is **structurally unable** to gate a per-replica compile failure, which is precisely the failure mode a roll produces. No amount of anchoring, `partitions:all`, or grant-widening fixes it — those address reason 1 only. Core #4290 (merged today) closed the false-zero hole and added `coverage.partitions`; this is a different hole and it is still open.

## What the reading should have been

Per AGENTS.md's own rule, `coverage.partitions` absent ⇒ **"M unknown", and the sweep does not pass on that zero.** It was not a zero, it was 8 — but the same rule applies to the verdict: with M unknown and a truncated companion count, that reading was **not a pass**, and its number was a floor. The instruction says this; it is easy to read "8 findings, none of them mine" as a pass anyway.

## Direction

[...]

PR #4531, opened 2026-09-16
docs(#4471): the compile cascade with nothing edited — a landed-but-not-loaded module, and a chained lambda as one compile unit

rbuergi:
Commits the #4471 root-cause finding into the doc tree, where the conserve-work-products rule says it belongs — it currently lives only in an issue thread.

## What it adds

A new subsection of `NodeTypeCompilation.md` → *"That recompile can FAIL — and nothing upstream can warn you"*, placed directly after **The failure cascade**.

That section documents the cascade entered from a **deleted framework API**. This adds the second entry point, which needs **no edit to anything**: an in-mesh source `using`s a store-delivered module that is *landed but not loaded* on the replica, so the byte-identical source that compiled on the previous pod does not resolve on this one.

Three things it records that no existing page does (checked: `landed but not loaded`, `pending_module_activation` and `restart activates` are already covered across five pages, so this extends rather than duplicates; `chained lambda` / `one compile unit` had zero hits):

1. **A chained `configuration` lambda is one compile unit.** One unresolvable name removes every capability on the chain together — on `Hosting/PlatformBuildInbox` that was the fleet watch, build queue, operational-space provisioning and the inbox drain. Adding a capability to a chain adds its dependencies to every other capability on it.
2. **Why nothing gates it.** The failure is per-replica; the shared `compilationStatus`, `bake-report`'s `previouslybroken` and the `DynamicTypePreWarmer` readiness gate all read the shared record, so they report healthy (table in the section). Only `pending_module_activation` names the cause — and nothing consumes it.
3. **Measured vs. inferred, stated explicitly.** The module state, the `using` and the chain are each read off the live system; that the lambda fails to compile *on that replica* is inferred, because the one instrument that would show it directly cannot by construction (#4320).

## Verification

Doc-only, internal architecture page — not user-facing, so no What's New entry (per the pullrequest skill, step 0.5). No new page (topic map unchanged) and no new links. Ran the guards that cover an Architecture body edit, on this branch: `DocumentationLinkIntegrityTest`, `DocumentationEmbedIntegrityTest`, `NoConflictMarkersGuard`, `NoLiteralNulInSourceGuard`, `ArchitectureTopicMapTest`.

Refs #4471 (does not close it: the restart is operational, and the two durable defects the section names — the chain's shared fate and the unconsumed detector — remain open).

Pairs-with: none — documentation only, no public surface changed.


rbuergi:
**Withdrawing this PR — its mechanism and its remedy are both refuted by #4536, which measured what this only inferred.**

This page asserted that `MeshWeaver.AI` being landed-but-not-loaded made `Hosting/PlatformBuildInbox`'s chained `configuration` lambda fail to compile, and that a restart activates the pending modules. #4536 refutes both from the live system:

- **Not a compile failure from the module.** `Hosting/TriageItem`, carrying the same `MeshWeaver.AI: min:3.0.0.0` entry, compiled *in that process* on the new pods at 01:10Z. And after the restart, both fresh pods **still** listed `MeshWeaver.AI` under `pending_module_activation` while the watcher drained 303 deliveries. So the module's load state was not what stopped the chain.
- **A restart does not activate them.** The restarted pods printed the same eight names at 21:33Z; the loader's own `[ModuleLoad] STALE PACK` line names **re-install**. I took "a restart activates them" from the health entry's text — the health check's promise is itself wrong, which #4536 records as a separate defect.
- **The actual mechanism** was an assembly **load** failure (`Failed to load assembly for Hosting/PlatformBuildInbox`, 00:49:15Z), after which the hub silently bound the mesh **default** configuration and served reads for twenty hours running none of its initialisers.

This page did mark "the lambda fails to compile on this replica" as inferred — but it still presented that inference as the mechanism and put a wrong remedy under it, and merged docs are what the next reader trusts. What it got right — five capabilities sharing one hub's fate, and a detector whose output has no reader — is already recorded in #4536's `AnUnloadableBuildIsNeverASilentDefault` page, so nothing is lost by closing this.

<sub>2026-09-16 — Claude Opus 5.</sub>

Issue #4528, opened 2026-09-16
Bundle transfers: the 120 s budget bounded SIZE, not responsiveness — #4549 fixed the shape; the roll, the byte/rate measurement and the inverted 3 min/5 min budgets are still owed


rbuergi:
## What is failing

The **`plugin-registry-bundles`** resilience pipeline — the one `PluginBundleClient` resolves, with a
**120 s attempt / 5 min total** budget — is logging `OnTimeout`. An attempt on that pipeline exceeding
120 s is a different defect from #4222, which is about the 30 s `plugin-registry` pipeline and the
catalog listing.

## The measurement

Read from `Admin/_LogIncident/9ca334c1e8dad9ca` on memex.systemorph.com, **2026-09-16T20:5xZ**
(version 1692, `occurrences` 2,078, `lastSeen` 2026-09-16T18:53:25Z).

🚨 **That incident's fingerprint masks the `Source:` value**, so every Polly `OnTimeout` on every
pipeline folds onto it. Of its **10 retained samples**:

| samples | pipeline | when |
|---|---|---|
| **6** | `plugin-registry-bundles-standard//Standard-AttemptTimeout` | 2026-09-15T20:50:45Z – 20:55:17Z |
| 4 | `plugin-registry-standard//Standard-AttemptTimeout` (#4222's) | 2026-09-16 06:45Z, 08:25Z, 18:52Z, 18:53Z |

The six are a burst inside **four and a half minutes** across **four pods**
(`memex-portal-deployment-6546c779d4-4ngqv`, `-7fd9668b69-4qtkx`, `-7fd9668b69-4zzzs`,
`-7fd9668b69-jnwlv`) — i.e. several replicas at once, which is the shape of a roll or a reconcile
wave, not of one unlucky request.

#4222's triage on 2026-09-15 saw this at **n=1** and said outright that if bundle transfers keep
exceeding 120 s "that wants its own issue". It is now n=6 in one window. This is that issue.

## Why it matters beyond the latency

**It contaminates #4222's closing condition.** That issue closes on *"`occurrences` stops advancing
over a full week"* on this very incident — but the counter is shared with this defect, so the
incident can keep advancing after #4222 is completely fixed, and has already produced reopens that
read as "#4222 regressed". Anyone reading a future reopen must read `samples[]` before believing it.

## Where to look

`plugin-registry-bundles` is resolved by `PluginBundleClient`
[...]

PR #4549, opened 2026-09-17
The bundle transfer's budget measured its SIZE, and recorded nothing (#4528)

rbuergi:
## The issue's premise was mine, and the evidence does not support it

I filed #4528 saying bundle-pipeline timeouts were *"6 of 10 retained samples … one 4½-minute
window across four pods"*. That was an artefact of the incident node retaining only ten samples.
Measured properly through the control instance's read-only `Logs` action, 2026-09-17:

| | |
|---|---|
| module adopts that failed | **18** — 14 in one pass on 2026-09-15 (19:58:24Z → 20:54:45Z), 4 more on 2026-09-16 (14:20:44Z → 14:33:11Z) |
| spacing | **exactly 180 s** for 10 of them — the reconciler's `PerPackageAdoptBudget` |
| cause, every time | `The operation has timed out.` |
| packages | AI, Anthropic, AppleIntelligence, Chat, Essentials, Import, Mcp, Northwind, Notifications, OgCard, OpenAI, OpenStreetMap, Publish, Radzen — the pass walking the catalog in order |
| attempt timeouts on the 120 s transfer pipeline | **11**, all inside the 65-minute 09-15 episode, 10 of them on ONE pod |
| `Bundle fetch for …` lines beside them | **0** |
| byte counts or durations anywhere in the fleet | **none** |

It is **not a burst and not a roll wave**: it is a sequential reconcile pass in which every package
dies at the same outer bound, and it recurs.

**Coverage was established, not assumed.** `sinceMinutes` is a request, so a control query for every
Polly resilience event over the same window returned 27 events spanning 2026-09-15T19:46:02Z →
2026-09-16T18:53:25Z, the last of which matches the incident's own `lastSeen` exactly. One caveat
recorded rather than hidden: the incident retains five samples on pods `7fd9668b69-*` for which
Loki holds no events at all, so the two instruments disagree and Loki's copy is the narrower one.

## The defect

`PluginBundleClient.DownloadOverHttp` sent with `HttpClient`'s **buffering default**, so the whole
archive was downloaded *inside* `SendAsync` — inside the Polly attempt. A per-attempt budget then
measures **size ÷ throughput** instead of *"is the registry answering?"*, which is the only question
it can answer. A large bundle and an unresponsive registry produced the identical timeout.

`OciRegistryClient`, in the same assembly, has always used `ResponseHeadersRead` and streamed. 🚨 And
that path is **dead code on every deployment**: it runs only when a catalog entry carries an
`artifact`, and the platform default `IPublicationArtifacts` records none — which is also why the
one log line that *did* carry a byte count never fires, and why no production size was recoverable.

## The change — shape and measurement, never the numbers

- **Headers first, then stream.** The attempt now bounds responsiveness; the body streams outside
  it; the operation as a whole stays bounded by the caller's own budget. No hang is introduced.
- **Record what moved.** Bytes, elapsed and rate on completion; on a transfer that does *not*
  complete, how much had arrived — then the fault is rethrown untouched, never swallowed to produce
  a log line. Zero bytes after two minutes accuses the registry; most of a large archive accuses the
  size. They want opposite fixes, and until now nothing could tell them apart.
- **A rate over a near-zero interval reports `0`**, not a fabricated figure.

**No budget was raised and no retry added.**

## Evidence

- `BundleTransferIsStreamedAndMeasuredTest` — 6 cases, green. It pins the completion option, that
  the body is streamed (`CopyToAsync`, not `ReadAsByteArrayAsync`), that both the completed and the
  incomplete line exist, and the throughput arithmetic including the near-zero guard.
- **Negative control run, not assumed**: reverting the streaming shape makes it fail (1 of 6);
  restored, rebuilt from the restored source, green again.
- **The guard is scoped deliberately.** Its first version asserted over the whole file and failed on
  `FetchIndex`, which reads an 8.7 KB JSON document where buffering is correct — a statement about
  the wrong transfer. It now reads only `DownloadOverHttp`'s own body.
- Doc guards green: `DocumentationLinkIntegrityTest`, `ArchitectureTopicMapTest`,
  `WhatsNewEntryIntegrityTest`.
- `dotnet build -c Release -warnaserror`, one project per invocation, `0 Warning(s) 0 Error(s)`:
  `MeshWeaver.PluginCatalog`, `Memex.Portal.Shared`, `Memex.Portal.Shared.Test`,

## Found and deliberately NOT fixed

- 🚨 **The two budgets are inverted.** `RegistryUpdateReconciler.PerPackageAdoptBudget` is **3 min**
  while the transfer pipeline's own `TotalRequestTimeout` is **5 min**, so the retry is structurally
  unable to finish and the HTTP layer's cause is always discarded for a bare `TimeoutException` —
  which is exactly why 11 attempt timeouts produced 0 client-side lines. The finite outer bound is
  right (*a hang is worse than a failure*); two independently authored budgets that contradict each
  other is the problem, and the fix is to derive one from the other rather than raise either. That
  is a separate change with its own risk, not a rider on this one.
- 🚨 **The incident fingerprint masks `Source:`**, so the listing pipeline (#4222) and this one share
  a counter and neither can be settled on *"occurrences stopped advancing"*. That formula lives in
  the log watcher in MeshWeaver.Plugins.
- Unrelated and not conflated: `meshweaver.azurecr.io` refusing GitHub-runner connections (#4526) is
  the image path, not the plugin-registry bundle path.

## Scope of the claim

**This does not prove a bundle fits 120 s, and it does not cure the 18 failures** — no production
bundle size was obtainable, precisely because nothing recorded one. What it changes is that the
budget now bounds responsiveness rather than size, and that the next occurrence will say which of
the two it was. That evidence is what #4528 still needs, so this is deliberately **not** marked as
resolving it.

Refs #4528

Design conserved as `Doc/Architecture/BundleTransferBudget`.


rbuergi:
## Review addressed — all three, and one of them was a hang I was about to introduce

All three findings were valid. Two of them are exactly the kind I most want caught: one defeats the
purpose of the change, the other trades the defect for a worse one.

### 1. The zero-byte case was still silent — the one diagnosis this change exists to enable

Correct, and the sharpest of the three. My clock started **after** `SendAsync` returned headers, so a
registry that never answered would be cancelled at the 120 s attempt *before* `startedAt`, `buffer`
or the catch existed — emitting the same bare Polly timeout as before. The whole claim of this PR is
that *"zero bytes after two minutes accuses the registry"*, and that was precisely the case still
producing no evidence.

The timestamp is now taken **before the request**, and the request itself is inside the try, so a
header-stage failure logs `0 of <declared> byte(s) had arrived` with its elapsed time.
`TheMeasurement_StartsBeforeTheRequest_SoAZeroByteFailureIsStillReported` pins the ordering.

### 2. The streamed body could hang the two callers that have no deadline — fixed with a STALL bound

Correct, and I verified it rather than taking it on faith: neither
`InstanceAutoRegistrationService` (`bundles.Adopt(package.Id)`) nor
`CatalogLayoutAreas.InstallPackage` (`bundles.AdoptModule(...)`) wraps its adopt in a `.Timeout(...)`
— only `RegistryUpdateReconciler` does, via `PerPackageAdoptBudget`. So moving the body outside the
attempt would have left a headers-then-stall registry hanging the manual install and the default
install indefinitely. **A hang is worse than a failure** (Plugins#959), and my PR body's claim that
the operation "stays bounded by the caller's own budget" was true on exactly one of the three paths.
That was wrong and is corrected in the doc page too.

The body is now bounded **where it is read**, and deliberately on **silence rather than total
duration**: a linked `CancellationTokenSource` armed with `TransferStallBudget`, **reset by every
chunk that arrives**. A transfer still making progress is never cut off however large it is — which
is the defect being removed — while one that goes quiet fails with a named `TimeoutException`
(*"the registry sent no data for 120 s … N byte(s) had arrived"*), distinguished from the caller
cancelling by `when (!ct.IsCancellationRequested)`. `TheStreamedBody_IsBoundedBySILENCE_NotByTotalDuration`
asserts both `CancelAfter` calls, because losing the reset would silently restore a total-duration
bound — the very thing this PR removes.

### 3. The incomplete line now names the registry

Fixed as raised. An instance may have several registries configured, and a byte count that cannot be
attributed to an endpoint does not say which one needs fixing.

### One more defect, in my own test rather than the code

Worth recording: `TheMeasurement_StartsBeforeTheRequest` failed on its first run because it searched
for the word `SendAsync`, which also appears in the **comment** above the call — so it compared the
clock against a sentence rather than against the request. It now matches `.SendAsync(request,`. A
source-text guard that matches prose is a guard measuring the wrong thing, and it happened to fail
loudly here only by luck.

All 8 cases green; `dotnet build -c Release -warnaserror` clean on `MeshWeaver.PluginCatalog` and
`Memex.Portal.Shared.Test`; doc guards re-run.

Issue #4161, opened 2026-09-13
A store-installed module copy overrides the image's own copy on the strength of a DLL existing — a 2026-08-27 rc8 pack displaced a same-day source build and every form fell back to ToString() (self-registry, Plugins#1483 residual)


rbuergi:
## The defect

`ModuleActivation.ComputeEffectiveModuleEntries` (Pass 1, `src/MeshWeaver.PluginCatalog/ModuleActivation.cs` ~L1087–1113 on `main`) makes **every enabled persisted store entry whose landed DLL exists** an override of the same-named `Modules:Assemblies` baseline entry. The only gate is the DLL's existence; the platform floor is advisory (#3648, correctly), and nothing compares the store copy's **framework identity** (or age) with the image's own copy. So the store copy wins by construction, even when it was built for a different platform than the one that is booting.

## Measured (MeshWeaver.Plugins#1483, 2026-09-08 and 09-09 comments)

memex-local, self-registry, portal built from source (`3.0.0-ci.0`), after Plugins#1486 landed the view packs in the image:

```
2026-09-07  [ModuleLoad] MeshWeaver.Blazor.Views ← /app/MeshWeaver.Blazor.Views.dll
              (source=appsettings, mvid=8c833e20, written=2026-09-07 08:30:50Z)
2026-09-08  [ModuleLoad] MeshWeaver.Blazor.Views ← /tmp/meshweaver-pinned-modules/…/MeshWeaver.Blazor.Views@96af4c59/…
              (source=store, mvid=a87ef06f, written=2026-08-27 14:00:29Z)
[ModuleActivation] store-installed module 'MeshWeaver.Blazor.Views' declares platform >= 3.0.0-rc8 but this deployment runs 3.0.0-ci.0 — advisory (#3648)
```

A 2026-08-27 store build displaced the image's own 2026-09-07 copy on a platform built that day. The link probe passed (the module's linked types exist), the module loaded, and the view registrations inside no longer matched the control types the platform emits — so the Subscribe panel's **outermost** control rendered as `StackControl { … }` text: a whole-tree `ToString()` fallback, not a missing-view error. Before #1486 the failure was narrower (three packs missing); after it the failure got *broader*, because the packs that now load are the stale ones. A locally built portal and store-served view packs were mutually incompatible with no combination that rendered.

[...]

PR #4553, opened 2026-09-17
fix(modules): a declined landed pack no longer reports "a restart activates them" (#4550)

### LINKED ISSUE #4550: Store-delivered modules the image also ships are declined on every boot — and reported as "landed but not yet loaded, a restart activates them" (two identity schemes, never equal)
**Measured on memex.systemorph.com, 2026-09-16 (image `3.0.0-ci.8710`), and it survives a restart.**

`/health` reports:

> `pending_module_activation: Degraded — 8 module(s) are landed but not yet loaded in this process — a restart activates them: MeshWeaver.AI, MeshWeaver.Blazor.Chat, MeshWeaver.Blazor.EntityViews, MeshWeaver.Blazor.Graph, MeshWeaver.Hosting.Instance, MeshWeaver.Markdown.Collaboration, MeshWeaver.Markdown.Export, MeshWeaver.Mcp`

The deployment was restarted at 21:17Z on the same image (`Ops/Actions/restart-memex-20260916-2120`). The NEW pods reported the same eight at 21:33Z. **"A restart activates them" is false for these modules**, and two operators (one of them this session) acted on it.

### What is actually happening

The pod's own log, from the boot that produced that state:

```
[ModuleActivation] SKIPPED store-installed module 'MeshWeaver.AI': declined: built for another
platform (framework gce971b2cd3c81912748cb3786490ee2035addd29; this deployment runs
s4b2836629f0c58ad4ed0aa683190981a) — the image's own copy runs instead.
```

`gce971b2c…` is a **commit** identity — a core commit (merged 2026-09-15 19:24Z). `s4b28366…` is the **API-surface** identity a portal resolves for itself. `ModuleActivationBoot.ComputeEffectiveModuleEntries` (#4161) compared them with ordinal equality.

**No module bundle in the fleet states an `s<hash>` identity** (every pack lane reads the platform's `MeshWeaver.Compiler.dll`, or states `g$sha` — `node-repo-module-pack.yml`), and **no portal resolves a `g<sha>` one** as `LiveFrameworkMvid`. So for every module the image also ships, the store copy is declined on **every** boot, whatever the registry publishes: the registry lane is structurally dead for those modules, and #4161's "self-healing — publish it and the next reconcile lands it and it wins again" cannot happen.

Three surfaces then describe that state, each with a different and wrong remedy:

1. `/health` / the package card — *restart* (the next boot re-runs the same comparison).
2. `[ModuleLoad] STALE PACK … no usable store-installed entry claims this name … Re-install the module` — there IS an enabled entry with its landed DLL present, and re-installing lands the same bytes.
3. The activation report has no state for "declined", so it folds into `Pending`.

### Denominator / counter-example

memex-cloud (image `3.0.0-ci.8411`) is the control: its 09-13 pod reported `pending_module_activation: Degraded — 20 module(s) …` (`Ops/Status/memex-cloud` v1224, sampled 2026-09-16T00:19:30Z), the rolling restart at 21:24Z replaced the pods, and the three new pods on the SAME image report no `pending_module_activation` entry at all (measured by `/health` at 04:41Z and in `Ops/Status/memex-cloud` v1237). So the check works, prints when Degraded on that image, and a genuinely pending module IS activated by a restart — the memex case is a different state wearing the same words.

### Fix

- Compare like with like: `ModuleFrameworkIdentity` (scheme-aware), and let the platform state BOTH of its readings — the surface identity and `FrameworkBuildIdentity.ProducerStatedIdentity` (what a packer reading this platform's anchor states). A bundle packed by the build that produced this image then MATCHES and is adopted.
- Report the decline as its own state (`ModuleActivationReport.Declined`), subtracted from `Pending`, with the remedy that actually clears it; same for the STALE PACK line and the package card.

### Residual (separate change)

Even compared like-with-like, the commit scheme is per-COMMIT: a store module released after the image, with no rebuild against it, stays declined. Making the pack lanes state the **surface** identity (stable across commits that do not move the API surface) would let such a module be adopted. That is a fleet-wide change to what every bundle states and is tracked separately.

rbuergi:
Fixes #4550

## What was measured

`memex.systemorph.com` (image `3.0.0-ci.8710`), 2026-09-16:

```
/health → pending_module_activation: Degraded — 8 module(s) are landed but not yet loaded in this
process — a restart activates them: MeshWeaver.AI, MeshWeaver.Blazor.Chat,
MeshWeaver.Blazor.EntityViews, MeshWeaver.Blazor.Graph, MeshWeaver.Hosting.Instance,
MeshWeaver.Markdown.Collaboration, MeshWeaver.Markdown.Export, MeshWeaver.Mcp
```

The deployment was restarted at 21:17Z on the same image (`Ops/Actions/restart-memex-20260916-2120`); the new pods reported the same eight at 21:33Z. Two operators restarted a production portal on that sentence.

The pod's own log, from that boot:

```
[ModuleActivation] SKIPPED store-installed module 'MeshWeaver.AI': declined: built for another
platform (framework gce971b2cd3c81912748cb3786490ee2035addd29; this deployment runs
s4b2836629f0c58ad4ed0aa683190981a) — the image's own copy runs instead.
```

`gce971b2c…` is a **commit** identity (a core commit); `s4b28366…` is the **API-surface** identity a portal resolves for itself. #4161 compared them with ordinal equality. **No module bundle in the fleet states an `s<hash>` identity** (every pack lane reads the platform's `MeshWeaver.Compiler.dll`, or states `g$sha` — `node-repo-module-pack.yml`) and **no portal resolves a `g<sha>` one** as `LiveFrameworkMvid` — so the store copy of every module the image also ships was declined on **every** boot, whatever the registry published, and #4161's "self-healing" could never happen.

**Control (memex-cloud, image `3.0.0-ci.8411`):** its 09-13 pod reported `pending_module_activation: Degraded — 20 module(s) …` (`Ops/Status/memex-cloud` v1224); the 21:24Z rolling restart replaced the pods and the three new pods on the SAME image report no such entry (`/health` at 04:41Z, `Ops/Status/memex-cloud` v1237). The check works and prints when Degraded on that image — a genuinely pending module IS activated by a restart. memex's eight were a different state wearing the same words.

## The change

- **`ModuleFrameworkIdentity`** — one scheme-aware comparison (`Matches` / `Differs` / `NotComparable` / `NotStated`), used by the boot discriminator and by the activation report, so the two cannot reach different verdicts about one entry.
- **`FrameworkBuildIdentity.ProducerStatedIdentity`** — what a packer reading *this* platform's anchor would state, i.e. the scheme every bundle is written in. `ModuleActivationBoot.ComputeEffectiveModuleEntriesForPlatform` takes **every** reading a platform states about itself, so a bundle packed by the build that produced this image MATCHES and is adopted. (A distinct NAME rather than an overload: a positional `null` seventh argument in a dependent would otherwise become CS0121, and bare `cref`s would become CS0419.)
- A bundle from **another** build still loses to the image's copy, and so does one with **no comparable reading** — un-declining the latter would reinstate Plugins#1483 (a store pack that linked fine and rendered the Subscribe panel as a whole-tree `ToString()`). What changes there is the wording: it no longer asserts a difference nothing measured, and it names the remedy that actually clears it.
- **`ModuleActivationReport.Declined`** — the state's own row (init-only, binary-compatible), subtracted from `Pending`, with the sentence *"the module RUNS, from the image's own copy … A RESTART DOES NOT CHANGE THIS and neither does re-installing"*. The package card gets the same line (localized, en+de) instead of "restart required", and the `[ModuleLoad] STALE PACK` remediation stops telling the reader to re-install a pack that was declined.
- The report is given the same inputs the boot decided on: `Modules:Assemblies` names through the new `MeshBuilderModuleActivation.BaselineModuleNames`, and the platform's two readings — both supplied by core's own registration, so the honest message lands with this image and needs no host change.

## Tests

`test/MeshWeaver.Compiler.Pipeline.Test/LandedPackDeclinedNeverSaysRestartTest.cs` — real `ModuleLandingService`, real `ModuleActivationBoot`, real `PendingModuleActivations` reading that landing off the volume, no mocks. The control asserts FIRST that the entry is enabled, its landed DLL exists, and `ModuleActivationStatus.NotYetLoaded` genuinely sees it as not-loaded, so "it is not pending" cannot pass having measured nothing. The differential control is the same landed bytes with an empty baseline — a Store-only module still reports pending, because for it a restart is the truth.

Green locally: the new suite + `ModuleIdentityDiscriminatorTest` (13/13), the module-activation suites in `Memex.Portal.Shared.Test` (76/76), `LocalizationTest` (60/60), and the four doc guards (8/8). Every touched project builds `-c Release -warnaserror` with `0 Error(s)`.

## What this does NOT close

Compared like with like, the commit scheme is per-COMMIT: a store module released **after** the image with no rebuild against it stays declined in favour of the image's copy. Making the pack lanes state the **surface** identity (stable across commits that do not move the API surface) is the change that would adopt it — fleet-wide, to what every bundle states, and tracked on #4550's "Residual" section rather than smuggled in here.

**Both halves are in this repo.** The boot caller is `memex/Memex.Portal.Shared/MemexConfiguration.cs` (every portal host composes its module union through it), so the decision and the reporting move together in this PR — no cross-repo pairing, and no window in which the report and the boot disagree about an entry.

Pairs-with: none — this change adds public surface and removes none.

Mirror-sync: `npm run sync:i18n -- --ref <merged core sha>` — one new key (`ui.moduleRunsImageCopy`, en+de); tracked on Systemorph/MeshWeaver#4550.

Doc: [Two Identity Schemes, One Comparison](src/MeshWeaver.Documentation/Data/Architecture/ModuleIdentitySchemes.md) · What's New entry included.


rbuergi:
Copilot's review was produced against the first commit (`3644cc4`), and four of its five findings were already addressed by the commits that followed it — they are the same three gaps, found independently:

1. **"the host still calls the legacy overload … a same-build `g<sha>` pack is declined while the report says Matches"** — fixed in `512ff87`: `MemexConfiguration.ConfigureMemexMesh` now calls `ComputeEffectiveModuleEntriesForPlatform` with `[LiveFrameworkMvid, ProducerStatedIdentity]`. The boot caller is in this repo (`memex/Memex.Portal.Shared`), so there is no window in which the boot and the report disagree.
2. **"`ReportUnloadedActivatedModules` constructs its own reader with empty inputs"** — fixed in `512ff87`: that reader is given the same `ImageShippedModules` + `PlatformIdentities`, and prints the declined line instead of "landed but not loaded … until a restart".
3. **"the report omits the `LandedDllExists` gate, so a lost landed DLL could be Declined instead of Unresolvable"** — fixed in `1c4932a`: the classification applies the boot's order, DLL existence first.
4. (the second copy of finding 1, on the registration.)

5. **"`BaselineModuleNames` returns a mutable `HashSet` behind `IReadOnlySet`"** — valid, and the fix is written (`ToImmutableHashSet`). It is **not in this PR**: the branch is already in the merge queue, where pushes are refused, and dequeuing a green PR for a change with no behavioural effect would cost the queue a full cycle for nothing. It lands as a one-line follow-up immediately after this merges, referenced from #4550.

PR #4587, opened 2026-09-17
ci(gate): the bake artifact comes back as an OPT-IN — Education was reading it

rbuergi:
## The regression

822c292631 (*"ci(gate): stop uploading the bake nothing ever downloaded"*, landed today ~09:27Z) removed `bake-<sha>` and `bake-<sha>-shard-<n>` from `node-repo-gate.yml`, on the finding that

> an exhaustive search over MeshWeaver{,.Plugins,.Education,.Reinsurance,.SocialMedia,.Manufacturing,.Crm} and Memex finds exactly three references to these artifacts: the two uploads and the one download that feeds the second upload.

**The saving is real** — 66.8 GB of live Actions storage, two copies of identical bytes — and for six of the seven callers nothing was lost.

**The seventh was reading it.** `MeshWeaver.Education/.github/workflows/ci.yml`, twice (`e2e-install` at line 826 and `e2e-mesh` at line 1529):

```yaml
      - name: Fetch the prebuilt NodeType assemblies (the lane's bake)
        if: steps.upstream.outputs.refetch != 'true'
        uses: actions/download-artifact@v8
        with:
          name: bake-${{ github.sha }}
          path: ${{ runner.temp }}/prebuilt
```

and `e2e/mesh/bundle-coverage.sh` **fails** when no bundle reaches the mesh, so the download is load-bearing, not an optimisation. All five of that repo's mesh jobs died on

```
##[error]Unable to download artifact(s): Artifact not found for name: bake-28b37ee663f23b60d868f4746521dc0023077341
##[error]HARNESS FAILURE — the 'portal' container does not exist, so the mesh under test never started.
```

on run [35212656116](https://github.com/Systemorph/MeshWeaver.Education/actions/runs/35212656116), taking `Disposable-mesh gate` — a **required** context on that repo's main — with them. It surfaced two hours late only because those jobs had been skipped all morning behind a separate runner defect (Systemorph/MeshWeaver.Education#343).

The search looked for `bake-run-id` and for the uploads; it did not look for `name: bake-${{ github.sha }}` under `download-artifact`, which is where the reader was.

## The fix

The two uploads come back **behind a new `upload-bake` input, default `false`** — so the saving stays for the six callers who were never reading it, and the one who is says so at its own call site. Education's opt-in lands in its own PR alongside this.

The input's description carries the whole story — the 66.8 GB, who reads it, and what happened when it went — so the next person to measure this storage finds the reader before deleting it again.

## Verified

`actionlint .github/workflows/node-repo-gate.yml` reports exactly the same **9** pre-existing findings before and after (all `SC2129`/style on untouched scripts).


rbuergi:
Superseded: Systemorph/MeshWeaver@5527a7ef1e (`fix(gate): the bake HAD a consumer — restore the upload, opt out per caller`) landed the same fix on main while this was open, and landed it better — `upload-bake` defaults to **true**, so a caller this lane has never heard of cannot be broken by a default, and the five repos that provably do not read it pass `false` and keep ~64.9 GB of the 66.8.

Same finding, reached from the other end: MeshWeaver.Education's five disposable-mesh jobs died on `Artifact not found for name: bake-28b37ee…` on run 35212656116, taking its required `Disposable-mesh gate` with them. Nothing further needed here.

PR #4590, opened 2026-09-17
Address the review on #4576: the skip order, a hex ref, the landing activity, and a silent hold

rbuergi:
## Addresses the Copilot review on #4576 — all four findings were real

#4576 (hole 3 of #3845 + the webhook lane) merged before its review was answered. All four comments were correct; this is the follow-up, one fix per finding, each with a test that fails without it.

### 1. The webhook's per-commit skips ran BEFORE the seal gate

`SkipReason` answered *"already at this commit"* about `head_sha`, ahead of `DecideBuild`. A Space sitting on an **unsealed tip** — where a pre-gate tip import had put it — was therefore skipped on a build **of that same commit** and never rolled back to the sealed one, and the per-landing checks after the gate were unreachable.

`SkipReason` is split into the half that is about the CONFIGURATION (null config, `ExportOnly`, branch mismatch — asked before the gate, since no commit can change those answers) and the half that is about ONE COMMIT (`CommitSkipReason` — asked about `BuildImport.Commit`, the commit that actually lands). For an ordinary delivery that is `head_sha` and the behaviour is byte-for-byte what it was; for a redirect it is the sealed commit, so "already holds that tree" and "final verdict on that tree" mean what they say.

### 2. A typed ref was compared to the seal by a seven-character prefix

`SameCommit` matches on a 7-char prefix — right for two machine-produced shas, wrong for a ref a person typed, because **a branch name may legally be hex**. `abcdef1` would have read as the sealed commit: fetched AS a branch with no redirect, and eligible to attribute a marker-less seal. Only a FULL 40-hex sha is a coordinate now (`IsFullCommitSha`); anything shorter redirects onto the sealed commit — the same tree for a shortened sha, the protection kept for a branch.

### 3. The landing activity called the sealed commit "the built commit"

The webhook's redirect ran `UpdateToProvenCommitFromGitHub`, whose activity title and progress line name their commit *"the built commit"* / *"the commit the build proved"* — for a redirect the one thing that is not true, with the explanation only in the server log. New surface `UpdateToSealedCommitFromGitHub` titles it *"the sealed commit"* and carries the gate's own viewer-localized `Notice` onto the activity, before the fetch. A separate method rather than an optional parameter: widening a shipped signature is a binary break for every assembly compiled against it, prebuilt bundles included.

### 4. The unreadable-index hold dropped its direction line

It returned one line and discarded the `newerLine` the caller had just computed — the one hold of four that said nothing about what releases it, in exactly the case (readable release markers, unreadable identity directory) where the roll is the actionable half.

## Verified

- `dotnet build -c Release -warnaserror`, one project per invocation: GitSync, Hosting.Test, Documentation.Test — each `0 Warning(s) 0 Error(s)`.
- New real-mesh fact `ASourceOnAnUnsealedTip_IsRolledBackAndTheActivitySaysSo` — covers findings 1 and 3 on the instrument that matters (the ref reaching `Fetch`, then the activity's own message keys).
- New pure facts: a hex branch name never reads as the sealed commit; a shortened sealed sha lands on the full one and says so; a typed prefix does not attribute a marker-less seal; an unreadable index names the direction.
- Existing suites re-run: Documentation.Test gate slice and the Hosting.Test sync/seal slice.

Mirror-sync: `npm run sync:i18n -- --ref <merged core sha>` in MeshWeaver.Plugins `clients/react/src/i18n/` for the new `activity.gitsync.updateToSealedCommit.title` key — handed over here, tracked on MeshWeaver#3845.

Refs #3845 — the review follow-up for hole 3. Hole 4 is next.


rbuergi:
Answering the one suppressed review note (the English `ArgumentException` in `UpdateToSealedCommitFromGitHub`): keeping it, deliberately, and it is not a localization gap.

That throw is a **caller-contract violation**, not user-facing copy: it fires only when a machine trigger asks for an import at the sealed commit without naming one, which is the refusal `SyncRefContract` calls load-bearing ("'we could not establish a commit' and 'the branch tip' are different answers"). Its two siblings — `UpdateToProvenCommitFromGitHub` and `ReconcileAtProvenCommitFromGitHub` — carry the same sentence in English for the same reason, and this repo's convention is explicit about where such text lands: `ActivityRunner` logs a command's failure message "verbatim and therefore English: it is an upstream exception message … not a sentence this platform composes, so no catalog can carry it (#3281)", with the terminal STATUS beside it as the language-independent signal.

Everything a person is actually shown on this path IS keyed: `activity.gitsync.updateToSealedCommit.title`, `activity.gitsync.seal.fetching`, and the gate's own `landsOnSeal` / `advanceByRoll` / `advanceBySeal` lines, all en + de. If this exception ever reached a viewer it would mean a caller passed no commit — a defect to fix in the caller, not a string to translate.

PR #4644, opened 2026-09-17
ci(gate): the platform refusal names the remedy that works, not the one that cannot

rbuergi:
## The defect

`resolve-gate-platform.sh`'s *"below the OLDEST set kept"* refusal ended with:

> Either way no future refresh will install it: **re-run the job so it re-resolves.**

Measured 2026-09-17 on `MeshWeaver.Plugins#2040`, that instruction is wrong in both directions:

| attempt | how | resolved set | outcome |
|---|---|---|---|
| 1 | original run | 3.0.0-ci.8820 | failed |
| 2 | `rerun-failed-jobs` | **3.0.0-ci.8820** | failed identically |
| 3 | `rerun` (ALL jobs) | **3.0.0-ci.8820** | failed identically |

* `rerun-failed-jobs` **cannot re-resolve at all** — the job that resolved the set *succeeded*, so it is not re-run and its output is replayed verbatim.
* A **full** re-run *does* re-resolve, and returns the **same set** for as long as the repository's `main` has not PASSED on a newer one — which, while `main` is red, is forever.

So the message sent every reader — human and agent — to the one action that cannot work in the case that actually occurs. It did that while MeshWeaver.Plugins was red **repo-wide** (`main` and all seven branches with recent runs) on exactly this refusal. I followed it myself for two attempts before measuring.

## Also stale on the mechanism

The same sentences claimed the refresh *"only ever installs the NEWEST sealed set"* and *"keeps the 3 newest"*. Since Memex#329 it **also installs back** the set each subscribed repository's `main` last passed, and the window moved `3` → `16` on 2026-09-17. A number written into a diagnostic goes stale the first time the cluster moves it — which is how this one came to misinform three sessions in one day. All three refusals now name the **rule** and deliberately no number.

## What each refusal says now

The remedy turns on the one question that decides it — *is this the set your `main` last passed, and is your repository subscribed?*

* **yes** → the next tick (≤10 min) installs it back; wait, then re-run the failed jobs and the *replayed* resolution succeeds unchanged.
* **not subscribed** → no refresh will ever install it; add the repository to `MAIN_PASSED_REPOS`. That, and not a re-run, is the fix.
* and in both cases, plainly: a re-run helps only *after* the volume carries the set again.

## Doc

`FleetCiProcess.md` carried both stale claims (points 2 and 4) and is corrected with the same measurements, plus the **retention form of its own documented deadlock**: while `main` is red its main-passed set stops moving, so rule 2 pins the set its PRs need in place instead of losing it — the longer main stays red, the more firmly the set is held. And committing that rule is not running it: it sat in git from 09-14 and reached the cluster on 09-17, which is why the volume kept the 3 newest for three days after the fix was written.

## Verification

`resolve-gate-platform.sh --self-test` → `OK (1 container case, 2 volume cases, 8 refusals, 3 usage refusals, 7 wait cases)`. The four assertions that pin this refusal's content (`below the OLDEST set kept`, `PURGED`, `OVERTAKEN`, `Memex#329`) all still hold.

Pairs-with: none — no public type or member leaves `src/`; the diff is one CI script and one doc page.
Mirror-sync: none — no i18n catalog key is added or re-worded.


rbuergi:
All three review findings are valid and are addressed in **#4650**, stacked on this branch (this one is already in the merge queue, and dequeuing a shared queue to amend a strictly-better diff costs everyone a queue build).

Briefly, because the third one is a genuine miss:

1. **The self-test did not pin the new contract** — it asserted only the four markers the *old* message already carried, so the diagnostic could regress all the way back to the wrong remedy with every assertion green. Four assertions added, plus a *structural* refusal of any retention number (`keep(s|ing)? (only )?the [0-9]+ newest`) rather than one phrasing. Negative control run: breaking only the message turns the self-test red naming that assertion. (My first attempt at that control was malformed — `sed` without a line number rewrote the assertion *and* the message together, so it passed having proved nothing.)

2. **"Wait for the tick" was the wrong instruction** — an install is ~200 s of extraction *after* the tick that starts it, over an SMB share that caches a negative lookup for up to 30 s, and a rule-2 install runs after the newest set's. Re-running the moment a tick fires can still land before `.complete` is published, and the reader would conclude the fix does not work. The remedy now says to wait until the volume actually lists the digest.

3. **I fixed three of five sites.** The wait-loop's overtaken refusal and the file header's `GONE` description still asserted *"the refresh only ever takes the NEWEST sealed set, so this one will never be installed"* — on exactly the path a shard takes when the volume advances *while it waits*. The three clauses are now hoisted into `$KEPT`/`$REMEDY`/`$RERUN` and shared by all four refusals, which is the durable fix: three independent copies are how the stale claim outlived Memex#329 in the first place.

Issue #4601, opened 2026-09-17
Typed node content is written with no schema validation — a string-declared member accepts and persists a JSON object, and the record stops deserializing


sierragolflima:
## Why this repo

The missing validation is in the core write path, not in a content model. `ClientContent` (`Systemorph/MeshWeaver.Crm`, `Crm/Source/Records.cs:99`) is only the record this was measured on — it declares its members correctly and is not at fault. Any node type's typed content reaches the store through the owning hub's write pipeline in core, and that is what accepts a payload contradicting the declared schema. Filing against `Systemorph/MeshWeaver.Crm` would put a core write-pipeline change in a module's queue.

Prior art checked, none covers this: #3623 (issue, closed — a failed *read* defaulting over content on write; this is its mirror image, a write that stores what no read can parse), #1388 (issue, closed — Article create path dropping fields, the opposite behaviour), #4597 (issue, open — a cast failure that is correctly detected and logged), #3645 (issue, closed). The `create` tool's own contract warns that content written without a `$type` "is stored intact and then reads as if the node were empty" — the same silence, one level up.

## Symptom

The mesh write API accepts typed node content that violates the declared schema of its content type, versions it, and stores it **verbatim**. A member declared `string?` will hold a JSON object. No validation error, no warning. The record then no longer deserializes as its declared type, and every reader that does `ContentAs<T>` gets `null` — which is the documented trigger for the data-loss class in #3623.

## Measured — 2026-09-17, memex.systemorph.com

A scratch `Crm/Client` node (`ZzClientContentProbe`, created and deleted for this test) through the mesh MCP write tools.

**1. A well-formed payload persists every field.** `create` with all six members correctly typed, read back at v1:

```json
{"$type":"ClientContent","website":"https://example.com/","industry":"Crm/Industry/MedicalTechnology",
 "country":"Crm/Country/CH","segment":"Crm/Segment/Enterprise","owner":"rsalzmann","body":"…"}
```

All six present. **2. `patch` with well-formed values also persists correctly** — `{"content":{"country":"Crm/Country/DE","segment":"…"}}` at v2 changed `country` and preserved the rest.

**3. A schema-invalid payload is stored verbatim.** `patch` with:

```json
{"content":{"country":{"$type":"CountryReference","id":"CH"},
            "industry":{"$type":"IndustryReference","id":"MedicalTechnology"}}}
```

returned `Patched: ZzClientContentProbe (v2 → v3)` and read back at v3 as:

```json
[...]

PR #4658, opened 2026-09-17
fix(content): a whole-document bind failure is CAUGHT, not escaped (#4601 follow-up)

rbuergi:
**A guard that promised to admit a shape was failing every write of it — because "we do not judge this" was written as a `catch` that does not catch.**

`ContentSchemaValidator` (#4601, merged this evening in `4e13e8d0e9`) tests proposed content by deserialising it:

```csharp
catch (JsonException ex) when (MemberOf(ex.Path) is { } member)                       // refuse: a member contradicts its type
catch (Exception ex) when (ex is NotSupportedException or InvalidOperationException)  // no contract here: say nothing
```

A **missing `required` member reports `Path` = `$`**, which names no member. That exception matched **no clause at all**, left the validator, and failed the write with System.Text.Json's own message — the guard refusing, untranslated and naming no node, the exact shape its class doc calls *"the ordinary partial-content shape a legitimate writer produces"*.

## Measured, on the set carrying the guard (`3.0.0-ci.8856`)

| where | what it looked like |
|---|---|
| MeshWeaver.Plugins in-mesh licensing tests, creating `Admin/Keys/…` | `Unexpected error: JSON deserialization for type 'MeshWeaver.Markdown.MarkdownContent' was missing required properties including: 'content'.` → `GATE FAILED — tests: Store/Licensing` |
| the AI plugin's `update` tool | same message, `MeshPluginTest` ×2 → `Module bundle (MeshWeaver.AI)` red ([Plugins#2049](https://github.com/Systemorph/MeshWeaver.Plugins/pull/2049)) |

Both are shapes this guard exempts **twice over**: once as partial content, and once for the `[JsonExtensionData]` buffer `MarkdownContent` carries, which the unmapped-member arm skips outright. Neither was ever meant to be refused.

## The fix

A whole-document failure is now **caught** and carried into the member census — the one judgement that still applies to content that did not bind. That census answers `Valid` for a type with an extension-data buffer and for any payload naming at least one declared member, so the only thing it can still refuse is content **none** of whose members the type knows: the #4601 shape. For a type with a `required` member that case now reaches the guard's own localized refusal, where previously it could only ever escape.

Nothing the guard refused before is admitted now.

## Verification — a negative control, not a green run

With the validator change reverted and the test project **rebuilt** (`0 Error(s)`, and the rebuilt assembly checked for the new log literal so the run could not be answered by a stale binary), exactly the two new cases fail:

```
MeshWeaver.Graph.Test  Total: 6, Failed: 2
  ContentSchemaValidationTest.Create_OmittingARequiredMember_WhileNamingADeclaredOne_StillLands
  ContentSchemaValidationTest.Create_WithNoDeclaredMember_OnATypeWithARequiredOne_IsRefusedByTheGuard_NotThrownFromIt
```

With the fix restored and rebuilt: **6/6**.

The control asserts the **absence** of System.Text.Json's wording in the refusal message — that is precisely what separates a judgement from an escape, and a test that only checked "something failed" would pass on the bug.

## Why this is urgent

`MeshWeaver.Plugins` main is dark, and this is one of its roots. The satellite halves (#2039, Plugins#2047, Plugins#2049) do not reach it: the in-mesh licensing failure is in content the platform image writes, so it clears only when a CD wave seals a set carrying this commit.


rbuergi:
Closing as a duplicate: [`5d70d480c3`](https://github.com/Systemorph/MeshWeaver/commit/5d70d480c3) (#4648) landed the same fix on main while this was in the queue — an unconditional `catch (JsonException)` with the judgement inside it, falling through to the member census rather than returning Valid, which is the same reasoning and the same behaviour as this PR.

#4648 is strictly more thorough: four new cases to my two (it adds the Update side and a buffered-required type), the same `NotContain("required properties")` control that separates a judgement from an escape, plus the architecture page and a What's New entry. Nothing here is worth carrying over.

Recorded for the record, since this PR's description has the fleet-side measurement #4648 does not: the escape reached further than core's own CD. It also failed MeshWeaver.Plugins' in-mesh licensing gate — `GATE FAILED — tests: Store/Licensing`, on `Admin/Keys/…` creates — which is one of the roots MeshWeaver.Plugins main is dark on tonight. That clears when a CD wave seals a set carrying `5d70d480c3`.

rbuergi:
Overlap note — not a request to close, and I am not the owner of this PR.

This and **#4657** ("fix(content): a case the guard declines to judge must still be CAUGHT (#4648)", merged 2026-09-17T20:29:31Z) are the same defect with the same root cause, opened ~14 minutes apart: #4657 at ~19:53Z, this at 20:07:35Z. Both close the escape in `src/MeshWeaver.Graph/Security/ContentSchemaValidator.cs` where #4624 wrote its "not judged" exemption as an exception FILTER — `catch (JsonException ex) when (MemberOf(ex.Path) is { } member)` — so a whole-document failure (`JsonException.Path == "$"`, which is what a missing `required` member raises) matched no catch at all and escaped the validator.

They differ only in shape: this one adds a second `catch (JsonException ex)` clause after the filtered one; #4657 makes the single catch unconditional and branches inside, falling THROUGH to the declared-member rule on `$` rather than returning `Valid` (an early `Valid` would exempt every type with a `required` member from the rule #4601 exists for, because such a type throws before `UnmappedMemberHandling.Skip` can apply).

Since #4657 has merged, this PR now conflicts on `ContentSchemaValidator.cs`. Whether to close it or rebase it onto what landed is its owner's call — flagging it so the decision is made deliberately rather than by a merge conflict discovered later. Worth keeping from this PR either way: any test case here that #4657 does not already cover.

PR #4721, opened 2026-09-18
fix(#3542): the node-content editor binds by JSON key, so an admin's update policy reaches the record

### LINKED ISSUE #3542: SelfUpdate orders candidates by semver — the 3.1.0-ci.7841 slip tag outranks every sealed 3.0.0 wave, and it rolled memex-cloud to it on a policy record that lost its own policy
## What is measured (memex.meshweaver.cloud, 2026-09-07 06:19–06:27Z)

The portal's self-updater rolled the instance on its own to **`3.1.0-ci.7841`**:

```
[SelfUpdate] check (PolicyChange): applied update 3.1.0-ci.7841 (was 3.0.0-rc9.ci.7693; last rolled 2026-09-03T17:06:49Z). UNVERIFIED — no combo verification has been recorded for '3.1.0-ci.7841' on this instance
[SelfUpdate] rolled 3.1.0-ci.7841 WITHOUT combo clearance …
Admin/UpdatePolicy: latestAvailableTag=3.1.0-ci.7841, lastCheckVerdict="no newer release: 500 tag(s) listed, none newer than the installed 3.1.0-ci.7841."
```

`3.1.0-ci.7832 … 3.1.0-ci.7841` are the tags from the version-bump slip on main (build 7841 = core `34cdff3c4`, main of 2026-09-05 21:17Z, CD run 33992997386 — sealed and promoted, but three days behind). ACR also holds `3.0.0-ci.7955 / 7962 / 7964 / 7977`, every one of them newer AND sealed. The updater orders candidates by **semver**, so the slip tags outrank every sealed 3.0.0 wave, and its own record now says "installed 3.1.0-ci.7841, none newer" — the instance will never move to a 3.0.0 wave by itself, and any pipeline roll to one (`helm upgrade` from the Memex overlay) will be undone at the next check because a "newer" 3.1.0 exists. `newest` in `main-cd.yml` is computed by ACR time order; the updater's `VersionSelect.IsNewer` is not.

Consequences seen on the instance: 7841 lacks `e89fa2c67` (payment contract, 09-06 — `Store/Plugin` stays in compile Error), `3eefcf3cf` (#3391 — the Store git feed 401s again two hours after each pod start) and `b8accc2b2` (#3475). The roll itself started from a `PolicyChange` trigger at 06:19:55Z whose write nobody has claimed; the `Admin/UpdatePolicy` record had read `policy: None` all night and now carries no policy field at all, after ~1300 refused cross-hub writes of its own `lastCheckedAt` today (`[MergeGuard] Admin/UpdatePolicy: refused stale/reordered cross-hub write`). The rollout stalled at 7/8 with the third new pod unscheduled while the first two rebuild 350 NodeTypes.

## Proposal

1. **Order candidates by sealed publication, never by version string.** A candidate is eligible only if its tag was promoted by a sealed CD run (the `Promote: tag the full set` + `Register the publication` pair), and "newer" is CD order (run number / promotion time), not semver. A tag whose semver is higher than the newest sealed one is exactly the slip case and must lose.
2. **"Installed" is what the pod RUNS** (`/api/version`, the image tag of the running Deployment), never the record's `latestAvailableTag` — after a manual roll-back (09-03) the record kept claiming a version no pod ran.
3. **The policy record must not lose its policy under its own bookkeeping writes**: `lastCheckedAt` belongs on a separate node or a per-pod key, and an absent `policy` must read as `None`, not as "enabled" (the 08-30 WhatsNew already fixed the absent-node case; the absent-field case is what fired today).
4. Retag or delete the `3.1.0-ci.783x/7841` tags in ACR (fleet decision; the images are sealed but mislabelled).

Related: Memex#165 (the roll decision), Memex#212 (pins record what runs), MeshWeaver#3538 (a module newer than its platform).

rbuergi:
## The defect

`MeshNodeEditorField.FromType` derived each editor field's JSON `Key` from the CLR **property
name** (`p.Name.ToCamelCase()`), ignoring `[JsonPropertyName]`. `MeshNodeContentEditorView`
(MeshWeaver.Plugins) uses that key **verbatim, on both sides of the binding**, as a key into the
node content's JSON object:

```csharp
var value = ToJsonObject(node.Content)?[f.Key];   // LoadValues
obj[key] = value;                                 // Persist
```

So a property renamed behind the attribute binds the control to **a field that does not exist**.

`Admin/UpdatePolicy` is exactly that case. #3607 — itself a fix for this issue — renamed
`UpdatePolicyContent.Policy` to `DeclaredPolicy` + `[JsonPropertyName("policy")]` so that an
*absent* declaration could fail closed to `None` instead of reading as "roll to anything". From
that commit the Updates settings tab's strategy dropdown wrote `declaredPolicy`:

- **the read** misses, so an install that HAS a policy renders the dropdown unset;
- **the write** lands under a key the record's deserializer ignores, so the value goes nowhere.

🚨 **And it is worse than "written but unread" — the value never reaches storage at all.** The owning
per-node hub materialises the content as `UpdatePolicyContent` and re-serialises it, so the unknown
`declaredPolicy` is dropped on that round trip. Measured: the control's wait for `Stable` to appear
on the record under *any* key spends its whole convergence budget and times out.

What the operator sees is a **transient** success. `MeshNodeContentEditorView` keeps the chosen value
in its own `_text[f.Key]` field state, so the dropdown shows it until the next stream emission runs
`LoadValues`, which finds nothing under that key and silently reverts the control to unset. Set it,
navigate away, come back — the field is empty again, with nothing to say why.

Nothing throws. Nothing is logged. The write genuinely succeeds.

### Why it latches

An install whose `policy` field was lost fails closed to `None`. Under `None` the poller records
*"updates are disabled on this install"* and **evaluates nothing at all** (#3795) — no check, no
hold, no verdict. The Updates tab is the ONE surface built to turn updates back on, and it had
become a no-op that looks like a success. Nothing else on the install ever disagrees with it,
because under `None` nothing else runs.

Measured on memex.meshweaver.cloud, 2026-09-18: `Admin/UpdatePolicy` carries no `policy` field, its
`lastCheckVerdict` reads *"updates are disabled on this install (Admin/UpdatePolicy = None); the
registry was not listed"*, and its `heldAt` is frozen at **2026-09-11T08:57Z** — the last evaluation
that ran while a policy was still on the record. Every `lastCheckedAt` since is a check that decided
nothing.

🚨 **No live instance's `Admin/UpdatePolicy` is touched by this PR.** This fixes the code that
discards what an operator sets; restoring the value is an operations decision.

## What changed (three root causes, no band-aids)

1. **The key is the JSON name.** `[JsonPropertyName]` wins; the camelCase property name stays the
   fallback **unchanged on purpose** — it is the key every stored record was already written under,
   and re-deriving it differently would re-key live content.
2. **`[JsonIgnore]` is not editable.** Such a property is one the record does not persist, so a
   control over it can only ever write a key the type drops on read. The computed
   `UpdatePolicyContent.Policy` was reflected as a **second, unlabelled** enum dropdown beside the
   broken one, with nothing to tell them apart. Only the default `Always` form is skipped —
   `[JsonIgnore(Condition = Never)]` means the opposite and stays editable, which
   `NotificationSettings` and `GitHubSyncConfig` rely on.
3. **`UpdatePolicyContent.RequireCiGreen` gets `[JsonIgnore(Condition = Never)]`.** It defaults to
   `true` through a property initializer while the hub serializer omits `WhenWritingDefault` — and
   `false` IS the CLR default for a bool. Unticking *"Only update to CI-verified (green) builds"*
   wrote a value the serializer dropped, and the next read re-applied the initializer's `true`. A
   checkbox that could be unticked and never stayed unticked.

## Refuse the write, or write a genuine partial patch? — decided per writer

Both answers appear here, and which one is right depends on what the writer knows:

- **The bookkeeping writers refuse.** They own a typed record and must replace a field on it, so
  "I could not read what is there" has no safe continuation: #3619 gave all five of them
  `Update<UpdatePolicyContent>`, whose `null` means ABSENT and only absent and which faults with a
  `MeshNodeStreamException` — **the write does not happen**. Verified unchanged and green here.
- **The editor writes a genuine partial patch, and that was already correct.** `Persist` sets ONE
  key on the record's own current JSON and hands it back, so the merge patch the framework computes
  names only that key and cannot clobber a field the operator never touched — which is exactly why
  the fix is a *key* correction and not a change of write shape. The patch was always minimal; it
  was aimed at a field that did not exist.

Its remaining refusal case — content present but unreadable, where the patch degenerates into a
whole-record replacement — is the one instance this PR does not close, below.

## The control

`AdminChosenPolicyReachesTheRecordTest` (`test/Memex.Portal.Shared.Test`) drives
`MeshNodeContentEditorView.Persist`'s **exact** per-field write against a real mesh — the same
`obj[key] = value` read-modify-write through `GetMeshNodeStream(path).Update(...)`, with the key the
backend declares. `MeshNodeEditorFieldTest` pins the contract itself, including the denominator:
**every declared `Key` must appear in the JSON the serializer actually produces for that type.**

Measured by reverting only the two source files (`git apply -R`), rebuilding, and proving the loaded
`MeshWeaver.Graph.dll` carried the pre-fix code before running:

| | pre-fix | post-fix |
|---|---|---|
| `AdminChosenPolicyReachesTheRecordTest` (5) | **4 failed**, 1 passed | 5 passed |
| `MeshNodeEditorFieldTest` (3) | **2 failed**, 1 passed | 3 passed |

The pre-fix failures, in the test's own words:

```
strategy field key = 'declaredPolicy', label = 'Update strategy'
  → the wait for `Stable` on the record times out: it is never stored, under any key.

strategy dropdowns: declaredPolicy ('Update strategy'), policy ('Policy')
Expected collection to have 1 item(s) ... but found 2.

dropdown would show: '<unset>'
Expected value to be "None" ... but found <null>.
```

🚨 **The second commit here is CI catching a real ordering race in my own test, not a flake.**
The first version waited on `c.DeclaredPolicy is not null` — which the *seeded* record
(`policy: "None"`) already satisfies, so the wait returned the pre-write snapshot and the assertion
ran against a state the write never produced. It passed locally and failed on CI (shard 5, run
13126) with the fix demonstrably working in the same output (`strategy field key = 'policy'`). It
now waits on the write's own footprint — the value written, read off the raw stored bytes. That is
the exact trap `UnreadablePolicyRecordIsNotClobberedTest` documents on its own positive controls,
and correcting it is what produced the sharper finding above.

**What it would have caught:** #3607's rename. The property moved behind a `[JsonPropertyName]` and
every test still passed, because nothing compared what the editor DECLARES against what the
serializer DOES. That comparison is now an assertion.

🚨 **The positive control is the load-bearing half.** `AFieldWithNeitherTrapLands` writes `pattern`
— a field whose CLR name already equals its JSON name and whose value is not a CLR default — through
the identical path, and it **passed before the fix as well as after**. Without it, "the edit did not
stick" would be equally satisfied by a harness that never wrote anything.

## The sweep — every writer, counted

**7 write surfaces touch `Admin/UpdatePolicy`.** Measured on this tree, not remembered:

| # | writer | shape | state |
|---|---|---|---|
| 1 | `UpdatePolicyNodeType.EnsureExists` | `CreateNode`, create-on-absent | leaves an existing record untouched |
| 2 | `UpdatePolicyNodeType.RecordVerification` | `Update<UpdatePolicyContent>` | already typed (#3619) |
| 3 | `SelfUpdateHostedService.RecordHandover` | `Update<UpdatePolicyContent>` | already typed |
| 4 | `SelfUpdateHostedService.RecordHold` | `Update<UpdatePolicyContent>` | already typed |
| 5 | `SelfUpdateHostedService.RecordCheck` | `Update<UpdatePolicyContent>` | already typed |
| 6 | `SelfUpdateHostedService.RecordAvailable` | `Update<UpdatePolicyContent>` | already typed |
| 7 | **the admin editor** (`UpdatePolicySettingsTab` → `ForType` → `Persist`) | per-field JSON patch | **broken — fixed here** |

`grep 'GetMeshNodeStream(UpdatePolicyNodeType.NodePath)' -A2 | grep '\.Update('` returns **zero**
untyped writes on that node. The five typed ones are `Update<UpdatePolicyContent>` at
`UpdatePolicyNodeType.cs:416` and `SelfUpdateHostedService.cs:{1315,1361,1551,1578}`. Writer 7 is
the one the #3619/#3631 sweep missed, because it is not a `.Update` lambda in this repo at all — it
is a field list this repo declares and another repo writes through.

## The `(cur ?? new X())` class — quantified, not hand-waved

A brace-matched scan of every `.Update(...)` call containing a lambda across `src/`, `memex/`,
`tools/` and `samples/` (comments and string literals masked, so the rule's own prose does not
self-match):

- **119** untyped `Update(node => …)` lambda call sites — the denominator;
- **12** typed `Update<T>(…)` call sites;
- **1** hit for "reads content AND defaults it": `src/MeshWeaver.Data/DataExtensions.cs:2600`,
  `store ?? new EntityStore()` — the stream's own state, **not** `MeshNode.Content`, where `null`
  means absent and only absent. Not this class.

So the `(cur ?? new X()) with { … }` shape on node content is at **zero** in this repo, and stays
there: `DefaultedContentReadInsideAnUpdateLambdaGuard` is the ratchet.

🚨 **One instance of the class does survive, in another repo, and this PR does not close it.**
`MeshNodeContentEditorView.Persist` opens `var obj = ToJsonObject(node.Content) ?? new JsonObject()`,
and `ToJsonObject` returns `null` for two different facts: "there is no content" (correct —
create-on-absent) and "I could not read the content", which its `catch` swallows. In the second case
the per-field write becomes a **whole-record replacement by `{ "<key>": value }`**. It is narrower
than the key defect (it needs content present and not serialisable to a JSON object) and it is
invisible to the guard, whose regex looks for `ContentAs<T>(…) ?? new` and cannot see a defaulting
read behind a helper method. Diagnosis and cure are written up in the doc page; filed separately
rather than folded in, so this change stays one repo.

## What the brief's stated control already covers

*"An `UpdatePolicy` whose content does not materialize as the CLR type, written by a bookkeeping
update, must not end up with `policy` lost"* — that control **exists and passes on `main` today**:
`UnreadablePolicyRecordIsNotClobberedTest`, landed by #3619 / #3631 on 2026-09-07, four bookkeeping
writes × (unreadable record refused, readable record written) plus positive controls. It is green in
this PR's run. That half of #3542 is settled; the half this PR closes is the operator-facing one it
left open.

🚨 **#3542 was already closed as completed on 2026-09-09**, on those earlier halves. It was closed
correctly for what it had been read as, and the fault it describes was still live on the public
instance nine days later — because the half nobody had looked at was not a *writer* of the record at
all, but the control an operator uses to set it.

## `.ContentAs<T>()` vs. "deserialize where the type IS registered" — which case this is

The ABSOLUTE rule says to prefer `.ContentAs<T>(hub.JsonSerializerOptions)` over a cast, **but
first** to ask whether a payload is untyped because the read is happening on a hub that never
registered the type — in which case `.ContentAs<T>` would paper over a routing mistake.

**This is the second case, and the architecture already answers it correctly — which is why the fix
is not a `ContentAs<T>` call.** The Blazor client hub does not register `UpdatePolicyContent` and
should not have to: `MeshNodeContentEditorView` is deliberately type-free and works on the JSON DOM,
so that one control can edit any node's content with **no client-side type registry**. The type
knowledge lives where the type IS registered — on the backend, in `MeshNodeEditorField.FromType`,
which reflects the content record and ships the resulting field list on the control.

So the work was already in the right place. What was wrong is what that place SHIPPED: a list of CLR
names where the consumer needs wire names. Adding a `ContentAs<T>` on the GUI side would have
required registering every editable content type in the client hub — moving the work to the wrong
place to fix a defect that was one attribute lookup deep in the right one.

## Conserved

- [`Doc/Architecture/EditorFieldKeys`](src/MeshWeaver.Documentation/Data/Architecture/EditorFieldKeys.md)
  — the binding contract, both failure modes with their measurements, the denominator over the five
  editor-bound content types, the half that lives in MeshWeaver.Plugins, and a checklist for the
  next editor-bound type. Linked from the Architecture index.
- What's New entry, `Category: Fix` — says plainly that a strategy set before this release did not
  take effect and should be set again.

## Verification

- `dotnet build -c Release -warnaserror`, one project per invocation: `MeshWeaver.Graph`,
  `Memex.Portal.Shared`, `MeshWeaver.Graph.Test`, `Memex.Portal.Shared.Test`,
  `MeshWeaver.Documentation`, `MeshWeaver.Documentation.Test` — **0 Warning(s), 0 Error(s)** each.
- `MeshWeaver.Graph.Test`, whole project: **1791 passed, 0 failed**.
- `Memex.Portal.Shared.Test`, the whole self-update + settings surface (15 classes):
  **268 passed, 0 failed**.
- `Memex.Portal.Shared.Test`, this control + the three existing `UpdatePolicy` regression classes:
  **18 passed, 0 failed**.
- `DocumentationLinkIntegrityTest`, `DocumentationFrontMatterGuard`,
  `DefaultedContentReadInsideAnUpdateLambdaGuard`, `WhatsNewEntryIntegrityTest`: green — and the
  embedded `MeshWeaver.Documentation.dll` under the test bin was confirmed to carry the new page
  before the run, so the link check really validated it.

## Recycle / deploy note

`Admin/UpdatePolicy`'s per-node hub serves content, not the field list: the editor's `Fields` are
computed per render on the settings tab's layout area, so a rolled image renders the corrected keys
with no recycle. What a roll does **not** do is re-set a policy that was lost — an operator has to
pick the strategy again on the Updates tab, which now works.

Closes #3542


rbuergi:
The surviving instance of the same class, in the other repo, is now tracked as Systemorph/MeshWeaver.Plugins#2082 — `MeshNodeContentEditorView.Persist`'s `ToJsonObject(node.Content) ?? new JsonObject()`, where a failed read becomes a whole-record replacement. The doc page's "The half that is not in this repo" section carries the diagnosis and the cure.

Noted there as well: `DefaultedContentReadInsideAnUpdateLambdaGuard` structurally cannot see it — its regex matches `ContentAs<T>(…) ?? new` and the defaulting read here is behind a helper method, and its `ScannedRoots` are core's `src`/`memex` only. The guard is green and the instance is live; that is a gap in the ratchet, not only in the file.

rbuergi:
**Blocked on a Copilot quota refusal, not on findings — and the remedy is a maintainer action the gate explicitly bars agents from taking.**

`Automatic review answered` is RED. Its log:

```
the automatic review has not landed — the reviewer posted, but not a review:
review 5247394785 at 2026-09-18T11:45:24Z is a refusal:
"Copilot was unable to review this pull request because the user who requested
 the review has reached their quota limit."
threads opened by the automatic reviewer: 0, answered by a person: 0
```

So there is no thread to answer — `pulls/4721/comments` is empty and the single review is `COMMENTED` with no comments. Replying on a thread cannot clear this, and neither can re-running the check: it reads live state, and the live state is that no review exists.

Per the gate's own instruction, going green needs **a maintainer** to either re-request the Copilot review or apply the `review-waived` label — *"never an agent, and never automatically"*. I am not doing either, and I am not merging past a red required context (and not reaching for `--admin`).

Everything else is green or in flight; I will report the tests verdict here when `Consolidate test results` lands. Local verification on this exact commit: 1791 + 268 + 18 tests passing, 336 source guards, 50 ratchets, and every touched project clean under `-c Release -warnaserror`.

rbuergi:
Pushed `c3228e8a` — **CI caught a real ordering race in my own test, and correcting it sharpened the finding.**

The first version of the control waited on `c.DeclaredPolicy is not null`. The *seeded* record already carries an explicit `policy: "None"`, so that predicate is true of the **pre-write snapshot**: the wait returned the state the write had not touched yet and the assertion ran against something it never observed. It passed locally and failed on CI (shard 5, run 13126) — with the fix visibly working in the same output (`strategy field key = 'policy'`). That is precisely the trap `UnreadablePolicyRecordIsNotClobberedTest` documents on its own positive controls, and I walked into it.

It now waits on the write's own footprint: the VALUE written, read off the raw stored bytes.

**What that changed in the diagnosis.** With the corrected predicate the pre-fix run *times out* — `Stable` never appears on the record under **any** key. The owning per-node hub materialises the content as `UpdatePolicyContent` and re-serialises it, so the unknown `declaredPolicy` is dropped on that round trip and never reaches storage.

So the edit is not "written but unread" — it is **never stored**. Three places in this change set said the junk key "echoes back permanently" and the UI therefore reads as applied; that was an inference, it is wrong, and the doc page, the XML docs and the What's New entry now say what was measured: the control keeps the value in its own field state, so the tab shows it until the next emission and then silently reverts to unset.

Negative control re-run with the corrected predicate, reverting only the two source files and rebuilding (assembly 1347584 reverted vs 1348096 fixed): **4 of 5 failed, and the one that passed is the positive control.** Post-fix: 13/13 green with the three existing `UpdatePolicy` regression classes, doc guards 4/4.

rbuergi:
**CI on `c3228e8a`: 32 success, 4 skipped, 1 failure — and the one failure is the Copilot quota gate, nothing else.**

`Consolidate test results` = **success**. The test race my previous push fixed is confirmed closed on CI, not just locally.

`Automatic review answered` = failure, same cause as before and unchanged by the new head:

```
check-review-answered: #4721 (open) head c3228e8ad5 — RED
threads opened by the automatic reviewer: 0, answered by a person: 0
the automatic review has not landed — review 5247394785 is a refusal:
"Copilot was unable to review this pull request because the user who requested
 the review has reached their quota limit."
```

This is fleet-wide, not specific to this change — measured by REST just now: **#4719 and #4720 are red on the same gate; older #4713 is green.** The gate's own self-test in this very run asserts `quota refusal is not a review → expected=red`, so the red is by design.

Its two sanctioned remedies are both maintainer-only — re-request the Copilot review, or apply the `review-waived` label — *"never an agent, and never automatically"*. I have done neither, and I am not merging past a red required context or reaching for `--admin`.

**This PR is ready from my side and needs a maintainer to clear the review gate.**

PR #4725, opened 2026-09-18
feat(i18n): the harness-disconnect status lines, in both languages

rbuergi:
The core half of Systemorph/MeshWeaver.Plugins#2074. It lands **first**, deliberately — see Ordering.

## What it is for

`ThreadChatView.ClearHarnessCredential` (the `/logout` path, in MeshWeaver.Plugins) carries four hard-coded English literals. One of them is not a label but a **claim**:

```csharp
MeshQuery.DeleteNode(providerPath).Subscribe(
    _ => ShowSkillStatus($"{harness.Definition.DisplayName} disconnected — its stored credential was removed.", false),
    …
```

The `onNext` arm discards the value. Since #4668 that value distinguishes *removed it* from *there was nothing there*, so a second Disconnect — or one after another device revoked the credential — reports a removal that did not happen. Same family as the API-tokens screen fixed in #4703.

This PR adds only the text the Plugins-side fix binds to. The behaviour change is in that repo.

| key | en | de |
|---|---|---|
| `threads.harnessDisconnected` | `{0} disconnected — its stored credential was removed.` | `{0} getrennt — die gespeicherten Zugangsdaten wurden entfernt.` |
| `threads.harnessDisconnectedNothingStored` | `{0} disconnected — there was no stored credential to remove.` | `{0} getrennt — es waren keine Zugangsdaten gespeichert.` |
| `threads.harnessDisconnectFailed` | `Couldn't disconnect: {0}` | `Trennen nicht möglich: {0}` |
| `threads.harnessDisconnectNotSignedIn` | `Nothing to disconnect — you don't appear to be signed in.` | `Nichts zu trennen — Sie scheinen nicht angemeldet zu sein.` |

"Disconnected" is the honest half and holds either way, so it stays in both variants; only the removal clause is conditional. The other three were plain i18n bugs — a German viewer saw them in English — and are fixed in passing because they are literals in the same method.

## Ordering — why core is first and not second

An unknown key does not fall back to English, it renders **as the raw key**:

> `Lookup` — *"Visible-but-harmless: the raw key surfaces in the UI so a gap is obvious in review."*

So the Plugins half merging first would put `threads.harnessDisconnected` on screen, in both languages, until a set carrying these keys sealed. Core first removes that window entirely. This is the same inverted ordering `Implementers:` and `Mirror-sync:` use, and for the same reason.

## Verification

`LocalizationTest` — **`total: 60, failed: 0`** (it fails when a language is missing any English key, which is the guard that matters for a two-catalog change).

Both catalogs re-read after the edit: valid JSON as **plain UTF-8, no BOM**, **1630 keys each**, all four keys present in both. Checked explicitly because a BOM introduced by an editing script broke the i18n gate on #4703 earlier today.

## Cross-repo

Pairs-with: none — nothing is removed; four keys are added.

Mirror-sync: MeshWeaver.Plugins runs `npm run sync:i18n -- --ref <this merge sha>` for the four keys this adds. The mirror's guard compares against a PINNED core commit rather than core's `main`, so nothing reds anywhere until that pin moves — which is exactly why the handover is declared here rather than left to be noticed.


rbuergi:
**Drive-by diagnosis of the shard-5 red — not my PR, no action taken. Ignore if you have it.**

`Run tests (shard 5)` fails on one test, and the signature says it is the harness budget rather than an assertion:

```
MeshWeaver.Layout.Test.RefusalNamesWhichEndTheStreamMetTest.AStreamTheOwnerReaped_IsNamedAsSuch [FAIL]
  Test execution timed out after 30000 milliseconds
```

**30,000 ms exactly is xUnit's flat `methodTimeout`** (`test/xunit.runner.json`), not a `.Timeout(...)` inside the test. So nothing asserted — the method was cut off.

🚨 **The known shape here, worth checking first:** a **negative** assertion always spends its *whole* window, and `TestTimeouts` are **CI-scaled (~1.7×)** while `methodTimeout` is a flat 30 s. So a `NotEmit(TestTimeouts.Quick)`-style wait that is comfortable locally can exceed the method budget on CI and die **anonymously** — exactly like this, with no failing assertion to read. If this test waits to prove something does *not* happen, that is the first thing to look at.

If it genuinely needs to wait that long, the durable fix is to make the question **structural** rather than time-bounded (assert the state that proves the reap, rather than waiting out a window). Raising `methodTimeout` is not the fix — a run that hits a cap is stuck, not slow.

**Context that may narrow it:**
- **`main` is green** — run #13140 (`c36c185d7`), `MeshWeaver Build and Test` success, **0 failed jobs** — so this is not a live red on main that your branch merely surfaced.
- Everything else on your head is the fleet-wide `Automatic review answered` / `copilot-pull-request-reviewer` failure from the Copilot **quota** refusal (*"the user who requested the review has reached their quota limit"*), which is affecting #4719, #4720, #4721, #4723 and #4728 as well. Its remedy is maintainer-only — re-request the review or apply `review-waived` — so nothing on your side clears it.

🚨 And per this repo's rules: please don't re-run it to see whether it was a flake. If it reproduces, it is a real timing bug; if it does not, the re-run has told you nothing and cost the evidence.

rbuergi:
This PR's red is **not caused by its diff**, and I've filed the cause separately so it isn't rediscovered.

`Consolidate test results` fails on shard 5 with one test:

```
MeshWeaver.Layout.Test.RefusalNamesWhichEndTheStreamMetTest.AStreamTheOwnerReaped_IsNamedAsSuch
[FAIL] Test execution timed out after 30000 milliseconds
```

Thirty seconds, **no output beyond `TEST START`**. This PR adds four keys to `strings.{en,de}.json` plus a What's New file — it adds no key that test reads, changes no existing value, and cannot reach a Layout stream-refusal path. `main` was green on the same test at 13:42Z (run #13140).

So the failure is **unattributed**: a pre-existing intermittent race surfacing here. Filed as **#4741**, with the evidence that pins it as state- or timing-dependent rather than inherent — the sibling `AStreamIdReusedAfterAReleaseIsJudgedOnItsSecondLife`, which calls the same helper **twice** and awaits `DisposalCompleted` **twice**, passed in 0.41 s five milliseconds earlier in the same run.

🚨 **Please do not re-run this to clear it.** A re-run would either hide a real race or spend another 30 minutes reproducing it, and neither answers the question. It is also not in `.github/known-flakes.json`, so the merge-queue steward will not treat it as a catalogued flake.

Two prerequisites before this PR can merge on its own merits:

1. **#4741** — the race itself.
2. **#4740** — every shared `TestTimeouts` budget exceeds the flat 30 s `methodTimeout` (`Convergence` is 36 s locally, **108 s on CI**), so the test's own `because` clause is structurally unable to print. Until that is fixed, each occurrence reports an anonymous timeout instead of naming which await did not complete — which is why #4741 carries a hypothesis rather than a diagnosis.

Separately, the `Automatic review answered` red on this PR is the gate defect fixed by **#4738**: Copilot reviewed this PR at 16:32:28Z with `### 🟢 Approval recommended` and **zero findings**, but the old predicate recognised a review only by the literal string `Pull request overview`, which this body does not contain. Verified against live data: under the fixed predicate this PR reads **GREEN** on the review gate. That half clears automatically once #4738 lands — the gate checks out `refs/pull/N/merge`, so no rebase is needed here.

